PR #1496, opened 2021-02-22
feat(python): add possibility to enter PIN via letters

prusnak:
When I enter the PIN on computer without the numeric keypad (most of the laptops), the operation is quite tedious.

Let's allow also PIN entry via (lowercase) letters.

I chose the following layout for couple of reasons:

```
e r t
d f g
c v b
```

* this layout avoids using Y/Z keys which are commonly swapped on localized keyboard layouts (qwerty vs qwertz)
* this layout avoids using AS/QW keys which are commonly swapped on localized keyboard layouts (qwerty vs azerty)


matejcik:
please fix style and add a changelog entry, otherwise LGTM

prusnak:
Force pushed a94732249764b6d4d78f5957388f00ae56849ea2

matejcik (approved):

Issue #1491, opened 2021-02-18
Allow decreasing the output value in RBF transactions


prusnak:
Currently we don't allow user to decrease output values during the RBF process.

We could allow this providing the following is met:
* all output values are either staying the same or they are decreased, i.e. no output value increase allowed
* for each changed output we'll show a confirmation dialog with output (address), delta amount and final amount

andrewkozlik:
Here are some screenshots of the current WIP. In this example we are taking 0.0001 from an external output and 0.0009428 from a change output to bump the fee from 0.0000378 to 0.0010806.

![image](https://user-images.githubusercontent.com/42678794/109039302-ead27880-76cc-11eb-92e1-3a68c6c40d07.png) ![image](https://user-images.githubusercontent.com/42678794/109039314-edcd6900-76cc-11eb-94b7-cc62d7b52afd.png) ![image](https://user-images.githubusercontent.com/42678794/109039320-f02fc300-76cc-11eb-98f8-a0bbb62b10f4.png) ![image](https://user-images.githubusercontent.com/42678794/109039326-f32ab380-76cc-11eb-804b-2a0f63a3e9d5.png)

It's hard to fit the output modification on a single screen, because the address can take up more than two lines. Here is an earlier attempt, where the new amount can easily overflow the screen:

![image](https://user-images.githubusercontent.com/42678794/109040420-29b4fe00-76ce-11eb-896b-02cbaef5d361.png)

**The same flow on TT:**

![image](https://user-images.githubusercontent.com/42678794/109039911-a09dc700-76cd-11eb-9623-74fcdcbba70c.png) ![image](https://user-images.githubusercontent.com/42678794/109039922-a3002100-76cd-11eb-8be3-fc091c1fd179.png) ![image](https://user-images.githubusercontent.com/42678794/109040070-d04ccf00-76cd-11eb-8d84-11fe493055ba.png) ![image](https://user-images.githubusercontent.com/42678794/109039940-a85d6b80-76cd-11eb-9e31-9208a25a0949.png)

prusnak:
I like these screens 👍

andrewkozlik:
[...]

PR #1500, opened 2021-02-25
Allow decreasing the output value in RBF transactions

andrewkozlik:
Implements https://github.com/trezor/trezor-firmware/issues/1491.


core/src/trezor/ui/layouts/tt.py
```diff
@@ -389,13 +390,43 @@ def confirm_replacement(
...
+    page1 = Text("Modify amount", ui.ICON_SEND, ui.GREEN)
+    page1.normal("Address:")
+    page1.br_half()
+    page1.mono(*_split_address(address))
+
+    page2 = Text("Modify amount", ui.ICON_SEND, ui.GREEN)
+    if sign < 0:
+        page2.normal("Decrease amount by:")
+    else:
+        page2.normal("Increase amount by:")
```
andrewkozlik, line 409: Note that this branch is unreachable, so as an alternative  we could raise an exception here or rename the function to `confirm_decrease_output()`.

legacy/firmware/layout2.c
```diff
@@ -502,6 +502,37 @@ void layoutConfirmReplacement(const char *description, uint8_t txid[32]) {
...
+                               int page) {
+  if (page == 0) {
+    render_address_dialog(coin, out->address, _("Modify amount for"),
+                          _("address:"), NULL);
+  } else {
+    char *question = NULL;
+    uint64_t amount_change = 0;
+    if (orig_out->amount < out->amount) {
+      question = _("Increase amount by:");
+      amount_change = out->amount - orig_out->amount;
```
andrewkozlik, line 516: Same as in the Python implementation, this branch is currently unreachable, so as an alternative we could rename the function to `layoutConfirmDecreaseOutput()`.

andrewkozlik:
As usual, I would like two pairs of eyes to review each of the implementations for T1 and TT. I expect @matejcik and @onvej-sl to do the review, since they are already familiar with the topic, but if @prusnak would like to review one or both implementations instead of @matejcik or @onvej-sl , then feel free to do so. Since T1 doesn't support external inputs, the implementation there is a bit simpler. For more information on the TT implementation see https://github.com/trezor/trezor-firmware/issues/1491#issuecomment-785895490.

onvej-sl (approved):
I reviewed both core and legacy.

core/src/trezor/ui/layouts/tt.py
matejcik: i'd keep this as-is, this is rather far away from the code that knows about the unreachability

matejcik (commented):
minor comment request, otherwise LGTM

core/src/apps/bitcoin/sign_tx/approvers.py
```diff
@@ -44,6 +44,9 @@ def __init__(self, tx: SignTx, coin: CoinInfo) -> None:
 
         self.amount_unit = tx.amount_unit
 
+    def is_payjoin(self) -> bool:
+        return self.external_in != self.orig_external_in
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/1500/commits/21bfdc993dd1e4f7dcc332303d604f461bdff11e

matejcik (approved):

PR #1502, opened 2021-02-25
fix(core/cardano): Allow stake pool registrations with zero margin

refi93:
Motivation: Stake pools with zero margin are currently being wrongfully rejected by Trezor, throwing an `"Invalid certificate"` validation error in this case.

Changes: This PR relaxes trezor's validation logic to allow zero stake pool margin and adds a test fixture to cover that path of logic

Testing:
* `pytest tests/device_tests -k "test_cardano_sign_tx"` (also verified it failed on "Invalid certificate" error before the change)
* updated UI fixtures with `make test_emu_ui_record`


matejcik:
merging into a Cardano branch, will merge to master once CI checks go through

matejcik:
in master now

PR #1507, opened 2021-03-02
Make crypto library thread-safe, remove static variables.

catenocrypt:
In our project (https://trustWallet.com) we have found that Trezor crypto lib is not thread-safe, produces wrong results in a multi-threaded environment.  Values get mixed up due to static variables.  We use this lib in mobile environment, and e.g. our Android app uses multiple threads (we have no control over this).

In this PR I changed variables to non static.

We also added some multi-threaded tests that catch such errors (e.g. https://github.com/trustwallet/wallet-core/blob/ar/trezor-new1-1/tests/Binance/TWAnySignerTests.cpp#L71).

If static variables are needed (e.g. due to some HW/platform considerations), a conditionally define could be used, like this:

```C
#if USE_STATIC
#define STATIC static
#else
#define STATIC
#endif
```

Note: the commit to fix this in our downstream version:
https://github.com/trustwallet/wallet-core/pull/1294/commits/7c55c06cea23387a7ed0a8c25832806d10d40f09


prusnak:
This is a NACK. Most of the changes does not make even sense (e.g. CONFIDENTIAL has no effect without static, switching static const to non-static variable, etc.). I suggest you investigate where you Android app crashes and fix that particular places, not just blindly remove static without understanding what it does.

catenocrypt:
OK, understood.  We'll keep this difference in our downstream version.
(In some places I replaced 'static const' mistakenly to 'static' instead of 'const'; I fixed in next commit, but PR is closed, not relevant any more).

KaycoinsInc (approved):

PR #1509, opened 2021-03-02
Add support for Schnorr crypto (BCH variant)

Fabcien:
Support Schnorr signature and verification defined in
https://github.com/bitcoincashorg/bitcoincash.org/blob/master/spec/2019-05-15-schnorr.md.

The implementation is based on https://github.com/Bitcoin-ABC/secp256k1
as well as the test vectors.

This is a first step toward adding support for Schnorr signature and verification to trezorlib.

### DIFF
--- a/crypto/schnorr.c
+++ b/crypto/schnorr.c
@@ -1,0 +1,213 @@
+/**
+ * Copyright (c) 2021 The Bitcoin ABC developers
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES
+ * OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include "schnorr.h"
+#include "hmac_drbg.h"
+#include "memzero.h"
+#include "rfc6979.h"
+
+#include <assert.h>
+#include <stdio.h>
+#include <string.h>
+
+static int jacobi(const bignum256 *_n, const bignum256 *_k) {
+  assert(!bn_is_zero(_k) && bn_is_odd(_k));
+
+  bignum256 n_copy = {0};
+  bignum256 *n = &n_copy;
+  bn_copy(_n, n);
+
+  bignum256 k_copy = {0};
+  bignum256 *k = &k_copy;
+  bn_copy(_k, k);
+
+  int t = 0;
+  while (!bn_is_zero(n)) {
+    while (bn_is_even(n)) {
+      // jacobi(2 * n, k) = jacobi(n, k) if k = 1 (mod 8) or k = 7 (mod 8)
+      // jacobi(2 * n, k) = -jacobi(n, k) if k = 3 (mod 8) or k = 5 (mod 8)
+      uint32_t r = k->val[0] & 0x07;
+      t ^= (r == 3 || r == 5);
+      bn_rshift(n);
+    }
+
+    if (bn_is_less(n, k)) {
+      // jacobi(n, k) = jacobi(k, n) if k = n = 1 (mod 4)
+      // jacobi(n, k) = -jacobi(k, n) if k = n = 3 (mod 4)
+      t ^= ((n->val[0] & k->val[0] & 3) == 3);
+      bignum256 *temp = n;
+      n = k;
+      k = temp;
+    }
+
+    // jacobi(n, k) = jacobi(n - k, k)
+    bn_subtract(n, k, n);
+  }
+
+  int k_is_one = bn_is_one(k);
+
+  // Cleanup
+  memzero(&n_copy, sizeof(n_copy));
+  memzero(&k_copy, sizeof(k_copy));
+
+  // Map t: [0] => 1, [1] => -1
+  t = -2 * t + 1;
+
+  return k_is_one * t;
+}
+
+static int is_non_quad_residue(const bignum256 *n, const bignum256 *prime) {
+  return jacobi(n, prime) == -1;
+}
+
+static int generate_k_schnorr(const ecdsa_curve *curve, const uint8_t *priv_key,
+                              const uint8_t *hash, bignum256 *k) {
+  rfc6979_state rng = {0};
+  uint8_t hmac_data[SHA256_DIGEST_LENGTH + 16] = {0};
+
+  /*
+   * Init the HMAC with additional data specific to Schnorr. This prevents from
+   * leaking the private key in the case the same message is signed with both
+   * Schnorr and ECDSA.
+   */
+  memcpy(hmac_data, hash, SHA256_DIGEST_LENGTH);
+  memcpy(hmac_data + SHA256_DIGEST_LENGTH, "Schnorr+SHA256  ", 16);
+  hmac_drbg_init(&rng, priv_key, 32, hmac_data, SHA256_DIGEST_LENGTH + 16);
+
+  for (int i = 0; i < 10000; i++) {
+    generate_k_rfc6979(k, &rng);
+    // If k is too big or too small, we don't like it
+    if (bn_is_zero(k) || !bn_is_less(k, &curve->order)) {
+      continue;
+    }
+
+    memzero(&rng, sizeof(rng));
+    return 0;
+  }
+
+  memzero(&rng, sizeof(rng));
+  return 1;
+}
+
+// e = H(Rx, pub_key, msg_hash)
+static void calc_e(const ecdsa_curve *curve, const bignum256 *Rx,
+                   const uint8_t pub_key[33], const uint8_t *msg_hash,
+                   bignum256 *e) {
+  uint8_t Rxbuf[32] = {0};
+  SHA256_CTX ctx = {0};
+  uint8_t digest[SHA256_DIGEST_LENGTH] = {0};
+
+  bn_write_be(Rx, Rxbuf);
+
+  sha256_Init(&ctx);
+  sha256_Update(&ctx, Rxbuf, sizeof(Rxbuf));
+  sha256_Update(&ctx, pub_key, 33);
+  sha256_Update(&ctx, msg_hash, SHA256_DIGEST_LENGTH);
+  sha256_Final(&ctx, digest);
+
+  bn_read_be(digest, e);
+  bn_fast_mod(e, &curve->order);
+  bn_mod(e, &curve->order);
+}
+
+int schnorr_sign_digest(const ecdsa_curve *curve, const uint8_t *priv_key,
+                        const uint8_t *digest, uint8_t *sign) {
+  uint8_t pub_key[33] = {0};
+  curve_point R = {0};
+  bignum256 e = {0}, s = {0}, k = {0};
+
+  ecdsa_get_public_key33(curve, priv_key, pub_key);
+
+  // Compute k
+  if (generate_k_schnorr(curve, priv_key, digest, &k) != 0) {
+    memzero(&k, sizeof(k));
+    return 1;
+  }
+
+  // Compute R = k * G
+  scalar_multiply(curve, &k, &R);
+
+  // If R.y is not a quadratic residue, negate the nonce
+  bn_cnegate(is_non_quad_residue(&R.y, &curve->prime), &k, &curve->order);
+
+  bn_write_be(&R.x, sign);
+
+  // Compute e = H(Rx, pub_key, msg_hash)
+  calc_e(curve, &R.x, pub_key, digest, &e);
+
+  // Compute s = k + e * priv_key
+  bn_read_be(priv_key, &s);
+  bn_multiply(&e, &s, &curve->order);
+  bn_addmod(&s, &k, &curve->order);
+  memzero(&k, sizeof(k));
+  bn_mod(&s, &curve->order);
+  bn_write_be(&s, sign + 32);
+
+  return 0;
+}
+
+int schnorr_verify_digest(const ecdsa_curve *curve, const uint8_t *pub_key,
+                          const uint8_t *digest, const uint8_t *sign) {
+  curve_point P = {0}, sG = {0}, R = {0};
+  bignum256 r = {0}, s = {0}, e = {0};
+
+  bn_read_be(sign, &r);
+  bn_read_be(sign + 32, &s);
+
+  // Signature is invalid if s >= n or r >= p.
+  if (!bn_is_less(&r, &curve->prime) || !bn_is_less(&s, &curve->order)) {
+    return 1;
+  }
+
+  if (!ecdsa_read_pubkey(curve, pub_key, &P)) {
+    return 2;
+  }
+
+  // Compute e
+  calc_e(curve, &r, pub_key, digest, &e);
+
+  if (bn_is_zero(&e)) {
+    return 3;
+  }
+
+  // Compute R = sG - eP
+  bn_subtract(&curve->order, &e, &e);
+  scalar_multiply(curve, &s, &sG);
+  point_multiply(curve, &e, &P, &R);
+  point_add(curve, &sG, &R);
+
+  if (point_is_infinity(&R)) {
+    return 4;
+  }
+
+  // Check r == Rx
+  if (!bn_is_equal(&r, &R.x)) {
+    return 5;
+  }
+
+  // Check Ry is a quadratic residue
+  if (is_non_quad_residue(&R.y, &curve->prime)) {
+    return 6;
+  }
+
+  return 0;
+}
--- a/crypto/schnorr.h
+++ b/crypto/schnorr.h
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) 2021 The Bitcoin ABC developers
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES
+ * OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#ifndef __SCHNORR_H__
+#define __SCHNORR_H__
+
+#include "ecdsa.h"
+
+/* A Schnorr signature is always 64 bytes */
+#define SCHNORR_SIG_LENGTH 64
+
+// sign/verify returns 0 if operation succeeded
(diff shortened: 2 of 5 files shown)


onvej-sl:
Are you able to tell the difference between these signatures and Schnorr signatures defined in https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki, @Fabcien ?

Fabcien:
The goal is the same, but it differs in some choices being made:
 - The computation uses the 33 bytes compressed public key format (vs 32 bytes X only for BIP340)
 - The R.y is chosen as the one being a quadratic residue (vs the one being even for BIP340)

prusnak:
> The goal is the same, but it differs in some choices being made:

Would it be possible to rewrite your code in such a way it supports both BCH Schnorr and BIP-340 Schnorr? For example by adding `schnorr_type` enum with values `SCHNORR_BCH` and `SCHNORR_BIP340`

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+int schnorr_verify_digest(const ecdsa_curve *curve, const uint8_t *pub_key,
+                          const uint8_t *digest, const uint8_t *sign) {
+  curve_point P, sG, R;
+  bignum256 r, s, e;
+
+  bn_read_be(sign, &r);
+  bn_read_be(sign + 32, &s);
+
+  // Signature is invalid if s >= n or r >= p.
+  if (bn_is_zero(&r) || bn_is_zero(&s) || !bn_is_less(&r, &curve->prime) ||
```
onvej-sl, line 220: Is (r != 0 and s != 0) actually needed? It's not written in the specification.

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+#include "memzero.h"
+#include "rfc6979.h"
+
+#include <assert.h>
+#include <stdio.h>
+#include <string.h>
+
+int jacobi(const bignum256 *_n, const bignum256 *_k) {
+  int t = 0;
+  int k_is_one;
```
onvej-sl, line 34: Please initialise _all_ local variables. It's our convention.

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+                   const uint8_t pub_key[33], const uint8_t *msg_hash,
+                   bignum256 *e) {
+  uint8_t Rxbuf[32];
+  SHA256_CTX ctx;
+  uint8_t digest[SHA256_DIGEST_LENGTH];
+
+  bn_write_be(Rx, Rxbuf);
+
+  sha256_Init(&ctx);
+  sha256_Update(&ctx, Rxbuf, 32);
```
onvej-sl, line 134: `sizeof(Rxbuf)` is better than 32

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+  while (!bn_is_zero(&n)) {
+    while (bn_is_even(&n)) {
+      bn_rshift(&n);
+      r = k.val[0] & 0x07;
+      t ^= (r == 3 || r == 5);
+    }
+
+    bn_cmov(&temp, 0, &n, &n);
+    bn_cmov(&n, 0, &k, &k);
+    bn_cmov(&k, 0, &temp, &temp);
```
onvej-sl, line 54: Why don't you use `bn_copy` instead of `bn_cmov`?

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+  assert(!bn_is_zero(_k) && bn_is_odd(_k));
+
+  bn_copy(_n, &n);
+  bn_copy(_k, &k);
+
+  bn_mod(&n, &k);
+
+  while (!bn_is_zero(&n)) {
+    while (bn_is_even(&n)) {
+      bn_rshift(&n);
```
onvej-sl, line 47: You should comment a bit what is happening here. Something like:
jacobi(2 * n, k) = jacobi(n, k) if k == 1 (mod 8) or k == 7 (mod 8)
jacobi(2 * n, k) = -jacobi(n, k) if k == 3 (mod 8) or k == 5 (mod 8)

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+      r = k.val[0] & 0x07;
+      t ^= (r == 3 || r == 5);
+    }
+
+    bn_cmov(&temp, 0, &n, &n);
+    bn_cmov(&n, 0, &k, &k);
+    bn_cmov(&k, 0, &temp, &temp);
+
+    t ^= ((n.val[0] & k.val[0] & 3) == 3);
```
onvej-sl, line 56:
```
jacobi(n, k) = jacobi(k, n) if k = n = 1 (mod 8)
jacobi(n, k) = -jacobi(k, n) if k = n = 3 (mod 8)
```

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+/*
+ * Init the HMAC with additional data specific to Schnorr. This prevents from
+ * leaking the private key in the case the same message is signed with both
+ * Schnorr and ECDSA.
+ */
+void init_rfc6979_schnorr(const uint8_t *priv_key, const uint8_t *hash,
+                          rfc6979_state *state) {
+  uint8_t hmac_data[SHA256_DIGEST_LENGTH + 16];
+
+  memcpy(hmac_data, hash, 32);
```
onvej-sl, line 100: `SHA256_DIGEST_LENGTH` is better than 32.

crypto/schnorr.c
```diff
@@ -0,0 +1,254 @@
...
+  if (generate_k_schnorr(curve, &k, &rng) != 0) {
+    memzero(&k, sizeof(k));
+    return 1;
+  }
+
+  // Compute R = k * G
+  point_multiply(curve, &k, &curve->G, &R);
+
+  // If R.y is not a quadratic residue, negate the nonce
+  bn_normalize(&k);
```
onvej-sl, line 173: I think `k` is already normalized. Maybe you wanted to reduce it?

crypto/schnorr.c
Fabcien: Removed in bc6267c

crypto/schnorr.c
Fabcien: Done in bc6267c

crypto/schnorr.c
Fabcien: Done in bc6267c

crypto/schnorr.c
Fabcien: Done in bc6267c

crypto/schnorr.c
Fabcien: Done in bc6267c

crypto/schnorr.c
Fabcien: Done in bc6267c

crypto/schnorr.c
Fabcien: Done in bc6267c

crypto/schnorr.c
Fabcien: Neither is necessary, k is guaranteed to be < order. Removed in bc6267c

Fabcien:
@prusnak It can probably be done, but there are enough differences that it might be cleaner to just create new functions for BIP340. Code factorization can happen at this time to avoid duplication as needed.

prusnak:
> that it might be cleaner to just create new functions for BIP340

OK

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+    return 2;
+  }
+
+  // Compute e
+  calc_e(curve, &r, pub_key, digest, &e);
+
+  // Compute R = sG - eP
+  bn_cnegate(1, &e, &curve->order);
+  bn_mod(&e, &curve->order);
+  point_multiply(curve, &s, &curve->G, &sG);
```
andrewkozlik, line 233: Use `scalar_multiply()`, it's specially written to be faster.
```suggestion
  scalar_multiply(curve, &s, &sG);
```

andrewkozlik (changes requested):

onvej-sl:
@Fabcien, please, leave conversation resolving to reviewers.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+
+  // Compute k
+  init_rfc6979_schnorr(priv_key, digest, &rng);
+  if (generate_k_schnorr(curve, &k, &rng) != 0) {
+    memzero(&k, sizeof(k));
+    return 1;
+  }
+
+  // Compute R = k * G
+  point_multiply(curve, &k, &curve->G, &R);
```
andrewkozlik, line 174: Use `scalar_multiply()`, see https://github.com/trezor/trezor-firmware/pull/1509/files#r589447096.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+  int k_is_one = 0;
+  uint32_t r = 0;
+  bignum256 n = {0}, k = {0}, temp = {0};
+
+  assert(!bn_is_zero(_k) && bn_is_odd(_k));
+
+  bn_copy(_n, &n);
+  bn_copy(_k, &k);
+
+  bn_mod(&n, &k);
```
andrewkozlik, line 43: Same comment as https://github.com/trezor/trezor-firmware/pull/1509/files#r588659565 applies to this `bn_mod` too, but in fact this `bn_mod` seems redundant, since `n < k` is guaranteed given that `_n` is always a point coordinate and `_k` is the order of the finite field. I'd just add `bn_is_less(_n, _k)` to the `assert` above.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+                   bignum256 *e) {
+  uint8_t Rxbuf[32] = {0};
+  SHA256_CTX ctx = {0};
+  uint8_t digest[SHA256_DIGEST_LENGTH] = {0};
+
+  bn_write_be(Rx, Rxbuf);
+
+  sha256_Init(&ctx);
+  sha256_Update(&ctx, Rxbuf, sizeof(Rxbuf));
+  sha256_Update(&ctx, pub_key, 33);
```
onvej-sl, line 139: `sizeof(pub_key)` is better than `33`.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+
+  if (!ecdsa_read_pubkey(curve, pub_key, &P)) {
+    return 2;
+  }
+
+  // Compute e
+  calc_e(curve, &r, pub_key, digest, &e);
+
+  // Compute R = sG - eP
+  bn_cnegate(1, &e, &curve->order);
```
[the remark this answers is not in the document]
onvej-sl: > bn_subtract(&curve->order, &e, &e); seems more appropriate than bn_cnegate(), but I leave it to @onvej-sl to comment.

I totally agree.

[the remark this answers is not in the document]
onvej-sl: > I am also wondering what to do if e comes out zero, in which case anyone can easily forge a signature. The likelihood of that happening by chance is infinitesimal, but it could be induced by a fault injection. I don't think we will be using Schnorr signatures in any situation where fault injections would be a concern, so I don't think we need to handle e==0, but I am asking for comments just in case anyone has a different opinion.

This is analogous to https://github.com/Fabcien/trezor-firmware/blob/add_schnorr_crypto/crypto/ecdsa.c#L1044_L1051. Is there a reason to handle it differently?

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  // Compute R = k * G
+  point_multiply(curve, &k, &curve->G, &R);
+
+  // If R.y is not a quadratic residue, negate the nonce
+  bn_cnegate(is_non_quad_residue(&R.y, &curve->prime), &k, &curve->order);
+  bn_mod(&k, &curve->order);
+
+  bn_mod(&R.x, &curve->order);
```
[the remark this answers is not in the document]
onvej-sl: You probably wanted `bn_mod(&R.x, &curve->prime)` but this is not necessary because the coordinates of the result of `scalar_multiply` (and `point_multiply`) are already fully reduced. (I agree it is badly documented).

Interesting is that no test failed but this is because `log2((prime - order) / prime)` is about `-128`, so it's very unlikely the coordinate was actually reduced.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+    memzero(&k, sizeof(k));
+    return 1;
+  }
+
+  // Compute R = k * G
+  point_multiply(curve, &k, &curve->G, &R);
+
+  // If R.y is not a quadratic residue, negate the nonce
+  bn_cnegate(is_non_quad_residue(&R.y, &curve->prime), &k, &curve->order);
+  bn_mod(&k, &curve->order);
```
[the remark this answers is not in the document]
onvej-sl: This is true, since `k` goes to `bn_addmod` which cope with even unreduced numbers.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+  calc_e(curve, &R.x, pub_key, digest, &e);
+
+  // Compute s = k + e * priv_key
+  bn_copy(&e, &s);
+  bn_read_be(priv_key, &private_key_scalar);
+  bn_multiply(&private_key_scalar, &s, &curve->order);
+  memzero(&private_key_scalar, sizeof(private_key_scalar));
+  bn_addmod(&s, &k, &curve->order);
+  memzero(&k, sizeof(k));
+  bn_fast_mod(&s, &curve->order);
```
onvej-sl, line 193: Calling `bn_fast_most` is not necessary since an output of `bn_addmod` is partly reduced.

crypto/schnorr.c
onvej-sl: > Same comment as https://github.com/trezor/trezor-firmware/pull/1509/files#r588659565 applies to this bn_mod too, but in fact this bn_mod seems redundant, since n < k is guaranteed given that _n is always a point coordinate and _k is the order of the finite field.

Yes, this is because the coordinates of a point that is an output of `scalar_multiply`, `point_multiply` or `point_add` are fully reduced.

> I'd just add bn_is_less(_n, _k) to the assert above.

I agree.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+
+    bn_copy(&n, &temp);
+    bn_copy(&k, &n);
+    bn_copy(&temp, &k);
+
+    // jacobi(n, k) = jacobi(k, n) if k = n = 1 (mod 4)
+    // jacobi(n, k) = -jacobi(k, n) if k = n = 3 (mod 4)
+    t ^= ((n.val[0] & k.val[0] & 3) == 3);
+
+    bn_mod(&n, &k);
```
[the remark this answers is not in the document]
onvej-sl: I thought I commented it too, but I probably didn't...

The problem is that the function `bn_fast_mod(n, k)` you want to use works only for `2^256 - 2^224 <= k <= 2^256`. So we don't have a function for general reduction.

> so you might as well replace this line with bn_subtract(&n, &k, &n),

You can't, try `jacobi(7, 1)`.

I think the algorithm works correctly. My experimental results show that it needs about three times more iterations in the outer loop for 256-bit numbers. If the quadratic reciprocity is used only if `n < k` (so it either subtracts or uses the quadratic reciprocity) the algorithm needs only two times more iterations. That's not so bad.

crypto/schnorr.c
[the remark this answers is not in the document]
andrewkozlik: We use ECDSA to verify signatures in the bootloader, where fault injections would be a concern, but I am not sure that we will be using Schnorr signatures for anything similar. On the other hand it wouldn't hurt to fail if `e==0`, so let's err on the side of caution and do so.

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+      bn_rshift(&n);
+      // jacobi(2 * n, k) = jacobi(n, k) if k = 1 (mod 8) or k = 7 (mod 8)
+      // jacobi(2 * n, k) = -jacobi(n, k) if k = 3 (mod 8) or k = 5 (mod 8)
+      r = k.val[0] & 0x07;
+      t ^= (r == 3 || r == 5);
+    }
+
+    bn_copy(&n, &temp);
+    bn_copy(&k, &n);
+    bn_copy(&temp, &k);
```
andrewkozlik, line 56: Swapping pointers is more efficient than copying.

crypto/schnorr.c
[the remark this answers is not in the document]
andrewkozlik: I never suggested using `bn_fast_mod()`. My point is that the `bn_mod()` here is highly misleading, because it will result in a different flow than one would expect and even when one realizes this, it's still far from obvious that the return value will be correct.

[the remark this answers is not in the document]
andrewkozlik: Fixed in https://github.com/trezor/trezor-firmware/pull/1509/commits/2bf51f3a2c2b6e39e88f3face3e9a3be67a634ee. @onvej-sl please review and resolve.

crypto/schnorr.c
andrewkozlik: Done in https://github.com/trezor/trezor-firmware/pull/1509/commits/2bf51f3a2c2b6e39e88f3face3e9a3be67a634ee.

crypto/schnorr.c
andrewkozlik: Actually I take back my comment about adding the assert, because the algorithm works correctly even if the assertion doesn't hold, so it would be misleading to have it there. I just removed the `bn_mod()` in https://github.com/trezor/trezor-firmware/pull/1509/commits/2bf51f3a2c2b6e39e88f3face3e9a3be67a634ee.

crypto/schnorr.c
Fabcien: I don't think this can work since `pub_key` is a parameter, this would be `sizeof(uint8_t *)`

crypto/schnorr.h
```diff
@@ -0,0 +1,42 @@
...
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#ifndef __SCHNORR_H__
+#define __SCHNORR_H__
+
+#include "ecdsa.h"
+
+/* A Schnorr signature is always 64 bits */
```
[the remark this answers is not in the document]
Fabcien: Good catch, done in 1c7bd82

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+#include "schnorr.h"
+#include "hmac_drbg.h"
+#include "memzero.h"
+#include "rfc6979.h"
+
+#include <assert.h>
+#include <stdio.h>
+#include <string.h>
+
+int jacobi(const bignum256 *_n, const bignum256 *_k) {
```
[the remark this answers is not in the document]
Fabcien: Done in 1c7bd82

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+  rfc6979_state rng = {0};
+
+  ecdsa_get_public_key33(curve, priv_key, pub_key);
+
+  // Compute k
+  init_rfc6979_schnorr(priv_key, digest, &rng);
+  if (generate_k_schnorr(curve, &k, &rng) != 0) {
+    memzero(&k, sizeof(k));
+    return 1;
+  }
```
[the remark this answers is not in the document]
Fabcien: I inlined `init_rfc6979_schnorr()` as suggested in 1c7bd82

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+  bn_read_be(digest, e);
+  bn_fast_mod(e, &curve->order);
+  bn_mod(e, &curve->order);
+}
+
+int schnorr_sign(const ecdsa_curve *curve, const uint8_t *priv_key,
+                 const uint8_t *msg, const uint32_t msg_len, uint8_t *sign) {
+  uint8_t digest[SHA256_DIGEST_LENGTH] = {0};
+
+  double_sha256(msg, msg_len, digest);
```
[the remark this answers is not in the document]
Fabcien: Removed in 1c7bd82

crypto/schnorr.c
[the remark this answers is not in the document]
Fabcien: Removed in 1c7bd82

crypto/schnorr.c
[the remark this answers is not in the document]
Fabcien: Removed in 1c7bd82

crypto/schnorr.c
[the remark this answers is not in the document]
Fabcien: I added the check in 1c7bd82

crypto/schnorr.c
```diff
@@ -0,0 +1,252 @@
...
+  bn_write_be(&R.x, sign);
+
+  // Compute e = H(Rx, pub_key, msg_hash)
+  calc_e(curve, &R.x, pub_key, digest, &e);
+
+  // Compute s = k + e * priv_key
+  bn_copy(&e, &s);
+  bn_read_be(priv_key, &private_key_scalar);
+  bn_multiply(&private_key_scalar, &s, &curve->order);
+  memzero(&private_key_scalar, sizeof(private_key_scalar));
```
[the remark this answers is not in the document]
Fabcien: Nice ! Done in 1c7bd82

crypto/schnorr.c
Fabcien: Done in 1c7bd82

crypto/schnorr.c
Fabcien: Removed in 1c7bd82

Fabcien:
Thanks for the great review! I hopefully addressed all your comments in 1c7bd82. This is a fairly large change, please let me know if you want me to squash the commits.

crypto/schnorr.c
onvej-sl: Absolutely right.

crypto/schnorr.c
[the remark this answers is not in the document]
onvej-sl: > I never suggested using bn_fast_mod().

And I didn't mean it's you how want to use it. :)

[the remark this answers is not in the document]
onvej-sl: Looks good!

andrewkozlik (approved):
> Thanks for the great review! I hopefully addressed all your comments in 1c7bd82. This is a fairly large change, please let me know if you want me to squash the commits.

Good work @Fabcien. Thank you for your contribution. I already squashed the commits in https://github.com/trezor/trezor-firmware/tree/add_schnorr_crypto and we'll see if the CI tests pass: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/267555129. If they succeed, then I will force-push the squashed branch into this PR and we'll merge.

andrewkozlik:
CI tests successful. Ready to merge once @onvej-sl and @prusnak approve.

onvej-sl:
I approve (I'm not able to find the "Add review" button).

prusnak:
Thanks! Great work everyone!

PR #1510, opened 2021-03-02
feat(core/cardano): Implement bech32 asset ids based on CIP-0014

refi93:
Motivation: [CIP-0014](https://github.com/cardano-foundation/CIPs/pull/64) specifying a human-readable token fingerprint has reached a consensus, it has been picked up by block explorers such as cardanoscan.io already, hence we are updating Trezor's UI to show it instead of the raw token policy id and name (which may or may not be printable)

Changes:
* add utility function to format the asset id as specified in CIP-0014, cover it with basic unit tests
* updated UI test fixtures to reflect the changes

Testing:
* `pytest tests/device_tests -k "test_cardano_sign_tx"` passes
* updated UI fixtures via `make test_emu_ui_record`

Affected UI fixtures before: https://drive.google.com/drive/u/0/folders/1n2nHWf7ssMOPzrxW5cHsZKrxiHdF0cnf
Affected UI fixtures after: https://drive.google.com/drive/u/0/folders/1jNKn3ts-txOYhjySaM6wbAny4kCm8O6j


matejcik:
merging into a Cardano branch, will merge to master once CI checks go through

matejcik:
in master now

PR #1512, opened 2021-03-03
feat(core): refactor display_init_seq from display_init

prusnak:


hiviah (approved):
LGTM, works on device.

PR #1515, opened 2021-03-04
style(core): fix bootloader formatting

mmilata:
Fixes https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1073573197


prusnak:
Thanks!

PR #1418, opened 2021-01-16
WebAuthn on port 21326

andrewkozlik:
When running `emu.py` (debug) and `emu.py -d` (production), the WebAuthn UDP port is different in each case, 21326 and 21325, respectively. There are two problems:
1. This is an unwelcome complication when running `hid-bridge`, because you have to set the port manually.
2. `emu.py -d` shows the wrong port number:
```
TREZOR_PATH=udp:127.0.0.1:21324
TREZOR_PROFILE_DIR=/var/tmp
TREZOR_UDP_PORT=21324
TREZOR_FIDO2_UDP_PORT=21326  # <---- here
```

We can easily unify the WebAuthn port for both debug and production by reserving the second interface for debug. That way, if WebAuthn is enabled, it will always be on the third interface, meaning that by default the emulator will run WebAuthn on port 21326 every time (except for bitcoin-only).

PR #1517, opened 2021-03-05
allow using all four USB endpoints at the same time

matejcik:
In PR #1418, @mcudev seems to have fixed the limitation that prevented us from enabling the VCP interface when Webauthn is also enabled.

Someone who understands the low-level code please look this over. We should probably ignore the changes to `usb.py` and keep the original, except allow VCP whenever debug is also allowed. We will also want #1418 at the same time, in order to have the UDP port offsets fixed; otherwise either webauthn or debuglink have non-fixed ports.

### DIFF
--- a/core/emu.py
+++ b/core/emu.py
@@ -223,7 +223,7 @@ def cli(
         TREZOR_PATH=f"udp:127.0.0.1:{emulator.port}",
         TREZOR_PROFILE_DIR=str(profile_dir.resolve()),
         TREZOR_UDP_PORT=str(emulator.port),
-        TREZOR_FIDO2_UDP_PORT=str(emulator.port + 2),
+        TREZOR_FIDO2_UDP_PORT=str(emulator.port + 1),
         TREZOR_SRC=str(SRC_DIR),
     )
     os.environ.update(emulator_env)
--- a/core/src/usb.py
+++ b/core/src/usb.py
@@ -3,9 +3,9 @@ from trezor import io, utils
 
 _iface_iter = iter(range(5))
 
-ENABLE_IFACE_DEBUG = __debug__
+ENABLE_IFACE_DEBUG = True
 # change to False to enable VCP, see below
-ENABLE_IFACE_WEBAUTHN = not utils.BITCOIN_ONLY
+ENABLE_IFACE_WEBAUTHN = True
 
 # We only have 10 available USB endpoints on real HW, of which 2 are taken up by the USB descriptor.
 # iface_wire, iface_debug and iface_webauthn also consume 2 each, iface_vcp uses 3.
@@ -13,9 +13,7 @@ ENABLE_IFACE_WEBAUTHN = not utils.BITCOIN_ONLY
 # By default, iface_vcp is only enabled on bitcoin_only firmware, where iface_webauthn
 # is disabled. Implementation-wise, we check if any of the previous ifaces is disabled
 # in order to enable VCP.
-ENABLE_IFACE_VCP = __debug__ and (
-    utils.EMULATOR or not ENABLE_IFACE_DEBUG or not ENABLE_IFACE_WEBAUTHN
-)
+ENABLE_IFACE_VCP = True
 
 # interface used for trezor wire protocol
 id_wire = next(_iface_iter)
@@ -65,7 +63,7 @@ if not utils.BITCOIN_ONLY and ENABLE_IFACE_WEBAUTHN:
         # fmt: on
     )
 
-if __debug__ and ENABLE_IFACE_DEBUG:
+if ENABLE_IFACE_DEBUG:
     # interface used for debug messages with trezor wire protocol
     id_debug = next(_iface_iter)
     iface_debug = io.WebUSB(
@@ -74,7 +72,7 @@ if __debug__ and ENABLE_IFACE_DEBUG:
         ep_out=0x01 + id_debug,
     )
 
-if __debug__ and ENABLE_IFACE_VCP:
+if ENABLE_IFACE_VCP:
     # interface used for cdc/vcp console emulation (debug messages)
     id_vcp = next(_iface_iter)
     id_vcp_data = next(_iface_iter)
@@ -99,7 +97,7 @@ bus = io.USB(
 bus.add(iface_wire)
 if not utils.BITCOIN_ONLY and ENABLE_IFACE_WEBAUTHN:
     bus.add(iface_webauthn)
-if __debug__ and ENABLE_IFACE_DEBUG:
+if ENABLE_IFACE_DEBUG:
     bus.add(iface_debug)
-if __debug__ and ENABLE_IFACE_VCP:
+if ENABLE_IFACE_VCP:
     bus.add(iface_vcp)
--- a/core/src/usb.py
+++ b/core/src/usb.py
@@ -36,15 +36,6 @@ iface_wire = io.WebUSB(
 # Therefore, each of the following needs to include the respective static expression
 # so that it can be correctly excluded from the resulting build.
 
-if __debug__ and ENABLE_IFACE_DEBUG:
-    # interface used for debug messages with trezor wire protocol
-    id_debug = next(_iface_iter)
-    iface_debug = io.WebUSB(
-        iface_num=id_debug,
-        ep_in=0x81 + id_debug,
-        ep_out=0x01 + id_debug,
-    )
-
 if not utils.BITCOIN_ONLY and ENABLE_IFACE_WEBAUTHN:
     # interface used for FIDO/U2F and FIDO2/WebAuthn HID transport
     id_webauthn = next(_iface_iter)
@@ -74,6 +65,15 @@ if not utils.BITCOIN_ONLY and ENABLE_IFACE_WEBAUTHN:
         # fmt: on
     )
 
+if __debug__ and ENABLE_IFACE_DEBUG:
+    # interface used for debug messages with trezor wire protocol
+    id_debug = next(_iface_iter)
+    iface_debug = io.WebUSB(
+        iface_num=id_debug,
+        ep_in=0x81 + id_debug,
+        ep_out=0x01 + id_debug,
+    )
+
 if __debug__ and ENABLE_IFACE_VCP:
     # interface used for cdc/vcp console emulation (debug messages)
     id_vcp = next(_iface_iter)
@@ -97,9 +97,9 @@ bus = io.USB(
     usb21_landing=False,
 )
 bus.add(iface_wire)
-if __debug__ and ENABLE_IFACE_DEBUG:
-    bus.add(iface_debug)
 if not utils.BITCOIN_ONLY and ENABLE_IFACE_WEBAUTHN:
     bus.add(iface_webauthn)
+if __debug__ and ENABLE_IFACE_DEBUG:
+    bus.add(iface_debug)
 if __debug__ and ENABLE_IFACE_VCP:
     bus.add(iface_vcp)
--- a/core/embed/trezorhal/usbd_conf.c
+++ b/core/embed/trezorhal/usbd_conf.c
@@ -427,7 +427,7 @@ USBD_StatusTypeDef  USBD_LL_Init (USBD_HandleTypeDef *pdev)
     HAL_PCD_Init(&pcd_hs_handle);
 
     // use 130 + 149 * 6 = the 1024 32-bit words in the USB OTG_HS data RAM
-    HAL_PCDEx_SetRxFiFo(&pcd_hs_handle, 130); // 128 32-bit words
+    HAL_PCDEx_SetRxFiFo(&pcd_hs_handle, 130); // 130 32-bit words
     HAL_PCDEx_SetTxFiFo(&pcd_hs_handle, 0, 149); // 149 32-bit words
     HAL_PCDEx_SetTxFiFo(&pcd_hs_handle, 1, 149); // 149 32-bit words
     HAL_PCDEx_SetTxFiFo(&pcd_hs_handle, 2, 149); // 149 32-bit words
--- a/core/embed/trezorhal/usbd_conf.c
+++ b/core/embed/trezorhal/usbd_conf.c
@@ -426,13 +426,13 @@ USBD_StatusTypeDef  USBD_LL_Init (USBD_HandleTypeDef *pdev)
     /* Initialize LL Driver */
     HAL_PCD_Init(&pcd_hs_handle);
 
-    HAL_PCDEx_SetRxFiFo(&pcd_hs_handle, 0x2e);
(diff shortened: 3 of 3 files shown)


core/src/usb.py
```diff
@@ -3,19 +3,17 @@
 
 _iface_iter = iter(range(5))
 
-ENABLE_IFACE_DEBUG = __debug__
+ENABLE_IFACE_DEBUG = True
```
prusnak, line 6: We don't want debug interface in the production build

core/src/usb.py
```diff
@@ -3,19 +3,17 @@
...
 # change to False to enable VCP, see below
-ENABLE_IFACE_WEBAUTHN = not utils.BITCOIN_ONLY
+ENABLE_IFACE_WEBAUTHN = True
```
prusnak, line 8: We don't want WebAuthn interface in the Bitcoin-only firmware

core/src/usb.py
```diff
@@ -3,19 +3,17 @@
...
 # We only have 10 available USB endpoints on real HW, of which 2 are taken up by the USB descriptor.
 # iface_wire, iface_debug and iface_webauthn also consume 2 each, iface_vcp uses 3.
 # That is a grand total of 11. That means that we can't enable everything at the same time.
 # By default, iface_vcp is only enabled on bitcoin_only firmware, where iface_webauthn
 # is disabled. Implementation-wise, we check if any of the previous ifaces is disabled
 # in order to enable VCP.
-ENABLE_IFACE_VCP = __debug__ and (
-    utils.EMULATOR or not ENABLE_IFACE_DEBUG or not ENABLE_IFACE_WEBAUTHN
-)
+ENABLE_IFACE_VCP = True
```
prusnak, line 16: We don't want VCP in the production build

prusnak (changes requested):
Low-level changes seem OK, but we need to test them like hell (on all platforms including Android) before merging (as this also affects the bootloader).

I left some comments about the higher level changes.

matejcik:
if bootloader is affected, let's get this to the next release if at all possible

mcudev:
I'll be sure to review this PR. I was just getting back around to this, so this is good timing.
I have a Beagle USB 12 protocol analyzer that I'll test this through as well (before I tested with WireShark USB logging).

mcudev:
I was just trying to figure out which model T USB interface orderings to use. What do you guys want to use?

For production, it's easy: always wire first, webauthn second (if not bitcoin only).
For debug, if bitcoin only: always wire first, debug and vcp immediately sequentially after
For debug, if NOT bitcoin only: always wire first, then we have the choice of putting debug and vcp before or after webauthn

So, I went through looking for dependencies on the interface ordering.
It looks like the python trezorlib transports assume that the debug interface will be the interface immediately following the wire interface. Thus the trezorctl util assumes that too.
The trezor bridge has the same assumption: https://github.com/trezor/trezord-go/blob/d0b5fb041d2a8e5be35899655949e154b3b7845b/usb/libusb.go#L46
The hid-bridge and emu.py assume the same thing in debug mode (which was the reason for #1418).
There's might be more that I didn't find yet.

Also, regarding the debug interface, it looks like just a second interface that is very similar to wire.
You can send debug and regular protocol messages over both. It's not like debug is limited to only debug messages.
Similarly, wire will process debug messages when the firmware is compiled in debug mode. Is there anything particularly special about the debug interface? Am I missing something? Is it just an interface to interact with the device if wire gets blocked?

matejcik:
> Is it just an interface to interact with the device if wire gets blocked?

This, yes. We sometimes need to send a debug request and get a debug response _while waiting for a different response_ on the normal interface.
Also the debug interface behaves slightly differently with regards to workflow management, debug messages are not treated as workflows.



> I was just trying to figure out which model T USB interface orderings to use. What do you guys want to use?

I see no reason to not keep what is there:
1. normal
2. (if debug) debug
3. webauthn
4. (if debug) vcp

mcudev:
that original ordering should work too. i guess i just grouped them in my mind because they are the debug interfaces. i think i'm ok with leaving that order if you guys are too.

are you guys thinking that we should basically rebase this onto the commits for #1418?

core/emu.py
```diff
@@ -223,7 +223,7 @@ def cli(
         TREZOR_PATH=f"udp:127.0.0.1:{emulator.port}",
         TREZOR_PROFILE_DIR=str(profile_dir.resolve()),
         TREZOR_UDP_PORT=str(emulator.port),
-        TREZOR_FIDO2_UDP_PORT=str(emulator.port + 2),
+        TREZOR_FIDO2_UDP_PORT=str(emulator.port + 1),
```
mcudev, line 226: assuming that we build on top of #1418 changes to this file can go away

mcudev:
I put together a better commit for this now that we're building on top of #1418. Here it is:

https://github.com/trezor/trezor-firmware/compare/master...mcudev:usb_updates

One note: in the existing production code, `MICROPY_HW_USB_VBUS_DETECT_PIN` is not defined, so we'd have: `pcd_hs_handle.Init.vbus_sensing_enable = 0;`

But, in the TT schematic (docs/hardware/model-t/schematic.png), I see `PB13` is connected to `HS_VBUS`. So, I wonder if we should continue without vbus sensing enabled? Was it an oversight before? It might not matter, and it's ok to keep it disabled, because the device is a bus powered peripheral. I'm not sure right now. It's something to keep in mind, so I left a `TODO` comment in my latest example commit.

prusnak:
> But, in the TT schematic (docs/hardware/model-t/schematic.png), I see `PB13` is connected to `HS_VBUS`. So, I wonder if we should continue without vbus sensing enabled? Was it an oversight before? It might not matter, and it's ok to keep it disabled, because the device is a bus powered peripheral. I'm not sure right now. It's something to keep in mind, so I left a `TODO` comment in my latest example commit.

Right. The PIN is connected and we don't use VBUS sensing for anything (so it is disabled). I think it doesn't matter whether we enable it or not, so we might as well keep it disabled.

mcudev:
Great, thanks for the info. I've updated the comment and will test my latest example commit more deeply tomorrow (I'll wait to see if any other changes are mentioned before then).

mcudev:
I've been testing this a lot today. I still have a lot to test tomorrow.

So far, all the USB analyzer traces looked clean. There are obviously some differences in how the Windows 10 and Ubuntu USB host code controls and queries the TT device. But, I didn't see any isssues, just differences. The expected interface endpoints were being used and everything functionally worked fine from a user perspective too.

Below are notes on what I've tested so far.

All tests today were performed on a production Trezor T running the offical SatoshiLabs bootloader version 2.0.3. I'll run tests tomorrow that test the bootloader USB stack changes on unlocked dev hardware (because the prod boardloader won't accept a dev bootloader).

I used Trezor Bridge version 2.0.27, where applicable.
I used Trezor Suite 21.2.3 on Ubuntu and 21.3.1 on Windows 10, where applicable.

The USB analyzer that I used is a Beagle USB 12 protocol analyzer (https://www.totalphase.com/products/beagle-usb12/ and https://www.totalphase.com/products/data-center/)

All testing done against commit https://github.com/mcudev/trezor-firmware/commit/c7c8b8175fdc6a03a3f18f37f027c586db4b0d92

I built a full non-debug firmware (built with `make clean vendor build BITCOIN_ONLY=0 PYOPT=1`)
Ran `trezorctl device wipe --bootloader`
Ran `trezorctl firmware-update -f build/firmware/firmware.bin`
Then I connected the TT to the host via the protocol analyzer and recorded my testing.

On Ubuntu 20.10
  After connecting the TT check for any errors with `journalctl -ke`
  check descriptors with `lsusb --verbose -d 1209:`
  test the wire interface:
    `trezorctl get-features`
    `trezorctl device setup -t 256 -r -p -s`
  test the hid interface: webauthn with dummy test code against gemini.com
```
      navigator.credentials.create({
        publicKey: {
        "rp": {
          "name": "Gemini Trust Company, LLC",
          "id": "gemini.com"
        },
        user: {
          id: new Uint8Array([1]),
          name: "user@email.test",
          displayName: "whatever"
        },
        challenge: new Uint8Array([1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]),
        "pubKeyCredParams": [{"alg": -7, "type": "public-key"}, {"alg": -257, "type": "public-key"}],
        "authenticatorSelection": {"requireResidentKey": false, "userVerification": "preferred"},
        "attestation": "none",
        "extensions":{}
      }
      });
```
  test the wire interface: trezor suite (with and without bridge). performed a bunch of different user actions.
  test the wire interface: https://wallet.trezor.io/ (with and without bridge in chrome) performed a bunch of different user actions.
  `trezorctl device wipe`

Then I hooked the TT device up to the USB analyzer on Windows 10 home x64 version 20H2 build 19042.844:
  test the wire interface: https://wallet.trezor.io/ (without bridge in chrome) loaded the device and performed a bunch of different user actions.
  test the hid interface: webauthn with dummy test code against gemini.com
  test the wire interface: trezor suite (without bridge) performed a bunch of different user actions.


Tomorrow, I'll work on testing:
- the same full non-debug firmware on Chrome OS, Android 8 phone, and an Android 11 phone
- a bitcoin only non-debug firmware (`make clean vendor build BITCOIN_ONLY=1 PYOPT=1`) on Ubuntu 20.10, Windows 10, Chrome OS, Android 8 phone, and an Android 11 phone. I don't have a Mac or an iPhone to test with.
- a full debug firmware (`make clean vendor build BITCOIN_ONLY=0 PYOPT=0`) on Ubuntu 20.10
- a bitcoin only debug firmware (`make clean vendor build BITCOIN_ONLY=1 PYOPT=0`) on Ubuntu 20.10
- the bootloader usb changes on dev hardware against Ubuntu 20.10, Windows 10, Chrome OS, Android 8 phone, and an Android 11 phone.

mcudev:
I didn't get as much done today as I wanted. But, I got through testing the same full non-debug firmware on Chrome OS, an Android 8 phone, and an Android 11 phone. They were all fine. I saw in the protocol analyzer logs that the Android devices reset the USB bus fairly frequently. None of my other tests have shown that. The bus resets did not affect the apparent functionality of the device -- everything was normal from a user experience. I went back and put the latest production SatoshiLabs firmware on the TT and ran the tests and got the same results. It's just a difference in the Android USB host code. Not a problem, just another difference. The Chrome OS logs were pretty similar to the Ubuntu logs.

More testing coming tomorrow.

mcudev:
So far today I've tested a bitcoin only non-debug firmware (`make clean vendor build BITCOIN_ONLY=1 PYOPT=1`) on Ubuntu 20.10, Windows 10, Chrome OS, Android 8 phone, and an Android 11 phone.
The testing went fine and was similar to what I did for the previous firmware. The major difference was that the webauthn HID interface was not present and thus not available for the related tests. The descriptors looked correct too.

I also tested the bootloader usb changes on dev hardware against Ubuntu 20.10, Windows 10, Chrome OS, Android 8 phone, and an Android 11 phone. I successfully wiped a firmware and loaded a firmware on all. I did some various other testing too and it all worked fine. I kind of wonder if the usb resets that I was seeing on android are due to a less perfect connection since I have an extra adapter on the analyzer cable to convert to the usb-c plug. The traces all looked reasonable. I didn't experience any user level problems either.

I just have this testing left:
- a full debug firmware (`make clean vendor build BITCOIN_ONLY=0 PYOPT=0`) on Ubuntu 20.10
- a bitcoin only debug firmware (`make clean vendor build BITCOIN_ONLY=1 PYOPT=0`) on Ubuntu 20.10

I'm pretty sure that will go fine because I've tried it a lot already, I just have to re-test the same commit for consistency and to make sure that I didn't screw up something obvious. I'll finish this today or tomorrow.

Just wanted to reiterate that I've been testing https://github.com/mcudev/trezor-firmware/commit/c7c8b8175fdc6a03a3f18f37f027c586db4b0d92 So, that commit is what will need to come over to this PR.

mcudev:
I finished the testing that I planned. I think what I've got is good. The things I noticed while testing the debug firmware builds are that the debug mode seems to slow down some user interface interactions to the point that it seems like the device froze. Eventually, those interactions finish after 30-90 seconds. I think it has to do something with the micropython and/or event loop stuff. It happens sometimes and works at regular speed. Unplugging and replugging the device usually helps. I don't think it's related to this work. Also, I noticed that when you get a USB reset, the `screen` command that is watching the VCP interface will terminate. I don't think the USB resets are a result of these changes. They could be maybe related to less than optimal physical connections (I'm not sure).
Overall, I'm pretty comfortable with the changes in https://github.com/mcudev/trezor-firmware/commit/c7c8b8175fdc6a03a3f18f37f027c586db4b0d92. If something comes up, we can dig into specific bugs deeper. 

Below is more info on what I tested (mostly for reference):

---

a full debug firmware (`make clean vendor build BITCOIN_ONLY=0 PYOPT=0`) on Ubuntu 20.10
`trezorctl device wipe --bootloader`
`trezorctl firmware-update -f build/firmware/firmware.bin`
`journalctl -ke`
`lsusb --verbose -d 1209:`

test the wire interface:
  `trezorctl get-features`
  `trezorctl device setup -t 256 -r -p -s`
test the hid interface: webauthn with dummy test code against gemini.com
```
navigator.credentials.create({
  publicKey: {
  "rp": {
    "name": "Gemini Trust Company, LLC",
    "id": "gemini.com"
  },
  user: {
    id: new Uint8Array([1]),
    name: "user@email.test",
    displayName: "whatever"
  },
  challenge: new Uint8Array([1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]),
  "pubKeyCredParams": [{"alg": -7, "type": "public-key"}, {"alg": -257, "type": "public-key"}],
  "authenticatorSelection": {"requireResidentKey": false, "userVerification": "preferred"},
  "attestation": "none",
  "extensions":{}
}
});
```
test the wire interface: trezor suite (with and without bridge)
test the wire interface: https://wallet.trezor.io/ (with and without bridge in chrome)
also with `sudo screen /dev/ttyACM0` at times
`trezorctl device wipe`

---

a bitcoin only debug firmware (`make clean vendor build BITCOIN_ONLY=1 PYOPT=0`) on Ubuntu 20.10
`trezorctl device wipe --bootloader`
`trezorctl firmware-update -f build/firmware/firmware.bin`
`journalctl -ke`
`lsusb --verbose -d 1209:`
test the wire interface:
  `trezorctl get-features`
  `trezorctl device setup -t 256 -r -p -s`
verify no HID fido interface seen by the gemini.com test
test the wire interface: trezor suite (with and without bridge)
test the wire interface: https://wallet.trezor.io/ (with and without bridge in chrome)
also with `sudo screen /dev/ttyACM0` at times
`trezorctl device wipe`

matejcik:
@mcudev this all sounds very good. could you open a PR from the appropriate branch?

closing this one

PR #1520, opened 2021-03-06
remove stay_in_bootloader to get past bootloader info screen for good firmware

mcudev:
this fixes what is probably a recently introduced bug.
the bootloader was always booting to the bootloader info screen and staying in the usb loop due to `stay_in_bootloader` always being true now.
this removes `stay_in_bootloader` and takes the touch check out of the else clause into it's own separate `if`.

### DIFF
--- a/core/embed/bootloader/main.c
+++ b/core/embed/bootloader/main.c
@@ -260,7 +260,6 @@ int main(void) {
 
   vendor_header vhdr;
   image_header hdr;
-  secbool stay_in_bootloader = sectrue;  // flag to stay in bootloader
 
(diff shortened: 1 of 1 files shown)


prusnak:
Yes, this was a indeed a bug. The `stay_in_bootloader ` value should have been `secfalse`

> this removes `stay_in_bootloader` 

I'd like to keep that flag in the code, although it is set to constant now. Once the feature is implemented in the future, we'll just use this constant.

Fixed in ebfd51d818ff8a67e74965fb24840d335d9f8994

Thanks!

PR #1522, opened 2021-03-07
fix(debug): fix show-text import and parameter exceptions

mcudev:
i was testing the usb debug stuff and noticed that this had a stale import. after that was fixed, an exception was due to the positional parameter to `Success`.

reproduce with:

`trezorctl debug show-text blah`

### DIFF
--- a/core/src/apps/debug/show_text.py
+++ b/core/src/apps/debug/show_text.py
@@ -1,8 +1,9 @@
 import trezor.messages.DebugLinkShowTextStyle as S
+import trezor.ui.components.common.text as text
 from trezor import ui, wire
 from trezor.messages.DebugLinkShowText import DebugLinkShowText
 from trezor.messages.Success import Success
-from trezor.ui import style, text
+from trezor.ui import style
 from trezor.ui.components.tt.text import Text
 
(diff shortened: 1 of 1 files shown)


core/src/apps/debug/show_text.py
```diff
@@ -1,8 +1,9 @@
 import trezor.messages.DebugLinkShowTextStyle as S
+import trezor.ui.components.common.text as text
```
prusnak, line 2: Seems like this should be `components.tt.text` (that's what we use in the rest of the codebase) - but let's wait for @matejcik - he knows more than me on this topic

```suggestion
import trezor.ui.components.tt.text as text
```

but I

matejcik:
i'm somewhat tempted to throw this code out completely, I don't think we will need it anymore. let's discuss that on standup

mcudev:
i only came across this because i was trying to test the debug usb interface with trezorctl. after looking deeper, trezorctl was using the wire interface anyways, and so was not testing what i wanted to.

AaronWashington (changes requested):
> ``

matejcik:
closing in favor of #1531

PR #1529, opened 2021-03-15
common/defs/fido: add namecheap

prusnak:


common/defs/fido/namecheap.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Namecheap",
+  "u2f": [
+    {
+      "app_id": "857101361b20a9544cdb9bef65858b6bac7013550d8f84f7efee252b96fa7c1e",
+      "label": "www.namecheap.com"
+    }
+  ],
+  "webauthn": ["www.namecheap.com"]
```
andrewkozlik, line 9:
```suggestion
  "webauthn": ["www.namecheap.com"],
  "no_icon": true
```

prusnak: icon is there, but for some weird reason github shows that `core/src/apps/webauthn/res/icon_namecheap.toif` is `Empty file.`

andrewkozlik (approved):
utACK

PR #1533, opened 2021-03-18
style(core): use new type annotation syntax

mmilata:
This brings the latest fashion of python type annotations to core, namely:

* using `None` default value for function argument no longer makes the argument `Optional`, i.e. this is no longer legal:
    ```
    def foo(x: int = None) -> None: ...
    ```
    https://www.python.org/dev/peps/pep-0484/#union-types
* `X | Y` is now used instead of `Union` and `Optional` types [PEP604](https://www.python.org/dev/peps/pep-0604/)
* `dict`, `list`, `set`, `type` are now used instead of `Dict`, `List`, `Set`, `Type` [PEP585](https://www.python.org/dev/peps/pep-0585/)

~It looks like PEP585 doesn't work for type aliases so there is still few places where the capitalized types are used~. Clarification: pipe operator doesn't work for defining type aliases, and `tuple[]` doesn't work with `...` so lines like following exist:
```
TextContent = Union[str, int]
Indices = Tuple[int, ...]
```

Protobuf messages keep the old syntax because they need to be compatible with various python versions.


core/src/protobuf.py
```diff
@@ -181,26 +187,26 @@ def readinto(self, buf: bytearray) -> int:
...
-        Type[BytesType],
-        Type[UnicodeType],
-        Type[MessageType],
+        type[BytesType],
+        type[UnicodeType],
+        type[MessageType],
     ]
-    FieldDef = Tuple[str, MessageTypeDef, Any]
-    FieldDict = Dict[int, FieldDef]
+    FieldDef = tuple[str, MessageTypeDef, Any]
```
matejcik, line 198: why does this work here but not in other places?
why is the above `Union` but has `type`s inside?

core/src/trezor/sdcard.py
```diff
@@ -1,13 +1,13 @@
...
 if False:
-    from typing import Any, Callable, Optional, TypeVar
+    from typing import Any, Callable, TypeVar
 
     T = TypeVar("T", bound=Callable)
 
 
 class FilesystemWrapper:
-    _INSTANCE = None  # type: Optional[FilesystemWrapper]
+    _INSTANCE = None  # type: FilesystemWrapper | None
```
matejcik, line 10: this should be a new-style type declaration

core/src/protobuf.py
matejcik: ah i see the updated PR description

matejcik:
> Clarification: pipe operator doesn't work for defining type aliases, and `tuple[]` doesn't work with `...`

do you know if this is by design or is it a mypy bug?

matejcik:
LGTM overall, please get this ready to merge (feel free to set it up before or after the simple-confirms PR, i'm not sure which is mroe appropriate)

core/src/trezor/sdcard.py
mmilata: Fixed in 6f68820:
```python
_INSTANCE: "FilesystemWrapper" | None = None
```

mmilata:
> do you know if this is by design or is it a mypy bug?

this i can't figure out:/

Squashed, rebased & resolved conflicts.

Issue #741, opened 2019-12-05
Parse timestamp to datetime


tsusanka:
In Stellar and most likely in some other apps as well the host is sending a timestamp to Trezor. We currently do not have a function to format it into something more readable. Since we probably (?) do not want to include the whole datetime micropython library we can either:

1. use C and its time.h library (implemented in `tsusanka/timestamp`, but it fails on `multiple definition of snprintf`)
2. write it manually
3. add the datetime library

prusnak:
[datetime.py](https://github.com/micropython/micropython-lib/blob/master/datetime/datetime.py) in case we want to go with option 3

prusnak:
Another place where we could use this function is here:

https://github.com/trezor/trezor-firmware/blob/3d50bb5ed8ddb6025e4492731fbeaaa4649e23d9/core/src/apps/wallet/sign_tx/layout.py#L92 (when `lock_time` is interpreted as timestamp and not the blockheight).

prusnak:
Code in C: https://stackoverflow.com/questions/20826847/convert-between-date-time-and-time-stamp-without-using-std-library-routines

bosomt:
**QA OK**

fw trezor-fw-regular-2.4.3-221977ad.bin
fw trezor-fw-regular-1.10.4-221977ad.bin

PR #1535, opened 2021-03-18
Timestamp to human readable date&time string (Martin Poloch home assignment)

Snapyy:
Code uses utime module which in current version offers only timezone-dependant function to convert timestamp to datetime (utime.localtime). Tested only on core emulator with computer timezone set to UTC.
The mentioned function also looses precision of the conversion with increasing timestamp. The precision is much worse than 1 second around current timestamps and with future timestamps (in more than 100 years), the precision is less than a minute and even less.
Workaround was implemented so the current precision is 1 second for the next few decades but I still suggest further inspection.


core/src/trezor/strings.py
```diff
@@ -64,3 +70,57 @@ def format_duration_ms(milliseconds: int) -> str:
...
+    to the correct datetime
+    Example:
+        timestamp                   1616057224
+        correct datetime            "2021-03-18 08:47:04"
+        utime-calculated datetime   "2021-03-18 08:46:56"
+        midnight datetime           "2021-03-18 00:00:00"
+        returned date               (2021, 3, 18, 0, 0, 0, 3, 77, 0)
+        correction                  31624 (~ 8:47:04)
+    """
+    approx_datetime = utime.localtime(timestamp)
```
andrewkozlik, line 89: On the actual device this fails with:
```
AttributeError: 'module' object has no attribute 'localtime'
```

andrewkozlik: Even if we got this working, I suspect that it might give an incorrect time on the device, because the STM32 implementation seems to be using 2000-01-01 as the beginning of the epoch as opposed to the UNIX implementation which uses 1970-01-01. It's hard to believe, but that's what I see in the C code. I didn't actually test.

core/src/trezor/strings.py
```diff
@@ -64,3 +70,57 @@ def format_duration_ms(milliseconds: int) -> str:
         divisor = 1
 
     return format_plural("{count} {plural}", milliseconds // divisor, unit)
+
+
+def _calculate_timestamp_correction(timestamp: int) -> (Tuple[tuple, int]):
+    """
+    utime module can't convert timestamp to datetime with seconds precision.
```
andrewkozlik, line 77: It looks like this problem occurs only in the UNIX implementation, because micropython quite pointlessly converts the timestamp to a C `float` which has a mantissa of 23 bits, causing the lower bits to get lost. The STM32 implementation reads it as an integer.

andrewkozlik: This is actually fixed upstream.

prusnak: Yes, we need to merge MicroPython 1.14 in before merging in this PR.

core/src/trezor/strings.py
```diff
@@ -64,3 +70,57 @@ def format_duration_ms(milliseconds: int) -> str:
...
+    >>> format_timestamp_to_human(0)
+    '1970-01-01 00:00:00'
+    >>> format_timestamp_to_human(1616051824)
+    '2021-03-18 07:17:04'
+    """
+    date, correction = _calculate_timestamp_correction(timestamp)
+    # negative correction = substract 12 hours and calculate again
+    if correction < 0:
+        date, correction = _calculate_timestamp_correction(timestamp - 43_200)
+        correction += 43_200
```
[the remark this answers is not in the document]
Snapyy: The correction is not always positive. Calling the function with for example timestamp=1547078340. In the UNIX implementation, this returns date=2019-1-10, correction=-60. Caused by rounding of the utime.localtime function.

core/src/trezor/strings.py
Snapyy: Yes, from my limited sources, I expect the behavior you described. Sorry for not mentioning it in the PR description.

core/src/trezor/strings.py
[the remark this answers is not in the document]
andrewkozlik: Oh I see, it rounds to the nearest value. For some reason I assumed it truncates, i.e. rounds down.

core/src/trezor/strings.py
andrewkozlik: If the two implementations are really using different epochs, then I have doubts whether we should be using `utime` at all. We would then have to have a piece of code like "If not EMULATOR, then subtract 30 years", which just seems very wrong.

Secondly, there is https://github.com/trezor/trezor-firmware/issues/741, which indicates that there may have been some opposition to adding the utime module to the firmware build.

prusnak: I just added PR https://github.com/trezor/trezor-firmware/pull/1548 to update uPy to 1.14 - among other things it introduces `utime.gmtime()` function which is exactly what we need here I suppose.

matejcik: fwiw #741 is not opposed to adding utime, but adding full python's `datetime`. At that time there was probably no date processing in `utime`, but now there is.

matejcik:
we now have uPy 1.17 so we are able to continue on this

apart from Bitcoin lock-time, this should be used in Stellar timebounds

PR #1541, opened 2021-03-23
Add Rust Protobuf

jpochyla:
Builds in top of #1540

WIP, needs many fixes in the Python side:
- `p2py` `build_protobuf`
  - separate messages and enums, put enums into `trezor.enums`, put messages to `mocks`.
- python unit tests

PR #1542, opened 2021-03-23
Add Rust UI and layouts

jpochyla:
PoC, builds on top of #1541 now, for no particular reason.


core/embed/rust/src/protobuf/obj.rs
```diff
@@ -67,7 +67,7 @@ impl MsgObj {
         match attr {
             Qstr::MP_QSTR_MESSAGE_WIRE_TYPE => {
                 // Return the wire ID of this message def, or None if not set.
-                Ok(self.msg_wire_id.map_or(Obj::const_none(), |wire_id| wire_id.into()))
+                Ok(self.msg_wire_id.map_or_else(Obj::const_none, Into::into))
```
matejcik, line 70: why do you prefer `map_or_else` here?

jpochyla: So that `Obj::const_none` will get called only in the `None` case. In this case it's a matter of style, really, buy there's also a lint for this that would complain with more sensitive setup.

core/embed/rust/src/ui/component/component.rs
```diff
@@ -0,0 +1,107 @@
...
+        Self {
+            timers: Vec::new(),
+            next_token: 1,
+        }
+    }
+
+    pub fn request_timer(&mut self, deadline: Duration) -> TimerToken {
+        let token = self.next_timer_token();
+        if self.timers.push((token, deadline)).is_err() {
+            // The timer queue is full. Let's just ignore this request.
```
mmilata, line 93: I wonder if panic!ing here would make it easier to track down when this actually happens, i suppose ignoring it will cause some part of UI to get stuck?

mmilata (commented):
This is shaping up nicely!

I'm wondering how we can use different component (and layout) directories when compiling for T1 and for TT. Can we just wrap two versions of module in e.g. `#[cfg(feature = "model_1")]` or does this need to be handled by `build.rs`?

I'm a bit tempted to start implementing the T1 components since we don't have all of them in python yet. Then we could do it like this?
* implement T1 components in rust, keep TT components in python, both layouts in python
* swap TT components for rust versions
* eventually rewrite layout functions to rust

We'll probably still need a python layer of layouts to handle ButtonRequests/other async stuff.

Is there anything missing that would be needed to implement e.g. `confirm_action` layout using the rust ui?

core/embed/rust/src/ui/component/button.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub fn enable(&mut self) {
+        self.set(State::Initial);
+    }
+
+    pub fn disable(&mut self) {
+        self.set(State::Disabled);
+    }
+
+    pub fn is_disabled(&self) -> bool {
+        matches!(self.state, State::Disabled)
```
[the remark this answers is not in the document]
jpochyla: In this case it is exactly the same. `matches!` comes useful in more complicated scenarios, and also `==` requires `PartialEq`, which we have derived for `State`, but it's not always the case.

core/embed/rust/src/ui/component/component.rs
jpochyla: We can `unwrap` here, yes. I've tried to avoid doing this automatically for the UI stuff, as most of the failures are not critical.

jpochyla:
> This is shaping up nicely!

<3

> I'm wondering how we can use different component (and layout) directories when compiling for T1 and for TT. Can we just wrap two versions of module in e.g. `#[cfg(feature = "model_1")]` or does this need to be handled by `build.rs`?

It doesn't need any changes in the build script. In case you want to export same (or very compatible) interface, but use different impl under each feature (or a platform), see [here for an example](https://github.com/Sinono3/souvlaki/blob/master/src/platform/mod.rs), it's a rather common Rust pattern. If you just want to conditionally include a module, under it's usual path, `#[cfg(feature = "model_1")]` is enough.

> I'm a bit tempted to start implementing the T1 components since we don't have all of them in python yet. Then we could do it like this?
> 
>     * implement T1 components in rust, keep TT components in python, both layouts in python
> 
>     * swap TT components for rust versions
> 
>     * eventually rewrite layout functions to rust
> 
> 
> We'll probably still need a python layer of layouts to handle ButtonRequests/other async stuff.
> 
> Is there anything missing that would be needed to implement e.g. `confirm_action` layout using the rust ui?

I don't think anything substantial is missing. Although I haven't done any work re. T1 components, they should be much simpler than the TT counterparts, so we can try.

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -0,0 +1,350 @@
...
+        // painting by now, and we're prepared for a paint pass.
+
+        // Drain any pending timers into the callback.
+        while let Some((token, deadline)) = inner.event_ctx.pop_timer() {
+            let token = token.try_into();
+            let deadline = deadline.try_into();
+            if let (Ok(token), Ok(deadline)) = (token, deadline) {
+                inner.timer_fn.call_with_n_args(&[token, deadline])?;
+            } else {
+                // Failed to convert token or deadline into `Obj`, skip.
```
mmilata, line 144: Would it make sense to panic here at least in debug mode, so that we catch such errors in CI?

mmilata (commented):
Looking good, I appreciate moving the components under `model_tt`, gonna make the T1 work easier. The diff is large but there are lots of formatting/import changes. I haven't done super in-depth review of the TT components since our customers currently won't see them, though the code generally looks good.

From my point of view two things are missing, which I can add if @jpochyla is working on something else:
* additional feature flag (`ui`?) that would allow us not to build the `ui` module for production builds, but still allow us to use the code in CI, for T1, and for TT experiments,
* adding support for `LayoutObj` to matejcik's memory analyzer so that layouts are not reported as unused memory.

Also not sure about the fuzz crate, if I understand correctly it's not used in normal builds so we can keep it in this PR even though it's not really related?

mmilata:
I've implemented the changes suggested above. They are probably not straightforward enough so I won't push them directly, feel free to cherry pick them if they make sense:
* 062e9210040807ed0603c28d9e7fb74b2a505806 adds `ui` feature flag that is only enabled on T1 or when `NEW_UI=1` variable is set when invoking SCons
* 65657b7046851ea0bf2c8c9cedc2e92e7338bb86 extends the memory analyzer to recognize rust layouts, this is very basic since it reports no children (e.g. pointers to strings held in embedded `struct Buffer`s

PR #1538, opened 2021-03-19
layouts: altcoin apps

mmilata:
This is a follow-up of #1480 which modifies all altcoin apps to use the layout abstraction. To support various multiple page confirms it adds the `paginate_paragraphs` function which is a generalization of `paginate_text` that allows mixing fonts in the paginated output, with the restriction that different font text must start on a new line.

In the previous PR we discussed the `truncate=` parameter which is used here to emulate the old behavior wrt. text that doesn't fit on a single screen. It's used quite a lot to keep the UI diff small (or skip the pagination logic if we know the text fits), even in some cases where pagination might be desirable. Ideas?

My aim was to keep the [UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1113283338/artifacts/master_diff/index.html) reasonably small even though there are some differences especially for Eos and Monero.

Not covered in this PR, to be done in the next (final) one:
* `apps.management` and `apps.webauthn`
* splitting `trezor.ui.layouts.tt` to multiple modules to make it smaller
* review of `br_type` values to have them consistent & useful to Suite

PR #1545, opened 2021-03-25
layouts: webauthn and management

mmilata:
This should be the final major PR in the _layouts_ series as pretty much all UI now works using layouts.

Previously all TT layouts were in a single file (`trezor/ui/layouts/tt.py`), this PR splits them into multiple files but does not add the "import magic" (i.e. `trezor/ui/layouts/__init__.py`) because it's not clear it will be needed (or if we're gonna RIIR the layouts).

Regarding memory this branch passes device tests (w/ `--random-order-bucket=none`) with minimal `--heap-size` set to 459K while master needs 461K. Might be because this branch diverged from master quite some time ago, needs rechecking.

TODO after #1538 approved: go over changes once more, consider test coverage, post UI diff


matejcik:
I rebased this and made some minor tweaks (most notably 073b412) in a separate branch: https://github.com/trezor/trezor-firmware/compare/matejcik/layouts-4?expand=1

I still want to do a detailed review, but it looks good overall.

@mmilata want to look it over if the rebase is right from your POV? or should I forcepush directly into your branch?

mmilata:
Rebase ack, please push with great force.

core/src/apps/management/change_wipe_code.py
```diff
@@ -56,27 +53,41 @@ async def change_wipe_code(ctx: wire.Context, msg: ChangeWipeCode) -> Success:
     return Success(message=msg_wire)
 
 
-def _require_confirm_action(
+async def _require_confirm_action(
```
matejcik, line 56: any particular reason you made this `async` (as opposed to using return type `Awaitable[None]`)?

matejcik:
I made some fixups and a minor UI change, please review. Also I'd like to fixup the `async` thing unless there's reason for it.

Otherwise LGTM. Not sure if this requires a changelog entry?

matejcik:
i'd like to wait with merging until QA has had a go at the U2F changes

bosomt:
**QA OK**

core/src/apps/management/change_wipe_code.py
mmilata: no reason, feel free to change it

mmilata (commented):
Fixups ACK. Not sure about the changelog as there are no functional and very little visual changes?

matejcik:
Thanks. Squashed, rebased, added changelog (because previous layout changes also have it)

bosomt:
already tested, moving to approved

Issue #1131, opened 2020-07-24
Firmware spends a lot of time looking up qstrings


hiviah:
**Describe the bug**
Signing code (and probably other code) spends abou 33% in `qstr_find_strn`.

**Firmware version and revision**
2.3.1 with synchronous protobuf branch

**To Reproduce**
Use `pytest tests/device_tests/test_msg_signtx_segwit.py::TestMsgSigntxSegwit::test_lots_of_segwit_inputs` test from `tsusanka/streaming-timing` branch. 

Probably any signing test will suffice, I can see the qstring find function when running other signing tests taking 10-25%.

Measurements from ITM trace PC sampling:

Total cumulative time spent in functions when test runs:

        % FUNCTION
    30.46 qstr_find_strn
    15.76 gc_collect_end
     7.54 gc_alloc
     6.45 gc_mark_subtree
     5.62 mp_map_lookup
     4.33 mp_execute_bytecode
     3.76 sha256_Transform
     1.78 mp_obj_get_type
     1.51 find_qstr
     1.43 memset
     0.70 mp_binary_op
     0.65 qstr_hash

Sample of 1 second running signing code:

     33.41%     3473 qstr_find_strn
     17.98%     1870 gc_mark_subtree
      8.77%      912 gc_alloc
      5.90%      614 mp_map_lookup
      4.62%      481 mp_execute_bytecode
      4.20%      437 sha256_Transform
      3.57%      372 gc_sweep
      2.47%      257 gc_deal_with_stack_overflow
      1.87%      195 mp_obj_get_type
      1.64%      171 memset
      1.56%      163 find_qstr
      0.68%       71 mp_setup_code_state
      0.64%       67 mp_binary_op
      0.57%       60 mp_load_method_maybe
      0.56%       59 qstr_hash
    -----------------
     88.44%     9202 of 10395 Samples

     Ex |   Count  |  MaxD | TotalTicks  |  AveTicks  |  minTicks  |  maxTicks 
    ----+----------+-------+-------------+------------+------------+------------
     15 |     1014 |     2 |          0  |          0 |         0  |         0
     93 |     1149 |     1 |          0  |          0 |         0  |         0

    [---H] Interval = 1004mS

prusnak:
I just did a small experiment, I patched our copy of the micropython using the following (removing `qstr_find_strn` lookups from `objstr.c`:

```patch
diff --git a/py/objstr.c b/py/objstr.c
[...]

PR #1548, opened 2021-03-26
chore(vendor): bump micropython to 1.14

prusnak:
Full changelog: https://github.com/micropython/micropython/releases/tag/v1.14

Relevant changes:
* rename `sys` module to `usys` - but `sys` should still work with both unix and stm32 ports as they both enable `MICROPY_MODULE_WEAK_LINKS`
* add `utime.gmtime()` function - can be used in https://github.com/trezor/trezor-firmware/pull/1535

### DIFF
--- a/core/embed/firmware/mpconfigport.h
+++ b/core/embed/firmware/mpconfigport.h
@@ -186,7 +186,12 @@ typedef long mp_off_t;
 
 #define MICROPY_BEGIN_ATOMIC_SECTION()     disable_irq()
 #define MICROPY_END_ATOMIC_SECTION(state)  enable_irq(state)
-#define MICROPY_EVENT_POLL_HOOK            __WFI();
+#define MICROPY_EVENT_POLL_HOOK \
+    do { \
+        extern void mp_handle_pending(bool); \
+        mp_handle_pending(true); \
+        __WFI(); \
+    } while (0);
 
 #define MICROPY_HW_BOARD_NAME "TREZORv2"
 #define MICROPY_HW_MCU_NAME "STM32F427xx"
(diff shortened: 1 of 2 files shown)


matejcik:
I'm +1 on merging but i have not reviewed the diff in detail. Should we do that?

matejcik:
4559bcb4679e04e0a5e24030675676ff6a9803f2 introduces an infinite recursion in the emulator because we call `mp_hal_delay_ms` in `MICROPY_EVENT_POLL_HOOK` (upstream calls `mp_hal_delay_us`)
so far it seems we can fix it by doing the upstream thing too

prusnak:
@matejcik did you apply the full change from this PR? (including the https://github.com/trezor/trezor-firmware/pull/1548/files#diff-ee670ff4c07d3ed9a091c5b02e7f9ebae1236ae46c196f109645527fde76ac4cR229-R234)

I hit the same infinite recursion, but it seemed to me that the change above fixed it - I might be wrong ofc.

matejcik:
you're right. i was commenting on upy 1.14 in general, did not notice that this is already fixed here

matejcik:
meanwhile 1.15 was released
https://github.com/micropython/micropython/releases/tag/v1.15

prusnak:
Switching to draft - we want 1.15

matejcik:
while we're at this, let's consider inclusion of https://github.com/trezor/micropython/commit/a1c8b941448ad1ea0e3a6922de782fbd63ecdc58
(details: #1131 )

prusnak:
1.17 released recently: https://github.com/micropython/micropython/releases/tag/v1.17

matejcik:
superseded and fixed in #1805

PR #1549, opened 2021-03-26
Show nLockTime in human-readable form

andrewkozlik:
`nLockTime` = 500000000:
image

`nLockTime` = 499999999:
image

I also fixed a bug due to which all bitmaps we flipped incorrectly.


prusnak (approved):
Code looks good - utACK

If this builds and works on the device I am fine with the change

Issue #946, opened 2020-04-14
Run device tests on an actual TT in CI


tsusanka:
Related to #935. We need to add support to [tpmb](https://github.com/mmahut/tpmb) for TT first.

tsusanka:
On today's meeting we have realized we can do this also without tpmb (with the debug mode on of course).

We could modify the bootloader and:
- Always enter the bootloader, do not require any screen touch to enter.
- Do not overwrite the unofficial bootloader using firmware.
- Do not wait for the click when the firmware is loading up even if the firmware header says so.

cc @prusnak

hiviah:
Also the firmware should be immediately booted after loading it via bootloader and not reset (otherwise we would get bootloader bootloop).

hiviah:
The bootloader will delete storage on boot to make testing deterministic.

tsusanka:
Let's wait for the "workflow separation" task currently being done by @matejcik .

mmilata:
monero tests take about 10m to run on hardware, should we add them too?

matejcik:
yeah, let's

matejcik:
with #1565, the only tests that fail on hardware are the following:
```
FAILED ../../tests/device_tests/test_msg_signtx_bcash.py::TestMsgSigntxBch::test_send_bch_multisig_change
1198FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_bitcoin_gold_nochange
1199FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_p2sh
1200FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_attack_change_input
1201FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_multisig_1
1202FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_btg_multisig_change
1203FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_btg_external_presigned
1204FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_mixed_inputs
[...]

PR #1551, opened 2021-03-26
ci: run monero tests on hardware

mmilata:
Merges the two `shell.nix` files since the test needs both the monero binary and uhubctl.

Part of #946.


ci/shell.nix
```diff
@@ -61,6 +69,10 @@ stdenv.mkDerivation ({
     darwin.apple_sdk.frameworks.Metal
     darwin.libobjc
     libiconv
+  ] ++ lib.optionals hardwareTest [
+    (callPackage ./hardware_tests/uhubctl.nix { }) # HACK FIXME replace this with just uhubctl once pinned version of nixpkgs is updated
```
prusnak, line 73: Just update to the latest nixpkgs-unstable commit that succeeded in hydra (from https://status.nixos.org) so we can drop this hack

prusnak (changes requested):

ci/shell.nix
mmilata: Tried that but it breaks tests: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1131566999 https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1131567000 will investigate.

tsusanka (approved):

prusnak (approved):

ci/shell.nix
mmilata: So I was only able to drop the hack by introducing another hack in c76aee0. Without diving into tox i don't know how to create a minimal reproducer, not even 100% sure if it's a bug in nixpkgs or tox.

PR #1557, opened 2021-03-31
feat(core/emulator): JSON memory map dump

matejcik:
use `trezor.utils.mem_dump("somefile.json")` in a key place, then
`./tools/analyze-memory-dump.py src/somefile.json` to look at what is going on. clickable analysis is in `memorymap.html`

it's pretty nice :)

~~requires the following micropython hack: [`7f15dc3` (#6)](https://github.com/trezor/micropython/pull/6/commits/7f15dc330830ccae5a49a7a50acd51ab51544f0f)~~


core/embed/extmod/trezorobj.h
```diff
@@ -27,6 +27,12 @@
 #error Use MPZ for MicroPython long int implementation.
 #endif
 
+#ifdef TREZOR_EMULATOR
+#define STATICTYPE
```
prusnak, line 31: Is there a reason why not to just override STATIC completely for emulator?

prusnak (commented):
We need to merge this PR first - https://github.com/trezor/trezor-firmware/pull/1548/files

Since it includes changes necessary for uPy 1.14 version bump (and your PR is bumping uPy to 1.14 too)

core/embed/extmod/trezorobj.h
matejcik: i don't understand reasons for `STATIC` in the first place so I made the change as non-invasive as possible.

of course if you think we can switch off `STATIC` that works too

matejcik: for that matter, it would also remove the need for the micropython hack mentioned above, which would be good.

question is, won't removing STATIC get the emulator farther from the real thing in a noticeable way?

matejcik:
this is now ready to merge, no additional micropython patches required

@prusnak do you want to look at it again or should I just merge?

prusnak (approved):

PR #1562, opened 2021-04-03
Fix for crypto tests when sizeof(int) < 8

xloem:
`ck_assert_int_eq` accepts `int`s, but this test is comparing the high bits of `uint64_t` values.

This pull request adds a `ck_assert_uint64_eq` macro which compares both high and low portions of values.

Previously only the lowest 32 bits were being tested, on my systems.

### DIFF
--- a/crypto/tests/test_check.c
+++ b/crypto/tests/test_check.c
@@ -23,6 +23,7 @@
 
 #include <check.h>
 #include <inttypes.h>
+#include <limits.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


andrewkozlik:
`ck_assert_int_eq` uses `intmax_t` since 2013, see https://github.com/libcheck/check/commit/9b74b6390219f477f4da30777390891115c2beb2. So if your system supports `int64_t` and `intmax_t` is "capable of representing any value of any signed integer type" as per C99/7.18.1.5, then `ck_assert_int_eq` should work for 64-bit arguments. So it seems to me that your libcheck is rather out of date. Could that be the problem?

What I think we should do though is change `ck_assert_int_eq` to `ck_assert_uint_eq`.

prusnak:
Superseded by https://github.com/trezor/trezor-firmware/pull/1610

prusnak:
Closing in favor of merged https://github.com/trezor/trezor-firmware/pull/1610

xloem:
Thanks for taking the time to consider this (although it addresses a different issue than #1610 does).  I thought I had checked with libcheck's HEAD commit and still ran into a compiler warning using it, but I seem to have misplaced my terminal log to see what really happened.

andrewkozlik:
@xloem you are correct that #1610 addresses a different issue. If you are able to reproduce the problem above with the latest version of libcheck, then please open an issue with detailed steps to reproduce, compiler version, platform information and the build log and we will investigate. Currently we are unable to reproduce it and it appears that libcheck should work correcty out of the box without defining any additional macros.

PR #1564, opened 2021-04-12
feat(core/cardano): add support for catalyst voting registration

gabrielKerekes:
Motivation: Add support for [Catalyst voting registration](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0015/CIP-0015.md) - currently the CIP is slightly out of date, it should be updated soon.

Changes:
- `metadata` field has been removed and replaced with `auxiliaryData` to reflect lastest [CDDL](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L212)
- auxiliary data are now sent as an object containing either a blob (this leads to the same results as including metadata previously did) or an object containing Catalyst registration parameters
- auxiliary data hash gets shown to the user while signing a transaction regardless of their type - we haven't been showing the hash at all before
- catalyst registration gets returned as part of the transaction
- due to the memory optimizations done previously we have removed the auxiliary data (metadata) size restriction
- we renamed all the `create_*` functions in `trezorlib/cardano.py` which deal with JSON objects to `parse_*` - mainly because of the `_parse_address_parameters` function which parses the data from JSON and uses `create_address_parameters` to construct the protobuf object.

Notes:
- we had to modify `pb2py.py` to support enums with one value. The `CardanoMetadataType` which we added currently has only one value - `CATALYST_REGISTRATION`. We decided to introduce and keep the enum so that the request for Catalyst registration is more explicit and also to make it easier to add possible future types without impacting the existing structures.
- the Catalyst registration auxiliary data get wrapped in a tuple and an empty tuple is added to be in accordance with the [CDDL](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L212) spec and compatible with other Cardano tools. E.g. cardano-cli formats auxiliary data as `[{61284:....},[]] ` instead of just `{61284:...}` which would also be valid.
- UI test fixtures have not yet been generated

Testing:
- `pytest tests/device_tests -k "test_cardano_sign_tx"` passes
- updated UI fixtures via `make test_emu_ui_record`


python/src/trezorlib/cardano.py
```diff
@@ -315,6 +321,69 @@ def create_withdrawal(withdrawal) -> messages.CardanoTxWithdrawalType:
...
+
+            catalyst_registration_parameters = (
+                messages.CardanoCatalystRegistrationParametersType(
+                    voting_public_key=bytes.fromhex(
+                        catalyst_registration["voting_public_key"]
+                    ),
+                    staking_path=tools.parse_path(
+                        catalyst_registration["staking_path"]
+                    ),
+                    nonce=catalyst_registration["nonce"],
```
gabrielKerekes, line 369: We've noticed here that it is possible to send numbers larger than uint64 max (18446744073709551615) and they simply go through to the Firmware, although the protobuf type is uint64. Cardano transaction signing then fails because the CBOR implementation can't (and probably shouldn't) handle larger numbers. This is true for all integers which aren't checked for upper bounds in the FW itself.

We are not sure if this is something we should be handling in trezorlib/FW or whether this should be handled somewhere else. Could you please provide some insight into this?

If a larger value is sent via Trezor Connect, FW receives uint64 max value instead of the value which has been sent.

matejcik: I'm OK with not handling this and letting the CBOR error bubble up as a FirmwareError or whatever is the final result. If you are worried about someone doing this by mistake, the place to check is here in trezorlib.

The firmware is currently not capable of enforcing protobuf numeric field sizes. I am not sure if that can be a problem anywhere; in the usual case, the final encoder should be using fixed size field and the failure will appear there.

gabrielKerekes:
Hi @matejcik, when do you expect you will be able to fully review this PR please? It would be awesome to get this into the May release.

matejcik:
That is the plan :) I'll do it sometime this week, hopefully tomorrow.

Note that there is still a possibility that we will skip the May release too.

gabrielKerekes:
> Note that there is still a possibility that we will skip the May release too.
Oh, that would be a shame, we are hoping that won't happen.

common/protob/messages-cardano.proto
```diff
@@ -218,6 +241,24 @@ message CardanoSignTx {
...
+        required CardanoAddressParametersType reward_address_parameters = 3;
+        required uint64 nonce = 4;
+    }
+
+    message CardanoTxMetadataType {
+        required CardanoMetadataType type = 1;
+        optional CardanoCatalystRegistrationParametersType catalyst_registration_parameters = 2;
+    }
+
+    message CardanoTxAuxiliaryDataType {
```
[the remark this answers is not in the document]
matejcik: reading through the details, it seems that you expect more auxiliary data types, of which "metadata" is the only one supported right now?

matejcik (commented):
looks reasonable overall

the only thing i'm concerned about is the nesting of the auxiliary data types

common/protob/messages-cardano.proto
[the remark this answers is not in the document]
gabrielKerekes: According to the [CDDL spec](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L215) `auxiliary_data` can even now contain `auxiliary_scripts` - which we don't need to support here yet. `auxiliary_scripts` is thus one field that might be added to `CardanoTxAuxiliaryDataType` later, but more might be potentially introduced. The `metadata` and `auxiliary_scripts` fields can be both present at the same type, so they aren't exclusive types of `auxiliary_data`. So when it comes to `CardanoTxAuxiliaryDataType` only new fields are probably going to be added, not `CardanoAuxiliaryDataType`s.

However, when it comes to `CardanoTxMetadataType`, we expect this to expand further with other `CardanoMetadataType`s and other optional fields.

Thus it seemed best to us to separate it this way so both can easily be extended without bloating `CardanoSignTx`.

> I'm wondering if we could also leave out blob here and keep the metadata field in the original transaction?

This could be a bit misleading, since what gets sent in the `blob` is now `auxiliary_data` not just the `metadata` part of it. So if someone would send `metadata` and `auxiliary_data` at the same time, expecting `auxiliary_data` to be built from these two parts, only one of those would actually be added to the transaction. Of course, we could validate it, but it seems confusing. This way the client's intent is explicit even at the price of backwards compatibility. Also, metadata probably haven't been widely used yet so it shouldn't be too much trouble.

common/protob/pb2py
```diff
@@ -88,6 +88,10 @@ class ProtoField:
         elif field.type == field.TYPE_ENUM:
             value_dict = descriptor.enum_types[type_name]
             valuestr = ", ".join(str(v) for v in value_dict.values())
+            # append trailing comma if enum has only one value so that valuestr
```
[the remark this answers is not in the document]
gabrielKerekes: Done. https://github.com/trezor/trezor-firmware/pull/1564/commits/85d4305c67b15e45f752db9a9274a872cef10f9d

core/src/apps/cardano/helpers/metadata_dict_keys.py
```diff
@@ -0,0 +1,2 @@
+CATALYST_REGISTRATION = 61284
```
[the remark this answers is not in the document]
gabrielKerekes: Done. https://github.com/trezor/trezor-firmware/pull/1564/commits/55873618f04653e8991d86bede937def841b2f85

common/protob/messages-cardano.proto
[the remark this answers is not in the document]
gabrielKerekes: On second thought - perhaps it could indeed be simplified. We're going to have an internal discussion on the topic and I'll get back to you.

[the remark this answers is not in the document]
matejcik: > So if someone would send `metadata` and `auxiliary_data` at the same time, expecting `auxiliary_data` to be built from these two parts, only one of those would actually be added to the transaction.

the way I'm reading it, you're **either** sending a blob that represents all of the auxiliary data, **or** some structured data. Which actually seems to be an argument _in favor_ of keeping the old field. True, you need to check that the host does not send both, but that's what I'd expect.
We can (and probably should) rename the `metadata` field to `auxiliary_data_blob`, which breaks API compatibility (can't use old app with new Connect) but not wire compatibility (old app with old Connect still works)

core/src/trezor/messages/CardanoTxMetadataType.py
```diff
@@ -0,0 +1,34 @@
+# Automatically generated by pb2py
+# fmt: off
+import protobuf as p
+
+from .CardanoCatalystRegistrationParametersType import (
```
matejcik, line 5: the formatting does not match what `make gen` produces

common/protob/messages-cardano.proto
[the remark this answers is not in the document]
gabrielKerekes: We discussed this and ended up with the simplification you proposed. Dropping `CardanoTxMetadataType` and moving `catalyst_registration_parameters` to `CardanoTxAuxiliaryDataType` and also removing `CardanoAuxiliaryDataType` and serializing based on whether `blob` or `catalyst_registration_parameters` is filled.

However, we would still remove the `metadata` field from the transaction and move it to `CardanoTxAuxiliaryDataType`. This way there's only one field related to auxiliary data in the transaction and it makes it more obvious that the fields are somewhat related. The impact on existing users should be close to zero as metadata haven't been widely used by any major Cardano tools.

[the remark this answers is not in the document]
matejcik: :+1: go ahead

core/src/trezor/messages/CardanoTxMetadataType.py
gabrielKerekes: Yes, is there something I can do about this? Running `make pystyle` reformats the file.

Although maybe it's irrelevant since the file is going to be removed.

matejcik: probably :) `make pystyle` should not be touching the file since it's explicitly saying `# fmt: off`
so that might be a bug in your local `black` copy or something?

matejcik: oh, but it's `isort` doing this. weird.

common/protob/messages-cardano.proto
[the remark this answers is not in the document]
gabrielKerekes: Restructured in https://github.com/trezor/trezor-firmware/pull/1564/commits/c82d4d141bda52e6da40a9fcb938a1830d43e5e8.

core/src/trezor/messages/CardanoTxMetadataType.py
gabrielKerekes: The problem is still there but moved to `core/src/trezor/messages/CardanoTxAuxiliaryDataType.py`. Is there something I should do about it?

gabrielKerekes:
If the restructuring is approved we don't need the one value enums fix in pb2py (https://github.com/trezor/trezor-firmware/pull/1564/commits/e085aa4c66db84ab5dd3c4a85d4f2435a71126d9). Should I keep it anyways or revert it?

matejcik:
> If the restructuring is approved we don't need the one value enums fix in pb2py ([e085aa4](https://github.com/trezor/trezor-firmware/commit/e085aa4c66db84ab5dd3c4a85d4f2435a71126d9)). Should I keep it anyways or revert it?

let's keep it, might be useful in the future

core/src/trezor/messages/CardanoTxMetadataType.py
matejcik: I think if you stuff `# isort:skip_file` into the pb2py template this will go away.

gabrielKerekes: Done in https://github.com/trezor/trezor-firmware/pull/1564/commits/5eb535fbe212ee55bed20d11ee0bc450065a09eb. Interestingly enough `# isort: skip_file` (with a space after `:`) didn't seem to work although that's what's used in the [isort docs](https://pycqa.github.io/isort/docs/configuration/action_comments/) - maybe different version?

core/src/apps/cardano/auxiliary_data.py
```diff
@@ -0,0 +1,241 @@
...
+
+AUXILIARY_DATA_HASH_SIZE = 32
+CATALYST_VOTING_PUBLIC_KEY_LENGTH = 32
+CATALYST_REGISTRATION_HASH_SIZE = 32
+
+METADATA_KEY_CATALYST_REGISTRATION = 61284
+METADATA_KEY_CATALYST_REGISTRATION_SIGNATURE = 61285
+
+
+def validate_auxiliary_data(
```
matejcik, line 39: this should also check that not more than one of the fields is set

proposal for a simple check:
```
data_provided = 0
if blob:
    data_provided += 1
    ...
elif params:
    data_provided += 1
    ...

...
if data_provided != 1:
    raise
```

core/src/apps/cardano/auxiliary_data.py
```diff
@@ -0,0 +1,241 @@
...
+        # validation to prevent CBOR injection and invalid CBOR
+        # we don't validate data format, just that it's a valid CBOR
+        cbor.decode(auxiliary_data_blob)
+    except Exception:
+        raise INVALID_AUXILIARY_DATA
+
+
+def _validate_catalyst_registration_parameters(
+    keychain: seed.Keychain,
+    catalyst_registration_parameters: CardanoCatalystRegistrationParametersType | None,
```
matejcik, line 72: should not be optional, you're calling that only in the "if params is not falsey" branch

core/src/apps/cardano/auxiliary_data.py
```diff
@@ -0,0 +1,241 @@
...
+    protocol_magic: int,
+    network_id: int,
+) -> None:
+    if not catalyst_registration_parameters:
```
matejcik, line 76: should not be optional, you're calling that only in the "if params is not falsey" branch

core/src/apps/cardano/auxiliary_data.py
```diff
@@ -0,0 +1,241 @@
...
+            protocol_magic,
+            network_id,
+        )
+
+    auxiliary_data_bytes = get_auxiliary_data_cbor(
+        keychain, auxiliary_data, protocol_magic, network_id
+    )
+
+    auxiliary_data_hash = hash_auxiliary_data(bytes(auxiliary_data_bytes))
+    await show_auxiliary_data_hash(ctx, auxiliary_data_hash)
```
matejcik, line 122: wondering if this is necessary if you show the parsed parameters. but feel free to keep it if you want

matejcik (changes requested):
two final things.

core/src/apps/cardano/auxiliary_data.py
gabrielKerekes: Sure. I wanted to write a device test for this, but the parsing in trezorlib has the same if/elif structure, so it only includes one of the fields. Should we maybe also update `python/src/trezorlib/cardano.py` to include both fields if they are passed in? Or should trezorlib also validate that only one of the fields is present?

Or maybe the client libraries (trezorlib/connect) should be the ones validate this? Because this doesn't really result in an invalid transaction, since the serialization does a if/elif for the fields.

matejcik: We don't want to accept the bad data on the device side, because then it's unclear which will be used if you send both.

You can update trezorlib to do either; raising an exception is nicer for the end-user, including both fields would allow us to test the device behavior. I don't consider it trezorlib's job to validate data this deep so I don't have a strong opinion about what happens.

Regarding that, ISTM the main reason most of the code in `trezorlib/cardano.py` exists at all is to be able to call `parse_path` on paths that are deep inside the nested structure. Instead, we could get rid of it and use `protobuf.dict_to_proto`, if we added smart handling when expected field is `repeated uint32` and received field is `str` on which `parse_path` does not fail.

Let's not make that a part of this PR, but if you confirm that otherwise we don't need the kitchen sink of `parse_*` functions, I'll make the change in a followup PR.

core/src/apps/cardano/auxiliary_data.py
gabrielKerekes: I think it's better to show it to the users, just in case they might want to verify the hash.

core/src/apps/cardano/auxiliary_data.py
gabrielKerekes: I thought the purpose of those functions is also validation of the data, but I suppose `dict_to_proto` handles that (and as you mentioned, it's maybe not even trezorlib's job)? Some validations would perhaps get lost, but nothing which isn't also handled in FW. It seems to me there isn't any special data handling/processing, which wouldn't allow the use of `dict_to_proto`.

@refi93 what do you think about this please?

gabrielKerekes: We discussed this offline with @refi93 and we don't see any issues with using `dict_to_proto` - it would simplify the parsing quite some bit. So if you could make that change, that would be awesome.

Also we realized that the same either/or structure is present in transaction outputs and it isn't correctly validated in the code - there's also just an if/elif there. So I'll add the `data_provided` pattern there also as a new commit here if that's OK. 

And I would rename the variable to `fields_provided` if that's also OK, it seems like a better fit to me.

core/src/apps/cardano/auxiliary_data.py
gabrielKerekes: Removed. https://github.com/trezor/trezor-firmware/pull/1564/commits/a2e97e9525c3f9afe7edf9ee89056ba766ebb69d

core/src/apps/cardano/auxiliary_data.py
gabrielKerekes: `fields_provided` added to auxiliary data validation https://github.com/trezor/trezor-firmware/pull/1564/commits/a2e97e9525c3f9afe7edf9ee89056ba766ebb69d

gabrielKerekes: On second thought, I won't add the output validation in this PR. We have another PR opened (just internally atm) which refactors address validation so I will add it there as it directly touches that code.

matejcik (approved):
LGTM now.
if this is all from you, please squash&rebase

gabrielKerekes:
Squashed, rebased, UI fixtures regenerated. :+1: Thank you for the review.

Is there perhaps any update on the May release please?

vantuz-subhuman:
Hi! This PR and change is not mentioned in the changelog for 2.4.0 here: https://github.com/trezor/trezor-firmware/blob/master/core/CHANGELOG.md

But am I getting it right that it IS available in that version?

refi93:
@vantuz-subhuman - yes it's included in fw 2.4.0 which has already been released but due to issue https://github.com/trezor/trezor-firmware/issues/1659 as explained in https://github.com/trezor/connect/pull/830#issuecomment-859582380 the update it's not being offered automatically in connect.trezor.io, but it still is in wallet.trezor.io

PR #1565, opened 2021-04-12
App separation and workflow restarts

### LINKED ISSUE #946: Run device tests on an actual TT in CI

### LINKED ISSUE #1114: model T responds with firmware error when loading white noise homescreen
**Describe the bug**
model T with 2.3.2 responds with firmware error when loading white noise homescreen
`Homescreen change failed.
Error details: Firmware error`
Default homescreens: `Success
Homescreen has been changed.`

with production FW 2.3.1 works fine

FW 2.3.2 Revision e70811c

 - Wallet software: web wallet b3f18eb73765f92dd35eff2c562ef1edb5dd9fa3 
 - OS: Nixos
 - Browser CHromium
 - Version 83.0.4103.116

**To Reproduce**
1. wallet/suite/trerzorctl 
2. connect device with 2.3.2 revision e70811c
3.try to load custom background

**Expected behavior**

**Screenshots**

**Additional context**

wallet log: 
["log","Mon, 20 Jul 2020 12:17:16 GMT","[trezor.js] [call] Received","Failure",{"code":"Failure_FirmwareError","message":"Firmware error"}]
["log","Mon, 20 Jul 2020 12:17:16 GMT","[trezor.js] [session] releasing"]

trezorctl error:

ash-4.4$ trezorctl set homescreen -f ./Downloads/white-noise.toif 
Error: FirmwareError: Firmware error
bash-4.4$ trezorctl set homescreen -f ./Downloads/white-noise.toif 
Error: FirmwareError: Firmware error

matejcik:
should fix #1114

also fixes #946 

I recommend you review this not as a whole but going by individual commits. That way it's much more readable what is going on:

5fa959a27e38680a31fd556804b82f9e6b7a62f1 ensures that buffers in `trezor.wire` are allocated at import time, i.e., in a more predictable place

aa2504dba5ed27da9815a3bacd12b8e3e3e33c3f just moves `empty_bytearray` to `trezor.utils` where it actually fits much better

d6d4c0d9af39815f33f483645e2e0d5ca32e03f9 refactors `storage.cache` to preallocate all the space it needs at import time. It also modifies the API, cache now only stores `bytes`, which needed some downstream modifications.
This commit breaks the CoinJoin authorization function, which is fixed separately later.

b9652545943b44d96e9ef0b3e5078b088a349c66 updates Cardano to use the new caching mechanism. For now we cache plaintext of the passphrase, because that's the easiest (and smallest) thing to store for Cardano's custom root derivation.

751688ebc0b65fe927e4c6023f4b41e25b617bfb moves the need to use `storage.device.get_device_id()` to a later point, because we moved the `usb` import before `boot` process, but the storage is not unlocked until `boot` is done.

26366b9755b6da5f6330e0620637b02e80cffe65 removes `apps.*.boot()` in favor of a single file with all the handlers registered in an if-elseif sequence. This removes the big handler dict from RAM and eliminates the need to import (and then unimport) apps.*, which generates a lot of garbage.
For now, handlers must be registered manually in the if-elif sequence.
This also decouples workflow registration from wire handling. `wire` already exposed `find_handler` to be overriden for pinlock, now the mechanism is used even for the normal handlers.

259ff3e1a9ee3fe4191f889ce37cc28667e51ee7 is the meat and potatoes:
One, we keep a small hand-curated list of imports always loaded. We must make sure that these don't hold references to anything at runtime, which can't be checked automatically unfortunately. (a later commit a3d23d4 adds more detailed docs)
Two, app registration and event loop setup is moved out into `session` module, so that `main` can repeatedly import and unimport this module and all the loaded things with it.

4e130e02cac6ba64bb46785a05666a8221fbf519 reflects this in the Python library: the emulator no longer responds _instantly_ because the event loop is stopped and restarted. Hence we allow for up to 1s for the emulator to catch up.

0bcf3a022610f47486a47cf8c272e7b65cc33bf1 is another important part that reduces run-time cross-dependencies until they are actually needed.
Most importantly, the always-loaded `apps.base` no longer imports the full kitchen sink of functionality at import time, instead deferring to when a particular handler is called. (This does not need to be done for most apps, because the handlers are imported dynamically, whereas `apps.base` always exists.)

8efc9250473475060b6dc0b8a4a2bff7de52d0a9 fixes debuglink functionality. Right now the `apps.debug` app is unloaded after every workflow, but certain things must survive that reload. This commit moves these things to an always-active module which is not reloaded.

9b8a09bfb24615c923ad6b8e582bffef4c447d9f is unrelated, it fixes a couple tests that are throwing deprecation warnings

684380eb113d2a351ec270dce72999db4dee7fb2 defragments the memory used by PathSchema, mainly in the `apps.bitcoin` app. Parsing a string into a PathSchema leaves behind a lot of GC trash. For this reason a "copy" is implemented that will create a new compact instance, so that the previous one can be collected along with the trash.

99050e256b045c146676c2fed34d9aa67bd4703a is another defrag measure: in bitcoin, pre-fill all possible kinds of TxAck* message into protobuf's `field_cache`, so that it's done in one place and hopefully sequentially, instead of ad-hoc in the middle of processing

ac7cd156f5ebd658fe2d3a378113b1eb6f945f08 fixes a bug this introduced, where `apps.debug.boot()` wipes the storage ..... which now happens after every workflow. Made debugging sort of inconvenient.

98f17e873daa411abe57c8fb50beaacd21c71574 is a refactor-for-clarity of `trezor.wire.handle_session()`. The original implementation was convoluted because it needed to juggle a lot of async balls. With new assumptions (synchronous protobuf, restart after every workflow), most of that is no longer necessary.
TODO fix comments

32e9a61a1b21f3288eb1ddef9aa9bcc235ac44a1 updates CoinJoin to work with the new cache.
We are now caching the original `AuthorizeCoinJoin` message, serialized in protobuf. The authorization code must check the message type, decode it and hand it over to the requested handler.
Because we now cache a serialization instead of a live object, we must manually update the cache when data changes, i.e., when a CoinJoin operation is authorized.
We also unfortunately don't cache the derived keychain and need to re-derive it from seed. This will come up again if we encrypt the session cache.

49447e8c368506f277ddfbe54c1a71147b67edfc makes sure that `experimental_features` behave as before; otherwise the value would get lost when the storage becomes locked.

257e0076258586f1c8a5f03783daed41cc927110 is a trick to force micropython to intern all symbols that might come up, even if they don't explicitly exist in the source code.

6bbffb22a1f58d2720a1d9e6c920d4eb3c2891e0 resolves "flickers" when homescreen is repeatedly restarted. We shove a run-time state into `storage.cache`, sort of awkwardly, and then look for "is the homescreen currently drawn?" information when deciding whether to fadeout/render/fadein.

46666c4ce27b19a7d1b9ca13f098610ac8a95c12 makes loading `apps.fido2` faster by moving once-per-powerup keys to an always-active module

1d329ac132d6e854f589d48bbe44e5aa3edf3a43 fixes a race condition in UI test generation

400bce4f139a443aa2e934ab5f9161395e67b513 bumps unit test heap size. We don't really want unlimited, because otherwise something could run wild, but otherwise we don't really care how much the unit tests eat.

fbf8cd2c6d2e0a1655377dccb6e0c0997f76ec2b ensures that the allocated space for `sys.modules` is big enough for all usecases and so that it doesn't get reallocated into the middle of a GC arena at some point.

13835d7e652a9a8e084396f43cd1967b55736964 more trickery: when loading the always-active modules, which are not cleared from memory after a workflow, we don't want _them_ to get reallocated to the middle of a GC arena. So we artificially inflate their `__dict__`s right at the beginning.

6f08c6a05b373d1379c69c30bce14a36f280157e updates UI test fixtures.
There is A LOT of recorded changes, but only like two of them are actual changes; the others are removing the final homescreen from the fixture, because due to timing this no longer draws fast enough to be picked up by the test. A later commit (452fd5a2e890df8579e633e01c6d984b3204263e) makes this visible in the master diff.

0927aecd0b8900d9b236030b6dce76d02902b4db adds the Monero test output to artifacts; this was a pretty significant thing missing previously.

e3895d3972411486a2108ec706356bb08cc1513c is a nicety: `./emu.py -e` will erase the existing profile so you don't need to enter PIN if the last test you ran set it.

452fd5a2e890df8579e633e01c6d984b3204263e should be removed before merging. It detects the UI testcases that only differ in the missing last homescreen.

a3d23d430e6c8bb362a1483061f8717d7c38cdbb improves comments in `main.py`


matejcik:
Hardware tests should now be stable wrt failures. `test_15_of_15` is still (or again?) failing on hardware, but I looked into the memory layout and the fragmentation is _reasonable_. I don't think there are any more big gains to be won. We will see if Rust protobuf helps (it very well might), or use @andrewkozlik's improvements.

matejcik:
It appears that the UI tests are failing in CI. I'll investigate, but perhaps it's the race condition with last homescreen being drawn or not. Resolving this will not touch the bulk of this. Please go ahead with the review.

matejcik:
per https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1173367716 there is only a small number of tests that fail on hardware:
```
FAILED ../../tests/device_tests/test_msg_signtx_bcash.py::TestMsgSigntxBch::test_send_bch_multisig_change
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_bitcoin_gold_nochange
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_p2sh
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_attack_change_input
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_multisig_1
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_btg_multisig_change
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_btg_external_presigned
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_mixed_inputs
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_p2sh_witness_change
FAILED ../../tests/device_tests/test_msg_signtx_bgold.py::TestMsgSigntxBitcoinGold::test_send_bitcoin_gold_change
```
plus of course `test_15_of_15`.
the bgold tests were excluded previously; interestingly, the `send_bch_multisig_change` was not.

the device now _should_(tm) behave more reasonably, so I re-enabled test randomization and removed "fail-early" flag in a3e173c

matejcik:
In a twisty sequence of changes, I modified the UI tests to force existence of the final homescreen snapshot -- even in tests where it would "organically" not appear. This hopefully avoids the race.

The big UI diff 6f08c6a05b373d1379c69c30bce14a36f280157e is reverted (1e1e59a) and replaced by a much smaller diff 4645617

the "missing homescreen definition check" is also reverted, but that was never intended to go into master so :woman_shrugging:

core/src/trezor/wire/__init__.py
```diff
@@ -150,12 +128,16 @@ async def wait(self, *tasks: Awaitable) -> Any:
...
+    WIRE_BUFFER_DEBUG = bytearray(PROTOBUF_BUFFER_SIZE)
+
 
 class Context:
-    def __init__(self, iface: WireInterface, sid: int) -> None:
+    def __init__(self, iface: WireInterface, sid: int, buffer: bytearray) -> None:
         self.iface = iface
         self.sid = sid
-        self.buffer = bytearray(PROTOBUF_BUFFER_SIZE)
+        self.buffer = buffer
```
mmilata, line 140: Would it make sense to zero the buffer here, just to be on the safe side?

core/src/trezor/wire/__init__.py
```diff
@@ -301,189 +283,163 @@ def __init__(self, msg: codec_v1.Message) -> None:
...
-            utils.unimport_end(modules)
+            try:
+                next_msg = await _handle_single_message(
+                    ctx, msg, use_workflow=not is_debug_session
+                )
+            finally:
+                if not __debug__ or not is_debug_session:
+                    # Unload modules imported by the workflow.  Should not raise.
+                    # This is not done for the debug session because the snapshot taken
+                    # in a debug session would clear
```
mmilata, line 413: Clear what?:)

core/src/apps/common/passphrase.py
```diff
@@ -41,6 +38,9 @@ async def _request_from_user(ctx: wire.Context) -> str:
 
 
 async def _request_on_host(ctx: wire.Context) -> str:
+    from trezor.messages.PassphraseAck import PassphraseAck
+    from trezor.messages.PassphraseRequest import PassphraseRequest
```
mmilata, line 42: Are there any guidelines on when to import at the module level and when to import in a function like this?

mmilata (commented):
This is some serious magic! I'm very happy about the hardware tests and the session handling simplification is cool too. Can't say I fully understand the coinjoin stuff, will go over it once more tomorrow, sending what I have for now.

core/src/all_modules.py.mako
```diff
@@ -0,0 +1,67 @@
...
+# explanation:
+# uPy collects string literals and symbol names from all frozen modules, and converts
+# them to qstrings for certain usages. In particular, it appears that qualified names
+# of modules in sys.modules must be qstrings. However, the collection process is
+# imperfect. If `apps.common.mnemonic` is always imported as `from ..common import mnemonic`,
+# the string "apps.common.mnemonic" never appears in source code, is never collected,
+# but then is generated and interned at runtime.
+# A similar thing happens in reverse: if module `storage.cache` is always imported as
+# this name, then "storage.cache" is collected but neither "storage" nor "cache" alone.
+# Which is a problem, because "cache" is a symbol that is added to `storage`'s dict.
```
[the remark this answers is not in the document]
matejcik: That's a good idea. I'll file a bug, but I'm not sure if it can reasonably be solved the way qstrings are collected currently.

core/src/all_modules.py.mako
```diff
@@ -0,0 +1,67 @@
+# generated from resources.py.mako
```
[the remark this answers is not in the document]
matejcik: 82e2b41a

core/src/apps/bitcoin/authorization.py
```diff
@@ -1,64 +1,59 @@
...
-    ) -> None:
-        self.coordinator = msg.coordinator
-        self.remaining_fee = msg.max_total_fee
-        self.fee_per_anonymity = msg.fee_per_anonymity or 0
-        self.address_n = msg.address_n
-        self.keychain = keychain
-        self.coin = coin
-        self.script_type = msg.script_type
-
-    def __del__(self) -> None:
```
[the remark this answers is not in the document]
matejcik: good catch, removed in e5378e2a

core/src/apps/cardano/__init__.py
```diff
@@ -1,10 +0,0 @@
-from trezor import wire
-from trezor.messages import MessageType
-
-CURVE = "ed25519"
```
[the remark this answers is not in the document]
matejcik: Correct. Only apps that are using `@auto_keychain` need the global parameters.
(A missing one would have been caught in device tests too.)

core/src/trezor/wire/__init__.py
matejcik: there is no reason, the buffer is write-only from Context's point of view.

ceb21517 clarifies this -- aparently there's some leftovers that were never even used

core/src/trezor/wire/__init__.py
matejcik: comments fixed in f55ecce5

core/src/apps/common/passphrase.py
matejcik: well that would be something good to document, right?

In general, "avoid importing big modules unless you know that everyone who imports this file either (a) already has them or (b) actually needs them".

It seems `passphrase.py` is done wrong, too; hiding `PassphraseRequest` and `PassphraseAck` is probably not necessary, but importing `ui` at top level is not a good thing -- `apps.common.passphrase` is pulled in from somewhere important, but the user might not even be using passphrase, or the current session might have it cached, etc.

The most important general pattern is avoiding importing `ui` on top level in files that could reasonably be used in a silent workflow.

mmilata: Yeah, making this part of documentation would certainly decrease the odds of someone like me accidentally making the fragmentation worse again.

core/src/all_modules.py.mako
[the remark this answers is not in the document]
matejcik: https://github.com/micropython/micropython/issues/7132

core/src/apps/common/passphrase.py
matejcik: imports in `apps.common.passphrase` fixed in 3aa8fe4

docs still TBD

matejcik:
two more features:

470d99b removes the dict of `MessageType`s from RAM

720d2c9 fixes error handling in case message reading from wire fails

tsusanka (approved):
I was not doing 100% thorough review - rather a quick check - but it LGTM!

tsusanka:
One more thing - maybe we could add some documentation to `docs/` regarding this?

matejcik:
abd54c0 fixes an issue where homescreen would not redraw when FIDO is done, so it looks like Trezor is frozen
ec30d94 fixes an issue where FIDO forgets the last time Trezor was successfully unlocked with a PIN

mmilata (approved):
New changes LGTM.

matejcik:
Bitcoin Gold tests no longer fail after c558f5c, which leaves only `test_15_of_15` as problematic.

PR #1570, opened 2021-04-15
fix(legacy): show dialog before rebooting to bootloader

prusnak:


tsusanka (approved):
LGTM. Maybe @hiviah could review as well?

tsusanka:
I am also pinging @matejzak for the copy.

tsusanka:
Anthonny and Matej came up with

> Restart device in bootloader mode?

to avoid the double "boot" in "reboot bootloader". What do you think? We can discuss on Slack further..

hiviah (approved):

prusnak:
Changed the wording in 3b770c5ff1ba4a7a50edfa3362a7ed0509916530

tsusanka (approved):

PR #1575, opened 2021-04-20
Use towncrier for generating changelogs

### LINKED ISSUE #1441: Enforce changelog entries via CI
something along the lines of:
if there's a change in X subdir, AND the change is not isolated to `messages` (i.e. regenerated protobuf), there MUST be a change in the `X/CHANGELOG.md` file, otherwise the task fails

This would check that we add proper changelogs, and we could decide to merge a PR even if a changelog is missing, in case that is appropriate.

There is an option to set this CI task to "allowed to fail", but that would mean that the CI pipeline would show green even with missing changelogs, right?

mmilata:
Fixes #1441. Usage: [docs/misc/changelog.md](https://github.com/trezor/trezor-firmware/blob/mmilata/towncrier/docs/misc/changelog.md)

Notes:
* [Towncrier docs](https://github.com/twisted/towncrier)
* `make changelog` was removed, `make changelog_check` now runs `towncrier build --draft` on all subprojects which is probably not that useful so maybe I should remove it too.
* The CI job has `master` hardocoded as the base branch, AFAIK there's no way to for us to obtain it in gitlab CI?
* The `tools/generate-changelog.py` stages changes to git. This is towncrier behavior which I'm not fan of, can send them a patch if we don't want this.
* Not sure whether to use something non-hidden instead of `.changelog.d`.
* I didn't convert the existing unreleased changelog entries of `legacy/bootloader` to the fragment files because there are not always issue numbers. Next release?

Examples:
* [Example release commit](https://github.com/trezor/trezor-firmware/commit/063d02d74de48f110d6852e2ace2ec5f65b7dd6b)
* [core changelog](https://github.com/trezor/trezor-firmware/blob/mmilata/towncrier-generated/core/CHANGELOG.md)
* [python changelog](https://github.com/trezor/trezor-firmware/blob/mmilata/towncrier-generated/python/CHANGELOG.md) (link to GH diff doesn't work because there's no corresponding tag)
* [legacy fw changelog](https://github.com/trezor/trezor-firmware/blob/mmilata/towncrier-generated/legacy/firmware/CHANGELOG.md)
* [CI job for branch without changelog](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1195503796)


matejcik:
I didn't realize this at first, but I strongly dislike that we are losing the "unreleased" section of changelogs. This is very useful to see changes that are already in master but not yet in a release.

What are our options?

 * generate the section with every PR (i presume this is possible?)
   - drawback: we get changelog conflicts again
     they are trivial to solve by rerunning "make changelog", but it's still a manual step. OTOH, most PRs include a manual step for the final rebase, so maybe?
 * install some sort of automation that regenerates the changelog when merging to master. Would this be possible?
 * other ideas?

docs/misc/changelog.md
```diff
@@ -0,0 +1,50 @@
...
+
+* `added`
+* `changed`
+* `deprecated`
+* `removed`
+* `fixed`
+* `security`
+* `incompatible`, used only by `python`
+
+Entries are added by creating files in the `.changelog.d` directory where the
```
matejcik, line 34: can't we use the `towncrier` tool to do this? Including instructions for that would be preferable.

matejcik:
> * other ideas?

along the lines of: is there a good other way to see "unreleased changelog" besides looking into CHANGELOG.md ?

mmilata:
Currently `make changelog_check` prints out the unreleased sections of all subprojects. You can manually do this with `towncrier build --draft --version unreleased` per-subproject, perhaps we can wrap this in a script. But I can't think of anything that doesn't create merge conflicts or pollutes git history.

tools/generate-changelog.py
```diff
@@ -0,0 +1,147 @@
...
+        return today.strftime("%Y-%m-%d")
+    else:
+        daysuffix = {1: "st", 2: "nd", 3: "rd"}.get(today.day % 10, "th")
+        return today.strftime(f"%d{daysuffix} %B %Y")
+
+
+@click.command()
+@click.argument(
+    "project",
+    type=click.Path(exists=True, dir_okay=True, resolve_path=True),
```
matejcik, line 94: should probably be `file_ok=False` ?

matejcik:
does towncrier list all sections even if they are empty?

if not, i'd suggest adding `incompatible` to the other changelogs as well, so that we have an identical template.

please add the release branch check, and update the docs to use `towncrier` tool

mmilata:
> does towncrier list all sections even if they are empty?

This can be configured, currently it doesn't list empty sections. I'll merge the templates.

Regarding release branch, I'm still not sure how they work. Looking at v2.3.6, it seems that:
* release branch is created
* fw goes to QA, fixes are committed to master and cherry-picked to the release branch
* releases are tagged, version bump may occur on the release branch
* on master, version is bumped after the commit where the release branch starts

My questions are:
* when should the `generate-changelog.py` be run? whenever version is changed, regardless if it's on release branch or master?
* what should the release branch check do?

tsusanka:
I'll document it soon, I promise.

> release branch is created
> fw goes to QA, fixes are committed to master and cherry-picked to the release branch
> releases are tagged, version bump may occur on the release branch
> on master, version is bumped after the commit where the release branch starts

Yes, correct.

> My questions are:
> when should the generate-changelog.py be run? whenever version is changed, regardless if it's on release branch or master?

We will run it just before creating release branch?

> what should the release branch check do?

Check on `release/YY.MM` branches that the command does not generate anything - therefore the changes were committed?

mmilata:
> Check on release/YY.MM branches that the command does not generate anything - therefore the changes were committed?

So each cherry-pick (which presumably contains changelog fragments) to release branch must be followed by version bump & new changelog section?

tsusanka:
Version bump? Version bump not but very good point about the cherry-picks. I am honestly not sure if the CI job makes sense then :thinking:. We can discuss at tomorrow's standup.

matejcik:
my original thought was, if you cherry-pick onto a release branch, you need to regenerate the changelog. but that probably won't work just like that, because towncrier would add a new version for that, right?

we'd actually need to patch the already generated section -- so either do that manually, or write a script to do it

docs/misc/changelog.md
mmilata: Added mention in 5b80e81f4887f153875e7e37d4d81490cf414c12.

tools/generate-changelog.py
mmilata: Fixed in 5b80e81f4887f153875e7e37d4d81490cf414c12.

mmilata:
Fixups:
* 5b80e81f4887f153875e7e37d4d81490cf414c12: add `incompatible` category to all subdirs, paragraph about release branches in docs
* 79f3bcc853b5edbf1bd244cff91c3a7540ee713f: adds `CHANGELOG.unreleased` scripts that print unreleased changes
* f90c7b44ce9f2a41d2b078f6909bc89efb6ffd0c: adds CI check for `release/...` branches

PTAL

I still owe the script for bumping versions.

mmilata:
Added script for updating versions. Should it increment patch version if no version number is provided? And should it have flag for also updating `FIX_VERSION_(MAJOR|MINOR|PATCH)`?

tsusanka:
@mmilata awesome! I think it is okay to keep it as it is for now and we can fine-tune it along the way :).

matejcik (approved):
lgtm let's do this

mmilata:
Rebased. Also removed `.gitattributes` because we no longer need `merge=union` for changelogs I think.

tsusanka (approved):

PR #1580, opened 2021-04-22
feat(legacy): make sure that known_bootloader check contains the hash of our bundled bootloader

prusnak:
This kicked me in the nuts during the last singing round.

Not ideal way how to do it, but can't think of a prettier method.


matejcik (approved):
utACK

Issue #1371, opened 2020-11-26
TestMultisig::test_15_of_15 fails on device due to fragmented heap


mmilata:
**Describe the bug**
While there is multiple places where I've seen the device returning an error caused by fragmented memory, I chose to write down what I found out about this one because it is easy to trigger reliably.

* On normal firmware in debug mode it suffices to run the test right after boot for it to fail.
* On btconly firmware in debug mode you need to run couple of other tests before this one for the bug to manifest:

```
28209056 trezor.wire ERROR exception:
Traceback (most recent call last):
  File "trezor/wire/__init__.py", line 387, in handle_session
  File "trezor/loop.py", line 526, in __iter__
  File "trezor/loop.py", line 191, in _step
  File "apps/bitcoin/keychain.py", line 184, in wrapper
  File "apps/bitcoin/sign_tx/__init__.py", line 70, in sign_tx
  File "apps/bitcoin/sign_tx/bitcoin.py", line 58, in signer
  File "apps/bitcoin/sign_tx/bitcoin.py", line 194, in step4_serialize_inputs
  File "apps/bitcoin/sign_tx/bitcoin.py", line 576, in sign_nonsegwit_input
  File "apps/bitcoin/sign_tx/bitcoin.py", line 728, in input_derive_script
  File "apps/bitcoin/scripts.py", line 68, in input_derive_script
  File "apps/bitcoin/scripts.py", line 406, in input_script_multisig
  File "apps/common/writers.py", line 12, in empty_bytearray
MemoryError: memory allocation failed, allocating 1602 bytes
```

Allocations for the single test: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/875883108/artifacts/core/prof/memperf-html/index.html

**Firmware version and revision**
2020-11-26 master / 01d105804821c881b24a2d69b58e6094eb1f5f2c

See also #590.

mmilata:
Also happens with `test_msg_signtx_segwit.py`, see #1479. This wasn't the case before f1382bf.

mmilata:
Also `test_op_return.py`. See `$TESTS_SKIP` in [`ci/test-hw.yml`](https://github.com/trezor/trezor-firmware/blob/master/ci/test-hw.yml) for more complete list.

mmilata:
Update: there was a massive improvement in heap fragmentation thanks to #1541 and #1565, however `test_15_of_15` remains one test that we still skip.

[...]

PR #1581, opened 2021-04-23
Reduce string manipulations in BTC signing

andrewkozlik:
This PR aims to reduce memory allocations and fragmentation mainly in Bitcoin signing by using writer semantics and memoryviews instead of string manipulations. I am not sure how effective the individual changes are, for example the global static buffer for Tx serialization might be pointless now that we have the workflow restarts.

### DIFF
--- a/core/src/apps/common/writers.py
+++ b/core/src/apps/common/writers.py
@@ -1,6 +1,7 @@
 from trezor.utils import ensure
 
 if False:
+    from typing import Union
     from trezor.utils import Writer
 
 
@@ -68,7 +69,7 @@ def write_uint64_be(w: Writer, n: int) -> int:
     return 8
 
 
-def write_bytes_unchecked(w: Writer, b: bytes) -> int:
+def write_bytes_unchecked(w: Writer, b: Union[bytes, memoryview]) -> int:
     w.extend(b)
     return len(b)
 
--- a/crypto/tests/test_check_cashaddr.h
+++ b/crypto/tests/test_check_cashaddr.h
@@ -55,9 +55,17 @@ START_TEST(test_cashaddr) {
     rawdata_len = base58_decode_check(valid_cashaddr[i].legacy, HASHER_SHA2D,
                                       rawdata, sizeof(rawdata));
     ck_assert_uint_eq(rawdata_len, 21);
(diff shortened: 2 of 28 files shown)


crypto/tests/test_check_cashaddr.h
```diff
@@ -55,9 +55,16 @@ START_TEST(test_cashaddr) {
...
-                              : rawdata[0] == 5 ? 0x08
-                                                : -1);
+
+    int addr_type = -1;
+    if (rawdata[0] == 0)
+      addr_type = 0x00;  // P2PKH
+    else if (rawdata[0] == 5)
+      addr_type = 0x08;  // P2SH
+    else
+      ck_abort();
```
prusnak, line 65:
```suggestion
    if (rawdata[0] == 0) {
      addr_type = 0x00;  // P2PKH
    } else if (rawdata[0] == 5) {
      addr_type = 0x08;  // P2SH
    } else {
      ck_abort();
    }
```

prusnak (changes requested):

crypto/tests/test_check_cashaddr.h
prusnak: ff2b819

mmilata:
Rebased onto current master as there were some conflicts, mostly in protobuf message imports.

mmilata:
I haven't done an in depth review but tested it with T1 and it makes several tx signing device tests pass that previously failed. So the changes are definitely an improvement in this regard.

matejcik:
might fix #1371

core/src/apps/common/writers.py
```diff
@@ -1,4 +1,4 @@
-from trezor.utils import ensure
+from trezor.utils import empty_bytearray, ensure  # noqa: F401
```
mmilata, line 1: I think `empty_bytearray` is not needed anymore? Maybe it was before I rebased.

mmilata (approved):
Reviewed & did not found any real problem, good work! Confirming that it fixes #1371.

core/src/apps/common/writers.py
andrewkozlik: You are right, I removed it and the tests are passing just fine.

PR #1583, opened 2021-04-25
Python formatting with poetry

dankonino:
Using github actions, for installing python & poetry.
And then running `poetry run make pystyle_check` which does check if everything is properly formatted.

Task for interview with @tsusanka

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,8 @@
+# Github actions workflow
+
+Enable **github actions** in your repo.
+Config file is  `.github/workflows/main.yml` 
+
+### Poetry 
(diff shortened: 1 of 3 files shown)


tsusanka:
Thanks @dankonino! Please:
- Investigate why the NixOS job fails (tip: investigate " vs “ :)).
- Squash commits into one nice [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) and force push.

dankonino:
Yep squash makes a lot of sense, sorry :D
And as for the investigation 
```
“poetry install”
error: getting status of '/home/runner/work/trezor-firmware/trezor-firmware/install”': No such file or directory```

I think it's looking for a package that doesn't exist there, will try few things.

PR #1584, opened 2021-04-26
pystyle_check with actions

dankonino:
### GH actions for pyhon formatting
Clean up the commits
NixOS I still need to test, so I will do the separate one for that.


tsusanka:
Thanks, we will rework the PR a little bit.

PR #1587, opened 2021-04-30
feat(ci): add checks to github actions

vdovhanych:
Added the rest of the checks from travis-ci to github actions, checks will run on pull requests.

### DIFF
--- a/.github/workflows/prebuild.yml
+++ b/.github/workflows/prebuild.yml
@@ -1,0 +1,44 @@
+name: "Prebuild checks"
+
+on: [pull_request]
+
+jobs:
+
+  style_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: cachix/install-nix-action@v13
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+        name: "Run script"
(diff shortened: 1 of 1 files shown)


.github/workflows/prebuild.yml
```diff
@@ -0,0 +1,44 @@
+name: "Prebuild checks"
+on:
+  pull_request:
```
tsusanka, line 3: I am bit surprised this is a valid syntax without any letters follwing the `:`. Shouldn't it be
```
on:
  - pull_request
```
?

tsusanka (commented):
Thanks! Looks great, I have one tiny question.

.github/workflows/prebuild.yml
vdovhanych: Based on the [docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on)there seems to be multiple options but none is with - should be correct. Maybe it could be changed to ```
on: [pull_request]
```

tsusanka: Ok maybe let's change it to that? Looks better to me :upside_down_face:

vdovhanych: Ok, I changed it to the `on: [pull_request]`.

PR #1596, opened 2021-05-05
Create Syscoin Testnet

claudiovb:
We added [syscoin](https://syscoin.org) testnet


claudiovb:
Also Enabled support for both sys and sys testnet to trezor-connect
@sidhujag 
@matejcik

prusnak:
I am against adding of testnets of arbitrary coins. If you need this testnet, feel free to build your own firmware with the changes you suggested.

sidhujag:
> I am against adding of testnets of arbitrary coins. If you need this testnet, feel free to build your own firmware with the changes you suggested.

I don't understand, you have a bunch of coins with testnets that are much newer than Syscoin?

prusnak:
Two wrongs don't make a right. 

Why do you need the Testnet included in the official build? Can't you build a firmware with Testnet included by yourself?

claudiovb:
@prusnak 
We have done a build with Testnet included and are using the firmware for testing purposes,  the point for having testnet included  in the official build is due to the fact we're developing a wallet that have testnet as optional network and then developers( the ones that will switch to test network) would be able to code on top of that( using our wallet sdk) with all accounts in testnet, trezor included... 

But anyways if you think its better we can  just make a new PR that only add SYS mainnet support to connect since that's the necessary for our application to work with trezor support ( but this will require us exclude  the trezor account when a user switch to test network due lack of support) .

sidhujag:
Yup that's fine. I just wasn't sure about the policy for testnet's but thanks @prusnak for clearing that up. Would have been awesome to be in the official build for easier testing/deployment process but it's all good otherwise. Thanks

PR #1599, opened 2021-05-05
fix(core): change logic of vendor header comparison

prusnak:
Previously we checked whether the current vendor header and
the new vendor header are the same by comparing the embedded keyset.

What originally looked like a good idea is not that good, because
this disallows us from ever changing the vendor header signing keys
without causing erasure of the storage during the version update.

This commit fixes that by changing the logic to comparing just the
vendor string.

Change of function names is purely cosmetic:
* vendor_keys_hash -> vendor_header_hash
* check_vendor_keys_lock -> check_vendor_header_lock

### DIFF
--- a/common/protob/messages-management.proto
+++ b/common/protob/messages-management.proto
@@ -75,7 +75,7 @@ message Features {
     optional uint32 fw_minor = 23;              // reported firmware version if in bootloader mode
     optional uint32 fw_patch = 24;              // reported firmware version if in bootloader mode
     optional string fw_vendor = 25;             // reported firmware vendor if in bootloader mode
-    optional bytes fw_vendor_keys = 26;         // reported firmware vendor keys (their hash)
+    // optional bytes fw_vendor_keys = 26;      // obsoleted, use fw_vendor
     optional bool unfinished_backup = 27;       // report unfinished backup (equals to Storage.unfinished_backup)
     optional bool no_backup = 28;               // report no backup (equals to Storage.no_backup)
     optional bool recovery_mode = 29;           // is recovery mode in progress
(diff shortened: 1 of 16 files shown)


matejcik:
@andrewkozlik could you take a look from security PoV ?

matejcik:
@prusnak please add changelog

prusnak:
I will add the changelog once https://github.com/trezor/trezor-firmware/pull/1652 is merged

prusnak:
Added changelogs in e149625f92b7f378013381d5da2564b15e746dd0

andrewkozlik (commented):
My apologies for taking so long to review this.
- ACK from security PoV.
- `FLASH_OTP_BLOCK_VENDOR_KEYS_LOCK` might deserve to be renamed to `FLASH_OTP_BLOCK_VENDOR_HEADER_LOCK` or `FLASH_OTP_BLOCK_VENDOR_STRING_LOCK`, because it no longer has anything to do with keys.
- The `fw_vendor_keys` field in the `Features` message also doesn't represent a hash of the keys any longer. Either it should be fixed to return a hash of the pubkeys or it should be deprecated because it's basically just a hash of the `fw_vendor` string, which makes it redundant.

prusnak:
> * `FLASH_OTP_BLOCK_VENDOR_KEYS_LOCK` might deserve to be renamed to

Done in https://github.com/trezor/trezor-firmware/pull/1599/commits/9b1efff16542072bf252dd52234275c76c3fe642

> * The `fw_vendor_keys` field in the `Features` message also doesn't represent a hash of the keys any longer

Deprecated the field in de1378d

andrewkozlik (approved):

prusnak:
Rebased and squashed

bosomt:
we can probably move this change to January sprint as there is no Trezor model T release this month

Issue #1588, opened 2021-05-02
RuntimeError('FirmwareError: firmware error') when trying to sign a simple transaction


hakonamatata:
I am starting to get a bit worried. I can not sign transactions and spend/send. I have tried one input to one output. One input to an output and a change address. Basic transactions. Nothing works, I get this firmware error every time.

**Firmware version and revision**

Using Trezor model T

Firmware version
2.3.6

**Desktop/smartphone setup (please complete the following information):**
Tried using Trezor Suite and Electrum, same issue. Does not matter what wallet I use. 

**To Reproduce**
Any transaction with Firo it seems. Fails with Firmware error every time. 

**Screenshots**
![image](https://user-images.githubusercontent.com/4003969/116810016-92db4680-ab41-11eb-8100-241fe652a223.png)

**Additional context**
I am using Firo [https://www.firo.org/](https://www.firo.org/)

All my coins are locked with Trezor, but if I can't spend that is a big problem.

prusnak:
What Firo wallet is on the screenshot? Is that Electrum-Firo? Which version?

hakonamatata:
> 
> 
> What Firo wallet is on the screenshot? Is that Electrum-Firo? Which version?

Firo Electrum Version 4.0.9.2

prusnak:
> Firo Electrum Version 4.0.9.2

Please report the issue to their issue tracker - https://github.com/firoorg/electrum-firo/issues - they are sending coin name `Zcoin` even though the coin was renamed to `Firo`: https://github.com/firoorg/electrum-firo/blob/21052243d57b3c51fbd6aecaf3353e1babef7454/electrum_dash/plugins/trezor/trezor.py#L210

a-bezrukov:
@prusnak,
Pavol,
Zcoin was rebranded to Firo on March 25 trezor/trezor-firmware#1544 (comment)
The latest firmwares for Trezor are dated 15 February 2021 (Model T) and 10 February 2021 (Trezor One). So, this change should not be in the firmwares yet.
Am I right thinking that Model T firmware 2.3.6 should have coin = "Zcoin"?

matejcik:
regardless, bad coin name should not result in "Firmware error"

prusnak:
> regardless, bad coin name should not result in "Firmware error"

How about missing input value? Should that return also a more meaningful error?

matejcik:
[...]

PR #1602, opened 2021-05-06
fix(common): remove Firo

prusnak:
using Firo leads to unspendable coins, let's remove this coin until we can spend the coins again

context:
- https://github.com/trezor/trezor-firmware/issues/1588
- https://github.com/firoorg/electrum-firo/issues/3

(kind of) fixes https://github.com/trezor/trezor-firmware/issues/1588


matejcik:
I'd suggest delisting Firo from tio/coins, but keeping firmware support until the next release, so that people who are helped by #1565 can send their coins out?

yura-pakhuchiy:
Please consider https://github.com/trezor/trezor-firmware/issues/1588#issuecomment-833520062 before making decision

matejcik:
superseded by #1647

a-bezrukov:
Hi Guys,
We are reworking these large transactions and moving the long data into the extra payload area. It's hard to provide any ETA on this, but this work is starting right now.
Thanks.

PR #1612, opened 2021-05-13
refactor(core): disable SD card, SBU, fatfs for T1 build

### LINKED ISSUE #1163: Disable Core features for T1 port
Currently the T1 port includes the following features which are not available on T1:

  - `core/embed/extmod/modtrezorio/modtrezorio-fatfs.h`
  - `core/embed/extmod/modtrezorio/modtrezorio-sdcard.h`
  - `core/embed/extmod/modtrezorio/modtrezorio-sbu.h`

We should remove these from `core/embed/extmod/modtrezorio/modtrezorio.c` and make sure these (`trezor.io.fatfs`, `trezor.io.sdcard`, `trezor.io.sbu`) are never called from MicroPython while running on T1.

mmilata:
Fixes #1163.

### DIFF
--- a/core/.changelog.d/1163.fixed
+++ b/core/.changelog.d/1163.fixed
@@ -1,0 +1,1 @@
+Disable TT features (SD card, SBU, FAT) for T1 build.
--- a/core/SConscript.firmware
+++ b/core/SConscript.firmware
@@ -139,10 +139,13 @@ if FEATURE_FLAGS["SECP256K1_ZKP"]:
 
 # modtrezorio
 SOURCE_MOD += [
-    'embed/extmod/modtrezorio/ff.c',
-    'embed/extmod/modtrezorio/ffunicode.c',
     'embed/extmod/modtrezorio/modtrezorio.c',
 ]
+if TREZOR_MODEL == "T":
+    SOURCE_MOD += [
+        'embed/extmod/modtrezorio/ff.c',
+        'embed/extmod/modtrezorio/ffunicode.c',
+    ]
 
 # modtrezorui
 CPPPATH_MOD += [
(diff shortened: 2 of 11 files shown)


matejcik:
the `sdcard_tt` redirect is weird.

why can't we exclude `trezor/sdcard.py` directly, and fix this at import sites?

mmilata:
I wanted to postpone changing `apps/common/request_pin.py` until we get to implementing T1 pin entry (which needs layouts). Still, I suppose it can be done better, let me see.

matejcik:
`apps.common.request_pin` should not be importing `storage.sd_salt` on top level (ideally not at all, but it seems it's the best way to implement `can_lock_device` so the model difference should be handled there)

`apps.common.sdcard` seems like a very good place for the redirect of the sort you did

mmilata:
I've moved the redirect to `apps/common/sdcard.py`, PTAL. Also relevant parts of trezorhal are now built only for TT. It's getting a bit messy but at least it shaves ~30K off the fw image size.

matejcik:
> at least it shaves ~30K off the fw image size.

does this also factor the Python sources?

I was comparing the earlier version with the current one, and ISTM the earlier approach from a199eb2 is more correct. That is, localizing the difference to `trezor.sdcard`, and have the module available but returning different data.

What do you think about this? https://github.com/trezor/trezor-firmware/compare/matejcik/sdcard?expand=1
(also removing the `trezor.fatfs` redirect, it didn't seem to help anything)

mmilata:
> does this also factor the Python sources?

Not sure what you mean, it was the difference in size of firmware.bin files.

> What do you think about this?

Looks reasonable, merged it with the C/buildsystem changes and force-pushed, hopefully that's what you had in mind. PTAL

matejcik (approved):
utACK

mmilata:
Resolved conflicts, please take a look when you have time @prusnak.

alex-jerechinsky:
@prusnak please review when possible

prusnak (approved):

prusnak (approved):
LGTM

PR #1613, opened 2021-05-17
fix(core): imports needed for T1 startup

mmilata:
Currently core on T1 errors out on ImportError before the firmware initializes USB. While this is mostly solved by #1545 it's probably not going to be in master soon so I'd like to propose this change to at least make T1 fully boot.


matejcik (approved):
LGTM, please merge when CI passes

PR #1615, opened 2021-05-19
feat(python/trezorctl): new firmware commands (update, download, verify), CLI tests

### LINKED ISSUE #1258: extend `trezorctl firmware` subcommand
it would be useful to convert `trezorctl firmware-update` to `trezorctl firmware update` (keeping aliases as-is of course)

with additional subcommands:

`trezorctl firmware check FIRMWARE_FILE`
* prints info about file, same as `firmware update -n` does. Exits 0 on valid or 1 on invalid firmware.
* `--quiet` - do not print info
* `--fingerprint` - only print fingerprint
* `--remote`??? look at SL servers and check if the fingerprint matches?

`trezorctl firmware download`
* download FW from SL servers and save it locally
* usual options like `--version`, `--beta` and `--bitcoin-only` apply
* model is taken from currently connected (or selected) Trezor, or specify `--model=1|T`

grdddj:
Adding new CLI command - (firmware), with three subcommands (update, download, verify).

Creating a prototype of CLI testing under cli/tests.

Fixes #1258

### DIFF
--- a/python/src/trezorlib/cli/firmware.py
+++ b/python/src/trezorlib/cli/firmware.py
@@ -14,13 +14,17 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
+import os
 import sys
+from typing import BinaryIO
+from urllib.parse import urlparse
 
 import click
 import requests
 
 from .. import exceptions, firmware
-from . import with_client
+from ..client import TrezorClient
+from . import TrezorConnection, with_client
 
 ALLOWED_FIRMWARE_FORMATS = {
     1: (firmware.FirmwareFormat.TREZOR_ONE, firmware.FirmwareFormat.TREZOR_ONE_V2),
@@ -28,12 +32,34 @@ ALLOWED_FIRMWARE_FORMATS = {
 }
 
 
-def _print_version(version):
+def _print_version(version: dict) -> None:
     vstr = "Firmware version {major}.{minor}.{patch} build {build}".format(**version)
     click.echo(vstr)
 
 
-def validate_firmware(version, fw, expected_fingerprint=None):
+def _is_bootloader_onev2(client: TrezorClient) -> bool:
+    """Check if bootloader is capable of installing the Trezor One v2 firmware directly.
+
+    This is the case from bootloader version 1.8.0, and also holds for firmware version
+    1.8.0 because that installs the appropriate bootloader.
+    """
+    f = client.features
+    version = (f.major_version, f.minor_version, f.patch_version)
+    bootloader_onev2 = f.major_version == 1 and version >= (1, 8, 0)
+    return bootloader_onev2
+
+
+def _get_file_name_from_url(url: str) -> str:
+    """Parse the name of the file being downloaded from the specific url."""
+    full_path = urlparse(url).path
+    return os.path.basename(full_path)
+
+
+def print_firmware_version(
+    version: str,
+    fw: firmware.ParsedFirmware,
+) -> None:
+    """Print out the firmware version and details."""
     if version == firmware.FirmwareFormat.TREZOR_ONE:
         if fw.embedded_onev2:
             click.echo("Trezor One firmware with embedded v2 image (1.8.0 or later)")
@@ -47,9 +73,20 @@ def validate_firmware(version, fw, expected_fingerprint=None):
         click.echo("Trezor T firmware image.")
         vendor = fw.vendor_header.text
         vendor_version = "{major}.{minor}".format(**fw.vendor_header.version)
-        click.echo("Vendor header from {}, version {}".format(vendor, vendor_version))
+        click.echo(f"Vendor header from {vendor}, version {vendor_version}")
         _print_version(fw.image.header.version)
 
+
+def validate_signatures(
+    version: str,
+    fw: firmware.ParsedFirmware,
+) -> None:
+    """Check the signatures on the firmware.
+
+    Prints the validity status.
+    In case of Trezor One v1 prompts the user (as the signature is missing).
+    Exits if the validation fails.
+    """
     try:
         firmware.validate(version, fw, allow_unsigned=False)
         click.echo("Signatures are valid.")
@@ -68,24 +105,63 @@ def validate_firmware(version, fw, expected_fingerprint=None):
         click.echo("Firmware validation failed, aborting.")
         sys.exit(4)
 
+
+def validate_fingerprint(
+    version: str, fw: firmware.ParsedFirmware, expected_fingerprint: str = None
+) -> None:
+    """Determine and validate the firmware fingerprint.
+
+    Prints the fingerprint.
+    Exits if the validation fails.
+    """
     fingerprint = firmware.digest(version, fw).hex()
-    click.echo("Firmware fingerprint: {}".format(fingerprint))
+    click.echo(f"Firmware fingerprint: {fingerprint}")
     if version == firmware.FirmwareFormat.TREZOR_ONE and fw.embedded_onev2:
         fingerprint_onev2 = firmware.digest(
             firmware.FirmwareFormat.TREZOR_ONE_V2, fw.embedded_onev2
         ).hex()
-        click.echo("Embedded v2 image fingerprint: {}".format(fingerprint_onev2))
+        click.echo(f"Embedded v2 image fingerprint: {fingerprint_onev2}")
     if expected_fingerprint and fingerprint != expected_fingerprint:
-        click.echo("Expected fingerprint: {}".format(expected_fingerprint))
+        click.echo(f"Expected fingerprint: {expected_fingerprint}")
         click.echo("Fingerprints do not match, aborting.")
         sys.exit(5)
 
 
-def find_best_firmware_version(
-    bootloader_version, requested_version=None, beta=False, bitcoin_only=False
-):
-    url = "https://data.trezor.io/firmware/{}/releases.json"
-    releases = requests.get(url.format(bootloader_version[0])).json()
+def check_device_match(
+    version: str,
+    fw: firmware.ParsedFirmware,
+    bootloader_onev2: bool,
+    trezor_major_version: int,
+) -> None:
+    """Validate if the device and firmware are compatible.
+
+    Prints error message and exits if the validation fails.
+    """
+    if (
+        bootloader_onev2
+        and version == firmware.FirmwareFormat.TREZOR_ONE
+        and not fw.embedded_onev2
+    ):
+        click.echo("Firmware is too old for your device. Aborting.")
+        sys.exit(3)
+    elif not bootloader_onev2 and version == firmware.FirmwareFormat.TREZOR_ONE_V2:
+        click.echo("You need to upgrade to bootloader 1.8.0 first.")
+        sys.exit(3)
+
+    if trezor_major_version not in ALLOWED_FIRMWARE_FORMATS:
+        click.echo("trezorctl doesn't know your device version. Aborting.")
+        sys.exit(3)
+    elif version not in ALLOWED_FIRMWARE_FORMATS[trezor_major_version]:
+        click.echo("Firmware does not match your device, aborting.")
+        sys.exit(3)
+
+
+def get_all_firmware_releases(
+    bitcoin_only: bool, beta: bool, major_version: int
+) -> list:
+    """Get sorted list of all releases suitable for inputted parameters"""
+    url = f"https://data.trezor.io/firmware/{major_version}/releases.json"
+    releases = requests.get(url).json()
     if not releases:
         raise click.ClickException("Failed to get list of releases")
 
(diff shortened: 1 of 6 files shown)


matejcik (changes requested):
Thank you for your contribution!

The submission looks reasonable overall. Most of my comments are concerned with style, some are about the resulting functionality.

Don't worry about the unrelated changes to `OPTIONS.rst`, these are because the file in master is not up-to-date with recent changes.

As noted in comments, please remove the test file completely. If we introduce CLI tests, these should be based on pytest and the Click testing functionality.

Please also add a changelog entry (see https://docs.trezor.io/trezor-firmware/misc/changelog.html)

---

Note: when reviewing, I use imperative mood a lot: "please change this, rename that", etc. In this particular case, I want to stress again that responding to this review is _not_ part of the assignment. It is up to you if you want to continue working on the PR; if not, someone else (maybe me) will pick up later.

In any case, feel free to also add yourself to the `CONTRIBUTORS` file.

grdddj:
Thank you for a very thorough review! All those points make perfect sense.

From those suggestions, I made a second commit. The tests were deleted from repo, but I still have them in a working state locally, which may come in handy later when migrating them to pytest.

core/.changelog.d/1258.added
```diff
@@ -0,0 +1 @@
+CLI: two new firmware commands - "trezorctl firmware download" and "trezorctl firmware verify"
```
matejcik, line 1: both changelogs should be in the `python/` directory

python/src/trezorlib/cli/firmware.py
```diff
@@ -230,98 +268,54 @@ def version_str(version):
...
-        version=version,
-        fw=fw
-    )
-    validate_fingerprint_on_firmware(
-        version=version,
-        fw=fw,
-        expected_fingerprint=fingerprint,
-    )
-    click.echo("Firmware is well signed and fingerprinted.")
+    common_kwargs = {'version': version, 'fw': fw}
```
matejcik, line 300: please instead just past `version, fw` as positional arguments in all of the calls below

python/src/trezorlib/cli/firmware.py
```diff
@@ -357,52 +351,43 @@ def cli():
...
     """
-    if not filename:
-        click.echo("Please include the filename by '-f' or '--filename'.")
-        sys.exit(1)
-
-    # Deciding if to validate device integrity
-    if skip_device_check:
+    # Deciding if to take the device into account
+    if check_device:
+        client = get_default_client()
```
matejcik, line 372: this way the client won't respect the passed in `--path` parameter and possibly some other things.

instead, please use `obj.client_context()`, see `trezorctl.py::get_session` for a correct example

matejcik (changes requested):

python/src/trezorlib/cli/firmware.py
```diff
@@ -81,9 +127,57 @@ def validate_firmware(version, fw, expected_fingerprint=None):
...
-):
+    client: TrezorClient,
+    version: str = None,
+    beta: bool = False,
+    bitcoin_only: bool = False
+) -> tuple:
+    """Get the url from which to download the firmware and its expected fingerprint."""
+    f = client.features
+    model = f.model or "1"
+    bootloader_version = [f.major_version, f.minor_version, f.patch_version]
```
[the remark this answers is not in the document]
grdddj: I am not sure what are the implications here, is it only the name of the variable that should change?

Theoretically, this X.X.X version does not change depending on being in BL mode or not. When running `trezorctl get-features`, it contains the same three version numbers in both cases.

We may use `f.major_version` instead of `bootloader_version[0]` and only initialize the whole list when comparing it with `release["min_bootloader_version"]`, it is its only full usage.

grdddj:
Thanks for the feedback. I made the fixes, apart from the BL version, where I am not sure how to proceed - see the comment above.

Some changes were also done by black formatting, thank you for introducing that to me, it seems really helpful.

python/src/trezorlib/cli/firmware.py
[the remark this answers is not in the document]
matejcik: > Theoretically, this X.X.X version does not change depending on being in BL mode

This is an incorrect assumption. Bootloader has its own version. And it is important for the selection algorithm below - namely that certain firmware versions (1.7.0 and up) require certain minimum bootloader version (1.6.0). The releases.json contains `min_bootloader_version` field which specifies that.

So if the device is _not_ in bootloader mode, we need to look at the firmware version, lookup its `bootloader_version` in the json, and then use that for comparison with `min_bootloader_version`.
...alternately, I just noticed there is also a `min_firmware_version`, which seems to match the result of this lookup. So i.e., I didn't verify this, but it should be possible to write "if is_bootloader: compare version to min_bootloader_version; else: compare version to min_firmware_version"

python/src/trezorlib/cli/firmware.py
```diff
@@ -369,9 +373,11 @@ def verify(
     """
     # Deciding if to take the device into account
     if check_device:
-        client = get_default_client()
-        bootloader_onev2 = _is_bootloader_onev2(client)
-        trezor_major_version = client.features.major_version
+        # Make sure the previous session is not resumed
+        obj.session_id = None
```
matejcik, line 377: this is not appropriate here. we want to reuse the session.

python/src/trezorlib/cli/firmware.py
```diff
@@ -15,25 +15,49 @@
...
+def _print_version(version: dict) -> None:
     vstr = "Firmware version {major}.{minor}.{patch} build {build}".format(**version)
     click.echo(vstr)
 
 
-def validate_firmware(version, fw, expected_fingerprint=None):
+def _is_bootloader_onev2(client: TrezorClient) -> bool:
+    """Check if bootloader is capable of installing the Trezor One v2 firmware directly."""
+    f = client.features
+    bootloader_version = (f.major_version, f.minor_version, f.patch_version)
```
matejcik, line 45: This is no longer necessarily _bootloader_ version too.

AFAICT it doesn't affect functionality (firmware version 1.8.0 also must have bootloader 1.8.0 or better), but please rename the variable and modify docstring so that this is documented.

python/src/trezorlib/cli/firmware.py
```diff
@@ -81,11 +125,41 @@ def validate_firmware(version, fw, expected_fingerprint=None):
...
+    elif version not in ALLOWED_FIRMWARE_FORMATS[trezor_major_version]:
+        click.echo("Firmware does not match your device, aborting.")
+        sys.exit(3)
+
+
+def get_all_firmware_releases(
+    bitcoin_only: bool, beta: bool, major_version: int
+) -> list:
+    """Retrieve all the firmware releases suitable for inputted parameters"""
+    url = "https://data.trezor.io/firmware/{}/releases.json".format(major_version)
```
matejcik, line 161: new-style f-string is preferred:
`url = f".../{major_version}/releases..."`

python/src/trezorlib/cli/firmware.py
```diff
@@ -102,20 +176,53 @@ def find_best_firmware_version(
...
+    """
 
     def version_str(version):
         return ".".join(map(str, version))
 
-    want_version = requested_version
+    f = client.features
+
+    releases = get_all_firmware_releases(bitcoin_only, beta, f.major_version)
+    releases.sort(key=lambda r: r["version"], reverse=True)
```
matejcik, line 203: please return the releases already sorted

python/src/trezorlib/cli/firmware.py
```diff
@@ -102,20 +176,53 @@ def find_best_firmware_version(
...
+        want_version = [int(x) for x in version.split(".")]
+        if len(want_version) != 3:
+            click.echo("Please use the 'X.Y.Z' version format.")
+        if want_version[0] != f.major_version:
+            model = f.model or "1"
+            click.echo(
+                "Warning: Trezor {} firmware version should be {}.X.Y (requested: {})".format(
+                    model, f.major_version, version
+                )
+            )
```
matejcik, line 216:
```suggestion
            click.echo(
                f"Warning: Trezor {model} firmware version should be "
                f"{f.major_version}.X.Y (requested: {version})"
            )
```

python/src/trezorlib/cli/firmware.py
```diff
@@ -174,31 +230,261 @@ def version_str(version):
...
+    click.echo("Firmware is well signed and fingerprinted.")
+
+    if bootloader_onev2 is not None and trezor_major_version is not None:
+        validate_firmware_for_device_and_version_consistency(
+            version=version,
+            fw=fw,
+            bootloader_onev2=bootloader_onev2,
+            trezor_major_version=trezor_major_version,
+        )
+        click.echo("Device and firmware version are well compatible.")
```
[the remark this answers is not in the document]
matejcik: please change the wording as suggested

python/src/trezorlib/cli/firmware.py
```diff
@@ -166,39 +281,212 @@ def version_str(version):
...
 
-@click.command()
+
+@cli.command()
 # fmt: off
-@click.option("-f", "--filename", type=click.File("rb"))
-@click.option("-u", "--url")
-@click.option("-v", "--version")
+@click.argument("filename", type=click.File("rb"))
+@click.option("-c", "--check-device", is_flag=True, help="Validate the device compatibility")
```
matejcik, line 382: please remove "the"

python/src/trezorlib/cli/firmware.py
```diff
@@ -166,39 +281,212 @@ def version_str(version):
...
+    if not fingerprint:
+        fingerprint = fp
+    if not output:
+        output = open(_get_file_name_from_url(url), "wb")
+
+    output.write(firmware_data)
+    output.close()
+    click.echo("Firmware saved under {}.".format(output.name))
+
+    if not skip_check:
```
matejcik, line 456: validate first, write file only if validation succeeds

python/src/trezorlib/cli/trezorctl.py
```diff
@@ -76,9 +76,10 @@
     "xlm": stellar.cli,
     "xtz": tezos.cli,
     # firmware-update aliases:
-    "update-firmware": firmware.firmware_update,
-    "upgrade-firmware": firmware.firmware_update,
-    "firmware-upgrade": firmware.firmware_update,
+    "update-firmware": firmware.update,
+    "upgrade-firmware": firmware.update,
+    "firmware-upgrade": firmware.update,
+    "firmware-update": firmware.update,
```
matejcik, line 82: while you're touching this, please also add a `fw` alias for the `firmware` subcommand :)

matejcik (changes requested):
Great, this is almost ready to merge. I just have a couple of style-related or other minor requests, please apply them, then rebase on master and squash into one commit.

python/src/trezorlib/cli/firmware.py
grdddj: Sure, I also prefer them, so will change them in the whole file

python/src/trezorlib/cli/firmware.py
grdddj: At first I had it like that, but then it seemed to me it could be confusing to implicitly sort, unless the function name was explicitly named "get_all_firmware_releases_sorted" or something like that (or the variable name "releases_sorted")

python/src/trezorlib/cli/firmware.py
grdddj: Is it enough to rename the word "bootloader" with "device" both in the variable name and in docstring?

matejcik: "This is the case from bootloader version 1.8.0, and also holds for _firmware_ version 1.8.0 because that installs the appropriate bootloader."

python/src/trezorlib/cli/trezorctl.py
grdddj: Added in 2ae4

python/src/trezorlib/cli/firmware.py
grdddj: Modified in 2ae4

python/src/trezorlib/cli/firmware.py
grdddj: Fixed in 2ae4

python/src/trezorlib/cli/firmware.py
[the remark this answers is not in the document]
grdddj: Changed in 2ae4

python/src/trezorlib/cli/firmware.py
grdddj: Changed in 2ae4

python/src/trezorlib/cli/firmware.py
grdddj: Changed in 2ae4

python/src/trezorlib/cli/firmware.py
grdddj: Modified in 2ae4

python/src/trezorlib/cli/firmware.py
```diff
@@ -166,39 +281,212 @@ def version_str(version):
...
+    )
+
+    firmware_data = download_firmware_data(url)
 
-    return url, fingerprint
+    # When fingerprint and output file were not supplied, take them from SL servers
+    if not fingerprint:
+        fingerprint = fp
+    if not output:
+        output = open(_get_file_name_from_url(url), "wb")
```
matejcik, line 450: please also move this after the validation. otherwise you'd create a new file and leave it empty if validation failed

grdddj: Fixed in 015edd3

matejcik:
style checks are now failing, please run `make style`

also please regenerate `OPTIONS.rst`, i think it's drifted.
(note to self, it should be part of `make gen` probably)

grdddj:
> style checks are now failing, please run `make style`
> 
> also please regenerate `OPTIONS.rst`, i think it's drifted.
> (note to self, it should be part of `make gen` probably)

Fixed in aac03c9

grdddj:
Linking this PL with the issue - #1258

matejcik:
and we are go :rocket: thanks!

Issue #334, opened 2019-07-15
FIDO2 Support txAuthSimple extension


andrewkozlik:
The txAuthSimple extension allows for a simple form of transaction authorization. A Relying Party can specify a prompt string, which will be displayed on the Trezor. If the user confirms, then the prompt string will be signed using the credential private key.

Currently I am not aware of any site that would use this, so this is low priority.

We have to be careful about displaying arbitrary prompts on the screen. Since the display is considered trusted by the user, an attacker could misuse this to persuade the user into taking some dangerous action. So we should say something like "Do you wish to authorize the following transaction from <RPID>?". The prompt should have a different color or some other visual distinction. The user will be able to swipe up and down until they reach the end of the prompt and then confirm or decline, similar to the way the recovery seed is displayed.

andrewkozlik:
Unfortunately this feature has been removed from WebAuthn, because none of the browsers implemented it. What a shame :-(. This could have been used to securely confirm withdrawals from cryptocurrency exchanges.
https://groups.google.com/a/fidoalliance.org/g/fido-dev/c/nIA0gb3Px6c/m/ag1C9TCCCQAJ?pli=1

prusnak:
The extension was removed via https://github.com/w3c/webauthn/issues/1386

petrdvorak:
Help us get it back, @prusnak! https://github.com/w3c/webauthn/pull/2020

PR #1616, opened 2021-05-19
FIDO2: Added missing "txAuthSimple" in the getInfo

herrjemand:


prusnak:
Did we really implement `txAuthSimple`? I don't think so. 

Plus I thought it was removed from WebAuthn - see https://github.com/trezor/trezor-firmware/issues/334#issuecomment-766725741

herrjemand:
@prusnak The metadata statement says.

Now doing testing, I think it's metadata that is wrong.

prusnak:
@herrjemand Ah yes, that is the case. I added the extension to the JSON by accident.

I removed it in d358503676304377ccdf25f92829087bbb1e6622 and the latest definition is available from: https://raw.githubusercontent.com/trezor/trezor-firmware/master/core/src/apps/webauthn/metadata/trezor-ctap2.json

PR #1617, opened 2021-05-19
build: fixes for T1 support in core

mmilata:
* fix incorrect architecture for T1 which is `thumbv7m-none-eabi` but we used `thumbv7em-none-eabi`
* explicitly handle `.ARM.exidx*` sections in the linker script since otherwise the resulting section has semi-random name which is unusable for calculation of FW size included in header


prusnak (approved):
utACK for non-rust changes

matejcik (commented):
utACK for the rest

maybe we shouldn't be adding a changelog for this?

matejcik (approved):

PR #1621, opened 2021-05-20
Disable Nightly Rust

matejcik:
When we need it, we will re-enable.

Also add some build documentation.

@mmilata is there anything to be done on the nix side? Not sure if it's worth changing the image when we expect we might be going back to nightly again.

### DIFF
--- a/docs/core/build/embedded.md
+++ b/docs/core/build/embedded.md
@@ -11,7 +11,7 @@ You will also need Python dependencies for signing.
 ### Debian/Ubuntu
 
 ```sh
-sudo apt-get install scons gcc-arm-none-eabi libnewlib-arm-none-eabi
+sudo apt-get install scons gcc-arm-none-eabi libnewlib-arm-none-eabi llvm-dev libclang-dev clang
 ```
 
 ### NixOS
@@ -33,6 +33,15 @@ For other users:
 2. Follow the [install instructions](https://launchpadlibrarian.net/287100883/readme.txt)
 3. To install OpenOCD, run `brew install open-ocd`
 4. Run `make vendor build_boardloader build_bootloader build_firmware`
+
+## Rust
+
+Install the appropriate target with [`rustup`](https://rustup.rs/):
+
+```sh
+rustup target add thumbv7em-none-eabihf  # for TT
+rustup target add thumbv7m-none-eabi     # for T1
+```
 
 ## Building
 
--- a/docs/core/build/emulator.md
+++ b/docs/core/build/emulator.md
@@ -11,13 +11,13 @@ Install the required packages, depending on your operating system.
 * __Debian/Ubuntu__:
 
 ```sh
-sudo apt-get install scons libsdl2-dev libsdl2-image-dev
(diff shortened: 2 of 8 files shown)


matejcik:
please ignore the changelog failure, this should not be adding changelog entries

docs/core/build/embedded.md
```diff
@@ -34,6 +34,15 @@ For other users:
 3. To install OpenOCD, run `brew install open-ocd`
 4. Run `make vendor build_boardloader build_bootloader build_firmware`
 
+## Rust
+
+Install the appropriate target with [`rustup`](https://rustup.rs/):
+
+```sh
+rustup target add thumbv7em-none-eabihf  # for TT
+rustup target add thumbv7m-none-eabihf   # for T1
```
mmilata, line 43:
```suggestion
rustup target add thumbv7m-none-eabi     # for T1
```

prusnak:
> @mmilata is there anything to be done on the nix side?

We need to modify the `shell.nix` file.

@mmilata Is there a way how to get rid of the mozilla overlay? or better put: Is there a way how to install rustc for all 3 targets we need via nixpkgs?

docs/core/build/emulator.md
```diff
@@ -52,6 +52,17 @@ brew install scons sdl2 sdl2_image pkg-config
...
+
+You will require Rust and Cargo. The currently supported version is 1.52 stable.
+The recommended way to install both is with [`rustup`](https://rustup.rs/). If you
+are installing `rustup` for the first time, the stable toolchain will be installed
+for you automatically. Otherwise, make sure you are up to date:
+
+```sh
+rustup update
+```
+
```
mmilata, line 65: Perhaps also mention that clang/libclang is needed by [bindgen](https://rust-lang.github.io/rust-bindgen/requirements.html)?

mmilata:
I can take care of `shell.nix` modifications. The mozilla overlay can be switched to stable channel but I'll look into using the compiler directly from nixpkgs.

mmilata:
@prusnak there doesn't seem to be support for additional targets in the rustc nix expression, there's some support for crosscompilers to be used to crosscompile nixpkgs but I don't think we can use that.

There also exists different overlay which claims to be more reproducible than the one from mozilla: https://github.com/oxalica/rust-overlay

mmilata:
Pushed 7488107 which updates shell.nix to use stable rust-1.52.1 binary. Switches to https://github.com/oxalica/rust-overlay which is more reproducible and also faster than mozilla overlay. @prusnak can you please check that it works on darwin?

prusnak (approved):
Confirm I was able to initialize the nix environment on darwin as specified by the new definitions in `shell.nix`

docs/core/build/embedded.md
mmilata: Fixed in a823ddf.

docs/core/build/emulator.md
mmilata: Fixed in a823ddf.

mmilata:
Fixed documentation & rebased, will merge later if no objections.

PR #1622, opened 2021-05-20
SystemView/RTT documentation

hiviah:
No code change here, just documentation on SystemView and RTT usage added.


matejcik:
please add link to the document from `docs/index.md`
also shouldn't this go under `docs/core` ?

Issue #1619, opened 2021-05-19
Device not detected by computer after upgrade to latest 1.10.0 Intermediary firmware


bosomt:
**Describe the bug**
When you update to latest provided Intermediary FW device display will blink forever after reconnection and won't be recognized by computer till you switch it to bootloader version. 
Works fine after you downgrade bootloader to older Intermediary firmware 1.10.0
EDIT: this issue can be fixed by wiping device in bootloader mode via `trezorctl wipe-device -b  ` after that you can install standard firlmware 1.10.0

**Firmware version and revision**
1.10.0 - intermediary
Firmware fingerprint: 9c97590fd86b3c9bdcaa7421f58344ba9524fa207a2673dd22ca902449401c2d

**Desktop/smartphone setup (please complete the following information):**
 - Wallet software: trezorctl
 - OS: macOS

**To Reproduce**
1. Upgrade to new intermediary firmware
2. Reconnect device
3. Observe device
4. Try to load full firmware to device

**Screenshots**

```
trezorctl firmware_update -f trezor-inter-1.10.0.bin
WebUSB transport is disabled: dlopen(libusb-1.0.dylib, 6): image not found
Trezor One firmware with embedded v2 image (1.8.0 or later)
Firmware version 1.10.0 build 0
Signatures are valid.
Firmware fingerprint: 9c97590fd86b3c9bdcaa7421f58344ba9524fa207a2673dd22ca902449401c2d
Embedded v2 image fingerprint: e0bd7ce4643f9c98c376a827022a001cbac8894ec7c88593525e84e19d7b69d4
Extracting embedded firmware image.
Please confirm the action on your Trezor device
testing@testings-MacBook-Air Downloads % trezorctl get_features                              
WebUSB transport is disabled: dlopen(libusb-1.0.dylib, 6): image not found
Failed to find a Trezor device.

```
https://user-images.githubusercontent.com/31506317/118867022-ec35ca80-b8e2-11eb-94b7-c011940f7c67.mp4

bosomt:
same behavior for 1.9.4/1.8.0  device /blinking + not found/

```
trezorctl get_features
WebUSB transport is disabled: dlopen(libusb-1.0.dylib, 6): image not found
Features (166 bytes) {
    bootloader_hash: 32 bytes 0xf7fa165be6d780f3e1af00abc07df8b3076bcdad72d70da22a63d8896b6391d8,
    capabilities: [
        Bitcoin (1),
[...]

PR #1623, opened 2021-05-20
fix(legacy): set STAY_IN_BOOTLOADER_FLAG before the reboot of intermediate_fw

prusnak:
Fixes https://github.com/trezor/trezor-firmware/issues/1619

### DIFF
--- a/legacy/intermediate_fw/trezor.c
+++ b/legacy/intermediate_fw/trezor.c
@@ -31,6 +31,7 @@
 #include "oled.h"
 #include "rng.h"
 #include "setup.h"
+#include "supervise.h"
 #include "timer.h"
 #include "util.h"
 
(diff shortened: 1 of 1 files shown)


hiviah (approved):

prusnak:
It is up to discussion whether to replace `erase_sector(FLASH_CODE_SECTOR_FIRST, FLASH_CR_PROGRAM_X32);` with just flash_program that nukes (set to 0) the first 4 bytes of the sector.

andrewkozlik (approved):
utACK. This should solve the blinking behavior which occurs when the storage is initialized, but in all likelihood it won't solve the issue with the black screen which occurs when the storage is not initialized.

andrewkozlik:
> It is up to discussion whether to replace `erase_sector(FLASH_CODE_SECTOR_FIRST, FLASH_CR_PROGRAM_X32);` with just flash_program that nukes (set to 0) the first 4 bytes of the sector.

As explained here https://github.com/trezor/trezor-firmware/issues/1619#issuecomment-845261713, I am no longer sure that this will fix the issue with the black screen, but it still might be worth a shot, because right now I have no other clue what might be causing the issue.

Issue #1620, opened 2021-05-19
Inconsistent revision of FW between TT and T1


tomasklim:
**Describe the bug**
On T1 the revision of firmware is in UTF-8/ASCII. 
However, on TT the revision of firmware consists from raw bytes and after hex encoding to UTF-8 I get the format I have on T1 from beginning. 

**Firmware version and revision**
T1: `1.10.0` & `4424ece1ccb7fc0d6cad00ff840fac287a34f07`
TT: `2.3.6` & `623139636266363763`

**Expected behavior**
I would expect to have these revisions united not only by encoding but also by its length.

prusnak:
> TT: `2.3.6` & `623139636266363763`

This is correct. Trezor returns the revision as bytes, which got encoded in javascript to hex string. `623139636266363763` => `b"b19cbf67c"` => valid git revision for firmare 2.3.6

> T1: 1.10.0 & 4424ece1ccb7fc0d6cad00ff840fac287a34f07

This however seems totaly broken.

matejcik:
this is probably a copy-paste error, the right revision is **f**4424ece1ccb7fc0d6cad00ff840fac287a34f07 -> f4424ece1ccb7fc0d6cad00ff840fac287a34f07

Anyway this is a mess in any case. The `revision` field is `bytes`. T1 sends the full hash of the commit, while TT sends _string_ output of `git describe`.

We should:
* either flip the field to `string`, but that seems like a backwards compatibility can of worms,
* or send bytes from core as well, even if it's just a byte encoding of the short version

prusnak:
I am all for consistency here.

Using string has the advantage of being able to express stuff such as `b19cbf67c-dirty`, but I am not sure we are really in desperate need of this.

So maybe sending the revision as binary bytes makes sense the most and we can decide whether we want to send the whole thing (20 bytes) or just a prefix to save some space.

tsusanka:
As discussed we want to change TT behavior to T1 behavior.

PR #1624, opened 2021-05-21
fix(core): unify Features.revision reporting with legacy

prusnak:
Fixes https://github.com/trezor/trezor-firmware/issues/1620

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -42,8 +42,7 @@ FIRMWARE_P1_MAXSIZE = 786432
 FIRMWARE_P2_MAXSIZE = 917504
 FIRMWARE_MAXSIZE    = 1703936
 
-GITREV=$(shell git describe --always --dirty | tr '-' '_')
-CFLAGS += -DGITREV=$(GITREV)
+CFLAGS += -DSCM_REVISION='\"$(shell git rev-parse HEAD | sed 's:\(..\):\\x\1:g')\"'
 
 TESTPATH = $(CURDIR)/../tests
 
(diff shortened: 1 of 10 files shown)


core/Makefile
```diff
@@ -42,7 +42,7 @@ FIRMWARE_P1_MAXSIZE = 786432
 FIRMWARE_P2_MAXSIZE = 917504
 FIRMWARE_MAXSIZE    = 1703936
 
-CFLAGS += -DSCM_REVISION='"$(shell git rev-parse HEAD | sed 's:\(..\):\\x\1:g')"'
```
prusnak: This was taken from https://github.com/trezor/trezor-firmware/blob/3976ef7261add5e513699d82412d6efe242a5fe1/legacy/firmware/Makefile#L131

Do we need to fix it there as well?

matejcik: in legacy makefile is using CFLAGS implicitly

in core, they're explicitly passed to scons via `CFLAGS="$(CFLAGS)"` which is where the `"`s get mangled

bosomt:
**QA OK**

trezorctl get_features


```
    major_version: 1,
    minor_version: 10,
    model: '1',
    needs_backup: False,
    no_backup: False,
    passphrase_protection: False,
    patch_version: 2,
    pin_protection: False,
    revision: 20 bytes 0x01c1ae426fe911619f0312fa417ecdd7e102848c,

    major_version: 2,
    minor_version: 4,
    model: 'T',
    needs_backup: False,
    no_backup: False,
    passphrase_always_on_device: False,
    passphrase_protection: True,
    patch_version: 1,
    pin_protection: False,
    recovery_mode: False,
    revision: 20 bytes 0x01c1ae426fe911619f0312fa417ecdd7e102848c,

```

PR #1625, opened 2021-05-21
build(core): skip rust library for now

mmilata:
Disable linking in of `trezor_lib.a` into the final image, for this release.

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -3,15 +3,11 @@
  }:
 
 let
-  rustOverlay = import (builtins.fetchTarball {
-    url = "https://github.com/oxalica/rust-overlay/archive/4d5d8e4288a8e0efd074e56a448e450b0f8df975.tar.gz";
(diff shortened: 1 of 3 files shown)


tsusanka (approved):
LGTM, but I did not watch Rust implementation that closely to be 100% sure it is all it needs.

matejcik:
LGTM as well.

noting that **this should not be applied to master** but to the release branch, @tsusanka.

please edit the target branch when appropriate.

tsusanka:
We have discussed with @mmilata that he'll try to do the release process (Monday's fine). Let's not forget to add `[RELEASE_ONLY]` to the commit message btw so the release commit check passes.

prusnak:
@mmilata Do I understand it correctly, we won't be needing rust at all for this release? If yes, can we please also remove the `rustOverlay` and friends from the `shell.nix` in the release branch?

mmilata:
Yes, we've decided the rust protobuf code is not polished enough for this release and without it the FFI in master is not used. Will remove it completely from the build process for the release.

tsusanka:
@mmilata I guess we can close this since we have it (correctly) in the release branch?

PR #1626, opened 2021-05-21
fix(legacy/intermediate_fw): don't erase the whole firmware sector

prusnak:
Erase just the first firmware word

### DIFF
--- a/legacy/intermediate_fw/trezor.c
+++ b/legacy/intermediate_fw/trezor.c
@@ -139,9 +139,14 @@ int main(void) {
   }
 
   if (sectrue == storage_initialized) {
-    // don't erase
+    // Storage probably contains a seed so leave the firmware intact.
+    // Invalidating it would cause the bootloader to wipe the storage before
+    // installing the target firmware. User will need to confirm installation.
     reboot_device();
   } else {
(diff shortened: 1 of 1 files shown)


legacy/intermediate_fw/trezor.c
```diff
@@ -143,7 +142,7 @@ int main(void) {
     // don't erase
     reboot_device();
   } else {
```
andrewkozlik, line 144:
```suggestion
    // Storage probably contains a seed so leave the firmware intact.
    // Invalidating it would cause the bootloader to wipe the storage before
    // installing the target firmware. User will need to confirm installation.
    reboot_device();
  } else {
    // New device. Invalidate the intermediate firmware so that after reboot
    // the bootloader will install the target firmware without asking for user
    // confirmation.
```

andrewkozlik (approved):
LGTM. Note that since we are no longer erasing the firmware, we don't need the `__attribute__((noinline, section(".data")))` and we don't need to define our own `_flash_program_word()` function. However, I don't see a problem with leaving it as is.

hiviah:
Just for documentation purposes, that strange attribute was to keep code linked in RAM and not flash when it was being erased. Dirty hack, but sometimes you have to fight the compiler/linker.

Issue #1627, opened 2021-05-21
Support safety_checks parameter in ApplySettings on T1


andrewkozlik:
In 568da98199 which was released with T1 firmware v1.9.2 we introduced a safety check that doesn't allow spending LTC from a Bitcoin path, because LTC doesn't have strong replay protection via SIGHASH_FORKID. In order to claim coins sent to a wrong path users had to downgrade to firmware 1.9.1 or 1.9.0 and then use Electrum to recover the coins. Now that we have a new bootloader with firmware 1.10.0, it is no longer possible to downgrade to earlier firmware versions, i.e. older firmware will not run at all. So users have no way of claiming these coins other than buying a TT and entering the seed there or entering it directly into Electrum, which are both pretty harsh options. The solution is to allow users to disable safety checks in T1 like they can on TT.

prusnak:
Quick fix would be replacing the following

https://github.com/trezor/trezor-firmware/blob/4f1274c9c4eb3c74feb8f9605bace2285b3199c3/legacy/firmware/signing.c#L687-L690

with the code taken from

https://github.com/trezor/trezor-firmware/blob/4f1274c9c4eb3c74feb8f9605bace2285b3199c3/legacy/firmware/fsm_msg_coin.h#L186-L196

but adapted to `return false;` instead of `return;`

Not sure if we even want/need to add the whole `safety_checks` thing. But surely we can if we want ...

andrewkozlik:
> Quick fix would be [...]

Yes, that's a very good point that we can just behave in T1 as if safety_checks == PromptAlways. But I think I already have it implemented:

![image](https://user-images.githubusercontent.com/42678794/119170185-304fd900-ba63-11eb-8198-4c9d99fe946c.png) ![image](https://user-images.githubusercontent.com/42678794/119170126-1dd59f80-ba63-11eb-9dc7-306b1a101011.png)

I could also implement the `PromptAlways` setting fairly quickly, but I am not sure whether it's desired.

tsusanka:
We should also not forget reporting the safety_checks value in Features. Suite depends on it (see my comment [here](https://github.com/trezor/trezor-suite/issues/3405#issuecomment-846100197)).

prusnak:
> I could also implement the `PromptAlways` setting fairly quickly, but I am not sure whether it's desired.

[...]

PR #1628, opened 2021-05-21
Safety checks setting in T1

andrewkozlik:
Fixes https://github.com/trezor/trezor-firmware/issues/1627.

I implemented only the settings `Strict` and `PromptTemporarily`, because `PromptAlways` is not available in Suite or trezorctl, so it doesn't seem like an efficient use of time to implement it now. Note that this PR does not implement any safety checks. It only implements the setting to disable safety checks. So large fees etc. are still treated leniently in T1 as opposed to TT.

There are some differences in how TT and T1 handle `GetAddress` when it's **_not_** shown on the display. T1 will always return the address even if the path is non-standard. TT will not return the address if the path is non-standard, unless safety-checks are disabled. Also T1 checks that the script type matches the purpose field whereas TT does not. These discrepancies are unwelcome, but syncing the behavior is not a priority at the moment and I don't think there is anything wrong with the way T1 handles these cases, so I left it as is. If anyone feels otherwise, then create a separate issue. I am writing this mainly to explain some of the changes in the device tests.


prusnak (approved):
utACK changes in legacy code

andrewkozlik:
@matejcik do you want to have a look at the device tests or shall we merge this?

tests/device_tests/test_msg_getaddress.py
```diff
@@ -269,15 +269,14 @@ def test_unknown_path_tt(client):
 def test_unknown_path_t1(client):
```
mmilata, line 269: It should be possible to drop this testcase and use `test_unknown_path_tt` for both models except that `get_address` behaves differently with strict setting and `show_display=False`. In this case TT validates the path but T1 doesn't. Or we can cheat and use `show_display=True` in the call below `# account number is too high`.

mmilata (approved):
LGTM, got some suggestions wrt TT compatibility but not sure how feasible they are.

tests/device_tests/test_msg_getaddress.py
mmilata: Same with `test_invalid_path`.

prusnak (approved):

legacy/firmware/signing.c
```diff
@@ -686,7 +687,18 @@ bool compile_input_script_sig(TxInputType *tinput) {
...
                              tinput->address_n, false)) {
-    return false;
+    if (config_getSafetyCheckLevel() == SafetyCheckLevel_Strict) {
+      return false;
+    }
+
+    layoutDialogSwipe(&bmp_icon_warning, _("Abort"), _("Continue"), NULL,
+                      _("Wrong address path"), _("for selected coin."), NULL,
+                      _("Continue at your"), _("own risk!"), NULL);
+    if (!protectButton(ButtonRequestType_ButtonRequest_Other, false)) {
```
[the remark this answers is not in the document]
andrewkozlik: Good point. Done in 3fcaebd.

tests/device_tests/test_msg_getaddress.py
andrewkozlik: Done in f97c6e2e9c6e4d7900dae78ca58236d0990482a7. I changed to `show_display=True` so that we can unify the tests. In terms of security `show_display=True` if the more important case anyway.

mmilata (approved):

PR #1629, opened 2021-05-25
ci: improve changelog check, longer timeout for T1 hw test

mmilata:
`check_changelog.sh`:
- ignore `master` branch
- treat `secfix/` same as `release/` branches
- allow skiping check with `[no changelog]` in commit message

### DIFF
--- a/ci/check_changelog.sh
+++ b/ci/check_changelog.sh
@@ -4,13 +4,27 @@ base_branch=master
 fail=0
 subdirs="core python legacy/firmware legacy/bootloader"
 
+changed_files=$(mktemp)
+trap 'rm -- $changed_files' EXIT
+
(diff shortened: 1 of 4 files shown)


mmilata:
Per @matejcik's request I've updated the script so that `[no changelog]` only skips checking the commit it is in, as opposed to turning off the check entirely for the branch.

The check continues to work on whole branches, e.g. it is legal to have three commits modifying files in `core/` but only one of the commits adding a file to `core/.changelog.d`. I think this is quite common scenario and requiring the change log to be in the same commit as the change seems to be too restrictive.

matejcik (approved):
utACK

tsusanka (approved):
LGTM

matejcik:
@mmilata please squash&merge

PR #1632, opened 2021-05-26
feat(core): avoid restarting session for select messages (fixes #1631)

### LINKED ISSUE #1631: Device repeatedly flashes after reconnecting during Shamir backup and blocks seed recovery
**Describe the bug**
When user disconnects the device safely (i.e. after completing at least one share) during Shamir backup, device then repeatedly flashes and makes it unable to complete the seed recovery. 
This repeated blinking is continuous in Wallet and user cannot complete the recovery.
In case of Suite, bliking will stop when Suite fully loads. 
Every time the device flashes, it returns the user back to the starting screen (the one that informs the user how many shares are needed for seed recovery).

Firmware TT 2.4.0 d80d5456

 - Wallet software: Trezor Wallet 249daf21e2f8380c89c3e4b46c976851c50afec0 
                              Trezor Suite 21.5.1 4d09d88476dab2e6b2fbfb833b749e9ac62251c2
 - OS: nixos 21.03
 - Browser Chromium Version 88.0.4324.182 (Official Build) (64-bit)

**To Reproduce**
1. Wipe device
2. Enter first share of shamir seed
3. Disconnect and reconnect device
4. Try to unsuccessfully complete the seed recovery

**Expected behavior**
User will be able to complete the Shamir recovery. 

**Videos**
[here](https://drive.google.com/drive/folders/1xprv89jeds5Rd2cUn3U-AJo6Rnz7SA79?usp=sharing)

matejcik:
Introduces `wire.AVOID_RESTARTING_FOR` and the recovery workflow will set it whenever it runs.

This also allows us to e.g. avoid restarting for GetPublicKey and thus make the old Wallet significantly faster, at the cost of some memory fragmentation. I am not doing that now, but it would be easy if we actually needed it.

### DIFF
--- a/core/src/apps/management/recovery_device/homescreen.py
+++ b/core/src/apps/management/recovery_device/homescreen.py
@@ -6,7 +6,7 @@ from trezor import strings, utils, wire, workflow
 from trezor.crypto import slip39
 from trezor.crypto.hashlib import sha256
 from trezor.errors import MnemonicError
-from trezor.messages import BackupType
+from trezor.messages import BackupType, MessageType
 from trezor.messages.Success import Success
 from trezor.ui.layouts import show_success
 
@@ -31,6 +31,7 @@ async def recovery_homescreen() -> None:
 
 
 async def recovery_process(ctx: wire.GenericContext) -> Success:
+    wire.AVOID_RESTARTING_FOR = (MessageType.Initialize, MessageType.GetFeatures)
     try:
         return await _continue_recovery_process(ctx)
     except recover.RecoveryAborted:
--- a/core/src/session.py
+++ b/core/src/session.py
@@ -1,4 +1,4 @@
-from trezor import loop, utils, wire, workflow
+from trezor import log, loop, utils, wire, workflow
 
 import apps.base
 import usb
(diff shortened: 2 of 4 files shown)


matejcik:
no changelog because this is not a difference from the previously released version

core/src/apps/management/recovery_device/homescreen.py
```diff
@@ -31,6 +31,7 @@ async def recovery_homescreen() -> None:
 
 
 async def recovery_process(ctx: wire.GenericContext) -> Success:
+    wire.AVOID_RESTARTING_FOR = (MessageType.Initialize, MessageType.GetFeatures)
```
mmilata, line 34: When recovery finishes, `wire` is unimported and the value is initialized to () next session, right?

mmilata (approved):
Verified that the change resolves the screen blinking issue. Solution looks reasonable & test is cool.

core/src/apps/management/recovery_device/homescreen.py
matejcik: exactly.

matejcik:
Waiting for re-test from @wendys-cats before merging. Afterwards this will also need to be cherry-picked to the release branch.

PR #1633, opened 2021-05-26
Fix bitcoin-only strings check, make legacy conform

mmilata:
I propose merging this after *core* is made conformant, otherwise the `core fw btconly build` is gonna be red in every CI pipeline run.

Regarding legacy ISTM that the check should stay in the `BITCOIN_ONLY`, so only the error messages differ. Perhaps we can use the generic one even in full firmware?

### DIFF
--- a/legacy/firmware/signing.c
+++ b/legacy/firmware/signing.c
@@ -1070,9 +1070,14 @@ static bool signing_validate_output(TxOutputType *txoutput) {
 
 static bool signing_validate_bin_output(TxOutputBinType *tx_bin_output) {
   if (!coin->decred && tx_bin_output->has_decred_script_version) {
+#if BITCOIN_ONLY
+    fsm_sendFailure(FailureType_Failure_DataError,
+                    _("Unsupported output type."));
+#else
     fsm_sendFailure(
         FailureType_Failure_DataError,
         _("Decred details provided but Decred coin not specified."));
+#endif
     signing_abort();
     return false;
   }
--- a/tools/check-bitcoin-only
+++ b/tools/check-bitcoin-only
@@ -2,11 +2,17 @@
 RETURN=0
 
 # dump all coins except the first 3 (Bitcoin, Testnet, Regtest)
-./common/tools/cointool.py dump -l -p -t | grep '"name"' | cut -d '"' -f 4 | tail -n +4 | while read altcoin; do
+ALTCOINS=$(./common/tools/cointool.py dump -l -p -t | grep '"name"' | cut -d '"' -f 4 | tail -n +4)
+# split on newlines only
+OLDIFS=$IFS
+IFS="
+"
+for altcoin in $ALTCOINS; do
     # echo :"$altcoin":
-    if strings $1 | grep "$altcoin" | grep -v TEXT_MARGIN_LEFT | grep -v _MIN_MNEMONIC_LENGTH_WORD ; then
+    if strings "$1" | grep "$altcoin" | grep -v TEXT_MARGIN_LEFT | grep -v _MIN_MNEMONIC_LENGTH_WORD ; then
         RETURN=1
     fi
 done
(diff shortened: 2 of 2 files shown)


matejcik:
> Regarding legacy ISTM that the check should stay in the `BITCOIN_ONLY`, so only the error messages differ. Perhaps we can use the generic one even in full firmware?

what would be the point of the check in BITCOIN_ONLY ?

I believe bitcoin-only firmware should ignore the existence of the field (iow, treat its value as if it were an unknown field), and protobuf messages containing unknown fields are valid (their values are ignored when decoding)

mmilata:
It wasn't immediately clear to me what allowing these additional valid inputs would do in the firmware. After looking at it a bit more I'm now quite sure the field is not used anywhere except [1](https://github.com/trezor/trezor-firmware/blob/legacy/v1.10.0/legacy/firmware/transaction.c#L507) and [2](https://github.com/trezor/trezor-firmware/blob/legacy/v1.10.0/legacy/firmware/transaction.c#L755) both of which check that we're processing decred tx in some way. Fixed in 2b5cd0a4c35ff198da3576e41ab8947696599d72.

matejcik (approved):
utACK

matejcik:
ready to merge code-wise but depends on #1654

mmilata:
Added cff1d5e that mostly fixes altcoins leaking into core by moving the `include_in_bitcoin_only` option from protobuf message level to file level where it is easier to mark all such messages/enums. Hopefully the `import` statements added to .proto defs are not a problem.

"Mostly fixes" means that there are still mentions of Decred. I'm tempted to allow it as an exception since it seems to be tightly bound to the bitcoin logic and separating it doesn't seem worthwhile.
```
Decred details provided but Decred coin not specified.
DecredStakingSpendType
)C#trezor.enums.DecredStakingSpendType
.Q&trezor/enums/DecredStakingSpendType.py
UiConfirmDecredSSTXSubmission
trezor/enums/DecredStakingSpendType.py
```

common/protob/pb2py
```diff
@@ -65,18 +65,24 @@ INT_TYPES = (
...
+LengthDelimited = c.Struct(
+    "len" / c.VarInt,
+    "bytes" / c.Bytes(c.this.len),
+)
+
 ListOfSimpleValues = c.GreedyRange(
     c.Struct(
         "key" / c.VarInt,
-        "value" / c.VarInt,
+        "value" / c.Switch(c.this.key & 0b111, {0: c.VarInt, 2: LengthDelimited}),
```
mmilata, line 76: Needed to not choke on the `java_package` and `java_outer_classname` file options.

common/protob/messages-bootloader.proto
```diff
@@ -5,6 +5,10 @@ package hw.trezor.messages.bootloader;
 option java_package = "com.satoshilabs.trezor.lib.protobuf";
 option java_outer_classname = "TrezorMessageBootloader";
 
+option (include_in_bitcoin_only) = true;
```
matejcik, line 8: is this necessary? bootloader is not affected by BITCOIN_ONLY, no?

matejcik (approved):
nicely done.

I think keeping Decred is OK. We _could_ export the Decred types into a separate `messages-bitcoin-decred.proto` file, and wrap mentions in `if not BITCOIN_ONLY` , but it's probably not worth the effort.

I'm surprised that the _field names_ (`decred_tree`, `decred_staking_spend_type`) don't come up in the strings check.

Do you know why?

mmilata:
> I'm surprised that the field names (decred_tree, decred_staking_spend_type) don't come up in the strings check.

Very good point. Turns out the check was case-sensitive and the altcoin names are capitalized. In c465f21 I've added a whitelist which notably includes `decred` and `omni` due to the nontrivial integration into bitcoin code, and `ripple` and `dash` because they are BIP-39 words.

PR #1219, opened 2020-08-24
Fix undefined behavior in shamir unbitslice()

andrewkozlik:
Fixes https://github.com/trezor/trezor-firmware/issues/1151

PR #1637, opened 2021-05-31
crypto: improve fuzz testing code, harnesses, documentation and scripts

invd:
This PR extends the self-contained fuzzer harness collection for trezor-crypto. For some background, see the previous #1090 PR.

The added code is stand-alone and adds no functional changes to the library itself. The fuzzer improvements have been relevant in #1219, c5f26beeeae315faf8e4be4476cf92949048101c and the SLIP39 find() topic.

### DIFF
--- a/crypto/fuzzer/fuzzer.c
+++ b/crypto/fuzzer/fuzzer.c
@@ -20,16 +20,25 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
+#include <assert.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
 // necessary for the target functions
+#include "aes/aes.h"
 #include "bignum.h"
 #include "ecdsa.h"
 #include "hasher.h"
+#include "nist256p1.h"
 #include "rand.h"
+#include "secp256k1.h"
+
+#include "ed25519-donna/ed25519-donna.h"
+#include "ed25519-donna/ed25519.h"
+#include "nem.h"
+#include "shamir.h"
 
 /* fuzzer input data handling */
 const uint8_t *fuzzer_ptr;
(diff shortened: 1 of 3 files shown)


crypto/fuzzer/fuzzer.c
```diff
@@ -358,6 +372,339 @@ int fuzz_nem_validate_address() {
...
+  memcpy(&major, fuzzer_input(sizeof(uint32_t)), sizeof(uint32_t));
+  memcpy(&minor, fuzzer_input(sizeof(uint32_t)), sizeof(uint32_t));
+
+  xmr_get_subaddress_secret_key(output, major, minor, m);
+
+  return 0;
+}
+
+extern void xmr_derive_private_key(bignum256modm s, const ge25519 *deriv,
+                                   uint32_t idx, const bignum256modm base);
```
andrewkozlik, line 421: Why are there so many external functions declared in the code? Usually external functions are declared by including the appropriate header file. Declaring them this way would make sense only if they were not in a header file, i.e. they are considered internal to the c-file, but then they would probably be static anyway and even if they weren't static, it would be questionable if such functions should be fuzzed.

andrewkozlik (approved):
utACK

crypto/fuzzer/fuzzer.c
invd: This is a valid question. My original rationale here was that it helps to have an explicit reference to which function is fuzzed including the function parameter definition. Using them as part of the `extern` definition made the reference functional, vs. just having a comment.

However, I agree that this notation is not very elegant as the number of target functions grows, and developers with a reasonably advanced IDE have automated access to this information anyway. So I've changed this to regular `#include` definitions already in the next iteration of this fuzzer collection.

@andrewkozlik : I'll update this PR. Thanks for the review so far.

invd: @andrewkozlik : I've changed the declaration style via ccac25e, so this PR is ready for review again.

matejcik:
@andrewkozlik please merge this if you think it's mergeable, or tag one of the other reviewers if you think they should look at it

PR #1641, opened 2021-06-02
Bump versions

mmilata:
Bumping minor version ok? `FIX_VERSION_...` is the compatible storage and doesn't need to be touched, right?


tsusanka:
LGTM!

PR #1642, opened 2021-06-02
Avoid accidental build with broken stack protector, especially for bundled bootloader.bin

hiviah:
Per https://blog.inhq.net/posts/faulty-stack-canary-arm-systems/ ARM gcc versions 9.2.1-10.2 have broken stack smash protector. Unfortunately these are still included in distros.

The main purpose of the PR is to prevent build of bootloader.bin with broken gcc which gets bundled with firmware since that is done usually manually and not in CI.

### DIFF
--- a/core/embed/trezorhal/compiler_traits.h
+++ b/core/embed/trezorhal/compiler_traits.h
@@ -1,0 +1,38 @@
+/*
+ * This file is part of the Trezor project, https://trezor.io/
+ *
+ * Copyright (c) SatoshiLabs
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef __COMPILER_TRAITS_H__
+#define __COMPILER_TRAITS_H__
+
+/*
+ * Avoid accidental build with gcc versions having broken stack protector 9.2.1 - 10.2
+ */
+#if defined(__GNUC__) && !defined(__llvm__)
+
+#define GCC_VERSION (__GNUC__ * 10000 \
+                     + __GNUC_MINOR__ * 100 \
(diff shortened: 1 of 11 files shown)


invd:
As discussed directly, I think this is a decent approach.

It is likely that there will be fixed backports of GCC for ARM without the stack canary compiler issue some time in the future, at which point the checks may need some adjusting, but I think this is acceptable.

For the record, to my knowledge only  the original `10.2` release (so `10.2.0`) is affected. `10.2.1` is the first fixed version.
AFAIK the current check is correct, I'm primarily mentioning this since `10.2` can also be misunderstood as the whole `10.2.x` series.

prusnak:
> I'm primarily mentioning this since `10.2` can also be misunderstood as the whole `10.2.x` series.

Addressed in 240891e9e2ebfb8160e3efd1c52f5869aed2827b

Thanks for spotting this.

matejcik:
what is the reason this is a draft? @hiviah @prusnak

core/embed/trezorhal/compiler_traits.h
```diff
@@ -0,0 +1,40 @@
...
+#if defined(__GNUC__) && !defined(__llvm__)
+
+#define GCC_VERSION \
+  (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
+
+#if GCC_VERSION >= 90201 && GCC_VERSION <= 100200
+#pragma message \
+    "Only remove this GCC check if you are sure your compiler is patched or not used for production."
+#error \
+    "ARM GCC versions 9.2.1 - 10.2 have broken stack smash protector, aborting build."
```
prusnak, line 36:
```suggestion
    "ARM GCC versions 9.2.1 - 10.2.0 have broken stack smash protector, aborting build."
```

legacy/compiler_traits.h
```diff
@@ -0,0 +1,40 @@
...
+#if defined(__GNUC__) && !defined(__llvm__)
+
+#define GCC_VERSION \
+  (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
+
+#if GCC_VERSION >= 90201 && GCC_VERSION <= 100200
+#pragma message \
+    "Only remove this GCC check if you are sure your compiler is patched or not used for production."
+#error \
+    "ARM GCC versions 9.2.1 - 10.2 have broken stack smash protector, aborting build."
```
prusnak, line 36:
```suggestion
    "ARM GCC versions 9.2.1 - 10.2.0 have broken stack smash protector, aborting build."
```

prusnak (changes requested):
I just noticed we changed the comment, but not the error messages itself. Left 2 suggestions. ACK once these 2 have been addressed.

prusnak (approved):

PR #1646, opened 2021-06-03
Extra BIP32 test (new test vector #4)

catenocrypt:
An issue was found with some BIP32 implementation, causing incorrect keys with certain seeds (starting with zeroes), see
  https://blog.polychainlabs.com/bitcoin,/bip32,/bip39,/kdf/2021/05/17/inconsistent-bip32-derivations.html

Trezor implementation is not affected.

Nonetheless the newly added test vector can be included as an extra test (BIP32 test vector #4).
Test vector 4 is being added to the BIP32 BIP here:
  https://github.com/bitcoin/bips/pull/1030

### DIFF
--- a/crypto/tests/test_check.c
+++ b/crypto/tests/test_check.c
@@ -1898,6 +1898,32 @@ START_TEST(test_bip32_vector_4) {
   memcpy(&node3, &node, sizeof(HDNode));
   memzero(&node3.private_key, 32);
   ck_assert_mem_eq(&node2, &node3, sizeof(HDNode));
+
+  // [Chain m/0'/1']
+  fingerprint = hdnode_fingerprint(&node);
+  r = hdnode_private_ckd_prime(&node, 1);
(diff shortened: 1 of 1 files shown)


prusnak:
For the reference, suggested test vector is not yet merged into the BIP32, but it will be a matter of days before it does. It got several high-profile ACKs already - see https://github.com/bitcoin/bips/pull/1030

prusnak:
@catenocrypt can you please add the test for path m/0<sub>H</sub>/1<sub>H</sub> as well? It is included in the test vector, but not included in your code.

catenocrypt:
> @catenocrypt can you please add the test for path m/0H/1H as well? It is included in the test vector, but not included in your code.

I have included this too now.  The existing vector 1--3 do not include all paths, that's why I have not included all.

prusnak (approved):

andrewkozlik (approved):

prusnak:
Thanks!

PR #1648, opened 2021-06-03
fix(core): spaces in Shamir info texts 

### LINKED ISSUE #1643: Spaces lost in Shamir info texts
**Describe the bug**
See screenshots. Looks like we lost all spaces in Shamir info texts.

major_version: 2,
    minor_version: 4,
    model: 'T',
    needs_backup: False,
    no_backup: False,
    passphrase_always_on_device: False,
    passphrase_protection: True,
    patch_version: 0,
    pin_protection: False,
    recovery_mode: False,
    revision: 9 bytes b'3204fd682',
    sd_card_present: False,
    sd_protection: False,
    unfinished_backup: True,
    unlocked: True,
    vendor: 'trezor.io',
    wipe_code_protection: False,

**Info:**
 - Suite version: web 21.5.1 (4d09d88476dab2e6b2fbfb833b749e9ac62251c2)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:88.0) Gecko/20100101 Firefox/88.0
 - OS: MacIntel
 - Screen: 1824x1140
 - Device: model T 2.4.0 Bitcoin only

**To Reproduce**
1. Start Shamir backup
2. Click on INFO button during backup
3. Observe

**Screenshots**
obrazek

obrazek

mmilata:
Fixes #1643.

adv00000000adv00000001adv00000002emu00000003emu00000004

### DIFF
--- a/core/src/trezor/ui/components/tt/info.py
+++ b/core/src/trezor/ui/components/tt/info.py
@@ -37,7 +37,7 @@ class InfoConfirm(ui.Layout):
         style: InfoConfirmStyleType = DEFAULT_STYLE,
     ) -> None:
         super().__init__()
-        self.text = text.split()
+        self.text = [text]
         self.style = style
         panel_area = ui.grid(0, n_x=1, n_y=1)
         self.panel_area = panel_area
@@ -66,7 +66,7 @@ class InfoConfirm(ui.Layout):
                 max_lines=6,
                 offset_y=y + TEXT_LINE_HEIGHT,
                 offset_x=x + TEXT_MARGIN_LEFT - ui.VIEWX,
-                line_width=w,
+                line_width=w - TEXT_MARGIN_LEFT,
                 fg=fg_color,
                 bg=bg_color,
             )


matejcik (approved):
LGTM

out of curiosity, wouldn't the automatic linebreaker do the right thing?

mmilata:
@matejcik with `break_lines=True` and tweaked `line_width` it ends up like this:

emu00000000

matejcik:
do you mean `break_words`? we want that False of course

mmilata:
It's actually the same regardless of `break_words` value, IIRC because of how we treat "words".

Btw the dashes where words are broken are not visible because they're similar color as background.

core/src/trezor/ui/components/tt/info.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         style: InfoConfirmStyleType = DEFAULT_STYLE,
     ) -> None:
         super().__init__()
-        self.text = text.split()
+        self.text = text
```
matejcik, line 40:
```suggestion
        self.text = [text]
```

matejcik: very subtle but it's a bug! `render_text` expects an iterable of strings. if we pass the original text, the iterable is `["h", "e", "l"....]`, instead of `["hello"]` which we want

with this fixed and bounds tweaked, the automatic linebreaker does a decent job

image

matejcik (changes requested):
see suggestion

matejcik:
FWIW the output with automatic linebreaker _should_ be identical to the previous code, because calling `text.split()` effectively triggered the same line-fitting algorithm

please verify that and if it confirms, leave out the newline changes

core/src/trezor/ui/components/tt/info.py
mmilata: oof, good catch, i feel retarded ... should be fixed in 38a48e2b374c0ff8461160f76c3c31cbad9f7741

mmilata:
Yep, looks the same.

emu00000000emu00000001emu00000002

bas00000000bas00000001

matejcik (approved):
utACK

PR #1649, opened 2021-06-03
build(core): improve GITREV format

mmilata:
Currently the git revision that we store in the firmware is something like `v0-broken0-2849-g767e7b8e8-dirty` because the most recent tag on master is `v0_broken0` from 2019. This change whitelists `core/...` tags which we may be present on release branches, on master the revision will fall back to `767e7b8e8-dirty`.

As the revision gets included in the firmware, building from the same commit in a repository without tags produces image with different checksum than one built in repo with these tags.


mmilata:
Ah, cool, didn't see that one.

PR #1652, opened 2021-06-04
docs(core+legacy): update changelogs to new format

prusnak:


matejcik:
should this also include towncrier configs and extension of the changelog checker?

prusnak:
> should this also include towncrier configs and extension of the changelog checker?

I think that would be nice, but I am not capable of adding it. So let's add it only if you have spare cycles to add an extra commit with this.

mmilata:
Added towncrier configs, it's a bit complicated but hopefully we won't add new changelogs often.

matejcik (approved):

PR #1655, opened 2021-06-07
ci: add duration timestamps on all test jobs

vdovhanych:
The option will allow us to see the duration of each step in test jobs we run on ci. This is mainly for better debugging of legacy test job, which sometimes takes 30 minutes to finish.

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -66,6 +66,7 @@ stdenv.mkDerivation ({
     rustStable
     wget
     zlib
+    moreutils
   ] ++ lib.optionals (!stdenv.isDarwin) [
     procps
     valgrind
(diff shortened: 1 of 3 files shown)


matejcik (approved):
utACK, ping @mmilata to look over the nix side of things (although it seems harmless enough)

tsusanka (approved):

mmilata:
[Docs for |&](https://www.gnu.org/software/bash/manual/html_node/Pipelines.html#Pipelines) since most of searches lead to stackoverflow which is down today.

vdovhanych:
> What do you think about piping both stdout and stderr to ts using |& instead of |?

Not really sure if it is necessary, to timestamp the errors right or am I wrong? If the error message occurs it will show it and that's it. Do you think we should also add it to possible error output? Personally, I'd leave it like this with only standard output having time stamps.

mmilata (approved):

PR #1663, opened 2021-06-10
Merge 21.06 release branch

mmilata:
Rust-related buildsystem changes were dropped.

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 2.4.0 [24th May 2021]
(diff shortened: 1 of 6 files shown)


tsusanka (approved):
I am missing:
- `build(core): skip rust library for now` but that was release-only so :heavy_check_mark: 
-  `feat(core): avoid restarting session for select messages` but that is cherry-picked from 86089fa5ad2d5fd19e8a0cf6957d2cae59af5516 which is in `master` so :heavy_check_mark: 
- `fix(core): spaces in Shamir info texts` but that is cherry picked from commit 767e7b8e8de3cd677211406d52c68a3da10915b7 which is in `master`  so :heavy_check_mark:

mmilata:
Yeah, `changelog prebuild` needs to be modified to handle this case as well. Thanks for review, merging!

PR #1669, opened 2021-06-14
Make RDI work also on T1

hiviah:
This requires moving the DRBG initialization in `reset_handler` while interrupts are still forbidden, otherwise SysTick will fail when called before DRBG is initialized.

Tested both on T1 and TT.

Needs a bit more though, as I found out disabling RDI from SConscript still leaves some code behind, e.g. T1 will trip up on `drbg_generate` even if RDI is off. So probably some `#ifdef` is missing somewhere.

### DIFF
--- a/core/embed/firmware/main.c
+++ b/core/embed/firmware/main.c
@@ -52,6 +52,8 @@
 #include "touch.h"
 
 int main(void) {
+  random_delays_init();
+
 #ifdef RDI
   rdi_start();
 #endif
--- a/core/embed/firmware/main.c
+++ b/core/embed/firmware/main.c
@@ -52,8 +52,6 @@
 #include "touch.h"
 
 int main(void) {
-  random_delays_init();
-
 #ifdef RDI
   rdi_start();
 #endif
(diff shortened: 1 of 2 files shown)


onvej-sl:
> Needs a bit more though, as I found out disabling RDI from SConscript still leaves some code behind, e.g. T1 will trip up on drbg_generate even if RDI is off. So probably some #ifdef is missing somewhere.

This should be fixed in https://github.com/trezor/trezor-firmware/pull/1676/commits/db06ab00f6ab92f47d3bd669d748eb43bffdf136.

onvej-sl:
I think there is no reason to move `random_delays_init` to `startup.S`.

The only think we need is to disable all exceptions (or at least the systick) before the interrupt vector table is switched and reenable them after the RAM is populated with initialized static variables (.data segment). Otherwise an exception handler can jump in and access uninitialized static variables which could lead to another exception.

Furthermore, I'm afraid that disabling the exceptions with `cpsid if` won't work on a production T1, since the firmware on T1 is started in unprivileged mode.

onvej-sl (approved):

PR #1670, opened 2021-06-16
test: add device tests for non-standard paths

### LINKED ISSUE #1665: Add device tests for non-standard paths
We have repeatedly had issues with breaking applications that use non-standard paths (Unchained Capital, Casa and GreenAddress), see below. A set of device tests is required in order to avoid this in the future. The tests should cover all messages that have an `address_n` field, but most importantly `GetAddress` and the transaction signing messages.
https://github.com/trezor/trezor-firmware/blob/74cf309a9311ee6ec33cea6f7b7bab1c5a15aedf/core/src/apps/bitcoin/keychain.py#L48-L62

mmilata:
Fixes #1665.


tests/device_tests/test_nonstandard_paths.py
```diff
@@ -0,0 +1,255 @@
...
+            messages.InputScriptType.SPENDWITNESS,
+            messages.InputScriptType.SPENDP2SHWITNESS,
+        ),
+    ),
+    # Casa m/49/coin_type/account/change/address_index
+    (
+        "m/49/0/63/0/255",
+        (messages.InputScriptType.SPENDP2SHWITNESS,),
+    ),
+    # Unchained hardened m/45'/coin_type'/account'/[0-1000000]/change/address_index
```
matejcik, line 76: are you sure UC paths apply for single-sig?

mmilata: Don't know. I copied whatever `validate_path_against_script_type` allows in `apps/bitcoin/keychain.py`.

matejcik: ah, but that branch only applies if `multisig` is set, because in that case `SPENDADDRESS` is equivalent to `SPENDMULTISIG`

this btw maybe indicates a problem either in keychain or in the test -- these paths should not be allowed for single-sig

mmilata: Uh, right, should be fixed in 1102ebdddbc2158b1f950045e4b6db04e2816a60.

matejcik (approved):

tests/device_tests/test_nonstandard_paths.py
mmilata: ISTM that failing to validate path against script type only results in warning dialog for `get_address` as well as `sign_tx`.

PR #1671, opened 2021-06-17
Send ButtonRequests for paging events

matejcik:
Previously, when the user was swiping through a paginated screen, this would be completely invisible to the host. This presented some challenges, mainly in our test suite, where we couldn't reliably detect that the current screen is paginated or how many pages there are to swipe through.

Two fields were added to ButtonRequest. The fields are optional - if present, the current screen is paginated. `page_number` indicates which page will be shown after the ButtonRequest is ack'd.

The test suite will use the new ButtonRequests to detect and auto-swipe-through paginated screens. This guarantees that UI tests cover all paginated screens, unless a custom input_flow is used.

cc @szymonlesisz for protobuf & behavior changes


matejcik:
[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1358863213/artifacts/master_diff/index.html), I like it a lot

jpochyla:
Does it add any noticable delay to the page transitions?

mmilata (approved):
LGTM! Nice improvement re: tests.

I suppose this can break integration with third parties. Didn't have success running HWI tests but it seems they don't do much on buttonrequests so we're probably safe here: https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/trezor.py#L202

UI diff lgtm

matejcik:
> Does it add any noticable delay to the page transitions?

no, if there's any delay, it's imperceptible

matejcik:
if you found this from the changelog link: the main feature (sending BRs on every swipe) was reverted in #1696

Issue #1674, opened 2021-06-20
Intermediary FW asks for FW confirmation on new devices


tsusanka:
If I install Intermediary 1.10.0 on brand new device from the factory (BL 1.4.0) I am prompted for firmware update confirmation which goes against the intermediary [spec](https://www.notion.so/satoshilabs/Intermediary-FW-3815baff0d4d4682b40dc7a3bcefbb9a). The spec says there should no confirmation needed on user's side.

This is what Trezor looks like after Intermediary FW is installed:
![image](https://user-images.githubusercontent.com/1835345/122678194-f0a51a00-d1e5-11eb-886f-0bbf225b594a.png)

If I install normal FW afterwards this is what I get:

![image](https://user-images.githubusercontent.com/1835345/122678047-51802280-d1e5-11eb-84d6-12fed0ede4d3.png)

But the FW should be installed right away without any confirmation because the intermediary FW should delete itself. It should therefore equal to a device on which I run `trezorctl wipe-device --bootloader`, which looks like this and does not prompt before FW install:

![image](https://user-images.githubusercontent.com/1835345/122678204-0ca8bb80-d1e6-11eb-989f-1386483dff9e.png)

I believe this is not a blocker for Suite's release - there is simply just an extra confirmation which is not ideal but not blocking as Suite handles it. It would be nice to fix this though so the extra confirmation is removed.

Note that as described in the spec the Intermediary FW **can not erase storage for initialized devices**. Having prompt there is OK as written in the spec.

Full video [here](https://drive.google.com/file/d/10JxEPD1nv5cGpYH1z5YgF-O7kIv_kwAG/view?usp=sharing).

prusnak:
The issue here is that bootloader 1.10.0 still detects the firmware via `firmware_present_new`


https://github.com/trezor/trezor-firmware/blob/69d1465e0862aebf52640bf6f2c95a3a486df0ef/legacy/bootloader/signatures.c#L124-L137

although we invalidate it here:

https://github.com/trezor/trezor-firmware/blob/69d1465e0862aebf52640bf6f2c95a3a486df0ef/legacy/intermediate_fw/trezor.c#L73-L85

hiviah:
I'd guess the word is not written either because the unlock is not correct or needs erase first. There was something with using `memory_write_unlock` (hack that might be needed just for replacing bootloader) or missing erasing the sector first.

prusnak:
[...]

PR #1675, opened 2021-06-21
Fix for #1674 - always invalidate firmware magic 4 bytes in intermediate FW

hiviah:
Always invalidate first magic 4 bytes of firmware header, no matter whether storage was initialized or not.

This fixes the confirmation of "Install new firmware" and also avoids storage erase.

### DIFF
--- a/legacy/intermediate_fw/trezor.c
+++ b/legacy/intermediate_fw/trezor.c
@@ -138,17 +138,10 @@ int main(void) {
     }
   }
 
-  if (sectrue == storage_initialized) {
-    // Storage probably contains a seed so leave the firmware intact.
-    // Invalidating it would cause the bootloader to wipe the storage before
-    // installing the target firmware. User will need to confirm installation.
-    reboot_device();
-  } else {
-    // New device. Invalidate the intermediate firmware so that after reboot
-    // the bootloader will install the target firmware without asking for user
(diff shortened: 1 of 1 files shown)


prusnak:
Are you 100% sure this will not erase the storage on subsequent firmware update?

hiviah:
I think I can't really test/debug that part to be 100% sure since the signature check `signatures_new_ok` will always fail with unsigned firmware, both at update time for new and also for old stored firmware.

hiviah:
I tried to prove that changing the 4 magic bytes won't change the outcome, but actually proved the opposite.

Because of this FW magic check

https://github.com/trezor/trezor-firmware/blob/f39abc8356d18a50e3201f534cb22c3ca72b5c0f/legacy/bootloader/usb.c#L233

the variable `old_was_signed` remains false (`SIG_FAIL`) and this following check on it will cause firmware erase:

https://github.com/trezor/trezor-firmware/blob/f39abc8356d18a50e3201f534cb22c3ca72b5c0f/legacy/bootloader/usb.c#L391

So we can't get rid of the confirmation without changing bootloader.

prusnak:
Closing, we don't want this solution. A better solution is decribed in https://github.com/trezor/trezor-firmware/issues/1674#issuecomment-872062490

Issue #1658, opened 2021-06-09
Successful wipe code triggers RSOD


wendys-cats:
**Describe the bug**
After entering wipe code, device will successfully wipe, but will display RSOD. After reconnecting the device is wiped and functions. Applies only for TT device 2.4.0 firmware, T1 works as intended. 

Text
FATAL ERROR
expr: drbg_initialized
file: embed/trezorhal/random&delays.c:81
func: drbg_generate
rev:  revisbelow

**Firmware version and revision**
TT 2.4.0 usigned ea3596a
TT 2.4.0 signed ea3596ad8

**Desktop/smartphone setup (please complete the following information):**
 - Wallet software: Trezor Suite app and staging wallet 3a553b24e498c56a2e03897fded706dffb4a6d6b
 - OS: Windows 10  and NixOS 21.03
 - Browser none and Chromium 
 - Version APP and Version 88.0.4324.182 (Official Build) (64-bit)

**To Reproduce**
1. Enable PIN protection
2. Enable wipe-code protection via trezorctl
3. Enter wipe-code instead of PIN

**Expected behavior**
Device will wipe itself without displaying errors.

**Screenshots**

![photo_2021-06-09_14-33-24](https://user-images.githubusercontent.com/58593111/121359147-7ba72a00-c933-11eb-8a80-68a8c5fd258c.jpeg)

prusnak:
This is a regression from 2.3.x, assigning P1

STew790:
I was able to reproduce this with exactly the same setups. 

**Firmware version and revision**
TT 2.4.0 usigned ea3596a
TT 2.4.0 signed ea3596ad8

**Desktop/smartphone setup (please complete the following information):**

Wallet software: Trezor Suite app and staging wallet 3a553b24e498c56a2e03897fded706dffb4a6d6b
OS: Windows 10 and NixOS 21.03
Browser none and Chromium
Version APP and Version 88.0.4324.182 (Official Build) (64-bit)

andrewkozlik:
[...]

PR #1676, opened 2021-06-22
Fix of red screen

onvej-sl:
This PR fixes https://github.com/trezor/trezor-firmware/issues/1658.

From my point of view, the cause of the issue is that `shutdown()` is called from the unprivileged mode (see [this comment](https://github.com/trezor/trezor-firmware/issues/1658#issuecomment-865883049)). To solve that, I implemented a SVC handler that calls `shutdown()` from the privileged mode.

### DIFF
--- a/core/embed/trezorhal/common.c
+++ b/core/embed/trezorhal/common.c
@@ -25,6 +25,7 @@
 #include "display.h"
 #include "flash.h"
 #include "rand.h"
+#include "supervise.h"
 
 #include "stm32f4xx_ll_utils.h"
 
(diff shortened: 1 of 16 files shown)


core/embed/trezorhal/common.c
```diff
@@ -25,13 +25,23 @@
...
 
-#define COLOR_FATAL_ERROR RGB16(0x7F, 0x00, 0x00)
+void shutdown(void) {
+#ifdef USE_SVC_SHUTDOWN
+  svc_shutdown();
+#else
+  // It won't work properly unless called from the privileged mode
+  shutdown_privileged();
+#endif
+}
```
prusnak, line 44: Maybe one suggestion to consider using define instead of a function here.

But maybe that's over-optimizing stuff and compiler does this anyway ...

```suggestion
#ifdef USE_SVC_SHUTDOWN
#define shutdown svc_shutdown
#else
// It won't work properly unless called from the privileged mode
#define shutdown shutdown_privileged
#endif
```

prusnak (commented):
PR looks good, left one suggestion to consider

hiviah (approved):

core/embed/trezorhal/common.c
onvej-sl: > But maybe that's over-optimizing stuff and compiler does this anyway ...

Yes, compiler inlines it. I'd leave it as it is.

prusnak (approved):

PR #1677, opened 2021-06-22
ci/hardware: enable SD card tests and btc tests on regular build

mmilata:
Increase timeout since 4h isn't always enough.

### DIFF
--- a/ci/test-hw.yml
+++ b/ci/test-hw.yml
@@ -5,8 +5,7 @@ image: registry.gitlab.com/satoshilabs/trezor/trezor-firmware/trezor-firmware-en
 # Currently it's not possible to run all regular TT tests without getting into
 # a state where the micropython heap is too fragmented and allocations fail
 # (often manifesting as a stuck test case). For that reason some tests are
-# skipped and the rest is divided into several groups with device reboots in
-# between them.
+# skipped.
 # See also: https://github.com/trezor/trezor-firmware/issues/1371
 hardware core regular device test:
   stage: test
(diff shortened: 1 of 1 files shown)


matejcik:
The reason SD card tests are disabled is that they are _eating_ SD cards. There is no wear leveling and so every test run is hammering the same sectors. I'm not sure if we want to be reopening this particular can of worms again?

matejcik:
also honestly not sure we want to run btc tests on regular build? we might but it's saving significant time that we don't do that?

mmilata:
Admins bought some kind of "endurance" SD card (+couple spares) that are meant for constantly recording cameras, we can see how long they last.

Agreed that btc tests on regular build probably improve coverage negligibly compared to the increase in test duration (72m -> 216m). I thought we can afford it for the nightly runs but it might be too much in case we want to do multiple runs in a single day.

tsusanka:
- ACK for sd card tests, I would add them even if we would need to replace the card once in a while. Let's try it..
- Regarding bitcoin tests on regular build: I do not have a strong preference here. If we add it, do we still complete a complete pipeline for master over one night? I mean it would be nice to have a result in the morning.

mmilata:
Currently the nightly pipeline takes almost 6h, if we add the btc tests it will take a bit over 8h. I'd say it's still feasible to run overnight, although we should set the start to earlier than 1 am so that it's finished when people start working in the morning.

tsusanka:
We can set it to let's say 11 pm. So let's force push and merge this?

tsusanka (approved):

Issue #1601, opened 2021-05-06
Cardano streamed transaction signing


leanil:
**Is your feature request related to a problem? Please describe.**
The unserialized python representation of complete Cardano transactions can already exceed Trezor’s memory if they contain too many items (inputs, outputs, asset groups, etc.). Also new elements are continuously being added to transactions.
Independently, returning completely valid serialized transactions from Trezor won't be possible as we introduce multisig, making a rework of outgoing messages inevitable.

**Describe the solution you'd like**
The Vacuumlabs team is preparing a larger refactoring of the Cardano transaction signing process to stream the upload of transaction data to Trezor (i.e. multiple small incoming messages rather than a single `CardanoSignTx` request). We've described the proposed changes in more detail in [this document](https://docs.google.com/document/d/1K5Q8XCHt9Y2vgDNGAZed1aOJ7HokSWENT95wWBq_bJ4/edit?usp=sharing), and we would like to hear your opinion and receive feedback on it before we start implementing them.

**Describe alternatives you've considered**
No alternatives

**Additional context**
- We won’t be returning the whole serialized transaction anymore. We will only return the transaction hash, the witnesses and auxiliary data supplement. This also makes more sense for multisig, but is a breaking change for both Trezor Connect and Trezorlib.
-  The full auxiliary data as a blob will not be sent to Trezor anymore, only its hash will be - our reasoning is explained in the document.
- Could you briefly describe the intents behind Bitcoin’s implementation? E.g. what's the benefit of explicitly requesting transaction parts, rather than simply sending the parts in a predefined order?

matejcik:
> * Could you briefly describe the intents behind Bitcoin’s implementation? E.g. what's the benefit of explicitly requesting transaction parts, rather than simply sending the parts in a predefined order?

This lets the host implement a conceptually simple algorithm (send the part when requested), while the firmware can contain more complex logic of requesting certain parts of the transaction at the appropriate times. It also allows greater flexibility in processing on the firmware side.

Examples:
- Legacy signing needs to restream the whole transaction, while segwit signing only needs to restream the outputs.
- We were able to move prevtx streaming after the user confirmation, by simply requesting the outputs in advance and then requesting the previous txes afterwards -- with no change to the host-side libraries

matejcik:
> we don’t see the benefit of explicitly requesting the transaction parts (like the btc signing does)
> or even introducing empty ack messages

[...]

Issue #67, opened 2018-07-10
Race condition when signing NEM transactions on emulator with DebugLink


mlejva:
I managed to reproduce this race condition approximately 6 times out of total 10 runs.

I'm using [connect](https://github.com/trezor/connect/tree/tests/connect) together with an emulator. The emulator is [loaded](https://github.com/trezor/connect/blob/tests/connect/src/__tests__/init_device.py) with DebugLink. 
Transport is [started](https://github.com/trezor/connect/blob/tests/connect/src/__tests__/run.sh#L150-L164) with both the emulator and DebugLink support  - `./trezord-go -e 21324 -e 21325`.

Once the emulator is ready and the `device__connect` event is received `nemSignTransaction` method is [called]() via connect with the following payload

```javascript
{
    timeStamp: 1,
    fee: 2000000,
    type: 4100,
    deadline: 74735615,
    otherTrans: {
        timeStamp: 2,
        amount: 2000000,
        fee: 15000,
        recipient: 'TALICE2GMA34CXHD7XLJQ536NM5UNKQHTORNNT2J',
        type: 257,
        deadline: 67890,
        message: {
            payload: '746573745f6e656d5f7472616e73616374696f6e5f7472616e73666572',
            type: 1,
        },
        version: -1744830464,
        signer: 'c5f54ba980fcbb657dbaaa42700539b207873e134d2375efeab5f1ab52f87844',
    },
    version: -1744830464,
}
```

After last confirmation button is pressed emualtor returns 'Failure_Firmware'.
This is output from the emulator
```
Traceback (most recent call last):
 	 File "/trezor-core/src/trezor/wire/__init__.py", line 120, in session_handler
	 File "/trezor-core/src/trezor/wire/__init__.py", line 117, in session_handler
	 File "/trezor-core/src/trezor/wire/__init__.py", line 153, in protobuf_workflow
	 File "/trezor-core/src/trezor/wire/__init__.py", line 153, in protobuf_workflow
	 File "/trezor-core/src/trezor/wire/__init__.py", line 142, in protobuf_workflow
	 File "/trezor-core/src/apps/nem/signing.py", line 32, in sign_tx
	 File "/trezor-core/src/apps/nem/multisig/__init__.py", line 26, in aggregate_modification
```

Similar issues happens when calling `nemSignTransaction` with the following payload
```javascript
{
    timeStamp: 74649215,
    fee: 2000000,
    type: 16386,
    deadline: 74735615,
    message: {},
[...]

PR #1683, opened 2021-06-30
feat(cardano): streamed transaction signing

gabrielKerekes:
As already announced in https://github.com/trezor/trezor-firmware/issues/1601 we have been working on “refactoring” the Cardano transaction signing process to be streamed. This PR is the result of our work.

**Important points:**
- Trezor will no longer expect all the tx data at once but will expect separate messages for tx elements (inputs, outputs, assets, tokens, certificates, …).
- Trezor will no longer return the whole serialized transaction but will return just the transaction hash, data for witnesses construction (`pub_key`, `signature`, `chain_code`) and auxiliary data supplement, which is for now only used for Catalyst registration metadata. This means that clients will then have to assemble the transaction on their own.
- In the case of arbitrary auxiliary data (not Catalyst registration) only the hash of auxiliary data is being sent to Trezor.
- Inputs no longer contain `path`s on the FW level - `path`s are still passed into clients and used by them to determine which witnesses should be requested.
- It is no longer determined by the FW which witnesses should be added to the transaction/returned. Witness requests are passed in in a separate message during the tx signing process instead. This means that clients (mainly trezorlib, Trezor Connect) “gather” these witnesses from the transaction parameters and send them to Trezor. Unreasonable witness paths are shown to the user.
- Hiding of “change” outputs is now independent of input paths and is determined purely on the address path being reasonable.
- Everything else that would require comment is hopefully documented in the code.

As a result, Trezor is now able to sign a Cardano transaction with 65 inputs, 40 outputs, another 10 outputs with 12 policies and 10 assets each. The multiasset outputs were all over the 4kB max size given by the protocol. The whole transaction was 62kB (current protocol tx size limit is 16kB). It could also probably handle much much more, but these tests already took long enough to complete. So we should be ready for future updates.

I plan on generating the ui test fixtures as well as adding changelog entries and updating the Cardano README after the review process. Accompanying Trezor Connect PR is being worked on and should also be ready soon. Hopefully the way the PR is split into commits will simplify the review process at least a little.


tests/device_tests/cardano/test_sign_tx.py
```diff
@@ -102,26 +106,84 @@ def test_cardano_sign_tx_with_multiple_chunks(client, parameters, result):
 
     expected_responses = [
         messages.PassphraseRequest(),
-        # XXX as many ButtonRequests as paginations. We are relying on the fact that
-        # there is only one fixture whose pagination is known.
-        # If that changes, we'll need to figure out something else.
+        *[messages.CardanoTxItemAck()] * 4,
```
gabrielKerekes, line 109: I'm not a 100% convinced this is a good idea. It looked a bit simpler before rebasing on top of the `ButtonRequest` changes. But perhaps it's worth it to have such a test and it won't probably change too often.

core/src/apps/cardano/helpers/lazy_cbor_collection.py
```diff
@@ -0,0 +1,130 @@
...
+    ) -> Iterator[bytes | Type[LazyCborCollection.PauseIteration]]:
+        yield _get_cbor_header(_CBOR_TYPE_ARRAY, self.size)
+        yield self.PauseIteration
+
+        for element in self:
+            yield from cbor.encode_streamed(element)
+            yield self.PauseIteration
+
+
+def _get_cbor_header(typ: int, l: int) -> bytes:
```
gabrielKerekes, line 114: We weren't sure what the best approach here was. Either to make `cbor._header` public and thus making the cbor library less independent or to have a copy of the function.

gabrielKerekes (commented):
Leaving a few comments to point out the things we were unsure about. Also `cbor.OrderedMap` and `cbor.encode_chunked` are no longer used now. Do we want to delete them or should we rather keep them in since they are already there?

gabrielKerekes:
I'm not sure why `style_check` is failing in the CI, everything seems to be ok locally.

gabrielKerekes:
While adding support for token minting I noticed that the auxiliary data weren't handled in an ideal way - adding another message after auxiliary data wasn't straightforward and `auxiliary_data_supplement` would stay in memory until sent back. I've updated the code to send back `auxiliary_data_supplement` right after processing auxiliary data so that other transaction elements can easily be sent after auxiliary data.

If you haven't started reviewing the code yet, I would squash the fixups.

matejcik:
> If you haven't started reviewing the code yet, I would squash the fixups.

please go ahead

python/src/trezorlib/cli/cardano.py
```diff
@@ -83,10 +91,26 @@ def sign_tx(client, file, protocol_magic, network_id, testnet):
...
+            "type": witness["type"],
+            "pub_key": witness["pub_key"].hex(),
+            "signature": witness["signature"].hex(),
+            "chain_code": witness["chain_code"].hex()
+            if witness["chain_code"] is not None
+            else None,
+        }
+        for witness in sign_tx_response["witnesses"]
+    ]
+    if auxiliary_data_supplement := sign_tx_response.get("auxiliary_data_supplement"):
```
matejcik, line 106: please don't use the walrus in trezorlib, we're still keeping python 3.6 compatibility.

(it's ok to use it in the firmware itself)

tests/device_tests/cardano/test_sign_tx.py
matejcik: The original purpose of this test was to ensure that the response is properly chunked. I don't think that applies anymore, so feel free to drop the test altogether.

python/src/trezorlib/cli/cardano.py
gabrielKerekes: Ok, I've removed it.

tests/device_tests/cardano/test_sign_tx.py
gabrielKerekes: I basically deleted the test but in the end I decided to move it to `sign_tx.json` - so there would be a test which tests a transaction with many of the parameters set. The communication checks are gone though.

gabrielKerekes:
> please go ahead

Squashed.

matejcik:
please rebase again so we get the pesky style check green

gabrielKerekes:
Nice, all green now.

gabrielKerekes:
@matejcik will there be an August release please? If so, do you think we could fit this in?

matejcik:
We're not currently planning to release in August. I'll let you know if this changes.

gabrielKerekes:
@matejcik what is the review status on this PR please? I see it's been moved to `In progrees` 13 days ago. I'm asking because we have another larger PR coming (https://github.com/vacuumlabs/trezor-firmware/pull/66) which I've just put up for internal review and it would be desirable to get both of these into the September release if possible. I'm aware that FW freeze should happen on Aug 20 so the schedule is quite tight.

python/src/trezorlib/cli/cardano.py
```diff
@@ -40,20 +40,27 @@ def cli():
 @cli.command()
 @click.argument("file", type=click.File("r"))
 @click.option("-f", "--file", "_ignore", is_flag=True, hidden=True, expose_value=False)
+@click.option(
+    "-s",
+    "--signing-mode",
+    required=True,
+    type=click.Choice(cardano.SIGNING_MODE_IDS.keys(), case_sensitive=False),
```
matejcik, line 47: instead:
```
ChoiceType({m.name: m for m in messages.CardanoTxSigningMode})
```
(modify `ChoiceType` if you need the `case_sensitive` flag?)
then you can remove `SIGNING_MODE_IDS` entirely

tests/device_tests/cardano/test_sign_tx.py
```diff
@@ -34,6 +34,7 @@
     "cardano/sign_tx.slip39.json",
 )
 def test_cardano_sign_tx(client, parameters, result):
+    signing_mode = cardano.SIGNING_MODE_IDS[parameters["signing_mode"]]
```
matejcik, line 37:
```
signing_mode = messages.CardanoSigningMode.__members__[parameters["signing_mode"]]
```

tests/device_tests/cardano/test_sign_tx.py
```diff
@@ -92,46 +95,31 @@ def test_cardano_sign_tx_failed(client, parameters, result):
...
 
 
-@parametrize_using_common_fixtures("cardano/sign_tx.chunked.json")
-def test_cardano_sign_tx_with_multiple_chunks(client, parameters, result):
-    inputs = [cardano.parse_input(i) for i in parameters["inputs"]]
-    outputs = [cardano.parse_output(o) for o in parameters["outputs"]]
-    certificates = [cardano.parse_certificate(c) for c in parameters["certificates"]]
-    withdrawals = [cardano.parse_withdrawal(w) for w in parameters["withdrawals"]]
-    auxiliary_data = cardano.parse_auxiliary_data(parameters["auxiliary_data"])
+def _transform_expected_result(result):
```
matejcik, line 98: side note: this (and the whole `trezorlib.cardano` changeset) is a strong argument for doing the trezorlib part _somehow differently_, but I honestly don't have any idea how. Maybe follow how Connect does it?

No change requested (I don't think anyone other than the test suite is actually using the Python API), but noting it for posterity.

matejcik (changes requested):

common/tests/fixtures/cardano/sign_tx.json
```diff
@@ -845,11 +1070,279 @@
...
+          {
+            "type": 1,
+            "pub_key": "5d010cf16fdeff40955633d6c565f3844a288a24967cf6b76acbeb271b4f13c1",
+            "signature": "a1f28828038757949ac319ca24bcd41d6af41ced07db20d5b6ed7392e4b6aa663ee92e873fa571a257ddeef945bee76694107da50edbd4f6c12b8654e0a22b02",
+            "chain_code": null
+          }
+        ]
+      }
+    },
+    {"description": "transaction with everything set except pool registration certificate",
```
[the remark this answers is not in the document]
gabrielKerekes: 👍 https://github.com/trezor/trezor-firmware/pull/1683/commits/92746053baa1620dd30160814c64f94c213c7413

tests/device_tests/cardano/test_sign_tx.py
gabrielKerekes: https://github.com/trezor/trezor-firmware/pull/1683/commits/92746053baa1620dd30160814c64f94c213c7413 (I've fixed this up in the wrong commit, I'll fix this during rebasing)

core/src/apps/cardano/__init__.py
```diff
@@ -0,0 +1 @@
+CURVE = "ed25519"
```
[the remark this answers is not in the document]
gabrielKerekes: > any reason for this?

Probably not then. https://github.com/trezor/trezor-firmware/pull/1683/commits/45b53e8adebc8240d508e07df799b5e1a1de3ee8

core/src/apps/cardano/certificates.py
```diff
@@ -74,32 +74,36 @@ def cborize_certificate(
             certificate.pool,
         )
     elif certificate.type == CardanoCertificateType.STAKE_POOL_REGISTRATION:
-        pool_parameters = certificate.pool_parameters
+        assert False, "Use cborize_initial_pool_registration_certificate_fields instead"
```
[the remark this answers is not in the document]
gabrielKerekes: It was intentional and could only be hit by programmer error. I removed it though as suggested - https://github.com/trezor/trezor-firmware/commit/45b53e8adebc8240d508e07df799b5e1a1de3ee8.

core/src/apps/cardano/helpers/cbor_hash_builder.py
```diff
@@ -0,0 +1,87 @@
+from apps.cardano.helpers.lazy_cbor_collection import (
+    LazyCborCollection,
+    LazyCborDict,
+    LazyCborList,
+)
+
+if False:
+    from typing import Any, Protocol
+
+    class HashFn(Protocol):
```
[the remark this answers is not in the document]
gabrielKerekes: 👍 https://github.com/trezor/trezor-firmware/commit/45b53e8adebc8240d508e07df799b5e1a1de3ee8

core/src/apps/cardano/layout.py
```diff
@@ -371,44 +360,29 @@ async def confirm_stake_pool_metadata(
...
-        "confirm_pool_network",
+        "confirm_pool_final",
         title="Confirm transaction",
         props=[
-            ("Network:", protocol_magics.to_ui_string(protocol_magic)),
-            (
-                "Valid since: %s" % format_optional_int(validity_interval_start),
-                None,
-            ),
+            ("Confirm signing the stake pool registration as an owner\n\n\n", None),
```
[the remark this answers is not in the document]
gabrielKerekes: I removed the newlines and I also changed the other properties to use the tuple instead of string formatting - it seems more correct in the code and doesn't mess up the UI too much. https://github.com/trezor/trezor-firmware/commit/45b53e8adebc8240d508e07df799b5e1a1de3ee8

The newlines are used similarly multiple times in the file though so that's where I got it from --> https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/cardano/layout.py#L153. This was added when transitioning to the new layout system.

python/src/trezorlib/cli/cardano.py
gabrielKerekes: :+1: https://github.com/trezor/trezor-firmware/commit/45b53e8adebc8240d508e07df799b5e1a1de3ee8 The `case_sensitive` flag is not really needed.

core/src/apps/cardano/helpers/lazy_cbor_collection.py
```diff
@@ -0,0 +1,130 @@
...
+_CBOR_TYPE_MAP = const(0b101 << 5)
+
+
+class LazyCborCollection(Generic[T]):
+    """
+    A base class for LazyCborCollections. Contains basic functionality for a lazy collection. Enables iteration over
+    and CBOR serialization of its items as they are being added (one by one) thus never storing them all at once.
+    """
+
+    class PauseIteration:
```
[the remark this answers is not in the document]
gabrielKerekes: Yes, it's pretty much just to simplify the type hints. I tried converting these to objects but I wasn't able to figure out the typing then - do you have any suggestion?

If it's OK to keep the classes, should we be returning the type or an instance of the class [here](https://github.com/trezor/trezor-firmware/pull/1683/commits/c567c40df696b6d5ebe2ceca24e727cccd086c4c#diff-26f6db3cc7dd537dd5ff942a3bd922c6f33f67457356fdc6d1a6116f154acff8R99)?

core/src/apps/cardano/helpers/cbor_hash_builder.py
```diff
@@ -0,0 +1,87 @@
...
+        self._update_hash()
+
+    def _update_hash(self) -> None:
+        while (chunk := next(self.serializer)) is not LazyCborCollection.PauseIteration:
+            # for type checking
+            assert isinstance(chunk, bytes)
+            self.hash_fn.update(chunk)
+
+    def get_hash(self) -> bytes:
+        assert (
```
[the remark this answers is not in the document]
gabrielKerekes: Since the class is "protected" by `tx_builder` the asserts are IMHO safe and can only break by programmer error. However if we remove/update `tx_builder` as you propose then these might need to be actual errors.

core/src/apps/cardano/helpers/lazy_cbor_collection.py
gabrielKerekes: Rather than exposing the internal `_header` function I've created `create_array_header` and `create_map_header` functions which are "public". It seems more correct than exposing both the `_header` function and the type constants. Also the `_header` IMHO has a name which isn't descriptive enough for the outside world. I also added tests for these functions.

https://github.com/trezor/trezor-firmware/pull/1683/commits/45b53e8adebc8240d508e07df799b5e1a1de3ee8#diff-bce249b14d43a52eb8502d517376e3eebf19dcce888961d10dad2ac1218578caR315

core/src/apps/cardano/helpers/tx_builder_state_machine.py
```diff
@@ -0,0 +1,166 @@
+from micropython import const
+
+if False:
+    ActionType = int
+
+
+class TxBuilderStateMachine:
```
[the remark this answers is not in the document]
gabrielKerekes: I will have to take a closer look at this - but it does sound reasonable and doable.

core/src/apps/cardano/layout.py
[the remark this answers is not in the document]
matejcik: Yeah, those shouldn't be there either, they're from a previous iteration of the layout system. We'll be cleaning up all of it within this year.

:+1: for tuple usage. The reason we left the original formatting was to keep the UI differences to a minimum, but this is another thing we'll be flipping over soon.

core/src/apps/cardano/helpers/cbor_hash_builder.py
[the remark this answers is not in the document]
gabrielKerekes: I replaced the asserts to 'if's raising a `RuntimeError` - I couldn't find a more suitable error. I've also added unit tests for the class. https://github.com/trezor/trezor-firmware/pull/1683/commits/fef973a76a34acf4d0118f2f078cb3382f78c6b6

core/src/apps/cardano/helpers/tx_builder_state_machine.py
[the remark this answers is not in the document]
gabrielKerekes: In the end I've inlined the calls to the hash builder and thus `tx_builder` and the state machine have been deleted. I tried implementing it via the generator but the inner structure of the generator was basically a duplicate of the branching in the `process_xxx` calls and also lacked type and other guarantees. It also worsened readability significantly, which could perhaps be improved in a few iterations, but IMHO it's not worth it. If you're interested you can take a look https://github.com/vacuumlabs/trezor-firmware/pull/67 - I'd be interested in hearing some alternatives/improvements to how it can written with a generator - perhaps I went the wrong way with the function yielding.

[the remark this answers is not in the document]
matejcik: wrt generator usage, what I had in mind was sort of the inverse of what you are trying at [`a5e01a1` (#67)](https://github.com/vacuumlabs/trezor-firmware/pull/67/commits/a5e01a15ab2ab4b15bfc7982928a813af2e8334f)

simplified pseudocode:
```
def build_tx_hash(params):
	hash_builder = HashBuilder()
	hash_builder.add(params.foo)
	hash_builder.add(params.bar)
	for _ in range(params.num_inputs):
		inp = yield
		assert isinstance(inp, CardanoInput)
		yield from build_tx_hash_input(hash_builder, inp)

def build_tx_hash_input(inp):
	hash_builder.add_collection("input")
	hash_builder.add(inp.amount)
	for _ in range(inp.metadata_len):
		meta = yield
		assert isinstance(meta, CardanoInputMeta)
		hash_builder.add(meta)

def sign_tx(msg):
	builder = build_tx_hash(msg)
	for inp in get_inputs():
		builder.send(inp)
		builder.send(inp.meta1)
		builder.send(inp.meta2)
	for out in get_outputs():
		builder.send(out)
```

it's much more readable, but it still has the properties that (a) it copies the layout of the `process_xxx` calls and (b) lacks type guarantees for callers.

---

i'm perfectly OK with the current version, and unless you have compelling reasons to go the generator route, i'd keep the inlined version

matejcik:
i've been playing with a slightly different approach to the lazy collections, see https://gist.github.com/matejcik/27c66f71e2ce5a1e1875172cd2df7dab

usage would be:
```
input_list: LazyList[SomeInputType]
with root_dict.add(TX_BODY_KEY_INPUTS, LazyList(inputs_count)) as input_list:
	_process_inputs(ctx, input_list, inputs_count)
```

advantages are:
* shorter and IMHO more understandable serialization code (plus you wouldn't need tx builder class, stack is managed implicitly)
* context manager = you don't need to call finalize manually
* you can have typed lazy collections

I haven't tried to use it in `sign_tx` yet though so maybe there are problems I don't currently see.
Let me know what you think.

(to be clear I don't insist on using this. Just, as long as we're tweaking...)

matejcik:
> * you can have typed lazy collections

not sure if that's of any practical use; you'd need to express the full shape of the structure in the type hints. The one thing that might be practical is to declare that all dict keys are of type `int`, but item type is probably always `Any`

gabrielKerekes:
I am working on adding it to sign_tx and I'm almost done actually. It seems to be working alright and it IMHO did remove some complexity :+1: The typing isn't perfect, yes, but at least some of the transaction parts can by typed quite nicely. I'm just refactoring it all a bit - will push in a while.

gabrielKerekes:
So `HashBuilderCollection` is added in https://github.com/trezor/trezor-firmware/pull/1683/commits/8945f3d0d7a33e9aeecbba77ff41cf40b0bded58. I renamed some things and added the `finish_and_get_hash` function as I wasn't sure how better to get the hash. I changed `_insert_into_parent` to `_insert_child` as I was confused what was happening before - it seems more intuitive to me the other way around.

core/src/apps/cardano/helpers/hash_builder_collection.py
```diff
@@ -0,0 +1,110 @@
...
+    from trezor.utils import HashContext
+
+    T = TypeVar("T")
+    K = TypeVar("K")
+    V = TypeVar("V")
+else:
+    T = 0  # type: ignore
+    K = 0  # type: ignore
+    V = 0  # type: ignore
+    Generic = [object]  # type: ignore
```
matejcik, line 14: you could do: `Generic = {T: object, (K, V): object}`

core/src/apps/cardano/helpers/hash_builder_collection.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def start(self, hash_fn: HashContext) -> "HashBuilderCollection":
+        self.hash_fn = hash_fn
+        self.hash_fn.update(self._header_bytes())
+        return self
+
+    def _insert_child(self, child: "HashBuilderCollection") -> None:
+        child.parent = self
+        assert self.hash_fn is not None
+        child.hash_fn = self.hash_fn
```
matejcik, line 34: `child.start(self.hash_fn)`

core/src/apps/cardano/helpers/hash_builder_collection.py
```diff
@@ -0,0 +1,110 @@
...
+        if self.remaining != 0:
+            raise RuntimeError  # not all items were added
+        if self.parent is not None:
+            self.parent.has_unfinished_child = False
+        self.hash_fn = None
+        self.parent = None
+
+    def finish_and_get_hash(self) -> bytes:
+        assert self.hash_fn is not None
+        hash = self.hash_fn.digest()
```
matejcik, line 64: you should first `self.finish()` and then get digest

the alternative is to avoid this function and do:
```
hash_fn = blake2b(...)
tx_dict.start(hash_fn)
with tx_dict:
	...
digest = hash_fn.digest()
```

matejcik (commented):
couple of quick notes

core/src/apps/cardano/helpers/hash_builder_collection.py
```diff
@@ -0,0 +1,110 @@
...
+        self.finish()
+
+        return hash
+
+    def __enter__(self) -> "HashBuilderCollection":
+        assert self.hash_fn is not None
+        return self
+
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        if exc_val:
```
matejcik, line 75: exception will fall through if `__exit__` returns a falsey value, so this is a very weird thing to do explicitly

if you want to make sure that `finish()` is not called in case of exception, do instead:
```
if exc_type is None:
	self.finish()
```

gabrielKerekes: Can `exc_val` be a falsey value? I think I need to do:
```
if exc_val is not None:
    raise exc_val
    
self.finish()
```
I need this, because otherwise the errors thrown inside the `with` statement are never propagated which causes tests to fail because of invalid errors - the errors are simply `RuntimeError`s then and also the stack trace isn't helpful for development then.

core/src/apps/cardano/helpers/hash_builder_collection.py
gabrielKerekes: https://github.com/trezor/trezor-firmware/pull/1683/commits/ed157b3fd72aa0fa789980af3597d201ce21672a

core/src/apps/cardano/helpers/hash_builder_collection.py
gabrielKerekes: https://github.com/trezor/trezor-firmware/pull/1683/commits/ed157b3fd72aa0fa789980af3597d201ce21672a

core/src/apps/cardano/helpers/hash_builder_collection.py
matejcik:
```python3
class Ctx:
    def __enter__(self):
        return self

    def __exit__(self, a, b, c):
        if a is None:
            raise RuntimeError("this is from finish()")


with Ctx():
    raise RuntimeError("this is from outside the class")
```

```
  File "test.py", line 11, in <module>
    raise RuntimeError("this is from outside the class")
RuntimeError: this is from outside the class
```

gabrielKerekes: 🤦 Aaah, I see now. I was confused by what's happening. It was indeed failing because of an error being thrown inside of `finish`. My bad, sorry. https://github.com/trezor/trezor-firmware/pull/1683/commits/90a0ad9192bd9ab7cc3e3523526ea6bf376c2ba9

gabrielKerekes:
@matejcik what is your status on this please?

As I mentioned before we have another PR (https://github.com/vacuumlabs/trezor-firmware/pull/66) almost ready - I would be able to create a PR into this repo today/tomorrow. Do you think it's at all possible we could get https://github.com/vacuumlabs/trezor-firmware/pull/66 into the September release?

matejcik:
> Do you think it's at all possible we could get [vacuumlabs#66](https://github.com/vacuumlabs/trezor-firmware/pull/66) into the September release?

We can try but it's unlikely, the PR is lather large.

core/src/apps/cardano/sign_tx.py
```diff
@@ -743,31 +854,38 @@ async def _show_change_output_staking_warnings(
...
-        if (
-            len(output) != BIP_PATH_LENGTH
-            or output[ACCOUNT_PATH_INDEX] != inp[ACCOUNT_PATH_INDEX]
-            or output[ACCOUNT_PATH_INDEX] > MAX_SAFE_ACCOUNT_INDEX
-            or output[-2] >= 2
-            or output[-1] >= MAX_SAFE_CHANGE_ADDRESS_INDEX
-        ):
-            return False
+def _should_hide_output(path: list[int]) -> bool:
+    """Return whether the output address is from a safe path, so it could be hidden."""
```
matejcik, line 858: this means that any path from any account is considered safe, so in particular, transfer between my own accounts will have no visible output confirmation screen?

matejcik: i propose the following modification:

- if there are "internal" outputs (specified by address_params instead of address), at least one "external" output must exist, otherwise transaction is rejected at end of outputs stage
- all provided `address_n`s (for outputs AND witnesses) must match the same account, otherwise transaction is rejected at first non-matching entry

matejcik:
(don't mind the CI failure, it will go away after rebase to master. please don't do that yet though.)

core/src/apps/cardano/helpers/hash_builder_collection.py
```diff
@@ -0,0 +1,98 @@
...
+        self.has_unfinished_child = False
+
+    def start(self, hash_fn: HashContext) -> "HashBuilderCollection":
+        self.hash_fn = hash_fn
+        self.hash_fn.update(self._header_bytes())
+        return self
+
+    def _insert_child(self, child: "HashBuilderCollection") -> None:
+        child.parent = self
+        assert self.hash_fn is not None
```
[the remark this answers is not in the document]
gabrielKerekes: No, but it gets passed into `child.start` where it can't be `None` so that's what the assert is here for.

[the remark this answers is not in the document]
matejcik: ah, right, because you inverted it and this is called in the `parent`! ok disregard my confusion :)

core/src/apps/cardano/sign_tx.py
gabrielKerekes: Before `_should_hide_output` is called in `_show_output`, `address_n` is checked to match against `SCHEMA_ADDRESS` - if it doesn't match, signing will either fail or a warning will be shown - depending on `safety_checks.is_strict()`.

We have refactored `_show_output` in the multi-sig PR so I wouldn't want to touch it here if there are no serious security implications and if there are, I would rather deal with those in a separate PR. The `_should_hide_output` function has also been there before tx streaming. It's just been modified to not take inputs into consideration as that is not possible anymore and checking against the inputs didn't really add much guarantees anyways.

Given all this, I'm don't think the proposed modifications are necessary.

core/src/apps/common/cbor.py
```diff
@@ -62,7 +62,9 @@ def _header(typ: int, l: int) -> bytes:
 
 
 def _cbor_encode(value: Value) -> Iterator[bytes]:
-    if isinstance(value, int):
+    if hasattr(value, "cbor_serialize"):
```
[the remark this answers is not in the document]
gabrielKerekes: I fixed it up in https://github.com/trezor/trezor-firmware/pull/1683/commits/8ed3c0161ac30af1d13a54d8204c8cf79b09dab8, but the whole commit can actually be reverted. I'll do that when cleaning up.

python/src/trezorlib/cli/cardano.py
```diff
@@ -60,7 +60,7 @@ def sign_tx(client, file, signing_mode, protocol_magic, network_id, testnet):
         protocol_magic = cardano.PROTOCOL_MAGICS["testnet"]
         network_id = cardano.NETWORK_IDS["testnet"]
 
-    signing_mode = messages.CardanoTxSigningMode.__members__[signing_mode]
+    signing_mode = messages.CardanoTxSigningMode[signing_mode]
```
matejcik, line 63: no, i mean, this whole line is superfluous because already `isinstance(signing_mode, messages.CardanoTxSigningMode)`

gabrielKerekes: Yeah, I just noticed and quickly force pushed the commit 🤦 Sorry.

gabrielKerekes:
In https://github.com/trezor/trezor-firmware/pull/1683/commits/8a9c7d9bd6a8ca55c6a824ee975c0932eac1bd46 and https://github.com/trezor/trezor-firmware/pull/1683/commits/615d88829c02a7935f3688e962decff19f05277c I renamed `CardanoSignTxInit.witnesses_count` to `witness_requests_count` and related variables similarly. `witness_request` is a much more suitable name as what's being sent aren't really witnesses yet.

core/src/apps/cardano/sign_tx.py
matejcik: The protection was not really strict before this, but at least there was guarantee that all change is going back into the same account. Now the only guarantee is that it matches `SCHEMA_ADDRESS` (fwiw you could basically replace the whole function by `return SCHEMA_ADDRESS.match(...)`).

This change basically throws out the account isolation model. Anything counts as "change output" as long as it's going into one of your accounts. Malicious software can deanonymize your spending by splitting up change into all your accounts, for example.

> not take inputs into consideration as that is not possible anymore

You can still match against witness paths, as proposed. (unless it is a known scenario to request witnesses from multiple accounts at the same time?)

> I would rather deal with those in a separate PR.

If it makes it easier for you, we can finalize this PR and commit it into a separate branch.

gabrielKerekes: Ok, I think I get your point.
> If it makes it easier for you, we can finalize this PR and commit it into a separate branch.

Let's do it here.

> if there are "internal" outputs (specified by address_params instead of address), at least one "external" output must exist, otherwise transaction is rejected at end of outputs stage

Why is this restriction needed?

I agree with the rest, although rather than failing, maybe showing a warning would be enough? Or failing/warning should depend on `safety_checks.is_strict()`.

matejcik: > > if there are "internal" outputs (specified by address_params instead of address), at least one "external" output must exist, otherwise transaction is rejected at end of outputs stage
> 
> Why is this restriction needed?

After thinking about it some more, it probably isn't. This is purely a UX problem: you'll be confirming a transaction with zero total, but that's basically correct information.

> rather than failing, maybe showing a warning would be enough? Or failing/warning should depend on `safety_checks.is_strict()`.

This should be a hard failure, because it's fully avoidable from the host side by sending a "correct" transaction: instead of multiple change outputs going to multiple accounts, specify them via addresses (that will be displayed for the user to confirm).

gabrielKerekes: Hm, it seems I'll have to consult the issue with IOHK - I'll get back to you ASAP.

gabrielKerekes: I'm not sure simply rejecting the transaction is a good idea. Until now the mixing of accounts was possible, although yes, the outputs would be shown. We might discreetly break some integrations if we decide to start failing from now on.

IMHO if we show a warning accompanying before outputs and witnesses which contain a different path than the first item containing a path (output or witness), it should be safe enough for the users. IMHO showing the outputs as address parameters + the address is better/less confusing than showing an encoded address.

> Malicious software can deanonymize your spending by splitting up change into all your accounts, for example.

Is deanonymization an issue if a malicious wallet can easily bulk export any of your public keys and do whatever with them? Or is bulk export limited to a single account?

Regardless of whether we end up failing or showing, we need to store the account of the first encountered path. Do you have a suggestion for storing this account?

One alternative is that a `transaction_account` variable would get passed around from the top level `sign_tx` function and it would also have to be returned from the functions which can modify it e.g. `process_outputs`, `process_certificates`, .... 

Another is to introduce a `CardanoTxContext`class which would store the `transaction_account` and also `network_id` and `protocol_magic`. The ugly thing is that `transaction_account` would be modified at some point during tx signing and not at the start.

Another would be to determine `transaction_account` in Connect and pass it into Trezor or let it be passed into Connect and forwarded to Trezor - which doesn't really seem right and it would perhaps unnecessarily change the protocol.

matejcik: > Is deanonymization an issue if a malicious wallet can easily bulk export any of your public keys and do whatever with them? Or is bulk export limited to a single account?

This way the Trezor itself signs a cryptographic proof that the accounts are linked; the best the user can claim is that the accounts are not theirs, but they deliberately sent them money. With bulk export, the attacker has no way of proving to a 3rd party that the XPUBs come from the same device.

Another issue which does not strictly fall under "security" is that an attacker can basically freely move funds between user's accounts.

> IMHO if we show a warning accompanying before outputs and witnesses which contain a different path

Mixing witness accounts, if accompanied by a warning, is OK (there's still some space left for a deanonymization attack, but at that point it's mostly theoretical) --- as long as _no outputs were hidden_ previously. The logic goes like this:
(a) either you're spending from a single account and sending back change to that same account, in which case change detection kicks in and the change output is not shown. (FWIW in Bitcoin we also show a warning when more than two change outputs are detected, because that indicates suspicious activity)
(b) or you're spending from multiple sources and/or sending to multiple accounts, in which case it's impossible to determine which outputs are "change", and all outputs must be shown.

I don't think "warning: some outputs were not displayed" is something that makes any sort of sense, so hard failure is the only option here.

We previously checked all outputs against _first input_, which at least ensures that the change is not scattered, but still doesn't necessarily display the right thing. I consider it a security-adjacent bug. If any existing integrations relied on this bug, I'm OK with breaking them by fixing the bug.

--- 

> Regardless of whether we end up failing or showing, we need to store the account of the first encountered path. Do you have a suggestion for storing this account?

Have a look at Bitcoin's matcher: https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/bitcoin/sign_tx/matchcheck.py
It's not a full "transaction context", you would probably only need to pass it into "process outputs" and "process witnesses" and not any deeper, from a quick look.

(if applicable, you could copy-paste the code and we'd resolve extracting back common functionality in some other PR)

If we need to pass this deeper somewhere, i'd consider making the TxContext and placing a matcher as one of its properties.

gabrielKerekes: Alright, I'll implement the matchchecker today and so transaction signing would fail if multiple accounts are used. I think we also need to add certificates and withdrawals to the matchchecker though, because they require witnesses which are included along the other witness requests and we have no way to distinguish which came from where. This means that basically all the paths (change output payment path, certificate path, withdrawal path, input/witness path) in a transaction will have to belong to a single account. Do you agree?

tsusanka:
JFYI we'll review the Connect PR https://github.com/trezor/connect/pull/841 after this is merged and trezor-common is bumped (we'll do that).

gabrielKerekes:
> JFYI we'll review the Connect PR trezor/connect#841 after this is merged and trezor-common is bumped (we'll do that).

Good, thank you. I think I'll also have to go through the Connect PR once more after this is merged to make sure everything's good. At least some renaming will definitely be necessary.

core/src/apps/cardano/sign_tx.py
matejcik: I'll have to go with your opinion here because i don't really know what certificates and withdrawals _do_ -- but yes, presumably the same logic (single account yes/no) would apply too.

gabrielKerekes: I've implemented the `AccountPathChecker`. All relevant paths are being included in it including pool owners as the witness for that is also being passed in from the outside. I had to change `SCHEMA_ADDRESS` to `SCHEMA_PAYMENT` in order to simplify the validation for hiding outputs (as I mentioned before, this has already been refactored in the mutli-sig PR anyways so I wouldn't want to further modify it here if not necessary). I had to modify some tests which didn't follow the single account rule and I added more tests which verify that transactions with mixed accounts indeed do fail.

matejcik: ACK on the matcher usage, but I believe the "Byron to Shelley" scenario must be supported -- isn't that what a past issue was about? Trezor wrongly showing a change output in case Byron and Shelley inputs are mixed?

At least some Cardano wallets consider Byron and Shelley accounts to be the same account. (I presume it makes more sense in Cardano than in Bitcoin, if Shelley is basically a "rolling upgrade"? But you'd need to confirm with IOHK, i'm the less knowledgeable party here)

So presumably to fix that, instead of picking `to_account_path` as the matcher value, you'd just pick `path[ACCOUNT_PATH_INDEX]` ?

common/tests/fixtures/cardano/sign_tx.json
```diff
@@ -114,36 +138,62 @@
...
-              "prev_index": 0
+            "path": "m/44'/1815'/0'/0/2",
+            "prev_hash": "a34dc95d806a3b206aab5e0c2aaa5ff0704f84868fe65793053f6ae9a7970979",
+            "prev_index": 0
           },
           {
-              "path": "m/1852'/1815'/0'/0/0",
-              "prev_hash": "3b40265111d8bb3c3c608d95b3a0bf83461ace32d79336579a1939b3aad1c0b7",
-              "prev_index": 0
+            "path": "m/44'/1815'/0'/0/3",
```
matejcik, line 151: this `1852' -> 44'` should be reverted

common/tests/fixtures/cardano/sign_tx.json
```diff
@@ -114,36 +138,62 @@
...
+            "prev_hash": "a34dc95d806a3b206aab5e0c2aaa5ff0704f84868fe65793053f6ae9a7970979",
+            "prev_index": 0
           },
           {
-              "path": "m/1852'/1815'/0'/0/5",
-              "prev_hash": "33ad5e2a8f298053da804c30c9f72836bfac0a58a30aef2ff87656418b01f70b",
-              "prev_index": 0
+            "path": "m/44'/1815'/0'/0/4",
```
matejcik, line 156: this `1852' -> 44'` should be reverted

common/tests/fixtures/cardano/sign_tx.json
```diff
@@ -1038,59 +1038,6 @@
         ]
       }
     },
-    {
-      "description": "Byron to Shelley transfer",
```
matejcik, line 817: this testcase should be kept and should succeed

common/tests/fixtures/cardano/sign_tx.json
```diff
@@ -1232,11 +1179,6 @@
             "path": "m/1852'/1815'/0'/0/0",
             "amount": "7120787"
           },
-          {
-            "addressType": 8,
-            "path": "m/44'/1815'/0'/0/1",
```
matejcik: not sure about this one but presumably should also be kept?

matejcik (commented):
marking the places where the byron/shelley distinction should still pass

core/src/apps/cardano/sign_tx.py
gabrielKerekes: We've confirmed with IOHK that Daedalus and Yoroi that they don't combine the two wallets. I.e. if there are Byron inputs then Shelley outputs are external e.g. migrating the whole wallet from Byron to Shelley. However, it seems Adalite can combine them in some instances. But IMHO it doesn't matter if we're talking about `44'/1815'/0'` vs `1852'/1815'/0'` or `1852'/1815'/0'` vs `1852'/1815'/1'`. The same issues arise. Since Daedalus and Yoroi don't mix the two wallets, the user wouldn't see the funds potentially sent to the `44'/...` (hidden) change output. We'll discuss this internally ASAP. Sorry this is taking so long.

matejcik: > he same issues arise. Since Daedalus and Yoroi don't mix the two wallets, the user wouldn't see the funds potentially sent to the `44'/...` (hidden) change output.

This is a good point. Thanks for confirming. I'm OK with keeping the implementation as-is if you agree.

gabrielKerekes:
We decided to keep the possibility of Byron to Shelley migration. However it's rather constrained as both Byron and Shelley accounts need to be 0 so it should be safe. https://github.com/trezor/trezor-firmware/pull/1683/commits/e630dc2ae01dd8d3b74b3126076cdee61a57b72a

I also discovered that it was possible to include a pool registration certificate in an `ORDINARY_TRANSACTION` and the call would crash. It was also possible to include a different certificate in a `POOL_REGISTRATION_AS_OWNER` transaction. I've fixed that in https://github.com/trezor/trezor-firmware/pull/1683/commits/f4f4729e943b881bd0d9129f3f41517f1805f43f and updated the tests for this in https://github.com/trezor/trezor-firmware/pull/1683/commits/ab1ed990958f56dad04548909cd15ca974d23625.

I've updated the Cardano README - https://github.com/trezor/trezor-firmware/pull/1683/commits/ddcc0e3ecf88aa3e1fa993242048194235ecc699.

I've added changelog entries - https://github.com/trezor/trezor-firmware/pull/1683/commits/0b131026dd1971f38c1d1e20369af60f170d0f29

And while waiting for the Byron/Shelley discussion to happen I already rebased and squashed everything on top of master - I figured it would be OK since the previous stuff was already reviewed.

When it comes to the UI test fixtures I'm again having trouble generating them, I'll investigate a little. In nix-shell and poetry shell I run `pytest tests/device_tests -m cardano --ui=record` as you've previously advised. This however asks for an emulator, which it, as you previously mentioned, probably shouldn't. So I run the emulator with animations enabled (or should they be disabled?) but all the fixtures change anyways.

matejcik:
> This however asks for an emulator, which it, as you previously mentioned, probably shouldn't.

I don't think I ever said that :) If you run via pytest, you'll need to start the emulator separately. Animations must be *disabled*, that is probably why every fixture changes.

gabrielKerekes:
> I don't think I ever said that :) If you run via pytest, you'll need to start the emulator separately.

Yes, you are right, in that case we were talking about `make test_emu_ui_record`. Sorry.

However, I tried with animations disabled and the results are still different. I thinks it's a macOS related issue. It seems the recorded images are encoded differently on MacOS as one of the images is only 15kB on mac vs. 22kB on Linux and thus the hashes differ. I'll be able to reliably generate the fixtures later today on a Linux machine.

matejcik:
Yes, there's a difference between what Mac / Linux generates, we already ran into it somewhere. I created an issue #1769 

for now if you can regenerate on Linux that would be nice, otherwise I'll regenerate locally tomorrow

matejcik:
new commits LGTM. please squash&rebase again when you're ready

gabrielKerekes:
Rebased. I will update the UI fixtures later today/early tomorrow. Changelog checks are failing, I guess I'll add a changelog to the Python folder also.

gabrielKerekes:
UI Fixtures updated https://github.com/trezor/trezor-firmware/pull/1683/commits/584fec718b18ec9b9c5187c9167f7b9744723e6b - sign_tx results have almost all changed because total transaction amount is not shown to the user anymore.

I've also added a changelog entry for Python - https://github.com/trezor/trezor-firmware/pull/1683/commits/d131ae828e177251c86c75a004685eb1a8ac4b35. I've force pushed the changes.

matejcik:
and we're all green. merged, thanks!

gabrielKerekes:
Awesome! @matejcik thank you for all the help and suggestions!

gabrielKerekes:
@matejcik please, just to make sure, the upcoming FW release is happening on September 8th, right?

matejcik:
that is the plan, yes

PR #1688, opened 2021-07-06
chore(ci): update alpine-linux to 3.14.0 and nix to 2.3.14

prusnak:


tsusanka (approved):

matejcik:
style check is failing. did clang-format version change?

prusnak:
> clang-format version change?

I think it did, although I don't understand how this could happen.

mmilata:
It's failing in master too, fix: #1689.

mmilata (commented):
There's another `ALPINE_VERSION` set inside `ci/environment.yml`. I wonder if we can drop it from there and download the minirootfs in the Dockerfile?

prusnak:
> I wonder if we can drop it from there and download the minirootfs in the Dockerfile?

Might be possible too, but for now I just changed the `ci/environment.yml` too.

PR #1690, opened 2021-07-07
layouts: basic T1 support

mmilata:
homescreen signing-t1 wipe

rust port status:
* most of bitcoin-only fw layouts implemented
* no setup/recovery, no pagination, no animations, looks ugly
* works on hardware, signing sometimes OOMs


mmilata:
Rewrote to use Rust UI infrastructure. Requires #1892.

matejcik:
is this PR still relevant?

mmilata:
IIRC there's a bunch of screens implemented (in outdated way) that are not in master. Let me close the PR and keep the branch for some more time in case this is useful.

PR #1695, opened 2021-07-09
update nixpkgs, nanopb

prusnak:


matejcik (approved):
utACK

mmilata:
Rust build is failing in CI, investigating.

mmilata (approved):
Added fixup that makes CI pass. Went through rust release notes and there doesn't seem to be anything that affects us.

PR #1704, opened 2021-07-12
refactor(core/ethereum): improve API of the rlp module

matejcik:
- improve length calculation so that we don't need to actually encode to get the length
- instead of returning bytes, directly write into a provided Writer

the test vectors in `test_trezor.crypto.rlp.py`, and the Ethereum device tests, show that the new implementation produces results identical to the old one


core/src/trezor/crypto/rlp.py
```diff
@@ -1,53 +1,94 @@
...
+
+STRING_HEADER_BYTE = const(0x80)
+LIST_HEADER_BYTE = const(0xC0)
+
+
+def _byte_size(x: int) -> int:
+    if x < 0:
+        raise ValueError  # only unsigned ints are supported
+    for exp in range(64):
+        if x < 0xFF ** exp:
```
prusnak, line 18: This seems weird. Maybe you meant:

```suggestion
        if x < 0x100 ** exp:
```

matejcik: > Also this function returns 0 for x=0. Not sure if that is the intention.

This is intentional, yes

But good point about the equality test. I'll need to add an edge value to the testcases first as it's apparently not covered.

matejcik: 3d4d186

core/tests/test_trezor.crypto.rlp.py
```diff
@@ -27,6 +29,12 @@ class TestCryptoRlp(unittest.TestCase):
          '7f'),
         (128,
          '8180'),
+        (254,
+         '81fe'),
+        (255,
+         '81ff'),
+        (256,
+         '820100'),
         (1000,
```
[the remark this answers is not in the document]
matejcik: 9f35f44

core/src/trezor/crypto/rlp.py
```diff
@@ -1,53 +1,94 @@
...
-    elif l < 256 ** 8:
-        bl = int_to_bytes(l)
-        return bytes([len(bl) + offset + 55]) + bl
+from micropython import const
+
+if False:
+    from typing import Union
+    from trezor.utils import Writer
+
+    RLPItem = Union["RLPItem", bytes, int]
```
mmilata, line 7: What does the recursion do here?

Also, should there be bytearray since it's checked against in the code?

mmilata (approved):
One question, otherwise looks better and more efficient than the original version.

core/src/trezor/crypto/rlp.py
matejcik: Ah, I meant to run this though mypy and forgot. Good catch, this is supposed to be something like `Union[list["RLPItem"], bytes, int]`

(also I suspect that mypy won't actually allow a recursive type)

matejcik: 4a009c8

PR #1706, opened 2021-07-14
Add support for recursive protobuf messages

gabrielKerekes:
Hello, I'm currently implementing the next Cardano updates. As a part of this I've come upon an issue I was able to solve but since it touches `pb2py` and the mako templates I would appreciate your feedback on the proposed solution or perhaps you will be able to come up with a more elegant solution.

The issue is that I'm adding a recursive protobuf message to `messages-cardano.proto`:
```
message CardanoScript {
    required CardanoScriptType type = 1;
    repeated CardanoScript scripts = 2;

    optional bytes key_hash = 3;
    repeated uint32 key_path = 4;
    optional uint32 required_signatures_count = 5;
    optional uint64 invalid_before = 6;
    optional uint64 invalid_hereafter = 7;
}
```
Without any changes `pb2py` gets stuck in an infinte loop - I've fixed that [here](https://github.com/trezor/trezor-firmware/pull/1706/files#diff-c20641a3c126c8673a1b3524439ddcb77cb667475f6381820eb3cb4ec567c314R402). Then when the message is already generated Python and mypy don't recognize the type of the recursive fields since it's nested in the class. So I've updated the templates to wrap recursive types in double quotes. I couldn't do the same thing for the types in the `FIELDS` field since the type of the field would then get interpreted as a string in `trezorlib/protobuf.py:L163`. That's why I've moved the `FIELDS` declaration after the recursive class declaration in the trezorlib template so that the recursive type would be recognized.

### DIFF
--- a/core/src/trezor/_proto_messages.mako
+++ b/core/src/trezor/_proto_messages.mako
@@ -41,20 +41,20 @@ def member_type(field):
     class ${message.name}(protobuf.MessageType):
 % if message.fields:
 % for field in message.fields:
-        ${field.name}: ${member_type(field)}
+        ${field.name}: "${member_type(field)}"
 % endfor
 
         def __init__(
             self,
             *,
 % for field in required_fields:
-            ${field.name}: ${field.python_type},
+            ${field.name}: "${field.python_type}",
 % endfor
 % for field in repeated_fields:
-            ${field.name}: list[${field.python_type}] | None = None,
+            ${field.name}: "list[${field.python_type}] | None" = None,
 % endfor
 % for field in optional_fields:
-            ${field.name}: ${field.python_type} | None = None,
+            ${field.name}: "${field.python_type} | None" = None,
 % endfor
         ) -> None:
             pass
--- a/python/src/trezorlib/_proto_messages.mako
+++ b/python/src/trezorlib/_proto_messages.mako
@@ -21,22 +21,13 @@ class ${enum.name}(IntEnum):
 required_fields = [f for f in message.fields if f.required]
 repeated_fields = [f for f in message.fields if f.repeated]
 optional_fields = [f for f in message.fields if f.optional]
-
-
-def type_name(field):
-    if field.type_object is not None:
-        return field.type_name
-    else:
-        return '"' + field.type_name + '"'
-
-
 %>\
 class ${message.name}(protobuf.MessageType):
     MESSAGE_WIRE_TYPE = ${message.wire_type}
 % if message.fields:
     FIELDS = {
 % for field in message.fields:
-        ${field.number}: protobuf.Field("${field.name}", ${type_name(field)}, repeated=${field.repeated}, required=${field.required}),
+        ${field.number}: protobuf.Field("${field.name}", "${field.type_name}", repeated=${field.repeated}, required=${field.required}),
 % endfor
     }
 
@@ -44,13 +35,13 @@ class ${message.name}(protobuf.MessageType):
         self,
         *,
 % for field in required_fields:
-        ${field.name}: ${field.python_type},
+        ${field.name}: "${field.python_type}",
 % endfor
 % for field in repeated_fields:
-        ${field.name}: Optional[List[${field.python_type}]] = None,
+        ${field.name}: Optional[List["${field.python_type}"]] = None,
 % endfor
 % for field in optional_fields:
-        ${field.name}: Optional[${field.python_type}] = ${field.default_value_repr},
+        ${field.name}: Optional["${field.python_type}"] = ${field.default_value_repr},
 % endfor
     ) -> None:
 % for field in repeated_fields:
--- a/python/src/trezorlib/protobuf.py
+++ b/python/src/trezorlib/protobuf.py
@@ -144,7 +144,7 @@ REQUIRED_FIELD_PLACEHOLDER = object()
 @attr.s(auto_attribs=True)
 class Field:
     name: str
-    type: Union[str, "MessageType", IntEnum]
+    type: str
     repeated: bool = attr.ib(default=False)
     required: bool = attr.ib(default=False)
     default: object = attr.ib(default=None)
@@ -154,10 +154,11 @@ class Field:
         if self.type in WIRE_TYPES:
             return WIRE_TYPES[self.type]
 
-        if safe_issubclass(self.type, MessageType):
+        field_type_object = get_field_type_object(self)
+        if safe_issubclass(field_type_object, MessageType):
             return WIRE_TYPE_LENGTH
 
-        if safe_issubclass(self.type, IntEnum):
+        if safe_issubclass(field_type_object, IntEnum):
             return WIRE_TYPE_INT
 
         raise ValueError(f"Unrecognized type for field {self.name}")
@@ -269,6 +270,15 @@ class CountingWriter:
         return nwritten
 
 
+def get_field_type_object(field: Field) -> Optional[type]:
+    from . import messages
+
+    field_type_object = getattr(messages, field.type, None)
+    if not safe_issubclass(field_type_object, (IntEnum, MessageType)):
+        return None
+    return field_type_object
+
+
 def decode_packed_array_field(field: Field, reader: Reader) -> List[Any]:
     assert field.repeated, "Not decoding packed array into non-repeated field"
     length = load_uvarint(reader)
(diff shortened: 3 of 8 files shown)


matejcik:
I am not sure that we even want to support this.
What is your usecase for a recursive message? Couldn't the same thing be solved by Trezor requesting the sub-field separately? (this would also be much better in terms of memory management)

gabrielKerekes:
> I am not sure that we even want to support this.

That's what I was afraid of.

> What is your usecase for a recursive message?

We're adding a `get-script-hash-call` which should verify the hash of a [Cardano native script](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L228). The whole script needs to be sent to Trezor to be shown to the user. The input to Trezor looks something like this:

```
{
  "type": 1,
  "scripts": [
    {
      "type": 0,
      "key_hash": "c4b9265645fde9536c0795adbcc5291767a0c61fd62448341d7e0386"
    },
    {
      "type": 0,
      "key_path": "m/1854'/1815'/0'/0/0"
    },
    {
      "type": 2,
      "scripts": [
        {
          "type": 0,
          "key_path": "m/1854'/1815'/0'/0/0"
        },
        {
          "type": 0,
          "key_hash": "0241f2d196f52a92fbd2183d03b370c30b6960cfdeae364ffabac889"
        }
      ]
    },
    {
      "type": 3,
      "required_signatures_count": 2,
      "scripts": [
        {
          "type": 0,
          "key_path": "m/1854'/1815'/0'/0/0"
        },
        {
          "type": 0,
          "key_hash": "0241f2d196f52a92fbd2183d03b370c30b6960cfdeae364ffabac889"
        },
        {
          "type": 0,
          "key_hash": "cecb1d427c4ae436d28cc0f8ae9bb37501a5b77bcc64cd1693e9ae20"
        }
      ]
    },
    {
      "type": 4,
      "invalid_before": 100
    },
    {
      "type": 5,
      "invalid_hereafter": 200
    }
  ]
}
```
The message can be seen in the PR.

> Couldn't the same thing be solved by Trezor requesting the sub-field separately?

Yes, I think it could. It would just unnecessarily complicate the call - a rolling hash would need to be calculated. Memory shouldn't be too much of a concern in this case as the scripts shouldn't be too large. So I was hoping we could avoid this, but I agree it's definitively a less invasive solution.

matejcik:
I see. We'll need to discuss this internally.

> The message can be seen in the PR.

which one?

...if we did allow recursive messages, I'd do it a bit differently:
- provide _all_ typehints as strings. it's legal and simplifies the template code a lot.
- modify `protobuf.py` to always accept string for the type name, fetch the classes/enums with getattr. (I was actually considering that while writing the current version, it would simplify some of the code that deals with them.) That way there's no need to move the `FIELDS` declaration out of the class, and it allows dependency loops in addition to simple recursion.
- haven't looked at `pb2py` yet, but we might even be able to get rid of the dependency ordering.

matejcik:
> * provide _all_ typehints as strings. it's legal and simplifies the template code a lot.

(definitely all class/enum names, possibly also simple type names, whichever makes for simpler template code)

gabrielKerekes:
> which one?

Sorry, I meant this PR and [this protobuf message](https://github.com/trezor/trezor-firmware/pull/1706/files#diff-18106a190950e664af1b274351237e61e56e58186b7f2671a56e028e2aefcf3eR64).

> ...if we did allow recursive messages, I'd do it a bit differently:

Sounds good!

Please let me know about the next steps after the internal discussion.

matejcik:
so the result is, simply allowing support for recursive messages in the tooling is OK. we'll need to be on the lookout for actual usages, this one seems OK at first look.

so feel free to proceed

gabrielKerekes:
Just to be sure - we can proceed with the solution proposed in this PR, right?

matejcik:
in general yes, but see my comment https://github.com/trezor/trezor-firmware/pull/1706#issuecomment-880536577 for an outline of a cleaner solution.

python/src/trezorlib/_proto_messages.mako
```diff
@@ -22,7 +22,6 @@ required_fields = [f for f in message.fields if f.required]
 repeated_fields = [f for f in message.fields if f.repeated]
 optional_fields = [f for f in message.fields if f.optional]
 
-
 def type_name(field):
```
matejcik, line 25: you can drop this function altogether

core/src/trezor/_proto_messages.mako
```diff
@@ -30,12 +30,12 @@ optional_fields = [f for f in message.fields if f.optional]
 
 def member_type(field):
     if field.required:
-        return field.python_type
+        return f"\"{field.python_type}\""
```
matejcik, line 33: instead, put the quotes around the whole thing at the usage site. `list["SomeType"]` is equivalent to `"list[SomeType]"`

Speaking of, for firmware we're in python 3.10 mode so I'm wondering if we even need the quotes here at all. I didn't try that though.

python/src/trezorlib/protobuf.py
```diff
@@ -420,16 +430,17 @@ def dump_message(writer: Writer, msg: MessageType) -> None:
...
                 dump_uvarint(writer, counter.size)
                 dump_message(writer, svalue)
 
-            elif safe_issubclass(field.type, IntEnum):
-                if svalue not in field.type.__members__.values():
+            elif safe_issubclass(field_type_object, IntEnum):
+                if svalue not in field_type_object.__members__.values():
                     raise ValueError(
-                        f"Value {svalue} in field {field.name} unknown for {field.type.__name__}"
+                        f"Value {svalue} in field {field.name} unknown for {field_type_object.__name__}"
```
matejcik, line 443: `unknown for {field.type}` because that's now guaranteed to be a string

python/src/trezorlib/protobuf.py
```diff
@@ -269,6 +270,12 @@ def write(self, buf: bytes) -> int:
         return nwritten
 
 
+def get_field_type_object(field: Field) -> Optional[type]:
+    from . import messages
+
+    return getattr(messages, field.type, None)
```
matejcik, line 276: there should be ` if not issubclass(thing, (IntEnum, MessageType)): return None` here to avoid accidentally fetching a builtin or something like that

matejcik (commented):
> I’m also not sure if/how the rest of the code could now be simplified.

I'm not sure either honestly. I hoped that we could at least avoid `safe_issubclass`, but that's not gonna work either. So I guess it makes sense how you wrote it.

core/src/trezor/_proto_messages.mako
gabrielKerekes: I tried leaving the quotes out however mypy (and also PyCharm) complains then - I'm not sure if mypy is still on old Python or if this isn't supported even in 3.10 though.

matejcik:
this looks good now. feel free to rebase / convert to full PR / merge with something else / whatever you originally intended :)

gabrielKerekes:
I've rebased. Converted to normal PR. I think it would be best to merge this now as it has already been reviewed. However I noticed that in the [trezorlib _proto_messages.mako](https://github.com/trezor/trezor-firmware/pull/1706/files#diff-43677503b596b62a69ff95885a41b51c55b2a3ea5f94233cac7ab4223f4b60adR41) not the whole types are in quotes. Should I update it so that it's consistent with the other template?

matejcik:
I'd leave it as-is, there is no reason to touch it.

For posterity: I just now realized that this couples `trezorlib.protobuf` with `trezorlib.messages`, IOW, you can't use custom protobuf classes from anywhere else.

I'm leaving that as-is also, it will be relatively easy to allow passing actual type objects in `Field` if we ever need it.

matejcik:
> For posterity: I just now realized that this couples `trezorlib.protobuf` with `trezorlib.messages`, IOW, you can't use custom protobuf classes from anywhere else.

Aaaannnd this breaks python tests (`cd python; pytest tests`)

Please fix that. Either by re-allowing type objects directly, or (probably simpler) by stuffing the objects into `trezorlib.messages` in the test case:
```
from trezorlib import messages

messages.SimpleEnum = SimpleEnum
# ...
```

gabrielKerekes:
Sorry it took so long - I've updated the tests to include the messages in `messages` module. I've also added a test case for a recursive message.

gabrielKerekes:
I also added a changelog entry - https://github.com/trezor/trezor-firmware/pull/1706/commits/38c05d02df70883e0a3f58c40807d8ae0dd37f8e.

matejcik:
I don't think these changelog entries are needed. I'll need to check whether the changelog checker is broken, it should be ignoring generated files.

matejcik:
but the templates are also changed, which might be appropriate to trigger... hmmm

in any case, I don't think changelogs are appropriate here. For trezorlib, this is already subsumed in "refactor of protobuf codec". For core, there's essentially no change except changing typehints to strings, and that's invisible to both users and developers (and so needless noise in the changelog).

This would be appropriate in a `pb2py` changelog, but we don't have that so :woman_shrugging:

Please remove the changelogs. You can add the string `[no changelog]` to the commit message to silence the failing check.

gabrielKerekes:
I've removed the changelogs and I've also rebased on top of the current master.

matejcik:
rebased on current master (with conflicts resolved) at 4f7c6b3586683cfc57b02076621b54e8e7f163f1

matejcik:
pipeline: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/348413867

matejcik:
merged as 4f7c6b3 via #1750

Issue #810, opened 2020-01-27
Evaluate Python 3.5 compatibility of trezorlib


matejcik:
Construct 2.10 has dropped Python 3.5 compatibility now, which causes some of our CI runners to fail the python test, which still tests Python 3.5.

We've been using 3.6 features in the test suite and elsewhere for quite some time now, but have been keeping 3.5 compatibility in trezorlib because it didn't cost us anything. That might be changing now.

We could fix Construct on version 2.9 and postpone the problem for a couple more months, but perhaps that is not worth it.

Debian stable version now has Python 3.7. We need to evaluate which systems could still be on 3.5 and whether it's worth keeping compatibility, or refering them to an older version of trezorlib.

If we decide to drop compatibility, the first (and the only required) step is dropping `py35` target from tox config. An optional second step is using f-strings everywhere, and third step is variable type annotations.

matejcik:
the problem that is breaking CI is actually a Construct packaging bug https://github.com/construct/construct/pull/834
Hopefully this will be resolved on upstream side, so we will not need to pin the version.

after a review on Repology, it turns out that pretty much all relevant distros already have Python 3.6 or 3.7. But Debian Stretch (oldstable) still has 3.5, and this is presumably a rather large install base, so I'm choosing to stay on 3.5 as well for now.

Dropping priority and adding a "deferred" flag, I'm sure we will come back to this sooner or later :)

matejcik:
#1266 breaks Py3.5 compatibility again: in `def foo(*, a, b,)` the trailing comma is a syntax error

we could resolve this if we want -- in any case, this issue must be resolved one way or the other before trezorlib 0.13 is released

PR #1707, opened 2021-07-14
keyctl-proxy fixes

prusnak:
* don't use format strings in keyctl-proxy
* fix typo in keyctl-proxy help
* rename fw_or_type to image_type in keyctl-proxy to avoid confusion

### DIFF
--- a/core/.changelog.d/1707.fixed
+++ b/core/.changelog.d/1707.fixed
@@ -1,0 +1,1 @@
+Don't use format strings in keyctl-proxy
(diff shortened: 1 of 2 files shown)


matejcik:
huh, so with this change keyctl proxy still works on 3.5? despite trezorlib being gradually upgraded to 3.6 ?

prusnak:
> huh, so with this change keyctl proxy still works on 3.5? despite trezorlib being gradually upgraded to 3.6 ?

Didn't go as far as to check this :-/

matejcik:
i guess it should be usable with an older version of trezorlib in any case

prusnak:
> i guess it should be usable with an older version of trezorlib in any case

What's the latest version of trezorlib that works with Python 3.5 ?

matejcik:
the currently released 0.12.2 works with py3.5
it's only 0.13 in master that broke compatibility

matejcik:
#810

prusnak:
That's great news, thanks!

matejcik (approved):

Issue #1656, opened 2021-06-08
Support Taproot


tsusanka:
Support Taproot in firmware.

Tasks:
- [x] Make secp256k1-zkp default:
  - https://github.com/trezor/trezor-firmware/issues/1534
  - PR https://github.com/trezor/trezor-firmware/pull/1678
- [x] Support new transaction format
- [x] Decide BIP-44 paths
- [x] Implement Bech32m address format (bip-0350)

Docs:
- [BIP 340: Schnorr Signatures](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki)
- [BIP 341: Taproot: SegWit version 1 spending rules](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)
- [BIP 342: Validation of Taproot Scripts](https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki)
- [BIP 350: Bech32m format for v1+ witness addresses](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki)

prusnak:
FTR:  Taproot softfork will activate in November.

tsusanka:
@prusnak could you fill in the last bullet point please? I did catch it properly what was it.

prusnak:
> @prusnak could you fill in the last bullet point please? I did not catch it properly what was it.

Done

andrewkozlik:
Additional tasks:
- [ ] In `SignTx` implement the BIP-341 "common signature message" computation.
- [ ] In `SignTx` disable streaming of previous transactions if at least one input is taproot.
- [ ] Implement taproot support in `scripts.py`, i.e. `output_script_native_p2tr()` , `witness_p2tr()`, `parse_witness_p2tr()` etc.
- [ ] Decide how best to implement multisig for taproot in Trezor
    - MuSig (interactive signing) vs. unagreggated multisig (non-interactive).
    - Consider that taproot allows encoding more complicated schemes, e.g. with four keys A, B, C and D, we can have something like "A or (B and C) or (B and D)". This would require modifications to `MultisigRedeemScriptType`.
- [ ] Implement multisig scripts for taproot. 
- [ ] Use new Bech32m format in `GetAddress`.
- [ ] Support taproot scripts and Schnorr signatures in `SignatureVerifier`.
- [ ] Support taproot scripts and Schnorr signatures in `GetOwnershipProof` and in `verify_nonownership()`.
[...]

PR #1710, opened 2021-07-15
Taproot: add support for Taproot descriptors

prusnak:
This is the first of long series of PRs that will implement Taproot on Trezor. Part of https://github.com/trezor/trezor-firmware/issues/1656

This one takes care of implementing new enum fields to Input and Output scripts and adds possibility to generate a Taproot account descriptor. Taproot does not use new prefix for XPUBs and recommends using descriptors instead.


matejcik (approved):
utACK

PR #1711, opened 2021-07-15
Merge 21.07 release branch

mmilata:


prusnak:
Can't we just merge the branch instead of rebasing the change?

mmilata:
I tried that previous release but the repo is configured to enforce linear history iirc.

prusnak:
> I tried that previous release but the repo is configured to enforce linear history iirc.

I think admins can override this rule on merge. I'd prefer merge commit in this case, to make it obvious we are just merging the release branch.

tsusanka:
I agree it would be simpler. I have turned it off for now could you do that @mmilata please? Will turn it on then.

tsusanka (approved):
I confirm all commits are present.

PR #1716, opened 2021-07-16
feat(core): in debug mode, dump contents of received protobuf messages

matejcik:
Useful for debugging, but makes the log much chattier.

Probably the usefulness outweighs the chattiness.

image

### DIFF
--- a/core/src/trezor/utils.py
+++ b/core/src/trezor/utils.py
@@ -35,6 +35,8 @@ if False:
         Set,
     )
 
+    from trezor.protobuf import MessageType
+
 
 def unimport_begin() -> Set[str]:
     return set(sys.modules)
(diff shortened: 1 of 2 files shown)


prusnak (approved):
utACK

matejcik:
there's some bugs though

matejcik:
and hopefully fixed

matejcik:
note: the [no changelog] mark is way over git's first line limit.

maybe we could allow adding "no changelog" to the message contents too?

mmilata:
@matejcik it's supposed to work in the commit message contents as well

mmilata (commented):
Changes look good. If the chattiness becomes too annoying perhaps we can introduce some kind of switch. I'm mildly triggered by not dumping also the sent messages but I suppose these are usually not as interesting. Can/should the code handle repeated (list) fields?

matejcik:
> Can/should the code handle repeated (list) fields?

absolutely should, i'll go fix it

> If the chattiness becomes too annoying perhaps we can introduce some kind of switch.

i originally intended to do that (adding another envvar reader to `trezor.utils` seems the way to go), but tbqh the log is already very chatty and of dubious general usefulness, this doesn't seem to change the situation too much.
What I'm most worried about is performance of debug fw on hardware. Do you think maybe we should stick a "if emulator" there?

> not dumping also the sent messages

with this code in place we can easily add that at any time we need it

matejcik:
> Can/should the code handle repeated (list) fields?

image

mmilata (approved):
Perhaps `if EMULATOR` is not a bad idea, seeing as the messages are converted to one long string.

matejcik:
added `and utils.EMULATOR` in https://github.com/trezor/trezor-firmware/pull/1716/files#diff-8be600ec13beb037f92f905b366e48d31a6f3f2bef7f7a0ef410cb790b2900b2R105

waiting for CI to go green

PR #1717, opened 2021-07-17
Taproot: add taproot field to coin definitions

prusnak:
+ add taproot field to coin definitions
+ enable for Bitcoin Testnet and Bitcoin Regtest


matejcik (approved):
utACK

PR #1723, opened 2021-07-19
feat(core/emulator): support protobuf messages in memory dumps

mmilata:
Followup of #1557. Not sure whether shortval is the right place for the message name.

### DIFF
--- a/core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
+++ b/core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
@@ -33,6 +33,7 @@
 #include "py/objtype.h"
 
 #include "embed/extmod/trezorobj.h"
+#include "embed/rust/librust.h"
 #include "embed/trezorhal/usb.h"
 
 #include <string.h>
@@ -301,6 +302,11 @@ typedef struct _mp_obj_HID_t {
   usb_hid_info_t info;
 } mp_obj_HID_t;
 
+typedef struct _mp_obj_protomsg_t {
+  mp_obj_base_t base;
+  mp_map_t map;
+} mp_obj_protomsg_t;
+
 void dump_bound_method(FILE *out, const mp_obj_bound_meth_t *meth) {
   print_type(out, "method", NULL, meth, false);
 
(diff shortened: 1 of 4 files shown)


core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
```diff
@@ -487,6 +493,14 @@ void dump_trezor_vcp(FILE *out, const mp_obj_VCP_t *vcp) {
   fprintf(out, ",\n");
 }
 
+void dump_protomsg(FILE *out, const mp_obj_protomsg_t *value) {
+  print_type(out, "protomsg", NULL, value, false);
+  fprintf(out, ",\n\"shortval\": ");
+  dump_short(out, protobuf_debug_name(value));
```
matejcik, line 499: i would think something like this could work, instead of defining a function?
```
mp_obj_t name[2] = {MP_OBJ_NULL, MP_OBJ_NULL};
mp_obj_type_t *type = mp_obj_get_type(value); // or protobuf_debug_msgtype();
type->attr(value, MP_QSTR_MESSAGE_NAME, name);
dump_short(out, name[0]);
```

core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
```diff
@@ -487,6 +493,14 @@ void dump_trezor_vcp(FILE *out, const mp_obj_VCP_t *vcp) {
   fprintf(out, ",\n");
 }
 
+void dump_protomsg(FILE *out, const mp_obj_protomsg_t *value) {
+  print_type(out, "protomsg", NULL, value, false);
+  fprintf(out, ",\n\"shortval\": ");
```
matejcik, line 498: `print_type` will print shortval by itself, so this will create a duplicate key in json.
json doesn't mind, I think, but still not great.

shortval is probably not best for this anyway. I'd stick it in `"message_name"`, i don't see any reason to match this attribute to anything else?

matejcik:
looks good overall, please also add support for `MsgDef`

core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
mmilata: nice trick, fixed in 88d5f81

core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
mmilata: right, i wasn't sure how exactly is shortval special ... fixed in 88d5f81, please take a look

core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
```diff
@@ -494,13 +494,28 @@ void dump_trezor_vcp(FILE *out, const mp_obj_VCP_t *vcp) {
...
 void dump_protomsg(FILE *out, const mp_obj_protomsg_t *value) {
+  mp_obj_t name[2] = {MP_OBJ_NULL, MP_OBJ_NULL};
+  mp_obj_type_t *type = protobuf_debug_msg_type();
+  type->attr((mp_obj_t)value, MP_QSTR_MESSAGE_NAME, name);
+
   print_type(out, "protomsg", NULL, value, false);
-  fprintf(out, ",\n\"shortval\": ");
-  dump_short(out, protobuf_debug_name(value));
+  fprintf(out, ",\n\"message_name\": ");
+  dump_short(out, name[0]);
```
matejcik, line 503: isn't this missing the trailing `,` in the resulting json?

mmilata: `dump_map_as_children` prints `,` as the first character

mmilata:
I was getting `MemoryError` because `min_ptr` was something like `0x620c60`. I noticed the analyzer ignores addresses starting with 0x5 so I extended this to 0x6 also in 95ef4a7, even though I have no idea what it means.

matejcik:
my guess is linker is placing Rust static objects somewhere starting with 0x6

matejcik (approved):
lgtm, let's merge when CI is green

PR #1724, opened 2021-07-19
core: explicitly refresh display on T1

mmilata:
It is my understanding that we need to explicitly push pixel data to the OLED display on T1, while the TT display is mapped to memory and does not need to be told about changes.

Such explicit refresh is already implemented for emulator where it is invoked after every event loop step. This patch mostly adds tracking whether the pixel contents were modified since last refresh so that we don't spend time redrawing display needlessly.


matejcik (approved):
LGTM, but @prusnak has more say in this

prusnak (approved):
utACK

PR #1725, opened 2021-07-19
Share wire buffer between wire and debuglink

mmilata:
Saves about 8KiB, which is 6.25 % of T1 memory (debug build only). I dislike the changes in codec_v1.py because they introduce overhead for production builds, please advise.

### DIFF
--- a/core/src/trezor/wire/__init__.py
+++ b/core/src/trezor/wire/__init__.py
@@ -130,8 +130,6 @@ DUMMY_CONTEXT = DummyContext()
 PROTOBUF_BUFFER_SIZE = 8192
 
 WIRE_BUFFER = bytearray(PROTOBUF_BUFFER_SIZE)
-if __debug__:
-    WIRE_BUFFER_DEBUG = bytearray(PROTOBUF_BUFFER_SIZE)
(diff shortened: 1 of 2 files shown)


matejcik:
to completely avoid the overhead, make it a decorator that returns the function unmodified in production and wrapped in a try/finally on debug?

matejcik:
that also avoids allocating an additional object every time the function is called

mmilata:
Yeah that would be a nice solution, however I need to "lock" the buffer somewhere after the `loop.wait()` call meaning I'd have to split the functions. That makes the code less readable and introduces a couple of allocations which we better avoid.

matejcik:
I see. Let's not complicate things then.

But in any case please use a common instance of the BufferLock object instead of creating a new one for the with-block. (then you can also keep the lock attribute on the instance as opposed to class, which makes it more readable)

matejcik (approved):

mmilata:
CI green, ok to merge?

romanz:
FTR, it was undone in https://github.com/trezor/trezor-firmware/pull/2572.

PR #1726, opened 2021-07-20
feat(ci): add sha checksum for alpine download and change to https

### LINKED ISSUE #1714: Be stricter about downloading the Alpine container
https://walletscrutiny.com/hardware/trezorOne/ points out that we just download a container from Alpine CDN, without even using https.

We should at least be checking its hash against a known value.

vdovhanych:
This should add sha checksum and changes the download of alpine to https. For our GitLab ci and also inside the 
`build-docker.sh ` script. Both should fail with a message `sha256sum: WARNING: 1 computed checksum did NOT match` script and ci will fail if the sha of the file does not match the one in the script. 
Resolves #1714

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -18,8 +18,22 @@ if [ -z "$ALPINE_ARCH" ]; then
   esac
 fi
 
+if [ -z "$ALPINE_CHECKSUM" ]; then
+  case "$ALPINE_ARCH" in
+    aarch64)
+      ALPINE_CHECKSUM="bc541e148463b3dde10fdbb1af8eac4e34706eae8883c6d126263db07a9a9c42"
+      ;;
+    x86_64)
+      ALPINE_CHECKSUM="bcdf5a4e58637b9228f8e474547a3de9ea02a05a5fa68a2495b0657ada7e65f6"
+      ;;
+    *)
+      exit
+  esac
+ fi
+
+
 CONTAINER_NAME=${CONTAINER_NAME:-trezor-firmware-env.nix}
-ALPINE_CDN=${ALPINE_CDN:-http://dl-cdn.alpinelinux.org/alpine}
+ALPINE_CDN=${ALPINE_CDN:-https://dl-cdn.alpinelinux.org/alpine}
 ALPINE_RELEASE=${ALPINE_RELEASE:-3.14}
 ALPINE_VERSION=${ALPINE_VERSION:-3.14.0}
 ALPINE_TARBALL=${ALPINE_FILE:-alpine-minirootfs-$ALPINE_VERSION-$ALPINE_ARCH.tar.gz}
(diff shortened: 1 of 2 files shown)


matejcik (approved):

build-docker.sh
```diff
@@ -19,10 +19,11 @@ if [ -z "$ALPINE_ARCH" ]; then
 fi
 
 CONTAINER_NAME=${CONTAINER_NAME:-trezor-firmware-env.nix}
-ALPINE_CDN=${ALPINE_CDN:-http://dl-cdn.alpinelinux.org/alpine}
+ALPINE_CDN=${ALPINE_CDN:-https://dl-cdn.alpinelinux.org/alpine}
 ALPINE_RELEASE=${ALPINE_RELEASE:-3.14}
 ALPINE_VERSION=${ALPINE_VERSION:-3.14.0}
 ALPINE_TARBALL=${ALPINE_FILE:-alpine-minirootfs-$ALPINE_VERSION-$ALPINE_ARCH.tar.gz}
+ALPINE_CHECKSUM=${ALPINE_CHECKSUM:-bcdf5a4e58637b9228f8e474547a3de9ea02a05a5fa68a2495b0657aaa7e65f6}
```
prusnak, line 26: This will break on `aarch64`. Let's use the similar trick as above:

```
if [ -z "$ALPINE_CHECKSUM" ]; then
  case "$ALPINE_ARCH" in
    aarch64)
      ALPINE_CHECKSUM ="bc541e148463b3dde10fdbb1af8eac4e34706eae8883c6d126263db07a9a9c42"
      ;;
    x86_64)
      ALPINE_CHECKSUM ="bcdf5a4e58637b9228f8e474547a3de9ea02a05a5fa68a2495b0657aaa7e65f6"
      ;;
    *)
      echo "Unsupported arch"
      exit
  esac
 fi
```

prusnak (changes requested):

build-docker.sh
vdovhanych: You're right didn't think of that. Pushed commit with your suggestion, and it should work as expected.

build-docker.sh
```diff
@@ -18,8 +18,22 @@ if [ -z "$ALPINE_ARCH" ]; then
   esac
 fi
 
+if [ -z "$ALPINE_CHECKSUM" ]; then
+  case "$ALPINE_ARCH" in
+    aarch64)
+      ALPINE_CHECKSUM="0cc1dbad694c93b686bd76a0a5db3aaae2e1357e1b687a99b1bd37df9bae8692"
```
prusnak, line 24: Where did you get this hash from? I suggested a different one.

vdovhanych: I downloaded the wrong one, sorry for that. Yours was indeed the right one, somehow got confused by the file names. Should be fixed now.

prusnak (approved):

tsusanka (approved):
I confirm the hashes equal to the ones found on https://aplinelinux.org [here](https://dl-cdn.alpinelinux.org/alpine/v3.14/releases/x86_64/alpine-minirootfs-3.14.0-x86_64.tar.gz.sha256) and [here](https://dl-cdn.alpinelinux.org/alpine/v3.14/releases/aarch64/alpine-minirootfs-3.14.0-aarch64.tar.gz.sha256).

Issue #1651, opened 2021-06-04
Add changelog checker to GitHub actions


matejcik:

vdovhanych:
Hi, maybe it's very straightforward, but I'm not sure what it is that should be checked here. Assume it's something like [this](https://github.com/marketplace/actions/changelog-checker). Thnx @matejcik

matejcik:
we have a gitlab job for this, and a custom checker script in `ci/check-changelog.sh`
we just want to run that in github actions (as well), mainly so that external contributors get the result separately and straight away

cc @mmilata who made the script

matejcik:
this changelog check should not block merging, we sometimes want to deliberately ignore the result

matejcik:
on #1706 the github action is passing and the gitlab job is failing. the github action should fail as well.

vdovhanych:
Interesting, the only thing I can think of is that GitHub is cloning the repository is differently from GitLab. @mmilata maybe that variable is indeed needed for the script?

mmilata:
That shouldn't be the problem, the script correctly detected that it's a feature (not release) branch. Maybe GH actions point `origin/master` to something else than GitLab does?

vdovhanych:
Could this be the problem @mmilata? 

Checkout pull request HEAD commit instead of merge commit
```yaml
- uses: actions/checkout@v2
  with:
    ref: ${{ github.event.pull_request.head.sha }}
```

Now it checks out HEAD of the merge commit would changing to this help?

mmilata:
Hmm, I tested that and the script seems to work fine on merge commits. Looking at the options I wonder if the lack of history due to default `fetch-depth: 1` isn't the problem. Let's try with `fetch-depth: 0`?

PR #1727, opened 2021-07-20
feat(ci): add changelog check to github actions and small improvments

vdovhanych:
This adds a changelog check to GitHub actions for external pull requests. I also adjusted the naming of other checks while on it. I can do it in a separate commit if that would be better. 

Resoles #1651

### DIFF
--- a/.github/workflows/prebuild.yml
+++ b/.github/workflows/prebuild.yml
@@ -3,19 +3,20 @@ name: "Prebuild checks"
 on: [pull_request]
 
 jobs:
-
   style_check:
+    name: Style check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
       - uses: cachix/install-nix-action@v13
         with:
           nix_path: nixpkgs=channel:nixos-unstable
-        name: "Run script"
+        name: "Run style check"
       - run: nix-shell --run "poetry install"
       - run: nix-shell --run "poetry run make style_check"
 
   defs_check:
+    name: Defs check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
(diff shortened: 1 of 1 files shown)


matejcik (approved):
utACK

.github/workflows/prebuild.yml
```diff
@@ -3,19 +3,20 @@ name: "Prebuild checks"
...
   style_check:
+    name: Style check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
       - uses: cachix/install-nix-action@v13
         with:
           nix_path: nixpkgs=channel:nixos-unstable
-        name: "Run script"
+        name: "Install poetry and run style check"
```
tsusanka, line 14: I would drop the "Install poetry" as it is a prerequisite not something we test per-se.

tsusanka (commented):
One nit but LGTM. So we do this because external PRs do not run in Gitlab right? @matejcik shouldn't we also remove the check from Gitlab then? So we do not have it twice.

matejcik:
> shouldn't we also remove the check from Gitlab then

style, defs and gen checks also run in gitlab.
I'd keep it, we also use gitlab internally, it can be useful to get the changelog check before you create a PR

tsusanka:
Ok!

.github/workflows/prebuild.yml
vdovhanych: Ok removed the install poetry and left only the description "Run style check" or other checks.

mmilata (approved):
LGTM! The script uses the value of `$CI_COMMIT_BRANCH` which I'm not sure github actions set. However when it's not set the script does the right thing (assumes it's not a release branch), so I guess that's ok.

tsusanka:
@mmilata We could use [GITHUB_REF](https://docs.github.com/en/actions/reference/environment-variables) for that. But I agree as that is solely for the release branch (and we do not file PR there) we do not have to worry about it..

Issue #1705, opened 2021-07-13
FIDO2 Binance icon missing after origin was changed


bosomt:
**Describe the bug**
Biance changed their FIDO2 origin from www.binance.com to binance.com
As you can see on screenshot there is no icon attached to new origin in current firmware version.

**Firmware version and revision**
2.4.1 regular commit 0x46e05307645e4a446e7c64708f38042aea6e845d

**Desktop/smartphone setup (please complete the following information):**
 - Browser Chrome
 - Version Version 91.0.4472.114 (Official Build) (arm64)

**To Reproduce**
1. Open Chrome and login to Binance
2. Go to https://www.binance.com/en/my/security and add new token
3. Observe Binance icon

**Screenshots**

![obrazek](https://user-images.githubusercontent.com/31506317/125403846-49676d00-e3b6-11eb-918b-e4676e8439a6.png)

andrewkozlik:
Please download the T1 and TT firmwares from:
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/339348790

For both T1 and TT test the following:

Test case 1:
1. Try to find an old credential that was created when the origin was www.binance.com.
2. Authenticate using the old credential.
3. Verify that the icon shows up correctly (TT only) and that the authentication is successful.

Test case 2:
1. Register a new credential at binance.com.
2. Verify that the icon shows up correctly (TT only), that the origin reads "binance.com", i.e. without the "www", and that the registration is successful.
3. Authenticate using the new credential.
4. Verify that the icon shows up correctly (TT only) and that the authentication is successful.

bosomt:
@andrewkozlik ON IT

bosomt:
Test case 1:

i was not able to find correct combination of device and account that was registered before they changed origin.
Is it possible to load registration to device via trezorctl or other way ?

Test case 2:

TT **QA OK**
T1  **QA OK**

andrewkozlik:
> Test case 1:
> 
> i was not able to find correct combination of device and account that was registered before they changed origin.
> Is it possible to load registration to device via trezorctl or other way ?

These credentials are derived purely from the seed. They do not need to be loaded to the device separately. It should be working fine on our end, so I think we don't need to pursue this.

> Test case 2:
> 
> TT **QA OK**
> T1 **QA OK**

Thanks!

PR #1729, opened 2021-07-21
Add new rpId to Binance's FIDO definition.

andrewkozlik:
Fixes https://github.com/trezor/trezor-firmware/issues/1705.


matejcik (approved):
utACK but I'd prefer to wait until gitlab is up and green

PR #1743, opened 2021-07-30
Avoid busy loop in legacy emulator

matejcik:
1. I use `poll()` to check data on normal and debug interfaces at the same time...
2. which allows me to use timeout on the `poll()` call...
3. so I can implement `usbSleep(millis)` in terms of that timeout, instead of busy-looping over `usbPoll()`

So I changed current `usbPoll()` to `usbSleep()`, and implement `usbPoll()` as `usbSleep(0)`

The more important busy loop is in `main.c`. In order to make that sleep sensibly, I had to modify the `usbSleep()` function so that it will send/receive the whole protobuf message in a single call -- otherwise we'd be inserting delays in the middle of messages.

### DIFF
--- a/legacy/emulator/emulator.h
+++ b/legacy/emulator/emulator.h
@@ -30,7 +30,8 @@ void emulatorPoll(void);
 void emulatorRandom(void *buffer, size_t size);
 
 void emulatorSocketInit(void);
-size_t emulatorSocketRead(int *iface, void *buffer, size_t size);
+size_t emulatorSocketRead(int *iface, void *buffer, size_t size,
+                          int timeout_ms);
 size_t emulatorSocketWrite(int iface, const void *buffer, size_t size);
 
 #endif
--- a/legacy/emulator/udp.c
+++ b/legacy/emulator/udp.c
@@ -19,6 +19,7 @@
 
 #include <arpa/inet.h>
 #include <errno.h>
+#include <poll.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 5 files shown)


matejcik:
Btw @prusnak are there deeper reasons for busy-looping this way (i.e., per USB packet)? Is there something that might need to be done in the middle of receiving or sending a packet?

Or is it just how it was implemented at first and there was no reason to change it?

This might be important for core, we're considering reading from USB synchronously.

prusnak:
> Or is it just how it was implemented at first and there was no reason to change it?

Yes, I guess the naive implementation worked and never had to be changed earlier.

Your code looks so much better and if it works I'd merged that in! 👍

prusnak (approved):

matejcik:
@prusnak to clarify, my question was about the code that runs _on hardware_.

IOW, if we wanted to remove the "if emulator", do you know of any reason why a similarly implemented usbPoll/usbSleep would break?

invd:
@matejcik I noticed this PR while doing T1 fuzzer changes today.

While the new emulator code is more efficient in different aspects, I think the old emulator code was closer to how the USB packet handling works in hardware. For example, the number of potentially accepted "USB" packets per `USBPoll()` call has changed significantly, which _might_ be relevant in specific bug edge cases.

In my understanding, the goal of the emulator is not to be the as-close-as-possible low-level simulation of the hardware behavior, so this is just an observation from the somewhat different perspective of the fuzzer, where this topic is more relevant.

matejcik:
@invd this is good to know. The optimization is important for our day-to-day because it seems to affect CI speed and stability, but it might be worth making it configurable in order to be able to match HW behavior.

PR #1745, opened 2021-08-01
refactor(python/stellar): Use Stellar Python SDK to parse Stellar transactions.

overcat:
This PR mainly introduces the [Stellar Python SDK](https://github.com/StellarCN/py-stellar-base) to reduce the maintenance cost of the Stellar code.

### Incompatible changes

- `trezorlib.stellar.parse_transaction_bytes` was removed. (We can keep it if needed.)
- `trezorlib.stellar` now relies on stellar-sdk, please check the README.md for more information.

### Added

- `trezorlib.stellar.parse_transaction` was added, it includes support for the Stellar [TransactionV1](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md#xdr) format transaction.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -55,7 +55,14 @@ units) will not be recognized, unless you install HIDAPI support (see below).
   pip3 install trezor[ethereum]
   ```
 
-To install both, use `pip3 install trezor[hidapi,ethereum]`.
+* **Stellar**: To support Stellar signing from command line, additional packages are
+  needed. Install with:
+
+  ```sh
+  pip3 install trezor[stellar]
+  ```
+
+To install all three, use `pip3 install trezor[hidapi,ethereum,stellar]`.
 
 ### Distro packages
 
--- a/python/setup.py
+++ b/python/setup.py
@@ -22,6 +22,7 @@ extras_require = {
     "ethereum": ["rlp>=1.1.0", "web3>=4.8"],
     "qt-widgets": ["PyQt5"],
     "extra": ["Pillow"],
+    "stellar": ["stellar-sdk>=4.0.0,<5.0.0"],
 }
 
 extras_require["full"] = sum(extras_require.values(), [])
--- a/python/src/trezorlib/cli/stellar.py
+++ b/python/src/trezorlib/cli/stellar.py
@@ -83,8 +83,17 @@ def sign_transaction(client, b64envelope, address, network_passphrase):
         click.echo()
         click.echo("  pip install stellar-sdk")
         sys.exit(1)
+    try:
+        envelope = parse_transaction_envelope_from_xdr(b64envelope, network_passphrase)
+    except Exception:
+        click.echo(
+            "Failed to parse XDR.\n"
+            "Make sure to pass a valid TransactionEnvelope object.\n"
+            "You can check whether the data you submitted is valid TransactionEnvelope object "
+            "through XDRViewer - https://laboratory.stellar.org/#xdr-viewer\n"
+        )
+        sys.exit(1)
 
-    envelope = parse_transaction_envelope_from_xdr(b64envelope, network_passphrase)
     if isinstance(envelope, FeeBumpTransactionEnvelope):
         click.echo("FeeBumpTransactionEnvelope is not supported")
         sys.exit(1)
--- a/python/src/trezorlib/cli/stellar.py
+++ b/python/src/trezorlib/cli/stellar.py
@@ -15,11 +15,20 @@
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
 import base64
+import sys
 
 import click
 
 from .. import stellar, tools
 from . import with_client
+
+try:
+    from stellar_sdk import (
+        parse_transaction_envelope_from_xdr,
+        FeeBumpTransactionEnvelope,
+    )
+except ImportError:
+    pass
 
 PATH_HELP = "BIP32 path. Always use hardened paths and the m/44'/148'/ prefix"
 
@@ -68,8 +77,20 @@ def sign_transaction(client, b64envelope, address, network_passphrase):
     For testnet transactions, use the following network passphrase:
     'Test SDF Network ; September 2015'
     """
+    if not stellar.HAVE_STELLAR_SDK:
+        click.echo("Stellar requirements not installed.")
+        click.echo("Please run:")
+        click.echo()
+        click.echo("  pip install stellar-sdk")
+        sys.exit(1)
+
+    envelope = parse_transaction_envelope_from_xdr(b64envelope, network_passphrase)
+    if isinstance(envelope, FeeBumpTransactionEnvelope):
+        click.echo("FeeBumpTransactionEnvelope is not supported")
+        sys.exit(1)
+
     address_n = tools.parse_path(address)
-    tx, operations = stellar.parse_transaction_bytes(base64.b64decode(b64envelope))
+    tx, operations = stellar.from_envelope(envelope)
     resp = stellar.sign_tx(client, tx, operations, address_n, network_passphrase)
 
     return base64.b64encode(resp.signature)
--- a/python/src/trezorlib/stellar.py
+++ b/python/src/trezorlib/stellar.py
@@ -13,13 +13,44 @@
 #
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
-
-import base64
-import struct
-import xdrlib
+from decimal import Decimal
+from typing import Union
 
 from . import exceptions, messages
 from .tools import expect
+
+try:
+    from stellar_sdk import (
+        AccountMerge,
+        AllowTrust,
+        Asset,
+        BumpSequence,
+        ChangeTrust,
+        CreateAccount,
+        CreatePassiveSellOffer,
+        HashMemo,
+        IdMemo,
+        ManageData,
+        ManageSellOffer,
+        Operation,
+        PathPaymentStrictReceive,
+        Payment,
+        ReturnHashMemo,
+        SetOptions,
+        TextMemo,
+        TransactionEnvelope,
+        TrustLineEntryFlag,
+        Price,
+        Network,
+    )
+    from stellar_sdk.xdr.signer_key_type import SignerKeyType
+
+    HAVE_STELLAR_SDK = True
+    DEFAULT_NETWORK_PASSPHRASE = Network.PUBLIC_NETWORK_PASSPHRASE
+
+except ImportError:
+    HAVE_STELLAR_SDK = False
+    DEFAULT_NETWORK_PASSPHRASE = "Public Global Stellar Network ; September 2015"
 
 # Memo types
 MEMO_TYPE_NONE = 0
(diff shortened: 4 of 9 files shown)


python/setup.py
```diff
@@ -22,6 +22,7 @@
     "ethereum": ["rlp>=1.1.0", "web3>=4.8"],
     "qt-widgets": ["PyQt5"],
     "extra": ["Pillow"],
+    "stellar": ["stellar-sdk>=4.0.0,<5.0.0"],
```
overcat, line 25: Do I need to add it to pyproject.toml in the root directory? If necessary, is it to be added under `cointool`?

python/src/trezorlib/stellar.py
```diff
@@ -33,309 +54,193 @@
...
     it from the G... string
     """
-    decoded = base64.b32decode(address_str)
-
-    # skip 0th byte (version) and last two bytes (checksum)
-    return decoded[1:-2]
+    return StrKey.decode_ed25519_public_key(address_str)
 
 
-def parse_transaction_bytes(tx_bytes):
```
overcat, line 80: If needed, we can keep this function.

pyproject.toml
```diff
@@ -63,6 +63,9 @@ pyasn1 = "*"
 # legacy
 setuptools = ">=24.2.0"
 
+# python
```
overcat, line 66: I did not find an entry corresponding to the Python folder, so I added it.

prusnak: You should use optional dependencies: https://python-poetry.org/docs/pyproject/#extras

`stellar-sdk = { version = ">=4.0.0,<5.0.0", optional = true }`

and then add the following at the end of the file:

```
[tool.poetry.extras]
stellar = ["stellar-sdk"]
```

python/README.md
```diff
@@ -55,7 +55,14 @@ units) will not be recognized, unless you install HIDAPI support (see below).
...
 
-To install both, use `pip3 install trezor[hidapi,ethereum]`.
+* **Stellar**: To support Stellar signing from command line, additional packages are
+  needed. Install with:
+
+  ```sh
+  pip3 install trezor[stellar]
+  ```
+
+To install both, use `pip3 install trezor[hidapi,ethereum,stellar]`.
```
prusnak, line 65:
```suggestion
To install all three, use `pip3 install trezor[hidapi,ethereum,stellar]`.
```

overcat: 80f62f09ea2e278395de508d8e53fb6435644386

pyproject.toml
overcat: 80f62f09ea2e278395de508d8e53fb6435644386

python/src/trezorlib/stellar.py
matejcik: we don't :)

python/.changelog.d/1745.incompatible.2
```diff
@@ -0,0 +1 @@
+`trezorlib.stellar` now relies on stellar-sdk, please check the README.md for more information.
```
matejcik, line 1: why is this an incompatible change entry?

pyproject.toml
matejcik: no, given that `test_stellar.py` uses this, it should be part of the main pyproject.toml, not an extra. please revert this part

poetry.lock
```diff
@@ -1,3 +1,37 @@
+[[package]]
```
matejcik, line 1: This upgraded some things that don't seem necessary. Did you use `poetry update stellar-sdk` ?

python/.changelog.d/1745.incompatible.2
overcat: Because the lib did not rely on it before, if the user directly upgrades trezorlib without installing stellar-sdk, then he cannot use trezorlib.stellar normally. Do you have any other suggestions?

matejcik (commented):
LGTM overall, but please add some testcases for the V1 transaction format.

python/.changelog.d/1745.incompatible.2
matejcik: I see. OK, let's leave this.

poetry.lock
overcat: No, I added this package by the following command

```
> poetry add "stellar-sdk>=4.0.0,<5.0.0" 

Updating dependencies
Resolving dependencies... (0.6s)

Writing lock file

Package operations: 8 installs, 8 updates, 0 removals

  - Updating idna (2.10 -> 3.2)
  - Installing multidict (5.1.0)
  - Updating typing-extensions (3.7.4.3 -> 3.10.0.0)
  - Installing async-timeout (3.0.1)
  - Updating attrs (20.3.0 -> 21.2.0)
  - Updating certifi (2020.12.5 -> 2021.5.30)
  - Updating urllib3 (1.26.3 -> 1.26.6)
  - Installing yarl (1.6.3)
  - Installing aiohttp (3.7.4.post0)
  - Updating cffi (1.14.4 -> 1.14.6)
  - Updating requests (2.25.1 -> 2.26.0)
  - Updating six (1.15.0 -> 1.16.0)
  - Installing aiohttp-sse-client (0.2.1)
  - Installing pynacl (1.4.0)
  - Installing stellar-base-sseclient (0.0.21)
  - Installing stellar-sdk (4.1.0)

```

pyproject.toml
overcat: a78c9c112688c8522704e019ec993ae294b59d2b

overcat:
@matejcik PTAL 🙏

Some new commits have been added, please check them. 

It’s worth mentioning that the XDRs in the original tests seemed to be invalid, so I removed and refactored them. You can use [this tool](https://laboratory.stellar.org/#xdr-viewer?input=AAAAAAGUcmKO5465JxTSLQOQljwk2SfqAJmZSG6JH6wtqpwhAAABLAAAAAAAAAABAAAAAAAAAAEAAAALaGVsbG8gd29ybGQAAAAAAwAAAAAAAAAAAAAAABbxCy3mLg3hiTqX4VUEEp60pFOrJNxYM1JtxXTwXhY2AAAAAAvrwgAAAAAAAAAAAQAAAAAW8Qst5i4N4Yk6l%2BFVBBKetKRTqyTcWDNSbcV08F4WNgAAAAAN4Lazj4x61AAAAAAAAAAFAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABLaqcIQAAAEBKwqWy3TaOxoGnfm9eUjfTRBvPf34dvDA0Nf%2BB8z4zBob90UXtuCqmQqwMCyH%2BokOI3c05br3khkH0yP4kCwcE&type=TransactionEnvelope&network=test) to check whether XDR is valid.

Thanks!

python/src/trezorlib/stellar.py
```diff
@@ -97,13 +98,25 @@ def parse_transaction(xdr: str, network_passphrase: str = DEFAULT_NETWORK_PASSPH
...
-        tx.memo_text = memo.memo_text
-    if isinstance(memo, IdMemo):
+        # In the ledger, memo_text is stored in bytes, so in theory, memo_text should be defined as bytes.
+        # For forward compatibility, we try to decode here. If the decoding fails, a decoding error will be
+        # displayed on the device.
+        try:
+            tx.memo_type = MEMO_TYPE_TEXT
+            tx.memo_text = memo.memo_text.decode("utf-8")
+        except UnicodeDecodeError:
+            tx.memo_text = "Unable to display the text memo"
```
matejcik, line 108: This seems trivially incorrect - you're changing the transaction context. Trezor will happily sign a transaction with the memo `"Unable to display the text memo"`, but it's probably a different transaction from the one submitted.

matejcik:
> the XDRs in the original tests seemed to be invalid

they're valid but [they're not envelopes](https://laboratory.stellar.org/#xdr-viewer?input=AAAAABXWSL%2Fk028ZbPtXNf%2FYylTNS4Iz90PyJEnefPMBzbRpAAAAZAAAAAEAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAEAAAAAXVVkJGaxhbhDFS6eIZFR28WJICfsQBAaUXvtXKAwwuAAAAABVEVTVAAAAAAphJYCwg5YNl8SPBLYehykVQ0QzSGwrg4Y1E4%2BVv1qFQAAAAAdzxaYAAAAAA%3D%3D&type=Transaction&network=test) but raw Transaction objects.

i'm wondering how that happened, but it probably doesn't matter, the way you implemented the testcases is better.

python/src/trezorlib/stellar.py
overcat: I haven't read the firmware code yet, do you mean that Trezor will return the wrong transaction signature in this case?

matejcik: It's the right transaction signature for what you're submitting ;) but it's wrong for the decoded transaction.

Also the Laboratory claims that "Unable to display the text memo" is too long. I'm wondering where this limit is enforced, and whether we should also validate it on the firmware side.

overcat: I checked the code and it seems to have an impact, so in this case, can we throw an exception first and prompt that the transaction cannot be processed? Then we fix it in another PR.

matejcik: this whole snippet, including the comment, is very confusing and seems wrong.

In XDR the memo text is defined as `string`, which per XDR RFC must be ASCII, per Stellar docs probably UTF. In any case what you get from the parsed XDR should be of type `str`, right? If the XDR parser returns `bytes`, then the `decode()` call is correct, but ... if the decoding fails we need to raise an exception anyway because the incoming XDR is incorrect. So maybe remove the whole try/except block and let the original UnicodeDecodeError fall through?

also the "forward-compatibility" is probably the wrong term here. Suggest only something like:
```
# memo_text is specified as UTF-8 string, but returned as bytes from the XDR parser
```

matejcik: > Then we fix it in another PR (fix memo_text to bytes type.).

`memo_text` is [defined as `string`](https://github.com/stellar/stellar-core/blob/master/src/xdr/Stellar-transaction.x#L520) so I believe there's nothing to fix on the firmware side

overcat: Thanks for your explanation, I don't want this to cause the entire PR to be blocked, so I removed it first. I will make an in-depth test later.

4ad2a3e496abd530dcc142d0614d6630cdf15f48

matejcik:
The current code has the problem that you need `stellar-sdk` to execute trezorctl, even if you don't use stellar commands. (`trezorctl` imports `trezorlib.cli.stellar` imports `trezorlib.stellar` imports `stellar-sdk`)

We need to break this cycle somehow:
- either in `trezorlib.cli.stellar`, wrap `from .. import stellar` in a try/except (and then fix usage of the constants `stellar.DEFAULT_BIP32_PATH` and `DEFAULT_NETWORK_PASSPHRASE`)
- or in `trezorlib.stellar`, wrap `from stellar_sdk import ...` in a try/except, and then again deal with contents of the constants

matejcik:
> * or in `trezorlib.stellar`, wrap `from stellar_sdk import ...` in a try/except, and then again deal with contents of the constants

this seems like the better option: then you can do something like:
```
try:
	from stellar_sdk import (...)
	HAVE_STELLAR_SDK = True
except ImportError:
	HAVE_STELLAR_SDK = False

# ...

def parse_transaction(...):
	if not HAVE_STELLAR_SDK:
		raise RuntimeException("Stellar SDK not available")
```

...while the other functions (notably `sign_tx` and `get_address` remain available if someone wants to construct their arguments in a different way

then in `trezorlib.cli.stellar::sign_tx()` you should also do:
```
if not stellar.HAVE_STELLAR_SDK:
	click.echo("Don't have stellar SDK, install it with pip, blabla")
```
(see `trezorlib.cli.ethereum` usage of `HAVE_SIGN_TX`)

the one problem remaining is what to do about the DEFAULT_NETWORK_PASSPHRASE constant. Probably use the literal that was there before, in case we can't use the stellar sdk version?

matejcik:
this also leads to an idea: instead of `parse_transaction(xdr: str, passphrase: str)`, there could be a `from_envelope(envelope: TransactionEnvelope)` which accepts the parsed envelope -- or a result of `TransactionBuilder.build()`, so that you can integrate `trezorlib` into a Python program already using `stellar_sdk` without going through XDR.

Then you'd also need to import `stellar_sdk` (conditionally) in `trezorlib.cli.stellar` and call `parse_transaction_envelope_from_xdr` in the CLI part, so that you can call `trezorlib.stellar.from_envelope(parsed_te)`

matejcik:
> this also leads to an idea

with this split design, there is no need for `DEFAULT_NETWORK_PASSPHRASE` to exist in `trezorlib.stellar` at all. Callers would import it from `stellar_sdk` on their own.

overcat:
> > this also leads to an idea
> 
> with this split design, there is no need for `DEFAULT_NETWORK_PASSPHRASE` to exist in `trezorlib.stellar` at all. Callers would import it from `stellar_sdk` on their own.

@matejcik I'm sorry I need to go out for a while, I will deal with this problem when I come back, I really appreciate your support 
❤️

python/src/trezorlib/stellar.py
```diff
@@ -33,309 +55,215 @@
...
-
-
 DEFAULT_BIP32_PATH = "m/44h/148h/0h"
 # Stellar's BIP32 differs to Bitcoin's see https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0005.md
-DEFAULT_NETWORK_PASSPHRASE = "Public Global Stellar Network ; September 2015"
+DEFAULT_NETWORK_PASSPHRASE = Network.PUBLIC_NETWORK_PASSPHRASE
 
 
-def address_from_public_key(pk_bytes):
+def address_from_public_key(pk_bytes: bytes) -> str:
```
matejcik, line 63: this is not used anywhere, you can remove it

python/src/trezorlib/stellar.py
```diff
@@ -33,309 +55,215 @@
...
-    final_bytes.extend(pk_bytes)
-    # checksum
-    final_bytes.extend(struct.pack("<H", _crc16_checksum(final_bytes)))
-
-    return base64.b32encode(final_bytes).decode()
+    return StrKey.encode_ed25519_public_key(pk_bytes)
 
 
-def address_to_public_key(address_str):
+def address_to_public_key(address_str: str) -> bytes:
```
matejcik, line 68: this is not used anywhere, you can remove it

python/src/trezorlib/stellar.py
```diff
@@ -33,309 +55,215 @@
...
-    # Check for and parse optional source account field
-    source_account = None
-    if unpacker.unpack_bool():
-        source_account = unpacker.unpack_fopaque(32)
-
-    # Operation type (See OP_ constants)
-    type = unpacker.unpack_uint()
-
-    if type == OP_CREATE_ACCOUNT:
+def _read_operation(parsed_operation: Operation):
```
matejcik, line 124: suggest renaming `parsed_operation` to `op`, it's significantly shorter and still very clear

python/src/trezorlib/stellar.py
```diff
@@ -33,309 +55,215 @@
...
+    # TODO: Let's add muxed account support later.
+    source_account = parsed_operation.source
+    if source_account:
+        source_account = source_account.account_id
```
matejcik, line 128: please make this clearer:
```
if op.source:
    source_account = op.source.account_id
else:
    source_account = None
```

python/src/trezorlib/stellar.py
```diff
@@ -33,309 +55,215 @@
...
-            send_max=unpacker.unpack_hyper(),
-            destination_account=_xdr_read_address(unpacker),
-            destination_asset=_xdr_read_asset(unpacker),
-            destination_amount=unpacker.unpack_hyper(),
+            send_asset=_read_asset(parsed_operation.send_asset),
+            send_max=_read_amount(parsed_operation.send_max),
+            destination_account=parsed_operation.destination.account_id,
+            destination_asset=_read_asset(parsed_operation.dest_asset),
+            destination_amount=_read_amount(parsed_operation.dest_amount),
             paths=[],
```
matejcik, line 150: would this work? `paths=[_read_asset(asset) for asset in operation.paths]`

matejcik (changes requested):

overcat:
> > this also leads to an idea
> 
> with this split design, there is no need for `DEFAULT_NETWORK_PASSPHRASE` to exist in `trezorlib.stellar` at all. Callers would import it from `stellar_sdk` on their own.

The current `sign_tx` function is like this:
```python
def sign_tx(
    client, tx, operations, address_n, network_passphrase=DEFAULT_NETWORK_PASSPHRASE
):
    pass
```

If we remove `DEFAULT_NETWORK_PASSPHRASE` from `trezorlib.stellar`, do we still need to set a default value for `network_passphrase` parameter? If not, we need to mark this as a breaking change.

----

In addition, we have the following code in `trezorlib.cli.stellar`. I think we can't change `default=stellar.DEFAULT_NETWORK_PASSPHRASE` to `default=Network.PUBLIC_NETWORK_PASSPHRASE`, which will still cause import-related exception.

```python
@click.option(
    "-n",
    "--network-passphrase",
    default=stellar.DEFAULT_NETWORK_PASSPHRASE,
    required=False,
    help="Network passphrase (blank for public network).",
)
@click.argument("b64envelope")
@with_client
def sign_transaction(client, b64envelope, address, network_passphrase):
    pass
```

-------

In summary, I think we should keep `DEFAULT_NETWORK_PASSPHRASE = "Public Global Stellar Network ; September 2015"`

matejcik:
> In summary, I think we should keep `DEFAULT_NETWORK_PASSPHRASE = "Public Global Stellar Network ; September 2015"`

agreed, but only in case we can't import `stellar_sdk`

python/src/trezorlib/stellar.py
overcat: 75ebbce0d081dde71c79022c1822943c892efe32

python/src/trezorlib/stellar.py
overcat: 75ebbce0d081dde71c79022c1822943c892efe32

python/src/trezorlib/stellar.py
overcat: 75ebbce0d081dde71c79022c1822943c892efe32

python/src/trezorlib/stellar.py
overcat: 75ebbce0d081dde71c79022c1822943c892efe32

python/src/trezorlib/stellar.py
```diff
@@ -13,14 +13,45 @@
...
+        TrustLineEntryFlag,
+        Price,
+        Network,
+    )
+    from stellar_sdk.xdr.signer_key_type import SignerKeyType
+
+    HAVE_STELLAR_SDK = True
+    DEFAULT_NETWORK_PASSPHRASE = Network.PUBLIC_NETWORK_PASSPHRASE
+
+except ImportError:
```
overcat, line 51: dd8d28e8242f7caec586b060cfe6068195f839d5

python/src/trezorlib/stellar.py
overcat: 75ebbce0d081dde71c79022c1822943c892efe32

matejcik:
LGTM if tests pass now. Please rebase&autosquash.

Also please change the incompatible changelog entries to one, something like "trezorlib.stellar was reworked to use stellar-sdk instead of providing local implementations"

overcat:
> LGTM if tests pass now. Please rebase&autosquash.
> 
> Also please change the incompatible changelog entries to one, something like "trezorlib.stellar was reworked to use stellar-sdk instead of providing local implementations"

Done.

matejcik:
it seems you left out the commit that fixes `test_stellar.py` after going from parse_transaction -> from_envelope

matejcik:
also I have a fun error for you :)
```
% trezorctl stellar sign-transaction -n m/44h/148h/0h AAAAANSPYkfVpAotFQB7TXGiovZ+yuPpykHGaXvF0OZ8Fp3FAAAAZAAG40cAAAABAAAAAQAAAAAAAAAAAAAAAGDt9IAAAAAAAAAAAQAAAAAAAAAKAAAAC3RyZXpvcl90ZXN0AAAAAAEAAAAFdmFsdWUAAAAAAAAA

Aborted!
```
the XDR in question is raw Transaction (same format as was used in the test cases). What actually happens is `parse_transaction_envelope_from_xdr` raises `EOFError` and that passes through Click as the "Aborted!" string.

Not sure why. However, in any case, it will be necessary to wrap the parse_envelope call in a try/except and raise something meaningful (and visible to the user). "Failed to parse XDR. Make sure to pass a valid TransactionEnvelope object." perhaps?

matejcik (approved):
:+1: looking nice.

matejcik:
Thanks for your work. Will merge this if/when automated checks go green. Unfortunately GitHub Actions seem to have problems right now so that'll need to wait for tomorrow.

overcat:
@matejcik Thank you for your effort :-)

python/src/trezorlib/cli/stellar.py
```diff
@@ -68,8 +77,28 @@ def sign_transaction(client, b64envelope, address, network_passphrase):
...
+        click.echo("Stellar requirements not installed.")
+        click.echo("Please run:")
+        click.echo()
+        click.echo("  pip install stellar-sdk")
+        sys.exit(1)
+    try:
+        envelope = parse_transaction_envelope_from_xdr(b64envelope, network_passphrase)
+    except Exception:
+        click.echo(
+            "Failed to parse XDR.\n Make sure to pass a valid TransactionEnvelope object.\n "
```
overcat, line 90: It looks like there are two extra spaces here, I will fix it, and force push.

python/tox.ini
```diff
@@ -8,6 +8,7 @@ envlist =
 [testenv]
 deps =
     -rrequirements.txt
+    -rrequirements-optional.txt
```
overcat, line 11: c5982132a5b64c67edd4c1e0713c451bf55f0469 makes ci happy.

pyproject.toml
overcat: @matejcik The current tests under the `python` folder use the environment provided by tox, so I think we can mark stellar-sdk as optional here.

https://github.com/trezor/trezor-firmware/pull/1745#discussion_r683903844

matejcik:
i see you deleted a comment about making `stellar_sdk` optional in pyproject.toml ? :)

i believe we could remove stellar_sdk from pyproject.toml entirely in that case?

overcat:
> i see you deleted a comment about making `stellar_sdk` optional in pyproject.toml ? :)
> 
> i believe we could remove stellar_sdk from pyproject.toml entirely in that case?

I'm not sure if I should do this. Should pyproject.toml contain all the dependencies required by the development environment? If we remove it, the developer needs to install stellar-sdk manually.

overcat:
Optional dependencies such as `rlp` and `web3` are not included in pyproject.toml. If this is the expected behavior, then I will remove stellar-sdk from pyproject.toml.

matejcik:
> Optional dependencies such as `rlp` and `web3` are not included in pyproject.toml. If this is the expected behavior, then I will remove stellar-sdk from pyproject.toml.

please do that.

overcat:
> > Optional dependencies such as `rlp` and `web3` are not included in pyproject.toml. If this is the expected behavior, then I will remove stellar-sdk from pyproject.toml.
> 
> please do that.

f96bf8f0f7c2bac45f56a85a47c759f08d33f221

overcat:
It looks like all the tests have passed, can I perform a rebase operation now? @matejcik

matejcik:
yes please

overcat:
> yes please

done

matejcik:
and merged. Thanks again for your contribution!

PR #1752, opened 2021-08-06
feat(ci): Suite e2e tests in CI pipeline testing current emulator

grdddj:
Trying to include Suite e2e tests into CI, which would be testing current emulators

Connected with:
- https://github.com/trezor/trezor-suite/pull/4121
- https://github.com/trezor/trezor-user-env/issues/49

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -49,6 +49,8 @@ stdenv.mkDerivation ({
     check
     clang-tools
     clang
+    docker
+    docker-compose
     editorconfig-checker
     gcc
     gcc-arm-embedded
--- a/ci/test.yml
+++ b/ci/test.yml
@@ -342,7 +342,7 @@ core unix memory profiler:
 suite e2e tests:
   stage: test
   variables:
-    COMPOSE_FILE: ./docker/docker-compose.suite-e2e-tests.yml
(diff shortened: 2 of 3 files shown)


grdddj:
Fixup commit 1dd5ebb:
- fixing the docker-compose file location

The CI job will still fail, as the image under `registry.gitlab.com/satoshilabs/trezor/trezor-suite/trezor-suite-e2e` is not existing yet - the Suite CI will need to create it.

Also, the `develop` branch of `Suite` currently does not support Cypress environmental variables, and `tenv` is currently not able to accept emulator version as URL.

So because of these and other reasons, it is still work in progress, and requires quite some coordination (on firmware, Suite and tenv side). Before doing all these, we should also **agree if this is a good way of doing it**

matejcik:
ISTM the firmware container doesn't have Docker inside it.

I don't want to add docker to the firmware env. Can a job use a different environment?

Converting the PR to draft for now.

grdddj:
> Can a job use a different environment?

No idea about this. If it could reuse `Suite` environment, it would be nice, but it could lose its independence

tsusanka:
Let's discuss on a meeting what we want to achieve with this.

Hannsek:
Closing this as we will have to rebuild this if we ever want this.

PR #1753, opened 2021-08-07
refactor(tests): modernize test_msg_getaddress_segwit_native

prusnak:
The fixture changes in `tests/ui_tests/fixtures.json` need to be re-generated by someone running Linux.

There seem to be some differences how SDL saves stuff to PNGs on Linux and macOS (most probably the default compression level is different).


prusnak:
I regenerated UI fixtures on a Linux machine now, but they are again different than the original ones. 🤔

grdddj:
> I regenerated UI fixtures on a Linux machine now, but they are again different than the original ones. thinking

Can it be because there is an **extra passphrase screen** in new UI tests? Therefore it would generate different hashes. We have encountered it in https://github.com/trezor/trezor-firmware/pull/1784#issuecomment-910438906

The `core device ui test` in your CI pipeline is skipped, so it cannot be verified

prusnak:
I regenerated the fixtures with disabled animations and the PR is now ready to be reviewed.

mmilata (approved):
LGTM, ui diff ACK

PR #1754, opened 2021-08-07
chore(nix): update rust to 1.54

prusnak:
https://blog.rust-lang.org/2021/07/29/Rust-1.54.0.html

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -3,10 +3,10 @@
  }:
 
 let
-  # the last commit from master as of 2021-07-09
+  # the last commit from master as of 2021-08-07
   rustOverlay = import (builtins.fetchTarball {
-    url = "https://github.com/oxalica/rust-overlay/archive/76732f3ddba766b1fe772fca80aafbc9cccd78dd.tar.gz";
-    sha256 = "0lqwxn28malsjxw8hxp4iwax5v0mlwi4l5q0cxfacq153gi6j0f8";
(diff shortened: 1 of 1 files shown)


ci/shell.nix
```diff
@@ -23,7 +23,7 @@ let
     ${nixpkgs.patchelf}/bin/patchelf --set-interpreter "$(cat $NIX_CC/nix-support/dynamic-linker)" "$out"
     chmod -w $out
   '';
-  rustStable = nixpkgs.rust-bin.stable."1.53.0".default.override {
+  rustStable = nixpkgs.rust-bin.stable."1.54.0".minimal.override {
```
prusnak, line 26: I am trying to change `default` to `minimal` here (mainly to exclude rust-docs), but I am not sure whether this would work (minimal also does not include e.g. rustfmt and clippy).

mmilata (approved):
There don't seem to be any major changes in rust/cargo that should affect us. The minimal profile seems to work fine. LGTM

matejcik:
> minimal profile seems to work fine.

if `rustfmt` and `clippy` is excluded, won't it be a problem if the nixshell is your dev environment?

(on that note but not in this PR, we should add rustfmt to style-check)

mmilata:
It can be explicitly added while still excluding `rust-docs`. However we currently need rustfmt with nightly features, and rustfmt version in nixpkgs has them (even though it's 1.53) so #1802 uses that. Did not try anything with clippy yet.

Issue #1691, opened 2021-07-07
Stellar: add operations


quietbits:
Hello Trezor Team,

The following operations are not supported in the Stellar app:
- [Path Payment Strict Send](https://github.com/trezor/trezor-firmware/blob/d6776d988e4f5b7d1a0db1ea3db0f422660cbe25/core/src/apps/stellar/operations/__init__.py#L34) [Path Payment Strict Receive works]
- [Create Claimable Balance](https://developers.stellar.org/docs/start/list-of-operations/#create-claimable-balance)
- [Claim Claimable Balance](https://developers.stellar.org/docs/start/list-of-operations/#claim-claimable-balance)
- [Begin Sponsoring Future Reserves](https://developers.stellar.org/docs/start/list-of-operations/#begin-sponsoring-future-reserves)
- [End Sponsoring Future Reserves](https://developers.stellar.org/docs/start/list-of-operations/#end-sponsoring-future-reserves)
- [Revoke Sponsorship](https://developers.stellar.org/docs/start/list-of-operations/#revoke-sponsorship)
- [Set Trustline Flags](https://developers.stellar.org/docs/start/list-of-operations/#set-trustline-flags)
- [Clawback](https://developers.stellar.org/docs/start/list-of-operations/#clawback)
- [Clawback Claimable Balance](https://developers.stellar.org/docs/start/list-of-operations/#clawback-claimable-balance)

We're adding Trezor support to [Laboratory](https://laboratory.stellar.org/) and it would be nice if all operations could be signed.

JFWooten4:
Yeah this is pretty serious. Even setting account clawback flag fails

stepmes:
Thank you @quietbits, we're noting this down and will get back to it as soon as we have the capacity to work on this area.

JFWooten4:
@stepmes  Please don't close :( Create claimable balance results in `Failed to execute 'postMessage' on 'Window': function t(e,n){return function(e,t){if(!(e instanceof t))throw new TypeError(\"Cannot call a class as a functio...<omitted>...)} could not be cloned` error, which I don't think is covered by #1756. Can't sing these from Suite to my knowledge

JFWooten4:
Any timeline on when this gets fixed? Mission-critical for many use cases

bjorn:
It appears claimable balances are getting used more and more. Why not add support for the "Claim Claimable Balance" operation in official firmware? See also the StellarTerm issue linked above.

[...]

Issue #1698, opened 2021-07-09
Stellar: support muxed accounts


quietbits:
Muxed accounts are defined in [CAP-27](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md) and [SEP-23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md). Payment to the destination, which is a muxed account, can't be signed with Trezor on [Stellar Account Viewer](https://laboratory.stellar.org/). The same error happens on [Stellar Laboratory](https://laboratory.stellar.org/) as well.

Trezor device shows the correct transaction information on the device, but it fails to sign it:

![image](https://user-images.githubusercontent.com/7346473/125114963-ad3d1d80-e0b8-11eb-96da-8bbdd44ebb12.png)

matejcik:
Muxed accounts are not supported in firmware. This means that if the device shows "correct transaction information," it's
(a) not showing the muxed account at all?
(b) showing account differently -- in which case i don't see how it's showing "correct information"?
(c) displaying the muxed account correctly, and behaving as if it is non-muxed? in which case, muxed account is indistinguishable (on user and protocol level) from non-muxed?

please explain

matejcik:
I tried generating a muxed account in Laboratory and sending to it from Account Viewer, but failed:
![image](https://user-images.githubusercontent.com/596073/125445672-510f4224-7568-47d6-85f2-54d5794d0bd9.png)
So I can't verify the behavior you describe.

stepmes:
Thank you @quietbits for this feedback. We're closing the issue now, but we'll keep it in our backlog for future review.

overcat:
> Muxed accounts are not supported in firmware. This means that if the device shows "correct transaction information," it's
> (a) not showing the muxed account at all?
> (b) showing account differently -- in which case i don't see how it's showing "correct information"?
> (c) displaying the muxed account correctly, and behaving as if it is non-muxed? in which case, muxed account is indistinguishable (on user and protocol level) from non-muxed?
> 
> please explain

MuxedAccount was introduced in [CAP-27](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md), and [SEP-23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md) introduced how to use ASCII to represent this type of address.


The definition of `MuxedAccount` in CAP-27 is as follows:
```
union MuxedAccount switch (CryptoKeyType type) {
[...]

PR #1755, opened 2021-08-09
Strong typing for Stellar

### LINKED ISSUE #2: Test each and every operation in Stellar
Some operations are not tested. I belive those are:

- StellarAllowTrustOp
- StellarPathPaymentOp
- StellarChangeTrustOp
- StellarCreatePassiveOfferOp
- StellarManageDataOp
- StellarManageOfferOp
- StellarPathPaymentOp

Could avoid https://github.com/trezor/connect/issues/382

### LINKED ISSUE #1699: Stellar: invalid signature if timebounds are set to 0/0
Trezor (both T1 and TT) conflate timebounds set to 0/0 with timebounds unset, and produce a signature for "timebounds unset". This is of course invalid if the incoming parameters are "set to 0/0"

---

### Original report:

**Describe the bug**
On [Stellar Laboratory](https://laboratory.stellar.org/) created a payment transaction without the upper time bound. The transaction was shown correctly on the device and signed. When validating the signature on the Laboratory, it was invalid. The signature validates correctly if the upper time bound is set.

Trezor One v1.10.1

 - Wallet software: Trezor Wallet
 - OS: Windows and MacOS (Trezor Bridge v2.0.27)
 - Browser: Chrome v91.0.4472.124

**To Reproduce**
1. Go to [Laboratory](https://laboratory.stellar.org/)
2. Select `public` network (top right corner)
2. Click on "Build Transaction" tab and fill in:
- Source Account
- Transaction Sequence Number (click on Fetch... button)
- Select Payment in Operation Type dropdown
- Destination
- Asset `native`
- Amount
3. Click "Sign in Transaction Signer" at the bottom
4. Scroll to the bottom of the page and click "Sign with Trezor"
5. Follow the Trezor instructions
6. Should see "Success!" message that the signature was added
7. Click on "Submit in Transaction Submitter" at the bottom of the page
8. Scroll to the bottom of the page to view checked signatures

image

**Expected behavior**
The upper time bound should not be required to validate the signature.

matejcik:
Changes most fields in Stellar protobuf to `required`. This mostly matches the Stellar XDR, so it should not introduce incompatibilities on caller sides.

The notable exception to "mostly matches" are the timebounds fields. In Stellar XDR, they are optional, and in the past, we would treat them as such. This PR makes timebounds required, and so rejects transactions without timebounds.
Per https://github.com/stellar/js-stellar-base/issues/438 omitting timebounds is strongly discouraged by documentation and existing implementations, so I feel that we do not need to support this feature.
Incidentally, this fixes #1699 

Several fields (`memo_type`, `signer_type`, `asset.type`) now use enums. There is another incompatibility: `StellarAssetType` is renamed to `StellarAsset`, so that `StellarAssetType` can be used as the name of the enum.

Enum values are binary-compatible with the previously existing fields.

ping @szymonlesisz or @matejkriz for the above, which require modifications in Connect:
1. if the incoming data structure does not have `timebounds`, it must be rejected
2. rename `StellarAssetType` to `StellarAsset` as appropriate. Or let me know if this is a bigger problem and we'll figure out some other way to name this.

---

all testcases have been regenerated with timebounds, and signatures checked against Stellar Laboratory. I renamed the testcases, which unfortunately breaks the UI diff.
fixes #2


common/protob/messages-stellar.proto
```diff
@@ -175,10 +199,10 @@ message StellarChangeTrustOp {
  */
 message StellarAllowTrustOp {
     optional string source_account = 1;     // (optional) source account address
-    optional string trusted_account = 2;    // The account being allowed to hold the asset
-    optional uint32 asset_type = 3;         // 1 = 4-character, 2 = 12-character
+    required string trusted_account = 2;    // The account being allowed to hold the asset
+    required StellarAssetType asset_type = 3;
     optional string asset_code = 4;         // human-readable asset code
-    optional uint32 is_authorized = 5;
+    required bool is_authorized = 5;
```
overcat, line 205: I don’t think it should be a bool value here.

https://github.com/stellar/stellar-core/blob/master/src/xdr/Stellar-transaction.x#L269

matejcik: good catch, but it's treated in code as bool anyway:
https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/stellar/operations/serialize.py#L30
https://github.com/trezor/trezor-firmware/blob/master/legacy/firmware/stellar.c#L989

i'm wondering why it isn't defined as an enum in the XDR?

overcat: It was indeed bool before, but it has changed with the update of the protocol. https://github.com/stellar/stellar-core/commit/29669910a7ef57db6b21dc86ee207fda3cbc1c39#diff-11f6a6577ee4d25e4eb1b1834bd71486cc73c26ad4eda68345286a27f5d74307R236

> i'm wondering why it isn't defined as an enum in the XDR?

I don't know why, but my personal guess is to support bitwise operators?

overcat: > good catch, but it's treated in code as bool anyway:
> https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/stellar/operations/serialize.py#L30
> https://github.com/trezor/trezor-firmware/blob/master/legacy/firmware/stellar.c#L989

Do you plan to modify it in this PR? At present, if passing `is_authorized` is not 0 or 1, the user cannot get the correct signature.

matejcik: > I don't know why, but my personal guess is to support bitwise operators?

originally yes, apparently, but the comment is pretty explicit and there's even a [review comment](https://github.com/stellar/stellar-core/pull/2881#discussion_r560476438) about it

i'll probably do an enum in our protobuf anyway, but it would be nicer to be in line with upstream about it

matejcik: As mentioned in your PR, I won't be changing this for an operation that is deprecated anyway. We'll leave that to if/when we implement the additional Stellar operations.

overcat:
Hi @matejcik, any updates here? If this PR can be merged, I can resolve #1691 and #1698 on this basis.

matejcik:
This is waiting for an internal review and presumably will be merged in September.

Please hold off on implementing the mentioned issues, it is currently unclear if we would accept such PRs.

overcat:
> This is waiting for an internal review and presumably will be merged in September.
> 
> Please hold off on implementing the mentioned issues, it is currently unclear if we would accept such PRs.

Thank you for your quick reply, I will discuss with you before implementation.

JFWooten4:
How to update our firmware should this merge to main?

core/src/apps/stellar/layout.py
```diff
@@ -44,18 +50,28 @@ async def require_confirm_init(
...
-async def require_confirm_timebounds(ctx, start: int, end: int):
-    await confirm_timebounds_stellar(ctx, start, end)
+async def require_confirm_timebounds(ctx: Context, start: int, end: int) -> None:
+    await confirm_properties(
+        ctx,
+        "confirm_timebounds",
+        title="Confirm timebounds",
+        props=(
+            ("Valid from (UTC)", str(start) if start > 0 else "[no restriction]"),
+            ("Valid to (UTC)", str(end) if end > 0 else "[no restriction]"),
```
mmilata, line 60: I know it's a change wrt to original code, but shouldn't we show human readable time here like T1 does?

mmilata (approved):
LGTM! UI diff ack. Thanks for getting rid of `confirm_timebounds_stellar`.

core/src/apps/stellar/layout.py
matejcik: we can't currently because we don't have epoch-to-datetime decoding.

and we will get it with uPy 1.14 (see #1535) so this will be fixed then

common/protob/messages-stellar.proto
```diff
@@ -5,14 +5,22 @@ package hw.trezor.messages.stellar;
 option java_package = "com.satoshilabs.trezor.lib.protobuf";
 option java_outer_classname = "TrezorMessageStellar";
 
+// https://github.com/stellar/stellar-core/blob/02d26858069de7c0eefe065056fb0a19bf72ea56/src/xdr/Stellar-ledger-entries.x#L25-L31
+enum StellarAssetType {
```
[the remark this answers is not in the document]
matejcik: correct

core/src/trezor/crypto/base32.py
```diff
@@ -1,4 +1,4 @@
-# Base32 implementation taken from the micropython-lib's base64 module
+# Base32 implementation taken from the micropython-lib'data base64 module
```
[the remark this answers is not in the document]
matejcik: fixed in rebase

matejcik:
rebased, fixed conflict with master.

this is now waiting for input from Connect side

matejcik:
TODO:
* [ ] fix tests
* [ ] add changelogs

bosomt:
@matejcik anything that can/need to be checked on our side ?

matejcik:
is Stellar Account Viewer in your scope? if yes, then general operation of Stellar

matejcik:
in particular please check that https://github.com/trezor/trezor-firmware/issues/1699 is fixed

bosomt:
@matejcik Stellar is not in our scope, 
but of course we will keep it in needs QA column and we will test it during release testing

PR #1765, opened 2021-08-18
Remove Lisk

matejcik:
Per https://lisk.com/blog/development/announcing-lisk-mainnet-v3-migration, the currently existing Trezor support for Lisk will stop functioning. This PR removes all Lisk-related functionality from Trezor codebases.

I'm keeping the entry in `misc.json` and `support.json` as historical data.


matejcik:
Cc @szymonlesisz for protobuf changes.

I removed entries from `MessageType`, instead of marking them `deprecated`. Will Connect be able to handle this, or should those entries be kept?

legacy/firmware/fsm_msg_common.h
```diff
@@ -75,7 +75,6 @@ bool get_features(Features *resp) {
   resp->capabilities[1] = Capability_Capability_Bitcoin_like;
   resp->capabilities[2] = Capability_Capability_Crypto;
   resp->capabilities[3] = Capability_Capability_Ethereum;
-  resp->capabilities[4] = Capability_Capability_Lisk;
```
andrewkozlik, line 78: It seems to me we should reduce the capability count by one and shift the indices of NEM, Stellar and U2F.

matejcik: 9c0a16a

szymonlesisz:
> I removed entries from `MessageType`, instead of marking them `deprecated`. Will Connect be able to handle this, or should those entries be kept?

@matejcik i guess it's ok to remove them. i will remove those methods from connect or we will make some placeholder with information about removed support, so those messages will never be called

tsusanka (approved):
Did not test, but LGTM.

prusnak (approved):
utACK

PR #1741, opened 2021-07-30
Updating the chainId to be 64bits to cater to large chain Id

joshuafernandes:
## Issues:
The chainId of the palm.io network is slightly larger than the maximum allowed by Trezor, this seeks to allow larger chain ids

PR #1771, opened 2021-08-19
feat(core,legacy): add support for Ethereum 64-bit chain_id

arbitrarylink:
Adds support for Ethereum 64-bit chain_id  This also makes sure that chain_id of 0 is rejected on core the same way it is rejected in legacy.

[Test-Results.txt](https://github.com/trezor/trezor-firmware/files/7017982/Test-Results.txt)

### DIFF
--- a/core/src/apps/ethereum/sign_tx.py
+++ b/core/src/apps/ethereum/sign_tx.py
@@ -17,7 +17,7 @@ from .layout import (
 )
 
 # maximum supported chain id
-MAX_CHAIN_ID = (0xFFFF_FFFF_FFFF_FFFF - 35) / 2
+MAX_CHAIN_ID = 2147483629
 
 
 @with_keychain_from_chain_id
--- a/core/src/apps/ethereum/sign_tx.py
+++ b/core/src/apps/ethereum/sign_tx.py
@@ -17,7 +17,7 @@ from .layout import (
 )
 
 # maximum supported chain id
-MAX_CHAIN_ID = 2147483629
+MAX_CHAIN_ID = (0xFFFF_FFFF_FFFF_FFFF - 35) / 2
 
 
 @with_keychain_from_chain_id
@@ -165,12 +165,16 @@ def check(msg: EthereumSignTx):
     if msg.tx_type not in [1, 6, None]:
         raise wire.DataError("tx_type out of bounds")
 
+    check_chain_id(msg)
     check_data(msg)
 
     # safety checks
     if not check_gas(msg) or not check_to(msg):
         raise wire.DataError("Safety check failed")
 
+def check_chain_id(msg: EthereumSignTx):
+    if msg.chain_id < 1:
+        raise wire.DataError("chain_id out of bounds")
 
 def check_data(msg: EthereumSignTx):
     if msg.data_length > 0:
--- a/core/src/apps/ethereum/sign_tx.py
+++ b/core/src/apps/ethereum/sign_tx.py
@@ -172,9 +172,11 @@ def check(msg: EthereumSignTx):
     if not check_gas(msg) or not check_to(msg):
         raise wire.DataError("Safety check failed")
 
+
 def check_chain_id(msg: EthereumSignTx):
     if msg.chain_id < 1:
         raise wire.DataError("chain_id out of bounds")
+
 
 def check_data(msg: EthereumSignTx):
     if msg.data_length > 0:
--- a/core/src/apps/ethereum/sign_tx.py
+++ b/core/src/apps/ethereum/sign_tx.py
@@ -165,17 +165,11 @@ def check(msg: EthereumSignTx):
     if msg.tx_type not in [1, 6, None]:
         raise wire.DataError("tx_type out of bounds")
 
-    check_chain_id(msg)
     check_data(msg)
 
     # safety checks
     if not check_gas(msg) or not check_to(msg):
         raise wire.DataError("Safety check failed")
-
-
-def check_chain_id(msg: EthereumSignTx):
-    if msg.chain_id < 1:
-        raise wire.DataError("chain_id out of bounds")
 
 
 def check_data(msg: EthereumSignTx):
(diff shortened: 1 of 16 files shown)


arbitrarygbg:
This supersedes #1741

core/src/apps/ethereum/sign_tx.py
```diff
@@ -165,13 +165,19 @@ def check(msg: EthereumSignTx):
...
 
+    check_chain_id(msg)
     check_data(msg)
 
     # safety checks
     if not check_gas(msg) or not check_to(msg):
         raise wire.DataError("Safety check failed")
 
 
+def check_chain_id(msg: EthereumSignTx):
```
matejcik, line 176: After reviewing the corresponding code in legacy, I don't think this is necessary (and actually modifies a feature).

When `chain_id` is 0 or unset, it is not hashed in (per https://github.com/trezor/trezor-firmware/pull/1771/files#R75), producing a pre-EIP155 signature. This check would disable pre-EIP155 signatures.

tests/device_tests/test_msg_ethereum_signtx.py
```diff
@@ -185,21 +185,22 @@ def test_ethereum_signtx_nodata(self, client):
             sig_v, sig_r, sig_s = ethereum.sign_tx(
                 client,
                 n=parse_path("44'/60'/0'/0/100"),
+                chain_id=1,
```
matejcik, line 188: please revert all these, with `chain_id` unset these are all pre-EIP155 signatures.

tests/device_tests/test_msg_ethereum_signtx_eip155.py
```diff
@@ -101,6 +101,46 @@
...
+    (
+        11297108109,
+        1,
+        (
+            22594216253,
+            "cccccef8f6636f316e69a1239436f9d61021501ba20ed72b6528c84b2f727980",
+            "614cb533a5d73cbf92ba4683c87c9238d0547572ee8f0f821ab1decdfc9639b8",
+        ),
+    ),
+    # Max uint64
```
matejcik, line 134: please also add a test for current `MAX_CHAIN_ID`, and `MAX_CHAIN_ID+1`

prusnak:
I think we could use this opportunity to make `MAX_CHAIN_ID` do what the name suggests. That is, we could throw an error when provided value in protobuf exceeds the maximum and simplify this:

https://github.com/trezor/trezor-firmware/blob/a39870449ffae07b53080a3a9e758ebb0e48a374/core/src/apps/ethereum/sign_tx.py#L153-L156

into this:

``` python
if msg.chain_id: 
  req.signature_v += 2 * msg.chain_id + 8 
```

and this:

https://github.com/trezor/trezor-firmware/blob/f93a8514e8e25b17b02360d9955c0a30999adf68/legacy/firmware/ethereum.c#L211-L217

into this:

``` c
if (chain_id) { 
  msg_tx_request.signature_v = v + 2 * chain_id + 35; 
} else { 
  msg_tx_request.signature_v = v + 27; 
} 
```

What do you think @matejcik?

matejcik:
My thoughts were in the opposite direction: always require chain_id, and then just send a single bit for `signature_v` and let the host compute the correct value. (both [Connect](https://github.com/trezor/connect/blob/83af30f73f4cfa7c099c55b2b0f8a103abc299c8/src/js/core/methods/helpers/ethereumSignTx.js#L30-L34) and [trezorlib](https://github.com/trezor/trezor-firmware/blob/master/python/src/trezorlib/ethereum.py#L78-L81) already do this).

I'm not sure if there are any Ethereum networks out there not using EIP-155, but we certainly don't support any, seeing as our Ethereum network database is keyed on chain_id.

At that point, perhaps we could go the extra mile and introduce `chain_id_bytes` field in case someone decides that 64 bits are not enough. (Ethereum seems to internally use 256-bit numbers, and all other numeric values in the message are `bytes` already.)

prusnak:
> My thoughts were in the opposite direction: always require chain_id

I let you decide on this. Either seems fine to me. I just want to decrease the number of source lines in our codebases.

> At that point, perhaps we could go the extra mile and introduce `chain_id_bytes` field in case someone decides that 64 bits are not enough.

I prefer not to do this, because this would break all existing implementations. OTOH uint32 and uint64 look exactly the same on the wire.

arbitrarylink:
Would it make sense for me to just remove the check_chain_id function, revert the tests to no longer set chain_id=1, add the two tests for MAX_CHAIN_ID and MAX_CHAIN_ID+1 and then leave all of the other suggestions for another future PR?

matejcik:
@arbitrarylink yes, that makes sense. Please do that.

matejcik:
thanks for your work. I'm merging this to a separate Ethereum branch for now, I want to tweak it a little and make it a part of a larger change set (see also #1781)

martinlsanchez:
@matejcik  quick question, are you planning to merge `ethereum-typing` into `master` for the next firmware release?

matejcik:
yes

matejcik:
@arbitrarylink could you share more details about `chain_id==0` ?
Reading the EIP-155 spec, it seems that the value is technically allowed, so I'm inclined to unify the behavior, instead of disallowing zero in core, allowing it in legacy.

Do you know of any reason why that would be a bad idea?

arbitrarylink:
> @arbitrarylink could you share more details about `chain_id==0` ?
> Reading the EIP-155 spec, it seems that the value is technically allowed, so I'm inclined to unify the behavior, instead of disallowing zero in core, allowing it in legacy.
> 
> Do you know of any reason why that would be a bad idea?

Currently Trezor One (Legacy) disallows chain_id of 0.  It returns an error of "Error: DataError: ChainId out of bounds".  This comes from the code ./legacy/firmware/ethereum.c: 444.

The code in this PR I had earlier was to disallow chain_id 0 in Trezor T (core) was to make the core behave like the legacy.  This code has been reverted as requested.

If you want both devices to support a chain_id of 0 you will have to modify the Legacy code.

matejcik:
i'm well aware.

I'm asking if you know something about one or the other behavior being "more correct" in some sense.

arbitrarylink:
There is an EIP that is not yet accepted that will explicitly make chain_id of 0 invalid.   https://eips.ethereum.org/EIPS/eip-3788

matejcik:
Thanks for the info!

This means that currently transactions with chain_id 0 are valid and actually can be used.

Issue #1764, opened 2021-08-17
Cardano sign transaction - button request other


vladimirvolek:
**Describe the bug**
All button requests are `ButtonRequest_Other` In Cardano sing transaction flow.
In the same flow in Bitcoin it's `ButtonRequest_SignTx` and `ButtonRequest_ConfirmOutput`

**Firmware version and revision**
2.4.1

**Desktop/smartphone setup (please complete the following information):**
 - Wallet software: Trezor Suite
 - OS: macOS
 - Browser Chrome
 - Version 92.0.4515.131

**Expected behavior**
No `ButtonRequest_Other` in the sign flow.

tsusanka:
We want to rework the button requests altogether. How critical is it for you? Also @gabrielKerekes would this be okay on your side?

matejcik:
related: #35

gabrielKerekes:
> Also @gabrielKerekes would this be okay on your side?

To be honest, I've never had to deal with the Button requests other than in the tests at some point - but we've already removed the expected responses. I think changing it shouldn't cause any problems on our end.

tsusanka:
@vladimirvolek do you think you could open a PR for this? It should be straightforward, just search for `.Other` inside the cardano app. Example here: https://github.com/trezor/trezor-firmware/blob/395324a8ad9399bacba2ebb8740d72971842d761/core/src/apps/cardano/layout.py#L85

It would be better if you do it, so it fits your needs.

tsusanka:
Let's solve this via #35 (the new button requests).

komret:
Seems like #35 is not moving forward and this bug is forcing some [weird workarounds](https://github.com/trezor/trezor-suite/pull/7964/commits/f243277668880fa4e58fb206a05f1ec9e075c157) in Suite. Can someone look into it, @Hannsek?

Hannsek:
Does Suite need this right now or it can wait few weeks?

komret:
Not urgently, we have the workaround, it's just ugly.

ibz:
Reopened due to #18501 which happened as a follow-up of https://github.com/trezor/trezor-firmware/commit/b8e2709ca8f141a5ded1ffdd5d20d4b388fe49d6

Issue #35, opened 2019-04-16
Rethink ButtonRequests


mroz22:
Hi guys, 

I might have found some inconsistencies worth refactoring, see:

ResetDevice call
- for model T responds with `ButtonRequest_ResetDevice`
- for model one responds with `ButtonRequest_ProtectCall`

BackupDevice call
- for model T responds with `ButtonRequest_ResetDevice`
- for model one responds with `ButtonRequest_ConfirmWord`

I would suggest to change it this way: 

ResetDevice call
- for model T responds with `ButtonRequest_ResetDevice`
- for model one responds with `ButtonRequest_ResetDevice`

BackupDevice call
- for model T responds with `ButtonRequest_ProtectCall`
- for model one responds with `ButtonRequest_ConfirmWord`

prusnak:
Let's also fix the case mentioned in https://github.com/trezor/trezor-firmware/issues/18

tsusanka:
@TomasZorvan also encountered this. I'm setting higher priority because of that.

tsusanka:
We will:

- Remove the `data` field, because it most likely not being used
- `code` stays
- We might add some some optional structures based on the usage (Shamirdata etc.), to be defined
- replace Other with Unknown and add a check that we do not send such ButtonRequestType (does it make sense to have it then?)
- investigate and add new types to ButtonRequestType
- considering adding an index to denote that we are on X-th such screen

- make Layout more intelligent and auto send the ButtonRequests - tracked in #

prusnak:
Removed the `ButtonRequest.data` field in https://github.com/trezor/trezor-firmware/pull/644

matejcik:
on TT it would be useful to have a distinct (set of?) ButtonRequest codes that precede PIN entry and possibly passphrase entry.

matejcik:
I'm currently thinking about "semantic ButtonRequests" that we talked about at some points.

From the layouting thing, ISTM the solution is to extend the `ButtonRequest` so:
```protobuf
message ButtonRequest {
	optional ButtonRequestType code = 1;  // technically deprecated but we probably want to keep this?
	required string type = 2;
	repeated KeyValue content = 3;

	message KeyValue {
		required string key = 1;
		optional string value = 2;
	}
}
```
IOW, reduce every layout to a set of string keys and string (**only!**) values.
in addition, a string (not enum) for "type", so that this is easily extensible and can be more descriptive.

no further structuring.
let's review the other possible types:
[...]

PR #1777, opened 2021-08-23
fix: Cardano button requests

vladimirvolek:
resolves: https://github.com/trezor/trezor-firmware/issues/1764

### DIFF
--- a/core/src/apps/cardano/layout.py
+++ b/core/src/apps/cardano/layout.py
@@ -79,7 +79,7 @@ async def confirm_sending(
         width_paginated=17,
         to_str="\nto\n",
         to_paginated=True,
-        br_code=ButtonRequestType.Other,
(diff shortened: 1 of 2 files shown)


matejcik:
Let's not add 20 new entries to the ButtonRequestType enum. Instead please see propsed solution in  #35, in particular https://github.com/trezor/trezor-firmware/issues/35#issuecomment-903836036

matejcik:
fwiw we could probably move this forward by using Cardano as a testbed for the feature, and then propagating the new usage back to other coins

tsusanka:
Ok let's do it the "new way" which we yet need to fine-tune. Apologies @vladimirvolek. @matejcik let's close both this and the issue?

vladimirvolek:
Closing after the discussion with @tsusanka.

PR #1781, opened 2021-08-24
remove Wanchain special handling

matejcik:
at some point this year, the Wanchain network hard-forked and introduced its own EIP-155 chainID distinct from the Ethereum network.

This means that we can remove logic which detects Wanchain based on presence of `tx_type` values, and rely on the common method of finding the data via looking up chainID.


prusnak (approved):
code utACK

matejcik:
superseded by #1794

Issue #1350, opened 2020-11-18
Extract fixtures to the common project


tsusanka:
Extract fixtures for device tests to _common_. We have already [done this for cardano](https://github.com/trezor/trezor-firmware/tree/e871a1c4ae56180c48a61a7ce9e48e0e5b7a762d/common/tests/fixtures/cardano), now we need to extend it to all other tests. Let's do this gradually, it's a lot of work.

somdipto:
hey , i would like to work on it , plz assign me this

PR #1784, opened 2021-08-25
Fixtures for Ethereum device tests

grdddj:
First version (draft how it could theoretically look like) of addressing https://github.com/trezor/trezor-firmware/issues/1350:
- refactoring (vectorizing) test_msg_ethereum_signtx.py
- NOTES for this file:
  - 2 test cases (below) are currently failing
  - `test_ethereum_signtx_newcontract 2` (testcase 8) is failing on Assertion error on `sig_r_hex` - in current state, it is being run after one failing test, it could be the reason
  - `test_ethereum_sanity_checks 4` is still failing (it is commented out in the current state - so it is not a problem)
  - there is a lot of possibilities for improvement (TODOs in code)


grdddj:
Notes after creating JSON fixtures for all the Ethereum tests (apart from `test_getpublickey.py`, where we have only one testcase):
- `test_verifymessage.py` and `test_signmessage.py` have a global variable `ADDRESS` (the latter also `PATH`), which seems to be influencing the tests, could be also put into the fixture, to have consistent results
- `test_signtx_eip1559.py` has two unvectorized testcases with `access_list` - if we would vectorize it, it would probably need to be included in its own fixture (to create the access_list info)
- `test_signtx_eip155.py` has one special unvectorized test with `data` - I was not sure whether to include it into fixture (however, it is only adding `data` field which will be always empty apart from this one)

I had one general **problem** in vectorizing the latest files - not being able to run the `pytest` tests reliably (and therefore validating all the testcases):
- for example running `pytest tests/device_tests/ethereum/test_signmessage.py`
- most of the time, the tests did not even start, or froze after couple of testcases (couple green dots or red F's appearing in the terminal)
- the rootcause seems to be a unstable connection with the Trezor device
  -  for hardware Trezor, all I am getting is the "No debuggable device found", no matter if it is T1 or TT
  - running it with Emulator through `tenv` is better for T1 (the first file `test_signtx.py` was running quite fine sometimes),  for TT it does not run at all
  - when looking at the emulator during the tests, some actions were happening there, and when the tests froze, there was mostly some dialog asking for input (like "Really send 10 ETH to 0xabc?")
  - I tried to run bridge from `tenv` (2.0.31) and also the default bridge on NixOS
  - as for the TT emulators, all the newest ones and master were tried, with no difference
  - @matejcik I guess it will be quicker for me to connect with you directly tomorrow to try resolving this
- looking for documentation of tests, I found out that there are couple of links to `tests/device-tests.md`, which does not exist

grdddj:
Commit a651dcd:
- addressing the feedback from @matejcik

common/tests/fixtures/ethereum/verifymessage.json
```diff
@@ -5,72 +5,72 @@
...
+                "address": "0xEa53AF85525B1779eE99ece1a5560C0b78537C3b"
             },
             "result": {
                 "sig": "752d283b3aea1eb44fd09203f4d5c430a6544e399b8500b02722b54325f6d8d457fd83460a31045cb0d6e8356240954ba072fdfe5cdb3f16d416e2acf1a180a51c"
```
matejcik, line 22: the `sig` is an input, not a result

`result` should probably be empty?

common/tests/fixtures/ethereum/signmessage.json
```diff
@@ -5,72 +5,80 @@
...
             "result": {
                 "sig": "f2c3d9c81fe35a6bc68f0ace01bd7209baf11e455e7335903e7ad2879ce5d7bc2199bf8bafc1efe1ff2cfdf2963cd5767911a641c9445cca6f37dc6da72fd1611b"
             }
         },
         {
-            "name": "7",
             "parameters": {
-                "msg": "This message has length 100!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
+                "msg": "This message has length 100!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!",
+                "address": "0x73d0385F4d8E00C5e6504C6030F47BF6212736A8",
```
matejcik, line 70: the `address` is not an input, it's a result

common/tests/fixtures/ethereum/getaddress.json
```diff
@@ -0,0 +1,71 @@
+{
+    "setup": {
+        "mnemonic": "all all all all all all all all all all all all",
+        "passphrase": ""
+    },
+    "tests": [
+        {
+            "name": "1",
```
matejcik, line 8: please remove `name`s here too

tests/device_tests/ethereum/test_sign_verify_message.py
```diff
@@ -0,0 +1,58 @@
...
+    "ethereum/verifymessage.json",
+)
+def test_verify(client, parameters, result):
+    res = ethereum.verify_message(
+        client, parameters["address"], bytes.fromhex(result["sig"]), parameters["msg"]
+    )
+    assert res is True
+
+
+def test_verify_invalid(client):
```
matejcik, line 48: let's modify this so that it's clearer what is going on

1. pick one of the vectors in JSON, use the data here
2. call `verify_message` on the unmodified vector, assert result is `True`
3. change one byte of `signature`, call `verify_message`, assert result is `False`
4. change the message but use the original signature, call `verify_message`, assert false

matejcik: > * pick one of the vectors in JSON, use the data here

pick the first vector and say in comment that it's the first vector from JSON

common/tests/fixtures/ethereum/verifymessage.json
grdddj: FIxed in 59c7eb7, the result made empty object

common/tests/fixtures/ethereum/signmessage.json
grdddj: Fixed in 59c7eb7

common/tests/fixtures/ethereum/getaddress.json
grdddj: Removed in 59c7eb7

tests/device_tests/ethereum/test_sign_verify_message.py
grdddj: Modified in 59c7eb7

matejcik: please also add the second failing condition: same signature, changed message

matejcik: (both are important. first checks that there exists a signature that can fail, the second checks that the signature depends on the message)

grdddj: What about also changing the address?

matejcik: yes, that's also a good idea

tests/device_tests/ethereum/test_sign_verify_message.py
```diff
@@ -50,26 +50,34 @@ def test_verify(client, parameters, result):
...
-    # Changing the signature, after which the verification should fail
-    parameters["sig"] = parameters["sig"][:-2] + "00"
+    # Changing the signature, expecting failure
     res = ethereum.verify_message(
         client,
-        parameters["address"],
-        bytes.fromhex(parameters["sig"]),
-        parameters["msg"],
+        address,
+        sig[:-1] + bytes.fromhex("00"),
```
matejcik, line 71: instead: `+ b"\x00"`

tests/device_tests/ethereum/test_sign_verify_message.py
grdddj: Message-changed-test-case in 0fdabfa

tests/device_tests/ethereum/test_sign_verify_message.py
grdddj: Changed in 1125280

matejcik (approved):
code LGTM

there's two broken tests on legacy. I believe you will need to add the `skip_t1` marker to them.

you will also need to regenerate the UI test fixtures

grdddj:
Notes from manual review of generated UI screenshots:
- https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1553557823/artifacts/master_diff/index.html
- `verifymessage` and `getpublickey`look exactly the same as before
- the rest also, but it has always has an **extra screen with passphrase entry** (I am not sure why)

matejcik:
> * the rest also, but it has always has an **extra screen with passphrase entry** (I am not sure why)

that's all right, it's because the json fixtures enable passphrase entry even if the passphrase used is `""`

matejcik:
(and this doesn't trigger for `verifymessage` because it's not using Trezor's keys so it doesn't need the passphrase to operate)

tests/device_tests/ethereum/test_sign_verify_message.py
```diff
@@ -0,0 +1,92 @@
...
+from trezorlib import ethereum
+from trezorlib.tools import parse_path
+
+from ...common import parametrize_using_common_fixtures
+
+pytestmark = [pytest.mark.altcoin, pytest.mark.ethereum]
+
+
+@parametrize_using_common_fixtures(
+    "ethereum/signmessage.json",
```
matejcik, line 28: for future reference (let's not touch this PR now):

This looks like it would fit on a single line.
The reason it's formatted like this is the trailing `,` -- if you have a trailing comma in the argument list, `black` will put each argument on a separate line and keep it that way. If you removed the comma, `black` would create a single line from this.

Generally we want to keep things on one line if possible.

matejcik:
merged, thanks :100:

grdddj:
> merged, thanks 100

Thank you as well for the guidance :)

Issue #1787, opened 2021-08-28
mimatched bound error when compiling trezor core on linux


achow101:
When compiling trezor core for linux (with `make build_unix`), I get the following build errors:

```
vendor/trezor-crypto/sha2.c:580:46: error: argument 2 of type ‘sha2_byte[]’ {aka ‘unsigned char[]’} with mismatched bound [-Werror=array-parameter=]
  580 | void sha1_Final(SHA1_CTX* context, sha2_byte digest[]) {
      |                                    ~~~~~~~~~~^~~~~~~~
In file included from vendor/trezor-crypto/sha2.c:33:
vendor/trezor-crypto/sha2.h:95:28: note: previously declared as ‘uint8_t[20]’ {aka ‘unsigned char[20]’}
   95 | void sha1_Final(SHA1_CTX*, uint8_t[SHA1_DIGEST_LENGTH]);
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
vendor/trezor-crypto/sha2.c:634:40: error: argument 2 of type ‘char[]’ with mismatched bound [-Werror=array-parameter=]
  634 | char *sha1_End(SHA1_CTX* context, char buffer[]) {
      |                                   ~~~~~^~~~~~~~
In file included from vendor/trezor-crypto/sha2.c:33:
vendor/trezor-crypto/sha2.h:96:27: note: previously declared as ‘char[41]’
   96 | char* sha1_End(SHA1_CTX*, char[SHA1_DIGEST_STRING_LENGTH]);
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1791, opened 2021-09-01
Fix GCC 11.1.0 compile errors in core

onvej-sl:
Fixes https://github.com/trezor/trezor-firmware/issues/1787.


prusnak (commented):
ACK `crypto/sha2.c`

not sure about `crypto/ed25519-donna/ed25519.c`

andrewkozlik (approved):

PR #1794, opened 2021-09-02
Type-checking for Ethereum app

matejcik:
includes #1771 which was already reviewed (2b3b75d)

includes and supersedes #1781 which already had a review from @prusnak 

there are following protobuf changes:
* from #1771, `chain_id` is enlarged to `uint64`. This is fully backwards-compatible.
* `chain_id` is set to `required`. This **removes support for non-EIP-155 networks**. Our support database is keyed to EIP-155 chain_id, so we do not officially support any non-EIP-155 network. It's unclear if any exist at all.
  - TODO: `core` should _probably_ disallow `chain_id==0`. It's not _invalid_ per se, far as I could find out, but per [this comment](https://github.com/trezor/trezor-firmware/pull/1741#issuecomment-894365490) it will probably simply not work with libraries. T1 already disallows it.
* `gas_price` and `gas_limit` set to `required`. Both T1 and TT had checks that would reject messages with `gas_price` or `gas_limit` unset, so while it's an incompatible change on the protobuf level, we know that no app could possibly have been sending such messages.
* a bunch of other fields are set to `required` that we know were required already (parameters to sign/verify, data chunk in TxAck...)

UI diff: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1556349947/artifacts/master_diff/index.html
the data_streaming case differs because we're specifying chain_id 1 so the transfer registers as Ethereum (previously UNKN)

TODO:
* [x] add restriction that `chain_id >= 1`
* [x] add correct changelogs
* [x] enable Palm network


matejcik:
> * `core` should _probably_ disallow `chain_id==0`. It's not _invalid_ per se, far as I could find out, but per [this comment](https://github.com/trezor/trezor-firmware/pull/1741#issuecomment-894365490) it will probably simply not work with libraries. T1 already disallows it.

ISTM the reason for disallowing `chain_id==0` was to simplify implementation, so that `0` internally indicates "EIP-155 disabled". I'm inclined to simplify code and enable chain_id 0 for T1 as well.

mmilata (approved):
I haven't been able to spot any problem. Together with the Stellar PR this makes core more than 70% typed:ok_hand: What about the zero chain_id?

matejcik:
i'm trying to ask around if there are reasons for/against chain_id 0

matejcik:
further information on chain_id 0:
https://ethereum-magicians.org/t/discussion-to-eip-3788-strict-enforcement-of-chainid/7001
https://github.com/ethereum/go-ethereum/issues/23152
https://eips.ethereum.org/EIPS/eip-3788

haven't read through it all yet

matejcik:
Conclusion seems to be: `geth` also uses `chain_id == 0` to indicate pre-EIP-155 transactions (codebase confirms this), so while nothing explicitly _disallows_ this value, realistically nothing works with that value either.

Pre-EIP-155 transactions can currently be used deliberately to produce cross-chain-replayable transactions. So by forcing EIP-155 on Trezor, we disallow the cross-chain usecase.
The above linked EIP-3788 aims to disallow this practice, but it's very new so it's unclear if it's going anywhere.

arbitrarylink:
If this PR gets approved, will the next release of the Firmware also include an updated version of Trezor Connect?

matejcik:
Final decision about chain_id:
- we are keeping `chain_id` as required field, thus disallowing non-EIP-155 cross chain replayable transactions. This is actually a security feature, we _don't_ want to be signing cross chain replayable transactions.
- we are explicitly disallowing `chain_id == 0` to prevent confusion if someone wanted to try signing a non-EIP-155 transaction

I also reorganized the sanity checks somewhat, so that `chain_id == 0` is checked in EIP1559 as well (arguably it's unnecessary there, but better safe than sorry)

@mmilata PTAL at the new commits

mmilata (approved):
New changes ACK.

Issue #1769, opened 2021-08-19
Mac and Linux generate different UI fixtures


matejcik:
as reported in #1683 and #1753 

Per @prusnak's investigation, different PNG encoding library is used, presumably with different compression level.

Options:
1. figure out the compression levels and try to force the same one
2. explicitly use libpng with same parameters on both platforms
3. instead of raw PNG data, calculate hashes of decoded image data

option (3) seems most straightforward

prusnak:
> 3\. instead of raw PNG data, calculate hashes of decoded image data

+1 for trying option 3; options 1 and 2 are far from being straightforward

matejcik:
@grdddj to implement option 3, you would need to look into `tests/ui_tests/__init__.py` function `_hash_files`, modify it to decode the image files and hash the decoded contents

PR #1795, opened 2021-09-02
fix(tests): decoding PNG files to fix Linux/Mac differences in encoding

grdddj:
Addressing https://github.com/trezor/trezor-firmware/issues/1769:
- when creating a file hash for UI fixtures, accounting for possible difference in PNG encoding algorithm (between Linux and Mac), so decoding the "real" image data and hashing them instead, to be consistent on both platforms


tests/ui_tests/__init__.py
```diff
@@ -9,6 +9,7 @@
 from _pytest.outcomes import Failed
 
 from .reporting import testreport
+from helpers import get_bytes_from_png
```
matejcik, line 12: use relative import, presumably from `.helpers` should work

tests/ui_tests/helpers.py
```diff
@@ -0,0 +1,110 @@
+import zlib
+import struct
+
+
+# Taken from https://pyokagan.name/blog/2019-10-14-png/, made into function and blacked
+def get_bytes_from_png(png_file: str) -> bytes:
```
matejcik, line 6: instead of implementing raw PNG reading (fragile), please use the Pillow library which we already have in dependencies

matejcik: this seems suitable: https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.tobytes

matejcik: Pillow-based impl should be small enough so that you don't need to create a new file for it, i think?

matejcik (changes requested):
please also regenerate the fixtures so that someone can check that they come out the same on Mac

grdddj:
Thanks for the suggestions, the latest force-push includes the simplifications and regenerated `fixtures.json`, so it can be tried by somebody with Mac

matejcik:
@prusnak can you confirm that Mac produces the same hashes now?

prusnak:
I tried the changes on macOS and it seems this helped for some cases (e.g. `test_u2f_counter.py`) but did not help for others (e.g. `test_op_return.py`).

I guess that

matejcik:
can you upload the recorded directory somewhere?

prusnak:
here: [screens.zip](https://github.com/trezor/trezor-firmware/files/7099133/screens.zip)

grdddj:
Thanks for providing the screenshots, there is a **difference between the UI screenshots** themselves from what I can see.

Below is the comparison of `test_opreturn`, first in "my" Linux environment and second in "Mac" - there are more screens there.

Hash of Linux one is `592e849159d5e02fed1c353030089106213951876221bd1d67e12b6ac8c3d077` and when I do hashing of the Mac screenshots, I get `a607ae83192cf7445d17270a42b9523c3865eefa18e29de1e6a608dc873cb4d4`

@prusnak - does the last hash correspond with what you are getting locally?

For the `test_u2f_counter` I also get the same hash on both screenshot folders - `19f77e2d284431da5fadac938f5822c1a6b17c07ee6c801085efd84974f3163f`

image

image

matejcik:
@prusnak these look like they are recorded with animations enabled.
please make sure you run `emu.py -a` for UI tests

prusnak:
> please make sure you run `emu.py -a` for UI tests

Argh! That was it.

grdddj:
> > please make sure you run `emu.py -a` for UI tests
> 
> Argh! That was it. The hashes match now.

Nice, thanks for verifying.

One other thing seems to be broken, the CI Pipeline's last step failed for some reason connected with UI tests:
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1557001773

prusnak:
It seems this PR is not necessary after all - it was my fault for not disabling the animations when generating the fixtures. 🤦 

However, I'd still merge the changes because SDL can optionally be linked against macOS-shipped PNG encoder and we might hit this issue again later.

grdddj:
> It seems this PR is not necessary after all - it was my fault for not disabling the animations when generating the fixtures. facepalm
> 
> However, I'd still merge the changes because SDL can optionally be linked against macOS-shipped PNG encoder and we might hit this issue again later.

From this https://github.com/trezor/trezor-firmware/pull/1683#issuecomment-901892448 it seems even with animations disabled, the hashes still differed between Linux/Mac

The disadvantage of merging it could be higher computational overhead of having to decode PNG, but these things are running on servers/computers and not on Trezor, so no big slowdown.

matejcik:
@gabrielKerekes could you verify that (a) you can still reproduce the problem with UI fixtures on a Mac, and (b) this PR solves it?

matejcik:
> One other thing seems to be broken, the CI Pipeline's last step failed for some reason connected with UI tests:
> [gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1557001773](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1557001773)

this is caused by an earlier failure that was not reported: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1557001748

the `prepare_ui_artifacts.py` file has its own copy of `_hash_files`.

@grdddj please modify the file so that it uses `_hash_files` from `tests.ui_tests`. It might be sufficient to do `from tests.ui_tests import _hash_files`; if not, you'll need to do some `sys.path` trickery.

matejcik:
you can also use other things from `tests.ui_tests`, namely `FILE_HASHES` (after calling `read_fixtures()` and `UI_TESTS_DIR / "screens"`

grdddj:
> this is caused by an earlier failure that was not reported: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1557001748

Does it also mean we will need to do something  with this, so next time the AssertionError will fail the CI?

matejcik:
That's a good idea, but i'm not sure what. Can you look into it?

grdddj:
> That's a good idea, but i'm not sure what. Can you look into it?

Sure. I imagine we could be somehow catching the errors in the definition of `core device ui test` CI stage

grdddj:
So it turns out this UI artifacts are being run in `after_script`, which does not fail the job, even if the script fails

Possibility:
- adding all the `after_script` stuff into `script`

https://github.com/trezor/trezor-firmware/blob/6f1005b155a7d79482dfee33506f7892b4e81c48/ci/test.yml#L12-L20

gabrielKerekes:
> @gabrielKerekes could you verify that (a) you can still reproduce the problem with UI fixtures on a Mac, and (b) this PR solves it?

I can confirm that by checking out a previous commit (cec1f39a90b8eff65fa2672f258829464ba28fd0) and running `pytest tests/device_tests -m cardano --ui=record` all the Cardano hashes changed. With this branch checked out no fixtures changed. I also tried changing how a stake delegation certificate is displayed and only the hashes of fixtures containing the certificate changed. So this PR does solve the problem 👍

matejcik (approved):
perfect, thanks for testing.

@grdddj please squash the fixups and rebase on current master so we can get it merged.

let's leave the CI issue aside now, it's unlikely that we would run into it again

PR #1797, opened 2021-09-06
New Lisk V3 support for Model T

hirishh:


core/src/apps/lisk/lisk32.py
```diff
@@ -0,0 +1,129 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+"""Reference implementation for Lisk32 addresses. (based on bech32)"""
```
matejcik, line 22: is this different from Bitcoin's bech32, apart from using a different `CHARSET`?

core/src/apps/lisk/sign_tx.py
```diff
@@ -0,0 +1,134 @@
...
+
+    try:
+        await _require_confirm_by_type(ctx, tx, asset)
+    except AttributeError:
+        raise wire.DataError("The transaction has invalid asset field")
+
+    amount = _get_tx_amount(tx, asset)
+    await layout.require_confirm_fee(ctx, amount, tx.fee)
+
+    txbytes = dump_message_buffer(tx)
```
matejcik, line 40: please add check that `txbytes == msg.transaction` and raise a `DataError` if there is mismatch

core/src/apps/lisk/lisk32.py
hirishh: The difference with bech32 are basically:
- The charset 
- The checksum algorithm is slightly different 

About the checksum: in bitcoin it includes also the HRP prefix. Instead in Lisk the HRP is just a prefix and is not part of the data being passed to the checksum function.. So that's why I had to adapt the bech32 code.

core/src/apps/lisk/sign_tx.py
hirishh: Good idea yes. will do

PR #1798, opened 2021-09-07
Fix Click 8.x problem when a ChoiceType argument has a default value

matejcik:
the following option definition:
```
@click.option("-t", "--script-type", type=ChoiceType(INPUT_SCRIPTS), default="address")
```
```
Options:
  -t, --script-type [address|segwit|p2shsegwit|pkh|wpkh|sh-wpkh]
```
works fine in Click 7.x
and we use this in several places

in 8.x, if `-t` is not specified, the following error appears:
```
Error: Invalid value for '-t' / '--script-type': 0 is not one of 'address', 'segwit', 'p2shsegwit', 'pkh', 'wpkh', 'sh-wpkh'.
```
This is because Click 8.0 first converts the _default value_ to the destination type (which is `int` for trezorlib 0.12.x and `IntEnum` in master), and then runs `ChoiceType.convert` on it again as a part of some subsequent step.

This is not strictly a Click bug, as their docs explicitly state that "The custom type should check at the top if the value is already valid and pass it through to support those cases."

### DIFF
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 _At the moment, the project does **not** adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). That is expected to change with version 1.0._
 
+## [0.12.4] - 2021-09-07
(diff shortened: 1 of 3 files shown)


matejcik:
CI running in corresponding master PR #1799

Issue #1800, opened 2021-09-07
Some device tests  are working even without enabling passphrase in setup_client


grdddj:
**Describe the bug**
In some tests (for example the first testcases in `tests/device_tests/test_protection_levels.py`), it is possible to delete the `passphrase=True` kwarg from the `setup_client` mark, and the tests are still passing. 

**Expected behavior**
Either this kwarg should not be there if it works without it, or it should be failing when we disallow the "passphrase mode".

grdddj:
Same happens in https://github.com/trezor/trezor-firmware/blob/f083beb3c26000033c41af2c0f5f1a0c3ae93669/tests/device_tests/test_protect_call.py#L52-L53

matejcik:
In case of `test_exponential_backoff_with_reboot`, I indeed see no reason to enable passphrase -- the test is checking that PIN prompt will exponentially increase wait time. This is not touching passphrase at all.

There's apparently a "FIXME TODO add passphrase tests" :)) But I believe that's sufficiently covered elsewhere.

Please modify `test_protect_call.py` as follows:
1. rename to `test_pin`
2. remove the class
3. remove `skip_t2` marker and make sure this runs on TT as well
4. replace `_some_protected_call` with `get_test_address`. No need to check the result, we are interested in the `expected_responses`

wrt (3), there is a difference between T1 and TT: on PIN query, T1 sends `PinMatrixRequest` and TT sends `ButtonRequest(code=PinDialog)`

you can build expected_responses like this:
```
is_t1 = client.features.model == "1"
client.set_expected_responses([
	(is_t1, messages.PinMatrixRequest),
	(not is_t1, messages.ButtonRequest(code=...)),
	messages.Address,
])
```

i'm not sure if TT also returns `PinException` so you'll need to test that

matejcik:
in case of `test_protection_levels`, the test is testing that the appropriate protections are executed, so setting passphrase there is right. We need to check that it's there though.

please do the following modifications:
1. remove class
2. add `assert pin_protection is True` and `assert passphrase_protection is True` to every test case where it applies; and `is False` where it doesn't
3. move the `setup_client` marker to top level `pytestmark = ...` so that all testcases are covered unless they have their own different marker
4. try modifying this so that it works on TT as well. instead of the above mentioned trick with `is_t1`, a more appropriate approach here would be something like:
   ```
   def pin_request(client):
	   if client.features.model == "1":
           return PinMatrixRequest
       else:
           ...
[...]

Issue #1110, opened 2020-07-15
Replace use_passphrase with setup_client in tests


tsusanka:
After #1107 is merged, let's replace `client.use_passphrase("passphrase")` with `setup_client(passphrase="passphrase")` where applicable.

tsusanka:
Also let's drop support for "True" in the passphrase argument and replace all usages of `passphrase=True` with strings. If there are tests that need passphrase enabled but not prefilled let's make them do that call explicitly.

matejcik:
done in #1801

PR #1801, opened 2021-09-07
Using setup_client mark for setting passphrase in device tests

grdddj:
Trying to address https://github.com/trezor/trezor-firmware/issues/1110:
- where possible, using `@pytest.mark.setup_client(passphrase="passphrase")` instead of calling `client.use_passphrase("passphrase")`
- also replacing some `passphrase=True` arguments into `setup_client` mark with `passphrase=""`, to explicitly set empty passphrase
  - `passphrase=True` remained in cases where we are setting passphrase to some value later
- during doing this,  issue https://github.com/trezor/trezor-firmware/issues/1800 was created, and these testcases were not touched at all


tests/device_tests/test_msg_recoverydevice_slip39_advanced_dryrun.py
```diff
@@ -65,7 +65,7 @@ def input_flow():
     )
 
 
-@pytest.mark.setup_client(mnemonic=MNEMONIC_SLIP39_ADVANCED_20, passphrase=True)
+@pytest.mark.setup_client(mnemonic=MNEMONIC_SLIP39_ADVANCED_20, passphrase="TREZOR")
```
matejcik, line 68: is `passphrase` even used here?

tests/device_tests/test_msg_recoverydevice_slip39_basic_dryrun.py
```diff
@@ -35,7 +35,7 @@
 ]
 
 
-@pytest.mark.setup_client(mnemonic=SHARES_20_2of3[0:2], passphrase=True)
+@pytest.mark.setup_client(mnemonic=SHARES_20_2of3[0:2], passphrase="TREZOR")
```
matejcik, line 38: is passphrase even used here?

tests/device_tests/test_session_id_and_passphrase.py
```diff
@@ -97,7 +97,7 @@ def test_session_with_passphrase(client):
     assert _get_xpub(client, passphrase="A") == XPUB_PASSPHRASES["A"]
 
 
-@pytest.mark.setup_client(passphrase=True)
+@pytest.mark.setup_client(passphrase="")
```
matejcik, line 100: please keep `passphrase=True` in this file

tests/device_tests/test_msg_wipedevice.py
```diff
@@ -25,7 +25,7 @@
 PIN4 = "1234"
 
 
-@pytest.mark.setup_client(passphrase=True)
+@pytest.mark.setup_client(passphrase="TREZOR")
```
matejcik, line 28: please keep `passphrase=True` here

matejcik (commented):
Maybe my comments here contradict something I said earlier, sorry about that.

I think the policy should be something like:
- if the test is using a specific passphrase (even `""`), that should be used as the argument
- if the test is setting multiple passphrases explicitly, then `passphrase=True`
- if the test wants passphrase enabled, but is not relying on a specific value (such as in `test_session_and_passphrase`), then `passphrase=True`

does this make sense to you, after seeing the usages?

---

I believe the high-level goal was to enable passphrase by default for every test (and explicitly disable in tests that don't want it). If we get that far, then removing every `passphrase=True` _with no other modification_ would result in correctly passing tests.

Let's not do that right now though.

tests/device_tests/test_msg_recoverydevice_slip39_advanced_dryrun.py
grdddj: Apparently not, the test works even when deleting the `passphrase` argument

tests/device_tests/test_msg_recoverydevice_slip39_basic_dryrun.py
grdddj: It is not, neither in the second testcase in this file

matejcik: i don't see a reason why passphrase _should_ be enabled, so please remove them (and the second file as well)

tests/device_tests/test_msg_recoverydevice_slip39_advanced_dryrun.py
grdddj: Fixed in 40530e8

tests/device_tests/test_msg_recoverydevice_slip39_basic_dryrun.py
grdddj: Fixed in 40530e8

tests/device_tests/test_session_id_and_passphrase.py
grdddj: Fixed in 40530e8

tests/device_tests/test_msg_wipedevice.py
grdddj: Fixed in 40530e8

grdddj:
Commit b0eb544:
- allowing all testcases from `test_protection_levels.py` and `test_pin.py` (renamed from `test_protect_call.py`) to be run on TT
- there is quite a lot of differences in expected responses, incorrect PIN handling etc. between T1 and TT
- implementing steps and ideas from comments in this issue: https://github.com/trezor/trezor-firmware/issues/1800

tests/device_tests/test_pin.py
```diff
@@ -0,0 +1,123 @@
...
+            [
+                messages.ButtonRequest(code=messages.ButtonRequestType.PinEntry),
+                messages.ButtonRequest(code=messages.ButtonRequestType.PinEntry),
+                messages.Address,
+            ]
+        )
+        get_test_address(client)
+
+
+def _test_backoff(attempts, start):
```
matejcik, line 82: ideally do not call non-test function "test"-anything. `_check_backoff_time` is better here

tests/device_tests/test_pin.py
```diff
@@ -0,0 +1,123 @@
...
+    """Helper to assert the exponentially growing delay after incorrect PIN attempts"""
+    if attempts <= 1:
+        expected = 0
+    else:
+        expected = (2 ** (attempts - 1)) - 1
```
matejcik, line 87:
```suggestion
    expected = (2 ** attempts) - 1
```

tests/device_tests/test_pin.py
```diff
@@ -0,0 +1,123 @@
...
+    got = round(time.time() - start, 2)
+    assert got >= expected
+
+
+@pytest.mark.skip_t2
+def test_exponential_backoff_t1(client):
+    for attempt in range(1, 4):
```
matejcik, line 94:
```suggestion
    for attempt in range(3):
```

together with the above suggestion the results will be the same

tests/device_tests/test_protection_levels.py
```diff
@@ -32,239 +33,362 @@
...
-    @pytest.mark.setup_client(pin=PIN4, passphrase=True)
-    def test_change_pin(self, client):
-        with client:
+pytestmark = pytest.mark.setup_client(pin=PIN4, passphrase=True)
+
+# General NOTE: we need to assert the value of client.features.passphrase_protection
+# at the end, because in case of TT, this value gets filled correcly only after
+# unlocking the device with the PIN.
+# Also, it may be needed to update the client info by client.refresh_features().
+# There are cases when it is even impossible to check the correct TT passphrase_protection,
```
matejcik, line 42: if the test doesn't unlock by itself, please use `client.ensure_unlocked()`. you might need to specify a pin sequence for the unlock.

maybe an overall better solution would be to create something like:
```
def _assert_protection_enabled(client):
	with client:
		client.use_pin_sequence(...)
		client.ensure_unlocked()
		assert pin_protection is True
		assert passphrase_protection is True
	client.clear_session()
```
and then call it at start of every test?

tests/device_tests/test_protection_levels.py
```diff
@@ -32,239 +33,362 @@
...
+            ]
+        )  # TrezorClient reinitializes device
+        device.apply_settings(client, label="nazdar")
+        assert client.features.passphrase_protection is True
+
+
+def test_change_pin(client):
+    with client:
+        assert client.features.pin_protection is True
+        # Models differ both in PIN sequence (TT requires one more input)
```
matejcik, line 83: that suggests that this should be two tests

tests/device_tests/test_protection_levels.py
```diff
@@ -32,239 +33,362 @@
...
+        device.change_pin(client)
+
+    assert client.features.passphrase_protection is True
+
+
+@pytest.mark.setup_client(pin=None, passphrase=False)
+def test_ping(client):
+    with client:
+        assert client.features.pin_protection is False
+        assert client.features.passphrase_protection is False
```
matejcik, line 117: we usually put these asserts before the with-block, and do only the things the with-block needs inside (so expected responses, pin sequences, passphrases, and the call (or calls) to Trezor)

tests/device_tests/test_protection_levels.py
```diff
@@ -32,239 +33,362 @@
...
+        client.set_expected_responses([messages.ButtonRequest, messages.Success])
+        client.ping("msg", True)
+
+
+def test_get_entropy(client):
+    with client:
+        assert client.features.pin_protection is True
+        # In case of TT, we need to supply PIN
```
matejcik, line 125: please split to two tests

tests/device_tests/test_protection_levels.py
```diff
@@ -32,239 +33,362 @@
...
+    # Even the order of responses is different here
+    if client.features.model == "1":
+        expected_responses = [
+            messages.ButtonRequest,
+            messages.PinMatrixRequest,
+            messages.PassphraseRequest,
+            messages.MessageSignature,
+        ]
+    else:
+        expected_responses = [
```
matejcik, line 261: please split into two tests.

please mark which of the ButtonRequests is the PIN request -- either by using _pin_request() or by specifying a code

tests/device_tests/test_protection_levels.py
```diff
@@ -32,239 +33,362 @@
...
+                messages.PinMatrixRequest,
+                messages.PinMatrixRequest,
+                messages.PinMatrixRequest,
+                messages.Success,
+                messages.Features,
+            ]
+        else:
+            client.use_pin_sequence([PIN4, PIN4, PIN4, PIN4])
+            expected_responses = [
+                messages.ButtonRequest,
```
matejcik, line 98: please mark which of these are PIN requests

tests/device_tests/test_protection_levels.py
```diff
@@ -32,239 +33,362 @@
...
-            amount=390000 - 10000,
-            script_type=messages.OutputScriptType.PAYTOADDRESS,
+    client.refresh_features()
+    assert client.features.passphrase_protection is True
+
+
+def test_verify_message(client):
+    with client:
+        assert client.features.pin_protection is True
+        # In case of TT, we need to supply PIN
```
matejcik, line 283: please split into two tests, and indicate which ButtonRequest is the PIN request

matejcik (changes requested):

tests/device_tests/test_protection_levels.py
matejcik: Tests that do not have both can have the assertions listed explicitly. If PIN is disabled, TT will correctly report `passphrase_protection` status at start.

If PIN is enabled and passphrase is not, we should probably try to unlock and check if passphrase protection is False? So maybe this function should have an argument, `passphrase_expected=True` that you can explicitly specify as False?

tests/device_tests/test_protection_levels.py
grdddj: Yes, I like the idea of the `_assert_protection()` function, so I will call it before the `with` block in all the tests, so after these assertions there are only test-relevant things

tests/device_tests/test_pin.py
```diff
@@ -0,0 +1,123 @@
...
+from trezorlib.exceptions import PinException
+
+from ..common import get_test_address
+
+PIN4 = "1234"
+BAD_PIN = "5678"
+
+pytestmark = pytest.mark.setup_client(pin=PIN4)
+
+# FIXME TODO Add passphrase tests
```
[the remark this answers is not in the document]
grdddj: Fixed in 699ae28

tests/device_tests/test_pin.py
grdddj: Fixed in 699ae28

tests/device_tests/test_pin.py
grdddj: Fixed in 699ae28

tests/device_tests/test_protection_levels.py
grdddj: Fixed in 699ae28

tests/device_tests/test_protection_levels.py
grdddj: Fixed in 699ae28

tests/device_tests/test_protection_levels.py
grdddj: Fixed in 699ae28

tests/device_tests/test_protection_levels.py
grdddj: Fixed in 699ae28

tests/device_tests/test_protection_levels.py
grdddj: Fixed in 699ae28

tests/device_tests/test_protection_levels.py
grdddj: Fixed in 699ae28

tests/device_tests/test_pin.py
```diff
@@ -0,0 +1,118 @@
...
+        with client, pytest.raises(PinException):
+            client.use_pin_sequence([BAD_PIN])
+            get_test_address(client)
+        _check_backoff_time(attempt, start)
+
+
+@pytest.mark.skip_t1
+def test_exponential_backoff_t2(client):
+    # Defining how long we expect each step to last from the beginning
+    # TODO: we should have some automatic way, but there seems to be no clear pattern
```
matejcik, line 100: This is because you are always checking against a fixed `START`, so the time consumed is cumulative.

The pattern is the same as with T1, so `device_wait_time = 2 ** attempt`, and `emulator_wait_time = device_wait_time - calculation_time`
The calculation time for T1 is 0 and for TT it seems to be 0.3.

Please modify so that the `start` time is updated before every attempt.
(you probably don't need to export the `start` variable out of the input_flow, the final `assert` can be left out.)

grdddj: Changed in 908b52d

grdddj:
When running the UI fixtures regeneration by `make test_emu_ui_record`, it failed in the T2 backoff test, when it finished quicker than we expected. It seems that the calculation time in TT should not be considered.

I will delete the 0.2 seconds we are supplying as the calculation_time

image

matejcik (approved):
LGTM, except the UI test will probably fail because you didn't delete the old fixture entries.

please squash&rebase and fix the fixtures during that, if necessary

PR #1802, opened 2021-09-08
ci(core/rust): check formatting, run unit tests

mmilata:


ci/test.yml
```diff
@@ -8,6 +8,9 @@ core unit test:
     - core unix regular build
   script:
     - nix-shell --run "poetry run make -C core test | ts -s"
+    # we need to generate qstrings for rust unit tests to work
+    - nix-shell --run "poetry run make -C core build_unix | ts -s"
```
matejcik, line 12: it would be better to add the required files to the artifact

mmilata: Fixed in 4aa1201.

core/Makefile
```diff
@@ -78,6 +78,9 @@ emu: ## run emulator
 test: ## run unit tests
 	cd tests ; ./run_tests.sh $(TESTOPTS)
 
+test_rust: ## run rs unit tests
+	cd embed/rust ; cargo test -- --test-threads=1
```
mmilata, line 82: btw this needs to be changed to `cargo test --features="test ui_debug" -- --test-threads=1` when the UI branch is merged

matejcik (approved):

Issue #1808, opened 2021-09-14
common: ethereum test fixtures followup


szymonlesisz:
Follow up after implementation common fixtures in [trezor-connect](https://github.com/trezor/connect/commit/087c6c2a3497c057abaac1617fd803f8f912b9d3)

- [ ] sign_tx.json (and variations) - consider using hexed values/strings in "value" field. javascript doesnt handle big numbers well (like 12345678901234567890)

- [ ] consolidate sign_tx vs sign_tx_eip1559 result:  `sig_r/sig_s` vs `sig_r_hex/sig_s_hex`

- [ ] consolidate paramaters.path: sometimes there is `m/` (getaddress.json) somtimes there is not (the rest)

- [ ] missing `EthereumGetPublicKey` fixtures

matejcik:
> * [ ]  missing `EthereumGetPublicKey` fixtures

we only have one :( if you have more, please contribute

szymonlesisz:
we do have 3, on seed `alkohol woman...`
https://github.com/trezor/connect/blob/develop/tests/__fixtures__/ethereumGetPublicKey.js

PR #1810, opened 2021-09-15
Improving consistency of ethereum fixtures

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1808:
- addressing consistency issues in Ethereum fixtures in device tests

What is still missing is  `sign_tx.json (and variations) - consider using hexed values/strings in "value" field. javascript doesnt handle big numbers well (like 12345678901234567890)`
- @szymonlesisz  is it enough to have `"value": "0XAB54A98CEB1F0AD2"` instead of `"value": 12345678901234567890`?


common/tests/fixtures/ethereum/getaddress.json
```diff
@@ -6,31 +6,31 @@
     "tests": [
         {
             "parameters": {
-                "path": "m/44'/60'/0'"
+                "path": "44'/60'/0'"
```
matejcik, line 9: Please modify this the other way.

Having the path start with `m/` makes it easier to auto-detect that it's a derivation path, and Cardano fixtures also use `m/` everywhere.

tests/device_tests/ethereum/test_getpublickey.py
```diff
@@ -17,29 +17,19 @@
...
-        )
-        assert (
-            res.xpub
-            == "xpub6D54vV8eUYHMVBZCnz4SLjuiQngXURVCGKKGoJrWUDRegdMByLTJKfRs64q3UKiQCsSHJPtCQehTvERczdghS7gb8oedWSyNDtBU1zYDJtb"
-        )
+
+@parametrize_using_common_fixtures("ethereum/getpublickey.json")
+def test_ethereum_getpublickey(client, parameters, result):
+    res = ethereum.get_public_node(client, parse_path(parameters["path"]))
+    assert res.node.depth == len(parameters["path"].split("/"))
```
matejcik, line 30:
```suggestion
	path = parse_path(parameters["path"])
    res = ethereum.get_public_node(client, path)
    assert res.node.depth == len(path)
```

matejcik:
> * is it enough to have `"value": "0XAB54A98CEB1F0AD2"` instead of `"value": 12345678901234567890`?

looking at https://github.com/trezor/connect/blob/develop/tests/__fixtures__/ethereumSignTransaction.js, it seems that's exactly what Connect is using. 

but please always use lowercase `0x` to indicate hex values.

tests/device_tests/ethereum/test_getpublickey.py
grdddj: Changed in 68e03e7

common/tests/fixtures/ethereum/getaddress.json
grdddj: Fixed in 68e03e7

grdddj:
Commit 68e03e7:
- changing the majority of `integer` fields to `hexadecimal` to be consistent with the abovementioned Connect file
- considerations/notes:
  - `chain_id` and `sig_v` are still `int`
  - `sig_r` and `sig_s` are `hex`, but without `0x` at the beginning
  - `data` is hex, but also without `0x`

szymonlesisz:
> is it enough to have "value": "0XAB54A98CEB1F0AD2" instead of "value": 12345678901234567890?

with or without 0x prefix, doesn't matter.
currently its hardcoded to make tests running:
https://github.com/trezor/connect/blob/develop/tests/__fixtures__/ethereumSignTransaction.js#L39

grdddj:
> all changed values are fine with me

Thanks for confirming. Is there something else we can do to improve these fixtures for Connect (in this PR and outside of it)?

matejcik (approved):
LGTM from my side. Let's wait a bit if there's anything else @szymonlesisz wants.

szymonlesisz:
Everything works great.
Got one nitpick in getpublickey.
`fingerprint` and `chain_id` are returned as number from the device and from connect. we can store them as numbers (so you can avoid this `int(result["child_num"], 16)` and i can avoid `parseInt`)

grdddj:
> `fingerprint` and `chain_id` are returned as number from the device and from connect. we can store them as numbers (so you can avoid this `int(result["child_num"], 16)` and i can avoid `parseInt`)

Sure, I have changed them to be `int`

szymonlesisz:
lgtm, thanks

PR #1811, opened 2021-09-15
implement exception handling for Rust

### LINKED ISSUE #1598: Add exception safety to Rust code
Currently Rust bindings calling Micropython functions do not install exception handlers. An exception raised by Micropython will skip and throw away the Rust context, which might lead to incorrectness in certain cases.

matejcik:
fixes #1598


core/embed/rust/src/micropython/list.rs
```diff
@@ -2,25 +2,27 @@ use core::convert::TryFrom;
...
     }
 
-    pub fn append(&mut self, value: Obj) {
+    pub fn append(&mut self, value: Obj) -> Result<(), Error> {
         unsafe {
             let ptr = self as *mut Self;
             let list = Obj::from_ptr(ptr.cast());
-            ffi::mp_obj_list_append(list, value);
+            // EXCEPTION: Will raise if allocation fails.
+            catch_exception(|| ffi::mp_obj_list_append(list, value)).map(|_| ())
```
jpochyla, line 25: Why the `map`?

core/embed/rust/src/micropython/obj.rs
```diff
@@ -134,12 +145,19 @@ impl TryFrom<Obj> for i32 {
...
+        //  - `int` is a mutable pointer of the right type.
+        //  - `obj` can be anything uPy understands.
+        // EXCEPTION: Can raise if `obj` is int but cannot fit into `cty::mp_int_t`.
+        let result =
+            catch_exception(|| unsafe { ffi::mp_obj_get_int_maybe(obj.as_ptr(), &mut int) });
+        match result {
+            Ok(true) => Ok(int.try_into()?),
+            Ok(false) => Err(Error::TypeError),
+            Err(_) => {
+                let int64: i64 = obj.try_into()?;
```
jpochyla, line 157: Why don't we just return an error?

core/embed/rust/src/micropython/obj.rs
```diff
@@ -150,10 +168,14 @@ impl TryFrom<Obj> for i64 {
     fn try_from(obj: Obj) -> Result<Self, Self::Error> {
         let mut ll: cty::c_longlong = 0;
 
+        // SAFETY:
+        //  - `ll` is a mutable pointer of the right type.
```
jpochyla, line 172: `ll` is not a pointer here.

core/embed/rust/src/micropython/runtime.rs
```diff
@@ -1,10 +1,90 @@
-use cstr_core::CStr;
+use core::mem::MaybeUninit;
 
-use super::ffi;
+use crate::error::Error;
 
-pub fn raise_value_error(msg: &'static CStr) -> ! {
+use super::{ffi, obj::Obj};
+
+pub fn raise_exception(err: Error) -> ! {
```
jpochyla, line 7: Nit, I'd rather make this unsafe.

core/embed/rust/src/micropython/runtime.rs
```diff
@@ -1,10 +1,90 @@
...
+    // doesn't play too well with Rust, we setup the non-local return pads in C, and
+    // execute `func` through a callback.
+
+    unsafe {
+        // First, we craft a wrapping closure that calls `func`. Because we are generic
+        // over the return type, we cannot pass the returned value over the FFI
+        // boundary, so we assign it explicitly in `wrapper`.
+        let mut result = MaybeUninit::zeroed();
+        let mut wrapper = || {
+            result = MaybeUninit::new(func());
```
jpochyla, line 33: We can use `MaybeUninit::write()` in 1.55.

core/embed/rust/src/protobuf/decode.rs
```diff
@@ -277,7 +282,7 @@ impl<'a> InputStream<'a> {
         let buf = self
             .buf
             .get(self.pos..self.pos + len)
-            .ok_or(Error::Missing)?;
+            .ok_or(error::end_of_buffer())?;
```
jpochyla, line 285: `ok_or_else` would be better here, same in the bottom two cases.

core/embed/rust/src/protobuf/defs.rs
```diff
@@ -127,7 +127,7 @@ pub fn find_name_by_msg_offset(msg_offset: u16) -> Result<u16, Error> {
         .filter(|def| def.msg_offset == msg_offset)
         .next()
         .map(|def| def.msg_name)
-        .ok_or(Error::Missing)
+        .ok_or(Error::KeyError(msg_offset.into()))
```
jpochyla, line 130: `ok_or_else`

core/embed/rust/src/protobuf/encode.rs
```diff
@@ -220,7 +219,7 @@ impl<'a> OutputStream for BufferStream<'a> {
                 *pos += len;
                 buf.copy_from_slice(val);
             })
-            .ok_or(Error::Missing)
+            .ok_or(error::end_of_buffer())
```
jpochyla, line 222: `ok_or_else`

core/embed/rust/src/protobuf/encode.rs
```diff
@@ -231,6 +230,6 @@ impl<'a> OutputStream for BufferStream<'a> {
                 *pos += 1;
                 *buf = val;
             })
-            .ok_or(Error::Missing)
+            .ok_or(error::end_of_buffer())
```
jpochyla, line 233: `ok_or_else`

core/embed/rust/src/protobuf/obj.rs
```diff
@@ -150,11 +152,12 @@ pub struct MsgDefObj {
 }
 
 impl MsgDefObj {
-    pub fn alloc(def: MsgDef) -> Gc<Self> {
+    pub fn alloc(def: MsgDef) -> Result<Gc<Self>, Error> {
         Gc::new(Self {
             base: Self::obj_type().to_base(),
             def,
         })
+        .map_err(Into::into)
```
jpochyla, line 160: I'd rather write it with `?` and a `let`. I suppose it would also be a bit more consistent with the rest of the codebase. Small nit.

core/embed/rust/src/protobuf/obj.rs
```diff
@@ -191,33 +194,38 @@ impl TryFrom<Obj> for Gc<MsgDefObj> {
...
 
 unsafe extern "C" fn msg_def_obj_attr(self_in: Obj, attr: ffi::qstr, dest: *mut Obj) {
     util::try_or_raise(|| {
-        let this=  Gc::<MsgDefObj>::try_from(self_in)?;
+        let this = Gc::<MsgDefObj>::try_from(self_in)?;
         let attr = Qstr::from_u16(attr as _);
 
-        if unsafe { dest.read() } != Obj::const_null() {
-            return Err(Error::InvalidOperation);
+        if !(unsafe { dest.read() }).is_null() {
```
jpochyla, line 207: Could we split this so it's a bit more readable?

jpochyla (commented):
Left a few codestyle comments. Overall I like it, nice!

core/embed/rust/src/error.rs
```diff
@@ -1,46 +1,59 @@
...
-    NotInt,
-    InvalidOperation,
+    MissingKwargs,
+    AllocationFailed,
+    CaughtException(Obj),
+    KeyError(Obj),
+    AttributeError(Qstr),
+    ValueError(&'static CStr),
+    ValueErrorParam(&'static CStr, Obj),
 }
```
jpochyla, line 17: It's a bit unfortunate that this ties the `Error` type to MicroPython, and pollutes all the modules with it, but I suppose that a single error type is a bit of a (very convenient) anti-pattern anyway.

core/embed/rust/src/micropython/list.rs
jpochyla: More generally, if we don't need the return value, I'd prefer throwing it away *in* the closure (`;` is an operator for that), and not only here, but in other uses of `catch_exception` as well.

core/embed/rust/src/protobuf/obj.rs
matejcik: fixed during rebase

core/embed/rust/src/error.rs
matejcik: true, but it turns out that we don't do basically any error handling within Rust and the Error type is almost always supposed to fall through to uPy

core/embed/rust/src/micropython/obj.rs
matejcik: Ah, because I got confused and thought that `mp_obj_get_int_maybe` only returns the smallint range. Good catch.

core/embed/rust/src/micropython/list.rs
matejcik: fixed along with one other case in b406bd0a0ff986392c2fccef336a7b9201734023

core/embed/rust/src/micropython/obj.rs
matejcik: fixed in b406bd0a0ff986392c2fccef336a7b9201734023

core/embed/rust/src/micropython/obj.rs
matejcik: fixed in b406bd0a0ff986392c2fccef336a7b9201734023

core/embed/rust/src/micropython/runtime.rs
matejcik: fixed in b406bd0a0ff986392c2fccef336a7b9201734023

core/embed/rust/src/protobuf/decode.rs
matejcik: fixed in b406bd0a0ff986392c2fccef336a7b9201734023

i'd honestly expect the compiler to do this for me -- esp. in cases where it's `ok_or_else(|| something(foo.into()))`

core/embed/rust/src/protobuf/defs.rs
matejcik: fixed in b406bd0a0ff986392c2fccef336a7b9201734023

core/embed/rust/src/protobuf/encode.rs
matejcik: fixed in b406bd0a0ff986392c2fccef336a7b9201734023

core/embed/rust/src/protobuf/encode.rs
matejcik: fixed in b406bd0a0ff986392c2fccef336a7b9201734023

core/embed/rust/src/protobuf/obj.rs
matejcik: fixed in 82f9995f1fa7aa44bf8535345f713925c2f788be

core/embed/rust/src/micropython/runtime.rs
matejcik: done in aa7f5c39289700db09d8b5a7fd351b8087061f37

mmilata (commented):
Looks great! Will approve after WIP commits are finished. Gonna be a bit tricky to track which upy functions might throw, but still massive improvement over no exception handling.

core/embed/rust/src/micropython/runtime.rs
```diff
@@ -1,10 +1,94 @@
...
+type ProtectedCallback = unsafe extern "C" fn(ProtectedArgument);
+
+fn split_func_into_callback_and_argument<F>(func: &mut F) -> (ProtectedCallback, ProtectedArgument)
+where
+    F: FnMut(),
+{
+    // Here we mono-morphize a version of `trampoline` for each type `F`, so it
+    // calls the correct `FnMut` impl, and cast `func` into its data part to use
+    // as the argument.
+    (trampoline::<F>, func as *mut _ as *mut _)
```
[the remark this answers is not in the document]
matejcik: my guess is that you must first cast `&mut T` to `*mut T`, without changing `T`, and only the second cast can go from `*mut A` to `*mut B` ?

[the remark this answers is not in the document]
jpochyla: Yup, what matejcik said.

matejcik:
@mmilata fixed the WIPs as 63b7936, PTAL

afterwards I'll rebase to master to fix conflicts

mmilata (approved):

mmilata:
Builds are failing because in micropython 1.17 the function `mp_obj_new_exception_arg1` became `static inline` micropython/micropython@022b8a7fea5e74b6b83895b2ff555ceab9af478c.

matejcik:
i'll probably hand-call it where needed

matejcik:
it seems that the Rust unit tests are now broken because symbols are missing:
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1599995792
@jpochyla can you take a look?

mmilata:
My understanding of the problem: previously we did not call any uPy code from unit tests but now we do so we need to link uPy objects to the test binary. I think we need to call `cargo test` from SCons and either:

* pass all the required `.o` files (is that what's in `obj_program`? or can we just use `trezor-emu-core` somehow?) as parameters to it, or
* build the test binary ourselves, like `trezor-emu-core` but with the test driver as `main()`.

jpochyla:
The `test` feature is missing here: `cargo test -- --test-threads=1`

jpochyla (approved):

mmilata:
Nice, I forgot that the UI branch already had this using `cc` crate. Looks like it still needs some path tweaks: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1605105556

PR #1821, opened 2021-09-22
refactor(core): decouple T1 button handling from touch

mmilata:
Hopefully there won't be Trezor with buttons *and* touch screen otherwise the handling in emulator would have to be rewritten.


prusnak:
> Hopefully there won't be Trezor with buttons _and_ touch screen otherwise the handling in emulator would have to be rewritten.

🤔 😆

prusnak (approved):
` core/embed/trezorhal` changes utACK

matejcik (approved):
LGTM

PR #1823, opened 2021-09-23
add cducoin

MachaUniverse:

PR #1824, opened 2021-09-23
add bitcoin clone

MachaUniverse:
cducoin is a bitcoin clone


prusnak:
Duplicate of https://github.com/trezor/trezor-firmware/pull/1823

MachaUniverse:
https://wiki.trezor.io/Developers_guide:Contributing

"If the coin you are adding is a fork of Bitcoin or other cryptocurrency we already support (in other words, new app is not needed) you can modify the definitions in the trezor-firmwarerepository and file a PR. In such case the coin does not have to be in TOP30"

prusnak:
Quoting from the same wiki:

> At the moment, we do not have the capacity to add new coins that do not fit the aforementioned category. Our current product goal is to unite what we support in firmware and in Trezor Suite and since firmware is way ahead of Suite we want to pause for a bit, implement the remaining coins into Suite and then consider adding new coins. This effectively means that our team will not be accepting any requests to add new cryptocurrencies.

prusnak:
Which CDU wallet supports Trezor so we can test the integration?

MachaUniverse:
Lookout for the cdu wallet via email as the coin is not publicly released yet.

PR #1829, opened 2021-09-29
Fix docker image build in CI

mmilata:
Looks like busybox's `sha256sum` needs two spaces. Updated alpine and nix versions while poking into this.


prusnak (approved):
LGTM (also checked Alpine hashes)

PR #1844, opened 2021-10-08
Improve Github issue templates

tsusanka:
Improves Github issue templates. It adds few more fields as per our [Backlog project](https://github.com/orgs/trezor/projects/42).

I am still not sure if we need labels for code/product/bug/community feedback, because we'll be doing it twice (once assign a column and then label). I will now try to find out if we can automate that.


matejcik:
I'm not convinced that we should have a "Community Feedback" template. We will categorize issues here, but we don't want to indicate that GitHub is a forum for community feedback.

tsusanka:
Feel free to check out what it looks like in my fork: https://github.com/tsusanka/trezor-firmware/issues/new/choose .

PR #1853, opened 2021-10-15
chore(common): update coins support status

mmilata:
Please review carefully, I'm doing this for the first time.

Was asked to add SHIB ERC20 token but there's no mention of it in our repo nor https://github.com/ethereum-lists/tokens.git master?


yura-pakhuchiy:
@matejcik I wonder why this PR removes support for xdai chain? It is rather popular eth side chain, eg. it is used for https://poap.xyz/, POAPs distributed on virtually every ETH related event nowadays.

matejcik:
@yura-pakhuchiy the problem is that the `xDAI` (100) symbol is in collision with Arbitrum-on-xDAI (200). We could support either only xDAI (100) and exclude Arbitrum, or if you have any sources on why the Arbitrum thing is "legit" (in the sense of, cannot cause user confusion between transacting on normal xDAI and Arbitrum).

The only metric available to us is marketcap, where xDAI is not winning any awards. So if you have a suggestion for a more relevant metric, please also let us know.

yura-pakhuchiy:
@matejcik xDAI (https://www.xdaichain.com/) and Arbitrum on xDai (https://www.xdaichain.com/for-developers/developer-resources/aox-arbitrum-on-xdai) are is the same project, so I would say both of them a legit. However I feel you might be confusing regular Arbitrum (which is quite large) with Arbitrum on xDai (which seems to be just a experimental network).

matejcik:
that honestly sounds like we could exclude xDAI(200) and keep xDAI(100)

PR #1861, opened 2021-10-21
Ethereum EIP-1559 transaction support in legacy

### LINKED ISSUE #1834: Support EIP-1559 in legacy
It seems MetaMask is using EIP-1559 for ETH transactions which are not supported by legacy at this moment, which makes T1 unusable with MetaMask. @matejcik before we decide if we want this could you or someone from the team please guesstimate how many mandays this would take?

mmilata:
Fixes #1834. TT implementation: #1653.

Since the access lists are serialized after streamed `data` we need to copy them out of the initial `EthereumSignTxEIP1559` message. With the current array length limits this eats <3KiB of static memory but since it compiles I guess we can afford it.

Please advise regarding dialog layout:
ethfee1 ethfee2 ethfee3

### DIFF
--- a/legacy/firmware/ethereum.c
+++ b/legacy/firmware/ethereum.c
@@ -51,6 +51,12 @@ static uint64_t chain_id;
 static bool eip1559;
 struct SHA3_CTX keccak_ctx = {0};
 
+static uint32_t signing_access_list_count;
+static EthereumAccessList signing_access_list[8];
+_Static_assert(sizeof(signing_access_list) ==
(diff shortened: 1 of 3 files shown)


ddaws:
Great to see this has been merged, thank you! Where can I see when this is marked for release and the release version?

prusnak:
> and the release version?

1.10.4 which will be released around the middle of December

ketys-from-meiro:
Really need this, cannot claim some airdrops and loosing money meanwhile.

PR #1868, opened 2021-10-25
crypto big endian support

2128675309:
This patch fixes multiple errors in the crypto routines, related to lack of big-endian support.  Correctness is measured by using test_check.

Big-endian builds already needed `CFLAGS="-DSPH_BIG_ENDIAN=1 -DSPH_LITTLE_ENDIAN=0 -DBYTE_ORDER=BIG_ENDIAN"` for some existing endian support; this does not change.

## Before Patch

x86_64 (little-endian, 64-bit): `test_check` passes 100%
MIPS (big-endian, 32-bit): `test_check` **passes 81%, 28 failures** (and 0 errors)

## After Patch

x86_64 (little-endian, 64-bit): `test_check` passes 100%
MIPS (big-endian, 32-bit): `test_check` passes 100%

## Change descriptions

- All adjustments re-use the `BIG_ENDIAN` / `LITTLE_ENDIAN` / `BYTE_ORDER` defines and swap functions already present in `sha2.h`, for consistency
- Fixed endian issues in `sha3` (this fixes many failed tests, as sha3 is a dependency to much other stuff)
- Fixed endian issues in `blake2b` and `blake2s`
- Added `test_check` coverage for blake2b-256-personalized using ZCash test vectors, to ensure the big-endian fixes to blake2b personalize are correct (`test_check` only covered Key, not Personalize)
- Fixed endian issue in `xmr_base58` and `xmr_subaddress_get_secret_key`
- Fixed bug in `test_check` slip39 word to index test (test was using `sizeof(char*)` instead of `strlen(char*)` as `word_length`)
- ~~Because the codebase allows easy include / exclude of different crypto functions, I opted to mildly replicate the endian defines into each `.h` that needs them~~
- Common `byte_order.h`, using the code from `sha2.h`, was created

~~A common header + consistent setting for all these endian defines and endian detection would eventually be ideal, but that further pulls in groestl rework (which used it's own set of defines, separate from sha2).  For now, the changes are the bare minimum just to get big endian systems to produce correct crypto calculations.~~

## Testing

All existing (little-endian) systems should be unaffected by any change. 

To test the big endian changes, you need to just build and run `test_check` on a big-endian system.  QEMU MIPS is sufficient, you can loosely follow steps to set it up here: https://aircrack-ng.blogspot.com/2018/10/to-be-or-not-to-be-using-qemu-to-run.html

If you test on QEMU, you will need to extend the default test timeout or you will likely get test timeout errors.

```
cd crypto/
export CK_DEFAULT_TIMEOUT=20
export CFLAGS="-DSPH_BIG_ENDIAN=1 -DSPH_LITTLE_ENDIAN=0 -DBYTE_ORDER=BIG_ENDIAN"
make tests/test_check
tests/test_check
```

### DIFF
--- a/crypto/blake2_common.h
+++ b/crypto/blake2_common.h
@@ -1,20 +1,39 @@
+
+#include "byte_order.h"
+
 static inline uint32_t load32(const void *src) {
   uint32_t w;
   memcpy(&w, src, sizeof w);
+#if BYTE_ORDER == BIG_ENDIAN
+  REVERSE32(w, w);
+#endif
   return w;
 }
 
 static inline uint64_t load64(const void *src) {
   uint64_t w;
   memcpy(&w, src, sizeof w);
+#if BYTE_ORDER == BIG_ENDIAN
+  REVERSE64(w, w);
+#endif
   return w;
 }
 
 static inline void store16(void *dst, uint16_t w) { memcpy(dst, &w, sizeof w); }
 
-static inline void store32(void *dst, uint32_t w) { memcpy(dst, &w, sizeof w); }
+static inline void store32(void *dst, uint32_t w) {
+#if BYTE_ORDER == BIG_ENDIAN
+  REVERSE32(w, w);
+#endif
+  memcpy(dst, &w, sizeof w);
+}
 
-static inline void store64(void *dst, uint64_t w) { memcpy(dst, &w, sizeof w); }
+static inline void store64(void *dst, uint64_t w) {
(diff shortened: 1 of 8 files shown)


andrewkozlik:
I ran `make style`, rebased over master, squashed the two commits, fixed the commit message and fixed a typo in the comments.

andrewkozlik (approved):

andrewkozlik:
@2128675309 thank you for your contribution!

PR #1873, opened 2021-10-26
Cardano derivation improvements

### LINKED ISSUE #1231: Should Cardano usage emit a new session id?
An example from @szymonlesisz:

1. I am using BTC:
```
-> Initialize(session_id=None)
<- PassphraseReq()
-> PassphraseAck(passphrase="A")
<- Features(session_id="ABCD")
```

2. I switch to Cardano:
```
-> Initialize(session_id="ABCD")  # I use the very same session_id received previously
<- PassphraseReq()  # passphrase is requested again though because of Cardano
-> PassphraseAck(passphrase="B")  # I enter different one
<- Features(session_id="ABCD")  # I get the same session id
```

@szymonlesisz was quite surprised the session id has not changed because under the `session_id="ABCD"` - we currently have two passphrases (one for BTC, one for Cardano).

Should we:
1. Create a new session anytime Cardano is used for the first time. In another words we would need to track whether the session is "normal" one or "Cardano" one.
2. Simply derive the Cardano seed anytime passphrase is being exchanged for other coins. This is ugly but it will simplify so many things and is properly a requirement for Suite implementation at some point anyway.
3. Not deal with it at the moment because there is not (hopefully) a mixed wallet yet anyway.

cc @matejcik

### LINKED ISSUE #1783: Fix Cardano compatibility issues
There are a number of different seed derivation schemes used in Cardano, i.e. the way in which the mnemonic is converted into a seed. The officially recommended derivation scheme is Icarus, see [1]. In Trezor we implement Icarus, but the implementation has a bug which manifests in case of a 24 word seed, see https://github.com/trezor/trezor-firmware/issues/1387#issuecomment-798790044. So someone who is trying to import or export a 24-word Icarus seed between Trezor and another Cardano wallet will fail [2]. Ledger has their own derivation scheme, which is BIP-39 based, and incompatible with everything else [3].

It would be helpful to users if Trezor supported all three schemes and in Suite users would be able to see all their accounts, similarly like we do with legacy, segwit and native segwit accounts in Bitcoin. The default account type is native segwit, but if a user has coins on a legacy account then they can still see them and spend them or create a legacy account if they wish. Similarly, in Cardano the native account type should come from Icarus derivation, but Suite would automatically show the other two account types if the user has previous transactions on them.

The way to implement this would be to add a field called `CardanoDerivationType` to each Cardano protobuf message that has an `address_n` field. The derivation type would dictate how to obtain the seed.
```
enum CardanoAddressType {
    ICARUS = 0;  // The future default.
    ICARUS_LEGACY = 1;  // The current implementation with bug.
    LEDGER = 2;
}
```

This issue is related to https://github.com/trezor/trezor-firmware/issues/1231 but is orthogonal. They can be implemented independently.

[1] https://github.com/cardano-foundation/CIPs/blob/master/CIP-0003/CIP-0003.md
[2] https://forum.cardano.org/t/all-ada-cardano-seem-gone-account-shows-0-with-trezor-model-t-ledger-nano-s/35784
[3] https://forum.cardano.org/t/swapping-from-ledger-to-trezor/50513/7

### LINKED ISSUE #1807: TT: passphrase_protection should be None when Trezor is locked
**Describe the bug**
Presumably due to some historical coincidence, the `passphrase_protection` field on TT is private, so its value is not accessible when Trezor is locked.

It would be nice to change this, but it's probably not worth the hassle with upgrading storage version.

When TT is locked (`Features.unlocked is False`), the value of `Features.passphrase_protection` is always set to `False` regardless of the true value. This is a bug. The field should _not_ be reported, i.e., the value should be `None`, to indicate that it is unknown.

### LINKED ISSUE #1842: Device with Shamir backup does not show progress bar
The `progress` callback while deriving seed is completely missing in the Shamir case.

**To reproduce:**

1. configure your device with Shamir backup, e.g., `trezorctl device setup -b shamir`
2. unplug and re-plug the device to clear any caches
3. use the seed, e.g., `trezorctl btc get-address -n m/44h/0h/0h/0/0`

**Expected result:**

Screen shows "Please wait..." and a spinner that slowly fills up.

(see what happens on a device with Standard backup)

**Actual result:**

Screen shows "Please wait..." heading, but the rest of the screen is empty for several seconds until the result is returned.

Related: #671 because repeated derivation of the mnemonic is cached on the C level so the progress bar won't appear the second time.

### LINKED ISSUE #1843: Deriving Cardano seed does not show progress bar
As of now, Cardano roots are not cached between invocations.

While waiting for the Cardano derivation (that might take several seconds of real time), Trezor screen should show "Please wait..." and a spinner.
Instead, Trezor shows the homescreen until the derivation is done.

matejcik:
With this PR, the same passphrase is used for Cardano derivation and the standard derivation (fixes #1231). If enabled via `Initialize(derive_cardano=True)`, the Cardano derivation runs at the same time as the standard derivation, which means that the start-up time is longer.

This is only relevant for BIP-39 seeds. With SLIP-39, a different derivation is used and it is available even without `derive_cardano=True`.

Additionally, the Ledger and Icarus derivations are implemented. Ledger derivation is available even in sessions that don't have `derive_cardano=True`. Icarus and Icarus-Trezor differ only when the mnemonic is 24 words long, in which case there is an additional delay at start of the session.

Every Cardano call now has an additional required field `derivation_type` that must specify one of Ledger, Icarus, Icarus-Trezor. This fixes #1783.

Also fixes #1843 as part of the wider Cardano derivation redesign.

**Small fixes folded into the PR:**
SLIP-39 progress display - 391c12f fixes #1842 
passphrase_protection incorrectly set to False - aa16042 fixes #1807

The changes in `crypto/` were already reviewed in #1849. @andrewkozlik, please have a quick look at 793b123db98751d8c6156798d1d9ab56fa1097ca which I had to do to in order to use the upstream test vector with 15-word mnemonic.

**INCOMPATIBILITY:**

This PR breaks all existing Cardano wallets. There is two parts to this:
1. If the user has BIP-39 seed, and `Initialize(derive_cardano=True)` is not sent, all Cardano calls will fail because the root secret will not be available.
   The only way to preserve compatibility here would be to use `True` as default -- but that would incur start-up delay for _all other_ wallets, which is undesirable.
2. The `cardano_derivation` argument is required on every Cardano call.
   We _could_ set a default value here. But it would have to be Icarus-Trezor (otherwise the wallets would silently start using a different derivation for people with 24-word seeds). And we kind of want the default to be Icarus going forward --- _especially_ for new users creating new accounts. (This is why `trezorctl` already defaults to Icarus.)
   This way, we pass the "default value" question onto Connect. Personally I believe that Connect should also require the argument and pass the decision to individual wallets, but this is up for debate with the Connect team.

TODO:
* [x] add test for Ledger derivation with `derive_cardano=False`
* [x] add test for SLIP-39 `derive_cardano=False`
* [x] add "fixes #1842" to 391c12f commit message


matejcik:
JFYI: as part of the rebase, I [tweaked](https://github.com/trezor/trezor-firmware/pull/1873/commits/f09db2283cb243b77da362a0906af8ab42c65d3f#diff-f014d9246998ed95f3b2fcc4df4ecb494f6b0652938fa4b83f91bf4194811a97L36-R39) the protobuf to account for a historic field. This has otherwise zero impact on the code.

PR #1881, opened 2021-11-01
chore(common): update chains and tokens

mmilata:
Note: `ethereum/chains` was bumped to only `master^` because ethereum-lists/chains@45635a60f90f52415de1433c89a9675595241866 adds chain with `eth:ETH` symbol which completely fucks up `coins_details.json` (replacing every reference to Suite w/ MEW+mycrypto). I'll figure out a workaround next time.


tsusanka (approved):

tsusanka:
LGTM. I hope @matejcik won't mind that I am merging this so we can deploy it.

PR #1882, opened 2021-11-01
chore(proto): backwards compatibility changes for protobuf defs

mroz22:
At the moment trezor-connect holds [4 sets of protobuf definitions](https://github.com/trezor/connect/tree/develop/src/data/messages) and loads them dynamically based on the version of currently connected device.

We would like to upgrade to the [new trezor-link version](https://github.com/trezor/trezor-link/pull/44). To be able to do that we kind of need to get rid of multiple messages.json files. 

So we started digging into "what got wrong" to find out why we needed to introduce this workaround in the past. Obviously it was because of some backwards incompatible changes to .proto files and our lack of assertiveness probably :joy_cat: 

We (me and @szymonlesisz) propose re-introducing some of the fields that used to exist in past. ack @matejcik 

So far I have added those that seem to be the most important to us. There were also some backwards incompatible changes in altcoins specific .proto files (eth, stallar) but we are thinking of increasing minimal supported version for methods specific to these coins in trezor connect instead. This should be discussed with @tsusanka probably?

### DIFF
--- a/common/protob/messages-management.proto
+++ b/common/protob/messages-management.proto
@@ -34,8 +34,7 @@ enum SafetyCheckLevel {
  */
 message Initialize {
     optional bytes session_id = 1;     // assumed device session id; Trezor clears caches if it is different or empty
-    // optional bool skip_passphrase = 2;  // removed as part of passphrase redesign
-    reserved 2;
+    optional boolean _skip_passphrase = 2 [deprecated=true]; // removed as part of passphrase redesign
     optional bool derive_cardano = 3;  // whether to derive Cardano Icarus root keys in this session
 }
 
@@ -67,7 +66,7 @@ message Features {
     optional bytes bootloader_hash = 14;        // hash of the bootloader
     optional bool imported = 15;                // was storage imported from an external source?
     optional bool unlocked = 16;                // is the device unlocked? called "pin_cached" previously
-//    optional bool passphrase_cached = 17;       // is passphrase already cached in session?  DEPRECATED
+    optional bool _passphrase_cached = 17 [deprecated=true]; // is passphrase already cached in session?
     optional bool firmware_present = 18;        // is valid firmware loaded?
     optional bool needs_backup = 19;            // does storage need backup? (equals to Storage.needs_backup)
     optional uint32 flags = 20;                 // device flags (equals to Storage.flags)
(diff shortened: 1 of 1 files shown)


tsusanka:
Thanks for bringing this up :+1:.

With respect to increasing minimal supported version, I am definitely up for it. Do you have an idea how old firmware we would stop supporting? Doesn't have to be exact, just to get an idea.

I also want to bring attention to https://github.com/trezor/trezor-suite/issues/3728. It is ACKed by Product that we will support only two years of firmwares in Suite (but all bootloaders!) if that could help with this issue in anyway.

 I'll leave the protobuf changes up to @matejcik.

common/protob/messages-common.proto
```diff
@@ -50,7 +50,8 @@ message Failure {
  */
 message ButtonRequest {
     optional ButtonRequestType code = 1;  // enum identifier of the screen
-    optional uint32 pages = 2;            // if the screen is paginated, number of pages
+    optional string _data = 2 [deprecated=true];
```
matejcik, line 53: I don't believe `data` was ever used for anything at all. Do you have usages in codebase that say something else?

common/protob/messages-management.proto
```diff
@@ -140,12 +141,20 @@ message ApplySettings {
     optional string label = 2;
     optional bool use_passphrase = 3;
     optional bytes homescreen = 4;
-//    optional PassphraseSourceType passphrase_source = 5;  DEPRECATED
+    optional PassphraseSourceType _passphrase_source = 5 [deprecated=true];
```
matejcik, line 144: would it work to declare this as `uint32` so that we can avoid the enum?

matejcik (commented):
we can do this (see comments in particular places), but my main question would be, why do you need these fields to exist?

wasn't the upgrade supposed to get us to a more sane version of protojs, where if an unrecognized field is received, it is ignored?

or do you have usages in code for these fields?

common/protob/messages-management.proto
mroz22: that should do imho.

common/protob/messages-common.proto
mroz22: Trezor connect does not appear to touch it directly but it simply [forwards any button request](https://github.com/trezor/connect/blob/1ea2accd177eaa0c2c2c9348b47b515161ae3193/src/js/device/DeviceCommands.js#L406) further to clients. So if it was ever used, it might still be somewhere out in the wild. So what is the optimal solution? keep the field there as deprecated? or make it reserved?

matejcik: but this is [a recent change](https://github.com/trezor/connect/commit/2faa0bcf5c639ed6740eb760381d896d758c0331), in particular, _after_ this field was already `pages`.

If you don't have any usages of `_data` in code, I'd prefer to leave this as-is.

matejcik (approved):

matejcik:
please squash the fixups so we can merge

PR #1883, opened 2021-11-01
Mypy tests for Monero app

grdddj:
Because Monero app is huge and contains a lot (initially 1468) mypy errors, creating a separate PR outside of https://github.com/trezor/trezor-firmware/pull/1876

First commit went through all easy-to-understand functions and their arguments and return values, and reduced number of mypy errors to 918


grdddj:
Closing as all the type-improvements suggested here were merged in https://github.com/trezor/trezor-firmware/pull/2219

PR #1890, opened 2021-11-03
chore(common/defs/fido): add coinbase u2f

### LINKED ISSUE #1866: Add coinbase's U2F app_id
**Is your feature request related to a problem? Please describe.**
Please add Coinbase U2F app_id to Trezor model T firmware to make it log better during U2F dialogues. 

image

matejcik:
Fixes #1866 

I was unable to verify validity of this data, because coinbase needs to analyze DNA samples of my ancestors up to the fourth generation in order to even complete registration. @bosomt please check if the result of this build works for you.


bosomt:
@matejcik ON it with branch 2cdf006c6ff63bb8a6e9368ebfc03eb30378897a build

bosomt:
@matejcik see video, 
tested with T1 no problem
tested with TT not able to login: https://youtu.be/apOEodNRN0w 

both same commit
```
trezor-fw-regular-1.10.4-2cdf006c.bin
trezor-fw-regular-2.4.3-2cdf006c.bin
```

common/defs/fido/coinbase.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "Coinbase",
+  "u2f": [
+    {
+      "app_id": "e27d61b4e99de0ed98163cb38b7af933c6665e5509e8490837055813778e236a",
+      "label": "coinbase.com"
+    }
+  ]
```
prusnak, line 8: I _think_ coinbase use FIDO2 not U2F:

```suggestion
  "webauthn": ["coinbase.com"],
```

prusnak:
Also this logo seems more appropriate: https://static.wikia.nocookie.net/logopedia/images/9/9a/Coinbase_2021_App.png

common/defs/fido/coinbase.json
matejcik: it would make sense, per the app_id being `sha256("coinbase.com")` which is what FIDO does -- BUT the authentication runs U2F because that's what the TT screen says

i'll probably have to test it myself, but coinbase doesn't allow me to go to settings just yet

prusnak: > BUT the authentication runs U2F because that's what the TT screen says

The question is whether TT even tries FIDO2 when there is an explicit u2f entry in the coinbase.json. Maybe the server supports both? 🤷

matejcik: the original report also shows "U2F Register" when the id is not known yet

matejcik: i'll just try both entries and see what happens

matejcik:
logo changed per suggestion

@bosomt please try with  e03b110 when the pipeline finishes

common/defs/fido/coinbase.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Coinbase",
+  "u2f": [
+    {
+      "app_id": "e27d61b4e99de0ed98163cb38b7af933c6665e5509e8490837055813778e236a",
+      "label": "coinbase.com"
+    }
+  ],
```
prusnak, line 8: I think you can drop this part (as can be seen from `core/src/apps/webauthn/knownapps.py` where exactly the same block is emitted twice)

prusnak (changes requested):

common/defs/fido/coinbase.json
matejcik: good point but in that case this modification is a no-op and we'll see the same result with TT

prusnak: Yep, for testing purpose it is fine to keep it.

bosomt:
> logo changed per suggestion
> 
> @bosomt please try with [e03b110](https://github.com/trezor/trezor-firmware/commit/e03b110433385f96cb425ea9331d15078d6b8c2e) when the pipeline finishes

**QA NOK**

tested with `trezor-fw-regular-2.4.3-e03b1104.bin `

Firefox: U2F dialogue is displayed over and over again after login is confirmed via device
Chrome: FIDO2 dialogue is displayed over and over again after login is confirmed via device dialogue.
Coinbase icon is displayed.

matejcik:
i'm afraid i'll have to hand this over to @andrewkozlik

common/defs/fido/coinbase.json
andrewkozlik: The choice of U2F vs. FIDO is up to the browser and the site. Trezor doesn't make any choice in this respect. The U2F is probably Firefox.

common/defs/fido/coinbase.json
andrewkozlik: I think the "u2f" can be dropped. IIRC it is implied by the "webauthn".

andrewkozlik:
My first guess is that a different Trezor/seed was used during registration in combination with incorrect error handling by CoinBase. Let's first check this by removing all registered tokens from the account and re-registering. If that doesn't help, then I will investigate later.

Is there some way that I could test this without having to undergo colonoscopy by CoinBase?

andrewkozlik:
The issues that @bosomt encountered may have been caused by the U2F counter.

prusnak:
Thanks Andrew. So let's drop the extraneous `u2f` section from the JSON and this is ACK from me.

andrewkozlik (approved):
utACK

prusnak (approved):
utACK

PR #1894, opened 2021-11-04
chore(vendor): update QR-code-generator

### LINKED ISSUE #1639: Update QR-code-generator
https://github.com/nayuki/QR-Code-generator

couldn't find a changelog, reviewing all commits shouldn't be too difficult though

I also notice they have a Rust version! and it's available as a crate too. Maybe investigate switching to that?

mmilata:
Fixes #1639. The only non-cosmetic change seems to be the improved finder pattern detection. Bumped to master since there are only variable name changes since the 1.7.0 release. Code size seems to stay the same.

~~Waiting for nayuki/QR-Code-generator#127.~~

### DIFF
--- a/core/.changelog.d/1639.changed
+++ b/core/.changelog.d/1639.changed
@@ -1,0 +1,1 @@
+Update QR-code-generator library version.
--- a/legacy/firmware/.changelog.d/1639.changed
+++ b/legacy/firmware/.changelog.d/1639.changed
@@ -1,0 +1,1 @@
+Update QR-code-generator library version.


matejcik:
why does legacy build fail but core succeeds? is it because core doesn't have `-Wunused-parameter` ?

mmilata:
> because core doesn't have `-Wunused-parameter` ?

Precisely. Adding it makes it fail in multiple places in our code though.

prusnak (approved):
utACK

PR #1896, opened 2021-11-04
fix(python/build_tx): make build_tx work again after blockbook protection upgrades

matejcik:
Also improve script type selection.

### DIFF
--- a/python/tools/build_tx.py
+++ b/python/tools/build_tx.py
@@ -26,6 +26,16 @@ from trezorlib.cli import ChoiceType
 from trezorlib.cli.btc import INPUT_SCRIPTS, OUTPUT_SCRIPTS
 from trezorlib.protobuf import to_dict
 
+SESSION = requests.Session()
+SESSION.headers.update({"User-Agent": "trezorlib"})
+
+BITCOIN_CORE_INPUT_TYPES = {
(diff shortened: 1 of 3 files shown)


matejcik:
folded a subrepo push here just to get the block_fixup check

python/tools/build_tx.py
```diff
@@ -26,6 +26,16 @@
...
 from trezorlib.protobuf import to_dict
 
+SESSION = requests.Session()
+SESSION.headers.update({"User-Agent": "trezorlib"})
+
+BITCOIN_CORE_INPUT_TYPES = {
+    "scripthash": messages.InputScriptType.SPENDADDRESS,
+    "pubkeyhash": messages.InputScriptType.SPENDP2SHWITNESS,
+    "witness_v0_keyhash": messages.InputScriptType.SPENDWITNESS,
+    "witness_v1_taproot": messages.InputScriptType.SPENDP2SHWITNESS,
```
prusnak, line 36: Shouldn't this be `SPENDTAPROOT`?

prusnak: Also `SPENDADDRESS` and `SPENDP2SHWITNESS` are mixed up?

matejcik: well :hankey:

Issue #1874, opened 2021-10-26
Add the secp256k1-zkp library to legacy


andrewkozlik:
See https://github.com/trezor/trezor-firmware/pull/1678#pullrequestreview-776263927

onvej-sl:
Implemented in https://github.com/trezor/trezor-firmware/pull/1897.

PR #1897, opened 2021-11-04
Onvej sl/legacy zkp

onvej-sl:
Implements https://github.com/trezor/trezor-firmware/issues/1874.

Regarding 05351eaf0136c203c9ab9b3f157e3b2527cb467f: The size of the context depends on architecture, it's 184 for 32-bit architecture and 208 for 64-bit architecture. `SECP256K1_CONTEXT_SIZE` must be at least the size (but doesn't have to be equal). In 05351eaf0136c203c9ab9b3f157e3b2527cb467f I set `SECP256K1_CONTEXT_SIZE` to 184 for device build and to 208 for emulator build. There is no problem to have `SECP256K1_CONTEXT_SIZE` same for all builds if you think it's better.

@hiviah Please review f6268db435cffb16c9dce97b29ce48cfd5c6243d and the changes in `legacy/firmware/trezor.c`
@andrewkozlik Please review the rest.


legacy/Makefile.include
```diff
@@ -74,6 +75,7 @@ CFLAGS   += $(OPTFLAGS) \
             -DRAND_PLATFORM_INDEPENDENT=1 \
             -I$(TOOLCHAIN_DIR)/include \
             -I$(TOP_DIR) \
+            -I$(TOP_DIR)/.. \
```
andrewkozlik, line 78: This looks like it's adding the root directory of the repository to the include paths. If that is the case, then why? We shouldn't need to be doing that.

legacy/Makefile.include
```diff
@@ -84,6 +86,8 @@ LDFLAGS  += -L$(TOP_DIR) \
             $(CPUFLAGS) \
             $(FPUFLAGS)
 
+ZKP_PATH = ../../vendor/secp256k1-zkp
```
andrewkozlik, line 89: Although this works, to be consistent I think we should respect the current convention of referencing 3rd party code through the `legacy/vendor` directory, which should have a symlink to the root vendor directory.

onvej-sl: Good point, fixed in e981bc7f1bed31e9f349aaf1b77c803a1f474f20.

legacy/Makefile.include
onvej-sl: Adding symlink to the root vendor directory fixes it. See e981bc7f1bed31e9f349aaf1b77c803a1f474f20.

crypto/zkp_bip340.c
```diff
@@ -40,33 +40,42 @@ int zkp_bip340_get_public_key(const uint8_t *private_key_bytes,
...
-    } else {
-      result = -1;
+    if (secp256k1_ec_pubkey_create(context_writable, &public_key,
+                                   private_key_bytes) != 1) {
+      result = 1;
     }
   }
 
+  if (context_writable) {
+    zkp_context_release_writable();
```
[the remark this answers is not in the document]
onvej-sl: Done in 3f98059e585e948f626d899ccd6fa86798bd63db and 71052bb86ca8e56400680d59a5a074dd5493600a.

andrewkozlik (approved):
LGTM. Haven't given much attention to the build script, but seems OK too.

invd:
@onvej-sl : I've noticed a side problem in the current Makefile that is touched by 0b29dbc887197b62061c250cfc00c830634f9429 :

```
In file included from ../vendor/secp256k1-zkp/src/secp256k1.c:12:
In file included from ../vendor/secp256k1-zkp/src/assumptions.h:12:
../vendor/secp256k1-zkp/src/util.h:75:12: fatal error: 'valgrind/memcheck.h' file not found
#  include <valgrind/memcheck.h>
           ^~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

From what I can see, the default Makefile settings assume that Valgrind libraries are installed on the system. `VALGRIND` is set to 1, which makes the crypto test code `#include` it. 
https://github.com/trezor/trezor-firmware/blob/fd0d1ed8718a7299f7057423a7b6d61674c67dd5/crypto/Makefile#L53-L57
Up until the secp256k1-zkp addition, the "normal" build did not require Valgrind libraries outside of the tests.

This has now changed when delivering the `$(CFLAGS)` to secp256k1-zkp since it also reacts to the `VALGRIND` define:
https://github.com/ElementsProject/secp256k1-zkp/blob/6db00f5b2e0986add852fa808ab485a25a9278f7/src/util.h#L74

Note that secp256k1-zkp includes the libraries on `#if defined(VALGRIND)` whereas the Trezor code uses `#if VALGRIND`, so the behavior for `VALGRIND=0` or an undefined `VALGRIND` is conflicting between them.

I think it is reasonable to require the Valgrind libraries for the regular builds but recommend
1. documenting the dependency
1. evaluating if the `VALGRIND` logic can be improved

prusnak:
I am OK with changing our logic from `#if VALGRIND` to `#ifdef VALGRIND` to make things simpler.

hiviah (approved):
Changes in `firmware/trezor.c` LGTM.

onvej-sl:
@hiviah  Did you review f6268db as well?

onvej-sl:
>  Up until the secp256k1-zkp addition, the "normal" build did not require Valgrind libraries outside of the tests.

I'm not sure, what you mean by "normal" build. Unless I mess something up, this build of secp256k1-zkp shouldn't be used anywhere other than in tests.

invd:
> > Up until the secp256k1-zkp addition, the "normal" build did not require Valgrind libraries outside of the tests.
> 
> I'm not sure, what you mean by "normal" build. Unless I mess something up, this build of secp256k1-zkp shouldn't be used anywhere other than in tests.

You're right, I was thinking of other legacy Makefile usage that is unrelated. Other parties may be using the crypto code for more than testing (it was formerly distributed as trezor-crypto for that purpose, after all) but I understand that this is not really in scope. We previously discussed this area in the internal ticket no 86, but I think that is not as relevant to this PR.

hiviah:
> @hiviah Did you review [f6268db](https://github.com/trezor/trezor-firmware/commit/f6268db435cffb16c9dce97b29ce48cfd5c6243d) as well?

Yes, seems fine.

onvej-sl:
> You're right, I was thinking of other legacy Makefile usage that is unrelated. Other parties may be using the crypto code for more than testing (it was formerly distributed as trezor-crypto for that purpose, after all) but I understand that this is not really supported.

I understand. Are you OK with https://github.com/trezor/trezor-firmware/commit/562164a328a8f0c1b3bb5c1ec8ded2d3ca93b73c? If someone wants to compile the code without valgrind, they must use `VALGRIND=0`, which I think is intuitive.

invd:
> I understand. Are you OK with [562164a](https://github.com/trezor/trezor-firmware/commit/562164a328a8f0c1b3bb5c1ec8ded2d3ca93b73c)? If someone wants to compile the code without valgrind, they must use `VALGRIND=0`, which I think is intuitive. However, valgrind is on by default, since the main use case is testing.

I think the `ifdef valgrind` doesn't work as intended since
* the variable use is case-sensitive
* for `ifdef VALGRIND` in combination with `VALGRIND ?= 1`, `VALGRIND=0` doesn't skip the ifdef. `VALGRIND=` works but is less intuitive, I think.

`ifeq ($(VALGRIND),1)` looks like a more straightforward way to test against the `1` default value.

onvej-sl:
@invd Is that OK? https://github.com/trezor/trezor-firmware/commit/dc785662fb94a2ab44bf9da9f596041d84eff579

invd:
> @invd Is that OK? [dc78566](https://github.com/trezor/trezor-firmware/commit/dc785662fb94a2ab44bf9da9f596041d84eff579)

Yes, I think that works. Thanks for the followup!

Issue #1862, opened 2021-10-22
emulator build fails on arm


vdovhanych:
**Describe the bug**
When building `build_unix` on the arm version of Linux (tried ubuntu and nixos), the build fails. 

I was building it using the docker container and those images from the docker hub. I have not tried the build natively on my M1 mac. 

Build fails on this step (both nixos and ubuntu)
```.....
.....
/home/project/core/embed/unix/main.c:110: undefined reference to `nlr_push'
/nix/store/4hn3w66fq9c4mdy5xl3mhm0y796zynhl-binutils-2.35.1/bin/ld: /home/project/core/embed/unix/main.c:146: undefined reference to `nlr_jump'
/nix/store/4hn3w66fq9c4mdy5xl3mhm0y796zynhl-binutils-2.35.1/bin/ld: build/unix/embed/unix/main.o: in function `main_clean_exit':
/home/project/core/embed/unix/main.c:411: undefined reference to `nlr_jump'
/nix/store/4hn3w66fq9c4mdy5xl3mhm0y796zynhl-binutils-2.35.1/bin/ld: build/unix/vendor/micropython/ports/unix/unix_mphal.o: in function `sighandler':
/home/project/core/vendor/micropython/ports/unix/unix_mphal.c:55: undefined reference to `nlr_jump'
collect2: error: ld returned 1 exit status
scons: *** [build/unix/trezor-emu-core] Error 1
make: *** [Makefile:160: build_unix] Error 2
(.venv) bash-4.4# 

```
I have tried to add suggested fix from this [issue](https://github.com/micropython/micropython/issues/4666) 
But it still failed at that exact step. 

**Firmware version and revision**
Emulator build from the current master

maxgillett:
I was eventually able to build natively on an M1 using clang 12.0.5, but only after forcing usage of the setjmp-based implementation of the [`gc_helper_get_regs` function]( https://github.com/micropython/micropython/blob/master/shared/runtime/gchelper_generic.c#L164).  Otherwise, clang errors complain that the [register variables](https://github.com/micropython/micropython/blob/master/shared/runtime/gchelper_generic.c#L129-L139]) need to be explicitly initialized (which I assume would impact garbage collection, as these variables would no longer reflect the current register variables).

prusnak:
> I was eventually able to build natively on an M1 using clang 12.0.5

Can you please send a pull request with the fix?

PR #1899, opened 2021-11-06
fix(core): fix (emulator) build on Apple silicon

maxgillett:
See https://github.com/trezor/trezor-firmware/issues/1862#issuecomment-962494531


prusnak:
Thanks! I superseded this PR with PR https://github.com/trezor/trezor-firmware/pull/1900 which uses the same way as upstream: https://github.com/micropython/micropython/blob/ff4f1f3ab395a6be426cc1b6ce0839bf65574f45/ports/unix/mpconfigport.h#L189-L192

Issue #1586, opened 2021-04-30
AOPP support


tsusanka:
Website: https://aopp.group
Short demo video: https://www.youtube.com/watch?v=heP-ufUke-g

21isenough:
Hey, dear Trezor Team

We are Pocket Bitcoin. A growing Bitcoin DCA-platform from Switzerland 🇨🇭

Many of our clients sign up to Pocket Bitcoin with their Trezors and receive their bitcoin straight to their cold storage. In order to sign up, the users needs to sign a message with their Trezor and for this particular case, AOPP support would be super convenient. 

Very happy to find this open issue here on the topic. Is there a timeline or an eta for AOPP support?

dspicher:
@andrewkozlik see the PR linked above for an initial attempt. Feedback would be greatly appreciated.

andrewkozlik:
Firmware PR ready: https://github.com/trezor/trezor-firmware/pull/1903

@qa Once the PR has been merged, please test that the following features work correctly:
- T1 support of pagination in SignMessage and Verify message for Bitcoin and Ethereum.
- T1 and TT show the address in SignMessage and VerifyMessage for Bitcoin and Ethereum.

PR #1903, opened 2021-11-08
AOPP support

andrewkozlik:
Implements https://github.com/trezor/trezor-firmware/issues/1586.
- Adds a new `bool` parameter to `SignMessage` in T1 and TT called `no_script_type`, which is `false` by default and when set to `true` `SignMessage` does not include script type information in the recovery byte of the signature, same as in Bitcoin Core.
- Implements pagination in `SignMessage` on T1 so that up to 1024 characters can be displayed.
- Adds a screen showing the address being used for signing in `SignMessage` and `EthereumSignMessage` on T1 and TT.
- Improves message signing UI to better distinguish the externally provided text from internally generated messages by:
  - Adding a "Confirm message" subheader to `SignMessage` and `VerifyMessage`.
  - Unifying font for `SignMessage` and `VerifyMessage` to monospaced. (`SignMessage` used to show the message in normal font and `VerifyMessage` in monospaced.)

I didn't tweak the message color as originally proposed in https://github.com/trezor/trezor-firmware/issues/1586#issuecomment-848653922, because after discussion with @matejzak we concluded that users probably wouldn't understand the color distinction. Nevertheless, I already added a parameter to `paginate_text()` so that changing the text color will be a one-line change if we choose to do this in the future.


common/protob/messages-bitcoin.proto
```diff
@@ -152,6 +152,7 @@ message SignMessage {
     required bytes message = 2;                                         // message to be signed
     optional string coin_name = 3 [default='Bitcoin'];                  // coin to use for signing
     optional InputScriptType script_type = 4 [default=SPENDADDRESS];    // used to distinguish between various address formats (non-segwit, segwit, etc.)
+    optional bool no_script_type = 5 [default=false];                   // don't include script type information in the recovery byte of the signature, same as in Bitcoin Core
```
prusnak, line 155: Isn't `[default=false]` redundant here?

mmilata (changes requested):
Overall this looks very good, however there seems to be a problem with the legacy pagination because it shows garbage at the end of `TestMsgVerifymessage::test_message_long`:

verylong

The PR changes some ButtonRequests from `ProtectCall` to `Other`, we might want to notify Suite of this change. TT currently always uses `Other`.

UI diff ack.

core/src/trezor/ui/components/tt/scroll.py
```diff
@@ -301,22 +301,33 @@ def create_tasks(self) -> tuple[loop.Task, ...]:
 def paginate_text(
     text: str,
     header: str,
+    subheader: str | None = None,
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/1903/commits/0b237b28d78658e4f7cb6267a47d97d76ac78f85

common/protob/messages-bitcoin.proto
andrewkozlik: Removed https://github.com/trezor/trezor-firmware/pull/1903/commits/ed627c4fc57bf6f621f53b99d76b37284866e324.

andrewkozlik:
> there seems to be a problem with the legacy pagination because it shows garbage at the end of `TestMsgVerifymessage::test_message_long`:

There was a bug in `split_message()`, which didn't correctly handle strings that are not null-terminated. Fixed https://github.com/trezor/trezor-firmware/pull/1903/commits/e7ed41b35661a813131ffeb73bf25e6ac502946a.

> The PR changes some ButtonRequests from `ProtectCall` to `Other`, we might want to notify Suite of this change. TT currently always uses `Other`.

Yes, I wanted to unify the T1/TT behavior. Do we have some agreed-upon way of communicating such changes? Just posting it on slack?

legacy/firmware/layout2.c
```diff
@@ -189,18 +189,17 @@ const char **split_message(const uint8_t *msg, uint32_t len, uint32_t rowlen) {
...
   memzero(str, sizeof(str));
-  strlcpy(str[0], (char *)msg, rowlen + 1);
-  if (len > rowlen) {
-    strlcpy(str[1], (char *)msg + rowlen, rowlen + 1);
-  }
-  if (len > rowlen * 2) {
-    strlcpy(str[2], (char *)msg + rowlen * 2, rowlen + 1);
+  for (int i = 0; i < 4; ++i) {
+    size_t show_len = strnlen((char *)msg, MIN(rowlen, len));
+    memcpy(str[i], (char *)msg, show_len);
```
prusnak, line 196: I am not against the change, just want to understand:

Was there anything wrong with the `strlcpy` approach?

andrewkozlik: From the man pages:
> Also note that strlcpy() and strlcat() only operate on true ''C'' strings. This means that for strlcpy() src must be NUL-terminated and for strlcat() both src and dst must be NUL-terminated.

prusnak (approved):

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -983,33 +982,26 @@ async def confirm_sign_identity(
...
-        text.mono(*chunks_intersperse(address, MONO_ADDR_PER_LINE))
-        await raise_if_cancelled(
-            interact(ctx, Confirm(text), br_type, ButtonRequestType.Other)
-        )
     else:
         header = f"Sign {coin} message"
-        font = ui.NORMAL
         br_type = "sign_message"
 
+    text = Text(header, new_lines=False)
```
matejcik, line 994: this should be using `confirm_address()` layout

matejcik: ...i wanted to make an issue to fix consistency of this, but it's probably not worth it while we expect to move to Rust UI.

Issue #556, opened 2019-09-23
Rename and order device tests


tsusanka:
We have all device tests in one folder and many of them are called `test_msg_*` even though they test a bit more than just the single message.

I suggest to structure this a little bit better by splitting the tests up to more subfolders (e.g. reset, recovery, ethereum, etc.) or we could follow the src/apps' structure.

matejcik:
the following would also be nice while we're messing with this:

* get rid of classes
* rename all `proto` usages to `messages`

PR #1908, opened 2021-11-10
Device tests cleanup

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/556

Final goals:
- separating the device tests into folders with common tests
- deleting the testing classes
- renaming all `proto` usages to `messages`


tests/device_tests/recovery/test_bip39.py
```diff
@@ -0,0 +1,215 @@
+# This file is part of the Trezor project.
```
matejcik, line 1: if this file is `skip_t2` as a whole, and we have a corresponding `_t2` variant, please also rename it to `test_bip39_t1`

there might be other cases that i'm not seeing now. If you know of similar others (i.e., (a) whole file is model-specific and (b) there is a corresponding file for the other model), we might want to rename them too

tests/device_tests/test_multisig_change.py
```diff
@@ -84,70 +86,70 @@ class TestMultisigChange:
...
         amount=55500000,
         prev_hash=TXHASH_b0946d,
         prev_index=0,
-        script_type=proto.InputScriptType.SPENDMULTISIG,
+        script_type=messages.InputScriptType.SPENDMULTISIG,
         multisig=multisig_in3,
     )
 
-    def _responses(self, inp1, inp2, change=0):
+    def _responses(inp1, inp2, change=0):
```
matejcik, line 140: this ... looks like it works, but it's wrong. the function should be declared a `@staticmethod`, otherwise the first argument only works by coincidence.

anyway please remove the whole class and instead move the elements to top level (inputs should probably be named in uppercase, `_responses` can stay as is)

tests/device_tests/recovery/test_bip39.py
grdddj: Couple files found and renamed in d15ee5c

matejcik:
looking great overall.
please rebase and fix conflicts.

let's also move the rest of the tests to the appropriate folders:

`bitcoin`:
* descriptors
* authorize coinjoin, get ownership proof
* getaddress_*
* getpublickey_*
* signmessage_*, verifymessage_*
* signtx_*
   * additionally, please rename signtx_<altcoin> to just test_<altcoin>
* multisig_*
* op_return
* nonstandard_paths

`misc`:
* sign identity
* cosi
* cipherkeyvalue
* getentropy
* get ecdh session key

`webauthn`:
* webauthn
* u2f counter

please move `test_stellar` to its own folder for consistency

i think "reset + backup + recovery" should live in the same folder, but i'm not sure what to call it. maybe just `reset_recovery` ?

tests/device_tests/test_multisig_change.py
grdddj: Class removed and global variables capitalized in 946823a

grdddj:
Force-pushed after rebase on `master`

matejcik:
`signtx.py` can be moved to `bitcoin`, it's a helper used only there (iirc)

otherwise the files in root look basically OK, or I can't think of a better structure for them at the moment

grdddj:
> `signtx.py` can be moved to `bitcoin`, it's a helper used only there (iirc)

It is also used in `test_protection_levels.py`, so under current conditions it should stay in the root folder

matejcik:
> EDIT: or would it be worth putting in into `bitcoin` and calling `from bitcoin.signtx import ...` in `test_protection_levels.py`?

nah, let's keep it as-is then

matejcik:
rebased on the SuperShamir PR, cleaned up the git history a little.

matejcik:
CI failures seem related to the clang/gcc issue tracked in #1973

Issue #1913, opened 2021-11-12
The mktable tool does not exit cleanly on unexpected curve values


invd:
The [mktable](https://github.com/trezor/trezor-firmware/tree/fd0d1ed8718a7299f7057423a7b6d61674c67dd5/crypto/tools#mktable) tool runs into a segmentation fault if the first command line parameter is not matching a curve specified in [curves.h](https://github.com/trezor/trezor-firmware/blob/fd0d1ed8718a7299f7057423a7b6d61674c67dd5/crypto/curves.h). 

As far as I'm aware, this is a rarely used internal tool with no direct impact on any firmware, so this is a low-priority issue.

```
./mktable 
Usage: ./mktable CURVE_NAME

./mktable incorrect-algorithm
Segmentation fault
```

The issue is caused by the following lines:
https://github.com/trezor/trezor-firmware/blob/fd0d1ed8718a7299f7057423a7b6d61674c67dd5/crypto/tools/mktable.c#L19-L21

`get_curve_by_name()` returns a `NULL` pointer, so the unchecked use of `info` for `info->params` triggers the issue. A null check with error message should fix this problem.

prusnak:
Fix in https://github.com/trezor/trezor-firmware/pull/1914

PR #1914, opened 2021-11-12
fix(crypto): fix wrong input handling in mktable

prusnak:
Fixes https://github.com/trezor/trezor-firmware/issues/1913


invd:
Looks good to me :+1: 

Side note for testing: `./mktable ed25519` is a case where the curve exists, but has no `params`.

PR #1916, opened 2021-11-14
fix(legacy): add Taproot account handling to GetAddress dialog

prusnak:
As reported by Boso, BIP86 is not handled in the GetAddress dialog

image (5)

image (6)

### DIFF
--- a/legacy/firmware/layout2.c
+++ b/legacy/firmware/layout2.c
@@ -79,19 +79,25 @@ static const char *address_n_str(const uint32_t *address_n,
     return _("Path: m");
   }
 
-  // known BIP44/49 path
+  // known BIP44/49/84/86 path
   static char path[100];
   if (address_n_count == 5 &&
       (address_n[0] == (0x80000000 + 44) || address_n[0] == (0x80000000 + 49) ||
-       address_n[0] == (0x80000000 + 84)) &&
+       address_n[0] == (0x80000000 + 84) ||
+       address_n[0] == (0x80000000 + 86)) &&
       (address_n[1] & 0x80000000) && (address_n[2] & 0x80000000) &&
       (address_n[3] <= 1) && (address_n[4] <= BIP32_MAX_LAST_ELEMENT)) {
+    bool taproot = (address_n[0] == (0x80000000 + 86));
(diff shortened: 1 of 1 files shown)


andrewkozlik (approved):

bosomt:
**QA OK**

FW: trezor-fw-regular-1.10.4-5178613d.bin 

i noticed one thing, BTC test account does not say TEST but REGTEST and it causes that account # does not fit into device display.
Checked old firmware and it said TEST and account # was displayed.
But not in latest firmware release 1.10.3


obrazek

obrazek

andrewkozlik:
> i noticed one thing, BTC test account does not say TEST but REGTEST and it causes that account # does not fit into device display.

This is because `coinBySlip44()` picks the first coin that has slip44 `coin_type` equal to 1. The ordering of the list is probably not deterministic. This is only a problem for testnets, which all have type 1, but mainnets shouldn't have collisions, which is ensured by `make defs_check`. I think we should treat `coin_type == 1` explicitly in `coinBySlip44()`.

andrewkozlik:
> Checked old firmware and it said TEST and account # was displayed.

Interesting. It seems to me that that firmware must have been at east two years old, because Regtest was added in 0f6e06e1e5 and Regtest comes before Testnet alphabetically.

bosomt:
> > Checked old firmware and it said TEST and account # was displayed.
> 
> Interesting. It seems to me that that firmware must have been at east two years old, because Regtest was added in [0f6e06e](https://github.com/trezor/trezor-firmware/commit/0f6e06e1e52fb29ef524cd26c8c45beae705e50d) and Regtest comes before Testnet alphabetically.

yes i compared it with 1.6.3 and 1.10.3 :) looks like noone noticed for two years ? :)

andrewkozlik:
> I think we should treat `coin_type == 1` explicitly in `coinBySlip44()`.

Done in https://github.com/trezor/trezor-firmware/commit/34ac730f93d6ecf634efa36ed9524cb2faf4cb68.

bosomt:
**QA OK**

test account number is displayed 
FW trezor-fw-regular-1.10.4-3b18c810.bin

PR #1917, opened 2021-11-15
Taproot in Trezor 1

andrewkozlik:
This PR implements the main Taproot functionality (https://github.com/trezor/trezor-firmware/issues/1656) in trezor-legacy and fixes some small bugs:
- Implements `GetAddress` for Taproot.
- Implements the BIP-341 common signature message computation and Taproot signing.
- Implements support for replacement transactions with Taproot inputs.
- Disables prevtx streaming if all internal inputs are Taproot.
- Adds warning dialog in `SignMessage` if a non-standard path is used.
- Fixes a bug in the transaction weight calculation for multisig inputs.
- Fixes a bug in the Testnet account label in the GetAddress dialog.

This essentially replicates https://github.com/trezor/trezor-firmware/pull/1891 in T1, except that T1 does not support external inputs.

This PR does not implement:
- Message signing and verification for Taproot.
- Multisig support for Taproot.

@matejcik please review the code overall.
@onvej-sl please review the code from a security perspective.

Ignore the first commit which will be removed during rebase after https://github.com/trezor/trezor-firmware/pull/1897 is merged into master.


legacy/firmware/signing.c
```diff
@@ -1390,15 +1541,13 @@ static bool signing_check_orig_input(TxInputType *orig_input) {
     // (aka BIP-143 script code), which is what our code expects here in order
     // to properly compute the legacy transaction digest or BIP-143 transaction
     // digest.
-    if (!fill_input_script_sig(orig_input)) {
-      fsm_sendFailure(FailureType_Failure_ProcessError,
-                      _("Failed to derive public key."));
-      signing_abort();
+    if (!derive_node(orig_input) || !fill_input_script_sig(orig_input)) {
```
matejcik, line 1544: couldn't we start using `script_pubkey` here?
(probably in a separate PR though)

or is there a possible conflict with host-provided script_pubkey?

matejcik (approved):
LGTM

legacy/firmware/.changelog.d/1656.added.3
```diff
@@ -0,0 +1 @@
+Show warning dialog in SignMessage if a non-standard path is used.
```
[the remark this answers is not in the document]
andrewkozlik: 0d42506df79bff5abcd5b3486522f1e7c2f264c2

legacy/firmware/signing.c
andrewkozlik: > couldn't we start using `script_pubkey` here? (probably in a separate PR though)

I think the problem is that the BIP-143 script code is not the actual scriptPubKey of this input, it's like the _legacy_ scriptPubKey that would correspond to the public key. I know, weird. See details of Item 5 in https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification.

> or is there a possible conflict with host-provided script_pubkey?

That wouldn't be a problem, because the host-provided script_pubkey is only for external inputs. For one we (logically) don't sign external inputs, for another we don't even support external inputs in T1.

matejcik: right, this thing!
it would have been much better if this was a separate global, but that's neither here nor there and might not be worth messing with anyway.

andrewkozlik: It actually makes some sense to have this in script_sig, because when you compute the transaction digest for legacy signing, you put this scriptPubKey exactly in the spot where the scriptSig would belong in the serialized transaction.

onvej-sl (approved):

prusnak:
Needs rebase after https://github.com/trezor/trezor-firmware/pull/1897 has been merged

PR #1924, opened 2021-11-18
feat(shell.nix): adding pyright as a dependency

grdddj:
Because of new type-tests, adding pyright as a dependency

PR #1925, opened 2021-11-18
style(all): reformat with clang-format 13.0

mmilata:
New clang-format does weird thing with `} else // comment<lf> if (...) {`: https://github.com/trezor/trezor-firmware/commit/7a35984591e2efbaa38bd6ef5a96221a89052677 . I don't know [which new option](https://releases.llvm.org/13.0.0/tools/clang/docs/ReleaseNotes.html#clang-format) disables this behavior, if any, so I moved the comments inside the `if` body instead.

Requires #1924.

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -65,6 +65,7 @@ stdenv.mkDerivation ({
     pkgconfig
     poetry
     protobuf3_6
+    pyright
     rustfmt
     rustStable
     wget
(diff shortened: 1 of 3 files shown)


matejcik:
to me personally this placement of comments looks better so utACK

prusnak:
Please resolve the merge conflict

prusnak (approved):

PR #1927, opened 2021-11-19
Fix CI check for protobuf problems, fix protobuf problem

matejcik:


tools/build_protobuf
```diff
@@ -71,6 +72,7 @@ do_rebuild() {
 do_check() {
     # rebuild protobuf in tmpdir and check result against specified directory
     local TMPDIR=$(mktemp -d proto-check.XXXXXX)
+    trap "rm -r $TMPDIR" RETURN
```
mmilata, line 75:
```
In tools/build_protobuf line 75:
    trap "rm -r $TMPDIR" RETURN
                ^-----^ SC2064: Use single quotes, otherwise this expands now rather than when signalled.
```

mmilata (approved):
Shellcheck complains, otherwise looks good.

tools/build_protobuf
matejcik: AFAICT the usage is correct -- TMPDIR never changes so I am perfectly OK with expanding now
or do you see another issue?

prusnak: We want to have it expanded now, since the variable is local

mmilata: (it's not a problem just fyi)

matejcik: but it traps `RETURN` so on function exit so this should exist anyway
interesting point though. would `trap RETURN` run in the function scope?

prusnak: We can add `# shellcheck disable=SC2064` on the preceding line

matejcik:
CI fails only on the nixpkgs thing, so I'm merging this

PR #1935, opened 2021-11-22
ci: also build legacy firmware with MEMORY_PROTECT=1

mmilata:
Runs the "production" build, then make clean, then the developer build. No artifacts retained. I tried to avoid adding 4 more jobs.

@hiviah suggested to always build bootloader with `MEMORY_PROTECT=1` but this raises question how to build the files common among firmware and bootloader. Also I would prefer not to introduce another parameter passed to the build, nor do some major build refactoring if possible.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -187,7 +187,7 @@ legacy fw regular build:
   needs: []
   script:
     - nix-shell --run "export MEMORY_PROTECT=1 && poetry run legacy/script/cibuild"
-    - nix-shell --run "make -C legacy clean"
+    - nix-shell --run "poetry run legacy/script/setup"
     - nix-shell --run "export MEMORY_PROTECT=0 && poetry run legacy/script/cibuild"
     - nix-shell --run "poetry run make -C legacy/demo"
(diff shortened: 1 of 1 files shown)


hiviah (approved):
I think this is good solution. The extra build adds just 2 seconds when I test it locally. It is possible to build bootloader and firmware with separate MEMORY_PROTECT flags, but this is cleaner and the 2 seconds are not worth extra complexity.

prusnak (approved):

ci/build.yml
```diff
@@ -185,10 +185,10 @@ crypto build:
 legacy fw regular build:
   stage: build
   needs: []
-  variables:
-    MEMORY_PROTECT: "0"
   script:
-    - nix-shell --run "poetry run legacy/script/cibuild"
+    - nix-shell --run "export MEMORY_PROTECT=1 && poetry run legacy/script/cibuild"
+    - nix-shell --run "make -C legacy clean"
```
matejcik, line 190: `make clean` is not enough.
that should _probably_ be fixed, but for now you should run `legacy/script/setup`
dtto below

mmilata: Fixed in 81e9d28.

prusnak (approved):

matejcik (approved):

PR #1938, opened 2021-11-23
Keep bootloader image in non-production builds

hiviah:
This is supposed to catch early cases when `PRODUCTION=1` build would overflow, but doesn't overflow for developer since linker normally drops the bootloader image, as `check_and_replace_bootloader` is ifdef-ed out.

Obviously the size still differs a bit from full production build, but not by so much.


prusnak:
Quote from Slack:

> Currently free `12280` bytes in `flash` and `57344` bytes in `flash2` for the production build.

So it does not make sense to move bootloader (which is currently 94208 bytes) from flash1 to flash2.

mmilata (approved):

PR #1940, opened 2021-11-24
ci: introduce gitlab cache for vnev

vdovhanych:
This introduces caching of `.venv` with GitLab cache. Related to this [issue](https://github.com/satoshilabs/devops/issues/49)

I hope I didn't miss any job that is installing poetry and would benefit from this. Also, this type of cache is per branch it means that the first time branch runs, the ci will create cache for it and use that next time that branch runs in the ci.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -1,5 +1,12 @@
 image: registry.gitlab.com/satoshilabs/trezor/trezor-firmware/trezor-firmware-env.nix
 
+# Caching
+.gitlab_caching: &gitlab_caching
+  cache:
+    key: "$CI_COMMIT_REF_SLUG"
+    paths:
+      - .venv/
+
 variables:
   SDL_VIDEODRIVER: "dummy"
   XDG_RUNTIME_DIR: "/var/tmp"
@@ -8,6 +15,7 @@ variables:
 
 core fw regular build:
   stage: build
+  <<: *gitlab_caching
   needs: []
   script:
     - nix-shell --run "poetry run make -C core build_boardloader"
@@ -25,6 +33,7 @@ core fw regular build:
 
 core fw regular debug build:
   stage: build
+  <<: *gitlab_caching
   needs: []
   script:
     - nix-shell --run "PYOPT=0 poetry run make -C core build_firmware"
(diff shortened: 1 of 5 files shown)


matejcik:
any idea what this is about?
```
WARNING: .venv/bin/python: chmod .venv/bin/python: no such file or directory (suppressing repeats) 
71
```

but it seems to work anyway so LGTM

matejcik (approved):

vdovhanych:
> any idea what this is about?
> 
> ```
> WARNING: .venv/bin/python: chmod .venv/bin/python: no such file or directory (suppressing repeats) 
> 71
> ```
> 
> but it seems to work anyway so LGTM

@matejcik where is that? I think it could be that I added the caching for .venv even though we do not install poetry there. If that is not the case i don't know why :/

matejcik:
it happens e.g. here: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1815367953
at start when the cache is being unpacked, before `poetry install` is called

vdovhanych:
Strange, honestly I have no idea why that might happen 🤷‍♂️

mmilata (approved):
Looking good, let's merge it & see how much it speeds pipelines.

mmilata:
Btw do the caches expire or do they need to be deleted manually? Haven't found this in the manual.

mmilata:
Warning probably means that gitlab-runner tried to chmod a symlink: https://gitlab.com/gitlab-org/gitlab/-/issues/14613

vdovhanych:
> Btw do the caches expire or do they need to be deleted manually? Haven't found this in the manual.

They should get deleted after 30 days on the branches that won't run in that time. Should behave like artifacts.

vdovhanych:
I kind of thought of a better caching solution in the meantime and thus I'm not sure of the improvements this brings. Gitlab cache has an option to specify if that job should pull or push or push-pull policy for the cache. With the option to push and pull on the first stage in the ci we would create the cache and the rest of the stages would only pull that cache without creating it at the end. This could also speed it up a bit as the jobs won't need to cache it at the end. @matejcik @mmilata This is just a thought but let me know if I should try and implement it (shouldn't be that hard)

matejcik:
If it's trivial to set up, then let's try and see how much it helps.
If not, I believe the curent setup is good enough, so maybe keep this in mind for future improvements.

mmilata:
Agreed, let's try it and see.

PR #1948, opened 2021-11-26
ci: auto add new issues to backlog project

vdovhanych:
This GitHub workflow automatically adds every newly created issue to the Inbox column in the Backlog project. This is the same setup as we have for the trezor-suite.


matejcik (approved):
utACK but @tsusanka take a look too

tsusanka (approved):
LGTM as well

PR #1951, opened 2021-11-26
perf(legacy/bootloader): use slow but small implementation of bn_inverse

onvej-sl:
|            | bootloader size | `ecdsa_verify_digest` time |
|------------|-----------------|----------------------------|
| before     | 32760 B         | 230 ms                     |
| after      | 32060 B         | 390 ms                     |
| difference | -700 B          | +160 ms                    |

The execution times were measured on TT in python and multiplied by `168/120`. The bootloader verifies 3 signatures on startup, which means the total execution time increases from 700 ms to 1200 ms.

Do you think it's worth it?

### DIFF
--- a/legacy/bootloader/Makefile
+++ b/legacy/bootloader/Makefile
@@ -11,6 +11,7 @@ OBJS += ../gen/fonts.small.o
 # overrides from trezor-crypto
 CFLAGS += -DUSE_PRECOMPUTED_IV=0
 CFLAGS += -DUSE_PRECOMPUTED_CP=0
+CFLAGS += -DUSE_INVERSE_FAST=0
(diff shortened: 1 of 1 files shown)


hiviah (changes requested):
I'd really want not to use this shortcut, even though I am thankful you've found it. Unless really necessary.

hiviah:
OTOH -700 B saved on bootloader size and only + 160ms added time could be palatable.

prusnak:
Let's keep this open as a reminder and not merge this right now.

prusnak:
Let's close this in favor of https://github.com/trezor/trezor-firmware/pull/1957 which has been merged already.

Afsoning (approved):

PR #317, opened 2019-07-08
core/secp256k1_zkp: add blinding and rangeproof API

romanz:
Allows rangeproof generation and rewinding, surjection proof generaion and Pedersen commitments' balancing on the device.

Notes to reviewers:
- The current code requires significant stack size increase (to ~40kB), so it depends on https://github.com/trezor/trezor-firmware/issues/174  (or increasing the stack size via `memory_T.ld`) to run on the TREZOR T.
- The tests can run on the emulator via:
```
$ make build_unix
$ ./tests/run_tests.sh test_trezor.crypto.curve.secp256k1.py
```

PR #1952, opened 2021-11-27
chore(core/embed): refactor trezor_obj_get_uint64 into trezorobj.h

prusnak:
Inspired by the change in https://github.com/trezor/trezor-firmware/pull/317

Also simplified `trezor_obj_get_uint64` to look more like `trezor_obj_get_uint`

### DIFF
--- a/core/embed/extmod/trezorobj.h
+++ b/core/embed/extmod/trezorobj.h
@@ -75,6 +75,24 @@ static inline uint8_t trezor_obj_get_uint8(mp_obj_t obj) {
   return u;
 }
 
+static inline uint64_t trezor_obj_get_uint64(mp_const_obj_t obj) {
+  if (MP_OBJ_IS_SMALL_INT(obj)) {
+    mp_int_t i = MP_OBJ_SMALL_INT_VALUE(obj);
+    mp_uint_t u = i;
+    return u;
+  } else if (MP_OBJ_IS_TYPE(obj, &mp_type_int)) {
+    uint64_t u = 0;
+    mp_obj_int_t *self = MP_OBJ_TO_PTR(obj);
+    if (self->mpz.neg != 0) {
+      mp_raise_TypeError("value is negative");
+    }
+    mpz_as_bytes(&self->mpz, MP_ENDIANNESS_BIG, sizeof(uint64_t), (byte *)&u);
+    return u;
+  } else {
+    mp_raise_TypeError("value is not int");
+  }
+}
+
 bool trezor_obj_get_ll_checked(mp_obj_t obj, long long *value);
 
 mp_obj_t trezor_obj_call_protected(void (*func)(void *), void *arg);
--- a/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-monero.h
+++ b/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-monero.h
@@ -21,6 +21,8 @@
 #include "py/objint.h"
 #include "py/objstr.h"
 
+#include "embed/extmod/trezorobj.h"
+
 #include "bignum.h"
 #include "memzero.h"
 #include "monero/monero.h"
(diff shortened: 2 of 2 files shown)


prusnak:
I am tagging @jpochyla because (IRRC) he wrote the original trezorobj code and @onvej-sl because the refactor is relevant to Monero only.

prusnak:
One question stands:

- `trezor_obj_get_uint` does not accept False/True objects
- `trezor_obj_get_uint64` does accept False/True objects

Does Monero code actually use this conversion from boolean?

core/embed/extmod/trezorobj.h
```diff
@@ -75,6 +75,25 @@ static inline uint8_t trezor_obj_get_uint8(mp_obj_t obj) {
...
+  } else if (obj == mp_const_true) {
+    return 1;
+  } else if (MP_OBJ_IS_SMALL_INT(obj)) {
+    mp_int_t i = MP_OBJ_SMALL_INT_VALUE(obj);
+    mp_uint_t u = i;
+    return u;
+  } else if (MP_OBJ_IS_TYPE(obj, &mp_type_int)) {
+    uint64_t u = 0;
+    mp_obj_int_t *self = MP_OBJ_TO_PTR(obj);
+    mpz_as_bytes(&self->mpz, false, 8, (byte *)&u);
```
onvej-sl, line 90: I'm not sure if it's important:

  * Since `mpz_as_bytes` [accepts](https://github.com/micropython/micropython/blob/97a7cc243b028833bdcb8ce0bc19b2bce7545851/py/mpz.c#L1603) negative numbers, `trezor_obj_get_uint64` accepts them too (whereas `trezor_obj_get_uint` does not).
  * I guess `trezor_obj_get_uint64` is not endianness agnostic.

prusnak: The `false` argument to `mpz_as_bytes` says this is all little-endian.

onvej-sl:
> Does Monero code actually use this conversion from boolean?

I think it doesn't, since [this commit](https://github.com/trezor/trezor-firmware/commit/d68251d23a5425ab20ca38ce2059a1f5fe60b141) is [passing](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/418388228).

core/embed/extmod/trezorobj.h
onvej-sl: > The false argument to mpz_as_bytes says this is all little-endian.

Which means it won't work on big-endian architecture. (Honestly, I'm not sure whether we want to support such architectures).

prusnak:
> I think it doesn't, since [this commit](https://github.com/trezor/trezor-firmware/commit/d68251d23a5425ab20ca38ce2059a1f5fe60b141) is [passing](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/418388228).

Thanks! Removed the bool conversion in 676925e72ddf7857cfe89467b04889b6d32d5cb0

core/embed/extmod/trezorobj.h
prusnak: Fixed endianness in 4f896bd256e2289af073df37835e8da80a9c5eb4

matejcik:
i'm still worried about the negative number support.
but i guess it's easy enough to peek into `self->mpz->neg` and raise if set though (or is there maybe `mpz_is_negative` or something?)

prusnak:
> i'm still worried about the negative number support.

`mpz_as_uint_checked` checks the neg field of mpz_t structure, so let's do the same:

``` diff
   } else if (MP_OBJ_IS_TYPE(obj, &mp_type_int)) {
     uint64_t u = 0;
     mp_obj_int_t *self = MP_OBJ_TO_PTR(obj);
+    if (self->mpz->neg != 0) {
+        mp_raise_TypeError("value is negative");
+    }
     mpz_as_bytes(&self->mpz, MP_ENDIANNESS_BIG, sizeof(uint64_t), (byte *)&u);
     return u;
   } else {
```

matejcik (approved):

prusnak:
Rebased on top of current master

core/embed/extmod/trezorobj.h
```diff
@@ -75,6 +75,24 @@ static inline uint8_t trezor_obj_get_uint8(mp_obj_t obj) {
...
 
+static inline uint64_t trezor_obj_get_uint64(mp_const_obj_t obj) {
+  if (MP_OBJ_IS_SMALL_INT(obj)) {
+    mp_int_t i = MP_OBJ_SMALL_INT_VALUE(obj);
+    mp_uint_t u = i;
+    return u;
+  } else if (MP_OBJ_IS_TYPE(obj, &mp_type_int)) {
+    uint64_t u = 0;
+    mp_obj_int_t *self = MP_OBJ_TO_PTR(obj);
+    if (self->mpz->neg != 0) {
```
matejcik, line 86: CI says this should be `self->mpz.neg`

matejcik:
also please run clang-format

PR #1955, opened 2021-11-29
docs(core/cardano): add derivation scheme docs

### LINKED ISSUE #1920: cardano derivation_type documentation
In cardano documentation https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/cardano/README.md, I was unable to find information reasons on why cardano `derivation_type` protobuf field exists and what are the differences between these types of derivations and their implications.

I would therefore like to humbly ask @matejcik for his contribution here. This would also help us with our efforts on other layers (trezor-connect, suite).

matejcik:
fixes #1920 
@mroz22 please take a look


mroz22 (approved):
:+1:

PR #1957, opened 2021-11-29
Build legacy bootloader with size optimization

hiviah:
Some objects from libtrezor were compiled with `-O3` even when used in bootloader. This changes their compilation to size optimization `-Os`, saving currently 1116 bytes in legacy bootloader.


prusnak (approved):
untested ACK

Feel free to merge if everything works on the device as expected.

hiviah:
Tested on device, everything seems to work fine - installed FW few times, finished setup, works with Suite.

Issue #1737, opened 2021-07-27
PIN/Passphrase input for trezorctl doesn't work in some environments


vietnamese-chowder:
**Describe the bug**
I'm trying to run `trezorctl` to sign a Qtum transaction built in Java alongside a Trezor device, by creating a process through ProcessBuilder that runs the command `trezorctl sign-tx <name_of_file>.json`. After running it, I'm feeding it the PIN (corresponding to the layout) that I acquire from the user on the frontend.

I am inputting the PIN to the process using a `BufferedWriter(OutputStreamWriter(proc.getOutputStream()))` by calling `write(pin)`, then `newLine()`, then `flush()`. Everything works on Linux and on Mac when running the program through Intellij, but when I build the jar file and run that, it doesn't. `trezorctl` doesn't receive the input and my program hangs at the next `stdInput.next()`. On Windows this behaviour also occurs on Intellij, and on Git Bash.

When I run `trezorctl` on the command line on Windows, it works on cmd, but not on Git Bash, exhibiting the same behaviour. I've also tried running my program with sudo, so I don't think it's a permission issue. I'm pretty sure this is an issue with trezorctl's IO and not my application, because I tested the exact same method on a dummy Python script I wrote and it worked, even on Windows.

**Firmware version and revision**
1.10.2

**Desktop/smartphone setup (please complete the following information):**
 - Wallet software: trezorctl
 - OS: Windows
 - Browser: None, Git Bash
 - Version: 0.12.2

**To Reproduce**
1. Make sure your Trezor has either PIN or Passphrase enabled
2. Open Git Bash on Windows
2. Type any trezorctl command i.e. `trezorctl get-address -n "m/49'/0'/0'/0/0"`
3. Try typing your PIN/Passphrase (according to the layout) and press enter

**Expected behavior**
The mask doesn't work, so you'll see what you're typing, and pressing enter does nothing.

**Screenshots**
![image](https://user-images.githubusercontent.com/88055026/127214267-abca0291-277f-42e1-8412-f32159620a83.png)

matejcik:
This seems to be an issue either with the click library, or with Python itself.

possibly related: #1682 

please try manually upgrading click: `pip install --upgrade click` and see if the problem goes away afterwards

matejcik:
The issue is actually a [bug](https://bugs.python.org/issue44762) in Python's `getpass` function, which does not work correctly on Windows unless you're using a real terminal.

[...]

PR #1959, opened 2021-11-30
fix(python): allowing PIN/passphrase input for Git Bash

grdddj:
Trying to solve https://github.com/trezor/trezor-firmware/issues/1737

Allowing the input for PIN/passphrase on Windows in Git Bash, which has issues with hidden input.

Recognizing the situation (through `sys.stdin.isatty()`) and in negative case disallowing the hidden feature and reporting this to the user.

Screenshot from Git Bash on Windows:

image

### DIFF
--- a/python/src/trezorlib/ui.py
+++ b/python/src/trezorlib/ui.py
@@ -15,6 +15,7 @@
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
 import os
+import sys
 from typing import Any, Callable, Optional, Union
 
 import click
@@ -54,6 +55,12 @@ PIN_CONFIRM = PinMatrixRequestType.NewSecond
 WIPE_CODE_NEW = PinMatrixRequestType.WipeCodeFirst
 WIPE_CODE_CONFIRM = PinMatrixRequestType.WipeCodeSecond
 
+# Workaround for limitation of Git Bash
+# getpass function does not work correctly on Windows when not using a real terminal
+# (the hidden input is not allowed and it also freezes the script completely)
+# Details: https://bugs.python.org/issue44762
+CAN_HANDLE_HIDDEN_INPUT = sys.stdin and sys.stdin.isatty()
+
 
 class TrezorClientUI(Protocol):
     def button_request(self, br: messages.ButtonRequest) -> None:
@@ -70,8 +77,12 @@ def echo(*args: Any, **kwargs: Any) -> None:
     return click.echo(*args, err=True, **kwargs)
 
 
-def prompt(*args: Any, **kwargs: Any) -> Any:
-    return click.prompt(*args, err=True, **kwargs)
+def prompt(text: str, *, hide_input: bool = False, **kwargs: Any) -> Any:
+    # Disallowing hidden input and warning user when it would cause issues
+    if not CAN_HANDLE_HIDDEN_INPUT and hide_input:
+        hide_input = False
+        text += " (WARNING: will be displayed!)"
+    return click.prompt(text, hide_input=hide_input, err=True, **kwargs)
 
 
 class ClickUI:
@@ -144,6 +155,9 @@ class ClickUI:
                     default="",
                     show_default=False,
                 )
+                # In case user sees the input on the screen, we do not need confirmation
+                if not CAN_HANDLE_HIDDEN_INPUT:
+                    return passphrase
                 second = prompt(
                     "Confirm your passphrase",
                     hide_input=True,
--- a/python/.changelog.d/1959.fixed
+++ b/python/.changelog.d/1959.fixed
@@ -1,0 +1,1 @@
+Fix PIN and passphrase entry in certain terminals on Windows


prusnak:
It seems to me that `click.prompt` uses `getpass.getpass` to obtain the input when `hide_input` is set.

According to [this post](https://stackoverflow.com/questions/38878741/getpasswarning-can-not-control-echo-on-the-terminal-when-running-from-idle) `getpass` should be printing a warning if that is not possible. Is that not the case for us?

grdddj:
> `getpass.getpass` should be printing a warning if echo free input is not possible. Is that not the case for us?

It was not the case for me, neither for the reporter of the issue - I have replicated the same behavior. There was no warning, and the script literally froze and had to be killed.

prusnak:
Sorry, I now see the upstream Python bug mentioned earlier: https://bugs.python.org/issue44762

python/src/trezorlib/ui.py
```diff
@@ -71,6 +78,11 @@ def echo(*args: Any, **kwargs: Any) -> None:
 
 
 def prompt(*args: Any, **kwargs: Any) -> Any:
+    # Disallowing hidden input when it would cause issues
+    if not CAN_HANDLE_HIDDEN_INPUT:
+        if "hide_input" in kwargs and kwargs["hide_input"]:
+            kwargs["hide_input"] = False
+            print("(Unfortunately this terminal cannot handle hidden input.)")
```
prusnak, line 85: 1. `click.echo` is probably what we want instead of `print`
2. let's try a slightly different wording

```suggestion
            click.echo("(Warning: This terminal can't handle hidden input. The input may be shown!)")
```

prusnak (changes requested):

python/src/trezorlib/ui.py
grdddj: After the change, the output looks like this. Is it alright?

image

prusnak: I am fine with the outcome, but let you and @matejcik have the final word.

python/src/trezorlib/ui.py
```diff
@@ -71,6 +78,11 @@ def echo(*args: Any, **kwargs: Any) -> None:
 
 
 def prompt(*args: Any, **kwargs: Any) -> Any:
+    # Disallowing hidden input when it would cause issues
```
matejcik, line 81: instead of the `kwargs` shenanigans below:
```
def prompt(text: str, prompt_hidden: bool = False, **kwargs: Any)
```
we never use `*args` so we can instead fetch the positional `text` parameter directly, which is useful (see below)

python/src/trezorlib/ui.py
matejcik: not "input _may_ be shown", we know it _will_ be shown.

As an alternative to the `click.echo` call, maybe modify the `text`, e.g.: `text += " (WARNING: will be displayed!)`
But I'm not super happy about the wording just yet.

python/src/trezorlib/ui.py
```diff
@@ -54,6 +55,12 @@
 WIPE_CODE_NEW = PinMatrixRequestType.WipeCodeFirst
 WIPE_CODE_CONFIRM = PinMatrixRequestType.WipeCodeSecond
 
+# Workaround for limitation of Git Bash
+# getpass function does not work correctly on Windows when not using a real terminal
+# (the hidden input is not allowed and it also freezes the script completely)
+# Details: https://bugs.python.org/issue44762
+CAN_HANDLE_HIDDEN_INPUT = sys.stdin.isatty()
```
matejcik, line 62: `sys.stdin and sys.stdin.isatty()` (from the linked issue - `sys.stdin` can also be none, in which case we probably don't handle the hidden input either

python/src/trezorlib/ui.py
matejcik: also even if we decide to go with echo, it must be just `echo`, not `click.echo` -- because `echo` goes to stderr while `click.echo` without `err=True` goes to stdin.

matejcik (commented):
additional comments:

Consider looking at `CAN_HANDLE_HIDDEN_INPUT` in `get_passphrase` and in that case, only ask once. No point in retyping to confirm if the user can see typos on screen.

This solves the issue for Git bash, but I'm not sure that it solves it for usage in scripts. I was considering a separate `ScriptUI` implementation:
* uses stdin instead of stderr (so `click.echo` and `click.prompt` without the `err` override)
* never uses hidden inputs
* prints prompts on single line, so instead of:
   ```
   prompt("Enter PIN")
   ```
   it would show
   ```
   click.echo("Enter PIN")
   pin = click.prompt()
   ```
* you can select `ScriptUI` by something like `trezorctl --script`
* it does not show PIN help text
* it prints out button request code (and in the future also name/index), PIN request type
* instead of auto-returning PASSPHRASE_ON_DEVICE, it asks you
  * (and of course only asks for passphrase once)
* does not use the envvar - the script is responsible for providing it.
* shows you something like "enter pin: code=None, max_length=50"
  * and then returns whatever you entered directly, the script is responsible for validating

In essence, this mode would turn `trezorctl` into an interactive tool that allows you to integrate Trezor into, say, Java, without writing a Java library.

grdddj:
Thanks for the suggestions. After c2e5955 it looks like:

image

The creation of `ScriptUI` sounds interesting - should we create a separate issue for it?

matejcik:
> The creation of `ScriptUI` sounds interesting - should we create a separate issue for it?

I consider #1737 to be it -- if you read the issue description, it concerns script use, the git bash thing is a sub-issue.
We can accept this PR as a partial fix.

matejcik (approved):
lgtm, please squash&rebase

matejcik:
also please add a changelog entry: "Fix PIN and passphrase entry in certain terminals on Windows"

grdddj:
Sorry for a lot of force-pushes, I had issues with Windows line-endings and formatting

PR #1961, opened 2021-11-30
Rust UI: loaders, hold to confirm, pagination for model T

mmilata:
Followup of #1960.

### DIFF
--- a/core/embed/rust/src/ui/model_tt/component/loader.rs
+++ b/core/embed/rust/src/ui/model_tt/component/loader.rs
@@ -67,8 +67,12 @@ impl Loader {
     pub fn progress(&self, now: Instant) -> Option<u16> {
         match &self.state {
             State::Initial => None,
-            State::Growing(animation) | State::Shrinking(animation) => Some(animation.value(now)),
+            State::Growing(anim) | State::Shrinking(anim) => Some(anim.value(now)),
         }
+    }
+
+    pub fn is_started(&self) -> bool {
+        matches!(self.state, State::Growing(_) | State::Shrinking(_))
     }
 
     pub fn is_finished(&self, now: Instant) -> bool {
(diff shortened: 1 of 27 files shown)


mmilata (commented):
Looks good, lots of stuff we're gonna need. Loader works smoothly. I propose to merge this after #1960, bugs will probably be found & fixed when we write some layouts using these components. Also let's change `chore` in commit messages to `fix`/`feat`, iirc chores are meant to have no source code changes.

Things I'd like to work on based on this PR:
- decouple `Paginated` from TT - factor out theme, write `Page` component for T1
- `impl Paginate for Paragraphs`

core/embed/rust/src/ui/model_tt/component/loader.rs
```diff
@@ -0,0 +1,201 @@
...
+                    ctx.request_anim_frame();
+                }
+            }
+        }
+        None
+    }
+
+    fn paint(&mut self) {
+        // TODO: Should we paint based on the time in `Self::event`, or based on the
+        // current instant?
```
matejcik, line 137: current instant seems significantly more appropriate (time between generating the event and receiving it goes away)
also currently events don't propagate time, no?

matejcik:
LGTM except the two comments

matejcik:
squashed & rebased onto master (with #1960 merged)

I had to re-add `Mul<f32> for Duration`, at least now I know why it's safe, and added explanation in comments: 9c2387a

core/src/trezor/loop.py
```diff
@@ -558,3 +558,16 @@ def is_running(self) -> bool:
...
+
+    def __init__(self) -> None:
+        self.task: Task | None = None
+
+    def handle(self, task: Task) -> None:
+        self.task = task
+
+    def schedule(self, deadline: int, value: Any) -> None:
+        if self.task is not None:
+            schedule(self.task, value)
```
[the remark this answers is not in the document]
matejcik: 77804024

core/embed/rust/src/ui/model_tt/component/loader.rs
matejcik: 67a4dc37

Issue #1944, opened 2021-11-25
Implement ownership proofs for Taproot


andrewkozlik:
Implement support for Taproot scripts and Schnorr signatures in `GetOwnershipProof` and in `verify_nonownership()` in trezor-core only. Needed for CoinJoin.

andrewkozlik:
Implemented in https://github.com/trezor/trezor-firmware/pull/1965.

PR #1965, opened 2021-12-01
Ownership proofs for Taproot

andrewkozlik:
Implements https://github.com/trezor/trezor-firmware/issues/1944.

For key path spending the implementation is straightforward and I think the SLIP-0019 specification does not need any modification, except for adding a test vector (https://github.com/satoshilabs/slips/pull/1206). I considered changing SLIP-0019 to use the tagged hash for the sighash computation in case of Taproot, but I see no reason to do so. For script path spending, which we do not support yet in Trezor, we might need to consider adding `ext_flag` to the `flags` byte in the proof body and append the tapscript message extension to the sighash computation in SLIP-0019. However, I am not sure yet whether this change will be necessary. Anyway, this PR adds ownership proof support for Taproot key path spending and some tests, most importantly a transaction with an external Taproot input that has an ownership proof.


prusnak (approved):

tests/tx_cache.py
```diff
@@ -107,7 +107,9 @@ def cli(tx, coin_name):
 
     click.echo(f"Fetching from {tx_url}...")
     try:
-        tx_src = requests.get(tx_url).json(parse_float=Decimal)
+        tx_src = requests.get(tx_url, headers={"user-agent": "x"}).json(
```
onvej-sl, line 110: Would you add a short comment on why this is necessary? Something like blockbook servers refuses requests with empty user agent.

prusnak (approved):

tests/tx_cache.py
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/1965/commits/ca7ff4cdf4518ab201b943cddee527ee74779cb1

onvej-sl (approved):

bosomt:
@andrewkozlik any tests needed by QA ?

andrewkozlik:
> @andrewkozlik any tests needed by QA ?

Thanks, no QA needed for this PR.

PR #1449, opened 2021-02-04
python: Allow overriding some Trezor specific things

achow101:
When working with Trezor clones, it is useful to use trezorlib over their forks of the library. As such, we need to be able to specify alternative vendors, minimum firmware versions, and alternative message classes to use for those devices. This PR adds attributes to `TrezorClient` that allow a caller to override those things.

Specifically, `msg_type_to_class_override` is a map of wire types to message classes. When decoding a message, the wire type is looked up in this message first before falling back to the default Trezor messages. `vendors` is added for the vendors check. This defaults to the `VENDORS` tuple, but can be overridden with another tuple by the caller. Lastly `minimum_versions` replaces `MINIMUM_FIRMWARE_VERSION` in the `is_outdated` check. This defaults to `MINIMUM_FIRMWARE_VERSION` but can be overridden by the caller.

PR #1448, opened 2021-02-04
python: Allow specifying USB ids in HID and WebUsb enumerate

achow101:
When working with Trezor clones, it's useful to be able to use trezorlib instead of the fork of it that those clones made. As such, it is necessary to allow `HidTransport` and `WebUsbTransport` be able to enumerate them. To allow this, I've added an optional `usb_ids` parameter to their `enumerate` method. By default, this use the Trezor USB IDs. However callers can specify alternative IDs to use, or alternatively, `None` can be provided to enumerate any and all usb devices.

PR #1442, opened 2021-02-02
python: Raise ValueError for non-existant tx hashes

achow101:
If the device asks for a tx_hash which is not present in prev_txes,
raise a ValueError with some more detailed messaging about the missing
hash rather than the default dictionary lookup failure of KeyError.

PR #1967, opened 2021-12-02
Assorted trezorlib improvements

matejcik:
Includes #1442 and changes that implement #1448 and #1449. @achow101 please take a look.

In addition, a firmware upload progress bar for trezorctl is also here.

TODO: add changelog entry for 49f2e69

### DIFF
--- a/python/src/trezorlib/transport/hid.py
+++ b/python/src/trezorlib/transport/hid.py
@@ -1,6 +1,6 @@
 # This file is part of the Trezor project.
 #
-# Copyright (C) 2012-2019 SatoshiLabs and contributors
+# Copyright (C) 2012-2022 SatoshiLabs and contributors
 #
 # This library is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Lesser General Public License version 3
--- a/python/src/trezorlib/transport/hid.py
+++ b/python/src/trezorlib/transport/hid.py
@@ -17,10 +17,11 @@
 import logging
 import sys
 import time
-from typing import Any, Dict, Iterable, List
+from typing import Any, Dict, Iterable, List, Optional
 
 from ..log import DUMP_PACKETS
-from . import DEV_TREZOR1, UDEV_RULES_STR, TransportException
+from ..models import TREZOR_ONE, TrezorModel
+from . import UDEV_RULES_STR, TransportException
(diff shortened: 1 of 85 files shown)


python/src/trezorlib/transport/hid.py
```diff
@@ -132,11 +133,17 @@ def get_path(self) -> str:
         return f"{self.PATH_PREFIX}:{self.device['path'].decode()}"
 
     @classmethod
-    def enumerate(cls, debug: bool = False) -> Iterable["HidTransport"]:
+    def enumerate(
+        cls, models: Optional[Iterable["TrezorModel"]] = None, debug: bool = False
```
prusnak, line 137: Isn't it better to use the following?
```suggestion
        cls, models: Iterable["TrezorModel"] = {TREZOR_ONE}, debug: bool = False
```
and drop the `if models is None: models = ...` below?

python/src/trezorlib/transport/webusb.py
```diff
@@ -114,15 +115,21 @@ def get_path(self) -> str:
         return f"{self.PATH_PREFIX}:{dev_to_str(self.device)}"
 
     @classmethod
-    def enumerate(cls, usb_reset: bool = False) -> Iterable["WebUsbTransport"]:
+    def enumerate(
+        cls, models: Optional[Iterable["TrezorModel"]] = None, usb_reset: bool = False
```
prusnak, line 119: Isn't it better to use the following?
```suggestion
        cls, models: Iterable["TrezorModel"] = TREZORS, usb_reset: bool = False
```
and drop the `if models is None: models = ...` below?

python/src/trezorlib/transport/hid.py
matejcik: The problem with this is that passthrough from `trezorlib.transport.enumerate()` will not work. The general `enumerate()` method doesn't know which models are appropriate for the specific transports, so it will pass in `None`.

prusnak: Okay, thanks

python/src/trezorlib/transport/webusb.py
prusnak: Same as above

Issue #1934, opened 2021-11-22
Package ‘libcxx-13.0.0’ in /nix/store/.../llvm/13/libcxx/default.nix:42 is marked as broken, refusing to evaluate


Sjors:
**Describe the bug**
`nix-shell` fails with an error about a broken llvm package.

**Firmware version and revision**
Master branch at d448d7b8416f2ee985ed078f482bf05ac49214c3 with up to date submodules.

**Desktop/smartphone setup (please complete the following information):**
 - OS: macOS Monterey 12.0.1
 - NixOS 2.4

**To Reproduce**

- `git submodule update --init --recursive --force`
- `poetry shell`
- `nix shell`

prusnak:
Yes, I encountered this as well. Clang is broken on macOS darwin unfortunately.

Checkout this commit and it should work: 6eac12e80e622298d9f9f6a1a1aca7df55e69860

hynek-jina:
This issue should be solved by upcoming release.

prusnak:
It's not fixed yet, so please, let's keep this open.

prusnak:
Fix in https://github.com/trezor/trezor-firmware/pull/1968

prusnak:
> Fix in #1968

Edit: we will need to resolve this differently (wait until staging is merged into master in nixpkgs).

prusnak:
Another PR in https://github.com/trezor/trezor-firmware/pull/1969

PR #1968, opened 2021-12-03
ci(nix): remove clang-tools

prusnak:
We have added clang-tools explicitly earlier because we needed
clang-format v9 which was only in that package (while clang shipped v7).

Things are different now and clang package now ships clang-format v11.

Fixes https://github.com/trezor/trezor-firmware/issues/1934

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -55,7 +55,6 @@ stdenv.mkDerivation ({
     autoflake
     bash
     check
-    clang-tools
     clang
     editorconfig-checker
     gcc


prusnak:
@mmilata can you please check whether `make cstyle` works as expected after this change on linux? (it performs as expected - no changes are produced - when I run this on aarch64-darwin).

mmilata (approved):
Works fine on linux.

mmilata:
btw I noticed `clang` and `llvmPackages` in nix are the oldest supported version, in our case LLVM/clang 7. I wonder if we should use `llvmPackages_latest` instead or choose some more recent version.

prusnak:
`llvmPackages_latest` is broken on master for darwin at the moment (see https://github.com/trezor/trezor-firmware/issues/1934). At the same time clang points to LLVM/clang 11 on darwin. But I agree that maybe we should be explicit about the llvm version if llvm resolves to version 7 on linux. Will update this PR.

prusnak:
Argh, this is messier than I thought. 

`clang-tools` always uses `llvmPackages_latest` and using override would probably cause a rebuild on user's machine.

Let's wait until llvmPackages_latest is fixed on master for darwin (after merging staging to master) and I'll send a new PR then.

mmilata:
If we only care about `clang-format` then it seems to be included in the `llvmPackages.clang` package as well.

prusnak:
> be included in the `llvmPackages.clang` package as well.

Is it?

```
stick ~ > nix-shell -p llvmPackages_12.clang       

[nix-shell:~]$ clang-format --version
clang-format version 11.1.0

[nix-shell:~]$ which clang-format
/nix/store/3zy8k5xl5s9cmcghgd90c05dga0j5h6a-clang-11.1.0/bin/clang-format
```

mmilata:
Then it probably only works on linux:(

```
% nix-shell -p llvmPackages_12.clang

[nix-shell:~]$ clang-format --version
clang-format version 12.0.1

[nix-shell:~]$ which clang-format
/nix/store/zl1iynf8afbpzfc8ia9fd2xc774br4f0-clang-12.0.1/bin/clang-format
```

prusnak:
Okay, let's do this then: https://github.com/trezor/trezor-firmware/pull/1969

PR #1969, opened 2021-12-03
ci(nix): use explicit versions for clang and gcc

### LINKED ISSUE #1934: Package ‘libcxx-13.0.0’ in /nix/store/.../llvm/13/libcxx/default.nix:42 is marked as broken, refusing to evaluate

prusnak:
Supersedes https://github.com/trezor/trezor-firmware/pull/1968

Fixes #1934

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -41,7 +41,7 @@ let
   llvmPackages = nixpkgs.llvmPackages_12;
 in
 with nixpkgs;
-stdenv.mkDerivation ({
+stdenvNoCC.mkDerivation ({
   name = "trezor-firmware-env";
   buildInputs = lib.optionals fullDeps [
     # install other python versions for tox testing
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -37,6 +37,8 @@ let
     # to use official binary, remove rustfmt from buildInputs and add it to extensions:
     extensions = [ "clippy" ];
   };
+  gcc = nixpkgs.gcc11;
+  llvmPackages = nixpkgs.llvmPackages_12;
 in
 with nixpkgs;
 stdenv.mkDerivation ({
@@ -55,8 +57,7 @@ stdenv.mkDerivation ({
     autoflake
     bash
     check
-    clang-tools
-    clang
+    llvmPackages.clang
     editorconfig-checker
     gcc
     git


mmilata:
Looks good, pulls clang-format 12 for me. I'd also use the same llvmPackages for libclang (`LIBCLANG_PATH` attribute definition) so that we don't have to download two versions of LLVM.

prusnak:
> Looks good, pulls clang-format 12 for me. I'd also use the same llvmPackages for libclang (`LIBCLANG_PATH` attribute definition) so that we don't have to download two versions of LLVM.

Ah, good catch. PR reworked in cba8bf273326cb929fb6b8cb398ef23f58629d6e

ci/shell.nix
```diff
@@ -37,6 +37,8 @@ let
     # to use official binary, remove rustfmt from buildInputs and add it to extensions:
     extensions = [ "clippy" ];
   };
+  gcc = nixpkgs.gcc11;
+  llvmPackages = nixpkgs.llvmPackages_12;
 in
 with nixpkgs;
```
mmilata, line 43: > bindings introduced by with do not shadow bindings introduced by other means
didn't know this

mmilata (approved):
LGTM

However this is weird:
```
% cd trezor-firmware
% nix-shell
$ gcc --version
gcc (GCC) 10.3.0
```

10.3 is the version of gcc-arm-embedded :thinking: 

It happens also in master so was not introduced in this PR.

prusnak:
```
$ gcc --version
gcc (GCC) 10.3.0
```

What does `which gcc` say?

mmilata:
Ok, this is unrelated to gcc-arm-embedded as it also happens outside our repo:
```
% nix-shell -p gcc11

[nix-shell:~]$ gcc --version
gcc (GCC) 10.3.0

[nix-shell:~]$ which gcc
/nix/store/da9l6ifizmh883j94gzx5n51h8lrrp00-gcc-wrapper-10.3.0/bin/gcc
```
I'm on NixOS unstable channel.

prusnak:
> It happens also in master so was not introduced in this PR.

On master this is legit, because gcc points to gcc10 on Linux and gcc10 is on 10.3.0

prusnak:
@mmilata what you see is exactly the same issue I see here on darwin: https://github.com/trezor/trezor-firmware/pull/1968#issuecomment-985420969

The gcc installed via Linux stdenv trumps gcc installed in nix-shell, the same as llvm installed via Darwin stdenv trumps llvm installed in nix-shell.

mmilata:
Huh, that's weird o_O. There has to be some way to use custom gcc version on linux though, I suppose we're not the first to want it.

prusnak:
> I suppose we're not the first to want it.

The fix is trivial :-) We need to use stdenvNoCC: f727691348aa22e815a1d9fdc7ef77838792f19d

mmilata (approved):
That kinda makes sense, yeah:D

Issue #1932, opened 2021-11-22
Run Connect tests against our branches in CI


tsusanka:
Add jobs to our CI that would run the current firmware against Connect master. Let's also discuss what jobs should be run on every push.

Maybe let's do this along with https://github.com/trezor/trezor-user-env/issues/49.

hynek-jina:
Blocked by https://github.com/trezor/trezor-user-env/issues/49

grdddj:
This could be theoretically closed, as the `Connect` tests were already added into our `CI` in above-mentioned https://github.com/trezor/trezor-firmware/pull/1972 @hynek-jina @matejcik (if we do not want any improvements over current situation)

matejcik:
right, let's close this

PR #1972, opened 2021-12-03
Connect tests against current firmware branch in CI

grdddj:
Starting to experiment with https://github.com/trezor/trezor-firmware/issues/1932

Goal is to get the `connect` repository and run all its tests in our CI using out freshly-build emulator.

For quicker feedback loop, I have disabled all other CI stages apart from the new `connect`.

To be done:
- `tenv` image being able to run a local emulator (as we will have the build emulator as artifact on CI server and we could use it to the tests)
  - other possibility is just sending it the URL with the CI artifact (this is already implemented [in this PR](https://github.com/trezor/trezor-user-env/pull/120)

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -57,6 +57,7 @@ stdenvNoCC.mkDerivation ({
     autoflake
     bash
     check
+    curl  # for connect tests
     editorconfig-checker
     gcc
     gcc-arm-embedded
--- a/ci/test.yml
+++ b/ci/test.yml
@@ -373,3 +373,27 @@ core unix memory profiler:
       - core/prof/memperf-html
     expire_in: 1 week
     when: always
+
+# Connect
+
+connect test core:
+  image: registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env:latest
+  stage: test
+  needs:
+    - core unix frozen debug build
+  variables:
+    SDL_VIDEODRIVER: "dummy"
+  before_script:
+    - cp /builds/satoshilabs/trezor/trezor-firmware/core/build/unix/trezor-emu-core /trezor-user-env/src/binaries/firmware/bin/trezor-emu-core-v2.99.99
+    - chmod u+x /trezor-user-env/src/binaries/firmware/bin/trezor-emu-core-v2.99.99
+    - nix-shell -p autoPatchelfHook SDL2 SDL2_image --run "autoPatchelf /trezor-user-env/src/binaries/firmware/bin/trezor-emu-core-v2.99.99"
+  script:
+    - /trezor-user-env/run.sh &
+    - nix-shell --run "tests/connect_tests/connect_tests.sh 2.99.99"
+  after_script:
+    - cp /trezor-user-env/logs/debugging.log trezor-user-env-debugging.log
+  artifacts:
+    paths:
+      - trezor-user-env-debugging.log
+    expire_in: 1 week
+    when: always
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -1,2 +1,3 @@
 junit.xml
 trezor.log
+connect_tests/connect
--- a/tests/connect_tests/connect_tests.sh
+++ b/tests/connect_tests/connect_tests.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+# Cloning connect repository and running the tests from there
+
+FILE_DIR="$(dirname "${0}")"
+cd ${FILE_DIR}
+
+CONNECT_DIR="connect"
+
+# For quicker local usage, do not cloning connect repo if it already exists
+if [[ ! -d "${CONNECT_DIR}" ]]
+then
+    git clone https://github.com/trezor/connect.git
+    cd ${CONNECT_DIR}
+    git submodule update --init --recursive
+else
+    cd ${CONNECT_DIR}
+fi
+
+# Taking an optional script argument with emulator version
+if [ ! -z "${1}" ]
+then
+    EMU_VERSION="${1}"
+else
+    EMU_VERSION="2-master"
+fi
+echo "Will be running with ${EMU_VERSION} emulator"
+
+# Using -d flag to disable docker, as tenv is already running on the background
+nix-shell --run "yarn && tests/run.sh -d -f ${EMU_VERSION}"


grdddj:
After a big help from @vdovhanych (thanks a lot!), now it seems we have a working pipeline of `Connect` tests in our `CI`.

Successful `CI` is at https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/424136708 - the `build` job is creating a debug emulator and the `connect` job is taking this emulator, putting it into `tenv` and running some basic tests on it.

In the end there is no need to `docker cp` the emulator inside the `tenv` image, as the `CI` job is configured to use the `tenv` image and clone the `firmware` code with artifacts into it - so instead of `docker cp`, we can just `cp` the emulator to appropriate location in `tenv` filesystem and when `tenv` runs then, it will see it.

Current issues/next steps:
- we need to name the new emulator `2-master` (or theoretically some `X.Y.Z`, like `2.99.99`), as `2-custom` or any other version fails the setup in `Connect` tests: https://github.com/trezor/connect/blob/eadacf99637b6dabd54165879184724dc96d032e/tests/common.setup.js#L52-L56
  -  we could make a small change in `Connect` to allow for `2-custom` or some other names 
- we need to find out how to call all the existing tests at once, currently we call just one using `-i applySettings` argument to the testing `tests/run.sh` script
  - we may need to hardcode the list of tests or somehow split the tests into multiple parts as is done in `Connect CI` - https://github.com/trezor/connect/blob/develop/.gitlab-ci.yml
- new `connect` `CI` stage should be merged into `test` stage
- decide whether to test `T1` emulators as well
- deleting the `Dockerfile` changes to our `environment` image, which were in the end not used
- reverting all the `CI` changes

matejcik:
> * https://github.com/trezor/connect/blob/eadacf99637b6dabd54165879184724dc96d032e/tests/common.setup.js#L52-L56

This honestly sounds like something `tenv` should be abstracting away. Connect should just be able to call "allow-unsafe-paths" and tenv would either do it, if the version is high enough, or do nothing on versions that don't have safety checks.

> we need to find out how to call all the existing tests at once, currently we call just one using `-i applySettings` argument to the testing `tests/run.sh` script

i believe you can just run `tests/run.sh` without this and it will run everything

> * decide whether to test `T1` emulators as well

yes, definitely

grdddj:
> This honestly sounds like something `tenv` should be abstracting away. Connect should just be able to call "allow-unsafe-paths" and tenv would either do it, if the version is high enough, or do nothing on versions that don't have safety checks.

`Tenv` is already ignoring this for `1.X.Y` versions. Is `Connect` logic correct, and only versions  `>= 2.3.2` do have safety-checks, or is there some other way how to recognize it (somehow read it from `client.features`)?

https://github.com/trezor/trezor-user-env/blob/master/src/emulator.py#L430-L438

matejcik:
> `Tenv` is already ignoring this for `1.X.Y` versions.

And that's wrong as well! T1 does have safety checks. They're less strict though, maybe Connect is not hitting them?
It would be interesting to see which Connect test fail.

I would believe the Connect numbers, but perhaps a better strategy is to _always_ call the `apply_settings`, but catch and ignore the exception if the setting doesn't exist.

grdddj:
> And that's wrong as well! T1 does have safety checks. They're less strict though, maybe Connect is not hitting them?
> It would be interesting to see which Connect test fail.

It seems `Connect` really does not care about `T1`, as running it with `1-master` fails at the same place as the `2-custom`.

I also tried to run that basic `applySettings` tests on `1.10.2` and it failed the `Change auto_lock_delay back` test on `"code": "Failure_DataError", "error": "fsm_msg_common.h:367:No setting provided",`

grdddj:
We have the first results of running all the `Connect` tests in our `CI`:
- **T1**
  - under https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1858984591
  - **a lot of errors** - `Tests:       115 failed, 1 skipped, 249 passed, 365 total`
  - it seems that `Connect` tests were never supported for `T1`, otherwise I cannot explain so many broken things
  - examples of error messages:
    -  `"error": "ui-device_firmware_unsupported"`
    - `"error": "messages.c:257:Unknown message"`
- **T2**
  - under https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1858984585
  - **everything was successful**, the job is green, its logs were just too long that it is not showing the end result with details

matejcik:
it would be nice to get T1 tests running, but that's probably an issue of Connect.

You could try excluding tests manually -- many coins being tested aren't supported on T1, also Connect probably doesn't have the auto-pagination mechanism?
(we could add it - Connect now passes through the `pages` field, so you could add something like "scroll-pages <n>" to tenv for Connect to call)

In any case, let's keep only the TT pipeline for this PR, and continue on the T1 thing in a separate issue/pr

grdddj:
Rebased on `master` and force-pushed.

The commit history (with 42 commits) does not make sense, as it is a lot of experimenting and reverting the changes - I will definitely squash it into probably one commit after we agree.

The first complete `CI` run with the new `connect` job was run in https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/424511062 - everything fine apart from that `clang` thing, should be fixed with the rebase.

The one disadvantage is that the `connect` job itself - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1858984585 - takes 20 minutes to complete, which is making the `CI` run longer. The whole pipeline took 38 minutes, which seems much more than the average (I saw one pipeline finished in 23 minutes) - but the time is probably variable with `CI` load and we would need to run it more times to see and compare the average time.

matejcik:
please go ahead and squash&rebase so that we get a clean slate here

grdddj:
> afterwards, a thing we need to check is deliberately breaking Connect tests and checking that they fail.

Yes, that would be nice. Should I create a new branch for it? If we can somehow get the `CI` run for any branch - and if not, we would need to create a new `PR` for it - or doing it inside this `PR` and then reverting the changes.

grdddj:
I have created a separate branch - `grdddj/failing_connect_ci_tests`, where the ethereum's  `get_address` was modified and `CI` for it is running under https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/425118821. Some device and maybe even unittests should fail together with the new `connect` tests

grdddj:
In `grdddj/redirecting_connect_output_to_logfile` I am trying to send `connect` logs to a logfile instead of `Gitlab` screen, as the output is so long it does not even show the final result

grdddj:
Pipeline https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/425118680 verified that these `connect` tests would fail when some mistake is introduced in `firmware` (unfortunately we cannot see the real results there, just that it failed - to be solved below)

Pipeline https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/425195602 is introducing redirect of all `connect-specific` logs into a logfile and making it an artifact under https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1862813513/artifacts/browse. It shows the final test output, and surprisingly in this case we have some errors for `Bitcoin` tests, it seems like a `timeout` issue with the emulator

`tenv` PR https://github.com/trezor/trezor-user-env/pull/124 is introducing above-mentioned idea with `safety-checks`

`tenv` PR https://github.com/trezor/trezor-user-env/pull/123 is trying to solve the issue with `emulator` and `bridge` logs spamming the `stdout` in all `Gitlab` jobs using `tenv`

matejcik (approved):

PR #1973, opened 2021-12-03
GCC 11 emulator fixes

mmilata:
In https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1846713702 clang is picked over gcc (and ends with error) because both have `cc` symlink. `hiPrio`/`lowPrio` don't seem to work for nix shells but the order of the packages in the list matters.

Also the new gcc uncovered a bug in crypto benchmarks.

Also wrong parameter to `strnlen` in legacy: 78c6d275f039e73cc3080bc146010f712f3b0025.

Also had to add `-fPIE` to emulator CFLAGS otherwise `cargo test` fails: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1847861609


legacy/firmware/config.c
```diff
@@ -319,7 +319,7 @@ static secbool config_upgrade_v10(void) {
   storage_unlock(PIN_EMPTY, PIN_EMPTY_LEN, NULL);
   if (config.has_pin) {
     storage_change_pin(PIN_EMPTY, PIN_EMPTY_LEN, (const uint8_t *)config.pin,
-                       strnlen(config.pin, MAX_PIN_LEN), NULL, NULL);
+                       strnlen(config.pin, sizeof(config.pin)), NULL, NULL);
```
prusnak, line 322: I think the original code was intentional. Wasn't it @andrewkozlik ?

What is the GCC11 error being thrown here?

mmilata: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1847730745

```
config.c: In function ‘config_upgrade_v10.isra’:
config.c:321:5: error: ‘strnlen’ specified bound 50 exceeds source size 10 [-Werror=stringop-overread]
  321 |     storage_change_pin(PIN_EMPTY, PIN_EMPTY_LEN, (const uint8_t *)config.pin,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  322 |                        strnlen(config.pin, MAX_PIN_LEN), NULL, NULL);
      |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from config.c:31:
config.h:66:18: note: source object declared here
   66 |   STORAGE_STRING(pin, 10)
      |                  ^~~
config.h:33:8: note: in definition of macro ‘STORAGE_STRING’
   33 |   char NAME[SIZE];
      |        ^~~~
cc1: all warnings being treated as errors
```

prusnak: It is correct that the values in `struct _Storage` and macros `MAX_*_LEN` are untangled, because the struct is used only for importing the old versions of the storage. New storage even does not store the PIN at all.

I assume what andrew meant with the code was the following:

```
MIN(strnlen(config.pin, sizeof(config.pin)), MAX_PIN_LEN)
```

and while your suggested change works just fine, it will stop working as intended when `sizeof(config.pin)` for some reason would become larger than `MAX_PIN_LEN` in the future (it most probably won't).

That's why I suggest to use the fix above or something among those lines, but only if @andrewkozlik confirms I am not completely mistaken.

andrewkozlik: This got introduced in [cbb0d82999](https://github.com/trezor/trezor-firmware/commit/cbb0d829994838313483d6934f869371df33b4bb#diff-c6290b9a209a5ca74c273f883bda833ce8e2e6238b0378b3217e554655157ce1R319-R320), where I changed the arguments of `storage_change_pin()`. I think it's just a mistake that I didn't use `sizeof(config.pin)`. Both of the proposed solutions are fine. I cannot imagine that `sizeof(config.pin)` would ever become larger than `MAX_PIN_LEN`, because `sizeof(config.pin)` is constant. The only way that it could happen is if we decided to reduce `MAX_PIN_LEN`, which would create all sorts of problems. So I am happy with @mmilata's original solution.

mmilata: Can't imagine shortening the pin limit either but weirder things happened and correctness here doesn't cost much so I added it in 3a25cf0.

mmilata:
Fixed the last build issue in bd3ef91, please take a look.

andrewkozlik (approved):

Issue #1983, opened 2021-12-06
Protobuf message is interpreted twice


tsusanka:
**Describe the bug**
Firing `RebootToBootloader` followed by `GetFeatures` instantly lead to a peculiar behaviour where the user dialog (Do you want to restart...) is shown again (and `ButtonRequest` is sent again as well).

**Firmware version and revision**
I have encountered this with 1.10.1 but I think this will concern all >1.10.0 devices.

**Desktop/smartphone setup (please complete the following information):**
- trezorlib on current master (4419572437f7f3f4239a03725cc1902f5803fc1e)
- macOS m1

**To Reproduce**
```
from trezorlib.client import get_default_client
from trezorlib.log import enable_debug_output
from trezorlib import messages
from time import sleep

enable_debug_output()
client = get_default_client()
client.call(messages.RebootToBootloader())

client.call(messages.GetFeatures())
```

Output log can be found [here](https://gist.github.com/tsusanka/59fe58a68b9d13bcea980a579de7053f).

**Expected behavior**
Firmware responds with `Features` on `GetFeatures` message.

**Additional context**
- Note that I believe this issue can only be replicated on fast computers (whatever that means) when the second message is fired instantly. 
- I am not sure if this is bound to the `RebootToBootloader` message, maybe this is a generic problem.
- We solve this issue in Connect by some sleep so it is not a time critical issue, however if it can be solved, it would be quite nice as the results are unexpected.

hiviah:
Yes the sleep is there to flush response message out (but there's no "flush and USB is now stopped". On first look the usbSleep should only wait. Will investigate further why it seems to recurse handler.

hiviah:
The issue is caused by combination of two things:

We can't do

  1. receive message
  2. reset
  3. send response

So 2. and 3. are swapped.

Using usbSleep really does process messages, but incorrectly, as we broke the FSM by calling `dwc_poll` while still inside FSM.

Dialog shown first time:

```
[0] from 0x0801a684 in fsm_msgRebootToBootloader+0 at fsm.c:351
[1] from 0x08018e56 in msg_process+66 at messages.c:229
[2] from 0x08018f36 in msg_read_common+182 at messages.c:283
[3] from 0x0804ac6a in dwc_poll+470 at ../../usb/usb_dwc_common.c:401
[4] from 0x08018b50 in usbPoll+12 at usb.c:410
[5] from 0x0801a176 in main+166 at trezor.c:185
```

[...]

PR #1985, opened 2021-12-06
Prevent recursing in handling RebootToBootloader by USB flush

hiviah:
This fixes showing reboot dialog twice from #1983 by explicitly calling USB flush (write and wait) instead of USB sleep which has side effect of handling messages.

The client will still get LIBUSB error when sending messages too quickly.


prusnak (approved):
utACK

matejcik (approved):

PR #1986, opened 2021-12-07
build: update rust to 1.57

mmilata:
https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html
https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html


prusnak (approved):
ACK, firmware builds, did not try on the real device

PR #1992, opened 2021-12-08
fix(crypto): avoid warnings and document bitwise operator use

invd:
Discovered via clang-14 warnings for -Wbitwise-instead-of-logical
Closes https://github.com/satoshilabs/trezor-firmware/issues/129

### DIFF
--- a/crypto/bignum.c
+++ b/crypto/bignum.c
@@ -271,7 +271,8 @@ int bn_is_equal(const bignum256 *x, const bignum256 *y) {
 //   &truecase == &falsecase or &res == &truecase == &falsecase
 void bn_cmov(bignum256 *res, volatile uint32_t cond, const bignum256 *truecase,
              const bignum256 *falsecase) {
-  assert((cond == 1) || (cond == 0));
+  // Intentional use of bitwise OR operator to ensure constant-time
+  assert((int)(cond == 1) | (int)(cond == 0));
 
   uint32_t tmask = -cond;   // tmask = 0xFFFFFFFF if cond else 0x00000000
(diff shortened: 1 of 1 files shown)


andrewkozlik:
Hmm, I just realized that maybe the `|` instead of `||` is intentional to avoid short-circuiting and ensure constant time operation.

invd:
@andrewkozlik : I may be missing something here, but in my understanding the `||` operator only skips the second check as part of lazy evaluation if the first check is true. If that is the case, the assert will trigger. 
An attacker with good time observation can figure out if the assert triggered after one or after two checks, so this still leaks information about the exact `cond` value, but AFAIK this is acceptable in the error case.

We can cast the checks to `(int)` and keep the `|` to suppress the warning and keep the non-short-circuit behavior if we're sure that this is functionally correct.

andrewkozlik:
> We can cast the checks to `(int)` and keep the `|` to suppress the compiler warning and keep the non-short-circuit behavior if we're sure that this is functionally correct.

Agreed, and also add a comment explaining that it is intentional, because otherwise it look really weird. But first, let's figure out whether the `|` is intentional or not, or to be more precise whether constant-time operation is required for `cond`. The assertion was introduced in 24ceb0ab6b, so @onvej-sl should weigh in.

onvej-sl:
The `|` is definitely intentional and constant-time operation for `cond` is required.

prusnak (approved):

andrewkozlik (approved):

onvej-sl (approved):

PR #2003, opened 2021-12-08
chore(legacy): rename MEMORY_PROTECT flag to PRODUCTION to match core naming

prusnak:
Please review carefully 🙏


docs/common/reproducible-build.md
```diff
@@ -78,4 +78,4 @@ _Note: Fingerprints displayed for T1 at the end of `build-docker.sh` do not matc
 official firmware due to the legacy header._
 
 _Note: T1 firmware built this way won't boot because unsigned firmware needs to be built with
-[`MEMORY_PROTECT=0`](../legacy/index.md#combining-bootloader-and-firmware-with-various-memory_protect-settings-signedunsigned)._
+[`PRODUCTION=0`](../legacy/index.md#combining-bootloader-and-firmware-with-various-memory_protect-settings-signedunsigned)._
```
mmilata, line 81:
```suggestion
[`PRODUCTION=0`](../legacy/index.md#combining-bootloader-and-firmware-with-various-production-settings-signedunsigned)._
```

hiviah (approved):
Checked by disasembly+bindiff also the non-production binaries generated since I use them for debug and they LGTM.

legacy/startup.S
```diff
@@ -20,7 +20,7 @@ memset_reg:
   .type reset_handler, STT_FUNC
 reset_handler:
 
-#if MEMORY_PROTECT
+#if PRODUCTION
```
hiviah, line 23: In future I'd really like to replace this instance of `PRODUCTION` for some other variable. Since it's the most confusing instance. You always want that `#if` block included unless using `PRODUCTION=1` bootloader and unsigned `PRODUCTION=1` firmware.

But I'll put that into another PR. It needs some though about how to do it correctly.

Issue #2011, opened 2021-12-11
Broken links in changelogs


prusnak:
Github stopped redirecting from
e.g. [`https://github.com/trezor/trezor-firmware/issues/1227`](https://github.com/trezor/trezor-firmware/issues/1227)
to
[`https://github.com/trezor/trezor-firmware/pull/1227`](https://github.com/trezor/trezor-firmware/pull/1227)
😢 
(the other direction still works) 

We kind of relied on it in our changelogs as we always used the `/issues/` in our changelog URLs.

What needs to be fixed now:
 
- [ ] fix links in `core/CHANGELOG.md`
- [ ] fix links in `legacy/firmware/CHANGELOG.md`
- [ ] fix links in `python/CHANGELOG.md`
- [ ] update `tools/generate-changelog.py` to generate `/pull/` in the URL instead of `/issues/` when the referenced number is PR and not an issue

The rest of the changelogs do not contain broken links.

Carlosbogo:
Hi, I would like to work on this. Is there anyone doing it already?

prusnak:
@Carlosbogo Not at this moment, please, go ahead!

PR #2013, opened 2021-12-12
Docs: Update links in changelog

Carlosbogo:
Updates the links in the changelogs so that they are not broken anymore due to changes from Github, as mentioned in #2011.


prusnak:
What do you think @matejcik?

It (mis)uses the fact that `/pull/` -> `/issues/` redirect still works on GitHub and IMO that's good enough at least for now.

Carlosbogo:
I actually checked all the files mentioned in #2011 and the only ones that had broken links are the ones I changed. If I missed something please tell me and I'll be glad to help.

prusnak:
> and the only ones that had broken links are the ones I changed.

You missed `python/CHANGELOG.md`, but you are right about the rest that no changes were necessary.

Carlosbogo:
Sorry about that, I just added those changes too. Thanks for pointing it out.

matejcik:
Good enough for now, I suppose. In the future we'll want to do something like, ping `/issues/` and use `/pull/` if it fails

prusnak:
@Carlosbogo thank you for your contribution!

PR #2016, opened 2021-12-14
feat: add Cardano testnet definitions

vladimirvolek:


common/defs/misc/misc.json
```diff
@@ -87,6 +87,22 @@
             "Daedalus": "https://daedaluswallet.io"
         }
     },
+    {
+        "name": "Cardano Testnet",
+        "shortcut": "tADA",
+        "slip44": 1815,
```
prusnak, line 93: Every testnet should use 1 for SLIP44.

```suggestion
        "slip44": 1,
```

prusnak (changes requested):

matejcik:
Why do you need this? Just to display `tADA` on trezor.io/coins ?

mroz22:
> Why do you need this? Just to display `tADA` on trezor.io/coins ?

Connect runs this script https://github.com/trezor/connect/blob/develop/scripts/coins-build.sh in order to generate coins.json file which is later used in connect runtime. My understanding is that without the change proposed here, we won't have cardano testnet in the resulting json file.

matejcik:
there are missing support infos
please add them, see https://github.com/trezor/trezor-common/tree/filter-repo/tools#maintaining-support-status

matejcik (approved):

PR #2017, opened 2021-12-14
test for: path warning displayed instead of failing immediately

### LINKED ISSUE #1895: Path warning displayed when the same transaction fails later due to strict checks
**To Reproduce**

make sure you have strict security: `trezorctl set safety-checks strict`

save the following json:
```
{
  "coin_name": "Testnet",
  "details": {
    "lock_time": 0,
    "version": 2
  },
  "inputs": [
    {
      "address_n": [
        0
      ],
      "amount": 4977040,
      "prev_hash": "a5cd2a706d680587e572df16a8ce5233139a094ebbd148cc66a8004dcc88819c",
      "prev_index": 0,
      "script_type": "SPENDWITNESS",
      "sequence": 4294967293
    }
  ],
  "outputs": [
    {
      "address_n": [
        2147483732,
        2147483648,
        2147483648,
        1,
        0
      ],
      "amount": 4977040,
      "script_type": "PAYTOWITNESS"
    }
  ],
  "prev_txes": {
    "a5cd2a706d680587e572df16a8ce5233139a094ebbd148cc66a8004dcc88819c": {
      "bin_outputs": [
        {
          "amount": 4977040,
          "script_pubkey": "0014820d4a343a44e915c36494995c2899abe3741893"
        },
        {
          "amount": 0,
          "script_pubkey": "6a24aa21a9ed6edfd0a9431aad16549cf318c0d1d237e6a0a87a6403f177caa74fe05546a98f"
        }
      ],
      "inputs": [
        {
          "prev_hash": "0000000000000000000000000000000000000000000000000000000000000000",
          "prev_index": 4294967295,
          "script_sig": "030c15203a205468697320626c6f636b20776173206d696e65642077697468206120636172626f6e206e6567617469766520706f77657220736f75726365201209687a200909200901b34d00641f8d000000",
          "sequence": 4294967295
        }
      ],
      "lock_time": 0,
      "version": 1
    }
  }
}
```

try to sign it: `trezorctl btc sign-tx repro.json`

**Expected behavior:**

When the address `m/0` is encountered, signing should fail immediately with "Forbidden key path"

**Actual behavior:**

First a path confirmation [is displayed](https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/bitcoin/sign_tx/approvers.py#L104-L106). Only afterwards when trying to actually use the keychain, signing fails.

**Ideas to fix:**

Approver (or the `validate_path_against_script_type` helper) should check the value of safety_checks and fail early.

This would disable spending from paths that don't match the script type but are otherwise valid. I'm not sure how big a problem that is.

mmilata:
Fixes #1895.

I've chose to `skip_t1` instead of using Litecoin path for the input because T1 does not return error on first `TxAckInput` like TT does.


tests/device_tests/bitcoin/test_signtx_invalid_path.py
```diff
@@ -215,3 +219,30 @@ def attack_processor(msg):
...
+    )
+
+    out1 = messages.TxOutputType(
+        address_n=parse_path("84h/0h/0h/1/0"),
+        amount=4977040,
+        script_type=messages.OutputScriptType.PAYTOWITNESS,
+    )
+
+    with client, pytest.raises(TrezorFailure) as exc:
+        client.set_expected_responses(
```
matejcik, line 242: does this fail the right way if the expected responses don't match?

maybe a better way would be to just `try: btc.sign_tx; except: pass`, and leave only the `expected_responses` as a thing we're testing

matejcik (approved):

andrewkozlik (approved):

PR #2019, opened 2021-12-15
Blind EIP-712 for core

mmilata:
Needs #1970.


matejcik:
superseded for now by https://github.com/trezor/connect/pull/983

PR #2023, opened 2021-12-16
ScriptUI for trezorctl

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1737 and using suggestions in https://github.com/trezor/trezor-firmware/pull/1959

Creating a possibility of more reliable usage of `trezorctl` by other scripts.

Making a new `ScriptUI` class that has a lot in common with `ClickUI`. 

It is not final by any means, the `ScriptUI` will need to be changed quite a lot and in the end we may somehow connect it with `ClickUI` - or create a `BaseUI` and both will inherit from it.

New (temporary) file `python/src/trezorlib/cli/test_script_client.py` allows for quick testing and developing - currently it only tries `get-address` without the passphrase.

### DIFF
--- a/python/src/trezorlib/cli/__init__.py
+++ b/python/src/trezorlib/cli/__init__.py
@@ -23,10 +23,13 @@ import click
 
 from .. import exceptions
 from ..client import TrezorClient
-from ..transport import Transport, get_transport
-from ..ui import ClickUI
+from ..transport import get_transport
+from ..ui import ClickUI, ScriptUI
 
 if TYPE_CHECKING:
+    from ..transport import Transport
+    from ..ui import TrezorClientUI
+
     # Needed to enforce a return value from decorators
     # More details: https://www.python.org/dev/peps/pep-0612/
     from typing import TypeVar
@@ -50,13 +53,18 @@ class ChoiceType(click.Choice):
 
 class TrezorConnection:
     def __init__(
-        self, path: str, session_id: Optional[bytes], passphrase_on_host: bool
+        self,
+        path: str,
+        session_id: Optional[bytes],
+        passphrase_on_host: bool,
+        script: bool,
     ) -> None:
         self.path = path
         self.session_id = session_id
         self.passphrase_on_host = passphrase_on_host
+        self.script = script
 
-    def get_transport(self) -> Transport:
+    def get_transport(self) -> "Transport":
         try:
             # look for transport without prefix search
             return get_transport(self.path, prefix_search=False)
@@ -68,8 +76,14 @@ class TrezorConnection:
         # if this fails, we want the exception to bubble up to the caller
         return get_transport(self.path, prefix_search=True)
 
-    def get_ui(self) -> ClickUI:
-        return ClickUI(passphrase_on_host=self.passphrase_on_host)
+    def get_ui(self) -> "TrezorClientUI":
+        if self.script:
+            # It is alright to return just the class object instead of instance,
+            # as the ScriptUI class object itself is the implementation of TrezorClientUI
+            # (ScriptUI is just a set of staticmethods)
+            return ScriptUI  # type: ignore [Expression of type "Type[ScriptUI]" cannot be assigned to return type "TrezorClientUI"]
+        else:
+            return ClickUI(passphrase_on_host=self.passphrase_on_host)
 
     def get_client(self) -> TrezorClient:
         transport = self.get_transport()
--- a/python/src/trezorlib/cli/trezorctl.py
+++ b/python/src/trezorlib/cli/trezorctl.py
@@ -156,6 +156,12 @@ def configure_logging(verbose: int) -> None:
     help="Enter passphrase on host.",
 )
 @click.option(
+    "-S",
+    "--script",
+    is_flag=True,
+    help="Use UI for usage in scripts.",
+)
+@click.option(
     "-s",
     "--session-id",
     metavar="HEX",
@@ -170,6 +176,7 @@ def cli_main(
     verbose: int,
     is_json: bool,
     passphrase_on_host: bool,
+    script: bool,
     session_id: Optional[str],
 ) -> None:
     configure_logging(verbose)
@@ -181,7 +188,7 @@ def cli_main(
         except ValueError:
             raise click.ClickException(f"Not a valid session id: {session_id}")
 
-    ctx.obj = TrezorConnection(path, bytes_session_id, passphrase_on_host)
+    ctx.obj = TrezorConnection(path, bytes_session_id, passphrase_on_host, script)
 
 
 # Creating a cli function that has the right types for future usage
@@ -189,10 +196,14 @@ cli = cast(TrezorctlGroup, cli_main)
 
 
 @cli.resultcallback()
-def print_result(res: Any, is_json: bool, **kwargs: Any) -> None:
+def print_result(res: Any, is_json: bool, script: bool, **kwargs: Any) -> None:
     if is_json:
         if isinstance(res, protobuf.MessageType):
-            click.echo(json.dumps({res.__class__.__name__: res.__dict__}))
+            res = protobuf.to_dict(res, hexlify_bytes=True)
+
+        # No newlines for scripts, pretty-print for users
+        if script:
+            click.echo(json.dumps(res))
         else:
             click.echo(json.dumps(res, sort_keys=True, indent=4))
     else:
--- a/python/src/trezorlib/ui.py
+++ b/python/src/trezorlib/ui.py
@@ -172,6 +172,56 @@ class ClickUI:
                 raise Cancelled from None
 
 
+class ScriptUI:
+    """Interface to be used by scripts, not directly by user.
+
+    Communicates with a client application using print() and input().
+
+    Lot of `ClickUI` logic is outsourced to the client application, which
+    is responsible for supplying the PIN and passphrase.
+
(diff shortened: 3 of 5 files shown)


grdddj:
It could be ready for review now, also being rebased and squashed.

Thanks @matejcik for writing the reference client, the `ScriptUI` implementation then is quite straightforward. I think we should keep it as a part of the code, it is currently in the `cli` folder.

It was tested both on `T1` and `TT` (emulators, not the physical device - __yet__). The one "problem" I encountered is that the `button-requests` arrive only after the dialogs are finished on the `trezor` side, so the client cannot tell user "There is a dialogue" - but user can see it clearly on the device.

Apart from `trezorctl --script get-address -n "m/49'/0'/0'/0/0"` I did not try any other functionality - we should probably verify some more cases, even though the interaction will be very similar/same.

python/src/trezorlib/ui.py
```diff
@@ -172,6 +172,52 @@ def get_passphrase(self, available_on_device: bool) -> Union[str, object]:
...
+
+    Communicates with a client application using print() and input().
+
+    Lot of `ClickUI` logic is outsourced to the client application, which
+    is responsible for supplying the PIN and passphrase.
+
+    Reference client implementation can be found under `cli/test_script_client.py`.
+    """
+
+    def __init__(self) -> None:
```
grdddj, line 186: As `self` is not being used in any method, can it be somehow simplified?

python/src/trezorlib/cli/__init__.py
```diff
@@ -68,8 +73,11 @@ def get_transport(self) -> Transport:
         # if this fails, we want the exception to bubble up to the caller
         return get_transport(self.path, prefix_search=True)
 
-    def get_ui(self) -> ClickUI:
-        return ClickUI(passphrase_on_host=self.passphrase_on_host)
+    def get_ui(self) -> Union[ClickUI, ScriptUI]:
```
matejcik, line 76: this should presumably be returning `TrezorClientUI`

python/src/trezorlib/ui.py
matejcik: you can make everything a `@staticmethod` and pass in `ScriptUI` class object instead of `ScriptUI()` instance

python/src/trezorlib/cli/trezorctl.py
```diff
@@ -181,20 +188,25 @@ def cli_main(
...
-            click.echo(json.dumps({res.__class__.__name__: res.__dict__}))
+            res_dict = protobuf.to_dict(res, hexlify_bytes=True)
+            click.echo(json.dumps({res.__class__.__name__: res_dict}))
         else:
-            click.echo(json.dumps(res, sort_keys=True, indent=4))
+            # Printing the result on one line for scripts, with indent for CLI users
+            if script:
+                click.echo(json.dumps(res))
+            else:
+                click.echo(json.dumps(res, sort_keys=True, indent=4))
```
matejcik, line 209: this seems complicated

IMHO the flow should be:
```
if isinstance(MessageType):
    res = protobuf.to_dict(...)

if script:
    json.dumps(res)  # no newlines
else:
   json.dumps(res, indent=4)  # pretty print
```

python/src/trezorlib/ui.py
```diff
@@ -172,6 +172,52 @@ def get_passphrase(self, available_on_device: bool) -> Union[str, object]:
...
+        pass
+
+    def button_request(self, br: messages.ButtonRequest) -> None:
+        name = br.code.name if br.code else None
```
matejcik, line 190: are there cases when `br.code` is None?

if not, `code` should be `br.code.name`, and `name` should be `br.name` -- this doesn't exist yet, so maybe we can leave it out for now

python/src/trezorlib/ui.py
```diff
@@ -172,6 +172,52 @@ def get_passphrase(self, available_on_device: bool) -> Union[str, object]:
...
+        print(f"?BUTTON code={br.code} name={name} pages={br.pages}")
+
+    def get_pin(self, code: Optional[PinMatrixRequestType] = None) -> str:
+        print(f"?PIN code={code}")
```
matejcik, line 194: this should be `?PIN` if code is None and `?PIN code={code.name}` otherwise

matejcik (commented):
looks good overall. i'll read it once more tomorrow.

the test should be moved somewhere else -- `tools/` maybe? and should not be called `test_` when it's not pytest-able

python/src/trezorlib/ui.py
grdddj: Good idea. Unfortunately, from type-checking view it is problematic, as `def get_ui(self) -> TrezorClientUI:` signature is invalid if we would return just `ScriptUI` (in addition to `ClickUI(...)`) - once we return class, once instance.

(_Maybe_ it is not an issue, as running  the newest `pyright` (`1.1.204`) does not see any problems there. In `Pylance` - probably having an older `pyright` version - it is flagged as an issue)

matejcik: yeah, this is both correct and [already fixed](https://github.com/microsoft/pyright/issues/2729)

python/src/trezorlib/ui.py
grdddj: I could not see any case where it is `None`, but it is marked `Optional` in protobuf.

`br.name` is indeed not defined, so I will delete the `name=xxx` part. We could probably leave it in the client, it is `Optional` there

python/src/trezorlib/ui.py
grdddj: If this `UI` is to be used by scripts, should we send `code.name` (a human readable string) instead of just `code` (an int)? The second one allows for easy parsing on the client (`if int(code) == PinMatrixRequestType.Current`). 

But yeah, even in the first case, there can be probably a comparison `if code == PinMatrixRequestType.Current.name` and client can directly print it to the user

matejcik: It's an enum so we should be exposing the _symbol_, not the wire value. If you want to get back the enum value (in Python), you can do `PinMatrixRequestType.__members__[code]`. But the scripts might not be Python and might not have the enum library available -- in which case it is much nicer to rely on "human-readable" symbols than on magic numbers.

Also what I expect the script author to do is `if code == "Current"` without messing with the enum definition

grdddj: Yes, that is indeed  better. Changed in e5670a4

python/src/trezorlib/ui.py
grdddj: Name part deleted in e5670a4

python/src/trezorlib/cli/trezorctl.py
grdddj: Done like this in e5670a4

python/src/trezorlib/cli/__init__.py
grdddj: Done in e5670a4

python/src/trezorlib/ui.py
grdddj: Changed in e5670a4

grdddj:
> the test should be moved somewhere else -- `tools/` maybe? and should not be called `test_` when it's not pytest-able

Moved under `python/tools/trezorctl_script_client` in e5670a4

grdddj:
`Pyright` in `CI` is failing on that inconsistency between class and instance. The version of `pyright` cannot be even seen in the `CI` log - we might want to include it
```
/builds/satoshilabs/trezor/trezor-firmware/python/src/trezorlib/cli/__init__.py
  /builds/satoshilabs/trezor/trezor-firmware/python/src/trezorlib/cli/__init__.py:81:20 - error: Expression of type "Type[ScriptUI]" cannot be assigned to return type "TrezorClientUI"
    "ScriptUI" is incompatible with protocol "TrezorClientUI"
      "button_request" is an incompatible type
        Type "(br: ButtonRequest) -> None" cannot be assigned to type "(self: Self@TrezorClientUI, br: ButtonRequest) -> None"
          Function accepts too many positional parameters; expected 1 but received 2
            Parameter name mismatch: "self" versus "br"
      "get_pin" is an incompatible type
        Type "(code: PinMatrixRequestType | None = None) -> str" cannot be assigned to type "(self: Self@TrezorClientUI, code: PinMatrixRequestType | None) -> str"
          Function accepts too many positional parameters; expected 1 but received 2
    ... (reportGeneralTypeIssues)
```

matejcik:
> Is it even worth doing

the way I see it, `TrezorClientUI` is a collection of three _functions_ (not methods) `button_request()`, `get_pin()` and `get_passphrase()`. Originally `ClickUI` was also a set of staticmethods, before I added things like "show button request only once". Semantically, it's not a base class, it's a protocol at best. The following is a valid "implementation" of UI:
```
# def button_request(): ...
# def get_pin(): ...
# def get_passphrase(): ...

SomeUI = SimpleNamespace(button_request=button_request, get_pin=get_pin, get_passphrase=get_passphrase)
```

It can be slightly confusing for the reader of `trezorctl` (so please make sure to add a comment pointing out the subtlety, something along the lines "the class object itself is the implementation". But that is not a strong reason to enforce conditions on how `ScriptUI` is implemented

grdddj:
Pushed explanation of the issue and also `# type: ignore` of it, so the `CI` will be green.

In the future we could get rid of the ignore (after `pyright` version increases), it would be flagged by `pyright-tool` to be unused ignore

matejcik (approved):
utACK, please squash fixups

PR #2025, opened 2021-12-17
ci: add arm build for core unix and legacy emulators

vdovhanych:
This PR adds a build for arm emulators needed for trezor-user-env running on arm version of nix. See why [here](https://github.com/trezor/trezor-user-env/pull/114).
This is now done with our m1 runner as docker executor.  I used my nixos image as nixos/nix is completely broken since they switched from alpine to native nixos one. See [here](https://github.com/NixOS/nix/issues/5777#issuecomment-996242619). Once the upstream image is fixed, we can change to that.

This pr also adds native fixes manual macOS build using our runner and built on x86_64 via rosseta2.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -153,6 +153,27 @@ core unix frozen debug build:
     untracked: true
     expire_in: 1 week
 
+core unix frozen debug build arm:
+  image: vdovhanych/nixos
+  stage: build
+  <<: *gitlab_caching
+  needs: []
+  only:
+    - master
+    - /^release\//
+    - /^secfix\//
+  variables:
+    PYOPT: "0"
+  script:
+    - nix-shell --run "poetry run make -C core build_unix_frozen"
+    - mv core/build/unix/trezor-emu-core core/build/unix/trezor-emu-core-arm
+  artifacts:
+    name: "$CI_JOB_NAME-$CI_COMMIT_SHORT_SHA"
+    untracked: true
+    expire_in: 1 week
+  tags:
+    - docker_darwin_arm
+
 core unix frozen btconly debug t1 build:
   stage: build
   needs: []
@@ -174,9 +195,9 @@ core macos frozen regular build:
   needs: []
   when: manual
   tags:
-    - darwin
+    - darwin_arm
   script:
-    - nix-shell --run "poetry run make -C core build_unix_frozen"
+    - nix-shell --option system x86_64-darwin --run "poetry run make -C core build_unix_frozen"
     - export NAME="trezor-emu-core.darwin"
     - cp -v core/build/unix/trezor-emu-core ./$NAME
     - chmod +x $NAME
@@ -300,6 +321,29 @@ legacy emu regular debug build:
     paths:
       - legacy/firmware/trezor.elf
     expire_in: 1 week
+
+legacy emu regular debug build arm:
+  image: vdovhanych/nixos
+  stage: build
+  <<: *gitlab_caching
+  needs: []
+  only:
+    - master
+    - /^release\//
+    - /^secfix\//
+  variables:
+    DEBUG_LINK: "1"
+    EMULATOR: "1"
+  script:
+    - nix-shell --run "poetry run legacy/script/cibuild"
+    - mv legacy/firmware/trezor.elf  legacy/firmware/trezor-arm.elf
+  artifacts:
+    name: "$CI_JOB_NAME-$CI_COMMIT_SHORT_SHA"
+    paths:
+      - legacy/firmware/trezor-arm.elf
+    expire_in: 1 week
+  tags:
+    - docker_darwin_arm
 
 legacy emu btconly debug build:
   stage: build
--- a/ci/deploy.yml
+++ b/ci/deploy.yml
@@ -182,12 +182,16 @@ release core unix debug deploy:
   before_script: []  # no poetry
   needs:
     - core unix frozen debug build
+    - core unix frozen debug build arm
   script:
     - export VERSION=$(./tools/version.sh core/embed/firmware/version.h)
     - DEST="$DEPLOY_PATH/trezor-emu-core-v$VERSION"
-    - echo "Deploying to $DEST"
+    - DEST_ARM="$DEPLOY_PATH/trezor-emu-core-arm-v$VERSION"
+    - echo "Deploying to $DEST and $DEST_ARM"
     - nix-shell -p patchelf --run "patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 core/build/unix/trezor-emu-core"
+    - nix-shell -p patchelf --run "patchelf --set-interpreter /lib/ld-linux-aarch64.so.1 core/build/unix/trezor-emu-core-arm"
     - rsync --delete -va core/build/unix/trezor-emu-core "$DEST"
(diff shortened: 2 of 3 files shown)


matejcik (approved):
utACK

ci/build.yml
```diff
@@ -153,6 +153,22 @@ core unix frozen debug build:
     untracked: true
     expire_in: 1 week
 
+core unix frozen debug build arm:
+  image: vdovhanych/nixos
```
prusnak, line 157: this should work now (new arm64 images have been pushed to docker hub recently:

```suggestion
  image: nixos/nix:latest-arm64
```

ci/build.yml
```diff
@@ -301,6 +317,24 @@ legacy emu regular debug build:
       - legacy/firmware/trezor.elf
     expire_in: 1 week
 
+legacy emu regular debug build arm:
+  image: vdovhanych/nixos
```
prusnak, line 321: same here:

```suggestion
  image: nixos/nix:latest-arm64
```

prusnak (changes requested):

ci/build.yml
vdovhanych: This is the one that is broken (nix 2.5.0) that was pushed to docker hub 3 days ago. The latest image was reverted to old nix on alpine (2.3.16) which doesn't have an arm version.

prusnak: OK

prusnak (approved):

vdovhanych:
The new nixos/nix image was just pushed, I will change it to that and we can merge after that.

prusnak:
> The new nixos/nix image was just pushed

`latest` alias still points to `2.5.0` and not to the fixed `2.5.1` - I am asking on Matrix whether this is intentional

prusnak (approved):

vdovhanych:
So the new image of `nixos/nix` is unable to build core emulator as the image is missing `/var/tmp` folder 👍 the same goes for tenv as its also unable to be built without creating that folder in `Dockerfile`.

prusnak:
> we can just make our base ci image multi-arch one and use that later on.

What would we need to do to make our CI image multi-arch? The Dockerfile is already prepared for multiarch (you can change the arch via the `ALPINE_ARCH` arg).

vdovhanych:
> > we can just make our base ci image multi-arch one and use that later on.
> 
> What would we need to do to make our CI image multi-arch? The Dockerfile is already prepared for multiarch (you can change the arch via the `ALPINE_ARCH` arg).

I'm unsure if it will be as simple as adding buildx and setting two targets to build. We would need to modify the `Dockerile` a bit.  The image [`vdovhanych/nixos`](https://github.com/vdovhanych/nixos-docker-arm64/blob/master/Dockerfile) is pretty similar to our base ci image, so there shouldn't be an issue with that.

prusnak:
Ok. Let's merge this with `vdovhanych/nixos` as base image for arm64 emulator builds and discuss how to optimize later. Thanks

tsusanka (commented):
LGTM for now, but as discussed we need to discuss our approach to arm a bit more "conceptually". But let's keep that for Q1 :).

Btw is there a reason to run the jobs for every branch? I would run it for schedules and 
 `/^release\//` /  `/^secfix\//` branches only?

Agree that we don't need to run it on every branch. Should be fine to have it run only on master. as we want to have latest master emulators in tenv right ?

vdovhanych:
> LGTM for now, but as discussed we need to discuss our approach to arm a bit more "conceptually". But let's keep that for Q1 :).
> 
> Btw is there a reason to run the jobs for every branch? I would run it for schedules and `/^release\//` / `/^secfix\//` branches only?

Agree that we don't need to run it on every branch. Should be fine to have it run only on master. as we want to have latest master emulators in tenv right ?

tsusanka:
Yes, master and also releases/secfix which we want to be deployed to firmware.corp servers (tenv fetches the firmwares from there so new firmware versions are added automatically).

Scheduled is not needed, although I like that it gives us information that "everything is fine from the CI point of view".

vdovhanych:
> Yes, master and also releases/secfix which we want to be deployed to firmware.corp servers (tenv fetches the firmwares from there so new firmware versions are added automatically).
> 
> Scheduled is not needed, although I like that it gives us information that "everything is fine from the CI point of view".

Updated the configuration and also added the jobs for releasing arm emulators to our server.

ci/deploy.yml
```diff
@@ -195,6 +195,27 @@ release core unix debug deploy:
...
+    - DEST="$DEPLOY_PATH/trezor-emu-core-arm-v$VERSION"
+    - echo "Deploying to $DEST"
+    - nix-shell -p patchelf --run "patchelf --set-interpreter /lib/ld-linux-aarch64.so.1 core/build/unix/trezor-emu-core"
+    - rsync --delete -va core/build/unix/trezor-emu-core "$DEST"
+  only:
+    - /^core\//
+  except:
+    - branches  # run for tags only
+  tags:
+    - deploy
```
tsusanka, line 217: Sorry for delaying the PR but is there a reason not to make it part of `release core unix debug deploy`? We can _need_ both of the builds and deploy at the same time. Yes, it will break if one of the builds fails, but I think that is acceptable. We should try to make the jobs compact.

vdovhanych: Both artifacts have the same folder structure and naming, not sure if it won't get overwritten. My first idea was to add it there, but I think it won't work like that.

tsusanka: Oh, I see. Hmm, let's rename it to `*-arm` after the build here? 
https://github.com/trezor/trezor-firmware/pull/2025/files#diff-fad8dced0db200747490daa0aa2fb66c1d30378b9cf5ba032ae96e2e5d8825c1R168

vdovhanych: I tried to do it like that but I'm not sure if its ok that it will overwrite other files but I think that this way we should have both emualtors and we should be able to deploy it with one job.

ci/deploy.yml
```diff
@@ -182,33 +182,16 @@ release core unix debug deploy:
   before_script: []  # no poetry
   needs:
     - core unix frozen debug build
+    - core unix frozen debug build arm
   script:
     - export VERSION=$(./tools/version.sh core/embed/firmware/version.h)
     - DEST="$DEPLOY_PATH/trezor-emu-core-v$VERSION"
-    - echo "Deploying to $DEST"
+    - DEST_ARM="$DEPLOY_PATH/trezor-emu-core-arm-v$VERSION"
+    - echo "Deploying to $DEST and "$DEST_ARM"
```
tsusanka, line 190: echo "Deploying to $DEST and $DEST_ARM"

(extra ")

tsusanka: Feel free to force-push then!

ci/deploy.yml
tsusanka: What other files it would overwrite?

vdovhanych: For the core, we have all of the contents of `core/build/unix` but that is not needed for emulator or tenv. Should be fine 👍

ci/deploy.yml
vdovhanych: fixed in 6f7b404d0

tsusanka (approved):

PR #2027, opened 2021-12-17
build(crypto): make valgrind flag compatible with secp256k1-zkp

prusnak:
It seems to me we have someone forgotten to apply this patch from https://github.com/trezor/trezor-firmware/pull/1897#issuecomment-971600801 = https://github.com/trezor/trezor-firmware/commit/dc785662fb94a2ab44bf9da9f596041d84eff579

This PR fixes this


andrewkozlik (approved):
utACK, but let's wait for @onvej-sl before merging.

onvej-sl:
Thank you, @prusnak. It seems to me that I committed it to a wrong branch.

PR #2029, opened 2021-12-19
fix(crypto): refactor ethereum_address_checksum to prepend "0x"

prusnak:
This is just a simple refactor which moves `resp->address[0] = '0'; resp->address[1] = 'x';` into the `ethereum_address_checksum` function.


andrewkozlik (approved):
utACK

onvej-sl (approved):

PR #2034, opened 2021-12-22
Fix proof of ownership sighash computation

andrewkozlik:
Fixes a bug in the computation of the sighash in ownership proofs. SLIP-0019 specifies that the `scriptPubKey` and `commitmentData` should be hashed as length-prefixed strings [1], but the trezor-core implementation omitted to hash the length.

This also means that we have incorrect test vectors in the SLIP. Update here: https://github.com/satoshilabs/slips/pull/1223.

The connect tests are failing, but should work with https://github.com/trezor/connect/pull/995 (untested).

[1] https://github.com/satoshilabs/slips/blob/master/slip-0019.md#proof-footer


matejcik (approved):
utACK. not sure if we need to wait for the Connect fix, probably not?

andrewkozlik:
> utACK. not sure if we need to wait for the Connect fix, probably not?

I think we should merge the two simultaneously. If we merge this PR prematurely, then `connect test core` will fail in the CI in master.

mroz22:
Feel free to merge this without wating for connect. At the moment, connect tests are running against currently released firmware. Master firmware runs only in the nightly jobs.

matejcik:
@mroz22 we have Connect master tests running in our pipeline too, so we need to merge in sync.

onvej-sl (approved):

andrewkozlik:
If I merge this now, then the master branch and anything that is forked-off of master will fail in the CI due to the connect test failing. We need https://github.com/trezor/connect/pull/995 to be merged, but then trezor-connect's own CI will start failing, because it's running against the currently released firmware. Options:
1. Run trezor-connect's CI against trezor-firmware's master branch.
2. Disable the connect tests in trezor-firmware's CI and run them only in trezor-firmware's nightly build.
3. Allow failure of the connect tests in trezor-firmware's CI https://docs.gitlab.com/ee/ci/yaml/#allow_failure.
4. Some combination of the above.

matejcik:
"allow failure" might be the best choice. We want the Connect tests status to be informative, not necessarily a stopper.

mroz22:
I agree with choice 3. This is quite expectable that changes on lower firmware level might be breaking for trezor-connect.

PR #2036, opened 2021-12-23
fix(core,legacy): Fix domain-only ethTypedData

aloisklink:
When doing Ethereum signTypedData, and the primary type = "EIP712Domain", we should completely ignore the "message" part and only sign the domain.

According to the community, this is technically allowed by the spec, and may be used by ETH smart contracts to save on gas (Ethereum smart contract computational cost).

It's also been supported by MetaMask since https://github.com/MetaMask/metamask-extension/pull/6896 (or July 2019).

Test case generated by [@MetaMask/eth-sig-util](https://github.com/MetaMask/eth-sig-util)'s library, using [this gist by @alisinabh](https://gist.github.com/alisinabh/96bd50fd2f9bf827544585cee40e2f7c).

See https://ethereum-magicians.org/t/eip-712-standards-clarification-primarytype-as-domaintype/3286 for community discussion.

### Todo

- [ ] legacy tests: I couldn't figure out how to run legacy tests on an emulator, so I'll need to wait for CI to test it for me.
- [x] create `towncrier` changelog entries once I know the number of this PR (rebased)


common/tests/fixtures/ethereum/sign_typed_data.json
```diff
@@ -4,6 +4,28 @@
...
     "tests": [
+        {
+            "name": "domain_only_message",
+            "comment": "Bare minimum EIP-712 message",
+            "parameters": {
+                "path": "m/44'/60'/0'/0/0",
+                "metamask_v4_compat": true,
+                "data": {
+                    "types": {
+                        "EIP712Domain": []
```
grdddj, line 15: Theoretically not necessary, but I would also welcome a test-case with non-empty domain-data

grdddj (changes requested):
Some small notes/suggestions on the TT code. As for CI, I will transfer your branch into our repo so that we can run it.

EDIT: CI could be seen here - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/440076528

core/src/apps/ethereum/sign_typed_data.py
```diff
@@ -71,32 +71,46 @@ async def generate_typed_data_hash(
...
+                member_path=[1],
+                show_data=show_message,
+                parent_objects=[primary_type],
+            )
+        )
+
+    full_bytes_to_hash = b"".join(parts)
 
-    await confirm_hash(ctx, message_hash)
+    await confirm_hash(ctx, full_bytes_to_hash)
```
grdddj, line 111: This way even the `\x19\x01` prefix and also the `domain_separator` will be shown on the screen, instead of just showing the `message_hash` as before. We probably do not want that (or at least it changes the previous state).

Hard to say what is the best way of doing it. We could handle the confirmation dialog in if/else depending on `primary_type == "EIP712Domain"`. In that if/else, it could be also possible to set `message_hash = b""` when we do not care about it, so we can keep the below return statement unchanged (and maybe more importantly also get rid of the array-creating-and-appending-and-joining logic)

legacy/firmware/ethereum.c
```diff
@@ -987,12 +987,27 @@ static void ethereum_typed_hash(const uint8_t domain_separator_hash[32],
...
+}
+
 void ethereum_typed_hash_sign(const EthereumSignTypedHash *msg,
                               const HDNode *node,
                               EthereumTypedDataSignature *resp) {
   uint8_t hash[32] = {0};
-  ethereum_typed_hash(msg->domain_separator_hash.bytes, msg->message_hash.bytes,
-                      hash);
+
+  if (msg->message_hash.size) {
```
grdddj, line 1005: As uncovered by CI in https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1934782475, we cannot use this condition, as current state expects all the hashes to be size 32, and will fail otherwise. The check is at https://github.com/trezor/trezor-firmware/blob/master/legacy/firmware/fsm_msg_ethereum.h#L223-L226

We could theoretically just extend that check with `&& msg->message_hash.size != 0` to allow for this new possibility

core/src/apps/ethereum/sign_typed_data.py
matejcik: I didn't notice this originally, but we definitely do want the domain hash to also show. 

Showing the full thing, including `\x19\x01` makes sense.

matejcik:
We might want to do a UX tweak for TT. I am not sure what exactly, but the user should be notified that there is only the domain and if they want to inspect the data, that's what they should look into.

OTOH there is the final confirmation screen so the user always has a chance to back out.

What do you think @aloisklink ?

matejcik:
maybe `should_show_domain` would use title "confirm message" instead of "confirm domain" in that case?

legacy/firmware/ethereum.c
```diff
@@ -987,12 +987,27 @@ static void ethereum_typed_hash(const uint8_t domain_separator_hash[32],
   keccak_Final(&ctx, hash);
 }
 
+// When primaryType="EIP712Domain", we ignore message_hash completely
+static void ethereum_domain_only_typed_hash(
```
matejcik, line 991: instead, add a boolean parameter to `ethereum_typed_hash()`

legacy/firmware/fsm_msg_ethereum.h
```diff
@@ -262,12 +262,17 @@ void fsm_msgEthereumSignTypedHash(const EthereumSignTypedHash *msg) {
...
-  layoutConfirmHash(&bmp_icon_warning, _("EIP-712 message hash"),
-                    msg->message_hash.bytes, 32);
-  if (!protectButton(ButtonRequestType_ButtonRequest_Other, false)) {
-    fsm_sendFailure(FailureType_Failure_ActionCancelled, NULL);
-    layoutHome();
-    return;
+
+  // No message hash when setting primaryType="EIP712Domain"
+  // https://ethereum-magicians.org/t/eip-712-standards-clarification-primarytype-as-domaintype/3286
+  if (msg->message_hash.size) {
```
matejcik, line 268: see above, check `msg->has_message_hash` instead

core/src/apps/ethereum/sign_typed_data.py
aloisklink: Personally, I'd agree with @matejcik. It makes more sense to show the `0x1901` prefix and `domain_seperator`, even though it changes the previous state. I'll add it in, but if we want to keep it using the same state, I can always revert back to the old behaviour.

common/tests/fixtures/ethereum/sign_typed_data.json
aloisklink: Added in 82e1e9693a57ab77a819abf2f6e180925e47c620, although I haven't tested whether it works with Trezor One yet, since I still haven't had the time to figure out how to run tests for it. I'll let CI do it.

core/src/apps/ethereum/sign_typed_data.py
aloisklink: I've removed the array-creating-and-appending-and-joining logic so the code should be more similar to how it previously was, so the total lines changed should be less.

I've made the changes in a `sqaush! ...` commit in https://github.com/trezor/trezor-firmware/pull/2036/commits/5024b0165fe0f45c43af2f9032334d545bdbebd5, so that it will be put in the commit message when we `--autosquash`. It seems like a minor UI change, but I'd also be happy to put it in the CHANGELOG entry too.

I have however kept the change that shows the `0x1901` and `domain_sepeaator` on the screen. I think it makes it more clear, since now it will show the full data that is being hashed, especially since the `domain_separator` is variable data, but I'm happy to revert it if you disagree.

Screenshot below shows it still shows on one screen, so no scrolling required.
emu00000048

legacy/firmware/ethereum.c
aloisklink: Done in https://github.com/trezor/trezor-firmware/pull/2036/commits/2d1a145b172ce6e20cce496e9ba2ba05e4863f76 (although I haven't tested it yet locally, so I'll let CI test it). I've made it a `squash! ...` commit so that the commit message will be updated.

legacy/firmware/ethereum.c
aloisklink: Added as `bool domain_only_hash` in https://github.com/trezor/trezor-firmware/pull/2036/commits/2d1a145b172ce6e20cce496e9ba2ba05e4863f76, although it's untested, so I'll CI test it for me.

legacy/firmware/fsm_msg_ethereum.h
aloisklink: Done in https://github.com/trezor/trezor-firmware/pull/2036/commits/2d1a145b172ce6e20cce496e9ba2ba05e4863f76, although it's untested, so I'll CI test it for me.

aloisklink:
> We might want to do a UX tweak for TT. I am not sure what exactly, but the user should be notified that there is only the domain and if they want to inspect the data, that's what they should look into.
> 
> OTOH there is the final confirmation screen so the user always has a chance to back out.

My gut feeling is that with EIP712 typed signing being so common nowadays in the Ethereum ecosystem, it's pretty likely people would just skip reading the title of transactions, and just do a "tap, tap, tap" while only looking at the contents.

Instead I've changed the "Confirm message" screen so it just says "No message field" in https://github.com/trezor/trezor-firmware/pull/2036/commits/23be403a7947e09e4319d4c80894e342f690b6d4 (see screenshot below). That way there is the same amount of screens to do an EIP712 typed signing.

I'd admit that UI/UX isn't something I'm skilled at, so I'd be happy to change it, or let you guys change it.

emu00000893

grdddj:
Good job! CI could be seen at https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/442570184

Everything is fine there, except for the `legacy test` stage - the `T1` tests did not even run because the testing code failed when parsing the `"message_hash": null` in test vectors in 
https://github.com/trezor/trezor-firmware/blob/33c243be8dae968e591cc481c03c1dca988ce8c5/tests/device_tests/ethereum/test_sign_typed_data.py#L53
(It does not account for possibility of empty `message_hash`)
We could do 
```python
ethereum.decode_hex(parameters["message_hash"]) if parameters["message_hash"] else None
```

--------
The `UI/UX` for this special case seems fine for me - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1947225717/artifacts/master_diff/added/ethereum-test_sign_typed_data.py::test_ethereum_sign_typed_data[full_domain_empty_message].html

There is also a UI change in all the "normal" EIP712 test-cases where the `message_hash` is nonzero. It causes the final hash to be longer and it currently does not fit on one screen. Now it shows the whole hash, but at the expense of user having to swipe the screen once to confirm it. I am fine with it, even though regular users would maybe find it troublesome (no possibility of "tap tap tap", they would now need to "tap tap swipe tap" :))
Example: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1947225717/artifacts/master_diff/diff/ethereum-test_sign_typed_data.py::test_ethereum_sign_typed_data[basic_data].html

-------
With regards to running the `T1` tests locally, I also do not know how to do it, but will try to find out :)

matejcik:
> Instead I've changed the "Confirm message" screen so it just says "No message field"

this is perfectly reasonable, thanks.

matejcik:
> (It does not account for possibility of empty `message_hash`)

should we perhaps modify `decode_hex` to also accept `None` ?

legacy/firmware/ethereum.c
```diff
@@ -976,23 +976,34 @@ int ethereum_message_verify(const EthereumVerifyMessage *msg) {
...
 }
 
 void ethereum_typed_hash_sign(const EthereumSignTypedHash *msg,
                               const HDNode *node,
                               EthereumTypedDataSignature *resp) {
   uint8_t hash[32] = {0};
-  ethereum_typed_hash(msg->domain_separator_hash.bytes, msg->message_hash.bytes,
-                      hash);
+
+  if (msg->has_message_hash) {
```
matejcik, line 1001: i believe you don't need this `if`. The field `message_hash.bytes` still exists and is of the right type

so just pass `msg->has_message_hash` as the 3rd argument

matejcik (changes requested):
ACK on the UI changes.

please modify the test `show_more_button` so that it scrolls through the final hash -- you'll need to insert a `client.debug.swipe_up()` in the appropriate place
(https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1947225717/artifacts/master_diff/diff/ethereum-test_sign_typed_data.py::test_ethereum_sign_typed_data_show_more_button.html)

matejcik:
you can run T1 tests locally the normal way :)
1. build the T1 emulator:
```sh
export EMULATOR=1 DEBUG_LINK=1
cd legacy
./script/setup
./script/cibuild
```
2. run the T1 emulator: `./firmware/trezor.elf`
3. run the tests: from repository root, `pytest tests/device_tests`

grdddj:
> should we perhaps modify `decode_hex` to also accept `None` ?

That could be done, but currently putting `None` into it is so rare that I think it is not worth complicating the `decode_hex` by taking and returning `Optional`. (It could also mean having to handle some type-checking issues)

grdddj:
> 



> you can run T1 tests locally the normal way :)
> 
>     1. build the T1 emulator:
> 
> 
> ```shell
> export EMULATOR=1 DEBUG_LINK=1
> cd legacy
> ./script/setup
> ./script/cibuild
> ```
> 
>     2. run the T1 emulator: `./firmware/trezor.elf`
> 
>     3. run the tests: from repository root, `pytest tests/device_tests`

Thanks :) After getting rid of the `None`-error, all the `T1` tests are fine locally.

One needed change is to modifiy `trezorlib::ethereum::sign_typed_data_hash`'s signature to include `message_hash: Optional[bytes]` for type-checking purposes (even though the tests are not being type-checked).

Also the `CLI` function `trezorlib::cli::ethereum::sign_typed_data_hash` needs to account for the possibly-empty-message-hash inputted from user (possibly with the same `ethereum.decode_hex(message_hash_hex) if message_hash_hex else None`)

aloisklink:
`trezorlib::ethereum::sign_typed_data_hash` and `trezorlib::cli::ethereum::sign_typed_data_hash` have been updated in https://github.com/trezor/trezor-firmware/pull/2036/commits/80853316dd1beacd96078208a6351393a02763f6

I managed to get the T1 emulator working! Thanks for those instructions @matejcik (btw, it's probably worth adding it into the docs: https://docs.trezor.io/trezor-firmware/index.html)

I've also tested the CLI for T1 and it also works:

```bash
```

legacy/firmware/ethereum.c
aloisklink: Done in https://github.com/trezor/trezor-firmware/pull/2036/commits/34b0008275b4fcfbad801e4ffa4416fbe561a772 and I managed to confirm it works locally :smile:

aloisklink:
> There is also a UI change in all the "normal" EIP712 test-cases where the message_hash is nonzero. It causes the final hash to be longer and it currently does not fit on one screen. Now it shows the whole hash, but at the expense of user having to swipe the screen once to confirm it. I am fine with it, even though regular users would maybe find it troublesome (no possibility of "tap tap tap", they would now need to "tap tap swipe tap" :))
Example: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/1947225717/artifacts/master_diff/diff/ethereum-test_sign_typed_data.py::test_ethereum_sign_typed_data[basic_data].html

Good catch, I didn't notice it. I've slightly changed this screen by paging it. Since the hash is created from the domain/message screens, I though it'd be quite unlikely that somebody would want to view the full hash, so using pagination seems to be better than swiping (plus it keeps it as a "tap tap tap" :) )

I've also updated the `show_more_button` test so it swipes through the new show more.

Changes made in https://github.com/trezor/trezor-firmware/pull/2036/commits/0cec5c9755a16f3e34ede89b9fb8b5a25d4b0285

emu00000000

matejcik:
The final screen should ​be hold-to-confirm.
(We use normal confirms in the Bitcoin and Ethereum SignMessage, and arguably we should have a hold-to-confirm there as well. Given how important EIP-712 seems to be for smart contracts, I'm guessing there is potential for coin loss if you sign a message that you didn't intend, so there should be at least this.

I don't think the "show more" dialog supports this at the moment. However, UX-wise, the hash doesn't seem to be great content for the hold-to-confirm anyway.

I realize that I originally said that we want to keep the hash, but in the end it does seem pointless -- nobody is ever going to be verifying the hash against anything, except perhaps as a debugging tool (which should not be in the face of normal user's workflow).

I'll confirm this with the team, but right now it seems better to not show it at all, and instead put something like "really sign EIP-712 message?" at end

aloisklink:
> I don't think the "show more" dialog supports this at the moment. However, UX-wise, the hash doesn't seem to be great content for the hold-to-confirm anyway.

I read your mind and spotted the same issue and added it in https://github.com/trezor/trezor-firmware/pull/2036/commits/7c23e86543ee9cf3d6f473ab4cc79696575344a9 :smile:

emu00000069

matejcik:
> I read your mind

well done :)

> avoid confusing a user with too much information.

Exactly, we don't want to train users to click through meaningless dialogs.

For now let's keep it as it is now, I'll confirm what we want to do here and get back to you.

matejcik:
@aloisklink final decision: please replace the hash screen with a simple "Really sign EIP-712 message?" and a hold-to-confirm.

the wording is not awesome, so if you have a better idea, feel free.

matejcik:
> Would you like me to add a commit reverting [7c23e86](https://github.com/trezor/trezor-firmware/commit/7c23e86543ee9cf3d6f473ab4cc79696575344a9)?

yes please. even if we wanted to keep the code, we would want it in a separate commit



> Maybe the title could be "Confirm typed data" with the message as "Confirm signing EIP-712 typed data?"

sure, that makes sense.

aloisklink:
I've reverted https://github.com/trezor/trezor-firmware/commit/7c23e86543ee9cf3d6f473ab4cc79696575344a9 and https://github.com/trezor/trezor-firmware/commit/0cec5c9755a16f3e34ede89b9fb8b5a25d4b0285 since the changes in them don't matter anymore. I'll manually remove them in the final `git rebase -i --autosquash` before merging.

The new final confirm screen in https://github.com/trezor/trezor-firmware/pull/2036/commits/30c84dce85601784b791d333ce905021212ac9ce looks like the following:

emu00000000

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -694,6 +694,7 @@ async def confirm_text(
     br_code: ButtonRequestType = ButtonRequestType.Other,
     icon: str = ui.ICON_SEND,  # TODO cleanup @ redesign
     icon_color: int = ui.GREEN,  # TODO cleanup @ redesign
+    hold: bool = False,  # TODO Only supported if NOT paginating
```
aloisklink, line 697: Let me know if you want me to add support for `hold=True` when paginating too (it's just one line of code, but it won't be tested).

matejcik:
> I added a hold parameter to confirm_text() to make it happen, see

`confirm_action` should be used here, which already has hold-to-confirm

core/src/apps/ethereum/layout.py
```diff
@@ -110,12 +111,11 @@ def require_confirm_data(ctx: Context, data: bytes, data_total: int) -> Awaitabl
...
 
 async def confirm_typed_data_final(ctx: Context) -> None:
-    await confirm_text(
+    await confirm_action(
         ctx,
         "confirm_typed_data_final",
         title="Confirm typed data",
-        data="",
-        description="Confirm signing EIP-712 typed data?",
+        action="Confirm signing\nEIP-712 typed data?",
```
matejcik, line 118: didn't confirm_action auto-wrap the text?

aloisklink: In my opinion, manually wrapping it looks better, but I'm happy to let it auto-wrap if that's preferable.

(for some reason the "p" button didn't take a screenshot, so here's a manual screenshot)
image

core/src/apps/ethereum/layout.py
```diff
@@ -109,16 +110,26 @@ def require_confirm_data(ctx: Context, data: bytes, data_total: int) -> Awaitabl
...
-        title="Confirm hash",
-        data="0x" + hexlify(message_hash).decode(),
+        "confirm_typed_data_final",
+        title="Confirm typed data",
+        action="Confirm signing\nEIP-712 typed data?",
         hold=True,
     )
 
 
+def require_confirm_empty_typed_message(ctx: Context) -> Awaitable[None]:
```
matejcik, line 123: please rename to just `confirm_...` (the `require_` is an older naming convention that is not used anymore, everything is "require" nowadays)

aloisklink: Fixed in https://github.com/trezor/trezor-firmware/pull/2036/commits/ec280a2cab209f3c6700541b3daf0efaafdd7b42

For some reason, the fixtures seemed to have changed too, but maybe I misrecorded them in the earlier commit (e.g. by running the emulator with animations on)
image

core/src/trezor/ui/layouts/tt/__init__.py
matejcik: let's not add functionality that we don't need. the current state of code is OK

aloisklink:
Sorry, forgot to add a towncrier changelog entry when we added changes to the `python` folder and that seems to have prevented CI tests from running.

Added in https://github.com/trezor/trezor-firmware/pull/2036/commits/1cb29d3477889e6e23c0fe70baf672b20f00f758

Any chance you can restart the CI job? :smile:

aloisklink:
It seems like CI failed at `connect test core` in the tests "TrezorConnect methods › TrezorConnect.signTransaction › Testnet (P2WPKH): with proof" and "TrezorConnect methods › TrezorConnect.signTransaction › Testnet (P2TR): with proof"

I'm guessing this is a false positive since I didn't touch any Bitcoin code?
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1972352609

matejcik:
I squashed and rebased the code, with one more tweak: the "hold to confirm" button should read "Hold to confirm".

after rebase, the Connect test will very likely be OK (or at least cause an expected failure)

matejcik:
and we're green!
thanks for your contribution :100:

aloisklink:
> and we're green! thanks for your contribution

And thank you for being patient and giving great code reviews :rocket:

PR #2038, opened 2021-12-26
Update chains, coins & support status

mmilata:
I'd like to ask for brief review since `eth:ETH` has become `eth:ETH:1` due to duplicate key with ethereum-lists/chains#528.


yura-pakhuchiy:
Boba chain is legit, top4 roll-up by TVL according to https://l2beat.com/.

mmilata:
I'm gonna merge this now so as not to hold up the release.

@yura-pakhuchiy unfortunatelly we currently don't support networks that use the same shortcut as other existing network (ETH in this case). If you want to use Boba with Trezor can you please open an issue?

PR #2040, opened 2021-12-27
fix fujicoin params

fujicoin:
I want to modify "xpub_magic_multisig_segwit_p2sh" and "xpub_magic_multisig_segwit_native" to prefix them to "Ypub" and "Zpub" respectively.

Fujicoin completed taproot activation in May.
https://www.fujicoin.org/fujicoincore_release_history.html


prusnak:
Please provide test vectors including a Taproot transaction generated by Trezor and accepted by Fujicoin.

Related question: Which Fujicoin wallet supports Taproot and Trezor at the same time?

fujicoin:
@prusnak 
Fujicoin core v2 2.0 supports Taproot and Trezor at the same time.
The next transaction is a remittance from the descriptor wallet to which Trezor is connected to the bech32m (P2TR) address.
https://explorer.fujicoin.org/tx/ab0a59e10a37152aebbdaff234fb71aa6ac0007861c44e82dcf8a9cd384e1898

Currently, you can send money to a P2TR address, but Trezor will not generate a P2TR descriptor, so a wallet with Trezor connected will not be able to generate a P2TR address.

Fujicoin core v2 2.0 and HWI can be downloaded from the following.
https://www.fujicoin.org/en/download_core

If you want to test it, you can use Fujicoin's raw DB to sync to the network right away.
https://www.fujicoin.org/en/download_db

Currently you need to use a command on the console to generate a P2TR address.
`getnewaddress" "" bech32m "`

Bitcoin core does not yet support P2TR address generation, but Fujicoin has already supported it.

prusnak:
Please provide P2TR unit test similar to https://github.com/trezor/trezor-firmware/blob/33b23ea6a0cfc5360f7bfcb92c2b3cdf241ead24/tests/device_tests/bitcoin/test_grs.py#L222-L244

prusnak:
> so a wallet with Trezor connected will not be able to generate a P2TR address.

Sure it can. Just compile your own firmware using the PR you just provided.

fujicoin:
@prusnak 
I created a Trezor Emulator using my PR and confirmed the transaction with the P2TR address in the Trezor Suite with Fujicoin implemented.
See the figure below.

trezor_suite1

trezor_suite2

trezor_suite3

trezor_suite4
382366-551e-4c8a-b06b-0bee9e3fcd21.png)

Check the transaction in the blockbook.
https://explorer.fujicoin.org/tx/ed5e2655a8bdaf29c7e1c40e7033db140fc519b43dd5b04fbb6e6966acb369fa

I confirmed it on the actual mainnet, so I don't think it is necessary to write a test, but what about?

fujicoin:
I have created a binary for Fujicoin Trezor Suite. I think it's easy to test with this.
https://github.com/fujicoin/trezor-suite/releases/tag/f21.12

matejcik:
> I confirmed it on the actual mainnet, so I don't think it is necessary to write a test

both steps are important:

(a) that we have internal confirmation that the transaction signature is generated on Trezor correctly, even if something changes in the future,
(b) and that the transaction, generated by Trezor, has been accepted by network (so we know that the data in the test is valid)

Usually you would want to create a testnet transaction, so that you can use the default `all all all...` x12 seed and don't have to put actual funds on it.

fujicoin:
Added P2TR test.

fujicoin:
I'm running a test, but test_fujicoin.py fails.
The actual transaction is as follows:
[real_tx.txt](https://github.com/trezor/trezor-firmware/files/7805210/real_tx.txt)

The transaction generated by the test is as follows:
[test_tx.txt](https://github.com/trezor/trezor-firmware/files/7805211/test_tx.txt)

I think "txid" is different and natural.
Why are the "sequence" numbers so different?
There is a difference in "txinwitness", but it seems that they will not be the same if the root key is different in the first place.

What should I do now?

fujicoin:
Rewrote serialized_tx.hex () so that test passes.

fujicoin:
Recreated test_fujicoin.py based on a new transaction

・The emulator was newly started with the following command.
`./emu.py --mnemonic" all all all all all all all all all all all all "-p test`
・Created a transaction with RBF turned off in Fujicoin Trezor Suite.

The test now passes because the emulator and the test root key are the same.

matejcik:
thanks for the test case, merged now

PR #2049, opened 2022-01-05
tests(crypto): update wycheproof testcase source and parsing

invd:
The JSON schema of the test cases has slightly changed.
The "curve" field has moved to the "testGroups" level and the X25519 algorithm has been renamed, which requires adjustments in the parsing code.

A duplicate definition of "ecdh_vectors" has also been removed.


onvej-sl (approved):

onvej-sl:
Well done, @invd !

PR #2051, opened 2022-01-05
Update to pyright-1.1.202

mmilata:
So far we've been using pyright from the pinned nixpkgs commit specified in `ci/shell.nix`. This way it's not possible (afaik) to specify the precise version, all we can do is bump the nixpkgs version to latest unstable and get pyright version which was on NPM last time it was synced with nixpkgs.

This change moves the package definition to trezor repo while using the [same tooling](https://github.com/svanderburg/node2nix). Unfortunately updating pyright now involves generating some `.nix` files so README.md is needed to document the process.


prusnak:
Shouldn't we just update pyright to latest (1.1.203) in nixpkgs and use pinned unstable once the update is merged to master?

prusnak:
Ah, I know see it is very hard to update an individual package and whole nodePackages needs to be updated at once ... :-/

mmilata:
Yeah, IME you need to let node2nix download half of NPM, pray that no package breaks, then find someone willing to review such PR. Probably not flexible enough given how often pyright releases new versions.

ci/shell.nix
```diff
@@ -39,6 +39,8 @@ let
   };
   gcc = nixpkgs.gcc11;
   llvmPackages = nixpkgs.llvmPackages_12;
+  # see pyright/README.md for update procedure
+  pyright = nixpkgs.callPackage ./pyright/default.nix {};
```
prusnak, line 43: small nit-pick - `default.nix` is not required because it is default

```suggestion
  pyright = nixpkgs.callPackage ./pyright {};
```

prusnak (approved):
tested ACK

mmilata:
CI now failing on:
```
python/src/trezorlib/cli/firmware.py:287:66 - error: "closest_version" is possibly unbound (reportUnboundVariable)
```

prusnak:
> ```
> python/src/trezorlib/cli/firmware.py:287:66 - error: "closest_version" is possibly unbound (reportUnboundVariable)
> ```

That's a valid finding - @matejcik any ideas how to fix that in the cleanest way?

matejcik:
i have a fix in the typing branch

in the meantime .204 was released which fixes the issue that kept me on .202, so I'll go ahead and try to bump the version here

matejcik:
thanks, I will merge this as part of #1939 (https://github.com/trezor/trezor-firmware/pull/1939/commits/f54a65f29337438ec8d0d2253996940143599f52)

PR #2046, opened 2022-01-04
fix(crypto): fix incorrect sig buffer size

invd:
The function expects `sig` to be larger, see `// signature_bytes has 64 bytes`.
Correcting the buffer resolves memory safety issues during the test run.

Closes https://github.com/satoshilabs/trezor-firmware/issues/130.

PR #2052, opened 2022-01-06
ci: Use address sanitizer in CI.

andrewkozlik:
Issue: https://github.com/satoshilabs/trezor-firmware/issues/130.

I added the address sanitizer to most of the test builds. What remains to be done, @mmilata:
1. For all the core `test_emu*` tests in the nightly build replace the standard emulator with the emulator that uses the address sanitizer. This emulator can be build using `build_unix_debug` as described in https://github.com/trezor/trezor-firmware/blob/master/docs/tests/device-tests.md#building-for-debugging-emulator-only.
2. For the `legacy test` and `legacy btconly test` in the nightly build replace the standard emulator with the emulator that uses the address sanitizer. This can be done by defining the environment variable `ADDRESS_SANITIZER=1` during the build.

Note that the crypto tests are failing right now, because there is a bug, which will be fixed in https://github.com/trezor/trezor-firmware/pull/2046.


mmilata:
Pushed my changes and now CI is failing, looking into it ...

invd:
@andrewkozlik small potential improvement: there is an optional `-fsanitize-address-use-after-scope` mode flag [in clang](https://clang.llvm.org/docs/AddressSanitizer.html#id1) and [gcc](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.htmlhttps://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html) which may catch additional bugs at a small overhead cost. I think this can be added to the PR but have not re-tested the proposed changes with it.

mmilata:
Hmm, the `core unit test` has bunch of:
```
Direct leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7fd246be5a7f in __interceptor_malloc (/nix/store/r6hnfzkm1q6awqmgr04nc8nfxrpbjvsq-gcc-10.3.0-lib/lib/libasan.so.6+0xaca7f)
    #1 0x7fd2468bdce3 in SDL_malloc_REAL (/nix/store/crw7z48h4ni8imy5nay5xqp6l8j6z29m-SDL2-2.0.14/lib/libSDL2-2.0.so.0+0xa2ce3)

Direct leak of 21 byte(s) in 1 object(s) allocated from:
    #0 0x7fd246be5a7f in __interceptor_malloc (/nix/store/r6hnfzkm1q6awqmgr04nc8nfxrpbjvsq-gcc-10.3.0-lib/lib/libasan.so.6+0xaca7f)
    #1 0x7fd245d63357 in __vasprintf_internal (/nix/store/ff88p8pnhdmf8bflzbxldys21djw9dp0-glibc-2.33-56/lib/libc.so.6+0x7b357)
```
Perhaps we can fix the SDL-related one but not sure about the libc. Can we just turn off malloc/free leak checking as it's not really interesting for us?

`core unit test t1` failure seems rust-related, investigating.

andrewkozlik:
> Can we just turn off malloc/free leak checking as it's not really interesting for us?

I think it would be good to at least try to fix them. If we can't, then they can be disabled:
`ASAN_OPTIONS=detect_leaks=0 ./run_tests.sh`

andrewkozlik:
> @andrewkozlik small potential improvement: there is an optional `-fsanitize-address-use-after-scope` mode flag [in clang](https://clang.llvm.org/docs/AddressSanitizer.html#id1) and [gcc](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.htmlhttps://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html) which may catch additional bugs at a small overhead cost. I think this can be added to the PR but have not tested the proposed changes with it.

@invd this check is enabled by default [1]. Thanks to it I already fixed a bug https://github.com/trezor/trezor-firmware/pull/2052/commits/dd411d9eca46d6f655d3026455af2b2e8bd8c0f8. 

I wasn't able to fix the SDL-related memory leak, so it looks like we will need to disable leak detection, unless someone else is able to figure it out.

The runtime errors, such as:
`runtime error: null pointer passed as argument 1, which is declared to never be null`
are due to micropython which sometimes calls `memcpy` or `memset` with `NULL` pointers and relies on compilers treating it as a no-op, which they do. However, the behavior is undefined according to the C standard.

[1] https://github.com/google/sanitizers/wiki/AddressSanitizerUseAfterScope

mmilata:
Looks like everything is green in CI except the `crypto test`. During final rebase please uncomment the `only:\n- schedules` lines in ci/{build,test}.yml so it only runs on nightly.

mmilata:
Out of scope of this PR but i tried clang+ASan and it makes emulator crash on start, any ideas?
```
Waiting for emulator to come up...
AddressSanitizer:DEADLYSIGNAL
=================================================================
==768125==ERROR: AddressSanitizer: stack-overflow on address 0x7ffe7aa3f0b0 (pc 0x7ffe7aa3f0b0 bp 0x7ffe7aa3eef0 sp 0x7ffe7aa3eef8 T0)
    #0 0x7ffe7aa3f0b0  ([stack]+0x1f0b0)

SUMMARY: AddressSanitizer: stack-overflow ([stack]+0x1f0b0) 
==768125==ABORTING
```

andrewkozlik:
> Looks like everything is green in CI except the `crypto test`. During final rebase please uncomment the `only:\n- schedules` lines in ci/{build,test}.yml so it only runs on nightly.

Something is not right. It looks like the unit tests and device tests (and maybe other tests) are running against the regular emulator. We should at least be seeing the `runtime error: null pointer passed as  argument...` errors.

mmilata:
I didn't add CI jobs for unit tests, will do so. But `core device asan test` should be using the asan build, the warnings are written to trezor.log.

andrewkozlik:
> I didn't add CI jobs for unit tests, will do so.

Well the weird thing is that in https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/443749132 the unit tests ran against the sanitizing emulator and yet the `core unit test` definition didn't change at all during the entire time. So it would seem that the emulator which gets chosen is not very well defined.

> But `core device asan test` should be using the asan build, the warnings are written to trezor.log.

Oh yeah I missed that.

mmilata:
Yes I had a bug in SConstruct.unix that made *all* jobs sanitized in that pipeline. Sorry, should have mentioned that.

invd:
> > 
> 
> @invd this check is enabled by default [1]. Thanks to it I already fixed a bug [dd411d9](https://github.com/trezor/trezor-firmware/commit/dd411d9eca46d6f655d3026455af2b2e8bd8c0f8).

Ah, nice. Typical compiler documentation is unfortunately not very clear on the default behavior for individual versions when it comes to flag details like this.

mmilata (approved):

matejcik:
this can be merged, correct @mmilata ? There are pending auto-reviews, but don't seem relevant.

mmilata:
yep i think we can merge it

PR #2054, opened 2022-01-09
fix(ci): copy pyright directory in Dockerfile

prusnak:
follow-up to https://github.com/trezor/trezor-firmware/pull/1939


matejcik (approved):
utACK

Issue #1971, opened 2021-12-03
Add test for Info button in Shamir and SuperShamir backup


matejcik:
Similar to #527, an Info button is present in the backup flow, both for Shamir and SuperShamir.
We also need to have UI tests for it.

Afsoning:
> Similar to #527, an Info button is present in the backup flow, both for Shamir and SuperShamir.
> We also need to have UI tests for it.

PR #2061, opened 2022-01-11
feat(tests): clicking Info button in (Super)Shamir backup device tests

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1971

Backup Shamir and SuperShamir device testing functions extended to click the `Info` button where possible, so we have it in the `UI` tests.

Considerations/possible improvements:
- Not sure if the `click_through_info_button_shamir` should be in `common`, if it is used only by one file currently
- We are currently using `click_through(client.debug, screens=XX, code=B.ResetDevice)` repeatedly, where `screens=XX` is the only changing parameter, we could create a `click(screens: int)` helper function to wrap it and remove duplication


matejcik:
please replace usages of `click_through` with simple `yield; client.debug.press_yes()`, and please inline the explanatory comments ([like so](https://github.com/trezor/trezor-firmware/blob/master/tests/device_tests/test_msg_backup_device.py#L42-L49))
feel free to leave out the assert on the buttonrequest type, it is checked by the expected responses anyway.

The purpose of `click_through(screens=N)` is to click _through N screens_, so that we can run 16 clicks in one line. This is no longer the case because of the info clicks -- the longest N is 3, which is the _only case_ where it might make sense to do
```python
        # 1. Threshold (3)
        # 2. Checklist
        # 3. Confirm show seeds
        yield from click_through(screens=3)
```
(numbering is there to explain why the following `click_through` call has exactly N screens)
in all the other places, clicking explicitly is more readable and means you can put the comment on the yield statement

please declare the info button in `buttons.py`, something like `INFO_SHAMIR_RECOVERY = (LEFT, BOTTOM)` will probably work?
please rename `click_through_info_shamir` to just `click_info` (it's not clicking _through_ anything) and move it to `test_msg_backup_device`

what happens if you leave out the `wait_layout` calls? I would expect that just the `yield` will do the right thing, but I'm not sure.

the UI tests show some screen corruption. please verify that this is not what really happens :) i believe it's a result of interplay of `press_yes()` and `click()` in the same sequence, and we'll wait for the Rust UI to resolve this.

grdddj:
Suggestions from the comment above done in cac49d3

> what happens if you leave out the wait_layout calls? I would expect that just the yield will do the right thing, but I'm not sure.

I have tried omitting the `wait_layout` calls, but they did not solve the situation, leaving the first one caused the test to freeze on the info screen. `press_yes()` was still not possible in any case.

> the UI tests show some screen corruption. please verify that this is not what really happens :) i believe it's a result of interplay of press_yes() and click() in the same sequence, and we'll wait for the Rust UI to resolve this.

At least this is consistent locally and in `CI`, so we do not get inconsistent results. I can confirm that when you click the `Info` button manually, there is no corruption on the screen

matejcik (approved):
wonderful, please rebase & merge

mmilata:
It seems that this change is breaking nightly device tests [sometimes](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2003537360). It occurs more often on the AddressSanitizer-instrumented builds which is interesting.

I've been able to reproduce it locally with:
```
cd core
PYOPT=0 ADDRESS_SANITIZER = 1 make build_unix_frozen
export TESTOPTS="-x -v -k test_msg_backup_device.py"
export PYTEST_TIMEOUT=15 # by default pytest stays frozen
while make test_emu; sleep1; done
```
Output: https://gist.github.com/mmilata/d9e2e40b40b8870dc1b7ebad6710b4d7

matejcik:
> It occurs more often on the AddressSanitizer-instrumented builds

so a race condition, in all likelihood something to do with `wait_layout` which is evil and horrible.

@grdddj would you mind trying to reproduce the issue, and then see if it goes away if you read layouts as they come?
i.e., call `debuglink.watch_layout()` unconditionally, and follow all calls to `press_yes()` with `wait_layout()`, same as the `click()` calls
(if this helps, we can add `wait` argument to `press_yes` so that the code is less verbose)

grdddj:
I was able to reproduce the issue, it failed after 6 successful attempts, with the same Timeout error.

I have only removed the conditional `client.watch_layout()` - so it is called every time - and now it has run 20 times in a row without failure. I will let it run for longer, but it seems it could be the solution. (I did not change anything else, no `wait_layout()`s)

matejcik:
That is very weird though. Does it mean that the builds freeze when watch_layout is not set?

grdddj:
> Does it mean that the builds freeze when watch_layout is not set?

Hard to say. I still did not push the "fix", because when trying it again and again, it still sometimes fails. Funnily enough, when I had the longest streak of around 50 successes (which I killed manually), I found out that in `test_backup_slip39_advanced` which is the problematic one, I did not do any changes (I made the change only in `test_backup_slip39_basic`).

I have increased the timeout to 25 seconds from 15, because the test is long and some other delay could cause it to fail. However, all the timeout-fails show the `Success` message at the end.

grdddj:
After increasing timeout, it achieved 60 successes in a row, so I made a PR with the changes: https://github.com/trezor/trezor-firmware/pull/2084

If you could, please try to verify it also locally, there are a lot of unknowns

mmilata:
I've been running e6dc2008ba7e262b922670169a36e158b7fb30c9 in a loop and it got stuck several times, e.g. on `test_backup_slip39_basic[click_info]`. Some time I've seen following error in the log, bot not every time:
```
620447180259 trezor.wire ERROR exception:
  File "trezor/wire/__init__.py", line 401, in handle_session
  File "trezor/wire/codec_v1.py", line 75, in read_message
  File "trezor/wire/codec_v1.py", line 32, in __enter__
AssertionError: global buffer already used by another context
```
This can indicate that the firmware receives DebugLink message while parsing WebUSB message, or vice-versa.

matejcik:
perhaps a better approach is to pick @andrewkozlik's commit https://github.com/trezor/trezor-firmware/pull/2069/commits/20337a24483ccf695549c6af7315e806d8ced156, fix the popup to accept debuglink click, and then remove the whole watch_layout thing?

matejcik:
if the global buffer is the issue, i'm wondering why we didn't see it sooner / elsewhere

grdddj:
> perhaps a better approach is to pick @andrewkozlik's commit 20337a2, fix the popup to accept debuglink click, and then remove the whole watch_layout thing?

Good idea. I would to it, but have no clue how to `fix the popup to accept debuglink click`

matejcik:
implant this: https://github.com/trezor/trezor-firmware/blob/trezor_obj_get-negative/core/src/trezor/ui/components/common/confirm.py#L68-L76
into this: https://github.com/trezor/trezor-firmware/blob/trezor_obj_get-negative/core/src/trezor/ui/components/tt/info.py#L29

Issue #1954, opened 2021-11-29
NEM tests: investigate difference between T1 and TT


matejcik:
Originally we needed separate test casees for T1 and TT NEM implementation, because the TT variant needed pagination input flows.

These are now automated, so we should be able to merge the T1 and TT variants.

PR #2062, opened 2022-01-11
chore(tests): unifying NEM T1 and TT device tests

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1954

Using the same test functions for signing NEM mosaics in both `T1` and `TT`


matejcik (approved):
utACK, that seems to have gone well :)

PR #2063, opened 2022-01-11
feat(crypto): improve trezor-crypto fuzzer

invd:
Introduce fuzzing harnesses for zkp* and other functions
Adapt some differential fuzzing and other logical checks
Add a new dictionary extraction program
Additional documentation and minor cleanup
Add a temporary workaround for clang-14 and more explicit Makefile behavior

### DIFF
--- a/crypto/Makefile
+++ b/crypto/Makefile
@@ -2,6 +2,13 @@ ifeq ($(FUZZER),1)
 CC       ?= clang
 LD       ?= $(CC)
 SANFLAGS += -fsanitize=fuzzer
+CFLAGS   += -fsanitize-ignorelist=fuzzer/sanitizer_ignorelist.txt
+
+# this works around clang optimization issues in relation with -fsanitize=undefined
+# TODO is there a better solution, for example by disabling a specific optimization technique?
+blake2b.o: OPTFLAGS += -O0
+blake2s.o: OPTFLAGS += -O0
+
 else ifeq ($(ADDRESS_SANITIZER),1)
 SANFLAGS += -fsanitize=address,undefined
 endif
--- a/crypto/Makefile
+++ b/crypto/Makefile
@@ -1,15 +1,33 @@
+# CLANG_VERSION is empty if the compiler is not clang-based
+CLANG_VERSION = $(shell $(CC) --version | grep -Po "clang version \K[\d\.]+")
+CLANG_VERSION_MAJOR = $(shell echo $(CLANG_VERSION) | cut -f1 -d.)
+
+# determine specific version ranges
+ifneq ($(CLANG_VERSION),)
+$(if $(shell [ $(CLANG_VERSION_MAJOR) -ge 13 ] && echo "OK"), \
+    $(eval CLANG_AT_LEAST_13 := true), \
+    $(eval CLANG_AT_LEAST_13 := false))
+$(if $(shell [ $(CLANG_VERSION_MAJOR) -ge 14 ] && echo "OK"), \
+    $(eval CLANG_AT_LEAST_14 := true), \
+    $(eval CLANG_AT_LEAST_14 := false))
+endif
+
 ifeq ($(FUZZER),1)
 CC       ?= clang
 LD       ?= $(CC)
 SANFLAGS += -fsanitize=fuzzer
 
-# TODO gcc as well as older clang versions <= 12 do not support this feature
+# only clang versions >= 13 support this feature
+ifeq ($(CLANG_AT_LEAST_13),true)
 $(info "info: using -fsanitize-ignorelist")
-SANFLAGS   += -fsanitize-ignorelist=fuzzer/sanitizer_ignorelist.txt
+SANFLAGS += -fsanitize-ignorelist=fuzzer/sanitizer_ignorelist.txt
+else
+$(info "info: not using -fsanitize-ignorelist")
+endif
 
 # TODO is there a better solution, for example by disabling a specific optimization technique?
 # there is a clang optimization issue in relation with the blake2 code at -fsanitize=undefined
-$(warning "warning: disable optimization on blake2 code as workaround")
+$(warning "warning: disabling optimization on blake2 code as workaround")
 blake2b.o: OPTFLAGS += -O0
 blake2s.o: OPTFLAGS += -O0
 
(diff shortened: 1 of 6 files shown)


invd:
I've rebased and squashed the changes down to a few commits. Let me know if other changes are requested.

crypto/Makefile
```diff
@@ -3,9 +3,18 @@ CC       ?= clang
 LD       ?= $(CC)
 
 SANFLAGS += -fsanitize=fuzzer
-
 CFLAGS   += $(SANFLAGS)
 LDFLAGS  += $(SANFLAGS)
+
+# TODO gcc as well as older clang versions <= 12 do not support this feature
+$(info "info: using -fsanitize-ignorelist")
+CFLAGS   += -fsanitize-ignorelist=fuzzer/sanitizer_ignorelist.txt
```
mmilata, line 11: This is going to break CI asan builds which are currently using gcc. Can you please make this conditional for clang?

mmilata: are currently -> will be when #2052 is merged

invd: Note that this setting is only used for custom fuzzer builds that satisfy `ifeq ($(FUZZER),1)`. AFAIK those are currently not touched by the CI.

I've thought about the conditional behavior since "silently" enabling or disabling the ignorelist functionality can have unexpected results for the user, for example by hiding bugs or creating performance differences.

Correctly figuring out if the compiler is clang is an involved process as the `$(CC)` program names vary a lot (for example via honggfuzz compiler aliases) and Makefiles don't have native language support for the necessary version string handling and version number comparison. I've added another commit which should cover this in a _reasonable_ fashion.

mmilata: Ah, sorry, I missed the condition. Since it doesn't affect CI the clang detection logic doesn't have to be included.

invd: I think the detection logic is an improvement and makes sense to merge.

Benefits:
* developers who have a clang 14.x behind the `clang` binary otherwise run into the `-Wbitwise-instead-of-logical` issues
* comparing different clang < 13, clang >= 13 versions without Makefile modifications gets easier
* better support for future conditional Makefile behavior related to newly introduced clang functionality

crypto/Makefile
```diff
@@ -50,11 +77,21 @@ ZKP_CFLAGS = \
...
+
 VALGRIND ?= 1
 ifeq ($(VALGRIND),1)
 CFLAGS += -DVALGRIND
 endif
 
+ifeq ($(VALGRIND),1)
+CFLAGS += -DVALGRIND
+endif
+
```
andrewkozlik, line 94: This seems to be a duplicate.

crypto/Makefile
```diff
@@ -1,11 +1,38 @@
...
+# only clang versions >= 13 support this feature
+ifeq ($(CLANG_AT_LEAST_13),true)
+$(info "info: using -fsanitize-ignorelist")
+CFLAGS   += -fsanitize-ignorelist=fuzzer/sanitizer_ignorelist.txt
+else
+$(info "info: not using -fsanitize-ignorelist")
+endif
+
+# TODO is there a better solution, for example by disabling a specific optimization technique?
+# there is a clang optimization issue in relation with the blake2 code at -fsanitize=undefined
```
andrewkozlik, line 32: Can you elaborate on this? Maybe there is a legitimate bug in the code.

crypto/Makefile
invd: You're right, fixed by changing the last commit.

crypto/Makefile
invd: From what I can tell at the moment, some parts of the `blake2b.c` and `blake2s.c` code are very challenging for the clang compiler optimization under certain sanitizer conditions. The result is either a significant slowdown of compilation or a complete abort, neither of which is desirable during frequent re-compilation, hence the workaround.
I think this is more of a problem on the part of clang than anything done wrong in the code (e.g. undefined behavior).  Ultimately I plan to isolate a good reproducer and report this to LLVM, but since clang is not used to produce production firmwares, I haven't prioritized this topic yet.

In my experience, having `-fsanitize=undefined` without `-fsanitize=address` in combination with a high optimization level is a good way to trigger the worst-case behavior, especially on newer clang versions.

Example:
* `CC=clang-14 FUZZER=1 SANFLAGS="-fsanitize=undefined" OPTFLAGS="-O3 -g0" make fuzzer` 
```
[...]
1.	<eof> parser at end of file
2.	Code generation
3.	Running pass 'Function Pass Manager' on module 'blake2s.c'.
4.	Running pass 'Greedy Register Allocator' on function '@blake2s_compress'
[...]
clang: error: clang frontend command failed with exit code 139 (use -v to see invocation)
```
The compilation failure is with the very new `14.0.0-++20220108071805+3e9919cf5750-1~exp1~20220108071855.114`, but the general effect can also be observed with older versions such as `clang 9.0.1-16.1`, so this is not a recent regression.

Notably, the problem is not libFuzzer-related and can also be triggered on "normal" builds, for example via `CC=clang-14 CFLAGS="-fsanitize=undefined" OPTFLAGS="-O3 -g0"  make`. In my experience, it's less problematic there in practice since adding ASan into the mix makes the problem go away, and on regular builds there are much less reasons to _only_ build with UBSan.

andrewkozlik: Thanks for the explanation. Not what I expected :-). LGTM.

invd: Additional debugging shows that the performance problems are also present with the simple `clang-14 -fsanitize=undefined -O1 -o blake2b.o -c blake2b.c` compilation run, so they're not related to any other compiler flags or Trezor-specific settings that we could change.

Adding `__attribute__ ((optnone))` to `blake2b_compress()` makes the problem effectively disappear, compilation times go down from ~30s to less than ~1s for the file at `-O1` for `clang-14`. Gcc `11.2.0` shows no comparable effects when marking the compress function with `__attribute__ ((optimize("0")))`. 

I think this is additional evidence that it is a clang-only problem which does not negatively affect the production binaries.

andrewkozlik (approved):
utACK. I didn't review the fuzzer itself, but I don't think it's necessary given that it does not impact the firmware.

invd:
Adapted the code to be compatible with 56126b75a7fdadf9c7fe41a5e1090c2e555b71a7 and 2880a2fa33104e0be9c6b79648d0b97212e284dd .

invd:
I've removed the schnorr function tests after #2090.

invd:
Minor additions:
* start using the BIP32 & BIP39 cache clearing functions 
* add some documentation

@andrewkozlik @mmilata @onvej-sl are there other changes you would like to see before this can be merged?

invd:
@andrewkozlik @mmilata @onvej-sl @matejcik : could you let me know how we can make progress with this?
From what I can tell, there are no merge conflicts or other important changes to the `crypto` logic since the last push in Februrary, so this PR should be ready to merge.

matejcik:
this seems to completely avoid the crypto code, so if you don't have anything against @andrewkozlik, please just merge?

invd:
Thanks!

Issue #1430, opened 2021-01-25
Payment requests


tsusanka:
Product Spec: [Notion](https://www.notion.so/satoshilabs/Payment-Requests-a8e8b9cca3274e2ea86179873f0edc51)
Tech Spec: [Google Doc](https://docs.google.com/document/d/1gnN2faUZ9Ymc2-QeKD5EcY6aBEOnyzx4h3fhD_N2H3o/edit#heading=h.a30oxabks97l)

andrewkozlik:
@matejzak regarding the design of showing output addresses: Note that there could be multiple outputs in the payment request, but the mainstream use-case is just one output. I imagine it could look something like this:
```
Confirm sending
0.05 TEST to
trezor.io

Buying 15.9636 DASH.

[grey "Details"] [green check-mark]
```

If there is just one output, then instead of "Address #1" we would show just "Address". Pressing "Details" would show:
```
Address #1
0.02 TEST to
tb1qkvwu9g3k2pdxew
fqr7syz89r3gj557l3
uuf9r9

[red cancel-mark] [grey "Next"]
```

Pressing "Next" would show:
```
Address #2
0.03 TEST to
tb1qldlynaqp0hy4zc
2aag3pkenzvxy65sae
sxw3wd

[red cancel-mark] [grey "Done"]
```

Pressing "Done" would show the initial "Confirm sending" screen again, where the user could confirm.

andrewkozlik:
I have been giving some more thought to the implementation and what I proposed above might be unnecessarily complicated, because it allows the user to repeatedly cycle through the outputs. Currently we don't even allow the user to go back and forth with normal outputs, so this might not be worth the effort at the moment. The simplest thing I can think of is allowing the user to just "unpack" the outputs for one pass-through:

```
Confirm sending
0.05 TEST to
trezor.io

Buying 15.9636 DASH.

[grey "Details"] [green check-mark]
```

Pressing "Details" would show:
```
Confirm sending
0.02 TEST to
tb1qkvwu9g3k2pdxew
fqr7syz89r3gj557l3
uuf9r9

[red cancel-mark] [green check-mark]
```

Pressing the cancel-mark would cancel the entire signing flow. Pressing the check-mark would show the next output in the payment request:
```
Confirm sending
0.03 TEST to
tb1qldlynaqp0hy4zc
2aag3pkenzvxy65sae
sxw3wd

[red cancel-mark] [green check-mark]
```

[...]

Issue #1394, opened 2020-12-18
Verify CoinJoin coordinator's URL


tsusanka:
In the CoinJoin authorization workflow we display coordinator's URL. As discussed in Product we would like to have this cryptographically signed so Trezor can verify the URL and potentially allow/disallow it. 

![image](https://user-images.githubusercontent.com/1835345/102595075-59b4b380-4117-11eb-81e9-c8cf46788c48.png)

Our general idea is that the coordinator would have some kind of "header" which is signed by SatoshiLabs. What does @andrewkozlik think?

andrewkozlik:
Just to explain how this works at them moment: The name of the service provider gets included in the *commitmentData* of the ownership proof(s) generated by Trezor. So the purpose of this screen is to affirm the user's intent to participate in a CoinJoin transaction with the given service provider. This acts as a mitigation against denial-of-service attacks, so that an attacker cannot reuse the proof of ownership to register an input into a different CoinJoin transaction than the one for which the proof was intended.

So currently this confirmation protects the system as a whole against DoS, but doesn't provide any assurance to the user that this is indeed where the CoinJoin will take place. We could add this feature by using the payment request mechanism that I recently proposed. Once we implement that, it seems pretty simple to apply it to CoinJoin. The signed payment request would get sent to Trezor in the final phase of CoinJoin when the transaction is getting signed.

The added value to the user is not exceptionally significant. The current implementation already protects against coin loss. Furthermore Suite will presumably be using a TLS connection with the coordinator, which will give the user assurance about the coordinator's authenticity. So this feature would just give the user a second level of assurance about who the coordinator is. The coordinator's authenticity is certainly something to be concerned about, because the user trusts the coordinator to mix the user's input with the inputs of other honest users. A malicious coordinator could mix the user's input with the coordinator's own inputs only, giving them a false sense of anonymity.

So I agree this is a good idea once we implement signed payment requests.

tsusanka:
Will be solved by #1429.

tsusanka:
Since we have #1430 for payment requests I do not think we need this one.

PR #2069, opened 2022-01-19
Payment requests in TT

andrewkozlik:
Implements: https://github.com/trezor/trezor-firmware/issues/1430 and https://github.com/trezor/trezor-firmware/issues/1394
Old draft PR: https://github.com/trezor/trezor-firmware/pull/1429
Original proposal (some parts may be outdated): https://docs.google.com/document/d/1gnN2faUZ9Ymc2-QeKD5EcY6aBEOnyzx4h3fhD_N2H3o/
Draft public specification (should be up to date): https://github.com/satoshilabs/slips/blob/master/slip-0024.md

I tried to resolve all comments in the old draft PR, I hope I didn't miss anything. I added more tests and made several additional changes:
- The maximum number of nonces is restricted to one nonce per session in order to minimize the size of the session cache. In the forseeable future we don't have any use-cases for multiple payment requests per transaction. We can change this when such use-cases arise. We could also fairly easily save space by using the session ID + a counter as a nonce.
- Trezor will require the payment request to use a nonce if any memos are present. We can loosen this requirement in the future, but for now it seems wiser to protect against replay when memos are present. In fact this property can be ensured by the trusted party which issues the signatures, so it's no big deal.
- Refactor: Renamed `*_bitcoin_varint` to `*_compact_size` so that we use the same terminology as Bitcoin core, see https://github.com/bitcoin/bitcoin/blob/d0bf9bb6a539f151ec92725d20a2b6c22cb095a5/src/serialize.h#L233-L273. Varint actually means something else in Bitcoin core, see https://developer.bitcoin.org/reference/transactions.html#compactsize-unsigned-integers.


core/src/apps/bitcoin/sign_tx/payment_request.py
```diff
@@ -0,0 +1,88 @@
...
+        writers.write_bytes_prefixed(self.h_pr, nonce)
+        writers.write_bytes_prefixed(self.h_pr, msg.recipient_name.encode())
+        writers.write_compact_size(self.h_pr, len(msg.memos))
+        for memo in msg.memos:
+            writers.write_uint32(self.h_pr, memo.type)
+            if memo.type == MemoType.COIN_PURCHASE:
+                assert memo.amount is not None  # checked by sanitizer
+                assert memo.coin_name is not None  # checked by sanitizer
+                assert memo.mac is not None  # checked by sanitizer
+                memo_coin = coininfo.by_name(memo.coin_name)
```
matejcik, line 56: This still only works for Bitcoin-like coins. I don't think we have, or even _can_ have, a general method of fetching the coin symbol given the SLIP44 identifier. In particular, such method would fail for ERC20 tokens.

ISTM we could simply include the symbol in the payment request? We have no way to identify the amount anyway, so the only thing the use of SLIP44 serves (if i understand right) is cross-verification that the address matches the coin -- but, as pointed out, this can't really work.

Not sure about the SLIP-24 text and the address MAC calculation either. Adding SLIP-44 is really weak.
Perhaps adding the full derivation path instead?

core/src/apps/bitcoin/sign_tx/layout.py
```diff
@@ -84,6 +96,41 @@ async def confirm_decred_sstx_submission(
...
+            memo_coin = coininfo.by_name(memo.coin_name)
+            memo_texts.append(
+                "Buying "
+                + format_coin_amount(memo.amount, memo_coin, amount_unit)
+                + "."
+            )
+        else:
+            raise wire.DataError("Invalid memo type in payment request memo.")
+
+    if msg.amount is None:
```
matejcik, line 121: i'd prefer this to be validated in `BasicApprover.add_payment_request()`, and only asserted here

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -518,6 +520,31 @@ async def confirm_output(
...
+    text.normal(amount + " to")
+    text.normal(recipient_name)
+    text.br_half()
+    for memo in memos:
+        text.normal(memo)
+
+    result = await raise_if_cancelled(
+        interact(
+            ctx,
+            InfoConfirm(text, info="Details"),
```
matejcik, line 539: what is supposed to happen if you click Details?

also given that there can be multiple memos, the screen should probably be paginated right from the start

andrewkozlik: > what is supposed to happen if you click Details?

It results in `show_payment_req_details` being set to `true`, which means each output in the payment request is confirmed individually, see https://github.com/trezor/trezor-firmware/pull/2069/files#diff-5d7c5fcd1f383543bfa486901d66e95191cb54deb21fbd3449a4489dffddf9c9R190-R193.

> also given that there can be multiple memos, the screen should probably be paginated right from the start

You mean using `paginate_paragraphs()`?

core/src/apps/bitcoin/sign_tx/layout.py
andrewkozlik: The rationale behind this is that if someone adds a new memo type and forgets to implement it in the approver or more specifically in `PaymentRequestVerifier`, then the signature verification will fail, so there is no need to check it there. However, if they forget to implement it in the layouts, then the memo won't get shown to the user and the signing process will succeed. The assertion will only uncover it if the implementer also adds a device test. It would have to be an `ensure()` rather than `assert`.

core/src/apps/bitcoin/keychain.py
```diff
@@ -43,10 +43,15 @@
         | VerifyMessage
     )
 
-    class MsgWithAddressScriptType(Protocol):
+    class _MsgWithAddressScriptType(Protocol):
         address_n: Bip32Path
         script_type: InputScriptType
 
+    class _MsgWithAddressOptScriptType(Protocol):
```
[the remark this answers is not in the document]
andrewkozlik: Seems like a leftover from rebasing. https://github.com/trezor/trezor-firmware/pull/2069/commits/6d640283b1265f004fda956f21368d6cfe96d067

common/protob/messages-management.proto
```diff
@@ -421,3 +421,22 @@ message CancelAuthorization {
...
+ */
+message GetNonce {
+    option (unstable) = true;
+}
+
+/**
+ * Response: Contains a random nonce
+ * @end
+ */
+message Nonce{
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/af2bcf654694efba8e38fd74da0a656572440fe2

common/protob/messages-bitcoin.proto
```diff
@@ -414,6 +419,32 @@ message PrevOutput {
     optional uint32 decred_script_version = 3;      // only for Decred
 }
 
+/** Data type of a payment request for a set of outputs.
+ * @embed
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/af2bcf654694efba8e38fd74da0a656572440fe2

core/src/apps/common/address_mac.py
```diff
@@ -0,0 +1,32 @@
...
+
+def check_address_mac(
+    address: bytes, mac: bytes, coin: CoinInfo, keychain: Keychain
+) -> None:
+    expected_mac = get_address_mac(address, coin, keychain)
+    if len(mac) != hashlib.sha256.digest_size or not utils.consteq(expected_mac, mac):
+        raise wire.DataError("Invalid address MAC.")
+
+
+def get_address_mac(address: bytes, coin: CoinInfo, keychain: Keychain) -> bytes:
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/1b42dd3abf091de4a0a81071e4cca24a3a239754 (note this will cause device tests to fail until further fixes are added)

python/src/trezorlib/btc.py
```diff
@@ -250,6 +255,7 @@ def sign_tx(
     outputs: Sequence[messages.TxOutputType],
     details: Optional[messages.SignTx] = None,
     prev_txes: Optional["TxCacheType"] = None,
+    payment_reqs: List[messages.TxAckPaymentRequest] = [],
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/611741250deb48a5fbbe0d398fff6fb020a73b2f

python/src/trezorlib/messages/GetNonce.py
```diff
@@ -0,0 +1,15 @@
+# Automatically generated by pb2py
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/c7c1ae2211934ba2465a2e022ce7ded8206392ca

tests/device_tests/bitcoin/payment_req.py
```diff
@@ -0,0 +1,84 @@
+from collections import namedtuple
+from hashlib import sha256
+
+from ecdsa import SECP256k1, SigningKey
+
+from trezorlib import btc, messages
+
+TextMemo = namedtuple("TextMemo", ["text"])
+CoinPurchaseMemo = namedtuple(
+    "CoinPurchaseMemo", ["amount", "coin_name", "slip44", "address_n"]
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/ba88565f81b8eedd1d055e722649e6d2e2486537

tests/device_tests/bitcoin/payment_req.py
```diff
@@ -0,0 +1,84 @@
...
+
+def make_payment_request(
+    client, recipient_name, outputs, change_addresses=None, memos=None, nonce=None
+):
+    slip44 = 1  # Testnet
+
+    h_pr = sha256(b"SL\x00\x24")
+
+    if nonce:
+        h_pr.update(bytes([len(nonce)]))
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/b34bbd6770fb0f0fe6fcc51eb93185e15d922cdc

common/protob/messages-bitcoin.proto
```diff
@@ -414,6 +419,32 @@ message PrevOutput {
...
+        UTF8_TEXT = 0;
+        COIN_PURCHASE = 1;
+    }
+
+    message PaymentRequestMemo {
+        required MemoType type = 1;    // the type of data in the data field
+        required bytes data = 2;       // encoded memo data, as included in the paymentRequestDigest
+        optional string coin_name = 3; // if the data is an address, the coin it belongs to
+        optional uint64 amount = 4;    // if the data is an address, the amount it will receive
+        optional bytes mac = 5;        // if the data is an address, the MAC returned by GetAddress
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/1c5fb4284e8a34e69362c92e714572bd7b9f1f2b (functionality not implemented yet, please take a look first)

core/src/apps/bitcoin/sign_tx/payment_request.py
andrewkozlik: The SLIP44 coin type serves to verify that the address matches the coin, e.g. ETH, not necessarily the token. As discussed, we will specify the token/symbol using a string. I think we should specify the amount itself as a string, because we might have a problem determining the number of decimal places the coin/token uses. See https://github.com/trezor/trezor-firmware/pull/2069/commits/1c5fb4284e8a34e69362c92e714572bd7b9f1f2b#diff-4568999ce3d84d878396d98f640d2cab8ee735bc568324e7e387ef4254c840a5R448-R452.

core/src/apps/bitcoin/sign_tx/layout.py
matejcik: i specifically mean `msg.amount`, which the protobuf description claims is required for non-coinjoin

common/protob/messages-bitcoin.proto
```diff
@@ -414,6 +419,40 @@ message PrevOutput {
...
+        optional RefundMemo refund_memo = 2;
+        optional CoinPurchaseMemo coin_purchase_memo = 3;
+    }
+
+    message TextMemo {
+        required string text = 1;      // plain-text note explaining the purpose of the payment request
+    }
+
+    message RefundMemo {
+        required string address = 1;   // the address where the payment should be refunded if necessary
```
matejcik, line 445: don't we also need the MAC (and slip44 maybe) here?

common/protob/messages-bitcoin.proto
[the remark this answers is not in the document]
matejcik: the shape of it LGTM, details in subsequent comments

core/src/trezor/ui/layouts/tt/__init__.py
matejcik: Oh, I understand now. This is confusing because the function is called `confirm_...` so one would assume that the boolean result is confirmed yes/no.

Please either change the function to return `result` (and then `show_payment_req_details = result is INFO`), or rename to `should_show_details` or something, like `should_show_more` below.

> You mean using `paginate_paragraphs()`?

yes

common/protob/messages-bitcoin.proto
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/commits/a1663be359d0e7b7a628f8653780d57f82f5ca37

core/src/trezor/ui/layouts/tt/__init__.py
andrewkozlik: > Please either change the function to return `result` [...]

https://github.com/trezor/trezor-firmware/pull/2069/commits/741087054d607bec797ce465dab1716e2d0e783a

> also given that there can be multiple memos, the screen should probably be paginated right from the start

https://github.com/trezor/trezor-firmware/pull/2069/commits/21962b8fae72ac58017c14cb2d82c828006be4ea

common/protob/messages-bitcoin.proto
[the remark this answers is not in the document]
andrewkozlik: functionality implemented in https://github.com/trezor/trezor-firmware/pull/2069/commits/94cef9693a8a69b1753eeae69e894f6bd52b9833

core/src/apps/bitcoin/sign_tx/layout.py
andrewkozlik: Done as part of https://github.com/trezor/trezor-firmware/pull/2069/commits/94cef9693a8a69b1753eeae69e894f6bd52b9833

andrewkozlik:
I also implemented refund memos as part of https://github.com/trezor/trezor-firmware/commit/94cef9693a8a69b1753eeae69e894f6bd52b9833 and fixed the device tests and added device tests for refund memos in https://github.com/trezor/trezor-firmware/pull/2069/commits/9b93875c8c75b4042bea8473470b6eb6d0091ad2.

matejcik (commented):
LGTM now. Please squash&rebase on master, I want to check the UI diff after the "good test vectors" changes are factored in.

core/src/apps/bitcoin/sign_tx/payment_request.py
```diff
@@ -0,0 +1,97 @@
...
+        writers.write_bytes_prefixed(self.h_pr, msg.recipient_name.encode())
+        writers.write_compact_size(self.h_pr, len(msg.memos))
+        for m in msg.memos:
+            if m.text_memo is not None:
+                memo = m.text_memo
+                writers.write_uint32(self.h_pr, _MEMO_TYPE_TEXT)
+                writers.write_bytes_prefixed(self.h_pr, memo.text.encode())
+            elif m.refund_memo is not None:
+                memo = m.refund_memo
+                check_address_mac(memo.address, memo.mac, coin.slip44, keychain)
```
matejcik, line 61: one very small nit: either here or in the protobuf definition we should mention in a comment that this is using the SLIP44 ID of the same coin for which the payment request applies -- unlike the purchase memo, where the SLIP44 may (and usually will) differ

andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2069/files#diff-df31c4d0ad39cb3c36fb3d14a52b556c582b45fb1f5d199a49ce47c6304563e1R61

matejcik (approved):
:100:

PR #2070, opened 2022-01-19
chore(core/embed): check for negative values in trezor_get_uint

prusnak:
Followup to https://github.com/trezor/trezor-firmware/pull/1952

Let's see if the CI goes through with this added extra checks

### DIFF
--- a/core/embed/extmod/modtrezorio/modtrezorio-poll.h
+++ b/core/embed/extmod/modtrezorio/modtrezorio-poll.h
@@ -52,7 +52,15 @@ STATIC mp_obj_t mod_trezorio_poll(mp_obj_t ifaces, mp_obj_t list_ref,
     mp_raise_TypeError("invalid list_ref");
   }
 
-  const mp_uint_t timeout = trezor_obj_get_uint(timeout_ms);
+  // The value `timeout_ms` can be negative in a minority of cases, indicating a
+  // deadline overrun. This is not a problem because we use the `timeout` only
+  // to calculate a `deadline`, and having deadline in the past works fine
+  // (except when it overflows, but the code misbehaves near the overflow
+  // anyway). Instead of bothering to correct the negative value in Python, we
+  // just coerce it to an uint. Deliberately assigning *get_int* to *uint_t*
+  // will give us C's wrapping unsigned overflow behavior, and the `deadline`
+  // result will come out correct.
+  const mp_uint_t timeout = trezor_obj_get_int(timeout_ms);
   const mp_uint_t deadline = mp_hal_ticks_ms() + timeout;
   mp_obj_iter_buf_t iterbuf = {0};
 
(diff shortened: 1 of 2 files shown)


matejcik:
this is fun! the emulator won't even start. let's look into why.

matejcik:
so it turns out that when calculating the wait time for `io.poll`, the result can be (and often is) negative.
```python
delay = utime.ticks_diff(_queue.peektime(), utime.ticks_ms())
```
presumably this means overrun, i.e., `_queue.peektime()` is in the past

this does not matter because the value is used to calculate a `deadline` which is always positive -- and the unsigned overflow wraps around to give us the correct result
https://github.com/trezor/trezor-firmware/blob/trezor_obj_get-negative/core/embed/extmod/modtrezorio/modtrezorio-poll.h#L55-L56
```c
  const mp_uint_t timeout = trezor_obj_get_uint(timeout_ms);
  const mp_uint_t deadline = mp_hal_ticks_ms() + timeout;
```
it seems we can fix this by calling `timeout = trezor_obj_get_int(timeout_ms)`. This will not raise on negative values, gets assigned into an `uint_t` and triggers the same wrapping overflow behavior.
alternately (and perhaps more cleanly), it should be `mp_int_t timeout`, and only add it if it is non-negative. but I don't think this is necessary. the code behaves badly near the uint_t overflow anyway

core/embed/extmod/modtrezorio/modtrezorio-poll.h
```diff
@@ -52,7 +52,7 @@ STATIC mp_obj_t mod_trezorio_poll(mp_obj_t ifaces, mp_obj_t list_ref,
     mp_raise_TypeError("invalid list_ref");
   }
 
-  const mp_uint_t timeout = trezor_obj_get_uint(timeout_ms);
+  const mp_uint_t timeout = trezor_obj_get_int(timeout_ms);
```
prusnak, line 55: If we are going to proceed with this fix I think this really deserves a comment in the source code describing why this is intended.

prusnak:
> it seems we can fix this by calling `timeout = trezor_obj_get_int(timeout_ms)`. This will not raise on negative values, gets assigned into an `uint_t` and triggers the same wrapping overflow behavior.
> alternately (and perhaps more cleanly), it should be `mp_int_t timeout`, and only add it if it is non-negative. but I don't think this is necessary. the code behaves badly near the uint_t overflow anyway

What do you think @jpochyla ?

jpochyla:
> What do you think @jpochyla ?

This will fail for timeouts bigger than 2^31-1, do we care?

matejcik:
> This will fail for timeouts bigger than 2^31-1, do we care?

I don't think we do. The maximum legitimate timeout is 2^29.

jpochyla:
Seems legit then.

PR #1993, opened 2021-12-08
Rust UI: pagination for T1

mmilata:
emu00000000 emu00000001 emu00000002

TODO
- [x] pagination of `Paragraphs`
- [ ] tests

Issue #1122, opened 2020-07-23
Implement new fonts


tsusanka:
For T1 (but in Core!) we want to introduce PixelOperator8 (regular and bold) and for TT we want to introduce Hoves (regular, medium and bold). <s>These fonts will completely replace all the current ones.</s> These fonts will replace the current ones with one exception and that is Roboto Mono - we will keep that for addresses etc.

[fonts.zip](https://github.com/trezor/trezor-firmware/files/4964573/fonts.zip)

prusnak:
What about the monospace variant for TT? The font we currently use (Roboto) has both non-proportional (monospace) and proportional variants. TT Hoves does not seem to have one.

tsusanka:
Very good points.

> What about the monospace variant for TT? Wwe need this for addresses, etc. The font we currently use (Roboto) has both non-proportional (monospace) and proportional variants. TT Hoves does not seem to have one. Will we stick with Roboto for monospace?

I'll clear that up with Petr and let you know.

> Are there any Trezor screen mockups using the new font? I am not sure how TT Hoves is readable on small displays (while Roboto was designed exactly for this purpose).

All screens are on [zeplin](https://app.zeplin.io/projects?pid=5ec4dfb6824fd12f62639331). I have uploaded [this screen](https://app.zeplin.io/project/5ec4dfb6824fd12f62639331/screen/5eea10f3fa85b246e9c7afb5) to TT via  `git checkout toif-upload`
and `tools/toif_upload image.png` and it seems readable just fine _but_ it is true some of the font is quite small. Since this is an important change we should definitely decide first if this is the way we want to go before doing any code changes. I have scheduled a meeting with Ondra (the other designer) to evaluate this and I will bring it up on future Product meeting as well.

tsusanka:
> I'll clear that up with Petr and let you know.

Ok, sorry for the confusion, Petr was inaccurate. We want to add the aforementioned fonts and keep Roboto Mono exactly for addresses, amounts etc. I will edit it above.

prusnak:
Added PixelOperator font (for T1) to Trezor Core in https://github.com/trezor/trezor-firmware/commit/47e75504239a988db7c82a871a2bf0075c2393bb

prusnak:
Trying new fonts for core in the `new-fonts` branch - it currently fails, but one can see the new screenshots in the `ui tests` artefacts.

tsusanka:
Petr's library:

[...]

PR #2071, opened 2022-01-19
Redesigned confirm_action layout for TT

mmilata:
Depends on #1993. Several changes needed to implement basic auto-paginated `confirm_action` layout:

* add TT Hoves fonts (#1122)
* extend `model_tt.component.SwipePage` to display buttons on last page and to hide scrollbar when content fits single page
* add `UI2` build parameter to enable new UI, add `trezor.ui.layouts.tt_v2` as third UI implementation in Core

send
pag0 pag1

* add `bounds()` method to layout objects that draws wireframes for UI debugging

sendw
pagw0 pagw1


core/embed/rust/src/ui/model_tt/component/button.rs
```diff
@@ -238,3 +252,65 @@ pub struct ButtonStyle {
     pub border_radius: u8,
     pub border_width: i32,
 }
+
+pub struct ButtonArray<T> {
```
jpochyla, line 256: I have a feeling that the number of these container types will grow, and in some cases a bit of generalisation would be useful. IIRC i've tried to do this with the tuple components, and the `Map`ping wrapper. What do you think?

core/embed/rust/src/ui/model_tt/component/page.rs
```diff
@@ -44,6 +55,18 @@ where
...
 
+    fn make_content(layout: &PageLayout, content: impl FnOnce(Rect) -> T) -> T {
+        // Check if content fits on single page.
+        let mut content = content(layout.content_single_page);
+        if content.page_count() <= 1 {
+            return content;
+        }
+
+        // Reduce area to make space for scrollbar if it doesn't fit.
+        content.set_area(layout.content);
```
jpochyla, line 66: Very very small nit:
```suggestion
        if content.page_count() > 1 {
            // Reduce area to make space for scrollbar if it doesn't fit.
            content.set_area(layout.content);
        }
```

core/embed/rust/src/ui/model_tt/component/page.rs
```diff
@@ -58,14 +81,26 @@ where
         // paint.
         self.fade = Some(theme::BACKLIGHT_NORMAL);
     }
+
+    fn paint_hint(&mut self) {
+        display::text_center(
+            Point::new(self.pad.area.center().x, self.pad.area.bottom_right().y - 3),
```
jpochyla, line 87: I don't remember now if we have a `Rect::bottom_center()` so the params could use the same source point, but if not, maybe we should!

core/embed/rust/src/ui/model_tt/component/page.rs
```diff
@@ -179,18 +245,60 @@ impl Component for ScrollBar {
...
-            display::rounded_rect(
-                Rect::from_center_and_size(dot, Self::DOT_SIZE),
-                if i == self.active_page {
-                    theme::FG
-                } else {
-                    theme::GREY_LIGHT
-                },
+        if self.has_previous_page() {
+            display::icon(
+                dot - Offset::new(0, Self::ARROW_SPACE),
```
jpochyla, line 250: Maybe we should add `Offset::x()` and `Offset::y()` constructors for these cases?

core/embed/rust/src/ui/model_tt/component/page.rs
```diff
@@ -179,18 +245,60 @@ impl Component for ScrollBar {
...
+    pub buttons: Rect,
+}
+
+impl PageLayout {
+    const BUTTON_SPACE: i32 = 6;
+    const SCROLLBAR_WIDTH: i32 = 10;
+    const SCROLLBAR_SPACE: i32 = 10;
+
+    pub fn new(area: Rect) -> Self {
+        let (content, buttons) = area.hsplit(-Button::<&str>::HEIGHT);
```
jpochyla, line 289: I'd personally opt for `hor_split` and `vert_split`, I sometimes get lost in the `v`/`h` in these calculations. Or, even better, left/right + top/bottom instead of positive/negative values,  because we usually want one or the other. Just a personal preference, feel free to ignore :)

jpochyla (commented):
Adding some ideas.

core/embed/rust/src/ui/model_tt/component/button.rs
```diff
@@ -11,15 +11,18 @@ pub enum ButtonMsg {
...
+    content: ButtonContent<T>,
     styles: ButtonStyleSheet,
     state: State,
 }
 
-impl Button {
-    pub fn new(area: Rect, content: ButtonContent) -> Self {
+impl<T> Button<T> {
+    pub const HEIGHT: i32 = 38;
+    pub const BASELINE_OFFSET: i32 = -4;
```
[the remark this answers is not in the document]
jpochyla: I.e. "Positive values move the button text downward, negative upward" or something.

core/embed/rust/src/ui/model_tt/theme.rs
```diff
@@ -119,6 +149,24 @@ impl DefaultTextTheme for TTDefaultText {
...
+pub const CONTENT_BORDER: i32 = 5;
+
+/// +----------+
+/// |    13    |
+/// |  +----+  |
+/// |10|    | 5|
+/// |  +----+  |
+/// |    14    |
+/// +----------+
+pub fn borders() -> Rect {
```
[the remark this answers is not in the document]
mmilata: Added in e0abc6c.

core/embed/rust/src/ui/component/label.rs
```diff
@@ -24,7 +24,7 @@ where
 {
     pub fn new(origin: Point, align: Alignment, text: T, style: LabelStyle) -> Self {
         let width = display::text_width(&text, style.font);
```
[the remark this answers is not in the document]
mmilata: We actually have it so I deleted this function in e0abc6c.

core/embed/rust/src/ui/display.rs
```diff
@@ -53,6 +53,13 @@ pub fn rect(r: Rect, fg_color: Color) {
     display::bar(r.x0, r.y0, r.width(), r.height(), fg_color.into());
 }
 
+pub fn hollow_rect(r: Rect, fg_color: Color) {
```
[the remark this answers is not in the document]
mmilata: Good idea, fixed in e0abc6c.

core/embed/rust/src/ui/model_tt/component/page.rs
mmilata: fixed in e0abc6c.

core/embed/rust/src/ui/model_tt/component/page.rs
mmilata: Definitely! Fixed in e0abc6c.

core/embed/rust/src/ui/model_tt/component/page.rs
mmilata: Yeah, I've also had to pause and think about which is which. Should be fixed in e0abc6c, please take a look.

core/embed/rust/src/ui/model_tt/component/page.rs
mmilata: D'oh ... f4dc04a

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -102,6 +120,23 @@ where
...
+                &paragraph.content.as_ref()[char_offset..],
+                &mut paragraph.layout.initial_cursor(),
+                &mut TextNoOp,
+            );
+            if matches!(fit, LayoutFit::OutOfBounds { .. }) {
+                break;
+            }
+            char_offset = 0;
+        }
+    }
```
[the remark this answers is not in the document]
mmilata: In 68a7be9 I refactored paragraph handling and the function is a bit shorter now.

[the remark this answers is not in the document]
mmilata: and faeb4dc

core/embed/rust/src/ui/model_tt/component/button.rs
mmilata: That's a valid concern. I didn't understand what `Map` is for but now I think I do - please take a look at c345d66 which attempts to implement button array in terms of tuples and `Map`.

core/embed/rust/src/ui/model_tt/component/button.rs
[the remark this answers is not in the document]
mmilata: Thanks, fixed in bc7cc34.

core/embed/rust/src/ui/model_tt/component/button.rs
```diff
@@ -175,8 +181,9 @@ impl Component for Button {
 
         match &self.content {
             ButtonContent::Text(text) => {
+                let text = text.as_ref();
                 let width = display::text_width(text, style.font);
-                let height = display::text_height();
+                let height = style.font.text_height() + Self::BASELINE_OFFSET;
```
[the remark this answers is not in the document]
mmilata: Fixed in bc7cc34.

core/embed/rust/src/trace.rs
```diff
@@ -12,6 +14,7 @@ pub trait Tracer {
 /// Value that can describe own structure and data using the `Tracer` interface.
 pub trait Trace {
     fn trace(&self, d: &mut dyn Tracer);
+    fn bounds(&self, _sink: &dyn Fn(Rect)) {}
```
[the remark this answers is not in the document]
mmilata: Fair enough, self-contained tracer makes sense. I've moved the method to trait `Component` in fa5602a.

mmilata:
Rebased to master & squashed bunch of commits.

mmilata:
@prusnak can you maybe take a look at cbf0c48114a656b38020b52555bccc6b101361d9 if you have time? I took it from your branch and changed it not to delete the old fonts.

prusnak:
cbf0c48114a656b38020b52555bccc6b101361d9 LGTM 👍

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -2,20 +2,34 @@ use core::convert::{TryFrom, TryInto};
...
     util,
 };
 
 use super::{
-    component::{ButtonMsg, DialogMsg, HoldToConfirm, HoldToConfirmMsg},
+    component::{Button, ButtonMsg, DialogMsg, HoldToConfirm, HoldToConfirmMsg, SwipePage, Title},
     theme,
 };
 
+impl<T> TryFrom<PageMsg<T, bool>> for Obj {
```
matejcik, line 21: vscode is notifying me of "conflicting implementation" when t1 and tt are both enabled -- which indicates that this might be better moved to common code?

core/embed/rust/src/ui/display.rs
```diff
@@ -49,11 +49,18 @@ pub fn fade_backlight(target: i32) {
...
 
-pub fn rounded_rect(r: Rect, fg_color: Color, bg_color: Color, radius: u8) {
+pub fn rect_stroke(r: Rect, fg_color: Color) {
+    display::bar(r.x0, r.y0, r.width(), 1, fg_color.into());
+    display::bar(r.x0, r.y0 + r.height() - 1, r.width(), 1, fg_color.into());
+    display::bar(r.x0, r.y0, 1, r.height(), fg_color.into());
+    display::bar(r.x0 + r.width() - 1, r.y0, 1, r.height(), fg_color.into());
+}
+
+pub fn rect_fill_rounded(r: Rect, fg_color: Color, bg_color: Color, radius: u8) {
```
matejcik, line 63: FWIW we should check somewhere that `radius` is 2, 4, 8, or 16. That's what the C code does, but the check is silent. An `assert!` in `display::bar_radius` probably?

core/mocks/generated/trezorui2.pyi
```diff
@@ -2,21 +2,21 @@ from typing import *
 
 
 # extmod/rustmods/modtrezorui2.c
-def layout_new_example(text: str) -> None:
+def layout_new_confirm_action(
+    *,
+    title: str,
```
matejcik, line 7: the `*`s got lost in rebase

core/embed/rust/src/ui/model_tt/component/title.rs
```diff
@@ -0,0 +1,80 @@
+use super::theme;
+use crate::ui::{
+    component::{Child, Component, ComponentExt, Event, EventCtx},
+    display,
+    geometry::{Insets, Rect},
+};
+
+pub struct Title<T, U> {
```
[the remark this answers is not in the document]
mmilata: Will rename it to `Frame`. Not sure if I understand correctly but if you don't need a title you can just use the thing directly without a wrapper.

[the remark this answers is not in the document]
matejcik: i mean more like, "Frame" is a default top-level component for every "layout" (where "layout" is a screen as designed in Figma). But in case we have layouts that don't have title, we might need a different top-level component in case just sticking, e.g., a ButtonPage, is not appropriate. (consider a pop-up that is a grey bubble on black background)

core/embed/rust/src/ui/model_tt/component/passphrase.rs
```diff
@@ -23,9 +23,9 @@ pub enum PassphraseKeyboardMsg {
 pub struct PassphraseKeyboard {
     page_swipe: Swipe,
     textbox: Child<TextBox>,
-    back_btn: Child<Button>,
-    confirm_btn: Child<Button>,
-    key_btns: [[Child<Button>; KEYS]; PAGES],
+    back_btn: Child<Button<&'static str>>,
+    confirm_btn: Child<Button<&'static str>>,
+    key_btns: [[Child<Button<&'static [u8]>>; KEYS]; PAGES],
```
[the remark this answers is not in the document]
mmilata: We can change it to `str` but then we'd have to do `Vec<char>`*->`&[u8]` conversion before passing the phrase into `display::text()`. Since we don't support non-ASCII passphrases, using `&[u8]` directly seems like path of least resistance?

_*_ basically UTF-32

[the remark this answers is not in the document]
matejcik: i did a weird thing in 8670ad8f391d8f6c2bdca2b58138b22c9230cddc where `TextBox` is still `Vec<u8>` which internally guarantees that it's ASCII only, and easily converts ` -> &[u8] -> &str` for calling the modified `display::text()`

let's leave this as-is until @jpochyla has a chance to comment

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: Wanted to solve this as part of "sane rs->py message passing" but this one is actually semi-sane, so fixed in 0292fbfee125d7ed2f201c52447e074c9a1bb2ad.

core/embed/rust/src/ui/display.rs
mmilata: Added in be6ecd2b9848979c589a79c28b43d0ced9ff376a.

core/embed/rust/src/ui/model_tt/component/title.rs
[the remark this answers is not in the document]
mmilata: Fixed in 32e5653c2e53c24f0979d36cc39d3d30f5f6fa82 and be6ecd2b9848979c589a79c28b43d0ced9ff376a.

Issue #1018, opened 2020-05-26
Good test vectors for signed transactions


matejcik:
Many transaction test vectors in our testsuite aren't actually valid transactions: they are spending outputs that do not exist, have mismatched amounts, or belong to different addresses.

In general, the majority of transaction test vectors only validate our implementation in core vs legacy, and check that no regressions in the signing code happened. As there is currently no good way to validate transactions against the real cryptocurrency implementation, there could be invisible serialization bugs and such.

The goal of this issue is to improve the test suite in this regard.

The Platonic ideal of a test vector is one that you can copy-paste from the test suite and spend on-chain directly. This is of course impossible.

The next best thing is validating that a Trezor-generated transaction comes out identical to a transaction that has already been accepted by the network. We have several such testcases, but we've lost many of them because, e.g., they are using invalid BIP32 paths.
It is rather difficult to _generate_ such transactions as needed, because one needs to create live transaction to serve as inputs, a live transaction that spends them, that costs money and time and some amount of bookkeeping.

As an intermediate goal, it would be good to have a "full feature set" of transactions that have been accepted by Testnet (or the respective testnets of different altcoins). This removes the need to spend real money and limits the amount of necessary bookkeeping to "get coins from faucet, return coins to faucet".

Another option is to run a blockchain internally.

Finally, we can explore ways to take advantage of [`testmempoolaccept`](https://bitcoincore.org/en/doc/0.17.0/rpc/rawtransactions/testmempoolaccept/) (and equivalents?), possibly coupled with reverting the node to a block at particular height, so that the UTXOs being used are unspent at that time. This would allow us to independently check the validity of the test vectors.

tsusanka:
Please note that we now have Bitcoin regtest in trezor-user-env (https://github.com/trezor/trezor-user-env/issues/99) and also a first test using this in Suite 🎉  (https://github.com/trezor/trezor-suite/pull/4540). But also note that @karliatto was responsible for making the first test (in another words for the infrastructure) and I would like the QA to expand the tests furthermore after our new Head of QA starts in ~February.

matejcik:
some notes:

(a) **Using regtest is impractical.**  
* Everything we actually need to test, we can test with static vectors. Static vectors are also easier to work with.
* Each developer would need a regtest instance, a rather heavy dependency.
* Not mentioning that this is not scalably applicable for altcoins.
* Suite will already have end-to-end tests with regtest, so big-picture we got that part covered.

[...]

PR #2075, opened 2022-01-21
Better test vectors for transaction signing - BTC legacy addresses

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1018

Trying to use real transactions from mainnet/testnet wherever possible (from our `all all all...` seed), so that the `script_pubkey` check is not failing.

So far only `device_tests/bicoin/test_signtx.py` was modified, as it was the easiest to start with.
These tests were changed to use real blockchain transactions (the transactions already existed or it was viable to create them on testnet):
```
test_not_enough_vouts
test_two_two
test_one_one_fee
test_one_three_fee
test_attack_change_outputs
test_p2sh
test_lots_of_outputs
test_lots_of_inputs
test_lots_of_change
test_lock_time[1-4294967295]
test_lock_time_blockheight
test_lock_time[500000000-4294967294]
test_lock_time_datetime[1985-11-05 00:53:20]
test_lock_time[499999999-4294967294]
test_lock_time_datetime[2048-08-16 22:14:00]
```

These are still missing, as it is rather hard to create these transactions
```
test_testnet_big_amount - needs at least 43 TEST to overflow uint32
test_spend_coinbase - we hardly have any coinbase transactions
test_testnet_fee_high_warning - would need at least 0.025 TEST fee
test_fee_high_hardfail - would need at least 0.25 TEST fee
test_fee_high_warning -  similar as above
```

In the process so far, at least two hopefully useful scripts were created:
- `/tools/unify_test_files.py` for improving the formatting and readability of the test files - already used on `device_tests/bicoin/test_signtx.py` - could be used for all the testing files
- `python/tools/sign_tx.py` for quick transaction creating on mainnet/testnet, inspired by the testing code - allows for creating arbitrarily big transactions easily, was used to generate all the testnet transactions for the tests

Next steps are looking at `segwit`, `multisig` and all the BTC-forks, which will probably not be so easy


tests/device_tests/bitcoin/test_signtx.py
```diff
@@ -885,7 +902,9 @@ def attack_processor(msg):
                 run_attack = True
             else:
                 msg.tx.outputs[0].address_n = []
-                msg.tx.outputs[0].address = "mwue7mokpBRAsJtHqEMcRPanYBmsSmYKvY"
+                msg.tx.outputs[
+                    0
+                ].address = "mwue7mokpBRAsJtHqEMcRPanYBmsSmYKvY"  # 44h/1h/4h/0/2
```
matejcik, line 907: please move the comment one line up so that this bad formatting does not happen

grdddj: fixed in 3d507a0

grdddj:
To the current `tx_matches()` function, where we are asserting `True` return value:
- when it fails, `pytest` does not show which of the checks failed, because we are asserting the return value, and it can return `False` on numerous occasions
- it _could_ be better to have `assert_tx_matches()` function which we would call from the test cases, and the assertions would be inside of that function - we would then see which _exact_ assertion got broken

matejcik:
> * it _could_ be better to have `assert_tx_matches()` function which we would call from the test cases, and the assertions would be inside of that function - we would then see which _exact_ assertion got broken

that is probably better

matejcik:
mostly LGTM

is assertion rewriting enabled for `tests/common.py`? i.e., did you check that when the assertion fails, it looks right in pytest output?

please also and add `assert_tx_matches` to the other test cases

please add documentation about the `CHECK_BLOCKBOOK_HEX` (maybe rename to something like ... uhhh ... `CHECK_ON_CHAIN` ?) to the appropriate place. we do have some documentation for device tests.

let's start the one-off script directory under `tools/snippets/`, add your scripts and add a README with one- or two-sentence description of what the script does

please change the git history so that:
* adding type hints to `signtx.py` and `tx_cache.py` is a separate commit
* adding `assert_tx_matches` to common is a separate commit
* all modifications to `test_signtx` and related additions/removals of the cache entries are a single commit

matejcik:
changelog check is failing but that will most likely go away after the rebase

matejcik:
oh! and please rebase this on top of `master`, _without_ Andrew's commits, so that we can merge this independently

grdddj:
> please also and add assert_tx_matches to the other test cases

You mean other test cases in `test_signtx.py`, or all other testcases in all files for transaction signing (I expect this to be done gradually, file by file, when fixing the pubkey problems)?

The issue sometimes is that the transaction does not exist in reality, so we cannot provide existing link to `Blockbook` (so in those cases I am using simple inline `assert`). When the transaction does not exist, we could theoretically just supply the tx_hash, and recognize it is not a link and not do the online validation.

grdddj:
> is assertion rewriting enabled for tests/common.py? i.e., did you check that when the assertion fails, it looks right in pytest output?

Thanks for pointing this out, I did not check previously, and when I did now, it only showed the assert line, not the actual values.

Fixed with `pytest.register_assert_rewrite("tests.common")` - will include it into one commit with the `common` changes

matejcik:
> You mean other test cases in `test_signtx.py`

only the other non-failing test cases in `test_signtx.py`

matejcik:
> the transaction does not exist in reality

In that case, please indicate in a comment that the resulting tx does not exist.
Which testcases are those, btw?

grdddj:
> Which testcases are those, btw?

Those where we are mainly testing the Trezor behaviour:
```
test_testnet_fee_high_warning
test_fee_high_warning
test_fee_high_hardfail
test_attack_change_input_address
test_spend_coinbase
```

grdddj:
Hopefully all the points above addressed in the force-push, the base PR branch also changed to `master`.

In the new `tools/snippets` folder, I have included also the path-to-address translation file for `all all...` seed, it could be potentially useful in the future as well

tests/device_tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+import pytest
+
+pytest.register_assert_rewrite("tests.common")
```
matejcik, line 3: could we just do this in `conftest.py` ?

grdddj: Docs say we could, but I saw we have the same thing in `tests/click_tests/__init__.py`, so I put it into `__init__` here as well

tools/snippets/address_cache_all_all_seed.json
```diff
@@ -0,0 +1,3002 @@
+{
```
matejcik: this is really not a good place for this data.

also fwiw you could have gotten this data from https://iancoleman.io/bip39/

tests/device_tests/__init__.py
matejcik: yeah but click_tests register a file in the click_tests subfolder, i.e., "their own"

matejcik:
please re-record UI fixtures

tests/device_tests/__init__.py
grdddj: Sure, so I will put it into `conftest.py`, as it is common for all tests - literally

grdddj: Moved in d2cee34

tools/snippets/address_cache_all_all_seed.json
grdddj: Deleted in 2816b51. True that if somebody wants this, they can generate it

grdddj:
> please re-record UI fixtures

Done in 4595828

tests/device_tests/bitcoin/test_signtx.py
```diff
@@ -392,17 +386,18 @@ def test_two_two(client):
                 request_output(0),
                 messages.ButtonRequest(code=B.ConfirmOutput),
                 request_output(1),
+                messages.ButtonRequest(code=B.ConfirmOutput),
```
matejcik, line 389: This is always a problem. Expected responses **must** stay identical except for txhash references.

The problem here is that your "change" output is not change -- it's from a different account. In this case you can fix it by switching the first output to change and the second to address.

tests/device_tests/bitcoin/test_signtx.py
```diff
@@ -322,12 +318,12 @@ def test_one_three_fee(client):
                 request_output(1),
                 messages.ButtonRequest(code=B.ConfirmOutput),
                 request_output(2),
+                messages.ButtonRequest(code=B.ConfirmOutput),
```
matejcik, line 321: This doesn't look fixable with this data: all three outputs go out to another account. Here we really want two normal and one change output.

tests/device_tests/bitcoin/test_signtx.py
matejcik: Correction: expected responses can of course change if the prevtx changes, or if you deliberately change the number of outputs/inputs, or similar.

But the number of confirmation _should_ stay the same, except in maybe some exceptional cases.

tests/device_tests/bitcoin/test_signtx.py
```diff
@@ -738,77 +705,81 @@ def test_p2sh(client):
...
+
+    assert_tx_matches(
+        serialized_tx,
+        hash_link="https://tbtc1.trezor.io/api/tx/890155c89905c86fb087eddbe9973fa845c99913836e872f41fbdfa3864b607d",
+        tx_hex="0100000001c03f07debba2c747751382873c8dcc2c41688e0cef1b5c63c2b49d9370004b07010000006b483045022100a2c79eaed632746fd514aa09eae51c0294bdf66c74619306e0273cd1a470c9e7022050c78dec6acd65b42150dba70e2b546dfd737d77f46975a97427999cb2b8280401210344e14b3da8f5fe77a5465d0f8fe089d64ed5517d1f1f989edd00f530938a2c22ffffffff01f065380c0100000017a91487dba64df7e9386d0b0f3ef557269833e12d1b7a8700000000",
     )
 
 
-@pytest.mark.setup_client(mnemonic=MNEMONIC12)
 def test_attack_change_outputs(client):
```
matejcik, line 756: this test should really fail when it doesn't trigger change detection.

please add `expected_responses` here, before the attack processor is installed.

i believe you can use the same data as in `test_two_two`, and the same expected_responses?

tests/device_tests/bitcoin/test_signtx.py
grdddj: Addressed in eaaa13742bdbe50f06a8452d8f6a93a52365a211 - created a new transaction to comply with this requirement

tests/device_tests/bitcoin/test_signtx.py
grdddj: Changed in eaaa13742bdbe50f06a8452d8f6a93a52365a211 - so that it recognizes change address

tests/device_tests/bitcoin/test_signtx.py
grdddj: Addressed in eaaa13742bdbe50f06a8452d8f6a93a52365a211 - taking the `test_two_two` details was alright

matejcik:
lgtm, please rebase as soon as CI passes

grdddj:
> lgtm, please rebase as soon as CI passes

Interestingly, device-tests failed, but just because of the Shamir tests - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2022797566

In master it should be fixed, so I am rebasing, squashing and force-pushing

Issue #2076, opened 2022-01-21
Check taproot script pubkeys


onvej-sl:
A taproot script pubkey has the following format (see [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#script-validation-rules)):
```
0x51 || 0x20 || output_key
```
Technically, the first byte (`0x51`) is the opcode for pushing `0x01` onto the stack, `0x01` is interpreted as the witness version. The second byte (`0x20`) is the opcode of a 32-byte push, `output_key` is a public key encoded as 32 bytes according to [BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#specification).

Half of 32-byte strings doesn't encode a (valid) public key, which means that the corresponding script pubkey cannot be spent.

The question is whether we want to check that the script pubkeys in transaction outputs are valid.

PR #2077, opened 2022-01-21
Check taproot script pubkeys

prusnak:
Fixes https://github.com/trezor/trezor-firmware/issues/2076


core/src/apps/bitcoin/scripts.py
```diff
@@ -238,6 +239,10 @@ def output_script_native_segwit(witver: int, witprog: bytes) -> bytearray:
     length = len(witprog)
     utils.ensure((length == 20 and witver == 0) or length == 32)
 
+    # check that P2TR address encodes a valid BIP340 public key
+    if witver == 1:
+        utils.ensure(bip340.verify_publickey(witprog))
```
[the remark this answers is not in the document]
matejcik: but maybe not here? istm this is not the right place for the check anyway, it should come at the address decoding step imo

onvej-sl (approved):
`crypto/zkp_bip340.{c,h}` and ` core/embed/extmod/modtrezorcrypto/modtrezorcrypto-bip340.h` LGTM

core/src/apps/bitcoin/scripts.py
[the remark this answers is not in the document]
andrewkozlik: I agree about raising `wire.DataError`. Technically is seems to make more sense to catch this in `common.decode_bech32_address()` rather than here, but doesn't make any difference at the moment, so I have no strong preference either way.

andrewkozlik (commented):
It might be good to also add a device test.

prusnak:
Rebased and addressed comments in 51847b1715e1418d276c9299640c696bc0ef94b1

legacy/firmware/transaction.c
```diff
@@ -288,6 +288,12 @@ static int address_to_script_pubkey(const CoinInfo *coin, const char *address,
     // push addr_raw (segwit_addr_decode makes sure addr_raw_len is at most 40)
     script_pubkey[0] = witver == 0 ? 0 : 80 + witver;
     script_pubkey[1] = addr_raw_len;
+    // check that P2TR address encodes a valid BIP340 public key
+    if (witver == 1 && addr_raw_len == 32) {
```
andrewkozlik, line 292: I am wondering about the purpose of checking `addr_raw_len == 32` at this point. What happens if the length is not 32 bytes? The public key check won't be performed and the address will pass validation, which is not what we want.

andrewkozlik: If we want to prevent the user from burning funds by sending them to an unrecognized address, then:
- We should be checking the witness program length somewhere. Either in `zkp_bip340_verify_publickey()`, i.e. remove the `addr_raw_len == 32` check above, or in `segwit_addr_decode()`, as we do for v0 witness programs.
- We should also be checking that we recognize the witness version 0 or 1, which we don't seem to be doing in T1

core/src/apps/bitcoin/common.py
```diff
@@ -133,6 +133,9 @@ def decode_bech32_address(prefix: str, address: str) -> tuple[int, bytes]:
         raise wire.ProcessError("Invalid address witness program")
     assert witver is not None
     assert raw is not None
+    # check that P2TR address encodes a valid BIP340 public key
+    if witver == 1 and len(raw) == 32 and not bip340.verify_publickey(bytes(raw)):
```
andrewkozlik, line 137: Same comment as for T1 (https://github.com/trezor/trezor-firmware/pull/2077/files#r792684749). Either we should skip not check `len(raw) == 32` and let `bip340.verify_publickey()` fail when the length is not 32 bytes or we should check that v1 witness programs are 32 bytes in `bech32.decode()`.

andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2077/commits/46113d6b81155452aa4d4acbbba4d1d1ff61ae8b

legacy/firmware/transaction.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2077/commits/ee590674cae5284489ab78aefa7ab2906d564252

andrewkozlik:
After rebasing the `test_codepoints` crypto tests started failing randomly. Does anyone have a clue what this is about? I don't see any obvious connection with the changes in the PR.

```
tests/test_check.c:6293:E:codepoints:test_codepoints_secp256k1:0: (after this point) Test timeout expired
tests/test_check.c:6305:E:codepoints:test_codepoints_nist256p1:0: (after this point) Test timeout expired
```

```
tests/test_check.c:6293:E:codepoints:test_codepoints_secp256k1:0: (after this point) Test timeout expired
tests/test_check.c:6308:P:codepoints:test_codepoints_nist256p1:0: Passed
```

matejcik:
the same failures are in master: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2095256947

matejcik:
could be related to the nixos update b9bbd5fcd78cca7126cf3773d2be8344f36d55e5 ?

andrewkozlik:
QA:
Check that Trezor 1 and Trezor T refuse to sign transactions that have an unrecognized Taproot or SegWit address such as:
`tb1pam775nxmvam4pfpqlm5q06k0y84e3x9w0xuhdpmxuna2qj3dfg6qy2pq29`
`tb1plycg5qvjtrp3qjf5f7zl382j9x6nrjz9n0dh50`
`tb1zlycg5qvjtrp3qjf5f7zl382j9x6nrjz9sdhenvyxq8c3808qxmusxanpxu`
`tb1pllllllllllllllllllllllllllllllllllllllllllllallllscqgl4zhn`

bosomt:
**QA OK**

1.10.6     revision: 20 bytes 0x0e27423cffb3b6047028c61890bec43bf626adc7,
2.4.4     revision: 20 bytes 0x0e27423cffb3b6047028c61890bec43bf626adc7

Tested in Suite app and Electrum app.
4x2x2 times address refused during Send dialogue.
I'm not able to confirm that addresses were checked by apps or devices, they were marked as invalid before i tried to sign transactions.

Screenshot 2022-02-17 at 12 46 36

Screenshot 2022-02-17 at 12 57 32

PR #2081, opened 2022-01-24
Better test vectors - multisig and segwit

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1018 and building on top of https://github.com/trezor/trezor-firmware/pull/2075

Making device tests pass for `multisig` and `segwit` transactions


grdddj:
I am not sure how to handle `test_attack_change_input_address` testcase - it fails on mismatch in expected errors - we expect `Transaction has changed during signing` but we get the new `Input does not match scriptPubKey`.

It seems to me it might be correct here, and we should change the expected error - as the test changes the address we sign the input with, which needs to trigger the `scriptPubKey` issue, as it seems to be checked before the `Transaction has changed`.

If the above holds true, and we still want to test `Transaction has changed` error, we would need to do something different. (Changing the script_type also generated `scriptPubKey` issue, changing the amount `Not enough funds`)

grdddj:
> this test case is testing a scenario that is completely prevented by the scriptPubKey

Looks like the similar situation is in `test_multisig,py::test_attack_change_input`, changed in 1cdec74.

I got rid of the `expected_responses` there, as it did not have any effect on the failed signing - even when I changed them, it did not make a difference. (Looks like the expected responses are evaluated only in successful case)

andrewkozlik:
> Looks like the expected responses are evaluated only in successful case

That seems a bit disturbing. I think the reason why we have the expected responses is to be sure that the test fails in the spot where we expect it to fail and not "accidentally" for another reason, which just happens to produce the same error message.

matejcik:
I'm wondering how to automatically verify this. The obvious problem is that we can't fail the test for two reasons at the same time.

For starters, we can flip the expected responses handler to ignore exceptions coming through it and always raise for bad expected responses. In cases like `with pytest.raises(...), client`, a test might pass because the exception is caught and properly checked by the `pytest.raises` handler. If `client` as the inner handler raised a _different_ exception, we should see a test failure.

grdddj:
> > Looks like the expected responses are evaluated only in successful case
> 
> That seems a bit disturbing. I think the reason why we have the expected responses is to be sure that the test fails in the spot where we expect it to fail and not "accidentally" for another reason, which just happens to produce the same error message.

I was wrong here, now it works for me - `expected_responses` added in e186a1d and when they change, the test fails.

In the "clean" path, I do not have them, as there would be `<ButtonRequestType.UnknownDerivationPath: 15>`, probably saying that the path `"48h/1h/0h/1h/0/0"` is not standard - however, the test should test other things.

andrewkozlik:
The purpose of `test_multisig,py::test_attack_change_input` was to ensure that MatchChecker's `check_input()` function catches the change in the account type. But now that we will be verifying the scriptPubKey I can't formulate a sensible attack scenario where changing the account type would lead to anything. In fact, it almost seems that checking the scriptPubKey makes `check_input()` obsolete. We can keep the test for legacy reasons and maybe mention it in a comment. The goal of the test was to replicate the "Malicious change in mixed transactions" attack in https://blog.trezor.io/details-of-firmware-updates-for-trezor-one-version-1-9-0-and-trezor-model-t-version-2-3-0-46deb141fc09.

andrewkozlik:
The same assessment applies to the `test_attack_change_input_address` tests and the other `test_attack_change_input` tests, except that the goal in these cases was a bit different. They all replicate an attack, where the user is coerced into unknowingly transferring funds from one account to another one of their accounts, resulting in privacy issues rather than coin theft.

grdddj:
Thanks @andrewkozlik  for the comments, I have included this note in 742406e 

As for all the other `test_attack_change_input`, if they now raise `Input does not match scriptPubKey`, I would just change the assert to account for this (when the error does not happen in the "clean" path obviously).

andrewkozlik:
> I have included this note in [742406e](https://github.com/trezor/trezor-firmware/commit/742406ec9e65c445d0211c990321fac4958339f7)

The check wasn't "replaced", you can still create a test which triggers a failure in `check_input()`, but the test scenario will no longer be an attack AFAICT. Maybe we could do that too or instead.

grdddj:
> The check wasn't "replaced",

Yeah, that is a bad wording. Could we say something like "Recently the scriptPubKey verification was introduced and it accounts for this instead"?

> Maybe we could do that too or instead.

There is quite a lot of tests for triggering `Transaction has changed during signing` already in other test files. Not sure if this one would be different, as it is multisig

grdddj:
Current situation with failing scriptPubKey tests:
- 45 errors to go, described below
- some of those seem easy to fix, some are tricky or would require bigger effort
```
tests/device_tests/bitcoin/test_signtx_segwit.py::test_attack_change_input_address
- probably just changing the expected error to the scriptPubKey issue
tests/device_tests/bitcoin/test_signtx_segwit.py::test_testnet_segwit_big_amount
- we would again need 43 TEST for this
tests/device_tests/bitcoin/test_signtx.py::test_spend_coinbase
- we do not have a coinbase transaction

tests/device_tests/bitcoin/test_signtx_external.py::test_p2wsh_external_presigned
tests/device_tests/bitcoin/test_signtx_external.py::test_p2wpkh_in_p2sh_presigned
tests/device_tests/bitcoin/test_signtx_external.py::test_p2wpkh_with_proof
- these could be fixed relatively easily

tests/device_tests/bitcoin/test_authorize_coinjoin.py::test_sign_tx
- not sure how to do this

tests/device_tests/test_protection_levels.py::test_signtx
- could be easy fix

tests/device_tests/bitcoin/test_signtx_invalid_path.py::test_invalid_path_pass_forkid
- Bcash
tests/device_tests/bitcoin/test_signtx_invalid_path.py::test_invalid_path_prompt
- Litecoin
tests/device_tests/bitcoin/test_signtx_invalid_path.py::test_attack_path_segwit
- Testnet, could be done

tests/device_tests/bitcoin/test_signtx_prevhash.py::test_invalid_prev_hash_in_prevtx
- 4 vectors
----- is fine when deleting the first signing - "btc.sign_tx(client, "Bitcoin", [inp0], [out1], prev_txes={tx_hash: prev_tx})"

tests/device_tests/bitcoin/test_nonstandard_paths.py::test_signtx
- 5 vectors, and each is run for three script types

tests/device_tests/bitcoin/test_nonstandard_paths.py::test_signtx_multisig
- 7 vectors

BTC-LIKE ALTCOINS:
tests/device_tests/bitcoin/test_bgold.py
- 9 tests
tests/device_tests/bitcoin/test_zcash.py
- 1 test
tests/device_tests/bitcoin/test_bcash.py
- 3 tests
tests/device_tests/bitcoin/test_decred.py
- 5 tests
```

andrewkozlik:
> There is quite a lot of tests for triggering `Transaction has changed during signing` already in other test files. Not sure if this one would be different, as it is multisig

Actually I think this one in particular is the more interesting one, precisely because it is multisig. It gets triggered by different code logic than the others.

I think the question boils down to whether we want the device tests to check only critical scenarios (normal use cases, attacks and warnings) or whether we want them to go in the direction of full code coverage. When it comes to Bitcoin signing I am more inclined to err on the side of caution.

grdddj:
> whether we want the device tests to check only critical scenarios or whether we want them to go in the direction of full code coverage

As Bitcoin signing is probably the most important thing Trezor does (at least in economical sense), I would also vote for as much coverage as possible

tests/device_tests/bitcoin/test_multisig.py
```diff
@@ -233,24 +231,31 @@ def test_missing_pubkey(client):
...
     In Phases 1 and 2 the attacker replaces a non-multisig input
     `input_real` with a multisig input `input_fake`, which allows the
     attacker to provide a 1-of-2 multisig change address. When `input_real`
     is provided in the signing phase, an error must occur.
+
+    NOTE: previously this test was making sure that MatchChecker's check_input()
+    function catches the change in the account type. Recently this check was replaced
+    by verifying the scriptPubKey.
+    The goal of the test was to replicate the "Malicious change in mixed transactions" attack in
+    https://blog.trezor.io/details-of-firmware-updates-for-trezor-one-version-1-9-0-and-trezor-model-t-version-2-3-0-46deb141fc09
```
andrewkozlik, line 246: I suggest:

The goal of this test is to replicate the "Malicious change in mixed transactions" attack in https://blog.trezor.io/details-of-firmware-updates-for-trezor-one-version-1-9-0-and-trezor-model-t-version-2-3-0-46deb141fc09
Previously this test was making sure that MatchChecker's check_input() function catches the change in the account type. After scriptPubKey verification was introduced, the attack is caught earlier by the scriptPubKey mismatch.

grdddj:
> Actually I think this one in particular is the more interesting one, precisely because it is multisig. It gets triggered by different code logic than the others.

How can it then be replicated in the current situation?

tests/device_tests/bitcoin/test_multisig.py
grdddj: Looks good, I will update it

grdddj:
In the last two commits with nonstandard paths (d198547 and f01fa06), there is a small mismatch in networks, but for our purpose, it seems to work:
- the addresses were generated on `Testnet` with safety checks turned off (as all those paths are valid only on  `Bitcoin`)
- one big transaction was generated on `Testnet`, sending something to those addresses
- when signing these UTXOs, `Bitcoin` parameters are used for the multisig creation and signing
- (we are only interested in "our" first signature, so the tx is incomplete, it cannot be broadcasted, even if our signature was valid - which is not, as it signs a different network)

Is it alright to "cheat" a little bit like this?

grdddj:
After last commit, [cbbb8f3](https://github.com/trezor/trezor-firmware/pull/2081/commits/cbbb8f36cc775c52d94ee7e0a6d3f5cccdc57e99), even the coinjoin should be alright. Thanks @andrewkozlik for consultation, in the end it was the same replacement of `script_pubkey` and `prev_hash` that helped, as everywhere else.

Now there are just two failing unittests, for `Bitcoin` and `Decred`, which need to be "fixed"

grdddj:
As we agreed, @matejcik, I am removing the `segwit_hash` argument from `assert_tx_matches` in [5ddef7f](https://github.com/trezor/trezor-firmware/pull/2081/commits/5ddef7fb29461998bec5c08ede2f1deb15c361d6)

That required a way how to get the `tx_id` from segwit transaction - in the end I compiled a minimum (de)serialization transaction logic with the use of Programming Bitcoin code - as it has some nice features and could be definitely useful elsewhere.

We could build the tx_id calculating logic using the `deserialize_tx.py` script in `python/tools` - if you thing it is better I can do it with it instead - this was just working out of the box.

grdddj:
In connection to the current segwit/nonsegwit logic, we have multiple possibilities how to assert the `tx_id`. I would say there is no big difference between them - if we are once relying on the `Tx.parse()`, then we could trust it with all the validation (option 3):

1 - current:
```python
parsed_tx = Tx.parse(serialized_tx)
if parsed_tx.segwit:
    assert tx_id == parsed_tx.id()
else:
    assert tx_id == tools.tx_hash(serialized_tx).hex()
```

2:
```python
parsed_tx = Tx.parse(serialized_tx)
if parsed_tx.segwit:
    tx_bytes = parsed_tx.serialize_legacy()
else:
    tx_bytes = serialized_tx
assert tx_id == tools.tx_hash(tx_bytes).hex()
```

3:
```python
assert tx_id == Tx.parse(serialized_tx).id()
```

grdddj:
Force-pushed after rebasing on `master`. There seem to be some newly added cases in `tests/device_tests/bitcoin/test_signtx_payreq.py::test_payment_request`, that are failing the `scriptpubkey` check.

I will try to resolve them.

grdddj:
I did not notice previously, but the `HWI` tests are also failing (not just) because of the `Input does not match scriptPubKey`

matejcik:
Finished my pass and tidied up the commit history.

Biggest change is introduction of `forge_prevtx()` function, plus related removal of several fake transactions from the cache.

I also introduced a naming convention: `TXHASH_abcdef` refers to real on-chain transactions, `FAKE_TXHASH_abcdef` to synthetic transactions that we keep in case using `forge_prevtx()` is not possible or not appropriate.

ci/test.yml
```diff
@@ -362,6 +362,7 @@ core hwi test:
   <<: *gitlab_caching
   needs:
     - core unix frozen debug build
+  allow_failure: true
```
grdddj, line 365: There are also `legacy` `HWI` tests which should be allowed to fail

matejcik:
Flipping to draft because this should not be merged as is. We will need to merge the test changes separately from the scriptPubKey checking changes, and possibly implement the safety check override.

matejcik:
squashed fixups and removed the scriptpubkey checks from this PR

ci/test.yml
matejcik: added in rebase

tests/device_tests/bitcoin/test_signtx_segwit.py
```diff
@@ -142,11 +148,19 @@ def test_send_p2sh_change(client: Client):
...
+    address = btc.get_address(
+        client,
+        "Testnet",
+        address_n,
+        script_type=messages.InputScriptType.SPENDP2SHWITNESS,
+    )
+    prev_hash, prev_tx = forge_prevtx([(address, 2 ** 32 + 1)], network="testnet")
 
     inp1 = messages.TxInputType(
         address_n=parse_path("m/49h/1h/0h/0/0"),
```
grdddj, line 161: We could use already-existing `address_n` variable from above. It could be also worth creating `script_type` variable, it is used two times and needs to be the same.

tests/device_tests/bitcoin/signtx.py
```diff
@@ -0,0 +1,136 @@
...
+            amount=amount,
+            script_pubkey=bytes(CBitcoinAddress(address).to_scriptPubKey()),
+        )
+        for address, amount in vouts
+    ]
+    tx = messages.TransactionType(
+        version=1,
+        # inputs_cnt=1,
+        inputs=[input],
+        # outputs_cnt=len(outputs),
```
grdddj, line 118: Are these comments real comments, or leftovers from debugging? (this and two lines above)

tests/device_tests/bitcoin/signtx.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def forge_prevtx(
+    vouts: Sequence[Tuple[str, int]], network: str = "mainnet"
+) -> Tuple[bytes, messages.TransactionType]:
+    """
+    Forge a transaction with the given vouts.
+    """
+    bitcoin.SelectParams(network)
+    input = messages.TxInputType(
```
grdddj, line 101: We are shadowing the built-in `input` function - we could name it `tx_input` - but not a big issue in this context

tests/device_tests/bitcoin/test_signtx_invalid_path.py
```diff
@@ -70,25 +55,24 @@ def test_invalid_path_fail(client: Client):
...
-# only changed the coin from Bitcoin to Litecoin and set safety checks to prompt.
 # Litecoin does not have strong replay protection using SIGHASH_FORKID, but
 # spending from Bitcoin path should pass with safety checks set to prompt.
 @pytest.mark.altcoin
 def test_invalid_path_prompt(client: Client):
-    # tx: d5f65ee80147b4bcc70b75e4bbf2d7382021b871bd8867ef8fa525ef50864882
-    # input 0: 0.0039 BTC
+    address_n = parse_path("m/44h/0h/0h/0/0")
+    address = btc.get_address(client, "Bitcoin", address_n)
+    prev_hash, prev_tx = forge_prevtx([(address, 390_000)])
```
grdddj, line 70: In this file we seem to have this same setup three times - we could create a global `PREV_TXES = {PREV_HASH: PREV_TX}` as in some other places

grdddj (commented):
The recent changes connected with `forge_prevtx` are very nice!

Seems the function is used everywhere it should, as only `FAKE_TXHASH...` connected with `Bitcoin` is currently the `coinbase` transaction. (I assume it is not possible to use that for altcoins, as `bitcoinlib` does not support them)

The `CI` is currently "failing", but just because of the `scriptPubKey` assertions that were removed - so everything alright

tests/device_tests/bitcoin/test_signtx_invalid_path.py
matejcik: done in 6abaf09

tests/device_tests/bitcoin/signtx.py
matejcik: removed comments in 8155783

tests/device_tests/bitcoin/test_signtx_segwit.py
matejcik: Yes, that was the intention, done while rebasing.
I'm keeping `script_type` explicit -- the very weak justification is that declaring `address_n` is saving one call to `parse_path` while declaring `script_type` only saves typing. However neither is really important in this context.

tests/device_tests/bitcoin/signtx.py
matejcik: right, that's not worth worrying about

PR #2084, opened 2022-01-25
fix(tests): using press_info instead of click for info buttons

grdddj:
Addressing issues discussed in https://github.com/trezor/trezor-firmware/pull/2061#issuecomment-1020341929

The new `click-info` Shamir tests sometimes caused `device-tests` to freeze, most likely because of some issues with watching/not watching layout.

After we are watching the layout regardless of whether we call `wait_layout()`, the errors seem to go away. (It was running consecutively for 30 minutes with around 30 seconds per test duration)


grdddj:
Still WIP - `debug.press_info() ` works now for clicking the INFO button. However, I am still unable to confirm the info popup with any other way than `debug.click(INFO_SHAMIR_RECOVERY)`

grdddj:
Seems to be better now, all the tests in `test_msg_backup_device` as fine. There was some error with nonexisting `self.content` for the info popup.

I did not try any loop to see if it does not fail occasionally.

As we are testing the INFO button even in some other test files, we should also replace it with this new way and get rid of the `wait_layout` stuff

matejcik:
> There was some error with nonexisting `self.content` for the info popup.

yeah, because `self.content` doesn't exist in that class. it would be `self.text` perhaps?



> As we are testing the INFO button even in some other test files, we should also replace it with this new way and get rid of the `wait_layout` stuff

:+1:

grdddj:
> yeah, because `self.content` doesn't exist in that class. it would be `self.text` perhaps?

Right, this works
 ```python
def read_content(self) -> list[str]:
    return self.text
```

grdddj:
Force-pushed with cleaned commit history.

There is a UI difference in UI tests, because when not using `client.click` for the `Info` button, the button will not be highlighted, so there is always one screen missing for each info-click.

matejcik (approved):
and the graphic artifacts are gone too, great!

Issue #2085, opened 2022-01-25
Trezor's RFC6979 operation diverges from that of Bitcoin core


guidovranken:
Bitcoin recently merged a PR which changes the RFC6979 logic slightly: https://github.com/bitcoin-core/secp256k1/pull/1064

This change manifested during differential fuzzing of both libraries. Here's an example input/output:

```
Operation:
operation name: ECDSA_Sign
ecc curve: secp256k1
private key: 22222222222222222222222222222222222222222222222222222222222222222222222222222
cleartext: {0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20} (32 bytes)
nonce source: RFC 6979
digest: NULL
Module secp256k1 result:
X: 84279556694870309735507709563194569800636593243238480998490490018474863819318
Y: 103352824344139482227715594130379040953147412873655551133553784360725214630958
R: 103054904516600407832745405359194456204833007267465256495413130796542780192747
S: 42057782440274948553724453459631334779856220983958300744152873085702181448330
Module trezor-firmware result:
X: 84279556694870309735507709563194569800636593243238480998490490018474863819318
Y: 103352824344139482227715594130379040953147412873655551133553784360725214630958
R: 99970470784785202501425791218580761315684492068983010033898911979393996459919
S: 54683905977549248098326634960786247631805780693695193563622382772855979798501
```

`cleartext` denotes the input message that is passed *unhashed* to `ecdsa_sign()`.

Is this something you would like to address? If not that's fine, but I need to know so I can change my fuzzer logic not to abort on RFC6979 differences between Trezor and Bitcoin.

prusnak:
I am for changing the behaviour in our library as well to match libsecp256k1.

@onvej-sl can you please have a look what needs to be changed in our rfc6979 code to match the new behaviour?

The change in libsecp256k1 was this commit: https://github.com/bitcoin-core/secp256k1/commit/45f37b650635e46865104f37baed26ef8d2cfb97

onvej-sl:
[...]

PR #2089, opened 2022-01-27
fix(crypto): use zero auxiliary data in BIP340

onvej-sl:
This PR:
  1. makes `zkp_bip340_sign_digest` use 32 zero bytes as auxiliary data if no auxiliary data is provided (see the rationale below)
  2. fixes tests accordingly

### Rationale
[BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#default-signing) says auxiliary data is a 32-byte array, which implicitly means the data cannot be empty. The function `secp256k1_schnorrsig_sign` from `secp256k1-zkp` incorrectly handles the case when it is provided with `NULL` as auxiliary data. The function, in that case, uses an empty array which is not in accordance with the BIP but it is completely OK from a security perspective.

The first solution that comes to mind is to use 32 zero bytes as auxiliary data if no auxiliary data is provided. This is what [NBitcoin](https://github.com/MetacoSA/NBitcoin) library does and [secp256k1-zkp](https://github.com/ElementsProject/secp256k1-zkp) does since [5324f89](https://github.com/ElementsProject/secp256k1-zkp/commit/5324f8942dd322448fae6c9b225ecac2854fa7e2).

Once we bump secp256k1-zkp submodule to the latest master, (1) can be removed from this PR.

Note that the [P2TR SLIP-19 test vector](https://github.com/satoshilabs/slips/blob/master/slip-0019.md#test-vector-5-p2tr) should be updated too.


onvej-sl:
> Maybe let's just directly update the secp256k1-zkp submodule in this PR?

I've already started working on it, since I need it to finish https://github.com/trezor/trezor-firmware/issues/2085, but I encountered minor problems. It'll be a part of of this PR or https://github.com/trezor/trezor-firmware/issues/2085.

tests/device_tests/bitcoin/test_fujicoin.py
```diff
@@ -45,5 +45,5 @@ def test_send_p2tr(client):
     # https://explorer.fujicoin.org/tx/a1c6a81f5e8023b17e6e3e51e2596d5b5e1d4914ea13c0c31cef90b3c3edee86
```
andrewkozlik, line 45: Obviously not valid anymore. At least add a comment to explain.

tests/device_tests/bitcoin/test_grs.py
```diff
@@ -240,5 +240,5 @@ def test_send_p2tr(client):
     # https://blockbook-test.groestlcoin.org/tx/c66a79075044aaab3dba17daffb23f48addee87d7c87c7bc88e2997ce38a74ee
```
andrewkozlik, line 240: Same comment as for Fujicoin.

tests/device_tests/bitcoin/test_fujicoin.py
andrewkozlik: The same also applies to the BTC P2TR tests, a lot of which are on the blockchain, but will now have a different signature. Evidently `test_p2tr_with_proof` and `test_p2tr_fee_bump`, but also for example `test_send_mixed`, which is 37aa642d3cc0e0993c32e0213c92231841d3a45fcf0d03e7057ace75dd786146, and probably the others too.

prusnak:
secp256k1-zkp has been updated in https://github.com/trezor/trezor-firmware/pull/2261

do we still need this PR?

onvej-sl:
This draft PR was replaced by https://github.com/trezor/trezor-firmware/pull/2175 (see https://github.com/trezor/trezor-firmware/pull/2175/commits/58d64a2ca963a3f84c903367a08e2cb7bd875f87).

PR #2090, opened 2022-01-27
Remove bitcoin cash Schnorr signatures

onvej-sl:
This PR removes bitcoin cash Schnorr signatures since
  * they are not used and
  * they are not compatible with the BIP340 signatures implemented in secp256k1-zkp.


prusnak (approved):

andrewkozlik:
So does this mean that it has been decided that we won't be implementing Schnorr signatures for BCH or that they can be easily implemented using secp256-zkp?

prusnak:
It's not on the current roadmap. If the need arises in the future, we can easily revert this commit.

andrewkozlik (approved):

PR #2091, opened 2022-01-28
Rust: Simplify LayoutObj

jpochyla:


matejcik (dismissed):

matejcik (commented):
there's some build errors

matejcik:
rebased on current master seems to have fixed the problems, let's see what CI has to say

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -100,7 +100,10 @@ struct LayoutObjInner {
 
 impl LayoutObj {
     /// Create a new `LayoutObj`, wrapping a root component.
-    pub fn new(root: impl ObjComponentTrace + 'static) -> Result<Gc<Self>, Error> {
+    pub fn new<T>(root: T) -> Result<Gc<Self>, Error> 
+    where
+        T: ComponentMsgObj + 'static,
```
jpochyla, line 105: 1. `impl` would be more idiomatic here instead of a named type param, as we don't use `T` anywhere.
2. the trait is wrong and should be `ObjComponentTrace`

jpochyla: Also the trait doesn't include Trace now, that's not correct.

matejcik:
how about this version @jpochyla ?

jpochyla:
Looks good!

PR #2092, opened 2022-01-28
Rust: Add obj_module! macro

jpochyla:


mmilata:
I stole the macro and now it's part of #2138. Let's reopen if 2138 won't get merged.

PR #2095, opened 2022-01-28
Basic typing and styling of test files

grdddj:
Type hints for the helper test functions and basic test typing and styling (mainly for device tests).

With the motto "Bugs hide in untyped code" and "Every untyped function is a lost static-checking opportunity".

All the device tests modifications were done automatically by the `unify_test_files.py` with the combination of `black` and `isort`. It only covers `client: TrezorClientDebugLink` typing, but just this has a big benefit of code-editor showing all the client docs and type-checking it.

When trying to run `show_results.py`, I noticed that the `Path` to the root folder was not resolving properly, so I have added `resolve()` functions every time after accessing `parent` - they also suggest that in their [docs](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parent)


matejcik:
> they also suggest that in their [docs](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parent)

I don't think that is what it says? If you mean this:

> If you want to walk an arbitrary filesystem path upwards, it is recommended to first call Path.resolve() so as to resolve symlinks and eliminate “..” components.

This to me reads that you're supposed to call `resolve()` once before using `.parent` any number of times.
so e.g. `ROOT = HERE.parent.parent` -> `ROOT = HERE.resolve().parent.parent`
in fact in most places we do `HERE = Path(__file__).parent.resolve()` which I know to be correct but it's a little cleaner to rewrite as `HERE = Path(__file__).resolve().parent`, but then you don't need to touch `HERE.parent.parent` in any way because it's resolved already

---

you seem to have removed the leading `m/` from some place? I don't think we need nor want that. If you do path format normalization, please make sure the leading `m/` is always there, it is a very nice indicator that the string is in fact a path.

---

I don't want to make the function signatures too long, so I'd propose a cosmetic change:
```python
from trezorlib.debuglink import TrezorClientDebugLink as Client

def test_foo(client: Client):
    ...
```

grdddj:
> This to me reads that you're supposed to call `resolve()` once before using `.parent` any number of times.

That is probably correct, we should not have any symbolic-links in the repository that would need to be resolved after each call to parent.

----

> make sure the leading m/ is always there, it is a very nice indicator that the string is in fact a path.

Yes, I agree with this - that is fortunately a trivial change in the unifying script. I think I will also commit the script changes in a commit previous to the applied changes, so we know the "conventions" we are enforcing.

------

> import TrezorClientDebugLink as Client

Good idea, I will do it like this

grdddj:
There is quite a lot of UI-tests renaming because of the path normalization - the testnames change because paths in them newly contain `h` instead of `'` as a sign of hardening and they also start with `m`.

I hope this is not an issue, as it means we will need to check if the new tests are still the same as the old ones. I did check some, and they correspond.

Also, at one point I was not sure if we want to use `h` instead of `'`, as the `'` is probably more official. I have no big preference here, we can relatively easily switch to any format.

matejcik:
> we will need to check if the new tests are still the same as the old ones.

you should be able to check that the hashes are unchanged.

grdddj:
> you should be able to check that the hashes are unchanged.

Good idea. Looking at them, they are alright.

tests/buttons.py
```diff
@@ -34,15 +36,15 @@ def grid(dim, grid_cells, cell):
...
     return grid(DISPLAY_WIDTH, 3, x), grid(DISPLAY_HEIGHT, 5, y)
 
 
-def grid34(x, y):
+def grid34(x, y) -> Tuple[int, int]:
     return grid(DISPLAY_WIDTH, 3, x), grid(DISPLAY_HEIGHT, 4, y)
 
 
-def type_word(word):
+def type_word(word: str) -> Generator[Tuple[int, int], None, None]:
```
matejcik, line 47: small nit: this would be better as `Iterator[Tuple[int, int]]`

matejcik:
basically LGTM

do you want to merge this before or after #2081 ?

grdddj:
> do you want to merge this before or after #2081 ?

I would do this one first, #2081 still might need some care, I want to fix unittests there.

I will address the `Iterator[Tuple[int, int]]` while rebasing on master

matejcik (approved):

PR #2097, opened 2022-02-01
Rust UI: confirm_action spacing fixes

mmilata:
confirm_action_v2a

### DIFF
--- a/core/embed/rust/src/ui/component/text/layout.rs
+++ b/core/embed/rust/src/ui/component/text/layout.rs
@@ -31,10 +31,10 @@ pub struct TextLayout {
 
     /// Additional space before beginning of text, can be negative to shift text
     /// upwards.
-    pub top_space: i32,
+    pub padding_top: i32,
     /// Additional space between end of text and bottom of bounding box, can be
     /// negative.
-    pub bottom_space: i32,
+    pub padding_bottom: i32,
 
     /// Background color.
     pub background_color: Color,
@@ -85,8 +85,8 @@ impl TextLayout {
     pub fn new<T: DefaultTextTheme>(bounds: Rect) -> Self {
         Self {
             bounds,
-            top_space: 0,
-            bottom_space: 0,
+            padding_top: 0,
+            padding_bottom: 0,
             background_color: T::BACKGROUND_COLOR,
             text_color: T::TEXT_COLOR,
             text_font: T::TEXT_FONT,
@@ -104,7 +104,7 @@ impl TextLayout {
     }
 
     pub fn initial_cursor(&self) -> Point {
-        self.bounds.top_left() + Offset::y(self.text_font.text_height() + self.top_space)
+        self.bounds.top_left() + Offset::y(self.text_font.text_height() + self.padding_top)
     }
 
     pub fn layout_ops<'o>(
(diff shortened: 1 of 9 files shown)


core/embed/rust/src/ui/component/text/layout.rs
```diff
@@ -29,6 +29,13 @@ pub struct TextLayout {
     /// Bounding box restricting the layout dimensions.
     pub bounds: Rect,
 
+    /// Additional space before beginning of text, can be negative to shift text
+    /// upwards.
+    pub top_space: i32,
+    /// Additional space between end of text and bottom of bounding box, can be
+    /// negative.
+    pub bottom_space: i32,
+
```
jpochyla, line 38: Personally I would call these `padding_top` and `padding_bottom`.

core/embed/rust/src/ui/component/text/layout.rs
```diff
@@ -107,7 +113,13 @@ impl TextLayout {
...
         sink: &mut dyn LayoutSink,
     ) -> LayoutFit {
-        let init_cursor: Point = *cursor;
+        let init_cursor = *cursor;
+        let height = move |end_cursor: &Point| {
+            self.top_space
+                + self.text_font.text_height()
+                + (end_cursor.y - init_cursor.y)
+                + self.bottom_space
+        };
```
jpochyla, line 122: Maybe it would be better to put this into a method? Something like `fn layout_height(&self, init_cursor: Point, end_cursor: Poin) -> i32`?

core/embed/rust/src/ui/component/text/layout.rs
jpochyla: Or maybe we should be using `Insets` for this?

jpochyla: The difference to just insetting `bounds` is that the returned text height includes the padding, right?

core/embed/rust/src/ui/geometry.rs
```diff
@@ -471,6 +475,31 @@ impl LinearLayout {
...
+    /// closure is called `count` times with top left point of each item as
+    /// argument. Items are centered along the cross axis.
+    pub fn arrange_uniform(
+        &self,
+        area: Rect,
+        count: usize,
+        size: Offset,
+        sink: &mut dyn FnMut(Point),
+    ) {
+        let item_size = self.axis.main(size.x, size.y);
```
jpochyla, line 489: `size.axis(self.axis)`?

core/embed/rust/src/ui/model_tt/component/page.rs
```diff
@@ -230,40 +233,42 @@ impl Component for ScrollBar {
     }
 
     fn paint(&mut self) {
-        let count = self.page_count as i32;
-        let interval = {
-            let available_height = self.area.height();
-            let naive_height = count * Self::DOT_INTERVAL;
-            if naive_height > available_height {
-                available_height / count
+        let layout = LinearLayout::vertical()
```
jpochyla, line 236: Niiiiiice

jpochyla (commented):
I got a bit confused about the spacing in layout, so left some comments.

core/embed/rust/src/ui/geometry.rs
mmilata: Good catch. Looks funny though, i wonder if there's a better name for the method:)

mmilata: Fixed in c1588ff.

core/embed/rust/src/ui/component/text/layout.rs
mmilata: Renamed the fields in 6dd100c.

core/embed/rust/src/ui/component/text/layout.rs
mmilata: Fixed in f24dca9. The text height might be different at the end compared to the beginning but we don't really support mixing font sizes anyway.

mmilata:
Merged into #2071.

PR #2098, opened 2022-02-03
Rust UI: Keyboards

jpochyla:


matejcik (approved):
LGTM, to be merged after #2109 which brings the new Rust we need

mmilata (commented):
Overall this looks great but is not working:
* in `PassphraseKeyboard` the character cycling doesn't seem to work,
* `Bip39` draws single button covering whole screen.

But these are probably some small bugs which I can iron out later as part of making this look like the new design.

Question: we're adding new timer for each key press, and have capacity for 4 timers. Can the user crash the firmware by typing really fast?

jpochyla:
The timers are drained after the event pass, so the capacity is for max timers per event pass, which will most probably be constant in all cases.

matejcik:
```
error: manual `!RangeInclusive::contains` implementation
```
whoa :)

PR #2101, opened 2022-02-04
fix(core): fix build with FEATURE_FLAGS.SECP256K1_ZKP=False

prusnak:
When SECP256K1_ZKP is set to False in SConscript.{firmware,unix} the build fails.

This PR fixes the issue. However, this still does not mean the firmware is usable as is, since we use the bip340 in our codebase unconditionally. 

The change is still beneficial IMHO because this allows development of a completely custom firmware (i.e. different `core/src`) with SECP256K1_ZKP not included in the build.

### DIFF
--- a/core/SConscript.firmware
+++ b/core/SConscript.firmware
@@ -10,7 +10,7 @@ NEW_UI = TREZOR_MODEL == '1' or os.environ.get('NEW_UI', '0') == '1'
 
 FEATURE_FLAGS = {
     "RDI": True,
-    "SECP256K1_ZKP": True,
+    "SECP256K1_ZKP": True,  # required for trezor.crypto.curve.bip340 (BIP340/Taproot)
     "SYSTEM_VIEW": False,
 }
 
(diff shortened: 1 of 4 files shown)


andrewkozlik:
I am not particularly happy about the fact that it allows one to break Taproot signing without giving any indication. You don't find out until you try to sign a Taproot input and get a firmware error. I'd at least add a comment in SConscript.unix and SConscript.firmware, like
```
FEATURE_FLAGS = {
    "SECP256K1_ZKP": True,  # required for Taproot
}
```

prusnak:
> it allows one to break Taproot signing without giving any indication.

Addressed in 2b9dd72292676f71a06b586aa160d80a570cb21a

andrewkozlik (approved):

prusnak:
Style check still needs to be fixed. But maybe rather than fixing the code in `core/src` we can just revert the change in `core/src/trezor/crypto/curve.py` and be done with it, since the code in `core/src` is not usable as-is anyway.

matejcik:
+1 on reverting the `curve.py` change. We can re-do it if we ever build code that is aware of the difference.

prusnak:
Reverted the `curve.py` change and rebased

PR #2103, opened 2022-02-07
fix(core/ethereum): implement hold-to-confirm for eip1559 transactions

### LINKED ISSUE #2020: EIP-1559 transactions on TT do not require hold 5 seconds to confirm
Transaction is confirmed immediately after user confirms totals fees. This is different from how confirmations works for legacy ETH transactions and for other coins and may lead to confirmation of unwanted transaction, due to user may expect one more screen for final confirmations, or due to poor quality touchscreen in TT (mine sometimes registers click in wrong location or double clicks instead of single click).

matejcik:
fixes #2020 

image


grdddj (approved):
Looks good, I like the UI change, a nice summary before sending.

mmilata (approved):
Looking good!

PR #2106, opened 2022-02-08
ci: always run UI tests

matejcik:
We don't seem to actually need a separate UI test pipeline. We will just run the UI tests in `core device test` pipeline.

This PR also removes `skip_ui` support -- that way we can be certain that no tests are skipped due to the UI run.


ci/test.yml
```diff
@@ -71,27 +71,6 @@ core device ui test:
     reports:
       junit: tests/junit.xml
 
-core device test:
-  stage: test
-  <<: *gitlab_caching
-  needs:
-    - core unix frozen debug build
-  variables:
-    TREZOR_PROFILING: 1
```
mmilata, line 80: We'll need to enable the coverage data collection in the new `core device test`.

PR #2107, opened 2022-02-08
Fix legacy technical debt in USB handling

hiviah:
This change is aimed to make USB handling in T1 legacy more readable and also remove states where handling of USB protobuf messages should not happen (even if not exploitable right now, it helps prevent accidental bugs in future).

FSM and message handling is already complicated due to call graph being complex.

### DIFF
--- a/legacy/common.c
+++ b/legacy/common.c
@@ -19,6 +19,7 @@
 
 #include "common.h"
 #include <stdio.h>
+#include <unistd.h>
 #include "bitmaps.h"
 #include "firmware/usb.h"
 #include "hmac_drbg.h"
@@ -83,7 +84,17 @@ void __assert_func(const char *file, int line, const char *func,
 }
 #endif
 
-void hal_delay(uint32_t ms) { usbSleep(ms); }
+void hal_delay(uint32_t ms) {
+#if EMULATOR
+  usleep(ms * 1000);
+#else
+  uint32_t start = timer_ms();
+
+  while ((timer_ms() - start) < ms) {
+    asm("nop");
+  }
+#endif
+}
 
 void drbg_init() {
   uint8_t entropy[48] = {0};
--- a/legacy/firmware/udp.c
+++ b/legacy/firmware/udp.c
@@ -18,6 +18,7 @@
  */
 
 #include <stdint.h>
+#include <unistd.h>
 
 #include "usb.h"
 
@@ -35,7 +36,7 @@ void usbInit(void) { emulatorSocketInit(); }
 #define _ISDBG ('n')
 #endif
 
-void usbSleep(uint32_t millis) {
+void waitAndProcessUSBRequests(uint32_t millis) {
   emulatorPoll();
 
   static uint8_t buffer[USB_PACKET_SIZE];
@@ -63,7 +64,7 @@ void usbSleep(uint32_t millis) {
 #endif
 }
 
-void usbPoll(void) { usbSleep(0); }
+void usbPoll(void) { waitAndProcessUSBRequests(0); }
 
 char usbTiny(char set) {
   char old = tiny;
@@ -71,4 +72,10 @@ char usbTiny(char set) {
   return old;
 }
 
-void usbFlush(uint32_t millis) { usbSleep(millis); }
+void usbFlush(uint32_t millis) {
+  const uint8_t *data;
+  while ((data = msg_out_data()) != NULL) {
+    emulatorSocketWrite(0, data, USB_PACKET_SIZE);
+  }
+  usleep(millis * 1000);
+}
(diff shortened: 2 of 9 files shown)


legacy/common.c
```diff
@@ -83,7 +84,17 @@ void __assert_func(const char *file, int line, const char *func,
 }
 #endif
 
-void hal_delay(uint32_t ms) { usbSleep(ms); }
+void hal_delay(uint32_t ms) {
+#if EMULATOR
+  usleep(ms * 1000);
```
prusnak, line 89: The reason why we used usbSleep here is that device responds to usb packets when it is sleeping.
Are we 100% sure this will not cause any problems (device randomly disconnecting)?

hiviah: With `usbSleep` device replies to *USB protobuf* messages.

Even with NOP loop (like in the PR), the USB itself works and doesn't cause disconnects, USB IRQ handlers put the data in FIFO. That `hal_delay` is only called when PIN is checked and user is "punished" for bad attempts (where no protobuf messages should be handled). Once normal operation is resumed, USB protobuf data is handled normally.

I've tried it on hardware T1 with Suite doing setup several times, also tried to change PIN, did a lot of wrong PIN inputs to increase the delay up to 16 seconds.

Everything works as it did before as far as I can tell.

prusnak: Thanks!

legacy/firmware/udp.c
```diff
@@ -63,12 +64,12 @@ void usbSleep(uint32_t millis) {
...
+void usbPoll(void) { waitAndProcessUSBRequests(0); }
 
 char usbTiny(char set) {
   char old = tiny;
   tiny = set;
   return old;
 }
 
-void usbFlush(uint32_t millis) { usbSleep(millis); }
+void usbFlush(uint32_t millis) { usleep(millis * 1000); }
```
matejcik, line 75: This doesn't flush anymore though. We actually need to call "process USB requests" to do that. We might want to separate "write out pending" from "read incoming" for the purpose of this function.

legacy/firmware/usb.h
```diff
@@ -25,8 +25,32 @@
 void usbInit(void);
 void usbPoll(void);
 void usbReconnect(void);
+
+/*
+ * Setting this value to 0 will limit the protobuf messages `usbPoll` and
```
invd, line 30: I think the comment should be extended since the protobuf USB messages are just one of three types of packets that are affected by this.

legacy/firmware/usb.h
```diff
@@ -25,8 +25,32 @@
...
+ */
 char usbTiny(char set);
-void usbSleep(uint32_t millis);
+
+/*
+ * This will wait given number of milliseconds for arrival of protobuf message.
+ * If it arrives, it will service it before returning.
+ *
+ * If you call this function from any function that is called from FSM,
+ * you must use `usbTiny(0)` before and `usbTiny(oldTinyValue)` after, otherwise
```
invd, line 45: I think this is not correct.

The pattern referenced here is as follows:
```C
oldTinyValue = usbTiny(1);
// do an operation that requires reading USB messages of the "tiny" type
usbTiny(oldTinyValue)
```

Technically, 
```C
usbTiny(1);
// do an operation that requires reading USB messages of the "tiny" type
usbTiny(0)
```
also works, but has different side effects on the global tiny state after processing.

In either case, a `usbTiny(1)` is required to go into the special mode, not `usbTiny(0)`.

legacy/common.c
hiviah: So far I've discovered that having `hal_delay` as NOP should work with legitimate clients (no recursion attempts which should not work anyway), malicious clients could just send bad PIN attempts to try device lockout anyway even now. However it's hard to try out all race conditions with client/T1 disconnects over all OSes to be 100% sure there is no edge case not covered (experimentally everything works as it's supposed even when trying disconnects, but...). So I'd probably have `hal_delay` to be a NOP sleep, but explicitly use the `waitAndProcessUSBRequests` in unlock as to not appear to be a no-op delay.

legacy/firmware/udp.c
hiviah: Fixed by writing out oustanding packets.

legacy/firmware/usb.h
hiviah: Updated comment to reflect those messages.

legacy/common.c
hiviah: After careful code analysis, `hal_delay` with NOP should work fine and I did test various race conditions induced manually as well. If busy loops broke USB low-level responsiveness, it would have also happened in long signing operations, for example, as USB low-level timeouts are very short.

matejcik (approved):
utACK

Issue #2093, opened 2022-01-28
UI tests for legacy


matejcik:
We actually have screenshotting capability in T1's debuglink. `DebugLinkState.layout` sends contents of the screen buffer.

We should be able to use this to create UI test recordings for legacy as well, with reasonable amount of effort.

The following snippet decodes the image format:
```python
                    assert len(state.layout) == 128 * 64 // 8
                    from PIL import Image

                    pxrows = [
                        state.layout[r * 128 : (r + 1) * 128] for r in range(64 // 8)
                    ]
                    pixels = []
                    for byteline in range(64 // 8):
                        offset = byteline * 128
                        row = state.layout[offset : offset + 128]
                        for bit in range(8):
                            pixels.extend(bool(px & (1 << bit)) for px in row)

                    im = Image.new("1", (128, 64))
                    im.putdata(pixels[::-1])
                    im.save(f"layout{self.counter:04d}.png")
```

A good place for it _seems_ to be the `DebugLinkUI.button_request` handler.

bosomt:
@matejcik i think we can move this from `Needs QA` column

PR #2108, opened 2022-02-08
UI tests for T1

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2093:
- enabling UI tests for `T1` by supporting screenshotting capability for it

Considerations:
- it was _not_ rebased on current `master`, as I actually found a use-case for the recently removed `skip_ui` mark - `test_protection_levels.py::test_reset_device` does not have consistent `UI` - it is generating always a new seed - so it is not possible to run UI tests on it (they will succeed in one case out of 2^128)
  - we might want to name the mark differently, maybe `inconsistent_ui`
- the screenshotting function and helper variables are defined on `DebugLink`, we might want to create a separate class, something like `ScreenshotterT1`
- the way we are passing the `model` (`T1/TT`) around does not look the best in both new cases - putting it into `Debuglink` and into `ui_tests` - there might be some better ways
- we could have had a smaller `fixtures.json` diff by prepending just `T1` tests with `T1_`, but prepending everything with `{model}_` makes it more explicit and we are ready to support a new model any time


tests/conftest.py
```diff
@@ -106,8 +108,8 @@ def client(request: pytest.FixtureRequest, _raw_client: Client) -> Client:
         request.session.shouldstop = "Failed to communicate with Trezor"
         pytest.fail("Failed to communicate with Trezor")
 
-    if run_ui_tests:
-        # we need to reseed before the wipe
+    if run_ui_tests and _raw_client.features.model == "T":
+        # we need to reseed before the wipe for TT
```
matejcik, line 112: i believe we will also need to reseed on T1

tests/ui_tests/__init__.py
```diff
@@ -99,10 +103,18 @@ def _process_tested(fixture_test_path: Path, test_name: str) -> None:
 
 @contextmanager
 def screen_recording(
-    client: Client, request: pytest.FixtureRequest
+    client: Client, request: pytest.FixtureRequest, model: str
```
matejcik, line 106: why not extract model from the `client`?

tests/device_tests/test_protection_levels.py
```diff
@@ -218,6 +218,7 @@ def test_wipe_device(client: Client):
 
 @pytest.mark.setup_client(uninitialized=True)
 @pytest.mark.skip_t2
+@pytest.mark.skip_ui
```
matejcik, line 221: `skip_ui` was removed

matejcik (commented):
This is looking very good, but we will need to resolve the `skip_ui` thing and reseeding.

Also please add screenshots to PIN and passphrase request handlers. But that will cause a lot of breakage if we don't solve reseeding.

T1 currently doesn't support `DebugLinkReseed`, right? We'll probably need to implement it.

matejcik:
> it is generating always a new seed - so it is not possible to run UI tests on it (they will succeed in one case out of 2^128)

this is why we need reseeding :) let's not bring back `skip_ui`

> * the screenshotting function and helper variables are defined on `DebugLink`, we might want to create a separate class, something like `ScreenshotterT1`

probably no need for that right now

> * prepending everything with `{model}_` makes it more explicit and we are ready to support a new model any time

agreed

tests/ui_tests/__init__.py
grdddj: Surely, I missed we have access to it

grdddj:
> Also please add screenshots to PIN and passphrase request handlers. But that will cause a lot of breakage if we don't solve reseeding.

PIN screenshots are fine, but the passphrase ones are really causing issues - the emulator freezes with the `Please enter your passphrase...` screen. How would reseeding help us here? I am not sure why the freeze happens, maybe the screenshot is a longer action working on `DebugLink` and it corrupts the `get_passphrase` function.

I checked and `DebugLinkReseedRandom` is really in the list of `MessageTypes` to be ignored by `T1`. How hard it would be to support it and implement in in `legacy`?

I still did not find out why the `Posttest` and `Deploy` phases are failing. It seems it is caused by renaming of test folders, and it cannot find the recorded screenshots. What is the process of generating them on the CI servers?

matejcik:
I implemented reseeding for T1 in https://github.com/trezor/trezor-firmware/compare/matejcik/t1-reseed?expand=1
Feel free to cherry-pick for your experiments, but I would like to review that change separately before building this PR on top of it.

I will look at the passphrase thing tomorrow.

grdddj:
> I implemented reseeding for T1 in https://github.com/trezor/trezor-firmware/compare/matejcik/t1-reseed?expand=1 Feel free to cherry-pick for your experiments, but I would like to review that change separately before building this PR on top of it.
> 
> I will look at the passphrase thing tomorrow.

Thanks for that.

I have created a new, rebased-on-master branch, which also includes yours two commits - https://github.com/trezor/trezor-firmware/commits/grdddj/t1_ui_tests_t1_reseed

Still (after many builds of T1 emulator), the UI tests for `T1` are freezing with the traceback at the "reseeding" line.

There might be issue with how I am building it, my steps are:
```sh
cd legacy
export EMULATOR=1 DEBUG_LINK=1
./script/setup
./script/cibuild
```

Tests in CI are also timing out - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2074335567

grdddj:
> i fixed the reseed branch, please cherry-pick from #2115

Thanks, the reseeding now seems to work alright everywhere. I needed to include it in a couple of tests that were "PIN-heavy" (more than one PIN dialogue was shown), as otherwise they were not stable.

The passphrase seems to be the last (but theoretically not critical/necessary) issue we have here

grdddj:
Interestingly, three `TT` UI tests broke because of the additional reseeding, so model-condition was added there.

Seems that in case of `TT`, the reseeding is valid for the whole session, and in `T1` case just for the next set of PIN screens.

matejcik:
> I needed to include it in a couple of tests that were "PIN-heavy" (more than one PIN dialogue was shown), as otherwise they were not stable.

We should basically never reseed _inside_ tests. You're patching over an underlying issue that will need to be investigated in T1 code.

matejcik:
got lucky with a reproducer -- it seems it's THE CACHES again :angry: #2125

matejcik:
Flipped back to draft because of f0b868b which should not be merged here but in #2125 

Apart from that, I redid the commit history a bit and enabled passphrase screen capture. Right now this is working for me locally, let's see what CI has to say.

PR #2109, opened 2022-02-08
chore: update dependencies

prusnak:


mmilata:
Looks like `black` is failing: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2069163728#L3039
Otherwise LGTM, will leave up to @matejcik if it's time to remove 3.6 yet.

matejcik:
As I said before, I would like to keep Python 3.6 compatibility in trezorlib for as long as it is not a burden.

If including Python 3.6 in CI is problematic, we can drop it, but I'd keep it in tox and keep the package requirements as-is

matejcik:
> Reworked the PR to keep Py3.6 compatibility

wasn't the previous version where pillow is 0.9 working for some reason?

prusnak:
> wasn't the previous version where pillow is 0.9 working for some reason?

Pillow 0.9.x is not compatible with Py3.6

Pillow 0.8.4 works on my machine, Pillow 0.8.1 (originally suggested) doesn't.

prusnak:
One more force push - pyproject.toml set to Py3.7, trezorlib kept at 3.6

prusnak:
@mmilata any idea why pyright is now missing?

matejcik:
do you maybe need to rerun the pyright generator script? https://github.com/trezor/trezor-firmware/blob/master/ci/pyright/README.md

prusnak:
> do you maybe need to rerun the pyright generator script? https://github.com/trezor/trezor-firmware/blob/master/ci/pyright/README.md

I tried it ... it refreshed lots of dependencies of pyright, but result is still the same - the environment does not have `pyright` in path.

mmilata:
Looks like pyright gets built but for some reason the `bin -> lib/node_modules/.bin` symlink disappeared, figuring out why ...

mmilata:
Ah, there's no `lib/node_modules/.bin` either, and pyright is `lib/node_modules/pyright/index.js`. Doesn't explain what happened thought.

mmilata:
Seems to be related to nodejs 14->16 update: nixos/nixpkgs#145432. My attempt for workaround: 91182cf3. Other possibility is to create the symlinks manually in `postInstall` hook.

prusnak:
> Seems to be related to nodejs 14->16 update: [NixOS/nixpkgs#145432](https://github.com/NixOS/nixpkgs/issues/145432). My attempt for workaround: [91182cf](https://github.com/trezor/trezor-firmware/commit/91182cf3d438192c026c1ea8794008ff8af4d1f8). Other possibility is to create the symlinks manually in `postInstall` hook.

Thanks, that did the trick. Cherry-picked and rebased.

mmilata (commented):
Looking good, though ti seems pylint became more stricter and fails: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2083350519

matejcik:
more like pylint changed the ignore list format AGAIN :angry: i'll fix it later if nobody else beats me to it

matejcik:
it's actually still the same issue that prevented me from updating the last time: https://github.com/PyCQA/pylint/issues/5491
fixed by removing leading `./` from paths passed in

mmilata (approved):
CI passes, let's merge.

Issue #2100, opened 2022-02-04
Support Electrum signatures in VerifyMessage


andrewkozlik:
Trezor supports message signing and verification in accordance with [BIP-137](https://github.com/bitcoin/bips/blob/master/bip-0137.mediawiki). The spec includes script type information to the first byte of the signature. Electrum does not include the script type information in the signature and is unable to verify message signatures produced by Trezor unless the `no_script_type` flag is set to `true` during signing (or unless the information is manually removed from the first byte). Similarly Trezor is unable to verify signatures generated by Electrum.

The script type information is actually not required in order to be able to verify the signature if the address is available to the verifier. In the case of Trezor the `address` field is required in `VerifyMessage`, so there is no reason why Trezor couldn't verify signatures generated by Electrum.

bosomt:
**QA OK**

_message generated via T1 latest public version in Electrum 4.1.5_

**T1 1.10.6 0x0e27423cffb3b6047028c61890bec43bf626adc7**
Signature is valid + `Message verification successful`
**TT 2.4.4 0x0e27423cffb3b6047028c61890bec43bf626adc7**
Signature is valid + `Message verification successful`

_message generated in Electrum 4.1.5 from standard local wallet_

**T1 1.10.6 0x0e27423cffb3b6047028c61890bec43bf626adc7**
Signature is valid + `Message verification successful`
**TT 2.4.4 0x0e27423cffb3b6047028c61890bec43bf626adc7**
Signature is valid + `Message verification successful`

only way to generate signed message via Trezor was 1.10.5  T1 firmware version

Issue #1880, opened 2021-11-01
"Verify message" does not display result on screen


matejcik:
This is the result of a counterintuitive flow: Trezor first verifies the message, and _only if the signature is OK_, shows the message and address on screen.

So if I'm seeing the message on Trezor screen, that means the signature is OK. An attacker cannot abuse this to claim that a bad signature is OK.
(They could abuse this to claim that an OK signature is bad, BUT they can do that anyway by corrupting the signature.)

However, from the user's point of view, the Trezor _never says that the signature is ok_.

A very simple improvement is to stick a pop-up "Signature is OK" at end.

In the new UI design, we might consider showing the message even if the signature is bad, so that the user gets feedback in such case. Currently, Trezor doesn't show anything and only Suite shows the failure.

bosomt:
**QA OK**

`Signature is OK` displayed on both models

1.10.6 revision: 20 bytes 0x0e27423cffb3b6047028c61890bec43bf626adc7,
2.4.4 revision: 20 bytes 0x0e27423cffb3b6047028c61890bec43bf626adc7

PR #2112, opened 2022-02-09
Support Electrum signatures in VerifyMessage.

andrewkozlik:
Implements:
- #2100 in T1 and TT.
- #1880 in T1 and TT.
- Generating Electrum-compatible message signatures in trezorctl.
- Removes Bitcoin Cash from T1 Bitcoin-only firmware.

Screenshots for #1880:
image image

Suite/Connect:
`VerifyMessage` and `EthereumVerifyMessage` will show one more confirmation screen at the end of the workflow, see screenshot, with button request type `success_verify` and code `Other`.

QA:
Check that T1 and TT are able to verify messages signed by Electrum. Test with Legacy, P2SH SegWit and native SegWit addresses.


core/src/apps/bitcoin/verify_message.py
```diff
@@ -18,9 +20,44 @@
...
+        raise wire.DataError("Invalid address")
+
+    if address_type.check(coin.address_type, raw_address):
+        # p2pkh
+        return InputScriptType.SPENDADDRESS
+    elif address_type.check(coin.address_type_p2sh, raw_address):
+        # p2sh
+        return InputScriptType.SPENDP2SHWITNESS
+
+    raise wire.DataError("Invalid address")
```
matejcik, line 58: small nit: in case the user passes an invalid address and an undetermined signature, they get `DataError` "Invalid address".
if they pass in an invalid address but one of the specific signatures, they get `ProcessError` "Invalid signature" (from recovered pkey != address)

To do this fully correctly, we would need to parse the address in all cases and then compare the `recid` with what we determined.
But I'm OK with leaving this as-is, just wanted to raise the point in case you see some sort of problem with it.

matejcik (approved):

core/src/apps/bitcoin/verify_message.py
matejcik: Yes. I meant, we parse address first (to report "malformed address" condition), then fold all subsequent errors (such as recid mismatch) into "invalid signature" condition.

It would not be too complicate to do the same on T1 -- `EXTERNAL` means "bad address".

I am still not sure if it's worth doing. But ISTM it is a good idea to change T1's `DataError` to `ProcessError` to unify the behavior -- invalid signature is not an error in data.

core/src/apps/bitcoin/verify_message.py
```diff
@@ -30,17 +67,16 @@ async def verify_message(ctx: wire.Context, msg: VerifyMessage) -> Success:
...
 
+    script_type = address_to_script_type(address, coin)
     recid = signature[0]
     if 27 <= recid <= 34:
-        # p2pkh
-        script_type = InputScriptType.SPENDADDRESS
-    elif 35 <= recid <= 38:
+        # p2pkh or no script type provided
+        pass  # use the script type from the address
+    elif 35 <= recid <= 38 and script_type == InputScriptType.SPENDP2SHWITNESS:
```
matejcik, line 75: will `verify_recover()` fail if we don't tweak the signature?

matejcik:
code-wise LGTM. assuming the answer to my comment about "verify_recover" is yes, please squash fixups & merge

core/src/apps/bitcoin/verify_message.py
andrewkozlik: Yes, it would fail. Why is this important?

matejcik: Because this is the only recid <-> script-type crosscheck. If it wouldn't fail, then the "and script_type == something" checks are pointless.

andrewkozlik: I don't think they would be pointless. If the `and script_type == ...` condition fails, then we end up raising a `ProcessError` on line 82, i.e. never reach `verify_recover()`.

matejcik: right, i somehow completely missed that. thanks.

PR #2115, opened 2022-02-10
Reseeding for T1

matejcik:
This replaces the `/dev/urandom` based randomness source for T1 **emulator** with the stock naive LCG from trezor-crypto -- which, importantly, we can reseed as needed.

In terms of security, Trezor emulators are not intended for production use anyway so "generate randomness securely" is not a feature we need to keep. Trezor T emulator is already using the same naive implementation.

Nothing changes for the hardware build.


prusnak (approved):
LGTM

PR #2116, opened 2022-02-10
ci: fix the environment build after changing runner config

vdovhanych:
Fixes the environment build after the configuration of the runner was changed to fix HWI tests.


ci/environment.yml
```diff
@@ -18,7 +18,9 @@
     - echo "${ALPINE_CHECKSUM}  ci/alpine-minirootfs-$ALPINE_VERSION-$ALPINE_ARCH.tar.gz" | sha256sum -c
     - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:latest --build-arg ALPINE_VERSION="$ALPINE_VERSION" --build-arg ALPINE_ARCH="$ALPINE_ARCH" --build-arg NIX_VERSION="$NIX_VERSION" --build-arg FULLDEPS_TESTING=1 ci/
     - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
-    - docker push $CONTAINER_NAME:latest
+    - docker push $CONTAINER_NAME:latest\
```
mmilata, line 21: Is the backslash intentional?

vdovhanych: typo...

Issue #1130, opened 2020-07-24
trezorctl: unify option arguments


matejcik:
due to the development history, similar subcommands in `trezorctl` have wildly different parameter conventions:

* To enable pin, use `trezorctl set pin`. To disable pin, use `trezorctl set pin -r`
* To enable passphrase, use `trezorctl set passphrase enabled`. To disable, `trezorctl set passphrase disabled`
* To enable SD-protect, use `trezorctl device sd-protect enable`. To disable, use `trezorctl device sd-protect disable`
* some commands accept a single required option, other accept a single argument
* etc.

we should come up with a set of guidelines and make sure all commands conform to them.

grdddj:
Commands for `QA` to try:

`trezorctl set pin`
`trezorctl set wipe-code`
`trezorctl set passphrase`
`trezorctl device sd-protect`

Check that all these commands take an arguments `on` and `off` as a way to specify the positive and negative effects (`trezorctl set pin on` vs `trezorctl set pin off`).

In the case of `trezorctl set passphrase`, also the `enabled/disabled` argument should work (to maintain backward compatibility) and be identical to `on/off`.

In the case of `trezorctl set pin` and `trezorctl set wipe-code`, the "legacy" `-r` flag is still supported as a way to "remove" - the same as `off` argument. These two cannot conflict with each other, therefore the usage `trezorctl set pin on -r` should return an error.

wendys-cats:
**QA NOK** - commands trezorctl set passphrase is not valid, nor is trezorctl device sd-protect

**Info**
**trezorctl** version 0.13.1 (via poetry from monorepo)
**device** used: TT 2.5.0
**OS:** Nix

**trezorctl device sd-protect**
![image](https://user-images.githubusercontent.com/58593111/161853389-bd079386-1716-40ff-a67c-bf910fe18bc2.png)

**trezorctl set passphrase**
![image](https://user-images.githubusercontent.com/58593111/161853455-c5b51493-189f-4487-aa52-e558fd4222d1.png)


**List of tried commands:**
 set pin OK
 set pin off OK
set pin on OK
set pin -r OK
set wipe-code OK
set wipe code on OK
set wipe code -r OK
set wipe code off OK
**set passphrase NOK**
set passphrase on OK
set passphrase off OK
set passphrase enabled OK
set passphrase disabled OK
**device sd-protect NOK**
device sd-protect on OK
device sd-protect off OK
device sd-protection refresh OK

[...]

PR #2123, opened 2022-02-11
feat(python): unify option arguments in trezorctl

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1130:
- unifying how the on/off, enable/disable, enabled/disabled, -/-r arguments look like
- currently `enabled/disabled`, with `enabled` being a default one does not seem bad
- I did not go through all the commands, just the basic ones from the issue

There are some changes in OPTIONS.rst, made by the helper script (from previous commits) - could these be run now, or are we always waiting for the `trezorlib` release?

### DIFF
--- a/python/src/trezorlib/cli/settings.py
+++ b/python/src/trezorlib/cli/settings.py
@@ -95,24 +95,28 @@ def cli() -> None:
 
 
 @cli.command()
-@click.option("-r", "--remove", is_flag=True)
+@click.argument(
+    "enable", type=ChoiceType({"enable": True, "disable": False}), default=True
+)
 @with_client
-def pin(client: "TrezorClient", remove: bool) -> str:
+def pin(client: "TrezorClient", enable: bool) -> str:
     """Set, change or remove PIN."""
-    return device.change_pin(client, remove)
+    return device.change_pin(client, remove=not enable)
 
 
 @cli.command()
-@click.option("-r", "--remove", is_flag=True)
+@click.argument(
+    "enable", type=ChoiceType({"enable": True, "disable": False}), default=True
+)
 @with_client
-def wipe_code(client: "TrezorClient", remove: bool) -> str:
+def wipe_code(client: "TrezorClient", enable: bool) -> str:
     """Set or remove the wipe code.
 
     The wipe code functions as a "self-destruct PIN". If the wipe code is ever
     entered into any PIN entry dialog, then all private data will be immediately
     removed and the device will be reset to factory defaults.
     """
-    return device.change_wipe_code(client, remove)
+    return device.change_wipe_code(client, remove=not enable)
 
 
 @cli.command()
@@ -218,7 +222,7 @@ def safety_checks(
 
 
 @cli.command()
-@click.argument("enable", type=ChoiceType({"on": True, "off": False}))
+@click.argument("enable", type=ChoiceType({"enable": True, "disable": False}))
 @with_client
 def experimental_features(client: "TrezorClient", enable: bool) -> str:
     """Enable or disable experimental message types.
@@ -240,7 +244,7 @@ def passphrase() -> None:
     # and "disable-passphrase". Otherwise `passphrase` would just take an argument.
 
 
-@passphrase.command(name="enabled")
+@passphrase.command(name="enable")
 @click.option("-f/-F", "--force-on-device/--no-force-on-device", default=None)
 @with_client
 def passphrase_enable(client: "TrezorClient", force_on_device: Optional[bool]) -> str:
@@ -250,7 +254,7 @@ def passphrase_enable(client: "TrezorClient", force_on_device: Optional[bool]) -
     )
 
 
-@passphrase.command(name="disabled")
+@passphrase.command(name="disable")
 @with_client
 def passphrase_disable(client: "TrezorClient") -> str:
     """Disable passphrase."""
--- a/python/docs/OPTIONS.rst
+++ b/python/docs/OPTIONS.rst
@@ -30,6 +30,7 @@ on one page here.
     -v, --verbose             Show communication messages.
     -j, --json                Print result as JSON object
     -P, --passphrase-on-host  Enter passphrase on host.
+    -S, --script              Use UI for usage in scripts.
     -s, --session-id HEX      Resume given session ID.
     --version                 Show the version and exit.
     --help                    Show this message and exit.
@@ -251,12 +252,13 @@ Ethereum commands.
     --help  Show this message and exit.
 
   Commands:
-    get-address      Get Ethereum address in hex encoding.
-    get-public-node  Get Ethereum public node of given path.
-    sign-message     Sign message with Ethereum address.
-    sign-tx          Sign (and optionally publish) Ethereum transaction.
-    sign-typed-data  Sign typed data (EIP-712) with Ethereum address.
-    verify-message   Verify message signed with Ethereum address.
+    get-address           Get Ethereum address in hex encoding.
+    get-public-node       Get Ethereum public node of given path.
+    sign-message          Sign message with Ethereum address.
+    sign-tx               Sign (and optionally publish) Ethereum transaction.
+    sign-typed-data       Sign typed data (EIP-712) with Ethereum address.
+    sign-typed-data-hash  Sign hash of typed data (EIP-712) with Ethereum address.
+    verify-message        Verify message signed with Ethereum address.
 
 FIDO2, U2F and WebAuthN management commands.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


python/src/trezorlib/cli/settings.py
```diff
@@ -95,24 +95,28 @@ def cli() -> None:
 
 
 @cli.command()
-@click.option("-r", "--remove", is_flag=True)
```
matejcik, line 98: Please keep the `-r` option functional for backwards compatibility.

All of these should work:
```
trezorctl set pin on
trezorctl set pin
trezorctl set pin off
trezorctl set pin -r
```
I'd ignore the `set pin off -r` case, but raise on `set pin on -r`

dtto for wipe-code

matejcik (changes requested):
Unfortunately `set <something> enable` is bad English -- there's two verbs, it should be "set <something> (to) enable**d**".

Let's change it to `on/off` everywhere. It is short and seems to work both with `device sd-protect off` and `set passphrase off`.

python/src/trezorlib/cli/settings.py
matejcik: that would be nice but tricky

alright, how about this:

- define a global choice group where "enable/enabled/on" match True and "disable/disabled/off" match False. Use it everywhere.
- convert `passphrase_enable()` and `passphrase_disable()` to a single command where the on/off thing is a parameter
- create a new command `passphrase_new_command()` in `trezorctl.py`, that will be aliased to `enable-passphrase` AND `disable-passphrase`, and it will only tell the user to use "set passphrase on/off" and exit with an error.

matejcik: make sure to set `hidden=True` for the `passphrase_new_command`

it would be cool if the help page still only listed `[on|off]`
maybe something like:
```python
def argument_enable(default: bool = True, **kwargs) -> Callable:
	return click.argument("enable", Choice({"on": True, ...), metavar="[on|off]", default=default)

@click.command()
@argument_enable()
def wipe_code(enable, ...):
    ...
```

matejcik: alternately, to keep the backwards compatibility even more: `passphrase_new_command` could read its own name from context (`@click.pass_context`) and invoke the appropriate passphrase call (https://click.palletsprojects.com/en/8.0.x/advanced/#invoking-other-commands)

grdddj: I will see what I can do.

So far [01b9d18](https://github.com/trezor/trezor-firmware/pull/2123/commits/01b9d18232dbcbdd74310daa1ffd6f1cc98491de) should contain the unification to `on/off`

matejcik:
we are already using a custom alias implementation in `TrezorctlGroup`. instead of introducing a dependency, it might be easier to refactor that into a base `AliasedGroup`:
```python


class AliasedGroup(click.Group):
    """Command group that handles aliases and Click 6.x compatibility.
    
    Click 7.0 silently switched all underscore_commands to dash-commands.
    This implementation of `click.Group` responds to underscore_commands by invoking
    the respective dash-command.

    Supply an `aliases` dict at construction time to provide an alternative list of
    command names:

    >>> @click.command(cls=AliasedGroup, aliases={"do_bar", do_foo})
    >>> def cli():
    >>>     ...
    """

    def __init__(self, aliases: Dict[str, click.Command], *args, **kwargs) -> None:
        super().__init__(*args, **kwargs)
        self.aliases = aliases

    def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Command]:
        cmd_name = cmd_name.replace("_", "-")
        # try to look up the real name
        cmd = super().get_command(ctx, cmd_name)
        if cmd:
            return cmd

        # look for a backwards compatibility alias
        if cmd_name in self.aliases:
            return self.aliases[cmd_name]

        return None

class TrezorctlGroup(AliasedGroup):
    """Command group that handles compatibility for trezorctl.

    With trezorctl 0.11.5, we started to convert old-style long commands
    (such as "binance-sign-tx") to command groups ("binance") with subcommands
    ("sign-tx"). The `TrezorctlGroup` can perform subcommand lookup: if a command
    "binance-sign-tx" does not exist in the default group, it tries to find "sign-tx"
    subcommand of "binance" group.
    """

    def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Command]:
        cmd = super().get_command(ctx, cmd_name)
        if cmd:
            return cmd

        # the subsequent lookups rely on dash-separated command names
        cmd_name = cmd_name.replace("_", "-")
        # look for subcommand in btc - "sign-tx" is now "btc sign-tx"
        cmd = btc.cli.get_command(ctx, cmd_name)
        if cmd:
            return cmd

        # Old-style top-level commands looked like this: binance-sign-tx.
        # We are moving to 'binance' command with 'sign-tx' subcommand.
        try:
            command, subcommand = cmd_name.split("-", maxsplit=1)
            # get_command can return None and the following line will fail.
            # We don't care, we ignore the exception anyway.
            return super().get_command(ctx, command).get_command(ctx, subcommand)  # type: ignore ["get_command" is not a known member of "None";;Cannot access member "get_command" for type "Command"]
        except Exception:
            pass

        return None
```

grdddj:
> we are already using a custom alias implementation in `TrezorctlGroup`. instead of introducing a dependency, it might be easier to refactor that into a base `AliasedGroup`:

I am unable to use the abovementioned, even after some modifications. How did you want to use this on our `passphrase` group and its two commands? I am probably trying to use it at the wrong place

(below is the original state)

```python
@cli.group()
def passphrase() -> None:
    """Enable, disable or configure passphrase protection."""
    # this exists in order to support command aliases for "enable-passphrase"
    # and "disable-passphrase". Otherwise `passphrase` would just take an argument.


@passphrase.command(name="enabled")
@click.option("-f/-F", "--force-on-device/--no-force-on-device", default=None)
@with_client
def passphrase_enable(client: "TrezorClient", force_on_device: Optional[bool]) -> str:
    """Enable passphrase."""
    return device.apply_settings(
        client, use_passphrase=True, passphrase_always_on_device=force_on_device
    )


@passphrase.command(name="disabled")
@with_client
def passphrase_disable(client: "TrezorClient") -> str:
    """Disable passphrase."""
    return device.apply_settings(client, use_passphrase=False)
```

matejcik:
while you're touching this, please change our requirements.txt and setup.py so that `click>=7,<8.1` because there's at least one breaking change in 8.1 that will need to be fixed

grdddj:
Thanks, I tried to do it in [444f4f4](https://github.com/trezor/trezor-firmware/pull/2123/commits/444f4f499372100184aa603568a0ca0019990797).

It is not perfect, as we need to define that `AliasedGroup` in `cli.__init__.py` to avoid circular dependencies between `trezorctl.py` and `settings.py`. We cannot define `TrezorctlGroup` there as well, as it depends on `btc.cli`, which is lives only in `trezorctl.py` (we could maybe inject it into `TrezorctlGroup`'s construction).

In the (only) usage of `TrezorctlGroup`, `pyright` was also complaining about some things, so I silenced them - but there could be also some better solution.

> while you're touching this, please change our requirements.txt and setup.py so that `click>=7,<8.1` because there's at least one breaking change in 8.1 that will need to be fixed

Done also in [444f4f4](https://github.com/trezor/trezor-firmware/pull/2123/commits/444f4f499372100184aa603568a0ca0019990797). Does it deserve a commit-on-its-own?

matejcik:
> Does it deserve a commit-on-its-own?

yes please.

please also try updating Poetry's click to 8.0.x (setting `click = "^8.0"` in pyproject.toml and then `poetry update click` should do it), remove `types-click` package, and see if some of the type errors don't go away.
if this works, you can put it in the same commit as the requirements change.

> is not perfect, as we need to define that `AliasedGroup` in `cli.__init__.py`

this is a perfectly reasonable solution: `AliasedGroup` is a thing that the whole of `trezorlib.cli` can use, while `TrezorctlGroup` is specific to the main `trezorctl` definition file.

grdddj:
Force-pushed with two complete commits.

The click-update did not resolve any type-issues.

We might have introduced one incompatible change in `SD-card` handling, with changing `disable/enable` into `off/on`
```python
SD_PROTECT_OPERATIONS = {
    "on": messages.SdProtectOperationType.ENABLE,
    "off": messages.SdProtectOperationType.DISABLE,
    "refresh": messages.SdProtectOperationType.REFRESH,
}
```

Would it be worth leaving both options there? (or potentially creating that dict connecting `enable/enabled/on` with `True` and the opposite with `False`)

```python
SD_PROTECT_OPERATIONS = {
    "enable": messages.SdProtectOperationType.ENABLE,
    "disable": messages.SdProtectOperationType.DISABLE,
    "on": messages.SdProtectOperationType.ENABLE,
    "off": messages.SdProtectOperationType.DISABLE,
    "refresh": messages.SdProtectOperationType.REFRESH,
}

```

grdddj:
`CI` has gone completely wild (even after a manual restart), all with the same message as here - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2192793228 - failing to do `nix-shell --run "poetry install"` because of `No such file or directory: '/builds/satoshilabs/trezor/trezor-firmware/.venv/bin/pip'`

Not sure what is the cause, but the most probable is the change in `poetry.toml` and `poetry.lock`, which forced the installation to start from scratch, while it was cached previously. It might mean some problem on the `CI` server.

What is also strange, is the new `package.dependencies` section in ffdc4c3, which has some mention about `Windows` and `python < 3.8`

Doing `poetry install` locally is fine

matejcik:
> The click-update did not resolve any type-issues.

That is because you didn't remove `types-click`. I see locally that the type:ignores no longer apply without it.

grdddj:
> That is because you didn't remove `types-click`. I see locally that the type:ignores no longer apply without it.

Right, after deleting it, I have these type-results:
```
WARNING: there are unused ignores!
File /home/jmusil/trezor-firmware/python/src/trezorlib/cli/trezorctl.py:131 has unused ignore. Substring: Untyped function decorator obscures type of function; ignoring decorator
File /home/jmusil/trezor-firmware/python/src/trezorlib/cli/trezorctl.py:134 has unused ignore. Substring: No parameter named "aliases"

ERROR: We have issues!

/home/jmusil/trezor-firmware/python/src/trezorlib/cli/__init__.py:44: - error: Argument of type "dict_keys[str, Any]" cannot be assigned to parameter "choices" of type "Sequence[str]" in function "__init__"
  "dict_keys[str, Any]" is incompatible with "Sequence[str]" (reportGeneralTypeIssues)
```

grdddj:
There seems to be a mismatch between `GitHub Actions` and `Gitlab CI` in the two `pyright` errors in the changed `trezorctl.py`. Locally I also get those errors (like in GH actions), even after deleting the `.venv` and `poetry install`. Probably even `nix-shell` restart will be needed

matejcik (changes requested):
found two more small problems.

also pushed a commit that resolves the github / gitlab CI discrepancy, so please fix the type problem

python/src/trezorlib/cli/settings.py
```diff
@@ -218,7 +237,7 @@ def safety_checks(
 
 
 @cli.command()
-@click.argument("enable", type=ChoiceType({"on": True, "off": False}))
+@click.argument("enable", type=ChoiceType({"on": True, "off": False}), default=True)
```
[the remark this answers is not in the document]
grdddj: Deleted in [8f312e9](https://github.com/trezor/trezor-firmware/pull/2123/commits/8f312e994237053703dc5bb729ecce90247152d3)

python/src/trezorlib/cli/settings.py
```diff
@@ -89,30 +89,49 @@ def image_to_tt(filename: str) -> bytes:
...
+
+
 @click.group(name="set")
 def cli() -> None:
     """Device settings."""
 
 
 @cli.command()
-@click.option("-r", "--remove", is_flag=True)
+@click.option("-r", "--remove", is_flag=True, help="DEPRECATED, use on/off argument")
```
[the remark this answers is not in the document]
grdddj: Changed in [8f312e9](https://github.com/trezor/trezor-firmware/pull/2123/commits/8f312e994237053703dc5bb729ecce90247152d3)

grdddj:
> also pushed a commit that resolves the github / gitlab CI discrepancy, so please fix the type problem

Done in [8f312e9](https://github.com/trezor/trezor-firmware/pull/2123/commits/8f312e994237053703dc5bb729ecce90247152d3). Good finding there!

grdddj:
Testing if GitHub comments are working.

Seems CI is green everywhere, I will force-push when it will be possible.

PR #1225, opened 2020-08-25
extended sign_message to also sign bytes, not only string

fametrano:
Bitcoin message signing is usually used to sign an input string which, as first step, is encoded to bytes. The rest of the implementation assumes bytes, operates on bytes, ecdsa-signs bytes.

This PR proposes to skip the string encoding if bytes, instead of a string, are provided as input: this allow to "bitcoin message sign" an arbitrary byte sequence (note that arbitrary byte sequences cannot be always decoded to strings).

See also https://github.com/bitcoin-core/HWI/pull/366

PR #2126, opened 2022-02-14
feat(python): extend sign_message to also sign bytes, not only string

grdddj:
Finalizing PR https://github.com/trezor/trezor-firmware/pull/1225:
- just some renaming and docs changes

### DIFF
--- a/python/.changelog.d/2126.changed
+++ b/python/.changelog.d/2126.changed
@@ -1,0 +1,1 @@
+Rename `normalize_nfc` to `prepare_message_bytes` in tools.py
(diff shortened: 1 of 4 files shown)


matejcik:
please add a changelog entry about the rename

grdddj:
> How do we show raw bytes message on the display?

The `CLI` can call the `btc.sign_message` only with string parameter, so somebody would need to call it themselves with bytes. I tried to simulate it with random bytes (`os.urandom(16)`) and it seems to be printing hex nicely:

image

prusnak:
> it seems to be printing hex nicely:

Nice. How about T1?

matejcik:
> Nice. How about T1?

https://github.com/trezor/trezor-firmware/blob/update-deps/legacy/firmware/fsm.c#L334-L340

prusnak (approved):

matejcik (approved):

PR #2127, opened 2022-02-14
feat(ci): add quick style checks suitable for git hooks

grdddj:
Creating quick-to-finish `make` targets to check the basic python style (`black` and `isort`), so that they could be used as git hooks (mostly `pre-commit` in this case) or a one-line CLI commands.

I would personally find this useful - it could reduce the number of fixup commits because of forgetting to run the style scripts locally. Not sure if it justifies a small duplication in `make` targets.


matejcik:
presumably @prusnak is using them as he authored the code :)

Code LGTM, i have no problem adding this if you find it useful. Please merge unless you want to add more to it.

WRT changelog check, that is somewhat problematic because in many cases you don't have the issue ID until you make a PR. But of course you can always skip the hooks so maybe? :woman_shrugging:

grdddj:
> Code LGTM, i have no problem adding this if you find it useful. Please merge unless you want to add more to it.

I think that is enough for now. I force-pushed with the `no-changelog` commit message (I assume merging that would break the `master`?) 

I have never merged anything myself, do we use `Squash and merge` or `Rebase and merge`? They seem similar, but the `Rebase` one looks better, also when I did not rebase on `master` locally

matejcik:
> I think that is enough for now. I force-pushed with the `no-changelog` commit message (I assume merging that would break the `master`?)

master branch doesn't run the changelog check -- it can't really, because changelog entries are applicable across a PR / branch.

> do we use `Squash and merge` or `Rebase and merge`

standard is "Rebase and merge"

"Squash and merge" is when you want to throw away the PR history completely and merge a single commit with all the changes, so we occassionally use that as well.

Issue #2110, opened 2022-02-09
Consider enabling USB by default on TT


hynek-jina:
USB will not automatically turn on when the device is plugged in (when using a PIN). Suite does not recognise then that TT is connected.

- [ ] Investigate pro and cons of  "semi-enabling"

Initial internal discussion: [Slack](https://satoshilabs.slack.com/archives/CL1D61PQF/p1643633067863079)

Issue #1026, opened 2020-05-27
trezorctl claiming "Failed to find device" when the device is in use by Suite


matejcik:
trezorctl coerces any errors while connecting to device into "Failed to find device".
Particularly in case of Bridge, we can detect that the device is acquired by a different user (by the presence of `session` field in enumeration) and show an appropriate response.

This also affects `trezorctl list`, which errors out in presence of an already acquired device.

Bonus points for implementing a command that force-releases the device. This could be merged with `trezorctl usb-reset` #225

tsusanka:
@szymonlesisz is this something high-priority for you?

STew790:
What can be done here by QA?

matejcik:
QA scenario:

1. Open Trezor Suite, go to Receive tab, show an address on Trezor.
2. **While the address is on screen** and Suite is waiting for user response, test the following:
3. run `trezorctl list`. You should not see a crash. Trezorctl should correctly report that device is in use.
4. try several different `trezorctl` commands. In each case, trezorctl should correctly respond that the device is in use.

STew790:
**QA NOK**
I tried all the listed commands on our wiki for trezorctl and here are the ones that show `Failed to find a Trezor device`
```
trezorctl get-address
trezorctl backup-device
trezorctl disable-passphrase
trezorctl reset-device
trezorctl set safety-checks
trezorctl get-features
trezorctl set-homescreen
```
**QA OK** 
All the other ones shows correctly `Device is in use by another process.`

The steps where the same as you mentioned.

matejcik:
Thanks for the report. That seems to be hitting a slightly different issue, so I'm going to move this one to approved.

Issue #973, opened 2020-04-30
Can't shut down emulator while waiting for a ButtonAck


matejcik:
While stuck waiting on `ctx.call()`, emulator does not respond to SIGINT or SIGTERM and must be SIGKILLed to shut it down.

bosomt:
@matejcik just emulator stuff, no need to test it on our side, right ?

matejcik:
correct!

PR #2129, opened 2022-02-15
WIP - opening empty USB connection before PIN

grdddj:
First reproducible results and signs of something connected with https://github.com/trezor/trezor-firmware/issues/2110:
- opening the USB even before user unlocks device with PIN

It seems that the USB is really open even before unlocking, however, something is still not right. Below are logs from some usb-tools and commands I found and tried to use:

(tested on a physical Trezor with the firmware in the state after this one commit)

----

**dmesg** command:
.. we connect the Trezor into USB and the locked screen is shown...
```
[161586.338683] usb 1-4: new full-speed USB device number 73 using xhci_hcd
[161586.466408] usb 1-4: config 1 has no interfaces?
[161586.467058] usb 1-4: New USB device found, idVendor=1209, idProduct=53c1, bcdDevice= 2.00
[161586.467064] usb 1-4: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[161586.467067] usb 1-4: Product: TREZOR
[161586.467070] usb 1-4: Manufacturer: SatoshiLabs
[161586.467073] usb 1-4: SerialNumber: 355C817510C0EABF2F147145
```
... we unlock the device via PIN...
```
[161614.583391] usb 1-4: USB disconnect, device number 73
[161614.887514] usb 1-4: new full-speed USB device number 74 using xhci_hcd
[161615.017634] usb 1-4: New USB device found, idVendor=1209, idProduct=53c1, bcdDevice= 2.00
[161615.017642] usb 1-4: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[161615.017647] usb 1-4: Product: TREZOR
[161615.017650] usb 1-4: Manufacturer: SatoshiLabs
[161615.017653] usb 1-4: SerialNumber: E52E08CF6B65FFE5E00B0167
[161615.024247] hid-generic 0003:1209:53C1.0033: hiddev96,hidraw4: USB HID v1.11 Device [SatoshiLabs TREZOR] on usb-0000:00:14.0-4/input2
[161615.024779] cdc_acm 1-4:1.3: ttyACM0: USB ACM device
```

The log is almost same in both cases, the first one just contains `config 1 has no interfaces?` (maybe we need to create some - there could be nothing by default). The second has extra `hid-generic` and `cdc_acm` lines with binging to `ttyACM0`, which seems to be a sign of successful connection - after that the device is visible on USB and can be used by `trezorctl`.

---------------

**find_usb_device.sh** script:
```bash
for sysdevpath in $(find /sys/bus/usb/devices/usb*/ -name dev); do
    (
        syspath="${sysdevpath%/dev}"
        devname="$(udevadm info -q name -p $syspath)"
        [[ "$devname" == "bus/"* ]] && exit
        eval "$(udevadm info -q property --export -p $syspath)"
        [[ -z "$ID_SERIAL" ]] && exit
        echo "/dev/$devname - $ID_SERIAL"
    )
done
```
after unlocking the device, one of the lines says 
```
/dev/ttyACM0 - SatoshiLabs_TREZOR_E52E08CF6B65FFE5E00B0167
```
, which shows the device is connected - it does not show it before unlocking

----
**lsusb** command:
shows the same line even before unlocking the device - which is a good sign
```
Bus 001 Device 075: ID 1209:53c1 Generic SatoshiLabs TREZOR
```


matejcik:
what does `trezorctl list -n` show in the locked state? how about `trezorctl list` without `-n`?

core/src/main.py
```diff
@@ -32,9 +32,9 @@
 # as a UI callback for storage, which can be invoked at any time
 import trezor.pin  # noqa: F401
 
-# === Prepare the USB interfaces first. Do not connect to the host yet.
-# usb imports trezor.utils and trezor.io which is a C module
-import usb
```
matejcik, line 37: for the next step, please make sure that the original `import usb` statement is kept as it was.

instead, modify `usb.py` to not add interfaces automatically, perhaps create a function `usb.add_interfaces()` which will be called from `main.py` before the second `open()`

matejcik: note that `usb.bus` as well as `usb.iface_*` should remain global, so that code can import them

grdddj:
> what does trezorctl list -n show in the locked state? how about trezorctl list without -n?

 ```
 tr list                          
Failed to enumerate WebUsbTransport. IndexError: No such interface: 0
  tr list -n
Failed to enumerate WebUsbTransport. IndexError: No such interface: 0

... unlocking the device ...

  tr list   
bridge:23 - My Trezor [Trezor T, DEDDF5B7F35EBBE7674A5334]
webusb:001:2 - My Trezor [Trezor T, DEDDF5B7F35EBBE7674A5334]
  tr list -n
webusb:001:2
bridge:23
```

**Comment by matejcik** [comment]
> Failed to enumerate WebUsbTransport. IndexError: No such interface: 0

where exactly does the error come from?

**Comment by grdddj** [comment]
> where exactly does the error come from?

Looks like from here:
https://github.com/trezor/trezor-firmware/blob/master/python/src/trezorlib/transport/__init__.py#L140

**Comment by matejcik** [comment]
> Looks like from here:

i'm more interested in where the exception is raised, not where it is reported

**Comment by grdddj** [comment]
> i'm more interested in where the exception is raised, not where it is reported

Traceback after not catching the exception:

```
  File "/home/jmusil/trezor-firmware/.venv/bin/trezorctl", line 33, in <module>
    sys.exit(load_entry_point('trezor', 'console_scripts', 'trezorctl')())
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/cli/trezorctl.py", line 247, in list_devices
    for transport in enumerate_devices():
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/transport/__init__.py", line 133, in enumerate_devices
    found = list(transport.enumerate(models))
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/transport/webusb.py", line 134, in enumerate
    if not is_vendor_class(dev):
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/transport/webusb.py", line 162, in is_vendor_class
    dev[configurationId][INTERFACE][altSettingId].getClass()
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/usb1/__init__.py", line 1581, in __getitem__
    raise IndexError('No such interface: %r' % (interface, ))
IndexError: No such interface: 0
```

**Comment by matejcik** [comment]
I see, thanks. This _could_ be fixable in trezorlib. That is possibly going to be a sub-issue -- make trezorlib and trezorctl capable of detecting "Trezor in locked state".

For now this looks good enough, please continue in this direction. Also please test how Bridge behaves with this device. It will _probably_ not be detected, the important part is, (a) will Bridge continue to work when you unlock the device? and (b) will Bridge detect another Trezor if there is also a locked one? (trezorctl will fail, as we have seen)

cc @mroz22 for Suite compatibility -- what do you think about this? Can you advise about webusb compatibility? We will probably need to modify Bridge, right?

**Comment by matejcik** [comment]
if this would be problematic, we could instead:
* set up an interface with no endpoints -- see https://github.com/trezor/trezor-firmware/blob/master/core/embed/trezorhal/usb_webusb-impl.h#L56
* set up a normal webusb interface without listening on it

File: core/src/main.py (hunk 1 above)
Comment (grdddj) [reply core/src/main.py:37]: When deleting the `usb_empty` and working just with the one `usb` (state below), I am again getting the Trezor red device with `FATAL ERROR ... core/embed/trezorhal/usb.c:104 ... check_desc_str(dev_info->manufacturer)` after unlocking the device. 

It could be necessary to also recreate the `bus` object in `add_interfaces()` - will try that

```python
# usb imports trezor.utils and trezor.io which is a C module
import usb
# opening empty usb
usb.bus.open("355C817510C0EABF2F147145")  # random 12byte hex

# create an unimport manager that will be reused in the main loop
unimport_manager = utils.unimport()

# unlock the device, unload the boot module afterwards
with unimport_manager:
    import boot
    del boot

# closing the empty USB, adding the interfaces and starting it again
usb.bus.close()
usb.add_interfaces()

import storage.device

usb.bus.open(storage.device.get_device_id())

```

grdddj: Recreating the `bus` helped to get rid of Trezor red screen - and it behaves the same as before - changes in [2e7553d](https://github.com/trezor/trezor-firmware/pull/2129/commits/2e7553d0f1f1c2ffb977596fcf346740c5491421)

matejcik: ah, I see the problem. the instance is not reopenable, per https://github.com/trezor/trezor-firmware/blob/grdddj/scriptpubkey-check-tests-segwit-multisig/core/embed/extmod/modtrezorio/modtrezorio-usb.h#L243-L262

please try to remove the `o->info` clearing (except `serial_number` should be cleared because that is set again by `open()`)

grdddj:
> Also please test how Bridge behaves with this device. It will probably not be detected, the important part is, (a) will Bridge continue to work when you unlock the device? and (b) will Bridge detect another Trezor if there is also a locked one? (trezorctl will fail, as we have seen)

a) Unlocked Trezor works fine with Bridge, I can use Suite, see funds on `all all` seed etc.
b) When I connect a `T1` (while having unlocked `TT` with current firmware), Bridge finds the `T1` immediately, and Suite will prompt me for PIN.

Probably not a surprise for you, but when the `T1` was waiting for PIN, `trezorctl list` was giving the error below:
```
Failed to enumerate WebUsbTransport. IndexError: No such interface: 0
  File "/home/jmusil/trezor-firmware/.venv/bin/trezorctl", line 33, in <module>
    sys.exit(load_entry_point('trezor', 'console_scripts', 'trezorctl')())
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jmusil/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/cli/trezorctl.py", line 248, in list_devices
    client = TrezorClient(transport, ui=ui.ClickUI())
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/client.py", line 126, in __init__
    self.init_device(session_id=session_id, derive_cardano=derive_cardano)
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/tools.py", line 284, in wrapped_f
    client.open()
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/client.py", line 130, in open
    self.transport.begin_session()
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/transport/bridge.py", line 153, in begin_session
    data = self._call("acquire/" + self.device["path"])
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/transport/bridge.py", line 138, in _call
    return call_bridge(uri, data=data)
  File "/home/jmusil/trezor-firmware/python/src/trezorlib/transport/bridge.py", line 47, in call_bridge
    raise TransportException(error_str)
trezorlib.transport.TransportException: trezord: acquire/27/null failed with code 400: wrong previous session
```

After putting PIN on `T1`, the output is
```
Failed to enumerate WebUsbTransport. IndexError: No such interface: 0
bridge:27 - My Trezor [Trezor 1, 044B63FAA3AFABD5A01694DD]
```

matejcik:
> ```
> trezorlib.transport.TransportException: trezord: acquire/27/null failed with code 400: wrong previous session
> ```

this is an instance of https://github.com/trezor/trezor-firmware/issues/1026

core/src/main.py
grdddj: Done in 5b9e37b - seems to be fine generally.

There is one difference from the previous state, there are some additional "reset" events going on with the device (output from the `dmesg` command)

```
[236266.736821] usb 1-2: USB disconnect, device number 98
[236267.042219] usb 1-2: new full-speed USB device number 99 using xhci_hcd
[236267.170750] usb 1-2: New USB device found, idVendor=1209, idProduct=53c1, bcdDevice= 2.00
[236267.170756] usb 1-2: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[236267.170760] usb 1-2: Product: TREZOR
[236267.170762] usb 1-2: Manufacturer: SatoshiLabs
[236267.170765] usb 1-2: SerialNumber: 8B70B8E9135C6E17B7322ACC
[236267.177926] hid-generic 0003:1209:53C1.003E: hiddev96,hidraw4: USB HID v1.11 Device [SatoshiLabs TREZOR] on usb-0000:00:14.0-2/input2
[236267.178439] cdc_acm 1-2:1.3: ttyACM0: USB ACM device
[236267.741257] usb 1-2: reset full-speed USB device number 99 using xhci_hcd
[236267.870002] cdc_acm 1-2:1.3: ttyACM0: USB ACM device
[236268.630570] usb 1-2: reset full-speed USB device number 99 using xhci_hcd
[236268.759911] cdc_acm 1-2:1.3: ttyACM0: USB ACM device
```

grdddj:
Good news from Windows:
- the system sees the connected device already before unlocking - and as soon as we unlock, Suite recognizes it and all seems to be fine
- also, when manually "ejecting" the device before unlocking, and then unlocking, it will connect nicely

With @sime we are trying to test it on MAC

sime:
Tested on OS X 12.2.1 on MacBook Air (M1, 2020).

* Checked out `grdddj/tt_enable_usb_by_default`, built and uploaded custom firmware.
* Set PIN
* Reconnected device.

Whilst device is 'not connected'.
```
$ trezorctl list
Failed to enumerate WebUsbTransport. IndexError: No such interface: 0
```

After device is connected.
```
$ trezorctl list
webusb:000:1:3:3:3 - (unnamed) [Trezor T, 63350A9F66CBDCB58A4AD8FC]
```

## Running Suite

Tracebacks occur when using `trezorctl` and the suite is open.

<details><summary>On passphrase prompt</summary><p>

```
$ trezorctl list
  File "/Users/smales/Development/trezor-firmware/.venv/bin/trezorctl", line 33, in <module>
    sys.exit(load_entry_point('trezor', 'console_scripts', 'trezorctl')())
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/cli/trezorctl.py", line 248, in list_devices
    client = TrezorClient(transport, ui=ui.ClickUI())
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/client.py", line 126, in __init__
    self.init_device(session_id=session_id, derive_cardano=derive_cardano)
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/tools.py", line 284, in wrapped_f
    client.open()
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/client.py", line 130, in open
    self.transport.begin_session()
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/transport/bridge.py", line 153, in begin_session
    data = self._call("acquire/" + self.device["path"])
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/transport/bridge.py", line 138, in _call
    return call_bridge(uri, data=data)
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/transport/bridge.py", line 47, in call_bridge
    raise TransportException(error_str)
trezorlib.transport.TransportException: trezord: acquire/1/null failed with code 400: wrong previous session
```
</p></details>
<details><summary>After entering passphrase</summary><p>

```
$ trezorctl list
bridge:1 - (unnamed) [Trezor T, 63350A9F66CBDCB58A4AD8FC]
  File "/Users/smales/Development/trezor-firmware/.venv/bin/trezorctl", line 33, in <module>
    sys.exit(load_entry_point('trezor', 'console_scripts', 'trezorctl')())
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/cli/trezorctl.py", line 248, in list_devices
    client = TrezorClient(transport, ui=ui.ClickUI())
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/client.py", line 126, in __init__
    self.init_device(session_id=session_id, derive_cardano=derive_cardano)
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/tools.py", line 284, in wrapped_f
    client.open()
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/client.py", line 130, in open
    self.transport.begin_session()
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/transport/protocol.py", line 117, in begin_session
    self.protocol.begin_session()
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/transport/protocol.py", line 85, in begin_session
    self.handle.open()
  File "/Users/smales/Development/trezor-firmware/python/src/trezorlib/transport/webusb.py", line 53, in open
    self.handle = self.device.open()
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/usb1/__init__.py", line 2034, in open
    mayRaiseUSBError(libusb1.libusb_open(self.device_p, byref(handle)))
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/usb1/__init__.py", line 127, in mayRaiseUSBError
    __raiseUSBError(value)
  File "/Users/smales/Development/trezor-firmware/.venv/lib/python3.9/site-packages/usb1/__init__.py", line 119, in raiseUSBError
    raise __STATUS_TO_EXCEPTION_DICT.get(value, __USBError)(value)
usb1.USBErrorNoDevice: LIBUSB_ERROR_NO_DEVICE [-4]
```

</p>
</details>

core/src/usb.py
```diff
@@ -29,79 +29,92 @@
...
-id_wire = next(_iface_iter)
-iface_wire = io.WebUSB(
-    iface_num=id_wire,
-    ep_in=0x81 + id_wire,
-    ep_out=0x01 + id_wire,
-    emu_port=UDP_PORT + WIRE_PORT_OFFSET,
-)
-bus.add(iface_wire)
+# All the interfaces are (conditionally) created and added in add_interfaces()
+iface_wire: io.WebUSB | None = None
```
matejcik, line 33: AFAIK creating the interfaces (without calling `bus.add()`) should be OK, so please create the interfaces on top level as before, and only add them in the `add_interfaces()` function.

grdddj: Sure. Small problem is that in `add_interfaces()` we will need to check which interfaces are defined.

We could use the same if-statements as when defining the interfaces, even though that would be some duplication. Probably still cleaner than checking something like `if  "iface_debug" in globals():`

grdddj: Done in [3822719](https://github.com/trezor/trezor-firmware/pull/2129/commits/38227199375ace354525361b2879f8b7eee43bc4)

matejcik: yes, reusing the same ifs is OK. that is how the code looked sometime ago actually :)

grdddj:
In [20e2041](https://github.com/trezor/trezor-firmware/pull/2129/commits/20e204163e0d15901440bd323c3ba32d14ca874d) I am initializing the `config` in `main.py` before the device is unlocked, so that we can call `storage.device.get_device_id()`. The ID of the USB connection is then the same before and after unlocking.

(Little surprisingly for me) the device ID is the same even after disconnecting and connecting again. Is that expected (and the ID is therefore persistently stored in storage)? It would make sense, as clients (Suite for example) can then recognize it is the same device it already knows.

@matejcik you were mentioning there is one special situation when the ID is not there, what situation is that? (Maybe when the device is "fresh" out of production)

matejcik:
> Is that expected (and the ID is therefore persistently stored in storage)?

Yes, the ID is randomly generated when you _first_ ask for it, and then saved in storage for subsequent access. It resets when you wipe Trezor

> what situation is that?

When Trezor is starting with uninitialized storage. You can reproduce it best with bootloader wipe:
- go to bootloader
- `trezorctl device wipe --bootloader`
- `trezorctl fw update ...`
- now when Trezor is booting, the storage is empty and the device ID must be generated

grdddj:
> * trezorctl device wipe --bootloader

I tried this with then uploading the custom firmware, and Trezor was not very happy - red screen #f03c15 saying 
```
Internal error
(HF) 

Please unplug the device
```

I will try to detect this condition and not do the ID setting in this case

matejcik:
Please investigate why this doesn't work before boot.py but it does work after.

I suspect that boot.py unlocks with empty PIN and then is able to set the device ID. We should probably handle this inside `get_device_id()`, or maybe in some other auxilliary function.

grdddj:
When trying to unlock the `storage` and `config` in `get_device_id()` (before setting the `dev_id`), I found some complications:

(my assumption is that I need to call `storage.init_unlocked() ` and `config.unlock(pin, salt)` to unlock)

- in order to get the `salt` properly, we need to call `request_sd_salt`, which is `async`, and therefore the whole **`get_device_id()` would need to be `async`** - and I guess that would not be very nice (or would even make the function completely unusable under current conditions)
- calling `storage.init_unlocked()` and `config.unlock("", None)` seemingly unlocked the storage on `emulator`, but then the emulator became completely unresponsive and had to be force-quitted
  - here it seemed that we cannot call those "unlocking" commands twice, but even when I commented them in `boot.py`, where they are called afterwards, it did not help
  - and theoretically even if it would help, it would be quite hard to "not call it again"
 
Conclusion is that I do not know (yet) how to unlock the config/storage in `get_device_id` as we agreed we will do

matejcik:
does it respond to trezorctl commands? such as `get-features` ?

grdddj:
> does it respond to trezorctl commands? such as `get-features` ?

Yes, it responds correctly to `trezorctl ping` and `trezorctl get-features`. When trying something UI-based, like `trezorctl device setup`, it seems to resurrect sometimes, but still is very unstable.

I am about to try it on the fresh hardware device

matejcik:
please push your current unstable version so that I can test it

> we could solve it (IMO) more easily without messing up with unlocking.

Maybe, but that is layering complexity to avoid an underlying problem. This makes the code slightly more fragile, and also the "underlying problem" here shouldn't exist in the first place, so I would strongly prefer figuring it out instead of papering over it.
Even if we end up doing something like what you suggest, it needs an explanatory comment better than "if we don't do this, the UI gets really unstable".

grdddj:
> please push your current unstable version so that I can test it

The basic changes are done in [e0e6b2b](https://github.com/trezor/trezor-firmware/pull/2129/commits/e0e6b2b282e52ab42796382566cdfd709e1fd422)

However, I found that emulator freezes even when being in previous state. I just ran `make clean` with the hope it was just something bad with cache

grdddj:
Rebasing on `master`, it should solve some of the `UI` tests failing in `CI`

grdddj:
In the current state - after rebasing on `master` - the steps to replicate the "unstable UI" are still the same:
- running the emulator with `-e`, which simulates the situation when the `device_id` is not there and we are trying to unlock the config in `get_device_id()`

matejcik:
the behavior I'm seeing is: UI works fine, except the final screen of a workflow hangs:
image
when the screen is stuck in this state, the emulator doesn't respond
however, if I press Ctrl+C on the emulator, and then issue a command which touches the UI (e.g. GetAddress(show_display=True) ), the UI will respond again, ctrl+c is processed and emulator cleanly shuts down.

can you confirm?

matejcik:
fwiw, the "stuck state" seems to be an instance of #973

grdddj:
> can you confirm?

I got something very similar - final screen is hanging.

My steps:
```sh
./emu.py -e -s   # wiping and then loading all all seed
<hangs on the screen 1 - impossible to click the green button>
trezorctl btc get-address -n 44h/0h/0h/0/0 -d
<click the green button>
<hangs on screen 2>
<press Ctrl+c in emulator's terminal two times>
trezorctl btc get-address -n 44h/0h/0h/0/0 -d
<it shows address on device and shuts down>
```
image
image

matejcik:
> ./emu.py -e -s   # wiping and then loading all all seed
> <hangs on the screen 1 - impossible to click the green button>

This is an identical situation -- observe that if you run a functioning emulator, you're not clicking anything, the answer is sent via debuglink. The difference is, if you click manually, the last screen refresh is with the button highlighted; with debuglink answer, it is not.

I tracked it down to the `config.init()` call in main.py. If I remove the argument, or supply some dummy function, everything works. Only using `show_pin_timeout` causes the weirdness. I'm investigating deeper.

grdddj:
> I tracked it down to the `config.init()` call in main.py. If I remove the argument, or supply some dummy function, everything works. Only using `show_pin_timeout` causes the weirdness. I'm investigating deeper.

I am confirming that deleting the `show_pin_timeout` argument gets rid of the problem.

My only guess so far is that `from trezor import ui` got imported "too soon"

matejcik:
In the meantime, it turns out that working with webusb is probably also a requirement.

Please check how the current solution behaves with webusb -- i.e., try using web Suite and "find Trezor" before it is unlocked. I expect that Trezor will not be shown. If that is the case, try hacking together something where we expose the interface but don't give it any endpoints (see [above](https://github.com/trezor/trezor-firmware/blob/master/core/embed/trezorhal/usb_webusb-impl.h#L56))
In the most basic variant, just modify the number of endpoints to be 0, and see how that device is detected on various OSes.

For next steps (if the results are promising), we'll probably need to indicate somehow whether the WebUSB endpoints should be registered. First idea is: `usb.bus.open(silent=True)` only works if the first interface is WebUSB (raises otherwise), only registers the first interface, and adds an argument to `usb_webusb_add()` so that it won't register the endpoints

matejcik:
> My only guess so far is that `from trezor import ui` got imported "too soon"

this is the case, yes; in fact, adding `import trezor.ui` anywhere before the first `with unimport_manager` block will cause the issue.

I'm looking into what is going wrong. I suspect it's something to do with a copy of the module incompletely unimported at some point. We have to be **super careful** with every single import line that happens in the main initialization path.
(in fact importing `trezor.config` top level in `storage/device.py` is a bad idea, although it seems to work in practice. you should move it inside the `get_device_id` function, and preferably import it as `trezorconfig` instead of `trezor.config`)

grdddj:
> Please check how the current solution behaves with webusb -- i.e., try using web Suite and "find Trezor" before it is unlocked. I expect that Trezor will not be shown. If that is the case, try hacking together something where we expose the interface but don't give it any endpoints (see [above](https://github.com/trezor/trezor-firmware/blob/master/core/embed/trezorhal/usb_webusb-impl.h#L56))
> In the most basic variant, just modify the number of endpoints to be 0, and see how that device is detected on various OSes.

(Not really knowing what happens on the background)
When I put `0` there instead of `2`, before unlocking it was not recognized on Windows desktop Suite nor on Linux in web Suite. Also, after unlocking, Windows was complaining that `device is not working correctly` and web Suite says `Device detected in incorrect state`

grdddj:
> what does trezorctl say about the (unlocked) device?

`trezorctl` (on Linux) does not say much. `trezorctl list` fails on `usb1.USBErrorBusy: LIBUSB_ERROR_BUSY [-6]` and `trezorctl get-features` gives `Failed to find a Trezor device.`

Suite screen after clicking `Use trezor here`:
image

The full Windows error popup:
```
USB device not recognized
The last USB device you connected to this computer malfunctioned, and WIndows does not recognize it.
```

grdddj:
> Just to make sure, please comment out all other interfaces except the main webusb one.

When I have just `bus.add(iface_wire)` in `add_interfaces()`, the results are the same.
(still having `d->iface.bAlternateSetting = 0;` )

hiviah:
I randomly found out that when using this firmware with Pixel 4a phone (stock ROM), that the phone unpowers TT after PIN is entered and verified, just when the USB descriptors change. I think it refers to the USB spec part where host may unpower device if it enables both pullups/pulldowns (looking for the precise part of the spec).

Since it does not happen with official FW, pretty sure it's linked to the change in USB.

hiviah:
I've tried the following:

1. pull on `PB15` (OTG_HS_DP) D+ pin to 0 for 5 ms instead of stopping USB fully
2. also commenting out `usb_deinit`
3. removing pulling on `PB15` completely, and also removing USB stop/deinit

Though if only second call to `usb_start` remains, it still causes unpower/reset on Pixel 4a. On other devices it seems to work as it should, device comes up with new address and it normally accessible via trezorctl after entering PIN, etc.

Looking at depth in the USB init code there is this though in `USB_DevConnect` in `stm32f4xx_ll_usb.c`:

`USBx_DEVICE->DCTL &= ~USB_OTG_DCTL_SDIS` (Clear soft disconnect - should not have any effect since we never cleared it though)

From the above I am inclined to believe that there is bug in the phone's USB stack handling, but since there is no TWRP for this model I can't really check various Android images.

matejcik:
> there is bug in the phone's USB stack handling

It would be interesting to find a different phone with Snapdragon 730G and see if the same bug manifests -- my guess is that it is not something that would randomly affect just a single model.

That is unfortunate, because I can't really see a workaround. Only thing the user can do is disable PIN, but if the phone is their only device, they won't be able to do that.

From the firmware side, we could have a setting to disable "early USB", and we can even detect the condition after the first time (set a storage value when USB goes up, reset it when it goes up the second time; if you're starting with the value set, do not start early USB).
Could we somehow detect the phone from the firmware side?

Furthermore, handling this would also complicate the minimal-responder variant, as we'd need to publish all interfaces the first time, including FIDO, so we'd have to have a separate minimal-responder for FIDO too.

hiviah:
I think only way to find out is using USB host mode on TT (not peripheral mode) so that you can see VID:PID of device, but on not sure how it's going to work with normal notebook (notebooks may not have peripheral mode, so...).

Using host mode also means you need to switch to peripheral mode which can trigger the same bug. Saving the value would probably make it just a single reboot, though. 

Secondly using host mode may require that host provides power, not sure about this how it is with USB-C, there are at least 5 different commonly used protocols (PD several versions, QC-2, QC-3, Samsung and Apple have their own). PD is the only standard one.

Hannsek:
Closing due to inactivity. If we want to do it anyway, we'll have to open new pr.

PR #2133, opened 2022-02-18
feat(core): headless emulator implies disabled animations

prusnak:
I was under impression that headless mode disables animation. However, this is not the case.

If we do not want this by default (for some reason I don't see/understand), feel free to close this PR.


matejcik (approved):
seems reasonable

PR #2138, opened 2022-02-19
Rust UI: improve messages returned to python

### LINKED ISSUE #2118: Improve messages returned from rust layouts

mmilata:
Built on top of #2092. Pyright fails since we no longer generate the `.pyi` files, can we do better than just ignoring these issues? Some other questions inline.

Fixes #2118.


core/embed/rust/src/ui/layout/result.rs
```diff
@@ -0,0 +1,34 @@
...
+        //  - There's nothing to mutate.
+        unsafe { Obj::from_ptr(self as *const _ as *mut _) }
+    }
+}
+
+// SAFETY: We are in a single-threaded environment.
+unsafe impl Sync for ResultObj {}
+
+static CONFIRMED_TYPE: Type = obj_type! { name: Qstr::MP_QSTR_CONFIRMED, };
+static CANCELLED_TYPE: Type = obj_type! { name: Qstr::MP_QSTR_CANCELLED, };
```
mmilata, line 27: Implemented this as two distinct types because it was the least comlex way to make python `print()` the value. Is there a reason not to do it like this.

core/embed/rust/src/ui/model_t1/layout.rs
```diff
@@ -60,12 +58,7 @@ extern "C" fn ui_layout_new_confirm_action(
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
-#[no_mangle]
-extern "C" fn ui_layout_new_confirm_text(
-    n_args: usize,
-    args: *const Obj,
-    kwargs: *const Map,
-) -> Obj {
+extern "C" fn new_confirm_text(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
```
mmilata, line 61: Note that I changed `*const Map` to `*mut Map` to pass type check.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
 };
 
 pub enum PinKeyboardMsg {
-    Confirmed,
+    Confirmed(Vec<u8, MAX_LENGTH>),
```
mmilata, line 21: Since the `pin()` method exists I guess it was supposed to be used to obtain the pin, but haven't figured out how to get a reference to the layout object at the time the result is constructed.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
```diff
@@ -145,7 +145,7 @@ impl Component for PinKeyboard {
 
     fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
         if let Some(Clicked) = self.confirm_btn.event(ctx, event) {
-            return Some(PinKeyboardMsg::Confirmed);
+            return Some(PinKeyboardMsg::Confirmed(self.digits.clone()));
```
mmilata, line 148: If we keep this, is there a way to zero out the pin being cloned in a way that doesn't get optimized out?

core/embed/rust/src/micropython/macros.rs
```diff
@@ -164,3 +170,23 @@ macro_rules! obj_type {
...
+        unsafe {
+            use $crate::micropython::ffi;
+
+            ffi::mp_obj_module_t {
+                base: ffi::mp_obj_base_t {
+                    type_: &ffi::mp_type_module,
+                },
+                // TODO: This is safe only if we pass in `Dict` with fixed `Map` (created by
+                // `Map::fixed()`, usually through `obj_map!`), because only then will
+                // MicroPython treat `globals` as immutable, and make the mutable cast safe.
```
mmilata, line 187: Would it be possible/difficult/ugly to just call the `obj_dict!(obj_map!(...))` from here instead?

core/embed/rust/src/ui/layout/result.rs
jpochyla: It takes a bit more flash but I wouldn't sweat it too much.

core/embed/rust/src/ui/model_t1/layout.rs
jpochyla: I can't check it now, but it should match the type the C-side expects.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
jpochyla: This is exactly the usecase the `ObjComponentMsg` trait (hope I remember the name right) is supposed to address.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
jpochyla: See above, we can get rid of PIN being stored in the message altogether. Re. reliable data zeroing, there are some crates, let's investigate how they work in a nostd env.

core/embed/rust/src/micropython/macros.rs
jpochyla: Not difficult, just ugly, we'd need to copy the field matchers from the macro definitions into this one. But I guess we should. Or we can maybe make a newtype for the static dicts, but I don't like it too much.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
jpochyla: If you omit the `TryInto` impl for the msg type, you should be able to impl the above trait and get a ref to the component for the conversion. Then you can pull the PIN out of it and convert into Obj.

jpochyla: If you omit the `TryInto` impl for the msg type, you should be able to impl the above trait and get a ref to the component for the conversion. Then you can pull the PIN out of it and convert into Obj.

mmilata: When I implement the trait like https://github.com/trezor/trezor-firmware/blob/31fc2a4270b01cdbd4d8aeb33b34ba45065c9cb6/core/embed/rust/src/ui/model_tt/layout.rs#L57-L65
then I get:
```
error[E0119]: conflicting implementations of trait `ui::layout::obj::ComponentMsgObj` for type `ui::model_tt::component::keyboard::pin::PinKeyboard`
  --> src/ui/model_tt/layout.rs:57:1
   |
57 |   impl ComponentMsgObj for PinKeyboard
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `ui::model_tt::component::keyboard::pin::PinKeyboard`
   |
  ::: src/ui/layout/obj.rs:37:1
   |
37 | / impl<T> ComponentMsgObj for T
38 | | where
39 | |     T: Component,
40 | |     T::Msg: TryInto<Obj, Error = Error>,
...  |
44 | |     }
45 | | }
   | |_- first implementation here
```
What am I missing?

mmilata: The `TryInto` implementation is not there anymore, perhaps the compiler is not smart enough to figure out that `T::Msg: TryInto<Obj, Error = Error>` is false?

matejcik:
we can extend `tools/build_mocks` relatively easily to also walk Rust files (maybe only `obj.rs` ?) and generate mocks in the same format as the `modtrezor*.h` files.

core/embed/rust/src/ui/layout/result.rs
matejcik: For a smaller version, we could just make two `mp_obj_object_t`s. The option to print the object name wouldn't be there but we don't _really_ need that.

But this is good enough I think.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
jpochyla: https://docs.rs/zeroize/latest/zeroize/ is such a crate.

core/embed/rust/src/ui/layout/result.rs
mmilata: Ok, I'll leave it like this now and rewrite it once we add more result constants.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
mmilata: As discussed, we may be hitting rust-lang/rust#20400. I rewrote all the `TryInto` conversions to explicit `ComponentMsgObj` impls, it's a bit ugly but uses less magic.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
mmilata: Not storing the pin in the message anymore so zeroing isn't needed (yet). But good to know such crate exists.

core/embed/rust/src/micropython/macros.rs
mmilata: Copied the field matchers in [681ac69](https://github.com/trezor/trezor-firmware/pull/2138/commits/681ac69b3f12cbb842e46f646cdd2047b6bcc00b).

mmilata:
I've addressed the comments and extended `build_mocks` to also scan `.rs` file for definitions, please take a look. I think this shouldn't conflict with #2155 much so marking as ready for review.

mmilata:
Also rebased to master.
cc @grdddj

core/embed/rust/src/micropython/macros.rs
```diff
@@ -164,3 +170,36 @@ macro_rules! obj_type {
...
+macro_rules! obj_module {
+    ($($key:expr => $val:expr),*) => ({
+        #[allow(unused_unsafe)]
+        unsafe {
+            use $crate::micropython::ffi;
+
+            static DICT: ffi::mp_obj_dict_t = ffi::mp_obj_dict_t {
+                base: ffi::mp_obj_base_t {
+                    /// SAFETY: FIXME extern static
+                    type_: unsafe { &ffi::mp_type_dict },
```
mmilata, line 184: @jpochyla can you advise if this is safe? Seems like a value that shouldn't be changing normally.

core/embed/rust/src/micropython/macros.rs
matejcik: the comment needs to be tweaked: "This is safe only because we are passing in a static dict with a fixed Map created by (...). That causes MicroPython to treat `globals` as immutable (...)"

matejcik (approved):
LGTM structure wise, leaving the details to @jpochyla

big :+1: on `build_mocks` handling

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
matejcik: maybe you should zero out the source buffer for the PIN though?

core/embed/rust/src/micropython/macros.rs
mmilata: 744f4d6bfafbc0b7676ff3bdead8160af6441bba

core/SConscript.unix
```diff
@@ -668,9 +668,13 @@ protobuf_blobs = env.Command(
 )
 env.Depends(protobuf_blobs, qstr_generated)
 
-RUST_PROFILE = 'dev' if ARGUMENTS.get('TREZOR_EMULATOR_DEBUGGABLE', 0) == '1' else 'release'
+if ARGUMENTS.get('TREZOR_EMULATOR_DEBUGGABLE', '0') == '1':
+    RUST_PROFILE = 'dev'
```
jpochyla, line 672: Why not "debug"?

core/embed/rust/src/micropython/macros.rs
jpochyla: Safe 👍

core/embed/rust/src/micropython/obj.rs
```diff
@@ -289,6 +289,22 @@ impl TryFrom<&'static CStr> for Obj {
...
 }
 
+impl TryFrom<(Obj, Obj)> for Obj {
+    type Error = Error;
+
+    fn try_from(val: (Obj, Obj)) -> Result<Self, Self::Error> {
+        // SAFETY:
+        //  - Should work with any micropython objects.
+        // EXCEPTION: Will raise if allocation fails.
+        let obj = catch_exception(|| unsafe { ffi::mp_obj_new_tuple(2, [val.0, val.1].as_ptr()) })?;
```
jpochyla, line 299: Could we please put `[val.0, val.1]` into a variable?

core/SConscript.unix
mmilata: With `debug` I'm getting:
```
$ make build_unix_debug
...
cd embed/rust; cargo build --profile debug --target-dir=../../build/unix/rust --no-default-features --features "model_tt"
error: profile name `debug` is reserved
To configure the default development profile, use the name `dev` as in [profile.dev]
```

According to https://doc.rust-lang.org/cargo/reference/profiles.html there's no default `debug` profile.

jpochyla: Facepalm, sorry.

core/embed/rust/src/micropython/obj.rs
mmilata: Sure, fixed in f17120e0d330692e053d902c01f648021702cec4.

core/embed/rust/src/micropython/macros.rs
mmilata: Thanks, updated comment in d128f680d9d2dd267407647e608b38f501a6d737.

mmilata:
Noticed that `rustc` doesn't like `///` for the mocks:
```
warning: unused doc comment
   --> src/ui/model_tt/layout.rs:153:5
    |
153 |     /// CONFIRMED: object
    |     ^^^^^^^^^^^^^^^^^^^^^
154 |     Qstr::MP_QSTR_CONFIRMED => CONFIRMED.as_obj(),
    |     ----------------------- rustdoc does not generate documentation for expressions
    |
    = note: `#[warn(unused_doc_comments)]` on by default
    = help: use `//` for a plain comment
```

I've added `#[allow(unused_doc_comments)]` but let me know if using different prefix is preferable.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
matejcik: 👍 
added the TODO

honestly, couldn't we just consume the layout object instead? `self` instead of `&mut self`
(probably not because something something reference in the Python-held object, but still something that seems semantically more correct)

core/embed/rust/src/ui/geometry.rs
```diff
@@ -447,7 +447,7 @@ impl Grid {
     }
 
     pub fn cells(&self, cells: GridCellSpan) -> Rect {
```
matejcik, line 449: I recall wanting to rename this to `cell_span` but apparently I didn't in my previous modifications? Please do that here.

matejcik: actually let's not do that, I will be doing a subsequent PR so i will do that there separately

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1,109 +1,124 @@
...
-            PinKeyboardMsg::Confirmed => 1.try_into(),
-            PinKeyboardMsg::Cancelled => 2.try_into(),
+impl<T> ComponentMsgObj for PinKeyboard<T>
+where
+    T: Deref<Target = [u8]>,
+{
+    fn msg_try_into_obj(&self, msg: Self::Msg) -> Result<Obj, Error> {
+        match msg {
+            PinKeyboardMsg::Cancelled => Ok(CANCELLED.as_obj()),
+            PinKeyboardMsg::Confirmed => self.pin().try_into(),
```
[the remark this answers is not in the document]
mmilata: good catch, fixed in [e75c848](https://github.com/trezor/trezor-firmware/pull/2138/commits/e75c848977b590b2a6a6d8a2878e3bbb39571c79)

matejcik (approved):

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
mmilata: #2552

PR #2144, opened 2022-02-25
Ensure input's script type and path match the scriptPubKey

matejcik:
TODO:
* [ ] if safety-checks are lowered, allow unverified EXTERNAL inputs
* [ ] tests


legacy/firmware/signing.c
```diff
@@ -72,6 +72,7 @@ static uint32_t idx1;      // The index of the input or output in the current tx
 static uint32_t idx2;  // The index of the input or output in the original tx
                        // (Phase 1), in the previous tx (Phase 2) or in the
                        // current tx when computing the legacy digest (Phase 2).
+static uint8_t external_inputs[64];  // bitfield of external input indices
```
matejcik, line 75: tiny hyperoptimization: wouldn't it be better to use `uint32_t` since we're running on a 32-bit processor?

(i don't know if the compiler is clever enough to optimize the bitmasking fun; it might be)

matejcik (commented):
LGTM, but please add `incompatible` changelog entries.
I wrote some previously:

> Trezor will refuse to sign UTXOs that do not match the provided derivation path (e.g., transactions belonging to a different wallet, or synthetic transaction inputs).

andrewkozlik:
> please add `incompatible` changelog entries.

https://github.com/trezor/trezor-firmware/pull/2144/commits/61186a17403e8216afcfa5084f25ae6064354f40 https://github.com/trezor/trezor-firmware/pull/2144/commits/f4388a129192045c2573c9adb6ae364da271c72a

legacy/firmware/signing.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2144/commits/37bf82fdcb855801da6a1805068658e56ef484d8

andrewkozlik:
In https://github.com/trezor/trezor-firmware/pull/2144/commits/54ac83072568eb02167b9a0ea8b4f765c5ae977f https://github.com/trezor/trezor-firmware/pull/2144/commits/f7aa680fc4ac2898918694c2c8a937fada632f68 I did a little heuristic around the TxWeight calculation, so that Trezor doesn't show the fee-over-threshold dialog when a lot of unverified external inputs are present.

In https://github.com/trezor/trezor-firmware/pull/2144/commits/86c6620010d356490717e4bab28edf3ce4d0da6c I replaced the operation in `check_internal_input()`, which is no longer needed, with a sanity check.

andrewkozlik:
Closes https://github.com/satoshilabs/trezor-firmware/issues/123, see also https://github.com/satoshilabs/trezor-firmware/pull/128.

matejcik:
LGTM, please rebase and merge

PR #2149, opened 2022-03-02
feat(core, python): create secure input functionality

grdddj:
When creating the hello-world feature, I was wondering what other easy-to-create things could be done with Trezor, and can be even useful.

Having a possibility to get a secure/secret input from device (via the passphrase keyboard) seems quite handy, as it could be used by other scripts to get sensitive info (password, PIN) from the user - without being vulnerable to keyloggers etc.

`trezorctl secure-input` returns a string to `stdout`, so a program can look like
```sh
trezorctl secure-input -p "Specify password" | ./authenticate.sh
```
, effectively inputting that password into the script without the password being visible anywhere.
(Here I am not really sure if it is "bulletproof", but seems more secure than typing anything on the keyboard)

It should be quite easy to create a public function in `trezorlib`, to be used in any python script:
```python
from trezorlib import get_secure_input()
password = get_secure_input("give me password")
authenticate(password)
```

Could be used by `Connect` as well to work in web-browser.

It is just a proof of concept, so I put all the definitions at the same places as `get_entropy`, as it is similar.


prusnak:
> without being vulnerable to keyloggers etc.

This protects the input from keyloggers, but not from malware that can read process memory. It's arguable that if an attacker can install a keylogger, they could probably install such malware.

However, if the returned data would be hashed in the device and only the resulting (sha256) hash would be returned, we'd have a nice little gadget. But I am not sure yet, how useful would that be ...

matejcik:
as presented, this is a super bad idea:
```
trezorctl secure-input -p "Enter hidden wallet passphrase"
```

matejcik:
also "Enter word 3 of your mnemonic", etc.

Even returning the input hashed with sha256 would not help because it opens the passphrase to brute-force attacks significantly faster than against the "full thing" (where you need to run some heavy calculations to get the final address to check for funds) -- not to mention mnemonic words, which you can reverse in a microsecond.

Pushing any sort of arbitrary on-device input back to the host, without very careful consideration, breaks several important assumptions. So I'm closing this.

It could make for an interesting blogpost about building apps for Trezor, accompanied by why this is a bad idea though .)

PR #2150, opened 2022-03-03
ci: fix HWI test bitcoind sources

mmilata:


grdddj (approved):
Looks good to me, `CI` is all-green and running `nix-shell --arg fullDeps true` locally started to install the `bitcoind` just fine

matejcik (approved):
utACK

PR #2155, opened 2022-03-04
Rust UI: Add dynamic place system

jpochyla:
- Experimental text layout stuff put to `text/iter.rs`, not used yet.
- `TextLayout` itself is still not unicode aware, I've just added some higher-level helpers.
- I'm not very confident about the text stuff (`FormattedText` and especially `Paragraphs`). About correctness, design of the system, and design of the APIs. To be revisited, possibly with some ideas from the iterator-based text layout, possibly not.
- `Maybe` still not fixed.
- Grouping of components to be revisited later. Related to `Maybe`.
- Some components `union` the returned rects of children, some just return their own bounds and assume the children won't paint outside.
- Haven't tested the visuals yet.
- We should figure out what to do with `LinearPlacement`, maybe it should operate on components, maybe we should keep the `Dimensions` trait. Anyway currently it's panicking on one-element input.


ci/shell.nix
```diff
@@ -114,7 +114,7 @@ stdenvNoCC.mkDerivation ({
     ffmpeg
     dejavu_fonts
   ];
-  LD_LIBRARY_PATH = "${libffi}/lib:${libjpeg.out}/lib:${libusb1}/lib:${libressl.out}/lib";
+  DYLD_LIBRARY_PATH = "${libffi}/lib:${libjpeg.out}/lib:${libusb1}/lib:${libressl.out}/lib";
```
jpochyla, line 117: Pushed by mistake, sorry.

jpochyla: Anyway this is what is needed on MacOS to compile, maybe there should be an if or something.

mmilata: We can just use both if setting `LD_LIBRARY_PATH` doesn't cause problems on MacOS.

mmilata (commented):
Looks great! I've pushed some trivial fixes in 48ac54d. Paragraphs are not worse than before and basic visuals are OK. CI is failing due to #2150 and should be green after rebase. I'd merge this soon and do the text layout stuff in another PR.

mmilata:
> We should figure out what to do with LinearPlacement, maybe it should operate on components, maybe we should keep the Dimensions trait.

Not sure if adding another method to `Component` is worth it. Can we  reduce `Dimensions` to just this single method and then operate on `[impl Component + Dimensions]`?

> Anyway currently it's panicking on one-element input.

Where? I wasn't able to trigger a panic, seems to be working correctly in this case.

core/embed/rust/src/ui/component/base.rs
```diff
@@ -194,11 +272,26 @@ impl EventCtx {
...
+            paint_requested: false, /* We also need to paint, but this is supplemented by
+                                    * `Child::marked_for_paint` being true. */
             anim_frame_scheduled: false,
         }
     }
 
+    /// Indicate that position or sizes of components inside the component tree
+    /// have changed, and we should perform a place pass before next event or
+    /// paint traversals.
+    pub fn request_place(&mut self) {
```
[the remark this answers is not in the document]
jpochyla: Currently we don't, but since the rest of the code is the same, I thought, why not.

jpochyla:
> I've pushed some trivial fixes...

Thanks! I'll go through it once more in the next few days and fix what I find.

mmilata:
> * Experimental text layout stuff put to `text/iter.rs`, not used yet.
> 
> * `TextLayout` itself is still not unicode aware, I've just added some higher-level helpers.
> 
> * I'm not very confident about the text stuff (`FormattedText` and especially `Paragraphs`). About correctness, design of the system, and design of the APIs. To be revisited, possibly with some ideas from the iterator-based text layout, possibly not.
> 
> * `Maybe` still not fixed.
> 
> * Grouping of components to be revisited later. Related to `Maybe`.
> 
> * Some components `union` the returned rects of children, some just return their own bounds and assume the children won't paint outside.
> 
> * ~~Haven't tested the visuals yet.~~
> 
> * We should figure out what to do with `LinearPlacement`, maybe it should operate on components, maybe we should keep the `Dimensions` trait. Anyway currently it's panicking on one-element input.

Still, these are mostly existing issues and the PR doesn't make any of this worse, right? So I'd propose to merge it and solve the rest in another PRs.

Also, the one-element LinearPlacement panci seems to be fixed in this branch.

matejcik (commented):
I'm halfway through the detailed review, found some nitpicks. Agree with @mmilata that we can probably merge this and fix things as they come up.

PR #2164, opened 2022-03-09
Automatically pick best QR encoding

### LINKED ISSUE #1751: Do not convert bech32 addresses to upper-case for QR code
Per https://www.qrcode.com/en/about/version.html, V3 QR code can fit 42 characters in BINARY mode, and V4 QR code can fit 62. This happens to match the length of Bitcoin segwit bech32 addresses for single-sig (42) and multisig (62) accounts.

Despite that, we convert bech32 addresses to uppercase. Ostensibly to lower complexity of the QR code, but this does not help in practice (V2 fits only 38 ALPHANUM so we need V3 anyway; conversely, V3 fits only 61 ALPHANUM so we need V4 anyway).

We should probably keep bech32 addresses as lowercase, for better compatibility.

How are the lengths of taproot addresses?

matejcik:
The `Qr` component will automatically figure out if it can make the QR code smaller ( = bigger pixels) by uppercasing the address.

If not, it will keep lower-case encoding to improve compatibility.

This fixes #1751 because Bech32 Bitcoin addresses fit into the same-size QR code regardless of whether BINARY or ALPHANUMERIC encoding is used.


mmilata (approved):
Nice! UI diffs lgtm, segwit address QR changes as expected.

PR #2166, opened 2022-03-09
Support Zcash v5 transaction format on TT

### LINKED ISSUE #2031: Support ZIP-244 TxIDs
The NU5 network upgrade in Zcash introduces a new TxID digest [1]. The legacy TxIDs will still be accepted by the network, but in order for Trezor to be able to spend UTXOs from transactions which use the new ZIP-244 digest, Trezor will apparently need to implement the new algorithm, so that it may verify previous transaction output amounts. The NU5 upgrade should activate on Zcash Mainnet on 18th April 2022 [2].

- In trezor-core the `Zcashlike` class will need to override `get_prevtx_output()` to compute the legacy TxID or ZIP-244 TxID depending on the transaction version. 
- The same will be needed in trezor-legacy.
- Updates may also be needed in trezor-suite and blockbook. Further investigation is needed.

[1] ZIP-244: https://zips.z.cash/zip-0244#txid-digest
[2] https://z.cash/support/schedule/

jarys:
Implements #2031 for Trezor T. (not for Trezor 1, so the this PR does not close the issue)

This PR implements

- new Zcash transaction format ([ZIP-225](https://zips.z.cash/zip-0225)) and
- new Zcash sighash algorithm ([ZIP-244](https://zips.z.cash/zip-0244))

to support Zcash Network Upgrade 5.

I moved Zcash code into the separated app to prepare it for future extensions (like unified addresses and shielded transactions).

I kept old Zcash code to preserve Zcash-like altcoins like Komodo and Koto.

### DIFF
--- a/core/src/apps/bitcoin/sign_tx/__init__.py
+++ b/core/src/apps/bitcoin/sign_tx/__init__.py
@@ -9,7 +9,8 @@ from ..keychain import with_keychain
 from . import approvers, bitcoin, helpers, progress
 
 if not utils.BITCOIN_ONLY:
-    from . import bitcoinlike, decred, zcash
+    from . import bitcoinlike, decred, zcash_v4
+    from apps.zcash.signer import Zcash
 
 if TYPE_CHECKING:
     from typing import Protocol
@@ -70,7 +71,10 @@ async def sign_tx(
         if coin.decred:
             signer_class = decred.Decred
         elif coin.overwintered:
-            signer_class = zcash.Zcashlike
+            if msg.version == 5:
+                signer_class = Zcash
+            else:
+                signer_class = zcash_v4.ZcashV4
         else:
             signer_class = bitcoinlike.Bitcoinlike
 
--- a/core/src/apps/bitcoin/sign_tx/tx_info.py
+++ b/core/src/apps/bitcoin/sign_tx/tx_info.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
         def create_hash_writer(self) -> HashWriter:
             ...
 
-        def create_sig_hasher(self) -> SigHasher:
+        def create_sig_hasher(self, tx: SignTx | PrevTx) -> SigHasher:
             ...
 
         def write_tx_header(
@@ -60,7 +60,7 @@ _MAX_BIP125_RBF_SEQUENCE = const(0xFFFF_FFFD)
 
 
 class TxInfoBase:
-    def __init__(self, signer: Signer) -> None:
+    def __init__(self, signer: Signer, tx: SignTx | PrevTx) -> None:
         # Checksum of multisig inputs, used to validate change-output.
         self.multisig_fingerprint = MultisigFingerprintChecker()
 
@@ -73,7 +73,7 @@ class TxInfoBase:
         self.h_tx_check = HashWriter(sha256())  # not a real tx hash
 
         # BIP-0143 transaction hashing.
-        self.sig_hasher = signer.create_sig_hasher()
+        self.sig_hasher = signer.create_sig_hasher(tx)
 
         # The minimum nSequence of all inputs.
         self.min_sequence = _SEQUENCE_FINAL
@@ -122,14 +122,14 @@ class TxInfoBase:
 # Used to keep track of the transaction currently being signed.
 class TxInfo(TxInfoBase):
     def __init__(self, signer: Signer, tx: SignTx) -> None:
-        super().__init__(signer)
+        super().__init__(signer, tx)
         self.tx = tx
 
 
 # Used to keep track of any original transactions which are being replaced by the current transaction.
 class OriginalTxInfo(TxInfoBase):
     def __init__(self, signer: Signer, tx: PrevTx, orig_hash: bytes) -> None:
-        super().__init__(signer)
+        super().__init__(signer, tx)
         self.tx = tx
         self.signer = signer
         self.orig_hash = orig_hash
--- a/core/src/apps/zcash/signer.py
+++ b/core/src/apps/zcash/signer.py
@@ -1,0 +1,112 @@
+from micropython import const
+from typing import TYPE_CHECKING
+
+from trezor.messages import SignTx
+from trezor.utils import ensure
+from trezor.wire import DataError, ProcessError
+
+from apps.bitcoin.common import ecdsa_sign
+from apps.bitcoin.sign_tx.bitcoinlike import Bitcoinlike
+from apps.common.writers import write_compact_size, write_uint32_le
+
+from .hasher import ZcashHasher
+
+if TYPE_CHECKING:
+    from typing import Sequence
+    from apps.common.coininfo import CoinInfo
+    from apps.bitcoin.sign_tx.tx_info import OriginalTxInfo, TxInfo
+    from apps.bitcoin.writers import Writer
+    from apps.bitcoin.sign_tx.approvers import Approver
+    from trezor.utils import HashWriter
+    from trezor.messages import (
+        PrevTx,
+        TxInput,
+    )
+    from apps.bitcoin.keychain import Keychain
+
+OVERWINTERED = const(0x8000_0000)
+
+
+class Zcash(Bitcoinlike):
+    def __init__(
+        self,
+        tx: SignTx,
+        keychain: Keychain,
+        coin: CoinInfo,
+        approver: Approver | None,
+    ) -> None:
+        ensure(coin.overwintered)
+        if tx.version != 5:
+            raise DataError("Expected transaction version 5.")
+
+        super().__init__(tx, keychain, coin, approver)
+
+    def create_sig_hasher(self, tx: SignTx | PrevTx) -> ZcashHasher:
+        return ZcashHasher(tx)
+
+    def create_hash_writer(self) -> HashWriter:
+        # Replacement transactions are not supported
+        # so this should never be called.
+        raise NotImplementedError
+
+    async def step3_verify_inputs(self) -> None:
+        # Replacement transactions are not supported.
+
+        # We don't check prevouts, because BIP-341 techniques
+        # were adapted in ZIP-244 sighash algorithm.
+        # see: https://github.com/zcash/zips/issues/574
+        self.taproot_only = True  # turn on taproot behavior
+        await super().step3_verify_inputs()
+        self.taproot_only = False  # turn off taproot behavior
+
+    async def step5_serialize_outputs(self) -> None:
+        await super().step5_serialize_outputs()
+
+    async def sign_nonsegwit_input(self, i_sign: int) -> None:
+        await self.sign_nonsegwit_bip143_input(i_sign)
+
+    def sign_bip143_input(self, i: int, txi: TxInput) -> tuple[bytes, bytes]:
+        signature_digest = self.tx_info.sig_hasher.hash_zip244(
+            txi, self.input_derive_script(txi)
+        )
+        node = self.keychain.derive(txi.address_n)
+        signature = ecdsa_sign(node, signature_digest)
+        return node.public_key(), signature
+
+    async def process_original_input(self, txi: TxInput, script_pubkey: bytes) -> None:
+        raise ProcessError("Replacement transactions are not supported.")
+        # Zcash transaction fees are very low
+        # so there is no need to bump the fee.
+
+    async def get_tx_digest(
+        self,
+        i: int,
+        txi: TxInput,
+        tx_info: TxInfo | OriginalTxInfo,
+        public_keys: Sequence[bytes | memoryview],
+        threshold: int,
+        script_pubkey: bytes,
+        tx_hash: bytes | None = None,
+    ) -> bytes:
+        return self.tx_info.sig_hasher.hash_zip244(txi, script_pubkey)
+
+    def write_tx_header(
+        self, w: Writer, tx: SignTx | PrevTx, witness_marker: bool
+    ) -> None:
+        # defined in ZIP-225 (see https://zips.z.cash/zip-0225)
+        assert tx.version_group_id is not None
+        assert tx.branch_id is not None  # checked in sanitize_*
+        assert tx.expiry is not None
+
+        write_uint32_le(w, tx.version | OVERWINTERED)  # nVersion | fOverwintered
+        write_uint32_le(w, tx.version_group_id)  # nVersionGroupId
+        write_uint32_le(w, tx.branch_id)  # nConsensusBranchId
+        write_uint32_le(w, tx.lock_time)  # lock_time
+        write_uint32_le(w, tx.expiry)  # expiryHeight
+
+    def write_tx_footer(self, w: Writer, tx: SignTx | PrevTx) -> None:
+        # serialize Sapling bundle
+        write_compact_size(w, 0)  # nSpendsSapling
+        write_compact_size(w, 0)  # nOutputsSapling
+        # serialize Orchard bundle
+        write_compact_size(w, 0)  # nActionsOrchard
--- a/tests/device_tests/zcash/test_sign_tx.py
+++ b/tests/device_tests/zcash/test_sign_tx.py
@@ -1,0 +1,258 @@
+# This file is part of the Trezor project.
+#
+# Copyright (C) 2012-2019 SatoshiLabs and contributors
+#
+# This library is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License version 3
+# as published by the Free Software Foundation.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the License along with this library.
+# If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
+
+import pytest
+
+from trezorlib import btc, messages
+from trezorlib.exceptions import TrezorFailure
+from trezorlib.tools import parse_path
+
+from ..bitcoin.signtx import request_finished, request_input, request_output
+
+B = messages.ButtonRequestType
+
+TXHASH_aaf51e = bytes.fromhex(
+    "aaf51e4606c264e47e5c42c958fe4cf1539c5172684721e38e69f4ef634d75dc"
+)
+TXHASH_e38206 = bytes.fromhex(
+    "e3820602226974b1dd87b7113cc8aea8c63e5ae29293991e7bfa80c126930368"
+)
+
+pytestmark = [pytest.mark.altcoin, pytest.mark.zcash, pytest.mark.skip_t1]
+
+
+def test_version_group_id_missing(client):
+    inp1 = messages.TxInputType(
+        # tmQoJ3PTXgQLaRRZZYT6xk8XtjRbr2kCqwu
+        address_n=parse_path("m/44h/1h/0h/0/0"),
+        amount=300000000,
+        prev_hash=TXHASH_e38206,
+        prev_index=0,
+    )
+    out1 = messages.TxOutputType(
+        address="tmJ1xYxP8XNTtCoDgvdmQPSrxh5qZJgy65Z",
+        amount=300000000 - 1940,
+        script_type=messages.OutputScriptType.PAYTOADDRESS,
+    )
+
+    with pytest.raises(TrezorFailure, match="Version group ID must be set."):
+        btc.sign_tx(
+            client,
+            "Zcash Testnet",
+            [inp1],
+            [out1],
+            version=5,
+        )
+
+
+def test_one_one(client):
+    inp1 = messages.TxInputType(
+        # tmQoJ3PTXgQLaRRZZYT6xk8XtjRbr2kCqwu
+        address_n=parse_path("m/44h/1h/0h/0/0"),
+        amount=300000000,
+        prev_hash=TXHASH_e38206,
+        prev_index=0,
+    )
+
+    out1 = messages.TxOutputType(
+        address="tmJ1xYxP8XNTtCoDgvdmQPSrxh5qZJgy65Z",
+        amount=300000000 - 1940,
+        script_type=messages.OutputScriptType.PAYTOADDRESS,
+    )
+
+    with client:
+        client.set_expected_responses(
+            [
+                request_input(0),
+                request_output(0),
+                messages.ButtonRequest(code=B.ConfirmOutput),
+                messages.ButtonRequest(code=B.SignTx),
+                request_input(0),
+                request_output(0),
+                request_finished(),
+            ]
+        )
+
+        btc.sign_tx(
+            client,
+            "Zcash Testnet",
+            [inp1],
+            [out1],
+            version=5,
+            version_group_id=0x892F2085,
+            branch_id=0x76B809BB,
+        )
+
+        # TODO: send tx to testnet
+        # TODO: check serialized_tx
+
+
+def test_one_two(client):
+    inp1 = messages.TxInputType(
+        # tmQoJ3PTXgQLaRRZZYT6xk8XtjRbr2kCqwu
+        address_n=parse_path("m/44h/1h/0h/0/0"),
+        amount=300000000,
+        prev_hash=TXHASH_e38206,
+        prev_index=0,
+    )
+
+    out1 = messages.TxOutputType(
+        address="tmNvfeKR5PkcQazLEqddTskFr6Ev9tsovfQ",
+        amount=100000000,
+        script_type=messages.OutputScriptType.PAYTOADDRESS,
+    )
+
+    out2 = messages.TxOutputType(
+        address_n=parse_path("m/44h/1h/0h/0/0"),
+        amount=200000000 - 2000,
+        script_type=messages.OutputScriptType.PAYTOADDRESS,
+    )
+
+    with client:
+        client.set_expected_responses(
+            [
+                request_input(0),
+                request_output(0),
+                messages.ButtonRequest(code=B.ConfirmOutput),
+                request_output(1),
+                messages.ButtonRequest(code=B.SignTx),
+                request_input(0),
+                request_output(0),
+                request_output(1),
+                request_finished(),
+            ]
+        )
+
+        btc.sign_tx(
+            client,
+            "Zcash Testnet",
+            [inp1],
+            [out1, out2],
+            version=5,
+            version_group_id=0x892F2085,
+            branch_id=0x76B809BB,
+        )
+
+        # TODO: send tx to testnet
+        # TODO: check serialized_tx
+
+
+def test_external_presigned(client):
+    inp1 = messages.TxInputType(
+        # tmQoJ3PTXgQLaRRZZYT6xk8XtjRbr2kCqwu
+        address_n=parse_path("m/44h/1h/0h/0/0"),
+        amount=300000000,
+        prev_hash=TXHASH_e38206,
+        prev_index=0,
+    )
+
+    inp2 = messages.TxInputType(
+        # tmQoJ3PTXgQLaRRZZYT6xk8XtjRbr2kCqwu
+        # address_n=parse_path("m/44h/1h/0h/0/0"),
+        amount=300000000,
+        prev_hash=TXHASH_aaf51e,
+        prev_index=1,
+        script_type=messages.InputScriptType.EXTERNAL,
+        script_pubkey=bytes.fromhex(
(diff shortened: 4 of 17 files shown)


jarys:
blocked by missing unit tests

core/src/apps/zcash/sig_hasher.py
```diff
@@ -0,0 +1,187 @@
...
+        h = HashWriter(blake2b(outlen=32, personal=b'ZTxIdTranspaHash'))
+
+        # only SIGHASH_ALL is supported in Trezor
+        write_uint8(h, SigHashType.SIGHASH_ALL)         # S.2a: hash_type                (1 byte)
+        write_hash(h, self.prevouts.get_digest())       # S.2b: prevouts_sig_digest      (32-byte hash)
+        write_hash(h, self.amounts.get_digest())        # S.2c: amounts_sig_digest       (32-byte hash)
+        write_hash(h, self.scriptpubkeys.get_digest())  # S.2d: scriptpubkeys_sig_digest (32-byte hash)
+        write_hash(h, self.sequence.get_digest())       # S.2e: sequence_sig_digest      (32-byte hash)
+        write_hash(h, self.outputs.get_digest())        # S.2f: outputs_sig_digest       (32-byte hash)
+        write_hash(h, txin_sig_digest)                  # S.2g: txin_sig_digest          (32-byte hash)
```
andrewkozlik, line 150: I think the comments should go on a separate line above the code, because the code formatter will break them (not only here). Run `make style` in the root directory to format the code.

core/src/apps/zcash/sig_hasher.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(self):
+        self.header = HeaderHasher()
+        self.transparent = TransparentHasher()
+        self.sapling = SaplingHasher()
+        self.orchard = OrchardHasher()
+
+        self.tx_hash_person = None
+        self.initialized = False
+
+    def initialize(self, tx: SignTx | PrevTx):
```
andrewkozlik, line 43: I am not too fond of these `initialize` methods. I think the proper way is for this to go in `__init__()`. I assume the reason why it's not there is because `Signer.create_sig_hasher()` doesn't have the `tx` parameter, so one solution would be:
- Add the `tx` parameter to `create_sig_hasher()` and `__init__()` and to remove it from `hash341()` and `hash143()`.

However, the sighasher doesn't actually need to be initialized with the `tx` parameter. It is only used to precompute the header digest and some personalization string, which can both be computed in `signature_digest()` and `txid_digest()` and optionally cached to avoid recomputing. So another option is:
- Add the `tx` parameter to `signature_digest()` and `txid_digest()` and cache the header digest.

core/src/apps/zcash/sig_hasher.py
```diff
@@ -0,0 +1,187 @@
...
+):
+    """Returns `S.2g: txin_sig_digest` field for signature digest computation."""
+    assert sighash_type == SigHashType.SIGHASH_ALL
+    h = HashWriter(blake2b(outlen=32, personal=b'Zcash___TxInHash'))
+
+    write_prevout(h, txi)                       # S.2g.i:   prevout      (field encoding)
+    write_uint64(h, txi.amount)                 # S.2g.ii:  value        (8-byte signed little-endian)
+    scripts.write_bip143_script_code_prefixed(  # S.2g.iii: scriptPubKey (field encoding)
+        h, txi, public_keys, threshold, coin
+    )
```
andrewkozlik, line 170: I think in the end the Zcash people decided on really using the scriptPubKey, not the scriptCode, didn't they?

andrewkozlik (commented):
Some comments to start out the review.

core/src/apps/zcash/sig_hasher.py
```diff
@@ -0,0 +1,187 @@
...
+        assert self.initialized
+
+        h = HashWriter(blake2b(outlen=32, personal=self.tx_hash_person))
+
+        write_hash(h, self.header.digest())       # T.1: header_digest       (32-byte hash output)
+        write_hash(h, self.transparent.digest())  # T.2: transparent_digest  (32-byte hash output)
+        write_hash(h, self.sapling.digest())      # T.3: sapling_digest      (32-byte hash output)
+        write_hash(h, self.orchard.digest())      # T.4: orchard_digest      (32-byte hash output)
+
+        return h.get_digest()
```
andrewkozlik, line 65: Note that this is dead code. We actually don't use or need `txid_digest()` anywhere, so I am wondering whether to keep the code for it. The only place that I can think of where it would be used are replacement transactions, which we do not support for Zcash, because fee bumping is not an issue. Is there any other possible use?

andrewkozlik: I can imagine that at some point in the future we will want to display the TXID on the Trezor screen at the end of the signing process. So I don't mind keeping the code for future use, just wanted to open a discussion.

core/src/apps/zcash/sig_hasher.py
```diff
@@ -0,0 +1,187 @@
...
+def write_hash(w: Writer, hash: bytes):
+    write_bytes_fixed(w, hash, TX_HASH_SIZE)
+
+
+def write_prevout(w: Writer, txi: TxInput):
+    write_bytes_reversed(w, txi.prev_hash, TX_HASH_SIZE)
+    write_uint32(w, txi.prev_index)
+
+
+class ZcashSigHasher:
```
andrewkozlik, line 33: We have two classes named `ZcashSigHasher`, which is confusing, even though they are in different modules. The original one can be renamed to `Zip243SigHasher` or `ZcashLegacySigHasher`. The newer one can be renamed to `ZcashHasher` or `Zip244Hasher`, because it does more than just signature hashes (assuming we keep `txid_digest()`). There is also the question of whether we need it to be distinct from `ZcashExtendedSigHasher`.

In summary:
Change original `ZcashSigHasher` to `Zip243SigHasher` or `ZcashLegacySigHasher`.
Change new `ZcashSigHasher` to `Zip244Hasher` or `ZcashHasher` or merge with `ZcashExtendedSigHasher`.
Change `ZcashExtendedSigHasher` to `Zip244SigHasher` or `ZcashSigHasher`.

core/src/apps/bitcoin/sign_tx/__init__.py
```diff
@@ -70,7 +71,13 @@ async def sign_tx(
...
             signer_class = decred.Decred
         elif coin.overwintered:
-            signer_class = zcash.Zcashlike
+            if (
+                coin.coin_name in ["Zcash", "Zcash Testnet"] and
+                msg.version == 5  # remove this condition after Zcash NU5 mainnet activation
+            ):
+                signer_class = zcash.signer.Zcash
+            else:
+                signer_class = zcashlike.Zcashlike
```
andrewkozlik, line 80: I take it that `zcashlike.py` is for the Komodo and Koto coins. Are these going to keep using the older transaction format? If so, then I think we should be calling this `zcash_legacy.py` or `zcash_v4.py`, because "zcashlike" would seem to be implementing a superset of Zcash functionality, which it will no longer be doing.

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,155 @@
...
+        await super().step5_serialize_outputs()
+        await self.serialize_sapling_bundle()
+        await self.serialize_orchard_bundle()
+
+    async def serialize_sapling_bundle(self):
+        write_compact_size(self.serialized_tx, 0)  # nSpendsSapling
+        write_compact_size(self.serialized_tx, 0)  # nOutputsSapling
+
+    async def serialize_orchard_bundle(self):
+        write_compact_size(self.serialized_tx, 0)  # nActionsOrchard
```
andrewkozlik, line 79: These do not need to be `async`.

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,155 @@
...
+        # were adapted in ZIP-244 sighash algortihm.
+        # see: https://github.com/zcash/zips/issues/574
+        self.taproot_only = True  # turn on taproot behavior
+        await super().step3_verify_inputs()
+        self.taproot_only = False  # turn off taproot behavior
+
+    async def step5_serialize_outputs(self):
+        await super().step5_serialize_outputs()
+        await self.serialize_sapling_bundle()
+        await self.serialize_orchard_bundle()
```
andrewkozlik, line 72: Semantically I think this actually belongs somewhere else. It's much more than just the outputs, it's both spends and outputs or entire actions. I'd suggest putting it in `write_tx_footer()` or `step7_finish()` or maybe even `step6_sign_segwit_inputs()`?

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,155 @@
...
+    async def serialize_orchard_bundle(self):
+        write_compact_size(self.serialized_tx, 0)  # nActionsOrchard
+
+    async def sign_nonsegwit_input(self, i_sign: int) -> None:
+        await self.sign_nonsegwit_bip143_input(i_sign)
+
+    async def process_original_input(self, txi: TxInput, script_pubkey: bytes) -> None:
+        raise ProcessError("Replacement transactions are not supported.")
+        # Zcash transaction fees are very low
+        # so there is no need to dump the fee.
```
andrewkozlik, line 87: `dump` -> `bump`

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,155 @@
...
+    def create_hash_writer(self) -> HashWriter:
+        # Replacement transactions are not supported
+        # so this should never be called.
+        raise NotImplementedError
+
+    async def step3_verify_inputs(self):
+        # Replacement transactions are not supported.
+
+        # We don't check prevouts, because BIP-341 techniques
+        # were adapted in ZIP-244 sighash algortihm.
```
andrewkozlik, line 63: algortihm -> algorithm

tests/device_tests/zcash/test_sign_tx.py
```diff
@@ -0,0 +1,256 @@
...
+            "Zcash Testnet",
+            [inp1],
+            [out1],
+            version=5,
+            version_group_id=0x892F2085,
+            branch_id=0x76B809BB,
+        )
+
+        # TODO: send tx to testnet
+        # TODO: check serialized_tx
```
andrewkozlik, line 104: Let's be sure that the transactions work on testnet, before merging the code.

tests/device_tests/zcash/test_sign_tx.py
```diff
@@ -0,0 +1,256 @@
...
+    with pytest.raises(TrezorFailure, match="Replacement transactions are not supported."):
+        btc.sign_tx(
+            client,
+            "Zcash Testnet",
+            [inp1],
+            [out1, out2],
+            version=5,
+            version_group_id=0x892F2085,
+            branch_id=0x76B809BB,
+        )
```
andrewkozlik, line 256: It would be nice to have one multisig test.

andrewkozlik (changes requested):
One last thing I am wondering about is the movement of the Zcash code up by one level. I assume this is because the amount of Zcash code is going to get bigger soon with orchard and with other changes in `GetAddress`. It makes sense to me to do it this way, but it is unusual to have this sort of arrangement in the present code. Until now the apps in `core/src/apps` did not interact amongst each other, with the exception of `src/apps/common`. However, in this PR the Bitcoin app is calling the Zcash app. @matejcik should approve this arrangement.

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(
+        self,
+        tx: SignTx,
+        keychain: Keychain,
+        coin: CoinInfo,
+        approver: Approver,
+    ) -> None:
+        ensure(coin.overwintered)
+        if tx.version != 5:
+            # seemless version upgrade
```
andrewkozlik, line 42: seemless -> seamless

andrewkozlik:
When is the NU5 upgrade going to be activated on Mainnet?

jarys:
> When is the NU5 upgrade going to be activated on Mainnet?

In the end of May 2022 according to the [Zcash timeline](https://z.cash/support/schedule/).

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,155 @@
...
+            log.warning(__name__, "Transaction format version {} is not supported.".format(tx.version))
+            log.warning(__name__, "Switching to version 5.".format(tx.version))
+            tx.version = 5
```
andrewkozlik, line 45: After discussion with @matejcik, we don't think that changing the version here is a good idea. If the version is not 5, then the signing should fail, because it indicates that the host application is expecting a different transaction version than what Trezor is going to be providing. BTW, note that the warning would only get logged in the debug firmware.

core/src/apps/bitcoin/sign_tx/__init__.py
```diff
@@ -70,7 +71,13 @@ async def sign_tx(
         if coin.decred:
             signer_class = decred.Decred
         elif coin.overwintered:
-            signer_class = zcash.Zcashlike
+            if (
+                coin.coin_name in ["Zcash", "Zcash Testnet"] and
+                msg.version == 5  # remove this condition after Zcash NU5 mainnet activation
```
andrewkozlik, line 76: @matejcik and I are inclined to choosing the signer based on `msg.version` only, i.e. ignore the coin name. Based on past experience we can expect the other Zcashlike coins to adopt NU5 sooner or later.

core/src/apps/zcash/sig_hasher.py
jarys: Yay , I didn't know we have a style formater. Ok, I will be using it.

andrewkozlik: More information/guidelines about contributing to Trezor: https://github.com/trezor/trezor-firmware/blob/master/docs/misc/contributing.md.

core/src/apps/zcash/sig_hasher.py
jarys: I don't like an idea of passing `tx` every time some digest is needed while using it only once. I will pass it as argument of `create_sig_hasher()`.

core/src/apps/zcash/sig_hasher.py
jarys: You are right. They decided to use *scriptPubKey* according to [this comment](https://github.com/zcash/zips/pull/577#discussion_r781645348). 

However notice that calling `write_bip143_script_code_prefixed` has currently only two possible results:
- P2PKH scriptPubKey for multiple possible script types
- multisig scriptPubKey for multisig

The function always return a scriptPubKey so it could be renamed to `write_bip143_script_pubkey_prefixed`. Or even to `write_script_pubkey_prefixed` since it is currently used only in Zcash and it is not further related to the BIP-143.

Or preferably just to `write_script_pubkey` with `prefixed: bool = true` argument for sake of universality.

jarys:
I made a significant change to the code when I was modifying `ZcashSigHasher` to use use script_pubkey instead of script_code.

Note that neither `hash143` nor `hash341` does not receive script_pubkey argument. I considered writing a new function `write_script_pubkey` but then I decided to assimilate the Taproot approach, where script_pubkey is derived by the method `signer.input_derive_script(txi)`.

This led me to unimplementing `ZcashExtendedSigHasher.hash143` and overriding all methods where `hash143` method occured, i.e. `sign_bip143_input` and `get_tx_digest`.

core/src/apps/zcash/sig_hasher.py
jarys: I will keep it for future use cases and for now I will add a comment that this is a dead code.

core/src/apps/zcash/sig_hasher.py
jarys: I

- rename original `ZcashSigHasher` to `Zip243SigHasher`
- rename new `ZcashSigHasher` to `ZcashHasher`
- merge `ZcashExtendedSigHasher` into `ZcashHasher`

in https://github.com/trezor/trezor-firmware/pull/2166/commits/9cec3b7ed08d2c613717db28d6079759ea3c2712

core/src/apps/zcash/signer.py
jarys: I agree.

tests/device_tests/zcash/test_sign_tx.py
jarys: I agree. I will submit these transactions as soon as the testnet will be relaunched, which should be in the beginning of April.

core/src/apps/zcash/sig_hasher.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/976d716b73f53089fda19981d27c8be3041c391c and https://github.com/trezor/trezor-firmware/pull/2166/commits/4abffe5215e1db8d972ae1ae29b81d8be4323c5f

core/src/apps/zcash/sig_hasher.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/060aa20c4d7490f209b520230ea2122f93dd4dcb

core/src/apps/zcash/sig_hasher.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/079f0a118983a7fbecc12686c7b78ade78d2f650

core/src/apps/zcash/signer.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/4bce23cffea5c3cf4b97e7a0061559cb8ae6e352

core/src/apps/zcash/signer.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/4bce23cffea5c3cf4b97e7a0061559cb8ae6e352

core/src/apps/bitcoin/sign_tx/__init__.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/71c85f9a8fe9a8a00a3ea779af87b71b1c384ed3

core/src/apps/bitcoin/sign_tx/tx_info.py
```diff
@@ -122,15 +124,13 @@ def get_tx_check_digest(self) -> bytes:
 # Used to keep track of the transaction currently being signed.
 class TxInfo(TxInfoBase):
     def __init__(self, signer: Signer, tx: SignTx) -> None:
-        super().__init__(signer)
-        self.tx = tx
```
andrewkozlik, line 126: I think there was a good reason for `self.tx = tx` to be assigned here instead of in the parent class. It had to do with type checking. (Note that `make pystyle` also runs a static type-checker.)

core/src/apps/bitcoin/sign_tx/__init__.py
andrewkozlik: Also the class name `Zcashlike` -> `ZcashV4`.

jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/7c467f6fd3372d8c184f8b0a3ab18dac1e6e62cc

core/src/apps/bitcoin/sign_tx/__init__.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/4b94cad001411de4a84864be9fa678c53f74025d

core/src/apps/zcash/signer.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/4b94cad001411de4a84864be9fa678c53f74025d

core/src/apps/zcash/signer.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/4b94cad001411de4a84864be9fa678c53f74025d

core/src/apps/bitcoin/sign_tx/tx_info.py
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/14148f6d577059c8739b6c40e4b746da0b3424d1

core/src/apps/bitcoin/sign_tx/tx_info.py
```diff
@@ -60,7 +60,9 @@ async def write_prev_tx_footer(
 
 
 class TxInfoBase:
-    def __init__(self, signer: Signer) -> None:
+    def __init__(self, signer: Signer, tx: SignTx | PrevTx) -> None:
+        self.tx = tx
```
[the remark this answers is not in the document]
jarys: https://github.com/trezor/trezor-firmware/pull/2166/commits/a9ef2a93a5e656fd6f5510ce3511502c842b58d3

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,113 @@
...
+        approver: Approver | None,
+    ) -> None:
+        ensure(coin.overwintered)
+        if tx.version != 5:
+            raise DataError("Expected transaction version 5.")
+
+        super().__init__(tx, keychain, coin, approver)
+
+    def create_sig_hasher(self, tx: SignTx | PrevTx) -> ZcashHasher:
+        assert SignTx.is_type_of(tx)
```
andrewkozlik, line 45: Why is this actually needed? Couldn't the sighasher work just as well with `PrevTx`?

jarys:
Nice! All CI checks are passing now.

core/src/apps/zcash/signer.py
jarys: Yes, it can. https://github.com/trezor/trezor-firmware/pull/2166/commits/f4fb43e219e29106fd6824772ba243112a76bcc1

I was just lazy to check, because currently we don't need it.

andrewkozlik (approved):
utACK

TODO:
- @matejcik please see https://github.com/trezor/trezor-firmware/pull/2166#pullrequestreview-920081522, when you have the time.
- When the new testnet activates:
    - on-chain device tests 
    - add multisig device test
- Now:
    - <s>please rebase over master and squash the commits as much as you can (especially let's not clutter the history with "make pystyle", "clean up ...", "fix typo" etc.)</s>
    - <s>push the rebased branch to the `trezor/trezor-firmware` repository to run the ([full](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines)) CI</s>

grdddj:
Update from creating test vectors:
- we are currently unable to create a `V5` transaction that would be understood by the testnet
  - we have tried https://sochain.com and https://blockbook-dev.corp.sldev.cz:19132 as providers to send the transactions, and in both cases it was rejected
- [grdddj/zcash_v5_testnet_tx_vectors](https://github.com/trezor/trezor-firmware/tree/grdddj/zcash_v5_testnet_tx_vectors) branch contains a `test_spend_real_output` "test", which should generate a valid and announcable transaction, as it spends a valid `UTXO` - https://blockbook-dev.corp.sldev.cz:19132/tx/259f886923e6bf4aa77f5e0c715bb8883be0aa4dd68f482e084af8de1aa20ae7 - owned by `tmQoJ3PTXgQLaRRZZYT6xk8XtjRbr2kCqwu`, which is our address under  `m/44h/1h/0h/0/0`
  - the resulting `tx_hex` is `0500008085202f89b4d0d6c2000000000000000001e70aa21adef84a082e488fd64daae03b88b85b710c5e7fa74abfe62369889f25000000006a4730440220462a4b71f50f473adb0a6c5069db368db3940994cc4020f23764609cd7ef69db0220133e5427b68b95c036235ca01dd8e9f5fd01effa68171320a78ea5845370c6c70121030e669acac1f280d1ddf441cd2ba5e97417bf2689e4bbec86df4f831bf9f7ffd0ffffffff01d0ae3f00000000001976a9141215d421cb8cec1dea62cbd9e4e07c01520d873f88ac000000`
  - putting it into https://blockbook-dev.corp.sldev.cz:19132/sendtx results in `-22: TX decode failed`
  - we have verified that blockbook currently indexes `v5` transactions

andrewkozlik:
@grdddj `version_group_id` needs to be set to `0x26a7270a`, see https://blockbook-dev.corp.sldev.cz:19132/tx/733df241dcfe7f03d4ef0e36450048f09685a42ee734543ffae220c42292eb99

core/src/apps/zcash/signer.py
```diff
@@ -0,0 +1,112 @@
...
+
+    async def step3_verify_inputs(self) -> None:
+        # Replacement transactions are not supported.
+
+        # We don't check prevouts, because BIP-341 techniques
+        # were adapted in ZIP-244 sighash algorithm.
+        # see: https://github.com/zcash/zips/issues/574
+        self.taproot_only = True  # turn on taproot behavior
+        await super().step3_verify_inputs()
+        self.taproot_only = False  # turn off taproot behavior
```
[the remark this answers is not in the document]
andrewkozlik: It works, but I agree, it's not pretty. I was thinking of changing it to `step3_verify_inputs(check_prev_tx: bool)`. In `bitcoin.py` it would be called with `step3_verify_inputs(not self.taproot_only)` and in `Zcash` (v5) we would have:
```py
    async def step3_verify_inputs(check_prev_tx: bool) -> None:
        await super().step3_verify_inputs(False)
```
However, I like @matejcik's proposal just as well, only not sure about the function naming. `verify_inputs_fast() -> verify_inputs_no_prevtx()`?

andrewkozlik:
> @grdddj will create test vectors.

I think we should have the following types of transactions in the device tests:
- v4 UTXO spent in a v5 transaction
- v5 UTXO spent in a v5 transaction
- v5 multisig UTXO spent in a v5 transaction

grdddj:
> `version_group_id` needs to be set to `0x26a7270a`

Nice, thanks for this! Is it the same in every transaction? And what about the `branch_id` value?

andrewkozlik:
> > `version_group_id` needs to be set to `0x26a7270a`
> 
> Nice, thanks for this! Is it the same in every transaction? And what about the `branch_id` value?

I think it should be the same in every transaction. I have no clue where the value comes from. They are not specified in the ZIPs, I just read them from the transactions in the blockchain.

grdddj:
I have force-pushed [grdddj/zcash_v5_testnet_tx_vectors](https://github.com/trezor/trezor-firmware/tree/grdddj/zcash_v5_testnet_tx_vectors) branch, and currently all the test-vector changes are in one commit - https://github.com/trezor/trezor-firmware/commit/1c618d9336f20148358e66ecb627076d281dc8f4

I was mistaken that we already have spend from `v4`, as even that [259f886923e6bf4aa77f5e0c715bb8883be0aa4dd68f482e084af8de1aa20ae7](https://blockbook-dev.corp.sldev.cz:19132/tx/259f886923e6bf4aa77f5e0c715bb8883be0aa4dd68f482e084af8de1aa20ae7) was already `v5`. Luckily, there was one history UTXO still unspent on `tmAgYbANTzZp7YoMkRbbaemQETgV5GkBEjF` ([4b6cecb81c825180786ebe07b65bcc76078afc5be0f1c64e08d764005012380d](https://blockbook-dev.corp.sldev.cz:19132/tx/4b6cecb81c825180786ebe07b65bcc76078afc5be0f1c64e08d764005012380d)), which I could spend.

Connected with spending from a multisig, I also included a test sending funds to multisig, which was a transaction I had to generate anyway.

There are currently two tests marked with `FAKE tx`:
- `test_one_one`, which could be theoretically removed, as we test this in a lot of other places
- `test_external_presigned` - as this is still spending invalid (already spend) inputs. I did not see a big worth recreating it
- the rest (apart from the `pytest.raises(...)`) are valid `Zcash Testnet` transactions

@matejcik also pointed out it would be nice to test spending of `v4` and `v5` in the same transaction, but I could not find any other old `v4` outputs and also failed to generate some myself. Maybe somebody else will be more lucky.

andrewkozlik:
> * `test_one_one`, which could be theoretically removed, as we test this in a lot of other places

I'd remove it, since we already have `test_spend_v4_input` and `test_spend_v5_input`, which do exactly this.

grdddj:
> > * `test_one_one`, which could be theoretically removed, as we test this in a lot of other places
> 
> I'd remove it, since we already have `test_spend_v4_input` and `test_spend_v5_input`, which do exactly this.

Removed in 1a97288 - a fixup commit to the above one

grdddj:
One more fixup - bf64793991e9c0f989d6ab2ad3c36b342ef67d53 - to fix the `fixtures.json` expected hashes. I made the mistake of generating the hashes with enabled animations. CI was correctly showing `UI` mismatch

matejcik:
will be merged in #2245

PR #2168, opened 2022-03-10
fix(core): support structs as array elements (fixes #2167)

### LINKED ISSUE #2167: EIP712 signing fails when struct is only used as an array element
**Describe the bug**
Hello guys, this is Bojan from @InjectiveLabs. We've been trying to implement native support for Trezor for the products that we are building but we ended up having an issue when trying to sign EIP712 typed data. This issue happens on Metamask as well when we try to sign the same EIP712 message using their native Trezor integration - which just shows that this is indeed an issue on your end. 

I'm using the latest firmware: `2.4.3` and I have Trezor Model T.

 - Wallet software: Trezor Suite installed
 - OS: MacOs
 - Browser: Chrome
 - Version: 99.0.4844.51

**To Reproduce**
- Try to sign the EIP712 typed data provided below using `TrezorConnect.ethereumSignTypedData`. You can even run the tests against 

**Expected behavior**
The nested types should be parsed properly and we should be able to see the nested typed data on display and the signature should be generated successfully. 

**Screenshots**

**Additional context**
I've tested with\without `message_hash`, `domain_separator_hash`, `metamask_v4_compat`, basically every possible scenario and it still doesn't work. If you try to click on "See the full message" on the screen on Trezor you get "Firmware Error", if you click accept you - again get "Firmware Error".

I've tried to remove the nested types (represent them as a single object and not an array of the type) and sign the typed data without them - it worked just fine. An example from the bellow typed data would be to replace Coin[] with Coin and in the Fee type - if we change this the data will be visible on the display and it can be signed. Obviously, we can't change the types of the typed data, instead we should have support for such types on Trezor.

Maybe you can reference my colleague's (@xlab) implementation that actually worked even with nested types.

```ts
const eip712Data = {
"types": {
    "Coin": [
        {
            "name": "denom",
            "type": "string"
        },
        {
            "name": "amount",
            "type": "string"
        }
    ],
    "EIP712Domain": [
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "version",
            "type": "string"
        },
        {
            "name": "chainId",
            "type": "uint256"
        },
        {
            "name": "verifyingContract",
            "type": "string"
        },
        {
            "name": "salt",
            "type": "string"
        }
    ],
    "Fee": [
        {
            "name": "amount",
            "type": "Coin[]"
        },
        {
            "name": "gas",
            "type": "string"
        }
    ],
    "Msg": [
        {
            "name": "type",
            "type": "string"
        },
        {
            "name": "value",
            "type": "MsgValue"
        }
    ],
    "MsgValue": [
        {
            "name": "delegator_address",
            "type": "string"
        },
        {
            "name": "validator_address",
            "type": "string"
        },
        {
            "name": "amount",
            "type": "TypeAmount"
        }
    ],
    "Tx": [
        {
            "name": "account_number",
            "type": "string"
        },
        {
            "name": "chain_id",
            "type": "string"
        },
        {
            "name": "fee",
            "type": "Fee"
        },
        {
            "name": "memo",
            "type": "string"
        },
        {
            "name": "msgs",
            "type": "Msg[]"
        },
        {
            "name": "sequence",
            "type": "string"
        },
        {
            "name": "timeout_height",
            "type": "string"
        }
    ],
    "TypeAmount": [
        {
            "name": "denom",
            "type": "string"
        },
        {
            "name": "amount",
            "type": "string"
        }
    ]
},
"primaryType": "Tx",
"sequence": "0",
"timeout_height": "8535057",
"domain": {
    "name": "Injective Web3",
    "version": "1.0.0",
    "chainId": 1,
    "verifyingContract": "cosmos",
    "salt": "1646876620490"
},
"message": {
    "account_number": "5712",
    "chain_id": "injective-1",
    "fee": {
        "amount": [
            {
            "amount": "200000000000000",
            "denom": "inj"
            }
        ],
        "gas": "400000"
    },
    "memo": "",
    "msgs": [
        {
        "type": "cosmos-sdk/MsgDelegate",
        "value": {
            "amount": {
                "amount": "100000000000000000",
                "denom": "inj"
            },
            "delegator_address": "inj17vy49gw9xnx700z8zwqqv4exl2rgdhanv75c4r",
            "validator_address": "injvaloper1w3psm8a9td2qz06s46cxss03mz5umxaxegvhhs"
            }
        }
    ]
}};

TrezorConnect.ethereumSignTypedData({
    path: "m/44'/60'/0'/0/0",
     metamask_v4_compat: true, // doesn't matter
    domain_separator_hash: "6192106f129ce05c9075d319c1fa6ea9b3ae37cbd0c1ef92e2be7137bb07baa1",
    message_hash: "c9e71eb57cf9fa86ec670283b58cb15326bb6933c8d8e2ecb2c0849021b3ef42",
    data: eip712Data
})
```

matejcik:


grdddj (approved):
Looks good. I verified those two new signatures by `Metamask` and they are identical.

PR #2171, opened 2022-03-10
Rust Storage: LRU Cache

jpochyla:


core/embed/rust/src/storage/cache.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    pub fn start_session(&mut self, received_session_id_bytes: Option<&[u8]>) -> SessionId {
+        // If we have received a session ID, take a look to the cache and return an ID
+        // of existing entry.
+        let received_session_id =
+            received_session_id_bytes.and_then(|bytes| SessionId::try_from(bytes).ok());
+        if let Some(received_id) = received_session_id {
+            if self.sessions.find(|s| s.id == Some(received_id)).is_some() {
+                return received_id;
```
matejcik, line 39: `self.active_id = received_session_id`

core/embed/rust/src/storage/cache.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl SessionCache {
+    fn new(id: SessionId) -> Self {
+        Self {
+            id: Some(id),
+            fields: Some(SessionFields::default()),
+        }
+    }
+
+    fn clear(&mut self) {
```
matejcik, line 77: rather than clearing the cache entry itself, let's add a `drop(predicate)` to the cache?

jpochyla: That would involve a rather large memcpy in this case, for no other benefit.

PR #2124, opened 2022-02-14
chore(python): loosen ethereum dependencies

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2111:
- some small changes in Ethereum dependencies
- change `rlp` to `simple-rlp` and importing `web3` only in the (one and only) function it uses it

It was not however without some linting issues. `flake8` was complaining that `rlp` is not in the list of requirements (as it comes from `simple-rlp` package) - I put a `noqa` comment there to silence it. Also `web3.Web3` could not be used as a type hint any more, as `web3` is no longer a whole-package import - I changed the argument type to `Any` in function using it.

PR #2177, opened 2022-03-15
fix(python): fix eth.sign-tx - correct input into rlp.encode()

grdddj:
Fixes issue introduced in https://github.com/trezor/trezor-firmware/pull/2124#:
- Ethereum signing with `trezorctl` was misbehaving because of the new `simple-rlp` dependency

Turned out `rlp.encode` needs a list of values, it cannot iterate through a tuple.


python/src/trezorlib/cli/ethereum.py
```diff
@@ -352,29 +352,37 @@ def sign_tx(
...
-                max_gas_fee,
-                gas_limit,
-                to,
-                amount,
-                data_bytes,
-                _format_access_list(access_list) if access_list is not None else [],
-            )
-            + sig
-        )
+        raw_transaction = [
```
matejcik, line 359: maybe rename to `transaction_items` ? this is not really a "raw" transaction

matejcik (approved):
almost LGTM, please go ahead and merge after fixing the variable name

PR #2181, opened 2022-03-18
chore(common): defs cleanup

prusnak:
- drop Gamecredits - they became an ERC-20 token in 2020
- drop Nix - dead coin
- drop Polis - coin became a BEP-20 token and later Olympus network in 2021


common/defs/support.json
```diff
@@ -2559,7 +2554,6 @@
       "eth:tPCTT": "(AUTO) exclude testnet",
       "eth:tPHOTON": "(AUTO) exclude testnet",
       "eth:tPLS": "(AUTO) exclude testnet",
-      "eth:tPOLIS": "(AUTO) exclude testnet",
```
matejcik, line 2562: we should probably keep `tPOLIS` out either way, given that we exclude other testnets

PR #2182, opened 2022-03-21
Make pyright_tool more user-friendly

matejcik:
* passed in directory respects absolute and relative paths instead of
  working "from repository root"
* we don't require the enableTypeIgnoreComments to be set so both
  `pyright` and `pyright_tool` can work with the same config at the same
  time
* use click's magic functionality to deal with missing / unreadable /
  unwriteable files
* read the error results via a pipe, do not write to filesystem unless
  requested
* simplified logic regarding "test mode"/"dev mode"
* renamed `--log` to more typical `--verbose`
* use pathlib more extensively


matejcik:
removed the comment in force-push.

will merge when CI is clear

Issue #2184, opened 2022-03-22
Optimize oledInvertDebugLink


matejcik:
The function `oledInvertDebugLink()` decides at runtime on every drawing cycle whether a debug build is running. We can instead conditionally compile in the function for debug builds only, and get rid of `is_debug_link` variable too.

Originally reported by @invd

prusnak:
Fix in PR https://github.com/trezor/trezor-firmware/pull/2186

PR #2186, opened 2022-03-22
fix(legacy): refactor oledInvertPixel to not use static bool variable, but ifdef constant

prusnak:
Fixes https://github.com/trezor/trezor-firmware/issues/2184

Before the change:
```
$ ls -l legacy/bootloader/bootloader-unaligned.bin 
-rwxr-xr-x 1 stick staff 31400 Mar 22 20:36 legacy/bootloader/bootloader-unaligned.bin
```

After the change:
```
$ ls -l legacy/bootloader/bootloader-unaligned.bin 
-rwxr-xr-x 1 stick staff 31276 Mar 22 20:32 legacy/bootloader/bootloader-unaligned.bin
```

Difference: -124 bytes

### DIFF
--- a/legacy/oled.c
+++ b/legacy/oled.c
@@ -65,7 +65,6 @@
  */
 
 static uint8_t _oledbuffer[OLED_BUFSIZE];
-static bool is_debug_link = 0;
 
 /*
  * macros to convert coordinate to bit position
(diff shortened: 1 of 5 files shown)


legacy/oled.c
```diff
@@ -181,23 +180,23 @@ void oledInit() {
 void oledClear() { memzero(_oledbuffer, sizeof(_oledbuffer)); }
 
 void oledInvertDebugLink() {
```
invd, line 182: In the current change, the function body is still defined in the production firmware build. However, the function isn't called, so the compiler likely ignores it for the bootloader/firmware build and this difference isn't important.

matejcik: it looks slightly cleaner to define out the whole function, but it's called twice so that would be ifdef'ing out two call sites. keeping the function empty and letting the emulator do the job seems better overall

PR #2187, opened 2022-03-23
ci: fix upload folders path on public storage

vdovhanych:
This fixes wrong path for uploading the ui_tests fixtures and emulators on data.trezor.io it was a mistake on my end and now it shoud be fixed.


matejcik (approved):
utACK, it seemed to work before but :woman_shrugging:

PR #2188, opened 2022-03-23
ci: public storage index upload fix

vdovhanych:
this is hopefully the last fix for uploading, last one was also broken as it was not uploading the index correctly


matejcik (approved):

matejcik:
why are github actions not running here...?

vdovhanych:
No idea but I think most likely it's due to the fact I pushed the already merged branch back to Github?

vdovhanych:
force push with changed commit message did not help.

matejcik:
Ah, it's a site-wide problem. I can't merge without the block-fixup check anyway so this will have to wait.

PR #2192, opened 2022-03-28
fix(core+legacy): fix SDL2 include path

prusnak:
The default include path for SDL2 is `/usr/include/SDL2` not `/usr/include`
```
$ pkg-config --cflags sdl2
-D_THREAD_SAFE -I/usr/include/SDL2
```


sime:
Oooh, I think I was battling this when trying to build it outside of Nix OS.

matejcik (approved):

matejcik:
CI failures are unrelated, merging

PR #2194, opened 2022-03-29
ci: use autoPatchelfHook from pinned nixpkgs for emulators

mmilata:
NixOS recently switched to different autoPatchelfHook implementation which doesn't work for us in NixOS/nixpkgs#149731. This PR forces the use of the older implementation as a temporary measure.

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -72,6 +72,7 @@ stdenvNoCC.mkDerivation ({
     SDL2
     SDL2_image
     autoflake
+    autoPatchelfHook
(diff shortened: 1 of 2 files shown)


matejcik (approved):
utACK

why isn't the new impl working, btw?

mmilata:
I'm having trouble comprehending what the shell code in NixOS/nixpkgs#149731 does but it seems that now running `autoPatchelf emulators/trezor-emu-*` does not modify the emulator binaries at all.

Issue #2193, opened 2022-03-29
Clean core unit tests


grdddj:
Unit tests in `core` contain some easy-to-solve inconsistencies/small issues - for example:
- archaic non-`black` styling
- occasional unused imports and unused variables through the tests

(generally running `flake8 core/tests` shows a lot of these issues)

It could be worth addressing this and increase the readability and maintainability of the tests.

Also some test structure could be done, like we did in `device-tests` - not having all the tests in one directory

PR #2196, opened 2022-03-29
Grdddj/core unit tests cleanup

grdddj:
Quick code and style improvements of `unit-tests` connected with https://github.com/trezor/trezor-firmware/issues/2193

We might not include the `black` changes, which have a very big diff - but they will decrease any future diffs when modifying the test cases


grdddj:
Force-pushed new changes done on the latest `master`.

Also, in addition to styling/code quality changes, allowing unit-tests for `TR` and adding this test into `CI`

obrusvit:
I rebased the branch on current `master` and re-run black and isort.

In order to add `core/tests` into `style` target, multiple more issues must be addressed. The most numerous warnings are:
- `F405` - star import (originating from `from common import *`)
- `E501` - long lines 
- `I900` -  'apps' not listed as a requirement 
- short variables, etc.

This PR also adds build and test targets specific for TS3.

If CI checks pass, I suggest to merge to partially improve the style of tests nad fix the issues later.

obrusvit (approved):
With the limitations highlighted in my previous comment, fine by me.

matejcik:
please also add this to Github Actions config. the gitlab CI is kind-of deprecated

obrusvit:
I added `T2B1` for python, rust and rust-client unit tests.

As per #2724 also `persistence` and `upgrade` tests should be added. However, those seem not prepared for `T2B1` so I consider it OOS for this PR.

matejcik:
rust-client is failing:
```
---- tests::test_emulator_features stdout ----
thread 'tests::test_emulator_features' panicked at 'assertion failed: `(left == right)`
  left: `"Safe 3"`,
 right: `"T"`', src/lib.rs:174:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
this is super uninteresting, let's just disable rust-client on T2B1 for now

firmware clippy tests are also failing:
```
error: this arithmetic operation will overflow
   --> src/ui/model_tt/component/homescreen/render.rs:147:30
    |
147 |                 &mut workbuf[2 * HOMESCREEN_TOIF_X_OFFSET
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `2_usize * 18446744073[70](https://github.com/trezor/trezor-firmware/actions/runs/7775914738/job/21202488123?pr=2196#step:5:71)9551608_usize`, which would overflow
    |
    = note: `#[deny(arithmetic_overflow)]` on by default
```

i've seen this before but I don't remember what it's about, investigate pls?

matejcik:
....on second thought, maybe first rebase on current main (feel free to squash all the fixups now) and see if the error goes away

obrusvit:
rebased (with `--autosquash`) on top of current `main`. Let's wait for the result of rust tests, then I investigate.

obrusvit:
Fixed the uninteresting part (`trezorctl`).

The 2nd one (rust unit tests) still fails.
It's probably something with `TREZOR_MODEL`, you can see `clippy` reaching to `src/ui/model_tt/component/homescreen/render.rs:` However,  if I build firmware for `TREZOR_MODEL=R`:

- `make clippy` works on my machine
- `make test_rust` fails
- `TREZOR_MODEL=R make test_rust` works fine

Btw. this line is hellish :laughing: `--features model_t$(shell echo $(TREZOR_MODEL) | tr "TR" "tr")`

matejcik (approved):

matejcik:
I briefly looked into this error and my suspicion is that it's a clippy version problem. I can randomly reproduce it locally when I manage to tell Rust to ignore my local versions of clippy and use the one from nix-shell, but even `nix-shell --pure` does not block access to my non-nix clippy so :woman_shrugging: 

let's mark the test as allowed-to-fail, and see if #3337 helps

obrusvit:
I rebased and excluded `T2B1` from `core_unit_rust_test` job with explanatory comment. 

Apparently, GH Actions currently don't have "allow failure" feature..? https://github.com/actions/runner/issues/2347

obrusvit:
rebased again because of omitted trailing whitespace in `core.yaml`

matejcik:
OK then, ship it

PR #2207, opened 2022-04-05
CI test fixes

mmilata:

PR #2210, opened 2022-04-05
test(storage): increase timeout for test_random_upgrade

mmilata:
Continuation of #2207.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -337,8 +337,8 @@ crypto build:
   only:
     changes:
       - .gitlab-ci.yml
-      - ci/**/*
-      - crypto/**/*
+      - ci/**
+      - crypto/**
   script:
     - cp -r crypto crypto_noasan
     - nix-shell --run "poetry run make -C crypto"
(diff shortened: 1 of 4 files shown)


mmilata:
Brand new error: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2293416551
```
00:10:38 E           hypothesis.errors.DeadlineExceeded: Test took 2516.78ms, which exceeds the deadline of 2000.00ms
```

andrewkozlik:
> cc @andrewkozlik, maybe we can loosen the deadline as well?

Agreed. It's basically the same thing, slow CI runners.

matejcik:
fwiw @mmilata https://github.com/trezor/trezor-firmware/commit/89bf7bca7d18663fd6d60dec48f0f97fea8ef873 this seems nicer than explicitly adding timeouts where none were before

mmilata:
Now gitlab isn't even showing the job at all, wut? Let me try again ...

PR #2212, opened 2022-04-07
ci(monero): update test executable, enable more tests

mmilata:
Due to `--heavy_tests` the test now takes additional 1 minute on emulator and 5 minutes on hardware.


matejcik (approved):
utACK

Issue #2170, opened 2022-03-10
Implement `storage.device` module in Rust


matejcik:
`storage/device.py` is very large in terms of function count, and very short in terms of actual code.

To lower memory pressure, we should implement these functions in Rust where they don't take up RAM space.

This will be the first step to rewriting whole `storage` into Rust.

Issue #1947, opened 2021-11-26
Investigate how to save flash space


prusnak:
We are slowly running out of Flash space. Let's collect some ideas how to save it:

- compressing the embedded bootloader ([see below](https://github.com/trezor/trezor-firmware/issues/1947#issuecomment-979883994))
- using just single sha256 implementation

----

The greatest offenders (generated by `bloaty -d sections,symbols build/firmware/firmware.elf`)

```
   7.6%   749Ki  42.7%   749Ki    .flash
    52.0%   389Ki  52.0%   389Ki    [3352 Others]
    23.7%   177Ki  23.7%   177Ki    [section .flash]
     4.8%  36.2Ki   4.8%  36.2Ki    nist256p1
     4.8%  36.2Ki   4.8%  36.2Ki    secp256k1
     3.2%  24.0Ki   3.2%  24.0Ki    ge25519_niels_base_multiples
     1.9%  13.9Ki   1.9%  13.9Ki    blake2b_compress
     1.1%  8.00Ki   1.1%  8.00Ki    wordlist
     0.9%  7.03Ki   0.9%  7.03Ki    img_loader
     0.9%  6.94Ki   0.9%  6.94Ki    secp256k1_sha256_write
     0.8%  5.93Ki   0.8%  5.93Ki    mp_qstr_const_pool
     0.6%  4.57Ki   0.6%  4.57Ki    ripemd160_process
     0.5%  4.10Ki   0.5%  4.10Ki    mp_execute_bytecode
     0.5%  4.00Ki   0.5%  4.00Ki    slip39_wordlist
     0.5%  4.00Ki   0.5%  4.00Ki    t_fl
     0.5%  4.00Ki   0.5%  4.00Ki    t_fn
     0.5%  4.00Ki   0.5%  4.00Ki    t_il
     0.5%  4.00Ki   0.5%  4.00Ki    t_im
     0.5%  4.00Ki   0.5%  4.00Ki    t_in
     0.5%  4.00Ki   0.5%  4.00Ki    words_button_seq
     0.5%  3.75Ki   0.5%  3.75Ki    ge25519_niels_sliding_multiples
     0.5%  3.68Ki   0.5%  3.68Ki    blake2s_compress

   8.4%   824Ki  46.9%   824Ki    .flash2
    50.8%   418Ki  50.8%   418Ki    [13139 Others]
    26.5%   218Ki  26.5%   218Ki    [section .flash2]
[...]

PR #2214, opened 2022-04-11
Storage in Rust

grdddj:
Implementing almost the whole `storage` logic in `Rust` and using that in `micropython`.
Connected with https://github.com/trezor/trezor-firmware/issues/2170

NOTES:
- `cache.py` is still in `micropython` - and all other functions using its functionality
- used constants from `device.py` were moved into `__init__.py` or stayed in `device_old.py`

DOES NOT WORK / IS MISSING:
- (latest CI - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/513831666)
- [firmware](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2317886014) cannot be built because `region 'FLASH' overflowed by 15368 bytes`
- [click tests](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2317886252) always fail at the last test-case, regardless of the order of the tests
- [upgrade tests](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2317886257) fail on some layout diff, locally I could not replicate it
- there is almost no documentation for the new functionality
- constants are not exported from the Rust code


grdddj:
Rebased on `master` to use the `StrBuffer` changes

mmilata:
Added [aa5cd4e](https://github.com/trezor/trezor-firmware/pull/2214/commits/aa5cd4e38be0bfe9bffdf9adae3b40539b860748) that moves readonly static data belonging to modetrezorui, like fonts, to flash2. Now the firmware should build and the segments be full like:
* `.flash`: `748KiB / 768 KiB` - 20 free
* `.flash2`: `878KiB / 896 KiB` - 18 free

matejcik:
speaking of the above result, @grdddj can your tool be used to do a complete breakdown of the flash contents? not just for Python apps but for other code as well?

matejcik:
please also investigate why we need the cache song-and-dance for `get_experimental_features`.

My first suspicion is that it is because we want the field to be public but it is private.

grdddj:
> speaking of the above result, @grdddj can your tool be used to do a complete breakdown of the flash contents? not just for Python apps but for other code as well?

There are currently two scripts. One for micropython - `analyze_micropython_code_size.py`, which looks at `build/firmware/frozen_mpy.o` (finds `550 kb`). Second one is mainly for the Rust  and C code - `analyze_firmware_code_size.py`, which looks currently at `build/unix/trezor-emu-core` (finds `822 kb`), but could be also used to analyze `build/firmware/firmware.elf`.

Running `analyze_firmware_code_size.py` for  `build/firmware/firmware.elf` tells that there is `1230 kb` of code size, so that is probably not everything, as `flash + flash2 = 1600 kb`.

To really get all the things in `flash` and `flash2`, there is a command from this [comment](https://github.com/trezor/trezor-firmware/issues/1947#issue-1064347794) -  `bloaty -d sections,symbols -n 0 build/firmware/firmware.elf` (it shows more than 50_000 lines). I could do some aggregating script from this, at least to get the overall info from there - `.flash2=838Ki, .flash=751Ki`

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+extern "C" fn storagedevice_set_mnemonic_secret(
+    n_args: usize,
+    args: *const Obj,
+    kwargs: *mut Map,
+) -> Obj {
+    let block = |_args: &[Obj], kwargs: &Map| {
+        let secret: Buffer = kwargs.get(Qstr::MP_QSTR_secret)?.try_into()?;
+        let backup_type: u8 = kwargs.get(Qstr::MP_QSTR_backup_type)?.try_into()?;
+        let needs_backup: bool = if kwargs.contains_key(Qstr::MP_QSTR_needs_backup) {
+            kwargs.get(Qstr::MP_QSTR_needs_backup)?.try_into()?
```
matejcik, line 182: just `kwargs.get(...).map(TryInto::try_into).ok_or(false)`

dtto below

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+
+extern "C" fn storagedevice_get_label() -> Obj {
+    let block = || LABEL.get().map_or(Ok(Obj::const_none()), |x| x.try_into());
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn storagedevice_set_label(label: Obj) -> Obj {
+    let block = || {
+        let label: StrBuffer = label.try_into()?;
+        LABEL.set(label.try_into()?)?.try_into()
```
matejcik, line 137: worth considering something like:

```
// storage_field.rs
pub trait FieldSetSlice<T> {
    fn set_slice(&self, slice: &T);
}

impl FieldSetSlice<str> for Field<String> { ... }
impl FieldSetSlice<[u8]> for Field<Vec<u8>> { ... }
```

then you can use `LABEL.set(label.as_ref())` without the additional `?`. it's a little less try_intos and we avoid the unnecessary `String` or `Vec` middle step

matejcik:
```rust
impl FieldGetSet<String> for Field<String> {
    // ...

    fn set(&self, value: String) -> StorageResult {
        self.set_slice(value.as_ref())
    }
}

File: core/embed/rust/src/storagedevice/storage_device.rs [hunk 3]
```diff
@@ -0,0 +1,690 @@
...
+
+extern "C" fn storagedevice_set_u2f_counter(count: Obj) -> Obj {
+    let block = || {
+        let key = helpers::get_appkey_u2f(APP_DEVICE, U2F_COUNTER, true);
+        storage::set_counter(key, count.try_into()?)?.try_into()
+    };
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn storagedevice_get_private_u2f_counter() -> Obj {
```
matejcik, line 439: leftover?

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+    let block = || {
+        U2F_COUNTER_PRIVATE
+            .get()
+            .map_or(Ok(Obj::const_none()), |x| Ok(x.into()))
+    };
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn storagedevice_delete_private_u2f_counter() -> Obj {
```
matejcik, line 448: leftover?

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+    let block = || EXPERIMENTAL_FEATURES.delete()?.try_into();
+    unsafe { util::try_or_raise(block) }
+}
+
+fn _normalize_autolock_delay(delay_ms: u32) -> u32 {
+    if delay_ms < AUTOLOCK_DELAY_MINIMUM {
```
matejcik, line 454: `delay_ms.max(AUTOLOCK_DELAY_MINIMUM).min(AUTOLOCK_DELAY_MAXIMUM)`

matejcik (changes requested):

core/embed/rust/src/storagedevice/helpers.rs
```diff
@@ -0,0 +1,71 @@
...
+pub fn get_appkey_u2f(app: u8, key: u8, writable_locked: bool) -> u16 {
+    let app = if writable_locked {
+        app | FLAGS_WRITE
+    } else {
+        app | FLAG_PUBLIC
+    };
+    ((app as u16) << 8) | key as u16
+}
+
+pub fn hexlify_bytes<const N: usize>(bytes: &[u8]) -> String<N> {
```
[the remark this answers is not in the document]
matejcik: in fact we probably won't need the helper, the one case where we do actually use the `String<N>` might be better handled by the proposed `set_slice()` call and a conversion from `[u8] -> str`

core/embed/rust/src/micropython/obj.rs
```diff
@@ -256,6 +258,46 @@ impl TryFrom<&[u8]> for Obj {
...
 
+impl TryFrom<()> for Obj {
+    type Error = Error;
+
+    fn try_from(_val: ()) -> Result<Self, Self::Error> {
+        catch_exception(Self::const_none)
+    }
+}
+
+impl<const N: usize> TryFrom<Vec<u8, N>> for Obj {
```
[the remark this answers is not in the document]
matejcik: In fact, (a) there is no reason to call unsafe code here, and (b) perhaps we don't actually need this code? There is a perfectly good `micropython::list` wrapper.

core/embed/rust/src/trezorhal/storage_field.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    pub const fn exact(self) -> Self {
+        Self {
+            max_len_is_exact: true,
+            ..self
+        }
+    }
+
+    // TODO: would it be worth calculating it once and store as an attribute?
```
mmilata, line 63: I don't think so.

core/embed/rust/src/trezorhal/storage_field.rs
```diff
@@ -0,0 +1,379 @@
...
+        field.delete()
+    }
+}
+
+impl<const N: usize> FieldGetSet<String<N>> for Field<String<N>> {
+    fn get(&self) -> Option<String<N>> {
+        let mut buf = [0u8; MAX_STRING];
+        let len = storage::get(self.appkey(), &mut buf);
+        len.ok().and_then(|len| {
+            let string = String::from(str::from_utf8(&buf[..len as usize]).unwrap());
```
mmilata, line 197: In micropython we raise an error after failing to decode a string so maybe returning `None` would be more appropriate here?

core/embed/rust/src/trezorhal/storage_field.rs
```diff
@@ -0,0 +1,379 @@
...
+            if self.is_len_ok(string.chars().count(), N) {
+                Some(string)
+            } else {
+                None
+            }
+        })
+    }
+
+    fn set(&self, val: String<N>) -> StorageResult<()> {
+        if self.is_len_ok(val.chars().count(), N) {
```
mmilata, line 207: Chars vs. bytes

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+
+// Length of SD salt auth tag.
+// Other SD-salt-related constants are in core/src/storage/sd_salt.py
+const SD_SALT_AUTH_KEY_LEN_BYTES: usize = 16;
+
+// TODO: how to export all public constants as integers?
+
+const DEVICE_ID: Field<String<24>> = Field::public(APP_DEVICE, 0x00).exact();
+const VERSION: Field<Vec<u8, 1>> = Field::private(APP_DEVICE, 0x01).exact();
+const MNEMONIC_SECRET: Field<Vec<u8, 256>> = Field::private(APP_DEVICE, 0x02);
```
mmilata, line 37: Maybe leave a comment that 3 was supposed to be LANGUAGE?

core/src/storage/device_old.py
```diff
@@ -0,0 +1,56 @@
...
+    from typing_extensions import Literal
+
+
+# fmt: off
+# Keys:
+DEVICE_ID                  = const(0x00)  # bytes
+# MOVE to storage FIDO2
+U2F_COUNTER                = const(0x09)  # int
+# TODO: move to init
+INITIALIZED                = const(0x13)  # bool (0x01 or empty)
```
mmilata, line 17: Why keep these here? It should be possible to expose the constants in `obj_module!`.

mmilata (commented):
Just a few observations, didn't review it fully as it is massive.

Also, what are the plans with `modtrezorcrypto`? Seems like it mostly contains logic related to storage, and not a lot of it.

matejcik:
> what are the plans with modtrezorcrypto?

did you mean `modtrezorconfig`? I suppose we could remove a couple functions that are only used through storage (get/set/delete/etc)

we still need to keep init/lock/unlock there unless we RiiR

mmilata:
Uhh right `modtrezorconfig`. In master most of the storage lookups (`common.py`) go through `modtrezorconfig` which contains additional validation (like rejecting `app == 0`). In this branch some functions in `modtrezorconfig` are left unused.

core/embed/unix/mpconfigport.h
```diff
@@ -198,6 +198,10 @@ extern const struct _mp_print_t mp_stderr_print;
 
 #define MICROPY_PY_TREZORCONFIG     (1)
 #define MICROPY_PY_TREZORCRYPTO     (1)
+#define MICROPY_PY_TREZORSTORAGEDEVICE (1)
```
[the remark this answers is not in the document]
grdddj: Done in [a51936b](https://github.com/trezor/trezor-firmware/pull/2214/commits/a51936bb7cdb83629e8351f30532e2580fdb8534)

core/tools/analyze_firmware_code_size.py
```diff
@@ -0,0 +1,253 @@
...
+CURRENT_DIR = Path(__file__).resolve().parent
+LATEST_RESULTS_FILE = CURRENT_DIR / "fw_size_profiling_latest.json"
+BENCHMARK_FILE = CURRENT_DIR / "fw_size_profiling_benchmark.json"
+
+# Makes sure the paths are relative to the `core` directory
+print("changing directory to `core`")
+os.chdir(CURRENT_DIR.parent)
+
+BUILD_CMD = "make build_unix"
+BIN_TO_ANALYZE = "build/unix/trezor-emu-core"
```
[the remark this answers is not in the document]
grdddj: Included in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/storagedevice/storage_device.rs
grdddj: Done in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/trezorhal/random.rs
```diff
@@ -14,3 +17,28 @@ pub fn shuffle<T>(slice: &mut [T]) {
         slice.swap(i, j);
     }
 }
+
+// TODO: it seems it gets the same results every time? (could be seeded with
+// device seed)
+pub fn get_random_bytes<const N: usize>() -> Vec<u8, N> {
```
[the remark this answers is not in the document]
grdddj: Done in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa) - and in the only current usage, there is really no need to convert it to `Vec`

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+            Err(Error::ValueError(cstr!("Not valid rotation")))
+        } else {
+            ROTATION.set(rotation)?.try_into()
+        }
+    };
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn storagedevice_get_label() -> Obj {
+    let block = || LABEL.get().map_or(Ok(Obj::const_none()), |x| x.try_into());
```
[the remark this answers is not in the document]
grdddj: In [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa) I could not make it work for all the situations, as some of those are not `try_into`s, but only `into`s (u8, u16). Also in case of `Vec<u8>`, to return bytes, we need to use `(&v as &[u8]).try_into()` to return bytes, opposed to list of integers.

There is probably a lot to improve here

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+    let block = || {
+        VERSION
+            .set(Buffer::try_from(version)?.try_into()?)?
+            .try_into()
+    };
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn storagedevice_is_initialized() -> Obj {
+    let block = || Ok(INITIALIZED.get().unwrap_or(false).into());
```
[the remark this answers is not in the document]
grdddj: Done in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+const LABEL_MAXLENGTH: usize = 32;
+
+// Length of SD salt auth tag.
+// Other SD-salt-related constants are in core/src/storage/sd_salt.py
+const SD_SALT_AUTH_KEY_LEN_BYTES: usize = 16;
+
+// TODO: how to export all public constants as integers?
+
+const DEVICE_ID: Field<String<24>> = Field::public(APP_DEVICE, 0x00).exact();
+const VERSION: Field<Vec<u8, 1>> = Field::private(APP_DEVICE, 0x01).exact();
```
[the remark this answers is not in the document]
grdddj: Done in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+const SD_SALT_AUTH_KEY: Field<Vec<u8, SD_SALT_AUTH_KEY_LEN_BYTES>> =
+    Field::public(APP_DEVICE, 0x12).exact();
+const INITIALIZED: Field<bool> = Field::public(APP_DEVICE, 0x13);
+const SAFETY_CHECK_LEVEL: Field<u8> = Field::private(APP_DEVICE, 0x14);
+const EXPERIMENTAL_FEATURES: Field<bool> = Field::private(APP_DEVICE, 0x15);
+
+const SAFETY_CHECK_LEVEL_STRICT: u8 = 0;
+const SAFETY_CHECK_LEVEL_PROMPT: u8 = 1;
+const DEFAULT_SAFETY_CHECK_LEVEL: u8 = SAFETY_CHECK_LEVEL_STRICT;
+const SAFETY_CHECK_LEVELS: [u8; 2] = [SAFETY_CHECK_LEVEL_STRICT, SAFETY_CHECK_LEVEL_PROMPT];
```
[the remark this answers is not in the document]
grdddj: Partly done in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa) - Enum is created, but I could not find how to convert it to iterable then

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+const ROTATION: Field<u16> = Field::public(APP_DEVICE, 0x0F);
+const SLIP39_IDENTIFIER: Field<u16> = Field::private(APP_DEVICE, 0x10);
+const SLIP39_ITERATION_EXPONENT: Field<u8> = Field::private(APP_DEVICE, 0x11);
+const SD_SALT_AUTH_KEY: Field<Vec<u8, SD_SALT_AUTH_KEY_LEN_BYTES>> =
+    Field::public(APP_DEVICE, 0x12).exact();
+const INITIALIZED: Field<bool> = Field::public(APP_DEVICE, 0x13);
+const SAFETY_CHECK_LEVEL: Field<u8> = Field::private(APP_DEVICE, 0x14);
+const EXPERIMENTAL_FEATURES: Field<bool> = Field::private(APP_DEVICE, 0x15);
+
+const SAFETY_CHECK_LEVEL_STRICT: u8 = 0;
```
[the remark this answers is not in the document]
grdddj: Enum created in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/trezorhal/storage_field.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+
+    // TODO: would it be worth calculating it once and store as an attribute?
+    fn appkey(&self) -> u16 {
+        let app = self.app | self.access as u8;
+        (app as u16) << 8 | self.key as u16
+    }
+
+    pub fn has(&self) -> bool {
```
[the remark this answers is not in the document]
grdddj: Done in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/storagedevice/helpers.rs
```diff
@@ -0,0 +1,71 @@
...
+
+const FLAG_PUBLIC: u8 = 0x80;
+const FLAGS_WRITE: u8 = 0xC0;
+
+pub fn get_appkey(app: u8, key: u8, is_public: bool) -> u16 {
+    let app = if is_public { app | FLAG_PUBLIC } else { app };
+    ((app as u16) << 8) | key as u16
+}
+
+pub fn get_appkey_u2f(app: u8, key: u8, writable_locked: bool) -> u16 {
```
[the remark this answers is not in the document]
grdddj: Done in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa) and moved into `storage_field.rs`

core/embed/extmod/rustmods/modtrezorstoragedevice.c
```diff
@@ -0,0 +1,29 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "py/runtime.h"
+
+#if MICROPY_PY_TREZORSTORAGEDEVICE
+
+#include "librust.h"
+
+MP_REGISTER_MODULE(MP_QSTR_trezorstoragedevice, mp_module_trezorstoragedevice,
```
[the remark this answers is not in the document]
grdddj: Done in [a51936b](https://github.com/trezor/trezor-firmware/pull/2214/commits/a51936bb7cdb83629e8351f30532e2580fdb8534)

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -179,6 +202,19 @@ impl From<&'static str> for StrBuffer {
     }
 }
 
+impl<const N: usize> TryFrom<StrBuffer> for String<N> {
+    type Error = Error;
+
+    fn try_from(str_buf: StrBuffer) -> Result<String<N>, Self::Error> {
+        let slice = str::from_utf8(&str_buf.0).unwrap();
```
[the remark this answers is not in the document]
grdddj: It works fine, changed in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -79,6 +81,27 @@ impl<const N: usize> From<&'static [u8; N]> for Buffer {
     }
 }
 
+// impl From<&'static str> for Buffer {
```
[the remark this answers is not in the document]
grdddj: Removed in [ba44c83](https://github.com/trezor/trezor-firmware/pull/2214/commits/ba44c834a8c0c826cc6480f6e5c1c84a09fc12fa)

core/embed/rust/src/storagedevice/storage_device.rs
```diff
@@ -0,0 +1,690 @@
...
+    };
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn storagedevice_is_initialized() -> Obj {
+    let block = || Ok(INITIALIZED.get().unwrap_or(false).into());
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn storagedevice_set_is_initialized(is_initialized: Obj) -> Obj {
```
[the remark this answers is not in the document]
grdddj: I think this needs to be exposed, as we use it in `storage/__init__.py::init_unlocked` - previously there was `common.set_bool(...)`, but that does not exist anymore

core/embed/rust/src/storagedevice/storage_device.rs
grdddj: This and the one below are actually used, these are some strange cases being called in `storage/__init__.py::_migrate_from_version_01()`. Not sure why exactly they are needed, and even how to test it - as the migration could be only possible on the physical device

core/embed/rust/storage.h
```diff
@@ -0,0 +1,75 @@
+/*
```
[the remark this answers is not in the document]
grdddj: It does not need :) Deleted in [39ad89d](https://github.com/trezor/trezor-firmware/pull/2214/commits/39ad89d8ff73f886b38ddfcc03ba4ff4527d3674)

core/embed/rust/src/micropython/obj.rs
```diff
@@ -256,6 +258,46 @@ impl TryFrom<&[u8]> for Obj {
...
+    type Error = Error;
+
+    fn try_from(val: Vec<u8, N>) -> Result<Self, Self::Error> {
```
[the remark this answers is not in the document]
grdddj: This whole function deleted in [697ab5b](https://github.com/trezor/trezor-firmware/pull/2214/commits/697ab5bea7c18530b0dc1a11944905691c0f7caf), replaced by `List::from_iter`

core/embed/rust/src/micropython/obj.rs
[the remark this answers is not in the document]
grdddj: This whole function deleted in [697ab5b](https://github.com/trezor/trezor-firmware/pull/2214/commits/697ab5bea7c18530b0dc1a11944905691c0f7caf), replaced by List::from_iter

core/src/storage/device_old.py
grdddj: These are actually leftovers from the past, they are no longer needed anymore - currently all the storage handling is done in Rust.

Some other constants like `HOMESCREEN_MAXSIZE` really could be exposed, will try that

core/embed/rust/src/trezorhal/storage_field.rs
grdddj: Changed to 
```rust
let string = match str::from_utf8(&buf[..len as usize]) {
        Ok(s) => String::from(s),
        Err(_) => return None,
};
```

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -179,6 +202,19 @@ impl From<&'static str> for StrBuffer {
     }
 }
 
+impl<const N: usize> TryFrom<StrBuffer> for String<N> {
+    type Error = Error;
+
+    fn try_from(str_buf: StrBuffer) -> Result<String<N>, Self::Error> {
+        let slice = str::from_utf8(&str_buf.0).unwrap();
+        if slice.chars().count() > N {
```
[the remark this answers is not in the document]
grdddj: Not completely sure, but with `slice.len()` it works the same, so changing to it (checked by `storage_field.rs::test_value_too_long_str` unittest)

core/embed/rust/src/trezorhal/storage_field.rs
grdddj: Changed to `val.len()`.

It is a question if this scenario can even happen, my guess is probably not, as `String<N>` needs to be already created, and if it is longer than `N`, then it will fail to create - with `String is too long to fit` error from `buffer.rs`.

Are we safe to delete the `is_len_ok` check for both the `String<N>` and `Vec<u8, N>`?

core/embed/rust/src/storagedevice/storage_device.rs
grdddj: > we avoid the unnecessary String or Vec middle step

Not sure how we could do it, as `String` and `Vec` still need to be in the `set()` signature

> it's a little less try_intos

We would need anyway to copy the `impl<const N: usize> TryFrom<StrBuffer> for String<N>` logic with length validation

core/embed/rust/src/storagedevice/helpers.rs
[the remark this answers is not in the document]
grdddj: > instead we can use [hex](https://lib.rs/crates/hex)?

I am afraid this dependency would take a lot of space - but will try it and see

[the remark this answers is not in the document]
grdddj: Comparison of the current `hexlify_bytes` and `hex` state:
```rust
    #[test]
    fn test_hexlify_bytes() {
        let bytes: &[u8; 6] = &[52, 241, 6, 151, 173, 74];
        let result: String<12> = hexlify_bytes(bytes);
        assert_eq!(result, String::<12>::from("34F10697AD4A"));
    }

    use hex;
    use core::str;
    #[test]
    fn test_hexlify_bytes_2() {
        let bytes: &[u8; 6] = &[52, 241, 6, 151, 173, 74];
        let mut result = [0u8; 6 * 2];
        hex::encode_to_slice(bytes, &mut result).unwrap();
        let mut string: String<12> = String::from(str::from_utf8(&result).unwrap());
        string.make_ascii_uppercase();
        assert_eq!(string, String::<12>::from("34F10697AD4A"));
    }
```

I am probably in favor of keeping the current `hexlify_bytes` and moving it into shared `helpers.rs` under `rust/src`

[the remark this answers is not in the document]
grdddj: Space-wise, after the change, `get_device_id` function increased by `116 b` in `build/firmware/firmware.elf`

core/embed/rust/storage.h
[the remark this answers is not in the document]
grdddj: > please also investigate why we need the cache song-and-dance for `get_experimental_features`.
> 
> My first suspicion is that it is because we want the field to be public but it is private.

I have found the connected [PR](https://github.com/trezor/trezor-firmware/pull/1565) and [commit](https://github.com/trezor/trezor-firmware/commit/96fd347ca895b6fc57bec7f3746fd791648d0931) and it seems to solve some issue that was in Trezor before that - something with workflow restarts

grdddj:
> please also investigate why we need the cache song-and-dance for `get_experimental_features`.
> 
> My first suspicion is that it is because we want the field to be public but it is private.

I have found the connected [PR](https://github.com/trezor/trezor-firmware/pull/1565) and [commit](https://github.com/trezor/trezor-firmware/commit/96fd347ca895b6fc57bec7f3746fd791648d0931) and it seems to solve some issue that was in Trezor before that - something with workflow restarts

core/embed/rust/src/micropython/obj.rs
```diff
@@ -256,6 +258,14 @@ impl TryFrom<&[u8]> for Obj {
     }
 }
 
+impl TryFrom<()> for Obj {
+    type Error = Error;
+
+    fn try_from(_val: ()) -> Result<Self, Self::Error> {
+        catch_exception(Self::const_none)
```
mmilata, line 265: `const_none` currently doesn't throw exceptions as it doesn't call micropython code, and I think it's reasonable to assume that it will continue in the future.

core/embed/rust/src/micropython/obj.rs
```diff
@@ -270,6 +280,17 @@ impl TryFrom<&str> for Obj {
...
 }
 
+impl<const N: usize> TryFrom<String<N>> for Obj {
+    type Error = Error;
+
+    fn try_from(val: String<N>) -> Result<Self, Self::Error> {
+        // SAFETY:
+        //  - `str` is guaranteed to be UTF-8.
+        // EXCEPTION: Will raise if allocation fails.
+        catch_exception(|| unsafe { ffi::mp_obj_new_str(val.as_ptr().cast(), val.len()) })
```
mmilata, line 290:
```suggestion
        (*val).try_into()
```
to call the `&str` impl

core/embed/rust/src/micropython/obj.rs
```diff
@@ -385,11 +406,39 @@ impl TryFrom<Obj> for usize {
     }
 }
 
-impl<T> From<Option<T>> for Obj
-where
-    T: Into<Obj>,
-{
-    fn from(val: Option<T>) -> Self {
+impl<const N: usize> TryFrom<Option<Vec<u8, N>>> for Obj {
```
mmilata, line 409: Can't we have both `From<Option<T>>` and `TryFrom<Option<T>>` defined in generic way? Why do it for each type separately?

core/tests/test_apps.management.recovery_device.py
```diff
@@ -28,8 +27,10 @@
 
 class TestSlip39(unittest.TestCase):
 
-    @mock_storage
     def test_process_slip39_basic(self):
+        config.init()
+        config.wipe()
```
mmilata, line 32: Wanna put this into `def setUp(self):`?

core/embed/rust/src/storagedevice/recovery.rs
```diff
@@ -0,0 +1,267 @@
...
+use heapless::Vec;
+
+// The maximum number of shares/groups that can be created.
+const MAX_SHARE_COUNT: usize = 16;
+const MAX_GROUP_COUNT: usize = 16;
+
+const DEFAULT_SLIP39_GROUP_COUNT: u8 = 1;
+
+const APP_RECOVERY: u8 = 0x02;
+const APP_RECOVERY_SHARES: u8 = 0x03;
```
mmilata, line 18: Unused?

Maybe we should turn all the app numbers into a single `enum`?

core/embed/rust/src/trezorhal/storage_field.rs
```diff
@@ -0,0 +1,391 @@
...
+impl<const N: usize> FieldGetSet<String<N>> for Field<String<N>> {
+    fn get(&self) -> Option<String<N>> {
+        let mut buf = [0u8; MAX_STRING];
+        let len = storage::get(self.appkey(), &mut buf);
+        len.ok().and_then(|len| {
+            let string = match str::from_utf8(&buf[..len as usize]) {
+                Ok(s) => String::from(s),
+                Err(_) => return None,
+            };
+            if self.is_len_ok(string.chars().count(), N) {
```
mmilata, line 192: Here as well. It's going to break for non-ASCII strings.

core/embed/rust/src/trezorhal/storage_field.rs
```diff
@@ -0,0 +1,391 @@
...
+        field.delete()
+    }
+}
+
+impl<const N: usize> FieldGetSet<String<N>> for Field<String<N>> {
+    fn get(&self) -> Option<String<N>> {
+        let mut buf = [0u8; MAX_STRING];
+        let len = storage::get(self.appkey(), &mut buf);
+        len.ok().and_then(|len| {
+            let string = match str::from_utf8(&buf[..len as usize]) {
```
mmilata, line 188: Maybe return the subslice directly from `storage::get(appkey: u16, data: &mut [u8]) -> StorageResult<&[u8]>`?

core/embed/rust/src/trezorhal/storage_field.rs
mmilata: Yes, `String<N>` is guaranteed to be at most `N` bytes long, though `is_len_ok` checks for exact length in some cases so we should leave it?

Hannsek:
Old PR, closing because inactivity. If needed, we can reopen.

PR #2218, opened 2022-04-13
Add Pulsar Coin Support

PulsarCoin:


matejcik:
Pulsar Coin is not supported in firmware so there is no point in adding a misc.json entry

PulsarCoin:
Would it be correct to add a file in trezor-firmware/common/defs/bitcoin?

Or is Pulsar support needed elsewhere first?

Thank you!
Mason Smith

matejcik:
If Pulsar is based on Bitcoin, then adding to `defs/bitcoin` would be appropriate, yes.

But be advised that we are currently not accepting new coin implementations anyway.

PulsarCoin:
Thank you for the response,

We will take that into consideration. Do you have a timeline of when you
are accepting new coins? We would really appreciate the opportunity to work
with you guys to implement our coin.

Let me know!

PR #2220, opened 2022-04-13
Fix flash and openocd rules in makefile

TychoVrahe:
When there are comments after make variable definition, make considers spaces between the comment and variable value as a part of the value, which breaks makefile rules when using default values (spaces are added to filename etc)

Therefore, in this PR, comments on openocd interface and transport default values are moved to separate line


prusnak:
What version of make do you use and under what OS?

TychoVrahe:
> What version of make do you use and under what OS?

GNU Make 4.2.1
Ubuntu 20.04.4 LTS

PR #2222, opened 2022-04-19
Fixing compilation for different trezor models 

TychoVrahe:
Fixing compilation for different trezor models (1, T, etc)

Compiling core for different trezor models works only partially. TREZOR_MODEL is defined in makefile (or via command line) as 1 or T, these are then passed to gcc and is interpreted as a number or a preprocessor token T, which is undefined. The preprocessor directives compare TREZOR_MODEL to either 1 or undefined token T. Undefined token ultimately evaluates to 0, thus the distinction works. But if another model is added, for example Model X, the undefined X in comparison would also evaluate to 0 and both conditions for T and X would be satisfied. 

To fix this, another -D macro (using SConscript) is added for gcc invocation - MODEL_T or MODEL_1, and in C files then use this token for evaluation what to build via `#if defined`. (It would be also possible to pass the constant to gcc as a char and then compare with char, but it only works for single character model names, and we would still need the original TREZOR_MODEL constant with the undefined T, since it is used for generating QSTR for micropython environment and it actually relies on the T symbol being undefined). 

Developers don't need to change anything when compiling, but for correct IDE indexing function, the corresponding MODEL_x macro has to be added.

This PR only affects C portion of the core code.

### DIFF
--- a/core/SConscript.boardloader
+++ b/core/SConscript.boardloader
@@ -131,6 +131,7 @@ env.Replace(
     ] + CPPPATH_MOD,
     CPPDEFINES=[
         ('TREZOR_MODEL', '$TREZOR_MODEL'),
+        'MODEL_'+TREZOR_MODEL,
         CPU_MODEL,
         'USE_HAL_DRIVER',
         ('STM32_HAL_H', '"<stm32f4xx.h>"'),
--- a/core/SConscript.bootloader
+++ b/core/SConscript.bootloader
@@ -159,6 +159,7 @@ env.Replace(
     ] + CPPPATH_MOD,
     CPPDEFINES=[
         ('TREZOR_MODEL', '$TREZOR_MODEL'),
+        'MODEL_'+TREZOR_MODEL,
         CPU_MODEL,
         'USE_HAL_DRIVER',
         ('STM32_HAL_H', '"<stm32f4xx.h>"'),
--- a/core/SConscript.bootloader_ci
+++ b/core/SConscript.bootloader_ci
@@ -159,6 +159,7 @@ env.Replace(
     ] + CPPPATH_MOD,
     CPPDEFINES=[
         ('TREZOR_MODEL', '$TREZOR_MODEL'),
+        'MODEL_'+TREZOR_MODEL,
         CPU_MODEL,
         'USE_HAL_DRIVER',
         ('STM32_HAL_H', '"<stm32f4xx.h>"'),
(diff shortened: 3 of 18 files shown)


prusnak:
Please use TREZOR_MODEL instead of MODEL.

matejcik:
> since it is used for generating QSTR for micropython environment and it actually relies on the T symbol being undefined

can you elaborate?

matejcik:
LGTM but please rename the new variables to `TREZOR_MODEL_*` as per https://github.com/trezor/trezor-firmware/pull/2222#issuecomment-1102237582

core/SConscript.boardloader
```diff
@@ -131,6 +131,7 @@ env.Replace(
     ] + CPPPATH_MOD,
     CPPDEFINES=[
         ('TREZOR_MODEL', '$TREZOR_MODEL'),
```
prusnak, line 133: We should remove these  
```suggestion
```


and only keep using `TREZOR_MODEL_X` defines

prusnak (changes requested):

core/SConscript.boardloader
matejcik: We need to know the model value for the Qstr. But I suppose we can make it conditional on each model def, if there are no other usecases like that.

prusnak: > But I suppose we can make it conditional on each model def, if there are no other usecases like that.

Good point. Let's do it like this.

prusnak:
As noted in https://github.com/trezor/trezor-firmware/pull/2230#issue-1214788304 we are still erroneously using `#if TREZOR_MODEL == x` in `core/embed/firmware/header.S` and `core/embed/firmware/startup.S`.

TychoVrahe:
> As noted in [#2230 (comment)](https://github.com/trezor/trezor-firmware/pull/2230#issue-1214788304) we are still erroneously using `#if TREZOR_MODEL == x` in `core/embed/firmware/header.S` and `core/embed/firmware/startup.S`.

fixed b03dd5a69f6c34886bd3ba7879aa21548ec94496

prusnak (approved):
untested ACK

Last nitpick: please fix commit messages to conform to format described in COMMITS.md.
We use `build(core): foo` not `build(core) - foo`.

TychoVrahe:
> please fix commit messages

Will have to force push, #2230 already builds on this, is force acceptable?

prusnak:
> is force acceptable?

Yes. #2230 needs to be rebased anyway and I don't expect any conflicts.

matejcik (approved):
utACK

matejcik:
I went ahead and squash-merged instead of waiting for the commit renames

PR #2223, opened 2022-04-19
fix(core/ui): add missing sources to build

mmilata:
In 2f987c3c5e9a8620c6333e6327057f1743cb5dac @ #2211 I failed to update `SConscript.firmware` which resulted in RSOD on startup. Also explains why nightly hardware tests are failing.


matejcik (approved):
utACK

PR #2227, opened 2022-04-21
Rust UI: PIN correction + long PIN

### LINKED ISSUE #1469: PIN entry correction on TT
**Is your feature request related to a problem? Please describe.**
The PIN entry screen on the TT shows dots instead of the entered digits. If the user is unsure whether they made a mistake while entering the PIN, then they have no way to check and have to either re-enter the entire PIN again or risk losing a PIN entry attempt. Even if the user knows that they made a mistake, pressing the "backspace" button erases the entire PIN, so they have to retype the whole thing. 

**Describe the solution you'd like**
- Allow the user to temporarily reveal the PIN that they are entering by touching the entry field. As soon as they remove their finger the field will show dots again.
- Change the behavior of the backspace button so that it erases one digit at a time. If pressed and held for 2 seconds or more, then make it erase the whole field.

**Additional context**
This is related to the implementation of the unlimited length PIN https://github.com/trezor/trezor-firmware/issues/1167. Long PINs make mistakes not only more likely but also more costly if the user doesn't have a way to correct them. Note that the entry field is currently center-aligned. For long PINs we want to make it behave like the passphrase entry field.

mmilata:
Fixes #1469.

pin

* yellow button now erases one character and holding it for 2s erases everything
* holding finger on dot display reveals the pin
* PIN can be up to 50 digits long and renders reasonably even if it doesn't fit screen
* overflowing PIN animated (yet) but using the same "twitching" as in original UI


mmilata:
Gonna piggyback on this PR and add [fdd7cee](https://github.com/trezor/trezor-firmware/pull/2227/commits/fdd7ceee7164ac27be23412703bb30b709046815) which fixes rendering of multiple overlapping `Maybe<T>` components like in mnemonic input layout.

grdddj:
I looked at it briefly and it works fine on the emulator.

In error cases, however (bad PIN, the second PIN is different than the first one), it throws `NotImplementedError` coming from `_show_modal` or `_show_popup` layout functions, these are missing.

Theoretically, I might borrow some of these things into the model `R` PIN entry - for example the `LongPressed` functionality, displaying of the very-long PIN, etc.

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -273,7 +273,7 @@ impl LayoutObj {
         static TYPE: Type = obj_type! {
             name: Qstr::MP_QSTR_Layout,
             locals: &obj_dict!(obj_map! {
-                Qstr::MP_QSTR_set_timer_fn => obj_fn_2!(ui_layout_set_timer_fn).as_obj(),
+                Qstr::MP_QSTR_attach_timer_fn => obj_fn_2!(ui_layout_attach_timer_fn).as_obj(),
```
matejcik, line 276: this makes it sound like we are "attaching a timer function", but instead this is "set timer function and attach to tree", right?

maybe we could call this `init` ?

matejcik (commented):
very pretty!

core/embed/rust/src/ui/layout/obj.rs
mmilata: That's correct. `init` sounds rather nebulous, what about something like `init_with_timer_fn` or `set_timer_fn_and_attach`?

matejcik: Ah, right, I understand the issue a little better now.

We need to send the `Attach` event at the time when the layout actually goes to screen. (this is now `handle_input_and_rendering`). The "set timer callback" operation is a functionally separate thing, but because we can't get the timer in constructor (where it seems to belong), we bundle the two things together.

("set timer callback" can be called arbitrarily and the implementation does not care for as long as it still receives the timer events when expected. "attach" can be called repeatedly to indicate that the same layout instance is put on screen again)

(speaking of, the current `Timer` implementation is very fragile. it doesn't clean up after itself, and if `handle_timers` did any await other than on that one timer, it would mess things up because it happily schedules the task multiple times. we should clean that up before this goes to production.
in fact a good opportunity would be allowing a single global layout with a single global timer that would reset all scheduled events every time something went on screen. this thing might then as well live mostly in Rust.)

considering the parenthesized content :) let's keep the name as-is and make it better in the future

matejcik (approved):
LGTM, please fix the conflict and merge

Issue #2216, opened 2022-04-12
Trezor R emulator up and running


matejcik:
The current T1-in-core code should be copied and modified for Model R.

We should use the existing image background, but modify the display to be twice as tall (so 128x128).

We should also port the existing T1 Rust layouts (basic confirmation, signtx flow) to the taller screen.

PR #2230, opened 2022-04-25
Basic R emulator

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2216

Creating a possibility to build and use `R` emulator via `TREZOR_MODEL=R make build_unix`.

It contains just some very basic layouts coming from `t1-in-core` rust layouts.
(Actually, it still uses `model_t1` rust feature, as the only thing that changed so far was `constants.rs::HEIGHT` from `64` to `128`. This is breaking the `T1` layout, but I was unable to properly create `model_tr`, there were issues with `#[cfg(feature`.)

All the `micropython` layouts defined in `core/src/trezor/ui/layouts/tr/__init__.py` are made just to work quickly, calling `trezorui2.confirm_action` and `trezorui2.confirm_text`.

It should support all the basic layouts needed to sign a `BTC` transaction via `Suite` and have basic `trezorctl` interaction.

Built on top of https://github.com/trezor/trezor-firmware/pull/2222 to be able to define and use a third `TREZOR_MODEL` - `R`.

Regarding the commit above, I have found two files where we are still using `TREZOR_MODEL == T` instead of `defined TREZOR_MODEL_T` - `core/embed/firmware/header.S ` and `core/embed/firmware/startup.S` - not sure if these files are working differently.


core/embed/unix/jpg_to_h.py
```diff
@@ -0,0 +1,34 @@
...
+h_file_template = """\
+// clang-format off
+unsigned char background_R_jpg[] = {content};
+unsigned int background_R_jpg_len = {length};
+"""
+
+with open(background_image, "rb") as f:
+    image_data = f.read()
+
+column_count = 11
```
prusnak, line 18: Other files (1, T) use 12 columns, not 11.

core/SConscript.boardloader
```diff
@@ -97,7 +97,7 @@ env.Replace(
 env.Replace(
     TREZOR_MODEL=TREZOR_MODEL, )
 
-if TREZOR_MODEL == 'T':
+if TREZOR_MODEL == 'T' or TREZOR_MODEL == 'R':
```
prusnak, line 100: SConscript file is Python so we can use the following:

``` python
if TREZOR_MODEL in ['T', 'R']:
```

core/src/apps/base.py
```diff
@@ -75,6 +75,11 @@ def get_features() -> Features:
             Capability.ShamirGroups,
             Capability.PassphraseEntry,
         ]
+
+    # Model R is not capable of PassphraseEntry
+    if utils.MODEL == "R":
+        f.capabilities.remove(Capability.PassphraseEntry)
+
```
prusnak, line 82: This is not very nice. How about removing `PassPhraseEntry` from two lists above and adding the following?

``` python
if utils.MODEL == "T":
    f.capabilities.append(Capability.PassphraseEntry);
```

core/src/trezor/ui/__init__.py
```diff
@@ -49,7 +49,7 @@ def refresh() -> None:
 
 
 # in both debug and production, emulator needs to draw the screen explicitly
-if utils.EMULATOR or utils.MODEL == "1":
+if utils.EMULATOR or utils.MODEL == "1" or utils.MODEL == "R":
```
prusnak, line 52: This is Python, so we can use the following:

``` python
if utils.EMULATOR or utils.MODEL in ["1", "R"]:
```

core/src/apps/homescreen/homescreen.py
```diff
@@ -53,7 +53,14 @@ def do_render(self) -> None:
 
         # homescreen with shifted avatar and text on bottom
         ui.display.avatar(48, 48 - 10, self.get_image(), ui.WHITE, ui.BLACK)
-        ui.display.text_center(ui.WIDTH // 2, 220, self.label, ui.BOLD, ui.FG, ui.BG)
+        if utils.MODEL == "1" or utils.MODEL == "R":
```
prusnak, line 56: This is Python, so we can use the following:

``` python
if utils.MODEL in ["1", "R"]:
```

prusnak (changes requested):
Left couple of nits

grdddj:
> Most of the comments revolve around `TREZOR_MODEL` in `SConscript` files or `utils.MODEL` in Python sources.

Yeah, it is beneficial for the future. All the suggestions done in [133289a](https://github.com/trezor/trezor-firmware/pull/2230/commits/133289a0d10b3b5d1cf90ad48498b7849ac69faf)

core/src/apps/homescreen/homescreen.py
matejcik: s/`[]`/`()`, tuples are preferred for static lists of options
or `in "1R"` works for single-letter models too :)

prusnak:
Since https://github.com/trezor/trezor-firmware/pull/2222 has been (squashed and) merged, can you please rebase on top of current master?

grdddj:
> Since #2222 has been (squashed and) merged, can you please rebase on top of current master?

Rebased and force-pushed now. It already includes the suggestion to compare tuples instead of lists.

core/embed/firmware/startup.S
```diff
@@ -48,7 +48,7 @@ reset_handler:
   // according to "ARM Cortex-M Programming Guide to Memory Barrier Instructions" Application Note 321, section 4.7:
   // "If it is not necessary to ensure that a pended interrupt is recognized immediately before
   // subsequent operations, it is not necessary to insert a memory barrier instruction."
-#if TREZOR_MODEL == T
+#if TREZOR_MODEL == T || TREZOR_MODEL == R
```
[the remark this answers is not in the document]
grdddj: Changes in the force-push [7569eb0](https://github.com/trezor/trezor-firmware/pull/2230/commits/7569eb069e0f62fedd229eeef6b5cbf7fca773ce)

core/SConscript.unix
```diff
@@ -558,7 +560,7 @@ if FROZEN:
             SOURCE_PY_DIR + 'trezor/enums/DebugMonero*.py',
             SOURCE_PY_DIR + 'trezor/enums/Eos*.py',
             SOURCE_PY_DIR + 'trezor/enums/Ethereum*.py',
-            SOURCE_PY_DIR + 'trezor/enums/Monero*.py',
+            SOURCE_PY_DIR ,+ 'trezor/enums/Monero*.py',
```
prusnak, line 563: typo?
```suggestion
            SOURCE_PY_DIR + 'trezor/enums/Monero*.py',
```

core/embed/rust/src/ui/model_t1/constant.rs
```diff
@@ -1,7 +1,7 @@
 use crate::ui::geometry::{Offset, Point, Rect};
 
 pub const WIDTH: i32 = 128;
-pub const HEIGHT: i32 = 64;
+pub const HEIGHT: i32 = 128;
```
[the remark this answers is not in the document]
grdddj: Done in [01363d6](https://github.com/trezor/trezor-firmware/pull/2230/commits/01363d64d9b4e8e28fe20980c480cd4b5f1a4887) - created new `model_tr` feature. Having this feature adds some complexity into conditional compilation - sometimes it (VSCode) thinks something is defined multiple times, for example in 
```rust
// core/embed/rust/src/ui/layout/obj.rs
#[cfg(feature = "model_t1")]
use crate::ui::model_t1::constant;
#[cfg(feature = "model_tr")]
use crate::ui::model_tr::constant;
#[cfg(feature = "model_tt")]
use crate::ui::model_tt::constant;
```

, it does not know only one of those can be true during compile-time

prusnak (changes requested):

core/SConscript.firmware
```diff
@@ -561,15 +561,15 @@ if FROZEN:
     if EVERYTHING:
         SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/altcoin.py'))
         SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/webauthn.py'))
-    if TREZOR_MODEL == 'T' and UI2:
+    if TREZOR_MODEL in ('T') and UI2:
```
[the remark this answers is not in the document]
grdddj: Thanks for noticing, it probably has a connection with the extra coma below. Both fixed in [a0820c1](https://github.com/trezor/trezor-firmware/pull/2230/commits/a0820c162b94f7422beb39273b18dd915e98d0ed)

grdddj:
Last fixup - [8bfc3b0](https://github.com/trezor/trezor-firmware/pull/2230/commits/8bfc3b0fe73b31ca9668308bd600cf2ccc07b5f4) - fixes CI's `gen` and `style` tests.

It also renames all components in `core/embed/rust/src/ui/model_tr` folder to use `TR` instead of `T1` as copy-pasted there

core/embed/firmware/memory_R.ld
```diff
@@ -0,0 +1,81 @@
+/* TREZORv2 firmware linker script */
```
mmilata, line 1: Will this file ever be different from memory_T.ld? Should we make it a symlink until then?

mmilata (approved):
Changes look good to me. A few comments inline. Did not test on hw.

core/embed/firmware/header.S
```diff
@@ -26,8 +26,8 @@ g_header:
     . = . + 8                        // reserved
     . = . + 512                      // hash1 ... hash16
 
-#if defined TREZOR_MODEL_T
-// model T header specifics
+#if defined TREZOR_MODEL_T || defined TREZOR_MODEL_R
```
matejcik, line 29: let's flip this, because the comments aren't strictly correct
```
#if !defined TREZOR_MODEL_1
// trezor-core header style
...
#else
// model 1 compatibility header
```

matejcik: please do that in a separate commit because it is not emulator related

core/embed/rust/src/ui/component/base.rs
```diff
@@ -4,6 +4,8 @@ use heapless::Vec;
 
 #[cfg(feature = "model_t1")]
 use crate::ui::model_t1::event::ButtonEvent;
+#[cfg(feature = "model_tr")]
```
matejcik, line 7: instead: `#[cfg(any(feature = "model_t1", feature = "model_tr"))]`
dtto in all ButtonEvent cases

core/embed/firmware/memory_R.ld
prusnak: +1 for symlink

python/src/trezorlib/models.py
```diff
@@ -33,7 +33,16 @@ class TrezorModel:
     default_mapping=mapping.DEFAULT_MAPPING,
 )
 
-TREZORS = {TREZOR_ONE, TREZOR_T}
+TREZOR_R = TrezorModel(
+    name="R",
+    # TODO: decide how to handle R versioning
```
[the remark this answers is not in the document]
grdddj: Sure. The version of `R` will be always the same as the version of `T`?

core/embed/rust/src/ui/component/base.rs
grdddj: Yeah, it would be the cleanest, but it is not possible, as the directory where the `event.rs` file lies is also different - `crate::ui::model_t1` vs `crate::ui::model_tr`.

During compilation, it would complain that in the case of `model_tr` feature the `model_t1` directory does not exist

grdddj: That brings the idea of putting it into some shared directory...

python/src/trezorlib/models.py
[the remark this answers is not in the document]
mmilata: IIRC there's some ongoing discussion regarding versioning but it's probably out of scope for this PR.

core/embed/rust/src/ui/display.rs
```diff
@@ -1,7 +1,9 @@
 use crate::{micropython::time, time::Duration, trezorhal::display};
 
-#[cfg(not(feature = "model_tt"))]
+#[cfg(feature = "model_t1")]
 use crate::ui::model_t1::constant;
+#[cfg(feature = "model_tr")]
+use crate::ui::model_tr::constant;
 #[cfg(feature = "model_tt")]
 use crate::ui::model_tt::constant;
```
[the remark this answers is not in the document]
matejcik: @grdddj does vscode actually break on this? For me it highlighted these lines as incorrect but otherwise the analyzer features worked fine. I'm OK with that result. I would go for explicit mutual exclusion only in case this breaks in a more serious way (e.g. impossible to follow definitions)

core/embed/rust/src/ui/component/base.rs
matejcik: right, I didn't notice that, sorry.

Yes, shared definitions are the way to go. In fact, it should be something like `ui/event.rs` defining both Button and Touch events. We might not even need these to be conditional on model, as long as the usages are conditional. What do you think @mmilata ?

mmilata: +1 for sharing

core/embed/rust/src/ui/display.rs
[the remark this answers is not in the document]
grdddj: It does not break anything, it just complains - no real error (not even warning) during compilation (which is completely outside of VSCode).

python/src/trezorlib/models.py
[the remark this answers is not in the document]
grdddj: Comment removed in [5cd212d](https://github.com/trezor/trezor-firmware/pull/2230/commits/5cd212d6b092437e82f102909ec1edd9837f82a3)

core/embed/firmware/memory_R.ld
grdddj: Symlink added in [5cd212d](https://github.com/trezor/trezor-firmware/pull/2230/commits/5cd212d6b092437e82f102909ec1edd9837f82a3)

core/embed/rust/src/ui/component/base.rs
grdddj: Shared `event.rs` created in [c042b45](https://github.com/trezor/trezor-firmware/pull/2230/commits/c042b4567c0ced0072ad166c7b12323cbdd1d6d9)

I am keeping the conditional use based on the model, otherwise, it would be complaining about unused imports

core/embed/firmware/header.S
grdddj: So in the end all the changes in `header.S`, `startup.S`, `SConscript.firmware` etc. - not touching emulator - should end up in a separate commit?

grdddj: Done in [91b1938](https://github.com/trezor/trezor-firmware/pull/2230/commits/91b1938c506c2ea02489c5c4ee561eab3cbe1f4b) - during squashing, I would transfer all non-emulator changes into this commit

matejcik (approved):
LGTM now. please squash & rebase, ideally moving non-emulator changes to their own commit, as you propose

grdddj:
Rebased, squashed, split into relevant commits, included changelogs

core/SConscript.unix
```diff
@@ -147,7 +147,7 @@ if FEATURE_FLAGS["SECP256K1_ZKP"]:
 SOURCE_MOD += [
     'embed/extmod/modtrezorio/modtrezorio.c',
 ]
-if TREZOR_MODEL == "T":
```
TychoVrahe, line 150: Any reason to use double quotes here unlike rest of the file?

core/SConscript.firmware
```diff
@@ -151,7 +151,7 @@ if FEATURE_FLAGS["SECP256K1_ZKP"]:
 SOURCE_MOD += [
     'embed/extmod/modtrezorio/modtrezorio.c',
 ]
-if TREZOR_MODEL == "T":
```
TychoVrahe, line 154: Any reason to use double quotes here unlike rest of the file?

TychoVrahe (commented):
Single/double quote usage in SConsrcipt files: inline

prusnak:
> Single/double quote usage in SConsrcipt files: inline

Once this is merged we should run `black core/SConscript.* core/SConstruct` and add that to the Python style-check.

matejcik:
ready to merge after tomorrow's release branching

PR #2231, opened 2022-04-26
feat(legacy/bootloader): report installed firmware version

matejcik:
This adds a capability to report installed firmware version to T1 bootloader.

The code is untested, because unfortunately I don't currently have means to test it.

### DIFF
--- a/legacy/bootloader/usb_send.h
+++ b/legacy/bootloader/usb_send.h
@@ -1,5 +1,3 @@
-#include <assert.h>
-
 static void send_msg_success(usbd_device *dev) {
   uint8_t response[64];
   memzero(response, sizeof(response));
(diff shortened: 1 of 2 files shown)


legacy/bootloader/usb_send.h
```diff
@@ -37,36 +39,71 @@ static void send_msg_failure(usbd_device *dev, uint8_t code) {
...
+    // msg_id
+    0x00, 0x11,
+    // msg_size
+    0x00, 0x00, 0x00, sizeof(feature_bytes) + (firmware_present ? sizeof(version_bytes) : 0),
+  };
+  // clang-format on
+
+  // Check that the response will fit into an USB packet, and also that the
+  // sizeof expression above fits into a single byte
+  static_assert(
```
prusnak, line 94: We use `_Static_assert` in the rest of the codebase. Let's use the same for consistency.

prusnak (changes requested):

matejcik:
merged as [768d577](https://github.com/trezor/trezor-firmware/pull/2270/commits/768d577d7823716a7becc93aca2cace387985c49) in #2270

PR #2233, opened 2022-04-27
On-device PIN input for model R

grdddj:
Creating a PIN-input dialogue on model R.

It is more of a proof-of-concept, it does not have all the features - and we probably do not yet have specs for it.

To make it work, a new event (`double-press`) is created and connected with a "middle" button on the screen.

Steps how to try it:
- `TREZOR_MODEL=R make build_unix` ... building the emulator
- `./emu.py -es` ... wiping the device and setting SLIP44 seed, so that we can set PIN
- `trezorctl set pin` ... initializing PIN-setting dialogue

On device:
- `TREZOR_MODEL=R make PYOPT=0 BITCOIN_ONLY=1 V=1 VERBOSE=1 build_firmware`
- `make upload`
- `trezorctl device load -s`
- `trezorctl set pin`

image

At the first screen, there is a possibility to `ACCEPT` the PIN, which will send the result to the client.
Choosing `NEXT` will increase the number by one, and on the second screen, the left button changes to `BACK`, which decreases the number - that is valid for all the digits lower than 9.

image

At the "last" screen with digit 9, the right button could be used to `CANCEL` the PIN-input process.

image

Pressing `OK` - both buttons (left and right arrow) - at any number will store this number into the PIN.

Apart from the `OK` button blinking, there is currently no visible indication that the digit was saved, which should be somehow improved.
Also, it would be nice to show the current length of the PIN and have the possibility to see it/edit it.


core/embed/rust/src/ui/model_tr/component/pin.rs
```diff
@@ -0,0 +1,149 @@
...
+    pin_buffer: String<50>,
+}
+
+// TODO: somehow show the length of the PIN
+// TODO: somehow allow for viewing the PIN
+// TODO: somehow allow for deleting the digits in PIN
+
+impl<T> PinPage<T>
+where
+    T: Paginate,
```
mmilata, line 30: No need to involve pagination or inner `T` component at all.

core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -13,12 +13,80 @@
     ExceptionType = BaseException | Type[BaseException]
 
 
+class DoublePressHandler:
```
mmilata, line 16: I'd prefer to handle the double press directly in the `Button` component unless there's reason not to.

mmilata (changes requested):

core/embed/rust/src/ui/model_tr/component/pin.rs
```diff
@@ -9,6 +9,8 @@ use crate::ui::{
 use super::{theme, Button, ButtonMsg, ButtonPos};
 use heapless::String;
 
+const MAX_LENGTH: usize = 9;
```
prusnak, line 12: Why not 50?

grdddj: Just copied it from `TT`'s PIN dialogue, which also has currently 9 - 
https://github.com/trezor/trezor-firmware/blob/18c830433922690cc9b793b1ee63675eb5e4a63d/core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs#L25

mmilata: #2227 increases this to 50

grdddj: Also increased it to 50 in [71740a4](https://github.com/trezor/trezor-firmware/pull/2233/commits/71740a44445495b4d36aca38074c199ce6fcffaf)

core/embed/rust/src/ui/model_tr/component/pin.rs
grdddj: In [71740a4](https://github.com/trezor/trezor-firmware/pull/2233/commits/71740a44445495b4d36aca38074c199ce6fcffaf) I got completely rid of the pagination, currently, I am showing the digit on screen using `display::text`

core/embed/rust/src/ui/geometry.rs
```diff
@@ -297,6 +297,14 @@ impl Rect {
         self.split_left(self.width() - width)
     }
 
+    /// Is meant to accommodate a very small button, for example "OK"
+    pub fn split_center(self, width: i32) -> Self {
+        // 42 % offset from right found out experimentally for "OK" button
+        self.split_right(width)
+            .1
+            .translate(Offset::new(-(self.width() as f64 * 0.42) as _, 0))
```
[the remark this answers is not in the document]
grdddj: I was not able to use it correctly to shift the button to the left. How would it be done here?

core/src/trezor/ui/layouts/tr/__init__.py
grdddj: Done in [9d58d42](https://github.com/trezor/trezor-firmware/pull/2233/commits/9d58d42848c1f531f88c2751b3cf4a42f376ad4c)

One "issue" is still that the first click is registered as a "normal click" and the appropriate button changes its visual state to `Pressed`. I am changing the visual state back to `Released` as soon as we register a double-click, but the quick visual state change can be disturbing for the user.

Can it be somehow resolved without bringing other disadvantages and complexity? Or is the current state OK?

matejcik:
one quick note: "double press" or "double click" usually means clicking the same button twice in a short time.

let's call two-button click something else. "Left" "Right" "Both" maybe?

grdddj:
> one quick note: "double press" or "double click" usually means clicking the same button twice in a short time.
> 
> let's call two-button click something else. "Left" "Right" "Both" maybe?

Good point, in [c8cd418](https://github.com/trezor/trezor-firmware/pull/2233/commits/c8cd418bc4a8fb49219fc37cebf47bc9b6f6e4d4) I am changing all the _double-press_ and _double-clicks_ to _both-button-press_

grdddj:
@sime as we agreed, I have done something comparable to Figma layouts - showing the previous and next option. Also with possible shuffling of digits (which needs to be asked for by the client). Feel free to build it and try

@TychoVrahe already verified the PIN entry works on a physical device :tada:

grdddj:
@sime I have pushed a new temporary branch specifically for the `Figma` design, as it is currently not using some here-introduced things like `hold-to-confirm` or `show-me-pin`.

After we agree everything is OK there, we will merge it back into this PR.

It is under [grdddj/r_pin_entry_figma_design](https://github.com/trezor/trezor-firmware/tree/grdddj/r_pin_entry_figma_design)

matejcik:
force-push to rebase on current version of the UI tests branch

core/embed/rust/src/ui/geometry.rs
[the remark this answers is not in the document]
mmilata: Something along the lines of.
```
self
    .inset(Insets::left((self.width() - width)/2))
    .inset(Insets::right((self.width() - width)/2))`
```

matejcik:
will be merged with #2277

PR #2234, opened 2022-04-27
build(core,legacy): remove gcc, use cc instead of gcc

prusnak:
... for firmware builds we still use arm-none-eabi-gcc via $(PREFIX)gcc

We should probably try this in the CI by rebuilding the environment image and running the pipeline.


mmilata (approved):
Both normal build and build-docker.sh works on linux. I think my system is quite similar to CI so it should work fine there as well.

matejcik (approved):
utACK

matejcik:
running environment now. if it breaks all builds, we will know :)

PR #2235, opened 2022-04-27
List::from_iter

matejcik:
create a MicroPython list from an iterator

### DIFF
--- a/core/embed/rust/src/micropython/list.rs
+++ b/core/embed/rust/src/micropython/list.rs
@@ -1,4 +1,4 @@
-use core::convert::TryFrom;
+use core::{convert::TryFrom, ptr};
 
 use crate::error::Error;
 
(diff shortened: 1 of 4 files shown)


grdddj (approved):
Looks good to me.

I have already cherry-picked those changes to https://github.com/trezor/trezor-firmware/pull/2214 to see if those two places using `List::from_iter` behave the same (`(*list).len = 0;` is the only change there)

matejcik:
> see if those two places using `List::from_iter` behave the same

perfect.

I sure hope they _don't_ behave the same. The previous code was broken and I still don't quite understand why it worked at all in one of the attempts.

matejcik:
> why it worked at all in one of the attempts.

could it be because in that case the iterator didn't know its max size? in which case it tries to preallocate 0 elements, and then it works as it is supposed to.

core/embed/rust/src/micropython/testutil.rs
```diff
@@ -0,0 +1,33 @@
...
+static mut MPY_INITIALIZED: bool = false;
+
+/// Initialize the MicroPython environment.
+///
+/// This is very hacky, in no way safe, and should not be used in production.
+/// The stack is configured on a best-effort basis and depending on from where
+/// this is called, you might get errorneous "recursion exceeded" problems.
+///
+/// This should only be called at start of your test function.
+#[cfg(test)]
```
mmilata, line 18: Is this needed when it's in mod.rs?

core/embed/rust/src/micropython/list.rs
```diff
@@ -17,6 +17,34 @@ impl List {
...
 
+    pub fn with_capacity(capacity: usize) -> Result<Gc<Self>, Error> {
+        // EXCEPTION: Will raise if allocation fails.
+        catch_exception(|| unsafe {
+            let list: *mut Self = ffi::mp_obj_new_list(capacity, ptr::null_mut())
+                .as_ptr()
+                .cast();
+            // By default, the new list will have its len set to n. We want to preallocate
+            // to a specific size and then use append() to add items, so we reset len to 0.
+            (*list).len = 0;
```
mmilata, line 28: `ffi::mp_obj_list_set_len` might be cleaner?

mmilata (approved):
LGTM, nice test

jpochyla:
Might be good to move this into a [FromIterator](https://doc.rust-lang.org/std/iter/trait.FromIterator.html) impls.

core/embed/rust/src/micropython/list.rs
matejcik: good find! i didn't notice that exists.

great comment too:
```
    // trust that the caller knows what it's doing
```

core/embed/rust/src/micropython/testutil.rs
matejcik: not _needed_ per se, I'm just being extra careful

matejcik:
> Might be good to move this into a [FromIterator](https://doc.rust-lang.org/std/iter/trait.FromIterator.html) impl.

that can't fail though

core/embed/rust/src/micropython/list.rs
matejcik: resolved in [d12f8c2](https://github.com/trezor/trezor-firmware/pull/2235/commits/d12f8c217e34c8875f9d15929055227be79bd7ae)

PR #2237, opened 2022-04-28
fix(core): re-enable passing of TREZOR_EMULATOR_RASPI flag

prusnak:
to re-enable build of Emulator for Raspberry Pi with a display

related to commit daf0d77 which removed the functionality from the makefile but we missed that there is now no way how to pass the flag to emit the `#define`


matejcik (approved):
utACK, this does not look like it breaks anything.

PR #2238, opened 2022-04-28
Render homescreen avatar for R and 1

grdddj:
Creating a "full" homescreen on both `R` and `1` emulators:
- rendering the label at the bottom, the avatar in the middle and the optional header at the top

Currently the `display.icon` function is used instead of `display.avatar`, and the picture is hardcoded there, with no possibility of change. Certainly room for improvement there.

image

image

image

image


core/src/trezor/res/resources.py
```diff
@@ -27,6 +28,10 @@ def load_resource(name: str) -> bytes:
         return b'TOIg\x18\x00\x18\x00\x97\x00\x00\x00E\xcf\xb1\r\xc20\x10\x85a\xdb(\x82\x08!\xb9\xa5\xa3\xa2\x85\r`\x03\xd8\x00\xc4\x04\x8c\xc0\x06\x88\x92\x8al\xc0\nl\x106\xc8\x08i@ @\xfe\xb1\x9dK\xfc\x1a\x7f\xb6N\xe7;\xa5B\x8cJ9\'\xe6\xef\xe4Kr\x9fo\xe7\x15L\x84\xba\x82B<\x00\x9e\xe2\x99w[t\x0c\xbe5\xae\x82_\x91\xbd@\\t\x16\xcd\xb2\xf9\tW\xc3^Z>\x16p\x90\xf7"\x17\xfb\xfa\xb9\x91\x1a\x83\xb3\xaaf\x1d\x1b\x95X]\xfb{\xc8\x86m\x86l0\xe4>jg\xd0\xe5\xe7\xfa\xb32\xe8\x18N\xddb\xd3]<\xfe'
     if name == "trezor/res/header_icons/wrong.toif":
         return b"TOIg\x18\x00\x18\x00\xaa\x00\x00\x00c`\xc0\x05\xaa\x1c@$\xcbr !v~3\x88\xad\xf56\x91\x81\xa1\xf7\xff/\x05\x06\x06\xa6\xf5\xffo00\xec\xff\xff\x7f\x11P\xf8\xff\xff\xdf\x0c\x0c\xd1@R\x01(\xfc\x7f+\x03\x033Pb\x02'P\xc0\x80\x01,\xfdr\xde\xff\xff`\x13\x98\x80\x12Pa\x06\x06\xab\xf7\xff\xff\xff[\x0c\xb1\x8cq\xfe\xff\xff\xbf\x04\xa06\x03\xd9?\xa1LM\x90\x9aI\x10%\xebAzA\x96\x83\xcd\xfc\xb5\x0el9\xc8%\xff\x17JA$\xbc@f\x83,_\xc2\xc0\xb0\x1el\xa56\xd8m\xbd`+\x81\x96\xdf\x04\xfa\xe5>\xd8%\xd6\x7f\x0b\x81Fv\x07\x80\xd8\xac;\x04p\x86\x02\x00"
+    if name == "trezor/res/homescreen_model_1.toif":
```
matejcik, line 31: Note: this is not great, it means we are bundling multiple images that won't be needed across models.

Not sure what to do about it. Maybe export the default homescreens to the `ui.components` module and exclude them here?
(But same problem exists in the opposite direction with FIDO icons.)

matejcik: #2273

matejcik (approved):

matejcik:
@grdddj looking good overall, please merge if it is in fact ready to merge

PR #2239, opened 2022-04-28
Implement firmware hashing

andrewkozlik:
Implements a new `GetFirmwareHash` message which hashes the firmware that is loaded on Trezor together with an optional challenge. The feature uses Blake2s with the challenge as the key for Blake2s. The way `blake2s_InitKey()` works is it hashes the key into the first block. (The key length is included in the Blake2s parameters block.)

On TT hardware the hashing operation takes 0.56 seconds for Blake2s. It would take 1.02s for SHA-256. The whole round-trip time using trezorctl and TT is 1.3 seconds for Blake2s. On T1 it's even faster. Although T1 has a slower processor, it also has smaller firmware.

**Possible TODOs**
- I didn't add any device tests, because they wouldn't work on hardware. Are we OK with that? I don't think we have anything like `@skip_hw`.
- <s>I should at least add a unit test for the `utils.firmware_hash()` in TT.</s>
- <s>Maybe add a simple progress dialog, like "please wait, computing" for the 1 second?</s>
- One more thing that would be nice to do (in another PR) is to fill the built firmware to its maximum size with randomly generated data. That can be a problem for reproducible builds. One idea that comes to mind is that we could have a file in our repository called `random_padding.bin` which would be used as the source of the random data. It's OK to reuse this data, it just has to be random to begin with, i.e. not deterministically generated.

Blake2s documentation: https://www.blake2.net/blake2.pdf


prusnak (approved):
Code review ACK

Did not test on real devices.

andrewkozlik:
> Did not test on real devices.

I did, both T1 and TT. But we should carefully retest with signed firmware too, in case there is some memory protection on official firmware that I am not aware of.

matejcik:
> * One more thing that would be nice to do (in another PR) is to fill the built firmware to its maximum size with randomly generated data. That can be a problem for reproducible builds. One idea that comes to mind is that we could have a file in our repository called `random_padding.bin` which would be used as the source of the random data. It's OK to reuse this data, it just has to be random to begin with, i.e. not deterministically generated.

❤️

core/embed/extmod/modtrezorutils/modtrezorutils.c
```diff
@@ -117,6 +119,49 @@ STATIC mp_obj_t mod_trezorutils_halt(size_t n_args, const mp_obj_t *args) {
...
+  if (chal.len != 0) {
+    if (blake2s_InitKey(&ctx, BLAKE2S_DIGEST_LENGTH, chal.buf, chal.len) != 0) {
+      mp_raise_msg(&mp_type_ValueError, "Invalid challenge.");
+    }
+  } else {
+    blake2s_Init(&ctx, BLAKE2S_DIGEST_LENGTH);
+  }
+
+  for (int i = 0; i < FIRMWARE_SECTORS_COUNT; i++) {
+    uint8_t sector = FIRMWARE_SECTORS[i];
```
matejcik, line 144: It seems this counts the unused spaces too? Is that a problem?

Am I correct that this relies on user storage being located before firmware? It would be nice to assert that somehow.

tools/firmware_hash.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
+
+from hashlib import blake2s
+from os import sys
```
grdddj, line 4: Just `import sys` is the common way, even though this also works, to my surprise

tools/firmware_hash.py
```diff
@@ -0,0 +1,29 @@
...
+
+for filename, size in ((FILE_T1, SIZE_T1), (FILE_T2, SIZE_T2)):
+    try:
+        data = open(filename, "rb").read()
+    except:
+        print(f"{filename} not found")
+        continue
+
+    if len(data) > size:
+        raise ValueError(fn, f"{filename} too big")
```
grdddj, line 20: `fn` is not defined here

tools/firmware_hash.py
```diff
@@ -0,0 +1,29 @@
...
+FILE_T1 = "legacy/firmware/trezor.bin"
+FILE_T2 = "core/build/firmware/firmware.bin"
+
+SIZE_T1 = (7 * 128 + 64) * 1024
+SIZE_T2 = 13 * 128 * 1024
+
+for filename, size in ((FILE_T1, SIZE_T1), (FILE_T2, SIZE_T2)):
+    try:
+        data = open(filename, "rb").read()
+    except:
```
grdddj, line 15: `except FileNotFoundError` would be more descriptive

grdddj (commented):
Just some small comments on `tools/firmware_hash.py`

core/embed/extmod/modtrezorutils/modtrezorutils.c
andrewkozlik: > It seems this counts the unused spaces too?

If you mean the unused space at the end of the firmware, then yes. The client that is checking the firmware needs to take that into account and pad the hashed data with `0xff` bytes, see `tools/firmware_hash.py`. The alternative would be to parse the vendor header and the firmware header to obtain the total size of the firmware. That's not too much trouble I think, but it might prove to be a pointless exercise if we go ahead with padding the firmware with random data.

andrewkozlik: > Am I correct that this relies on user storage being located before firmware?

I am not sure, you mean because of the MPU, so that it's readable? Shouldn't it always be readable?

tools/firmware_hash.py
andrewkozlik: [d429101](https://github.com/trezor/trezor-firmware/pull/2239/commits/d429101cb44669cedb377199a7e5939391e58594)

tools/firmware_hash.py
andrewkozlik: [d429101](https://github.com/trezor/trezor-firmware/pull/2239/commits/d429101cb44669cedb377199a7e5939391e58594)

andrewkozlik:
Added a unit test for the `utils.firmware_hash()` in TT in https://github.com/trezor/trezor-firmware/pull/2239/commits/fdfd3bc929701c855109d656a32d6aa75cf133aa, which is why I had to do some refactoring around flash initialization in https://github.com/trezor/trezor-firmware/pull/2239/commits/ace23afdda622e27522bd3190479bf9c7de8339e.

core/embed/extmod/modtrezorutils/modtrezorutils.c
matejcik: No, I misunderstood how `flash_sector_size()` calculates the size and thought that any holes in the sector space would also be counted. So the whole comment is moot.

That opens a different angle however: we actually **should** be counting the unused regions (sectors 12-16 on TT), otherwise that's 128 kB of completely hidden space.

Speaking of, the user storage is also 128 kB total hidden space. No idea what to do about that.

andrewkozlik: > That opens a different angle however: we actually should be counting the unused regions (sectors 12-16 on TT), otherwise that's 128 kB of completely hidden space.

Yes, ideally we should. But in order for it to make sense we would have to fill it with (known) random data. I'd also put these regions after the firmware in the hashing process, so that it's easier to compute on the client.

> Speaking of, the user storage is also 128 kB total hidden space. No idea what to do about that.

Me neither at the moment.

core/embed/extmod/modtrezorutils/modtrezorutils.c
```diff
@@ -117,6 +119,49 @@ STATIC mp_obj_t mod_trezorutils_halt(size_t n_args, const mp_obj_t *args) {
...
+  vstr_t vstr = {0};
+  vstr_init_len(&vstr, BLAKE2S_DIGEST_LENGTH);
+  if (blake2s_Final(&ctx, vstr.buf, vstr.len) != 0) {
+    vstr_clear(&vstr);
+    mp_raise_msg(&mp_type_RuntimeError, "Failed to finalize firmware hash.");
+  }
+
+  return mp_obj_new_str_from_vstr(&mp_type_bytes, &vstr);
+}
+STATIC MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mod_trezorutils_firmware_hash_obj, 0,
```
matejcik, line 162: this should actually be a one-argument (not variadic) function, checking for None

andrewkozlik: But then calling `utils.firmware_hash()` will result in `TypeError: function takes 1 positional arguments but 0 were given`. I meant the definition to be `/// def firmware_hash(challenge: bytes | None = None) -> bytes:`.

hiviah:
MPU makes firmware on TT OK to read, with T1 and usigned firmware you actually were in unprivileged mode with more MPU restrictions that official FW has, so that should be fine as well.

matejcik:
I implemented firmware dumping and on a TT it takes 30 seconds to run. Not horrible but not something to do on every startup either. Probably not worth adding at this point.

andrewkozlik:
1. I added a progress dialog in both Trezor models.
2. In Trezor T firmware the `Features` message will return the `fw_vendor` field same as in bootloader. Possible values are: "SatoshiLabs", "UNSAFE, DO NOT USE!", "EMULATOR".

prusnak:
> 1. I added a progress dialog in both Trezor models.

Do we really want this? Since it takes less than a second and this will be done on each start, it might be seen as very flashy / distracting / etc.

andrewkozlik:
> it might be seen as very flashy / distracting / etc.

Visually it takes approximately the same time on TT as seed derivation, so I wonder why one should have the progress dialog and the other shouldn't, but I had the same thought about flashiness. I remember my Trezor was quite flashy in the past at least with the old online wallet, but now Suite seems to have optimized the process.

It would be nice if we could somehow join the workflows together in the sense that the homescreen would not appear in between them. This can either be indicated in the messages with a new optional field `delay_homescreen` or it we can do it for certain/all messages automatically. In these cases Trezor would wait approximately 1 second after the workflow is finished before displaying the homescreen. If another message came in before the time expires, it would go directly to that message.

matejcik:
> It would be nice if we could somehow join the workflows together in the sense that the homescreen would not appear in between them.

cc @sime for UX design.
To me it seems that it would be nicer to not show anything, unless we know the user is actively looking at the device and waiting for something.

python/src/trezorlib/cli/firmware.py
```diff
@@ -590,3 +590,12 @@ def update(
         click.echo("Dry run. Not uploading firmware to device.")
     else:
         upload_firmware_into_device(client=client, firmware_data=firmware_data)
+
+
+@cli.command()
+@click.argument("hex_challenge", required=False)
+@with_client
+def get_hash(client: "TrezorClient", hex_challenge: Optional[str]) -> str:
```
matejcik, line 598: going forward, we should extend this command with the ability to compare to an existing fw image

core/embed/extmod/modtrezorutils/modtrezorutils.c
matejcik: ISTM nobody is calling it as `utils.firmware_hash()` anyway so it makes things a bit simpler to just require the argument.

matejcik (approved):

sime:
> > It would be nice if we could somehow join the workflows together in the sense that the homescreen would not appear in between them.
> 
> cc @sime for UX design. To me it seems that it would be nicer to not show anything, unless we know the user is actively looking at the device and waiting for something.

By uniting the workflow, I assume you mean the progress animation of 'Starting up'. But in favour of skipping an animation if it results in flickering.

mmilata (approved):
utACK

storage/norcow.c
```diff
@@ -299,7 +299,6 @@ static void compact(void) {
  * Initializes storage
  */
 void norcow_init(uint32_t *norcow_version) {
-  flash_init();
```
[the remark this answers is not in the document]
andrewkozlik: In core it's now called in [core/embed/unix/main.c](https://github.com/trezor/trezor-firmware/pull/2239/files#diff-36a131d6a2f4a443e9403737664175de26875424e952fe16dfeef01576466370) and in legacy it stayed the same as before, being called in legacy/emulator/setup.c.

Issue #2130, opened 2022-02-17
Do not initialize touch controller in firmware


matejcik:
Touch controller comes pre-initialized from bootloader. I believe we don't need to re-initialize in firmware's `main.c` (doing so incurs ~500 ms delay waiting for the controller to power down and boot back up).

Is there any reason not doing the initialization would be a problem? @hiviah @prusnak 

The only thing comes to mind is, bootloader might be using different configuration. Even in that case, it should be sufficient to (a) check that the controller is up and (b) send the configuration commands.

prusnak:
> Is there any reason not doing the initialization would be a problem?

I can't think of any.

matejcik:
QA instructions:
Just install this firmware and try to interact with it, unplug and replug a couple times. See if there are any touch issues (different than with regular firmware :) )

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 22.6.2 (f45c75705f153ec0dc44cb72fec2e5ec4a031b7c)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.6.2 Chrome/94.0.4606.81 Electron/15.3.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1680x1050
 - Device: model T 2.5.2 regular (revision c3b82fd651facd11ffdb481da4c64ff864b6ef7a)
 - Transport: bridge 2.0.31

matejcik:
we might want to set up configuration of the touch controller, ideally without restarting it.

this will prevent problems in case future bootloaders don't set the same parameters that we want.

Issue #1571, opened 2021-04-15
Jump from firmware to bootloader initiated by Host on TT


tsusanka:
For TTs with new boardloader (~July 2021) we can provide the "Go to bootloader" feature such as we did for T1 in #1047.

We need to:
- Support the new "go to bootloader" protobuf message, see #1461.
- Check whether the new boardloader is present.
- If no, abort and inform the host.
- If yes, show a confirmation dialog.
- Restart into bootloader.

tsusanka:
We should do this along with #1049.

tsusanka:
We have agreed with @prusnak, he will do the needed modification in bootloader. After that's done we can discuss who will do the rest.

mcudev:
I was just looking at this and saw that the soft reset check was removed from the boardloader by https://github.com/trezor/trezor-firmware/commit/3294a49cec25901f57d825ef0a7ba02bd7c3169e#diff-732ec3c652ad70179ea1d45a0736327ef215644155d9a4443e76c3efef1a8e4d. That was the main hurdle and I was hesitant to remove that from the boardloader, but since that bridge was crossed a while ago, maybe just carry through the fact that it was a soft reset (by not clearing the reset flags) and then look at those flags for the soft reset event in the bootloader as a way to know to stay in the bootloader that doesn't require user interaction. As far as I remember, that's all you really need since nothing else that I recall soft resets the system, and you just need a way to know to not leave the bootloader that doesn't require user interaction. My recollection from when I put that reset flag checking stuff in the boardloader was that I was trying to provide as consistent a system state as I could (and avoiding system states that were not strictly full power-on removed a lot of unknowns) and also make it harder to perform glitching attacks that target the post-boot ROM code.

prusnak:
> but since that bridge was crossed a while ago

The main challenge is how to implement this feature so it works with the older boardloader (that still halts when `reset_flags_check` tells it to). The commit you are linking to has never been deployed yet.

mcudev:
ah, not being deployed yet changes things. i don't like the approach of changing the reset flag check in the boardloader.
iirc, the reason that this feature is not implemented naturally by jumping from the firmware back to the bootloader is that the firmware runs with reduced privileges. maybe another way to backdoor into a solution is to ponder, "how do we get privileged execution back to the bootloader and in a way that makes sure it's a running an authentic prod bootloader?". it might be a dead end, but i'll think on it.

mcudev:
[...]

PR #2240, opened 2022-04-28
Do not initialize touch controller in firmware

marnova:
Resolves issue #2130


matejcik:
Blocked on #1571 where we need to review whether skipping the initialization breaks something.

hiviah:
So far this experimentally works with return-to-bootloader, but we'd still need more checks when the whole thing is complete.

hiviah (approved):
Tested with works stay-in-bootloader.

PR #2241, opened 2022-04-28
fix(core): leftover TREZOR_MODEL defines

mmilata:
Fixes [storage test](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2386665118). Related: #2222.


matejcik (approved):
utACK

PR #2243, opened 2022-05-02
Model R embedded implementation

TychoVrahe:
Implemented basic Model R functionality:

- display is functional
- buttons are working
- USB comm is working, possible to connect to suite and install (custom) firmware
- shrank loader so it fits the screen, but its not very pretty, probably round loader is not the way
- backlight implemented using display contrast settings, does not work to zero though.
- set colors for monochromatic display in bootloader
- UI needs work, current icons do not fit the display, etc
- using framebuffer in order to support current display functionality used in bootloader, firmware

(firmware is not implemented for R, works partially when fed model type as 'T' via QSTR, but still for example crashes when trying to create wallet, not sure why, maybe missing touch screen causes problems)
 
Built on top of #2230


TychoVrahe:
Fixed style issues & force pushed

matejcik:
please rebase on current master, merging the base branch did something bad to this PR

core/embed/trezorhal/button.c
```diff
@@ -1,25 +1,45 @@
...
+#define BTN_RIGHT_PORT GPIOE
+#define BTN_RIGHT_CLK_ENA __HAL_RCC_GPIOE_CLK_ENABLE
+#else
+#error Unknown Trezor model
+#endif
+
+static char last_left = 0, last_right = 0;
 
 void button_init(void) {
-  __HAL_RCC_GPIOC_CLK_ENABLE();
```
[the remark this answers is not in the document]
matejcik: right! totally didn't notice that the other macro calls it

TychoVrahe:
rebased onto master & force pushed

core/embed/bootloader/main.c
```diff
@@ -253,7 +264,14 @@ int main(void) {
   // delay to detect touch
   uint32_t touched = 0;
   for (int i = 0; i < 100; i++) {
+#if defined TREZOR_MODEL_T
     touched = touch_is_detected() | touch_read();
+#elif defined TREZOR_MODEL_R
+    button_read();
+    if (button_state_left() == 1 && button_state_right() == 1) {
```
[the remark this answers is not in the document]
TychoVrahe: fixed 2b27fbce08ea3ff90be1a71ec7f902db887fddde

matejcik (approved):
LGTM but I did not review the display driver (`display-stm32_R.h`) in too much detail.

@hiviah could you please take a look on that part specifically?

core/embed/bootloader/main.c
```diff
@@ -29,7 +29,12 @@
 #include "mpu.h"
 #include "random_delays.h"
 #include "secbool.h"
+#ifdef TREZOR_MODEL_T
 #include "touch.h"
+#endif
+#if defined TREZOR_MODEL_R || defined TREZOR_MODEL_1
```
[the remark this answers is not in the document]
TychoVrahe: fixed 572cc5b7a81e0a46b8db143ba423b7caa8449536 and 2e8dd0fcdb26cf90ddcaffbd48665ac75a056366. Though we still have model 1 variant in bootloader SConscript. 

However, this breaks the compilation of bootloader for model 1 variant. Since bootloader is part of build_all and build_embed, this also breaks. Which makes me wanna drop the commits. Or i can change these particular conditionals to not throw "Unknown Trezor model" errors.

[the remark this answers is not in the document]
matejcik: > can change these particular conditionals to not throw "Unknown Trezor model" errors.

that seems like the best option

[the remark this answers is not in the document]
matejcik: but I'm wondering if we could somehow noop-ize bootloader build for T1. Like, `Sconscript.bootloader` would just print "not building for T1" instead of doing anything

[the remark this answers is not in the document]
TychoVrahe: This can be achieved with this placed at the beginning of sconscript, seems like a better choice then running dummy compilation:
`if TREZOR_MODEL in ('1', ):
    # skip bootloader build
    env = Environment()
    def build_bootloader(target,source,env):
        print(f'Bootloader: nothing to build for Model {TREZOR_MODEL}')
    program_bin = env.Command(
                 target='bootloader.bin',
                 source=None,
                 action=build_bootloader
             )
    Return()
`
We can include this also in boardloader and bootloader_ci to skip those too.

core/embed/extmod/modtrezorui/display.c
```diff
@@ -366,10 +371,10 @@ void display_icon(int x, int y, int w, int h, const void *data,
...
   clamp_coords(x, y, w, h, &x0, &y0, &x1, &y1);
   display_set_window(x0, y0, x1, y1);
-  x0 -= x;
-  x1 -= x;
-  y0 -= y;
-  y1 -= y;
+  x0 -= MAX(x, 0);
+  x1 -= MAX(x, 0);
+  y0 -= MAX(y, 0);
+  y1 -= MAX(y, 0);
```
prusnak, line 377: Why this change?

core/embed/extmod/modtrezorui/display.c
```diff
@@ -384,8 +389,8 @@ void display_icon(int x, int y, int w, int h, const void *data,
     int st = uzlib_uncompress(&decomp);
     if (st == TINF_DONE) break;  // all OK
     if (st < 0) break;           // error
-    const int px = pos % w;
-    const int py = pos / w;
+    const int px = (pos * 2) % w;
+    const int py = (pos * 2) / w;
```
prusnak, line 393: Why this change?

core/embed/extmod/modtrezorui/display.c
```diff
@@ -426,27 +431,35 @@ bool display_toif_info(const uint8_t *data, uint32_t len, uint16_t *out_w,
...
+#else
+#define LOADER_SIZE_ADJ 1
+#endif
 #endif
 
 void display_loader(uint16_t progress, bool indeterminate, int yoffset,
                     uint16_t fgcolor, uint16_t bgcolor, const uint8_t *icon,
                     uint32_t iconlen, uint16_t iconfgcolor) {
-#if defined TREZOR_MODEL_T
+#if defined TREZOR_MODEL_T || defined TREZOR_MODEL_R
```
prusnak, line 446: I'd rather keep the `display_loader` only defined for `TREZOR_MODEL_T` and keep it undefined for other models for now. Suggested changes are okayish, but the code is very hard to understand even before the change.

prusnak (commented):
Left couple of questions / suggestions ...

core/embed/extmod/modtrezorui/display.c
```diff
@@ -426,27 +431,35 @@ bool display_toif_info(const uint8_t *data, uint32_t len, uint16_t *out_w,
   return true;
 }
 
-#if defined TREZOR_MODEL_T
+#if defined TREZOR_MODEL_T || defined TREZOR_MODEL_R
 #include "loader.h"
+#ifdef TREZOR_MODEL_R
+#define LOADER_SIZE_ADJ 3
```
prusnak, line 437: Nit: rename to `LOADER_SIZE_SCALE`

core/embed/extmod/modtrezorui/display.c
```diff
@@ -426,27 +431,35 @@ bool display_toif_info(const uint8_t *data, uint32_t len, uint16_t *out_w,
...
+#endif
 #endif
 
 void display_loader(uint16_t progress, bool indeterminate, int yoffset,
                     uint16_t fgcolor, uint16_t bgcolor, const uint8_t *icon,
                     uint32_t iconlen, uint16_t iconfgcolor) {
-#if defined TREZOR_MODEL_T
+#if defined TREZOR_MODEL_T || defined TREZOR_MODEL_R
+
+  int tmp_img_loader_size = img_loader_size / LOADER_SIZE_ADJ;
```
prusnak, line 448: Nit: rename
```suggestion
  const int loader_size = img_loader_size / LOADER_SIZE_SCALE;
```

core/embed/extmod/modtrezorui/display.c
TychoVrahe: I did it to achieve starting position (0,0) when negative coordinates were passed for some reason, but thinking about it now, there might be legitimate use cases for this -> displaying only certain part of the icon, so i will drop this.

prusnak: > displaying only certain part of the icon, so i will drop this.

I had exactly this on my mind. The change disallows you to show just a part of the icon.

core/embed/trezorhal/button.c
```diff
@@ -1,25 +1,45 @@
...
-  GPIO_InitStructure.Pin = BTN_PIN_LEFT | BTN_PIN_RIGHT;
-  HAL_GPIO_Init(GPIOC, &GPIO_InitStructure);
+  GPIO_InitStructure.Pin = BTN_LEFT_PIN;
+  HAL_GPIO_Init(BTN_LEFT_PORT, &GPIO_InitStructure);
+  GPIO_InitStructure.Pin = BTN_RIGHT_PIN;
+  HAL_GPIO_Init(BTN_RIGHT_PORT, &GPIO_InitStructure);
 }
 
 uint32_t button_read(void) {
-  static char last_left = 0, last_right = 0;
```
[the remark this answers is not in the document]
TychoVrahe: These were declared static, so no reset on `button_read` call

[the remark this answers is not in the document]
prusnak: Ah, of course. My bad. Ignore this :-)

core/embed/extmod/modtrezorui/display.c
TychoVrahe: The uncompress function returns data for 2 pixels which are fed to the display - therefore, the position on the display actually moves twice faster than the for loop.

Still not perfect probably, might be better to check the position of the pixels individually for cases where display dimensions are odd:
```
    int px = (pos * 2) % w;
    int py = (pos * 2) / w;
    if (px >= x0 && px <= x1 && py >= y0 && py <= y1) {
      PIXELDATA(colortable[decomp_out >> 4]);
    }
    px = (pos * 2 + 1) % w;
    py = (pos * 2 + 1) / w;
    if (px >= x0 && px <= x1 && py >= y0 && py <= y1) {
      PIXELDATA(colortable[decomp_out & 0x0F]);
    }
```

prusnak: Okay, I now see where the issue was. I assume we just never hit the problem, because we never rendered icon on the edge of the display.

> Still not perfect probably

We can safely assume the display dimensions will always be even (I've never seen a display in my life that has odd width). We also always render at even coordinate thanks to check line 365: `x &= ~1;  // cannot draw at odd coordinate`, so this is not a problem either.

So let's keep the change as it is. 👍

prusnak:
Now thinking more about changes in `display_loader` ... I think the better approach is the following:

1. revert changes done to the function
2. rename `core/embed/extmod/modtrezorui/loader.h` to `core/embed/extmod/modtrezorui/loader_T.h`
3. set `outer = 20` and `inner = 14` in `core/tools/codegen/gen_loader.py` and generate `loader_R.h` using the changed constants
4. include either `loader_T.h`or `loader_R.h` depending on model
5. update `core/embed/extmod/modtrezorui/display.h` so it defines `LOADER_ICON_SIZE` as `24` for Model R (and keep 64 for Model T)

This has the following advantages:
* less changes to already complicated code in `display_loader`
* saves some flash space on Model R because the `img_loader` structure is 9-times smaller

core/embed/extmod/modtrezorui/display.c
TychoVrahe: The issue also applies to icons with odd dimensions, which is more likely to happen

prusnak: > The issue also applies to icons with odd dimensions, which is more likely to happen

I just checked and there are no icons with odd width in the repository. Trying to use `./core/tools/toif_convert.py` on a grayscale image with odd width throws the following error:

``` python
  File ".../trezor-firmware/python/src/trezorlib/toif.py", line 70, in _from_pil_grayscale
    left, right = pixels[i], pixels[i + 1]
IndexError: image index out of range
```

My intention was never to allow this anyway -> sent PR https://github.com/trezor/trezor-firmware/pull/2259 to check the width of provided file.

The odd heights are present in the repository, but this should not cause any issues if I understand everything correctly.

TychoVrahe: Agree that we are safe as long as width is even. Lets keep the check simple then.

core/embed/bootloader/main.c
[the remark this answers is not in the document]
TychoVrahe: fixed 2ba1b65bf249d6a80497222e9670406d8b3f5acc

core/embed/extmod/modtrezorui/display.c
TychoVrahe: fixed 5c3c68610ab8813c49a5126aabd4c63d26adbf04

TychoVrahe:
> Now thinking more about changes in `display_loader` ... I think the better approach is the following:

Implemented in 55b5a47f3d2ac1ba329f34fb951dfe81ca55fddf

The render looks maybe a bit better, but its not a big margin and still likely not good enough for production

prusnak (approved):
Looks great, thanks!

hiviah (approved):
LGTM, looked at the display driver more in detail, looks sane, though I couldn't test it on HW.

matejcik:
@TychoVrahe please squash, rebase, fix style checks etc.

matejcik:
please add a changelog entry, something like Added: basic Trezor Model R hardware support.
(see misc/changelog.md for instructions)

TychoVrahe:
> T1 core build is now broken, please fix

fixed in 718be2026538002c8883c754444735d66fd9a740

matejcik:
everything seems to pass now so please squash the last fixup and let's get this merged

matejcik:
🎉

PR #2248, opened 2022-05-03
Firmware dumping

matejcik:
builds on @andrewkozlik's FW hashing

FW dumping takes ~30 secs on TT and ~10 secs on T1. I added a confirmation dialog, wording is up for discussion.


core/src/apps/misc/get_firmware.py
```diff
@@ -0,0 +1,46 @@
...
+from trezor.messages import FirmwareChunk, FirmwareChunkAck, GetFirmware, Success
+from trezor.ui.layouts import confirm_action, draw_simple_text
+
+from .get_firmware_hash import _render_progress
+
+if TYPE_CHECKING:
+    from trezor.wire import Context
+
+CHUNK_SIZE = const(1024 * 4)
+PROGRESS_TOTAL = const(13 * 128 * 1024)
```
andrewkozlik, line 14: preferrably 13 -> `utils.FIRMWARE_SECTORS_COUNT`

legacy/firmware/fsm_msg_common.h
```diff
@@ -554,3 +554,71 @@ void fsm_msgGetFirmwareHash(const GetFirmwareHash *msg) {
...
+  layoutProgress(_("Extracting..."),
+                 1000 * total_bytes_sent / FIRMWARE_PROGRESS_TOTAL);
+  msg_write(MessageType_MessageType_FirmwareChunk, resp);
+  return;
+
+end:
+  in_dump_firmware = false;
+  sector = 0;
+  sector_offset = 0;
+  total_bytes_sent = 0;
```
andrewkozlik, line 595: Nothing against `goto`, but it would be better to put this into `firmware_dump_abort()` and then also call that function in `fsm_abortWorkflows()`, see https://github.com/satoshilabs/trezor-firmware/commit/353ff85fd2.

tests/device_tests/test_firmware_dump.py
```diff
@@ -0,0 +1,27 @@
...
+}
+
+
+def test_firmware_dump_hash(client: Client) -> None:
+    data = firmware.get_firmware(client)
+    assert len(data) == FIRMWARE_LENGTHS[client.features.model]
+
+    if client.features.fw_vendor != "EMULATOR":
+        # check that the dumped data is not empty
+        assert not all(byte == "\xff" for byte in data)
```
andrewkozlik, line 18: We should try to check this against the actual firmware `.bin` file. Can be done later after this PR is merged.

andrewkozlik (approved):
Awesome! utACK

tests/device_tests/test_firmware_dump.py
matejcik: Exactly, also we should fix the core emulator to not send all empty data. T1 emulator doesn't seem to do that.

core/src/apps/misc/get_firmware.py
matejcik: done in [f1f4abe](https://github.com/trezor/trezor-firmware/pull/2248/commits/f1f4abe9811e527e8d179106bdaf2dac4f6cc275)

it might break `const()` but that is probably not important

legacy/firmware/fsm_msg_common.h
matejcik: created `firmware_dump_abort()` in [cdecdb8](https://github.com/trezor/trezor-firmware/pull/2248/commits/cdecdb8b4f4ed5468982ed5b7e41bc5b11b0dbf2)

abortWorkflows will need to be done separately, and as you say, it is unlikely to be a problem.

tests/device_tests/test_firmware_dump.py
matejcik: ...except it does, check fixed in [badb0e7](https://github.com/trezor/trezor-firmware/pull/2248/commits/badb0e71d64d23329b646862db8f872ef3aade92) and reporting added in [ea3db2a](https://github.com/trezor/trezor-firmware/pull/2248/commits/ea3db2adbfe1a9d7ab15bb6a73e6a08ca3b54b77)

mmilata (approved):
utACK

PR #2262, opened 2022-05-10
chore(core/rust): use clippy for all features

grdddj:
Makes `clippy` analyze all the Rust code by `--all-features` flag, as previously it was not checking some parts of code.

Also "fixing" some things that it was currently complaining about.


core/embed/rust/src/ui/display.rs
```diff
@@ -233,7 +237,7 @@ impl Color {
         self.0
     }
 
-    pub fn neg(self) -> Self {
+    pub fn negate(self) -> Self {
```
mmilata, line 240: Huh, what was the lint message?

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -372,12 +384,12 @@ extern "C" fn ui_layout_touch_event(n_args: usize, args: *const Obj) -> Obj {
     unsafe { util::try_with_args_and_kwargs(n_args, args, &Map::EMPTY, block) }
 }
 
-#[cfg(any(feature = "model_t1", feature = "model_tr"))]
+#[cfg(all(any(feature = "model_t1", feature = "model_tr", not(feature = "model_tt"))))]
```
mmilata, line 387: If I'm reading this correctly then `all` only has one argument and therefore does nothing?

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -372,12 +384,12 @@ extern "C" fn ui_layout_touch_event(n_args: usize, args: *const Obj) -> Obj {
...
 extern "C" fn ui_layout_touch_event(_n_args: usize, _args: *const Obj) -> Obj {
     Obj::const_none()
 }
 
-#[cfg(any(feature = "model_t1", feature = "model_tr"))]
+#[cfg(all(any(feature = "model_t1", feature = "model_tr", not(feature = "model_tt"))))]
```
mmilata, line 392: Same.

matejcik, line 392: similar to above:
cfg(any(feature = ...))
then
cfg(not(any(...)))

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -354,7 +362,11 @@ extern "C" fn ui_layout_set_timer_fn(this: Obj, timer_fn: Obj) -> Obj {
     unsafe { util::try_or_raise(block) }
 }
 
-#[cfg(feature = "model_tt")]
+#[cfg(all(
```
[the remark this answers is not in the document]
matejcik: in fact we should introduce `buttons` and `touch` features, then define `model_tt = ["touch"]` and `model_tr = ["buttons"]`

matejcik:
AFAICT the main point of contention is the constants section.

I dislike the way conditional compilation based on features selects exactly one, regardless of which features are enabled. That produces a strictly incorrect result that "arbitrarily" picks the model to build.

we should probably create `ui/constant.rs` that re-exports the selected constants, and put the complicated logic there.
another neat-ish alternative is using an envvar in `build.rs` to select the model (and generate `constant.rs` for us):
https://stackoverflow.com/a/37528134/222189
then we can put `include!(concat!(env!("OUT_DIR"), "/constant.rs"));` in `constant.rs` which only says `pub use model_t<?>::*;`

core/embed/rust/src/ui/layout/obj.rs
grdddj: Yeah, I had a mismatch in `()`.

Replaced with `#[cfg(not(feature = "model_tt"))]` and `#[cfg(feature = "model_tt")]` as suggested

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -22,14 +22,22 @@ use crate::{
 
 #[cfg(any(feature = "model_t1", feature = "model_tr"))]
 use crate::ui::event::ButtonEvent;
-#[cfg(feature = "model_tt")]
+#[cfg(all(
```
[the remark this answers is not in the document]
grdddj: Right, changed to  `#[cfg(not(feature = "model_tt"))]` and `#[cfg(feature = "model_tt")]` as suggested elsewhere

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -20,13 +20,9 @@ use crate::{
     util,
 };
 
-#[cfg(any(feature = "model_t1", feature = "model_tr"))]
+#[cfg(not(feature = "model_tt"))]
```
matejcik, line 23: this was correct before

Buttons are enabled if we're building for `model_t1` or `model_tr`.

Touch is enabled if we're building for `model_tt`.

If we are building _both_ for `model_t1` and `model_tt`, both touch and buttons is enabled. There is no mutual exclusion. You can have buttons and touch in the same build.

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -384,12 +376,12 @@ extern "C" fn ui_layout_touch_event(n_args: usize, args: *const Obj) -> Obj {
...
 }
 
-#[cfg(all(any(feature = "model_t1", feature = "model_tr", not(feature = "model_tt"))))]
+#[cfg(not(feature = "model_tt"))]
 extern "C" fn ui_layout_touch_event(_n_args: usize, _args: *const Obj) -> Obj {
     Obj::const_none()
 }
 
-#[cfg(all(any(feature = "model_t1", feature = "model_tr", not(feature = "model_tt"))))]
+#[cfg(not(feature = "model_tt"))]
```
matejcik, line 384: dtto this should be _enabled for_ t1/tr

instead of _disabled for_ tt

grdddj:
[3e2200c](https://github.com/trezor/trezor-firmware/pull/2262/commits/3e2200c60a36b5691808f8d53fa8b869db1ff877) creates the new `constants.rs` file and `touch` and `buttons` features. `Clippy` seems to be happy now

core/embed/rust/src/ui/constants.rs
```diff
@@ -0,0 +1,13 @@
+//! Reexporting the `constant` module according to the
```
matejcik, line 1: please call this `constant` so that it's the same as model_*.constant

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -376,12 +366,12 @@ extern "C" fn ui_layout_touch_event(n_args: usize, args: *const Obj) -> Obj {
     unsafe { util::try_with_args_and_kwargs(n_args, args, &Map::EMPTY, block) }
 }
 
-#[cfg(not(feature = "model_tt"))]
+#[cfg(all(feature = "buttons", not(feature = "touch")))]
```
matejcik, line 369: why is `buttons` here?

grdddj: Right, just simple `#[cfg(feature = "touch")]` and `#[cfg(not(feature = "touch"))]` works - the same below with "buttons"

Done in [d877e72](https://github.com/trezor/trezor-firmware/pull/2262/commits/d877e725bcb5b071562e0cfd469091a49affe7b0)

core/embed/rust/src/ui/constants.rs
grdddj: Done in [902c3b2](https://github.com/trezor/trezor-firmware/pull/2262/commits/902c3b207836315adfb0e5aad4ffcd987c9e7cfd)

matejcik:
looking good, please squash & rebase

matejcik (approved):

PR #2263, opened 2022-05-11
Strip old T1 header in firmware_hash.py

andrewkozlik:
The official T1 firmware has an extra 256 bytes at the beginning of the file, starting with "TRZR". This needs to be stripped when we compute the hash in the firmware_hash.py tool to get the right result. I am not exactly sure what this header is called, "old T1 header"? Also not entirely sure whether it's always 256 bytes. @prusnak 

I also made some improvements to the usability of the tool.

### DIFF
--- a/tools/firmware_hash.py
+++ b/tools/firmware_hash.py
@@ -6,24 +6,42 @@ from hashlib import blake2s
 FILE_T1 = "legacy/firmware/trezor.bin"
 FILE_T2 = "core/build/firmware/firmware.bin"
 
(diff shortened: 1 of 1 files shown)


matejcik:
It's always 256 bytes, that's the old header format. It is also something that is stripped by trezorctl when uploading on a new bootloader.

matejcik (approved):
utACK

prusnak:
It's always 256 bytes and always starts with TRZR

PR #2266, opened 2022-05-16
add Gather network

scream4ik:


sime:
@scream4ik Thanks for the issue.

Though, we are currently only seriously considering coins that are within the top 20 of market capitalisation.

PR #2268, opened 2022-05-17
Seed recovery for model R

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2265:
- allowing for seed recovery - `BIP39` only so far

How to try locally:
- `TREZOR_MODEL=R make build_unix`
- `./emu.py -ea`
- `trezorctl device recover` - starting recovery process

image
image
image
image
image

PR #2264, opened 2022-05-11
Seed backup for model R

grdddj:
Prototype of seed backup dialogs for model R.

How to try it locally:
- `TREZOR_MODEL=R make build_unix`
- `./emu.py -ea`
- `trezorctl device load -b -m 'alcohol woman abuse must during monitor noble actual mixed trade anger aisle'` - loading seed and marking the device for backup
- `trezorctl device backup` - starting backup process

image
image
image
image

PR #2277, opened 2022-05-19
Input methods for model R

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2267:
- passphrase entry for model R

Trying it locally:
- `TREZOR_MODEL=R make build_unix`
- `./emu.py` - does not need any erasing nor loading
- `trezorctl crypto get-entropy 1111` - I could not find a way how to spawn passphrase dialogue via `trezorctl`, so I have modified this one to do it

image
image
image
image
image


matejcik:
I changed the base to `trezor_r_ui_tests` and closed the other PRs
PIN screenshots: #2233 
Seed backup screenshots: #2264 
Seed recovery screenshots: #2268

core/embed/rust/src/ui/model_tr/component/pin.rs
```diff
@@ -0,0 +1,366 @@
...
+        self.display_text(Point::new(5, MIDDLE_ROW), previous);
+    }
+
+    fn show_next_digit(&self) {
+        let next = self.get_next_digit();
+        self.display_text(Point::new(115, MIDDLE_ROW), next);
+    }
+
+    /// Display bold white text on black background
+    fn display_text(&self, baseline: Point, text: &str) {
```
matejcik, line 209: dtto for moving helpers to top level or to a common file

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -36,6 +41,52 @@ where
...
+where
+    T: Deref<Target = str>,
+{
+    fn msg_try_into_obj(&self, msg: Self::Msg) -> Result<Obj, Error> {
+        match msg {
+            PassphrasePageMsg::Confirmed => self.passphrase().try_into(),
+            PassphrasePageMsg::Cancelled => Ok(CANCELLED.as_obj()),
+        }
+    }
+}
```
matejcik, line 88: lot of duplication here

how about
```
impl ComponentMsgObj for TextResultComponent {
    fn msg_try_into_obj(&self, msg: Self::Msg) -> Result<Obj, Error> {
        match msg {
            TextResultMsg::Confirmed => self.text_result().try_into(),
            TextResultMsg::Cancelled => Ok(CANCELLED.as_obj()),
        }
    }
}
```
and then implementing `TextResultComponent` as appropriate

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -108,6 +159,146 @@ extern "C" fn new_confirm_text(n_args: usize, args: *const Obj, kwargs: *mut Map
...
+            .into_child(),
+        )?;
+        Ok(obj.into())
+    };
+    unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
+}
+
+extern "C" fn show_share_words(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = |_args: &[Obj], kwargs: &Map| {
+        let share_words: StrBuffer = kwargs.get(Qstr::MP_QSTR_share_words)?.try_into()?;
```
matejcik, line 186: you should be accepting share words as `List`

core/embed/rust/src/util.rs
```diff
@@ -94,3 +96,12 @@ impl<T, E> ResultExt for Result<T, E> {
         }
     }
 }
+
+/// Concatenates arbitrary amount of slices into a String
+pub fn format_string<const M: usize, const N: usize>(slices: [&str; M]) -> String<N> {
```
matejcik, line 101: this would be much better as a macro
```
macro_rules! build_string {
    ($max:expr, $($string:expr),+) => {
        {
            let new_string = String::<$max>::new();
            $(new_string.push_str($string).unwrap();)+
            new_string
        }
    }
}

let string = build_string!(100, "hello", "world", number.as_str());
```

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -108,6 +159,146 @@ extern "C" fn new_confirm_text(n_args: usize, args: *const Obj, kwargs: *mut Map
...
+            middle_words.push_str(&line).unwrap();
+        }
+
+        let end_text: String<50> = util::format_string([
+            "I wrote down all ",
+            share_words_len_str.as_str(),
+            " words in order.",
+        ]);
+
+        let text_to_show: String<600> = util::format_string([
```
matejcik, line 216: why add this together instead of installing separate paragraphs?

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -108,6 +159,146 @@ extern "C" fn new_confirm_text(n_args: usize, args: *const Obj, kwargs: *mut Map
...
+
+extern "C" fn show_share_words(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = |_args: &[Obj], kwargs: &Map| {
+        let share_words: StrBuffer = kwargs.get(Qstr::MP_QSTR_share_words)?.try_into()?;
+        let title = "Recovery seed";
+
+        let share_words_len = share_words.split(',').count();
+        let share_words_len_str: String<10> = String::from(share_words_len as i16);
+
+        let beginning_text: String<50> = util::format_string([
```
matejcik, line 192: fwiw this might be a usecase for `FormattedText`

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -108,6 +159,146 @@ extern "C" fn new_confirm_text(n_args: usize, args: *const Obj, kwargs: *mut Map
...
+    };
+    unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
+}
+
+extern "C" fn request_word_count(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = |_args: &[Obj], kwargs: &Map| {
+        let title: StrBuffer = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
+        let text: StrBuffer = kwargs.get(Qstr::MP_QSTR_text)?.try_into()?;
+
+        let text: String<50> = String::from(text.trim());
```
matejcik, line 266: this is a strong argument for separating type `T` of major/minor prompt from type of choices.

here you could push `title` and `text` as the major/minor prompt, since `StrBuffer` is `AsRef<str>`, and completely separately, choices as `[&str]`

core/embed/rust/src/ui/event.rs
```diff
@@ -5,6 +5,7 @@ use core::convert::TryInto;
 pub enum PhysicalButton {
     Left,
     Right,
+    Both,
```
matejcik, line 8: we're not getting `Both` as a value from the physical layer, so this should not be here

core/src/apps/misc/get_entropy.py
```diff
@@ -20,6 +20,23 @@ async def get_entropy(ctx: Context, msg: GetEntropy) -> Entropy:
         br_code=ButtonRequestType.ProtectCall,
     )
 
+    # WARNING: Only for testing purposes to spawn the passphrase dialogue
```
matejcik, line 23: you couldn't spawn the passphrase dialog because TR is not sending `PassphraseEntry` as a capability. if you change that in `base.py`, you'll get passphrase dialogs wherever appropriate (so like `trezorctl btc get-address`)

please modify `base.py` and remove the TEMP commit

matejcik (changes requested):
main architecture notes:
* implement a ButtonController as opposed to letting individual buttons detect themselves
* modify ChoicePage to act as a subcomponent for everything
* investigate using TextBox for inputs
* use Frame to do titles for everything instead of doing them per component

`Page` doesn't seem like the right name? suggest renaming:
* `ChoicePage` => `Choices`
* `Bip39Page` => `Bip39Entry`
* `PinPage` => `PinEntry`
* `PassphrasePage` => `PassphraseEntry`

core/embed/rust/src/ui/model_tr/component/pin.rs
```diff
@@ -0,0 +1,366 @@
...
+        if self.page_counter == 0 {
+            if let Some(ButtonMsg::Clicked) = self.accept_pin.event(ctx, event) {
+                // Clicked ACCEPT. Send PIN to the client.
+                return Some(PinPageMsg::Confirmed);
+            }
+        } else if self.page_counter < 11 {
+            if let Some(ButtonMsg::Clicked) = self.ok.event(ctx, event) {
+                // Clicked CONFIRM. Append current digit to the buffer PIN string.
+                if !self.is_full() {
+                    self.append_current_digit();;
```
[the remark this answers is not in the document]
grdddj: CI warns about it, but does not fix it automatically. We however have a tool that notices this (`rustfmt`). It is "fixed" in the temporary branch for the seed recovery.

I imagine we will merge it here as soon as https://github.com/trezor/trezor-firmware/pull/2312 will be in `master`

grdddj:
Force-pushed to reflect the latest changes in `master` - most significantly using the `bip39 Wordlist`

@sime - this `PR` branch should currently support all your needs

sime:
Working for me.

grdddj:
The last two commits are focused only on code improvements, there are no new features - @sime, no need to build this for the UI/UX testing.

I tried to address most of the comments above, what is still remaining is the `ButtonController`

core/src/apps/misc/get_entropy.py
grdddj: Done in [3ef66f8](https://github.com/trezor/trezor-firmware/pull/2277/commits/3ef66f8bb111a8672374191796c192f7c430def4) and [548460c](https://github.com/trezor/trezor-firmware/pull/2277/commits/548460c338506a70e0addb1861b263d965b72e15)

core/embed/rust/src/ui/model_tr/layout.rs
grdddj: Done in [880cb15](https://github.com/trezor/trezor-firmware/pull/2277/commits/880cb1576cf15679cb98eee4181d1367a4fe929d) - the `title` and `subtitle` were moved to `Frame`, the choice components themselves to not show any description text

core/embed/rust/src/ui/model_tr/component/passphrase.rs
```diff
@@ -0,0 +1,543 @@
...
+    prev: Button<&'static str>,
+    next: Button<&'static str>,
+    accept: Button<&'static str>,
+    cancel: Button<&'static str>,
+    ok: Button<&'static str>,
+    reveal: Button<&'static str>,
+    del: Button<&'static str>,
+    page_counter: u8,
+    show_plain_passphrase: bool,
+    passphrase_buffer: String<MAX_LENGTH>,
```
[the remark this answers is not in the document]
grdddj: Done in [4630051](https://github.com/trezor/trezor-firmware/pull/2277/commits/4630051f31d4822486d01ac0fb02eb31dbd62845)

core/embed/rust/src/ui/model_tr/component/passphrase.rs
```diff
@@ -0,0 +1,543 @@
...
+    fn show_previous(&self) {
+        self.display_text(Point::new(5, MIDDLE_ROW), self.get_previous());
+    }
+
+    fn show_next(&self) {
+        self.display_text(Point::new(115, MIDDLE_ROW), self.get_next());
+    }
+
+    /// Display bold white text on black background
+    fn display_text(&self, baseline: Point, text: &str) {
```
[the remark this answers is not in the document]
grdddj: `ui::model_tr::component::common::display_bold`, `display_bold_center` and `display_bold_right` helpers created in [880cb15](https://github.com/trezor/trezor-firmware/pull/2277/commits/880cb1576cf15679cb98eee4181d1367a4fe929d)

core/embed/rust/src/ui/model_tr/component/passphrase.rs
```diff
@@ -0,0 +1,543 @@
...
+        self.get(self.page_counter - 1)
+    }
+
+    fn get_next(&self) -> &'static str {
+        self.get(self.page_counter + 1)
+    }
+
+    fn get(&self, index: u8) -> &'static str {
+        let index = index as usize;
+        match self.current_category {
```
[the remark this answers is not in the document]
grdddj: In the end, I made the `ChoicePage` simply reusable, because I did not manage to properly clean the previous instance when creating a new one

core/embed/rust/src/ui/model_tr/component/pin.rs
```diff
@@ -0,0 +1,366 @@
...
+const MIDDLE_ROW: i32 = 72;
+
+const MAX_LENGTH: usize = 50;
+const MAX_VISIBLE_DOTS: usize = 18;
+const MAX_VISIBLE_DIGITS: usize = 18;
+const HOLD_DURATION: Duration = Duration::from_secs(2);
+
+const DIGITS: [&str; 10] = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "0"];
+
+pub struct PinPage<T> {
```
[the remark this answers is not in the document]
grdddj: Done in [4630051](https://github.com/trezor/trezor-firmware/pull/2277/commits/4630051f31d4822486d01ac0fb02eb31dbd62845) and [880cb15](https://github.com/trezor/trezor-firmware/pull/2277/commits/880cb1576cf15679cb98eee4181d1367a4fe929d)

core/embed/rust/src/util.rs
grdddj: Done in [4630051](https://github.com/trezor/trezor-firmware/pull/2277/commits/4630051f31d4822486d01ac0fb02eb31dbd62845)

core/embed/rust/src/ui/model_tr/layout.rs
grdddj: I tried to create separate paragraphs, but then all the paragraphs need to have the same content-type - so all three would need to be `String<500>`, even though the beginning and end are very small.

I also did not succeed to combine the `Paragraphs` with `FormattedText`

core/embed/rust/src/ui/model_tr/layout.rs
grdddj: `FormattedText` are then not compatible with `Paragraphs` - but maybe there is a better how to display them

core/embed/rust/src/ui/model_tr/layout.rs
grdddj: I find this not so bad because each component is special - some allow `Cancelled`, some send result in `Result(String<X>)` etc.

core/embed/rust/src/ui/model_tr/component/choice.rs
```diff
@@ -0,0 +1,202 @@
...
+    pad: Pad,
+    prev: Button<&'static str>,
+    next: Button<&'static str>,
+    ok: Button<&'static str>,
+    page_counter: u8,
+}
+
+impl<T, const N: usize> ChoicePage<T, N>
+where
+    T: Deref<Target = str>,
```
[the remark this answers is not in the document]
grdddj: Done in [880cb15](https://github.com/trezor/trezor-firmware/pull/2277/commits/880cb1576cf15679cb98eee4181d1367a4fe929d)

core/embed/rust/src/ui/model_tr/component/choice.rs
```diff
@@ -0,0 +1,202 @@
...
+    choices: Vec<T, N>,
+    both_button_press: BothButtonPressHandler,
+    pad: Pad,
+    prev: Button<&'static str>,
+    next: Button<&'static str>,
+    ok: Button<&'static str>,
+    page_counter: u8,
+}
+
+impl<T, const N: usize> ChoicePage<T, N>
```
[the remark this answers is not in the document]
grdddj: Done in [880cb15](https://github.com/trezor/trezor-firmware/pull/2277/commits/880cb1576cf15679cb98eee4181d1367a4fe929d)

core/embed/rust/src/ui/model_tr/component/bip39.rs
```diff
@@ -0,0 +1,332 @@
...
+        } else {
+            self.offer_words = false;
+            self.letter_choices =
+                bip39::get_all_possible_letters(self.word_buffer.clone(), self.all_words.clone());
+        }
+
+        self.page_counter = 0;
+    }
+
+    fn render_header(&self) {
```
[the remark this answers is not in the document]
grdddj: Done in [880cb15](https://github.com/trezor/trezor-firmware/pull/2277/commits/880cb1576cf15679cb98eee4181d1367a4fe929d)

core/embed/rust/src/ui/model_tr/component/bip39.rs
```diff
@@ -0,0 +1,332 @@
...
+    offer_words: bool,
+}
+
+impl<T> Bip39Page<T>
+where
+    T: Deref<Target = str>,
+{
+    pub fn new(prompt: T) -> Self {
+        // Getting the initial choices (basically whole alphabet and all words)
+        let word_buffer = String::new();
```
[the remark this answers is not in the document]
grdddj: Done in [4630051](https://github.com/trezor/trezor-firmware/pull/2277/commits/4630051f31d4822486d01ac0fb02eb31dbd62845)

core/embed/rust/src/ui/model_tr/component/bip39.rs
```diff
@@ -0,0 +1,332 @@
...
+
+const LETTERS_ROW: i32 = 40;
+const MIDDLE_ROW: i32 = 72;
+
+/// Offer words when there will be fewer if them than this
+const WORD_THRESHOLD: usize = 10;
+
+pub struct Bip39Page<T> {
+    prompt: T,
+    letter_choices: Vec<String<1>, 50>, // TODO: could use char
```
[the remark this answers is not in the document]
grdddj: Done in [9ec03b2](https://github.com/trezor/trezor-firmware/pull/2277/commits/9ec03b2edf75c6d6e7f00b472f7c120520341602)

core/embed/rust/src/ui/model_tr/component/passphrase.rs
```diff
@@ -0,0 +1,543 @@
...
+];
+const MENU: [&str; 4] = ["abc", "ABC", "123", "*#_"];
+
+pub struct PassphrasePage<T> {
+    prompt: T,
+    max_len: u8,
+    lowercase_choices: [&'static str; 26],
+    uppercase_choices: [&'static str; 26],
+    digits_choices: [&'static str; 10],
+    special_choices: [&'static str; 30],
```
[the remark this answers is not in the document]
grdddj: Removed in [880cb15](https://github.com/trezor/trezor-firmware/pull/2277/commits/880cb1576cf15679cb98eee4181d1367a4fe929d)

grdddj:
Force-pushed after rebase on `master` and on https://github.com/trezor/trezor-firmware/pull/2339 - looking forward on that being merged! (I have included that PR as a single squashed commit)

NOTE on https://github.com/trezor/trezor-firmware/pull/2339: Here there are some changes to the `HoldToConfirm` and `Loader` code, making both components general with `T: AsRef<str>`, so they could be used with `ButtonController` and `ButtonContainer`, which are also `T: AsRef<str>`. I think it does not affect that PR, so it could be merged as-is and extended here.

Thanks @TychoVrahe for help with the screen flickering in the last commit, now it seems stable enough. We are currently re-drawing the buttons only when at least one of them changes (I was trying to change them individually, but failed, even with `pad` on each button and clearing it).

The next step is using the `ButtonController` in all user-facing UI components, not only in current input methods (PIN, passphrase etc.)

grdddj:
Force-pushed after rebase on `master` to have the [MacOS nix-shell fix](https://github.com/trezor/trezor-firmware/commit/888ff2318158515b60571ae9ab04e8d8306f22e5)

core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -234,6 +241,8 @@ async def confirm_backup(ctx: wire.GenericContext) -> bool:
         ctx=ctx,
         title="Success",
         data="New wallet created successfully!\n\nYou should back up your new wallet right now.",
+        verb="Back up",
```
sime, line 244: kkk

grdddj:
Force-pushing to add the recent `PR` with `printing`.

Last two commits include the new `Unifont` font and some changes needed because of this. There are still quite a lot of spaces where the new fonts are creating some trouble, as its size has changed.

As new fonts are also slightly larger in byte-size, currently we are overflowing flash on hardware in debug mode - for `500` bytes it says. **Therefore, please build the firmware with `BITCOIN_ONLY=1`**

Some screenshots with the new fonts:
image
image
image

sime:
> As new fonts are also slightly larger in byte-size, currently we are overflowing flash on hardware in debug mode - for `500` bytes it says. **Therefore, please build the firmware with `BITCOIN_ONLY=1`**

I have an inherit fear that with the fonts will blow away our flash space, please keep an eye on this if it becomes reality.

grdddj:
After the recent commits (dealing mostly with font magnification and some pixel-shiftings), my hardware device is experiencing `Internal error (HF)` during early stages of recovery. I have not encountered that in any other input methods.

I am investigating it.

grdddj:
[97e8180](https://github.com/trezor/trezor-firmware/pull/2277/commits/97e81801f78cf5473a4b4aab3f7e60ca5c623342) addresses the `Hard fault` during the `bip39` recovery process.

A more detailed description of the `HF` conditions:
- happened when `micropython` was requesting multiple words from `Rust` in a loop (for 12 words it needed to call the `request_word_bip39` 12 times in a row)
- first and (mostly) second requests were fine, but the third triggered that `HF` (almost without exception)
- trying to isolate the issue uncovered that the main factor in "having HF/not" is the number of items in `ChoicePage` - with lower numbers, it was possible to call the function repeatedly, with higher amounts it was failing even in the second request

I altered the process of retrieving the `bip39` words - all of them are retrieved in just one request, so all the `Rust` layout things need to be instantiated only once. It means that at least for now the `model R` and `model T` have different `Rust API` (it needs branching the caller code in `micropython`). Also, it might be fine to do it for `BIP39`, but we need to support `SLIP39` as well and there are some on-the-fly checks for `slip39` after each word - https://github.com/trezor/trezor-firmware/blob/426eae4dfcb853041419030bbb104cc2bbf6629f/core/src/apps/management/recovery_device/layout.py#L61, so these things would need to be transferred into `Rust` if we would also collect all `slip39` words at once. 

---

It probably does not address the root issue, but I did not succeed in finding out where exactly is the issue (with some memory profiling etc.).

One thing I could do is to tell exactly which commit made it stop working (it worked reliably before and not worked at all after) - cfc9e2e68 - which adds support for icons as buttons, and therefore probably increases the memory footprint of the `ChoiceItem`s that `ChoicePage` needs to allocate for.

---

What I still cannot fully understand is how it can work once or twice and fail the third time of the exact same call. The premise is that the `micropython` environment is still the same (as all the calls to `Rust` happen in one loop) - that would signal to me there is no cleaning of the "old" layouts when calling "new" one. 

Small experiment:
- replacing the call to `request_word` with `request_passphrase_on_device` also failed at the 2nd/3rd attempt
- on its own `request_passphrase_on_device` in a `while True` triggered `HF` _always_ after 5 attempts
- calling it one-by-one being triggered with `trezorctl` command did not throw `HF` even after much more attempts
- that would conclude to me that the homescreen is somehow "resetting/cleaning" the state

grdddj:
[02295a2](https://github.com/trezor/trezor-firmware/pull/2277/commits/02295a2cf1a4cc21e9feff0390c2d54234a80e5e) contains a functional `POC` of one component that can satisfy theoretically any step of actions to be done on the screen - and can be defined in a "Rusty" way of chaining the method calls on one object instead of that awkward format string.

There is actually not a lot of new functionality, the biggest differences between this and the previous "format" attempt are that it does not reuse `FormattedText`, does not use `format` string, and the `Op`s are stored directly in the component.

A lot of code in `flow_pages_poc_helpers.rs` is almost fully copypasted from `ui/component/text` - so that I could make some small changes there for this `POC`.

The calling code in `layout.rs` looks like this:
```rs
FlowPageMaker::new::<theme::TRDefaultText>(btn_layout, btn_actions)
    .icon_label_text(
        theme::ICON_USER,
        "Recipient".into(),
        truncated_address.as_ref().into(),
    )
    .newline()
    .icon_label_text(theme::ICON_AMOUNT, "Amount".into(), amount.as_ref().into())


```
to create two common patterns of icon, offset, normal-text label, newline and bold-text text

or 
```rs
FlowPageMaker::new::<theme::TRDefaultText>(screen.2.clone(), screen.3.clone())
    .text_bold(screen.0.into())
    .newline()
    .text_normal(screen.1.into())
```
to just have some normal text in combination with bold text

grdddj:
[9542554](https://github.com/trezor/trezor-firmware/pull/2277/commits/9542554531185a5ffd800dd97214f80e7c1d51e6) just makes it running on the hardware device. By previously allocating the same fixed amount (in `Vec<Op<String<100>>, 30>`) for each `FlowPage`, it failed to allocate space for device tutorial, which has 8 flow pages (a lot).

Temporarily solved by each `Flow` setting the suitable numbers for itself - a page with a small number of operations but longer text can choose for example `Vec<Op<String<120>>, 8>` and a page with shorter texts but more operations can define  `Vec<Op<String<25>>, 25>`.

It creates little bit of a mess with those `const X: usize`, as `Flow` currently looks like `Flow<T, const N: usize, const M: usize, const S: usize>`, as it also needs to choose the amount of the `FlowPage`s.

That is definitely not the cleanest and may be dangerous for rendering some outside-world strings, where somebody (some new cryptocurrency) would for example send `101` characters long address and we would only allocate `String<100>` (we could check the length before making a String out of it theoretically, but still not best).

It is also not very nice that the high-level code in `layout.rs` must be choosing these constants. We could create some components/helpers/primitives that would handle the allocation for the common tasks that we have - for example pairs of icon+label+text, but then we would need to support one pair, two pairs, three pairs...

I guess I am starting to almost dislike this one component for all screens, mostly because of the needed allocation issues, which can be really dangerous. I just cannot see another way right now how to generally support the use case "I want to place icons, I want to draw text and I do not know if it will fit one screen, therefore I need pagination".

PS: why do I use `Op::Text(String<N>)` instead of something else like `Op::Text(StrBuffer)`, which would be much more memory-efficient? (Am I right Imagining that in that case, the `Op` enum would not need almost any allocation, as `StrBuffer` is just a pointer to already existing data?) Currently for handling the page-break we need to manipulate the boundary text - when starting a new page, we need to cut out the already shown part and preserve only the new text-to-be-rendered. But maybe the component can store all the operations with `StrBuffer` and only create `String` Ops (or some other objects) for the sake of rendering the current page - here the concept of `Line` would be nice...

grdddj:
Replacing `String<N>` by `StrBuffer` in [5f80757](https://github.com/trezor/trezor-firmware/pull/2277/commits/5f807578381203dbfdd6563db27a79c7840a3813) makes me much more comfortable with it. Instead of modifying the `Op::Text` on the page break, I am setting the `length` parameter of how much is left to display from that StrBuffer - and it means the text content itself stays the same all the time, and we can still display slices of it of arbitrary length.

grdddj:
In [5ff1e44](https://github.com/trezor/trezor-firmware/pull/2277/commits/5ff1e4480fbbe63514985f56e7e8a1adb9680ab4) I am adding the possibility to get all the information from `Rust` layouts in `micropython` - via sending the `Trace` and processing it into useful data.

This is a starting point to implementing the automatic pagination and also will be useful for automatic UI tests.

Below is the example of the data that we have access to in `micropython` and can act upon it - the page indexes, which button goes next, screen text, button actions etc.

image

grdddj:
Current status regarding `flash` space:
- non-debug build is overflowing `.flash` for `23 kb`
- debug build overflows for `41 kb`

Such a big difference was probably caused by the recent `impl Trace`, which is `ui_debug`-only. We are probably still fine, however, as long as we manage to split the code into `.flash2`, where we should have enough space.

Estimate of free `.flash2` space (for model `R`):
- currently there are `50kb` free
- there is roughly `10-20` kb of `micropython` `UI` code that still gets included, so we can get rid of these (granted we have the functionality already implemented in `Rust`)
- if we get rid of the `ETH` definitions, that could save `50-70 kb`

Considering this, even the missing `41kb` for `debug` build do not look so bad.

grdddj:
[19e49d7](https://github.com/trezor/trezor-firmware/pull/2277/commits/19e49d7545d7f1fe28c292ff1afe2e75c017888e) adds two `CI` dashboards for quickly checking [all unique screens](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2885747410/artifacts/test_ui_report/all_unique_screens.html) and [all testcases in one place](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2885747410/artifacts/test_ui_report/all_screens.html). These two dashboards are also linked to each other by clicking the specific images.

grdddj:
Force-pushed after rebase on `master` so that I have the latest `success animation` changes there. Also changed the base branch to `master`, effectively merging https://github.com/trezor/trezor-firmware/pull/2252 into this `PR` as well.

The rebasing was not straightforward, so there is a possibility of some bug being introduced in it. I have tested some basic workflows on a hardware device and they were fine. For the user-testing, @sime, if there would be any issues on this "new" rebased version, the last "old" CI is [this](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/613892516) - so you can get the firmware from there, just in case (it already includes the double-PIN screen).

grdddj:
Force-pushed after rebase on `master` to include some latest changes.

What is unfortunate now is that on device almost all more advanced workflows (apart from seed backup) are throwing `Internal error (SO)` - in both the `debug` and `nondebug` mode. I am looking into it...

grdddj:
`Stack overflow` issues solved by lazy-loading the pages - `ChoicePage` requesting only what is needed right now, so just the current page needs to be stored on the stack.

Using
```rs
pub trait ChoiceFactory {
    fn get(&self, choice_index: u8) -> ChoiceItem;
    fn count(&self) -> u8;
}
```
to transmit information about all the screens - each specific input (`PIN` , `bip39`  etc.) implements this and returns the specific `ChoiceItem`  according to the current choice_index .

Similar was done for the `Flow`  pages - in `layout.rs`  we are defining a closure function specifying what `Page` corresponds to a specific index - without creating a `heapless::Vec`  which caused the `SO`.

---

[5125666](https://github.com/trezor/trezor-firmware/pull/2277/commits/512566640fda88b73809735cf579bedea3cb9daf) is introducing a way to get object sizes on the fly using the `debug` build and `gdb`.

It was created so that we can verify how much memory (stack) space all the layouts (and `Rust` code in general) are occupying, so we can avoid the `SO` issues.

From the results below it is apparent that the lazy-loading approach to page rendering works, the object sizes are much smaller than with the long `heapless::Vec`s with instantiated objects.

<details><summary>Long results with object/layout sizes</summary>
<p>

(some things like `FormattedText` or `Paragraphs` depends quite a lot on the input data - but others are fairly constant)

```
ChoicePage: 1079
ChoicePage.choices: 1
ChoicePage.pad: 20
ChoicePage.buttons: 1049
ChoicePage.page_counter: 1
ChoicePage.is_carousel: 1

ButtonPage: 1828
ButtonPage.content: 392
ButtonPage.scrollbar: 40
ButtonPage.pad: 20
ButtonPage.cancel_btn_details: 79
ButtonPage.confirm_btn_details: 79
ButtonPage.back_btn_details: 79
ButtonPage.next_btn_details: 79
ButtonPage.buttons: 1049

PassphraseEntry: 1147
PassphraseEntry.choice_page: 1079
PassphraseEntry.show_plain_passphrase: 1
PassphraseEntry.textbox: 64
PassphraseEntry.current_category: 1
PassphraseEntry.menu_position: 1

Paragraphs: 392
Paragraphs.area: 16
Paragraphs.list: 344
Paragraphs.placement: 8
Paragraphs.offset: 16
Paragraphs.visible: 8

FormattedText: 480
FormattedText.layout: 40
FormattedText.fonts: 16
FormattedText.format: 16
FormattedText.args: 200
FormattedText.icon_args: 200
FormattedText.char_offset: 8

SimpleChoice: 1254
SimpleChoice.choices: 88
SimpleChoice.choice_page: 1166

PassphraseEntry: 1147
PassphraseEntry.choice_page: 1079
PassphraseEntry.show_plain_passphrase: 1
PassphraseEntry.textbox: 64
PassphraseEntry.current_category: 1
PassphraseEntry.menu_position: 1

PinEntry: 1161
PinEntry.show_real_pin: 1
PinEntry.textbox: 64
PinEntry.choice_page: 1094

Bip39Entry: 1950
Bip39Entry.choice_page: 1374
Bip39Entry.letter_choices: 112
Bip39Entry.textbox: 32
Bip39Entry.pad: 20
Bip39Entry.offer_words: 1
Bip39Entry.bip39_words_list: 16
Bip39Entry.words: 392
Bip39Entry.word_count: 1

Flow: 2102
Flow.pages: 1
Flow.page_counter: 1
Flow.pad: 20
Flow.common_title: 24
Flow.current_page: 984
Flow.buttons: 1049

Page: 984
Page.ops: 608
Page.layout: 40
Page.btn_layout: 239
Page.btn_actions: 72
Page.current_page: 8
Page.page_count: 8
Page.char_offset: 8
```

</p>

grdddj:
Updated `flash` sizes for `Rust` [here](https://gist.github.com/grdddj/f75e8859fa074c1a5317e4e82a161c89)

It currently has `77 kb` in `debug` mode, which is not little.

Quite a lot (almost `15 kb`) is occupied by `Trace` and its implementations from all objects.

Some increases could also be seen due to `SO` mitigation. Instead of creating the `heapless::Vec` with objects on `stack`, the logic to create pages on demand is offloaded into functions living in `flash` space (obvious side-effect).

grdddj:
Replaced by https://github.com/trezor/trezor-firmware/pull/2610

PR #2279, opened 2022-05-20
ci: update arm jobs to run on tags and use the official nix image

vdovhanych:
This fixes the build job of arm emulators to run on tagged branches. Meaning the deployment after release should pass after this. I also changed the image to upstream nixos one as the issues with the old one were fixed.


prusnak (approved):

matejcik:
@vdovhanych please run `make gen`

matejcik (approved):

PR #2288, opened 2022-05-25
Type hints to emu.py

grdddj:
Improving the quality of an important development tool

### DIFF
--- a/core/emu.py
+++ b/core/emu.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+from __future__ import annotations
+
 import logging
 import os
 import platform
@@ -6,8 +8,8 @@ import signal
 import subprocess
 import sys
 import tempfile
-import time
 from pathlib import Path
+from typing import TextIO
 
 import click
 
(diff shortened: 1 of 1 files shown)


core/emu.py
```diff
@@ -1,13 +1,15 @@
 #!/usr/bin/env python3
+from __future__ import annotations
```
prusnak, line 2: Is this required?

I don't see any annotation where we'd need postponed evaluation, or I am wrong here?

grdddj: I am using "union pipe" two times  - `str | Path` - and without the `future` import this fails at runtime - as we have default `3.9` python, it works natively from `3.10`

prusnak: thanks, TIL

core/emu.py
```diff
@@ -47,13 +49,14 @@ def run_command_with_emulator(emulator, command):
...
 
-def run_emulator(emulator):
+def run_emulator(emulator: CoreEmulator) -> int:
     with emulator:
         signal.signal(signal.SIGINT, signal.SIG_IGN)
         return emulator.wait()
 
 
-def watch_emulator(emulator):
+def watch_emulator(emulator: CoreEmulator) -> Literal[0]:
```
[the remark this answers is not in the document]
grdddj: Yes, makes sense. Changed in [6783e62](https://github.com/trezor/trezor-firmware/pull/2288/commits/6783e62cb5a3ca04a0a8981dd5892b659737c4fe)

matejcik (approved):
👍  please squash fixup and merge

PR #2289, opened 2022-05-25
Implement SLIP-0025 CoinJoin accounts.

andrewkozlik:
Implements [SLIP-0025](https://github.com/satoshilabs/slips/blob/master/slip-0025.md) CoinJoin accounts.

Summary of PR:
- Any message which attempts to access the BIP32 subtree `m/10025'` needs to be preceded by `DoPreauthorized` or `UnlockPath`. These would most notably be `GetPublicKey`, `GetAddress` or interactive `SignTx` (when sending from a CoinJoin account or when sending to a CoinJoin account, if Suite references the destination using `address_n`).
- `AuthorizeCoinjoin` works only with SLIP-0025 path patterns, unless safety checks are disabled.
- `GetAddress` is not allowed to access the internal chain of the SLIP-0025 pattern, unless safety checks are disabled.

This PR adds a new type of message `UnlockPath` which can be used to enable access to SLIP-25 paths. It works similar to `DoPreauthorized`.
1. The client starts by sending `UnlockPath(address_n=[10025+0x80000000])`.
2. Trezor asks the user to confirm access to CoinJoin accounts (`ButtonRequest(code=Other)`).
3. Trezor responds with `UnlockedPathRequest` containing a `mac` value, which should be saved by the client for future use (call it `saved_mac`).
4. The client sends the `GetPublicKey`, `GetAddress` or interactive `SignTx` message, which requires access to a SLIP-25 account, and continues as usual.

The next time the client needs to access a SLIP-25 path the user will not need to confirm a dialog if the `saved_mac` is provided:
1. The client sends `UnlockPath(address_n=[10025+0x80000000], mac=saved_mac)`.
2. Trezor checks that the provided `mac` is correct and responds with `UnlockedPathRequest`.
3. The client sends the `GetPublicKey`, `GetAddress` or interactive `SignTx` message, which requires access to a SLIP-25 account.

The `saved_mac` is constant for the given combination of seed and passphrase. If the user chooses to remember their wallet, then Suite should store the `saved_mac` together with the SLIP-25 account XPUB so that the user does not have to confirm access to the CoinJoin account next time.

The design uses an authentication code as proof of user authorization. We wanted to use the XPUB for this, but it seems too complex, because there are a lot of parameters that go into an XPUB, e.g. the script type determines whether it's an `xpub`, `ypub` or a `zpub`.

UI changes:
1. The confirmation dialog for `UnlockPath`.
2. The `GetAddress` header overflows a bit, but that can happen with regular accounts too.

image image


python/src/trezorlib/device.py
```diff
@@ -220,6 +220,20 @@ def cancel_authorization(client: "TrezorClient") -> "MessageType":
...
+@expect(messages.UnlockedPathRequest, field="mac", ret_type=bytes)
+def unlock_path(client: "TrezorClient", n: "Address") -> "MessageType":
+    resp = client.call(messages.UnlockPath(address_n=n))
+
+    # Cancel the UnlockPath workflow now that we have the authentication code.
+    cancel_resp = client.call_raw(messages.Cancel())
+    if not isinstance(cancel_resp, messages.Failure):
+        raise RuntimeError("Invalid response")
+    if cancel_resp.code != messages.FailureType.ActionCancelled:
+        raise TrezorFailure(cancel_resp)
```
matejcik, line 232:
```suggestion
    try:
        client.cancel()
    except exceptions.Cancelled:
        return resp
    else:
        raise TrezorException("Unexpected response in UnlockPath flow")
```

core/src/apps/bitcoin/get_public_key.py
```diff
@@ -2,21 +2,39 @@
...
+        if auth_msg is not None and AuthorizeCoinJoin.is_type_of(auth_msg):
+            if not authorization.from_cached_message(auth_msg).check_get_public_key(
+                msg
+            ):
+                raise wire.ProcessError("Unauthorized operation")
+        elif auth_msg is not None and UnlockPath.is_type_of(auth_msg):
+            if auth_msg.address_n != msg.address_n[: len(auth_msg.address_n)]:
+                raise FORBIDDEN_KEY_PATH
+        else:
+            raise FORBIDDEN_KEY_PATH
```
matejcik, line 36: this would be much nicer linearized in a standalone function:
```
def should_unlock_slip25_path(address_n, auth_msg):
     if not address_n or address_n[0] != SLIP25:
          return False
     if auth_msg is None:
          return False
     ...
```

---

Optionally, you can make a custom schema object:
```
class Slip25AwareSchema:
     @staticmethod
     def match(path: Bip32Path) -> bool:
            return not path or path[0] != paths.SLIP25_PURPOSE

if authorization.is_slip25_unlocked(auth_msg):
      schema = paths.AlwaysMatchingSchema
else:
      schema = Slip25AwareSchema
```
or even integrate the auth msg and the unlock check into it. But the "if not unlocked: raise FORBIDDEN_KEY_PATH" is working here nicely, toying with schemas would only start being useful if used in more places.

core/src/apps/bitcoin/keychain.py
```diff
@@ -234,23 +243,44 @@ def get_coin_by_name(coin_name: str | None) -> coininfo.CoinInfo:
...
 
+def _get_unlock_pattern(
+    msg: MessageType, auth_msg: MessageType | None = None
+) -> str | None:
+    if AuthorizeCoinJoin.is_type_of(msg):
+        return PATTERN_SLIP25_TAPROOT
+
+    if (
+        auth_msg is not None
+        and (AuthorizeCoinJoin.is_type_of(auth_msg) or UnlockPath.is_type_of(auth_msg))
```
matejcik, line 263: fwiw you're not calling `check_get_public_key` here so hypothetically speaking the preauthorization could be using some other path.

maybe change `check_get_public_key` into `check_path_access` or something, and then reuse the function that I proposed for `get_public_key` here as well?

common/protob/messages-management.proto
```diff
@@ -441,6 +441,7 @@ message DoPreauthorized {
  * @start
  * @next SignTx
  * @next GetOwnershipProof
+ * @next GetPublicKey
```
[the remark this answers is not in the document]
andrewkozlik: It seems logical that if the user already authorized CoinJoin, then by doing so they also authorized access to the public key. On the other hand typically `UnlockPath` should be called before `AuthorizeCoinJoin` (only once for a remembered account), so I suppose calling GetPublicKey with DoPreauthorized is a bit superfluous.

core/src/apps/bitcoin/keychain.py
andrewkozlik: Something like this: https://github.com/trezor/trezor-firmware/pull/2289/commits/9c7d2307c1b74af6f78683f042411ccbe3513652

andrewkozlik: I pushed some more changes to improve the commentary https://github.com/trezor/trezor-firmware/pull/2289/commits/e0803c17fa2cdb54d1bb17505fb3dfab48f2e127.

common/protob/messages-management.proto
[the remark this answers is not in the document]
andrewkozlik: Removed the option to call GetPublicKey with DoPreauthorized. Seem like overengineering. https://github.com/trezor/trezor-firmware/pull/2289/commits/e0803c17fa2cdb54d1bb17505fb3dfab48f2e127 https://github.com/trezor/trezor-firmware/pull/2289/commits/b4aa745442ef91b21591d5c744d885c7eac06216

core/src/apps/bitcoin/get_public_key.py
andrewkozlik: TBH I don't understand the proposed simplification. Since I removed the option to call GetPublicKey with DoPreauthorized in https://github.com/trezor/trezor-firmware/commit/e0803c17fa2cdb54d1bb17505fb3dfab48f2e127 it seems simple enough now.

python/src/trezorlib/device.py
andrewkozlik: Fixup needed https://github.com/trezor/trezor-firmware/pull/2289/commits/c421a7fe61c7124419ca79a3727bdc40f63af212, since `client.cancel()` does not wait for the response.

core/src/apps/bitcoin/keychain.py
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2289/commits/47a80169b7d09af48410b4f9a53a843536380a99: I reverted the changes to `PathSchema.match()` made in https://github.com/trezor/trezor-firmware/commit/9c7d2307c1b74af6f78683f042411ccbe3513652 and instead introduced `PathSchema.restrict()`.

core/src/apps/base.py
```diff
@@ -183,6 +183,64 @@ async def handle_DoPreauthorized(
...
+    node = Slip21Node(seed)
+    node.derive_path(_KEYCHAIN_MAC_KEY_PATH)
+    mac = utils.HashWriter(hmac(hmac.SHA256, node.key()))
+    for i in msg.address_n:
+        write_uint32_le(mac, i)
+    expected_mac = mac.get_digest()
+
+    # Require confirmation to access SLIP25 paths unless already authorized.
+    if msg.mac:
+        if len(msg.mac) != len(expected_mac) or not utils.consteq(
```
matejcik, line 212: the length check is already covered by `consteq`, no?

core/src/apps/base.py
```diff
@@ -183,6 +183,64 @@ async def handle_DoPreauthorized(
...
+        )
+
+    wire_types = (
+        MessageType.Cancel,
+        MessageType.GetAddress,
+        MessageType.GetPublicKey,
+        MessageType.SignTx,
+    )
+    req = await ctx.call_any(UnlockedPathRequest(mac=expected_mac), *wire_types)
+    if Cancel.is_type_of(req):
```
matejcik, line 231: do we need explicit handling of Cancel? if you leave it out, receiving Cancel will raise `UnexpectedMessage` out of this handler, and invoke the `Cancel` handler, to the same effect istm

core/src/apps/base.py
```diff
@@ -183,6 +183,64 @@ async def handle_DoPreauthorized(
...
+        raise wire.ActionCancelled
+
+    assert req.MESSAGE_WIRE_TYPE is not None
```
matejcik, line 234: how about `assert req.MESSAGE_WIRE_TYPE in wire_types`, and then `assert handler is not None`, which is what should be true anyway?

core/src/apps/bitcoin/get_public_key.py
```diff
@@ -2,21 +2,33 @@
...
+    ctx: wire.Context, msg: GetPublicKey, auth_msg: MessageType | None = None
+) -> PublicKey:
     coin_name = msg.coin_name or "Bitcoin"
     script_type = msg.script_type or InputScriptType.SPENDADDRESS
     coin = coininfo.by_name(coin_name)
     curve_name = msg.ecdsa_curve_name or coin.curve_name
 
+    if msg.address_n and msg.address_n[0] == paths.SLIP25_PURPOSE:
+        # UnlockPath is required to access SLIP25 paths.
+        if auth_msg is None or not UnlockPath.is_type_of(auth_msg):
```
matejcik, line 25: i believe `is_type_of(None)` would correctly return False, but maybe the typechecker won't allow that? (which might be fixable by allowing an argument of type `Any` to `is_type_of`)

core/src/apps/base.py
```diff
@@ -183,6 +183,64 @@ async def handle_DoPreauthorized(
...
+    from apps.common.paths import SLIP25_PURPOSE
+    from apps.common.seed import Slip21Node, get_seed
+    from apps.common.writers import write_uint32_le
+
+    _KEYCHAIN_MAC_KEY_PATH = [b"TREZOR", b"Keychain MAC key"]
+
+    # UnlockPath is relevant only for SLIP-25 paths.
+    # Note: Currently we only allow unlocking the entire SLIP-25 purpose subtree instead of
+    # per-coin or per-account unlocking in order to avoid UI complexity.
+    if len(msg.address_n) != 1 or msg.address_n[0] != SLIP25_PURPOSE:
```
matejcik, line 199: `if msg.address_n != [SLIP25_PURPOSE]` would be cleaner

core/src/apps/base.py
andrewkozlik: No, the caller needs to guarantee that `len(sec) >= len(pub)`. This is important in case `len(sec)` is itself secret information. I actually think this is the most dangerous function that we have in python.

https://github.com/trezor/trezor-firmware/blob/0b4ccf45fc5a7de9ccff48dd33c8458bff1d20c9/core/embed/extmod/modtrezorutils/modtrezorutils.c#L50-L75

We could make it better by:
1. Adding `assert(secbuf.len >= pubbuf.len);` at the beginning, so that we at least have a better chance of catching programming bugs in the debug build.
2. Iterating only up to the minimum length, which can be computed in constant time:
    ```c
    // Compute minlen = min(secbuf.len, pubbuf.len) in constant time.
    uint32_t mask = -(secbuf.len < pubbuf.len);
    uint32_t minlen = (secbuf.len & mask) | (pubbuf.len & ~mask);
    ```
    However, if `secbuf.len < pubbuf.len`, then the timing of the loop reveals the length of `secbuf`. This is a problem if the caller actually knows that they can read past the buffer's length, because more bytes are allocated, e.g. in case of `bytearray`. BTW, in that case we could check `mp_obj_array_t::free`, but not with `mp_buffer_info_t` directly. The question is whether there actually is any place in our code that uses this feature.
3. Accessing `s` modulo `secbuf.len`, i.e.
    ```c
    diff |= s[i % secbuf.len] - p[i];
    ```
    Not sure how to deal with `secbuf.len==0` though. Also not sure whether the `%` operation is constant-time.

core/src/apps/base.py
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2289/commits/aafaadabaedac2e0f35e98a07985fb782bbd19b0

core/src/apps/base.py
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2289/commits/aafaadabaedac2e0f35e98a07985fb782bbd19b0

core/src/apps/base.py
andrewkozlik: The implicit handling of `Cancel` seems to work just fine. https://github.com/trezor/trezor-firmware/pull/2289/commits/aafaadabaedac2e0f35e98a07985fb782bbd19b0

core/src/apps/bitcoin/get_public_key.py
andrewkozlik: > i believe `is_type_of(None)` would correctly return False,

Yes, done in https://github.com/trezor/trezor-firmware/pull/2289/commits/aafaadabaedac2e0f35e98a07985fb782bbd19b0.

> which might be fixable by allowing an argument of type `Any` to `is_type_of`)

Done in https://github.com/trezor/trezor-firmware/pull/2289/commits/1eb3c7f945ed31224ab31ab0dbd000714a536769.

core/src/apps/base.py
andrewkozlik: I created a separate issue for this https://github.com/trezor/trezor-firmware/issues/2466/.

matejcik (approved):

andrewkozlik:
The core debug firmware no longer fits onto the flash memory of the device. In https://github.com/trezor/trezor-firmware/pull/2289/commits/eb182b923f39111d9307a25a86a49d395c6611b0 I reduced `ECMULT_GEN_PREC_BITS` from 4 to 2 to decrease the size of the precomputed tables for secp256k1-zkp and asked QA to verify if there are any significant changes in the signing time, see https://www.notion.so/satoshilabs/Signing-speed-test-377f09c64774480c8dd4a9e1727e5208. If they don't find any significant issues, then I will merge.

andrewkozlik:
According to the test report signing 100 Taproot inputs takes 50 seconds on this branch, which is the same time as on firmware release 2.5.2.

PR #2294, opened 2022-05-25
fix(legacy,core): use sat/vB only for segwit and sat/B otherwise

prusnak:
Follow-up to https://github.com/trezor/trezor-firmware/pull/2249


matejcik:
is it intentional that this is for _segwit-enabled_ coins (and not for segwit transactions, i.e., Bitcoin legacy transaction will still show sat/vB) ?

prusnak:
> is it intentional that this is for _segwit-enabled_ coins

Yes. For non-segwit transactions B = vB.

matejcik:
@marnova can you please take over? review the changes, regenerate fixtures etc.

marnova:
@matejcik ok

matejcik (approved):
utACK, didn't review every UI change but they look right

PR #2296, opened 2022-05-26
Record tests tweaks + upscale images for T1/TR

### LINKED ISSUE #2290: Make T1/TR screenshots larger in test reports
When generating the HTML report, it would be useful to 2x upscale the images coming from the small T1 / TR screens.

the following CSS works for me:
```css
img {
   width: 256px;
   image-rendering: pixelated;
}
```

marnova:
Also resolves #2290

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -165,7 +165,7 @@ def client(
 
 def pytest_sessionstart(session: pytest.Session) -> None:
     ui_tests.read_fixtures()
-    if session.config.getoption("ui") == "test":
+    if session.config.getoption("ui"):
         testreport.clear_dir()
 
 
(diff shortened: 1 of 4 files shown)


prusnak:
Maybe we can use `transform: scale(2);` as style instead of `width` and make the upscale argument configurable?

marnova:
@prusnak with `transform: scale(2);` I had a problem with the pictures (they were overlaying) and after some playing with it they were fine, but the table wouldn't adjust its size and I couldn't find a way how to set it up right. So 0771e8c3e94c36fcda947f2de23551880417f00d is a compromise.

matejcik (approved):

matejcik:
please squash the fixup and merge

PR #2301, opened 2022-05-26
ci: fix emulators deploy

vdovhanych:
This should fix emulators deploy after the version is tagged.

PR #2302, opened 2022-05-26
fix(tests): skip arm binaries during upgrade tests

mmilata:
Possibly related to #2301


mmilata:
Force-pushed version that drops the arm binaries during download instead of filtering names in test code.

vdovhanych (approved):
Nice

matejcik:
CI still failing?
tbh filtering out directories in the test seems like something we should _also_ do

mmilata:
Aaaaargh ... pushed one that at least passes on localhost. Wget flattens everything into single directory so there should be no subdirectories to filter out.

PR #2307, opened 2022-05-27
Basic animations for Model R

TychoVrahe:
Implements basic animations for Model R using a second framebuffer
- slide left
- slide right
- fast redraw of whole screen (this one would not need the second framebuffer)

Fast redraw enables to get rid of flickering when redrawing large part of the screen (could be useful for #2233 )

Slide effects are meant as replacement of fadeout-in effects which are not perfect on model R.

Also implemented faster initial display clear on model R as i was observing some partial screens after reset.

Noop for model 1 and T.

### DIFF
--- a/core/.changelog.d/2307.added
+++ b/core/.changelog.d/2307.added
@@ -1,0 +1,1 @@
+Added basic animations for Model R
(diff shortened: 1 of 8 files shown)


TychoVrahe:
rebased

TychoVrahe:
not needed anymore

PR #2308, opened 2022-05-27
Type hints fixes for older python versions

### LINKED ISSUE #2305: Broken build with python 3.8
Latest changes (#2286)  broke build with python 3.8 and older due to some type annotations.

As python 3.8 is a default in some widely used OSs ( Ubuntu 20.04.4 LTS), it would make sense to support this version.

Nice to have would be to also test with the minimal python version in CI pipelines

grdddj:
Fixes #2305:
- modifying type hints so that even older (`3.7` and `3.8`) python versions understand it
- also the test for this was created and added into `CI`
  - it is currently in the `Test` stage, but could be probably moved to `Prebuild` thanks to its nature (it is not dependent on `Build` and it is very similar to `style prebuild`)

In [this CI](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/549800707) the tests should fail, as it only has the first commit with tests - it is "fixed" in [subsequent CI](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/549805580), so the difference can be seen.

### DIFF
--- a/tests/test_python_support.py
+++ b/tests/test_python_support.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+"""
+Verifying that all the tools can be run even by older python versions.
+
+Uses `pyright --pythonversion 3.X <path>` output to check for substrings that
+indicate the type-hints in the code are not compatible with this version.
+"""
+
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+HERE = Path(__file__).resolve().parent
+ROOT_DIR = HERE.parent
+
+EXIT_CODE = 0
+
+os.chdir(ROOT_DIR)
+
+versions_to_check = [
+    "3.7",
+    "3.8",
+    "3.9",
+]
+
+dirs_to_check = [
+    "tools",
+    "common",
+    "core/tools",
+]
+
+signs_of_issues = [
+    "is unknown import symbol",  # we need to import some stuff from typing_extensions instead of typing
+    "will generate runtime exception",  # happens when using `dict` or `list` as a type alias
+]
+
+
+def check_directory(path: str, python_version: str) -> None:
+    global EXIT_CODE
+    cmd = (
+        "pyright",
+        "--pythonversion",
+        python_version,
+        path,
+    )
+
+    result = subprocess.run(cmd, stdout=subprocess.PIPE, text=True)
+    for line in result.stdout.splitlines():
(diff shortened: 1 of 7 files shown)


andrewkozlik:
Add `from __future__ import annotations` to `tools/generate_ci_docs.py`.

tools/pyright_tool.py
```diff
@@ -46,18 +46,24 @@
...
+from typing_extensions import (  # for python37 support, is not present in typing there
+    Final,
+    TypedDict,
+)
 
 import click
 
 if TYPE_CHECKING:
-    LineIgnores = list["LineIgnore"]
+    LineIgnores = List["LineIgnore"]
```
matejcik, line 60: type aliases hidden under `if TYPE_CHECKING` don't execute, so this should not be a problem?

tests/test_python_support.py
```diff
@@ -0,0 +1,60 @@
...
+]
+
+dirs_to_check = [
+    "tools",
+    "common",
+    "core/tools",
+]
+
+signs_of_issues = [
+    "Module 'typing' has no attribute",  # we need to import some stuff from typing_extensions
```
matejcik, line 34: can't we get the same info from pyright, if the python version is set appropriately?

grdddj:
> Add `from __future__ import annotations` to `tools/generate_ci_docs.py`.

Added in [ae9ca0d](https://github.com/trezor/trezor-firmware/pull/2308/commits/ae9ca0d0eb7ff964fce21fdf9cc6bdc8a9dbadaa), thanks for the trial. It was also discovered by `pyright`, even though `mypy` did not see it

tests/test_python_support.py
grdddj: It was indeed possible, changed in [29a778a](https://github.com/trezor/trezor-firmware/pull/2308/commits/29a778aca0103080a606ca39df559569ef408f10). It even correctly identified one more thing mentioned above. The only drawback is that it is slower than `mypy`, taking 30 seconds on my machine

tools/pyright_tool.py
grdddj: It should not be during runtime, but for the sake of the `pyright` tests, we need to also change it here, otherwise, it would complain about violation here

matejcik (approved):

matejcik:
very nice, please squash fixups & merge

PR #2312, opened 2022-05-31
Expose BIP-39 wordlist array from trezor-crypto to Rust code

matejcik:
For Trezor R, we need to be able to refine subsets of the wordlist, which is difficult to do without having the wordlist. We _could_ repeatedly call `mnemonic_get_word(i)` but that seems pointless.

Is there any particular reason the wordlist was not exposed in the original code?

tagging @andrewkozlik or @onvej-sl for review of the crypto-side changes
@mmilata and @grdddj for review of the Rust part

### DIFF
--- a/crypto/Makefile
+++ b/crypto/Makefile
@@ -1,5 +1,5 @@
 # CLANG_VERSION is empty if the compiler is not clang-based
-CLANG_VERSION = $(shell $(CC) --version | grep -Po "clang version \K[\d\.]+")
+CLANG_VERSION = $(shell $(CC) --version | sed -nr 's/^.*clang version ([0-9.]+).*$$/\1/p')
 CLANG_VERSION_MAJOR = $(shell echo $(CLANG_VERSION) | cut -f1 -d.)
 
 # determine specific version ranges
--- a/crypto/Makefile
+++ b/crypto/Makefile
@@ -98,7 +98,7 @@ ifdef SMALL
 CFLAGS += -DUSE_PRECOMPUTED_CP=0
 endif
 
-SRCS   = bignum.c ecdsa.c curves.c secp256k1.c nist256p1.c rand.c hmac.c bip32.c bip39.c pbkdf2.c base58.c base32.c
+SRCS   = bignum.c ecdsa.c curves.c secp256k1.c nist256p1.c rand.c hmac.c bip32.c bip39.c bip39_english.c pbkdf2.c base58.c base32.c
 SRCS  += address.c
 SRCS  += script.c
 SRCS  += ripemd160.c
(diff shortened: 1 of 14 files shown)


matejcik:
rebased to squash the fixups and get rid of the upgrade test failure

crypto/bip39_english.h
```diff
@@ -21,7 +21,9 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
```
andrewkozlik, line 22: `crypto/bip39_english.h` should be changed to `crypto/bip39_english.c` and added to the build. The `#include bip39_english.h` in `bip39.c` should then be removed.

andrewkozlik: Or you can create a new `bip39_english.h` file and move the `extern` declaration into that file.

crypto/Makefile
```diff
@@ -1,5 +1,5 @@
 # CLANG_VERSION is empty if the compiler is not clang-based
-CLANG_VERSION = $(shell $(CC) --version | grep -Po "clang version \K[\d\.]+")
+CLANG_VERSION = $(shell $(CC) --version | sed -nr 's/^.*clang version ([0-9.]+).*$$/\1/p')
```
andrewkozlik, line 2: I am unfamiliar with the `$$`, what does it do?

crypto/bip39_english.h
matejcik: done in [d97d3d3](https://github.com/trezor/trezor-firmware/pull/2312/commits/d97d3d3b4c7b736a308354eeeb3d05ca2f7e50df) and [eac8923](https://github.com/trezor/trezor-firmware/pull/2312/commits/eac89230d0b553fcbe1cc589ee333b34a1bc70f6)

andrewkozlik (approved):
Changes to `crypto` LGTM.

One final suggestion is I'd prefer to rename `BIP39_WORDS` -> `BIP39_WORD_COUNT`. This PR might be a good opportunity to change the naming before it proliferates.

matejcik:
sure, no problem

grdddj:
Locally I am experiencing the same Rust unit test issues as in CI - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2527954576 - however, I am able to build the `model R`, so it compiles well.

The `Wordlist` struct looks good, I will try to use it together with the seed recovery and see if it solves the issues on a physical device.

I just wonder if we need to always specify `iter().collect::<Vec<_>>()` when accessing the vector of words - cannot we have maybe `to_vec()` method that would do this internally (so that clients do not have to care about this complexity)?

sime:
@grdddj Just managed to successfully recover a 12 work seed.

mmilata:
AddressSanitizer claims there's buffer overflow when running the unit tests: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2565768011

matejcik:
Good catch, fortunately it appears to affect only the tests. I forgot to add 0-terminators. hopefully #2333 fixes this

matejcik:
wondering why the ASAN test didn't catch this when the PR was made....?

PR #2315, opened 2022-06-01
RGB LED driver for model R

### LINKED ISSUE #2300: RGB LED on model R
Implement RGB LED driver for Model R

TychoVrahe:
This implements RGB LED driver for model R, makes LED initialization in bootloader and firmware and exposes relevant functions to rust. 

Added small SRAM section to bootloader in order to be able to use DMA. Had to also add compiler flag to distinguish boardloader/bootloader/firmware to place DMA buffer correct section.

Fixes #2300

### DIFF
--- a/core/embed/bootloader/messages.c
+++ b/core/embed/bootloader/messages.c
@@ -358,8 +358,10 @@ void process_msg_FirmwareErase(uint8_t iface_num, uint32_t msg_size,
 }
 
 static uint32_t chunk_size = 0;
-// SRAM is unused, so we can use it for chunk buffer
-uint8_t *const chunk_buffer = (uint8_t *const)0x20000000;
+
+__attribute__((section(".buf"))) uint32_t chunk_buffer[IMAGE_CHUNK_SIZE / 4];
+
+#define CHUNK_BUFFER_PTR ((const uint8_t *const)&chunk_buffer)
 
 /* we don't use secbool/sectrue/secfalse here as it is a nanopb api */
 static bool _read_payload(pb_istream_t *stream, const pb_field_t *field,
@@ -375,7 +377,7 @@ static bool _read_payload(pb_istream_t *stream, const pb_field_t *field,
 
   if (offset == 0) {
     // clear chunk buffer
-    memset(chunk_buffer, 0xFF, IMAGE_CHUNK_SIZE);
+    memset((uint8_t *)&chunk_buffer, 0xFF, IMAGE_CHUNK_SIZE);
   }
 
   uint32_t chunk_written = offset;
@@ -390,7 +392,7 @@ static bool _read_payload(pb_istream_t *stream, const pb_field_t *field,
     }
     // read data
     if (!pb_read(
-            stream, (pb_byte_t *)(chunk_buffer + chunk_written),
+            stream, (pb_byte_t *)(CHUNK_BUFFER_PTR + chunk_written),
             (stream->bytes_left > BUFSIZE) ? BUFSIZE : stream->bytes_left)) {
       chunk_size = 0;
       return false;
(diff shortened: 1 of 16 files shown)


core/embed/trezorhal/rgb_led.c
```diff
@@ -0,0 +1,130 @@
...
+#if defined BOARDLOADER
+#error Not implemented for boardloader!
+#endif
+
+#if defined BOOTLOADER
+__attribute__((section(".buf")))
+#endif
+uint32_t rgb_led_data[RESET_DATA_LEN + DATA_LEN] = {0};
+
+static void rgb_led_set(uint32_t* start, uint8_t color) {
```
prusnak, line 21: What kind of protocol is this?

Maybe we can add some comments explaining what this function does?

Are we setting PWM delays with DMA? Why do constants `52` and `125` have ns equivalent? 

I have trouble understanding what exactly is going on. So maybe having a short explanation about the high-level architecture would be nice.

TychoVrahe: In the datasheet they call it "unipolar zeroing code" and basically they prescribe short pulse (200-400ns) for 0 and long pulse (580ns-1us) for 1. Reset is required before sending data: >80us without pulse. Then the data is sent -> RGB encoded as long and short pulses, so 24 bits (24 values of compare levels for PWM), and final value is zero to stop pulsing.  

PWM compare levels are set via DMA to avoid need for interrupt and manual change after every bit is sent.

I will add some explanation to the source also.

TychoVrahe: done 4f6c9f55d9168d4f72e7cfac7f302776fa4b4f3c

TychoVrahe:
Didn't know that SRAM was actually used in bootloader before for USB communication buffer and me adding the buffer for LED in bootloader broke firmware upload. Pushed a fix, i think its a bit cleaner this way but it changes important stuff a bit so please take good care during the review. @hiviah

prusnak:
> for USB communication buffer

It's not exactly USB communication buffer, we store the whole image chunk there - this is 128 KiB of data. I don't think this fact changes anything technically and the PR looks fine IMO, but it's important to understand the buffer is not just several hundreds bytes long as one could imagine when thinking about USB communication buffer.

core/embed/bootloader/messages.c
```diff
@@ -358,8 +358,10 @@ void process_msg_FirmwareErase(uint8_t iface_num, uint32_t msg_size,
 }
 
 static uint32_t chunk_size = 0;
-// SRAM is unused, so we can use it for chunk buffer
-uint8_t *const chunk_buffer = (uint8_t *const)0x20000000;
+
+__attribute__((section(".buf"))) uint32_t chunk_buffer[IMAGE_CHUNK_SIZE / 4];
+
+#define CHUNK_BUFFER_PTR (const uint8_t *const) & chunk_buffer
```
prusnak, line 364: Let's add brackets to be 100% sure this won't explode badly when used in some funky context.

```suggestion
#define CHUNK_BUFFER_PTR ((const uint8_t *const) & chunk_buffer)
```

TychoVrahe: fixed 8b066418a63c9c7989052501a17975e85c7e28d4

hiviah:
It seems there needs to be a tiny delay between after/at the end of `rgb_led_init`, otherwise if you put `rgb_led_set_color` right after it, it doesn't work right. The delay required is tiny enough that the result depends on compiler version.

hiviah (approved):

matejcik:
@TychoVrahe please rebase, fix conflicts, squash fixups and merge

Hannsek:
Cannot we delete this from the bootloader? Or can we use it for T3W1? @TychoVrahe

TychoVrahe:
its inconsequential. we will see what type of led if any will be used in T3W1 and what we can reuse there

Issue #2297, opened 2022-05-26
Rust UI: raw pixel access


matejcik:
We should be able to code equivalents of `display_text`, `display_loader` etc. in Rust, to avoid the need to write more C code.

This requires access to `PIXELDATA` macro/function from Rust.

hynek-jina:
No QA needed

PR #2317, opened 2022-06-01
Expose raw pixel access to rust

TychoVrahe:
Exposing to Rust functions needed for raw pixel access: pixeldata, pixeldata_dirty and set_window

Solves #2297

### DIFF
--- a/core/embed/extmod/modtrezorui/display-stm32_T.h
+++ b/core/embed/extmod/modtrezorui/display-stm32_T.h
@@ -30,11 +30,14 @@ const volatile uint8_t DISPLAY_ST7789V_INVERT_COLORS = 0;
 #define DISPLAY_MEMORY_BASE 0x60000000
 #define DISPLAY_MEMORY_PIN 16
 
-#define CMD(X) (*((__IO uint8_t *)((uint32_t)(DISPLAY_MEMORY_BASE))) = (X))
-#define ADDR                                           \
-  (*((__IO uint8_t *)((uint32_t)(DISPLAY_MEMORY_BASE | \
-                                 (1 << DISPLAY_MEMORY_PIN)))))
-#define DATA(X) (ADDR) = (X)
+__IO uint8_t *const DISPLAY_CMD_ADDRESS =
+    (__IO uint8_t *const)((uint32_t)DISPLAY_MEMORY_BASE);
+__IO uint8_t *const DISPLAY_DATA_ADDRESS =
+    (__IO uint8_t *const)((uint32_t)DISPLAY_MEMORY_BASE |
+                          (1 << DISPLAY_MEMORY_PIN));
+
+#define CMD(X) (*DISPLAY_CMD_ADDRESS = (X))
+#define DATA(X) (*DISPLAY_DATA_ADDRESS = (X))
 #define PIXELDATA(X) \
   DATA((X) >> 8);    \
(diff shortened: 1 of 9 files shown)


core/embed/extmod/modtrezorui/display-stm32_T.h
```diff
@@ -56,6 +56,8 @@ const volatile uint8_t DISPLAY_ST7789V_INVERT_COLORS = 0;
 // of ILI9341V datasheet
 #define DISPLAY_ID_ILI9341V 0x009341U
 
+void display_pixeldata(uint16_t c) { PIXELDATA(c); }
```
matejcik, line 59: wondering if we could `inline` this function (and then `#[inline(always)]` Rust's `pixeldata()`

core/embed/rust/src/ui/display.rs
```diff
@@ -183,6 +183,18 @@ pub fn text_right(baseline: Point, text: &str, font: Font, fg_color: Color, bg_c
...
 
+pub fn pixeldata(color: Color) {
+    display::pixeldata(color.into());
+}
+
+pub fn pixeldata_dirty() {
+    display::pixeldata_dirty();
+}
+
+pub fn set_window(x0: u16, y0: u16, x1: u16, y1: u16) {
```
matejcik, line 194: this should accept `Rect` argument

core/embed/extmod/modtrezorui/display-stm32_T.h
matejcik: I see, inlining across languages is problematic. There is something called Linker plugin LTO for Rust, but it's unclear if it even works with gcc.

But the function call overhead might hurt here. Wondering how best to expose the underlying thing to Rust.

Could we redefine `ADDR` to be something like `const _IO uint8_t* PIXELDATA_ADDR = (_IO uint8_t*)((uint32_t)...)`, then use the `PIXELDATA_ADDR` pointer manually?

matejcik:
```rust
#[cfg(feature = "model_tt")]
#[inline(always)]
pub fn pixeldata(c: u16) {
    unsafe {
        PIXELDATA_ADDR.write_volatile((c >> 8) as u8);
        PIXELDATA_ADDR.write_volatile((c & 0xff) as u8);
    }
}

#[cfg(not(feature = "model_tt"))]
#[inline(always)]
pub fn pixeldata(c: u16) {
    unsafe {
        display_pixeldata(c);
    }
}
```

matejcik:
[53fd148](https://github.com/trezor/trezor-firmware/pull/2317/commits/53fd148caf088591faa23a8f07f56c56265d3c5d) is a modified cherry-pick of #2322 -- which should be merged first and then this PR rebased on top of that

core/embed/extmod/modtrezorui/display-stm32_T.h
matejcik: [aea7578](https://github.com/trezor/trezor-firmware/pull/2317/commits/aea7578520e9bf2dc6fc6d923785fa0e0f3264d2) does the above

matejcik:
removed the "emulator" feature, we can use `target_arch = "arm"` in its place (some pieces of code already do that), and the logic of the check is nicer

PR #2322, opened 2022-06-06
feat(core/rust): add trezorhal bindings

matejcik:
Working off @jpochyla's commit from #2171, this replaces all manual ffi blocks with bindgen.

Also adds support for reading `TREZOR_MODEL` envvar in cargo.


matejcik:
bumped bindgen again because it fixes https://github.com/rust-lang/rust-bindgen/issues/2186

matejcik:
instead of emulator feature, we can use `target_arch = arm`

Issue #2208, opened 2022-04-05
Expose model information and metadata from boardloader


matejcik:
In order to support the same bootloader and/or firmware image on multiple Trezor models, we need some info exposed about the current hardware.

We definitely need to see **model identifier** (single character _might_ be sufficient).
HW revision and boardloader version could be useful. A nice-to-have would be a capability structure, like:
* does boardloader support reboot-to-bootloader?
* do we have SD card?
* is the SD card hot-swappable?

We don't currently have an exact list of things to put there, besides the model identifier. At this point it is mostly a research topic, i.e., how to expose this data, how big it can be, etc.

prusnak:
Check `core/embed/prodtest/main.c` lines 384-391. 

That's where we set the id to OTP in production (only model and batch initial date is stored). There has been only 1 boardloader version deployed, so all devices have the same bootloader IIRC.

hiviah:
OTP seems readable before dropping privileged mode and setting MPU, however we will still need to put some data in boardloader to put the capabilities in. So the OTP could be used as fallback in case we don't have the new boardloader with the structure.

matejcik:
closed via #2324

PR #2324, opened 2022-06-06
Boardloader capabilities structure

hiviah:
Implements #2208.

At the end of boardloader, 256 bytes are reserved for "super simple" version of ASN.1 structure, that is extendible with arbitrary data. For now, model name and boardloader version are stored there.

It is supposed to be both forward and backward compatible since boardloader is read-only. Adding a tag is possible, but using old firmware that doesn't understand the tag will be as well (tag will be skipped).


core/embed/boardloader/memory.ld
```diff
@@ -48,6 +48,9 @@ SECTIONS {
     . = ALIGN(4); /* make the section size a multiple of the word size */
   } >CCMRAM
 
+  /* Hard-coded address for cabalities structure */
```
prusnak, line 51: Typo:

```suggestion
  /* Hard-coded address for capabilities structure */
```

core/embed/trezorhal/board_capabilities.c
```diff
@@ -0,0 +1,81 @@
...
+
+const uint8_t *get_board_name() { return board_name; }
+
+const struct BoardloaderVersion *get_boardloader_version() {
+  return &boardloader_version;
+}
+
+void parse_boardloader_capabilities() {
+  uint8_t *pos = (uint8_t *)BOARD_CAPABILITIES_ADDR;
+  uint8_t *end = (uint8_t *)(BOARD_CAPABILITIES_ADDR + BOARD_CAPABILITIES_SIZE);
```
prusnak, line 39:
```suggestion
  const uint8_t *pos = (const uint8_t *)BOARD_CAPABILITIES_ADDR;
  const uint8_t *end = (const uint8_t *)(BOARD_CAPABILITIES_ADDR + BOARD_CAPABILITIES_SIZE);
```

core/embed/trezorhal/board_capabilities.c
```diff
@@ -0,0 +1,81 @@
...
+
+  if (memcmp(pos, CAPABILITIES_HEADER, 4) != 0) {
+    return;
+  }
+
+  pos += 4;
+
+  // -2 for possible tag without any data
+  while (pos <= end - 2) {
```
prusnak, line 48: Is the `-2` really necessary?
It should work correctly with the following, no?
And we could kill the terminator tag completely.

```suggestion
  while (pos < end) {
```

prusnak (changes requested):

core/embed/trezorhal/board_capabilities.c
hiviah: With `(pos < end)` you could read out-of-bound one byte behind the sector if `(pos == end-1)`.

Though we will use well-formed structure I thought then it's better be able to deal with malformed input to avoid any bugs.

The terminator is needed since the other two options are:

 * make and "empty tag" to have length until the end, which is kind of the same
 * rely that empty flash bytes are 0xFF behind the structure, so that it would be coincidentally seen as unknown tag of length 255 and cycle will end

It seems cleaner to have the terminator than the above two other options.

prusnak: Okay 👍

prusnak:
Pushed two more fixups, the first one eeb4b62 is obvious, the second one 35da445 inverses the logic when not to call `parse_boardloader_capabilities` - that way we don't forget to add new models. Are you OK with these @hiviah ?

hiviah:
OK with the changes. I'll squash it and finish it.

PR #2332, opened 2022-06-08
Make Rust buildable without Micropython

matejcik:
Features `micropython` and `protobuf` are defined. Protobuf implies micropython
because our protobuf impl is pretty much _for_ micropython.

The respective subdirs are included only if the matching feature is defined.

`util.rs` is moved to micropython because it mostly concerns micropython interop

`ResultExt`, useful only for ui_debug, is moved to `ui::util`.

micropython/time.rs is moved to trezorhal/time.rs. The functions
`mp_hal_ticks_ms` and `mp_hal_delay_ms` are still used, but they are not
specific to micropython; in fact, for firmware, they are defined in `delay.c`

`ui::*::layout` is conditional for micropython feature, because it only concerns
micropython layouts. If we want to reuse layouts defined there, we will need to
export them to not depend on Objs and Qstrs etc.


matejcik:
update: `micropython::time` is kept, and `trezorhal::time` is added.

`trezorhal` either re-exports the micropython functions (which have some niceties built-in), or uses raw `HAL_Delay` / `HAL_GetTick` functions from stmlib (or a corresponding implementation on Unix).

This seems to work out nicely.

PR #2333, opened 2022-06-09
Fix overflow in bip39.rs tests

matejcik:
by using `cstr!` to generate 0-terminated strings

### DIFF
--- a/ci/test.yml
+++ b/ci/test.yml
@@ -131,6 +131,8 @@ core device asan test:
     - core unix frozen debug asan build
   only:
     - schedules  # nightly build
+  variables:
+    PYTEST_TIMEOUT: "600"
   script:
     - nix-shell --run "poetry run make -C core test_emu | ts -s"
   artifacts:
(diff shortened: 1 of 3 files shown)


matejcik:
wondering why the ASAN test didn't catch it when this PR was made....?

matejcik:
`test_lots_of_inputs` fails consistently on the ASAN build.
is this something we saw before?

matejcik:
@mmilata ^^

mmilata:
Seems to be happening more often than not in recent nightly runs. It's exceeding the `PYTEST_TIMEOUT` limit - we can raise it to more than 5m, though looking at the [pipelines](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines?source=schedule) it seems that ~month ago around ( ~ bdfc4532) it started to happen more often.

matejcik:
> ~month ago

probably the switch to external runners

i'll increase the time limit in this PR

PR #2336, opened 2022-06-14
common/defs/fido: add cloudflare

mcudev:
this makes the fido security key functionality on the model t for dash.cloudflare.com more complete by showing the cloudflare logo.

if it helps for testing:
```
navigator.credentials.create({
  publicKey: {
    "rp": {
      "name": "Cloudflare Dashboard",
      "id": "dash.cloudflare.com"
    },
    user: {
      id: new Uint8Array([1]),
      name: "name",
      displayName: "display name"
    },
    challenge: new Uint8Array([1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]),
    "pubKeyCredParams": [{"alg": -7, "type": "public-key"}, {"alg": -257, "type": "public-key"}],
    "authenticatorSelection": {"requireResidentKey": false, "userVerification": "preferred"},
    "attestation": "none"
  }
});
```

PR #2342, opened 2022-06-17
Fix bitcoin only pipeline check

TychoVrahe:
#2336 added cloudflare, since there is a coin named 'flare', this causes bitcoin only check to fail. added 'cloudflare' to exceptions.


mmilata (approved):
LGTM, thanks!

PR #2344, opened 2022-06-20
Bump nixpkgs version, fix patching of emulator binaries

### LINKED ISSUE #2195: Fix patching of library paths of old emulator binaries
See #2194 and https://github.com/NixOS/nixpkgs/pull/149731. As autoPatchelfHook exists for different usecase maybe consider different approach like compiling the emulators with statically linked SDL.

mmilata:
Fixes #2195. The new autoPatchelf implementation doesn't seem to work with file arguments so we need to pass a directory to it. Protobuf 3.6 -> 3.19.


ci/shell.nix
```diff
@@ -88,7 +88,7 @@ stdenvNoCC.mkDerivation ({
     openssl
     pkgconfig
     poetry
-    protobuf3_6
+    protobuf3_19
```
matejcik, line 91: is protobuf always versioned? couldn't we just use `protobuf` or something?

mmilata: We can use `protobuf` which is the latest packaged version (3.19). Don't know if there was a reason to use explicit version in the first place.

PR #2347, opened 2022-06-22
ci: only use .venv from gitlab cache if commit matches

mmilata:
Nixpkgs bump #2344 caused poetry to fail when old .venv is present: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2625709443. Currently in GitLab CI we cache .venv across builds on the same branch so master branch keeps failing.

The PR changes the cache key so that the cache is only shared by jobs with the same git commit. This makes the cache less effective but improves isolation of builds on long-lived branches.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -306,6 +306,7 @@ core unix frozen debug build arm:
 
 core unix frozen btconly debug t1 build:
   stage: build
+  <<: *gitlab_caching
   needs: []
   variables:
     BITCOIN_ONLY: "1"
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -70,7 +70,6 @@ stdenvNoCC.mkDerivation ({
   ] ++ [
     SDL2
     SDL2_image
-    autoflake
     bash
     check
     curl  # for connect tests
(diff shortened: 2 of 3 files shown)


vdovhanych:
I'm not sure if in this case, the cache would even present any benefits as it would need to be created and downloaded per job. What about using the `clear runner cache button` when it starts to fail or after there were some bigger changes?

mmilata:
Can't see the button, can you hit it for me and trigger pipeline on master?

vdovhanych:
I cleared the cache and now jobs are green again https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/569907206

matejcik:
we could do `key: "$CI_COMMIT_REF_SLUG-$VENV_CACHE_VERSION"` for all jobs, keep `$VENV_CACHE_VERSION` set to 1, in `.gitlab-ci.yml`, and bump it whenever we do something like this.
that's essentially what the "clear caches" button is doing under the hood, but this way it is also explicitly controllable for us through git history.

for that matter would it be possible to inherit `.gitlab_caching` from the global `.gitlab-ci.yml` file?

matejcik:
fwiw, why isn't the cache configured globally for all jobs?

vdovhanych:
I think I've done it that way so we could easily add different paths for separate jobs. Btw should we do with this pr, do we need it?

matejcik:
i don't think so, @mmilata ?

mmilata:
Minor driveby fixes - one of the jobs did not use cache & shell.nix installed autoflake even though we are using the version from poetry. OK to merge?

vdovhanych (approved):
lgtm

matejcik (approved):

Issue #2338, opened 2022-06-16
potential ed25519 unsafe signing


kchalkias:
A recent security report revealed a potential private key leak issue when the `ed25519.sign` api takes a non-matching public key as input, see [MystenLabs/ed25519-unsafe-libs](https://github.com/MystenLabs/ed25519-unsafe-libs)

TL;DR Most projects with a dependency to donna library tend to allow this public api for their devs and an accidental misuse (i.e. provide non matching pub-priv keys as signing function inputs) can be exploited resulting to private key leak.

Trezor's implementation relies on the ed25519-donna api, which can be accidentally misused by Trezor devs, see here: https://github.com/trezor/trezor-firmware/blob/master/crypto/ed25519-donna/ed25519.c#L110

It is highly recommended that the sign api should not take the public key as input and:

-  it should be computed internally to the signing function (as the ed25519 RFC 8032 recommends) 
or 
- refactor your private key structure to include the public key bytes too - many projects take this approach with the aim to avoid recomputing the pubKey on each sig. request, i.e., similarly to https://github.com/penumbra-zone/ed25519-consensus/blob/main/src/signing_key.rs#L16

prusnak:
There are only 2 places where we call `ed25519_sign`:

1. https://github.com/trezor/trezor-firmware/blob/7d37109eb84ef029e7b7ef3eac55db3237af03f6/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-ed25519.h#L97-L98
2. https://github.com/trezor/trezor-firmware/blob/7d37109eb84ef029e7b7ef3eac55db3237af03f6/legacy/firmware/stellar.c#L1371

Case 1 calls `ed25519_publickey` to compute the public key from the secret key just beforehands
Case 2 uses the structure `node` which stores the corresponding public key to the private key.

So far I can tell the Trezor code is not affected.

kchalkias:
This is true and already cross checked that, there is no issue with current use, but the problem is by exposing this api there might a future one

andrewkozlik:
[...]

PR #2349, opened 2022-06-23
Remove public key from the ed25519 signing API

andrewkozlik:
Resolves https://github.com/trezor/trezor-firmware/issues/2338.

### DIFF
--- a/crypto/ed25519-donna/ed25519.h
+++ b/crypto/ed25519-donna/ed25519.h
@@ -21,9 +21,9 @@ void ed25519_publickey_ext(const ed25519_secret_key sk, const ed25519_secret_key
 #endif
 
 int ed25519_sign_open(const unsigned char *m, size_t mlen, const ed25519_public_key pk, const ed25519_signature RS);
-void ed25519_sign(const unsigned char *m, size_t mlen, const ed25519_secret_key sk, const ed25519_public_key pk, ed25519_signature RS);
+void ed25519_sign(const unsigned char *m, size_t mlen, const ed25519_secret_key sk, ed25519_signature RS);
 #if USE_CARDANO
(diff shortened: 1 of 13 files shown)


crypto/ed25519-donna/ed25519.h
```diff
@@ -21,9 +21,9 @@ void ed25519_publickey_ext(const ed25519_secret_key sk, const ed25519_secret_key
 #endif
 
 int ed25519_sign_open(const unsigned char *m, size_t mlen, const ed25519_public_key pk, const ed25519_signature RS);
```
kchalkias, line 23: Thanks for the refactoring, shall we also update this one?

andrewkozlik: This is the signature verification function, so I think there is nothing to update here. The naming is confusing, I know.

kchalkias: Oh I see, what about `ed25519_cosi_sign`?

andrewkozlik: CoSi is Cothority Collective Signing, it's an aggregate multisig scheme and the signing function needs the _collective public key_ `pk`, see https://github.com/dedis/doc/blob/master/cosi/cosi.md#signature-generation. So in this case having `pk` in the API is justified.

kchalkias (approved):
Thanks, great work and blazing fast response!

crypto/ed25519-donna/ed25519.h
kchalkias: Confirmed that in cosi signature scheme, we expect an extra randomness and the R value is not deterministically generated by M and sk as in the original ed25519. 

That said, assuming that the sign function will be invoked with a fresh tuple of <random nonce, pk> each time we invoke it, then the current implementation is safe. Can you confirm that if we ever retry cosi sign for the same message, the nonce is always a fresh random number?

prusnak (approved):
LGTM, thanks Andrew

PR #2352, opened 2022-06-29
Update trezorlib to 0.13.1

matejcik:
fixes several long-standing issues


python/requirements.txt
```diff
@@ -1,7 +1,7 @@
 ecdsa>=0.9
 mnemonic>=0.20
 requests>=2.4.0
-click>=7,<8.1
+click>=7,<8.2
```
prusnak, line 4: Maybe we don't want to be strict here, or do we expect that click will break the API again in 8.2?

```suggestion
click>=7
```

matejcik: > do we expect that click will break the API again in 8.2?

given their history, absolutely yes 👀

grdddj (approved):
I do not have any complaints about the changes. `Device is in use by another process` is really helpful. Looking forward to the release!

PR #2354, opened 2022-06-30
feat(cardano): add support for babbage features

gabrielKerekes:
[Babbage CDDL](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl)

Changes:
- new Babbage transaction elements serialization and  the corresponding logic and UI changes
  - `collateral_return` - an output which can be added to a transaction to contain unspent collateral. It would only be considered a real tx output if Plutus script validation fails and collateral would be spent. Useful so that any input can be used as collateral.
  - `total_collateral` - if the total ADA amount in collateral inputs doesn't match total collateral the transaction is invalid.
  - `reference_inputs` - inputs required for script validation.
  - output `inline_datum` - the whole serialized datum can now be included in the output instead of just the datum hash. Is sent as chunks since it can be quite long. Only first chunk is shown to the user.
  - output `reference_script` - used for script pre-publishing.  Is sent as chunks since it can be quite long. Only first chunk is shown to the user.
- "post-alonzo" output serialization as a map
  - Outputs can now be serialized in two formats. The legacy, array format or now also as a map. See [CDDL](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl#L75).
- `required_signers` allowed in ORDINARY and MULTISIG txs - there should be no harm in allowing this and some dapps misuse this field fir whatever they need.
- `datum_hash` allowed in non-script outputs - apparently this can be useful to "tag" inputs.


common/protob/messages-cardano.proto
```diff
@@ -210,6 +215,9 @@ message CardanoSignTxInit {
     optional bytes script_data_hash = 16;
     required uint32 collateral_inputs_count = 17;
     required uint32 required_signers_count = 18;
+    optional bool has_collateral_return = 19 [default=false];
```
gabrielKerekes, line 218: The `has_collateral_return`, `reference_inputs_count`, `CardanoTxOutput.format`, `CardanoTxOutput.inline_datum_size` and ``CardanoTxOutput.reference_script_size` fields should ideally be `required`.

But we're worried whether Connect will be released in time with the FW because of the new major Connect version and migration to the trezor-suite repo. So we rather made the fields optional with default values because otherwise the new FW would not work with old Connect since it would be expecting these fields.

What do you think about this please?

matejcik: Adding all new fields as optionals with defaults is correct IMHO, unless the field is actually required for some part of the logic. If the fields are functionally optional (say for older transaction type), AND if there is a default value that makes sense (such as `has_collateral_return = false` by default, similarly `*_count` added in the past can be implicitly 0), it's good that they are optional and older clients not using these features cleanly continue working.

That said, `total_collateral` is optional but has no default value. Would it make sense to set 0? Can a valid transaction be produced with 0 collateral?

matejcik:
> Outputs can now be serialized in two formats. The legacy, array format or now also as a map. See [CDDL](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl#L75).

are we required to support both?

common/protob/messages-cardano.proto
matejcik: or is `total_collateral` intended as optional?

gabrielKerekes: `total_collateral` is intended as optional so that it is left out of the transaction completely if it's not set i.e. 0 is probably not a valid default. I would expect e.g. ordinary transactions with `total_collateral` set to 0 to be invalid but I haven't verified this.

gabrielKerekes:
> are we required to support both?

The result of internal discussions was that yes, we need to support both, at least for the time being so that wallets and other apps would get a chance to migrate at least. Currently there is however no specific plan to drop the support from our side but it should be doable.

core/src/apps/cardano/sign_tx/signer.py
```diff
@@ -61,9 +66,22 @@
...
+BABBAGE_OUTPUT_KEY_AMOUNT = const(1)
+BABBAGE_OUTPUT_KEY_DATUM_OPTION = const(2)
+BABBAGE_OUTPUT_KEY_REFERENCE_SCRIPT = const(3)
+
+DATUM_OPTION_KEY_HASH = const(0)
+DATUM_OPTION_KEY_INLINE = const(1)
 
 POOL_REGISTRATION_CERTIFICATE_ITEMS_COUNT = 10
 
+MAX_CHUNK_SIZE = 256
```
matejcik, line 83: 1024 should be a safe value that is also used elsewhere (Ethereum data streaming, Bitcoin extra_data)

common/protob/messages-cardano.proto
```diff
@@ -252,6 +263,22 @@ message CardanoToken {
     optional sint64 mint_amount = 3;         // mint amount (can also be negative in which case the tokens are burnt)
 }
 
+/**
+ * Request: Transaction output inline datum chunk
+ * @next CardanoTxItemAck
+ */
+message CardanoTxInlineDatumChunk {
```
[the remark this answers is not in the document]
gabrielKerekes: Done in https://github.com/trezor/trezor-firmware/pull/2354/commits/7378e352f93ec465ab9b30690b312c96ea3185d2.

core/src/apps/cardano/sign_tx/signer.py
gabrielKerekes: Alright 👍 Updated in https://github.com/trezor/trezor-firmware/pull/2354/commits/2cb42cacb84efd27ae9bbe1b7caa53694dc0d252 and also the test case has been updated in https://github.com/trezor/trezor-firmware/pull/2354/commits/523417280374c4f4f729b2475b45fafc18a09175.

core/src/apps/cardano/sign_tx/signer.py
```diff
@@ -449,6 +609,62 @@ def _validate_token(
...
+                messages.CardanoTxItemAck(), messages.CardanoTxInlineDatumChunk
+            )
+            self._validate_chunk(
+                chunk.data,
+                chunk_number,
+                chunks_count,
+                wire.ProcessError("Invalid inline datum chunk"),
+            )
+            if chunk_number == 0 and should_show:
+                await layout.confirm_inline_datum(
```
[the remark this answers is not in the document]
gabrielKerekes: No, as it's impractical to verify the whole datum/reference script on the HW wallet if it's large and the users would probably not verify it anyways but would have to scroll and click a lot. The transaction id (hash) is displayed at the end of the flow so paranoid users can verify it.

core/src/apps/cardano/layout.py
```diff
@@ -256,6 +264,68 @@ async def confirm_sending_token(
...
+
+async def _confirm_data_chunk(
+    ctx: wire.Context, br_type: str, title: str, first_chunk: bytes, data_size: int
+) -> None:
+    MAX_DISPLAYED_SIZE = 56
+    displayed_bytes = first_chunk[:MAX_DISPLAYED_SIZE]
+    bytes_optional_plural = "byte" if data_size == 1 else "bytes"
+    props: list[tuple[str, bytes | None]] = [
+        (
+            f"{title} ({data_size} {bytes_optional_plural}):",
```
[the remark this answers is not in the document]
gabrielKerekes: If I am not mistaken the hexlified bytes are displayed? 🤔 
image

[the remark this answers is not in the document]
matejcik: right because you're passing a `bytes` type, forgot we have this feature. sorry about the noise .)

core/src/apps/cardano/sign_tx/signer.py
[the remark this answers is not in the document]
matejcik: that is exactly why we have the Details screen - see Ethereum usage. (you can scroll down as far as you like, and you can exit the scrollable at any time to go back to the short view)
but we don't need to change it in this PR, just something to consider for the future.

marnova (commented):
With my very limited knowledge about Cardano I have nothing to add that @matejcik didn't already mentioned.

matejcik (approved):
thanks @marnova 

@gabrielKerekes please squash the fixups so that this is ready to merge

gabrielKerekes:
> @gabrielKerekes please squash the fixups so that this is ready to merge

Done. I also squashed the `make Babbage protobuf elements optional` commit with the one adding types (the first one).

bosomt:
**QA OK**

`Device: model T 2.5.2 Universal (revision 0d87b55ba4fed7eecc72bf2a94ee473830b095e9)`
sent tx and test tx

https://adastat.net/transactions/735d5f748890d88e9ed60625899e2fc6d336f136a84164de5a9c5aedd6261e7a

https://explorer.cardano-testnet.iohkdev.io/en/transaction?id=8b86f06675ad581e1e44c83bf6eafb2f2a97f57433b973b2f55e80436a37d458

PR #2355, opened 2022-06-30
feat(cardano): add support for hiding some details during tx signing

gabrielKerekes:
**Note: Only the last three most recent commits are relevant since the PR is currently based on `cardano-refactors-2` although it's really based on the `cardano-babbage` changes, but a branch for that doesn't exist in the Trezor repository yet so I couldn't base the PR on top of that.**

Since transaction signing has become a UX pain with the introduction of Plutus, we're adding the option to show only a "Simple View" or "Show All" (details) during transaction signing (originally we called it "Expert mode"). The choice is given to the user at start of the transaction signing flow (see attached screenshot). If user chooses to "Show All" he will be shown all the transaction elements, otherwise only the most important ones will be shown. The idea is for ordinary users not to be bothered with items they wouldn't be able to verify anyways (inputs, hashes, datums, ...). [Here's a complete table](https://docs.google.com/spreadsheets/d/1fuikDJ6apUQ22oOoaKgSK86syp46BM_KzKFpTuziD4o/edit?usp=sharing) for what should be shown when.

We've also added tests which run with in the "Show All" mode. We didn't want to run the whole test suite for each mode, since this would double the number of tests (and there's already enough of them). So we copied relevant test cases from the other test files into `sign_tx.show_details.json` and created a separate test function for that. The tests should serve as UI tests. As with everything, if you come up with a better way to achieve the same results, we'll gladly modify the code but this seemed simplest to us.

image


core/src/apps/cardano/sign_tx/ordinary_signer.py
```diff
@@ -25,6 +25,8 @@ def __init__(
     ) -> None:
         super().__init__(ctx, msg, keychain)
 
+        self.signing_mode_title = "Confirming a transaction."
```
matejcik, line 28: maybe make this a classvar?

matejcik (approved):
one nit otherwise LGTM

core/src/apps/cardano/sign_tx/ordinary_signer.py
gabrielKerekes: Like this https://github.com/trezor/trezor-firmware/pull/2355/commits/c702e9db85293bc10b4f87bc1e01aca4d8e217d7? I didn't find a way to mark it abstract so that we could get rid of the assert 🤔

gabrielKerekes:
> CI is failing on UI fixtures, please regenerate them and add to the PR

UI fixtures and changelog added in https://github.com/trezor/trezor-firmware/pull/2355/commits/9fafbc86d61cea00ae4a0020865fea8c390a659a.

core/src/apps/cardano/sign_tx/ordinary_signer.py
matejcik: exactly.
I would expect `ClassVar[str]` without specifying the value would be valid?
(but maybe the typechecker would only allow that if the class was abstract via abcmeta, which we probably can't do in uPy)
but if not then let's keep it like this for now

gabrielKerekes: It is indeed valid. But the typechecker doesn't enforce the value to be set in the child signers so I'm not sure what the best approach here is.

I'd prefer keeping the type `ClassVar[str | None]` so that the assert in `_show_tx_init` is required and the app crashes "predictably" if `SIGNING_MODE_TITLE` is not set in any of the child classes. Also I would add a comment to the assert and perhaps also to the variable declaration in `Signer` if we decide to keep it.

matejcik: I filed a bug https://github.com/microsoft/pyright/issues/3652, but I'm not sure if it can be fixed in a practical manner.

I don't think that the assert is bringing any value except appeasing the type checker though. You are still asserting immediately before using the value, so the difference is only between getting an `AssertionError` in debug mode, or `AttributeError` in both debug and production. Using a semantically wrong type annotation does not help us do what we want, i.e., check this statically; and good test suite coverage can catch either option.

So please change to `ClassVar[str]` and remove the assert.

gabrielKerekes: Done in https://github.com/trezor/trezor-firmware/pull/2355/commits/3babd23acb7a4029e9e51b16fe734a194c1c8630.

gabrielKerekes:
> there are left-over(?) fixtures in the json, please delete them:

Sorry, deleted in https://github.com/trezor/trezor-firmware/pull/2355/commits/65ff6c7609dae690a8d665306dd73dacfb08ab8a.

matejcik (approved):
thank you. please squash & rebase on top of the previous PR, I will merge both via this one as soon as it passes CI

gabrielKerekes:
> please squash & rebase on top of the previous PR

Done 👍

> as soon as it passes CI

🤞

Issue #2340, opened 2022-06-16
Report availability of data connection


sime:
During the [debugging](https://www.notion.so/satoshilabs/32ca2fc003bf465a9b3c41ed7315a000?v=fabca552ea3a4dad919a93c94ccc0fcb&p=27d2299ec4584485a9f14cacd3f48656) of a Trezor that is not being recognised, users may connect a power only cable to the Trezor.

Is it technical feasible that the Trezor can distinguish if something is trying to 'talk' over the wire.

Potential use case: Trezor shows a 'no data connection' icon/label when a non-data friendly cable is connected. 

#### User Story
As a Trezor user
I want to know when I have connected a non data cable to my Trezor
So that I know my Trezor will not work with this cable

#### Accetpance criteria
Given I have a powerbank
When I connect my Trezor to it
Then I should see a 8x8 pixel blue (#0000FF) in the top left

Given I have power only USB data cable
And it is connected to a computer
When I connect my Trezor to it
Then I should see a 8x8 pixel blue (#0000FF) in the top left

Given I have a [USB-C hub](https://www.hypershop.com/collections/usb-c-hubs/products/hyperdrive-slim-usb-c-hub-for-macbook-chromebook-pc-usb-c-devices)
And it is connected to a power source
And no computer is connected to it
When I connect Trezor to the hub
Then I should see a 8x8 pixel blue (#0000FF) in the top left

#### Wireframe
![broken_cable](https://user-images.githubusercontent.com/216917/174846615-fdcb28fd-5a94-46dc-bc09-cc49c9c5fe9c.png)

matejcik:
it would also be nice to detect "dying" cables somehow, but I don't know how exactly they behave.
is it lost packets? unfinished data? ...etc

hiviah:
USB packets have CRC and retransmission (except isochronous mode for real-time like audio), so data won't seem corrupted I'd think.

Few times I've seen something was that device looked fine at first, but when you started more power intensive operation like firmware update, it got stuck

matejcik:
> CRC and retransmission

could we get access to, e.g., a retransmission counter?

sime:
Updated the issue description with acceptance criteria for proof of concept.

hiviah:
I looked over complete map of USB-related registers in SVD definition, but there doesn't seem to be any for counting retransmits.

sime:
@hiviah Could you point to a build where this can be tested?

cc: @Hannsek

[...]

PR #2366, opened 2022-07-04
Detect whether USB data pins are connected

hiviah:
Implementation of  #2340.

### DIFF
--- a/core/embed/unix/usb.c
+++ b/core/embed/unix/usb.c
@@ -27,6 +27,7 @@
 #include <time.h>
 #include <unistd.h>
 
+#include "profile.h"
 #include "touch.h"
 #include "usb.h"
 
@@ -252,7 +253,7 @@ void pendsv_kbd_intr(void) {}
 void mp_hal_set_vcp_iface(int iface_num) {}
 
 secbool usb_configured(void) {
-  if (access("/var/tmp/trezor.usb_data_disconnected", F_OK) == 0) {
+  if (access(profile_usb_disconnect_path(), F_OK) == 0) {
     return secfalse;
   }
 
--- a/core/embed/unix/usb.c
+++ b/core/embed/unix/usb.c
@@ -25,6 +25,7 @@
 #include <sys/poll.h>
 #include <sys/socket.h>
 #include <time.h>
+#include <unistd.h>
 
 #include "touch.h"
 #include "usb.h"
@@ -249,3 +250,11 @@ int usb_webusb_write(uint8_t iface_num, const uint8_t *buf, uint32_t len) {
 void pendsv_kbd_intr(void) {}
 
 void mp_hal_set_vcp_iface(int iface_num) {}
+
+secbool usb_configured(void) {
+  if (access("/var/tmp/trezor.usb_data_disconnected", F_OK) == 0) {
+    return secfalse;
+  }
+
+  return sectrue;
+}
--- a/core/src/apps/homescreen/homescreen.py
+++ b/core/src/apps/homescreen/homescreen.py
@@ -1,10 +1,11 @@
 import utime
 from micropython import const
+from typing import Tuple
 
 import storage
 import storage.cache
 import storage.device
-from trezor import config, ui, utils
+from trezor import config, io, loop, ui, utils
 from trezor.ui.loader import Loader, LoaderNeutral
 
 from apps.base import lock_device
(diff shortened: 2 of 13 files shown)


core/embed/unix/usb.c
```diff
@@ -249,3 +250,11 @@ int usb_webusb_write(uint8_t iface_num, const uint8_t *buf, uint32_t len) {
 void pendsv_kbd_intr(void) {}
 
 void mp_hal_set_vcp_iface(int iface_num) {}
+
+secbool usb_configured(void) {
+  if (access("/var/tmp/trezor.usb_data_disconnected", F_OK) == 0) {
```
matejcik, line 255: please add to `unix/profile.c` instead of hard-coding the name

core/src/apps/homescreen/homescreen.py
```diff
@@ -54,6 +67,9 @@ def do_render(self) -> None:
         # homescreen with shifted avatar and text on bottom
         # Differs for each model
 
+        if not utils.usb_data_connected():
```
matejcik, line 70: please render a header instead.
`ui.header_error("NO USB CONNECTION")` might fit?
this error is the highest priority so it should be the first check in the list

hiviah: Changed.

matejcik (approved):
LGTM
will merge once the CI goes green

(next time please don't resolve comments which you didn't start, the reviewer should do that)

we should add a test for this, but it's not really possible currently; we would either need to add emulator control to device_tests, or introduce UI recording to persistence_tests. I will create a separate issue for that.

hiviah:
Everything in CI is green now.

(don't understand why the click test got stuck before but everything looks OK now)

matejcik:
QA instructions:
- get a data-only USB cable; or plug the TT into a charger instead of a PC
- check that "NO USB CONNECTION" is shown on the homescreen (only after unlocking with PIN, i believe)
- check that it is _not_ shown when connecting to a normal PC with a good cable

matejcik:
@hiviah do we have any way for a normal user to trigger the _change_ in this parameter?
i dunno, get a powered USB hub, leave it unplugged from PC, then plug into PC?

bosomt:
No idea what happened here.
I'm using 2.5.3 for some time but i seen it for first time.
I was running CoinJoin so device was not able to go to sleep.
As soon i added another account in Trezor Suite message was gone

image


**Info:**
 - Suite version: desktop 22.11.1 (36517f2367056ec69f25858ddd96684364d1dbc5)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.11.1 Chrome/100.0.4896.160 Electron/18.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.3 Universal (revision 2f03ace311584988d5aeab58fd1acf24ef54711a)
 - Transport: bridge 2.0.32

hiviah:
Does it happen also when you set long autolock to prevent locking and just leave it connected for the same amount of time, without doing anything? How long did it take before the message showed up?

My guess is that OS deconfigured&suspended the device and once you try to connect to it again (make an account), bridge triggered OS to reenumerate USB devices so the device went back to "configured" (actively connected state).

bosomt:
@hiviah device/coinjoin was running all night long during Thursday night. So its possible it after autolock time. Found it like this at Friday morning.

bosomt:
it just happened to me again

- Suite reported device connected
- bridge reported device connected and no session
- device reported `no usb conection`

as soon i added account in suite it changed to `pin not set`
so its definitely not PIN/autolock related, no PIN is set in device 

**Info:**
 - Suite version: desktop 22.12.0 (e97f817b93cf74d122ad303a2572a8fb1a9bc3f0)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.12.0 Chrome/104.0.5112.124 Electron/20.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.3 Universal (revision 2f03ace311584988d5aeab58fd1acf24ef54711a)
 - Transport: bridge 2.0.32

PR #2369, opened 2022-07-05
chore(crypto): remove graphene base58 functions as they are not used anywhere

prusnak:


andrewkozlik (approved):
LGTM

BTW, there are some Graphene test vectors in `core/tests/test_trezor.crypto.base58.py`. The test uses the python implementation of base58, so it's unrelated to the code changes in this PR. I think it's OK to keep the Graphene test vectors in the core unit tests.

prusnak:
> I think it's OK to keep the Graphene test vectors in the core unit tests.

Right. At first I thought we should drop this as well, but then I realised we at least have test which overrides the digest function of `base58.{de,en}code_check` via the `digestfunc` argument.

PR #2371, opened 2022-07-06
Add support for Zcash unified addresses

jarys:
Specification: [ZIP-316](https://zips.z.cash/zip-0316)

Unified addresses feature is conditioned by `USE_ZCASH` flag.
- unix: `USE_ZCASH=1`
- firmware: `USE_ZCASH=BITCOIN_ONLY` , i.e. Zcash parasitizes on btconly free memory


core/src/apps/common/coininfo.py.mako
```diff
@@ -145,6 +151,15 @@ def by_name(name: str) -> CoinInfo:
             ${attr}=${func(coin[attr])},
             % endfor
         )
+% endfor
+    if utils.USE_ZCASH:
```
andrewkozlik, line 155: This should probably be `utils.USE_ZCASH or not utils.BITCOIN_ONLY`, or something similar, depending on https://github.com/trezor/trezor-firmware/issues/2375.

core/src/apps/zcash/addresses.py
```diff
@@ -0,0 +1,128 @@
+"""
+Implementation of encoding and decoding of Zcash
+unified addresses according to the ZIP-316.
+
+see: https://zips.z.cash/zip-0316
+"""
+
+from typing import TYPE_CHECKING
```
andrewkozlik, line 8:
```suggestion
from typing import Dict
```

jarys:
I realized why the flash was overflowing. Crate `f4jumble` depends on crate `blake2b_simd`, which uses enormous amount of inlining. After disabling this behavior by enabling `blake2b_simd/uninline_portable` feature, firmware fits into the flash again.

According to my analysis, dependencies are under 8.7 KiB.
```
  Size        Crate Name
3.7KiB blake2b_simd blake2b_simd::avx2::compress1_loop
1.3KiB blake2b_simd blake2b_simd::guts::Implementation::compress1_loop
1.1KiB blake2b_simd blake2b_simd::portable::round
  379B blake2b_simd blake2b_simd::Params::hash
  362B blake2b_simd blake2b_simd::Params::to_state
  318B     f4jumble f4jumble::State::g_round
  273B blake2b_simd blake2b_simd::State::update
  214B     f4jumble f4jumble::State::h_round
  179B    [Unknown] main
  171B blake2b_simd blake2b_simd::State::finalize
  140B blake2b_simd blake2b_simd::guts::final_block
  137B blake2b_simd blake2b_simd::State::fill_buf
   84B     f4jumble f4jumble::f4jumble_mut
   56B          std <core::ops::range::Range<usize> as core::slice::index::SliceIndex<[T]>>::index_mut
   45B          std <core::ops::range::Range<usize> as core::slice::index::SliceIndex<[T]>>::index
   35B blake2b_simd core::slice::<impl [T]>::copy_from_slice
   29B          std core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
   26B     f4jumble f4jumble::xor
8.7KiB              .text section size, the file size is 26.6KiB
```

prusnak:
Is this for the firmware or for the emulator?

I'm asking because e.g. `blake2b_simd::avx2` does not make sense for the firmware (we don't have Intel AVX2 instructions on ARM).

andrewkozlik:
> According to my analysis, dependencies are under 8.7 KiB.

Good job!

The code duplication between C and Rust is unfortunate. The C implementation of blake2b also uses a lot of inlining and is one of the biggest crypto implementations that we have https://github.com/trezor/trezor-firmware/issues/1947#issuecomment-1096746844. The only ones that are bigger are the ECC implementations that use precomputed tables.

If we want to get rid of the code duplication, then one option is implementing f4jumble in C, which should be very easy. It's just a 4-round Feistel scheme with blake2b at its core: https://github.com/daira/jumble/blob/main/f4jumble.py.

jarys:
> Is this for the firmware or for the emulator?

firmware

> So maybe we could switch to non-SIMD version of blake2b?

This is not so easy since we depend on `blake2b_simd` indirectly through `f4jumble` crate :/ Standard approach would be to fork `f4jumble` and then depend on this fork and maintain it.

> If we want to get rid of the code duplication, then one option is implementing f4jumble in C, which should be very easy.

This is possible optimization for sure.

core/src/apps/zcash/addresses.py
```diff
@@ -0,0 +1,128 @@
...
+from trezor.wire import DataError
+
+from apps.common.coininfo import CoinInfo
+from apps.common.readers import read_compact_size
+from apps.common.writers import write_bytes_fixed, write_compact_size
+
+if TYPE_CHECKING:
+    from typing import Dict
+
+    pass  # this `pass` resolves syntax error in addresses.i
```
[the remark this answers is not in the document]
jarys: 53874f4c4d9ca5ece5111dabdb8d3fdd48bd8d3b

core/src/apps/zcash/addresses.py
jarys: 53874f4c4d9ca5ece5111dabdb8d3fdd48bd8d3b

jarys:
I removed `USE_ZCASH` conditional compilation and enabled `blake2s_simd/uninline_portable` 6f2bf037689ce6f345f8b4be565512d29a0bff26.

But since importing `f4jumble` already soaked through many files, I will probably open a new PR to make a version based on C-blake2b implementation.

jarys:
replacing this PR by #2398

Issue #450, opened 2019-08-27
implement CoSi on Trezor T


matejcik:
The CoSi signing method is available on T1 (so it exists in trezor-crypto), but not on TT. It should be added at some point.

matejcik:
related, what occured to me when working on keyctl:

perhaps the sequence of CosiCommit and CosiSign should be considered a single operation with only one confirmation screen? Namely, if a CosiCommit is followed by a CosiSign with the same `address_n` and `data`, it could be considered a part of the same flow.

This however depends on exact security properties of the CoSi scheme.

1) Is confirmation for the commitment even required? Can an attacker abuse the commitment somehow?
2) Does it make sense to confirm commitment and signature separately? Does it make sense to want to commit to a digest and later decide not to sign it?
3) If (1) is yes, and (2) is no, could we use the commitment as a token that allows prompt-less signing?
4) Can we somehow check the validity of `global_commitment` and `global_pubkey`? Can an attacker substitute malicious values to obtain something interesting?

matejcik:
in light of the CoSi vulnerability, the flow should look as follows:

1. there is only *one* entry point, i.e., `CosiCommit`. `CosiSign` cannot be started separately.
2. generate a nonce and store it in sessionless cache. if there is already a cached nonce, reuse it.
   * reason: otherwise an attacker could pump out entropy via repeatedly CosiCommiting. this is _prooobably_ not a vulnerability but we disallow it just in case
   * sessionless cache to save memory; the CoSi flow can't work across sessions anyhow
4. send the commitment and wait for a `CosiSign`
5. load the cached nonce into a local variable and erase it from cache
6. show confirmation screen
7. produce and send the signature

matejcik:
3.5: check that the `address_n` in `CosiSign` is the same as in `CosiCommit`, abort otherwise

prusnak:
BTW, if we implement CoSi on Trezor T, we can drop the code completely from T1.

prusnak:
> BTW, if we implement CoSi on Trezor T, we can drop the code completely from T1.

Followup PR in https://github.com/trezor/trezor-firmware/pull/2675 but we are not merging this until this PR has been successfully tried in the wild (as in we were indeed able to sign the FW release with Model T).

PR #2372, opened 2022-07-06
feat(crypto): move from C/donna to rust/dalek ed25519

ryankurte:
hi folks! thanks for all your work designing and supporting such neat devices! 

we’ve been working to add support for @mobilecoinofficial, for which we need to add some new cryptography. rather than duplicating the existing primitives or rebuilding our cryptographic libraries and hopefully inline with your goals of unifying and rebuilding components in well-verified rust, we’ve begun a [rust implementation][1] of the `trezor-crypto` library based on the excellent [dalek-cryptography][2] project.

this provides a `trezor-crypto` compatible ABI to avoid duplication and allow continued updates in a piece-wise manner supported by your excellent collection of tests, while new rust components can use the included cryptography libraries directly.

our initial work aims to replace `ed25519-donna`  and the `ge` and `modm` components (including an initial hash/digest interface). this is tested via FFI against the upstream `donna` APIs to ensure compatibility, with the intent that the crate can be expanded to cover more algorithms going forwards (how great would it be to unify digest implementations!)

this PR actions this initial replacement. our library implementation is [here][2] (with docs [here][3] and a reserved name [here][4]), and if you would prefer to have ownership of these dependencies we’d be delighted to transfer the repository (and crate) to y’all along with this PR ^_^

apologies for the -size- of this commit / PR, though the additions are primarily in the cargo lockfile so hopefully this is not too ungainly, and please let us know if you need any changes!

### changes

- adds `trezor-crypto-lib` to `core/embed/rust/src/lib.rs` for linking
- ~~adds `dalek_` prefix to ed25519 methods~~
- adds custom RNG support for `getrandom` in `core/embed/rust/src/lib.rs`
- disables `cosi_combine_signatures` and `cosi_sign` functions
  (pending completion in `trezor-crypto-lib` https://github.com/mobilecoinofficial/rust-trezor-crypto/issues/3, assistance would be amazing)

relates to: https://github.com/trezor/trezor-firmware/issues/1947, https://github.com/trezor/trezor-firmware/issues/450

cc. @mobilecoinofficial/hardware-wallets @isislovecruft

[1]: https://github.com/mobilecoinofficial/rust-trezor-crypto
[2]: https://github.com/dalek-cryptography
[3]: https://mobilecoinofficial.github.io/rust-trezor-crypto/trezor_crypto_lib/
[4]: https://crates.io/crates/trezor-crypto-lib


prusnak:
Hi @ryankurte! Thank you for your contribution. 

Two high-level comments:

1. We have only recently removed public keys from the ed25519 signing APIs - see PR https://github.com/trezor/trezor-firmware/pull/2349 and issue https://github.com/trezor/trezor-firmware/issues/2338. I think it's required that your changeset does not re-introduce the same problem back again.

2. You remove the old ed25519-donna code (I am fine with that) but you reintroduce the functions with `dalek_` prefix (so e.g. `ed25519_sign_open` becomes `dalek_ ed25519_sign_open`). I don't see a good reason for that, considering you are dropping the old implementation. Let's drop the prefix and this will remove lots of unnecessary changes to `core/embed/extmod/modtrezorcrypto/modtrezorcrypto-*.h`

ryankurte:
thanks for the feedback! all sounds good to me. i have rebased, removed the public key argument from sign functions, and worked out how to remove the `dalek_` prefixes while retaining the [compatibility tests](https://github.com/mobilecoinofficial/rust-trezor-crypto/blob/main/tests/ed25519.rs#L11) ^_^

looks like there are a couple of bits and pieces still to clean up (depending on your opinions on maintaining trezor-crypto or matching donna nomenclature), as well as the unimplemented (and patched out) `cosi_combine_publickeys` and `cosi_sign` methods (https://github.com/mobilecoinofficial/rust-trezor-crypto/issues/3).

ryankurte:
fixed the remaining change to `ed25519_publickey_ext`, it looks like some part of the changes upstream have pushed us over flash size by 5128 bytes :-(

ryankurte:
hey @matejcik, thanks for the feedback. we do need to build wallet hardware support so are definitely be interested in how we can work with y’all to make this (or parts of this) suitable for merging!

> The change is too invasive with regards to our stack (we would need to introduce Rust build to legacy)
> 

> … make provisions for bootloader
> 

it is reasonably straightforward to include this as an alternative to the existing cryptography behind a experimental feature flag to avoid this if this would be more palatable?

> audit and perhaps vendor the Rust dependencies, esp. seeing that you have made private copies;
> 

it is definitely important to understand ones dependencies! the `dalek` crates are well reviewed and to be clear any patches are only required until the related upstream PRs have landed.

> you suggested in another issue that this actually consumes more flash space
> 

the comment on #2375 relates to the lack of headroom when implementing new coins.

for this PR (at the moment) flash use is slightly increased by the duplication of `sha2` , `sha3`, and `keccak` impls, i do not believe this is inherent to the implementation in any way it’s just an attempt to minimise the size of the changes.

> … and it brings no functionality that we need.
> 

the primary motivation for us is to add `ristretto`, but we’ve been trying to limit the scope of this wherever as possible 😅 with the side effect of supporting ecosystem-standard cryptography so one can use existing rust codebases rather than reimplementing each coin for the trezor.

> Note that we currently don't plan on adding new coin implementations, so we will not be accepting a MobileCoin implementation in the foreseeable future.
> 

just wondering whether / when y’all expect this to change and whether there’s anything we can do to smooth that process?

matejcik:
> it is reasonably straightforward to include this as an alternative to the existing cryptography behind a experimental feature flag to avoid this if this would be more palatable?

That would just needlessly enlarge the configuration matrix. It might be a reasonable short-term solution if/when we decide to port over the cryptography to Rust, but, again, there is currently no plan to do that.

Going forward, it might bring value to rely on 3rd party cryptography, provided that their nostd / no-alloc support is reasonable. We're running into some problems with that with the Zcash cryptography. Then again, to answer your other question -- there are currently no plans to bring support for _any_ new coins, and this won't change anytime soon. So we actually don't have any need for the new crypto. The company line is that we are a Bitcoin-first company and we do not want to splinter our focus (any more than it already is :) ).

---

On a separate note: given that the use of cryptography in core codebase is basically only through MicroPython, and we are already experimenting with Rust in bootloader, it might be reasonable to completely avoid the ABI compatibility layer and reimplement the Python bindings through Rust directly to the 3rd party libs. That way we would keep the C-based `trezor-crypto` in legacy, boardloader, possibly bootloader, and remove it (partially or completely) from core codebase.

edsimpsons83:
> 
> there are currently no plans to bring support for _any_ new coins, and this won't change anytime soon. So we actually don't have any need for the new crypto. The company line is that we are a Bitcoin-first company and we do not want to splinter our focus (any more than it already is :) ).

This is very unfortunate. I really wanted MobileCoin support on my Trezor.

PR #2374, opened 2022-07-07
chore(docs): Emulator disclaimer

### LINKED ISSUE #2281: Add disclaimer to Emulator build page
@slush0 says:
> Btw, some times ago there was a disclaimer for trezor emulator that it is intended for development purposes only. I don't see such disclaimer there anymore (https://github.com/trezor/trezor-firmware/blob/master/docs/core/emulator/index.md), which make me feel uneasy when I see people are building raspberry pi based Trezors on top of it.
>
> If I remember well, the emulator even don't use proper entropy source, which I don't see mentioned in the docs.

We should very explicitly document that:
* the emulator is intended for development only
* emulator build of trezor-crypto, both on legacy and core, uses a stupid LCG for its randomness source
* there is no consideration whatsoever to make the emulators secure in any way
* and most of the hardening is happening in hardware only
* we don't (currently) plan to change the situation because, as above, the emulator is intended for development only and we don't want to spend effort on maintaining it as a "platform"

sime:
Fixed #2281

### DIFF
--- a/docs/core/emulator/index.md
+++ b/docs/core/emulator/index.md
@@ -8,6 +8,12 @@ There is neither boardloader nor bootloader and no firmware uploads. Emulator ru
 
 Emulator significantly speeds up development and has several features to help you along the way.
 
+## ⚠️ Disclaimer ⚠️
+
+This emulator is for *development purposes only*. It uses a pseudo random number generator, and thus no guarantee on its entropy is made. No security or hardening efforts are made here. It is, and will continue to be, intended for development purposes only. Security and hardening efforts are only made available on [physical Trezor hardware](https://shop.trezor.io/).
+
+Any other usage of the emulator is discouraged. Doing so runs the risk of loosing funds.
+
 ## How to run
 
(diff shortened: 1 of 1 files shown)


docs/core/emulator/index.md
```diff
@@ -8,6 +8,12 @@ There is neither boardloader nor bootloader and no firmware uploads. Emulator ru
 
 Emulator significantly speeds up development and has several features to help you along the way.
 
+## ⚠️ Disclaimer ⚠️
+
+This emulator is for *development purposes only*. It uses a pseudo random number generator, and thus no guarantee on its entropy is made. No security or hardening efforts are made here. It is, and will continue to be, intended for development purposes only. Security and hardening efforts are only made available on [physical Trezor hardware](https://shop.trezor.io/).
+
+Any other usage of the emulator is discouraged. Doing so runs the risk of loosing funds.
```
matejcik, line 15:
```suggestion
Any other usage of the emulator is discouraged. Doing so runs the risk of losing funds.
```

PR #2376, opened 2022-07-08
Set output length in xmr_base58_decode()

andrewkozlik:
- Fixes `xmr_base58_decode()` to return the decoded data size. Resolves https://github.com/satoshilabs/trezor-firmware/issues/160.
- Introduces a new constant `max_bin_data_size=64`, which is sufficient for the spend key and the view key. Before there were two constants 64 and 128 for decoding and encoding.

### DIFF
--- a/crypto/monero/base58.c
+++ b/crypto/monero/base58.c
@@ -43,6 +43,7 @@ const size_t encoded_block_sizes[] = {0, 2, 3, 5, 6, 7, 9, 10, 11};
 const size_t full_block_size = sizeof(encoded_block_sizes) / sizeof(encoded_block_sizes[0]) - 1;
 const size_t full_encoded_block_size = 11; // encoded_block_sizes[full_block_size];
 const size_t addr_checksum_size = 4;
+const size_t max_bin_data_size = 64;
 const int decoded_block_sizes[] = {0, -1, 1, 2, -1, 3, 4, 5, -1, 6, 7, 8};
 #define reverse_alphabet(letter) ((int8_t) b58digits_map[(int)letter])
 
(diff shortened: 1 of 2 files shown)


andrewkozlik:
> 1. [5bfe481](https://github.com/trezor/trezor-firmware/commit/5bfe48145c2398ea0fb710a93e5c619ccbffbbee) sets output length correctly in `xmr_base58_encode` if the input length is zero.

LGTM

> 2. [b1bc5d3](https://github.com/trezor/trezor-firmware/commit/b1bc5d35607ecf07b528ca9cbb4287b9c834b423) fixes (lowers by one) minimum required output length in `xmr_base58_encode`. To be honest, I'm to sure about correctness of this commit.

LGTM, the original code uses `data.resize(data_size, 0);`, so it seems to make sense.

> 3. [67118fa](https://github.com/trezor/trezor-firmware/commit/67118faf6f24b9151649645196c25736125ac751) is not much important. `xmr_base58_encode` allows the variable into which the size of the output is stored to be `NULL`. This commit does the same for `xmr_base58_decode`.

- I think allowing `binsz==NULL` is unsafe design, because it is better for `xmr_base58_decode()` to check that the `data` array is sufficiently large for the output. The same argument applies to `xmr_base58_encode()`, but I am not sure I want to change existing behavior, in case some 3rd party relies on this.
- There are three other occurrences of `binsz` in `xmr_base58_decode()` which would need to be taken care of.

> Moreover, I'm not sure why the style checker is turned off for this file, but if would certainly be beneficial if it wasn't.

I think this might have been done to ease bug fixes coming in from upstream. However, the code is already so different, that patches couldn't be applied directly anyway. Applied formatting in [51ba2f5](https://github.com/trezor/trezor-firmware/pull/2376/commits/51ba2f5b99a5a3dd136806daec9b2bdf19cd9154).

PR #2396, opened 2022-07-21
Fix firmware initialization

### LINKED ISSUE #2130: Do not initialize touch controller in firmware

TychoVrahe:
This PR partially reintroduces touch controller initialization in firmware:
- turns out that if we want to change operation mode of the controller in future, we might have issues when settings from bootloader being incompatible with algorithm used in firmware - this could happen for example if new bootloader is used with intentionally downgraded firmware
- the removal of initialization also caused uninitialized i2c handler being used, which only accidentally didn't cause crashes due to fact that the touch_read function evaluated the uninitialized handler as being BUSY, and thus did I2C reset cycle, which fixed things - but this introduced some unnecessary delays
- this PR also fixes Model R firmware initialization, where some stuff wasn't called and some stuff was called twice or erroneously. A little reorganization of compilation conditions was make the process more clear.

closes #2130

### DIFF
--- a/core/.changelog.d/2130.changed
+++ b/core/.changelog.d/2130.changed
@@ -1,0 +1,1 @@
+Remove power-down power-up cycle from touch controller initialization in firmware
--- a/core/.changelog.d/2130.removed
+++ b/core/.changelog.d/2130.removed
@@ -1,1 +1,0 @@
-Don't initialize touch controller in firmware
--- a/core/embed/bootloader/main.c
+++ b/core/embed/bootloader/main.c
@@ -250,8 +250,8 @@ int main(void) {
   random_delays_init();
   // display_init_seq();
 #if defined TREZOR_MODEL_T
-  touch_init();
   touch_power_on();
+  touch_init();
 #endif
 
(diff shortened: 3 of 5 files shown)


hiviah (approved):
LGTM, though at first the ifdef regrouping was a bit confusing.

matejcik:
please remove or modify this changelog entry: https://github.com/trezor/trezor-firmware/blob/master/core/.changelog.d/2130.removed

TychoVrahe:
Modified the log entry.

matejcik (approved):

matejcik:
please squash the fixup and merge

PR #2398, opened 2022-07-23
Add support for Zcash unified addresses - memory optimized

jarys:
specification: https://zips.z.cash/zip-0316
optimization: f4jumble now uses C blake2b module


jarys:
This PR is ready for review. I don't understand why CI `core fw btconly build` fails. There is no error message.

core/src/apps/zcash/f4jumble.py
```diff
@@ -0,0 +1,58 @@
...
+    mask = blake2b(
+        personal=b"UA_F4Jumble_H" + bytes([i, 0, 0]),
+        outlen=len(left),
+        data=bytes(right),
+    ).digest()
+    xor(left, mask)
+
+
+def f4jumble(message: memoryview) -> None:
+    assert len(message) >= 22
```
prusnak, line 38: Why 22?

Upstream says the min length is 48 (and max length is 4194368): https://github.com/zcash/librustzcash/blob/37fc28634e811fdf8db9274a0080cff17c6b6a09/components/f4jumble/src/lib.rs#L67-L70

core/src/apps/zcash/unified_addresses.py
```diff
@@ -0,0 +1,122 @@
...
+def receiver_length(typecode: int):
+    """Byte length of a receiver."""
+    if typecode == Typecode.P2PKH:
+        return 20
+    if typecode == Typecode.P2SH:
+        return 20
+    if typecode == Typecode.SAPLING:
+        return 43
+    if typecode == Typecode.ORCHARD:
+        return 43
```
prusnak, line 30:
```suggestion
    if typecode in  [Typecode.P2PKH, Typecode.P2SH]:
        return 20
    if typecode in [Typecode.SAPLING, Typecode.ORCHARD]:
        return 43
```

core/src/apps/zcash/unified_addresses.py
```diff
@@ -0,0 +1,122 @@
...
+    if coin.coin_name == "Zcash":
+        return "u"
+    if coin.coin_name == "Zcash Testnet":
+        return "utest"
+    raise ValueError
+
+
+def padding(hrp: str) -> bytes:
+    assert len(hrp) <= 16
+    return bytes(hrp, "utf8") + bytes(16 - len(hrp))
```
prusnak, line 45: uPy supports only UTF-8

```suggestion
    return bytes(hrp) + bytes(16 - len(hrp))
```

prusnak (changes requested):

andrewkozlik:
> I don't understand why CI `core fw btconly build` fails. There is no error message.

At the end of the log you can see
```
o%trezor/enums/ZcashReceiverTypecode.py
trezor/enums/ZcashReceiverTypecode.py
```
This is the output of `tools/check-bitcoin-only`. We should clarify the error message.

If you run the following commands you will get the same result.
```
BITCOIN_ONLY=1 make build_firmware
strings core/build/firmware/firmware.bin | grep Zcash
```

I think what you need to do is look at `SConscript.firmware` and `SConscript.unix` and exclude `trezor/enums/Zcash*.py` from the Bitcoin-only build.

core/src/apps/zcash/f4jumble.py
jarys: You are right.

I started by copy-pasting implementation by daira. Then I found that it's outdated and I forgot to change this.

https://github.com/daira/jumble/blob/main/f4jumble.py#L18

jarys: https://github.com/trezor/trezor-firmware/pull/2398/commits/8dd8eff3e1e4c1fa0593987a0d3b697bcd5aed0a

core/src/apps/zcash/unified_addresses.py
jarys: https://github.com/trezor/trezor-firmware/pull/2398/commits/e54626dcad3d7fa22fc85c337bb44a760a024d99

jarys:
Thank you for clarification Andrew. I'm removing `trezor/enums/Zcash*.py` from btconly build in https://github.com/trezor/trezor-firmware/pull/2398/commits/2a18731b99d1ea9d8c30404f1efc95f629bb6c10 .

prusnak (approved):

jarys:
may I squash and merge , or waiting for another review ? @andrewkozlik

core/src/apps/zcash/unified_addresses.py
```diff
@@ -0,0 +1,118 @@
...
+        return "utest"
+    raise ValueError
+
+
+def padding(hrp: str) -> bytes:
+    assert len(hrp) <= 16
+    return hrp.encode() + bytes(16 - len(hrp))
+
+
+def encode(receivers: Dict[Typecode, bytes], coin: CoinInfo) -> str:
```
matejcik, line 44: please use py3.10-style annotations, i.e., in this case `dict[Typecode, bytes]` without importing `Dict`

core/src/apps/zcash/unified_addresses.py
```diff
@@ -0,0 +1,118 @@
...
+from apps.common.coininfo import CoinInfo
+from apps.common.readers import read_compact_size
+from apps.common.writers import write_bytes_fixed, write_compact_size
+
+from .f4jumble import f4jumble, f4unjumble
+
+
+def receiver_length(typecode: int):
+    """Byte length of a receiver."""
+    if typecode in [Typecode.P2PKH, Typecode.P2SH]:
```
matejcik, line 23: nit: these should be tuples `(Typecode.P2PKH, Typecode.P2SH)` not lists

matejcik (approved):
did a brief review, two small nitpicks.

@andrewkozlik do we want to add this to the upcoming release?

andrewkozlik:
> may I squash and merge , or waiting for another review ?

I will try to review today.

> do we want to add this to the upcoming release?

Is there any reason not to?

matejcik:
> Is there any reason not to?

the freeze is today; my question is, are you as the main reviewer confident enough in this code to include it?

common/protob/messages-zcash.proto
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Receiver typecodes for unified addresses.
+ * see: https://zips.z.cash/zip-0316#encoding-of-unified-addresses
+ */
+enum ZcashReceiverTypecode {
+  P2PKH = 0;
+  P2SH = 1;
+  SAPLING = 2;
+  ORCHARD = 3;
+}
```
andrewkozlik, line 17: Why do we need this `enum` in a protobuf definition? As far as I am able to tell this is never communicated over USB, so it could be defined more locally in `unified_addresses.py`. Will it be needed in the future in the protobuf messages? If yes, then how exactly?

core/tests/test_apps.zcash.unified_addresses.py
```diff
@@ -0,0 +1,91 @@
+from common import *
+
+from apps.common import coininfo
+from apps.zcash import unified_addresses
+from trezor.enums.ZcashReceiverTypecode import P2PKH, P2SH, SAPLING, ORCHARD
+
+# source: https://github.com/jarys/zcash-test-vectors/blob/trezor/unified_address.py
```
andrewkozlik, line 7: I found some official-looking test vectors at https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/test-vectors/json/unified_address.json. It looks like the ones in this file are different. Do these also come from some official source?

core/src/apps/zcash/unified_addresses.py
```diff
@@ -0,0 +1,118 @@
...
+            if length != expected_length:
+                raise DataError("Unexpected receiver length")
+        except ValueError:
+            pass
+
+        if r.remaining_count() < length:
+            raise DataError("Invalid receiver length")
+
+        receivers[typecode] = r.read(length)
+
```
andrewkozlik, line 117: > Consumers MUST reject Unified Addresses/Viewing Keys [...] that include both P2SH and P2PKH Transparent Addresses, or that contain only a Transparent Address.

Technically we should be checking this and ideally have a unit test for each of these cases.

core/src/apps/zcash/unified_addresses.py
```diff
@@ -0,0 +1,118 @@
...
+
+def padding(hrp: str) -> bytes:
+    assert len(hrp) <= 16
+    return hrp.encode() + bytes(16 - len(hrp))
+
+
+def encode(receivers: Dict[Typecode, bytes], coin: CoinInfo) -> str:
+    # multiple transparent receivers forbidden
+    assert not (Typecode.P2PKH in receivers and Typecode.P2SH in receivers)
```
andrewkozlik, line 46: Also
```python
assert Typecode.SAPLING in receivers or Typecode.ORCHARD in receivers
```
But I am wondering whether it is right to treat this with `assert`. It would seem more appropriate to use `if ... raise` at least until we have some guarantee that the calling code is checking this. (Right now `encode()` is not used.)

common/protob/messages-zcash.proto
andrewkozlik: The only place that I can think of where this could be used is for unified addresses in `GetAddress` if there would be a repeated field for the receivers. In that case I wonder if it wouldn't be better to recycle `InputScriptType` for this purpose in `GetAddress`.

andrewkozlik:
> the freeze is today; my question is, are you as the main reviewer confident enough in this code to include it?

I don't think we need to force this into the release, since Suite integration of unified addresses is not on the roadmap yet. So I am not sure this could be used in production by anyone in the nearest future.

On the other hand if we happen to merge this today somehow, then it's fine to include this in the freeze. (My apologies for not getting this reviewed earlier. I had a lot on my plate.)

prusnak:
Let's merge after the freeze if there is no strong reason to merge it before.

core/src/apps/zcash/unified_addresses.py
jarys: [fcc5fd1](https://github.com/trezor/trezor-firmware/pull/2398/commits/fcc5fd1cccbb92be5248f9179479e32e5856bb77)

core/src/apps/zcash/unified_addresses.py
jarys: [c536e71](https://github.com/trezor/trezor-firmware/pull/2398/commits/c536e71da57fdbcde00ff0c87405f3f51b4a25bf)

common/protob/messages-zcash.proto
jarys: No. This is not exposed and it will never be exposed. `t_address_n` and `z_address_n` fields will be used for getting unified addresses.

I will move this enum into unified_addresses.py

core/tests/test_apps.zcash.unified_addresses.py
jarys: I generated test vector by the same official library (zcash-test-vectors) with two minor changes:

- I generated ALL possible combinations of receivers. (The generated 10 random combinations.)
- I serialized test vectors differently to load them more easily in python.

Is it ok?

core/src/apps/zcash/unified_addresses.py
jarys: adding the check [c7b9b7d](https://github.com/trezor/trezor-firmware/pull/2398/commits/c7b9b7d3954062246288f47b6dac644da0ee82b2)

jarys: adding the second check [e780855](https://github.com/trezor/trezor-firmware/pull/2398/commits/e78085529fcc47d53fcf7045947f24d02837dcfc)

core/src/apps/zcash/unified_addresses.py
jarys: `assert` seems to fine to me since input of this function will be always created internally.

jarys: [6a410a8](https://github.com/trezor/trezor-firmware/pull/2398/commits/6a410a833c9ba28c91a0b364a7856b7fd08a5cbf)

core/tests/test_apps.zcash.unified_addresses.py
andrewkozlik: > - I generated ALL possible combinations of receivers. (They generated 10 random combinations.)

Generating more combinations is certainly good work. It would be nicer though if our test vectors were a superset of the official ones.

> - I serialized test vectors differently to load them more easily in python.

Absolutely OK.

common/protob/messages-zcash.proto
jarys: [5f33136](https://github.com/trezor/trezor-firmware/pull/2398/commits/5f33136b275176497f1e72d4651beb1aa3ce7d1d)

core/src/apps/zcash/unified_addresses.py
jarys: fixing typechecking [163f493](https://github.com/trezor/trezor-firmware/pull/2398/commits/163f493d46ecf31c7f10077a1a9b102145ad6cae)

core/tests/test_apps.zcash.unified_addresses.py
jarys: updated to official test vectors [163f493](https://github.com/trezor/trezor-firmware/pull/2398/commits/163f493d46ecf31c7f10077a1a9b102145ad6cae)

jarys: While testing on new testvectors I found a mistake :-( I reversed receivers order (which should be ascending). It is repaired now. I got confused by receivers priority order (which is descending).

jarys: [42f4215](https://github.com/trezor/trezor-firmware/pull/2398/commits/42f42155148673156a5275cead794d5ee18cba12)

core/src/apps/zcash/unified_addresses.py
```diff
@@ -0,0 +1,137 @@
...
+def encode(receivers: dict[Typecode, bytes], coin: CoinInfo) -> str:
+    # multiple transparent receivers forbidden
+    assert not (Typecode.P2PKH in receivers and Typecode.P2SH in receivers)
+    # at least one shielded address must be present
+    assert Typecode.SAPLING in receivers or Typecode.ORCHARD in receivers
+
+    length = 16  # 16 bytes for padding
+    for typecode, receiver_bytes in receivers.items():
+        length += 2  # typecode (1 byte) + length (1 byte)
+        length += receiver_length(typecode) or len(receiver_bytes)
```
andrewkozlik, line 64: nit: I'd simplify this for better legibility, since it's only to estimate the size for buffer preallocation.
```suggestion
    length = 16  # 16 bytes for padding
    for receiver_bytes in receivers.values():
        length += 2  # typecode (1 byte) + length (1 byte)
        length += len(receiver_bytes)
```

andrewkozlik (approved):
utACK

matejcik:
looks like this has all the approval it needs
please squash the fixups, rebase on current master, and fix the UI fixtures conflict

core/src/apps/zcash/unified_addresses.py
jarys: [b70a2c1](https://github.com/trezor/trezor-firmware/pull/2398/commits/b70a2c15d742251515ceb49d4b3a96a88cdf2d16)

andrewkozlik (approved):

jarys:
core fw regular debug build:
```
region `FLASH2' overflowed by 1024 bytes
```
:cry:

andrewkozlik:
Also, the address in `test_unified_address` has `Invalid receivers order`.

> region `FLASH2' overflowed by 1024 bytes

Strange, I still have 4 kB left when I build with `PYOPT=0 poetry run make -C core build_firmware` on my machine.

```
code_length: 1696256
-rw-rw-r-- 1 andrew andrew  1699840 Aug  9 10:14 firmware.bin
-rwxrwxr-x 1 andrew andrew   786432 Aug  9 10:14 firmware.bin.p1
-rwxrwxr-x 1 andrew andrew   913408 Aug  9 10:14 firmware.bin.p2
```

PR #2412, opened 2022-07-26
Optimize touch controller polling

### LINKED ISSUE #262: Reevaluate touch panel polling
It is possible that one of the causes behind the recent issues with the Trezor freezing (https://github.com/trezor/trezor-firmware/issues/144 and https://github.com/trezor/trezor-firmware/pull/217) may have been the way in which we get the touch events from the screen. We use the interrupt polling mode, see https://cdn-shop.adafruit.com/datasheets/FT6x06_AN_public_ver0.1.3.pdf. This means that when a finger is touching the screen we repeatedly send requests over I2C to get the current position of the finger. It amounts to something like over 1000 requests per second. See https://github.com/trezor/trezor-firmware/blob/master/core/embed/trezorhal/touch_t.h#L229. On its own, this doesn't seem to cause any problems, but when it takes place together with another resource intensive operation, such as generating 1000 random numbers every second, it might be causing the freeze issue we observed.

Let's consider some alternative solutions, e.g.:
- Use the touch panel in interrupt trigger mode.
- Use the touch panel in interrupt polling mode, but poll the touch panel no more than every 10 ms.

TychoVrahe:
This PR optimizes polling the touch controller by only polling when new data are ready, which significantly reduce cpu load when touch is active. 

Also removed the duplicate data detection, with reasoning that previously we filtered this out because we were literally polling the same data multiple times until new data were ready. We might still receive new data with same values though, not sure we would ever find this useful, so maybe the filtering should stay?

This could be also taken a bit further, using interrupts and DMA we could probably offload most of the work to peripherals, but it brings some issues to solve like polling done in the background even when no one wants the data, possibly synchronizing the access to the read buffer, and introducing new interrupt(s) brings some bootloader-firmware compatibility issues.

Builds on #2396 

closes #262

### DIFF
--- a/core/.changelog.d/262.added
+++ b/core/.changelog.d/262.added
@@ -1,0 +1,1 @@
+Optimize touch controller communication
--- a/core/embed/bootloader/.changelog.d/262.added
+++ b/core/embed/bootloader/.changelog.d/262.added
@@ -1,0 +1,1 @@
+Optimize touch controller communication
--- a/core/embed/trezorhal/touch.c
+++ b/core/embed/trezorhal/touch.c
@@ -39,6 +39,9 @@
 #define Y_POS_MSB (touch_data[5] & 0x0FU)
 #define Y_POS_LSB (touch_data[6])
 
+#define EVENT_OLD_TIMEOUT_MS 50
+#define EVENT_MISSING_TIMEOUT_MS 50
+
 static I2C_HandleTypeDef i2c_handle;
 
 static void touch_default_pin_state(void) {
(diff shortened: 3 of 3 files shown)


matejcik:
please fix line endings on the changelog entries (and please fix your editor config :) )

TychoVrahe:
fixed the changelog + rebased after merging #2396

matejcik:
note: this should not be merged before the upcoming freeze (Aug 2)

matejcik:
freeze is done. @hiviah please review, if it is OK by you then you can merge right away

hiviah (approved):

PR #2415, opened 2022-07-26
Ensure correct order when verifying external inputs in BTC signing

andrewkozlik:
Fixes a bug in verifying external inputs in Taproot-only transactions. `self.external` is a `set`, not a `list`, so the external inputs might be walked through in non-sequential order, which results in `h_check` not matching `self.h_external_inputs`.

### DIFF
--- a/core/.changelog.d/2415.fixed
+++ b/core/.changelog.d/2415.fixed
@@ -1,0 +1,1 @@
+Ensure correct order when verifying external inputs in Bitcoin signing.
--- a/core/src/apps/bitcoin/sign_tx/bitcoin.py
+++ b/core/src/apps/bitcoin/sign_tx/bitcoin.py
@@ -203,15 +203,15 @@ class Bitcoin:
             # the amounts and scriptPubKeys, because if an invalid value is provided then all
             # issued signatures will be invalid.
             expected_digest = self.h_external_inputs
-            for i in self.external:
+            for i in range(self.tx_info.tx.inputs_count):
                 progress.advance()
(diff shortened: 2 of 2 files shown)


matejcik:
could we do a testcase? two external inputs and check that the tx succeeds regardless of the order in which they are specified?

andrewkozlik:
> could we do a testcase? two external inputs and check that the tx succeeds regardless of the order in which they are specified?

To be precise, the order is not important, it's the indices of the external inputs. The first two cases below will succeed even with the faulty implementation. The third one will happen to fail. So we could add a test case where input 0 is internal and inputs 1 and 2 are external, but one thing I don't like about that is that it is python-implementation dependent. If micropython changes the implementation of `set`, then `list(set([1,2]))` might come out sorted, but some other choice of indices might not.
```python
>>> s=set(); s.add(0); s.add(1)
>>> [i for i in s]
[0, 1]
```
```python
>>> s=set(); s.add(0); s.add(2)
>>> [i for i in s]
[0, 2]
```
```python
>>> s=set(); s.add(1); s.add(2)
>>> [i for i in s]
[2, 1]
```

One thing that I forgot to note is that the fix could be implemented as a one-liner by adding `sorted()`:
```python
for i in sorted(self.external):
    ...
```
That might also be clearer to the reader, but it probably allocates a list on the heap.

matejcik:
yes, you are right that this is problematic. OK, let's leave it without a testcase.

matejcik (approved):

PR #2422, opened 2022-07-29
Fix rounding in fee rate computation

andrewkozlik:
This should make the fee rate appear the same on T1 and TT during signing. Someone ought to check the screens when the CI finishes, I haven't checked myself.


andrewkozlik:
https://github.com/trezor/trezor-firmware/pull/2422/commits/8d61f3e3506f3a3d7aacd9e2fc77482661bb9921 should avoid situations like in https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2792289454 where the critical information is spread over lines 47, 130, 173 and 178, which are interleaved with all the other output. (Related to https://github.com/trezor/trezor-firmware/pull/2422/commits/308d837424bfebe2bbaf48b7a2f0f27f9f51ee31 and https://github.com/trezor/trezor-firmware/pull/2422/commits/04686c2fc97ce68928acb01eb28b1fd41e1e5b78.)

matejcik:
waiting for QA confirmation that "round to nearest decimal" is what Suite wants.

matejcik:
there's a Suite-side issue to be fixed and meanwhile we can accept this to make sure the behavior is the same between T1 and TT

PR #2423, opened 2022-07-29
Better way to debug T1 combinations of debug/production combinations of bootloader and firmware

hiviah:
First, we try to avoid to have to abide by this table of allowed combinations, which exist due to historic artifacts (broken on-silicon write protection for F205):
  * https://docs.trezor.io/trezor-firmware/legacy/index.html#combining-bootloader-and-firmware-with-various-production-settings-signedunsigned

Firmware will check first whether it's in privileged mode before trying to set VTOR and stack, avoiding fault and making all relevant combinations possible.

Second, `PRODUCTION=0` bootloader will actively check that you are not trying to flash production firmware which would brick SWD/JTAG debug,


matejcik:
done in master

PR #2424, opened 2022-07-31
UFO (Uniform Fiscal Object) support

btclinux:
https://ufobject.com

https://www.coingecko.com/en/coins/uniform-fiscal-object

https://coinmarketcap.com/currencies/uniform-fiscal-object


matejcik:
only coins in top 30 by market cap AND coins that have a wallet with Trezor support are considered for acceptance

PR #2425, opened 2022-08-01
Add thousands separators to amounts

### LINKED ISSUE #2394: Thousand amount separator
With the use of `sat` units, there was a need to implement thousand separators for amount in Trezor Suite.
https://github.com/trezor/trezor-suite/issues/5545

The user needs to see the amount in the Suite and on the Trezor device in the same format for control when sending a transaction.

Most required is `en-us` formatting (like i.e. `123,456.78`) but more general solution would be even better for the future.

## Acceptance criteria
Given I view my coins in sats
When I prepare a transaction for 4000 sats
And I press Review & Send
Then I should see 'Confirm sending 4,000 sat' on the Trezor

### Model 1
Given I see 'Confirm sending 4,000 sat'
When I tap 'Confirm'
Then I should see 'Really send 4,153 sat'

### Model T
Given I see 'Confirm sending 4,000 sat'
When I tap 'Confirm'
Then I should see 'Total amount: 4,153 sat'

matejcik:
fixes #2394 

on trezor-core side, the thousands separators are only added to amounts, or more generally, anything formatted with `format_amount`. There are several more places where large integers are displayed which are not amounts. These are mostly out of scope however.

On legacy, there is no special "format_amount", so the general `bn_format` function is modified. I considered adding a parameter to specify whether the caller wants thousands separators or not, but decided against it in the end, as it would require changes in a lot of places and using thousands separators everywhere seems desirable anyway.


matejcik:
all amounts.

is there a requirement to _not_ do this for non-sats?

grdddj:
Looking at the big negative diff - looks like all the `TTui2...` test-cases from `fixtures.json` got deleted. Was this intentional?

matejcik:
> the general `bn_format` function is modified

this is a problem because it seems to be used internally in NEM code. it is weird that the NEM device tests did not crash on that though.

crypto/bignum.c
```diff
@@ -1781,10 +1781,17 @@ size_t bn_format(const bignum256 *amount, const char *prefix, const char *suffix
...
+    bool is_zero = false;
     do {
+      ++digits;
       bn_divmod10(&temp, &digit);
+      is_zero = bn_is_zero(&temp);
       BN_FORMAT_ADD_OUTPUT_CHAR('0' + digit)
-    } while (!bn_is_zero(&temp));
+      if (!is_zero && digits % 3 == 0) {
+        BN_FORMAT_ADD_OUTPUT_CHAR(',')
+      }
```
andrewkozlik, line 1793: The "Python prototype" documentation [above](https://github.com/trezor/trezor-firmware/pull/2425/files#diff-ae29a9bad321bf657961dd96f640a325907b9b1878fa746179e27909caefb95fR1675-R1694) should be updated to reflect the thousands separator.

sime:
> all amounts.
> 
> is there a requirement to _not_ do this for non-sats?

Originally yes. Clarified with @hynek-jina that Suite will thousand separate all cryptocurrencies. 

So please disregard my previous comment.

crypto/bignum.c
```diff
@@ -1781,10 +1781,17 @@ size_t bn_format(const bignum256 *amount, const char *prefix, const char *suffix
...
+    } while (!is_zero);
```
andrewkozlik, line 1794: I think this won't work correctly if `exponent > decimals`. We need to start counting the digits and adding separators in the block [above](https://github.com/trezor/trezor-firmware/pull/2425/files#diff-ae29a9bad321bf657961dd96f640a325907b9b1878fa746179e27909caefb95fR1777-R1779).

matejcik:
> looks like all the TTui2... test-cases from fixtures.json got deleted

fixed in [3d7ea3f](https://github.com/trezor/trezor-firmware/pull/2425/commits/3d7ea3fbe035ff38a8fd0a5c3185e6b987646561)

crypto/bignum.c
matejcik: fixed in [11db39c](https://github.com/trezor/trezor-firmware/pull/2425/commits/11db39c3720a03b4ac8b75b4aa4551a65d3c4fb3)

crypto/bignum.c
matejcik: fixed in [11db39c](https://github.com/trezor/trezor-firmware/pull/2425/commits/11db39c3720a03b4ac8b75b4aa4551a65d3c4fb3)

core/tests/test_trezor.strings.py
```diff
@@ -4,13 +4,20 @@
...
             (-4242, 7, "-0.0004242"),
+            (123, 5, "0.00123"),
+            (100, 5, "0.001"),
+            (123456789, 0, "123,456,789"),
+            (100000000, 5, "1,000"),
+            (100000001, 5, "1,000.00001"),
+            (100001000, 5, "1,000.01"),
+            (-100001000, 5, "-1,000.01"),
+            (123_456_789_123_456_789_123_456_789, 18, "123,456,789.123456789123456789"),
         ]
```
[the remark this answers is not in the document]
matejcik: added in [11db39c](https://github.com/trezor/trezor-firmware/pull/2425/commits/11db39c3720a03b4ac8b75b4aa4551a65d3c4fb3)

matejcik:
because of the NEM thing I had to change `bn_format` to accept an argument after all. There is now a separate commit for `crypto`, and `legacy` makes use of it through the new `bn_format_amount` function that hides a bunch of parameters.
When rebasing, I will need to move the "legacy" commit after the "crypto" commit.

andrewkozlik:
I am wondering whether `output_length` is always sufficiently large to accommodate the addition of the  the thousands separators. In case of `bn_format_uint64()` we typically allocate 32 bytes, which should always be sufficient. The most extreme example with `UINT64_MAX` that I can think of would be "-1,844,674,407,370,955,161.5 uDASH", which is 35 bytes including the terminating NULL, but I think there is no way that we can run into that considering that the maximum possible amounts in sats are nowhere near `UINT64_MAX`.

I didn't check the size of the allocated buffers thoroughly, because I assume that the device tests would be likely to uncover any problems especially if we check the UI test screens thoroughly.

marnova:
LGTM after fixes, you just missed one fixture https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2797235219/artifacts/test_ui_report/failed/TT_zcash-test_sign_tx.py::test_spend_v5_input.html

grdddj (approved):
Approve for the python part

marnova:
Don't forget to change the fixture for https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2797235219/artifacts/test_ui_report/failed/TT_zcash-test_sign_tx.py::test_spend_v5_input.html , otherwise approve

marnova (approved):

andrewkozlik (approved):

andrewkozlik:
Adding thousands separators in these cases looks a bit weird:
- `#1,337`
- `Memo (ID) 12,345,678,901,234`

https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2803415060/artifacts/master_diff/diff/T1_stellar-test_stellar.py::test_sign_tx[StellarManageBuyOfferOp].html
https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2803415060/artifacts/master_diff/diff/T1_stellar-test_stellar.py::test_sign_tx[StellarManageSellOfferOp].html
https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2803415060/artifacts/master_diff/diff/T1_stellar-test_stellar.py::test_sign_tx[memo_id].html

matejcik:
> Adding thousands separators in these cases looks a bit weird:

it does, but OTOH it's a correct rendering of numbers so 🤷‍♀️

PR #2433, opened 2022-08-02
Remove FW dumping

matejcik:
reverts #2248 

We are not going to use this functionality and there is no point in keeping it around.
The feature was never publicly available so the removal should have little impact.


matejcik:
The main drawback is that we can no longer test the firmware hashing against the direct dump from device.
OTOH, that was never reliable, and we should use an external image in any case.

Issue #2360, opened 2022-07-01
Conform to -Wstrict-prototypes


invd:
The codebase contains a number of C functions that do not take any parameters but are not marked with `void`. This style is deprecated, and compilers warn for it with `-Wstrict-prototypes`:
```C
zkp_context.c:54:21: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
int zkp_context_init() {
                    ^
                     void
```

With the upcoming `clang-15` compiler, `-Wstrict-prototypes` is now enabled by default, leading to build failures due to treating warnings as errors via `-Werror`. The relevant [change discussion](https://discourse.llvm.org/t/rfc-enabling-wstrict-prototypes-by-default-in-c/60521) also gives some context why the explicit `void` definition is better.

I propose switching all zero-parameter functions to the explicit void definition, and explicitly enabling `-Wstrict-prototypes` on gcc builds as well to ensure that no regressions are introduced. There are about a dozen cases in `zkp_context.c`, `oled.c`, `common.c`, `config.c`, `fsm_msg_common.h`, `stellar.c` and other places. 

This may help the compiler to produce better code but should otherwise introduce no functional changes.

invd:
Update: `-Wstrict-prototypes` is already explicitly enabled in the [legacy](https://github.com/trezor/trezor-firmware/blob/c5d22b43955f7bb431a5804ef0a9566592316119/legacy/Makefile.include#L58) and [crypto](https://github.com/trezor/trezor-firmware/blob/c5d22b43955f7bb431a5804ef0a9566592316119/crypto/Makefile#L50) build configurations. It appears that the warning behavior has changed with `clang-15`, though.

PR #2449, opened 2022-08-12
Improve trezor-crypto fuzzer

invd:
Improve existing fuzzer harnesses
Add new fuzz testing of Monero, BIP39, segwit address and ECDH related functions
Add special Memory Sanitizer checks
Use new fuzzer input marking functionality
Additional documentation
Remove build workaround for clang-14 that has been fixed in the code
Fix -Wstrict-prototypes issue in `zkp_context.c`

### DIFF
--- a/crypto/fuzzer/fuzzer.c
+++ b/crypto/fuzzer/fuzzer.c
@@ -150,7 +150,7 @@ void check_msan(void *pointer, size_t length) {
 // warning: use only if no manual memory cleanup is needed
 #define RETURN_IF_NULL(var_pointer) \
   if (var_pointer == NULL) {        \
-    return -1;                       \
+    return -1;                      \
   }
 
 void zkp_initialize_context_or_crash(void) {
@@ -324,12 +324,10 @@ int fuzz_base58_encode_check(void) {
 
   int ret = 0;
   // run multiple hasher variants for the same input
-  base58_encode_check(in_buffer, raw_inlen, HASHER_SHA2D, out_buffer,
-                            outlen);
-  base58_encode_check(in_buffer, raw_inlen, HASHER_BLAKED, out_buffer,
-                            outlen);
-  base58_encode_check(in_buffer, raw_inlen, HASHER_GROESTLD_TRUNC,
-                            out_buffer, outlen);
+  base58_encode_check(in_buffer, raw_inlen, HASHER_SHA2D, out_buffer, outlen);
+  base58_encode_check(in_buffer, raw_inlen, HASHER_BLAKED, out_buffer, outlen);
+  base58_encode_check(in_buffer, raw_inlen, HASHER_GROESTLD_TRUNC, out_buffer,
+                      outlen);
   ret = base58_encode_check(in_buffer, raw_inlen, HASHER_SHA3K, out_buffer,
                             outlen);
 


invd:
@andrewkozlik could you have a look at this PR? As with #2063, the changes are designed to avoid any functional changes in the production code.

crypto/fuzzer/fuzzer.c
```diff
@@ -96,82 +119,103 @@ void fuzzer_reset_state(void) {
...
+  (void)length;
+#endif
+#endif
+}
+
+// simplify the pointer check after a var_pointer = malloc()
+#define RETURN_IF_NULL(var_pointer) \
+  if (var_pointer == NULL) {        \
+    return 0;                       \
+  }
```
andrewkozlik, line 147: This should return `-1` I think.

crypto/fuzzer/fuzzer.c
```diff
@@ -96,82 +119,103 @@ void fuzzer_reset_state(void) {
...
+  memcpy(&decimals, fuzzer_input(4), 4);
+  memcpy(&exponent, fuzzer_input(4), 4);
+  trailing = (fuzzer_input(1)[0] & 1);
+
+  // IDEA allow prefix == NULL
+  char *prefix = malloc(prefixlen);
+  // IDEA allow suffix == NULL
+  char *suffix = malloc(suffixlen);
+  RETURN_IF_NULL(prefix);
+  RETURN_IF_NULL(suffix);
```
andrewkozlik, line 198: Technically, if the second `malloc()` fails, then we should free `prefix`.

crypto/Makefile
```diff
@@ -70,12 +70,13 @@ ZKP_CFLAGS = \
...
+# this is specific for 64-bit builds
 CFLAGS += -DSECP256K1_CONTEXT_SIZE=208
 
 # TODO remove this workaround once possible
-ifeq ($(CLANG_AT_LEAST_14),true)
-$(warning "warning: suppressing clang-14 compiler warning for secp256k1-zkp code")
-ZKP_CFLAGS += -Wno-bitwise-instead-of-logical
+ifeq ($(CLANG_AT_LEAST_15),true)
+$(warning "warning: suppressing clang-15 `-Wstrict-prototypes` compiler warning for `zkp_context.c` as workaround")
+zkp_context.o: CFLAGS += -Wno-strict-prototypes
```
andrewkozlik, line 79: Why do we need `no-strict-prototypes`? I don't have clang-15. Can you paste the compiler warning here?
If it's just a matter of changing `int zkp_context_init()` -> `int zkp_context_init(void)` etc., then I think we should fix the function definitions.

crypto/fuzzer/fuzzer.c
invd: Good point. `return -1` is a minor optimization on newer libFuzzer variants to signal the fuzz engine to discard this input. In the current "abort early because something is wrong" role of `RETURN_IF_NULL`, this is meaningful in theory. 

In practice this path is only used once `malloc()` returns `NULL` pointers. If the system is out of new memory to allocate, the fuzzer is in a broken state where we mainly aim to avoid not triggering deadly bugs and performance is not important. The primary edge case that is likely relevant here are `malloc(0)` calls where optimization is meaningful.

crypto/fuzzer/fuzzer.c
invd: Good catch. I've switched to a different ordering of `malloc` and `RETURN_IF_NULL` for code positions with more than one allocation to account for this.

crypto/Makefile
invd: For the crypto fuzzer, I often work around other code issues to move ahead and introduce as few controversial or incomplete changes to the main code as possible. 

When building the crypto fuzzer with clang, only the instances in `zkp_context.c` are seen by the compiler, but there are other instances in other files that should ideally get fixed as well. That's why I opened https://github.com/trezor/trezor-firmware/issues/2360 .

@andrewkozlik I see three options:
1) we stick with the Makefile workaround and fix the issues as part of a different PR
1) I fix `zkp_context.c` as part of this PR and drop the Makefile workaround
1) I fix all instances of this issue as part of this PR and drop the Makefile workaround

Which option do you prefer?

invd: Compiler warning on clang-15 and early clang-16:
```
zkp_context.c:54:21: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
int zkp_context_init() {
                    ^
                     void
zkp_context.c:81:25: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
void zkp_context_destroy() {
                        ^
                         void
zkp_context.c:90:51: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
const secp256k1_context *zkp_context_get_read_only() {
                                                  ^
                                                   void
zkp_context.c:97:48: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
secp256k1_context *zkp_context_acquire_writable() {
                                               ^
                                                void
zkp_context.c:108:34: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
void zkp_context_release_writable() {
                                 ^
                                  void
```

crypto/fuzzer/fuzzer.c
invd: Resolved in https://github.com/trezor/trezor-firmware/pull/2449/commits/66db14e0f96d2a21e39d3ebaab85bbe1365780f7.

crypto/fuzzer/fuzzer.c
invd: Resolved in https://github.com/trezor/trezor-firmware/pull/2449/commits/66db14e0f96d2a21e39d3ebaab85bbe1365780f7.

crypto/Makefile
andrewkozlik: I suggest Option 2.

When I try to build with clang 14.0.0 I get the error below. It's interesting that this doesn't trouble newer versions of the compiler.
```
fuzzer/fuzzer.c: In function ‘fuzz_ecdsa_recover_pub_from_sig_functions’:
fuzzer/fuzzer.c:1020:11: error: variable ‘zero_pubkey’ set but not used [-Werror=unused-but-set-variable]
 1020 |   uint8_t zero_pubkey[65] = {0};
      |           ^~~~~~~~~~~
```

invd: @andrewkozlik : that's curious, I cannot get my clang versions to detect the dead store even with `-Wunused-but-set-variable` and `-Wunused-variable` explicitly enabled. `zero_pubkey` is definitely an unused leftover that I should remove, so it's unexpected that it is not found.

Could you list the exact commands and flags that you used during your build?

crypto/fuzzer/fuzzer.c
invd: Correction: my changes don't help in the described scenario. Checking the `prefix` allocation result before allocating `suffix` does help if there's a problem with the `prefix` allocation, but not if the `prefix` allocation went fine and the `suffix` allocation has a problem.

@andrewkozlik : are you aware of an elegant, generic approach to deal with this in cases of two or more `malloc()` calls? My goal with the  `RETURN_IF_NULL` macro was to make the code more readable and avoid error-prone boilerplate code, but I think there's no escaping of C's less-than-ideal lack of good error cleanup patterns. I wanted to avoid `goto cleanup;` patterns in the code, but since we can't teach the macro generically which pointers it has to free, I don't see good alternatives to it.

invd: The verbose solution is in https://github.com/trezor/trezor-firmware/pull/2449/commits/2c5fcf1ab4c011d9f186de5350f4e759f147ab98.

crypto/Makefile
invd: https://github.com/trezor/trezor-firmware/pull/2449/commits/2c5fcf1ab4c011d9f186de5350f4e759f147ab98 removes the unused variable and some other dead stores.

andrewkozlik: I was wrong, it was actually GCC showing the error. I thought `make fuzzer` automatically uses clang as the default compiler.

andrewkozlik (approved):
utACK

crypto/Makefile
invd: That explains it. Although it's currently not important that GCC can build the fuzzer code since it lacks the fundamental `-fsanitize=fuzzer` libFuzzer support, it's still useful to look at what it complains about. As we've seen, it sometimes catches issues that clang doesn't find.

Therefore I've added https://github.com/trezor/trezor-firmware/pull/2449/commits/dda39e7fc684004334f68226f369a8df676d3cc9 to resolve the remaining code-specific warnings.

I'm aware that `FUZZER=1 make fuzzer` is not very intuitive (although it is [documented](https://github.com/trezor/trezor-firmware/blob/master/crypto/fuzzer/README.md#build)) and am open to suggestions on how to solve the conditional changes to the build more elegantly. In my understanding, target-specific Makefile variables can make the Makefile evaluation more complex if they have a global impact, so I've relied on external configuration of `FUZZER=1` for now that is straightforward to reason about.

crypto/fuzzer/fuzzer.c
andrewkozlik: Unfortunately I don't know of any pretty solution. `goto cleanup;` is fine, but I think we don't use it anywhere else. A sequence of `if (result == SUCCESS) {` blocks is sometimes used or nested `if` blocks would also work well, i.e.:
```c
int ret = 0;
char *out_buffer = malloc(outlen);
if (out_buffer != NULL) {
    char *in_buffer = malloc(inlen);
    if (in_buffer != NULL) {
        // code uses in_buffer and out_buffer here
        free(in_buffer);
    } else {
        ret = -1;
    }
    free(out_buffer);
} else {
    ret = -1;
}
return ret;
```
I'd leave it as is.

andrewkozlik:
I am not sure whether you see this @invd in the GitHub checks:
```
[CLANG-FORMAT]
./crypto/fuzzer/fuzzer.c ... 5 changes needed
make: *** [Makefile:73: cstyle_check] Error 1
Error: Process completed with exit code 2.
```
Run `make style`.

invd:
@andrewkozlik fixed via https://github.com/trezor/trezor-firmware/pull/2449/commits/99706492a32b9dd6fc8ac00779ef202beff3fc8d.

PR #2450, opened 2022-08-15
Small Rust size improvements

matejcik:
This PR modifies _all_ geometry operations to be `const`.

In addition, it introduces a nicer error handling: replace all `something.unwrap()` with `unwrap!(something)` to get information about file, line, function, all without invoking the built-in panic machinery.

There are also convenience macros `ensure!()` and `fatal_error!()` that are currently unused.

Open point: should we convert every panic! to this? My thought is probably yes, but that will also lose all traceback information.


grdddj:
I am OK with these changes, it seems to save `512 bytes` (output from `bin_size_tool` below).

In the `model R`, I am currently using `.unwrap()` quite a lot, mostly when appending to `String`s or `Vec`s - `String.push_str("xx").unwrap()`. Having `unwrap!(String.push_str("xx"))` does not look prettiest, but not a big deal.

```
.flash     788     1    L788     D0       [section .flash]
.flash     80      1    L80      D0       embed/rust/src/trezorhal/common.rs                           __fatal_error()
.flash     56      1    L56      D0       core::result::Result<T,E>::unwrap_or_fatal_error()
.flash     48      1    L48      D0       embed/rust/src/trezorhal/common.rs                           __fatal_error::as_cstr_buf()
.flash     40      1    L40      D0       embed/rust/src/protobuf/obj.rs                               msg_def_obj_attr()
.flash     36      1    L36      D0       embed/rust/src/micropython/map.rs                            set_obj()
.flash     32      1    L32      D0       embed/rust/src/micropython/buffer.rs                         StrBuffer::as_ref()
.flash     30      1    L30      D0       rust_begin_unwind
.flash     16      1    L16      D0       embed/rust/src/ui/model_tr/component/button.rs               Button<T>::place()
.flash     14      1    L14      D0       core::convert::Into<U>::into()
.flash     12      1    L12      D0       embed/rust/src/protobuf/decode.rs                            Decoder::decode_defaults_into()
.flash     8       2    L8       D0       embed/rust/src/ui/geometry.rs                                Rect::split_bottom()
.flash     8       2    L8       D0       embed/rust/src/ui/component/base.rs                          Child<T>::obj_place()
.flash     6       2    L6       D0       embed/rust/src/ui/geometry.rs                                Rect::split_right()
.flash     6       1    L6       D0       compiler_builtins::arm::__aeabi_memclr()
.flash     4       1    L4       D0       __aeabi_memclr
.flash     4       1    L4       D0       embed/rust/src/ui/component/text/layout.rs                   TextLayout::layout_text()
.flash     0       2    L0       D0       embed/rust/src/ui/geometry.rs                                Rect::center()
.flash     -2      1    L-2      D0       core::ptr::drop_in_place<core::str::error::Utf8Error>()
.flash     -4      1    L-4      D0       __aeabi_memset4
.flash     -8      27   L-8      D0       OUTLINED_FUNCTION
.flash     -28     1    L-28     D0       embed/rust/src/trezorhal/common.rs                           fatal_error()
.flash     -40     1    L-40     D0       core::cmp::Ord::clamp()
.flash     -48     1    L-48     D0       embed/rust/src/ui/geometry.rs                                Rect::split_top()
.flash     -50     1    L-50     D0       embed/rust/src/ui/component/text/paragraphs.rs               Paragraphs<T>::change_offset()
.flash     -50     1    L-50     D0       embed/rust/src/ui/geometry.rs                                Rect::split_left()
.flash     -170    1    L-170    D0       core::str::error::Utf8Error::fmt()
.flash     -310    1    L-310    D0       core::fmt::builders::DebugStruct::field()
.flash     -410    3    L-410    D0       core::fmt::Debug::fmt()
.flash     -580    1    L-580    D0       core::fmt::num::_<impl core::fmt::Debug for u32>::fmt()
SUMMARY: 30 rows, -512 bytes in total (L-512 D0).
```

matejcik:
the size saving is not expected to be that big in firmware, but it adds up in bootloader -- esp. with a separate feature "panic-immediate-abort" .... which I originally didn't intend to include here because I meant for it to be only in bootloader. But it also makes sense for firmware. So I'll add it too.

core/embed/rust/src/ui/geometry.rs
```diff
@@ -503,40 +539,45 @@ pub struct LinearPlacement {
...
+        Self {
+            align: Alignment::Center,
+            ..self
+        }
     }
 
-    pub fn align_at_end(mut self) -> Self {
-        self.align = Alignment::End;
-        self
+    pub fn align_at_end(self) -> Self {
```
TychoVrahe, line 572: const missing, also in Rect::split_right

core/embed/rust/src/lib.rs
```diff
@@ -51,17 +42,13 @@ fn panic_debug(panic_info: &core::panic::PanicInfo) -> ! {
...
-    // security reasons, as we do not always controls the message contents. We
-    // should also avoid printing "panic" or "rust" on the user screen to avoid any
-    // confusion.
-
-    // SAFETY: Safe because we are passing in \0-terminated literals.
-    let empty = unsafe { CStr::from_bytes_with_nul_unchecked(b"\0") };
-    let msg = unsafe { CStr::from_bytes_with_nul_unchecked(b"rs\0") };
-
-    trezorhal::common::fatal_error(empty, msg, empty, 0, empty);
+    // TODO: as of 2022, ignoring the `_info` parameter does not help with saving
```
TychoVrahe, line 45: maybe state compiler version instead/in addition to year

TychoVrahe (commented):
Left two comments. 

In my case, using this in bootloader produced cca 5.5kB of savings which is quite good. (though without converting unwrap to the macro it has no observable impact -maybe something below 512B alignment- so the saving is mostly the fmt machinery not being linked in)

core/embed/rust/src/lib.rs
matejcik: fixed in 863504278

core/embed/rust/src/ui/geometry.rs
matejcik: fixed in ab2ba1947

TychoVrahe (approved):

PR #2459, opened 2022-08-18
ci: fix data deploy job for release

vdovhanych:
Fixes the job deploy emulators to aws which causes jobs on tag to not run, this should solve the error.


matejcik (approved):
utACK

PR #2460, opened 2022-08-18
Switch to nightly Rust 

matejcik:
(again.)

We require nightly Rust compiler for the `panic_immediate_abort` feature of core libraries. This feature will compile all panic!s into a single fault instruction, which saves significant amount of space.

For now, we are not using any nightly code features. But now that we need nightly Rust anyway, we can start.

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -3,7 +3,7 @@
  }:
 
 let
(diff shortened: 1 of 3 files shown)


matejcik:
superseded by #2461 which is a branch rename only

PR #2461, opened 2022-08-18
Switch to nightly Rust 

matejcik:
We require nightly Rust compiler for the panic_immediate_abort feature of core libraries. This feature will compile all panic!s into a single fault instruction, which saves significant amount of space.

For now, we are not using any nightly code features. But now that we need nightly Rust anyway, we can start.

supersedes #2460 , branch rename so that we get HW tests on this

### DIFF
--- a/ci/shell.nix
+++ b/ci/shell.nix
@@ -3,7 +3,7 @@
  }:
 
 let
-  # the last commit from master as of 2022-02-08
+  # the last commit from master as of 2022-08-02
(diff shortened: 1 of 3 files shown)


mmilata (approved):
LGTM but I'd try rerunning https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2901819773 to see if it gets stuck on the same test

matejcik:
hardware test went well

Issue #2432, opened 2022-08-02
Run device tests in parallel


matejcik:
A single run of the device test suite takes ~10 minutes for trezor-core, or ~3 minutes for legacy. During this time, most of your CPU cores are idling. There is a low-hanging-fruit speedup to be gained by parallelizing the test suite.

Packages like [pytest-xdist](https://pypi.org/project/pytest-xdist/) can do that basically for free.

This is Not That Easy for a bunch of solvable reasons:
* Currently the tests do not launch their own emulator. (Also, for hardware tests, we don't _want_ tests to launch their own emulator.) This means only one instance is running, so only one test can be executed.
  We would need to add an ability to the test harness to launch multiple emulators on non-conflicting ports with non-conflicting profile directories (and separate UI test output).
* Certain parts of UI test collection rely on having all the data from the test run available. We would need to improve the whole thing.
* ??? more?

PR #2462, opened 2022-08-18
feat(tests): add possibility to run device/UI tests in parallel

grdddj:
Addresses https://github.com/trezor/trezor-firmware/issues/2432:
- add options to run device tests in parallel, with `make` commands like `test_emu_multicore` or `test_emu_ui_record_multicore`
- is for local purposes only, the `CI` still runs the "normal" serial tests

It does it using the suggested [pytest-xdist](https://pypi.org/project/pytest-xdist/), and running `pytest` with `-n auto` for using as many processes as possible.

The whole test suite finished in under 5 minutes for the simplest `test_emu_multicore`. The other versions including `UI` checks take longer.

Seems to be working fine with the `UI` tests, no changes needed to be done on that front.

Theoretically it can slow down the "normal" tests, as now the `_raw_client` fixture is not valid for the whole session, but is special/needs to be called by each function. That is because it will be called by multiple processes and each will need to be given its own `Client` available on a certain port.

There are probably still some small discrepancies when running the whole pipeline, I got occasional errors of `Failed to open debuglink for udp:127.0.0.1:XXXXX`, mostly for `wipe code` tests. However, that happened just 6 times out of more than 2000 test cases I ran with it.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -88,6 +88,9 @@ test_rust: ## run rs unit tests
 test_emu: ## run selected device tests from python-trezor
 	$(EMU_TEST) $(PYTEST) $(TESTPATH)/device_tests $(TESTOPTS)
 
+test_emu_multicore: ## run device tests using multiple cores
+	$(PYTEST) -n auto $(TESTPATH)/device_tests $(TESTOPTS) --control-emulators --model=core --random-order-seed=$(shell echo $$RANDOM)
+
 test_emu_monero: ## run selected monero device tests from monero-agent
 	cd tests ; $(EMU_TEST) ./run_tests_device_emu_monero.sh $(TESTOPTS)
 
@@ -105,8 +108,14 @@ test_emu_click: ## run click tests
 test_emu_ui: ## run ui integration tests
 	UI2="$(UI2)" $(EMU_TEST) $(PYTEST) $(TESTPATH)/device_tests --ui=test --ui-check-missing $(TESTOPTS)
 
+test_emu_ui_multicore: ## run ui integration tests using multiple cores
+	UI2="$(UI2)" $(PYTEST) -n auto $(TESTPATH)/device_tests $(TESTOPTS) --ui=test --ui-check-missing --control-emulators --model=core --random-order-seed=$(shell echo $$RANDOM)
+
 test_emu_ui_record: ## record and hash screens for ui integration tests
 	UI2="$(UI2)" $(EMU_TEST) $(PYTEST) $(TESTPATH)/device_tests --ui=record --ui-check-missing $(TESTOPTS)
+
+test_emu_ui_record_multicore: ## record and hash screens for ui integration tests using multiple cores
+	UI2="$(UI2)" $(PYTEST) -n auto $(TESTPATH)/device_tests $(TESTOPTS) --ui=record --ui-check-missing --control-emulators --model=core --random-order-seed=$(shell echo $$RANDOM)
 
 pylint: ## run pylint on application sources and tests
 	pylint -E $(shell find src tests -name *.py)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,8 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
+from __future__ import annotations
+
 import os
 from typing import TYPE_CHECKING, Generator
 
@@ -26,9 +28,11 @@ from trezorlib.transport import enumerate_devices, get_transport
 
 from . import ui_tests
 from .device_handler import BackgroundDeviceHandler
+from .emulators import EmulatorWrapper
 from .ui_tests.reporting import testreport
 
 if TYPE_CHECKING:
+    from trezorlib._internal.emulator import Emulator
     from _pytest.config import Config
     from _pytest.config.argparsing import Parser
     from _pytest.terminal import TerminalReporter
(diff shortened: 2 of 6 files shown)


matejcik:
ISTM `emulator` should be a session-scope fixture (see also some ideas in #2469), and `_raw_client` should remain a session-scope fixture that will (optionally, based on the "control-emulators" parameter) request the `emulator` fixture to get its clients

you should probably be using `EmulatorWrapper` to get the `Emulator` instance, this also takes care of generating profile-dir as a tempdir that is auto-cleaned afterwards

tests/conftest.py
```diff
@@ -36,29 +42,156 @@
...
-            transport = get_transport(path)
-            return Client(transport, auto_interact=not interact)
-        except Exception as e:
-            request.session.shouldstop = "Failed to communicate with Trezor"
-            raise RuntimeError(f"Failed to open debuglink for {path}") from e
-
+    # In case tests run in parallel, each process has its own path to the emulator.
+    if EMU is not None:
+        path = f"udp:127.0.0.1:{EMU.get_port()}"
+        return _client_from_path(request, path, interact)
```
matejcik, line 165: `Emulator` will set up `emu.client` for you which you can return directly -- but perhaps you'll need to add explicit `auto_interact` argument to `Emulator` so that the `debug` value is decoupled from it. We want to be able to run emulator in debug mode (to get console output) but keep manual interaction.

grdddj: Done in [00aa807](https://github.com/trezor/trezor-firmware/pull/2462/commits/00aa8071e1d9f549436b90ea80bbf6d813bfb691)

grdddj:
> ISTM `emulator` should be a session-scope fixture (see also some ideas in #2469), and `_raw_client` should remain a session-scope fixture that will (optionally, based on the "control-emulators" parameter) request the `emulator` fixture to get its clients
> 
> you should probably be using `EmulatorWrapper` to get the `Emulator` instance, this also takes care of generating profile-dir as a tempdir that is auto-cleaned afterwards

Done in [af78f86](https://github.com/trezor/trezor-firmware/pull/2462/commits/af78f86caec11cd7e1de78a6f7d9f2b768671f0f) - it was possible to simplify it quite a lot using an `emulator` fixture instead of a class together with `EmulatorWrapper`. All the setup and teardown logic now happens inside the fixture, no need to call it in `pytest_session_start` or `pytest_session_finish`

core/Makefile
```diff
@@ -88,6 +88,9 @@ test_rust: ## run rs unit tests
 test_emu: ## run selected device tests from python-trezor
 	$(EMU_TEST) $(PYTEST) $(TESTPATH)/device_tests $(TESTOPTS)
 
+test_emu_multicore: ## run device tests using multiple cores
+	$(PYTEST) -n auto $(TESTPATH)/device_tests $(TESTOPTS) --control-emulators --model=core --random-order-seed=$(shell echo $$RANDOM)
```
matejcik, line 92: is the explicit random seed so that all runners have the same one?
what happens when you leave it out?

grdddj: Yes, otherwise they would collect the tests in a different order and the master runner would complain about it.

Leaving it out produces:
```
pytest -n auto ../tests/device_tests/nem --control-emulators --model=core
============================================================================= test session starts ==============================================================================
platform linux -- Python 3.9.13, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
Using --random-order-bucket=module
Using --random-order-seed=20108

rootdir: /home/jmusil/trezor-firmware, configfile: setup.cfg
plugins: flaky-3.7.0, xdist-2.5.0, forked-1.4.0, ordering-0.6, random-order-1.0.4, asyncio-0.16.0, hypothesis-6.36.1, timeout-2.1.0
gw0 [17] / gw1 [17] / gw2 [17] / gw3 [17]
collecting 0 items / 1 error                                                                                                                                                   
==================================================================================== ERRORS ====================================================================================
_____________________________________________________________________________ ERROR collecting gw1 _____________________________________________________________________________
Different tests were collected between gw0 and gw1. The difference is:
--- gw0

+++ gw1

@@ -1,17 +1,17 @@

-tests/device_tests/nem/test_signtx_mosaics.py::test_nem_signtx_mosaic_supply_change
-tests/device_tests/nem/test_signtx_mosaics.py::test_nem_signtx_mosaic_creation_properties
....
```

tests/conftest.py
```diff
@@ -38,27 +42,91 @@
...
+            "Legacy emulator is not supported until it can be run on arbitrary ports."
+        )
+
+    def _get_port() -> int:
+        """Get a unique port for this worker process on which it can run.
+
+        Guarantees to be unique because each worker has a different name.
+        gw0=>20000, gw1=>20111, gw2=>20222, etc.
+
+        Assumes there will not be more than 10 workers.
```
matejcik, line 80: me and my 16 cores don't like this ;) 

how about instead:
```
assert worker_id.startswith("gw")
return 20000 + int(worker_id[2:]) * 3
```

matejcik (changes requested):

tests/conftest.py
matejcik: to expand: `*3` numerically because the emulator eats 3 consecutive ports: Normal link, Debug link, Webauthn fake interface.

grdddj: Works for me. Done in [55c719b](https://github.com/trezor/trezor-firmware/pull/2462/commits/55c719b512eaa92b53a42975378af244bcf68955)

matejcik (approved):
lgtm, please squash & merge.

i'm wondering if something can be done about the explicit seed though... would be nice if either pytest-xdist or pytest-random-order handled this

PR #2467, opened 2022-08-23
Fix uzlib window allocation in rust

TychoVrahe:
Fixing window allocation problem in Uzlib. Pins window location on stack.

### DIFF
--- a/core/embed/rust/src/trezorhal/uzlib.rs
+++ b/core/embed/rust/src/trezorhal/uzlib.rs
@@ -13,25 +13,25 @@ impl Default for ffi::uzlib_uncomp {
 pub struct UzlibContext<'a> {
     uncomp: ffi::uzlib_uncomp,
     src_data: PhantomData<&'a [u8]>,
-    window: Option<[u8; UZLIB_WINDOW_SIZE]>,
 }
 
 impl<'a> UzlibContext<'a> {
-    pub fn new(src: &'a [u8], use_window: bool) -> Self {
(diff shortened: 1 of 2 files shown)


matejcik:
@jpochyla care to take a look?

jpochyla:
@matejcik Uuuuuh I'm not sure what I was thinking there, but now that I look at it: with the window decoupled to a separate struct like this, and kept on stack, I think `Pin` is actually not needed here and we can use `&mut`.

matejcik:
it's kind of unwieldy though

another option would be to re-set the window pointer every time we call uncompress

TychoVrahe:
if we don't use Pin, we don't even need the window struct, just a slice in the caller? we need to never return the UzlibContext out of the caller function, otherwise it fails again (which is probably true even with the pin). 

resetting the pointer would discard the work done by C inside the window, wouldn't it?

I went for decoupling of the window because we can save some stack space if we don't need it. And also i couldn't make it compile when trying to Pin the whole struct :]

matejcik:
> we need to never return the UzlibContext out of the caller function, otherwise it fails again (which is probably true even with the pin).

add the `'a` lifetime to the window parameter and Rust won't let you return

> resetting the pointer would discard the work done by C inside the window, wouldn't it?

as long as Rust is actually moving the data -- which it should be -- then no, the pointer and the index into the window is separate

TychoVrahe:
second attempt: removed the pin, using slice in the caller and &mut.

matejcik (approved):

matejcik:
LGTM, please squash fixups & merge

PR #2468, opened 2022-08-23
refactor(core/rust/ui): text theme struct

mmilata:
Remove DefaultTextTheme brain damage.


core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -23,6 +23,7 @@ pub struct Paragraphs<T> {
     area: Rect,
     list: Vec<Paragraph<T>, MAX_PARAGRAPHS>,
     placement: LinearPlacement,
+    theme: TextTheme,
```
mmilata, line 26: Using `&'static TextTheme` would make the struct smaller but not sure if it's not premature optimization.

mmilata:
Updated as discussed with @matejcik in [cdaff88](https://github.com/trezor/trezor-firmware/pull/2468/commits/cdaff88e40c4f041f456cfd28bfef816585ed55d). Removes separate font setting for hyphens which causes [visual differences in line breaking](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2929570788/artifacts/master_diff/index.html). 

In `mmilata/text-theme-ref` @ e4f0bf4 I've tried putting the text style behind a reference to make the structs (e.g. Paragraphs) smaller. Due to `layout_ops` this causes bunch of our types to include `<'a>` which looks ugly so I'd revisit this again after we switch to the iterator based implementation?

matejcik:
> causes [visual differences in line breaking](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2929570788/artifacts/master_diff/index.html).

This affects only the monospace font, right?
Going forward I'd like the monospace theme to break wherever and not insert hyphens. I'll do that in the iterator PR, it would currently need a new `LineBreaking` entry

mmilata:
Previously (on TT) the hyphen font used for everything was BOLD, so it affects everything but the bold font. Agree with no hyphens for addresses etc.

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -419,8 +417,8 @@ async def confirm_output(
     color_to: int = ui.FG,  # TODO cleanup @ redesign
     to_str: str = " to\n",  # TODO cleanup @ redesign
     to_paginated: bool = False,  # TODO cleanup @ redesign
-    width: int = MONO_ADDR_PER_LINE,
-    width_paginated: int = MONO_ADDR_PER_LINE - 1,
+    width: int = 0,  # TODO cleanup @ redesign
```
matejcik, line 420: what's this about?

mmilata: The constants are imported from `...constants.tt` which we don't have in the frozen build.

core/embed/rust/src/ui/component/text/paragraphs.rs
mmilata: If we make the `TextStyle` reference in `TextLayout` &'static then `layout_ops` breaks because it mutates it. I don't see a simple fix aside from removing the reference from `TextLayout` entirely and pass it around separately.

matejcik: just make a copy of the font and color and mutate locally.
with the way `Op::skip_n_text_bytes` you should always get all the format mutating instructions at start, so the state of the locals will be the right thing before you get to any text

mmilata: I don't understand what you mean by mutate locally. `layout_ops` calls `layout_text` which reads the font and color from `TextLayout`. If we can't make up styles on the fly then we need to modify `layout_text` to accept font and color as a parameter?

matejcik: oooh right. OK let's leave that to the iterator PR

matejcik (approved):

Issue #2442, opened 2022-08-08
Fee rate not displayed when replacing transaction


matejcik:
**Describe the bug**
#1294 implements fee rate for "sign transaction" flow.
Fee rate should probably be displayed on "modify fee" screen as well.

screenshots:
https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2826526641/artifacts/test_ui_report/passed/TT_bitcoin-test_signtx_replacement.py::test_p2tr_fee_bump.html

sime:
We have the space to include the fee rate?

Concept: https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?node-id=3539%3A10964

matejcik:
this design is for the "total for transaction" screen, not for "modify fee" when replacing. screenshots: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/2930791419/artifacts/test_ui_report/passed/TT_bitcoin-test_signtx_replacement.py::test_p2pkh_fee_bump.html
relevant screen copied so it doesn't expire again:
![image](https://user-images.githubusercontent.com/596073/186647139-9afd6acf-772b-4923-9d6e-d4e91057b61e.png)

with that said, the same concept should work here -- the screen fits exactly _one_ additional line, so if the fee rate does not have a heading of its own, it will work fine

grdddj:
`QA` testing instructions:
- modify fee/replace transaction in `Suite` (or anywhere else) and check whether the `Modify fee` screen shows a fee rate in brackets below transaction fee - for example `(66.4 sat/vB)`

![image](https://user-images.githubusercontent.com/42543243/187422323-41ce3d3f-1d2f-43d6-9b36-8e94876bc0af.png)

grdddj:
We might consider doing it also for `legacy`, the fee-rate is not shown there when replacing.

There is also one free row left, where we could fit it. Question is whether it is worth the effort...

![image](https://user-images.githubusercontent.com/42543243/187440941-0e481ade-f623-45e1-8d7e-ba30788460d9.png)

matejcik:
this issue is also tagged legacy so I believe that applies, yes

grdddj:
> this issue is also tagged legacy so I believe that applies, yes

Done in https://github.com/trezor/trezor-firmware/pull/2481

STew790:
**QA OK**

Trezor T:
[...]

PR #2473, opened 2022-08-29
feat(core): show fee rate when replacing transaction

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2442:
- show fee rate when the transaction is replaced/when the fee is modified


matejcik (commented):
LGTM, waiting for green CI to see UI diff

matejcik:
(the failure in legacy is unrelated and I have a suspicion why it's there, but hopefully it will go away after the re-run)

matejcik (approved):

PR #2475, opened 2022-08-30
Do not show "USB NOT CONNECTED" if device enters USB suspend state

hiviah:
Linux has by default 2 seond autosuspend for USB devices. This means on PCs with autosuspend supported, TT showed "USB NOT CONNECTED" when entering suspend.

We now treat switching to USB suspended from configured state as still connected.

This will still work for people plugging TT into charger or using cable without data pins to alert them.

### DIFF
--- a/core/embed/trezorhal/usb.c
+++ b/core/embed/trezorhal/usb.c
@@ -153,8 +153,16 @@ void usb_start(void) { USBD_Start(&usb_dev_handle); }
 void usb_stop(void) { USBD_Stop(&usb_dev_handle); }
 
 secbool usb_configured(void) {
-  USBD_HandleTypeDef *pdev = &usb_dev_handle;
+  const USBD_HandleTypeDef *pdev = &usb_dev_handle;
   if (pdev->dev_state == USBD_STATE_CONFIGURED) {
+    return sectrue;
+  }
+
+  // Linux has autosuspend device after 2 seconds by default.
+  // So a suspended device that was seen as configured is reported as
+  // configured.
+  if (pdev->dev_state == USBD_STATE_SUSPENDED &&
+      pdev->dev_old_state == USBD_STATE_CONFIGURED) {
     return sectrue;
   }
 


prusnak:
Since you are touching that code, can we maybe change

``` c
 USBD_HandleTypeDef *pdev = &usb_dev_handle;
```

into

``` c
 const USBD_HandleTypeDef *pdev = &usb_dev_handle;
```

hiviah:
Added the const-ness.

prusnak (approved):

core/.changelog.d/2475.fixed
```diff
@@ -0,0 +1 @@
+Do not show "USB NOT CONNECTED" if device enters USB suspend state from configured state
```
matejcik, line 1: please remove this (or reword the previous changelog entry, if needed)

this way it would needlessly spam the changelog for a feature that has not changed between releases

core/embed/trezorhal/usb.c
```diff
@@ -153,11 +153,19 @@ void usb_start(void) { USBD_Start(&usb_dev_handle); }
...
+  const USBD_HandleTypeDef *pdev = &usb_dev_handle;
   if (pdev->dev_state == USBD_STATE_CONFIGURED) {
     return sectrue;
   }
 
+  // Linux has autosuspend device after 2 seconds by default.
+  // So a suspended device that was seen as configured is reported as
+  // configured.
+  if (pdev->dev_state == USBD_STATE_SUSPENDED &&
+      pdev->dev_old_state == USBD_STATE_CONFIGURED) {
```
matejcik, line 165: at what point will `dev_old_state` get pushed out?
i.e., can we get suspended and then suspended again and push out the "previously configured" state?

core/.changelog.d/2475.fixed
hiviah: I'll add [no changelog] instead.

core/embed/trezorhal/usb.c
hiviah: You can't get "doubly suspended", you can enter from configured->suspended just once. Then again suspended->configured and back, but not suspended->suspended.

Here is the state machine:

image

hiviah:
Removed the changelog entry.

hiviah:
Not sure why two tests fail, no code is touched that could affect signatures or addresses.

Issue #2190, opened 2022-03-28
Do not convert bech32 addresses to upper-case for QR code (legacy)


hynek-jina:
Follow-up from https://github.com/trezor/trezor-firmware/issues/1751 fro Trezor One.

hynek-jina:
Currently T1 shows QR code with upper-case address (ie `BC1QUQGQ44WQ0ZJH6D920ZS42NSY4N4EV5VT8NXKE4`)
<img width="284" alt="Snímek obrazovky 2022-03-30 v 10 59 05" src="https://user-images.githubusercontent.com/26002916/160806177-2198bf06-9854-45de-a6ae-e7120fac9835.png">

While Suite shows different QR code with lower case address (ie `bc1quqgq44wq0zjh6d920zs42nsy4n4ev5vt8nxke4`)
<img width="1348" alt="Snímek obrazovky 2022-03-30 v 10 59 03" src="https://user-images.githubusercontent.com/26002916/160806327-51ab00a4-6d28-4d26-b12d-b33b77437c0c.png">

prusnak:
Will the Taproot QR code fit on a legacy screen if the address is not upper-cased? Let's test this (Taproot addresses are longer than Segwit v0).

Midar:
Isn't this more a bug in Suite that it's *not* uppercase there? I thought one of the ideas of BECH32 was that you all-uppercase the QR code to make it smaller?

hynek-jina:
Build the custom version of model one for QA to test the readability of QR code

grdddj:
I do not have good news here, the `taproot` address in lowercase does not fit in `V3` QR code (29x29), but requires `V4` (33x33). With our screen-height of `64 px`, it is not possible to fit `33*2` pixels on the screen, therefore it uses only one pixel per module, which is really badly visible.

(Very theoretically, we could display that 66x66 px QR code when the very top and very bottom row would be missing on the screen, it could still scan well given the error-corrections of QR codes.)

Below is the screenshot with magnified (`TREZOR_OLED_SCALE=2`) emulator in `Suite`.

On a hardware device, I was not able to scan this miniature `QR code`.

[...]

PR #2476, opened 2022-08-30
Not converting bech32 addresses to uppercase in legacy QR codes

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2190:
- do not convert `bech32` addresses to uppercase in `legacy` QR codes
- also adding an exception for (mostly) `taproot` addresses, which are too big to fit the screen - not showing last row in that case, which is fine thanks to the QR code's error correction


matejcik:
please also take into account https://github.com/trezor/trezor-firmware/issues/2190#issuecomment-1233464279, see if maybe we should handle taproot differently?

legacy/firmware/layout2.c
```diff
@@ -781,8 +781,12 @@ void layoutAddress(const char *address, const char *desc, bool qrcode,
...
 
     oledInvert(0, 0, 63, 63);
-    if (side > 0 && side <= 29) {
-      int offset = 32 - side;
+    if (side > 0 && side <= 33) {
+      // For QR code V4 (33x33), still maintaining two pixels per module
+      // so that it is nicely readable. Not showing the bottom row,
+      // which is not a problem thanks to the QR code's error correction.
+      // It is a use-case for BTC taproot addresses.
+      int offset = side <= 29 ? 32 - side : 0;
```
mmilata, line 789: Did you try using -2 instead of 0? Might look nicer aligned to center and we'd lose two half-rows instead of single full row:thinking:

grdddj: I tried `-1` for the `offset`, but the issue is that it also creates `x` offset, so even the left half-row is cut-off this way. Surely we could have two offsets, one for `x` (0) and one for `y` (-1), but that would complicate the code unnecessarily IMO

mmilata (approved):

Issue #2295, opened 2022-05-26
Update Connect tests, account for its move to Suite monorepo


grdddj:
`Connect` tests are being moved into Suite monorepo.

We should account for this in our way of running them - https://github.com/trezor/trezor-firmware/blob/master/tests/connect_tests/connect_tests.sh

sime:
Could it (or should it?) potentially support both?

Since Connect v8 will be support for some time.

grdddj:
> Could it (or should it?) potentially support both?
> 
> Since Connect v8 will be support for some time.

Not sure at all. I am not knowledgeable enough about `Connect`, I just heard it moves to another place, so our `Connect` tests would no longer be updated

sime:
@mroz22 @szymonlesisz Thoughts on how to maintain Connect tests in firmware?

grdddj:
Already done, `Connect` tests are currently fine in `master`

PR #2480, opened 2022-08-31
Migrate connect tests

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2295:
- updating `connect` tests to account for their move into `Suite` monorepo


matejcik (commented):
utACK if the CI passes, maybe @mroz22 can take a look if this seems right?

matejcik (approved):

PR #2481, opened 2022-08-31
feat(legacy): show fee rate when replacing transaction

grdddj:
Addresses https://github.com/trezor/trezor-firmware/issues/2442 for `legacy`:
- showing fee rate when replacing transaction for `model 1`


legacy/firmware/layout2.c
```diff
@@ -586,7 +586,9 @@ void layoutConfirmModifyOutput(const CoinInfo *coin, AmountUnit amount_unit,
 }
 
 void layoutConfirmModifyFee(const CoinInfo *coin, AmountUnit amount_unit,
-                            uint64_t fee_old, uint64_t fee_new) {
+                            uint64_t fee_old, uint64_t fee_new,
+                            uint64_t total_in, uint64_t total_out,
```
mmilata, line 590: `fee_new == total_in - total_out`, no need to pass these two

mmilata (changes requested):
Some of the arguments aren't required, otherwise lgtm.

legacy/firmware/layout2.c
grdddj: Good catch, it simplifies stuff a bit. Done in [3cfe21d](https://github.com/trezor/trezor-firmware/pull/2481/commits/3cfe21dd18626d57d1326b9f9ca1c79f0f88e33a)

mmilata (approved):
LGTM!

PR #2483, opened 2022-09-05
Enable bus fault and usage fault handlers in core

TychoVrahe:
The handlers were already there, this just enables them, which helps debugging the faults as you get different error message (hard fault occurs otherwise)

### DIFF
--- a/core/embed/firmware/main.c
+++ b/core/embed/firmware/main.c
@@ -96,6 +96,11 @@ int main(void) {
   // Init peripherals
   pendsv_init();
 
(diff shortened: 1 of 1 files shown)


prusnak:
Let's enable only for debug builds (`!PRODUCTION`) maybe?

TychoVrahe:
added condition for non-produciton builds: https://github.com/trezor/trezor-firmware/pull/2483/commits/63189c39f53e039f986ad0e73e4480f8015bc784

hiviah (approved):

PR #2485, opened 2022-09-05
StrBuffer::alloc

matejcik:
This PR adds the possibility of allocating strings on the GC heap -- useful whenever we need to perform some string manipulations and pass the results into a `StrBuffer`-based UI structure.

Example:
```rust
let somestr = String::<50>::new();
somestr.push_str("hello");
somestr.push_str("world");
let strbuf = StrBuffer::alloc(somestr.as_str())?;
```

---

This also removes the `Buffer` struct, because it is difficult to build it safely:
- in case of python `memoryview`s, an interior pointer would be stored as `ptr`
- it is difficult to get to the head pointer, because `get_buffer_info` does not return it and the original `Obj` is not it: the data pointer from, say, a `bytearray` object, might get reallocated between `Buffer` construction and use, so the `Buffer` stores stale pointer data.

If the two possibilities above happen at the same time, we would end up holding a stale interior pointer for `ptr`.

While it _seems_ that the above can't actually happen in practice, I have no guarantees. And doing this correctly would entail inspecting the passed in MicroPython type and manually extracting the correct head pointer, which I would generally like to avoid -- or calling `mp_get_buffer` at dereference time.

Given the general unsafety of manipulating MicroPython-provided buffers, I opted to remove `Buffer` and replace it by two `unsafe` functions `get_buffer` and `get_buffer_mut`. It is up to the caller to use them responsibly, with no claim to safety.

### DIFF
--- a/core/embed/rust/src/micropython/buffer.rs
+++ b/core/embed/rust/src/micropython/buffer.rs
@@ -1,151 +1,64 @@
-use core::{
-    convert::TryFrom,
-    ops::{Deref, DerefMut},
-    ptr, slice, str,
-};
+use core::{convert::TryFrom, ops::Deref, ptr, slice, str};
 
 use crate::{error::Error, micropython::obj::Obj};
 
 use super::ffi;
 
-/// Represents an immutable slice of bytes stored on the MicroPython heap and
-/// owned by values that obey the `MP_BUFFER_READ` buffer protocol, such as
-/// `bytes`, `str`, `bytearray` or `memoryview`.
+/// Represents an immutable UTF-8 string managed by MicroPython GC.
+/// This either means static data, or a valid GC object.
 ///
 /// # Safety
 ///
-/// In most cases, it is unsound to store `Buffer` values in a GC-unreachable
-/// location, such as static data. It is also unsound to let the contents be
-/// modified while a reference to them is being held.
-pub struct Buffer {
+/// We assume that MicroPython is handling strings according to Python
+/// semantics, i.e., that string data is immutable.
+/// Furthermore, we assume that string data is always either static or a GC head
+/// pointer, i.e., we can never obtain a pointer into the middle of a GC string.
+///
+/// Given the above assumptions about MicroPython strings, working with
+/// StrBuffers in Rust is safe.
+pub struct StrBuffer {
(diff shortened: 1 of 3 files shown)


core/embed/rust/src/micropython/buffer.rs
```diff
@@ -1,151 +1,61 @@
...
         }
     }
-}
 
-impl<const N: usize> From<&'static [u8; N]> for Buffer {
-    fn from(val: &'static [u8; N]) -> Self {
-        Buffer {
-            ptr: val.as_ptr(),
-            len: val.len(),
+    fn get_bytes(&self) -> &[u8] {
```
jpochyla, line 47: This would be usually named `as_bytes`.

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -1,151 +1,61 @@
...
-    fn from(val: &'static [u8]) -> Self {
-        Buffer {
-            ptr: val.as_ptr(),
-            len: val.len(),
+    pub fn alloc(val: &str) -> Result<Self, Error> {
+        // SAFETY:
+        // We assume that if `gc_alloc` returns successfully, the result is a valid
+        // pointer to GC-controlled memory of at least `val.len()` bytes.
+        unsafe {
+            let raw = ffi::gc_alloc(val.len(), 0) as *mut u8;
```
jpochyla, line 34: I'm not sure now, but IIRC uPy has some assumptions about strings being zero-terminated?

matejcik: True, but this string will never go back to uPy (except through `TryFrom<&str> for Obj` which makes a copy)

matejcik: that said, allocating +1 and writing a null terminator is cheap and makes it less error prone for any future modifications so :woman_shrugging:

matejcik: done in 79da7b3

core/embed/rust/src/micropython/buffer.rs
matejcik: done in 79da7b3

mmilata (approved):

PR #2496, opened 2022-09-08
ci(connect): exclude checkFirmwareAuthenticity test

mroz22:
This test expect actual firmware to have been released. At the moment version is set to 2.5.3 but it has not been released yet. So I suggest that we exclude this test from running in firmware repo.


matejcik:
this won't work on an emulator anyway, no?

mroz22:
I thought so, so the fixture expects "fail successfully" scenario, firmware hashes do not match, but connect returns some result.

matejcik (approved):

Issue #2493, opened 2022-09-07
tests: --ui=record does not work with multicore testing


matejcik:
AFAICT each runner tries to write its own results, they all write over each other, and the changes are lost. My understanding is that this is essentially random and _some subset_ of changed tests _would_ get written out given the right timing.

In any case, each runner must write their own fixture results, and the xdist controller is the one to merge them into the resulting fixtures.json.

General improvement suggestion: it would be nice if, internally, each test run produced a separate list of ui hashes (then the workers can dump that into `fixtures.gwXYZ.json`), and this list would then be _merged_ into the main list, keeping or removing pre-existing tests based on `--ui-check-missing`

grdddj:
Do we still want this, or is the current solution of 
```
test_emu_ui_record_multicore: ## quickly record all screens
	make test_emu_ui_multicore || echo "All errors are recorded in fixtures.json"
	make test_emu_accept_fixtures
```

enough? In that case we could close this

matejcik:
I think so, yes.

I was also thinking we could completely remove the `record` option at least internally, in favor of always running as if `test` and possibly running the "accept" code path afterwards.

PR #2502, opened 2022-09-09
Auto-paginate on v1 UI

matejcik:
When `debuglink.press_yes()` is sent, the UI will quickly scroll through all screens of the pagination before closing the current screen. This simplifies certain parts of UI tests.

It also causes a small difference in paginated tests where the scrollbar overflows the title, which did not seem worth fixing.

This also brings a small refactor of the debuglink feature that makes it simpler to register for debuglink events.

517cb6f9a9b01a3bc4e2724d86d1a19a379a76eb adds a private implementation of RIPEMD160 to trezorlib so that it works fine even with recent openssl that removes the RIPEMD160 algorithm

9cf81d4da9cd5057c8f7d3d0d8a8bce5471fb09b fixes a problem with `make test_emu_ui_multicore`, which is a sibling problem to #2493


matejcik:
UI difference:
image 
image

full diff: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3004946409/artifacts/master_diff/index.html

matejcik:
this will not be relevant anymore.

we might want to implement the same thing for UI2 though

PR #2504, opened 2022-09-09
feat(core,legacy): display only sat instead of sat BTC

marnova:


matejcik:
there is no UI diff for this change.

does it mean that we don't have tests looking at `sat BTC`? If that is true, please make some :)

matejcik:
you can take a simple signing test for the input data and add it to `test_signtx_amount_unit.py`, parametrized the same way the existing testcase that is already there.

tests/device_tests/bitcoin/test_signtx_amount_unit.py
```diff
@@ -73,3 +82,50 @@ def test_signtx(client: Client, amount_unit):
...
+                request_input(0, TXHASH_0dac36),
+                request_output(0, TXHASH_0dac36),
+                request_output(1, TXHASH_0dac36),
+                request_input(0),
+                request_output(0),
+                request_output(0),
+                request_finished(),
+            ]
+        )
+
```
matejcik, line 122: please exclude this block, that's not what we are interested in here so it's just needlessly spamming the test size

matejcik:
please fix style and squash everything. LGTM code-wise, just waiting for the UI test screens

matejcik (approved):
device tests are failing but they do show the difference that I was interested in, so :+1: on the result.

please merge when you get the CI green

tests/device_tests/bitcoin/test_signtx_amount_unit.py
marnova: 0dede9b8d613477ea75203e19d1448945b6fd6ad

PR #2506, opened 2022-09-09
feat(core/model_r): only refresh screen on display_refresh

matejcik:
Given that on Model R we have a frame buffer, let's not push out pixels one by one.

Takes code from #2307 to dump the whole frame buffer at once.


TychoVrahe (approved):
This looks fine, but we will need some more modifications for this to work though. 

Mainly boardloader and bootloader do not call `display_refresh` it their screens and error reports (didn't need until now as it wasn't needed - no emulator and T1 support there).

PR #2507, opened 2022-09-10
Speed up BTC signing

andrewkozlik:


andrewkozlik:
In https://github.com/trezor/trezor-firmware/pull/2507/commits/87bc82b405a895911f51d09cd87e90344e18e88f I tried changing `no_serialize` to `serialize` and the default to `true`. This change breaks device tests as well as unit tests. There seems to be something very wrong with the handling of default values in our code base.

`pytest ... -vv` log shows this:
```
[2022-09-19 18:42:28,612] trezorlib.client DEBUG: sending message: SignTx
SignTx (13 bytes) {
    coin_name: 'Testnet',
    inputs_count: 2,
    outputs_count: 1,
}
```

Trezor log shows this:
```
58574977910 trezor.wire DEBUG received message contents:
SignTx {
    inputs_count: 2
    lock_time: 0
    coin_name: 'Testnet'
    version: 1
    amount_unit: 0
    outputs_count: 1
    decred_staking_ticket: False
    timestamp: None
    branch_id: None
    version_group_id: None
    expiry: None
    serialize: False
}
```

This would indicate two bugs, neither of which I can explain:
- It's as if the `__init__()` method of `SignTx` in trezorlib [here](https://github.com/trezor/trezor-firmware/blob/87bc82b405a895911f51d09cd87e90344e18e88f/python/src/trezorlib/messages.py#L1193) isn't getting called, so `True` is never assigned to the `serialized` parameter on the trezorlib side.
- At the same time on the Trezor side the `serialized` parameter is assigned `False` instead of `True` as the default.

matejcik:
> * It's as if the `__init__()` method of `SignTx` in trezorlib [here](https://github.com/trezor/trezor-firmware/blob/87bc82b405a895911f51d09cd87e90344e18e88f/python/src/trezorlib/messages.py#L1193) isn't getting called, so `True` is never assigned to the `serialized` parameter on the trezorlib side.

It's not, and that is by design. (though very confusing)
The generated code is better thought of as a fakery for type-checking purposes than actual code; when we drop python 3.6 compatibility, the generated constructors will fly out the window.

> * At the same time on the Trezor side the `serialized` parameter is assigned `False` instead of `True` as the default.

Indeed, there seems to be a [bug in `pb2py`](https://github.com/trezor/trezor-firmware/blob/master/common/protob/pb2py#L641) where it expects the value to be a capital-T `True`. I'd need to check but I guess that shoud either be lowercase, or maybe a case-insensitive match?

matejcik:
> It's not, and that is by design. (though very confusing)

except that it does not matter and the functionality should be there: https://github.com/trezor/trezor-firmware/issues/2527

onvej-sl (approved):

core/src/apps/bitcoin/sign_tx/bitcoin.py
```diff
@@ -197,21 +213,25 @@ async def step2_approve_outputs(self) -> None:
...
-                        await self.verify_external_input(i, txi, txi.script_pubkey)
+            # All internal inputs are Taproot. We only need to verify presigned external inputs.
+            # We can trust the amounts and scriptPubKeys, because if an invalid value is provided
+            # then all issued signatures will be invalid.
+            if self.has_presigned:
+                expected_digest = self.h_external_inputs
+                for i in range(self.tx_info.tx.inputs_count):
+                    if i in self.external:
+                        progress.advance()
+                        txi = await helpers.request_tx_input(self.tx_req, i, self.coin)
```
andrewkozlik, line 227: > It seems to me that we can easily avoid streaming the inputs that have an ownership proof here. See [this draft](https://github.com/trezor/trezor-firmware/tree/onvej-sl/coinjoin-speed-step-3). Does it make sense?

The reason why I chose to use the `has_presigned` boolean value over a `set` of `presigned` indices is because it seemed like a premature optimization to use a `set`. It would optimize the signing speed only when all of the following conditions are satisfied:
- There is at least one presigned input.
- There are many external inputs with an ownership proof.
- All internal inputs are taproot.

So far we don't have a use-case for that. But it would be useful for CoinJoins where some participants don't have the ability to produce an ownership proof for their input and produce a presigned input. I am not against the optimization. I guess using a `set` vs. `bool` here won't make a relevant difference in terms of RAM.

andrewkozlik: We can add this commit to the PR: https://github.com/trezor/trezor-firmware/commit/2700aa24fec578497ed1d993fb53654574254690

core/src/apps/bitcoin/sign_tx/progress.py
```diff
@@ -1,31 +1,129 @@
-from trezor import ui, utils, workflow
+from typing import TYPE_CHECKING
+
+from trezor import ui, utils, wire, workflow
```
grdddj, line 3: Theoretically `wire` could be imported in `if __debug__` at the very bottom, it is not used in "non-debug" mode. (The reason is it occupies flash space even if it is unused).

Also `utils` and `workflow` are used only in a single function, so it can be imported there instead of top-level.

It is not so critical, each local import here would save around 5 bytes of flash space, but we might do these things consistently over the whole codebase - based on https://github.com/trezor/trezor-firmware/pull/2525

core/src/apps/bitcoin/sign_tx/progress.py
```diff
@@ -1,31 +1,129 @@
-from trezor import ui, utils, workflow
+from typing import TYPE_CHECKING
```
grdddj, line 1: As the size of this file increased a lot (from 30 to 130 lines), it might start making sense to create a `Progress` class instead of this collection of functions, that need to use the `global` keyword quite a lot.

(The `ProgressObj` would need to be created here, so it can be shared for all the places that call it.)

From size-perspective creating a class could be more costly, but the readability should be better. (And potentially modifying the instance variables could be less expensive than modifying global variables, which is happening quite often here.)

grdddj (commented):
Just a small review/some ideas, I cannot judge the signing details very much

core/src/apps/bitcoin/sign_tx/progress.py
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2507/commits/236640251228724ca3db76c8f3293fd33477d33c

core/src/apps/bitcoin/sign_tx/progress.py
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2507/commits/1d047e0b835765e76da42604237cea778cbcb131

mmilata (approved):
Changes to signing flow look good as far as my limited understanding of it goes. Stared a bit at the C changes and did not find any issues. Good job with the speedup!

core/src/apps/bitcoin/sign_tx/bitcoin.py
andrewkozlik: @onvej-sl?

I added the commit to this PR as https://github.com/trezor/trezor-firmware/pull/2507/commits/0c24201a844b290657d3982b5d483edb41075105.

core/src/trezor/crypto/bech32.py
```diff
@@ -156,17 +127,13 @@ def convertbits(
...
-    hrpgot, data, spec = bech32_decode(addr)
-    # the following two lines are strictly not required
-    # but they make typecheckers happy
-    if data is None:
-        return (None, None)
-    if hrpgot != hrp:
-        return (None, None)
     try:
+        hrpgot, data, spec = bech32_decode(addr)
         decoded = bytes(convertbits(data[1:], 5, 8, False))
```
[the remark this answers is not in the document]
andrewkozlik: `convertbits()` was my first suspicion, but as it turns out, it's nowhere near as significant as `bech32_decode()`. We might be able to save 5ms per external output, which could translate to a 13% improvement in the signing time if we are lucky.

core/src/apps/bitcoin/sign_tx/progress.py
```diff
@@ -1,31 +1,132 @@
...
 
-def advance(i: int = 1) -> None:
-    global _progress
-    _progress += i
-    report()
+class Progress:
+    def __init__(self):
+        self.progress = 0
+        self.steps = 0
+        self.prev_tx_step = 0
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2507/commits/6ec9b5a478d59b21e598fd81f7f46eb5ef50ff4a

matejcik:
> Is it OK to ignore the option in T1

Yes. This doesn't break compatibility in any way either, the caller is free to ignore the serialized chunk.

> Naming. I'd prefer `serialize`

this is resolved now, right?

core/src/trezor/crypto/bech32.py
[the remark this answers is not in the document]
andrewkozlik: So I tried implementing this and the improvement on the hardware is 1.65 seconds for the large transaction, i.e. 1%. Not worth reviewing and merging :-(.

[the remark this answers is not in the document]
mmilata: Agreed, thanks for testing it.

core/src/apps/bitcoin/sign_tx/bitcoin.py
```diff
@@ -122,27 +139,26 @@ def __init__(
         # stable device tests.
         self.orig_txs: list[OriginalTxInfo] = []
 
-        # The digests of the external inputs streamed for approval in Step 1. These are used
-        # to ensure that the inputs streamed for verification in Step 3 are the same as
+        # The digest of the presigned external inputs streamed for approval in Step 1. This is
+        # used to ensure that the inputs streamed for verification in Step 3 are the same as
```
matejcik, line 143: one thing I still don't understand is why we can skip over inputs that are not presigned.

matejcik:
> I tried to optimize the Python implementation, but it didn't help and I couldn't locate any particular operation that is responsible for the huge time difference. 

could it be the checksum calculation? native ints are only like 28 bits IIRC, so anything bigger is a mpz bigint

tests/ui_tests/fixtures.json
```diff
@@ -2058,21 +2054,21 @@
...
-"TTui2_bitcoin-test_verifymessage_segwit.py::test_message_testnet": "ee70f4a3410c5cb37a351587b38a625b489a801be0f5040fc3649c9b0fb7544d",
-"TTui2_bitcoin-test_verifymessage_segwit.py::test_message_verify": "b934503757f3974319ce43f4f9fad5e9d48291c17892b939bd4548ca122f4f76",
-"TTui2_bitcoin-test_verifymessage_segwit.py::test_verify_utf": "5daff0a6e6df88737ca1b479afd87df2e06e6e2233b86d9da123e2af8a1cecc3",
-"TTui2_bitcoin-test_verifymessage_segwit_native.py::test_message_long": "00c2b350d781b89b7a6df32b337798d85f54b76dd32c1faadd868cf90bd8249e",
-"TTui2_bitcoin-test_verifymessage_segwit_native.py::test_message_testnet": "dd9c50a6fb92ca728fe254111ef0a2d0392cc81ac21dd1ebc009314f24f07d0e",
-"TTui2_bitcoin-test_verifymessage_segwit_native.py::test_message_verify": "dfc35dc605f15bb3f391ae5c10870b3358ea49d1aea02ff9f3607798c15ca667",
-"TTui2_bitcoin-test_verifymessage_segwit_native.py::test_verify_utf": "91b659369711ecee12e7976ad56f6f1b8b748871be5f0d864b38c626109b5ce0",
+"TTui2_bitcoin-test_verifymessage.py::test_message_grs": "2064ccc2b13792fad9d0b5fe7f440dde0c4008e9bdb97836d6daa25abdd3b1f8",
```
matejcik, line 2057: something very strange is happening here. probably some sort of rebase artifact?

core/src/apps/bitcoin/sign_tx/bitcoin.py
andrewkozlik: There are three types of external inputs:
- with ownership proof - These used to be checked in Step 3, but are now checked in Step 1, see `process_external_input()`.
- presigned - These are still checked in Step 3.
- unverified - There is nothing to check for these. Only allowed when safety checks are disabled. Requires user confirmation if present, see `approvers.py`.

tests/ui_tests/fixtures.json
andrewkozlik: No clue where this came from. Reverted in https://github.com/trezor/trezor-firmware/pull/2507/commits/b4e4326e6dd8bf0993bed576f4c7d6acd1fb7240.

matejcik (approved):

andrewkozlik:
> > I tried to optimize the Python implementation, but it didn't help and I couldn't locate any particular operation that is responsible for the huge time difference.
> 
> could it be the checksum calculation? native ints are only like 28 bits IIRC, so anything bigger is a mpz bigint

I suppose it could be a part of the problem. When I was researching the speedup, I couldn't localize the time-demanding part to any particular step inside `bech32.decode()`.

PR #2517, opened 2022-09-15
Don't support preauthorized operations for get_public_node() and get_address() in trezorlib

andrewkozlik:
Trezorlib shouldn't support preauthorized operations for `get_public_node()` and `get_address()`. I forgot to remove this code during code review of the SLIP-25 implementation. The Trezor function `get_public_key()` doesn't work with `DoPreauthorized`. The Trezor function `get_address()` should work with `DoPreauthorized`, but it's more natural to use `UnlockPath` instead.

No QA needed.

### DIFF
--- a/python/.changelog.d/2517.added
+++ b/python/.changelog.d/2517.added
@@ -1,0 +1,1 @@
+Support SLIP-25 accounts in get-public-node and get-address.
--- a/python/src/trezorlib/btc.py
+++ b/python/src/trezorlib/btc.py
@@ -114,7 +114,6 @@ def get_public_node(
     coin_name: Optional[str] = None,
     script_type: messages.InputScriptType = messages.InputScriptType.SPENDADDRESS,
     ignore_xpub_magic: bool = False,
-    preauthorized: bool = False,
     unlock_path: Optional[List[int]] = None,
     unlock_path_mac: Optional[bytes] = None,
 ) -> "MessageType":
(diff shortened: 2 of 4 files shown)


andrewkozlik:
https://github.com/trezor/trezor-firmware/pull/2517/commits/f333fdb53d40845b916957e788f92eee65555214: Added support for SLIP-25 accounts in `trezorctl` commands `get-public-node` and `get-address`.

I am not sure how to deal with `get-descriptor`, because one can't specify a SLIP-25 account. I suppose redefine `-n` to mean BIP-32 path or account number.

```
Usage: trezorctl btc get-descriptor [OPTIONS]

  Get descriptor of given account.

Options:
  -c, --coin TEXT
  -n, --account INTEGER           account index (0 = first account)
                                  [required]
  -t, --script-type [address|segwit|p2shsegwit|taproot|pkh|wpkh|sh-wpkh|tr]
  -d, --show-display
  --help                          Show this message and exit.
```

matejcik (approved):
I believe we don't need changelog because there has not been a release in the meantime, right?

andrewkozlik:
> I believe we don't need changelog because there has not been a release in the meantime, right?

The last commit https://github.com/trezor/trezor-firmware/pull/2517/commits/1d114c6fd8cf8ea26dd0565e3cf05b64771b017d deserves a changelog.

PR #2518, opened 2022-09-15
Add Makefile commands to be able to flash T1 legacy bootloader and firmware via JLink

hiviah:
Add two commands for commandline flashing of T1 legacy bootloader and firmware with JLink.

Both commands are marked `.PHONY` so that it won't trigger rebuild, since (re)build needs to be done with `cibuild` script.


TychoVrahe (approved):
didn't test it, but looks fine

Issue #1922, opened 2021-11-18
TT Redesign


alex-jerechinsky:
Priority: Tasks that are required for Bitcoin only firmware

- [x] pagination
  - #1960 
  - #1961 
  - #1993 
- [x] #2118
- [x] enable debuglink for new UI
- [x] #2169
- [x] fonts #1122
- [ ] [clean up timer handling](https://github.com/trezor/trezor-firmware/pull/2227#discussion_r869337119): #2299
- [ ] implement layouts
  - [x] generic dialogs and confirmations
    - #2071
    - #2097
    - #2155
  - [x] #2215
  - [x] #2498 
  - [x] FIDO layouts #2532
  - [x] #2298
  - [x] #2142
    - #2098 
    - #2154
  - [ ] PIN/passphrase/recovery keyboard
    - #2104
    - #2154
    - #2227 
    - [x] fix visual glitches caused by overlapping `Maybe`s (needs Honza's placement code)
    - [x] show wrong pin warning (needs ability to start timer on layout init)
    - [x] long passphrases #2499
    - [ ] #2530
    - [x] funny button shape in PIN keyboard
  - [x] #2519
  - [x] altcoins #2500
- [x] `show_success` cannot break long lines, we need to add support for center-aligned paragraphs (in #2383)
- [x] #2531
- [x] `confirm_blob` doesn't support `ask_pagination` yet (bitcoin doesn't need it)
- [x] each icon (for IconDialog) takes 0.5K of flash, we need to either figure out how to represent three-color image more efficiently than 16bpp, or draw them procedurally
- [ ] unlocking and signing and verifying pin progress
- [x] fix device tests to work with UI2, possibly combined with dropping python UI entirely

Post-switch:
- [x] #2631
- [x] #2624 #2643
- [ ] #2629
- [ ] #2585
- [x] #2665 
- [x] #2670 
- [ ] rename `ui/layouts/tt_v2` to `tt`, merge and clean up `class Layout` + `class RustLayout`

Nice to have:
- [ ] switch to rust implementation of qrcodegen, this would make it possible to propagate errors to python which we currently don't do

grdddj:
While getting rid of all `UI1` stuff as a part of https://github.com/trezor/trezor-firmware/issues/2559, I noticed that `ui.Component` is being used in couple `monero` layouts, for example
[...]

PR #2519, opened 2022-09-20
Rust UI: confirm_properties layout

mmilata:
Part of #1922. The biggest change in this PR  is that `Paragraphs` no longer holds its contents in Vec with hardcoded capacity but any data structure implementing a simple trait can be used. I've attempted to put most of the logic into functions that won't be monomorphized in order to save code size. My current solution has the disadvantage that it needs to allocate large buffer on stack (can be moved to globals) when computing paragraph breaks, LMK if you can think of something better. [[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3053013611/artifacts/master_diff/index.html)]

TODO:
- [x] Put key and value on the same page if possible: [python impl](https://github.com/trezor/trezor-firmware/blob/core/v2.5.2/core/src/trezor/ui/layouts/tt/__init__.py#L814); done: 48becacd8c525fcc6a106c4db3efdddf87a59e9e.
- [x] Test on hardware.
- [ ] Consider renaming `Paragraph` to `Par`/`Para` to avoid mixups with `Paragraphs`.

cp1 cp2


jpochyla:
Niiiice

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -259,6 +262,23 @@ where
...
+        let mut iter_buf = IterBuf::new();
+        let iter = Iter::try_from_obj_with_buf(*self, &mut iter_buf);
+        let iter = unwrap!(iter);
+        for obj in iter {
+            let text: StrBuffer = unwrap!(obj.try_into());
+            let text = text.as_ref();
+            sink.extend(Some(Paragraph::new(&theme::TEXT_NORMAL, text)))
+        }
+    }
+}
```
mmilata, line 279: I didn't manage to implement taking the paragraphs directly out of `Obj` because I couldn't figure out the lifetimes:
```
error[E0597]: `text` does not live long enough
   --> src/ui/model_tt/layout.rs:275:24
    |
269 |     fn collect<'a>(&self, sink: &mut impl Extend<Paragraph<&'a str>>) {
    |                -- lifetime `'a` defined here
...
275 |             let text = text.as_ref();
    |                        ^^^^^^^^^^^^^ borrowed value does not live long enough
276 |             sink.extend(Some(Paragraph::new(&theme::TEXT_NORMAL, text)))
    |             ------------------------------------------------------------ argument requires that `text` is borrowed for `'a`
277 |         }
278 |     }
    |     - `text` dropped here while still borrowed

```

If we decide not to implement the trait for `Obj` then `ParagraphSource` can be redefined to provide random-access (instead of sequential) which would probably allow us to get rid of the large temporary array.

matejcik: doing this for `Paragraph<StrBuffer>` should be fine, no? the whole point of `StrBuffer` is to steal reasonable lifetimes from under the GC

> redefined to provide random-access (instead of sequential)

we can get random access from uPy lists too. it risks exceptions, but so does using the iterator.

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -417,14 +506,31 @@ where
...
+impl<T, const N: usize> VecExt<T> for Vec<Paragraph<T>, N>
+where
+    T: AsRef<str>,
+{
+    fn add(&mut self, paragraph: Paragraph<T>) -> &mut Self {
+        if paragraph.content.as_ref().is_empty() {
+            return self;
+        }
+        if self.push(paragraph).is_err() {
+            #[cfg(feature = "ui_debug")]
```
TychoVrahe, line 531: could benefit from debug-assertions also here using debug_panic!

TychoVrahe (commented):
The main issue i have here is with Swipe(Hold)Page and actually not caused in this PR, but nevertheless, it manifests here when testing confirm_properties on device so it seems like a good place to fix it
- the loader in SwipeHoldPage is not positioned correctly and collides with buttons. The place function should take into account height of the button area and probably also BUTTON_SPACING
- the buttons flicker when loader animation is in progres. This is fixed with this commit https://github.com/trezor/trezor-firmware/commit/a4d0ec3926607b54a15a712f2202f1a654a27607, but then the buttons are not redrawn when Hold to confirm is Cancelled. In on_page_change we should probably call request_complete_repaint  of the buttons too.
- when hold to confirm completes, the content is drawn on the remainders of the loader and briefly shown before next actions continue. It should probably just continue with fade out and move to next stuff right away.

mmilata:
Rebased to master & fixed the SwipeHoldPage issues in dc675ab825a29c417cfab891b95832a1cd26517a.

Switched `ParagraphSource` to random access in 9cd430fb2c2e04b28fccff05c5b5c19dced2e614.

Implemented key-value-on-same-page in 48becacd8c525fcc6a106c4db3efdddf87a59e9e.

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -338,17 +359,66 @@ extern "C" fn new_confirm_blob(n_args: usize, args: *const Obj, kwargs: *mut Map
...
+
+        let mut paragraphs = ParagraphVecLong::new();
+
+        let mut iter_buf = IterBuf::new();
+        let iter = Iter::try_from_obj_with_buf(items, &mut iter_buf)?;
+        for para in iter {
+            let [key, value, is_mono]: [Obj; 3] = iter_into_objs(para)?;
+
+            if let Some(key) = key.try_into_option::<StrBuffer>()? {
+                paragraphs.add(Paragraph::new(&theme::TEXT_BOLD, key));
```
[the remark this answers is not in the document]
mmilata: That was my plan but had no good name for the flag. Implemented in 48becacd8c525fcc6a106c4db3efdddf87a59e9e (only works for two subsequent paragraphs though so not sure about the reusability).

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -279,12 +381,34 @@ struct PageBreakIterator<'a, T> {
...
+        paragraphs: &[Paragraph<&'a str>],
+        mut offset: PageOffset,
+    ) -> Option<PageOffset> {
+        for paragraph in paragraphs.iter().skip(offset.par) {
+            let (next_offset, remaining_area, _layout) = offset.advance(area, paragraph);
+            if next_offset.par >= paragraphs.len() {
+                return None;
+            } else if let Some(remaining_area) = remaining_area {
+                #[cfg(feature = "ui_debug")]
+                assert!(next_offset.par == offset.par + 1);
```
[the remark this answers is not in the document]
mmilata: Fixed in dc675ab825a29c417cfab891b95832a1cd26517a. IIRC some time ago we decided to keep the debug assertions and PYOPT=0 separate.

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -205,60 +237,73 @@ pub mod trace {
...
     }
 }
 
+// NOTE: keep Paragraph<&str> small as we allocate long array of them on the
+// stack.
+#[derive(Clone, Copy)]
 pub struct Paragraph<T> {
     content: T,
-    layout: TextLayout,
+    style: &'static TextStyle,
```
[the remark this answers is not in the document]
mmilata: We'd have to include the lifetime in all types that contain the Paragraphs in them which hurts readability IMO. Let's do it once/if it's needed.

TychoVrahe:
One more issue with `SwipeHoldPage` after the fix: when the Hold to confirm button is pressed, or the loader shrinks, the buttons get cleared and redrawn, causing a flicker. Thinking that `SwipeHoldPage` should have its own `Pad` with size of content area returned by `place_get_content_area` which would be passed into `handle_hold_event` instead of `inner.pad`. Note that `place_get_content_area` doesn't currently return content + scrollbar like described in the comment so this needs to be fixed too in order for this to work properly (it returns `content_single_page` which is narrower by `theme::CONTENT_BORDER`, so maybe return `layout.content_single_page.union(layout.scrollbar)`)

mmilata:
Good catch! Do you just spot these or do you have some flicker-debugging mode implemented?

I've applied your suggestion in 66698ca and it seems to be working. If there are more similar bugs we might want to re-implement the component using `Maybe` + `PaintOverlapping` which were added some time after the initial `SwipePage` implementation.

TychoVrahe (approved):

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -279,12 +451,37 @@ struct PageBreakIterator<'a, T> {
     current: Option<PageOffset>,
 }
 
+impl<'a, T: ParagraphSource> PageBreakIterator<'a, T> {
```
matejcik, line 454: this should just be `impl<T> PageBreakIterator<'_, T>` no? as you're not using any of the type arguments in this block, it's basically just a namespace for `dyn_next`

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -338,17 +342,72 @@ extern "C" fn new_confirm_blob(n_args: usize, args: *const Obj, kwargs: *mut Map
...
+                } else {
+                    paragraphs.add(Paragraph::new(&theme::TEXT_BOLD, key));
+                }
+            }
+            if let Some(value) = value {
+                if is_mono.try_into()? {
+                    paragraphs.add(Paragraph::new(&theme::TEXT_MONO, value));
+                } else {
+                    paragraphs.add(Paragraph::new(&theme::TEXT_NORMAL, value));
+                }
```
matejcik, line 372: sidenote: we should violently throw out this flexibility of sometimes specifying only key and sometimes specifying only value. I believe only Cardano is using that, maybe a little bit of UX work is needed on that side.

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1021,6 +1092,17 @@ pub static mp_module_trezorui2: Module = obj_module! {
     ///     """Confirm byte sequence data."""
     Qstr::MP_QSTR_confirm_blob => obj_fn_kw!(0, new_confirm_blob).as_obj(),
 
+    /// def confirm_properties(
+    ///     *,
+    ///     title: str,
+    ///     items: Iterable[Tuple[str, str, bool]],
```
matejcik, line 1098: ...speaking of, this new type annotation disallows `None`s anyway?

matejcik (commented):
LGTM overall, one very small style nit

core/embed/rust/src/ui/model_tt/layout.rs
matejcik: what is the status of this? or should we merge as-is and create an issue for `ParagraphSource for Obj`?

core/embed/rust/src/ui/component/text/paragraphs.rs
mmilata: Oh, I didn't realize this was possible, fixed in 8a4c769cc2601aa8a3d01bd997a639cc6c5eaa51.

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -279,12 +451,37 @@ struct PageBreakIterator<'a, T> {
...
+    }
+}
+
 /// Yields indices to beginnings of successive pages. First value is always
 /// `PageOffset { 0, 0 }` even if the paragraph vector is empty.
-impl<'a, T> Iterator for PageBreakIterator<'a, T>
-where
-    T: AsRef<str>,
-{
+impl<'a, T: ParagraphSource> Iterator for PageBreakIterator<'a, T> {
```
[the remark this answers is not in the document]
mmilata: 8a4c769cc2601aa8a3d01bd997a639cc6c5eaa51

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: Good catch, fixed in 8a4c769cc2601aa8a3d01bd997a639cc6c5eaa51

matejcik (approved):

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: In #2531 I implemented `ParagraphSource for Vec<Paragraph<Obj>>` - it seemed more straightforward to me to determine the number of paragraphs and their styling at the beginning and only call to uPy to get the contents. If I'm measuring this correctly `ParagraphVecLong<Obj>` has 388 bytes in firmware.

mmilata: And `ParagraphVecLong<StrBuffer>` has 516, might save some memory by converting to `ParagraphVecLong<Obj>`. Quick & dirty way to measure this:

```rust
macro_rules! dump_size {
    ($x:expr $(,)?) => {
        #[allow(unknown_lints, eq_op)]
        const _: [(); 0 - $x as usize] = [];
    };
}
use core::mem;
dump_size!(mem::size_of::<ParagraphVecLong<Obj>>());
```

mmilata:
Rebased on master and fixed ui fixtures, will merge if CI passes.

PR #2520, opened 2022-09-20
Workaround failing build_unix_debug

mmilata:
Something needs the symbol even though we don't do unwinding (I think):
```
/bin/ld: build/unix/rust/debug/libtrezor_lib.a(core-05898138a596088a.core.4cbccd39-cgu.0.rcgu.o):(.data.DW.ref.rust_eh_personality[DW.ref.rust_eh_personality]+0x0): undefined reference to `rust_eh_personality'
```
Defining one that does nothing seems to avoid the problem, not sure what happens when it actually gets called. Needs nightly rustc. Possibly related: https://github.com/rust-lang/rust/issues/56152, https://docs.rust-embedded.org/embedonomicon/smallest-no-std.html#eh_personality

### DIFF
--- a/core/embed/rust/src/lib.rs
+++ b/core/embed/rust/src/lib.rs
@@ -3,6 +3,7 @@
 #![allow(clippy::new_without_default)]
 #![deny(unsafe_op_in_unsafe_fn)]
 #![allow(dead_code)]
+#![feature(lang_items)]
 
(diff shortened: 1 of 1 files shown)


matejcik (approved):
utACK?? :woman_shrugging: @jpochyla any opinion?

PR #2521, opened 2022-09-20
Boot sequence in rust

### LINKED ISSUE #2134: Firmware boot sequence in Rust
Due to memory layout optimization and the unimport mechanism, the contents of `main.py` are very fragile in terms of order of import statements: a wrongly placed import can significantly increase memory footprint, or in some cases even cause basic functions to fail (such as in #2129 experiments)

When the Rust UI is done, it should be relatively easy to implement `boot.py` fully in Rust, and that paves the way to implement all of `main.py` in Rust. This would remove the fragility and significantly shorten boot-up time. (which is currently ~3 seconds until the user sees _any_ content on screen).

In order to run the main boot sequence, we would also need to modify `modtrezorio-usb-*` and directly export the interface objects from Rust.

Open question: if the main sequence runs from Rust, when does the MicroPython interpreter start? A nice answer would be to restart it after every workflow, but we could only do that if the startup of the interpreter itself is fast enough.

TychoVrahe:
This PR reimplements boot sequence in rust.

Includes pin input, sd card salt validation, lockscreen including users avatar,+name, pin callback from storage, storage unlock.

~~Storage handling builds on #2536 , some other helper methods are implemented and should be usable like this, but when some complex storage solution is merged (e.g. #2214 ), this should be redone.~~

This unfortunately duplicates some python code, that is right now needed in both rust and python.


TychoVrahe:
rebased onto #2536 and done some cleaning

TychoVrahe:
rebased on top of #2571 , so that new homescreen is shown

TychoVrahe:
rebased

TychoVrahe:
rebased on current master and added support for T2B1. Few open issues in `todo`s still.

TychoVrahe:
I believe this is now ready to review.

The UI changes are caused by the different pin callback implementation, the remaining seconds message is shown earlier, which is likely more correct.

Merging this would probably make implementing stuff like https://github.com/trezor/trezor-firmware/issues/2110 more difficult, as the USB is handled by python which is not running yet when unlocking the device for first time.

PR #2522, opened 2022-09-20
added blackcoin support

CryptoLover705:


prusnak:
We are not adding new coins unless they are in TOP30 on Coinmarketcap.

koraycoding:
> We are not adding new coins unless they are in TOP30 on Coinmarketcap.

Trezor currently supports 1816 tokens and coins. No offense but your rejection argument doesn't make much sense. Blackcoin is one of the oldest and stable coins with good community and its acceptance will contribute positively to trezor and crypto world. I kindly ask you to reconsider on this matter.

matejcik:
I mean, BLK also has a 24 hour trading volume of $65 and I couldn't find a single non-reward transaction in as deep as the block explorer let me go :woman_shrugging:

PR #2523, opened 2022-09-20
fix flashing usb disconnected error

TychoVrahe:
This fixes flashing USB DISCONNECTED ERROR by adding 2s grace period after usb disconnection is detected. 

Change detection logic is removed from homescreen as it this object is rectreated sometimes and than the logic fails, futthermore it was redundant to the polling mechanism.

Also the header warning does not clear entire screen when it get drawn/removed, which eliminates flickering of the avatar.

The grace period is not only on startup but after every disconnection, as i was getting some brief warnings when suite started up - somehow the USB was for a small amount of time in `USBD_STATE_DEFAULT`.

### DIFF
--- a/core/embed/extmod/modtrezorio/modtrezorio.c
+++ b/core/embed/extmod/modtrezorio/modtrezorio.c
@@ -32,7 +32,7 @@
 #include "usb.h"
 
 // Whether USB data pins were connected on last check (USB configured)
-__attribute__((unused)) static bool usb_connected_previously = false;
+bool usb_connected_previously = true;
 
 #define CHECK_PARAM_RANGE(value, minimum, maximum)  \
   if (value < minimum || value > maximum) {         \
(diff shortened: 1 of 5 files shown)


hiviah (approved):

hiviah:
Not 100% sure about the removed "clear", but as far as I was able to test some scenarios, including locking or aborting the lock hold, it seems to work correctly with error and warning headers.

TychoVrahe:
The clear was clearing rest of the screen including avatar, which was then drawn again causing it to flash. U tried to prevent that, because the usb error can appear in arbitrary time and flashing the avatar is ugly. Setting the param to `False` would be an option too, but then `True` is never used afaik so i decided to remove it entirely. 

In case of loader abort, `display.clear` was added after @bosomt noticed the remnants of the loader. In other cases, the screen should be cleared by the layout management in the `_before_render` function. But i am not that familiar with the python part , so maybe @matejcik or @mmilata should check my logic here.

hiviah:
> but then True is never used afaik so i decided to remove it entirely

`True` was default for all of the `header_warning` and `header_error`, except the one changed. Since the calls appear at start of render function, the extra clear seems unnecessary though. So I guess this way it will flicker less in all cases.

matejcik (approved):

PR #2524, opened 2022-09-21
Fix flickering of hold to confirm on TT UI2

TychoVrahe:
Using Child components to eliminate flickering when animation is running or the other button is pressed.


mmilata (approved):
Thanks!

PR #2526, opened 2022-09-24
feat(cardano): add support for CIP36 Catalyst registration format

davidmisiak:
This PR adds support for [CIP36](https://cips.cardano.org/cips/cip36/) governance registration format. Changes:
- New field: `delegations` array (each delegation consists of a voting key and its voting power proportion) as an alternative to delegating the entire voting power to a single voting key. The array is embedded in the protobuf message due to simplicity (max. 32 delegations).
- New field: `voting_purpose` integer (currently 0 = Catalyst, 1 = other). For CIP36 registrations, this field is always serialized, (value 0 is used if not provided by the client).
- All `Catalyst` occurences are replaced by `governance`, since the format should be usable for other governance purposes, not only Catalyst.

The new 1694' derivation path and the vote cast call implementation is not included in this PR.


davidmisiak:
Sorry for confusion, this is not ready for a review yet (but should be soon).

Issue #2528, opened 2022-09-27
FIDO2 with invity.io logo


Hannsek:
"can we put a PNG invity logo in there?"

![image](https://user-images.githubusercontent.com/57008159/192571127-e0ecbb5a-5d9a-476d-8ecc-a3741690179f.png)

related: https://github.com/trezor/trezor-firmware/pull/1529

bosomt:
**QA OK**

**Info:**
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.10.1 Chrome/100.0.4896.160 Electron/18.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.3 Universal (revision e7ff43272ccae49fe4b1dcf198ed51c82d7c1472)
 - Transport: bridge 2.0.31

PR #2529, opened 2022-09-27
common/defs/fido: add invity.io

MrMebelMan:
closes https://github.com/trezor/trezor-firmware/issues/2528


MrMebelMan:
I have a few problems here:

First, gen_check fails because `Icon different from source: ./core/src/apps/webauthn/res/icon_invity.toif`.
I'm not quite sure what's the problem here, because I've made a `.toif` icon using
```bash
./core/tools/toif_convert.py ./common/defs/fido/invity.png ./core/src/apps/webauthn/res/icon_invity.toif
```

Second, tests don't pass if the PNG image is less than 128x128px. The problem is, when I use an image other than 64x64px, the logo ends up being too huge. Any tips? :pray: 

128x128px:
image

64x64px:
image

MrMebelMan:
Ok, I think I've messed up manual resizing, so I've temporarily removed this condition
https://github.com/trezor/trezor-firmware/blob/7046edd567fc28229c133766acfa328976a6b675/core/tools/build_icons.py#L54-L55
to allow gen_check to overwrite the `.toif` file with expected data and now everything works just fine.

P.S.: i wish I've RTFM and used `make gen` straight away :man_facepalming:

TychoVrahe:
note that recently merged dma2d pull request changed image format, please rebase and regenerate the icon

MrMebelMan:
> note that recently merged dma2d pull request changed image format, please rebase and regenerate the icon

Running `make gen` after the master rebase re-generates all the icons but mine :eyes: 

image

MrMebelMan:
> note that recently merged dma2d pull request changed image format, please rebase and regenerate the icon

Rebasing and running make gen [has lead to](https://github.com/trezor/trezor-firmware/pull/2529/commits/d0c5ff53954376f9c8e02f1ff00c3b40013a9fea) re-generation of all icons besides invity :eyes:
Any idea what's wrong here?

prusnak:
Did you pull master before rebase?

MrMebelMan:
> not sure what went wrong, but when i checkout your branch and run `make gen` all the icons including invity are returned back to the correct format

My bad, instead of using Poetry I've just nix-shelled the dependencies needed for `make gen`. There was probably some version mismatch in some of them. Oops!

MrMebelMan:
Hmm, no, even when I run it using Poetry, it generates icons that are different from master :<

andrewkozlik (approved):
utACK

PR #2525, opened 2022-09-22
Decrease micropython code size

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1947

Decreases the size of compiled `micropython` code by around **48 kb**, and `str1` symbol in `.flash2` by **6 kb**.

Makes a lot of code changes with the goal of making it more space-efficient, so the resulted compiled bytecode is smaller. It does it by consistently applying some strategies of code-size-reduction like inlining small functions or caching global variables in a local scope (more below).

The biggest decreases were achieved in the automatically generated data (`tokens.py`, `networks.py` or `coininfo.py`), where one small thing (like removing the keyword arguments) makes a huge difference.

Includes a tool for automatic detection of some size-decrease strategies (under `core/tools/upy_size`). Its [README](https://github.com/trezor/trezor-firmware/tree/grdddj/upy_size_decrease/core/tools/upy_size) covers all the used strategies in more details.

---

`Micropython` function sizes  [before](https://gist.github.com/grdddj/48fa9ea47f65d9705349a9650cab037f) and [after](https://gist.github.com/grdddj/d0d0a68b30ec0b6dc0fab585112bcdb4).

[Tree of size changes](https://gist.github.com/grdddj/90ac2478c3cd2e618f2a32d5432e3139) shows differences in specific directories/files.

It would be hardly possible without https://github.com/trezor/trezor-firmware/pull/2327, which offered quick feedback on any code change.

---

There is still some potential for further decreases (in order of `~ 5 kb`), which would mean inlining even more one-time functions (the longer ones) or making the `Message` types not enforce keyword-only arguments.

All the changes are up to discussion and we may decide to not include some of those, as optimization towards one side (size efficiency) worsens some other factors (like readability).

PR #2534, opened 2022-09-29
Redistribute contents in flash 

### LINKED ISSUE #2417: Alter distribution of code into .flash and .flash2 sections
Currently, with both the `TT` redesign and `model R` development, we are transferring `UI` code (and potentially other things) from `micropython` into `Rust`.

`Rust` code resides in `.flash` and `micropython` code in `.flash2` regions/sections. As we make the transition, `.flash` is overflowing while `.flash2` has (will have) a lot of free space (even more after implementing https://github.com/trezor/trezor-firmware/issues/15).

We might want/need to take some stuff from `.flash` and put them into `.flash2` to level it out (or most importantly avoid `.flash` overflowing).

TychoVrahe:
This PR redistributes content in flash a bit, so that all UI1 and UI2 on TT and TR can be linked. TT with UI2 and PYOPT=0 however does not fit, there just isn't enough space left in both sections to fit that right now. 

With logic that we are moving UI from python to rust, i moved the rust library into same section as python, so that overall size of that section is not changed too much when code is rewritten from py to rs.

closes #2417

### DIFF
--- a/core/embed/firmware/memory_T.ld
+++ b/core/embed/firmware/memory_T.ld
@@ -45,9 +45,11 @@ SECTIONS {
 
   .flash2 : ALIGN(512) {
     build/firmware/frozen_mpy.o(.rodata*);
-    build/firmware/vendor/secp256k1-zkp/src/secp256k1.o(.rodata*);
-    build/firmware/vendor/secp256k1-zkp/src/precomputed_ecmult.o(.rodata*);
-    build/firmware/vendor/secp256k1-zkp/src/precomputed_ecmult_gen.o(.rodata*);
+    build/firmware/vendor/trezor-crypto/aes/aestab.o(.rodata*);
+    . = ALIGN(4);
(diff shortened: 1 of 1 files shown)


matejcik (approved):
utACK. will this work with UI2 and #2525 ?

TychoVrahe:
@matejcik this works fine with that PR as it only decreases flash usage and even PYOPT=0 with UI2 will fit in flash. Since it decreases usage of flash2, there will be quite a lot of free space there and in flash1 we will still have relatively small reserve, so at that point we might want to move some other stuff to flash2. (other significant flash savings like upython 1.19 and token definitions also free flash2)

mmilata (approved):
Tested working.

prusnak (approved):

hiviah (approved):

hiviah:
BTW reran failing tests, but no idea why the "ui test fixtures" is still stuck.

PR #2536, opened 2022-10-03
Expose storage interface to rust

### LINKED ISSUE #2533: Rust Storage interface
while the #2214 PR is relatively large, it would be useful to make a smaller PR exposing only the `trezorhal::storage` module, for use in, e.g., #2521 

`trezorhal::storage_field` might be optionally included as well, but I am afraid that with what we know now about const generics and stack consumption, the `Field<String<N>>` and `Field<Vec<N>>` will need to be done differently.

TychoVrahe:
Cherry-picked storage interface from #2214

Closes #2533

### DIFF
--- a/core/embed/rust/src/trezorhal/storage.rs
+++ b/core/embed/rust/src/trezorhal/storage.rs
@@ -1,0 +1,333 @@
+use super::ffi;
+use crate::error::Error;
+use core::ptr;
+use cstr_core::{cstr, CStr};
+
+/// Result of PIN delay callback.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum PinCallbackResult {
+    /// Continue waiting for PIN unlock.
+    Continue,
+    /// Abort the unlock attempt before time is up.
+    Abort,
+}
+
+/// PIN delay callback function type.
+/// The storage layer will call this function while the PIN timeout is in
+/// progress. This is useful for showing UI progress bar.
+/// `wait` is the total number of seconds waiting.
+/// `progress` is a value between 0 and 1000, where 1000 indicates 100%.
+/// `message` is a message to show to the user.
+pub type PinDelayCallback = fn(wait: u32, progress: u32, message: &str) -> PinCallbackResult;
+
+pub type ExternalSalt = [u8; ffi::EXTERNAL_SALT_SIZE as usize];
+
+/// Static reference to the currently set PIN callback function.
+static mut PIN_UI_CALLBACK: Option<PinDelayCallback> = None;
+
+/// C-compatible wrapper for the Rust callback.
+unsafe extern "C" fn callback_wrapper(
+    wait: u32,
+    progress: u32,
+    message: *const cty::c_char,
+) -> ffi::secbool {
+    let message = unsafe { CStr::from_ptr(message as _) };
+    let result = unsafe {
+        PIN_UI_CALLBACK
+            .map(|c| c(wait, progress, message.to_str().unwrap_or("")))
+            .unwrap_or(PinCallbackResult::Continue)
+    };
+    if matches!(result, PinCallbackResult::Abort) {
+        ffi::sectrue
+    } else {
+        ffi::secfalse
+    }
+}
+
+/// Storage error type.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum StorageError {
+    /// Data is too long or not appropriate for the operation.
+    InvalidData,
+    /// Write failed.
+    WriteFailed,
+    /// Read failed.
+    ReadFailed,
+    /// Delete failed.
+    DeleteFailed,
+    /// Failed to perform the get-and-increment operation on a counter.
+    CounterFailed,
+}
+
+impl From<StorageError> for Error {
+    fn from(err: StorageError) -> Self {
+        match err {
+            StorageError::InvalidData => Error::ValueError(cstr!("Invalid data for storage")),
+            StorageError::WriteFailed => Error::ValueError(cstr!("Storage write failed")),
+            StorageError::ReadFailed => Error::ValueError(cstr!("Storage read failed")),
+            StorageError::DeleteFailed => Error::ValueError(cstr!("Storage delete failed")),
+            StorageError::CounterFailed => {
+                Error::ValueError(cstr!("Retrieving counter value failed"))
+            }
+        }
+    }
+}
+
+pub type StorageResult<T> = Result<T, StorageError>;
+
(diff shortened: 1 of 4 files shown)


core/embed/rust/src/trezorhal/storage.rs
```diff
@@ -0,0 +1,322 @@
...
+
+/// Get value for `appkey` from storage.
+/// The provided `data` buffer must be long enough to hold the value. If the
+/// buffer is shorter than the stored value, an error is returned. Use
+/// `get_length` to find the size of the value beforehand. On success, the data
+/// is copied to `data`, and length of copied data is returned. Data previously
+/// in the buffer beyond the copied data is not modified.
+/// If `appkey` is not present in storage, or the value is private and storage
+/// is locked, returns an error.
+pub fn get(appkey: u16, data: &mut [u8]) -> StorageResult<usize> {
```
mmilata, line 179:
```suggestion
pub fn get(appkey: u16, data: &mut [u8]) -> StorageResult<&[u8]> {
```
Returning a subslice of `data` might save the caller couple of lines of code.

core/embed/rust/src/trezorhal/storage.rs
```diff
@@ -0,0 +1,322 @@
...
+    } else {
+        Err(StorageError::CounterFailed)
+    }
+}
+
+/// Delete value for `appkey` from storage.
+/// Returns true if the value was successfully deleted.
+/// If the storage is locked, or the value does not exist in storage, returns
+/// false.
+pub fn delete(appkey: u16) -> bool {
```
mmilata, line 231: Why not `StorageResult<()>` like elsewhere?

mmilata (approved):
LGTM, got some bikeshed-level API suggestions.

core/embed/rust/src/trezorhal/storage.rs
matejcik: i think because the `false` result is not necessarily an "error"?

core/embed/rust/src/trezorhal/storage.rs
TychoVrahe: done https://github.com/trezor/trezor-firmware/commit/adb7063b5c166a675b04b3e09777a523735c9077

core/embed/rust/src/trezorhal/storage.rs
TychoVrahe: still the logic is similar to `get`, where locked storage or non present key returns an error

matejcik: Similar but not the same:
For `get`, the error `ReadFailed` indicates that we didn't get anything. This is an error in any case.
For `delete`, an error would mean that either the value isn't there (in which case it could be considered "success"), or we failed to delete it -- because storage is locked or because of i/o error.

But for API coherence, it seems more reasonable to declare and document that trying to delete nonexistent key _is an error_ (and then the higher level `storage_field` api will need to explicitly check if `storage.has(key)` before trying to delete it)

TychoVrahe: https://github.com/trezor/trezor-firmware/commit/f5839559c763ee55cd6d1e23bf0385da2dbdb02d

core/embed/rust/src/trezorhal/storage.rs
TychoVrahe: also tried to fix tests here https://github.com/trezor/trezor-firmware/commit/062ab2210c47c0a4fee96b02cdff31b481ab6bed but for some reason they still wont run, getting `signal: 11, SIGSEGV: invalid memory reference` on anything that actually accesses storage

core/embed/rust/src/trezorhal/storage.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+pub type StorageResult<T> = Result<T, StorageError>;
+
+/// Initialize the storage layer.
+/// This function must be called before any other storage function.
+pub fn init() {
+    unsafe {
+        ffi::storage_init(
+            Some(callback_wrapper),
```
mmilata, line 83: Seems to be crashing in this function. This line looks suspicious to me.

mmilata: Nevermind, the `Some` is actually fine.

mmilata: Maybe we need to call `flash_init` in the test setup?

core/embed/rust/src/trezorhal/storage.rs
```diff
@@ -0,0 +1,330 @@
...
+    } else {
+        Err(StorageError::ReadFailed)
+    }
+}
+
+/// Get value for `appkey` from storage.
+/// The provided `data` buffer must be long enough to hold the value. If the
+/// buffer is shorter than the stored value, an error is returned. Use
+/// `get_length` to find the size of the value beforehand. On success, the data
+/// is copied to `data`, and length of copied data is returned. Data previously
```
matejcik, line 178: please also update the comment ("length of copied data" -> subslice)

core/embed/rust/src/trezorhal/storage.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// Data is too long or not appropriate for the operation.
+    InvalidData,
+    /// Write failed.
+    WriteFailed,
+    /// Read failed.
+    ReadFailed,
+    /// Delete failed.
+    DeleteFailed,
+    /// Failed to perform the get-and-increment operation on a counter.
+    CounterFailed,
```
matejcik, line 59: Note: it kind of looks like that read/write/delete/counter could be folded into a single `OperationFailed`, because the only difference is error string. (`InvalidData` is different in that it is a problem with the data passed in, not with the storage operation)

but I am not sure how these are used in `storage_field` so let's keep them for now, or else we'll remove them when merging `storage_field`

core/embed/rust/src/trezorhal/storage.rs
matejcik: it's possible that it became necessary in the meantime .)

TychoVrahe: yeah that works! thanks https://github.com/trezor/trezor-firmware/commit/e6770444d6e4b67c5fcd7244030bb0f103e02639

core/embed/rust/src/trezorhal/storage.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/75587ee1b01dcd451c80a8142697dc5c6154965a

Issue #2217, opened 2022-04-12
Monero: blinking while refreshing progress bars


matejcik:
While displaying progress (such as "syncing" or "refreshing"), the Monero app repeatedly fades in/out the screen between steps.

This may have originally been a workaround for not properly redrawing the screen. However, in addition to being annoying, each fadein/fadeout adds an unnecessary 100ms delay to the whole operation.

E.g., in our tests, we have a 50-step liverefresh, which would incur real-life 5 second delay to the operation.

We should make sure that no unwanted blinking is happening.

matejcik:
details: Monero is implementing the loader as a `Component` and showing it in a `Popup`, leveraging the layout machinery. That is not appropriate in this usecase.

Instead, this should be converted to a directly painted progress tracking, same as in `bitcoin/sign_tx/progress.py`.

Issue #1298, opened 2020-10-06
Revisit warnings on TT homepage


tsusanka:
We have number of potential warnings: SEEDLESS, PIN NEEDED, BACKUP NEEDED, EXP FEATURES (#1297). We prioritize one over another. As part of the redesign we should consider redoing this, we might consider making it clickable and list all of them etc.

Related to #436.

PR #2538, opened 2022-10-03
Rust UI: homescreen/lockscreen/busyscreen

### LINKED ISSUE #2498: Rust UI: homescreen, lockscreen

mmilata:
Closes #2498.

Missing:
- [x] background image (avatar) - done in #2571 
- [ ] workflow restart integration isn't as good as in python (`trezorctl device setup --skip-backup --pin-protection` -> hold finger on touchscreen long enough for loader to show but lift it before locking -> `trezorctl ping foo` -> causes backlight fadeout-fadein)

homefail homepin homeexp busy lock


mmilata:
UI master diff: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3129304076/artifacts/master_diff/index.html (there's everything because homescreen is part of every testcase)

Device tests are failing due to #2539.

core/embed/rust/src/ui/model_tt/component/homescreen.rs
```diff
@@ -0,0 +1,266 @@
...
+                AREA,
+                icon,
+                notification.as_ref(),
+                color,
+            );
+        }
+    }
+
+    fn paint_loader(&mut self) {
+        let top_center = AREA.top_left().center(AREA.top_right());
```
TychoVrahe, line 83: could reuse Rect::top_center

core/embed/rust/src/ui/model_tt/component/homescreen.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+impl<T> Component for Homescreen<T>
+where
+    T: AsRef<str>,
+{
+    type Msg = HomescreenMsg;
+
+    fn place(&mut self, bounds: Rect) -> Rect {
+        self.pad.place(AREA.translate(LOADER_OFFSET));
```
TychoVrahe, line 153: seems wrong, probably the translated was meant to be used in loader

core/embed/rust/src/ui/model_tt/component/homescreen.rs
```diff
@@ -0,0 +1,266 @@
...
+pub enum HomescreenMsg {
+    Dismissed,
+}
+
+impl<T> Homescreen<T>
+where
+    T: AsRef<str>,
+{
+    pub fn new(label: T, notification: Option<(T, u32)>, hold_to_lock: bool) -> Self {
+        Homescreen {
```
TychoVrahe, line 43: `Self` seems less repetitive

core/embed/rust/src/ui/model_tt/component/frame.rs
```diff
@@ -187,3 +185,28 @@ where
         t.close();
     }
 }
+
+pub fn icon_text_center(
```
TychoVrahe, line 189: this function seems misplaced here, probable would be better in display::mod?

core/embed/rust/src/ui/model_tt/component/homescreen.rs
```diff
@@ -0,0 +1,266 @@
...
+            label,
+            notification,
+            hold_to_lock,
+            usb_connected: true,
```
TychoVrahe, line 47: i was removing this usb_connected change detection logic previously from python impl (see https://github.com/trezor/trezor-firmware/pull/2523) because when the object is recreated, the flag might be set wrongly and no painting might occur. this seems to reintroduce that issue, but i am not sure if this works in the same way

TychoVrahe (commented):
Just couple of quick comments, not thorough review.

Also cancelling the pin entry in bootscreen doesn't work (is fine in lockscreen)

core/embed/rust/src/ui/model_tt/component/homescreen.rs
```diff
@@ -0,0 +1,266 @@
...
+        event::{TouchEvent, USBEvent},
+        geometry::{Offset, Point, Rect},
+        model_tt::constant,
+    },
+};
+
+use super::{icon_text_center, theme, Loader, LoaderMsg, NotificationFrame};
+
+const AREA: Rect = constant::screen();
+const TOP_CENTER: Point = AREA.top_left().center(AREA.top_right());
```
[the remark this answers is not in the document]
mmilata: did not realize we already have bottom_center there, fixed [8dff0f7](https://github.com/trezor/trezor-firmware/pull/2538/commits/8dff0f7420698191712ea4e9a5e4b3bb7efba776)

core/embed/rust/src/ui/model_tt/component/homescreen.rs
mmilata: fixed [8dff0f7](https://github.com/trezor/trezor-firmware/pull/2538/commits/8dff0f7420698191712ea4e9a5e4b3bb7efba776)

core/embed/rust/src/ui/model_tt/component/homescreen.rs
mmilata: d'oh, fixed [8dff0f7](https://github.com/trezor/trezor-firmware/pull/2538/commits/8dff0f7420698191712ea4e9a5e4b3bb7efba776)

core/embed/rust/src/ui/model_tt/component/homescreen.rs
mmilata: I saw that PR and assumed whenever homescreen is started then USB must be connected, which is probably not always the case. Tried to improve this in [631c138](https://github.com/trezor/trezor-firmware/pull/2538/commits/631c1383a82da15db631f934c43a18542624c5d2).

mmilata:
Bootscreen fix in b680abdd0da8620a08f97e73b9ce493f334888f3

core/embed/rust/src/ui/model_tt/component/frame.rs
TychoVrahe: it is a better choice since it can be reused outside of TT, we can figure out something else if we gather more of these helpers or `ui::util` bloats too much.

TychoVrahe:
The homescreen is still redrawn sometimes, for example when suite wallet is opened. Not sure about why is that, but it should be avoided because with the image the rendering is relatively slow so it is annoying

matejcik:
> The homescreen is still redrawn sometimes

The Python code explicitly checks if the homescreen was not erased by anything since the last time it was shown, and in that case does not redraw.
If the same handling is not done here (I haven't looked at the code yet), then you would redraw every time the homescreen layout runs -- which happens after every silent call such as GetPublicKey.

mmilata:
Handling is [here](https://github.com/trezor/trezor-firmware/pull/2538/files#diff-79ba0a1db2566a1b482a077be5980c7b131ec9952629e76afa5f141bffb70e8fR36) but probably not good enough, gonna try it with suite.

mmilata:
Rebased to master so that @TychoVrahe can use this branch as a base and load the background image from storage (#2214).

mmilata:
> The homescreen is still redrawn sometimes, for example when suite wallet is opened.

I tried with suite and was able to reproduce one redraw and it is at the time suite asks for passphrase. Normally the device would show prompt to enter the passphrase but this is not implmented yet so the homescreen is kept on screen and then redrawn. I'm going to implement this prompt as part of #2217, if the issue persists I'll look into it again.

grdddj (commented):
Looking good, I do not have (m)any objections.

It just currently needs a rebase on `master` because of the `upy decrease` changes.

core/src/trezor/ui/layouts/tt_v2/homescreen.py
```diff
@@ -0,0 +1,139 @@
+from typing import Any, Tuple
+
+import storage.cache
+import storage.device
+from trezor import io, loop, ui, utils
+
+import trezorui2
+from apps.base import set_homescreen
+
+from . import _RustLayout
```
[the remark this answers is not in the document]
mmilata: I wanted to discourage use outside of `trezor.ui.layouts.tt_v2` but looks like this is not the way to do it, fixed in d7bc32ce0c12dc4e6a94df884bb8312ab31ca935.

core/mocks/generated/trezorui2.pyi
```diff
@@ -348,3 +348,36 @@ def show_remaining_shares(
...
+
+
+# rust/src/ui/model_tt/layout.rs
+def show_busyscreen(
+    *,
+    title: str,
+    description: str,
+    time_ms: int,
+    skip_first_paint: bool,
+) -> trezorui2.CANCELLED:
```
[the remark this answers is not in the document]
mmilata: Fixed in c28f9635e363c5bcd252447ce279e635e163a549. The types are not right anyway, the function doesn't return `CANCELLED` but returns `LayoutObj` whose event methods may return `CANCELLED`; same with the rest of the layout functions. It gets ignored by `interact()` so it mostly serves as documentation for programmer. No clear idea how to improve this.

I wouldn't add a suffix to the functions names, this should be solved by having different module name for different models.

core/embed/rust/src/ui/component/base.rs
```diff
@@ -104,6 +105,16 @@ impl<T> Child<T> {
         }
         result
     }
+
+    /// Do not draw on screen until an event requests paint. This is used by
+    /// homescreens to avoid flickerg when workflow restart happens.
```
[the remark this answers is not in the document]
mmilata: Fixed 06cd24715742be823b6f0f75e6b5c81d6f8e7baf.

core/embed/rust/src/ui/model_tt/component/homescreen.rs
```diff
@@ -0,0 +1,266 @@
...
+const LOCKED_Y: i16 = 101;
+const TAP_Y: i16 = 134;
+const HOLD_Y: i16 = 35;
+const LOADER_OFFSET: Offset = Offset::y(-10);
+const LOADER_DELAY: Duration = Duration::from_millis(500);
+const LOADER_DURATION: Duration = Duration::from_millis(2000);
+
+pub struct Homescreen<T> {
+    label: T,
+    notification: Option<(T, u32)>,
```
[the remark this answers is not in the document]
mmilata: Fixed in c28f9635e363c5bcd252447ce279e635e163a549.

TychoVrahe (approved):
I don't have any change requests, i will just reiterate that i am changing how the USB notification and preventing unwanted redraws work in the subsequent PR #2571 , since the way it works here is imo not sufficient. 

Also the way the notifications are passed from python to rust will probably need to be changed to something like 'pass list of active notifications' if we ever decide to work on this issue: 
https://github.com/trezor/trezor-firmware/issues/1298

matejcik (approved):
utACK, just two small comments

core/embed/rust/src/ui/event.rs
```diff
@@ -53,3 +53,15 @@ impl TouchEvent {
...
 }
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum USBEvent {
+    /// USB host has connected/disconnected.
+    Connected(bool),
+}
+
+impl USBEvent {
+    pub fn new(connected: bool) -> Self {
```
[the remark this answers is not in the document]
mmilata: [a6fa0d0](https://github.com/trezor/trezor-firmware/pull/2538/commits/a6fa0d0d03a725b5080a002d2c466cfc782bab9a)

tests/device_tests/ethereum/test_signtx.py
```diff
@@ -400,6 +400,11 @@ def input_flow_go_back(client: Client, cancel=False):
 )
 @pytest.mark.skip_t1
 def test_signtx_data_pagination(client: Client, flow):
+    # prevent this test from getting stuck on UI2
+    import os
+
+    assert os.getenv("UI2") != "1"
```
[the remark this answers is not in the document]
mmilata: [a6fa0d0](https://github.com/trezor/trezor-firmware/pull/2538/commits/a6fa0d0d03a725b5080a002d2c466cfc782bab9a)

matejcik (approved):

mmilata:
Rebased + fido test fix: f6847edba8dfe40a7bc5dd0c3f739cfc26898e48

PR #2539, opened 2022-10-04
Fix bug in SetBusy when running CI tests

andrewkozlik:
The resolution of the `ticks_ms()` is 62 bits in unix builds. On the CI machines which run continuously for days the value can overflow the 4 bytes that are available for it in the storage cache. This could cause `test_busy_state` to fail randomly.

- I increased the size of `APP_COMMON_REQUEST_PIN_LAST_UNLOCK` and `APP_COMMON_BUSY_DEADLINE_MS` to 8 bytes.
- I also introduced `set_int()` and `get_int()` which automatically deal with the encoding/decoding of integers in the storage cache. One of the reasons why this was so hard to debug is that in micropython `int.to_bytes()` does not fail when the value cannot be encoded into the specified number of bytes.

### DIFF
--- a/core/src/storage/cache.py
+++ b/core/src/storage/cache.py
@@ -227,6 +227,23 @@ def set(key: int, value: bytes) -> None:
     _SESSIONS[_active_session_idx].set(key, value)
 
 
+def set_int(key: int, value: int) -> None:
+    if key & _SESSIONLESS_FLAG:
+        length = _SESSIONLESS_CACHE.fields[key ^ _SESSIONLESS_FLAG]
+    elif _active_session_idx is None:
+        raise InvalidSessionError
+    else:
+        length = _SESSIONS[_active_session_idx].fields[key]
+
+    encoded = value.to_bytes(length, "big")
+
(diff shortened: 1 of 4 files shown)


core/src/storage/cache.py
```diff
@@ -227,6 +227,19 @@ def set(key: int, value: bytes) -> None:
...
+def set_int(key: int, value: int) -> None:
+    if key & _SESSIONLESS_FLAG:
+        length = _SESSIONLESS_CACHE.fields[key ^ _SESSIONLESS_FLAG]
+    elif _active_session_idx is None:
+        raise InvalidSessionError
+    else:
+        length = _SESSIONS[_active_session_idx].fields[key]
+
+    encoded = value.to_bytes(length, "big")
+    utils.ensure(int.from_bytes(encoded, "big") == value)
```
matejcik, line 239: I'm wondering whether this shouldn't be an `assert` instead of `ensure`. There seems to be very little value in this RSODing randomly in production?

matejcik (approved):

Issue #2542, opened 2022-10-04
Python typing: Include `py.typed` markers


unparalleled-js:
Currently, when using trezor as a dependency, the types information is not passed along when imported. When running `mypy` on consumers packages, you receive errors like:

```
error: Skipping analyzing "trezorlib.tools": module is installed, but missing library stubs or py.typed marker
```

https://dev.to/whtsky/don-t-forget-py-typed-for-your-typed-python-package-2aa3

PR #2543, opened 2022-10-04
fix: publish types to consumers and resolve downstream mypy issues

unparalleled-js:
fixes: #2542


matejcik (approved):
thanks!

matejcik:
please add a [changelog entry](https://docs.trezor.io/trezor-firmware/misc/changelog.html)

prusnak:
Late to the party, but is it intentional that the file `python/src/trezorlib/py.typed` is empty?

prusnak:
And I answer myself: yeah, it's OK since the file is just a marker.

Issue #2499, opened 2022-09-08
New UI: long passphrases


hynek-jina:

grdddj:
What are the requirements here? Looks like the passphrase already can handle 50 characters. @matejcik @mmilata

sime:
Yes, 50 character is the limit at least from the Trezor Suite POV.

grdddj:
Done in https://github.com/trezor/trezor-firmware/pull/2544

PR #2544, opened 2022-10-05
feat(core/rust/ui): support longer passphrases

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2499:
- enabling longer passphrases up to 50 characters - the limit was already there, just handling all the edge cases and visuals
- also fixing the change of the character categories - previously the swipes were not changing the buttons, as the new buttons were not placed - there is a need to re-place them on each swipe
  - we could solve it also by reusing the 10 already places buttons, and just changing the content of the buttons via `Button::set_content()` - in that case we would need to place them only once in the beginning and would not have to store the `PassphraseKeyboard::key_grid`, which is necessary for later placement

### DIFF
--- a/core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
+++ b/core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
@@ -85,19 +85,42 @@ impl PassphraseKeyboard {
         // Clear the pending state.
         self.input
             .mutate(ctx, |ctx, i| i.multi_tap.clear_pending_state(ctx));
-        // Make sure to completely repaint the buttons.
+        // Make sure to completely repaint the new buttons.
         for btn in &mut self.keys[key_page] {
             btn.request_complete_repaint(ctx);
         }
         // Reset backlight to normal level on next paint.
         self.fade = true;
+        // So that swipe does not visually enable the input buttons when max length
+        // reached
+        self.update_input_btns_state(ctx);
     }
 
+    /// Possibly changing the buttons' state after change of the input.
     fn after_edit(&mut self, ctx: &mut EventCtx) {
+        self.update_back_btn_state(ctx);
+        self.update_input_btns_state(ctx);
+    }
+
+    /// When the input is empty, disable the back button.
+    fn update_back_btn_state(&mut self, ctx: &mut EventCtx) {
         if self.input.inner().textbox.is_empty() {
             self.back.mutate(ctx, |ctx, b| b.disable(ctx));
         } else {
             self.back.mutate(ctx, |ctx, b| b.enable(ctx));
+        }
+    }
+
+    /// When the input has reached max length, disable all the input buttons.
+    fn update_input_btns_state(&mut self, ctx: &mut EventCtx) {
+        for btn in self.keys[self.scrollbar.active_page].iter_mut() {
+            btn.mutate(ctx, |ctx, b| {
+                if self.input.inner().textbox.is_full() {
+                    b.disable(ctx);
+                } else {
+                    b.enable(ctx);
+                }
+            });
         }
     }
 
@@ -131,18 +154,23 @@ impl Component for PassphraseKeyboard {
         self.scrollbar.place(scroll_area);
         self.scrollbar
             .set_count_and_active_page(PAGE_COUNT, STARTING_PAGE);
-        for (key, btn) in self.keys[self.scrollbar.active_page].iter_mut().enumerate() {
-            // Assign the keys in each page to buttons on a 5x3 grid, starting from the
-            // second row.
-            let area = key_grid.cell(if key < 9 {
-                // The grid has 3 columns, and we skip the first row.
-                key + 3
-            } else {
-                // For the last key (the "0" position) we skip one cell.
-                key + 1 + 3
-            });
-            btn.place(area);
+
+        // Place all the possible character buttons on all swipe-separated screens.
+        for swipe_screen in &mut self.keys {
+            for (key, btn) in swipe_screen.iter_mut().enumerate() {
+                // Assign the keys in each page to buttons on a 5x3 grid, starting
+                // from the second row.
+                let area = key_grid.cell(if key < 9 {
+                    // The grid has 3 columns, and we skip the first row.
+                    key + 3
+                } else {
+                    // For the last key (the "0" position) we skip one cell.
+                    key + 1 + 3
+                });
+                btn.place(area);
+            }
         }
+
         bounds
     }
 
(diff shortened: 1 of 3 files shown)


core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
```diff
@@ -124,25 +174,16 @@ impl Component for PassphraseKeyboard {
...
-            let area = key_grid.cell(if key < 9 {
-                // The grid has 3 columns, and we skip the first row.
-                key + 3
-            } else {
-                // For the last key (the "0" position) we skip one cell.
-                key + 1 + 3
-            });
-            btn.place(area);
-        }
+        self.place_current_buttons();
```
mmilata, line 186: Can you just place all the pages here? IIUC there's no need to redo this after every page change.

grdddj: Thanks for the suggestion, done in [f409b83](https://github.com/trezor/trezor-firmware/pull/2544/commits/f409b8386cbe5fdb4dc4ba93211f802acb213a62)

mmilata (changes requested):
Mostly looking good, however having 20 as a limit only really works with monospace font. It shouldn't be too difficult to find the longest suffix that fits the textbox (minus length of `...`) using given font.

emu00000000 emu00000001

grdddj:
> Having 20 as a limit only really works with monospace font. It shouldn't be too difficult to find the longest suffix that fits the textbox (minus length of `...`) using given font

Good point, changed in [b94299d](https://github.com/trezor/trezor-firmware/pull/2544/commits/b94299d0ab5eb804ef906278f3023cb53118837e)

In [6209817](https://github.com/trezor/trezor-firmware/pull/2544/commits/62098175a447a8c31e746ced114e990c4bc129b6) I have added some helper macros (copy-paste from `model R` branch) that are helpful during development, and I also already used the `build_string!` to concatenate the ellipsis with the right side of the text

core/embed/rust/src/ui/macros.rs
```diff
@@ -4,3 +4,35 @@ macro_rules! include_res {
...
+/// Transforms bool into string slice. For example for printing.
+macro_rules! booltostr {
+    ($bool:expr) => {{
+        if $bool {
+            "true"
+        } else {
+            "false"
+        }
+    }};
+}
```
mmilata, line 38: This can be just a regular function as opposed to macro, or am I missing something?

core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
```diff
@@ -252,20 +288,53 @@ impl Component for Input {
...
+            if style.font.text_width(text) <= self.area.width() {
+                // It can fit the whole passphrase, no need to do anything.
+                String::from(text)
+            } else {
+                // Passphrase is longer, showing its right end with ellipsis at the beginning.
+                let ellipsis = "...";
+                let ellipsis_width = style.font.text_width(ellipsis);
+                let awailable_width_for_text = self.area.width() - ellipsis_width;
+
+                // Finding out how many characters will fit in, starting from the right end.
```
mmilata, line 302: Nit: turning this into a method on `Font` like
```rust
fn longest_suffix(self, width: i16, text: &str) -> &str
// or
fn longest_suffix(self, width: i16, text: &str) -> usize
```
would make this a bit more readable and also potentially reusable.

Also there doesn't seem to be a need to build a string out of this, you can just display the two pieces of text directly?

grdddj: Good idea with creating a `Font` method for this.

With displaying two pieces of text, how can I do it without needing to call the `display::text` two times? (As that would create a lot of extra logic - sometimes calling it once, sometimes twice.)

grdddj: `Font` method done in [6e21a58](https://github.com/trezor/trezor-firmware/pull/2544/commits/6e21a58734413143eb851a2489463dfdc62e8f74)

mmilata: Why do you want to avoid calling it twice? Was thinking of something along the lines of:

```rust
let text_to_display = if fits_whole {
    text
} else {
    let ellipsis_width = ...;
    display_text(text_baseline, "...");
    text_baseline += ellipsis_width; // needed for paint_pending_marker
    text_suffix
}

display::text(text_baseline, text_to_display);
```

grdddj:
I have just noticed that in some cases the very-right pixel of `pending_marker` stays on the screen. It is because it paints itself one pixel longer than the last character, and can therefore be outside of area cleaned by `display::rect_fill`.

We could avoid it by not painting the marker longer than the letter, or by cleaning an area one pixel longer. My preference is probably the first option, as that respects the possible area boundary, and it cannot write into another area (at least not in `x` direction) - it would however make it look not-so-good.

image

core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
grdddj: Tried as an experiment in [2ef1b2a](https://github.com/trezor/trezor-firmware/pull/2544/commits/2ef1b2a7735b778d1a65cd4da17bfeae1b55b64c) - not sure if I like it, as it meant quite a lot of changes. But it removes the need for `heapless::String`, which is certainly helpful

core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
```diff
@@ -285,46 +283,52 @@ impl Component for Input {
...
+
+        // Find out how much text can fit into the textbox.
+        let text_to_display = if style.font.text_width(text) <= self.area.width() {
+            // It can fit the whole passphrase.
+            text
+        } else {
+            // Passphrase is longer, showing its right end with ellipsis at the beginning.
+            let ellipsis = "...";
+            let ellipsis_width = style.font.text_width(ellipsis);
+            let awailable_width_for_text = self.area.width() - ellipsis_width;
```
mmilata: available

mmilata (commented):
I don't really understand why we draw the marker 1px longer but judging by the comment it's intentional :shrug:
Otherwise this looks good to me, please drop the unused macros and squash.

core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
grdddj: Fixed in squash

mmilata (approved):

grdddj:
Force-pushed after rebase on `master`

bosomt:
**QA OK**

works as expected /both options - entering via host or via device/

I only noted that when you enter small characters like `,.` via device it looks strange that they dont fit to device display but they only take less than half of it ;) but its expected.

image

**Info:**
 - Suite version: desktop 22.10.3 (540db5426ce88c33bdda2fdae0ceb50f6cbc4a23)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.10.3 Chrome/100.0.4896.160 Electron/18.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.3 Universal (revision 8668eba9364b0dfb5fa5250a5366421d6ec1dde1)
 - Transport: bridge 2.0.31

grdddj:
> 
> image
> 

Looking at the photo, it seems to me this is the "legacy" `UI`, and not the "new" `UI2` - visible by the "ellipsis" at the beginning of very long text - old `UI` uses "<" (on photo) but the `UI2` uses "...". Also, we currently do not have a hard limit of characters fitting the screen, we calculate it based on character width.

(Unless you really typed "<" as the first character :))

Did you build the firmware with `UI2=1` option?

bosomt:
@grdddj no i just downloaded latest dev master build

maybe that's why i have not seen any change when i tried to test:
```
 -  TT redesign: special screens for altcoins #2500 
 -  each icon (for IconDialog) takes 0.5K of flash, we need to either figure out how to represent three-color image more efficiently than 16bpp, or draw them procedurally #2558 
```

where can i gat correct build that i can upload to device ?

mmilata:
Currently we don't even build it in CI. We're planning to switch to UI2 (#2580) shortly _after_ the upcoming release so binaries in the usual places will have it. New & exciting bugs to be expected at that point.

If you really want to test this now here's today's master with UI2=1 (bitcoin-only): [firmware.bin.zip](https://github.com/trezor/trezor-firmware/files/9867489/firmware.bin.zip)

bosomt:
@mmilata we will wait after current release. /but i will try that UI2.0 build :) /
Please move all UI related issues to new sprint when freezing current one or creating next one.

image

mmilata:
CC @hynek-jina ^

Also for the record UI2 = New UI = Rust UI (noted that this pr has neither in title).

Issue #2500, opened 2022-09-08
TT redesign: special screens for altcoins


hynek-jina:
most altcoin functionality is covered by `confirm_text` / `confirm_blob` / `confirm_properties`.

However, there are several custom screens:
* `confirm_total_ethereum` -> should be converted to a sequence of "confirm fee" -> "confirm fee" -> "confirm total" or something along these lines
* `confirm_total_ripple` -> should be removed in favor of "confirm output" -> "confirm fee" -> "confirm total"
* `confirm_transfer_binance` -> should be converted to `confirm_properties` for inputs, `confirm_output` for outputs. Unfortunately there doesn't seem to be a sensible way to show total, so some sort of "really confirm? (hold)" with no further info is probably the way to go.
* `confirm_decred_sstx_submission` -> figure something out? probably a `confirm_properties`?

grdddj:
Below are the custom screens in their current look.

To make sure, the goal here is to make it look exactly the same, or just "somehow make it work"?

![image](https://user-images.githubusercontent.com/42543243/189884974-dbfe21c7-fd3e-4d6d-80eb-abf17a6f52aa.png)
![image](https://user-images.githubusercontent.com/42543243/189885314-17aa4076-bf05-46c9-878e-8d991831c02a.png)
![image](https://user-images.githubusercontent.com/42543243/189885031-e92583c6-6325-40e6-b132-90d112b939e8.png)
![image](https://user-images.githubusercontent.com/42543243/189885058-dc2eb261-5d7d-4470-a096-fa9df5554462.png)
![image](https://user-images.githubusercontent.com/42543243/189885010-09f78cb1-84b3-42d6-910a-71eaf3b2b90b.png)

grdddj:
Done in https://github.com/trezor/trezor-firmware/pull/2545

Issue #2531, opened 2022-09-29
`confirm_blob` cannot directly render bytearray data, we currently hexlify it in python but encoding it on-the-fly would save some RAM


hynek-jina:

mmilata:
+ line breaking regardless of spaces and without hyphens

https://github.com/trezor/trezor-firmware/pull/2545#issuecomment-1277298750

PR #2545, opened 2022-10-05
TT redesign - altcoin screens

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2500:
- implementing some basic altcoin screens by reusing already existing components

We might even delete `altcoin.py` in the future (when switching to `UI2`) and call these things directly from the altcoin applications.

Some of the flows are a little different from the UX perspective, but given these are rarely used altcoins, it should not matter much. (And all could be changed if we want)

Comparison with "old" UI:
# confirm_total_ethereum
image
image image image

# confirm_total_ripple
image 
image image

# confirm_transfer_binance
image image
image image

# confirm_decred_sstx_submission
image
image


matejcik:
> confirm_total_ethereum

there's a "hold to confirm" in the first screen that should not be there
otherwise awesome, much better than the old version



> We might even delete `altcoin.py` in the future (when switching to `UI2`) and call these things directly from the altcoin applications.

yes, these should be inlined to `<app>/layout.py` which usually already exists

grdddj:
> there's a "hold to confirm" in the first screen that should not be there

Changed to classic "Next" in [3239509](https://github.com/trezor/trezor-firmware/pull/2545/commits/3239509b73cc6577a4572e44eab848605400a9fa)

image

mmilata (changes requested):
Looking good, can you please also implement `confirm_address` and `confirm_text` which are not in `altcoin.py` but only used by altcoins? Also `confirm_sign_identity` if you have the time:pray:

matejcik:
fwiw `confirm_blob` should NOT be using hyphens and should break by character (not by whitespace) in all cases

grdddj:
> fwiw `confirm_blob` should NOT be using hyphens and should break by character (not by whitespace) in all cases

Naive, but working solution suggested done in [2648ba8](https://github.com/trezor/trezor-firmware/pull/2545/commits/2648ba8697fa18549a4646cf32ff5e08176da43e) - it just does not include the hyphen, but leaves an empty space there (not something users would notice/care about). We surely can do something cleaner with more effort.

I have also changed the optional `challenge_visual` / date into `description` field of `confirm_blob`, so it is better visible and divided from the `identity`. It is a question if we want the `identity` to be in `mono`. Also, we could theoretically use `confirm_output`, which also renders one line in `TEXT_NORMAL` and one in `TEXT_MONO` - same as `confirm_blob` with `extra=None`.

image
image

mmilata:
Thanks for implementing the missing layouts. Instead of confirm_general I'd suggest merging confirm_output and confirm_total to something like confirm_value, or perhaps even dropping both and using confirm_blob.

I don't like [2648ba8](https://github.com/trezor/trezor-firmware/pull/2545/commits/2648ba8697fa18549a4646cf32ff5e08176da43e) very much, I think this should be handled by introducing a new variant to `enum LineBreaking`, handling it in `Span::fit_horizontally` and probably making it a default for `TEXT_MONO`? I can do this as part of #2531 if you want.

grdddj:
Reverting the "hyphen" changes for `confirm_blob` in [2e78e7f](https://github.com/trezor/trezor-firmware/pull/2545/commits/2e78e7fd2551ab9f3e0438354a4427c5d5ae3c1f)

---

> Instead of confirm_general I'd suggest merging confirm_output and confirm_total to something like confirm_value, or perhaps even dropping both and using confirm_blob.

Do you mean this on `micropython` or `Rust` side? They are all quite similar on both sides.

mmilata:
At least on rust side, possibly on the python side too, up to you.

grdddj:
Rebasing to `master` to get the `UI` recording feature.

[f6d003f](https://github.com/trezor/trezor-firmware/pull/2545/commits/f6d003f9f5b73cb74635f5b0e662b80e5281ecdc) includes some refactorings outlined above - merging `confirm_output` and `confirm_total` into `confirm_value` on the `Rust` side and reusing `confirm_value` in many `confirm_*` layouts in `micropython`.

It has the size benefit of `512` bytes (in reality a little more because of the alignment) which was achieved by reusing a lot of common code - mostly `_confirm_blob` helper  in `Rust`, being used by `confirm_blob` and `confirm_value` .

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -483,35 +495,25 @@ async def confirm_output(
...
-                description="To:",
-                value=address,
-            )
-        ),
+        title,
+        address,
+        "To:",
         "confirm_output",
         br_code,
+        verb="Next",
```
mmilata, line 505: Nit: make these uppercase for consistency with the rest of the file.

mmilata (approved):
LGTM, please squash before merging. Are you planning to add the Monero layouts to this PR?

core/src/trezor/ui/layouts/tt_v2/__init__.py
grdddj: Changed in force-push

grdddj:
> Are you planning to add the Monero layouts to this PR?

Probably not, I would do it in some future PR, to have this merged already.

grdddj:
Sorry for the second force-push, I have missed some `verb="Next"` uncapitalized arguments

grdddj:
Rebased on `master` and resolved conflicts in `layout.rs` - the new way of creating the paragraphs

mmilata:
UI diff https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3216882457/artifacts/master_diff/index.html

Issue #2547, opened 2022-10-06
Add trezorctl command for screen capture


grdddj:
To make screenshotting/UI visual testing easier, it is worth creating a `trezorctl` command that tells `Trezor` to save screenshots into a specified directory

prusnak:
Is that for emulator only?

grdddj:
> Is that for emulator only?

I think so. It would be really nice if it worked with hardware device in `debug` mode, but I do not think it does - it would need to send the data through `USB` and something would need to save it on the host. I will however try

grdddj:
Done in https://github.com/trezor/trezor-firmware/pull/2548

PR #2548, opened 2022-10-07
feat(python): add trezorctl possibility to record screen changes

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2547:
- allowing for `trezorctl` specifying the directory where all the screen changes from emulator will be saved

There are two possibilities how to do that, each with a little different use-case:

1. Via `-r / --record` option directly in `trezorctl` command, so it can be used together with any other action - e.g. `trezorctl -r screenshots ping "Hello world" -b`. This allows for recording separate `trezorctl` actions.
2. Via separate `record` command - `trezorctl record screenshots`, it also allows for stopping the recording with `trezorctl record -s/--stop`. This will save all the screens from the current "session" - until the emulator does not die.

Both will do basically the same - create `screenshots` directory in current `pwd`, if it does not exist, generate a unique subdirectory based on datetime (like `screenshots/2022-10-07_13-09-48`) and will start saving new screen refresh pictures into it. 

The reason of creating subdirectories for every call/every `emulator` session is that `emulator` names the screens `refresh-00000000.png`, `refresh-00000001.png` etc., resetting the counter at every emulator restart. When it would always save into the same directory, new screens would overwrite older ones.

Option number 1 currently does not stop recording after that `trezorctl` returns, so it will be recording further screens from emulator into the same dir (unless the next command specified their own `-r / --record`). It should be possible to stop recording at the end when recording started this way - probably with `@cli.set_result_callback()`?. 

I was not sure where to put the `record_screen` "helper" function (that gets called in two places), so I left it in `cli/trezorctl.py`. It might be moved outside of `cli`, maybe into `device.py` or `debuglink.py`. (It might be useful as a function to be called from scripts using `trezorlib`.)

### DIFF
--- a/python/src/trezorlib/cli/trezorctl.py
+++ b/python/src/trezorlib/cli/trezorctl.py
@@ -186,6 +186,11 @@ def configure_logging(verbose: int) -> None:
     help="Resume given session ID.",
     default=os.environ.get("TREZOR_SESSION_ID"),
 )
+@click.option(
+    "-r",
+    "--record",
+    help="Record screen changes into a specified directory.",
+)
 @click.version_option(version=__version__)
 @click.pass_context
 def cli_main(
@@ -196,6 +201,7 @@ def cli_main(
     passphrase_on_host: bool,
     script: bool,
     session_id: Optional[str],
+    record: Optional[str],
 ) -> None:
     configure_logging(verbose)
 
@@ -207,6 +213,10 @@ def cli_main(
             raise click.ClickException(f"Not a valid session id: {session_id}")
 
     ctx.obj = TrezorConnection(path, bytes_session_id, passphrase_on_host, script)
+
+    # Optionally record the screen into a specified directory.
+    if record:
+        debug.record_screen_from_connection(ctx.obj, record)
 
 
 # Creating a cli function that has the right types for future usage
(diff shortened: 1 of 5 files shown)


grdddj:
> Option number 1 currently does not stop recording after that trezorctl returns, so it will be recording further screens from emulator into the same dir (unless the next command specified their own -r / --record). It should be possible to stop recording at the end when recording started this way - probably with @cli.set_result_callback()?.

Stopping the recording when it was started as direct `trezorctl -r` option - [39e819b](https://github.com/trezor/trezor-firmware/pull/2548/commits/39e819be2c3a1bf14cc6375f98de1f4d4dcf7d21)

> I was not sure where to put the record_screen "helper" function (that gets called in two places), so I left it in cli/trezorctl.py. It might be moved outside of cli, maybe into device.py or debuglink.py. (It might be useful as a function to be called from scripts using trezorlib.)

Moving the main logic into `debuglink.py`, and also creating its first useful use-case by calling it from `emu.py`, where we can pass `-r screen_dir` and all the changes from emulator will save there - [39e819b](https://github.com/trezor/trezor-firmware/pull/2548/commits/39e819be2c3a1bf14cc6375f98de1f4d4dcf7d21)

python/src/trezorlib/cli/trezorctl.py
```diff
@@ -370,6 +400,29 @@ def wait_for_emulator(obj: TrezorConnection, timeout: float) -> None:
     LOG.info(f"Waited for {end - start:.3f} seconds")
 
 
+@cli.command()
+@click.argument("directory", required=False)
+@click.option("-s", "--stop", is_flag=True, help="Stop the recording")
+@click.pass_obj
+def record(obj: TrezorConnection, directory: Union[str, None], stop: bool) -> None:
```
matejcik, line 407: this command should live under `debug`

we should also enforce that recording is only available on emulator. IIRC we can detect the emulator via its vendor string in `client.features`

grdddj: Moved under `debug` and checking the emulator in [1b9bcb0](https://github.com/trezor/trezor-firmware/pull/2548/commits/1b9bcb06c8424f0902c19ff7298f18c270d425f7)

grdddj: Check for emulator moved deeper into the "main" `record_screen` function, so it also raises when used with hardware device - [5f97686](https://github.com/trezor/trezor-firmware/pull/2548/commits/5f976867b6826ec606a8b130ee9ac614f17c828e)

matejcik (approved):

PR #2549, opened 2022-10-10
Fix ASAN build

TychoVrahe:
This PR fixes asan builds, which are broken since the introduction of num_derive crate.


matejcik (approved):
utACK

mmilata (approved):
Builds fine, thanks!

mmilata:
Seems to be still broken in CI though:( https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3153022709

mmilata:
... because `test_rust` is invoked from Makefile and `--target` is missing there.

PR #2552, opened 2022-10-11
Rust UI: zero out memory after using pin/passphrase layout

### LINKED ISSUE #2530: Zero our memory of PIN/passphrase in the layout object

mmilata:
Fixes #2530. Originally I wanted to handle this in `msg_try_into_obj` but it would require making mutable a lot of references that previously weren't, and would make it easy to use the `Child` abstraction in a wrong way. Adding an event seems much less intrusive.

### DIFF
--- a/core/embed/rust/src/ui/util.rs
+++ b/core/embed/rust/src/ui/util.rs
@@ -1,3 +1,6 @@
+use heapless::{String, Vec};
+use zeroize::{DefaultIsZeroes, Zeroize};
+
 pub trait ResultExt {
     fn assert_if_debugging_ui(self, message: &str);
 }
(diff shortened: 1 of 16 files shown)


core/embed/rust/src/ui/util.rs
```diff
@@ -31,6 +34,20 @@ pub fn u32_to_str(num: u32, buffer: &mut [u8]) -> Option<&str> {
     }
 }
 
+pub fn zeroize_vec<T, const N: usize>(val: &mut Vec<T, N>)
+where
+    T: DefaultIsZeroes,
+{
+    let slice: &mut [T] = val.as_mut();
```
mmilata, line 41: This only gets the slice spanning the part of the buffer that is currently filled, i.e. typing your pin and backspacing it will result in it staying in the buffer after zeroize.

You can get the whole buffer with
```rust
let ptr = val.as_mut_ptr();
let slice = core::slice::from_raw_parts_mut(ptr, N * core::mem::size_of::<T>());
```
however `from_raw_parts_mut` requires that

> `data` must point to `len` consecutive properly initialized values of type `T`

and the values in heapless::Vec start as uninitialized. Is this to ensure that destructors are called upon overwriting the data (and therefore safe for u8 which doesn't have destructors), or could this still result in UB?

I realize this is rather obscure corner case, just curious what would be the right way to solve it.

matejcik: one thing comes to mind is to support this only for `Vec<T: Default>` and manually fill the vec up to capacity before zeroizing

of course, a bad thing about all of this is the generic explosion

mmilata: TBH we'll be fine with just `Vec<u8>`

mmilata: done in [1878b46](https://github.com/trezor/trezor-firmware/pull/2552/commits/1878b46b89510a06d16de8ec02b7db98c5b9784d)

matejcik:
closing, IIRC there were some problems with the whole approach and it's probably the wrong way to go about it. 

some ideas:
* maybe we should add zeroization directly to the GC
* maybe none of this matters if we plan to restart the micropython VM, so we can clear the memory then

PR #2553, opened 2022-10-12
Rust UI: replace TT Hoves Medium with DemiBold

mmilata:
Also remove TT Hoves vector files to avoid licensing issues.

[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3160774946/artifacts/master_diff/index.html)


TychoVrahe (approved):
LGTM. I am just wondering what is the reason to introduce fifth font type instead of using `MEDIUM` for this?

matejcik:
i think the enum contents are most telling: there are never five possibilities, only four.
please keep the four options, i have no opinion on whether to _rename_ "medium" to "demibold".

prusnak:
Are we keeping MEDIUM in Bootloader and DEMIBOLD is only used in the Firmware? If yes, why not change MEDIUM to DEMIBOLD in Bootloader too?

PR #2554, opened 2022-10-12
Fix rust tests with ASAN

TychoVrahe:
by adding target to rust tests invoked from makefile


mmilata (approved):

TychoVrahe:
added the target to `clippy` too, since it was failing on `ASAN` build as well.

mmilata (approved):

PR #2555, opened 2022-10-12
Re-introduce touch event filtering

TychoVrahe:
This reintroduces touch event filtering at the driver level.

The previous removal (in https://github.com/trezor/trezor-firmware/pull/2412) caused that UI was spammed with same events when user was just holding, making the UI less responsive to other events, like timers.


matejcik (approved):
utACK, makes sense

hiviah (approved):
utACK, the debounce makes sense.

Issue #2532, opened 2022-09-29
FIDO layouts


hynek-jina:

bosomt:
**QA OK**

Works as expected /tested on Google and GitHub account/
Noted one thing, that register / authenticate message and web log/name is not properly aligned but its probably expected /it matches screenshots from PR/

<img width="362" alt="image" src="https://user-images.githubusercontent.com/31506317/207592295-af1b5512-1c67-4745-a538-05afa03140da.png">

<img width="347" alt="image" src="https://user-images.githubusercontent.com/31506317/207592340-da8859bf-6a0d-4dcc-8327-a978d5efa9c3.png">


**Info:**
 - Suite version: desktop 22.12.1 (ddaacdb2245d7a909b3d2abfe918cff34d2e1609)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.12.1 Chrome/104.0.5112.124 Electron/20.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.4 Universal (revision ff35c77bad82c7acac1df6245940ebe0d6454642)
 - Transport: bridge 2.0.32

Hannsek:
@mmilata Did you fix the layout? (https://satoshilabs.slack.com/archives/C02BY51RD8R/p1669232972198689)

mmilata:
No

Hannsek:
Then please fix the layout.

Issue #703, opened 2019-11-13
Rename webauthn, u2f, etc. to "fido"


matejcik:
The functionality that was originally U2F and grew to include WebAuthN and FIDO2 is named ... inconsistently.

A good name for the whole complex is `fido`. This name should be used in protobuf messages, app/module names, etc.

This will be implemented in trezorlib/trezorctl. An open question is, should we rename it everywhere else too? Is it worth the hassle?

prusnak:
Places where a rename can/should take place:

- [ ] app name `core/src/apps/webauthn` -> `core/src/apps/fido`
- [ ] protobuf messages `{GetNext,Next,Set}U2FCounter` -> `{GetNext,Next,Set}FIDOCounter` (`messages-management.proto`)
- [ ] protobuf messages `WebAuthn*` -> `FIDO*` (`messages-webauthn.proto`)
- [ ] storage: `U2F_COUNTER` -> `FIDO_COUNTER`

matejcik:
> * [ ]  protobuf message names `{GetNext,Next,Set}U2FCounter` -> `{GetNext,Next,Set}FIDOCounter` (`messages-management.proto`)
> * [ ]  protobuf messages names `WebAuthn*` -> `FIDO*` (`messages-webauthn.proto`)

`*FIDOCounter` should be all moved to `messages-fido.proto` too

PR #2556, opened 2022-10-12
TT redesign: FIDO/webauthn layouts

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2532:
- implementing `webauthn` screens

MISSING: I was not able to replicate the situation when `confirm_webauthn` gets non-None `pageable` object. We do not have any tests for it and it is triggered from some very deep place in `fido` application. (throws `NotImplementedError` currently)

On the `Rust` side, I tried to reuse the `IconDialog`, which seems to be quite fine for the UI-tested use-cases. I am not sure how it handles the case when the information cannot fit a single line - in `UI1`, it is handled by `text_center_trim_left` and `text_center_trim_right`. We might add some longer string into the `UI` tests.

Also, unlike in all other current layouts, where the icon is hardcoded on the `Rust` side, here we need to send the icon bytes from `micropython`, as it has a database of all `webauthn` applications with their corresponding icons. Done with `unsafe fn get_buffer`, not sure if it is the best way.

image
image


TychoVrahe:
The webauthn `knownapps.py` db actually doesn't store the icon, just a filename. The icon is then loaded from universal `resources.py` where its searched for by that filename. I am wondering if it wouldn't be better to store the toifs in rust and have the lookup there? we could get rid of `resouces.py` (and related tools) eventually that way

grdddj:
> The webauthn `knownapps.py` db actually doesn't store the icon, just a filename. The icon is then loaded from universal `resources.py` where its searched for by that filename. I am wondering if it wouldn't be better to store the toifs in rust and have the lookup there? we could get rid of `resouces.py` (and related tools) eventually that way

Yes, we could store the icon data itself in `Rust`, I wanted to avoid it, however.

In the current state, all the "db" operations are nicely centralized in `micropython`, which takes care of everything. If we put the icon data in `Rust`, it would create unnecessary dependence between `upy/Rust` - adding any new `fido/webauthn` app would mean having to change both sides.

Is there some disadvantage/danger in sending the raw icon bytes (or pointer to it?) from `micropython`? I imagine it can be handy for some other future use-cases (when the icon we want to render is highly dynamic). "Just give me icon data and I will render it" seems quite attractive to me (given proper error handling).

It is true that _eventually_ (if we decide to get rid of `micropython`), it would move to `Rust` - all the generated `.mako` templates would need to be migrated into its `Rust` form.

Let's also ask what others think about this.

mmilata:
Haven't reviewed the code yet, just wanted to point out that there's no need to use templates, we have `include_res!` macro for pulling icons into firmware.

core/src/trezor/ui/layouts/tt_v2/webauthn.py
```diff
@@ -13,8 +19,39 @@ async def confirm_webauthn(
     info: ConfirmInfo,
     pageable: Pageable | None = None,
 ) -> bool:
-    raise NotImplementedError
+    if pageable is not None:
+        # TODO: how to even replicate this situation?
```
mmilata, line 23: A bit tricky, first apply this patch (will try to turn this into PR):
```patch
diff --git a/vendor/fido2-tests/tests/vendor/trezor/utils.py b/vendor/fido2-tests/tests/vendor/trezor/utils.py
index fa4ba3c..8339b57 100644
--- a/vendor/fido2-tests/tests/vendor/trezor/utils.py
+++ b/vendor/fido2-tests/tests/vendor/trezor/utils.py
@@ -9,11 +9,12 @@ from trezorlib.transport import enumerate_devices, get_transport
 
 
 def get_device():
+    interact = os.environ.get("INTERACT") == "1"
     path = os.environ.get("TREZOR_PATH")
     if path:
         try:
             transport = get_transport(path)
-            return TrezorClientDebugLink(transport)
+            return TrezorClientDebugLink(transport, auto_interact=not interact)
         except Exception as e:
             raise RuntimeError("Failed to open debuglink for {}".format(path)) from e
 
@@ -21,7 +22,7 @@ def get_device():
         devices = enumerate_devices()
         for device in devices:
             try:
-                return TrezorClientDebugLink(device)
+                return TrezorClientDebugLink(device, auto_interact=not interact)
             except Exception:
                 pass
         else:
```

Then start an emulator and run `INTERACT=1 make test_emu_fido2 TESTOPTS="-v -k multiple_rk"`.

The important part here is to create a layout that accepts list of (app_name, account_name) and looks like what you trezorui2.confirm_webauthn except it allows you to choose the account name by swiping, and needs to return an index of the account along with CONFIRMED/CANCELLED. I suppose we can worry about integrating it with `fido2.py` later.

core/src/trezor/ui/layouts/tt_v2/webauthn.py
```diff
@@ -13,8 +19,39 @@ async def confirm_webauthn(
...
+        return is_confirmed(
+            await interact(ctx, confirm, "confirm_webauthn", ButtonRequestType.Other)
+        )
 
 
 async def confirm_webauthn_reset() -> bool:
-    raise NotImplementedError
+    confirm = _RustLayout(
+        trezorui2.confirm_action(
+            title="FIDO2 Reset",
```
mmilata, line 51: In the new design all titles are uppercase.

core/src/trezor/ui/layouts/tt_v2/webauthn.py
mmilata: trezor/fido2-tests#2

core/src/trezor/ui/layouts/tt_v2/webauthn.py
grdddj: Changed in [3d79a2d](https://github.com/trezor/trezor-firmware/pull/2556/commits/3d79a2d30c9e4a6a8290fd9a32e1bfcf1355476f)

grdddj:
[3d79a2d](https://github.com/trezor/trezor-firmware/pull/2556/commits/3d79a2d30c9e4a6a8290fd9a32e1bfcf1355476f) moves all the icon data into `Rust`. From `micropython` we just send the `icon_name`.

Space-wise it is interesting that the `new_confirm_webauthn` has `2_676` bytes, which includes all the icon data (there are no mentions of `webauthn.rs` or `get_icon_data`).

Comparing it with the previous-state when icons resided in `micropython`, it seems their transition saves `1_679_952 - 1_677_904 ` ~= `2_000 bytes`

Looking at the exact diff of two firmware binaries:
```
.flash2    872     1    L872     D0       embed/rust/src/ui/model_tt/layout.rs                         new_confirm_webauthn()
.flash2    108     1    L108     D0       mp_qstr_frozen_const_pool
.flash2    8       162  L8       D0       OUTLINED_FUNCTION
.flash2    -1      1    L-1      D0       src/trezor/ui/layouts/tt_v2/webauthn.py                      confirm_webauthn()
.flash2    -4      1    L-4      D0       embed/rust/src/ui/model_tt/layout.rs                         new_request_pin()
.flash2    -39     1    L-39     D0       src/trezor/ui/components/common/webauthn.py                  ConfirmInfo.load_icon()
.flash2    -700    36   L-140    D-560    src/apps/webauthn/knownapps.py                               by_rp_id_hash()
.flash2    -835    1    L-835    D0       [section .flash2]
.flash2    -1_457  58   L-561    D-896    src/trezor/res/resources.py                                  load_resource()
SUMMARY: 9 rows, -2_048 bytes in total (L-592 D-1_456).
```

I have just realized that this `PR` probably should not delete the icons from `micropython`, as they are still in use by `UI1`. But at least we got this size comparison out of this mistake. I will need to revert the `micropython` deletions and make it somehow work for both UIs at the same time (unless we merge this into `TT-UI2-only` branch, which might be actually the best).

TychoVrahe:
well few hundreds bytes are saved by this transition (and second time in the lookup function) `icon="apps/webauthn/res/icon_dashlane.toif"` ->
            `icon_name="dashlane"`,

mmilata:
Let's not do `TT-UI2-only` yet please.

grdddj:
> Let's not do `TT-UI2-only` yet please.

Sure, [9f0aeea](https://github.com/trezor/trezor-firmware/pull/2556/commits/9f0aeeaac030f544c3ac8a0cdeb634867af176b8) brings back the `micropython` deletions.

Now it is compatible with both at the same time (at the small cost of defining both `icon` and `icon_name`)

grdddj:
[f5d3c65](https://github.com/trezor/trezor-firmware/pull/2556/commits/f5d3c6546f27a7e029f71d93a88a87d32283fdc6) implements the pagination possibility of choosing among many credentials.

image 
image image image
image


`confirm_webauthn` was modified to return a `tuple[bool, int]` to signal whether it was confirmed and on which screen index. "Normal" confirmation with just one credential reuses this as well, it just does not care about the `int` part. UI-wise, we are not showing the scrollbar when there is only one choice.

Considerations:
- it currently uses `scrollbar` to show users it is scrollable, instead of the arrows as in `UI1` - these can be easily added as well, maybe even with combination with the scrollbar
- the `fido2` tests are currently broken, because `debugmode` does not have the possibility to send `tuple[bool, int]` as its response from layout - it sends `trezorui.CONFIRMED` and that breaks the flow on `micropython` side
  - we may need to do something about it, or decide to ignore this, as no other layout have this "issue" currently

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -585,6 +602,62 @@ extern "C" fn new_show_error(n_args: usize, args: *const Obj, kwargs: *mut Map)
...
+
+        let controls = Button::cancel_confirm(
+            Button::with_icon(theme::ICON_CANCEL).styled(theme::button_cancel()),
+            Button::with_text("CONFIRM").styled(theme::button_confirm()),
+            2,
+        );
+
+        let fido_page = FidoPaginatedPage::new(pages, icon, controls);
+
+        let obj = LayoutObj::new(Frame::new(title, fido_page))?;
```
mmilata, line 655: The right border is too thin, fix with:
```suggestion
        let obj = LayoutObj::new(Frame::new(title, fido_page).with_border(theme::borders()))?;
```

Frame normally expects the inner component to have vertical scrollbar which is not the case here. This is quite ugly but using negative borders seemed uglier to me and couldn't think of anything else.

mmilata:
Regarding scrollbar vs. arrows, that should be up to UI people. Can we please have a design for this screen @Hannsek / @sime?

Title should be uppercase. Will review the code this week.

core/embed/rust/src/ui/model_tt/layout.rs
grdddj: Done in [e47408f](https://github.com/trezor/trezor-firmware/pull/2556/commits/e47408f74d63837b6d43ba0266f49ee74bd3f0e3)

All other places using `with_border` combine it with `into_child`, so I made it as well

mmilata: Yep, it should be everywhere (but isn't). I thought it wouldn't compile with anything else but actually got confused by the trait names. Will add this to my TODO.

sime:
Not familiar with the use case that FIDO is scrollable.

Browsers can ask to authenticate multiple times?

grdddj:
> Browsers can ask to authenticate multiple times?

It seems like there could be multiple credentials available for the same site/vendor/app, like for Google. It offers a scrollable choice of all your "accounts" and you choose one of them.

andrewkozlik:
> Browsers can ask to authenticate multiple times?

It is possible for resident credentials, i.e. credentials stored on Trezor. This can be used for passwordless logins. AFAIK only Microsoft uses resident credential at the moment. If you have multiple accounts for one service, then you get to choose which account to use by swiping between them.

Hannsek:
The design with swipe as it is now is fine. 
image

core/src/trezor/ui/layouts/tt_v2/webauthn.py
```diff
@@ -1,20 +1,73 @@
...
-) -> bool:
-    raise NotImplementedError
+) -> tuple[bool, int]:
+    # `trezorui2.confirm_webauthn` is expecting a list of `app_names`
+    # and `account_names` - to allow for the `pageable` situation.
+    # In case the list has only one element, it will not be paginated.
+    # Assuming the header and icon are the same for all pages.
+
+    app_names: list[str] = [info.app_name()]
+    account_names: list[str | None] = [info.account_name()]
```
mmilata, line 27: Consider using [list of tuples](https://docs.python.org/3/library/functions.html#zip) - then you don't need to handle the case where lists are of different length and there will be slightly less asserts on the receiving side.

See also https://github.com/trezor/trezor-firmware/blob/da83d985392d45eb5e3bb15cd2ff19692babf46a/core/embed/rust/src/ui/model_tt/layout.rs#L1064

core/embed/rust/src/ui/model_tt/component/webauthn.rs
```diff
@@ -0,0 +1,235 @@
...
+        let active_page = match swipe {
+            SwipeDirection::Left if self.scrollbar.has_next_page() => self.active_page() + 1,
+            SwipeDirection::Right if self.scrollbar.has_previous_page() => self.active_page() - 1,
+            _ => self.active_page(), // page did not change
+        };
+        self.scrollbar.go_to(active_page);
+
+        // Disable swipes on the boundaries. Not allowing carousel effect.
+        self.page_swipe.allow_right = self.scrollbar.has_previous_page();
+        self.page_swipe.allow_left = self.scrollbar.has_next_page();
```
mmilata, line 112: These doesn't seem to be updated before first swipe, so on the first screen you can probably swipe right even if you shouldn't?

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -585,6 +600,62 @@ extern "C" fn new_show_error(n_args: usize, args: *const Obj, kwargs: *mut Map)
...
+
+            let account_name: Option<StrBuffer> = {
+                let mut iter_buf = IterBuf::new();
+                let mut iter = unwrap!(Iter::try_from_obj_with_buf(account_name, &mut iter_buf));
+                iter.nth(page_index).unwrap().try_into_option().unwrap()
+            };
+
+            FidoPage::new(app_name, account_name)
+        };
+        let pages = FidoPages::new(get_page, page_count, "_tmp".into());
```
mmilata, line 639: Nit: you can make make `FidoPaginatedPage::new` accept `get_page` and `page_count` and turn `FidoPages` into private structure, without _tmp it doesn't serve any purpose to have it as part of API.

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -585,6 +600,62 @@ extern "C" fn new_show_error(n_args: usize, args: *const Obj, kwargs: *mut Map)
...
+        let app_names: Obj = kwargs.get(Qstr::MP_QSTR_app_names)?;
+        let account_name: Obj = kwargs.get(Qstr::MP_QSTR_account_names)?;
+
+        // Need to find the overall page count for pagination purposes.
+        let page_count = {
+            let mut count = 0;
+            let mut iter_buf = IterBuf::new();
+            let iter = Iter::try_from_obj_with_buf(app_names, &mut iter_buf)?;
+            for _ in iter {
+                count += 1;
```
mmilata, line 616: Should be possible to use [`Iterator::count`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count) here.

core/src/trezor/ui/layouts/tt/webauthn.py
```diff
@@ -13,17 +13,24 @@ async def confirm_webauthn(
...
     if pageable is not None:
         confirm: ui.Layout = ConfirmPageable(pageable, ConfirmContent(info))
     else:
         confirm = Confirm(ConfirmContent(info))
 
+    # NOTE: to be compatible with UI2's need to sending a tuple of (is_confirmed, page_index),
+    # sending a mock page_index of "-1", so that the caller knows it should not use it.
     if ctx is None:
-        return is_confirmed(await confirm)
+        return is_confirmed(await confirm), -1
```
mmilata, line 25: Why not just get it with `pageable.page()`?

core/src/trezor/ui/layouts/tt/webauthn.py
```diff
@@ -13,17 +13,24 @@ async def confirm_webauthn(
     ctx: wire.GenericContext | None,
     info: ConfirmInfo,
     pageable: Pageable | None = None,
-) -> bool:
+) -> tuple[bool, int]:
```
mmilata, line 16: Wouldn't `int | None` be a more pythonic choice? Or just int and exception when not confirmed.

core/embed/rust/src/ui/model_tt/component/mod.rs
```diff
@@ -8,6 +8,8 @@ mod number_input;
 mod page;
 mod scroll;
 mod swipe;
+mod webauthn;
+pub mod webauthn_icons;
```
mmilata, line 12: I think this doesn't need to be pub?

mmilata (changes requested):

core/embed/rust/src/ui/model_tt/component/webauthn.rs
```diff
@@ -0,0 +1,235 @@
...
+        },
+        constant,
+    },
+};
+
+use super::CancelConfirmMsg;
+
+pub enum FidoMsg {
+    Confirmed(u8),
+    Cancelled(u8),
```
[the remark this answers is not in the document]
grdddj: Yeah, theoretically the index is unused in that case (right now), but we want to always return the same value - `tuple[bool, int]` into `micropython`. I think sending it always like this is better than checking whether we have received `tuple[True, index]` or `False`. `src/trezor/ui/layuts/webauthn/confirm_webauthn` should also always return `tuple[bool, int]` so callers have an easy job.

Agree with changing `u8` to `usize`

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -16,6 +16,7 @@ use super::ffi;
 ///
 /// Given the above assumptions about MicroPython strings, working with
 /// StrBuffers in Rust is safe.
+#[derive(Eq, PartialEq, Clone, Copy)]
```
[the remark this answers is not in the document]
grdddj: Right, I have added `Eq, PartialEq` to compare two `StrBufers`, it was not working, but I forgot to delete it. I compare them with `str1.as_ref() != str2.as_ref()` now.

The same story with  `Copy, Cloned`, I added it during development as it was requested by the compiler, but it seems to be fine even without it.

core/src/apps/webauthn/knownapps.py.mako
```diff
@@ -39,7 +43,8 @@ def by_rp_id_hash(rp_id_hash: bytes) -> FIDOApp | None:
         # ${type} key for ${app.name}
         return FIDOApp(
             label=${black_repr(label)},
-            icon=${black_repr(app.icon_res)},
+            icon=${black_repr(app.icon_res)},  # UI1, TO BE REMOVED
```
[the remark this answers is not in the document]
grdddj: It would be possible to do this trick for `UI1`, but I think it is not needed, as very soon we will get rid of `UI1` - probably this code will never even be released with `UI1`.

core/embed/rust/src/ui/model_tt/component/mod.rs
grdddj: Yeah, it should not differ from the rest without a big reason.

Question is whether to even have `webauthn_icons` module, we could theoretically define all of that in `webauthn`, it should not be used anywhere else. Do you have some preference?

grdddj: `pub`deleted in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/src/trezor/ui/layouts/tt/webauthn.py
grdddj: Good idea, done in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/ui/model_tt/layout.rs
grdddj: Used in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/ui/model_tt/layout.rs
grdddj: Nice suggestion, in the end I got completely rid of `FidoPages`, as the `page_count` is stored in `scrollbar` and therefore `FidoPaginatedPage` can just store `get_page` function - [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/ui/model_tt/component/webauthn.rs
grdddj: Good catch, I am updating that during the object creation - [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/src/trezor/ui/layouts/tt_v2/webauthn.py
grdddj: Done in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/ui/model_tt/component/webauthn.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    fn place(&mut self, bounds: Rect) -> Rect {
+        self.page_swipe.place(bounds);
+
+        // Place the control buttons.
+        let controls_area = self.controls.place(bounds);
+
+        // Get the image and content areas.
+        let content_area = bounds.inset(Insets::bottom(controls_area.height()));
+        let (image_area, content_area) = content_area.split_top(70);
```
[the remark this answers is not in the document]
grdddj: Done in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/micropython/buffer.rs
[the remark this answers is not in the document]
grdddj: Whole `#[derive` deleted in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/ui/model_tt/component/webauthn.rs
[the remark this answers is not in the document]
grdddj: Changed to `usize` in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/micropython/obj.rs
```diff
@@ -185,6 +187,19 @@ impl From<bool> for Obj {
     }
 }
 
+impl TryFrom<(bool, u8)> for Obj {
```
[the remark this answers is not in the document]
grdddj: Deleted in [b557d76](https://github.com/trezor/trezor-firmware/pull/2556/commits/b557d76468ea90567cba981b4288e96502a5b73f)

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: #2581

core/src/apps/webauthn/knownapps.py.mako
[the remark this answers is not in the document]
mmilata: OK, then we need to keep in mind not to merge this PR before the upcoming release is forked off.

core/embed/rust/src/ui/model_tt/component/mod.rs
mmilata: I'd keep it in the separate file. Sorry for the insufficient explanation, what I had in mind is this:

* delete the `pub use webauthn_icons::get_webauthn_icon_data;` line in mod.rs as we're not using the function outside of the module
* in webauthn.rs import the function from `crate::ui::model_tt::component::webauthn_icons::get_webauthn_icon_data`

core/embed/rust/src/ui/model_tt/component/webauthn.rs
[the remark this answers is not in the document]
mmilata: Then just return `-1` in in `msg_try_into_obj`? Having the index in `Cancelled` here is completely useless.

core/embed/rust/src/ui/model_tt/component/webauthn.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    fn on_page_swipe(&mut self, ctx: &mut EventCtx, swipe: SwipeDirection) {
+        // Change the page number.
+        let active_page = match swipe {
+            SwipeDirection::Left if self.scrollbar.has_next_page() => self.active_page() + 1,
+            SwipeDirection::Right if self.scrollbar.has_previous_page() => self.active_page() - 1,
+            _ => self.active_page(), // page did not change
+        };
+        self.scrollbar.go_to(active_page);
```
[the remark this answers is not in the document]
mmilata: Still not using `go_to_previous_page` and `go_to_next_page` though? I see that I mixed up "`if`" with "`==`" in the snippet above, it should have been:

```suggestion
        match swipe {
            SwipeDirection::Left => self.scrollbar.go_to_next_page(),
            SwipeDirection::Right => self.scrollbar.go_to_previous_page(),
            _ => {}, // page did not change
        };
```

core/embed/rust/src/ui/model_tt/component/mod.rs
grdddj: Done like this in [f44f512](https://github.com/trezor/trezor-firmware/pull/2556/commits/f44f5126ddf39a27ba035f698f2512abf0ea72c8)

core/src/trezor/ui/layouts/tt/webauthn.py
grdddj: Changed to this in [f44f512](https://github.com/trezor/trezor-firmware/pull/2556/commits/f44f5126ddf39a27ba035f698f2512abf0ea72c8)

core/src/apps/webauthn/knownapps.py.mako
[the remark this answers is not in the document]
grdddj: Done in [f44f512](https://github.com/trezor/trezor-firmware/pull/2556/commits/f44f5126ddf39a27ba035f698f2512abf0ea72c8) - it was just one formatted string in the end

core/embed/rust/src/ui/model_tt/component/webauthn.rs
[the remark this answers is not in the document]
grdddj: Deleted the value from `Cancelled` in [f44f512](https://github.com/trezor/trezor-firmware/pull/2556/commits/f44f5126ddf39a27ba035f698f2512abf0ea72c8)

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -666,6 +680,52 @@ extern "C" fn new_show_error(n_args: usize, args: *const Obj, kwargs: *mut Map)
...
+            Button::with_text("CONFIRM").styled(theme::button_confirm()),
+            2,
+        );
+
+        let fido_page = FidoPaginatedPage::new(get_page, page_count, icon, controls);
+
+        let obj = LayoutObj::new(
+            Frame::new(title, fido_page)
+                .with_border(theme::borders())
+                .into_child(),
```
mmilata, line 722: Turns out I lied and LayoutObj::new wraps the component in Child<> automatically, sorry. See also #2581 #2600.

core/embed/rust/src/ui/model_tt/component/webauthn.rs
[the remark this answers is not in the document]
mmilata: Still seems like needless duplication of the logic but at least you're not computing the page number here directly anymore so whatever I guess.

core/embed/rust/src/ui/model_tt/component/webauthn_icons.rs
```diff
@@ -0,0 +1,70 @@
...
+const ICON_KEEPER: &[u8] = include_res!("model_tt/res/webauthn/icon_keeper.toif");
+const ICON_LOGIN_GOV: &[u8] = include_res!("model_tt/res/webauthn/icon_login.gov.toif");
+const ICON_MICROSOFT: &[u8] = include_res!("model_tt/res/webauthn/icon_microsoft.toif");
+const ICON_MOJEID: &[u8] = include_res!("model_tt/res/webauthn/icon_mojeid.toif");
+const ICON_NAMECHEAP: &[u8] = include_res!("model_tt/res/webauthn/icon_namecheap.toif");
+const ICON_PROTON: &[u8] = include_res!("model_tt/res/webauthn/icon_proton.toif");
+const ICON_SLUSHPOOL: &[u8] = include_res!("model_tt/res/webauthn/icon_slushpool.toif");
+const ICON_STRIPE: &[u8] = include_res!("model_tt/res/webauthn/icon_stripe.toif");
+const ICON_TUTANOTA: &[u8] = include_res!("model_tt/res/webauthn/icon_tutanota.toif");
+const ICON_WEBAUTHN: &[u8] = include_res!("model_tt/res/webauthn/icon_webauthn.toif");
```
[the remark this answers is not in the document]
grdddj: Good point, created `webauthn_icons.rs.mako` in [bfb822a](https://github.com/trezor/trezor-firmware/pull/2556/commits/bfb822a5922021ce56d053948525c28f809f677f). It already caught one bug, I somehow did not include the icon for `Kraken`

core/src/trezor/ui/layouts/tt_v2/webauthn.py
```diff
@@ -1,20 +1,66 @@
...
+    from ...components.tt.confirm import Pageable
 
 
 async def confirm_webauthn(
     ctx: wire.GenericContext | None,
     info: ConfirmInfo,
     pageable: Pageable | None = None,
-) -> bool:
-    raise NotImplementedError
+) -> int | None:
```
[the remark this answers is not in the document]
grdddj: Yes, right, it is slightly confusing that we cannot use `if not confirm_webauthn(...)`. `confirm_webauthn_choose` added in  [bfb822a](https://github.com/trezor/trezor-firmware/pull/2556/commits/bfb822a5922021ce56d053948525c28f809f677f)

mmilata (approved):

matejcik:
rebased on master and did quite a lot of changes:

* renamed the layout side from "webauthn" to "fido" according to #703 -- didn't touch app side because that might cause a larger change
* dropped `Paginate` and the rest of UI1 remnats, fixing #2639 along the way
* dropped FIDO's `ConfirmInfo` and refactored Python-side usage of `confirm_fido` to accept `app_name`, `icon` and an iterable of `account`s instead of messing with pagination simulation
* fixed an issue related to #2653
* added debuglink handling
* fixed and used the new list operations for random-access into the accounts list instead of repeatedly iterating
* modified `Label` to add ellipsis if it doesn't fit within its bounds
* modified the fido screen to use the new `Label` functionality instead of manually rendering text
* bumped the fido2 tests submodule with @mmilata's newest commit

lot of the new commits are called "fixup", all those should be squashed into the original "implement webauthn layouts" commit

core/src/trezor/ui/layouts/tt_v2/fido.py
```diff
@@ -0,0 +1,101 @@
...
+async def confirm_fido(
+    ctx: GenericContext | None,
+    header: str,
+    app_name: str,
+    icon_name: str | None,
+    accounts: list[str | None],
+) -> int:
+    """Webauthn confirmation for one or more credentials."""
+    # `trezorui2.confirm_webauthn` is expecting a list of tuples
+    # (app_name, account_name) to allow for the `pageable` situation.
```
mmilata, line 63: This comment is outdated.

core/src/apps/webauthn/add_resident_credential.py
```diff
@@ -49,8 +30,13 @@ async def add_resident_credential(
...
 
-    if not await confirm_webauthn(ctx, ConfirmAddCredential(cred)):
-        raise wire.ActionCancelled
+    await confirm_fido(
+        ctx,
+        "Import credential",
+        cred.app_name(),
+        cred.icon_name(),
+        [cred.account_name()],
+    )
```
mmilata, line 39: Needs fixture update: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3369467991

mmilata (commented):
Looks great, just 2 small things missing. Big up for getting rid of multiple inheritance in fido2.py.

grdddj (commented):
Very nice simplifications on the `upy` side!

Couple of small comments and questions

core/embed/rust/src/micropython/list.rs
```diff
@@ -92,6 +100,22 @@ impl List {
             core::slice::from_raw_parts_mut(items_ptr, len)
         }
     }
+
+    pub fn get(&self, index: usize) -> Result<Obj, Error> {
```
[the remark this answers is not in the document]
matejcik: the unsafe ones are unsafe :) `as_slice` is unsafe because the slice does not necessarily uphold the invariants that Rust promises during its whole lifetime.

`get` is safe because it is using `as_slice` exactly as safely as can be done, and the result _does_ uphold the relevant invariants.

core/src/trezor/ui/layouts/tt_v2/fido.py
```diff
@@ -0,0 +1,101 @@
...
+    else:
+        result = await interact(ctx, confirm, "confirm_fido", ButtonRequestType.Other)
+
+    # Returning None to indicate cancellation, otherwise the index of the
+    # chosen page.
+
+    if isinstance(result, int):
+        return result
+
+    from trezor.wire import ActionCancelled
```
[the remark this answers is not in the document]
matejcik: that was my thinking, yes. i can add a comment to make that clearer

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -661,6 +677,37 @@ extern "C" fn new_show_error(n_args: usize, args: *const Obj, kwargs: *mut Map)
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_confirm_fido(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let title: StrBuffer = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
+        let app_name: StrBuffer = kwargs.get(Qstr::MP_QSTR_app_name)?.try_into()?;
+        let icon: Option<StrBuffer> = kwargs.get(Qstr::MP_QSTR_icon_name)?.try_into_option()?;
+        let accounts: Gc<List> = kwargs.get(Qstr::MP_QSTR_accounts)?.try_into()?;
```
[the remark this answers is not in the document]
matejcik: I went through them and they all seemed more correct at the moment. The pattern is always "iterate once through a passed in iterable" -- unlike here, where we iterate once just to get the length, and then repeatedly to get the n-th element, which is much nicer to do with random access ability.

If we dive a little deeper, we might be able to update the other usages for random access by passing in the `Gc<List>` deeper into the layout. But just replacing the `Iter` loop by itself does not bring any value.

`try_from_obj_with_buf` could probably be made a bit better. I forget why the `IterBuf` is separate, my guess is something something lifetimes, but if that's the case, writing a macro might be useful.

core/embed/rust/src/ui/model_tt/component/fido.rs
```diff
@@ -0,0 +1,210 @@
...
+    get_account: F,
+    scrollbar: ScrollBar,
+    fade: bool,
+    controls: U,
+}
+
+impl<F, T, U> FidoConfirm<F, T, U>
+where
+    F: Fn(usize) -> T,
+    T: Deref<Target = str> + From<&'static str>,
```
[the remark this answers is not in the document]
matejcik: It's just because `Layout` is built with `Deref`, so either i would need to change `Layout` to `AsRef` (which propagates to a lot of places), or, easier, switch Fido to `Deref`.

More generally, it seems that `Deref` is the more appropriate of the two for us. IIUC, "Deref" means "this thing can be treated as a `&str`" whereas "AsRef" means "this thing is capable of giving you a `&str` if you ask for it". My ultimate goal is to replace `T: Deref` with a global type alias `StrType` which is going to be `StrBuffer` if we build with micropython, or `&str` if not, and for that case the `Deref`-based code seems to require less changes.

core/src/trezor/ui/layouts/tt_v2/fido.py
```diff
@@ -0,0 +1,101 @@
...
+
+    raise ActionCancelled
+
+
+async def confirm_fido_reset() -> bool:
+    confirm = _RustLayout(
+        trezorui2.confirm_action(
+            title="FIDO2 RESET",
```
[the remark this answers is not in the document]
matejcik: fwiw we probably don't need `confirm_fido_reset` here, this `confirm_action` can just as easily live in the caller

core/embed/rust/src/micropython/list.rs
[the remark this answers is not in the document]
matejcik: N.b.: marking function `unsafe` is completely independent of whether it contains any `unsafe` blocks inside, and in some weird cases a function can be `unsafe` despite having no `unsafe` blocks.

Marking a function `unsafe` is information for the caller: be careful and explicit (i.e., use an `unsafe` block) when calling this, because you _could_ be doing it wrong. Safe functions are, or should be, safe to call without thinking too much about it.

core/src/trezor/ui/layouts/tt_v2/fido.py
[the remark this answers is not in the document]
matejcik: ah, but the Rust-exposed functions are kwargs only so the kwargs are actually mandatory

core/src/apps/webauthn/add_resident_credential.py
matejcik: [f64ff81](https://github.com/trezor/trezor-firmware/pull/2556/commits/f64ff8116acb0107430297ea3ab072aa77714e7a)

core/src/trezor/ui/layouts/tt_v2/fido.py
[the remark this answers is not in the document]
matejcik: [e5359f4](https://github.com/trezor/trezor-firmware/pull/2556/commits/e5359f41f25496e1e6b1876a07eb4b39c2ea23f1)

core/src/trezor/ui/layouts/tt_v2/fido.py
```diff
@@ -0,0 +1,101 @@
...
+            accounts=accounts,
+        )
+    )
+
+    if ctx is None:
+        result = await confirm
+    else:
+        result = await interact(ctx, confirm, "confirm_fido", ButtonRequestType.Other)
+
+    # Returning None to indicate cancellation, otherwise the index of the
```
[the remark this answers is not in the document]
matejcik: [e5359f4](https://github.com/trezor/trezor-firmware/pull/2556/commits/e5359f41f25496e1e6b1876a07eb4b39c2ea23f1)

core/src/trezor/ui/layouts/tt_v2/fido.py
matejcik: [e5359f4](https://github.com/trezor/trezor-firmware/pull/2556/commits/e5359f41f25496e1e6b1876a07eb4b39c2ea23f1)

core/SConscript.unix
```diff
@@ -532,13 +532,13 @@ if FROZEN:
...
+        SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/fido.py'))
     if TREZOR_MODEL in ('T',):
         SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/tt_v2/__init__.py'))
         SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/tt_v2/reset.py'))
         SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/tt_v2/recovery.py'))
         if EVERYTHING:
-            SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/tt_v2/webauthn.py'))
+            SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/tt_v2/fido.py'))
     elif TREZOR_MODEL in ('1',):
         SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/ui/layouts/t1.py'))
```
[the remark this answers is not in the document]
matejcik: [af90635](https://github.com/trezor/trezor-firmware/pull/2556/commits/af90635feaaba8bb6e156d6f28e95ebd49b32a47)

mmilata (approved):

PR #2561, opened 2022-10-14
feat(cardano): add support for CIP-36 governance registration format

davidmisiak:
This PR adds support for [CIP-36](https://cips.cardano.org/cips/cip36/) governance registration format. Changes:
- New field: `delegations` array (each delegation consists of a voting key and its voting power proportion) as an alternative to delegating the entire voting power to a single voting key. The array is embedded in the protobuf message due to simplicity (max. 32 delegations allowed).
- New field: `voting_purpose` integer (currently 0 = Catalyst, 1 = other). In CIP-36 registrations, this field is always serialized, (value 0 is used if not provided by the client).
- All `Catalyst` occurences are replaced by `governance`, since the format should be usable for other governance purposes, not only Catalyst.
- The `ed25519_pk` bech32 prefix is [replaced](https://github.com/cardano-foundation/CIPs/pull/342) by `gov_vk` for the purpose of displaying voting public keys.

The new 1694' derivation path and the vote cast call implementation is not included in this PR.

---

EDIT - Added:
- Support for [new testnets](https://book.world.dev.cardano.org/environments.html) in the firmware and `cardanoctl`.


common/tests/fixtures/cardano/sign_tx.json
```diff
@@ -1060,7 +1060,150 @@
...
+        "auxiliary_data": {
+          "governance_registration_parameters": {
+            "staking_path": "m/1852'/1815'/0'/2/0",
+            "reward_address_parameters": {
+              "addressType": 0,
+              "path": "m/1852'/1815'/0'/0/0",
+              "stakingPath": "m/1852'/1815'/0'/2/0"
+            },
+            "nonce": 22634813,
+            "format": 1,
```
matejcik, line 1086: minor nit: you should be using `"CIP36"` string

matejcik (approved):
it seems the new tests are not included in fixtures.json? please regenerate that

common/tests/fixtures/cardano/sign_tx.json
davidmisiak: How so? The parser expects an int enum value, just like with output `format`. Or did you mean changing the parser to expect a string? (In that case, should I also change the output `format` parser?)

matejcik: my bad, I forgot that this is a custom parser and not the built-in json-to-protobuf. no reason to muck about with it then

davidmisiak:
> it seems the new tests are not included in fixtures.json? please regenerate that

Done, UI tests should be up to date. I also added changelog entries and rebased.

davidmisiak:
Also @matejcik, [here](https://github.com/vacuumlabs/trezor-firmware/compare/cardano-governance-cip36...vacuumlabs:trezor-firmware:cardano-add-new-testnets) is a commit that adds [new testnets](https://book.world.dev.cardano.org/environments.html). Should I add it here or open a separate PR?

matejcik:
feel free to add here. it has no effect on tests, right?

davidmisiak:
> it has no effect on tests, right?

I believe so.

Thanks, added.

matejcik:
that is the plan, yes

davidmisiak:
Awesome, thanks!

PR #2565, opened 2022-10-16
Change display type for model R

TychoVrahe:
Changes display type for model R.

Bootloader is modified to support displays that need pixeldata_dirty and refresh. Relevant for #2506 

Includes some refactoring/restruct, notably:
- moves display drivers to trezorhal
- changes the filenames to match display types, drop stm32 from the filename because more or less everything in trezorhal is stm32 specific so far
- the rename also enables to use same display without naming getting weird, like trezor R using display-stm32_1.c file.
- introduces 'board' concept, where the pinout is specified in e.g. `trezorhal/boards/trezor_t.h` so that driver implementations do not need to have messy ifdefs for every supported model type
- this also allows easy support for older board versions, like devkits for model R, which are perfectly good for many development use cases.


TychoVrahe:
in https://github.com/trezor/trezor-firmware/pull/2565/commits/032f3e601bf052f6f46576e1e153374e70a7e564 TR board v.4 is set as default and also introduced `display_reinit` function to lower the need for conditional compilations in `main`

hiviah (approved):
Finally got around to solder and flash the model with new display. Most of the changes are IIRC from previous PR, so ACK for this new display.

core/embed/extmod/modtrezorui/buffers.h
```diff
@@ -23,7 +23,7 @@
 #include <stdbool.h>
 
 #include "common.h"
-#include "display_defs.h"
+#include TREZOR_BOARD
```
matejcik, line 26: could we do `#include #TREZOR_BOARD` to avoid the awkward quoting from the def side?

matejcik: (otoh `MP_CONFIGFILE` also has quotes and it's easier to use it that way so i guess we'll keep it)

PR #2567, opened 2022-10-16
Optimize pin keyboard rendering for TT UI2

TychoVrahe:
Using pads + child components to get rid of text flickering when button is pressed when pin is empty.


mmilata (approved):
LGTM, thanks! I must admit I can barely spot the difference, wondering if years of using shitty Linux GPU drivers broke my brain ...

CI failure on click test is unrelated so this is ready to merge.

Issue #2513, opened 2022-09-14
T1 bootloader: verify firmware signatures based on SignMessage


matejcik:
The T1 bootloader verifies the firmware signature via a manual ECDSA 3-of-5 scheme:

1. let the message `m` be the firmware header with signature slots zeroed out.
   * (implementation-wise, take a SHA256 digest of the above data)
2. for each of the 3 signature slots:
    1. verify that the pubkey index points to one of the 5 baked in public keys
    2. verify that the pubkey index is different from the other two slots, i.e., all 3 slots are signed by different public keys
    3. verify that this is a valid signature of `m` for the selected pubkey
        * (implementation-wise, we call `ecdsa_verify_digest` instead of `ecdsa_verify`, because we already calculated the digest)

This entails signing arbitrary data with ECDSA, which cannot be done on Trezor directly.

We want to change the signing method so that it is **possible to verify the digest being signed on Trezor screen**.
The signer will calculate SHA256 digest of `m` (fw header minus signature slots) on host and send it to Trezor as data for Bitcoin `SignMessage`.
Trezor will display the digest on screen, and when confirmed, produce a signature of the following data:
`\x18Bitcoin Signed Message:\n\x20<32 bytes of digest>`.

On the bootloader side, the following is the updated algorithm:
1. let `m'` be the firmware header with signature slots zeroed out
2. let `m` be `SHA256("\x18Bitcoin Signed Message:\n\x20" + SHA256(m'))` (because Bitcoin does double-SHA everywhere it can, taking a signature of a SHA256 digest results in signature over a SHA(SHA(actual message)) )
3. continue with (2) from the original algorithm, i.e., the signatures are now verifying the changed `m`

hiviah:
This leaves a question which hash should now be displayed both at startup or update.

After reviewing possible paths of update/start/wipe combinations it makes probably most sense to show the `m'` hash - which is the same as shown with current scheme (confusingly called "new" in code).

Some points:

1. It will be compatible with the current hash computation, so there will be no need for Suite to make extra special cases
2. Hash shown of the same FW will be thus same before and after update (including various paths like wipe/update cycles)

matejcik:
you are absolutely correct, also showing hash of `m'` means showing the same thing the signer is confirming on Trezor

hiviah:
One note: because signing with Bitcoin SignMessage does double SHA-256, the value signed output from SignMessage and checked by FW is:

[...]

PR #2568, opened 2022-10-17
T1 bootloader: verify firmware signatures based on SignMessage, add signature debugging

hiviah:
* Removed oldest v1 style of firmware signature and presence checks. 
* Added debug helpers for T1 signatures.
* Support for v2 and v3 signatures, but can only update FW to v3-style signed. 
* Support for debugging T1 signatures.
* Scripts and README for debugging v2/v3 FW signing scheme. 
* Firmware in GetFeatures counts only v3 signatures as signed.

For clarity:

* v1 signatures - the "old" with `TRZR` header magic
* v2 signatures - the "new" (second) scheme and layout with `TRZF` magic
* v3 signatures - the scheme using SignMessage/VerifyMessage in #2513 .

Implements #2513 .

In `legacy/debug_signing/README.md` is HOWTO around testing signature schemes


hiviah:
NOTE: production public keys need to be filled in yet for v3 scheme.

(disregard the core click tests failing in gitlab, it's known)

TychoVrahe (commented):
Nice work. I only have a couple of super trivial comments.

I tested this on devkit with the QA bootloader/fw builds and seems to be working well, one issue is that for bootloaders 1.8.0 and 1.10.0 suite will refuse to install intermediate fw via custom firmware install and asks to upgrade bootloader to 1.8.0, which is more of an issue of how suite identifies the bootloader version (it estimates the bootloader version from firmware version, but these bootloaders do not have the firmware version in features, so it fails, see https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/utils/firmware/index.ts ) - thinking that we could sign the intermediate fw with v2 header also with v1 signature to make this work, or maybe it should be improved in suite if possible, or both. cc @matejcik

legacy/fw_signatures.h
```diff
@@ -59,11 +58,80 @@ typedef struct {
...
+ * @param store_fingerprint if non-NULL, store hash computed with chosen method
+ * here
+ * @param use_verifymessage false - use v2 signature scheme, true - use v3
+ * SignMessage/VerifyMessage scheme
+ * @return SIG_OK or SIG_FAIL
+ */
+int signatures_ok(const image_header *hdr, uint8_t store_fingerprint[32],
+                  bool use_verifymessage);
+/**
+ * Check is either v2 or v3 signature of header is valid.
```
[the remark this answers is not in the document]
hiviah: Fixed.

legacy/debug_signing/README.md
```diff
@@ -0,0 +1,77 @@
+# Debugging T1 signatures
```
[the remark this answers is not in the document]
hiviah: This is supposed to be a "separate chapter" about signatures debugging with documentation (`scripts` is more regarding build). Originally this contained also examples of unsigned and signed trezor.bin (both styles) and other outputs, which were made by hand in hex editor.
But in the end I decided not to add the trezor.bin files to not inflate the size of the repo, since now they can be created rather effortlessly as compared to before.

legacy/fw_signatures.h
```diff
@@ -59,11 +58,80 @@ typedef struct {
...
+ * @param hdr header to check
+ * @param store_fingerprint if non-NULL, store hash computed with chosen method
+ * here
+ * @param use_verifymessage false - use v2 signature scheme, true - use v3
+ * SignMessage/VerifyMessage scheme
+ * @return SIG_OK or SIG_FAIL
+ */
+int signatures_ok(const image_header *hdr, uint8_t store_fingerprint[32],
+                  bool use_verifymessage);
+/**
```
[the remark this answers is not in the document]
hiviah: Fixed.

matejcik:
yeah, this should not have been merged before we add the production keys :(

matejcik:
no matter, now we just need to be extra careful not to release this bootloader.

hiviah:
Sorry, totally forgot that the production keys were not there yet.

legacy/firmware/fsm_msg_common.h
```diff
@@ -24,7 +24,8 @@ bool get_features(Features *resp) {
 #else
   const image_header *hdr =
       (const image_header *)FLASH_PTR(FLASH_FWHEADER_START);
-  if (SIG_OK == signatures_new_ok(hdr, NULL)) {
+  // only v3 signatures count as signed
+  if (SIG_OK == signatures_ok(hdr, NULL, true)) {
```
matejcik, line 28: does this mean that a production firmware will return "UNSAFE, DO NOT USE" if the installed image has a valid v2 header? that does not seem correct

matejcik: i see we already discussed this in https://github.com/trezor/trezor-firmware/issues/2513#issuecomment-1279166574

but ISTM this should be `signatures_match` anyway

TychoVrahe: since this is in FW, at this point it should only have v3 header no?

matejcik: Not if this is a direct upgrade (no intermediate). But that is a good point. Maybe we want to always upgrade via intermediate?

TychoVrahe: right, direct upgrade could cause that.

i think it would be a bit preferable to upgrade only via intermediate, having different headers depending on how you upgraded doesn't seem very clean

hiviah: I think the logic was that the upgrade will be only via intermediate (which does not have any USB communication), so new firmware should always be signed only by v3 signature. But the change is trivial, so it may be changed to show both v2 and v3 as valid.

PR #2572, opened 2022-10-19
fix(core/debug): separate buffer for debuglink to prevent BufferLock deadlocks

matejcik:
taken out from #2335 

very likely fixes the current click-test problems.


mmilata (approved):
LGTM, thanks for looking into this.

PR #2575, opened 2022-10-20
Rust UI: on-the-fly conversion of bytes to hexadecimal representation

### LINKED ISSUE #2531: `confirm_blob` cannot directly render bytearray data, we currently hexlify it in python but encoding it on-the-fly would save some RAM

mmilata:
Builds on top of #2519. Closes #2531. Also monospace text is now broken without hyphens and regardless of word boundaries. [[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3203188320/artifacts/master_diff/index.html)]

mono1 mono2


core/embed/rust/src/ui/layout/util.rs
```diff
@@ -32,3 +35,98 @@ where
...
+    const HEX_LOWER: [u8; 16] = *b"0123456789abcdef";
+    let par = &objslice[index];
+
+    // Handle None as empty string.
+    if par.content() == Obj::const_none() {
+        return par.with_content("");
+    }
+
+    // Handle str.
+    if StrBuffer::try_from(par.content()).is_ok() {
```
matejcik, line 54: why not `if let strbuf = StrBuffer()...` and then `strbuf.as_ref()` instead of doing the get_buffer unsafe dance here?

mmilata:
```
error[E0515]: cannot return value referencing local variable `strbuf`
  --> src/ui/layout/util.rs:60:16
   |
59 |         let s = strbuf.as_ref();
   |                 --------------- `strbuf` is borrowed here
60 |         return par.with_content(&s[offset..]);
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function

```

mmilata: OK not really
```
error[E0106]: missing lifetime specifier
  --> src/micropython/buffer.rs:58:26
   |
58 |     fn into_ref(self) -> &str {
   |                          ^ expected named lifetime parameter
   |
   = help: this function's return type contains a borrowed value, but there is no value for it to be borrowed from
```

mmilata: But this works, not sure if good idea though.
```rust
    pub fn into_ref<'a>(self) -> &'a str {
        if self.ptr.is_null() {
            return ""
        }
        let slice = unsafe { slice::from_raw_parts(self.ptr, self.len) };
        unsafe { str::from_utf8_unchecked(slice) }
    }

```

matejcik: an alternative is to [teach `StrBuffer` about slicing](https://stackoverflow.com/a/57203324) while keeping the original str pointer, and make the inner Paragraphs based on `StrBuffer`s

mmilata: What if we wrapped this in a function that ties the lifetime of the result to the source `Obj` (needs `Paragraph::content` to return a reference)? Is it useless because `Obj` implements `Copy`?
```rust
pub fn obj_as_str(obj: &Obj) -> Result<&str, Error> {
    let strbuf = StrBuffer::try_from(*obj)?;
    Ok(strbuf.into_ref())
}
```

matejcik: > Is it useless because `Obj` implements `Copy`?

I think the general problem is not the `Copy`, but that `Obj` is a pointer and not an owner.
*But* what the code above seems to say is, as long as I hold _the pointer_, i.e., a piece of memory with an address of the uPy object, then the `&str` slice is valid.
Which seems to be exactly true?

One potential risk is that the obj in question is once removed from the actual string data, i.e., in theory, someone could mutate contents of the obj to point to another string. I don't think uPy allows that though.

Still, it feels like cheating, so I want @jpochyla to review it first.

matejcik: > in theory, someone could mutate contents of the obj to point to another string

even more specifically, since `Obj` is `Copy`, someone could try to grab another copy mutably from Rust. I am like 80 % sure that can't be done to a string either -- in general, the fact that this `Obj` is a string type is giving us a lot to work with.

mmilata:
Rebased to master [UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3218748446/artifacts/master_diff/index.html).

core/embed/rust/src/ui/layout/util.rs
jpochyla: Yeah the last suggestion seems like a best bet. The `into_ref` thing cannot really be a safe fn under the general safety guidelines (it's unsound -- possible to use unsafely from a safe code.) I don't really have time to make an authoritative opinion, though.

mmilata: Proposed solution in [0cc990d](https://github.com/trezor/trezor-firmware/pull/2575/commits/0cc990dd67e1b8b444a7e0a9a7d5992dfea7382d).

core/embed/rust/src/micropython/obj.rs
```diff
@@ -415,4 +415,8 @@ impl Obj {
     pub fn is_type_bytes(self) -> bool {
         unsafe { ffi::mp_type_bytes.is_type_of(self) }
     }
+
+    pub fn is_type_str(self) -> bool {
```
matejcik, line 419: let's change this to `is_str` and include the `is_qstr` check

mmilata: fixed a0c5483d6e986ee01bf895c0e4abf8ab4710a941

core/embed/rust/src/ui/layout/util.rs
```diff
@@ -32,3 +35,91 @@ where
...
+    }
+
+    fn size(&self) -> usize {
+        self.len()
+    }
+}
+
+impl<const N: usize> VecExt<Obj> for Vec<Paragraph<Obj>, N> {
+    fn add(&mut self, paragraph: Paragraph<Obj>) -> &mut Self {
+        if *paragraph.content() == Obj::const_none() {
```
matejcik, line 116: when can there be a `None` ?

mmilata: it's currently not used

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -384,19 +376,18 @@ extern "C" fn new_confirm_properties(n_args: usize, args: *const Obj, kwargs: *m
         let mut iter_buf = IterBuf::new();
         let iter = Iter::try_from_obj_with_buf(items, &mut iter_buf)?;
         for para in iter {
-            let [key, value, is_mono]: [Obj; 3] = iter_into_objs(para)?;
-            let key = key.try_into_option::<StrBuffer>()?;
-            let value = value.try_into_option::<StrBuffer>()?;
+            let [key, value]: [Obj; 2] = iter_into_objs(para)?;
```
matejcik, line 379: let's keep `is_mono`, it will come in handy when we want to indicate non-bytes properties that should be monospaced

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -384,19 +376,18 @@ extern "C" fn new_confirm_properties(n_args: usize, args: *const Obj, kwargs: *m
...
                     paragraphs.add(Paragraph::new(&theme::TEXT_BOLD, key).no_break());
                 } else {
                     paragraphs.add(Paragraph::new(&theme::TEXT_BOLD, key));
                 }
             }
-            if let Some(value) = value {
-                if is_mono.try_into()? {
+
+            if value != Obj::const_none() {
+                if value.is_type_bytes() {
```
matejcik, line 390:
```
if value.is_type_bytes() && !is_mono { return Error::ValueError }
if is_mono {
    ...
}
```

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -312,24 +310,19 @@ extern "C" fn new_confirm_action(n_args: usize, args: *const Obj, kwargs: *mut M
 
 fn _confirm_blob(
     title: StrBuffer,
-    data: Option<StrBuffer>,
-    description: Option<StrBuffer>,
-    extra: Option<StrBuffer>,
+    data: Obj,
+    description: Obj,
+    extra: Obj,
```
matejcik, line 315: The disadvantage of this approach is that it's _too_ flexible.
We want `data` to be `str | bytes`, but `description: str` and `extra: str`.

matejcik:
ACK on the scratch buffer and related changes, and also on the `ParagraphSource` trait.

what I'm very unconvinced of is `Paragraph<Obj>`.

i'm thinking of a more task-specific solution. rough draft:
```rust
enum StrOrBytes {
     Str(StrBuffer),
     Bytes(Obj),
}

impl StrOrBytes {
    pub fn as_para(&self, prototype: Paragraph, scratch: &mut [u8]) -> Paragraph<&str> {
         match self {
             Str(x) => x.as_ref(),
             Bytes(x) => self.hexlify(x, scratch),
        }
    }
}

impl TryInto<StrOrBytes> for Obj { ... }

struct ConfirmBlob {
    description: StrBuffer,
    extra: StrBuffer,
    data: StrOrBytes,
}

// === important part comes here

impl ParagraphSource for ConfirmBlob {
     fn at(...) {
         match index {
              0 => Paragraph::with_content(self.description, ...),
              1 => ...,
              2 => self.data.as_para(Paragraph(...), scratch),
         }
     }
}

// ==== even importanter part comes here

struct PropsList(Gc<List>);

impl ParagraphSource for PropsList {
      fn at(...) {
          let entry = self.0.get(index / 2);
          let [key, value] = ...;
          if index % 1 {
              convert_value_to_para(value)
          } else {
              convert_key_to_para(key)
          }
      }
}
```

mmilata:
Implemented suggested changes in [de9937d](https://github.com/trezor/trezor-firmware/pull/2575/commits/de9937d319bc2d35f957caaa04fbe789d6b2c307). Rebased to master to get `List::as_slice`.

There's a bit of a difficulty with e.g. `convert_key_to_para` because the `at()` methods returns a str reference that is twice removed from the `Gc<List>`, making the issue with `obj_as_str` even more obvious. I've tried to modify the helper to at least tie the lifetime of the &str with the list but it's probably not the best solution. Unfortunately I can only think of other bad solutions like always copying the string into the scratch buffer ...

core/embed/rust/src/ui/layout/util.rs
mmilata: no longer exists in de9937d

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: de9937d

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: de9937d

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: de9937d

mmilata:
Rebase broke some device tests, looking into that.

core/embed/rust/src/ui/layout/util.rs
```diff
@@ -32,3 +40,177 @@ where
...
+    }
+}
+
+impl TryFrom<Obj> for StrOrBytes {
+    type Error = Error;
+
+    fn try_from(obj: Obj) -> Result<Self, Error> {
+        if obj.is_str() {
+            Ok(StrOrBytes::Str(obj.try_into()?))
+        } else if obj.is_type_bytes() {
```
matejcik, line 101: we should probably add `obj.is_bytes()` that will check if `obj` is one of `bytes`, `bytearray` or `memoryview`

core/embed/rust/src/ui/layout/util.rs
```diff
@@ -32,3 +40,177 @@ where
...
+
+            if obj == Obj::const_none() {
+                return Ok(Paragraph::new(style, ""));
+            }
+
+            let para = if obj.is_str() {
+                // SAFETY:
+                // As long as self is visible to GC, the string will be also. The paragraph
+                // rendering does not keep the returned references for long so we can reasonably
+                // expect for the chain of references from self to the buffer not to be broken.
```
matejcik, line 191: Ooooh this one is fun because the "reasonable expectation" hinges on uPy not dropping the item out of our list, correct?

that's not great.
Let's keep it as is for this PR but we'll need to implement the fully correct solution (that is StrBuffer slicing) before release

matejcik (approved):
LGTM with some reservations in comments. But let's move forward with this and solve the &str lifetime issue by improving StrBuffer in a separate PR

core/embed/rust/src/ui/layout/util.rs
mmilata: ccfd8ee9f

core/embed/rust/src/ui/layout/util.rs
mmilata: #2665

mmilata:
[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3396174906/artifacts/master_diff/index.html)

PR #2180, opened 2022-03-16
Update CoinJoin transaction checks

andrewkozlik:
Closes https://github.com/trezor/trezor-firmware/issues/1160.

This PR implements the CoinJoin transaction checks that were agreed on in https://www.notion.so/satoshilabs/CoinJoin-checks-in-Trezor-35dfde067235476f945f7cec4173ff9b (see the items highlighted in green).

PR #2577, opened 2022-10-20
Implement CoinJoin requests

andrewkozlik:
We decided to strengthen the security guarantees that will be provided by the signed affiliate data in CoinJoin. Due to these changes it is becoming a little more complicated to reuse the solution that is based on payment requests, although technically it could still be done if we wanted to. Payment requests are meant for individual outputs or sets of outputs in a transaction, whereas the affiliate data is now working with the entire set of inputs and outputs. I'd call this a "transaction request" and in this particular case a `CoinJoinRequest`. Since there is only one such transaction request per transaction it makes sense to put it into `SignTx`, which is what I did here.

There is an optional commit d54758415ea8fabf995cd8c21db255f33b4797ac that I could add, which adds a fee tolerance of 200 sats on Mainnet. The rationale behind it is to prevent Trezor from declining to sign the transaction in case of some sort of rounding error in the fee computation. That would ruin the CoinJoin for all participants. But I think this shouldn't occur. Opinions?

@onvej-sl please review and feel free to comment on the CoinJoin request digest algorithm and the mask computation. It's implemented in three places actually (core, unit tests and device tests). `make_coinjoin_request()` in `tests/device_tests/bitcoin/payment_req.py` is pretty well legible.

@matejcik you might want to glance at the protobuf changes in 1de8f103cadfab5822593d810b845f9cc89bfc58, but I think it's in line with what we discussed IRL.

### DIFF
--- a/common/protob/messages-bitcoin.proto
+++ b/common/protob/messages-bitcoin.proto
@@ -198,6 +198,18 @@ message SignTx {
     optional AmountUnit amount_unit = 11 [default=BITCOIN];    // show amounts in
     optional bool decred_staking_ticket = 12 [default=false];  // only for Decred, this is signing a ticket purchase
     optional bool serialize = 13 [default=true];               // serialize the full transaction, as opposed to only outputting the signatures
+    optional CoinJoinRequest coinjoin_request = 14;            // only for preauthorized CoinJoins
+
+    /**
+     * Signing request for a CoinJoin transaction.
+     */
+    message CoinJoinRequest {
+        required uint32 fee_rate = 1;                // coordination fee rate in units of 10^-8 percent
+        required uint64 no_fee_threshold = 2;        // PlebsDontPayThreshold in Wasabi, the input amount above which the fee rate applies
+        required uint64 min_registrable_amount = 3;  // minimum registrable output amount
+        required bytes mask_public_key = 4;          // ephemeral secp256k1 public key used for masking coinjoin_flags, 33 bytes in compressed form
+        required bytes signature = 5;                // the trusted party's signature of the CoinJoin request digest
+    }
 }
 
 /**
@@ -306,6 +318,7 @@ message TxAck {
             optional uint32 orig_index = 17;                                    // index of the input in the original transaction (used when creating a replacement transaction)
             optional DecredStakingSpendType decred_staking_spend = 18;          // if not None this holds the type of stake spend: revocation or stake generation
             optional bytes script_pubkey = 19;                                  // scriptPubKey of the previous output spent by this input, only set of EXTERNAL inputs
+            optional uint32 coinjoin_flags = 20 [default=0];                    // bit field of CoinJoin-specific flags
         }
         /**
         * Structure representing compiled transaction output
@@ -359,6 +372,7 @@ message TxInput {
     optional uint32 orig_index = 17;                                    // index of the input in the original transaction (used when creating a replacement transaction)
     optional DecredStakingSpendType decred_staking_spend = 18; 	        // if not None this holds the type of stake spend: revocation or stake generation
     optional bytes script_pubkey = 19;                                  // scriptPubKey of the previous output spent by this input, only set of EXTERNAL inputs
+    optional uint32 coinjoin_flags = 20 [default=0];                    // bit field of CoinJoin-specific flags
 }
 
 /** Data type for transaction output to be signed.
@@ -427,7 +441,7 @@ message TxAckPaymentRequest {
     option (unstable) = true;
 
     optional bytes nonce = 1;              // the nonce used in the signature computation
-    required string recipient_name = 2;    // merchant's name or coordinator's name in case of CoinJoin
+    required string recipient_name = 2;    // merchant's name
     repeated PaymentRequestMemo memos = 3; // any memos that were signed as part of the request
     optional uint64 amount = 4;            // the sum of the external output amounts requested, required for non-CoinJoin
     required bytes signature = 5;          // the trusted party's signature of the paymentRequestDigest
--- a/common/protob/messages-bitcoin.proto
+++ b/common/protob/messages-bitcoin.proto
@@ -204,7 +204,7 @@ message SignTx {
      * Signing request for a CoinJoin transaction.
      */
     message CoinJoinRequest {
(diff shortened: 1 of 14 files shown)


core/src/apps/bitcoin/sign_tx/approvers.py
```diff
@@ -328,30 +335,84 @@ async def approve_tx(self, tx_info: TxInfo, orig_txs: list[OriginalTxInfo]) -> N
...
+
+        if not tx.coinjoin_request:
+            raise wire.DataError("Missing CoinJoin request.")
+
+        self.request = tx.coinjoin_request
         self.authorization = authorization
+        self.coordination_fee_base = 0
 
-        if authorization.params.coin_name != tx.coin_name:
-            raise wire.DataError("Coin name does not match authorization.")
```
andrewkozlik, line 344: Not needed. Already checked in `approve_tx()` by calling `self.authorization.approve_sign_tx(tx_info.tx)`.

andrewkozlik:
Added two commits based on discussion with @onvej-sl IRL:
[a4dbc3e](https://github.com/trezor/trezor-firmware/pull/2577/commits/a4dbc3e2fd1c051c210bf7a88d9db2b3da8f212b)
[34dc961](https://github.com/trezor/trezor-firmware/pull/2577/commits/34dc9617035e99f5eeddaa6dfd52935c4dac11af)

tests/device_tests/bitcoin/payment_req.py
```diff
@@ -98,3 +104,72 @@ def make_payment_request(
         nonce=nonce,
         signature=payment_req_signer.sign_digest_deterministic(h_pr.digest()),
     )
+
+
+def make_coinjoin_request(
```
onvej-sl, line 109: Please make this function deterministic so the tests that use the function are deterministic as well.

onvej-sl (changes requested):

tests/device_tests/bitcoin/payment_req.py
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/2577/commits/8f3bae4d87c0a0d62b145b3ec8b53bf7a3d9e7d6

common/protob/messages-bitcoin.proto
```diff
@@ -198,6 +198,18 @@ message SignTx {
     optional AmountUnit amount_unit = 11 [default=BITCOIN];    // show amounts in
     optional bool decred_staking_ticket = 12 [default=false];  // only for Decred, this is signing a ticket purchase
     optional bool serialize = 13 [default=true];               // serialize the full transaction, as opposed to only outputting the signatures
+    optional CoinJoinRequest coinjoin_request = 14;            // only for preauthorized CoinJoins
+
+    /**
+     * Signing request for a CoinJoin transaction.
+     */
+    message CoinJoinRequest {
+        required uint32 fee_rate = 1;                // coordination fee rate in units of 10^-8 percent
```
grdddj, line 207: This overflows `uint32` with 43 percent or more - not probable, but we might change it to `uint64` as protection.

grdddj (commented):
Haven't found anything bad, just one possible improvement.

common/protob/messages-bitcoin.proto
matejcik: reasonable.
(fwiw the firmware currently still ignores uint sizes so this is not an issue; however, that is going to be fixed at some point)

andrewkozlik: Firstly, a 43 % fee is crazy huge. However, I understand that we might want to represent percentages in other protobuf messages in the future and it would be nice to use a consistent format. The present scheme already allows 8 decimal digits of a percent, which seems way more precise than we might ever need. So rather than doubling the size of the data type, which would result in a maximum value of 184 billion %, I'd opt for decreasing the precision. If we go with 10<sup>–7</sup> or 10<sup>–6</sup>, then that can fit 429% or 4295% respectively and the precision is still great. We will probably never need more than 4 digits of percent precision anyway.

BTW note that the fee rate is specified in two places in protobuf `AuthorizeCoinJoin.max_coordinator_fee_rate` and `CoinJoinRequest.fee_rate`, so we definitely want to keep those consistent.

When CoinJoin was first implemented in July 2020, we had it at 10<sup>–9</sup>, see https://github.com/trezor/trezor-firmware/pull/1127. That means a maximum of 4.29 % could be represented. The meaning was also different (fee per anonymity unit vs. flat fee rate now). I recall that we discussed this at the standup back then and the argument was that 4 % (per anonymity unit) is huge.

In https://github.com/trezor/trezor-firmware/pull/2180 I changed the meaning to the flat fee rate and 8 digit precision to partially compensate the difference in meaning.

This is probably the last chance to easily change the precision, if we want to. It will also require changes in connect. After connect is deployed, it would probably be too complicated.

matejcik (approved):
ACK on protobuf changes.
didn't go through the rest of the code, i'll trust your reviews

common/protob/messages-bitcoin.proto
prusnak: +1 for 10^-6 precision. Let's not forget to change the `max_coordinator_fee_rate` too.

andrewkozlik: @prusnak interesting point, but I think there is a difference. Here we are talking about precision of 10^-6 of a _percent_, whereas in LND I understand the precision to be 10^-6 of the whole. In other words if we want to follow LND, then we should be talking about 10^-4 of a percent here.

prusnak: ah, right, never mind

andrewkozlik: So what will it be? 10^-4 (ppm, LND), 10^-6 (millionth of a percent), 10^-8 (current) etc.? To be honest, I have no strong preference. Anything between 4 and 8 is fine by me.

andrewkozlik: 10^-6 seems like a good middle ground.
Done in https://github.com/trezor/trezor-firmware/pull/2577/commits/52fca8cc50a4d8cb6f86959848d4f51c47e12ce0

onvej-sl (approved):

PR #2581, opened 2022-10-24
fix(core/rust/ui): add missing Child wrappers

mmilata:
I wonder if we should change `LayoutObj` so that it only wraps `Child<T>` as the top-level component. Or we can start returning `Child<SomeComponent>` from constructors like `SomeComponent::new`.

### DIFF
--- a/core/embed/rust/src/ui/model_tt/layout.rs
+++ b/core/embed/rust/src/ui/model_tt/layout.rs
@@ -296,14 +296,13 @@ extern "C" fn new_confirm_action(n_args: usize, args: *const Obj, kwargs: *mut M
         };
 
         let obj = if hold {
-            LayoutObj::new(
-                Frame::new(title, SwipeHoldPage::new(paragraphs, theme::BG)).into_child(),
-            )?
+            LayoutObj::new(Frame::new(title, SwipeHoldPage::new(paragraphs, theme::BG)))?
         } else {
             let buttons = Button::cancel_confirm_text(verb_cancel, verb);
-            LayoutObj::new(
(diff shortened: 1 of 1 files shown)


matejcik (approved):

matejcik:
wasn't there some work on `LayoutObj::new()` accepting `T` but auto-including `Child<T>` ?

matejcik:
well, no, it's actually [there already](https://github.com/trezor/trezor-firmware/blob/master/core/embed/rust/src/ui/layout/obj.rs#L122)

so why should we need this?

mmilata:
Huh, I must be going blind...

Means we need to remove the existing `.into_child()` calls in layout.rs?

matejcik:
probably.

would be nice to implement `new()` for `Child<T>` that just causes a compiler error, so that we don't miss things

matejcik (approved):

mmilata:
Is that even possible without something like negative trait bounds?

mmilata:
How would you create legit `LayoutObj` in that case?

matejcik:
by calling the more general `new<T>(root: T)` for any T that is not actually a `Child`?

mmilata:
Can you do that though? Using https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md or some other way?

matejcik:
you're right, that might work if it was a trait, but not if we're implementing things directly on the struct.

however it turns out that the solution is trivial: removing [this artifact](https://github.com/trezor/trezor-firmware/blob/master/core/embed/rust/src/ui/layout/obj.rs#L35-L42) causes `Child` not to implement the required `ComponentMsgObj` trait, and it doesn't seem to break anything else.

PR #2582, opened 2022-10-25
CoinJoin UI improvements

andrewkozlik:
Updates to the UI as discussed with @hynek-jina. Screens before and after. The last case has two alternatives, the latter with a new warning icon. @hynek-jina please ACK the revised design and say which alternative you prefer for the last one.

image image

image image

image image

image image image


hynek-jina:
Thanks. Yes.
And for the last screen use the gear please (_I think that the warning might scare someone into impulsively unplugging the Trezor_)

andrewkozlik:
> Maybe even waiting - clock icon or something like that could be better for the last screen.. But the gear is ok as well..

I left the read gear icon, I think it's fine.

core/src/apps/bitcoin/common.py
```diff
@@ -179,3 +180,20 @@ def tagged_hashwriter(tag: bytes) -> HashWriter:
...
+
+
+def format_fee_rate(
+    fee_rate: float, coin: CoinInfo, include_shortcut: bool = False
+) -> str:
+    # Use format_amount to get correct thousands separator -- micropython's built-in
+    # formatting doesn't add thousands sep to floating point numbers.
+    # We multiply by 100 to get a fixed-point integer with two decimal places,
+    # and add 0.5 to round to the nearest integer.
+    fee_rate_formatted = format_amount(int(fee_rate * 100 + 0.5), 2)
```
mmilata, line 192: Nit: why not `math.ceil`?

mmilata (approved):
LGTM, one nit in regarding existing code.

Adding `show_coinjoin` for UI2 to my TODO.

core/src/apps/bitcoin/common.py
prusnak: First of all, this is a refactor - the original code already exists in the master branch.

Second of all, `math.ceil` is not what `int(x + 0.5)` does. I guess you meant `round`? 

The original code was added in a41cfadbf1eff82e1e25c1ab7b960e73f1f0cdfa by @marnova. Martin: Did you have any special reason why not to use just `round(fee_rate, 2)`? We don't use this anywhere in the codebase, so there might be a reason why not to use it, but I don't know of any.

prusnak: First of all, this is a refactor - the original code already exists in the master branch.

Second of all, `math.ceil` is not what `int(x + 0.5)` does. I guess you meant `round`? 

The original code was added in 114376f7344b688dca3da65ec0ad9af16935ade8 by @matejcik. Matejcik: Did you have any special reason why not to use just `round(fee_rate, 2)`? We don't use this anywhere in the codebase, so there might be a reason why not to use it, but I don't know of any.

prusnak: Ah, stupid me - the comment above explains it: "Use format_amount to get correct thousands separator  - micropython's built-in formatting doesn't add thousands sep to floating point numbers."

Sorry for the noise.

andrewkozlik: `ceil` would round up, but we want to round to the nearest. Micropython doesn't implement `round()` AFAIK.

mmilata: It does, not in `math` but in `builtins`. Agreed that this is out of scope, apologies.

andrewkozlik: Oh interesting, that's good to know, thanks. Anyway, this particular line of code originated [here](https://github.com/trezor/trezor-firmware/commit/114376f7344b688dca3da65ec0ad9af16935ade8), so I felt no need to change it. Maybe @matejcik can provide insight as to why `int(x+0.5)` is better than `round(x)`, but TBH I think it doesn't matter.

matejcik: i vaguely remember that `round` still returns a float. i might be wrong though

PR #2584, opened 2022-10-26
fix(core/tests/monero): docker command line

mmilata:
Related to #2570. Also fixes shasum input format which requires two spaces on my system, hopefully it will work with two spaces with other implementations if there are any.


matejcik (approved):
utACK

Issue #2527, opened 2022-09-26
trezorlib: missing default value handling in protobuf


matejcik:
The `Field` type in protobuf.py [has a `default` slot](https://github.com/trezor/trezor-firmware/blob/master/python/src/trezorlib/protobuf.py#L153) which is [being used](https://github.com/trezor/trezor-firmware/blob/master/python/src/trezorlib/protobuf.py#L233) in the [appropriate locations](https://github.com/trezor/trezor-firmware/blob/master/python/src/trezorlib/protobuf.py#L363).

As it turns out, the values for that field are [never emitted](https://github.com/trezor/trezor-firmware/blob/master/python/src/trezorlib/_proto_messages.mako#L30).

(Somewhat related, the Rust protobuf descriptors [wrongly decode](https://github.com/trezor/trezor-firmware/blob/master/common/protob/pb2py#L641) `default=true` as false)

PR #2586, opened 2022-10-26
fix(python): add default value to protobuf.Field

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2527:
- add missing default value into creation of `protobuf.Field` in `python/src/trezorlib/messages.py`

There were two possible ways how to include the `default` kwarg :
1. when `field.default_value is not None`, as `None` is the "default-default" value
2. when `not field.required` - for all fields that do not enforce a value to be set

In the end, I chose `#2`, so seeing `default=...` is another sign that the field is not required.

> (Somewhat related, the Rust protobuf descriptors [wrongly decode](https://github.com/trezor/trezor-firmware/blob/master/common/protob/pb2py#L641) default=true as false)

I did not look into this, how can I replicate the issue?

### DIFF
--- a/common/protob/messages-nem.proto
+++ b/common/protob/messages-nem.proto
@@ -58,7 +58,7 @@ message NEMSignTx {
     message NEMTransfer {
         required string recipient = 1;              // Address of the recipient
         required uint64 amount = 2;                 // Amount of micro NEM that is transferred
-        optional bytes payload = 3 [default=""];    // Actual message data (unencrypted)
+        optional bytes payload = 3;                 // Actual message data (unencrypted)
         optional bytes public_key = 4;              // Public key of the recipient (for encrypted payloads)
         repeated NEMMosaic mosaics = 5;             // Attached mosaics
         /**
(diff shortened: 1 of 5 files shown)


matejcik:
> I did not look into this, how can I replicate the issue?

i believe https://github.com/trezor/trezor-firmware/pull/2507/commits/bd1324384ca533e4080409d9f9db2e9f8a1424ef is fixing that part

grdddj:
Unexpectedly, `legacy device tests` have failed because of the change, with couple `NEM` UI tests failing with [one extra screen](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3231519415/artifacts/test_ui_report/failed/T1_nem-test_signtx_transfers.py::test_nem_signtx_multiple_mosaics.html).

The reason was that `NEMTransfer::payload` had the default value of `b""`, and `Unencrypted message` layout was shown even in this case (`if (transfer->has_payload)`).

Fixed in [c2515bd](https://github.com/trezor/trezor-firmware/pull/2586/commits/c2515bd15db24be88761b5bbc3c16f6a6e703adf) by removing the explicit default from `messages-nem.proto`, which made `None` the default.

matejcik:
the NEM change should be a separate commit, otherwise LGTM and please get this ready to merge

matejcik:
there is still (or again?) a conflict

grdddj:
> there is still (or again?) a conflict

Yeah, `Coinjoin` previously, `Cardano` now. Resolved and force-pushed

PR #2587, opened 2022-10-26
Increase clock frequency to 180 MHz

TychoVrahe:
Increases CPU frequency to 180MHz for TT (and models based on STM32F42x)

On TT: for compatibility between older firmware/new bootloader and vice versa when switching layers the frequency is returned to original 168MHz.

Other models: frequency is set in boardloader.

Also the configuration for F405 / 120MHz was wrong, so i fixed that (wasn't used anywhere since T1 isnt using core boardloader).

### DIFF
--- a/core/.changelog.d/2587.changed
+++ b/core/.changelog.d/2587.changed
@@ -1,0 +1,1 @@
+CPU Frequency increased to 180 MHz
--- a/core/embed/boardloader/.changelog.d/2587.changed
+++ b/core/embed/boardloader/.changelog.d/2587.changed
@@ -1,0 +1,1 @@
+CPU Frequency increased to 180 MHz
--- a/core/embed/boardloader/main.c
+++ b/core/embed/boardloader/main.c
@@ -192,15 +192,6 @@ static secbool copy_sdcard(void) {
 }
 #endif
 
-// this function resets settings changed in boardloader, which might be
-// incompatible with older bootloader versions, where this setting might be
-// unknown
-void set_bld_compatible_settings(void) {
-#ifdef TREZOR_MODEL_T
-  display_set_big_endian();
-#endif
-}
-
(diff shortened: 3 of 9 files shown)


TychoVrahe:
in the fixup, i adjusted HSE division to achieve prescribed 1-2MHz VCO input frequency

hiviah (changes requested):
Seems that the clock changes work as expected on device, but we should

* put the `set_core_clock` into a header instead of having it as extern everywhere
* add comment to the function which clocks it sets how depending on parameter - main 168/180MHz, USB 48 MHz, do we use any other clock in clock network for display or something else based on this (?)

TychoVrahe:
@hiviah : https://github.com/trezor/trezor-firmware/pull/2587/commits/e5fe5e1f3ead7bbf26a7dab052923c2e64f8441d
1. added the header
2. tried to make the clock setting clearer and more versatile, inspired by how its done in legacy. unfortunately it still needs the defines since the config array is statically initialized after the SystemInit function runs for first time in boardloader. Let me know if this is better.
Basically all clocks are derived from the core clock, notably it affects:
SysTick - is reconfigured 
PWM for backlight - will be slightly faster, changing the configuration anyway here https://github.com/trezor/trezor-firmware/pull/2595
I2C for touch - will be slightly faster, currently set to 200kHz, the range is 10kHz-400kHz. Isn't reconfigured - the real frequency will be a bit different from what is in initialization. Maybe we should do the reconfig here, what do you think?
FMC for display data - will be slightly faster, seems to work fine
SDIO comm - will be slightly faster, seems to work fine
RGB LED on model R - i was actually counting with 180 MHz back then so this is corrected

Let me know if you can think of anything else that might be affected

hiviah (approved):
Retested, LGTM

PR #2589, opened 2022-10-28
Legacy docs: ImHex patterns for TRZF trezor.bin, TRZR+TRZF combined trezor.bin and bootloader binary images

### LINKED ISSUE #2562: Create firmware image pattern data for T1 firmware images for Kaitai or Imhex
When dealing with old existing firmware images, it's always "challenge" to decode what is where (especially there are two headers concatenated) manually.

There are 2 good cross-platform tools for this, [Kaiti Struct](https://kaitai.io/) (installable visualizer `ksv` or with [web IDE](https://ide.kaitai.io/)) and [Imhex](https://github.com/WerWolv/ImHex#screenshots). You can use structure viewer to decode various binary formats, better to look at screenshots to understand what it does.

It would be useful to have this for T1 FW images which luckily have fairly simple format, but still not much fun to compute offsets manually.

Imhex has also hex diff and decompiler, so pretty helpful to figure out diffs and decompile/understand old FW images. Even without pattern data it's already useful:

image

hiviah:
resolves #2562


matejcik (approved):

mmilata:
Any thoughts on upstreaming it to the standard library? I can put this on my TODO if there are no objections.

hiviah:
You could put it into the upstream patterns.

Each pattern needs a file for test, which should be as small as possible (the 46k intermediary would work best as samples, one with 2 headers, one with just TRZF header).

There is also question if using `ARMAddress` is better than `Address` - both `Address` and `ARMAddress.Address` via transform end up as correct pointer when used in expression e.g. in the reset handler, but `ARMAddress.Address` is shifted in pattern data view (value is shown before transform), which I found more confusing than using just Address and mask out the lowest Thumb bit.

PR #2592, opened 2022-10-31
fix(tests): fix instability of multicore runners

matejcik:
this resolves a problem when each xdist runner would clear the directory and stomp over one another


grdddj:
Does it fix https://github.com/trezor/trezor-firmware/issues/2493?

The changes look very familiar to me, haven't you done this somewhere else already?

grdddj (approved):
Looks good

PR #2593, opened 2022-10-31
UI2: Progress indicators

### LINKED ISSUE #2217: Monero: blinking while refreshing progress bars

### LINKED ISSUE #2574: Loader draws over homescreen when accessing wallet - TT UI2
When wallet is accessed and the `get_seed` function draws a loader, it does it on top of homescreen causing ugly looking mess.

Probably the display should be cleared first? 

Observing this on TT UI2. 

To reproduce just init the device and access standard wallet.

mmilata:
Fixes #2217 and fixes #2574. [[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3290566092/artifacts/master_diff/index.html)]

progress-loading progress-signing
progress-passphrase progress-cj

Monero

progress-xmrrefresh progress-xmrsign

### DIFF
--- a/core/embed/rust/src/ui/model_tt/layout.rs
+++ b/core/embed/rust/src/ui/model_tt/layout.rs
@@ -32,7 +32,9 @@ use crate::{
         layout::{
             obj::{ComponentMsgObj, LayoutObj},
             result::{CANCELLED, CONFIRMED, INFO},
-            util::{iter_into_array, iter_into_objs, ConfirmBlob, PropsList},
+            util::{
+                iter_into_array, iter_into_objs, upy_disable_animation, ConfirmBlob, PropsList,
+            },
         },
     },
 };
@@ -1174,6 +1176,10 @@ pub static mp_module_trezorui2: Module = obj_module! {
 
     /// INFO: object
     Qstr::MP_QSTR_INFO => INFO.as_obj(),
+
+    /// def disable_animation(disable: bool) -> None:
+    ///     """Disable animations, debug builds only."""
+    Qstr::MP_QSTR_disable_animation => obj_fn_1!(upy_disable_animation).as_obj(),
 
     /// def confirm_action(
     ///     *,
--- a/core/src/trezor/ui/layouts/tt_v2/__init__.py
+++ b/core/src/trezor/ui/layouts/tt_v2/__init__.py
@@ -24,6 +24,12 @@ CANCELLED = trezorui2.CANCELLED
 INFO = trezorui2.INFO
 
 
+if __debug__:
+    from trezor.utils import DISABLE_ANIMATION
+
+    trezorui2.disable_animation(bool(DISABLE_ANIMATION))
+
+
 class _RustLayout(ui.Layout):
     # pylint: disable=super-init-not-called
     def __init__(self, layout: Any, is_backup: bool = False):
--- a/core/embed/rust/librust_qstr.h
+++ b/core/embed/rust/librust_qstr.h
@@ -16,6 +16,7 @@ static void _librust_qstrs(void) {
   MP_QSTR_CONFIRMED;
   MP_QSTR_CANCELLED;
   MP_QSTR_INFO;
+  MP_QSTR_disable_animation;
   MP_QSTR_confirm_action;
   MP_QSTR_confirm_blob;
   MP_QSTR_confirm_properties;
--- a/core/embed/rust/src/ui/component/text/paragraphs.rs
+++ b/core/embed/rust/src/ui/component/text/paragraphs.rs
@@ -89,6 +89,10 @@ where
         self
     }
 
+    pub fn inner(&self) -> &T {
+        &self.source
+    }
+
     pub fn inner_mut(&mut self) -> &mut T {
         &mut self.source
     }
--- a/core/embed/rust/src/ui/layout/util.rs
+++ b/core/embed/rust/src/ui/layout/util.rs
@@ -6,10 +6,14 @@ use crate::{
         iter::{Iter, IterBuf},
         list::List,
         obj::Obj,
+        util::try_or_raise,
     },
-    ui::component::text::{
-        paragraphs::{Paragraph, ParagraphSource, ParagraphStrType},
-        TextStyle,
+    ui::{
+        component::text::{
+            paragraphs::{Paragraph, ParagraphSource, ParagraphStrType},
+            TextStyle,
+        },
+        util::set_animation_disabled,
     },
 };
 use cstr_core::cstr;
@@ -227,4 +231,12 @@ impl ParagraphStrType for StrBuffer {
     fn skip_prefix(&self, chars: usize) -> Self {
         self.offset(chars)
     }
+}
+
+pub extern "C" fn upy_disable_animation(disable: Obj) -> Obj {
+    let block = || {
+        set_animation_disabled(disable.try_into()?);
+        Ok(Obj::const_none())
+    };
+    unsafe { try_or_raise(block) }
 }
--- a/core/embed/rust/src/ui/model_t1/layout.rs
+++ b/core/embed/rust/src/ui/model_t1/layout.rs
@@ -13,6 +13,7 @@ use crate::{
         layout::{
             obj::{ComponentMsgObj, LayoutObj},
             result::{CANCELLED, CONFIRMED},
+            util::upy_disable_animation,
         },
     },
 };
@@ -115,6 +116,10 @@ pub static mp_module_trezorui2: Module = obj_module! {
 
     /// CANCELLED: object
     Qstr::MP_QSTR_CANCELLED => CANCELLED.as_obj(),
+
+    /// def disable_animation(disable: bool) -> None:
+    ///     """Disable animations, debug builds only."""
+    Qstr::MP_QSTR_disable_animation => obj_fn_1!(upy_disable_animation).as_obj(),
 
     /// def confirm_action(
     ///     *,
--- a/core/embed/rust/src/ui/model_tr/layout.rs
+++ b/core/embed/rust/src/ui/model_tr/layout.rs
@@ -13,6 +13,7 @@ use crate::{
         layout::{
             obj::{ComponentMsgObj, LayoutObj},
             result::{CANCELLED, CONFIRMED},
+            util::upy_disable_animation,
         },
     },
 };
@@ -115,6 +116,10 @@ pub static mp_module_trezorui2: Module = obj_module! {
 
     /// CANCELLED: object
     Qstr::MP_QSTR_CANCELLED => CANCELLED.as_obj(),
+
+    /// def disable_animation(disable: bool) -> None:
+    ///     """Disable animations, debug builds only."""
+    Qstr::MP_QSTR_disable_animation => obj_fn_1!(upy_disable_animation).as_obj(),
 
     /// def confirm_action(
     ///     *,
--- a/core/embed/rust/src/ui/model_tt/component/hold_to_confirm.rs
+++ b/core/embed/rust/src/ui/model_tt/component/hold_to_confirm.rs
@@ -3,6 +3,7 @@ use crate::{
     ui::{
         component::{Child, Component, ComponentExt, Event, EventCtx, FixedHeightBar, Pad},
         geometry::{Grid, Insets, Rect},
+        util::animation_disabled,
     },
 };
 
@@ -220,7 +221,7 @@ where
             loader.start_shrinking(ctx, now);
         }
         Some(ButtonMsg::Clicked) => {
-            if loader.is_completely_grown(now) {
+            if loader.is_completely_grown(now) || animation_disabled() {
                 return true;
             } else {
                 loader.start_shrinking(ctx, now);
--- a/core/embed/rust/src/ui/model_tt/component/loader.rs
+++ b/core/embed/rust/src/ui/model_tt/component/loader.rs
@@ -6,6 +6,7 @@ use crate::{
         display::{self, Color},
         geometry::{Offset, Rect},
         model_tt::constant,
+        util::animation_disabled,
     },
 };
 
@@ -131,7 +132,9 @@ impl Component for Loader {
         if let Event::Timer(EventCtx::ANIM_FRAME_TIMER) = event {
             if self.is_animating() {
                 // We have something to paint, so request to be painted in the next pass.
-                ctx.request_paint();
+                if !animation_disabled() {
+                    ctx.request_paint();
+                }
 
                 if self.is_completely_grown(now) {
                     return Some(LoaderMsg::GrownCompletely);
--- a/core/embed/rust/src/ui/util.rs
+++ b/core/embed/rust/src/ui/util.rs
@@ -31,6 +31,30 @@ pub fn u32_to_str(num: u32, buffer: &mut [u8]) -> Option<&str> {
     }
 }
 
+#[cfg(feature = "ui_debug")]
+static mut DISABLE_ANIMATION: bool = false;
+
+#[cfg(feature = "ui_debug")]
+pub fn animation_disabled() -> bool {
+    // SAFETY: single-threaded access
+    unsafe { DISABLE_ANIMATION }
+}
+
+#[cfg(feature = "ui_debug")]
+pub fn set_animation_disabled(disabled: bool) {
+    // SAFETY: single-threaded access
+    unsafe {
+        DISABLE_ANIMATION = disabled;
+    }
+}
+
+#[cfg(not(feature = "ui_debug"))]
+pub fn animation_disabled() -> bool {
+    false
+}
+#[cfg(not(feature = "ui_debug"))]
+pub fn set_animation_disabled(_disabled: bool) {}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/core/mocks/generated/trezorui2.pyi
+++ b/core/mocks/generated/trezorui2.pyi
@@ -4,6 +4,11 @@ CANCELLED: object
 
 
 # rust/src/ui/model_t1/layout.rs
+def disable_animation(disable: bool) -> None:
+    """Disable animations, debug builds only."""
+
+
+# rust/src/ui/model_t1/layout.rs
 def confirm_action(
     *,
     title: str,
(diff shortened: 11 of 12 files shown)


mmilata:
Seems to be triggering memory fragmentation issues on UI2 as well. Either there's some allocation in a loop somewhere or this is caused by module imports:thinking:

mmilata:
The problem was trezor.pin global variable holding reference to the layout which references trezor.ui, fixed in [6ac95da](https://github.com/trezor/trezor-firmware/pull/2593/commits/6ac95da36343ed5402c75bdf36333b75ebc66a8e).

Note to self: remove UI1 impl and leftovers in `core/src/trezor/ui/` after rebasing onto #2580.

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1339,6 +1372,17 @@ pub static mp_module_trezorui2: Module = obj_module! {
...
     Qstr::MP_QSTR_show_remaining_shares => obj_fn_kw!(0, new_show_remaining_shares).as_obj(),
+
+    /// def show_progress(
+    ///     *,
+    ///     title: str,
+    ///     indeterminate: bool = False,
+    ///     description: Callable[[], str] | None = None,
+    /// ) -> object:
+    ///    """Show progress loader. Please note that the number of lines reserved on screen for
+    ///    description is determined when object is place()d. If you want multiline descriptions
```
grdddj, line 1383: Typo: `place()d`

core/embed/rust/src/ui/model_tt/component/progress.rs
```diff
@@ -0,0 +1,136 @@
...
+        Self::AREA
+    }
+
+    fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
+        if let Event::Progress(new_value) = event {
+            if mem::replace(&mut self.value, new_value) != new_value {
+                if !animation_disabled() {
+                    ctx.request_paint();
+                }
+                let new_description = (self.description_func)();
```
grdddj, line 88: Do we currently have some use-case of `description_func` changing its return value between calls, or it is just a nice-to-have possibility for the future?

Bare function without arguments has not an easy job of tracking the progress easily - to return some appropriate description each time (it needs to store context in its global environment). What if we supplied the `new_value` into it, so the function could have some logic inside to response to changing (increasing) progress.

Simple example:
```python
def description(value: int) -> str:
    if value < 100:
        return "Beginning message"
    elif value < 900:
        return "Middle message"
    else:
        return "Finishing message"
```

grdddj (commented):
Good stuff!

I have some comments, but mostly questions about why we do certain things.

Can you also please give some easy way of triggering the progress in `Monero`?

I know that I do not do it myself, but it would be nice if all `PR`s included some quick instructions of how to trigger the changed functionality/screens (e.g. run this `trezorctl` command, run this `device test`, go to `Suite` and do X). Basically what the author of `PR` did in the development loop and while making sure it works. Something like `QA` instructions, that could be useful for reviewers.

Example for the `Bitcoin` progress bar:
```
$ ./emu.py  # enable animations, no "-a" flag
$ INTERACT=1 pytest ../tests/device_tests/bitcoin/test_signtx.py -k big_amount
```

Friendly reminder about the need of (little ugly) rebasing because of `micropython decrease` changes.

core/src/trezor/ui/__init__.py
```diff
@@ -124,7 +124,10 @@ def backlight_fade(val: int, delay: int = 14000, step: int = 15) -> None:
             display.backlight(val)
             return
     current = display.backlight()
-    if current > val:
+    if current < 0:
```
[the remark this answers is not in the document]
mmilata: This exists because of unit tests:
```
class TestSignSegwitTxNativeP2WPKH
  test_send_native_p2wpkh ... errored: Value must be between 0 and 255
  File "/home/mmilata/work/trezor-firmware-alt/core/tests/unittest.py", line 237, in run_class
  File "test_apps.bitcoin.segwit.signtx.native_p2wpkh.py", line 167, in test_send_native_p2wpkh
  File "../src/apps/bitcoin/sign_tx/bitcoin.py", line 54, in signer
  File "../src/apps/bitcoin/sign_tx/progress.py", line 34, in init
  File "../src/apps/bitcoin/sign_tx/progress.py", line 118, in report_init
  File "../src/trezor/ui/layouts/tt_v2/__init__.py", line 1127, in bitcoin_progress
  File "../src/trezor/ui/layouts/tt_v2/__init__.py", line 1100, in __init__
  File "../src/trezor/ui/__init__.py", line 130, in backlight_fade
ValueError: Value must be between 0 and 255
```

Eventually we should move all backlight handling to rust so there's no need to use it from python directly.

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -935,7 +947,14 @@ async def confirm_coinjoin(
 
 
 def show_coinjoin() -> None:
-    log.error(__name__, "show_coinjoin not implemented")
+    _RustLayout(
+        trezorui2.show_info(
+            title="CoinJoin in progress.",
+            description="Do not disconnect your Trezor.",
+            button="",
+            time_ms=1,
```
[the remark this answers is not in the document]
mmilata: `time_ms != 0` means the layout will not draw buttons. I wanted to avoid adding parameter for this but with `time_ms` it's not obvious ... re-doing this.

core/embed/rust/src/ui/model_tt/component/progress.rs
mmilata: PIN unlock progress changes the description to display number of seconds remaining, also Monero signing uses this.

The logical place for description updates would be in `Event::Progress(...)` but we currently don't have a way to send a python string along with an event and I didn't want to add it just for this. Using a closure for `description_func` still allows the caller (RustProgress) to provide the description when it changes.

grdddj: Right, I did not notice that client can be changing the description at any time in `RustProgress::report`, so  `(lambda: self.description)()` can return different results at any time.

core/src/trezor/ui/layouts/tt_v2/__init__.py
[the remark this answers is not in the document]
mmilata: See 50dd2cb12e9c263585b6b7698a41c0fa2fe73abd

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -1053,3 +1082,50 @@ async def request_pin_on_device(
...
+    ):
+        self.description = description or ""
+        self.layout: Any = trezorui2.show_progress(
+            title=title.upper(),
+            indeterminate=indeterminate,
+            description=None if description is None else lambda: self.description,
+        )
+        ui.backlight_fade(ui.style.BACKLIGHT_DIM)
+        ui.display.clear()
+        self.layout.attach_timer_fn(self.set_timer)
```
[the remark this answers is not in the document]
mmilata: The progress indicators should never set timers because they aren't integrated into the main loop like normal layouts. Using `NotImplementedError` here is confusing though, changed to `RuntimeError` in ea3e9a08fa2219e8f2129c107e4175f81c8c624a.

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -1084,6 +1087,12 @@ def draw_simple_text(title: str, description: str = "") -> None:
     ui.draw_simple(text)
 
 
+def request_passphrase_on_host() -> None:
```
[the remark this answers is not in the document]
mmilata: Good catch, updated in ea3e9a08fa2219e8f2129c107e4175f81c8c624a.

If we want pyright to check that layout functions with the same name take the same arguments across all models, having `__all__` in single file should suffice.

core/embed/rust/src/ui/model_tt/layout.rs
mmilata: Fixed in ea3e9a0.

grdddj:
ACK on the fixup changes.

Device tests are failing because of extra `Transaction loading` screen, the same as before the fixups.

core/src/trezor/pin.py
```diff
@@ -1,30 +1,27 @@
...
 
-from trezor import utils
+if TYPE_CHECKING:
+    from typing import Any
+    from trezor.ui.layouts.common import ProgressLayout
 
-_previous_progress: int | None = None
 _previous_seconds: int | None = None
+_previous_remaining: str | None = None
+_progress_layout: ProgressLayout | None = None
```
matejcik, line 9: ISTM we can now be super conservative in terms of imports.

it should be possible to:
(a) have a local copy of `backlight_fade()`
(b) import `trezorui` locally, instantiate a global `display = Display()` (it's a single block allocation that will remain static forever)
(c) import `trezorui2` locally and create a global `progress = trezorui2.show_progress()`
(d) run it via a bunch of plain functions

OTOH we are _almost_ at the point where we could skip the middleman and do this whole progress callback directly from Rust, so maybe instead I'll create a separate issue to do that

matejcik (commented):
ACK on Python side. I still want to take a look at the Rust side

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -159,6 +164,13 @@ def handle_timers(self) -> loop.Task:  # type: ignore [awaitable-is-generator]
     def page_count(self) -> int:
         return self.layout.page_count()
 
+    def draw_simple(self) -> None:
```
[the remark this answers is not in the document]
mmilata: Fixed in [3f8c350](https://github.com/trezor/trezor-firmware/pull/2593/commits/3f8c350063e4d25cb1a1004531bcdb28d36b82c9). Timers are not needed but we still should `Event::Attach` to all components.

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -1053,3 +1080,50 @@ async def request_pin_on_device(
...
+        ui.backlight_fade(ui.style.BACKLIGHT_DIM)
+        ui.display.clear()
+        self.layout.attach_timer_fn(self.set_timer)
+        self.layout.paint()
+        ui.backlight_fade(ui.style.BACKLIGHT_NORMAL)
+
+    def set_timer(self, token: int, deadline: int) -> None:
+        raise RuntimeError  # progress layouts should not set timers
+
+    def request_complete_repaint(self) -> None:
```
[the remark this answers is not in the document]
mmilata: Right, I thought this will be needed for cases when the progress is interleaved with confirmation dialogs, however this only happens in `apps/bitcoin/sign_tx/progress.py` and we just reinitialize whole layout object instead. 3f8c350

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1059,6 +1079,25 @@ extern "C" fn new_show_remaining_shares(n_args: usize, args: *const Obj, kwargs:
...
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let title: StrBuffer = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
+        let indeterminate: bool = kwargs.get_or(Qstr::MP_QSTR_indeterminate, false)?;
+        let description_callback: Obj = kwargs.get(Qstr::MP_QSTR_description)?;
+
+        let obj = if description_callback == Obj::const_none() {
+            LayoutObj::new(Progress::new(title, indeterminate, StrBuffer::empty))?
+        } else {
+            let callback = move || {
+                StrBuffer::try_from(description_callback.call_with_n_args(&[]).unwrap()).unwrap()
```
matejcik, line 1092: please use `unwrap!`

mmilata: 17e6630c75e14b600bf8f2fa458e771fcbe19949

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1059,6 +1079,25 @@ extern "C" fn new_show_remaining_shares(n_args: usize, args: *const Obj, kwargs:
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_show_progress(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let title: StrBuffer = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
+        let indeterminate: bool = kwargs.get_or(Qstr::MP_QSTR_indeterminate, false)?;
+        let description_callback: Obj = kwargs.get(Qstr::MP_QSTR_description)?;
+
+        let obj = if description_callback == Obj::const_none() {
```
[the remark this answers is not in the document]
mmilata: Closures are fun:)
```
error[E0308]: `if` and `else` have incompatible types
    --> src/ui/model_tt/layout.rs:1091:13
     |
1088 |            let callback = if description_callback == Obj::const_none() {
     |   ________________________-
1089 |  |             StrBuffer::empty
     |  |             ---------------- expected because of this
1090 |  |         } else {
1091 |  |             move || {
     |  |_____________^
1092 | ||                 StrBuffer::try_from(description_callback.call_with_n_args(&[]).unwrap()).unwrap()
1093 | ||             }
     | ||_____________^ expected fn item, found closure
1094 |  |         };
     |  |_________- `if` and `else` have incompatible types
     |
     = note: expected fn item `fn() -> StrBuffer {StrBuffer::empty}`
                found closure `[closure@src/ui/model_tt/layout.rs:1091:13: 1091:20]`
```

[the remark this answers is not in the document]
matejcik: oh great, so the two `Progress::new` calls are actually two different types.

could we declare the `description_func` as dyn somehow to avoid this? (and then probably make `move || StrBuffer::empty()` as the parameter. i mean we can probably do that in any case to make the above work, but that's not catching other uses of `Progress`)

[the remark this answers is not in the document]
mmilata: Because we currently don't pass any string in an event. We could add support, but then we'd have to add type parameter everywhere that touches it (or introduce dependency of ui to micropython).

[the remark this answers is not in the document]
mmilata: ... on the other hand we don't have to store the string or do weird things with it. Perhaps we can get away by adding `&'a str` to the event variant and convert from StrBuffer in `obj_event`. This will introduce lifetime parameter but that may not be all that bad. Will try it in the afternoon.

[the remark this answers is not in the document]
mmilata: Ok, that seems to work, just needs to make a copy of the description when it changes: 629528f79cd326dc7e3d9d79b06d8d60fa56a730

core/embed/rust/src/ui/model_tt/component/progress.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
+        if let Event::Progress(new_value, new_description) = event {
+            if mem::replace(&mut self.value, new_value) != new_value {
+                if !animation_disabled() {
+                    ctx.request_paint();
+                }
+                self.description.mutate(ctx, |ctx, para| {
+                    if para.inner_mut().content().as_ref() != new_description {
+                        para.inner_mut()
+                            .update((self.description_func)(new_description));
```
matejcik, line 91: this is very weird. how about instead we drop `description_func` altogether and give `Paragraph<StrBuffer>` a method `update_with(&mut self, &str)` (and later if we need it, we can add one for `Paragraph<&str>` too)

matejcik: ooohh but that would only work if we specified a trait for it, otherwise it won't build for `Progress<T>`, only for `Progress<StrBuffer>`

core/embed/rust/src/ui/model_tt/component/progress.rs
```diff
@@ -0,0 +1,136 @@
...
+where
+    F: Fn(&str) -> T,
+{
+    title: Child<Label<T>>,
+    value: u16,
+    loader_y_offset: i16,
+    indeterminate: bool,
+    description: Child<Paragraphs<Paragraph<T>>>,
+    description_pad: Pad,
+    description_func: F,
```
matejcik, line 28: ...but there _is_ an improvement we can do: change type of `description_func` from `F` to `fn(&str) -> Result<T, Error>` and pass in a static function (in our case `StrBuffer::alloc`)

and rename it to `update_description`? because "description func" was a function that _returns_ the description whereas this one will just allocate it

mmilata: Fixed in 489ddebfea522b4a339c9b1ef4e689978bcb1df1.

TIL that if you have a closure that doesn't capture any variables then it works with `fn(...) -> ...` syntax (not contained in this PR).

core/embed/rust/src/ui/model_tt/component/progress.rs
mmilata: We can add something like `fn from_str(&str) -> T` to the `ParagraphStrType` trait, or perhaps make it a supertrait of `From<&str>`/`TryFrom<&str>`, it's not too exotic a feature to require from a string type.

matejcik: leaving that up to you.
i think the fully correct solution here is "use StrBuffer for everything", moving `StrBuffer` to crate root and making it not depend on micropython, so I'd like to try that in any case (the main blocker being any possible adverse effect on bootloader code)

matejcik (approved):

PR #2597, opened 2022-11-02
Enable CoinJoin without experimental features enabled

andrewkozlik:
This PR:
- Removes the `unstable` flag from the `AuthorizeCoinJoin` message. This means I had to rework `test_experimental_features` a bit.
- Renames the `experimental` protobuf flag to `experimental_field`.
- Renames the `unstable` protobuf flag to `experimental_message`.
- Disables automatic activation of experimental features in device tests, which is almost never needed.
- Introduces `pytest.mark.experimental` to mark tests which actually need experimental features.
- The UI tests which involve PIN have changed due to a different pseudorandom layout of the pinpad and because EXPERIMENTAL MODE is no longer shown when the PIN is set.
- Seems to decrease the time taken by device tests by 3.5 %.


python/src/trezorlib/protobuf.py
```diff
@@ -188,7 +188,6 @@ def __init__(cls, name: str, bases: tuple, d: dict) -> None:
 
 class MessageType(metaclass=_MessageTypeMeta):
     MESSAGE_WIRE_TYPE: Optional[int] = None
-    UNSTABLE: bool = False
```
matejcik, line 191: why?

matejcik: ah because the mako does not generate it anyway

andrewkozlik: It looks unused to me. I think it's a relict of some no longer existing logic.

andrewkozlik: I think it ceased to be used in 266955ba65e419943f778019660517352630754f or somewhere around that commit.

PR #2603, opened 2022-11-04
trezorlib 0.13.4 release

matejcik:
No major changes, but I need to push out a version with the firmware parsing refactors


grdddj (approved):
Looks good

Issue #2612, opened 2022-11-04
Remove coordinator identifier screen


hynek-jina:
In current version we are displaying confusing name `CoinjoinCoordinatorIdentifier`.
We could hack it, but more clean solution is to remove this screen entirely. At least as long as we only have one coordinator...

![IMG_20221103_153031](https://user-images.githubusercontent.com/26002916/199987862-cfb4d1ed-e9b4-4c6c-898c-e18004286018.jpg)

andrewkozlik:
Internal issue: https://www.notion.so/satoshilabs/Do-not-confirm-coordinator-identifier-on-Trezor-f315bc707be4459e958c0f73ab2de6ca

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 22.12.1 (ddaacdb2245d7a909b3d2abfe918cff34d2e1609)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.12.1 Chrome/104.0.5112.124 Electron/20.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.4 Universal (revision ff35c77bad82c7acac1df6245940ebe0d6454642)
 - Transport: bridge 2.0.32

PR #2613, opened 2022-11-07
Do not confirm coordinator identifier on Trezor.

andrewkozlik:
See https://github.com/trezor/trezor-firmware/issues/2612.

Any additional warnings (path and mining fee) will appear after the user confirms, so that they don't pop up out of nowhere.


matejcik:
please fix the unused imports, otherwise LGTM

matejcik (approved):

PR #2614, opened 2022-11-07
fix layout flow in UI2

TychoVrahe:
This aims to fix two things:

- the backlight fade in was done before first rendering, which defeats the purpose
- when rust layout returns message, one more paint was done before moving on to fadeout and next layout/homescreen -> i think this is not useful and i have seen it cause some weirdness.


mmilata (approved):
Good catch, thanks!

I've force-pushed updated fixtures.json, can be merged once CI is green.

mmilata:
UI diff https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3287240062/artifacts/master_diff/index.html

PR #2632, opened 2022-11-11
make ChoicePage more flexible, fix middle button with arms position

TychoVrahe:
- making ChoicePage more flexible by making it accept implementation of Choice trait instead of ChoiceItem
- fixes positioning of middle button with arms -> x offset was used like with outline, but not needed -> button was somewhat uncentered

### DIFF
--- a/core/embed/rust/src/ui/model_tr/component/choice.rs
+++ b/core/embed/rust/src/ui/model_tr/component/choice.rs
@@ -1,9 +1,12 @@
+#[cfg(feature = "ui_debug")]
+use crate::trace::Trace;
 use crate::ui::{
     component::{Child, Component, Event, EventCtx, Pad},
     geometry::Rect,
+    model_tr::component::ButtonLayout,
 };
 
-use super::{theme, ButtonController, ButtonControllerMsg, ButtonPos, ChoiceItem};
+use super::{theme, ButtonController, ButtonControllerMsg, ButtonPos};
 
 pub enum ChoicePageMsg {
     Choice(u8),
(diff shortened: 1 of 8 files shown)


core/embed/rust/src/ui/model_tr/component/passphrase.rs
```diff
@@ -147,13 +147,7 @@ impl ChoiceFactoryPassphrase {
 }
 
 impl ChoiceFactory for ChoiceFactoryPassphrase {
-    fn get(&self, choice_index: u8) -> ChoiceItem {
```
grdddj, line 150: Do some of these moves of `get` within the same `impl` have some idea behind it? In some components it is done and not in others (and the order of `count` and `get` seems also random)

grdddj (approved):
Very nice, I appreciate the changes!

I do not have any change requests, just two points, that I can change myself in the future if I decide to.

Are you OK with merging it like this? (in the future these commits will probably get squashed together with the whole `model R` code)

core/embed/rust/src/ui/model_tr/component/passphrase.rs
TychoVrahe: no these are remainders of some experimentation, better to be reverted. the order should at least be same in all the components

core/embed/rust/src/ui/model_tr/component/choice.rs
```diff
@@ -168,7 +191,7 @@ where
 
         // Getting the remaining left and right areas.
         let (left_area, _center_area, right_area) =
-            available_area.split_center(self.current_choice().width_center());
+            available_area.split_center(self.choices.get(self.page_counter as u8).width_center());
```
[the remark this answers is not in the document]
TychoVrahe: The helper methods returned specific type `ChoiceItem` which is not possible anymore (the type is unknown at this struct), so i just went for deletion. You can probably use return type `&dyn Choice` if you want to reintroduce them

TychoVrahe:
feel free to merge it and squash it at will, if you don't mind being blamed for this in the future :)

PR #2633, opened 2022-11-11
fix success animation

TychoVrahe:
fixes success animation bugs (i16 overflow after changing coord type and comparison instead of shift)


grdddj (commented):
I am fine with the changes, but I would let someone more knowledgeable about the animations to approve it.

mmilata (approved):

PR #2642, opened 2022-11-15
Lower the default fees on Bitcoin Cash

danger89:
The default fees defined in `bcash.json` is an one-to-one copy of Bitcoin fees, which is _not_ correct.

As a Trezor user myself I **overpay** BCH fees in Trezor as well as losing some _privacy_ due to the high default fees.  

Thanks.

Best regards,
Melroy van den Berg

### DIFF
--- a/common/defs/bitcoin/bcash.json
+++ b/common/defs/bitcoin/bcash.json
@@ -28,10 +28,10 @@
   "fork_id": 0,
   "force_bip143": true,
   "default_fee_b": {
-    "Low": 10,
-    "Economy": 70,
-    "Normal": 140,
-    "High": 200
+    "Low": 1,
+    "Economy": 1,
+    "Normal": 1,
+    "High": 100
   },
   "dust_limit": 546,
   "blocktime_seconds": 600,


prusnak:
These values are used only if node does not report the fee values, so they are almost never used. 

So how exactly do you overpay fees?

danger89:
Ah I see. Well if indeed the backend node does not report the fee values, the current defined fallback fees are still too high for BCH.
In that case I will overpay. 

1sat/byte value is considered accepted min-fee and safe to use if the backend doesn't report any value and will reduce privacy compared to the current values. Everything you pay higher will not improve your privacy, since you (Trezor transactions) will stand-out above the rest.

cculianu:
Yeah these fees make trezor users stick out like a sore thumb on the network. They are the only txns you see really that pay this high a fee.

BCH doesn't have full mempools and fee pressure like BTC does.  1.0 sats/B is plenty.  We confirm txns every block.

matejcik:
where can I confirm these numbers? do you have a link to some sort of mempool stats for BCH?

matejcik:
> As a Trezor user myself I **overpay** BCH fees in Trezor as well as losing some _privacy_ due to the high default fees.

Also please note that accepting this PR in the firmware repo is unlikely to change the fees you're actually paying, because these are at the discretion of the wallet software that you use.
If you're using Suite, and Suite is actually looking at these numbers, then the real issue is either at Suite side or Blockbook side for not getting the right estimates from the backends.

ftrader:
> where can I confirm these numbers? do you have a link to some sort of mempool stats for BCH?

https://bch.loping.net/block-stats gives some idea in the "Fee Rates (avg, median, min)" graph.

There's also https://bitinfocharts.com/bitcoin%20cash/ whose numbers I find hard to interpret a.t.m. (not sure if correctly calculated) but essentially, wallets on BCH can take for granted that by paying the min fee of 1sat/byte , their transaction will be accepted and in nearly all cases, mined into the next block.

So I'd say the proposed change here is completely safe, esp. under the considerations mentioned that it's actually the backends which provide the actual feerate values if reachable.

matejcik:
Thanks.
The change is indeed safe, it's just that I want some sort of independent confirmation before including possibly completely nonsensical data.

matejcik (approved):

PR #2649, opened 2022-11-21
Prevent installing and running incompatible firmware

### LINKED ISSUE #2623: Add model information to firmware headers
trezor-core header should have a **single character** field `model` to indicate for which model the object is intended.

A good space for this is in the 8 reserved bytes after `fix_version`.

* legacy header should have model `'1'`. Legacy bootloader should check this field and refuse to install firmware if it's not there
  * ideally as part of #2513 
* core bootloader header should contain the field (set to 'R' or 'T'). There is no requirement for the boardloader to check this though; even if a mismatch is found, there is nothing the boardloader or the user can do about it. Just try to load it and hope for the best.
* core firmware header should contain the field (set to `'R'` or `'T'`). The bootloader must check this field and refuse to install firmware if there is a mismatch.
  * TT bootloader may want to accept older firmwares where this is set to `0`.
  * bootloader may look into the boardloader model info for the appropriate character (see #2324)
* for the hypothetical R/T multifirmware, we might choose a specific character (`'m'` for multi, perhaps?) or allow more characters

open question: should vendor header also be model-specific?

TychoVrahe:
Aims to disable running incompatible firmware to prevent
- installing incompatible bootloader and thus possibly bricking the device
- damaging HW by potentially setting the pins to something incompatible

Additionally, when installing bootloader, the bootloader monotonic version is checked to prevent downgrade, which would also cause bricking the device. 

Added 3 items to image header
- monotonic version
- hw model
- hw revision

Bootloader checks fw's image header to ensure compatible hw model and revision - both when installing and running the fw - this check is done agains bootloaders own HW_MODEL and HW_REVISION
Firmware checks bootloader's HW_MODEL agains board capabilities when installing bootloader
Firmware checks new bootloaders monotonic version against current bootloader's one. It also checks new bootloaders HW_MODEL and HW_REVISION against itself.

Expiry field in image header is abused to enable this functionality:
- bit 0 == 0 implies model T and disallow run/install on other models. The HW_MODEL and HW_REVISION are checked too, allowing model T values and zeros (for pre-check images compatibility).
- bit 0 == 1 implies this is implemented, check the HW_MODEL and HW_REVISION fields

Older model T bootloaders/boardloaders will run anything with expiry == 0.

builds on https://github.com/trezor/trezor-firmware/pull/2565 

closes #2623


core/embed/trezorhal/image.c
```diff
@@ -50,35 +50,71 @@ static secbool compute_pubkey(uint8_t sig_m, uint8_t sig_n,
...
+  memcpy(hdr->sig, data + IMAGE_HEADER_SIZE - IMAGE_SIG_SIZE + 1,
+         IMAGE_SIG_SIZE - 1);
+
   // TODO: expiry mechanism needs to be ironed out before production or those
   // devices won't accept expiring bootloaders (due to boardloader write
   // protection).
-  if (hdr->expiry != 0) return secfalse;
+  if ((hdr->expiry & 0xFFFFFFFE) != 0) return secfalse;
+
+  // abusing expiry field to implement model check
```
matejcik, line 78: the comment is IMHO misleading because the point of the expiry hack is to prevent old bootloaders from running TR firmwares, not to model-check in new bootloaders.

i would modify the logic like so:
(a) if  `expiry == hw_model == hw_revision == 0` then allow for model T, disallow for others
(b) if `hw_model` and `hw_revision` match our data, then allow

(of course this would be much nicer if we were allowed an early return for true, but we probably don't want to do that here)

this additionally disallows `expiry == 1` and missing model info which is IMHO a good thing.
OTOH it allows `expiry == 0` and model info filled in, which seems ok

core/embed/firmware/bl_check.c
```diff
@@ -97,6 +99,50 @@ void check_and_replace_bootloader(void) {
...
+    // reject downgrade
+    return;
+  }
+
+  const char *board_name = (const char *)get_board_name();
+  size_t board_name_len = strlen(board_name);
+  if ((board_name_len == 0) ||
+      strncmp("TREZORT", board_name, board_name_len) == 0) {
+    // no board capabilities, assume Model T
+    if ((new_bld_hdr.hw_model != 'T') && (new_bld_hdr.hw_model != 0)) {
```
matejcik, line 131: we're looking at the new bootloader being embedded in the new firmware, so ISTM we can ignore the case `new_bld_hdr.hw_model == 0`

TychoVrahe: if we are sure we will release new bootloader with next firmware

matejcik: that's a good point.
let's keep this as-is for now and I'll make a note somewhere to drop the other half of the condition with new bootloader release

core/embed/trezorhal/image.c
```diff
@@ -50,35 +50,71 @@ static secbool compute_pubkey(uint8_t sig_m, uint8_t sig_n,
   return sectrue * (0 == ed25519_cosi_combine_publickeys(res, keys, sig_m));
 }
 
-secbool load_image_header(const uint8_t *const data, const uint32_t magic,
-                          const uint32_t maxsize, uint8_t key_m, uint8_t key_n,
-                          const uint8_t *const *keys, image_header *const hdr) {
+secbool parse_image_header(const uint8_t *const data, const uint32_t magic,
```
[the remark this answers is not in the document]
TychoVrahe: yeah i didn't want to introduce the key check in firmware - it might, however, prevent a disaster if we in some weird state of mind release signed production firmware with unsigned bootloader and no other checks nor QA will catch that 

will try to rename things to make more sense.

core/embed/trezorhal/image.c
TychoVrahe: `expiry == 0` is not acceptable for non-TT images, as older TT bootloaders that dont check `hw_model` will be able to run those images.

So :
`expiry == hw_model == hw_revision == 0 ` allow for TT, reject for others

`expiry != 0 && hw_model == HW_MODEL && hw_revision == HW_REVISION` allow for all models, though expiry will still be set to 0 on TT, to allow running on existing boardloaders/bootloaders

since we still want to fill hw_model and hw revision into new TT images, we also need to allow:
`expiry == 0 && hw_model == HW_MODEL && hw_revision == HW_REVISION` 

that is basically current logic, apart from the fact that current logic allows also combinations of hw_model == 0 and hw_revision == HW_REVISION and vice versa, which might not be desirable, but also shouldn't happen.

Also using just one bit of expiry to do this, in order to allow more functionality in the `expiry` field in the future.

matejcik: my point was that branching on the expiry bit seems to muddle the logic (to my reading at least)

the current code, and your explanation too, flows in _three_ distinct branches:
```
if expiry_bit is 0 {
   if model_t and hw_model is 0 and hw_revision is 0 {
       ok
   } else if hw_model is expected and hw_revision is expected {
       ok
   }
} else if hw_model is expected and hw_revision is expected {
   ok
}
```
the way I'm reading it, "expiry_bit is 0" is _not_ the deciding factor, and there are _two_ options: either the header is old-style TT, or it is new-style anything.
```
if model_t and expiry_bit is 0 and hw_model is 0 and hw_revision is 0 {
   ok  /* special case for TT preexisting images */
else if hw_model is expected and hw_revision is expected { 
   ok  /* general case for new images from today on */
}
```

matejcik: or are you saying that the bootloader on TR should explicitly reject images that have `hw_model == `R` && expiry == 0` ? if yes, why?

TychoVrahe: because `expiry == 0` means TT image. So `hw_model == 'R' && expiry == 0` should never happen and something is broken

core/embed/trezorhal/image.c
```diff
@@ -50,35 +50,71 @@ static secbool compute_pubkey(uint8_t sig_m, uint8_t sig_n,
...
   if (hdr->hdrlen != IMAGE_HEADER_SIZE) return secfalse;
 
   memcpy(&hdr->expiry, data + 8, 4);
+  memcpy(&hdr->codelen, data + 12, 4);
+  memcpy(&hdr->version, data + 16, 4);
+  memcpy(&hdr->fix_version, data + 20, 4);
+  memcpy(&hdr->monotonic, data + 24, 1);
+  memcpy(&hdr->hw_model, data + 25, 1);
+  memcpy(&hdr->hw_revision, data + 26, 1);
+  memcpy(hdr->hashes, data + 32, 512);
```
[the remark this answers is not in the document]
prusnak: sigmask + signature are on a fixed offset, they start at `IMAGE_HEADER_SIZE - IMAGE_SIG_SIZE`

core/embed/trezorhal/image.c
TychoVrahe: the model_t condition is also valid for the `else if` branch

```
if expiry_bit is 0 {
   if model_t and hw_model is 0 and hw_revision is 0 {
       ok
   } else if model_t and hw_model is expected and hw_revision is expected {
       ok
   }
} else if hw_model is expected and hw_revision is expected {
   ok
}
```

matejcik: if the intention is to know that we don't produce `model == 'R' && expiry == 0` by mistake, a clearer way to write that would be to modify the `expiry & 0xfffe` on non-TT to explicitly check for 1

core/embed/firmware/bl_check.c
```diff
@@ -97,6 +99,50 @@ void check_and_replace_bootloader(void) {
...
+    // could also mean that the embedded bootloader is for another model
+    return;
+  }
+
+  image_header current_bld_hdr;
+  secbool old_header_valid =
+      parse_image_header(bl_data, BOOTLOADER_IMAGE_MAGIC,
+                         BOOTLOADER_IMAGE_MAXSIZE, &current_bld_hdr);
+
+  (void)old_header_valid;
```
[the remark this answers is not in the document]
TychoVrahe: Can't think of a scenario when this would be significant.

Even when using pre-check bootloader on TT devboard and flashing production TR firmware via st-link/j-link, this still shouldn't run because of the expiry mismatch. 

Otoh maybe there is some scenario a can't imagine right now, so i will introduce the check here. Unless we can think of scenario where overwriting invalid bootloader header would actually fix things. Like if bootloader runs the firmware, gets damaged somehow and this would repair it.. that probably can't happen though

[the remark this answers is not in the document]
matejcik: my first thought was along the lines of "we are on a TT hardware, the installed bootloader is for TR, and somehow we managed to load TT firmware with the correct bootloader in it.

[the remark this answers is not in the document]
matejcik: but yeah, that's super unlikely to happen through all the other measures that we have here.

core/embed/trezorhal/image.c
TychoVrahe: modified the conditions: 
https://github.com/trezor/trezor-firmware/commit/43d87b70a075d1093e2b7c0c9a96c13acfad3d1a

core/embed/firmware/bl_check.c
```diff
@@ -97,6 +99,50 @@ void check_and_replace_bootloader(void) {
       (const uint32_t *)&_binary_embed_firmware_bootloader_bin_start;
   const uint32_t len =
       (const uint32_t)&_binary_embed_firmware_bootloader_bin_size;
+
+  image_header new_bld_hdr;
+  secbool new_header_valid =
+      parse_image_header((uint8_t *)data, BOOTLOADER_IMAGE_MAGIC,
+                         BOOTLOADER_IMAGE_MAXSIZE, &new_bld_hdr);
+
+  if (new_header_valid != sectrue) {
```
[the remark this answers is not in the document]
TychoVrahe: added the abort: https://github.com/trezor/trezor-firmware/commit/43d87b70a075d1093e2b7c0c9a96c13acfad3d1a

core/embed/trezorhal/image.c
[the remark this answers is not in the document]
TychoVrahe: renamed and reshuffled:
https://github.com/trezor/trezor-firmware/commit/df3eccea7ad47dae23ec6f4c1d67990f19a3e39b

some basic checks were left in the `read_image_header` functions, since additional separation bloats the code elsewhere.

for vendor header the changes are smaller, left basically `read_vendor_header` with checks and `check_vendor_header_sig` for checking signature

core/embed/firmware/bl_check.c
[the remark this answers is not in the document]
TychoVrahe: added the check: https://github.com/trezor/trezor-firmware/commit/df3eccea7ad47dae23ec6f4c1d67990f19a3e39b#diff-64358366e8b14d6e519d7910beaa9a67feeb32d9b8f728456182cd3ac3be6d5d

core/embed/trezorhal/image.c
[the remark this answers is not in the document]
TychoVrahe: okay tried to use the struct approach, but it kinda leads to somewhat larger diff. 

unfortunately, for vendor_header this is not possible since the header has variable amount of pubkeys and the struct needs fixed size, so we could only do this for part of the struct before the pubkeys, which doesn't seem worth it.

so the result doesn't look entirely elegant because of the two different approaches.

https://github.com/trezor/trezor-firmware/commit/df3eccea7ad47dae23ec6f4c1d67990f19a3e39b

TychoVrahe:
tried to improve messages result also while at it: 
https://github.com/trezor/trezor-firmware/commit/2d8af5deb9de75b74783f2de894da1a39ab31ad5

TychoVrahe:
The check is added to boardloader when installing bootloader from SD card. Since we can't change this on produced devices it won't do much now.

matejcik:
waiting for Product feedback on "single character for model". looks like we will need more after all

core/embed/boardloader/main.c
```diff
@@ -229,18 +241,20 @@ int main(void) {
   }
 #endif
 
-  image_header hdr;
+  const image_header *hdr =
+      read_image_header((const uint8_t *)BOOTLOADER_START,
+                        BOOTLOADER_IMAGE_MAGIC, BOOTLOADER_IMAGE_MAXSIZE);
+
+  ensure(hdr != NULL ? sectrue : secfalse, "invalid bootloader header");
```
matejcik, line 248: again, by corrupting `hdr` a glitcher can continue to `check_image_header_sig` even if `read_image_header` returns an error.

would this even be a problem? a corrupted `hdr` is pointing to random memory, in the typical case outside the address space, right? so the chance of `check_image_header_sig` doing anything besides crashing is nil, right?

otoh checking that hdr == BOOTLOADER_START is a cheap way to avoid the issue in the first place without any obvious downsides

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/2649/commits/0c06166e66b15dc84c3f5ce1166ab7e7b36a028b

TychoVrahe:
model name implemented in: https://github.com/trezor/trezor-firmware/pull/2649/commits/286c5ab27bdb52420a0a350808e3cb1a2c9051a6

core/embed/bootloader/header.S
```diff
@@ -20,7 +24,10 @@ g_header:
     .byte FIX_VERSION_MINOR          // fix_vminor
     .byte FIX_VERSION_PATCH          // fix_vpatch
     .byte FIX_VERSION_BUILD          // fix_vbuild
-    . = . + 8                        // reserved
+    .word HW_MODEL                   // type of the designated hardware
+    .byte HW_REVISION                // revision of the designated hardware
+    .byte VERSION_MONOTONIC          // monotonic version
```
prusnak, line 29: Monotonic version of what? Of the hardware?

core/site_scons/tools.py
```diff
@@ -1,11 +1,46 @@
...
             'TREZOR_FONT_' + font_name + '_INCLUDE=\\"' + font.lower() + '.h\\"',
-            ]
+        ]
         sourcefile = 'embed/extmod/modtrezorui/fonts/' + font.lower() + '.c'
         if sourcefile not in sources:
             sources.append(sourcefile)
 
+
+def get_hw_model_as_number(hw_model):
+    return int.from_bytes(hw_model.encode(), 'little')
```
prusnak, line 13: Just to make sure - the resulting binary fill will contain `T2B1` not `1B2T`, right?

core/embed/bootloader/header.S
TychoVrahe: of the bootloader

core/site_scons/tools.py
TychoVrahe: yes.

ie for 'T2T1' this will convert to number 827601492, or 0x31543254, which is then stored on little endian arm in memory in this order: 0x54, 0x32, 0x54, 0x31, which is also the order how "T2T1" string would be stored.

core/embed/bootloader/header.S
prusnak: in that case please change the comment to `monotonic version of the bootloader`

TychoVrahe: i sort of don't want to do that:
- the `VERSION_xx` bytes don't comment this either
- this is also present in other headers (fw, reflash, prodtest), and adding the comment would cause unnecessary diff between the header files

Though i get that inserting the `HW_MODEL` between the `FIX_VERSION` and `VERSION_MONOTONIC` could be confusing (did this for alignment reasons), maybe something like: `monotonic version of the binary` so that there is no diff

core/embed/bootloader/header.S
```diff
@@ -10,7 +10,11 @@
 g_header:
     .byte 'T','R','Z','B'            // magic
     .word g_header_end - g_header    // hdrlen
+#ifdef TREZOR_MODEL_T
     .word 0                          // expiry
+#else
+    .word 1                          // expiry
```
prusnak, line 16: Why do we set this to 1 for model != T?

core/embed/bootloader/header.S
prusnak: OK

TychoVrahe:
https://github.com/trezor/trezor-firmware/commit/2d5c84861bcf54aa9ada6024166bb1c533fc34c8 commit adds dealing with legacy/model 1.

My logic is this:
- uploading older T1 images into older TT bootloaders - nothing we can do about that now.
- new T1 FW images wont work on old TT because this PR bumps T1 expiry. This is ok to do beacuse afaic T1 bootloader doesn't check expiry
- new T1 FW images wont work on new non-T1 bootloaders because they will check the HW_MODEL and HW_REVISION fields
- T1 bootloaders will check image model during upload and presence check, preventing installing and running incompatible images. Zeros in both fields will be accepted to allow installing older firmware.
- T1 FW doesn't check `board_capabilities` or bootloader image model when updating bootloader: T1 bootloader doesn't have that. 
- T1 FW won't probably run and try to replace bootloader on non-T1 hw anyway because it will fail to detect whitelisted bootloader

core/embed/bootloader/header.S
TychoVrahe: This breaks compatibility of non TT images with older TT bootloaders. So that even already released bootloaders will recognize TR firmware as invalid.

matejcik:
> uploading older T1 images into older TT bootloaders - nothing we can do about that now.

this never works anyway because T1 images are completely broken from TT bootloader point of view:
* vendor header is missing
* code hashes are wrong (sha256 vs blake2, different padding for empty sectors)
* signature is missing (T1 uses different fields for different type of signature from different keys)

TT images are typically too big for T1, and even if you stripped the vendor header and tried to push a TT image into T1, the hashes are wrong (see above)

so I don't think this is a concern at all, and I don't even think we need to make T1 bootloader larger by adding the "check model" code.

still, putting the model field into T1 image is a good idea, plus it will allow us to easily determine that the image is signed by the new keys, so let's make sure that it is a part of the SignMessage changeset

TychoVrahe:
ok sounds reasonable, removed the model check here https://github.com/trezor/trezor-firmware/pull/2649/commits/1a6417c7399e366cbccf35ed6efd916ce4777783

will cherrypick the rest when dealing with the SignMessage.

core/embed/bootloader/header.S
TychoVrahe: done here: https://github.com/trezor/trezor-firmware/pull/2649/commits/83fa07158a7c9ff7722740d1ec76b098e71982e6

prusnak:
About the version_monotonic ...

Maybe we could steal one byte (the last one) from expiry and use that as version_monotonic?

We don't need 32-bit resolution for expiration date. Heck, we might as well be fine with just 2 bytes for the expiration. So we could have 2 bytes for expiration, one byte version_monotonic and one byte reserved?

This would have a nice side-effect that if `version_monotonic > 0` it would not be possible to upload it into old version of the bootloader.

TychoVrahe:
@prusnak We can never use that in bootloader since TT boardloader already checks that whole expiry==0. 

Fw is already monotonic 1, so this would also break compatibility with current bootloaders immediately, but we could probably downgrade the fw monotonic, i don't think its used anywhere.

I don't know how useful would be preventing installing newer fw on old bootloaders, maybe if we need to do some processing during upgrade so to force people to go through intermediary? 

Not sure if we want different header structure in fw/bld because of this. We can achieve this be using another bit in the fw expiry if we need to.

We might also want to use the fw monotonic version the other way around, to prevent installing older official firmware on newer bootloaders. This might be necessary to do at some point, because if we ever need to use bootloader downgrade protection, we need to ensure that older firmware that don't check the monotonic version in the bootloader header cannot be installed anymore, as that would brick the device.

core/embed/trezorhal/board_capabilities.c
```diff
@@ -61,9 +60,7 @@ void parse_boardloader_capabilities() {
         // not used yet, just advance pointer
         break;
       case MODEL_NAME:
-        used_length = MIN(MODEL_NAME_MAX_LENGTH, length);
-        memcpy(board_name, pos, used_length);
-        board_name[MODEL_NAME_MAX_LENGTH] = 0;
+        memcpy((uint8_t *)&board_name, pos, sizeof(uint32_t));
```
matejcik, line 63: this should first check that `length == sizeof(uint32_t)`

TychoVrahe: fixed: https://github.com/trezor/trezor-firmware/commit/3172011eaa71ea86e25e94701f2cd9962a855891

matejcik (approved):
I believe this is ready, please rebase & merge unless you know of an outstanding issue?

TychoVrahe:
squashed and rebased. dropped the legacy changes since its done separately. will merge after ci passes

Issue #2635, opened 2022-11-14
TR design - header animation


Hannsek:
Animation in which long header bumps from left to right and conversely. 

See (the one on the right): https://www.figma.com/proto/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?page-id=3790%3A11545&node-id=5141%3A26329&viewport=11037%2C4757%2C1&scaling=min-zoom&starting-point-node-id=5141%3A26329

Keep in mind different sizes of scrollbar. https://github.com/trezor/trezor-firmware/issues/2630

PR #2654, opened 2022-11-22
Basic marquee implementation

TychoVrahe:
Implements basic bouncing marquee. 

Works for both model R and T. 

On T it would be possible to increase performance using DMA2D, but practically this seemed good enough, so went for simple solution for now. 

Partially implements #2635 , needs integration into model R branch.


TychoVrahe:
rebased

core/embed/rust/src/ui/component/marquee.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        let shift_width = if area_width > text_width {
+            area_width - text_width
+        } else {
+            text_width - area_width
+        };
+
+        let mut duration = (300 * shift_width as u32) / base_width as u32;
+        if duration < 300 {
+            duration = 300;
```
mmilata, line 137: Named constant please?

core/embed/rust/src/ui/component/marquee.rs
```diff
@@ -0,0 +1,229 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[cfg(feature = "ui_debug")]
+impl crate::trace::Trace for Marquee {
+    fn trace(&self, d: &mut dyn crate::trace::Tracer) {
+        d.open("Marquee");
```
mmilata, line 226: This should also contain the text so we can match on it in tests.

core/embed/rust/src/ui/component/marquee.rs
```diff
@@ -0,0 +1,229 @@
...
+impl Component for Marquee {
+    type Msg = Never;
+
+    fn place(&mut self, bounds: Rect) -> Rect {
+        let base_width = self.font.text_width("M");
+        let text_width = self.font.text_width(self.text.as_ref());
+        let area_width = bounds.width();
+
+        let shift_width = if area_width > text_width {
+            area_width - text_width
```
mmilata, line 130: When the text fits the area, is it going to just bounce from side to side? Maybe we should just not animate it in this case?

core/embed/rust/src/ui/component/marquee.rs
```diff
@@ -0,0 +1,229 @@
+use crate::{
+    micropython::buffer::StrBuffer,
```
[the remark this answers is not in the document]
TychoVrahe: agree, though here is my issue: this is used in TR branch from `Title` component (https://github.com/trezor/trezor-firmware/blob/grdddj/model_r_smaller/core/embed/rust/src/ui/model_tr/component/title.rs) , which also uses `StrBuffer`. So we would need to either:
- change `Title` (and components using it) not to use `StrBuffer`
- use `as_ref` when creating `Marquee` and add lifetimes to `Marquee` (and components that use it)
- possibly change `as_ref` of StrBuffer to return `&'static str`. Not sure whether this would be possible and correct.
- something else? any ideas?

There is additional related problem, many of the TR components use `StrBuffer`, but i need to use them from bootloader without micropython. So what i did in the TR bootloader PR is to actually move the `StrBuffer` out of micropython module and compile `StrBuffer` with or without micropython related stuff based on `#[cfg(feature = "micropython")]`. This would solve the problem here too. See here: https://github.com/trezor/trezor-firmware/blob/tychovrahe/rust_bootloader_/tr/core/embed/rust/src/strbuffer.rs (note that this PR is still very draftish, maybe it would be nicer not to duplicate the file and just conditionally compile specific functions or something)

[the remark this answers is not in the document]
TychoVrahe: Hmm also making Marquee generic would work and seems most painless

core/embed/rust/src/ui/component/marquee.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/2557830eb65a832fd1f7f475e9f442192ac5aae5

core/embed/rust/src/ui/component/marquee.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/2355d5755ccf6d652a2bd3d45b4d05caf3e097a8

core/embed/rust/src/ui/component/marquee.rs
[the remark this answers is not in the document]
TychoVrahe: did the generic approach: https://github.com/trezor/trezor-firmware/commit/a62e5de96358f93f2acdf409c55fbc3fe9878005

core/embed/rust/src/ui/component/marquee.rs
TychoVrahe: i actually solved this outside of marquee like this (see `needs_marquee` flag): https://github.com/trezor/trezor-firmware/blob/grdddj/model_r_smaller/core/embed/rust/src/ui/model_tr/component/title.rs

Advantage is that if you actually want bouncing text you can use this, disadvantage is that you need to deal with that in every place where you don't want the bouncing text. 

Not sure what is preferable

core/embed/rust/src/ui/component/marquee.rs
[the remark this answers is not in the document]
mmilata: yeah, making the str type generic is ugly but it's the simplest solution

core/embed/rust/src/ui/component/marquee.rs
mmilata: Can't imagine you'd ever want this, but maybe that's just problem with my imagination. I guess it really depends on the use cases ...

mmilata (approved):

STew790:
Nothing can be done here from QA side.

Issue #229, opened 2019-06-05
[feature] trezorlib API readthedocs


rfjakob:
Feature request: It would be nice to have the trezorlib API docs available via readthedocs.org or a similar service.

PR #2659, opened 2022-11-23
Trezorlib API on readthedocs

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/229:
- documents the most important parts of `trezorlib` API - e.g. `client`, `device`, `bitcoin`, `ethereum` modules
- creates the necessary infrastructure for automatic `readthedocs` documentation builds (through `mkdocs` and its friend packages)
- transforms all the `.rst` files into `.md`, so that they can be included in the docs
- current state of `readthedocs` documentation is available under https://grdddj-trezorlib-dev.readthedocs.io/en/latest/ - it is currently connected with my fork of the repo, which is needed so I can publish the docs (later `trezor` should take over this)

Next steps:
- `trezor` would need to "create" `readthedocs` account and setup the connection with `trezor/trezor-firmware` repository - and name it `trezorlib`
- decide which theme we would want to use - currently I am using the `material` theme, as I like its layout the most (below)
  - using the `readthedocs` theme does not have a good navigation/overview of all symbols
  
 
image

---

image

---

image

---

image


python/README.md
```diff
@@ -1,156 +1,3 @@
 # trezorlib
```
matejcik, line 1: could we instead symlink this to `python/docs/README.md` ? we need this file (a) for pypi page and (b) to show something nice in github view of the `python/` subfolder

matejcik:
while we're at this, could you take a brief look into `sphinx-click`? that might allow us to make a better documentation for trezorctl than what `make-options-md` is doing.

we'll make a separate issue if it looks like bigger piece of work

Hannsek:
In which state this is @matejcik? Maybe AI can help us to do this?

matejcik:
let's see! superseded by #6439

PR #2663, opened 2022-11-28
Funny shape erase button

### LINKED ISSUE #2661: funny button shape in PIN keyboard

TychoVrahe:
Implements funny shaped erase button.

I had to reshuffle the passphrase keyboard placement, since the resulting buttons were smaller and the icon didn't fit. 

The erase button is somewhat shorter than rest of the buttons, which is consistent with design, but not sure if desirable.

emu00000000

closes #2661


TychoVrahe:
passphrase diff: 
pp_before pp_after

prusnak:
Nitpick: disabled button looks weird (there is almost no contrast). I think there should be no button at all rather than showing it in disabled state (which I presume we don't have good designs for).

matejcik:
the design decision on that was that the button should have the same colors in enabled and disabled state.

matejcik:
> passphrase diff: pp_before pp_after

how does the entered passphrase text fit?

TychoVrahe:
Changed the icon color when disabled: https://github.com/trezor/trezor-firmware/commit/0f3affd306e72456abcddc4ef77c7c5cbaeaa0f8

empty

With some text:
text

TychoVrahe:
Added also to the mnemonc keyboard

emu00000000

mmilata (commented):
LGTM code-wise but please adjust the size of the buttons so that they form a regular grid with the other buttons, I believe that they are unaligned in figma by mistake.

mmilata:
Hopefully we won't add many new buttons like this, otherwise it would be better to implement drawing arbitrary polygons with rounded corners. For future reference: https://stackoverflow.com/questions/24771828/how-to-calculate-rounded-corners-for-a-polygon

TychoVrahe:
Adjusted the button width and also corner rounding on the right side, so it is the same as the regular buttons:

mnemonic
passphrase
pin

mmilata (approved):
Thanks, LGTM!

matejcik (approved):

PR #2668, opened 2022-11-30
optimize keyboard behavior

TychoVrahe:
This does some optimiazation/fixes of keyboards (mnemonic. pin and passphrase), namely:

1. adjusts long press to delete all input characters to 1500ms, to be tried
2. adds long press to delete to passphrase and mnemonic keyboards
3. fixes green button to confirm word in mnemonic keyboard
4. disallows activation of button when re-entered with touch move
5. changes font in passphrase input field to demibold (to match design)


matejcik (approved):
utACK

core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
```diff
@@ -192,31 +193,45 @@ impl Component for PassphraseKeyboard {
...
                 });
                 self.after_edit(ctx);
-                None
-            };
+                return None;
+            }
+            _ => {}
         }
 
+        if let Some(ButtonMsg::Clicked) = self.back.event(ctx, event) {}
```
mmilata, line 227: Leftover?

core/embed/rust/src/ui/model_tt/component/button.rs
```diff
@@ -247,11 +247,6 @@ where
             }
             Event::Touch(TouchEvent::TouchMove(pos)) => {
                 match self.state {
-                    State::Released if self.area.contains(pos) => {
-                        // Touch entered our area, transform to `Pressed` state.
-                        self.set(ctx, State::Pressed);
-                        return Some(ButtonMsg::Pressed);
-                    }
```
mmilata, line 254: Without this transition, does it still make sense to distinguish Released from Initial?

core/embed/rust/src/ui/model_tt/component/keyboard/passphrase.rs
TychoVrahe: removed: https://github.com/trezor/trezor-firmware/commit/0cf0232e6cb099bdff49411fc13a211c579ea247

core/embed/rust/src/ui/model_tt/component/button.rs
TychoVrahe: One thing i see is that Released -> Initial transition clears long timer when touch ends outside of buttons area. While it would work to clear the timer even in case when the button wasn't pressed, not sure if its cleaner solution

mmilata (approved):

PR #2669, opened 2022-12-01
UI2: support StrBuffer slicing

### LINKED ISSUE #2648: improve safety of buffer handling
Buffers used for DMA2D/rendering operations need special care when using - one needs to make sure not to use same buffer twice in the rendering function. 

Would be great if we had some mechanism to track buffer usage and panic/stop rendering when same buffer is requested more than once.

Originates from this discussion:
https://github.com/trezor/trezor-firmware/pull/2571#discussion_r1023031127

### LINKED ISSUE #2665: UI2: improve string handling safety
https://github.com/trezor/trezor-firmware/pull/2575#discussion_r1000686239
https://github.com/trezor/trezor-firmware/pull/2575#discussion_r1034816838

mmilata:
Fixes #2665. It seems to me we actually can't implement indexing by ranges as the type signature is `fn index(&self, index: Idx) -> &Self::Output` but we need to return new StrBuffer value, not a reference.

We can't create new `StrBuffer` from stack data so I had to drop the scratch buffer handling (introduced in #2575), the hexadecimal representation is now written to GC-allocated buffer. We can't do hexadecimal conversion without StrBuffer but I guess that shouldn't be a problem.

### DIFF
--- a/core/embed/rust/src/micropython/buffer.rs
+++ b/core/embed/rust/src/micropython/buffer.rs
@@ -16,9 +16,14 @@ use super::ffi;
 ///
 /// Given the above assumptions about MicroPython strings, working with
 /// StrBuffers in Rust is safe.
+///
+/// The `off` field represents offset from the `ptr` and allows us to do
+/// substring slices while keeping the head pointer as required by GC.
+#[repr(C)]
 pub struct StrBuffer {
     ptr: *const u8,
-    len: usize,
+    len: u16,
+    off: u16,
 }
 
 impl StrBuffer {
@@ -27,22 +32,41 @@ impl StrBuffer {
     }
 
     pub fn alloc(val: &str) -> Result<Self, Error> {
+        unsafe {
+            Self::alloc_with(val.len(), |buffer| {
+                // SAFETY: Memory should be freshly allocated and as such cannot overlap.
+                ptr::copy_nonoverlapping(val.as_ptr(), buffer.as_mut_ptr(), buffer.len())
+            })
+        }
+    }
+
+    pub fn alloc_with(len: usize, func: impl FnOnce(&mut [u8])) -> Result<Self, Error> {
         // SAFETY:
         // We assume that if `gc_alloc` returns successfully, the result is a valid
         // pointer to GC-controlled memory of at least `val.len() + 1` bytes.
         unsafe {
-            let raw = ffi::gc_alloc(val.len() + 1, 0) as *mut u8;
+            let raw = ffi::gc_alloc(len + 1, 0) as *mut u8;
             if raw.is_null() {
                 return Err(Error::AllocationFailed);
             }
-            // SAFETY: Memory should be freshly allocated and as such cannot overlap.
-            ptr::copy_nonoverlapping(val.as_ptr(), raw, val.len());
+
+            // SAFETY: GC returns valid pointers, slice is discarded after `func`.
+            let bytes = slice::from_raw_parts_mut(raw, len);
+            // GC returns uninitialized memory which we must make sure to overwrite,
+            // otherwise leftover references may keep alive otherwise dead
+            // objects. Zero the entire buffer so we don't have to rely on
+            // `func` doing it.
+            bytes.fill(0);
+            func(bytes);
+            str::from_utf8(bytes).map_err(|_| Error::OutOfRange)?;
+
             // Null-terminate the string for C ASCIIZ compatibility. This will not be
             // reflected in Rust-visible slice, the zero byte is after the end.
-            raw.add(val.len()).write(0);
+            raw.add(len).write(0);
             Ok(Self {
                 ptr: raw,
-                len: val.len(),
+                len: unwrap!(len.try_into()),
+                off: 0,
             })
         }
     }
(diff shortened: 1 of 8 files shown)


core/embed/rust/src/micropython/buffer.rs
```diff
@@ -16,9 +16,14 @@ use super::ffi;
...
 /// StrBuffers in Rust is safe.
+///
+/// The `off` field represents offset from the `ptr` and allows us to do
+/// substring slices while keeping the head pointer as required by GC.
+#[repr(C)]
 pub struct StrBuffer {
     ptr: *const u8,
-    len: usize,
+    len: u16,
+    off: u16,
```
mmilata, line 26: Not sure if premature optimization, now we can only have <64k length strings.

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -27,22 +32,48 @@ impl StrBuffer {
...
                 return Err(Error::AllocationFailed);
             }
-            // SAFETY: Memory should be freshly allocated and as such cannot overlap.
-            ptr::copy_nonoverlapping(val.as_ptr(), raw, val.len());
+
+            // SAFETY: GC returns valid pointers, slice is discarded after `func`.
+            let bytes = slice::from_raw_parts_mut(raw, len);
+            let initialized = func(bytes);
+            assert_eq!(initialized.as_ptr(), raw as *const u8);
+            assert_eq!(initialized.len(), len);
```
mmilata, line 68: Maybe instead fill the whatever's after initialized with zeroes? Need to make sure uninitialized memory doesn't escape.

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -27,22 +32,48 @@ impl StrBuffer {
...
+                ptr::copy_nonoverlapping(val.as_ptr(), buffer.as_mut_ptr(), buffer.len());
+                // SAFETY: Copy of valid UTF-8 string.
+                str::from_utf8_unchecked(slice::from_raw_parts(buffer.as_ptr(), buffer.len()))
+            })
+        }
+    }
+
+    /// SAFETY: The caller must ensure that `func` returns a slice that:
+    /// - spans the entire length of the given buffer, and
+    /// - contains valid UTF-8.
```
matejcik, line 48: This comment implies that `func` should be of type `-> ()`.
The return value is useless anyway because we only assert that it covers the whole slice and immediately throw it away. 

What we _could_ do is have it return `usize` so that `ret <= len`, allowing allocation above needed size. And if we wanted to make this function fully safe, we could then call `str::from_utf8`. ISTM we can afford that because it only triggers relatively rarely. Or we can have an unsafe internal helper used by both `alloc` and `alloc_with` that skips the validation, if this becomes a problem.

core/embed/rust/src/micropython/buffer.rs
matejcik: is an "uninitialized" `&[u8]` slice a problem?
my gut says no because whatever content is there, it's still makes for a valid `&[u8]` data.

mmilata: It may prevent other dead objects from being garbage collected: https://github.com/micropython/micropython/blob/master/py/gc.c#L689-L694

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -51,7 +82,18 @@ impl StrBuffer {
...
+    }
+
+    pub fn offset(&self, skip_bytes: usize) -> Self {
+        let off: u16 = skip_bytes.try_into().unwrap();
+        assert!(off <= self.len);
+        assert!(self.as_ref().is_char_boundary(skip_bytes));
+        Self {
+            ptr: self.ptr,
+            len: self.len.checked_sub(off).unwrap(),
+            off: self.off.checked_add(off).unwrap(),
```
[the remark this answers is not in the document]
mmilata: [57dfc3e](https://github.com/trezor/trezor-firmware/pull/2669/commits/57dfc3ea397eb5e29ba3d7c4d1dd6b4fc570acd0)

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -63,16 +52,19 @@ impl StrBuffer {
...
             let bytes = slice::from_raw_parts_mut(raw, len);
-            let initialized = func(bytes);
-            assert_eq!(initialized.as_ptr(), raw as *const u8);
-            assert_eq!(initialized.len(), len);
+            // GC returns uninitialized memory which we must make sure to overwrite, otherwise
+            // leftover references may keep alive otherwise dead objects. Zero the entire buffer so
+            // we don't have to rely on `func` doing it.
+            bytes.fill(0);
+            func(bytes);
+            unwrap!(str::from_utf8(bytes));
```
matejcik, line 60: more like `str::from_utf8(bytes)?;` no?

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -51,7 +74,18 @@ impl StrBuffer {
...
         } else {
-            unsafe { slice::from_raw_parts(self.ptr, self.len) }
+            unsafe { slice::from_raw_parts(self.ptr.add(self.off.into()), self.len.into()) }
+        }
+    }
+
+    pub fn offset(&self, skip_bytes: usize) -> Self {
+        let off: u16 = unwrap!(skip_bytes.try_into());
+        assert!(off <= self.len);
+        assert!(self.as_ref().is_char_boundary(skip_bytes));
```
matejcik, line 84: i'd prefer `ensure!` for both, these look important enough to do in production.
then it should be safe to leave off `checked_*` below: `self.len - off` is definitely safe if `off <= len`, `self.off + off` could only overflow if `self.off + self.len` could overflow, and given that `off` only advances by as much as `len` decreases, that should not be possible either.

core/embed/rust/src/micropython/buffer.rs
```diff
@@ -71,6 +105,7 @@ impl TryFrom<Obj> for StrBuffer {
             let new = Self {
                 ptr: bufinfo.buf as _,
                 len: bufinfo.len as _,
```
matejcik, line 107: this will need a check that `len` fits

core/embed/rust/src/micropython/buffer.rs
mmilata: 61244340b4d6a50adecb3e7c3313fb2fce53ffa3

core/embed/rust/src/micropython/buffer.rs
mmilata: 61244340b4d6a50adecb3e7c3313fb2fce53ffa3

core/embed/rust/src/micropython/buffer.rs
mmilata: `assert!` are checked in production, `ensure!` just adds fancy message, no?
Droped checked_ in 61244340b4d6a50adecb3e7c3313fb2fce53ffa3.

matejcik (approved):

PR #2671, opened 2022-12-05
Basic support for jpeg rendering

TychoVrahe:
Adds support for JPEG rendering.

Using http://elm-chan.org/fsw/tjpgd/00index.html , which needs complete review.


prusnak:
I think we can drop the `core/embed/extmod/modtrezorui/tjpgd/doc` folder

TychoVrahe:
superceded by https://github.com/trezor/trezor-firmware/pull/2673

PR #2673, opened 2022-12-08
Background jpg image for new homescreen

TychoVrahe:
supersedes https://github.com/trezor/trezor-firmware/pull/2571 by replacing upscaled toif with jpg.

builds on https://github.com/trezor/trezor-firmware/pull/2538 and https://github.com/trezor/trezor-firmware/pull/2671

This is somewhat slower, but looks definitely better. To make the slowness less disturbing, i adjusted the fade-out/fade-in mechanism to go to zero backlight when changing screens. 

The gradient dimm is replace by a rounded box for label, which also looks much better on the device.


core/embed/extmod/modtrezorui/tjpgd/tjpgdcnf.h
```diff
@@ -0,0 +1,33 @@
...
+/  2: Grayscale (8-bit/pix)
+*/
+
+#define	JD_USE_SCALE	1
+/* Switches output descaling feature.
+/  0: Disable
+/  1: Enable
+*/
+
+#define JD_TBLCLIP		0
```
prusnak, line 21: We should benchmark how faster/slower is using `JD_TBLCLIP == 1`

core/embed/extmod/modtrezorui/display.c
```diff
@@ -45,6 +46,15 @@
 #include "display_interface.h"
 
 static struct { int x, y; } DISPLAY_OFFSET;
+static uint8_t display_jpeg_work[3100];
```
prusnak, line 49: Can we extract `3100` as a constant or a define and describe how this is is calculated?

prusnak (changes requested):

core/embed/extmod/modtrezorui/tjpgd/tjpgdcnf.h
TychoVrahe: optimizaton -0s
TBLCLIP 1, FASTDECODE 1 - 95.3ms
TBLCLIP 0, FASTDECODE 1 - 85.0ms
TBLCLIP 0, FASTDECODE 0 - 143.8.0ms
TBLCLIP 0, FASTDECODE 2 - 81.0ms

optimizaton -03
TBLCLIP 1, FASTDECODE 1 - 87.3
TBLCLIP 0, FASTDECODE 1 - 77.3ms
TBLCLIP 0, FASTDECODE 0 - 86.2ms
TBLCLIP 0, FASTDECODE 2 - 71.5ms

Flash space usage is up about 1.5kB when optimizing on speed (with TBLCLIP=0, FASTDECODE=1).

Fastdecode 2 requires cca 6kB more RAM.

So: TBLCLIP = 1 is useless. 

Apart from this i am leaning towards setting this -03 and FASTDECODE=1, but we probably can live with anything but FASTDECODE=0 + -Os.

TychoVrahe: otoh maybe we could use the FASTDECOMP=2. I started using CCMRAM because of the other stuff anyway so i can place the work buffer there and enjoy a bit better performance. We can reclaim the ram anytime if we have better use for it.

prusnak:
I tried to compile this branch on macOS and I found this change has to be applied:

``` patch
diff --git a/core/embed/extmod/modtrezorui/buffers.c b/core/embed/extmod/modtrezorui/buffers.c
index 8c8b7a824..820acc62a 100644
--- a/core/embed/extmod/modtrezorui/buffers.c
+++ b/core/embed/extmod/modtrezorui/buffers.c
@@ -30,7 +30,7 @@
 #define BUFFER_SECTION
 #endif
 
-#if defined BOOTLOADER
+#if defined BOOTLOADER || TREZOR_EMULATOR
 #define NODMA_BUFFER_SECTION
 #else
 #define NODMA_BUFFER_SECTION __attribute__((section(".no_dma_buffers")))
```

Otherwise the build fails with the following error: 

```
embed/extmod/modtrezorui/buffers.c:52:1: error: argument to 'section' attribute is not valid for this target: mach-o section specifier requires a segment and section separated by a comma
NODMA_BUFFER_SECTION buffer_jpeg_t jpeg_buffers[BUFFERS_JPEG];
^
embed/extmod/modtrezorui/buffers.c:36:53: note: expanded from macro 'NODMA_BUFFER_SECTION'
#define NODMA_BUFFER_SECTION __attribute__((section(".no_dma_buffers")))
```

And we don't need to apply this section when building for the emulator, right?

core/embed/extmod/modtrezorui/display.c
TychoVrahe: added the comment: https://github.com/trezor/trezor-firmware/commit/c3f19f977e7e5a4c7f683f8789a774c8caca8dfa

TychoVrahe: added the comment: https://github.com/trezor/trezor-firmware/commit/2a3015b5100006d8dd6324fceb840be2b9dba594

TychoVrahe:
Yep for emulator we dont need to specify the section. no idea why it works on linux and not on mac. You are using arm cpu right? 

changed the placement: https://github.com/trezor/trezor-firmware/commit/6a01db74f3a7ce53da53ba8d675641ba58aae9fd

prusnak:
> no idea why it works on linux and not on mac

gcc vs clang

matejcik:
dunno if the code does it yet, but it needs to handle the stored homescreen not being recognized, and display the default one instead

TychoVrahe:
@matejcik yes this is handled: https://github.com/trezor/trezor-firmware/blob/ff35c77bad82c7acac1df6245940ebe0d6454642/core/embed/rust/src/ui/model_tt/component/homescreen/mod.rs#L292

core/embed/extmod/modtrezorui/tjpgd/tjpgd.c
```diff
@@ -0,0 +1,1157 @@
...
+	size_t len;
+	JRESULT rc;
+
+
+	memset(jd, 0, sizeof (JDEC));	/* Clear decompression object (this might be a problem if machine's null pointer is not all bits zero) */
+	jd->pool = pool;		/* Work memroy */
+	jd->sz_pool = sz_pool;	/* Size of given work memory */
+	jd->infunc = infunc;	/* Stream input function */
+	jd->device = dev;		/* I/O device identifier */
+  // TREZOR: moved from decomp function
```
mmilata, line 982: This is indented by spaces but the original file has tabs, is it intentional? Nitpick: would be nice to add trezor changes in separate commit to make porting to new upstream version easier.

mmilata (commented):
I'm not capable of reviewing tjpgd.c in a reasonable timeframe. Given the amount of pointer arithmetic we should probably assume that it contains errors. Can we put the jpeg homescreen into the encrypted part of storage?

In fact couple of problems were already found and the fixes published at https://github.com/cmumford/TJpgDec
There's also a fuzzer harness which would be nice to add to our fuzzing code, though this is probably out of scope of this PR.

Haven't reviewed render.rs in detail yet.

mmilata:
FWIW I tried running tjpgd through [C2Rust](https://c2rust.com/) and [the result](https://gist.github.com/mmilata/394ee34118741c52da1610d1f8219bc8) compiles but crashes on memory access when rendering. Probably fixable and the source can be much improved with simple changes ... but not sure how much effort would be to get rid of unsafe blocks entirely.

matejcik:
another option is gutting [jpeg_decoder](https://github.com/image-rs/jpeg-decoder)

matejcik:
> Can we put the jpeg homescreen into the encrypted part of storage?

I don't think that would help. The obvious attack vector is to get the user to download and install a slick homescreen, which is immediately decoded & displayed.

This sounds exactly as dangerous as we originally worried. Let's figure out if we can make the JPG decoder safe. If not, we'll go with TOIF (either avatar or fullscreen pixellated) for the upcoming release.

mmilata:
Ack, will try to see how far can i go with c2rustified tjpgd.

core/embed/extmod/modtrezorui/tjpgd/tjpgdcnf.h
```diff
@@ -0,0 +1,33 @@
...
+/* Specifies size of stream input buffer */
+
+#define JD_FORMAT		1
+/* Specifies output pixel format.
+/  0: RGB888 (24-bit/pix)
+/  1: RGB565 (16-bit/pix)
+/  2: Grayscale (8-bit/pix)
+*/
+
+#define	JD_USE_SCALE	1
```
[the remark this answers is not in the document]
TychoVrahe: disabled: https://github.com/trezor/trezor-firmware/commit/8162e230220a105c08f72327bb4cea8f38d44ee1

core/embed/extmod/modtrezorui/tjpgd/tjpgd.c
TychoVrahe: not intentional, fixed: https://github.com/trezor/trezor-firmware/pull/2673/commits/ec3313ded7dabd57cda40756139c15eafba6f62c

the trezor changes are in separate commit, ( https://github.com/trezor/trezor-firmware/commit/c590f555790f1b24874b0a46eaea3c547e279312 ) though not sure how helpful it will be. maybe we should consider making submodule?

core/embed/extmod/modtrezorui/display.h
```diff
@@ -61,9 +73,21 @@ void display_clear(void);
...
                        uint16_t *out_h, toif_format_t *out_format);
+
+bool display_jpeg_info(const uint8_t *data, uint32_t len, uint16_t *out_w,
+                       uint16_t *out_h, uint16_t *out_mcu_height);
+void display_jpeg(int x, int y, const uint8_t *data, uint32_t datalen);
+
+void display_jpeg_buffer_prepare(JDEC *jd, jpeg_context_t *jpg_context,
+                                 uint8_t *buffer, const uint8_t *data,
+                                 uint32_t datalen, uint32_t line_width);
+void display_jpeg_buffer_decomp(JDEC *jd);
```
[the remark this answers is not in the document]
TychoVrahe: Good idea. Rewritten to rust: https://github.com/trezor/trezor-firmware/commit/1fe1216f5f4c0a7520e843da758803aa22e2fe48 (and removed some leftovers https://github.com/trezor/trezor-firmware/commit/11f208e62522e7ca942e4f2e26326dde339394d6).

Moved the jpeg work data to buffers, but i am not happy about what became of that functionality. Originally i needed that because the buffers for rendering need to be placed in memory in a way accessible by the dma2d, so it needs special care in bootloader. Some of the buffers don't care and could be placed anywhere, they just don't fit on stack. Maybe the buffers for jpeg and blurring could be stored in heap/GC, as we most likely won't be using this in bootloader. 

Dma2d in also not used in bootloader yet, but that may change later.

TychoVrahe:
hmm translating the lib to rust looks promising.

since i rewrote functions `display_jpeg_buffer_prepare` etc to rust i tried to run this with the @mmilata converted code, it didn't crash for me but produced nonsense data. tried to make another translation and it actually works. 

in this branch is working example: https://github.com/trezor/trezor-firmware/tree/tychovrahe/homescreen/bg_jpg_rust

obviously the code needs a lot of love, but since now all of this is in rust, we can simplify the code:
- we won't need the `in` and `out` callbacks, it can just be specific functions that work for us
- we don't need the void* pointer to context, it can just be our specific struct with data we need
- we can get rid of the conversions to raw pointers etc

TychoVrahe:
Performance comparison:

C
flash - 1170952 B
homescreen render: 79.9ms
lockscreen render: 113.2ms

Rust:
flash - 1169416 B
homescreen render: 96.2ms
lockscreen render: 129.4ms

So it took some hit, but probably UX-wise still acceptable

mmilata:
Wondering if it's because array bounds checks got added or it's a difference between gcc vs. llvm.

TychoVrahe:
One thing is that the tjpgd C code was speed optimized which is not possible to do in rust on file level. Doesn't explain the whole difference though.

prusnak:
> One thing is that the tjpgd C code was speed optimized which is not possible to do in rust on file level.

This is not usable yet? https://rust-lang.github.io/rfcs/2412-optimize-attr.html

TychoVrahe:
hmm seems like its possible and i just didn't know it. even better. its experimental so far, but we are on nightly anyway. will do some measurement later

mmilata (commented):
Great work with the converted tjpgdec! Now that it uses minimal amount of `unsafe` we should consider for merging and improve it incrementally.

Overall looking good, added some comments mostly regarding style. Feel free to squash+rebase when addressing these.

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -465,6 +468,38 @@ extern "C" fn new_confirm_properties(n_args: usize, args: *const Obj, kwargs: *m
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_confirm_homescreen(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let title: StrBuffer = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
+        let data: Obj = kwargs.get(Qstr::MP_QSTR_image)?;
+        let buffer = unsafe { get_buffer(data) };
```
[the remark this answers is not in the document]
mmilata: Added an ugly fix for this in dcfa70c75d9bfbd25eb9853b1aeafc6a7eb18150. Feel free to drop it if it conflicts with your local changes.

core/SConscript.firmware
```diff
@@ -183,11 +183,17 @@ SOURCE_MOD += [
     'vendor/micropython/lib/uzlib/crc32.c',
     'vendor/micropython/lib/uzlib/tinflate.c',
 ]
+
+SOURCE_SPEED = [
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/954f8eb5a303e0e7a49d4fdc32f730abe97dfcf7

core/embed/rust/src/ui/layout/util.rs
```diff
@@ -240,3 +249,49 @@ pub extern "C" fn upy_disable_animation(disable: Obj) -> Obj {
...
+        }
+    };
+
+    unsafe { try_or_raise(block) }
+}
+
+#[cfg(feature = "jpeg")]
+pub extern "C" fn upy_jpeg_test(data: Obj) -> Obj {
+    let block = || {
+        let buffer = unsafe { get_buffer(data) };
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/26652abe81a8ce35c2508c92ef0ec84ea7e81403

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
```diff
@@ -0,0 +1,608 @@
...
+        if (line_num % mcu_height) == 0 && jpeg_ok {
+            jd_decomp(&mut jd, 0);
+        }
+    }
+
+    for y in 0..HEIGHT {
+        // several lines have been already decompressed before this loop, adjust for
+        // that
+        if y < HOMESCREEN_IMAGE_SIZE - (BLUR_RADIUS + 1) {
+            //dest = get_data(jd.buffer, line_num, mcu_height);
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/0430c934c9e03957798415294eb8fcef084589b8

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
```diff
@@ -0,0 +1,608 @@
...
+    } else {
+        next_text_idx += 1;
+        homescreen_position_text(unwrap!(texts.get(0)), text_buffer, &mut icon_data)
+    };
+
+    let (jpeg_size, mcu_height) = if let Some((size, mcu)) = jpeg_info(data) {
+        (size, mcu)
+    } else {
+        (Offset::zero(), 8)
+    };
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/c4436c75ce40f6e8d33b59fad644d23b43e49c0d

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
```diff
@@ -0,0 +1,608 @@
...
+                && x > DIM_BORDER
+                && x < WIDTH - DIM_BORDER)
+                || (y > HOMESCREEN_DIM_START
+                    && y < (HOMESCREEN_DIM_START + HOMESCREEN_DIM_HIEGHT)
+                    && x > DIM_BORDER + 1
+                    && x < WIDTH - (DIM_BORDER + 1))
+                || ((HOMESCREEN_DIM_START..=(HOMESCREEN_DIM_START + HOMESCREEN_DIM_HIEGHT))
+                    .contains(&y)
+                    && x > DIM_BORDER + 2
+                    && x < WIDTH - (DIM_BORDER + 2)))
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/c198e4fd018416b69e1be81f412ca55851043446

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
```diff
@@ -0,0 +1,608 @@
...
+    let mut line_num = 0;
+
+    set_window(screen());
+
+    let avgs_mem = unsafe { get_blurring_buffer(0, true) };
+    let avgs = &mut avgs_mem.buffer[0..DECOMP_LINES];
+    let mut avgs_totals = [[0_u16; HOMESCREEN_IMAGE_SIZE as usize]; COLORS];
+
+    let mut add_idx = 0;
+    let mut rem_idx = 0;
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/77d96ea8086046ac5a137e008db1874715e713ca

mmilata:
In 14a0df0b9a6e808a8336aacd82f83d5230921e22 I've refactored the code a bit and documented the `unsafe` conditions, please cherry-pick. It still has a long way to idiomatic code but we can do that in later PRs.

TychoVrahe:
Very nice improvement. It also increased decoding speed quite a lot - blurred homescreen now renders in about 160ms instead of 190ms. Guessing that removing the options from allocated slices helped.

I split the commit so i can squash later into separate commits.

I also increased the size of the work buffer a bit, because we needed to allocate a little bit more memory than C version to ensure non-overlapping slices - https://github.com/trezor/trezor-firmware/pull/2673/commits/89f56ef9f6e5d7d3c456998050a60155175e29f0

mmilata:
Did some more refactoring in 5ecf774ee30b0808a6c93f0a809666361d2f164e + a6f8e27d775d45657f610dc05d57b692d62d9acb, the decoder can now work with arbitrary data sources and sinks and is like 95% ready to be split off to a separate crate. The dynamic dispatch increases rendering time by 2-4ms and can be removed if needed. Also the API is more rustic now, sorry for big diff. Pls cherry-pick if you are ok with the changes.

TychoVrahe:
looks good, cherry picked.

mmilata (approved):
I think this is ready, please squash & rebase on top of master. Afterwards I'd like @matejcik to review at least the unsafe blocks.

TychoVrahe:
squashed & rebased, however the device tests are getting stuck, not sure why.

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
```diff
@@ -0,0 +1,621 @@
...
+            self.totals[GREEN_IDX][i] += self.lines[self.add_idx][GREEN_IDX][i];
+            self.totals[BLUE_IDX][i] += self.lines[self.add_idx][BLUE_IDX][i];
+        }
+    }
+
+    // adds one line and removes one line of averages to/from sliding total averages
+    fn vertical_avg(&mut self) {
+        for i in 0..HOMESCREEN_IMAGE_SIZE as usize {
+            self.totals[RED_IDX][i] +=
+                self.lines[self.add_idx][RED_IDX][i] - self.lines[self.rem_idx][RED_IDX][i];
```
mmilata, line 395: The test `test_apply_homescreen_toif` seems to panic here.

core/embed/rust/src/ui/display/tjpgd.rs
```diff
@@ -0,0 +1,1561 @@
...
+    f!(1.38704), f!(1.92388), f!(1.81226),       f!(1.63099), f!(1.38704), f!(1.08979), f!(0.75066),       f!(0.38268),
+    f!(1.30656), f!(1.81226), f!(1.70711),       f!(1.53636), f!(1.30656), f!(1.02656), f!(FRAC_1_SQRT_2), f!(0.36048),
+    f!(1.17588), f!(1.63099), f!(1.53636),       f!(1.38268), f!(1.17588), f!(0.92388), f!(0.63638),       f!(0.32442),
+    f!(1.00000), f!(1.38704), f!(1.30656),       f!(1.17588), f!(1.00000), f!(0.78570), f!(0.54120),       f!(0.27590),
+    f!(0.78570), f!(1.08979), f!(1.02656),       f!(0.92388), f!(0.78570), f!(0.61732), f!(0.42522),       f!(0.21677),
+    f!(0.54120), f!(0.75066), f!(FRAC_1_SQRT_2), f!(0.63638), f!(0.54120), f!(0.42522), f!(0.29290),       f!(0.14932),
+    f!(0.27590), f!(0.38268), f!(0.36048),       f!(0.32442), f!(0.27590), f!(0.21678), f!(0.14932),       f!(0.07612),
+];
+
+fn reinterpret_slice<T>(slice: &[i32]) -> &[T] {
```
matejcik, line 158: could we replace those calls with `slice.align_to()` / `align_to_mut`? that way it would be much clearer, at usage site, that the usage is safe
(fwiw as written these functions are very unsafe because they _internally_ rely on the type being integer

TychoVrahe: https://github.com/trezor/trezor-firmware/pull/2673/commits/dc8b6172198532b1920a3089e3cfb674855c71a8

bosomt:
**QA OK**

new UI2 is correctly detected


_2.5.3_
```
 ~/Downloads  trezorctl set homescreen Cat03.jpg                                                                                               
Error: Wrong size of image - should be 144x144
```

_2.5.4_
```
trezorctl get_features
    homescreen_format: <HomescreenFormat.Jpeg240x240: 2>,
0x0a955b4ad6d9fb5e7a1dbd3333548ffd8b33399a

trezorctl set homescreen scan.jpg                                                                                            
Error: Wrong size of image - should be 240x240
```

image

PR #2675, opened 2022-12-11
chore(legacy): remove CoSi functionality

prusnak:
Followup to https://github.com/trezor/trezor-firmware/issues/450

Let's not merge before we have successfully tested that the change above works on Trezor Model T!


common/protob/messages.proto
```diff
@@ -161,10 +161,13 @@ enum MessageType {
     MessageType_SignedIdentity = 54 [(bitcoin_only) = true, (wire_out) = true];
     MessageType_GetECDHSessionKey = 61 [(bitcoin_only) = true, (wire_in) = true];
     MessageType_ECDHSessionKey = 62 [(bitcoin_only) = true, (wire_out) = true];
+    reserved 71 to 74;
+    /*
```
matejcik, line 165: this would block the messages for TT too

obrusvit:
@Hannsek I suggest to close this PR in favor of #3551 where I `cherry-picked` also the commits from stick.

PR #2679, opened 2022-12-12
chore(ci): temporary allow_failure for connect tests

vdovhanych:
Lets allow it to fail for now until i figure it out. We are missing latest master builds in tenv due to this.


mmilata (approved):
Ack, let's do it for now.

PR #2681, opened 2022-12-13
UI tests: add button to update fixtures directly from report

TychoVrahe:
This enables convenient fixtures updating while checking results of UI tests.

### DIFF
--- a/tests/show_results.py
+++ b/tests/show_results.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import http.server
+import json
 import multiprocessing
 import os
 import posixpath
@@ -13,6 +14,7 @@ import click
 
 ROOT = Path(__file__).resolve().parent.parent
 TEST_RESULT_PATH = ROOT / "tests" / "ui_tests" / "reporting" / "reports" / "test"
+FIXTURES_PATH = ROOT / "tests" / "ui_tests" / "fixtures.json"
 
 
 class NoCacheRequestHandler(http.server.SimpleHTTPRequestHandler):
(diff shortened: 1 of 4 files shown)


matejcik (approved):
cool! utACK

tests/show_results.py
```diff
@@ -52,6 +54,29 @@ def translate_path(self, path: str) -> str:
...
+            length = int(self.headers.get("content-length"))
+            field_data = self.rfile.read(length)
+            data = json.loads(field_data)
+
+            with open(FIXTURES_PATH, "r") as jsonFile:
+                fixtures = json.load(jsonFile)
+
+            test_name = data.get("test", None)
+            test_hash = data.get("hash", None)
+            if test_name is not None and test_hash is not None:
```
grdddj, line 69: All the reading and writing of the file can be done in this if-statement - as when it is False, it will just write the same data, which is unnecessary work.

tests/show_results.py
```diff
@@ -52,6 +54,29 @@ def translate_path(self, path: str) -> str:
...
+        if self.path == "/fixtures.json" and os.path.exists(FIXTURES_PATH):
+
+            length = int(self.headers.get("content-length"))
+            field_data = self.rfile.read(length)
+            data = json.loads(field_data)
+
+            with open(FIXTURES_PATH, "r") as jsonFile:
+                fixtures = json.load(jsonFile)
+
+            test_name = data.get("test", None)
```
grdddj, line 67: `None` explicit default value is theoretically not needed, `get(key)` has `None` default by default.

tests/show_results.py
```diff
@@ -52,6 +54,29 @@ def translate_path(self, path: str) -> str:
             path += "/"
         return path
 
+    def do_POST(self) -> None:
+        if self.path == "/fixtures.json" and os.path.exists(FIXTURES_PATH):
```
grdddj, line 58: As `FIXTURES_PATH` is already a `Path` (not a string), it is possible to use `FIXTURES_PATH.exists()`

grdddj (commented):
Nice possibility! Did not test it, just some possible python improvements

tests/show_results.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/a2900711241302cb3b2b7f4d1cac0f538ecf4960

tests/show_results.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/a2900711241302cb3b2b7f4d1cac0f538ecf4960

tests/show_results.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/a2900711241302cb3b2b7f4d1cac0f538ecf4960

grdddj (approved):

PR #2682, opened 2022-12-13
refactor(core): Allow eth & btc signing at m / 45' / coin_type / account / change / address_index

0xBEEFCAF3:
## Toward an Unhardened Multisig Standard

Casa and other multisig providers have learned that the use of extensive key hardening causes usability issues. This is due to additional complexities that arise from having to coordinate the use of multiple geographically distributed keys. The degradation of user experience is not offset by any security improvements because the primary security of a multisignature wallet comes from preventing any given key from being a single point of failure. 

With that said, multisig coordinators should only need to harden up to the purpose path.
Mutlisig coordinators should sign at `m / 45' / coin_type / account / change / address_index`.

### DIFF
--- a/core/src/apps/common/paths.py
+++ b/core/src/apps/common/paths.py
@@ -338,6 +338,8 @@ PATTERN_SEP5 = "m/44'/coin_type'/account'"
 # https://github.com/trezor/trezor-firmware/issues/1749
 PATTERN_SEP5_LEDGER_LIVE_LEGACY = "m/44'/coin_type'/0'/account"
 
+PATTERN_CASA = "m/45'/coin_type/account/change/address_index"
+
 
 async def validate_path(
     ctx: wire.Context,
--- a/core/src/apps/ethereum/keychain.py
+++ b/core/src/apps/ethereum/keychain.py
@@ -44,16 +44,19 @@ PATTERNS_ADDRESS = (
     paths.PATTERN_BIP44,
     paths.PATTERN_SEP5,
     paths.PATTERN_SEP5_LEDGER_LIVE_LEGACY,
+    paths.PATTERN_CASA,
 )
 
 
 def _schemas_from_address_n(
     patterns: Iterable[str], address_n: paths.Bip32Path
 ) -> Iterable[paths.PathSchema]:
-    if len(address_n) < 2:
-        return ()
+    # Casa paths (purpose of 45) do not have hardened coin types
+    if address_n[0] == 45 | paths.HARDENED and not address_n[1] & paths.HARDENED:
+        slip44_hardened = address_n[1] | paths.HARDENED
+    else:
+        slip44_hardened = address_n[1]
 
-    slip44_hardened = address_n[1]
     if slip44_hardened not in networks.all_slip44_ids_hardened():
         return ()
 
--- a/common/tests/fixtures/ethereum/signmessage.json
+++ b/common/tests/fixtures/ethereum/signmessage.json
@@ -16,6 +16,16 @@
         },
         {
             "parameters": {
+                "msg": "This is an example of a signed message at a different path.",
+                "path": "m/45'/60/2/1/1"
+            },
+            "result": {
+                "address": "0x4C28fB5643a2F9B03736103188845C5B50300242",
+                "sig": "0x10cb53ab88112fa0a915ab3e477843ab534ba1a1caeb70747373a86cba69ca6c5f70367d8f9023e1d7cd9fe34279dd9266dd26d7390acbfdea777c4c436de4df1c"
+            }
+        },
+        {
+            "parameters": {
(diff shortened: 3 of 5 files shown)


core/src/apps/common/paths.py
```diff
@@ -338,6 +338,9 @@ def match(path: Bip32Path) -> bool:
 # https://github.com/trezor/trezor-firmware/issues/1749
 PATTERN_SEP5_LEDGER_LIVE_LEGACY = "m/44'/coin_type'/0'/account"
 
+PATTERN_CASA = "m/45'/[0-1000000]/[0-1000000]/[0-1]/[0-1000000]"
```
matejcik, line 341: any reason for not using the actual schema `"m/45'/coin_type/account/change/address_index"`?

core/src/apps/common/paths.py
```diff
@@ -338,6 +338,9 @@ def match(path: Bip32Path) -> bool:
...
+PATTERN_CASA_EXPORT = "m/45'"
```
matejcik, line 342: I don't think we need to consider "casa export" at all. You can export the public keys via `GetPublicKey` that has no path restrictions (and we're bringing `EthereumGetPublicKey` in line with that in #2410) and the export path should not be valid in any other context.

core/src/apps/ethereum/keychain.py
```diff
@@ -44,16 +44,25 @@
...
-    if len(address_n) < 2:
+    # Exclude Casa export paths (m/45')
+    if len(address_n) < 2 and not address_n[0] == 0x8000002D:
         return ()
 
-    slip44_hardened = address_n[1]
+    # Casa paths (purpose of 45) do not have hardened coin types
+    # Hardcode 60' for slip44 id
+    if address_n[0] == 0x8000002D:
+        slip44_hardened = 0x8000003c
```
matejcik, line 62:
```suggestion
    if address_n[0] == 45 | paths.HARDENED and not address_n[1] & paths.HARDENED:
        slip44_hardened = address_n[1] | paths.HARDENED
```

core/src/apps/common/paths.py
0xBEEFCAF3: Perfect! I'll try that out

core/src/apps/common/paths.py
0xBEEFCAF3: Adding anti-ransom limits. An attacker could derive a change address at a non-standard path -- assuming they have compromised the wallet software.

Happy to use the actual schema. Let me know what you think.

matejcik: I think you misunderstood me. Yes, I'm suggesting using the string `"m/45'/coin_type/account/change/address_index"` because `coin_type`, `account`, `change` and `address_index` are well defined aliases, in use on all the surrounding lines. My question is, is there any strong reason you made the limits explicit? If not, please use the aliases.

0xBEEFCAF3: Fixed! and updated the description

core/src/apps/ethereum/keychain.py
0xBEEFCAF3: Fixed

core/src/apps/ethereum/keychain.py
```diff
@@ -44,16 +44,24 @@
...
     paths.PATTERN_SEP5_LEDGER_LIVE_LEGACY,
+    paths.PATTERN_CASA,
 )
 
 
 def _schemas_from_address_n(
     patterns: Iterable[str], address_n: paths.Bip32Path
 ) -> Iterable[paths.PathSchema]:
-    if len(address_n) < 2:
+    # Exclude Casa export paths (m/45')
```
matejcik, line 54: i don't think this should be here anymore?

0xBEEFCAF3: Good catch

0xBEEFCAF3: Fixed

core/src/apps/ethereum/keychain.py
```diff
@@ -44,16 +44,20 @@
...
 )
 
 
 def _schemas_from_address_n(
     patterns: Iterable[str], address_n: paths.Bip32Path
 ) -> Iterable[paths.PathSchema]:
-    if len(address_n) < 2:
-        return ()
+    # Casa paths (purpose of 45) do not have hardened coin types
+    # Hardcode 60' for slip44 id
```
matejcik, line 55: this "hardcode" comment seems outdated

matejcik (changes requested):
you probably also want to add a test for Ethereum sign-tx on your path

0xBEEFCAF3:
> please also add test for EthereumSignTx, that seems to be the more important part -- or will you not be using EthereumSignTx with the new paths?

We will not be using EthereumSignTx for these new paths -- only message signing

matejcik (approved):

grdddj:
@matejcik seems like this merge has broken device/UI tests.

The addition of a new test at the second place in `signmessage.json` means the UI tests got out of sync - this has an easy fix of putting it as a last element and recording the fixtures.

However, a bigger issue seems that the new test doesn't even work:
```
E       AssertionError: assert '0x3beC5F707E...3cd089E8Ea02C' == '0x4C28fB5643...45C5B50300242'
E         - 0x4C28fB5643a2F9B03736103188845C5B50300242
E         + 0x3beC5F707Ef56057354f4c062C53cd089E8Ea02C
```

Gitlb job: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3621363252

The same happens on my machine as well.

Not sure whether there is a mistake in the assertion or in the code

0xBEEFCAF3:
Drafted a PR https://github.com/trezor/trezor-firmware/pull/2753

grdddj:
> @grdddj Apologies I believe the assertion is incorrect. I'll fix that and move the test fixture to the last element of that list. What do you mean by "recording the fixtures".

Thanks for correcting the asserted address in the PR.

I made another PR - https://github.com/trezor/trezor-firmware/pull/2754 - with that change and also with updating our UI test-suite. When the CI turns green, we will merge it into `master`.

That is what I meant by "recording the fixtures" - new test-case was added, but the screen recording of that test-case was not done, so the tests complained. We do that to have an overview of all screens during our testing, as when something fails later, it is a signal of something going wrong. Example of the failed test run with UI diff (from current `master`) is https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3625729086/artifacts/test_ui_report/index.html

matejcik:
i'm confused now.
@grdddj can you please integrate #2753 probably? so that #2754 is the only thing that needs to be merged and fixes all the problems

grdddj:
> i'm confused now. @grdddj can you please integrate #2753 probably? so that #2754 is the only thing that needs to be merged and fixes all the problems

#2754 already contains the changes from #2753. I could have took that commit, but it is in a remote fork, so I could not push anything on that branch anyway. #2753 could be closed if #2754 is merged

0xBEEFCAF3:
@matejcik Do you know roughly when these change are going to be deployed?

matejcik:
we're currently aiming for March release of T1 and April for TT

0xBEEFCAF3:
Thanks for Rolling out the T1 firmware. Everything looks great! Do you have a date set for the April release for TT? We are working on a release that is dependent on this, so if you have a target date, that would be very helpful. Thank you!

matejcik:
the release is scheduled for April 12 (early access) and April 19 (full public release).
be aware that if there are delays, we will push the dates back by a full month

0xBEEFCAF3:
@matejcik Great! Thank you for the update

Hannsek:
We won't release the fw in early access phase. Full public release is still valid.

0xBEEFCAF3:
Thanks for the update!

Issue #2686, opened 2022-12-15
Delete unused components from micropython UI - Popup, Loader, Pageable...


grdddj:
`core/src/trezor/ui/loader.py` and `core/src/trezor/ui/popup.py` contain UI components that are not used anymore (because of moving UI to `Rust`). Also `core/src/trezor/ui/components/common/confirm.py` has unused `Pageable` class.

Still having them in codebase is just increasing the binary size.

grdddj:
Turns out the `Loader` is still being used via `LoadingAnimation` in https://github.com/trezor/trezor-firmware/blob/550e2a79512b75f09e57878baaa52bb5310aefbe/core/src/apps/management/reset_device/__init__.py#L26-L28

@mmilata Did we miss this when moving into `Rust` loaders, or is this intentional?

mmilata:
You're right, I have missed it. Will move it to rust side.

grdddj:
Thanks, I have tagged you (and this issue) in https://github.com/trezor/trezor-firmware/pull/2688, as after we remove `Loader`, there will be more removing to be done

PR #2688, opened 2022-12-15
feat(core): delete unused code, mostly from micropython UI

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2686:
- deletes code that is currently unused (and IMO will not be used in the future) - mostly on the UI side

Next steps:
- [x] Look at the `Loader` being used as `LoadingAnimation` in https://github.com/trezor/trezor-firmware/blob/550e2a79512b75f09e57878baaa52bb5310aefbe/core/src/apps/management/reset_device/__init__.py#L26-L28
- [x] Once `Loader` is not there (@mmilata), delete some dependent things from `core/src/trezor/ui/style.py`
- [x] Delete whole `core/src/trezor/res` (also dependent on `Loader`, as it uses `trezor/res/check.toif`)
- [x] There might be some more unused functions in `core/src/trezor/ui/__init__.py` once we completely move to `UI2`


core/src/trezor/ui/components/common/confirm.py
```diff
@@ -1,11 +1,9 @@
 from typing import TYPE_CHECKING
```
matejcik, line 1: can't we remove this whole file? I was under the distinct impression that I removed it [here](https://github.com/trezor/trezor-firmware/pull/2556) but apparently not...?

grdddj: We certainly can, after the deletion, it contains just 
```
CONFIRMED
CANCELLED
INFO
def is_confirmed(x: Any) -> bool:
```
which each module could do on its own. We will get rid of one nested directory in a file structure

matejcik: which is the case for tt_v2 already anyway

grdddj: Done in [4bc7990](https://github.com/trezor/trezor-firmware/pull/2688/commits/4bc79909321a8d407551645b22f1043266507bbe)



Turned out there are two places that used it:
```
core/src/apps/bitcoin/sign_tx/approvers.py:215: - error: Import "trezor.ui.components.common.confirm" could not be resolved (reportMissingImports)

core/src/trezor/ui/layouts/tt_v2/__init__.py:549: - error: "confirm" is unknown import symbol (reportGeneralTypeIssues)
```

I solved it (little dirty) by sending a boolean saying if it was confirmed (True) of info was requested (False)

grdddj:
[608cf9d](https://github.com/trezor/trezor-firmware/pull/2688/commits/608cf9dcfdff0af2bab404996e48eef052ec6ea7) adds the possibility to run `make unused` in `core` to find the unused symbols in `core` `micropython` files.

`vulture` was added as a dependency. Looking at it now, it was automatically appended at the end of `pyproject.toml` under `# storage` header, which already contains some "non-storage" items like `inotify` or `yamllint`. We should probably put them in some other categories, like `# test`.

`poetry.lock` got regenerated because of that.

To silence the "non-problems", there is a `src/_vulture_ignore.txt` file to contain these symbols that appear to be unused, but we do not care (inspired by https://github.com/jendrikseipp/vulture). We also do not care about `messages.py` and `enums`, so excluding them from the search

core/src/storage/cache.py
```diff
@@ -220,10 +220,6 @@ def end_current_session() -> None:
     _active_session_idx = None
 
 
-def is_session_started() -> bool:
```
mmilata, line 223: Seems to be maybe used in tests/test_storage.cache.py?

mmilata (commented):
LGTM! I don't know why is there the loader that does not really do anything, just make you wait, maybe up for discussion with @Hannsek whether we want to keep it?

Also, what about deleting the toif counterparts in `core/assets`? Unfortunately the UI2 assests are also there so you can't just drop the directory.

core/src/storage/cache.py
grdddj: Right, I have defined it in a test-file in [05517d7](https://github.com/trezor/trezor-firmware/pull/2688/commits/05517d7683ce391682336bcc96ca6cf5397bd7f5)

grdddj:
> Also, what about deleting the toif counterparts in `core/assets`? Unfortunately the UI2 assests are also there so you can't just drop the directory.

In [7493f0d](https://github.com/trezor/trezor-firmware/pull/2688/commits/7493f0d5d2efcb6799e4ed635bee7c948c161966) I have deleted the old unused icons

mmilata (commented):
Let's also delete `NORMAL`, `BOLD`, `MONO`, `description_param_font` please?

mmilata:
Sorry, I suck at grepping, it actually is still used in couple of places.

matejcik:
it should be enough to render the "zero frame" of the loader. something with "draw_simple" or similar?

mmilata:
Is it this one https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/common/mnemonic.py#L95? Maybe there could be a way to initialize it before calling into C?

matejcik:
i believe it's actually this one https://github.com/trezor/trezor-firmware/blob/master/core/src/trezor/pin.py
and yes, we could do something like import it from `trezor.pin` (make it not private) and explicitly render it?

mmilata:
Right. The first render happens in pin_progress() constructor so that should be possible, just need to make sure it's called everytime before pin-unlock happens.

grdddj:
Thanks for the suggestions. Something along those lines done in [43013d9](https://github.com/trezor/trezor-firmware/pull/2688/commits/43013d9fa31d43e64de83554d2c64a296f7032ad)

It is surely not perfect, the "2 seconds left" is there for some time without the loader moving, and then with "1 second left" it starts moving quickly. However, this is something most regular users will see only once during their whole Trezor life-time.

grdddj:
It generated some UI diff - mostly one more recorded screen https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3645925550/artifacts/test_ui_report/index.html

However, there is one strange case where it caused a UI problem (drawing loader over PIN entry):

https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3645925550/artifacts/test_ui_report/failed/TT_test_msg_change_wipe_code_t2.py::test_set_remove_wipe_code.html

matejcik:
> It is surely not perfect, the "2 seconds left" is there for some time without the loader moving, and then with "1 second left" it starts moving quickly. However, this is something most regular users will see only once during their whole Trezor life-time.

ISTM not showing "2 seconds left" is the right thing to do? so there's "PROCESSING" that hangs for a bit and then the progress starts along with the timer.

for that matter, how long is the delay and which call is causing it?

core/src/trezor/pin.py
```diff
@@ -35,9 +49,12 @@ def show_pin_timeout(seconds: int, progress: int, message: str) -> bool:
     else:
         remaining = _previous_remaining
 
-    if progress == 0 or _progress_layout is None:
+    # create the layout if it doesn't exist yet or should be started again
+    if _progress_layout is None or (progress == 0 and not _started_with_empty_loader):
         _progress_layout = pin_progress(message, description=remaining or "")
+
```
matejcik, line 55: my guess is adding `_started_with_empty_loader = False` here will get rid of the rendering artifact

grdddj:
> 



> > It is surely not perfect, the "2 seconds left" is there for some time without the loader moving, and then with "1 second left" it starts moving quickly. However, this is something most regular users will see only once during their whole Trezor life-time.
> 
> ISTM not showing "2 seconds left" is the right thing to do? so there's "PROCESSING" that hangs for a bit and then the progress starts along with the timer.
> 
> for that matter, how long is the delay and which call is causing it?

Right, I will get rid of the "2 seconds left" text.

The delay is around half a second on the hardware device. However, debugging it on emulator, I could not find out which call is responsible for the delay. The second, "proper" loader is being called by `storage.reset()` -> `config.wipe()`. Really the only code between `render_empty_loader` and  `config.wipe()` are the imports in `storage::wipe` - `from trezor import config` and `from storage import cache` - so my best guess is that the delay is caused by these imports.

core/src/trezor/pin.py
grdddj: Right, adding that in [11c5e09](https://github.com/trezor/trezor-firmware/pull/2688/commits/11c5e0950446c20481b58853cd8b9c56150afcde) helped

matejcik: it should be outside the `if` though. it actually surprised me that this worked; apparently it is because the `if seconds == 0` clear the `_progress_layout`, but technically speaking that condition _may_ not happen at all, in which case you'd get the artifact the next time someone ""restarts"" after not finishing the previous loader.

grdddj: Moved outside of `if` and included a comment in a force-push together with rebase on `master`

matejcik:
there's a massive UI diff from the added "please enter passphrase" screen.

@Hannsek you were saying that we want to completely get rid of that screen, is that correct? if yes, this would be a very nice place to fold in that change.

Hannsek:
I think we want, but we need to discuss this with @sime.

matejcik (approved):
OK, for now let's go ahead with this as-is.

I didn't do a detailed review of the UI diff, just random samples, so hoping that everything is alright on that side .) if you didn't do that yet, @grdddj maybe locally remove the passphrase screen and see what differences are there without it?

otherwise please fix the conflict in poetry.lock and merge

grdddj:
> OK, for now let's go ahead with this as-is.
> 
> I didn't do a detailed review of the UI diff, just random samples, so hoping that everything is alright on that side .) if you didn't do that yet, @grdddj maybe locally remove the passphrase screen and see what differences are there without it?
> 
> otherwise please fix the conflict in poetry.lock and merge

I am confirming all the changes are screen additions due to two things:
- adding the passphrase dialogue
- drawing the empty loader with "PROCESSING"

matejcik (approved):

Issue #2616, opened 2022-11-07
Bootloader is built with wrong hashes and signature


hiviah:
Revision: 4e800b6ffde97862026c6196f7d983d62a05ea71

Current output at end of `make build_bootloader` build (or also `make build_embed` when it comes to bootloader part) :

```
Detected image type: bootloader
Firmware Header {
    magic: HeaderType.BOOTLOADER
    header_len: 1024
    expiry: 0
    code_length: 104448
    version: 2.0.5.0
    fix_version: 2.0.0.0
    hashes: [
        ❌ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
        ✔ 0000000000000000000000000000000000000000000000000000000000000000
    ]
    v1_signatures: [
        64 bytes 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
        64 bytes 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
        64 bytes 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
    ]
    v1_key_indexes: [0, 0, 0]
    sigmask: 3
    signature: 64 bytes 59d63653aae4c1ba23d35fa46106688331ce61f4c6f37492ea6410c2938765ea8e80ffb19a50e913eb5dc5fb9a53fac7332bbfc2097fb10c3ff6b0c983f24104
}
[...]

PR #2689, opened 2022-12-15
Headertool fixes

matejcik:
should fix both problems from #2616
@TychoVrahe please confirm

### DIFF
--- a/python/src/trezorlib/_internal/firmware_headers.py
+++ b/python/src/trezorlib/_internal/firmware_headers.py
@@ -53,12 +53,18 @@ def all_zero(data: bytes) -> bool:
     return all(b == 0 for b in data)
 
 
-def _check_signature_any(fw: "SignableImageProto", is_devel: bool) -> Status:
+def _check_signature_any(fw: "SignableImageProto", is_devel: bool = False) -> Status:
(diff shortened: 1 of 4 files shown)


python/src/trezorlib/_internal/firmware_headers.py
```diff
@@ -53,12 +53,19 @@ def all_zero(data: bytes) -> bool:
     return all(b == 0 for b in data)
 
 
-def _check_signature_any(fw: "SignableImageProto", is_devel: bool) -> Status:
+def _check_signature_any(fw: "SignableImageProto", is_devel: bool = False) -> Status:
+    print("checking signature of:", fw)
```
TychoVrahe, line 57: This prints the whole image, including the (binary) code, please limit the scope of printing or remove entirely

TychoVrahe (changes requested):
Works fine, just one change request

python/src/trezorlib/_internal/firmware_headers.py
matejcik: yeah this is a debug statement that shouldn't be there :( e9365f

TychoVrahe (approved):

PR #2699, opened 2022-12-19
fix UI test report generation

TychoVrahe:
Fixes a problem with report generation, where `data-actual-hash` was not filled in when generation was invoked via `make test_emu_ui`, which causes that links to next test are not properly generated. Generation via `python -m tests.ui_tests` worked fine.

Also fixes a bug where last failed tests failed to update fixture from report, as the window was closed before the promise was resolved and fixtures updated.


matejcik:
seemingly superseded by #2778

PR #2707, opened 2023-01-02
chore(build): use host network instead of the default (bridge)

prusnak:
It seems that bridge is not a good default because it does not always work out of the box.

see https://docs.docker.com/network/ for more info

edit: also [podman](https://podman.io) is a thing, so make it possible to override the docker command


matejcik (approved):
utACK

PR #2708, opened 2023-01-03
Introduce Toif / Icon / Image type

### LINKED ISSUE #2512: Rust UI: introduce Icon type
The code especially in #2471 is getting a little messy in places because we don't have a convenient `ui::Icon` type with `size()` and `pub data: &'static [u8]` and similar methods, which need to be laboriously explicitly written out in all usage sites.

Let's not block the above PR for that, but we should introduce `Icon` and factor out these things everywhere.

I believe there's some work on that in #2277 already?

TychoVrahe:
Introduces Icon and Image types, and more general Toif type.

Did this without `Copy`, but its a bit annoying at some places, so i left it as a fixup, see the difference with and without here: https://github.com/trezor/trezor-firmware/commit/cb2f10035829d7c82612a5132f336f89349dda43 

closes #2512


core/embed/rust/src/ui/util.rs
```diff
@@ -64,14 +65,13 @@ pub fn set_animation_disabled(_disabled: bool) {}
...
-    icon: &'static [u8],
+    icon: &Icon,
     space: i16,
     text: &str,
     style: TextStyle,
     text_offset: Offset,
 ) {
-    let toif_info = unwrap!(display::toif_info(icon), "Invalid TOIF data");
-    let icon_width = toif_info.0.y;
+    let icon_width = icon.width();
```
TychoVrahe, line 74: hmm there is a large UI diff because i incidentally fixed this bug affecting homescreen notification positioning
download1 download2 diff

TychoVrahe: updated the fixtures: https://github.com/trezor/trezor-firmware/pull/2708/commits/da5e028c772dd06e8e041ba4953ec5f0ccf844c5

TychoVrahe:
Rebased after the homescreen merge.

using the `Copy` now, as it was quite annoying already.

grdddj (commented):
Nice changes, the `Alignment2D` helps a lot!

I have got some suggestions about refactoring, mostly the `Toif`, `Icon` and `Image` structs.

Device tests are red because of the `master` branch, but it also seems the tests for `../tests/device_tests/test_msg_applysettings.py` are not recorded in this branch

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+                unwrap!(ctx.uncompress(&mut dest), "Decompression failed");
+            }
+        }
+    }
+
+    pixeldata_dirty();
+}
+
+/// Holding toif data and allowing it to draw itself.
+/// Lots of draw methods exist so that we can easily
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/cae8d31d36bf3e439d3447956f19d02c043f20ae

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+    // Text is useful for debugging purposes.
+    text: &'static str,
+    size: Offset,
+    format: ToifFormat,
+}
+
+impl Toif {
+    pub fn new(named_toif: NamedToif) -> Self {
+        let info = toif_info(named_toif.0);
+        let info = unwrap!(info);
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/cae8d31d36bf3e439d3447956f19d02c043f20ae

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+        Self {
+            data: named_toif.0[12..].as_ref(),
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/cae8d31d36bf3e439d3447956f19d02c043f20ae

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// Holding toif data and allowing it to draw itself.
+/// Lots of draw methods exist so that we can easily
+/// "glue" the toif together with other elements
+/// (text, display boundary, etc.) according to their position.
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub struct Toif {
+    data: &'static [u8],
+    // Text is useful for debugging purposes.
+    text: &'static str,
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/cae8d31d36bf3e439d3447956f19d02c043f20ae

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+        let toif = Toif::new(named_toif);
+        Self::check(&toif);
+        Self {
+            toif,
+            area: Rect::zero(),
+            precise_x: false,
+        }
+    }
+
+    pub fn from_slice(data: &'static [u8]) -> Self {
```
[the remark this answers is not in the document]
TychoVrahe: ultimately removed

core/embed/rust/src/ui/component/image.rs
```diff
@@ -1,24 +1,13 @@
...
-    display::{toif_info, Color},
-    geometry::{Alignment, Offset, Point, Rect},
+    display::{
+        toif::{Icon, Image},
+        Color,
+    },
+    geometry::{Offset, Point, Rect, CENTER},
 };
 
-pub struct Image {
```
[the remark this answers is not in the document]
TychoVrahe: moved https://github.com/trezor/trezor-firmware/commit/0a6698989084d2890f049603cabcb2083e4528ee

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+                bg_color.into(),
+            );
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub struct Image {
+    toif: Toif,
+    pub area: Rect,
```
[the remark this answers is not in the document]
TychoVrahe: moved https://github.com/trezor/trezor-firmware/commit/0a6698989084d2890f049603cabcb2083e4528ee

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+use heapless::String;
+
+use super::Color;
+
+/// Storing the icon together with its name
+/// Needs to be a tuple-struct, so it can be made `const`
+#[derive(Debug, Clone, Copy)]
+pub struct NamedToif(pub &'static [u8], pub &'static str);
+
+impl NamedToif {
```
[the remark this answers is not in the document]
TychoVrahe: done https://github.com/trezor/trezor-firmware/commit/4e3e114ce2d1b6124c79d540ea1de4cae5572739

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+        let toif = Toif::new(named_toif);
+        Self::check(&toif);
+        Self {
+            toif,
+            area: Rect::zero(),
+            precise_x: false,
+        }
+    }
+
+    pub fn from_slice(data: &'static [u8]) -> Self {
```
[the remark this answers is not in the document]
TychoVrahe: removed the from_slice: https://github.com/trezor/trezor-firmware/commit/1be1ae7093aae2729def9f98662dd195789726bf

ad loader, i am aware of this: when we merge the rust bootloader and remove the remaining loader calls from python we can delete this

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+            precise_x: false,
+        }
+    }
+
+    pub fn with_precise_x(mut self) -> Self {
+        self.precise_x = true;
+        self
+    }
+
+    fn check(toif: &Toif) {
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/1be1ae7093aae2729def9f98662dd195789726bf

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+            info.format,
+        ))
+    } else {
+        None
+    }
+}
+
+/// Aborts if the TOIF file does not have the correct grayscale flag, do not use
+/// with user-supplied inputs.
+pub fn toif_info_ensure(data: &[u8], format: ToifFormat) -> (Offset, &[u8]) {
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/b2c0be170cd9d0747fc15e2558dbec5969bd5a8c

matejcik:
high level note: it looks like the distinction between `Icon` and `Image` exists mostly to type-check that the data in the toif is the correct color depth, right? IOW, you're passing `Icon` or `Image` into a lower level drawing function -- which then looks at dimensions and raw Toif data, but doesn't care about the outer struct at all. technically you should be passing in the `Toif` struct directly.
the drawback of course is that you'd have to place the asserts on the color depth in code, whereas this way it won't even compile if we mix it up.
...question is, is the compiler even doing anything useful here? If I take a colored toif data and make an `Icon` instance out of it, I still get a panic, just in a different place.

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    fn size(&self) -> Offset {
+        self.toif().size
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub struct Icon {
+    toif: Toif,
+    area: Rect,
```
matejcik, line 138: area is unused for `Icon` because it's not a component

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub fn format(&self) -> ToifFormat {
+        self.format
+    }
+
+    pub fn name(&self) -> &'static str {
+        self.name
+    }
+}
+
+pub trait ToifAccessors {
```
matejcik, line 111: there's actually a very straightforward way to get rid of this: make the `toif` member public in both `Image` and `Icon` and use it directly `img.toif.size` instead of `img.size()`

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    /// Display the icon with baseline Point, aligned according to the
+    /// `alignment` argument.
+    pub fn draw(&self, baseline: Point, alignment: Alignment2D, fg_color: Color, bg_color: Color) {
+        let r = Rect::snap(baseline, self.size(), alignment);
+
+        if self.precise_x {
+            icon_precise(self, r.center(), fg_color, bg_color);
+        } else {
+            icon(
```
matejcik, line 166: is there actually any advantage of keeping this? could we always use `icon_precise` (and thus get rid of `precise_x` and redefine `Icon` as `struct Icon(Toif)` ?)

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -0,0 +1,276 @@
...
+                self.toif.data,
+                fg_color.into(),
+                bg_color.into(),
+            );
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub struct Image {
```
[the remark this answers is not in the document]
grdddj: `ToifAccessors` is nice, I don't know how to improve it further

core/embed/rust/src/ui/display/mod.rs
```diff
@@ -325,17 +227,14 @@ pub fn rect_rounded2_partial(
...
-        if toif_size.x <= MAX_ICON_SIZE && toif_size.y <= MAX_ICON_SIZE {
-            icon_area = Rect::from_center_and_size(center, toif_size);
+    if let Some((icon, icon_color)) = icon {
+        if icon.width() <= MAX_ICON_SIZE && icon.height() <= MAX_ICON_SIZE {
+            icon_area = Rect::from_center_and_size(center, icon.size());
             icon_area_clamped = icon_area.clamp(constant::screen());
-
-            let mut ctx = UzlibContext::new(toif_data, None);
+            let mut ctx = UzlibContext::new(icon.data(), None);
             unwrap!(ctx.uncompress(&mut icon_data), "Decompression failed");
```
matejcik, line 235: this could be a `fn uncompress(&self, buffer)` on `Toif`.
(and have it just straight out panic if it fails, I see no value in passing out an error only to have to unwrap! it one level up)

might actually be a tiny bit nicer to do `fn decompression_context(&self, window) -> UzlibContext` on Toif too, although the only thing that's achieving is hiding the raw `data`

core/embed/rust/src/ui/display/toif.rs
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/f05d93194cb77e02f35a87353a3614cd85fb1dbe

core/embed/rust/src/ui/display/toif.rs
TychoVrahe: done https://github.com/trezor/trezor-firmware/commit/f05d93194cb77e02f35a87353a3614cd85fb1dbe

core/embed/rust/src/ui/display/toif.rs
TychoVrahe: there are performance benefits of using the C functions (it does less boundaries checking and when DMA2D is enabled it uses that). 

though i think that we could easily live with more boundaries checking and i could implement the DMA2D variant in rust, and when bootloaders get merged we could drop the C functions

core/embed/rust/src/ui/display/mod.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/bc174f7ed04e27f751b67a4e52dd4985ae773c12

TychoVrahe:
rebased so we can see the CI & updated the remaining fixtures

ad the high level note: 
i think there is benefit in panicking earlier in the flow - i.e. in the later functions the icon might be displayed under some conditions, which could make the program not panic when testing and hide the bug for some time.

core/embed/rust/src/ui/display/toif.rs
TychoVrahe: implemented: https://github.com/trezor/trezor-firmware/commit/69294044a1e7b7239a80dad1e9045d7d561826c7

dma2d has its own limits (mainly can't render odd number of pixels) so in relatively rare situations this is needed, this defaults to non-dma2d rendering.

this happens when part of the icon is out of screen, and rendering starts (or ends) on odd coordinate -> this leads to odd number of pixels in line not shown, therefore odd number of total pixels is rendered since the grayscale toif is always even-wide.

TychoVrahe: removed the `precise_x` https://github.com/trezor/trezor-firmware/commit/8f4ec36f6c8163aa2ca9cb295a51992ecf334404 - didn't do the `struct Icon(Toif)` change yet - what's the benefit there? getting the toif via icon.0 seems less readable to me

removed the dma2d function and left only the precise rust rendering functiton: https://github.com/trezor/trezor-firmware/commit/d227728be3f4e7202784f12e4515afb0fd076223

there is large UI diff again https://github.com/trezor/trezor-firmware/commit/aeb475117b6ba3ae452713f8e312a8449e28e067 because making all icons render on precise location shifted 
- notification icon again
- scrollbar
- cancel button icon in pin keyboard
- checklist icon

scroll scroll2 scroll_diff

cancel cancel_after diff


checklist checklist1 checklist_diff

matejcik: > didn't do the struct Icon(Toif) change yet - what's the benefit there? getting the toif via icon.0 seems less readable to me

not much, it just makes it a little clearer that the `Icon` is just a newtype for a `Toif`. I'm ok with keeping the item named.

TychoVrahe:
anything else here?

grdddj (approved):

mmilata:
git bisect tells me that this PR causes #2779. I suspect it is because the passphrase keyboard struct has 40 Button instances, each of which has 2 Toif instances, each of which takes 17B+, and the keyboard layout constructor overflows stack.

Ideas:
* drop `name` from struct Toif
* use `&'static Icon` in the `ButtonContent::IconBlend` variant, or possibly more places
* make passphrase keyboard not use 40 buttons
* construct layouts directly on heap using experimental rust features (small structs would still desirable tho)

TychoVrahe:
- break the `ButtonContent` enum into `Component`s /or some other trait/ and make the `Button` generic over that

mmilata:
> break the ButtonContent enum into Components /or some other trait/ and make the Button generic over that 

FWIW this can potentially result in some of the methods being duplicated for each instance, increasing flash space usage, unless you jump some hoops to use `dyn` like in `Paragraphs`

PR #2698, opened 2022-12-19
Fix model T FMC timing

TychoVrahe:
Fixes settings of FMC - timing was too fast for the display driver, occasionally could cause packet loss.

PR #2714, opened 2023-01-04
fix model T display reinitialization crash

TychoVrahe:
Unfortunately the FMC timing fix #2698  introduced a bug that crashes firmware startup, because the FMC initialization accesses area of memory that is under MPU protection at that point.


mmilata (approved):
Fixes the crash, not sure if the change has any other effect.

PR #2715, opened 2023-01-04
tearing effect prevention improvement

### LINKED ISSUE #2546: Tearing effect prevention on TT is not working properly
Current implementation of tearing effect prevention on TT is not correct: it synchronizes mcu with the LCD after drawing is done via `display_refresh` function, but there is no guarantee when the next drawing cycle will begin, therefore, it doesn't prevent tearing when sending data into display in the middle of panel refresh.

One solution is to introduce `display_sync` function, which should be called before drawing, which would wait until panel refresh completes (TE signal goes HI). The `display_refresh` should wait probably function as a guarantee that everything that was sent to display is shown on the LDC, therefore should wait for TE to go LO and then HI. 

This is still not bulletproof, as our drawing makes little guarantees about how much time it will take, but it should make nice improvement.

The `display_sync` could theoretically help to eliminating some otherwise hard to solve flickering, like button changing color when pressed, if we can send data to display fast enough. 

Note that the cycle is about 13ms long, so neither refresh or sync should be overused.

TychoVrahe:
Introduces `display_sync` function that synchronizes writing to display RAM with panel refresh. This is called before every relevant paint so that optimally the paint is done before the corresponding panel part is refreshed.

`display_orientation` function is modified so that the panel refresh direction respects the orientation. This is only possible for north and south orientation with current display driver. 

For west and east orientation, the display refresh direction is perpendicular to the orientation so the tearing effect prevention is not working and the effect is actually diagonal, but at least the display is synchronized so the UI should behave less randomly in this regard.

closes #2546 , builds on #2714

### DIFF
--- a/core/embed/extmod/modtrezorui/display_interface.h
+++ b/core/embed/extmod/modtrezorui/display_interface.h
@@ -43,6 +43,7 @@ int display_backlight(int val);
 
 void display_init(void);
 void display_reinit(void);
+void display_sync(void);
 void display_refresh(void);
 const char *display_save(const char *prefix);
 void display_clear_save(void);
--- a/core/embed/rust/build.rs
+++ b/core/embed/rust/build.rs
@@ -293,6 +293,7 @@ fn generate_trezorhal_bindings() {
         .allowlist_function("display_pixeldata")
         .allowlist_function("display_pixeldata_dirty")
         .allowlist_function("display_set_window")
+        .allowlist_function("display_sync")
         .allowlist_var("DISPLAY_CMD_ADDRESS")
         .allowlist_var("DISPLAY_DATA_ADDRESS")
         .allowlist_type("toif_format_t")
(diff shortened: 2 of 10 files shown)


core/embed/extmod/modtrezorui/modtrezorui-display.h
```diff
@@ -62,6 +62,17 @@ STATIC mp_obj_t mod_trezorui_Display_clear(mp_obj_t self) {
 STATIC MP_DEFINE_CONST_FUN_OBJ_1(mod_trezorui_Display_clear_obj,
                                  mod_trezorui_Display_clear);
 
+/// def sync(self) -> None:
```
matejcik, line 65: Let's not expose it to Python at all and instead call it from Rust. `LayoutObj::paint_if_requested` seems like the right place.

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/aaf6e751fc1de967251054c09e8dfbdb9d379eff

TychoVrahe:
yeah it needs a rebase, homescreen image is not yet in this branch

PR #2726, opened 2023-01-09
integrate marquee into model R pages

TychoVrahe:
This adds marquess into titles, if the title doesn't fit.

Marquee is cherry picked from #2654 .

Wasn't happy about how scrollbar was handled, so i changed that a bit - i think this is more in line with how components should work and the base doesn't need the additional function, which was kinda non-systematic. Did this as separate commit for cleaner idea what changed (and easier revert if we decide against it): https://github.com/trezor/trezor-firmware/commit/f5900401b1c4a55b556f93dddbebe418fdbdd8ab


grdddj:
Awesome! Working fine on my emulator.

One unfortunate thing I noticed so far is that the animation triggers screenshot creation after each refresh, so it can create hundreds of screenshots during a user-interacted session (UI-tests are fine here).

Currently looking at the code changes and UI-diff. I found already one place where the header is missing:

image

core/embed/rust/src/ui/model_tr/component/title.rs
```diff
@@ -0,0 +1,102 @@
...
+    type Msg = Never;
+
+    fn place(&mut self, bounds: Rect) -> Rect {
+        self.area = bounds;
+        self.marquee.place(bounds);
+        bounds
+    }
+
+    fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
+        if !self.marquee.is_animating() {
```
grdddj, line 75: Seem these marquee events will be triggered even in case the title fits entirely.

We could do the calculation we do every time in `paint`
(`self.needs_marque = width > self.area.width()`) already in `place` and not do the `event` at all in case it is `false`

core/embed/rust/src/ui/model_tr/component/flow.rs
```diff
@@ -148,12 +150,16 @@ where
     fn event_consumed_by_current_choice(&mut self, ctx: &mut EventCtx, pos: ButtonPos) -> bool {
         if matches!(pos, ButtonPos::Left) && self.current_page.has_prev_page() {
             self.current_page.go_to_prev_page();
-            self.scrollbar.inner_mut().go_to_previous_page();
+            self.scrollbar
+                .inner_mut()
+                .set_active_page(self.page_counter as usize, ctx);
```
grdddj, line 157: I think this needs to be reverted, because in case there is "more pages withing a single page", the `self.page_counter` in this higher component is always the same, but the page has changed within `self.current_page` - see the screenshots in overall comment

grdddj (commented):
Couple of things to change, some of those might be resolved by the recent commit, I will try these again. You do not have to change all of these, I think I can handle those

Missing title on second screen and even on first screen when going back
`INTERACT=1 pytest ../tests/device_tests/bitcoin/test_bcash.py -k wrong`
image

Not updating the scrollbar when scrolling through the long content
image

show_warning()
missing scrollbar
image

show_share_words()
missing scrollbar
image

TychoVrahe:
fixed the integration in Flow here: https://github.com/trezor/trezor-firmware/pull/2726/commits/8aafdc16f0c666da627fb370385fd30692cd6c01

also adjusted the title position when its not marquee to be on top instead of one blank pixel line (to match design).

Previously i adjusted the scrollbar position to be vertically centered, but looking at design now, it seems ambiguous - somewhere the scrollbar is top aligned (e.g. confirm output), and somewhere its centered (confirm transaction). not sure whats right

core/embed/rust/src/ui/model_tr/component/title.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/pull/2726/commits/8043226ae84d4b6cb1720dc52b52960660f34fa0

but had to start with `true` otherwise it misses the first event which comes before the paint and does not start the animation. not sure it if doesn't defeat the purpose though

not sure this is the right way to start the animation either

TychoVrahe:
About the missing scrollbars in `show_warning` and `show_share_words`:

The `show_warning` is using confirm_action, the ScrollableFrame should be used even when no Title is provided, needs a bit of adjustment to enable empty title there. 

`show_share_words`: since the scrollbar was removed from ButtonPage, it should be added to ShareWords component and shown on first page. 

Wanna do this or should i?


(missing title and not updating scrollbar should be fixed by above fixups)

core/embed/rust/src/ui/model_tr/component/flow.rs
grdddj: Also a possibility, I agree with this.

I little preferred the scrollbar to have the option to paginate on its own - as there would be no need to calculate anything from outside, just tell it "go_to_previous_page" without any arguments

core/embed/rust/src/ui/model_tr/component/title.rs
grdddj: That is unfortunate, let's leave it like this for now, if it works - we can go back to it in further reviews with more people

grdddj:
> Wanna do this or should i?

Feel free to do that :) Thanks for the help

core/embed/rust/src/ui/model_tr/component/title.rs
grdddj: On a second thought - even if "first event which comes before the paint", we should be setting `self.needs_marque` in `place()` - which should probably be called before `event`. In `paint()`, we do not have to recompute it again

TychoVrahe: thats good idea, done here https://github.com/trezor/trezor-firmware/commit/314eb86056fae2fff1eb043b05fcd16c7c6bf17d

TychoVrahe:
show warning handled here: https://github.com/trezor/trezor-firmware/commit/38514c1c3273bd884d60d302c482b4f21fdfa18e
share words: https://github.com/trezor/trezor-firmware/pull/2726/commits/853e0d45d9281361bfd60a84777ee0e12f91e557

grdddj:
Included in https://github.com/trezor/trezor-firmware/pull/2610, closing this

Thanks for the help!

PR #2731, opened 2023-01-10
Differentiate build results for different models

### LINKED ISSUE #2351: Prepare a system to embed bootloader version according to model type in core
Currently we only have a bootloader binary for model T in Core source. We will need to design a way to do it for additional models. Obviously a solution like bootloader_T.bin and bootloader_R.bin is possible, but maybe there might be a better way to do it? something that will not bloat the repo over time with binaries, like pulling the bootloader binary from CI during build or something?

### LINKED ISSUE #2697: Differentiate build results for different models
Closely related to #2351

The build output of trezor-core should clearly indicate, via filename, which version and model is being built.
Presumably this could also help out QA when downloading CI builds, by locally naming the files something sensible instead of "trezor.bin".

This is then directly transferable to `build_docker.sh` that is supposed to build all variants. More work in basic build = less work for the outer script.
We need a naming convention anyway, for the purpose of deploying the files for Suite / Connect.

Simplest draft idea:
`T2B1-2.6.0.bin`
`T2T1-2.6.0-btconly.bin`
i.e., `<internal model>-<version>.bin`
for convenience of commands like `make upload`, we can always place a symlink `trezor.bin` to the latest build

TychoVrahe:
Adds a named binary file with added information about variant, model and version. Original files are left available too.

the filename is structured this way:
`<binary type>-<internal model>-<variant, if any>-<version>.bin`
i.e.
`firmware-T2T1-btconly-2.5.4.bin`
`bootloader-T2T1-2.0.5.bin`

Also a system enabling different embedded bootloaders according to model is implemented.

closes #2351 
closes #2697


matejcik:
please also expose these files in CI builds (both `firmware.bin` and `firmware-whatever.bin` should be exported as artifacts) and `build_docker.sh` (we can expose both, or just the one with the specifier)

matejcik:
looking at the CI artifacts, should we (a) call the default variant "regular", or (b) drop the "regular" symbol from the CI artifact?
also maybe put "debug" into the local names?

TychoVrahe:
there is also 'production' and commit hash in CI, not sure the name like
`firmware-T2T1-regular-production-debug-2.5.4-COMMITHASH.bin`
isn't overdoing it?

for `build-docker.sh`, i don't think there is anything to do for `core`. `legacy` copies the files explicitly so it needs adjustment, should i handle it here or leave it for https://github.com/trezor/trezor-firmware/issues/2732 ?

matejcik:
I believe the consensus was that we want
* model
* nothing or `btconly`
* version
* commit hash

in addition, CI is deploying things into the same directory and needs different file names for debug and non/debug variants.

I'm OK with either adding debug to our generated filename, or to keep this one difference between the build and CI

TychoVrahe:
- added the commit hash, and also added `-dirty` flag to local builds if there is some modification since the commit https://github.com/trezor/trezor-firmware/pull/2731/commits/b1f1ef8ef5a5cd9df6c44fccd3c90695ac45fbe8
- dropped the `regular` flag from CI https://github.com/trezor/trezor-firmware/pull/2731/commits/cbf53af169a4a90623e6ffbccf20b4e64a6621d4

PR #2733, opened 2023-01-11
Update blockchain_link.json

claudiovb:
Update the syscoin URL, our previous blockexplorer at sys1.bcfn.ca isn't operational any longer, making the trezor connection to syscoin fail.


matejcik (approved):
utACK

Issue #2728, opened 2023-01-10
Use lower case in the name `coinjoin`


hynek-jina:
Replace word `CoinJoin` with `coinjoin` wherever it appears.

Same applies if we use coinjoin as a verb or noun.

bosomt:
**QA OK**

 - Device: model T 2.5.4 Universal (revision ab65b18b07170c8a9e4f1506620af988d678453b)

bosomt:
**QA NOK**

https://github.com/trezor/trezor-firmware/blob/ab65b18b07170c8a9e4f1506620af988d678453b/core/src/trezor/ui/layouts/tt_v2/homescreen.py

<img width="366" alt="image" src="https://user-images.githubusercontent.com/31506317/214034126-7322cafc-2613-4645-a39b-67c205c83487.png">

**Info:**
 - Suite version: desktop 23.2.0 (47ff6c1b0068967d941c67675a9b6cc012464fb7)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.2.0 Chrome/104.0.5112.124 Electron/20.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.4 Universal (revision ab65b18b07170c8a9e4f1506620af988d678453b)
 - Transport: bridge 2.0.32

hynek-jina:
@bosomt This doesn't look like a mistake. It's a sentence... so it starts with capital `C`, right? Or what is the problem?

bosomt:
@hynek-jina not for me :)
+ dialogue headers are all caps so it says ` "AUTHORIZE COINJOIN",`
if its OK feel free to close the issue

https://github.com/trezor/trezor-firmware/blob/1359b21508c7efd1038beb921dfc3516a4a07967/core/embed/rust/src/ui/model_tt/layout.rs#L938

hynek-jina:
Yes, I think it is intended.

PR #2735, opened 2023-01-12
Use lowercase in the name coinjoin.

andrewkozlik:
Fixes https://github.com/trezor/trezor-firmware/issues/2728.


core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -981,16 +981,6 @@ async def confirm_coinjoin(
     )
 
 
-def show_coinjoin() -> None:
```
matejcik, line 984: why is this gone?

mmilata: This is handled by Busyscreen which directly calls the rust layout.

That said, the deleted code looks like:
deleted

And the busyscreen currently renders like this:
hs

The former looks nicer, perhaps I should modify the homescreen code to look like it?

mmilata (approved):

core/src/trezor/ui/layouts/tt_v2/__init__.py
andrewkozlik: `show_coinjoin()` is currently not used anywhere. The reference to it was removed in https://github.com/trezor/trezor-firmware/commit/5b3db7eca19eab2723f168563594b7eaacd4ef75, see the move from `core/src/apps/homescreen/busyscreen.py` to `core/src/trezor/ui/layouts/tt_v2/homescreen.py`. However, I also like the deleted screen better. I am wondering, should we add "Please wait " to it, or would it be too much?

mmilata: Like this?
cj

andrewkozlik: Hmm, not sure, this way it slightly overshadows the main message that we want to communicate, which is "Do not disconnect your Trezor". I was thinking of the "Please wait" as a title. But I am not too familiar with the new design, so I don't want to break consistency. I leave it to @hynek-jina.

mmilata: Dunno if I'm understanding this right:
cj2

Or did you mean to add third block of text above the icon?

andrewkozlik: I meant it above the icon (which is what might break consistency with the new design). We can also do it in reversed order: bold "Do not disconnect your Trezor." and then under that normal font "Coinjoin in progress. Please wait." But as I said, I leave it to Hynek.

mmilata: I can do this one right now as it uses existing layout (with lowercase w):
cj3

We currently have no layout with title-icon-text1-text2 so I'd prefer if you created issue for me to create one. I think @Hannsek is now in charge of UI2 product-wise instead of Hynek?

hynek-jina: Coinjoin screens for UI2 are here: https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?node-id=5752%3A43361&t=TWzAsAeUH4v2iwf8-4

Here is the discussed screen if I'm not mistaken.
Rounds

**So do we need another issue? Or what is the next step, please?**

_btw. it's little bit confusing with the responsibilities.. @Hannsek is taking care about whole UI2 and I'm responsible for coinjoin... so we have to cooperate. 🙂 Final word with trezor UI is on Hans_

mmilata: I wasn't aware of this screen, yes, please open an issue.

hynek-jina: Screens are not yet final. So ignore it for now.. @Hannsek will open related issues once is it ready for development.

matejcik:
OK, please fix the conflicts and make this ready to merge

matejcik:
all CI failures are expected

Issue #2440, opened 2022-08-06
T2B1: unite the Rust API with redesigned model T


grdddj:
It would be nice if the API between `micropython` and `Rust` was completely the same for `model T` and `model R`.

In that way we could reuse (literally reuse or at least copy-paste) almost all the calling and receiving code on both the `micropython` (`trezor/ui/layouts/<model>/__init__.py`) and `Rust` (`ui/<model>/layout.rs`) side - the only differences would be the `UI` handling on the `Rust` side.

It would mean uniting the strategies mostly for `string` handling - in the `model T` redesign, we tend to mostly pre-process them in `micropython` to avoid working with them in `Rust`, where we only show them. On the other hand, in `model R`, we are mostly processing them in `Rust` - which has its own issues.

As `model T` redesign will be "finished" sooner than `model R`, I am (from the `R` perspective) ready to "copy" anything that we will finalize for `model T` so that both models are as similar as possible.

matejcik:
outstanding issue:
`confirm_output_r` is a single flow that allows going back and forth between recipient & amount, no buttonrequest in the middle

`confirm_output` on T2T1 is two `confirm_value` calls each of which sends a buttonrequest.

not sure which way to unify here, an open question is also whether we should allow going back on TT too

matejcik:
same thing is true for `confirm_total` where the two screens are total amount + fee

matejcik:
T2B1:
* expose `confirm_value` to Python and delegate to `confirm_action` in Rust
* delegate `confirm_joint_total` on Rust side to `confirm_properties`, dtto for `confirm_modify_fee`
* `confirm_modify_output`: delegate to `confirm_properties` (not confirm_action) on Rust side
* `confirm_with_info` and `confirm_more` are missing designs, we will see if we need to implement them separately for T2B1 or delegate to something
* `confirm_coinjoin` can delegate to `confirm_properties` for now
* `confirm_reset_device` - delegate
* `confirm_recovery` - delegate
* `show_remaining_shares` -- only called from info button on `confirm_recovery` so never happens on T2B1. for now let's raise NotImplementedError from Python as it is now, in the future we can raise the exception from Rust if we need to.

matejcik:
`show_group_share_success` -- for now delegate to `confirm_action` in Rust, but maybe worth considering merging with `show_popup` on the T2T1 side?

grdddj:
> T2B1:
> 
[...]

PR #2741, opened 2023-01-12
feat(core): make first steps to unify Rust API for both models

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/2440
- model `T` changes we made in the office


matejcik (approved):

matejcik:
two changes that the UI tests uncovered:

1. "Enter PIN" has changed to "Enter old PIN" when changing PIN. that seems correct, let's record that into fixtures
2. the confirmation popup has changed for SLIP39 backup, e.g. here https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3595376778/artifacts/test_ui_report/failed/TT_reset_recovery-test_reset_slip39_basic.py::test_reset_device_slip39_basic_256.html

that's obviously a bug in the refactor around `_share_words_confirmed`

core/src/apps/management/reset_device/layout.py
```diff
@@ -184,11 +208,8 @@ async def slip39_basic_show_and_confirm_shares(
             await show_share_words(ctx, share_words, index)
 
             # make the user confirm words from the share
-            if await _confirm_share_words(ctx, index, share_words):
-                await _show_confirmation_success(ctx, index, len(shares))
+            if await _share_words_confirmed(ctx, index, share_words):
```
matejcik, line 211: looks like this should be `_share_words_confirmed(ctx, index, share_words, len(shares))` ?

grdddj: Correct, changed that in [2775364](https://github.com/trezor/trezor-firmware/pull/2741/commits/27753645e6ba1e4f925f28baeb25e78717936ffd) - together with recording the fixtures

matejcik (approved):

PR #2744, opened 2023-01-13
Device test features

grdddj:
Some device tests usability improvements done in `model R` branch, but wanting to move them to `master`, as `model R` branch will not be there any time soon.

1. Possibility to accept the UI diff (change hashes in `fixtures.json`) after running tests with `--ui=test`, without having to call them again with `--ui=record`
- by `make test_emu_accept_fixtures`
- the quickest workflow is then `make test_emu_ui_multicore` to generate the UI diff as quickly as possible, and immediately update those changes by abovementioned command

2. Quickening the tests in CI by not generating `index.html` after each test case
- by the usage of `--not-generate-report-after-each-test` `pytest` flag

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -351,6 +351,13 @@ def pytest_addoption(parser: "Parser") -> None:
         help="Which emulator to use: 'core' or 'legacy'. "
         "Only valid in connection with `--control-emulators`",
     )
+    parser.addoption(
+        "--not-generate-report-after-each-test",
+        action="store_true",
+        default=False,
+        help="Not generating HTML reports after each test case. "
+        "Useful for CI tests to speed them up.",
+    )
 
 
 def pytest_configure(config: "Config") -> None:
@@ -399,8 +406,11 @@ def pytest_runtest_teardown(item: pytest.Item) -> None:
     # Not calling `testreport.generate_reports()` not to generate
     # the `all_screens` report, as would take a lot of time.
     # That will be generated in `pytest_sessionfinish`.
-    if item.session.config.getoption("ui"):
-        testreport.index()
+
+    # Optionally generating `index.html` report unless turned off by `pytest` flag
+    if not item.session.config.getoption("not_generate_report_after_each_test"):
+        if item.session.config.getoption("ui"):
+            testreport.index()
 
 
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -249,6 +249,8 @@ def pytest_sessionstart(session: pytest.Session) -> None:
     ui_tests.read_fixtures()
     if session.config.getoption("ui") and _is_main_runner(session):
         testreport.clear_dir()
+        # Preparing a new empty file for UI diff
+        ui_tests.FIXTURES_DIFF.write_bytes(b"")
 
 
 def _should_write_ui_report(exitstatus: pytest.ExitCode) -> bool:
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -249,6 +249,9 @@ def pytest_sessionstart(session: pytest.Session) -> None:
     ui_tests.read_fixtures()
     if session.config.getoption("ui") and _is_main_runner(session):
         testreport.clear_dir()
+        # Preparing a new file for UI diff
+        with open(ui_tests.FIXTURES_DIFF, "w") as _:
+            pass
```
matejcik, line 255:
```suggestion
ui_tests.FIXTURES_DIFF.touch()
```

tests/ui_tests/__init__.py
```diff
@@ -18,6 +18,7 @@
 SCREENS_DIR = UI_TESTS_DIR / "screens"
 HASH_FILE = UI_TESTS_DIR / "fixtures.json"
 SUGGESTION_FILE = UI_TESTS_DIR / "fixtures.suggestion.json"
+FIXTURES_DIFF = UI_TESTS_DIR / "fixtures.diff.json"
```
matejcik, line 21: looks like the file is not valid json but one-json-per-line right? in that case let's call it `fixtures.json.diff` so that the suffix matches

tests/ui_tests/__init__.py
```diff
@@ -186,6 +195,31 @@ def write_fixtures_suggestion(
     )
 
 
+def update_fixtures_with_diff() -> int:
```
matejcik, line 198: could we use this to fix `test_emu_ui_record_multicore` somehow? (to recall: the issue is that the test records write over one another)

(although that would probably require more changes, i.e., saving a fixtures diff per runner and then finishing it up in the coordinator.)

tests/conftest.py
grdddj: That is unfortunately not possible here, as we want the file to be always empty at the beginning - not containing the old diff

tests/ui_tests/__init__.py
grdddj: Yes, they are individual JSON objects on each line.

Done like this to be "multicore-friendly", so that the file is append-only and there is no need for re-writing the whole file when adding a new record (which could cause some race-conditions when two processes/threads would do it simultaneously). 

Name changed in [c4d30db](https://github.com/trezor/trezor-firmware/pull/2744/commits/c4d30dbd1c5618f7dab668c584c819826c8dd230)

tests/ui_tests/__init__.py
grdddj: It really seems the issues with `test_emu_ui_record_multicore` are harder to solve.

We can define `test_emu_ui_record_multicore` = `test_emu_ui_multicore` + `test_emu_accept_fixtures` - I tried it in [c4d30db](https://github.com/trezor/trezor-firmware/pull/2744/commits/c4d30dbd1c5618f7dab668c584c819826c8dd230) and it works fine (we only need to catch the error code from the first command not to fail the whole "pipeline")

tests/conftest.py
matejcik: in that case `ui_tests.FIXTURES_DIFF.write_bytes(b"")`

tests/ui_tests/__init__.py
matejcik: looks fine, it's still technically possible for two writes to coincide but in practice it's super unlikely that two calls to `write()` with data this short would interleave at the kernel level

a small drawback is that this causes the write even if the tests fail, but eh :woman_shrugging:

tests/conftest.py
grdddj: Done like this in force-push

matejcik (approved):
utACK

Issue #2711, opened 2023-01-04
`trezorctl fw verify`: show fingerprints first and signature status afterwards


matejcik:
so that you can see information about unsigned firmware

bosomt:
**QA OK**

```
 ~/Downloads/trezorfwasuite/FWofiko  trezorctl fw verify -c trezor-2.5.3.bin                                                                    ✔  trezor-firmware 
Trezor T firmware image.
Vendor header from SatoshiLabs, version 0.0
Firmware version 2.5.3 build 0
Firmware fingerprint: 4f57dca1abc1a60d82c4fef7c96e86d784fc7a1e5e3da724dd2ae4d14c6350bf
Signatures are valid.
Firmware is appropriate for your device.
```

PR #2745, opened 2023-01-13
feat(python): show fingerprint before signatures in `trezorctl firmware verify`

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2711:
- showing fingerprint even when the signatures are invalid


grdddj:
> utACK, please add a changelog entry

Added in the force-push

matejcik (approved):
please merge when CI is done

Issue #1909, opened 2021-04-25
Remove EIP-712 single field size restriction


bonustrack:
Trezor is not triggered when trying to sign a message on snapshot.org, it seem to happen when the length of the message is too long. Is there a length limit of the message that can be signed with Trezor? If so, is it possible to increase that limit?

matejcik:
maximum length for `EthereumSignMessage` message is 1024 bytes and it is unlikely to be increased.

please provide more details, maybe steps to reproduce?

volkyeth:
> maximum length for `EthereumSignMessage` message is 1024 bytes and it is unlikely to be increased.
@matejcik could you explain why it's unlikely to be increased?

Some platforms are using arbitrarily long signed messages for publishing verified content in the format of posts. Snapshot is one of such.

I'm trying to sign an EIP-712 message, which is 7237 characters long, and Trezor Connect just blinks and closes up without displaying any error, presumably because this is a firmware limitation.

Can that limit be increased?
Is there any workaround for this?

matejcik:
EIP-712 is not affected by this limit. Instead, there is a soft limit of ~7 kB per a single text / bytes field. Which is unlikely to be increased either because there is no convincing reason to support 7-kilobyte text fields.

Signing Snapshot.org proposals generally works fine for me, and it seems that Trezor is only signing a hash of the proposal anyway. But you can try to link a proposal that is problematic and I can try to reproduce the problem.

More likely though, the issue is somewhere else. What method do you use to connect Trezor to Snapshot?

volkyeth:
My current example for this is actually https://prop.house
Sorry, but you won't be able to reproduce without spamming their platform 😅

They sign the whole content of your proposal. So, the message can be arbitrarily long.

Is there any drawback to allowing the Trezor to sign any messages as long as they fit in memory? I think UX-wise it would have to be a blind signing like for EIP-712 since it would have to paginate through hundreds of pages for long messages

Anyway, I see that this is a narrow edge case, so it's unlikely to be implemented

matejcik:
> Is there any drawback to allowing the Trezor to sign any messages as long as they fit in memory?

The "as long as they fit in memory" is the tricky part. Without going into too much detail, it's not as simple as feeding more and more of the message until it doesn't fit.

matejcik:
> Is there any drawback to allowing the Trezor to sign any messages as long as they fit in memory?

[...]

PR #2746, opened 2023-01-13
feat(core/ethereum): remove EIP-712 field size limitation of 1024 bytes

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/1909:
- removing the limitation of `EIP-712` message's fields to be smaller than 1024 bytes

### DIFF
--- a/core/src/apps/ethereum/sign_typed_data.py
+++ b/core/src/apps/ethereum/sign_typed_data.py
@@ -1,4 +1,3 @@
-from micropython import const
 from typing import TYPE_CHECKING
 
 from trezor.enums import EthereumDataType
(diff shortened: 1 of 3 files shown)


core/src/apps/ethereum/sign_typed_data.py
```diff
@@ -426,14 +421,12 @@ def _validate_value(field: EthereumFieldType, value: bytes) -> None:
...
-            raise DataError("Invalid length")
-    else:
-        if len(value) > _MAX_VALUE_BYTE_SIZE:
-            raise DataError(f"Invalid length, bigger than {_MAX_VALUE_BYTE_SIZE}")
+    # Checking if the size corresponds to what is defined in types.
+    # Not having any maximum field size - it is a responsibility of the client
+    # (and message creator) to make sure the data is not too large to cause problems
+    # on the Trezor side.
+    if field.size is not None and len(value) != field.size:
+        raise DataError("Invalid length")
```
matejcik, line 429: just one question, do we have a testcase that triggers this DataError?

grdddj: Haven't found one in `device_tests/ethereum`, but we will see for sure after CI finishes

matejcik (approved):
utACK, please merge if CI is good

Issue #2722, opened 2023-01-07
Centralize all common testing input flows


grdddj:
In case of bip39/slip39 reset/recovery, there seems to be a lot of duplication in `input_flow` functions.

When we are introducing `model R`, all these input flows need to be defined again and just slightly modified (somewhere more than slightly).

It seems like a nice opportunity to put them all in a separate file, it could be more documented and would decrease duplication.

Also, currently, we are using this lengthy pattern to get the appropriate flow.

```python
        if client.features.model == "T":
            client.set_input_flow(input_flow_tt)
        elif client.features.model == "R":
            client.debug.watch_layout(True)
            client.set_input_flow(input_flow_tr)
```

It could be then replaced by something like `client.set_input_flow(slip_39_recovery_input_flow(client.features.model))`

PR #2749, opened 2023-01-16
chore(tests): centralize all model-dependent input flows

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2722:
- defining all testing `input flows` at one place, providing some benefits
   - they can be easily refactored (as a lot of them are similar - mostly reset/recovery ones)
   - it is one place where all the model-specific differences could be well assessed/compared
   - device tests don't have to care about the model, they just request it for a specific case (e.g. bip39 dry run) and supply it with `client` and some optional data - and after the function runs, it can read its internal variables (e.g. with gathered mnemonic)

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -15,6 +15,7 @@
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
 import json
+import time
 from pathlib import Path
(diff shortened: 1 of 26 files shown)


matejcik:
will be resolved in #2803

PR #2751, opened 2023-01-18
Allow coinjoin request debugging key for testnets

andrewkozlik:
This PR allows the production firmware to accept coinjoin requests that are signed using the testing coinjoin request key if the transaction is for a testing network (SLIP-44 ID = 1), i.e. Testnet or Regtest.

I'd prefer not to make this change in the firmware, but the alternatives are worse:
- Use the production private key for signing coinjoin requests on testnet. This would increase the exposure of the production private key by making the server which holds it accessible to multiple clients. We should minimize the exposure of the key, which is easier if the only client is the production coordinator.
- Force testers, i.e. not just devs, but also people from within the company and possibly from outside the company who participate in the testing on Testnet, to use either a special custom firmware or the debug firmware. The debug firmware is completely impractical, because it erases the seed after every startup. The custom firmware is also impractical, because its installation and distribution would pose an obstacle to many testers, e.g. it also erases the seed, albeit only after the first installation.

### DIFF
--- a/core/src/apps/bitcoin/keychain.py
+++ b/core/src/apps/bitcoin/keychain.py
@@ -88,10 +88,10 @@ PATTERN_UNCHAINED_DEPRECATED = "m/45'/coin_type'/account'/[0-1000000]/address_in
 PATTERN_SLIP26_T1_FW = "m/10026'/49'/2'/0'"
 
 # SLIP-44 coin type for Bitcoin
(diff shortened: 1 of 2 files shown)


matejcik (approved):
utACK

matejcik:
test failures look related to #2682 so ignoring those

PR #2752, opened 2023-01-18
Add debugging tool for hashing operations

andrewkozlik:
This PR adds a simple tool to help debugging issues which involve incorrect digests being computed. The wrapper allows the developer to log what data the Trezor is hashing when computing the digest.

I don't know whether `utils.py` is the best place for this. I put it there, because `HashContext` and `HashWriter` are defined there. Possible alternatives:
1. Split off these hashing-related classes, including the new wrapper, into a separate file.
2. Introduce a file with debugging tools. There is `apps/debug/__init__.py`, but I think that's for handling debuglink messages.

### DIFF
--- a/core/src/trezor/utils.py
+++ b/core/src/trezor/utils.py
@@ -169,6 +169,42 @@ if TYPE_CHECKING:
             ...
 
 
+if False:
+
+    class DebugHashContextWrapper:
+        """
+        Use this wrapper to debug hashing operations. When digest() is called,
+        it will log all of the data that was provided to update().
+
+        Example usage:
+        self.h_prevouts = HashWriter(DebugHashContextWrapper(sha256()))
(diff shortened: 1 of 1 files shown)


core/src/trezor/utils.py
```diff
@@ -169,6 +169,42 @@ def extend(self, __buf: bytes) -> None:
             ...
 
 
+if __debug__:
```
matejcik, line 172: Make this `if False` please. We don't want dead code lying around, even in debug builds (which are larger just by themselves)

andrewkozlik: Rebased over master and force-pushed.

matejcik (approved):

matejcik:
will merge after CI goes green

PR #2753, opened 2023-01-18
fix(tests): Fix order of ethereum signmessage tests

0xBEEFCAF3:
[no changelog]

Both the assertion was incorrect and the order in list


matejcik:
superseded by #2754

PR #2754, opened 2023-01-19
fix(tests): correct ETH signmessage assertion and record the testcase

grdddj:
Fixes what was missing in https://github.com/trezor/trezor-firmware/pull/2682


matejcik:
it turns out that this was missing from the legacy codebase too, so I added it.
@andrewkozlik a quick review of a824b9c if you could?

andrewkozlik:
> it turns out that this was missing from the legacy codebase too, so I added it. @andrewkozlik a quick review of [a824b9c](https://github.com/trezor/trezor-firmware/commit/a824b9c8433f5409936d8e7f986211d9779bfe25) if you could?

LGTM

matejcik (approved):

PR #2757, opened 2023-01-19
fix(python): include private implementation of ripemd160

matejcik:
because recent openssl has dropped it and so it is gone from hashlib

### DIFF
--- a/python/src/trezorlib/tools.py
+++ b/python/src/trezorlib/tools.py
@@ -35,6 +35,8 @@ from typing import (
 
 import construct
 
+from ._ripemd160 import ripemd160
+
 if TYPE_CHECKING:
     from .client import TrezorClient
     from .protobuf import MessageType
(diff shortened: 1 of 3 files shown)


grdddj (approved):
Seems alright, if the CI test pass

python/src/trezorlib/tools.py
```diff
@@ -76,9 +78,7 @@ def tx_hash(data: bytes) -> bytes:
 
 
 def hash_160(public_key: bytes) -> bytes:
-    md = hashlib.new("ripemd160")
-    md.update(hashlib.sha256(public_key).digest())
-    return md.digest()
+    return ripemd160(hashlib.sha256(public_key).digest())
```
prusnak, line 81: Btw, the OpenSSL upstream reverted this stupid decision and re-introduced RIPEMD160 back to OpenSSL, so we might do it conditionally, but I am not sure it is worth the hassle.

```suggestion
    try:
        md = hashlib.new("ripemd160")
        md.update(hashlib.sha256(public_key).digest())
        return md.digest()
    except ValueError:
        from ._ripemd160 import ripemd160
        return ripemd160(hashlib.sha256(public_key).digest())
```

matejcik: Yeah, not worth it. This isn't really _used_ anywhere, i am in fact considering removing the failing function from trezorlib. I pushed this change to fix some failing tests, but it might be better to move the functionality into the test suite.

Issue #2639, opened 2022-11-14
Drop Trezor 1 UI from trezor-core


matejcik:
There is no reason to keep it because T1 and TR can use the same UI code.

matejcik:
I didn't notice that there's still `model_t1` on the Rust side of things. that should be dropped too.

grdddj:
It could be useful here - I have a commit for deleting `T1` completely in my `TR` branch - https://github.com/trezor/trezor-firmware/pull/2610/commits/e891734feba0e4dae8bd07d76eff150658f6d62b

PR #2761, opened 2023-01-20
chore(core/rust): delete model_1 code - it should use model_r

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2639:
- deleting `model_t1` from `Rust`


matejcik:
T1 builds in CI are broken, obviously.

I think we can remove them completely for now.

grdddj:
> T1 builds in CI are broken, obviously.
> 
> I think we can remove them completely for now.

They are broken because there is no `model_t1` feature in `Rust`. If we just supply `model_tr` in that case, it should build fine.

https://github.com/trezor/trezor-firmware/blob/e15a0ba1d31e891eb194d12506b3bf2b59fed8dc/core/SConscript.firmware#L718

Something like 
```
"model_tr" if TREZOR_MODEL == "1 else f'model_t{TREZOR_MODEL.lower()}'
```

matejcik:
yeah, that's how we'd fix the failure.
but
we don't really need the "T1 emulator build" because there is nothing different from TR there. it might be useful to keep the T1 firmware build around, but given the low priority of that, I believe we don't need to do that either, it's eating precious spots in the CI pipeline

grdddj:
> yeah, that's how we'd fix the failure. but we don't really need the "T1 emulator build" because there is nothing different from TR there. it might be useful to keep the T1 firmware build around, but given the low priority of that, I believe we don't need to do that either, it's eating precious spots in the CI pipeline

Sure, I will delete both of these from CI, but will do that one-line change in both `SConscripts`, so nobody is surprised later when it does not work.

On the note of deleting, I also made a commit of deleting all mentions of `model 1` in all `SConscripts` and almost started deleting all `#if defined TREZOR_MODEL_1` blocks from `core/embed` - but theoretically we may want to leave it there for future

matejcik:
> Sure, I will delete both of these from CI, but will do that one-line change in both `SConscripts`, so nobody is surprised later when it does not work.

:+1: 



> On the note of deleting, I also made a commit of deleting all mentions of `model 1` in all `SConscripts` and almost started deleting all `#if defined TREZOR_MODEL_1` blocks from `core/embed` - but theoretically we may want to leave it there for future

yes, we want to keep those. the intent was only to delete code that is definitely unused, i.e., implementations of T1-specific UI

grdddj:
Can we merge this? The approval came before some of the latest changes

Issue #2585, opened 2022-10-26
Usage of newlines in firmware strings


grdddj:
As a "relict" from `UI1`, there are quite a lot of places where we are inserting newlines `\n` into a to-be-displayed string to visually separate some parts of the message.

Example, where it makes sense, is `"The mining fee of\n{}\nis unexpectedly high."` However, mostly for longer sentences, it seems to be not so useful.

Also, considering that `TR` will be using the same strings, we may probably delete majority of these.

### Current:
```python       
content = "The following\ntransaction output\ncontains tokens."
```
![image](https://user-images.githubusercontent.com/42543243/198048260-2cd23071-2c59-4a01-83d6-fc139d86533c.png)

### Suggested:
```python
content = "The following transaction output contains tokens."
```
![image](https://user-images.githubusercontent.com/42543243/198048283-d489e7c3-d827-4f2a-be8d-27a512bfd725.png)

grdddj:
> Rather than a specific task, couldn't we do this as we go along?
> 
> i.e. Whenever you touch a particular layout, clean up the newline characters?

Also a possibility, but we can quite easily find out all the newlines in strings at once and decide what to do with them. (For example, delete all of them in one commit and see the `UI` diff.)

In regards to `TR`, it would be nice not having to do it there, as these changes would retrospectively change the `TT`design. (In the current `TR` branch, I am deleting the newlines in a special function, which is not very good - we want to probably keep some of the newlines even for `TR`.)

mmilata:
UI2 works around this in couple of places like
```
title=content.replace("\n", " ")
```
we should get rid of these after we switch.

PR #2763, opened 2023-01-20
Grdddj/usage of newlines in strings

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2585:
- deleting newlines that do not delimit some important data on the screen
- did not record the fixtures yet, so we can see the differing screens and discuss where to keep them and where not (cherry-picked and squashed commits that make it possible) - available at https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3636318493/artifacts/test_ui_report/differing_screens.html
- argument for deleting them everywhere is that some newlines could be beneficial on `TT`, but not so much on `TR` and vice-versa
- some way of "hard-space" (non-breakable space) as suggested by @TychoVrahe could be used


mmilata (approved):
8137eaf88597e244fe77d3b7df13ee34fc5f8c51 LGTM
I slightly prefer the "Continue with the next share." layout in the original version but it's a small difference.

mmilata:
Link for the lazy https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3636318493/artifacts/test_ui_report/differing_screens.html

grdddj:
> [8137eaf](https://github.com/trezor/trezor-firmware/commit/8137eaf88597e244fe77d3b7df13ee34fc5f8c51) LGTM I slightly prefer the "Continue with the next share." layout in the original version but it's a small difference.

I prefer it visually as well, but that way we would need to maintain two versions - one with newlines for TT and one without it (or on different places) for TR. We could do that if we want to, let's now unify it first.

There is CI failure in `unix ui changes` - created issue https://github.com/trezor/trezor-firmware/issues/2799.

That failure is not critical, merging anyway.

PR #2764, opened 2023-01-24
refactor(core): improve safety of buffer handling

### LINKED ISSUE #2648: improve safety of buffer handling

TychoVrahe:
This aims to improve safety of buffer handling in rust.

On the C side, i introduce tracking of allocated buffers. When attempting to get already allocated buffer, rust panics. Buffers need to be freed after use.

This prevents using the same buffer multiple times, but still is not safe because there is no mechanism implemented that would prevent using the buffer after freeing it.

One idea to further improve this would be to introduce wrappers around the buffers, only access the pointer from within the wrapper and free the buffer on `Drop`. Something like this is done in `BlurringContext`, but for the rest of the buffers this looks like a bit annoying thing to do, given how the buffers and subslices from them are used, also not sure how this would work with `dma2d_start_blend` which needs to get the pointer to the buffer anyway.

Furthermore, i have marked `dma2d_start_blend` as `unsafe` too, since the mcu uses the buffers until the transfer is complete, so care must be taken not to modify it.

closes #2648


mmilata:
I like this very much, even `#define BUFFER` is nice as far as C macros go;) Still I'm wondering if we can get rid of most of the C logic entirely? `display_icon` only seems to be used in the bootloader and `display_image` is used for the obsolete homescreens? Can we afford to just ignore these cases?

TychoVrahe:
In current master, `display_icon` is only used in `bootloader_ci`.

`display_image` is used in bootloader for displaying vendor image, which doesn't work now anyway due to the format change.

`display_image` is also used for FIDO icons. Its not used for the old homescreens.

Buffers themselves are also used in C functions `display_text_render_buffer` and `display_bar_radius_buffer`, which are used from rust

What i think is easily doable and worthwhile is 
- simplify C version of display icon to not use DMA2D
- rewrite DMA2D `display_image` to rust, use it there, and either - keep the C version for non-DMA2D usage (used in bootloader), or rewrite it too
- leave buffers.c in C for now

Before any additional work here i need to rebase this as its quite old

TychoVrahe:
@mmilata since your approval i added two more commits fixing an issue that the buffers were dropped while dma2d transfer was ongoing. I also added more safety comments to the `dma2d_start` functions, please have a look

mmilata (approved):

PR #2771, opened 2023-01-25
fix(core/bootloader): fix detection of installed firmware

TychoVrahe:
fixes a bug recently introduced, causing installed firmware always evaluated as new.


matejcik (approved):
utACK

PR #2773, opened 2023-01-26
chore(vendor): bump micropython to 1.19.1

### LINKED ISSUE #2341: Update Micropython to 1.19
https://github.com/micropython/micropython/releases/tag/v1.19

Probably the most interesting change is the redesigned bytecode which translates into 5% size reduction of frozen .mpy code

mmilata:
Current MicroPython patches: https://github.com/micropython/micropython/compare/v1.19.1...trezor:trezor-v1.19.1
Upstreamed micropython/micropython#10588, micropython/micropython#10589.
No changes in Rust bindings needed. Fixes #2341.

### DIFF
--- a/core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
+++ b/core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
@@ -253,17 +253,21 @@ void dump_dict_inner(FILE *out, const mp_obj_dict_t *dict) {
 
 void dump_function(FILE *out, const mp_obj_fun_bc_t *func) {
   print_type(out, "function", NULL, func, false);
-  fprintf(out, ",\n\"globals\": \"%p\"", func->globals);
   fprintf(out, ",\n\"code_alloc\": %ld", find_allocated_size(func->bytecode));
   fprintf(out, ",\n\"code_ptr\": \"%p\"", func->bytecode);
-  fprintf(out, ",\n\"const_table_alloc\": %ld",
-          find_allocated_size(func->const_table));
-  fprintf(out, ",\n\"const_table_ptr\": \"%p\"", func->const_table);
   mark(func->bytecode);
-  mark(func->const_table);
+  fprintf(out, ",\n\"context_ptr\": \"%p\"", func->context);
+  fprintf(out, ",\n\"qstr_table_ptr\": \"%p\"",
+          func->context->constants.qstr_table);
+  fprintf(out, ",\n\"qstr_table_alloc\": \"%ld\"",
+          find_allocated_size(func->context->constants.qstr_table));
+  // obj_table uses the same allocation as qstr_table
   fprintf(out, "},\n");
 
-  dump_value(out, func->globals);
+  // mp_module_context_t has mp_obj_module_t value at the beginning,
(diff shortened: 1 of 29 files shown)


prusnak:
Great! Do we have some basic benchmarks?

1. what's the firmware size before and after the change?
2. how long does the sign_tx test take before and after the change?

mmilata:
The upgrade saves about 15KiB of FLASH2 (833->818). However signing transactions now takes longer:

| version                     | 1.17       | 1.19       |
|-----------------------------|------------|-------------|
| test_lots_of_inputs | 12m10s| 12m51s |
| test_sign_tx_large   | 1m22s  | 1m45s   |

mmilata:
Force-pushed changes suggested in micropython/micropython#10588:
* dropped trezor/micropython@4ac33a573
* added dummy implementation of `soft_timer_gc_mark_all`
* removed `vendor/micropython/ports/stm32/softtimer.c` from build

core/embed/extmod/modtrezorutils/modtrezorutils-meminfo.h
```diff
@@ -253,17 +253,13 @@ void dump_dict_inner(FILE *out, const mp_obj_dict_t *dict) {
 
 void dump_function(FILE *out, const mp_obj_fun_bc_t *func) {
   print_type(out, "function", NULL, func, false);
-  fprintf(out, ",\n\"globals\": \"%p\"", func->globals);
+  fprintf(out, ",\n\"globals\": \"%p\"", func->context->module.globals);
```
matejcik, line 256: what kind of thing is `context`? if it's something that the meminfo can understand, we should be dumping the `context` instead of going through to globals

mmilata: Right, [context](https://github.com/micropython/micropython/blob/v1.19.1/py/bc.h#L209-L213) is module + constants. We can dump the module, [constants](https://github.com/micropython/micropython/blob/v1.19.1/py/bc.h#L199-L207) are in ROM I think and therefore not interesting for us?
c29a1a88a7fd312f750274e8934e299150125fe7

mmilata: Like 34af8044ed1b01da0f6a14490723c1e5307bdbc7? We could also pass `func->context` to `dump_value` because it has `mp_obj_module_t` at the beginning but this seems more explicit.

matejcik: i think we should _also_ pass the context to `dump_value`, and add support for the context type.
because this way you get the pointer but not content of the pointee -- in the rendered memory dump, you will not see how functions hold a reference to their module -- just a link to an opaque pointer, and then then module separately

matejcik (approved):

mmilata:
Gitlab CI failures are probably because of poetry cache. @vdovhanych can you please clear it? Just for this branch is fine, if that's even possible.

vdovhanych:
> Gitlab CI failures are probably because of poetry cache. @vdovhanych can you please clear it? Just for this branch is fine, if that's even possible.

Cleared

PR #2776, opened 2023-01-30
automatic generation of CMakeLists

TychoVrahe:
Generates `CMakeLists.txt` automatically during compilation, respecting what is currently being build (eg. firmware/ bootloader/emulator) and other settings like model, bitcoin only etc. This enables IDE to correctly parse the C source code

@hiviah since you probably use this the most, let me know if this is useful to you and doesn't break your workflow somehow.

### DIFF
--- a/core/.gitignore
+++ b/core/.gitignore
@@ -8,3 +8,5 @@ tests/trezor_monero_tests*
 .coverage.*
 htmlcov/
 mypy_report
+/CMakeLists.txt
+/cmake-build-debug/
--- a/core/Makefile
+++ b/core/Makefile
@@ -24,6 +24,7 @@ BITCOIN_ONLY ?= 0
 TREZOR_MODEL ?= T
 TREZOR_MEMPERF ?= 0
 ADDRESS_SANITIZER ?= 0
+CMAKELISTS ?= 0
 
 # OpenOCD interface default. Alternative: ftdi/olimex-arm-usb-tiny-h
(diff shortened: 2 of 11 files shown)


hiviah:
This is pretty cool.

Though could you add some parameter so that it won't always overwrite the `CMakeLists.txt` on each build, only when parameter is put on command line? Sometimes I change some of the definitions in `CMakeLists.txt` (usually when checking differences in code paths for `PRODUCTION=1`) or alternate between bootloader and firmware build?

TychoVrahe:
sounds reasonable: making this optional https://github.com/trezor/trezor-firmware/pull/2776/commits/3e5d1573bfbc4797d32e4dd01e175eee1ec94b9d

hiviah:
I think there is type confusion because Makefile sets `CMAKELISTS` to string `"0"` and not the integer `0` when retrieved via `ARGUMENTS.get('CMAKELISTS', 0)`.

Probably the simplest fix is to convert the value to int in SConscript which appears to make it work as intended (see patch - I zipped it because github wouldn't allow it as patch even though it's valid according its own rules)
[create_cmakelists.patch.zip](https://github.com/trezor/trezor-firmware/files/10546504/create_cmakelists.patch.zip)

TychoVrahe:
thanks, committed the patch: https://github.com/trezor/trezor-firmware/commit/cbe2fceb61352be540a89dc104cc4b9fa2049e94

hiviah (approved):

PR #2792, opened 2023-02-06
core: fix stack overflow when displaying passphrase keyboard

### LINKED ISSUE #2779: UI2: stack overflow in passphrase keyboard
**Describe the bug**
RSOD when assking for passphrase on device. Affects firmware but **not emulator**.

master

**To Reproduce**
`trezorctl set passphrase on && trezorctl btc get-public-node -n 'm/44h/0h/0h'`

mmilata:
Fixes #2779, related to #2708.

Size of `PassphraseKeyboard`:
* original: 4340 B
* after 24bcb74: 3668 B
* after 3487811: 2996 B
* after 6a4d470: 956 B

Not implemented:
* placement new (=construct object directly on micropython heap): seems tricky at the moment, would probably be feasible with global allocator
* further refactoring of `Button`: it's possible to reduce memory usage even further but the gains are not as big but with more effort

### DIFF
--- a/core/embed/rust/src/ui/display/loader.rs
+++ b/core/embed/rust/src/ui/display/loader.rs
@@ -43,10 +43,11 @@ pub fn loader_uncompress(
     const ICON_MAX_SIZE: i16 = constant::LOADER_ICON_MAX_SIZE;
 
     if let Some((icon, color)) = icon {
-        if icon.toif.width() <= ICON_MAX_SIZE && icon.toif.height() <= ICON_MAX_SIZE {
+        let toif_size = icon.toif.size();
+        if toif_size.x <= ICON_MAX_SIZE && toif_size.y <= ICON_MAX_SIZE {
             let mut icon_data = [0_u8; ((ICON_MAX_SIZE * ICON_MAX_SIZE) / 2) as usize];
             icon.toif.uncompress(&mut icon_data);
-            let i = Some((icon, color, icon.toif.size));
+            let i = Some((icon, color, toif_size));
             loader_rust(y_offset, fg_color, bg_color, progress, indeterminate, i);
         } else {
             loader_rust(y_offset, fg_color, bg_color, progress, indeterminate, None);
@@ -237,7 +238,7 @@ pub fn loader_rust(
                     let x_i = x_c - icon_area.x0;
                     let y_i = y_c - icon_area.y0;
 
-                    let data = unwrap!(icon_data).data()
+                    let data = unwrap!(icon_data).zdata()
                         [(((x_i & 0xFE) + (y_i * icon_width)) / 2) as usize];
                     if (x_i & 0x01) == 0 {
                         underlying_color = icon_colortable[(data & 0xF) as usize];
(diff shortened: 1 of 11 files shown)


core/embed/rust/src/ui/display/loader.rs
```diff
@@ -337,7 +338,7 @@ pub fn loader_rust(
 
             icon_buffer_used.buffer[icon_offset as usize..(icon_offset + icon_width / 2) as usize]
                 .copy_from_slice(
-                    &unwrap!(icon_data).toif.data[(y_i * (icon_width / 2)) as usize
+                    &unwrap!(icon_data).toif.zdata()[(y_i * (icon_width / 2)) as usize
```
mmilata, line 341: Don't we need to decompress the data here first? Are we even using loader with icon anywhere? Bootloader?

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -72,41 +67,46 @@ pub fn icon(icon: &Icon, center: Point, fg_color: Color, bg_color: Color) {
...
-            name: named_toif.1,
-            size: info.0,
-            format: info.1,
-        }
+    pub fn new(data: &'static [u8]) -> Self {
+        Self { data }
+    }
+
+    pub fn format(&self) -> ToifFormat {
+        let info = unwrap!(toif_info(self.data));
```
matejcik, line 79: unconvinced about those -- we're calling `toif_info` all over the place uselessly.

it seems slightly better to just implement `display_toif_info` in Rust here:

1. `new()` will assert the TOIF header
2. `format()` will match on the format character
3. `width()` and `height()` will just parse out the respective bytes
4. `size()` will combine `width()` + `height()`

all of those _look like_ they could be `const fn`s too, with the possible exception of `new()`?

matejcik (approved):
LGTM, like it a lot. 

ACK on the passphrase keyboard change and on removing `NamedToif`

OK with the `Toif` changes. I would like to see the changes I mentioned in the comments, but we can proceed like this if you think that would be better solved in a separate issue.

core/embed/rust/src/ui/display/loader.rs
TychoVrahe: yes there is a bug in here, the `loader_rust` is actually meant to receive already decompressed data, which allows to skip repeated decompression when updating progress. 

in bootloader the icons are used, so i made a fix there when i noticed this after rebase, but maybe you can cherry pick it here: https://github.com/trezor/trezor-firmware/commit/e8bfaaf0ce8931678427c9ef304bce3f83feda70

mmilata: Let's fix it in that PR if it's only used in bootloader.

core/embed/rust/src/ui/display/toif.rs
mmilata: Good idea, implemented in [9ecb7b0](https://github.com/trezor/trezor-firmware/pull/2792/commits/9ecb7b0e4bcd9437f6b9b763ff4a3238cad03e3d)

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -65,32 +50,45 @@ pub fn icon(icon: &Icon, center: Point, fg_color: Color, bg_color: Color) {
...
 #[derive(PartialEq, Eq, Clone, Copy)]
 pub struct Toif {
     data: &'static [u8],
 }
 
 impl Toif {
-    pub fn new(data: &'static [u8]) -> Self {
-        Self { data }
+    pub const fn new(data: &'static [u8]) -> Option<Self> {
+        if data.len() < TOIF_HEADER_LENGTH || data[0] != b'T' || data[1] != b'O' || data[2] != b'I'
```
matejcik, line 61: `&data[0..3] != b"TOI"` ?

mmilata: is not const

matejcik: oh, makes sense.

matejcik (approved):

Issue #2799, opened 2023-02-07
`unix ui changes` job in CI fails when there are some changes


grdddj:
`python -m tests.ui_tests.reporting master-diff` seems to be failing in case there are some UI changes or new tests - both for `device` and `click` tests.

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3726810931 from https://github.com/trezor/trezor-firmware/pull/2763
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3723274057 from https://github.com/trezor/trezor-firmware/pull/2610

matejcik:
what's weird is that there is a visible stack trace for a failed download, but then it goes on to download more...?

matejcik:
fix, hopefully, in above linked commit.

it looks like some fixtures were not deployed properly, so we'll need to watch this space and see if the problem magically fixes itself -- otherwise we'll get bad UI diffs and need to look deeper into it

grdddj:
Looks like there is another "edge-case" that we do not account for - when there is a rename of the test-case and therefore the older `master` test-case cannot be found in the new test-cases.

Example: "deleting" `test_recovery` test-case because of creating `test_recovery_bip39` and `test_recovery_slip39` 

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3762985591

PR #2806, opened 2023-02-08
core: pin/rotation wording chage

### LINKED ISSUE #2777: UI2 - customization + PIN
image

mmilata:
Fixes #2777.

cu4 cu3 cu2 cu1 cu0


tests/click_tests/test_autolock.py
```diff
@@ -110,10 +110,10 @@ def test_autolock_passphrase_keyboard(device_handler: "BackgroundDeviceHandler")
     assert layout.text == "< PassphraseKeyboard >"
 
     CENTER_BUTTON = buttons.grid35(1, 2)
-    # total sleep time: 11 * 1.1 = 12.1 seconds
+    # total sleep time: 11 * 1.5 = 16.5 seconds
```
matejcik, line 113: the intention here was to keep clicking for long enough to trigger the autolock, if it incorrectly ignored key presses.
so you can make it shorter (also please add the above to a comment so that we don't forget next time .)

mmilata: How much shorter? The timeout is 1s, below that we'll get a different letter. Something like 1.2s --> c63e7ff539930558e6dcad1faf168c2db626ef47?

matejcik: no I mean, if you want the click delay to be 1.5, you can do ceil(10/1.5) repetitions instead of 11 to get the total run time closer to the 10 (at which point an autolock would trigger if waking by keypress was incorrect)

mmilata: Ahh I see ... e233b74648687724a422b8fb5b4c9c14d209a13e?

tests/click_tests/test_autolock.py
```diff
@@ -132,10 +133,10 @@ def test_autolock_interrupts_passphrase(device_handler: "BackgroundDeviceHandler
     assert layout.text == "< PassphraseKeyboard >"
 
     CENTER_BUTTON = buttons.grid35(1, 2)
-    # total sleep time: 5 * 1.1 = 5.5 seconds
-    for _ in range(5):
+    # keep clicking for long enough to trigger the autolock if it incorrectly ignored key presses
```
matejcik, line 136: i don't think this is true? not sure what the original intention was with the `5` here.

obviously the first test case is checking that the passphrase keyboard keeps Trezor awake
and this other one is checking that if you auto-lock while the passphrase keyboard is up, it gets correctly cancelled.
why enter letters in the first place, i'm not sure. maybe something like, "even if you started typing, autolock must kill it"?
in which case the number of presses doesn't seem important

matejcik (approved):

tests/click_tests/test_autolock.py
```diff
@@ -110,10 +111,10 @@ def test_autolock_passphrase_keyboard(device_handler: "BackgroundDeviceHandler")
...
-    # total sleep time: 11 * 1.1 = 12.1 seconds
-    for _ in range(11):
+    # keep clicking for long enough to trigger the autolock if it incorrectly ignored key presses
+    for _ in range(math.ceil(11 / 1.5)):
         debug.click(CENTER_BUTTON)
-        time.sleep(1.1)
+        time.sleep(1.5)
 
     debug.click(buttons.OK, wait=True)
     assert device_handler.result() == "mzAZ4BgqmFHYxhPgdFH2pR2h1X7jJrNFSs"
```
matejcik, line 120: i'm assuming this won't be true anymore now that the number of presses comes out different?

mmilata: uhhhhhnnnnrrg [6685c10](https://github.com/trezor/trezor-firmware/pull/2806/commits/6685c10cbbe6dc2653279b08bf0558e096b52778)

tests/click_tests/test_autolock.py
mmilata: [84bfd40](https://github.com/trezor/trezor-firmware/pull/2806/commits/84bfd40b2fad843cda88447746dc17775cd23d6a)

matejcik:
squashed fixups and added a fix for #2799 so that we see it working.

mmilata:
works, thank you!

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 23.2.1 (bcfbc9ffb59b51f6c7ecb58d529769ca0db2f7b1)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.2.1 Chrome/104.0.5112.124 Electron/20.3.5 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.4 Universal (revision 9add2b0027048c617d780b4943c00718d72e8334)

PR #2807, opened 2023-02-08
ci(nix): use cached-nix-shell for speed

mmilata:
https://github.com/xzfc/cached-nix-shell

Decreases pipeline runtime by a few of minutes. Needs env image rebuild to be effective.


matejcik (approved):
utACK let's do this

matejcik:
environment is rebuilt, rerunning this pipeline to see if everything is in order

mmilata:
Force-pushed to see how long does a pipeline take without the env job.

PR #2813, opened 2023-02-10
ci: run `connect test core` manually

mmilata:
It has been failing for something like a year now? We deserve green pipelines


matejcik (approved):
GO GO GO

PR #2814, opened 2023-02-10
build: also deploy click-test fixtures

matejcik:
the master diff was failing in #2799 because of missing fixtures on the server.

those are in fact not actually deployed because the deploy job was not requesting results from the clicktest job


matejcik:
...runs but I don't see the hash in question in the list of uploads (`ce176c566cc895b2532e36752fe1c3d817073ca6f62587333bf13dc1834e0beb`)

matejcik:
...ah but that's because it does not exist in master anymore :woman_shrugging: this seems ready to go then

mmilata (approved):

PR #2816, opened 2023-02-11
ci: conditionally use cached-nix-shell in env setup

matejcik:
this should fix ARM failures


matejcik:
build succeeding on a fake tag: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3752061146

mmilata (approved):
let's try it

Issue #2717, opened 2023-01-05
Possibility of AssertionError in select_word


grdddj:
In the current UI2 implementation of `layouts.reset.select_word` we are asserting that the length of word choice is exactly 3.

ISTM this assertion can become False in some very unique situations - but nonetheless.

As I understand it (from the implementation of `management.reser_device.layout._confirm_share_words ` and ` _confirm_word`:
- our seed always has 12 words
- we split it into three parts, each has therefore 4 words
- for each of these three parts, we need to take three unique words to display as a choice to the user
- therefore, when any of the 4-word parts has less than three unique words, the assertion fails and the user receives a RuntimeError without doing anything wrong

The probability is fairly low (I would guess around 1 in 2048^2 cases ?), but it can happen three times for each seed backup.

In case we found there are just two unique elements (just one would be almost impossible), we could just append any of those into the list to make it a length of 3.

PR #2817, opened 2023-02-14
fix(core): remove the possibility of AssertionError in select_word

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2717


matejcik (approved):
utACK

PR #2821, opened 2023-02-14
ci: fix cached-nix-shell failure in hardware tests

mmilata:
Followup of #2816. Also remove `yaml prebuild` because it's a strict subset of `style prebuild`.


matejcik (approved):
right, this is a much better way to do it.

fwiw essentially every command is `$NIX_SHELL --run "poetry run <actual command>"`, couldn't we somehow instead make the whole env run inside nix shell + poetry shell?

matejcik:
here's what to do for poetry, couldn't we trick nix-shell into also giving us the env "inline"?

PR #2822, opened 2023-02-14
Fix ownership proof unit test to match test vector 2 from SLIP-19

andrewkozlik:
See https://github.com/satoshilabs/slips/blob/master/slip-0019.md#test-vector-2-p2wpkh-nested-in-bip16-p2sh.


matejcik (approved):
utACK

PR #2827, opened 2023-02-15
refactor(core): improve flexibility of build

TychoVrahe:
- improves flexibility of build by using USE_xxx defines instead of relying on MODEL everywhere
- allows different touch panel drivers used
- allows usage of STM32F429 mcu
- didn't do this for upython nor rust, but probably should be handled similarly


core/embed/trezorhal/flash.h
```diff
@@ -82,8 +82,10 @@ extern const uint8_t FIRMWARE_SECTORS[FIRMWARE_SECTORS_COUNT];
 // note: FLASH_SR_RDERR is STM32F42xxx and STM32F43xxx specific (STM32F427)
 // (reference RM0090 section 3.7.5)
 #ifndef STM32F427xx
+#ifndef STM32F429xx
```
mmilata, line 85: Nit: `#if !defined(STM32F427xx) && !defined(STM32F429xx)` or `#if !(defined(STM32F427xx) || defined(STM32F429xx))` would be more readable IMHO.

mmilata (commented):
The changes look good, got some very minor nitpicks.

Since this should be pure build system refactor I was wondering if the compiled files will be identical, turns out none of them are. Tried diffing hex dumps but didn't see anything obvious. `bootloader.bin` even has different size, 107520->109056.

Same with the included `bootloader_T2B1.bin`, it has 94208 B but when I build it it has 84992. Is it built using different toolchain than the one in shell.nix, or is it signed or something?

I'm not saying this is necessarily a problem, just find it a bit weird.

core/embed/trezorhal/mpu.c
```diff
@@ -153,7 +153,7 @@ void mpu_config_firmware(void) {
               LL_MPU_REGION_SIZE_1GB | LL_MPU_REGION_FULL_ACCESS |
               MPU_RASR_XN_Msk;
 
-#ifdef STM32F427xx
+#if defined STM32F427xx || defined(STM32F429xx)
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/ddc4be5379588242f44a9540599aee5e131100b4

core/embed/trezorhal/flash.h
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/ddc4be5379588242f44a9540599aee5e131100b4

TychoVrahe:
The T2B1 bootloader was actually just a copy of the T2T1 bootloader, i just copied that so that the T2B1 firmware could compile, didn't even meant to commit that frankly. Anyway, i replaced the T2B1 bootloader binary with one compiled from here, i think its more convenient that the firmware can compile, but it doesn't represent anything useful anyway at this point. 

One reason why the binaries will differ is that the order of files have changed a bit in sconscript, thus the functions are ordered differently in flash.

Nice catch with the bootloader size. The reason is that the sdcard.c is now compiled in bootloader, i thought this will be a non-issue as it is unused and linker removes unused functions. But turns out that `SDIO_IRQHandler` redefines the weak default handler and gets linked in, which causes quite a few things from HAL being linked in too. I am gonna need to figure out some solution to this before we can merge this.

TychoVrahe:
here is a proposed solution: https://github.com/trezor/trezor-firmware/commit/75378af8d7bf752b2423605a48afda319182c47a
basically sconscript asks the board for features it needs (and expects in source code). The board configuration adds what it can. The source code via ifdefs must handle situation when given feature is not available, or fail the build if it makes no sense.

not sure if its not getting overcompilcated

mmilata (approved):
Yea looks very reasonable. I'm imagining we may want to pass part of `FEATURES_AVAILABLE` directly to cargo, but that's out of scope here.

mmilata:
Squashed and rebased, will merged after CI passes.

PR #2830, opened 2023-02-16
fix(tests): make master_diff more reliable

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2799

There was a little mismatch in looping through the `same` testcases, because it also contained the `removed` cases - `master_tests.items() - removed_here.items()` did not work, as they had different keys - one with the `group` inside and one without it.

Also when we cannot find an arbitrary `actual` directory, defaulting to `recorded`, so that it can be run even after the `actual` directory is deleted (which happens every time a new UI-test suite is called).

Tested locally and as a part of https://github.com/trezor/trezor-firmware/pull/2803 - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3782160703

### DIFF
--- a/tests/ui_tests/reporting/master_diff.py
+++ b/tests/ui_tests/reporting/master_diff.py
@@ -66,14 +66,15 @@ def get_diff() -> tuple[dict[str, str], dict[str, str], dict[str, tuple[str, str
                 testname(test): current_tests[test]
                 for test in (current_tests.keys() - master_tests.keys())
             }
-            # items in both branches
-            same = master_tests.items() - removed_here.items() - added_here.items()
-            # create the diff
+            # create the diff from items in both branches
             diff_here = {}
-            for master_test, master_hash in same:
+            for master_test, master_hash in master_tests.items():
+                full_test_name = testname(master_test)
+                if full_test_name in removed_here:
+                    continue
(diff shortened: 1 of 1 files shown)


tests/ui_tests/reporting/master_diff.py
```diff
@@ -230,7 +232,10 @@ def tmpdir():
     for test_name, test_hash in added_tests.items():
         path = SCREENS_DIR / test_name / "actual"
         if not path.exists():
-            raise RuntimeError("Folder does not exist, has it been recorded?", path)
+            path = SCREENS_DIR / test_name / "recorded"
```
matejcik, line 235:
```
def get_screen_path(test_name: str) -> Path | None:
    path = SCREENS_DIR/ test_name / "actual"
    if path.exists():
       return path
    path = SCREENS_DIR / test_name / "recorded"
    if path.exists():
       print("WARNING: no actual screens for {test_name}, recording may be outdated: {path}")
       return path
    print("WARNING: missing screens for {test_name}. Did the test run?")
    return None
```

grdddj: Good idea, done in the force-push

matejcik (approved):

Issue #2838, opened 2023-02-20
UI2: Get rid of the loaders on startup


Hannsek:

grdddj:
Deeply connected with https://github.com/trezor/trezor-firmware/issues/2825

Issue #2825, opened 2023-02-15
UI2: Welcome screens (initialized)


Hannsek:
<img width="583" alt="image" src="https://user-images.githubusercontent.com/57008159/220160057-91c40b1b-48fb-498a-818f-b6bd2d25b1f0.png">

The second screen will be visible for app. 1,5 s.

Default name of the device is `Trezor Model T`

The first logo comes from the bootloader.
[locks.zip](https://github.com/trezor/trezor-firmware/files/10791600/locks.zip)

grdddj:
TODO: there is a need to get pixelated icon

Hannsek:
Please move the lock up so that it will be positioned 48 px from the top of the screen

PR #2840, opened 2023-02-21
Initialized welcome screen + removing startup loaders

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2825:
- show welcome screen with an icon and a model name, for at least `1500 ms`

Fixes https://github.com/trezor/trezor-firmware/issues/2838:
- do not show `Starting up` loader on startup, so that the welcome screen is visible all the time

In the end, I am calling `welcome_screen` from `boot.py::bootscreen`, as that fulfills the purpose quite well, and there seems to be no big delay even on the hardware device.

I was looking at https://github.com/trezor/trezor-firmware/pull/2521 - the possibility of calling that even earlier in `main.c`, but that would mean having it on two places (in `unix` and in `firmware`) and we could not call it from `upy` whenever we wanted. Also, we would still need to measure the time (`1500 ms`) in `boot.py::bootscreen`, so we show it long enough. But up to discussion...


core/src/boot.py
```diff
@@ -1,21 +1,47 @@
+import utime
```
matejcik, line 1: i would prefer to import and call `show_welcome_screen` directly from trezorui2 right here, to have it on screen before we start importing things. imports take a relatively long time so showing the screen before that would be nice.

we should skip fadein/fadeout, the screen is simple enough that it should not be a problem?

(we should also remove the fadein/fadeout sequence from later in this file, so that it does not hide the welcome screen)

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1343,6 +1350,14 @@ extern "C" fn new_show_busyscreen(n_args: usize, args: *const Obj, kwargs: *mut
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_show_welcome_screen(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
```
matejcik, line 1353: Rename to `draw_welcome_screen` and have it call `paint()` directly. We don't even need to instantiate a `LayoutObj` for it: just create the `WelcomeScreen` on stack, call `frame.place(consts::screen())` and `frame.paint()` and return.
(see also [this](https://github.com/trezor/trezor-firmware/blob/c995a9a359f57f9e7cdc19142e73380907945059/core/embed/rust/src/ui/model_tt/bootloader/mod.rs#L102-L115))

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1708,6 +1723,10 @@ pub static mp_module_trezorui2: Module = obj_module! {
     /// ) -> CANCELLED:
     ///     """Homescreen used for indicating coinjoin in progress."""
     Qstr::MP_QSTR_show_busyscreen => obj_fn_kw!(0, new_show_busyscreen).as_obj(),
+
+    /// def show_welcome_screen() -> object:
+    ///     """Show logo icon with the model name at the bottom."""
+    Qstr::MP_QSTR_show_welcome_screen => obj_fn_kw!(0, new_show_welcome_screen).as_obj(),
```
matejcik, line 1729: istm this does not need to be `obj_fn_kw`, can be `obj_fn_0` and not take arguments?

core/embed/rust/src/ui/model_tt/layout.rs
matejcik: (that way we will eventually be able to move the draw call to C without doing anything fancy)

it also prooobably doesn't need the `try` thing? if you don't try to allocate, there is nothing that could fail

core/embed/rust/src/ui/model_tt/layout.rs
grdddj: Would be certainly nice, but now we do not have `obj_fn_0!` and even though I see the definitions of `obj_fn_1` and `obj_fn_2`, I do not know how to create it.

Tried

```rs
macro_rules! obj_fn_0 {
    ($f:expr) => {{
        #[allow(unused_unsafe)]
        unsafe {
            use $crate::micropython::ffi;

            ffi::mp_obj_fun_builtin_fixed_t {
                base: ffi::mp_obj_base_t {
                    type_: &ffi::mp_type_fun_builtin_0,
                },
                fun: ffi::_mp_obj_fun_builtin_fixed_t__bindgen_ty_1 { _0: Some($f) },
            }
        }
    }};
}
``` 

with receiving 
```
error[E0308]: mismatched types
    --> src/ui/model_tt/layout.rs:1729:52
     |
1729 |     Qstr::MP_QSTR_show_welcome_screen => obj_fn_0!(draw_welcome_screen).as_obj(),
     |                                                    ^^^^^^^^^^^^^^^^^^^ incorrect number of function parameters
     |
    ::: src/micropython/macros.rs:12:75
     |
12   |                 fun: ffi::_mp_obj_fun_builtin_fixed_t__bindgen_ty_1 { _0: Some($f) },
     |                                                                           ---- arguments to this enum variant are incorrect
```

core/src/boot.py
grdddj: RIght, tried it and it is shown really very quickly.

Currently I have it like:

```python

# isort:skip_file

import utime
import trezorui2
from trezor import ui

# Showing welcome screen as soon as possible.
# Remembering time to control how long we show it.
trezorui2.show_welcome_screen().paint()  # type: ignore [Cannot access member "paint" for type "object"]
welcome_screen_start_ms = utime.ticks_ms()

# Setting up the display light.
# (For some reason needs to be done _after_ painting the welcome screen)
ui.display.backlight(ui.BACKLIGHT_NORMAL)

import storage
...
```

- it is strange that the backlight needs to be called after drawing

matejcik: what happens if you put it before drawing? :)

core/embed/rust/src/ui/model_tt/layout.rs
matejcik: here it is: https://github.com/trezor/trezor-firmware/pull/2521/files#diff-9cf76d30275f61c11d21878407648fcfeb12737698c9fb0faade3209ea0014cd

core/src/boot.py
grdddj: The screen is just blank, without showing anything - and after that 1500 ms it shows the green homescreen normally
image

core/embed/rust/src/ui/model_tt/layout.rs
matejcik: oh, that's exactly the same thing you have.

well the error seems to tell you that there's a wrong number of function parameters, indicating that you didn't remove parameters from `draw_welcome_screen` maybe?

core/src/boot.py
matejcik: i see, that might be an emulator-specific problem possibly?

in this case it would be great if you could also do the backlight call from Rust, because then you can postpone importing `trezor.ui`

matejcik: it's most likely a matter of not calling `display.refresh()` which the emulator requires and the real device is less stringent about.
that too should be done in Rust, after the paint call: sync -> paint -> refresh

core/embed/rust/src/ui/model_tt/layout.rs
grdddj: Both done in [5601deb](https://github.com/trezor/trezor-firmware/pull/2840/commits/5601debbbc46216c04794814e5199550cfdf9bca)

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -140,6 +141,12 @@ where
     }
 }
 
+impl ComponentMsgObj for WelcomeScreen {
```
matejcik, line 144: i think we don't need this impl when we don't put `WelcomeScreen` into a `LayoutObj` ?

core/src/trezor/pin.py
```diff
@@ -10,6 +10,18 @@
 _started_with_empty_loader = False
 keepalive_callback: Any = None
 
+_ignore_loader_messages: tuple[str, ...] | None = None
```
matejcik, line 13: to simplify, make it always a tuple and set to `()` when there are no ignored messages

grdddj: Sure, reverted in [00263df](https://github.com/trezor/trezor-firmware/pull/2840/commits/00263dfc3b161bd142a44d9cb0b17935784e35ae)

core/embed/rust/src/ui/model_tt/layout.rs
grdddj: Deleted in [3562258](https://github.com/trezor/trezor-firmware/pull/2840/commits/356225877c2db956f6fb0dbe782b139e2cdb528e)

matejcik (approved):

STew790:
**QA OK**
No loading screen on startup present and instead, "Trezor Model T" welcome screen shown.

**Info:**
 - Suite version: desktop 23.3.1 (8f0a5a67bd1cd0cf95724ee1468ee6ad70aead60)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.3.1 Chrome/104.0.5112.124 Electron/20.3.12 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: model T 2.5.4 Universal (revision 557e29769b09dd3abd7b67f9b4fe3d380db9f6c9)

Issue #2316, opened 2022-06-01
Enable jump-to-bootloader on TT with release of new bootloader


matejcik:
Jump-to-bootloader is currently gated by experimental features, as seen in [5c5dff3](https://github.com/trezor/trezor-firmware/pull/2284/commits/5c5dff38587bc0668d0297ff7e535a2fa5d9c7d6)

We should revert that commit when releasing the new bootloader.

Issue #2794, opened 2023-02-06
Interrupting firmware update at the wrong time erases seed


matejcik:
**To reproduce:**

1. set up device with seed
2. start firmware update
3. midway through, pull out the cable

**Desired result:**

After replugging, Trezor asks you to fix failed firmware update. When you install fw again, your settings and seed are there intact.

**Actual result:**

Device returns to factory state. Seed is wiped.

---

The above _seems_ technically possible.
One limitation comes to mind: bootloader would need to be able to determine which firmware vendor "owned" the storage, so that when you corrupt a signed firmware installation and then reinstall with unsigned, the storage is properly wiped.

For TT, this seems fixable. Store a vendor header identifier in storage when setting it up. If it matches the firmware being installed, keep storage, otherwise erase it.
This would require the bootloader to understand storage to some extent. Either properly read public fields, or figure out some sort of trick, like verify storage integrity keyed off the vendor identifier?? ( ¯\\\_(ツ)\_/¯)

---

For T1 a similar issue exists between signed / unsigned firmwares. We _might_ be able to get around it by declaring that only if a firmware is signed, it gets to keep storage. This would allow an attacker to "upgrade" storage from an unsigned firmware to a signed one -- however, that does not seem like a big problem, because an attacker can achieve exactly that by (1) installing another unsigned firmware that dumps the storage, (2) extracting all relevant information from it, brute-forcing the PIN if necessary, (3) installing a signed firmware, (4) recreating storage contents.
(there is still an opening in that a security issue might exist in signed firmware that can be triggered by invalid storage data. so an attacker installs unsigned firmware, writes bad data to storage, upgrades storage, signed firmware gets exploited. But the exploit would need to be persistent, otherwise the best the attacker can do is read back the storage contents they just installed :)) )

In any case, even if we plug this hole, there might not be enough space in the T1 bootloader to fit the additional logic.

matejcik:
cc @andrewkozlik for thoughts on the storage thing

matejcik:
> For TT, this seems fixable. Store a vendor header identifier in storage when setting it up. If it matches the firmware being installed, keep storage, otherwise erase it.

problem: this still allows unsigned firmware to install storage contents that match some other vendor header (i.e. the official one)

matejcik:
[...]

PR #2841, opened 2023-02-21
chore(core): enable reboot-to-bootloader without experimental features

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2316:
- not needing experimental features to trigger reboot-to-bootloader functionality

### DIFF
--- a/core/.changelog.d/2841.added
+++ b/core/.changelog.d/2841.added
@@ -1,0 +1,1 @@
+Add the possibility of rebooting the device into bootloader mode
--- a/core/src/apps/management/reboot_to_bootloader.py
+++ b/core/src/apps/management/reboot_to_bootloader.py
@@ -7,13 +7,9 @@ if TYPE_CHECKING:
 
 
 async def reboot_to_bootloader(ctx: Context, msg: RebootToBootloader) -> NoReturn:
-    import storage.device
-    from trezor import io, loop, utils, wire
+    from trezor import io, loop, utils
     from trezor.messages import Success
     from trezor.ui.layouts import confirm_action
-
-    if not storage.device.get_experimental_features():
-        raise wire.UnexpectedMessage("Experimental features are not enabled")
 
     await confirm_action(
         ctx,


core/.changelog.d/2841.changed
```diff
@@ -0,0 +1 @@
+Reboot to bootloader does not require experimental features
```
matejcik, line 1: changelog-wise, this is a "new" feature. so please move this to `added` and modify the text accordingly

grdddj: Done in the force-push

matejcik (approved):

Hannsek:
Can we change the copy on that screen? 😅

Hannsek:
@STew790 You'll need to wait a bit longer, we are waiting for another pr to merge. 

Answer to myself: Yes, we can change it. 😄

Hannsek:
After releasing this, when users will want to update fw, they click on "update fw" in suite and they will see this screen. Is that right @grdddj? Do we need the confirmation from the user?

grdddj:
> After releasing this, when users will want to update fw, they click on "update fw" in suite and they will see this screen. Is that right @grdddj? Do we need the confirmation from the user?

No idea, I do not know anything about `reboot-to-bootloader` functionality, but your suggestion makes sense

Hannsek:
@matejcik maybe?

Hannsek:
Why do we need it? I don't like the fact that the process of updating takes at least 3 clicks on Trezor screen.

matejcik:
In that case i'd much rather remove the initial confirmation from bootloader side than before executing the jump. I don't have a specific argument, but dropping the user to bootloader without any interaction reeks of badness.

it would make the UX flow make more sense too, IMHO: change wording to something like "proceed to firmware installation? y/n", then jump to bootloader blue screen "waiting for host", and then install.

while we're thinking about this, we could allow interaction-less upgrade, i.e., skip the "really install firmware by SatoshiLabs version 2.99" dialog, if (a) the vendor is unchanged, and (b) it is a strict **up**grade, and (c) click flag is set.
i.e.:
* reinstall of same version still gets a confirmation
* downgrade gets confirmation
* vendor change gets confirmation, obviously
* any operation with devel firmware gets confirmation

then you would, in firmware, click Yes on "proceed to installing firmware?", Trezor screen goes blue, waits for host, straight to  progress bar -> reboot -> you're done

Hannsek:
Hmm good idea, the less click for user the better. 
What is your time estimate?

In that case, we don’t have to do the happy path in bootloader blue. So the user won’t even see that he is in boot loader. I.e. waiting for host screen and progress bar would be black.

matejcik:
> Hmm good idea, the less click for user the better. What is your time estimate?

It's pretty easy actually. Skipping the "install firmware" button when coming from firmware is a matter of two or so lines. Finding the upgrade happy path is also easy to code, but I'd like to go through the implications better, so I'd defer that to the next bootloader release (along with #2794 hopefully)

> In that case, we don’t have to do the happy path in bootloader blue. So the user won’t even see that he is in boot loader. I.e. waiting for host screen and progress bar would be black.

Doable, but no point in doing it imho. The firmware upgrade process _is_ special.

Hannsek:
I agree. I will create another issue after thinking it through.

bosomt:
Tested yesterday with `bootloader_T2T1_qa-2.1.0.bin` via `trezorctl device reboot-to-bootloader` and it worked as expected 🎉 

I will retest with final version

Hannsek:
Can we test this also with Suite?

bosomt:
I think that we need to ask Suite developers to implement `reboot to bootloader` feature for TT.
@mroz22  right ? Do you want me to send Trezor Suite issue/feature request or its already covered?

Hannsek:
@mroz22 is not here this week. So probably @marekrjpolak. cc @matejkriz

Issue #2638, opened 2022-11-14
QR code of Taproot account can't be scanned from the T1 device


STew790:
**Describe the bug**
QR code for Taproot address shown on T1 can't be scanned on Android devices (works for iPhone). It works for the production version of the T1 firmware but not with the latest master T1 version.

**Steps to reproduce:**
1. Go to Taproot account in Suite
2. Click on "Receive"
3. On device, press "QR code"
4. Try to scan QR code from T1 on Android device.

**Info:**
 - Suite version: desktop 22.12.0 (ff01d767b3827ce73296a25fa36cfded68457074)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.12.0 Chrome/100.0.4896.160 Electron/18.3.5 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: model 1 1.11.3 Universal (revision 1852059e87d2070239cbfadb11b8ca516f80457c)
 - Transport: bridge 2.0.32

**Expected result:**
QR code of Taproot account can be scanned on T1 device.

**Actual result:**
QR code of Taproot account can't be scanned on T1 device.

matejcik:
how many different android devices and QR code scanner apps have you tried? :)

STew790:
Sorry, should have been more specific. We tried it on 2 Android devices (11 and 12). Tried it with Exodus app, our Trezor Suite in Chrome and a standard Android camera app. 

But @bosomt just tried to read it on MacOS via https://webqr.com/index.html and it didn't work for him there.
And I tried it with a standard iOS iPhone SE 2020 camera and it registered it immediately.

Did it work for you somewhere?

matejcik:
i believe @grdddj did test this on Android, but I don't know what app / phone / whatever.

our options now are:
(0) ignore this, say that Android QR libraries need to step up their game
(1) show taproot addresses in uppercase, which makes the QR code smaller (Suite will dislike us because they'll need to show tr addresses in uppercase for T1 but lowercase for TT)
(2) cut off the other part of the QR code. Maybe Android libraries dislike their pixels missing on one side but won't mind the other. We could also _rotate_ the QR code and cut off one of the sides.
(3) Make the blocks smaller -- use same density of data, but squish it into a half-size image. This will be scannable _if_ the phone's camera is good enough to focus on the smaller image.

[...]

PR #2842, opened 2023-02-22
fix(legacy): show complete taproot QR code

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2638:
- show the taproot QR code smaller, but complete

image

### DIFF
--- a/legacy/firmware/layout2.c
+++ b/legacy/firmware/layout2.c
@@ -790,12 +790,8 @@ void layoutAddress(const char *address, const char *desc, bool qrcode,
     }
 
     oledInvert(0, 0, 63, 63);
-    if (side > 0 && side <= 33) {
-      // For QR code V4 (33x33), still maintaining two pixels per module
-      // so that it is nicely readable. Not showing the bottom row,
-      // which is not a problem thanks to the QR code's error correction.
-      // It is a use-case for BTC taproot addresses.
-      int offset = side <= 29 ? 32 - side : 0;
+    if (side > 0 && side <= 29) {
+      int offset = 32 - side;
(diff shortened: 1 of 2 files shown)


matejcik:
umm shouldn't the background be black?

grdddj:
> umm shouldn't the background be black?

Haven't changed anything, all the QR codes on T1 have white background

image

matejcik:
Huh, indeed. It looks weird with the QR so small, but maybe we don't need to care. What do you think @Hannsek ?

Hannsek:
It is fine. However, it would be nice to have there black thick frame around the qr, and have the white background only "behind" the qr code.

Also, It will only be displayed as a tiny one?

matejcik (approved):

grdddj:
> It is fine. However, it would be nice to have there black thick frame around the qr, and have the white background only "behind" the qr code. But if it is not a "fruit lying already on the ground" I would let it be.
> 
> Also, It will only be displayed as a tiny one?

I would also not spend much time with this. Will merge it.

It is displayed so small because the bigger version does not fit whole on the page - it was scannable in the past, but apparently not anymore, therefore this "reverting"

Hannsek:
I can scan the big one now as well…

sime:
Note to QA: Ensure scanning of the QR code is tested on Android.

bosomt:
**QA OK**

works as expected /when enough light /
Tested on Android phone, Android tablet and on iPhone 11

 - Device: model 1 1.11.3 Universal (revision 00dde0e3125b5c3ca940f0987069a635587592f4)

Issue #2790, opened 2023-02-03
FIDO2 falling back to U2F


Hannsek:
https://satoshilabs.slack.com/archives/CL1D61PQF/p1675367498377149

mmilata:
Note: `core/tools/hid-bridge` might be useful to debug/bisect this.

Issue #2205, opened 2022-04-04
Unable to use Trezor Model T as security key on Google with Safari


sime:
**Describe the bug**
Adding Trezor Model T as a security key to Google fails.

**Firmware version and revision**
2.4.3

**Desktop/smartphone setup (please complete the following information):**
 - OS: macOS
 - Browser Safari 
 - Version 15.3

**To Reproduce**
1. Go to https://myaccount.google.com/signinoptions/two-step-verification
2. Authenticate
3. Click 'Add Security Key'
4. See 'Got your security key?' modal and click 'Next'
5. Press the green tick 

**Expected behavior**
Successful pair Model T to Google account

**Screenshots**
![P_20220330_151851(1)](https://user-images.githubusercontent.com/216917/161536893-77917b43-8371-4f1a-838b-3670e4383d30.jpg)
![Screenshot 2022-04-04 at 13 32 20](https://user-images.githubusercontent.com/216917/161536924-399e97f6-da1a-4111-abd4-37a10717f8c0.png)

**Additional context**
Model 1 works.

sime:
@mcudev Just to clarify, the issue is with Google.com ?

mcudev:
@sime yes, google.com. testing again today, i find that things are working a little better/differently with chrome version 101.0.4951.64 (Official Build) (64-bit).

now, i can get my key registered with the "FIDO2 Register" process and working with this google accounts + chrome combo, there is just one workaround that i had to do. when registering the key, i get this screen:

![allow_skip](https://user-images.githubusercontent.com/29890609/168420655-30bff5a3-3a7c-4d1e-89df-0d6fe24255a8.png)

if you click allow, the process fails with:

![could_not_connect](https://user-images.githubusercontent.com/29890609/168420665-07a10027-6f2f-4138-b271-cb5afa88f2f5.png)

when you click skip, you can name the key and continue on and things work. also, when logging in using the fido2 authenticate process, with this google + chrome combo, it actually lets me in instead of seeming like it works and then forcing me to use a fallback 2fa mechanism.

firefox still uses the u2f authenticate process and that still works too.

sime:
Can confirm it is failing for me on Chrome and Safari.

andrewkozlik:
Reopening. PR #2834 only fixes https://github.com/trezor/trezor-firmware/issues/2205#issuecomment-1109755886 for Chrome.

bosomt:
[...]

PR #2844, opened 2023-02-22
Fix FIDO issues with Safari browser and CBOR encoding

andrewkozlik:
Fixes https://github.com/trezor/trezor-firmware/issues/2205.
Safari browser changes the channel ID for every single FIDO message. We need to ignore the CID in U2F and keep up the same U2F dialog as long as the dialog corresponds to the same request data.

Fixes https://github.com/trezor/trezor-firmware/issues/2790.
A boolean value was incorrectly being encoded as an integer in CBOR encoding.

### DIFF
--- a/core/.changelog.d/2205.changed
+++ b/core/.changelog.d/2205.changed
@@ -1,0 +1,1 @@
+Ignore channel ID in U2F.
--- a/core/src/apps/webauthn/fido2.py
+++ b/core/src/apps/webauthn/fido2.py
@@ -211,8 +211,8 @@ _MAX_CRED_COUNT_IN_LIST = const(10)
 # The CID of the last WINK command. Used to ensure that we do only one WINK at a time on any given CID.
 _last_wink_cid = 0
 
-# The CID of the last successful U2F_AUTHENTICATE check-only request.
-_last_good_auth_check_cid = 0
+# Indicates whether the last U2F_AUTHENTICATE had a valid key handle.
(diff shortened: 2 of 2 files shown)


matejcik (approved):
utACK

bosomt:
I'm able to login to gmail in Safari browser on macARM.

but first FIDO2 screen does not display Google logo + you have to confirm login twice FIDO2 and then U2F 🤯 

image

image


**Info:**
 - Suite version: desktop 23.2.2 (00fe136fa75bc04628abf858e09223170b5cf188)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.2.2 Chrome/104.0.5112.124 Electron/20.3.12 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.5.3 Universal (revision 2f03ace311584988d5aeab58fd1acf24ef54711a)

andrewkozlik:
> you have to confirm login twice FIDO2 and then U2F

That's just one of several bugs in Safari's implementation of FIDO. My Yubikey also requires two confirmations in the same scenario. I can think of a pretty complicated hack to work around this, but it's not worth the effort and I think it would technically violate the protocol.

> first FIDO2 screen does not display Google logo

I might want to look into that, but the screen should not appear in the first place so low priority from my point of view.

PR #2848, opened 2023-02-27
T1 layout fixes

### LINKED ISSUE #1453: Verifying Stellar Wallet Address is cut off on Trezor One Device Screen
While using Exodus to receive Stellar coins on the Trezor One, viewing the wallet address on the device to receive causes the end of the address to be cut off while verifying

Trezor Firmware:  1.9.3.

 - Wallet software: Exodus
 - OS: Mac Catalina 10.15.7
 - Exodus Version: 21.1.29 

Reproduce:
1) Connect wallet and put in passphrase, testing with no passphrase
2) Click wallet link button to view contents
3) Click Stellar coins to receive
4) click receive then review on device
5) Address will be presented in the app
Screen Shot 2021-02-06 at 2 19 20 PM
6) Device will show cut off Address
trezorBug

Expected behavior:
The full address should be viewable on the trezor device

### LINKED ISSUE #2373: String of amount or fee amount not visible if too long
**Describe the bug**
If the string of amount or fee is too long it's not visible on Trezor's display and there's no line break

FW 1.11.1

**To Reproduce**
1. Send transaction with an amount of which string length exceeds the limit of Trezor One display
2. Confirm the transaction
3. During the on-device confirmation the whole string is not visible

**Expected behavior**
The whole string is visible 

**Screenshots**
Screenshot 2022-07-07 at 14 34 23

**Additional info**
Supposedly happening especially with ERC20 tokens

matejcik:
fixes #2373
fixes #1453 
two bugs that look very similar but in fact are in completely different code paths

also fixes a small bug in UI reports, and I added Ethereum address screens to the UI fixtures


matejcik:
new Stellar address screen:
image

Ethereum fees:
image
the numbers sometimes crawl off the screen because there is a fixed number of characters but the font is not monospace.
maybe we should make the line one char shorter to account for this?

---

aside: when I copy the image directly from the results page, it's tiny because its true resolution is 128x64
but when I screenshot it, it's huge because it's 2x scaled on the page and additional 2x scaled because i'm on a hidpi screen :woman_shrugging:

mmilata (approved):
LGTM, thanks for fixing the new testcases for ui tests.

CI fails because the `long_fees` fixture is mising for TT.

bosomt:
Still missing few pixels in last digit.

image

image

 - Suite version: desktop 23.2.2 (00fe136fa75bc04628abf858e09223170b5cf188)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.2.2 Chrome/104.0.5112.124 Electron/20.3.12 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model 1 1.11.3 Universal (revision be920d7cec2495d82496af500cbdeec6b945c505)

matejcik:
do you mean to tell me that Suite still doesn't use EIP1559 ??

anyway yes, this screen was not fixed, because there is no good way to fix it without completely reflowing the layout. one small improvement could be made by removing the icon, but even that might not be enough, esp. with long currency symbols...

you'll need to use Metamask (presumably) to trigger the EIP1559 flow where the individual fee values have their own screens

STew790:
I tried this on Metamask and I can see the cut off "ETH" on my screen as well. It's not that visible but I can clearly see the "T" crawling into the screen here.

**Screenshot:**
image


**Info:**
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.3.1 Chrome/104.0.5112.124 Electron/20.3.12 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: model 1 1.12.1 Universal (revision 7430850da5258e5d19ae55bac94a31447a9d4d14)

matejcik:
same as above, the layout is only changed on the fee screens. This is "recipient" screen, and there is currently nowhere to wrap the line to.
(looking more carefully at this screen, it might be possible to just put everything on one logical line and wrap that line, because there's a lot of space left behind from the address string. but that would be a separate issue)

STew790:
Ah, my bad, I read your comment wrong, sorry. Ok, so I tried to go through with the payment on MM and I got to this (unfortunately, the fee per gas is cut off):
image
image
image

**Info:**
Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.3.1 Chrome/104.0.5112.124 Electron/20.3.12 Safari/537.36
OS: Linux x86_64
Screen: 1920x1080
Device: model 1 1.12.1 Universal (revision 7430850da5258e5d19ae55bac94a31447a9d4d14)
Transport: BridgeTransport 2.0.32

matejcik:
i see it, i also commented on it on the original post

> the numbers sometimes crawl off the screen because there is a fixed number of characters but the font is not monospace.
> maybe we should make the line one char shorter to account for this?

i guess we can change it to one char shorter, the patch is tiny and the fw is not signed yet so maybe worth it @Hannsek ?

Hannsek:
Yes, I thought that you already did it. 😄 Let's do it now then.

But what about the coin name - ETH? Can we add there non-breaking space between the letters?

matejcik:
> Can we add there non-breaking space between the letters?

The line breaking is character based and there is no attempt to take whitespace into account. Changing that is a somewhat bigger task and I don't think worth doing in the legacy codebase, esp. seeing that there is a lot of other more easily solved problems -- such as the above QA examples of screens that can _also_ overflow that were not touched here.

Issue #2152, opened 2022-03-03
Show path for internal outputs in BTC signing for Model T


andrewkozlik:
When Trezor signs a transaction which is transferring BTC to another of the user's own accounts, it requires user confirmation of the destination address even if the destination account is under the same seed and passphrase. Trezor T doesn't show any indication that the address belongs to the user. Trezor 1 handles this better. It requires the user to confirm the output, but adds information that it belongs to the user’s own legacy/SegWit/Taproot account #...

![image](https://user-images.githubusercontent.com/42678794/156577290-84c6488a-355c-4f6d-a20c-1cc2d14fdee9.png)

We should add the same feature in Trezor T. This will facilitate account migration from old account types to Taproot accounts, which is especially interesting for CoinJoin. We need to figure out how to fit this information on the confirmation screen, which is already quite full:
![image](https://user-images.githubusercontent.com/42678794/156578685-fe5a159d-1608-42f3-9317-c99219531595.png)
For internal outputs we could:
- <s>remove the address and show only the human readable account description `BTC Taproot account #1`,</s>
- show both the address and account, but shorten the address to two/three lines max with `...`,
- <s>show both the address and account in full and split over multiple pages if necessary.</s>

sime:
Answering my own question, on the 'from' screen will show the amount.

matejcik:
this is a great point, I didn't realize that the designs don't have the separate amount screen.
but I believe it was there at some point?
one total amount is not enough if you have multiple recipients

andrewkozlik:
We certainly need to show the amount for each output, but to be clear I think both address and amount should be on one confirmation dialog, so that it's obvious to which address the amount belongs. (The dialog can be scrollable to multiple screens.)

I think we said that we don't need to show the currency in the account name "BTC L. SegWit" since the currency is the same for the entire transaction.

sime:
Paying to self screen with amount included.

![tt-paying-to-self](https://user-images.githubusercontent.com/216917/171186769-cd298586-7431-4360-a64c-5fb63e1a8ef7.png)

matejcik:
that works for pay-to-self but not for pay to others, where you want to show the full address.

sime:
The following is the alternative. It can be used for both use cases. Paying to self and others. When paying others, omit the last line, and complete the address.

![tt-pay-to-everyone](https://user-images.githubusercontent.com/216917/171203593-4f1ab35d-dde1-416c-9d78-8626c39654be.png)

[...]

PR #2851, opened 2023-02-28
Script type checks for change-outputs

matejcik:
Issue originally reported by @jhoenicke: Trezors do not do any script type checks on change-outputs, meaning that it is possible to invisibly send change to the right path but wrong script type. This would make the change address invisible in basically every existing wallet software.
This does not cause coin loss because it is relatively easy to recover such "hidden" funds.


matejcik:
`test_apps.bitcoin.approver.py` is failing because the outputs are all PAYTOWITNESS but expected to be PAYTOTAPROOT. is that a bug in the test? @andrewkozlik

legacy/firmware/signing.c
```diff
@@ -1487,6 +1553,14 @@ static bool signing_validate_input(const TxInputType *txinput) {
     return false;
   }
 
+  if (!coin->decred && txinput->has_decred_tree) {
```
matejcik, line 1556: should this be under `#if !BITCOIN_ONLY`, just for neatness' sake?

legacy/firmware/signing.c
```diff
@@ -1530,6 +1604,25 @@ static bool signing_validate_input(const TxInputType *txinput) {
...
 
+static bool signing_validate_prev_input(const TxInputType *txinput) {
+  if (txinput->prev_hash.size != 32) {
+    fsm_sendFailure(FailureType_Failure_DataError,
+                    _("Provided prev_hash is invalid."));
+    signing_abort();
+    return false;
+  }
+
+  if (!coin->decred && txinput->has_decred_tree) {
```
matejcik, line 1615: dtto

matejcik (commented):
LGTM overall.

my overarching thought is that sending to a mismatched script type really should be a safety-checks thing, but we don't need to go into that in this PR.

legacy/firmware/signing.c
andrewkozlik: Could be, but what should BTC-only firmware do then in `decred_tree` is specified? Should it ignore or return error?

matejcik: ignore bc from the btc-only point of view the field doesn't exist and the protobuf entry is just the kind of noise protobuf is explicitly allowed to have

andrewkozlik:
> sending to a mismatched script type really should be a safety-checks thing

I don't understand.

legacy/firmware/signing.c
andrewkozlik: `core` also doesn't take BTC-only into account in this check.

andrewkozlik:
On a different note, once https://github.com/trezor/trezor-firmware/issues/2152 is complete I'd start thinking about changing the logic that a transfer from a multisig account to a singlesig account on the same path is treated as a change-output.

andrewkozlik (approved):
utACK

legacy/firmware/signing.c
matejcik: and it probably also should. but ok, let's keep it the same on both sides

matejcik:
> sending to a mismatched script type really should be a safety-checks thing

like, only allowed when safety checks are lowered.

While there _are_ usecases for using a derivation path with a mismatched script type, I strongly suspect that the vast majority of cases are mistakes that are, for the layperson, difficult to recognize and recover from.
(My model of why this would happen is a misconfigured Electrum or some wallet of that sort. Then once you lose your PC, it will look like the coins are lost, even though recovering them is as simple as redoing the original broken configuration. This is a usecase for descriptors, but nobody is actually backing up their descriptors.)

This is the kind of footgun we should be removing now that we realize we have it lying around.
As a prototype of "good" firmware behavior, a transaction like that should be rejected, ideally before the user even sees anything (which is not currently possible ofc), unless the user has opted into doing weird things.

PR #2854, opened 2023-02-28
ci: hardware: wait longer for T1 firmware to install

mmilata:
The automation currently pushes yes button while the firmware is still being installed, probably because coinjoin and/or ethereum networks made the image larger. Also due to coinjoin we now run more tests so the job timeout needs to be increased.

Tested working in the private repo.


matejcik (approved):
utACK

PR #2856, opened 2023-03-02
fix(legacy): lower maximum of chars per line

matejcik:
so that we actually fit if all the biggest characters are used


matejcik:
if QA OKs this, please cherry-pick into the release branch.

mmilata (approved):
LGTM [[ui diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3862295438/artifacts/master_diff/index.html)]

bosomt:
- Device: model 1 1.11.3 Universal (revision 65e579c0418e6f706a0bbf208ecc7fa40c58c161)
 
Results:

+ still missing last pixels
+ on last picture i see different amount of tGor `...118` vs `...119` /not related but not sure if expected/


image
image

image

bosomt:
- Device: model 1 1.12.2 Universal (revision e2416bcec2072baf65dda2d3c2359cc7217753bc)
 
  
Results:

+ still missing last pixels
+ on last picture i see different amount of tGor `...118` vs `...119` /not related but not sure if expected/


image
image

image

matejcik:
you're still testing the wrong screens :woman_shrugging: 

i'll make a separate issue for that so that it is clearer.
this fix only affects EIP1559 screens (which, from what we've seen previously, only work in metamask)

matejcik:
the last one 8/9 is indeed weird, but it looks more like a Suite thing? Trezor is showing `...891189?`, i.e., there is _one more digit_ while Suite only shows `...89119`, so it looks like it's rounding in the last place.

Hannsek:
@matejkriz Does Suite round long numbers? Is there any max length of the number?

matejkriz:
> @matejkriz Does Suite round long numbers? Is there any max length of the number?

Yes, id does. But the maximum decimal length on that place should be 20.

PR #2857, opened 2023-03-02
Faster device/UI tests

grdddj:
Creating a draft from the experiments I tried.

The time-saving of all these changes _could_ be around 2-3 minutes per test, when nothing breaks.

WRT to `danger_no_wipe_device` marker, it is a really extreme measure, saves `0.1-0.2` seconds from each saved wipe-initialize cycle, but it takes some development time to try where it works and where not. 

Currently, the `skip_ui` option is not being used, as it made some UI tests break - but with some effort, it should be compatible.

### DIFF
--- a/core/tools/coverage-report
+++ b/core/tools/coverage-report
@@ -28,10 +28,14 @@ fi
 
 EXCLUDES="\
 src/all_modules.py,\
+src/typing.py,\
 src/apps/ethereum/tokens.py,\
+src/apps/webauthn/knownapps.py,\
+src/apps/common/coininfo.py,\
 src/trezor/messages.py,\
 src/trezor/enums/__init__.py"
 
+# Uses core/.coveragerc configuration file
 coverage html \
   --omit="$EXCLUDES" \
   --fail-under=${COVERAGE_THRESHOLD}
--- a/core/tools/coverage-report
+++ b/core/tools/coverage-report
@@ -1,10 +1,30 @@
 #!/bin/sh
 
-COVERAGE_THRESHOLD=${COVERAGE_THRESHOLD:-0}
+readonly COVERAGE_THRESHOLD=${COVERAGE_THRESHOLD:-0}
 
-coverage run --source=./src /dev/null 2>/dev/null
-mv .coverage .coverage.empty
-coverage combine .coverage.*
+if [ ${COVERAGE_THRESHOLD} -ne 0 ]; then
+  echo "COVERAGE_THRESHOLD set to ${COVERAGE_THRESHOLD}%"
+fi
+
+# Moving coverage files from src (tests usually save it there).
+mv -v src/.coverage* . 2>/dev/null
+
+# When there are some .coverage.* files, it is a sign
+# we need to combine them into one .coverage file.
+if ls .coverage.* 2>/dev/null; then
+  # Need to create an empty coverage file containing
+  # all existing code files, to get coverage over whole src.
+  echo "Creating .coverage.empty file"
+  coverage run --source=./src /dev/null 2>/dev/null
+  mv .coverage .coverage.empty
+  echo "Combining .coverage.* files"
+  coverage combine .coverage.*
+elif [ ! -f .coverage ]; then
+  echo "ERROR: .coverage file not found"
+  exit 1
+else
+  echo "Using already existing .coverage file"
+fi
 
 EXCLUDES="\
 src/all_modules.py,\
@@ -16,10 +36,23 @@ coverage html \
   --omit="$EXCLUDES" \
   --fail-under=${COVERAGE_THRESHOLD}
 
-if [ $? -eq 2 ]; then
-  echo "Code coverage is less than ${COVERAGE_THRESHOLD}%"
+EXIT_CODE=$?
+
+RESULT=$(grep pc_cov htmlcov/index.html | egrep -o '[0-9]{1,3}%')
+
+# Catching the case when coverage is less than threshold
+if [ ${EXIT_CODE} -eq 2 ]; then
+  echo "ERROR: Code coverage is less than ${COVERAGE_THRESHOLD}% - ${RESULT}"
+  echo "See core/htmlcov/index.html for details."
   exit 1
 fi
 
-RESULT=$(grep pc_cov htmlcov/index.html | egrep -o '[0-9]{1,3}%')
-echo "COVERAGE: ${RESULT}"
+# Catching all other errors
+if [ ${EXIT_CODE} -ne 0 ]; then
+  echo "ERROR: Coverage generation failed with exit code ${EXIT_CODE} - see output above"
+  exit ${EXIT_CODE}
+fi
+
+# Happy path
+echo "SUCCESS: COVERAGE: ${RESULT}"
+echo "See core/htmlcov/index.html for details."
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import os
+from pathlib import Path
 from typing import TYPE_CHECKING, Generator
 
 import pytest
@@ -36,6 +37,10 @@ if TYPE_CHECKING:
     from _pytest.config import Config
     from _pytest.config.argparsing import Parser
     from _pytest.terminal import TerminalReporter
+
+
+HERE = Path(__file__).resolve().parent
+
 
 # So that we see details of failed asserts from this module
 pytest.register_assert_rewrite("tests.common")
@@ -83,8 +88,23 @@ def emulator(request: pytest.FixtureRequest) -> Generator["Emulator", None, None
         # 1. normal link, 2. debug link and 3. webauthn fake interface
         return 20000 + int(worker_id[2:]) * 3
 
+    # So that we can generate coverage reports
+    profiling = os.environ.get("TREZOR_PROFILING") == "1"
+    if profiling:
+        core_dir = HERE.parent / "core"
+        profiling_wrapper = core_dir / "prof" / "prof.py"
+        main_args = [str(profiling_wrapper)]
+        # So that the coverage reports have the correct paths
+        os.environ["TREZOR_SRC"] = str(core_dir / "src")
+    else:
+        main_args = ["-m", "main"]
+
     with EmulatorWrapper(
-        model, port=_get_port(), headless=True, auto_interact=not interact
+        model,
+        port=_get_port(),
+        headless=True,
+        auto_interact=not interact,
+        main_args=main_args,
     ) as emu:
         yield emu
 
--- a/tests/emulators.py
+++ b/tests/emulators.py
@@ -17,7 +17,7 @@
 import tempfile
 from collections import defaultdict
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Sequence, Tuple
 
 from trezorlib._internal.emulator import CoreEmulator, Emulator, LegacyEmulator
 
@@ -75,6 +75,7 @@ class EmulatorWrapper:
         port: Optional[int] = None,
         headless: bool = True,
         auto_interact: bool = True,
+        main_args: Sequence[str] = ("-m", "main"),
     ) -> None:
         if tag is not None:
             executable = filename_from_tag(gen, tag)
@@ -107,6 +108,7 @@ class EmulatorWrapper:
                 port=port,
                 headless=headless,
                 auto_interact=auto_interact,
+                main_args=main_args,
             )
         else:
             raise ValueError(
--- a/ci/test.yml
+++ b/ci/test.yml
@@ -57,12 +57,12 @@ core device test:
   needs:
     - core unix frozen debug build
   variables:
-    TREZOR_PROFILING: 1
+    TREZOR_PROFILING: 1  # so that we get coverage data
     MULTICORE: 4  # more could interfere with other jobs
   script:
     - $NIX_SHELL --run "poetry run make -C core test_emu_ui_multicore | ts -s"
   after_script:
-    - mv core/src/.coverage core/.coverage.test_emu
+    - mv core/src/.coverage.* core  # there will be more coverage files (one per core)
     - mv tests/ui_tests/reports/test/ test_ui_report
     - $NIX_SHELL --run "poetry run python ci/prepare_ui_artifacts.py | ts -s"
     - diff -u tests/ui_tests/fixtures.json tests/ui_tests/fixtures.suggestion.json
--- a/ci/test.yml
+++ b/ci/test.yml
@@ -285,6 +285,7 @@ core click test:
       - tests/ui_tests/fixtures.suggestion.json
       - tests/trezor.log
       - tests/junit.xml
+      - core/.coverage.*
     reports:
       junit: tests/junit.xml
     expire_in: 1 week
--- a/ci/test.yml
+++ b/ci/test.yml
@@ -58,8 +58,9 @@ core device test:
     - core unix frozen debug build
   variables:
     TREZOR_PROFILING: 1
+    MULTICORE: 4  # more could interfere with other jobs
   script:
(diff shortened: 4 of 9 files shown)


grdddj:
Pushed just the changes of the build-phase (building `core unix frozen debug build` first) and switching `core device test` to multicore.

I am expecting it will not include `test_ui_report` into the artifacts, I still need to figure out why, but want to try it - previously the `core device test` was the only test in the whole CI

grdddj:
All the previous experiment can still be seen in https://github.com/trezor/trezor-firmware/commit/e9653a808eb4808d8cdb02ceaae8c4088e36d904

matejcik:
coverage is only created when you pass the appropriate flag to the emulator (or set an envvar)
because the pytest process is in control of creating the emulators, it presumably doesn't do that, but that should be fixable: the pytest setup should look at, presumably, `TREZOR_PROFILING`, and do what emu.py does (pass the profiling wrapper as an argument, it seems??)

I think we should be able to create coverage from the multicore test, but we might need to modify the code that creates it. see `prof.py`, we could change the code to generate `.coverage.???` where ??? is taken from an env var or something.

grdddj:
> coverage is only created when you pass the appropriate flag to the emulator (or set an envvar) because the pytest process is in control of creating the emulators, it presumably doesn't do that, but that should be fixable: the pytest setup should look at, presumably, `TREZOR_PROFILING`, and do what emu.py does (pass the profiling wrapper as an argument, it seems??)
> 
> I think we should be able to create coverage from the multicore test, but we might need to modify the code that creates it. see `prof.py`, we could change the code to generate `.coverage.???` where ??? is taken from an env var or something.

I did not know about `TREZOR_PROFILING`, added support for it in [4b9e079](https://github.com/trezor/trezor-firmware/pull/2857/commits/4b9e079ab83ef3b54f77ea74e4cc1db2bcd878bd)

Tried it locally and the `.coverage` was generated even in multicore, not sure if it contains the results from all 4 processes, we will see in CI - if not, I will look into `prof.py`

grdddj:
Moved from `draft` to `Ready for review` as both I and CI are satisfied with the state.

Apart things mentioned above, I have modified the coverage script a little bit to make it more user-friendly with local testing.

Also, looking at the resulting `HTML` report, it turned out that all the `TYPE_CHECKING` branches (and the import itself) were marked as untested, therefore lowering the resulting coverage percentage. I have included `# pragma: no cover` marks to these so that they do not count into coverage.

grdddj:
Some other "false alarms" in coverage reports:
- all `_LOCAL_CONSTANTS` - e.g. https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896087449/artifacts/core/htmlcov/src_apps_webauthn_fido2_py.html
- `typing.py` - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896087449/artifacts/core/htmlcov/src_typing_py.html
- all code `from typing import ...` - e.g. https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896087449/artifacts/core/htmlcov/src_trezor_log_py.html
- `TR` UI code when running `TT` and vice versa - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896087449/artifacts/core/htmlcov/src_trezor_ui_layouts_tr___init___py.html
- `fido/knownapps.py` - long if-if dictionary - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896087449/artifacts/core/htmlcov/src_apps_webauthn_knownapps_py.html
- generally code that does not run in `__debug__` mode - e.g. https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896087449/artifacts/core/htmlcov/src_boot_py.html

In cases of `typing.py` and `fido/knownapps.py` we could silence it by putting them into `EXCLUDES` list in `coverage-report`.

After we merge `TR`, it might make sense to always disregard the UI files for the other model somehow.

Local constants (`_FIELD_TYPE_INT16 = const(1) `) might be silences with defining some appropriate regex in `exclude_lines` `coverage` option.

Current code-coverage (with all the false-alarms above) is `88 %` - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3896087449 - so we might want to increase the  `COVERAGE_THRESHOLD` to e.g. `85 %`.

grdddj:
Current coverage report: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/index.html

Some other things I noticed when going through it:
- function definitions with decorators (even `@staticmethod` or `@classmethod`) are marked untested - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_storage_sd_salt_py.html
or 
https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_bitcoin_sign_tx_tx_weight_py.html#n52
- strange things are sometimes happening with `utils.BITCOIN_ONLY` - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_bitcoin_sign_tx___init___py.html#n7
- probably a majority of currently "untested" lines are raises - extreme example is https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_nem_validators_py.html
- I am surprised that two cases are untested in `BTC:SignatureVerifier` - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_bitcoin_verification_py.html#n80
- `parse_input_script_multisig` also looks like an important function - actually included in the untested branch above -  https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_bitcoin_scripts_py.html#n500
- `set_flags` looks like an important functionality, untested - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_storage_device_py.html#n216
- can we simulate the `usbcheck`? - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_trezor_ui_layouts_tt_v2_homescreen_py.html#n71
- `coininfo.py` - another dict-like if-if chain - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_common_coininfo_py.html
- this `trezor/ui/__init__.py` looks strange - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_trezor_ui___init___py.html
- this is also strange, the `_clear_tx_request` looks like not being called - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_bitcoin_sign_tx_helpers_py.html#n366
- this could even be a bug - there is no `raise` in front of `DataError` - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_bitcoin_sign_tx_payment_request_py.html#n48
- some animation code is untested - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_bitcoin_sign_tx_progress_py.html#n125
- do not we test this in storage tests? - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_storage___init___py.html
- `read_compact_size` would be worth at least a unit-test - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_common_readers_py.html#n7
- we might have a test cancelling the abortion from recovery homepage - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_management_recovery_device_layout_py.html#n146
- some of these errors are really never instantiated, could they be deleted not to take flash space? They take `100 bytes` each https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_trezor_wire_errors_py.html
- sometimes the coverage does not make sense, there are green lines right below red ones - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_common_keychain_py.html#n65
- looks like we do not have tests for changing rotation to some degrees - probably worth doing it - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_management_apply_settings_py.html#n191
- maybe worth a quick test (same with wipe-code) - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_apps_management_change_pin_py.html#n99
- module docstrings are being flagged - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3896730236/artifacts/core/htmlcov/src_trezor_crypto_bech32_py.html#n21

Could we also somehow combine `unit tests` into it? It might test some of the edge-cases which raise a lot

These patterns could be also probably ignored:
```
pass
assert False
raise RuntimeError
raise NotImplementedError
def mem_dump
def __repr__(self)
```

It looks s though `click tests` were not included in the result - as `.coverage.test_click` is not included in the printout - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3896730236 - and it really looks like the `core/.coverage.*` is not specified in `paths` of `core click test`. Will include it.

I take back what I said about the `TR` model, when both models will be part of our CI, we will create .coverage files for both models and combine them, so nothing should be missing. (The .coverage files will need different names, however.) When running just one model locally, though, it will currently miss the other model.

ci/build.yml
```diff
@@ -479,7 +501,8 @@ legacy emu btconly debug build:
 legacy emu btconly debug asan build:
   stage: build
   <<: *gitlab_caching
-  needs: []
+  needs:
+    - core unix frozen debug build  # core tests have priority
```
matejcik, line 505: this changeset will shutdown the whole pipeline in case core unix frozen debug build is broken. let's figure out something else

tests/conftest.py
```diff
@@ -83,8 +88,23 @@ def _get_port() -> int:
...
+    # So that we can generate coverage reports
+    profiling = os.environ.get("TREZOR_PROFILING") == "1"
+    if profiling:
+        core_dir = HERE.parent / "core"
+        profiling_wrapper = core_dir / "prof" / "prof.py"
+        main_args = [str(profiling_wrapper)]
+        # So that the coverage reports have the correct paths
+        os.environ["TREZOR_SRC"] = str(core_dir / "src")
+    else:
+        main_args = ["-m", "main"]
```
matejcik, line 100: i _think_ we can keep the main_args empty in this case? this `-m main` is a compatibility thing IIRC and we shouldn't be needing it anymore

tests/emulators.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         port: Optional[int] = None,
         headless: bool = True,
         auto_interact: bool = True,
+        main_args: Sequence[str] = ("-m", "main"),
```
matejcik, line 78: dtto

core/tools/coverage-report
```diff
@@ -1,25 +1,72 @@
...
-coverage run --source=./src /dev/null 2>/dev/null
-mv .coverage .coverage.empty
-coverage combine .coverage.*
+if [ ${COVERAGE_THRESHOLD} -ne 0 ]; then
+  echo "COVERAGE_THRESHOLD set to ${COVERAGE_THRESHOLD}%"
+fi
+
+# Optionally move coverage files from src
+# (tests usually save it there).
+if [ ${MOVE_FROM_SRC} -eq 1 ]; then
```
matejcik, line 14: why wouldn't we do this? isn't the "if ls" check sufficient?

core/tools/coverage-report
```diff
@@ -1,25 +1,72 @@
...
+    mv src/.coverage* .
+  fi
+fi
+
+# When there are some .coverage.* files, it is a sign
+# we need to combine them into one .coverage file.
+if ls .coverage.* 2>/dev/null; then
+  # Need to create an empty coverage file containing
+  # all existing code files, to get coverage over whole src.
+  if [ ${WHOLE_SRC_COVERAGE} -eq 1 ]; then
```
matejcik, line 26: in what situation would we want to skip this step?

ci/build.yml
grdddj: It would crash the whole pipeline, but when `frozen debug build` is broken, then the majority of the tests will not run anyway.

I am not against reverting this, it is really not so helpful now.

core/tools/coverage-report
matejcik: > are you suggesting completely removing the if statement?

yes, i'm not a fan of configuration options covering scenarios that we don't ever encounter :)

in the tests.yml case, you shouldn't have any coverages in the `core` subdir, so the "if ls" check covers you there; strictly speaking, you could just run the `mv src/.coverage.* . 2>/dev/null` and ignore any errors, right?

core/tools/coverage-report
matejcik: istm in such scenario an output that tells you the list of items that you are _not_ currently covering is overall more correct?

ci/build.yml
matejcik: please revert it then.

you're right about the "majority" of tests, but it's going to be annoying especially for commits that do something to legacy build while the core is not-yet-updated (or similar with btc-only variant)

core/tools/coverage-report
grdddj: Actually, in tests, we _have_ all the coverage files in `core` already, as each test moves it there from `src` in the artifacts - so nothing would be moved in that case.

Just `mv src/.coverage.* . 2>/dev/null` is certainly doable, but then we lose `echo` confirming the move was done, and more importantly the output of `ls` telling us what files were moved. Probably not vital, but if we can give this information, why hide it.

matejcik: yes, sorry, i meant `core/src`, because the local checkout didn't run any tests so everything we get must be coming from the artifacts

> then we lose echo confirming the move was done

`mv -v`? :)

(i don't have a problem with the "if ls" check per se, but now it turns out that `mv -v ...` would actually do mostly the same thing you want, so :woman_shrugging: )

tests/emulators.py
grdddj: Changed in [7543c1d](https://github.com/trezor/trezor-firmware/pull/2857/commits/7543c1d620a59dd353638d7254e3832d35c8ea12)

tests/conftest.py
grdddj: Changed in [7543c1d](https://github.com/trezor/trezor-firmware/pull/2857/commits/7543c1d620a59dd353638d7254e3832d35c8ea12) - works fine

ci/build.yml
grdddj: Reverted in [d8b5118](https://github.com/trezor/trezor-firmware/pull/2857/commits/d8b5118ec6e4b3680548f5641afe490ea60071ff)

core/tools/coverage-report
grdddj: Deleting the variable and usage of `ls` in [f246a1f](https://github.com/trezor/trezor-firmware/pull/2857/commits/f246a1f5f5a9f1ac1251e9026add3358696b9bef)

matejcik (approved):

grdddj:
Turned out the omission of `-m main` was working fine with _current_ emulators, it was not fine with _old_ emulators, and therefore the `upgrade tests` were failing.

I have put again the default value into `EmulatorWrapper` - `main_args: Sequence[str] = ("-m", "main")`

PR #2870, opened 2023-03-09
Show app icon in Fido2ConfirmNoCredentials

andrewkozlik:
This should fix the missing logo reported in https://github.com/trezor/trezor-firmware/pull/2844#issuecomment-1443747492
> first FIDO2 screen does not display Google logo

For reasons stated in https://github.com/trezor/trezor-firmware/pull/2844#issuecomment-1443823464, it does not fix:
> you have to confirm login twice FIDO2 and then U2F


matejcik (approved):
utACK

Issue #11, opened 2018-07-17
bootloader: jump_to_firmware can lead to undefined behavior


mcudev:
currently the `jump_to_firmware` function can lead to undefined behavior if an exception is taken after `SCB_VTOR` is set and before the firmware's `reset_handler` has completed. what i mean is that the firmware exception handlers are written in C and the sections that get initialized in SRAM are put there by the `reset_handler`. this is an informational issue right now because i don't see this being exploitable with the firmware's current set of exception handlers. the model t already considered this possibility and handles it differently.

to see this in action in a dev board, set a break point at https://github.com/trezor/trezor-mcu/blob/master/util.h#L65 and touch the reset pin on your board to cause clock instability which will trigger the `nmi_handler` for the clock security system. since the firmware's vectors are now in use, when you step in the debugger, you'll be transported to the exception handler at the firmware's flash memory address. thus, we're running firmware code before completing the bootloader and the subsequent firmware initialization.

EDIT: adding a note that the model t also clears out the registers just in-case anything sensitive was in them. reference https://github.com/trezor/trezor-core/blob/master/embed/trezorhal/util.s

tsusanka:
@prusnak is this still up to date? Is it something we would like to include in the next bootloader release?

prusnak:
@tsusanka Let's have @hiviah assess the issue.

mcudev:
Perhaps the most that is needed is to add a comment that warns of this possibility in the legacy T1 code where this informational issue could apply: `trezor-firmware/legacy/util.h` function `jump_to_firmware`.

The TT code already avoids this possible undefined behavior, so I'm not worried about it there.
Other than that, this can probably be closed. At the time that I thought of this less than obvious execution possibility, opening an issue was just the way to communicate it that came to mind.
Mitigating steps already in-place to reduce probability of this becoming a problem include:
1. `jump_to_firmware` is already called after the unofficial firmware user warning
1. `jump_to_firmware` already transitions control to untrusted firmware after mpu config
1. unless i missed it, i don't see the `SCB_VTOR` being changed from the bootloader's handlers for untrusted firmware. this would limit potential impacts to only officially signed firmware.
1. sram is zeroed before calling `jump_to_firmware`
1. i haven't seen any existing handlers in the officially signed firmware use data sections that may not be initialized yet (which is the biggest concern that can lead to undefined behavior). handlers don't change often either. so, it's mainly a don't shoot-yourself-in-the-foot kind of thing, i think.

[...]

Issue #10, opened 2018-08-21
Validate BIP32 paths according to BIP44, etc.


yura-pakhuchiy:
https://github.com/trezor/trezor-core/issues/326
I believe this should be fixed for Trezor 1 as well.

matejcik:
I think we can close this in favor of the core port, what do you think @tsusanka @prusnak ?

prusnak:
ACK

PR #2871, opened 2023-03-09
Improve protobuf varint handling

matejcik:
an unintentional panic was discovered by a pentest by cure53:

> During testing, it was discovered that the Trezor firmware is vulnerable to DoS attacks due to a flaw in the Rust Protobuf handling. A malicious user could leverage this vulnerability by sending a malformed USB packet to the Trezor firmware, which would result in a crash of the Trezor HSM. The issue was specifically found by a custom fuzzer which leverages the AFL++ framework.
>
> (...)
>
> While using the AFL++ fuzzer for testing, approximately 50 unique crashes emerged out of a total of 2 million executions performed with this test harness. All of the identified crashes had similar stacktraces. The following is an example of the stacktrace observed in these crashes:
>
> ```
> Program received signal SIGABRT, Aborted.
> __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
> 50 ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
> (gdb) bt
> #0 __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
> #1 0x00007ffff6777859 in __GI_abort () at abort.c:79
> #2 0x0000555555bdec22 in __fatal_error (expr=<optimized out>, msg=<optimized out>, file=<optimized out>, line=<optimized out>, func=<optimized out>) at embed/unix/common.c:83
> #3 0x000055555604f6d6 in trezor_lib::trezorhal::common::__fatal_error ()
> #4 0x00005555560736cf in rust_begin_unwind ()
> #5 0x0000555555bdec36 in core::panicking::panic_fmt () at library/core/src/panicking.rs:64
> #6 0x0000555555bded2c in core::panicking::panic () at library/core/src/panicking.rs:111
> #7 0x000055555605e68d in trezor_lib::protobuf::decode::InputStream::read_uvarint ()
> #8 0x0000555556054ee8 in trezor_lib::protobuf::decode::Decoder::message_from_stream ()
> #9 0x0000555556053ce1 in protobuf_decode ()
> #10 0x000055555604037b in fuzz (buffer=<optimized out>, len=<optimized out>) at embed/unix/harness.c:49
> #11 0x0000555555be3659 in main (argc=<optimized out>, argv=<optimized out>) at embed/unix/harness.c:103
> ```
>
> The trace indicates that a panic occurred in `trezor_lib::protobuf::decode::InputStream::read_uvarint` because of an overflow error, resulting in a fatal error that would render the device unusable.
>

As a result, a malicious application can cause the Trezor firmware to freeze.

Using Rust helps us here because the overflow triggers a panic (i.e., a shutdown), so an invalid value cannot propagate out.
(with that said, in this particular place, an attacker could not use the overflow for anything useful; the output would be a mis-decoded 64 bit number that the same attacker could instead just encode directly)

---

This PR fixes the panic and adds a testcase. I'm keeping it as a draft because there is space for further improvement here.

Namely:

* we should differentiate between (u/s)int32 and (u/s)int64
* we should have a special 32-bit decoder call
* we might want to try to use the 32-bit call even in 64-bit cases to improve performance (STM32 is a 32bit CPU) and only upgrade to the 64-bit code in case of overflow

### DIFF
--- a/core/embed/rust/src/protobuf/decode.rs
+++ b/core/embed/rust/src/protobuf/decode.rs
@@ -315,12 +315,24 @@ impl<'a> InputStream<'a> {
     pub fn read_uvarint(&mut self) -> Result<u64, Error> {
         let mut uint = 0;
         let mut shift = 0;
-        loop {
+        let mut last_byte = true;
+        while shift <= 64 - 7 {
+            // Shifting by 64 - 7 and then adding 7 bits is always safe.
             let byte = self.read_byte()?;
             uint += (byte as u64 & 0x7F) << shift;
             shift += 7;
             if byte & 0x80 == 0 {
+                last_byte = false;
                 break;
+            }
+        }
+        if last_byte {
+            // After reading 9 bytes, there is only one bit remaining to be set.
+            let byte = self.read_byte()?;
+            if byte > 1 {
+                return Err(Error::OutOfRange);
+            } else {
+                uint += (byte as u64) << shift;
             }
         }
         Ok(uint)
--- a/core/tests/test_trezor.protobuf.py
+++ b/core/tests/test_trezor.protobuf.py
@@ -1,7 +1,12 @@
 from common import *
 
 from trezor import protobuf
-from trezor.messages import WebAuthnCredential, Failure, SignMessage, DebugLinkMemoryRead
+from trezor.messages import (
+    WebAuthnCredential,
+    Failure,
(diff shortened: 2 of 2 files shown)


matejcik:
should be superseded by #3521

PR #2877, opened 2023-03-11
Add Nostr to misc.json

alexgleason:
This will enable Trezor users to sign Nostr events: https://github.com/nostr-protocol/nips/blob/master/06.md

### DIFF
--- a/common/defs/misc/misc.json
+++ b/common/defs/misc/misc.json
@@ -175,5 +175,16 @@
             "Homepage": "https://tether.to"
         },
         "wallet": {}
+    },
+    {
+        "name": "Nostr",
+        "shortcut": "NOSTR",
+        "slip44": 1237,
+        "curve": "secp256k1",
+        "decimals": 0,
+        "links": {
+            "Github": "https://github.com/nostr-protocol/nostr"
+        },
+        "wallet": {}
     }
 ]


prusnak:
No, this PR will not enable anything. The file basically is just a description what already has been implemented.

alexgleason:
Thanks for the quick reply. Can you point me to the correct place to add this? I'm able to call `TrezorConnect.getPublicKey()` and it works fine, but `TrezorConnect.signMessage()` complains that it doesn't know the coin for Nostr, so I think it needs to be added to coins.json somehow.

alexgleason:
A firmware change shouldn't be needed(?) for what I'm trying to do. Nostr keypairs are the same as Bitcoin, so maybe this should go under `bitcoin/` instead of `misc/`. It's just a different path (`m/44'/1237'/0'/0/0`). All I'm trying to do is bypass the error _"Forbidden key path"_ from TrezorConnect.

I could just pretend to be Bitcoin instead, but I think users would rather use the correct path.

prusnak:
Are you sure nostr uses the Bitcoin signature scheme?

alexgleason:
Yeah, I just able to successfully post a message signed with my Trezor: https://iris.to/note1mxm5x7hyswg39wapsn6rgjv0pzxt8qvfu2sufl05gehy6xs5chsqywasvd

Rather than pretending to be Bitcoin I just disabled security features on the Trezor and it worked.

alexgleason:
Actually, the signature is 130 bytes long when it should be 128, so maybe I'm wrong. The pubkey needs the first two bytes chopped off too.

alexgleason:
Finally figured out it's not currently possible. It wants a Schnorr signature, which I assumed was already the standard but isn't implemented here. Also I think Trezor might invisibly prepend extra text to the message, not sure. This sums it up: https://www.reddit.com/r/TREZOR/comments/vrftwn/comment/iexubo7/

Appreciate your time.

Issue #2173, opened 2022-03-11
Upgrade to isort 5


matejcik:
https://pycqa.github.io/isort/docs/major_releases/introducing_isort_5.html
https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0.html

It might be possible to use one of the new isort profiles instead of our custom configuration.

AdarshRawat1:
Can I work on this issue?

matejcik:
please go ahead!

AdarshRawat1:
Could you provide some direction on where to start?

matejcik:
I would start by upgrading isort in poetry, removing the current isort config completely (in setup.cfg i think?) and replacing it by a profile. I didn't look into which profiles are available, but I vaguely remember that there's a `black` based one, so that's where I would start.

then I'd try to run the `make pystyle` task and see what changes show up. the ideal result is no changes, in which case you're basically done.

what might happen is that the new isort will not run at all, in which case you'll need to modify the invocation (in top-level makefile and in python/Makefile) so that it works again

AdarshRawat1:
Please guide me on how I can run the `make pystyle` task.

AdarshRawat1:
> This _looks like_ windows has a problem with all our files specified with forward slashes, so there's a good chance this won't work on Windows at all. If you're willing to go the extra mile, try to do it in WSL.

Noted ! Will do it in WSL.Thanks.

matejcik:
something is wrong with your setup because no arguments are passed to isort/black

first of all you should be using `poetry` (see the last section [here](https://docs.trezor.io/trezor-firmware/core/build/index.html#poetry))

second, the makefile `PY_FILES` variable is empty
try running this command, which fills it:
`find . -type f -name '*.py'   | sed 'sO^\./OO' | grep -f ./tools/style.py.include | grep -v -f ./tools/style.py.exclude`

AdarshRawat1:
@matejcik I executed this command after following all your guidance and resources.
![image](https://user-images.githubusercontent.com/100958893/226417887-b9d35dc3-d7e9-48e7-aebf-eaa12231b166.png)

After this I tested if other changes are remaining - 
1.
![image](https://user-images.githubusercontent.com/100958893/226418013-84842d90-38cb-4776-93d4-6f517c369f3c.png)

2.
![image](https://user-images.githubusercontent.com/100958893/226418238-081e565f-1c3e-4470-aec9-cf2aa76752b1.png)

[...]

PR #2893, opened 2023-03-20
feat: Upgrade to isort 5.12.0

AdarshRawat1:
I have used the isort Python utility/library to sort imports alphabetically and automatically separate them into sections and by type. We have also used the black profile to format the code.


AdarshRawat1:
#2173

ci/common_sync/git-filter-repo
matejcik, line 1: please revert this

core/tests/test_apps.bitcoin.address.py
```diff
@@ -1,18 +1,20 @@
...
 from apps.bitcoin.keychain import validate_path_against_script_type
 from apps.bitcoin.writers import *
+from apps.common import coins
+from trezor import wire
+from trezor.crypto import bip32, bip39
+from trezor.enums import InputScriptType
+from trezor.messages import GetAddress
+from trezor.utils import HashWriter
+
+from common import *
```
matejcik, line 17: we actually need "common" to be first. but maybe this is not worth worrying about, just reverting the whole of `core/tests` is good enough

core/tests/production_tests/main.py
```diff
@@ -1,7 +1,6 @@
-import utime
-
 import trezorio as io
 import trezorui as ui
+import utime
```
matejcik, line 3: utime should also be in the "system builtins" group

tests/bip32.py
```diff
@@ -24,7 +24,6 @@
 from ecdsa.curves import SECP256k1
 from ecdsa.ellipticcurve import INFINITY, Point
 from ecdsa.util import number_to_string, string_to_number
-
 from trezorlib import messages, tools
```
matejcik, line 27: generally we want "trezorlib" to always be in the "1st party" section or whatever it is called, i.e., separated from click / pytest / ecdsa / etc.
there should be no diff here and in 90 % of `tests/`

setup.cfg
```diff
@@ -19,17 +19,6 @@ ignore =
...
-[isort]
-multi_line_output = 3
-include_trailing_comma = True
-force_grid_wrap = 0
-combine_as_imports = True
-line_length = 88
-not_skip=__init__.py
-known_standard_library = micropython,ubinascii,ustruct,uctypes,utime,utimeq,trezorio,trezorui,trezorutils,trezorconfig,trezorcrypto
-known_third_party = curve25519,ecdsa,hypothesis,pytest,mnemonic,shamir-mnemonic,trezor,storage,protobuf
-known_first_party = trezorlib,apps
```
matejcik, line 31: it looks like the new isort is good enough to figure out the "3rd party" stuff on its own, but we would like to keep both `known_standard_library` and `known_first_party`, see the other comments

matejcik (changes requested):
thank you for your work, this is shaping up nicely.

please look into tweaking the isort config as indicated, and revert some of the changes. it might actually be better if you only include files covered by `make pystyle` (as opposed to every .py file in the project)

AdarshRawat1:
@matejcik  I have made all the suggested changes.Please review!

AdarshRawat1:
@matejcik 
I couldn't make necessary changes on master branch as there were too many files, so I created a separate PR from different branch.

Issue #2865, opened 2023-03-07
Edit output from emulator to match colours in UI2


Hannsek:
Gamma correction would be needed.

This can be done after #2835.

grdddj:
Proof of concept done in https://github.com/trezor/trezor-firmware/commit/0ccf916c3744a1e0f86ba9fdde15adc566aafea0
- building on top of https://github.com/trezor/trezor-firmware/pull/2876 to have the latest designs
- increasing the brightness of colors in the emulator
- was trying to do it with gamma correction, but was not successful
- there are a lot of ways/algorithms how to increase the color brightness and there is hardly one that will work perfectly for all colors (from almost-black to almost-white)
  - currently, I am using this way
    - the maximum values of R, G and B are 31, 63 and 31 respectively (based on RGB565)
    - take the difference between current value and its maximum value, if R is 15, the diff is 31 - 15 = 16
    - multiply this difference by some percentage `INCREASE_BRIGHTNESS_PERCENT` in code, defaulting at `25`, getting 16*0.25 = 4
    - adding this to the current value, ending up at 15 + 4 = 19
- by pressing left and right arrows when in the emulator, the  `INCREASE_BRIGHTNESS_PERCENT` is getting decreased/increased by one percentage point (and printed into the terminal), so we can experiment with its various values (personally I liked the range between 20-25) - @Hannsek it it possible to play with it on `grdddj/emu_colors` branch

The current screens are visible at https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3967679098/artifacts/test_ui_report/all_unique_screens.html - compared to older ones in https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/3943227735/artifacts/test_ui_report/all_unique_screens.html

Other possibilities
- have some other "brightness-increasing" algorithm
- hard-coding the color transition for the emulator (there are not so many main colors actually)

prusnak:
> * was trying to do it with gamma correction, but was not successful

What does this mean? That you were not able to do it or that you did it but the results were not great?

grdddj:
> > * was trying to do it with gamma correction, but was not successful
> 
> What does this mean? That you were not able to do it or that you did it but the results were not great?

[...]

PR #2894, opened 2023-03-20
Emulator gamma correction

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2865:
- performs gamma correction on the emulator display, so the colors look more like on the physical device
- allows for changing the gamma value by left and right arrows


matejcik (approved):
utACK, seems right

grdddj:
Regenerated `fixtures.json` for the changed colors.

With @Hannsek we considered hardcoding the emulator colors (replacing them in `display_pixeldata`), which would give us the best colors, but at the expense of more complexity and mainly maintenance. So we decided this gamma-correction is good enough.

Will merge this after CI is green.

PR #2896, opened 2023-03-21
fix(core/bootloader): fix retries when reading from USB

TychoVrahe:


matejcik (approved):
utACK

prusnak (approved):
utACK

Any ideas how to test?

TychoVrahe:
tested this by disconnecting one wire of usb data line during firmware upload. 
- with the fix the error is shown after a while
- without the fix the device get stuck on the Installing firmware loader

Though this depends a bit on when exactly the communication breaks - the stuck can occur even after fix, if the bootloader waits for message outside of this retry function.

TychoVrahe:
Actually it appears that the stuck loader is moving very slowly and eventually it eventually shows another error when trying to send answer over the usb

bosomt:
Tried to reinstall 2.6.0/2.1.0 to same version so i cut one of data cables on USB port but loader never timed out and it got stuck in this screen forever.

image


image

Issue #2732, opened 2023-01-11
Use new artifact naming for legacy builds


matejcik:
to keep in sync with core builds (#2731), we need to rename the legacy build artifacts to `firmware-T1B1[-btconly]-<version>.bin`

mmilata:
#2731 also renames legacy artifacts, so we can close this unless we also want these names for locally built images?

matejcik:
> also want these names for locally built images?

yes, in particular for outputs of build-docker.sh

PR #2900, opened 2023-03-21
feat(legacy): use new artifacts naming for legacy builds

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2732:
- exports legacy firmware build under a descriptive name, for example `firmware-T1B1-bitcoinonly-1.12.2-cc5233e57-dirty.bin`
- adds this export also to `build-docker.sh`, where we need to account for older tags/branches not having this, and therefore ignoring the possible error


matejcik (approved):
utACK

PR #2901, opened 2023-03-21
feat(core): support old homescreen format

TychoVrahe:


core/embed/rust/src/ui/model_tt/component/homescreen/mod.rs
```diff
@@ -194,16 +203,40 @@ where
...
+                    let mut input = BufferInput(data.as_ref());
+                    let mut hs_img = HomescreenJpeg::new(&mut input);
+                    homescreen(
+                        &mut hs_img,
+                        &[text],
+                        notification,
+                        self.paint_notification_only,
+                    );
+                    show_default = false;
+                } else if data_ok && is_image_toif(data.as_ref()) {
```
mmilata, line 220: IMO `data_ok` is always true since we're inside `if let Ok(...) = res {` block and can be removed.

core/embed/rust/src/ui/model_tt/component/homescreen/mod.rs
```diff
@@ -290,10 +323,29 @@ where
...
         if let Ok(data) = res {
-            homescreen_blurred(data.as_ref(), &texts);
-        } else {
-            homescreen_blurred(IMAGE_HOMESCREEN, &texts);
+            if is_image_jpeg(data.as_ref()) {
+                let mut input = BufferInput(data.as_ref());
+                let mut hs_img = HomescreenJpeg::new(&mut input);
+                homescreen_blurred(&mut hs_img, &texts);
+                show_default = false;
+            } else if data_ok && is_image_toif(data.as_ref()) {
```
mmilata, line 335: Same here.

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
```diff
@@ -78,6 +82,99 @@ const RED_IDX: usize = 0;
...
+    }
+}
+
+impl<'i> HomescreenDecompressor for HomescreenToif<'i> {
+    fn get_height(&self) -> i16 {
+        1
+    }
+
+    fn decompress(&mut self) {
+        //self.jdec.as_mut().map(|dec| dec.decomp(&mut self.output));
```
mmilata, line 147: Leftover?

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
```diff
@@ -78,6 +82,99 @@ const RED_IDX: usize = 0;
...
+        // SAFETY: Aligning to u8 slice is safe, because the original slice is aligned
+        // to 16 bits, therefore there are also no residuals (prefix/suffix).
+        // The data in the slices are integers, so these are valid for both u16
+        // and u8.
+        if self.line >= HOMESCREEN_TOIF_Y_OFFSET
+            && self.line < HOMESCREEN_TOIF_Y_OFFSET + HOMESCREEN_TOIF_SIZE
+        {
+            let (_, workbuf, _) = unsafe { self.output.buffer().buffer.align_to_mut::<u8>() };
+            unwrap!(self.decomp_context.uncompress(
```
mmilata, line 156: Will this RSOD on malformed/short image data? Might be better to just draw black pixels in this case, that way device still remains usable?

mmilata (commented):
Looks good but would like to request testcase for this feature, and also for replacing toif->jpeg and jpeg->toif.

mmilata:
> replacing toif->jpeg and jpeg->toif

Right, we don't support that, still a testcase that renders toif homescreen would be nice.

core/embed/rust/src/ui/model_tt/component/homescreen/mod.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/efad3596e8d2fc3c208e13384990bda439334289

core/embed/rust/src/ui/model_tt/component/homescreen/mod.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/efad3596e8d2fc3c208e13384990bda439334289

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/efad3596e8d2fc3c208e13384990bda439334289

core/embed/rust/src/ui/model_tt/component/homescreen/render.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/efad3596e8d2fc3c208e13384990bda439334289

TychoVrahe:
Yeah current firmware doesn't support uploading toif. I would need to either enable it in debug, or a way to preinitialize storage with old data - not sure if there is a way to do that.

mmilata:
I guess extending `LoadDevice` would be the easiest.

matejcik:
you can either stuff it into upgrade_tests, and pick an old firmware that lets you install the homescreen
or go with persistence tests and just feed the emulator a pregenerated flash image

neither has UI tests currently but it should be easy to enable them for persistence_tests i think

bosomt:
Upgraded from to.

2.5.2 0d87b55b
f2.6.0 6e0c2578

Worked as expected /except that no backup or no pin overlays old homescreen/

image

PR #2903, opened 2023-03-23
Grdddj/emu gamma correction fix

grdddj:
Reverting the `fixtures.json` changes and improving the way of gamma-correcting the emulator display:
- creating a second buffer, `BUFFER_TO_DISPLAY`, that will be used to show the data
- it will be filled with `BUFFER` data and then all the pixels will be gamma-corrected
- doing the gamma-correction just for `TT` and not in case when the `SDL_VIDEODRIVER == dummy`
- nice side-effect is that the adjustments of gamma-correction value by keyboard arrows are immediately visible, without the need of changing the page (`display_refresh` is called each time the value updated)


matejcik (approved):
utACK

grdddj:
Storage test failed, but I believe it cannot have connection with the changes - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/3990558419 (OSError: /builds/satoshilabs/trezor/trezor-firmware/storage/tests/c/libtrezor-storage.so: undefined symbol: random32)

PR #2904, opened 2023-03-23
fix(cli): don't try to initialize device after wipe in bootloader

TychoVrahe:
what title says.

### DIFF
--- a/python/.changelog.d/2904.fixed
+++ b/python/.changelog.d/2904.fixed
@@ -1,0 +1,1 @@
+Removed attempt to initialize the device after wipe in bootloader mode
--- a/python/src/trezorlib/device.py
+++ b/python/src/trezorlib/device.py
@@ -101,7 +101,8 @@ def sd_protect(
 @session
 def wipe(client: "TrezorClient") -> "MessageType":
     ret = client.call(messages.WipeDevice())
-    client.init_device()
+    if not client.features.bootloader_mode:
+        client.init_device()
     return ret
 
 


matejcik:
I suspect it would be cleaner to try/except the `init_device` call and ignore any exception, no?

TychoVrahe:
i wouldnt necessarily call trying something that we know will fail and handling the exception cleaner, but mainly i dont know that its gonna fail with exception (it does for current USB bootloader since we deinit on the device, but current BLE implementation just doesnt answer and cli hangs)

matejcik:
> trying something that we know will fail and handling the exception

The point is that the caller isn't supposed to think about whether the implicit reinitialization will work or not; neither is there a reason to skip reinit unless you know in advance that it will fail.

> current BLE implementation just doesnt answer and cli hangs

this is an issue only in the bootloader, right?
so we could do "if client is bootloader then do not init"

TychoVrahe:
ok moved the logic here https://github.com/trezor/trezor-firmware/commit/42af4086b48b6a2ed13a5b62d8f1ee1e346638b6

matejcik:
perfect. please add a changelog entry too.

wondering what the failing upgrade test is about, it is unrelated here but...?
maybe a rebase will help?

matejcik (approved):

matejcik:
yes please:
run the bootloader wipe (`trezorctl device wipe -b`) over straight webusb and over bridge
check that in previous version this would cause an exception trace
check that it does not happen anymore with master version of trezorctl

bosomt:
**QA OK**

wiping 2.6.0/2.1.0 device:

**bridge running**

trezorctl 0.13.6
```
 ~/Trezor/trezor-firmware  on master !1  trezorctl wipe-device -b                                                                             trezor-firmware 
Wiping user data and firmware!
```

trezorctl 0.13.5
```
 ~/Trezor/trezor-firmware  on master !1  trezorctl wipe-device -b                                                                                                 
Wiping user data and firmware!
Error: trezord: read/16 failed with code 400: device disconnected during action
```

**bridge stopped**

trezorctl 0.13.6
```
 ~/Trezor/trezor-firmware  on master !1  trezorctl wipe-device -b                                                                              trezor-firmware 
Wiping user data and firmware!
```

trezorctl 0.13.5
```
 ~/Trezor/trezor-firmware  on master !1  trezorctl wipe-device -b                                                                                                
Wiping user data and firmware!
usb1.USBErrorIO: LIBUSB_ERROR_IO [-1]
```

PR #2908, opened 2023-03-26
isort 

AdarshRawat1:
[Reverted few files] #2893

I have used the isort Python utility/library to sort imports alphabetically and automatically separate them into sections and by type. We have also used the black profile to format the code.


core/src/apps/bitcoin/sign_tx/bitcoin.py
```diff
@@ -1,13 +1,12 @@
-from micropython import const
 from typing import TYPE_CHECKING
 
+from apps.common.writers import write_compact_size
+from micropython import const
```
matejcik, line 4: this is still not right. this change (and others in similar vein) should not exist. it looks like your config changes do not get picked up by isort, or maybe the options are called something else in the new version, or...?

AdarshRawat1: It is [as you suspected](https://pycqa.github.io/isort/docs/configuration/options.html#known-third-party:~:text=%2D%2Dfuture-,Known%20Third%20Party,-Force%20isort%20to) the options need to be defined in `.isort.cfg` or `pyproject.toml` file. I'll go ahead and add it to  the `pyproject.toml` file . 

Should I create a `.isort.cfg` file instead ?

core/src/apps/base.py
```diff
@@ -1,25 +1,24 @@
-from typing import TYPE_CHECKING
-
 import storage.cache as storage_cache
 import storage.device as storage_device
 from trezor import config, utils, wire, workflow
 from trezor.enums import HomescreenFormat, MessageType
 from trezor.messages import Success, UnlockPath
+from typing import TYPE_CHECKING
```
matejcik, line 6: this change should not be here

core/emu.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 from __future__ import annotations
 
+import click
```
matejcik, line 4: this change should not be here

core/mocks/generated/trezorconfig.pyi
```diff
@@ -1,6 +1,5 @@
 from typing import *
 
-
```
matejcik, line 3: as I said before, this whole directory should be excluded. in general, let's stick to what `make pystyle` produces and do not include any changes that are not touched by `make pystyle`

core/src/apps/bitcoin/authorize_coinjoin.py
```diff
@@ -1,14 +1,16 @@
 from micropython import const
+
```
matejcik, line 2: this should not be here

legacy/bootloader/firmware_sign_dev.py
```diff
@@ -1,9 +1,8 @@
 #!/usr/bin/env python3
 import argparse
+import ecdsa
```
matejcik, line 3: this change should not be here

matejcik (changes requested):
There are still unwanted changes.

As a general rule, we do not want any changes in top-level imports that were sorted "correctly" with the previous version. Reordering is allowed (a) in functions, (b) in "if type_checking" and other conditional blocks, and (c) in try blocks -- because those weren't managed by isort previously, so there might be mistakes.

I recommend going through https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0.html because it appears that some configuration options are not working as they used to.

core/src/apps/base.py
AdarshRawat1: > There are still unwanted changes.
> 
> As a general rule, we do not want any changes in top-level imports that were sorted "correctly" with the previous version. Reordering is allowed (a) in functions, (b) in "if type_checking" and other conditional blocks, and (c) in try blocks -- because those weren't managed by isort previously, so there might be mistakes.
> 
> I recommend going through https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0.html because it appears that some configuration options are not working as they used to.

Understood. I found the error. 
-[Changes](https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0.html#:~:text=isort%20settings%20no%20longer%20merge%20together%2C%20instead%20they%20override.) 
- [Solution
](https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0.html#:~:text=you%20can%20still%20get%20this%20behavior%20by%20moving%20your%20extra%20modules%20from%20the%20known_standard_library%20setting%20to%20extra_standard_library.)

AdarshRawat1:
@matejcik Please review, I have changed the configuration options according to new standards of isort. I have applied the new settings.

matejcik:
the config is looking nicely now. please have look at the failing CI check and fix the problem.

Next, let's talk about integration. Now that we have a +- okayish version of the isort config, we need to upgrade to isort in the preconfigured environments. That means bumping the isort version in `pyproject.toml`, regenerating poetry.lock via `poetry lock --no-update`, and seeing if `make pystyle` uses the right isort: e.g., right now, you can see [here](https://github.com/trezor/trezor-firmware/actions/runs/4577335788/jobs/8083408188?pr=2908) that the old isort 4.x is still in use in CI.
(notably, you will need to make sure you have poetry `1.1.x` to generate the right lockfile; we have a PR in the pipeline to update the version but it's not ready yet)

When that is done, you can check that it works by changing the order of imports somewhere, running `make pystyle`, and seeing that it goes back to the right version.

When you have _that_ down, I would like you to revert _all changes_ except for the config (that is, `pyproject.toml` + `setup.cfg` + `poetry.lock`), then run `make pystyle`, and commit _only_ the changes that were produced by `make pystyle`. This will, hopefully, take care of the problem with isort touching files that it shouldn't.

In particular: if `make pystyle` changes any file in `core/tests` subdirectory, that indicates a problem with how parameters are passed to `isort`, which we'll need to resolve.

AdarshRawat1:
> the config is looking nicely now. please have look at the failing CI check and fix the problem.
> 
> Next, let's talk about integration. Now that we have a +- okayish version of the isort config, we need to upgrade to isort in the preconfigured environments. That means bumping the isort version in `pyproject.toml`, regenerating poetry.lock via `poetry lock --no-update`, and seeing if `make pystyle` uses the right isort: e.g., right now, you can see [here](https://github.com/trezor/trezor-firmware/actions/runs/4577335788/jobs/8083408188?pr=2908) that the old isort 4.x is still in use in CI. (notably, you will need to make sure you have poetry `1.1.x` to generate the right lockfile; we have a PR in the pipeline to update the version but it's not ready yet)
> 
> When that is done, you can check that it works by changing the order of imports somewhere, running `make pystyle`, and seeing that it goes back to the right version.
> 
> When you have _that_ down, I would like you to revert _all changes_ except for the config (that is, `pyproject.toml` + `setup.cfg` + `poetry.lock`), then run `make pystyle`, and commit _only_ the changes that were produced by `make pystyle`. This will, hopefully, take care of the problem with isort touching files that it shouldn't.
> 
> In particular: if `make pystyle` changes any file in `core/tests` subdirectory, that indicates a problem with how parameters are passed to `isort`, which we'll need to resolve.

Thank you @matejcik, I have noted your valuable feedback. on it !

AdarshRawat1:
@matejcik am I allowed to make this change?
image

matejcik:
yes, please go ahed with what the solver is proposing

Hannsek:
@AdarshRawat1 are you going to continue working on this?

AdarshRawat1:
@Hannsek   I am working on resolving some issues on this PR, but my exams are taking up most of my time.

Hannsek:
Thanks, in case you wouldn't continue on this, please let us know.

Wishing you a good luck with the exams. :)

AdarshRawat1:
@Hannsek Thank you for your good luck wishes for my exams. I hope you are doing well too. 

I will let you know if I face any difficulties or delays. I value your feedback and cooperation on this project. Is this task urgent or do you have a deadline in mind?

Hannsek:
It is not urgent. 

Therefore, I will change it from pr to draft.

Hannsek:
@AdarshRawat1 Hey, are you planning to work on this anytime soon?

AdarshRawat1:
@Hannsek I'll resume in few days, sorry for the delay.

Hannsek:
@AdarshRawat1 so, you won’t continue on this?

PR #2916, opened 2023-03-28
Improve allocation report

grdddj:
Done as a secondary product of researching allocation differences between UI1 and UI2:
- typing improvements
- moving from custom-HTML creation into using `dominate`
- color highlights of most allocated lines
- list of most allocated lines

Comparison made between current `master` and `release/22.11` branches.

Steps (inspired by `core unix memory profiler` CI job):
- `TREZOR_MEMPERF=1 PYOPT=0 make build_unix_frozen`
- `TREZOR_MEMPERF=1 make test_emu`
- ... it takes a very long time ...
- `mkdir prof/memperf-html`
- `tools/alloc.py --alloc-data=src/alloc_data.txt html prof/memperf-html`

Results of the comparison:
- U1 data - https://grdddj.github.io/allocations_UI1/
- UI2 data - https://grdddj.github.io/allocations_UI2/

Note that the `Total allocations: XXX` is bigger in case of `UI1`, even though the number of tests in `release/22/11` is smaller - so it seems like a good sign (`UI2` using less of them).

For UI comparison, the relevant files are probably:
- `layouts/tt_v2/__init__.py` - https://grdddj.github.io/allocations_UI2/trezor/ui/layouts/tt_v2/init.py.html
- `layouts/tt/__init__.py` - https://grdddj.github.io/allocations_UI1/trezor/ui/layouts/tt/init.py.html

Seems no big changes there, most of the allocations are done by some `string` or `list` operations, and surprisingly the creation of layouts (`RustLayout`) does not mean allocation in most cases. The reliability of the data is another question, so who knows...

### DIFF
--- a/core/tools/alloc.py
+++ b/core/tools/alloc.py
@@ -1,14 +1,40 @@
 #!/usr/bin/env python3
+from __future__ import annotations
 
 from pathlib import Path
 from types import SimpleNamespace
+from typing import TYPE_CHECKING, Dict, Protocol, TextIO
+
+# for python37 support, is not present in typing there
+from typing_extensions import TypedDict
+
 import click
+from dominate import document
+from dominate.tags import *
+from dominate.util import raw
 
 HERE = Path(__file__).resolve().parent
 
 
-def parse_alloc_data(alloc_data):
-    parsed_data = {}
+if TYPE_CHECKING:
+
+    class LineAllocData(TypedDict):
+        total_allocs: int
+        total_calls: int
+        avg_allocs: float
+
+    # {filename:{lineno:LineAllocData}}
+    alloc_data_dict = Dict[str, Dict[int, LineAllocData]]
+
+    class SharedObject(Protocol):
+        data: alloc_data_dict
+        type: str
+
+
+def parse_alloc_data(
+    alloc_data: TextIO,
+) -> alloc_data_dict:
+    parsed_data: alloc_data_dict = {}
     for line in alloc_data:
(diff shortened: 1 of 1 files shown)


mmilata (approved):
changes lgtm

mmilata:
17 % total allocations down even though we added code & tests in the meantime is good i think!

Not sure what is the takeaway from the reports though. As an example I'm confused by https://grdddj.github.io/allocations_UI2/trezor/ui/layouts/tt_v2/init.py.html:
Why does `CONFIRMED = trezorui2.CONFIRMED` have 12603 allocations but `CANCELLED = trezorui2.CANCELLED` has zero? Why do some function definitions have 12603 allocations but others have 50412? And for some functions the line points to some random parameter in the argument list (`br_code` in `confirm_output`).

grdddj:
> Not sure what is the takeaway from the reports though. As an example I'm confused by https://grdddj.github.io/allocations_UI2/trezor/ui/layouts/tt_v2/init.py.html:
> Why does `CONFIRMED = trezorui2.CONFIRMED` have 12603 allocations but `CANCELLED = trezorui2.CANCELLED` has zero? Why do some function definitions have 12603 allocations but others have 50412? And for some functions the line points to some random parameter in the argument list (`br_code` in `confirm_output`).

I am also confused by some of the results. Maybe the allocation happens "randomly", for that we would need to run the same tests multiple times and see if we receive at least similar results - but probably not now.

I have fixed the failing `python_support_check` in the force-push and will merge when CI goes green

PR #2918, opened 2023-03-29
increase work buffer size for jpeg decompression

TychoVrahe:
Increasing the work buffer size a bit to make it more reliable.


matejcik (approved):
utACK

Issue #2680, opened 2022-12-13
UI2: send transaction style update


mmilata:
<img width="1002" alt="image" src="https://user-images.githubusercontent.com/85857/207273189-13fb5855-b1b4-4358-8701-18e8182111c8.png">

fixes #2151

Hannsek:
Figma file here: https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?node-id=4081%3A19006&t=VxjZTbyYG85Yz2AV-1

Hannsek:
New design is coming. For now, this is blocked.

mmilata:
Is this still blocked or is the figma design ready now?

Hannsek:
Still not ready, I hope it will be ready next week…

Hannsek:
<img width="873" alt="image" src="https://user-images.githubusercontent.com/57008159/217881822-106c2d04-be02-4889-926a-375ac4f28151.png">
<img width="534" alt="image" src="https://user-images.githubusercontent.com/57008159/217882317-4dfffff5-c7a3-4cda-bcc6-12b720938aaa.png">


Figma file: https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?node-id=6276%3A46984&t=fVkX6Rym50TQFKbi-4

Design is ready @mmilata. For now we won't use the chunks when displaying address. Later on we will use it.

andrewkozlik:
In the `recipient #1` screen:
- I think showing the currency in the account label is open for discussion. It should be the same as the currency that we are spending. If it's not, e.g. we are sending LTC to a BTC path, then it seems more appropriate to show the path, than trying to assign a name to it.
- I think we really need to use the word "account" in the account label. `L. SegWit #223` just seems unclear what it means, i.e. `L. SegWit account #223`.

Hannsek:
I agree with both of your points. 
<img width="212" alt="image" src="https://user-images.githubusercontent.com/57008159/219284473-7526a722-2cd0-4838-8504-44965fbb936e.png">

sending LTC to a BTC path is allowed only behind safety checks, right?

andrewkozlik:
> sending LTC to a BTC path is allowed only behind safety checks, right?

[...]

Issue #2353, opened 2022-06-29
Passphrase identification


andrewkozlik:
Trezor allows handling multiple passphrases at the same time. If the user enters two different passphrases and then sees a `GetAddress` dialog, then they can see the path of the address, but not the passphrase it belongs to.

If more than one passphrase is active in the session cache, then Trezor should show some indication of which one is being used in `GetAddress`, `SignTx` and when confirming access to a SLIP-25 account.

How should the passphrases be identified:
- **Temporary numeric label.** The sequence number of the passphrase as it was entered on Trezor. Trezor should be able to tell if the user entered the same passphrase as in another session. In that case both sessions should have the same passphrase number. Requires adding one `uint8_t` to the session cache. The assigned sequence number would need to be shared with Suite, so that both Suite and Trezor use the same number (e.g. Electrum might be using Trezor with passphrase 1 and then Suite connects to Trezor with passphrase 2).
- **Temporary text label from Suite.** Suite can send a text label for the entered passphrase and the user will just confirm that they are happy with referring to the entered passphrase using that label. The label will then either:
    - be stored in the session cache or
    - linked with the session ID using a temporary MAC via a temporary key generated in Trezor. In every `Initialize` the label and MAC will be sent to Trezor and the MAC verified.
- **Permanent text label.** A label that is configured on Trezor only once with a long-term MAC that can be stored and used at any time in the future, see #1244.

Related issues:
- https://github.com/trezor/trezor-firmware/pull/2289
- https://github.com/trezor/trezor-firmware/issues/1244
- https://github.com/trezor/trezor-firmware/issues/2151

Hannsek:
We can add the "name" of the wallet on the screen where the user sees the passphrase. Then we can add the "name" of the wallet also on the account detail screen in receive flow. 

"name" here stands for "hidden wallet #X"

![image](https://user-images.githubusercontent.com/57008159/213772371-6e6b7973-594c-40c9-9487-1d93723c33b0.png)

matejcik:
That looks reasonable, and I can't foresee any good attack scenarios either.

The only thing comes to mind, malware wants you to spend from account A, so while unlocking account A, the label shown will be account B, we cross our fingers that the user doesn't notice, then we get them to unlock account B and spend from account B, but the transaction actually spends from A (which now has the same label).
[...]

Issue #13, opened 2018-03-20
Dry-run recovery should display words on screen


slush0:
Dry-run recovery should display words on screen to prevent malware lying to device about used words.

sime:
We have on device dry runs for both T1 and TT. Can be closed?

andrewkozlik:
This issue is about dry-run recovery on T1. I assume that @slush0 meant that the Trezor should display the word that the user entered, which it does not do. It just shows:
![image](https://user-images.githubusercontent.com/42678794/214284300-56428105-60f2-427b-9d2f-f162db48bb3e.png)

I agree with @jhoenicke that this should not be limited to dry-run, but should apply to recovery always.

Hannsek:
Ahh, ok, misunderstood that, sorry.

PR #2925, opened 2023-03-30
Send flow style update

mmilata:
Fixes part of #2680, there are no changes to loaders or warning screens. I still need to write some tests but otherwise I think it's ready for review/testing.

 [UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4035728484/artifacts/test_ui_report/index.html), [differing screens](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4035728484/artifacts/test_ui_report/differing_screens.html)

b5da2d2913f970305cc4b8c0603754d9edc62ad905c78ecde620ad2955c315ec 19ebb69bc91bd43611b8546c6a05d4919dd58611ee4a9515d48064374c056939 62375c91ba7a9138e86dfebc2ea80adbd3a33ac787a534e47266b21b75a846b1 emu00000000 emu00000001 emu00000002

IMHO the _summary_ is a bit crowded, what if we moved the fee rate to the _spending from_ screen?


mmilata:
@andrewkozlik can you please review the changes to `apps/bitcoin/sign_tx/`?

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -440,42 +440,98 @@ extern "C" fn new_confirm_action(n_args: usize, args: *const Obj, kwargs: *mut M
...
+                SwipeHoldPage::new(paragraphs, theme::BG),
+            );
+            if let Some(subtitle) = self.subtitle {
+                frame = frame.with_subtitle(theme::label_subtitle(), subtitle);
+            }
+            if self.info_button {
+                frame = frame.with_info_button();
+            }
+            LayoutObj::new(frame)?
+        } else if let Some(verb) = self.verb {
```
grdddj, line 516: From 80 %, this `if` and `else if` branches are the same, I wonder if we can improve it.

Rough idea:
```rs
if !self.hold | self.verb.is_none() {
    panic!("Either `hold=true` or `verb=Some(StrBuffer)` must be specified");
}
let content = if self.hold {
    SwipeHoldPage::new(paragraphs, theme::BG)
} else {
    let buttons = Button::cancel_confirm_text(self.verb_cancel, Some(verb));
    SwipePage::new(paragraphs, buttons, theme::BG).with_cancel_on_first_page()
}
let mut frame = Frame::left_aligned(
    theme::label_title(),
    self.title,
    content,
);
if let Some(subtitle) = self.subtitle {
    frame = frame.with_subtitle(theme::label_subtitle(), subtitle);
}
if self.info_button {
    frame = frame.with_info_button();
}
let obj = LayoutObj::new(frame)?;
Ok(obj.into())
```

I am realising this will most probably not compile, because `SwipeHoldPage` and `SwipePage` are different types, but I believe it should be solvable somehow

core/embed/rust/src/ui/model_tt/component/button.rs
```diff
@@ -418,7 +418,11 @@ impl<T> Button<T> {
 
         let left = if let Some(verb) = left {
             left_is_small = verb.as_ref().len() <= 4;
-            Button::with_text(verb)
+            if verb.as_ref() == "^" {
```
grdddj, line 421: Thumbs up for this :)

grdddj (commented):
Couple comments.

Unit tests fail in CI because of changed function signatures - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/4035940725

mmilata:
> I cannot se the name of my receiving account. Should it be already implemented?

It should, and some tests trigger this screen.

emu00000002

Can you describe the steps after which you expect to see the screen but don't?

Hannsek:
I generate the address from an account in the same wallet. E.g. I am sending funds from Taproot #1 to Segwit #2. (Segwit #2 generated the address.) Then I should see the Segwit #2, no?

mmilata:
I tried this with (testnet) `SegWit #1` -> `Taproot #1`. Suite displayed `Bitcoin Testnet #1` but trezor didn't. This is because Suite didn't set the `TxOutput.address_n` field, as firmware can't detect internal outputs just from `TxOutput.address`.

Hannsek:
So is this something that needs to be done by Suite? That Suite will need to send different message?

mmilata:
Yes, Suite needs to send `TxOutput.address_n` instead of `TxOutput.address` for internal transfers in order for trezor to recognize them as such. Seems Suite is able to detect internal address this should probably not be difficult.

core/src/apps/bitcoin/sign_tx/approvers.py
```diff
@@ -113,6 +116,7 @@ def add_orig_change_output(self, txo: TxOutput) -> None:
 
     async def add_external_output(
         self,
+        tx_info: TxInfo,
```
andrewkozlik, line 119: If I understand this correctly, the reason why we need to add the TxInfo here is so that we can show the source account if the user presses the info button. But do we really want that screen to be accessible from the output confirmation dialog? Shouldn't it just be accessible from the Summary dialog?

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -534,37 +534,83 @@ async def confirm_output(
...
-        "confirm_output",
-        br_code,
-        verb="NEXT",
-    )
+    while True:
+        while True:
+            recipient_title = (
+                f"RECIPIENT #{output_index + 1}"
+                if output_index is not None
+                else title or "SENDING TO"
```
andrewkozlik, line 552: `output_index` overrides the `title`. I wonder if it shouldn't be the other way around.

Hannsek:
> I'd change the screen name to something more generic like "Details"

I agree that it `sending from` is not related to fee rate. However the `Details` won't tell the user that he is actually _sending from_ the account(s) below. It is not ideal, but I'd let there `sending from` as it is more user friendly solution.

> I think one problem with the summary screen that might be making it feel crowded is that the labels and values aren't distinguished by font. For example I'd expect "Total amount:" and "Including fee:" to be bold or different size from the amounts or the amounts to be bold.

There actually are different fonts.

> The "L. SegWit https://github.com/trezor/trezor-firmware/issues/13" in the heading seems to be larger than "RECIPIENT https://github.com/trezor/trezor-firmware/issues/2". This makes the heading look chaotic to me. Also, as I keep repeating frequently, the missing word "account" in "L. SegWit account https://github.com/trezor/trezor-firmware/issues/13" makes it unclear what the heading means.

The font used for the account name should be TT Hoves, Bold, 17. It actually looks different from the original design. @mmilata Is the font above used here?

As it is not so much important, _account_ won't be there. We can _eventually_ use marquee there. But not a priority now.

Hannsek:
> Yes, Suite needs to send `TxOutput.address_n` instead of `TxOutput.address` for internal transfers in order for trezor to recognize them as such. Seems Suite is able to detect internal address so this should probably not be difficult.

Ok, I will create an issue in Suite.

core/src/apps/bitcoin/sign_tx/approvers.py
Hannsek: Why it shouldn't be accessible from every screen?

andrewkozlik: It introduces a new dependency into the output confirmation and adds some complexity, which is always a reason to stop and think whether it's justified, because it can lead to more complexity in the long run. But I don't see any significant problem with it.

andrewkozlik: Also the button takes up room which could be better used to clearly communicate the account name when the output is to another of the user's own accounts.

Hannsek: Ok, we can put it only on the last screen if you wish.

There is unfortunately not enough space for it: 
image

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -534,37 +534,83 @@ async def confirm_output(
...
-        title,
-        address,
-        description,
-        "confirm_output",
-        br_code,
-        verb="NEXT",
-    )
+    while True:
+        while True:
+            recipient_title = (
```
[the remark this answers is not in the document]
mmilata: ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -534,37 +534,83 @@ async def confirm_output(
...
+                break
+            elif result is INFO and spending_label is not None:
+                result = await ctx.wait(
+                    RustLayout(trezorui2.show_spending_details(account=spending_label))
+                )
+                assert result is CANCELLED
+            else:
+                raise ActionCancelled
+
+        amount_title = (
```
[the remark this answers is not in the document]
mmilata: ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -534,37 +534,83 @@ async def confirm_output(
...
+                ctx,
+                RustLayout(
+                    trezorui2.confirm_value(
+                        title=recipient_title,
+                        subtitle=address_label,
+                        description=None,
+                        value=address,
+                        verb="CONTINUE",
+                        hold=False,
+                        info_button=(spending_label is not None),
```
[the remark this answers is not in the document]
mmilata: ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -840,47 +890,62 @@ async def confirm_total(
...
+        elif result is INFO and account_label is not None:
+            pass
+            result = await ctx.wait(
+                RustLayout(trezorui2.show_spending_details(account=account_label))
+            )
+            assert result is CANCELLED
+            layout.request_complete_repaint()
+            continue
+
+        raise ActionCancelled
```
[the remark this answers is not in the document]
mmilata: it's not unreachable, the if above it is not exhaustive

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -840,47 +890,62 @@ async def confirm_total(
...
+        pages=layout.page_count(),
     )
 
+    while True:
+        result = await ctx.wait(layout)
+
+        if result is CONFIRMED:
+            return
+        elif result is INFO and account_label is not None:
+            pass
```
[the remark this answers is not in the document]
mmilata: ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/apps/bitcoin/sign_tx/layout.py
```diff
@@ -49,8 +51,22 @@ def format_coin_amount(amount: int, coin: CoinInfo, amount_unit: AmountUnit) ->
     return f"{format_amount(amount, decimals)} {shortcut}"
 
 
+def account_label(coin: CoinInfo, address_n: Bip32Path | None):
```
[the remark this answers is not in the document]
mmilata: ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/apps/bitcoin/sign_tx/approvers.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, tx: SignTx, coin: CoinInfo) -> None:
         self.amount_unit = tx.amount_unit
         self.has_unverified_external_input = False
 
+        # output numbering to be used in confirmation dialogs
+        self.num_external_outputs = 0
```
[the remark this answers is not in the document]
mmilata: ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/trezor/ui/layouts/tt_v2/__init__.py
mmilata: Should be fixed in ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/apps/bitcoin/sign_tx/approvers.py
mmilata: Removed transaction details from confirm output layout, as well as passing `tx_info` around, in ba5f6490a6e802aa97a2441baef26decb8560c47.

mmilata:
> Is the font above used here?

Bold, 17 is the _RECIPIENT #..._ string. For the account name Figma has Monospace, 20, but the branch had Regular, 21. I changed the account name subtitle to Monospace in https://github.com/trezor/trezor-firmware/commit/ba5f6490a6e802aa97a2441baef26decb8560c47. It's still larger but not as much.

 emu00000002 emu00000003 emu00000001 emu00000000 97eefd6942ed1a365f5f4c2346b855e6af9c9fb321b3bed67f5dc3fd5e81d694

FWIW with the Regular, 21 subtitle the _account_ string fits (screenshot for size demonstration purposes only, normally the number would be at the end):
c9ecdb666bffc68cdc6c112971aeabd69a43ca733e2a4a7cf095e9cf66ff7834

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -840,47 +890,62 @@ async def confirm_total(
...
+            result = await ctx.wait(
+                RustLayout(trezorui2.show_spending_details(account=account_label))
+            )
+            assert result is CANCELLED
```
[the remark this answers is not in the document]
mmilata: After https://github.com/trezor/trezor-firmware/commit/ba5f6490a6e802aa97a2441baef26decb8560c47 this is no longer relevant.

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -534,37 +534,83 @@ async def confirm_output(
...
-        ctx,
-        title,
-        address,
-        description,
-        "confirm_output",
-        br_code,
-        verb="NEXT",
-    )
+    while True:
+        while True:
```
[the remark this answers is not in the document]
mmilata: simplified it considerably in ba5f6490a6e802aa97a2441baef26decb8560c47

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -534,37 +534,83 @@ async def confirm_output(
...
+            br_code,
+        )
+
+        if result is CONFIRMED:
+            return
+        elif result is INFO and spending_label is not None:
+            result = await ctx.wait(
+                RustLayout(trezorui2.show_spending_details(account=spending_label))
+            )
+            assert result is CANCELLED
```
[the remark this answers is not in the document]
mmilata: they're no longer there in ba5f6490a6e802aa97a2441baef26decb8560c47

Hannsek:
> For the account name Figma has Monospace, 20

Yes, My bad. It is `Roboto Mono, Medium, 20`

However, the font still looks differently. Is it _Medium_? 
image

core/src/apps/bitcoin/sign_tx/approvers.py
andrewkozlik: BTW, I just realized one nice example of how "it can lead to more complexity in the long run". If you want to add the fee rate to the info screen, then you wouldn't be able to do that in the output confirmation dialogue, because the fee is not known yet. It would require adding a new phase to the signing flow.

Hannsek:
> I don't see how "Spending from" is related to the fee rate. If we were to do this then I'd change the screen name to something more generic like "Details".

How about one more info screen @andrewkozlik ?

image

andrewkozlik:
> FWIW with the Regular, 21 subtitle the _account_ string fits (screenshot for size demonstration purposes only, normally the number would be at the end):
> c9ecdb666bffc68cdc6c112971aeabd69a43ca733e2a4a7cf095e9cf66ff7834

But why is the subtitle font larger than the title font? The title has cap height 12px and the subtitle 15px. As I already stated, it makes the heading look chaotic. The original mockup looked OK: https://github.com/trezor/trezor-firmware/issues/2680#issuecomment-1432586663 (except maybe for the space at the beginning of the line).

Hannsek:
> But why is the subtitle font larger than the title font? The title has cap height 12px and the subtitle 15px. As I already stated, it makes the heading look chaotic. The original mockup looked OK: https://github.com/trezor/trezor-firmware/issues/2680#issuecomment-1432586663 (except maybe for the space at the beginning of the line).

We have changed the fonts used in the design.

andrewkozlik:
> We have changed the fonts used in the design.

That is not a rational reason for the subtitle to be larger than the title.

andrewkozlik:
> > I don't see how "Spending from" is related to the fee rate. If we were to do this then I'd change the screen name to something more generic like "Details".
> 
> How about one more info screen @andrewkozlik ?
> 
> image

BTW, another option is for the fee + fee rate to be on the penultimate screen in the flow and the total amount to be on the last screen. I think we already had it like this in the design once.

Hannsek:
> BTW, another option is for the fee + fee rate to be on the penultimate screen in the flow and the total amount to be on the last screen. I think we already had it like this in the design once.

Yes, but then there is one more screen in the happy path that does not has to be there.

Regular users don't knows what `sat/vB` is (and they don't have to know), that is why I want to move it to the other screen.

mmilata:
Ok, RobotoMono_Medium_20 makes summary look better:

189f851bb9808da7e41f7c4722b307413feca74bad7cfb9802db8fd28da57b64 5817b0ac7a7adbd3edba13fcfb21c1344fb6862836dbf5e7eb73b71c41c907cf
41392a665ea0aef4236f13ff8d1212a9ec85313f3484deac7611757b56655c4e 392d46d7a8235ca4ff9f7e0211cce2f7299853f37ad7ea85c37959beb922745a

mmilata:
Since we currently don't have wallet identification I propose to keep the account and feerate on one screen and split it once #2353 is implemented:
emu00000000

mmilata:
[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4049603312/artifacts/master_diff/index.html) w/ updated Mono font

mmilata:
Device test + swipe to/from details in [16113c6](https://github.com/trezor/trezor-firmware/pull/2925/commits/16113c6052346e27c60ce3eeb31d7bd5820ff3f2).

@Hannsek if you have time can you please test if ButtonRequests work well with Suite? For every output we send two `ButtonRequest.ConfirmOutput` and I think both used to be on the same screen in UI1.

Hannsek:
Swipe works as expected. 👍🏻

I don't see any bad behaviour of the communication between Trezor and Suite.

@dahaca Can you confirm that it is ok in the new build?

mmilata:
Added fix for XRP, rebased, [UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4052613187/artifacts/master_diff/index.html).

core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -1590,24 +1698,36 @@ pub static mp_module_trezorui2: Module = obj_module! {
...
     ///     *,
-    ///     spending_amount: str,
-    ///     total_amount: str,
+    ///     title: str,
+    ///     items: List[Tuple[str, str]],
+    ///     info_button: bool = False,
     /// ) -> object:
-    ///     """Confirm total if there are external inputs."""
-    Qstr::MP_QSTR_confirm_joint_total => obj_fn_kw!(0, new_confirm_joint_total).as_obj(),
+    ///     """Transaction summary."""
```
grdddj, line 1729: I would probably add info that it is always HoldToConfirm, as that is an important property

grdddj (approved):
Code changes look good to me

bosomt:
**QA OK**

Tested on multiple coins like xrp,ada,test,btc,eth,

**Info:**
 - Suite version: desktop 23.4.1 (cff9e3ab255d4a7feebe7fb81ff3170146ee3537)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.4.1 Chrome/110.0.5481.208 Electron/23.2.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.6.0 Universal (revision 88e1f8c7a5c7615723664c64b0a25adc0c409dee)
 - Transport: BridgeTransport 2.0.27

PR #2929, opened 2023-04-03
support differently sized homescreen

TychoVrahe:
Support is added for non-square full-screen homescreen image


bosomt:
**QA OK**

 - Device: model T 2.6.0 Universal (revision 88e1f8c7a5c7615723664c64b0a25adc0c409dee)

PR #2931, opened 2023-04-03
Enable Trezor to work as a FIDO2 authenticator for Apple.

### LINKED ISSUE #2784: Trezor Model T does not work as an Apple Security Key
**Describe the bug**

I am able to add a Model One as an Apple Security Key.  
However, when I try to add a Model T, I am prompted to register on the Trezor screen, but when I confirm, 
I get the error "HTTPError: Your request could not be completed." on the "Add Keys" dialog.

Same issue reproduced in both MacOS and my iPhone.
Once the key is added in Model One, I can use my Model T (with the same seed) as a key as well.

**Info:**
-   OS: MacOS 13.2 and iOS 16.3
-   Firmware Version: Trezor Model T 2.5.3; Model One: 1.11.2

**How to reproduce**

1. Navigate to your iCloud Profile
1. Go to "Password and Security" screen in either iPhone or MacOS
1. Click "Security Keys"
1. Try to add Model T as the key

**Expected behavior**
Trezor is added as a Security Key

**Additional context**
Apple says that any "FIDO Certified" FIDO2 key is supported: https://support.apple.com/en-us/HT213154
Neither Trezor is FIDO certified, so why does the Model One work and the Model T does not work?

andrewkozlik:
Fixes #2784

Trezor T can't be used as an Apple security key, because the FIDO2 credentials are a bit larger than U2F credentials. Apple has some internal size limit on credentials and refuses to store them. As a result it's possible to use Trezor One as a security key with Apple, but not Trezor T. In this PR I introduce the `use_compact` option which reduces the amount of information stored in the credential to a minimum. After enabling this option for Apple, the FIDO2 registration works fine.

<s>Before merging @vdovhanych just needs to confirm that it works fine with `"use_sign_count": false`.</s>

### DIFF
--- a/common/defs/fido/apple.json
+++ b/common/defs/fido/apple.json
@@ -1,5 +1,7 @@
 {
   "name": "Apple",
   "webauthn": ["apple.com"],
-  "use_self_attestation": false
+  "use_sign_count": false,
+  "use_self_attestation": false,
+  "use_compact": true
 }
--- a/common/tools/coin_info.py
+++ b/common/tools/coin_info.py
@@ -138,6 +138,7 @@ class FidoApp(TypedDict):
     u2f: list[dict[str, str]]
     use_sign_count: bool
     use_self_attestation: bool
+    use_compact: bool
     no_icon: bool
 
     key: str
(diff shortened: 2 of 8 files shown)


matejcik (approved):
utACK

bosomt:
@vdovhanych thank you for confirmation 🎉 
i was not able to retest due to some Apple policy.

PR #2933, opened 2023-04-05
core: prevent USB events from waking autolock timer

### LINKED ISSUE #2651: Softlock behaves incorrectly with CoinJoin
from Slack:

> Apparently the soft lock during coinjoin behaves somewhat unexpectedly. I left the device alone with lock time 10 min (afaik default) and after few hours, the device is still not locked. Looks like the API calls around coinjoin are resetting internal counter, which they shouldn't IMO.

Presumably the device should not be usable, but CoinJoin messages should be allowed to run through.

With current softlock implementation any USB message will reset the timer.
We might want to differentiate "soft" lock touch (done by an USB message) from a "hard" touch (done by user interaction).
A "hard" touch resets the internal timer. A "soft" touch keeps the timer running, but if there are any "soft" touches registered at the end of the period, the locking will not shut down the current workflow.

To test: CoinJoin operations (via `DoPreauthorized`) are allowed to run even if the device is currently locked. Does this actually work, i.e., doesn't the signing process somehow implicitly rely on storage being unlocked?

mmilata:
~~Given the timeline and complexity I did not feel confident implementing the proposed solution of #2651. This PR introduces different solution which avoids touching the idle timer when processing message listed as allowed to be processed when device is locked.~~

~~The disadvantage is that with unfortunate timing the autolock interrupts coinjoin signing if there's one in progress.~~

Fixes #2651.

### DIFF
--- a/core/src/apps/base.py
+++ b/core/src/apps/base.py
@@ -24,18 +24,6 @@ if TYPE_CHECKING:
     )
 
 
-_ALLOW_WHILE_LOCKED = (
-    MessageType.Initialize,
-    MessageType.EndSession,
-    MessageType.GetFeatures,
-    MessageType.Cancel,
-    MessageType.LockDevice,
-    MessageType.DoPreauthorized,
-    MessageType.WipeDevice,
-    MessageType.SetBusy,
-)
-
-
 def busy_expiry_ms() -> int:
     """
     Returns the time left until the busy state expires or 0 if the device is not in the busy state.
@@ -313,17 +301,18 @@ def set_homescreen() -> None:
         set_default(homescreen)
 
 
-def lock_device() -> None:
+def lock_device(interrupt_workflow: bool = True) -> None:
     if config.has_pin():
         config.lock()
         wire.find_handler = get_pinlocked_handler
         set_homescreen()
-        workflow.close_others()
+        if interrupt_workflow:
+            workflow.close_others()
 
 
 def lock_device_if_unlocked() -> None:
     if config.is_unlocked():
-        lock_device()
+        lock_device(interrupt_workflow=workflow.autolock_interrupts_workflow)
 
 
 async def unlock_device(ctx: wire.GenericContext = wire.DUMMY_CONTEXT) -> None:
@@ -355,7 +344,7 @@ def get_pinlocked_handler(
         if iface is usb.iface_debug:
             return orig_handler
 
-    if msg_type in _ALLOW_WHILE_LOCKED:
+    if msg_type in workflow.ALLOW_WHILE_LOCKED:
         return orig_handler
 
     async def wrapper(ctx: wire.Context, msg: wire.Msg) -> protobuf.MessageType:
--- a/core/src/apps/bitcoin/sign_tx/bitcoin.py
+++ b/core/src/apps/bitcoin/sign_tx/bitcoin.py
@@ -1,6 +1,7 @@
 from micropython import const
 from typing import TYPE_CHECKING
 
+from trezor import workflow
 from trezor.crypto.hashlib import sha256
 from trezor.enums import InputScriptType
 from trezor.utils import HashWriter, empty_bytearray
@@ -75,6 +76,10 @@ class Bitcoin:
             self.tx_info.tx,
             self.orig_txs,
         )
+
+        # Following steps can take a long time, make sure autolock doesn't kick in.
+        # This is set to True again after workflow is finished in start_default().
+        workflow.autolock_interrupts_workflow = False
 
         # Verify the transaction input amounts by requesting each previous transaction
         # and checking its output amount. Verify external inputs which have already
--- a/core/src/storage/cache.py
+++ b/core/src/storage/cache.py
@@ -47,6 +47,10 @@ LOCKSCREEN_ON = object()
 BUSYSCREEN_ON = object()
 homescreen_shown: object | None = None
 
+# Timestamp of last autolock activity.
+# Here to persist across main loop restart between workflows.
+autolock_last_touch: int | None = None
+
 
 class InvalidSessionError(Exception):
     pass
@@ -338,8 +342,11 @@ def stored_async(key: int) -> Callable[[AsyncByteFunc[P]], AsyncByteFunc[P]]:
 
 def clear_all() -> None:
     global _active_session_idx
+    global autolock_last_touch
 
     _active_session_idx = None
     _SESSIONLESS_CACHE.clear()
     for session in _SESSIONS:
         session.clear()
+
+    autolock_last_touch = None
--- a/core/src/trezor/wire/__init__.py
+++ b/core/src/trezor/wire/__init__.py
@@ -194,8 +194,6 @@ class Context:
                 expected_type.MESSAGE_NAME,
             )
 
-        workflow.idle_timer.touch()
-
         # look up the protobuf class and parse the message
         return _wrap_protobuf_load(msg.data, expected_type)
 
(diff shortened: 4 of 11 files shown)


matejcik:
i'm not sure if that's better or worse than the current state.

what if, instead, we just force locked when going to the coinjoin busy state?
what do you think @andrewkozlik ?

andrewkozlik:
> i'm not sure if that's better or worse than the current state.

I think worse.

> what if, instead, we just force locked when going to the coinjoin busy state? what do you think @andrewkozlik ?

I think this is more of a product / UX design decision, but I personally am not in favor of the idea.

matejcik:
we're converging on the following:

1. keep the modification that preserves deadline across session restarts
2. remove "touch idle timer on usb message"
3. allow `DoPreauthorized` to set a flag, which, when set, causes the locking function to _not_ shutdown the running workflow (but otherwise lock the config, update the homescreen for next run, etc.)

core/src/trezor/workflow.py
```diff
@@ -178,16 +192,24 @@ async def _timeout_task(self, callback: IdleCallback) -> None:
...
 
-    def touch(self) -> None:
+    def touch(self, keep_deadline: bool = False) -> None:
         """Wake up the idle timer.
 
         Events that represent some form of activity (USB messages, touches, etc.) should
         call `touch()` to notify the timer of the activity. All pending callback timers
         will reset.
         """
+        if not dont_autolock:
```
matejcik, line 202: this seems wrong? means that in the weird case where a dont_autolock workflow is waiting for user confirmation, it will remain unlocked.

mmilata: Isn't that what we want though, for the whitelisted workflows, to remain unlocked? With locked storage.

I've used the the same whitelist for not waking up the timer because I'm not sure whether we can entirely ignore usb messages. That would break cases like `test_lots_of_inputs` with short autolock delay, which currently work fine?

mmilata: Ok, let's handle it in apps/bitcoin ...

matejcik: > That would break cases like test_lots_of_inputs with short autolock delay, which currently work fine?

right, which is why I wanted bitcoin signing to set its own exception to autolock too.

ignoring USB messages seems overall more correct, if we have a mechanism to let the workflow finish

core/src/trezor/wire/__init__.py
```diff
@@ -376,6 +375,9 @@ async def _handle_single_message(
                 log.exception(__name__, exc)
         res_msg = failure(exc)
 
+    finally:
+        workflow.autolock_interrupts_workflow = True
```
matejcik, line 379: this might be better in `workflow.start_default()` ?

mmilata:
Tests failing ... looks like preserving deadline across restarts is somehow broken. Did some brief testing on device and didn't catch this.

mmilata:
When user changed the autolock delay the value didn't propagate to the timer correctly as it used the cached value ... db44cb6aad4b7505f1f95e2c373b411c09cefd67

matejcik (approved):
preliminary ACK.

i would like to see some tests, let's talk tomorrow whether we can quickly add some or whether we'll defer that to a sepaeate issue.

core/src/apps/base.py
```diff
@@ -313,17 +301,18 @@ def set_homescreen() -> None:
         set_default(homescreen)
 
 
-def lock_device() -> None:
+def lock_device(interrupt_workflow: bool = True) -> None:
```
andrewkozlik, line 304: So this means that a `LockDevice` message will interrupt every workflow. Do we want that? I think yes, but just checking.

andrewkozlik: Actually sending any unexpected message in a workflow will probably interrupt it IIRC, won't it? So the above though is probably irrelevant. I was just trying to understand how this parameter is used. Anyway, it's OK as is.

core/src/apps/bitcoin/sign_tx/bitcoin.py
```diff
@@ -76,6 +77,10 @@ async def signer(self) -> None:
             self.orig_txs,
         )
 
+        # Following steps can take a long time, make sure autolock doesn't kick in.
+        # This is set to True again after workflow is finished in start_default().
+        workflow.autolock_interrupts_workflow = False
+
```
andrewkozlik, line 83: So now if a user leaves the Trezor hanging in an output confirmation screen, leaves the room and an attacker comes in, the attacker can finish the transaction confirmation even if the device has been autolocked. In comparison if the user leaves the Trezor hanging in a change label screen, i.e. a far less security-sensitive operation, the device will stop the workflow after the autolock timeout and won't let a potential attacker finish the label change.

matejcik:
pushed a bunch of tests confirming the new behavior in 97b6bde.

still not sure how to do the DoPreauthorized one -- maybe i'll need to extend one of the existing coinjoin tests

core/src/apps/bitcoin/sign_tx/bitcoin.py
andrewkozlik: Oh yeah, you are absolutely right. As usual I am doing two things at once and I overlooked that steps 1, 2 & approval are already complete at this stage.

> only if a prompt pops up afterwards, then it would remain open. (something something late path warning? or did we already resolve that?)

That would be a problem, but I think that's already resolved. All checks and warnings should pop up in Steps 1, 2 and the approval step. After that if some check fails, then the entire signing process should fail.

andrewkozlik: For example `BasicApprover.add_internal_input()` shows a warning but `BasicApprover.check_internal_input()` raises an exception. So I think we are fine.

core/src/storage/cache.py
```diff
@@ -47,6 +47,9 @@
 BUSYSCREEN_ON = object()
 homescreen_shown: object | None = None
 
+# Persist autolock deadline across restarts.
```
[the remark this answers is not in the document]
mmilata: [fba8e56](https://github.com/trezor/trezor-firmware/pull/2933/commits/fba8e5650137ff11d07e5a96270fa05add22dfea) changes this to
> Persist autolock deadline across main loop restart between workflows.

Hopefully that's better. We don't have anything about workflow restarts on docs.trezor.io, right?

core/src/trezor/workflow.py
```diff
@@ -187,7 +205,11 @@ def touch(self) -> None:
         """
```
andrewkozlik, line 205: This docstring probably needs an update. USB messages no longer count as activity.

andrewkozlik: Also please document `restore_deadline`.

core/src/trezor/workflow.py
```diff
@@ -215,7 +237,7 @@ def set(self, timeout_ms: int, callback: IdleCallback) -> None:
```
andrewkozlik, line 237: Again, the docstring above `idle_timer.set() also counts as an activity, so all running idle timers are reset.` needs updating if I understand it correctly.

andrewkozlik: Also the comment below the docstring `The reason for counting set() [...]` is probably invalidated by these changes.

core/src/trezor/workflow.py
andrewkozlik: Still trying to understand this. In practice there is only one task in `self.tasks`, the `lock_device_if_unlocked()`, right? But the function is written like there could be more tasks, each one with it's own timeout. Now we are somehow mixing up all the deadlines together with `storage.cache.autolock_deadline` even though they can be different?

mmilata:
6af361aeed156489bd654dbcfeaef746869d39e1 adds DoAuthorized click test

matejcik (approved):
LGTM now

PR #2935, opened 2023-04-05
CoSi paths on screen

### LINKED ISSUE #2932: Confirm CoSi path on screen
The new designs have moved the path away from the title.

We do need to see it, ideally with decoding its meaning. This would be great to fit into this release.

matejcik:
fixes #2932


mmilata (approved):

matejcik:
"ethereum" is throwing the bitcoin-only string check.

it's an open question whether CoSi should even be in bitcoin-only build, but for now i'm just changing "Ethereum definitions" to "External definitions"

Hannsek:
> it's an open question whether CoSi should even be in bitcoin-only build

It should not IMO.

PR #2941, opened 2023-04-07
bootloader: fix images smaller than one chunk

matejcik:
The new bootloader has a bug that disallows installing images smaller than one chunk, i.e., 128 kB minus whatever the size of vendor + firmware headers. For the first chunk, the bootloader always requests data up to the end of the chunk: if the host does not send enough data, the "received chunk size" check fails, and if the host pads the data out, the chunk hash will not match.

This PR corrects the calculation for the first chunk case, so that the right amount of data is requested.

Additionally, we make sure the prodtest image is built at least 128K big, to avoid the problem on the bugged bootloader version.


TychoVrahe (approved):
lgtm

PR #2943, opened 2023-04-08
tests: tweak wait time for test_autolock_ignores_getaddress

mmilata:
On _hardware_ this test is often failing, both on CI and my machine. After changing the wait time to 8s the test passes 20+ times in a row on my machine.

### DIFF
--- a/tests/device_tests/test_autolock.py
+++ b/tests/device_tests/test_autolock.py
@@ -156,13 +156,13 @@ def test_autolock_ignores_getaddress(client: Client):
     assert client.features.unlocked is True
 
     start = time.monotonic()
-    # let's continue for 9 seconds to give a little leeway to the slow CI
-    while time.monotonic() - start < 9:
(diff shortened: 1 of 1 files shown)


matejcik (approved):
utACK

matejcik:
although I'm wondering, what's the failure? does the GetAddress after 9 seconds request PIN? or does it get canceled? because if it requests PIN, then I'm curious about the timing: if I send a GetAddress at a 9-second mark, I would strongly expect the autolock to not engage yet.

and if it gets canceled, that's basically a success and we just need to add exception handling, no?

mmilata:
I've seen both.

matejcik:
OK, let's do it this way. who knows what the CI is doing with its cycles

mmilata:
Not sure if it's the CI, it was quite reproducible on my system using `@pytest.mark.flaky(max_runs=50, min_passes=20)`. Seems like something in our stack is slowing it down but no idea in which layer.

PR #2951, opened 2023-04-12
feat(tests): master diff creating different_screens.html report

grdddj:
Useful thing when finalizing PRs, to see all the unique screens that changed compared to `master`


mmilata (approved):
Awesome! Thanks!

Issue #2605, opened 2022-11-04
T2B1 design - bootloader


Hannsek:
Design: https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?node-id=3790%3A11545

<img width="540" alt="image" src="https://user-images.githubusercontent.com/57008159/203309111-0cb89413-3871-466f-8b70-e16911b09a6a.png">

<img width="729" alt="image" src="https://user-images.githubusercontent.com/57008159/203310452-d0431b69-9f2c-41dc-bae9-6d680c78ba66.png">

FW-less:
<img width="589" alt="image" src="https://user-images.githubusercontent.com/57008159/203311999-d011f034-dc8b-4729-85ec-27a2a4c5246b.png">

TychoVrahe:
Some feedback:
from firmware perspective, it would be beneficial to have the same flow as in model T - keeps things nicely united, less error prone and easier to maintain. This means either removing the initial bootloader warning on TR or adding it in TT, and then have screen with enter menu/ install fw actions like this:
![20221111_151818](https://user-images.githubusercontent.com/37884576/203006326-205064eb-e553-4a6d-8636-7633429003e1.jpg)
(this would also require addition icon design for exit menu item)

Confirming menu actions: we dont use hold to confirm on TT, no need to do it here (its unnecessary complication). Reboot action does not have to be confirmed at all.

See comparison of progress bar/spinner. The spinner is nice and the transition in success screen is smooth, but the beginning with no circle around the icon is weird.
![20221111_153715](https://user-images.githubusercontent.com/37884576/203008199-27602dae-4cb6-415f-803f-7ed0c8ffd1dd.jpg)
![20221111_153907](https://user-images.githubusercontent.com/37884576/203008316-bad2422b-84c5-4a49-885b-4091cda1dc09.jpg)

The mono font is problematic at some screens, see (mainly the phrase 'UNSAFE, DO NOT USE!' doesn't fit one line:
![font2](https://user-images.githubusercontent.com/37884576/203009494-1805b8e5-bffa-415e-92d0-0e1b0a693d55.jpg)
![font1](https://user-images.githubusercontent.com/37884576/203009497-3bdf0198-084c-4971-a04c-c341ad89d696.jpg)

Hannsek:
Noted, new design in progress.

Hannsek:
Design in original comment updated.
[...]

PR #2959, opened 2023-04-13
T2B1 bootloader

### LINKED ISSUE #2954: Prepare boardloader + bootloader + prodtest for T2B1
Prodtest is not needed for the first batch

TychoVrahe:
closes https://github.com/trezor/trezor-firmware/issues/2605

closes #2954


grdddj:
ACK on the TR non-BL changes, mostly https://github.com/trezor/trezor-firmware/pull/2959/commits/a87d052c94d1bbb5dd1e6aaf72eca6ee737d8e99 - all the device/click tests are fine (locally, not in broken CI) in this branch

Hannsek:
closes #2954

grdddj:
If you are not against @TychoVrahe, I will do the rebase on `master` here and get familiar with the bootloader UI code in the process

grdddj:
When I was getting familiar with the bootloader UI and playing with the emulator, I made a small refactor of all the components to make it more similar to firmware components.

[697b5a2](https://github.com/trezor/trezor-firmware/pull/2959/commits/697b5a208f7a2f349522edc88ccf7fc5cd531af7)

E.g.:
- using `bounds` from `place` and not `constant::screen`
- using just one `ButtonController` instance and swapping `ButtonLayout` there instead of having two instances
- reduce usage of `mut instance = Self...` and constructing `Self` in the final state
- add `with_info_screen` method to `Confirm` to improve the readability

I think we are ready to go from `Draft` to a "regular" `PR`

python/src/trezorlib/transport/udp.py
```diff
@@ -141,6 +141,7 @@ def write_chunk(self, chunk: bytes) -> None:
             raise TransportException("Unexpected data length")
         LOG.log(DUMP_PACKETS, f"sending packet: {chunk.hex()}")
         self.socket.sendall(chunk)
+        time.sleep(0.0001)
```
matejcik, line 144: why? this worked fine before without the delay

matejcik:
Please revert all changes to `model_tt`.

The result screen with Rust-based countdown string adds (a) a new size of String, and (b) Rust's own copy of int-to-str code. We want to keep that in C to keep the overall size down.
(and without this change, you don't need to introduce a separate feature for the success screen; generally I would prefer making the bootloader emulator support as minimal and non-intrusive as possible)

python/src/trezorlib/transport/udp.py
TychoVrahe: it did for TT, but actually didn't for TR. (at least on my machine)

TychoVrahe: it was failing during firmware upload to be specific

matejcik:
why is micropython module modified? glitch or intentional?

core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
```diff
@@ -37,9 +40,10 @@ pub trait Choice<T: StringType> {
 /// This way, no more than one item is stored in memory at any time.
 pub trait ChoiceFactory<T: StringType> {
     type Action;
+    type Item: Choice<T>;
```
matejcik, line 43: is this ever something other than `ChoiceItem`?

core/embed/rust/src/ui/model_tr/component/button.rs
```diff
@@ -465,6 +503,14 @@ impl<T> ButtonDetails<T> {
         self
     }
 
+    /// Second icon for the pressed state
+    pub fn with_pressed_icon(mut self, icon: Option<Icon>) -> Self {
```
matejcik, line 507: it doesn't seem that this belongs in this PR?

TychoVrahe:
Micropython change is definitely not intentional

core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
TychoVrahe: `MenuChoice` in `bootloader/menu.rs`

python/src/trezorlib/transport/udp.py
matejcik: That makes zero sense. I would like to look into this, but don't want to hold up the PR. Please remove this change from here, merge without it, and possibly open a separate PR for this.

core/embed/rust/src/ui/model_tr/component/button.rs
TychoVrahe: using this for proper inversion of some button icons - e.g. when it has added one column of pixels to satisfy even width requirement, you don't want to invert the extra column when pressed, hence the need for second icon. 

though @grdddj said that he was solving something related to this, so maybe it isnt needed anymore

matejcik: but i'm unable to lookup usages in the bootloader code because it's hidden like three layers of indirection away from actual callers :woman_shrugging:

core/embed/rust/src/ui/model_tr/bootloader/mod.rs
```diff
@@ -199,196 +190,206 @@ extern "C" fn screen_intro(
...
-        screen_install_success_initial(msg, complete_draw)
-    } else {
-        screen_install_success_bld(msg, complete_draw)
-    }
+extern "C" fn screen_welcome_model() {
+    let mut frame = WelcomeScreen::new();
+    show(&mut frame);
 }
 
+#[cfg(feature = "bootloader")]
```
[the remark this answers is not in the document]
TychoVrahe: yes it was. we can live without it, just the original emulator jump message was confusing, as it said something about firmware installation success, even if no installation was done

core/embed/rust/src/ui/model_tr/component/button.rs
TychoVrahe: thats true, the `button_controller` and `button` components are implemented that way

python/src/trezorlib/transport/udp.py
grdddj: Deleted in [abc042e](https://github.com/trezor/trezor-firmware/pull/2959/commits/abc042ee3467c443719f78fb1618c224d4d9a417)

grdddj:
> why is micropython module modified? glitch or intentional?

Micropython change reverted in [925bcfc](https://github.com/trezor/trezor-firmware/pull/2959/commits/925bcfc8b7b911f3efba3bf5290727b3fff8222e)

core/embed/rust/src/ui/model_tr/bootloader/mod.rs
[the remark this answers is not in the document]
grdddj: Not sure what to do here, it is the same as in `TT`code. Probably we will do this in future `PR`

core/embed/rust/src/ui/model_tr/component/button.rs
grdddj: Deleted in [a59c096](https://github.com/trezor/trezor-firmware/pull/2959/commits/a59c096a27d4d5ec653b1176d63ea95945e20a2f) - now we can invert even odd-wide icons precisely

core/embed/rust/src/ui/model_tr/bootloader/mod.rs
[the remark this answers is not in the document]
grdddj: In the end deleted in [459b78f](https://github.com/trezor/trezor-firmware/pull/2959/commits/459b78fc8fd013f7e3a176502a3b0bd3d4f01749) - I found out that even for TT it was added in this PR, so I deleted both while reverting the TT changes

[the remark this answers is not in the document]
grdddj: > Please revert all changes to `model_tt`.
> 
> The result screen with Rust-based countdown string adds (a) a new size of String, and (b) Rust's own copy of int-to-str code. We want to keep that in C to keep the overall size down. (and without this change, you don't need to introduce a separate feature for the success screen; generally I would prefer making the bootloader emulator support as minimal and non-intrusive as possible)

Revert done in [459b78f](https://github.com/trezor/trezor-firmware/pull/2959/commits/459b78fc8fd013f7e3a176502a3b0bd3d4f01749)

I do not plan to push any more commits - @TychoVrahe please look at the changes and I will then rebase/squash

python/src/trezorlib/transport/udp.py
grdddj: After deleting this, I am also _sometimes_ experiencing `USB error` during firmware upload

grdddj (approved):

Issue #2885, opened 2023-03-15
Add trezor color into device during assembling (T2B1)


TychoVrahe:
We need to differentiate colors of produced device. We need to add a color flag somewhere during production.
 
One way that comes to mind is to use OTP memory which would be programmed dur8ng device testing

prusnak:
Btw, OTP usage is documented here - https://github.com/trezor/trezor-firmware/blob/master/docs/core/misc/memory.md#otp

What is not explicitly documented in this file is the format of block 0.

This is mentioned here: https://github.com/trezor/trezor-firmware/blob/67f3af619ab7e91bf980e5fdec5949c5824ad049/core/embed/prodtest/main.c#L386

So all Trezors Model T manufactured today will contain `TREZOR2-230315` in OTP block 0.

It is trivial to extend this string identifier to something that makes sense. Each block has 32 bytes, so there is plenty of space to use for this.

PR #2961, opened 2023-04-14
support device variants/colors

TychoVrahe:
Uses one OTP block (block 4, 32 bytes) for device variant information- colors, btconly and whatever may come up in the future.

Color encoding is done by enumeration, one byte for now, should be enough for long time. 

First byte in this OTP block defines the structure:
byte 0: value 0x01 - defines structure of following data
byte 1: value 0x01-0x?? - colors
byte 2: value 0x00 - not set/universal, 0x01 - BTC only device.
rest: zeroes.

This partially solves #2885 , but the prodtest part is not done yet.


core/embed/extmod/modtrezorutils/modtrezorutils.c
```diff
@@ -257,6 +288,12 @@ STATIC const mp_rom_map_elem_t mp_module_trezorutils_globals_table[] = {
      MP_ROM_PTR(&mod_trezorutils_reboot_to_bootloader_obj)},
     {MP_ROM_QSTR(MP_QSTR_usb_data_connected),
      MP_ROM_PTR(&mod_trezorutils_usb_data_connected_obj)},
+    {MP_ROM_QSTR(MP_QSTR_device_variant_present),
+     MP_ROM_PTR(&mod_trezorutils_device_variant_present_obj)},
+    {MP_ROM_QSTR(MP_QSTR_device_color),
+     MP_ROM_PTR(&mod_trezorutils_device_color_obj)},
+    {MP_ROM_QSTR(MP_QSTR_device_btconly),
+     MP_ROM_PTR(&mod_trezorutils_device_btconly_obj)},
```
mmilata, line 296: What do you think of dropping `device_variant_present` and instead doing
```
def device_color() -> int | None:
def device_btconly() -> bool | None:
```

common/protob/messages-management.proto
```diff
@@ -36,6 +36,15 @@ enum HomescreenFormat {
     Jpeg240x240 = 2;
 }
 
+
+/**
+ * Device color
+ */
+enum DeviceColor {
```
mmilata, line 43: **device** is one of these words we use so much it's sometimes hard to tell what it means. What about `VariantColor`? Or `ChassisColor`? Though that wouldn't work with btconly, if I understand it correctly that we want to be able to enforce btconly firmware based on OTP contents.

mmilata (commented):
Otherwise lgtm

common/protob/messages-management.proto
```diff
@@ -123,6 +132,8 @@ message Features {
     optional bool busy = 41;                    // is the device busy, showing "Do not disconnect"?
     optional HomescreenFormat homescreen_format = 42;   // format of the homescreen, 1 = TOIf 144x144, 2 = jpg 240x240
     optional bool hide_passphrase_from_host = 43;   // should we hide the passphrase when it comes from host?
+    optional DeviceColor device_color = 44;          // color of the device
+    optional bool device_btconly = 45;          // device is intended as bitcoin only
```
matejcik, line 136: do we need this? we already cover btc-onliness in Capabilities

common/protob/messages-management.proto
```diff
@@ -123,6 +132,8 @@ message Features {
     optional bool busy = 41;                    // is the device busy, showing "Do not disconnect"?
     optional HomescreenFormat homescreen_format = 42;   // format of the homescreen, 1 = TOIf 144x144, 2 = jpg 240x240
     optional bool hide_passphrase_from_host = 43;   // should we hide the passphrase when it comes from host?
+    optional DeviceColor device_color = 44;          // color of the device
```
matejcik, line 135: making it an enum does not scale with future devices. I see two options: (a) int color id copied straight out of otp, Suite will lookup model + color id to display the right color
(b) string with possible marketing name of the color ("galaxy black", "apple red") and the firmware has a lookup table per model

TychoVrahe: i think a) is preferable as it will allow adding more colors without needing to update firmware/bootloader

core/embed/extmod/modtrezorutils/modtrezorutils.c
TychoVrahe: https://github.com/trezor/trezor-firmware/pull/2961/commits/b1bcaab232c2776b142d3bdda7b5e58215e71029

common/protob/messages-management.proto
TychoVrahe: Well the enum itself is gone as part of https://github.com/trezor/trezor-firmware/pull/2961/commits/adb743ef4444e9a2ff05e8a73928ad70173a9f6a , but the term `device` is still used everywhere else in this feature - and i think its rightful usage in this case, as the data/variants are bound to the specific device.

Though there is i.e. `device_id` which is bound to the storage instance and gets regenerated on wipe, so yeah, there is room for confusion.

One alternative term i can think of is 'unit', so that would result in unit_color and unit_btconly, not sure if that is better. 

Or i can drop the device word and just call the whole thing variant, but thats kinda opaque.

Any preference?

mmilata: _Unit_'s good, but I can live with _device_ also. We can discuss on standup but I'd leave the final choice to you.

TychoVrahe: going with unit: https://github.com/trezor/trezor-firmware/commit/4b3267b66829098c17fd6416832ebf746e62413b

mmilata (approved):

common/protob/messages-management.proto
mmilata: (aa) if the colors can reasonably fit into [a-z] we can store ASCII letter in the OTP and represent it as a string for protobuf

TychoVrahe: what is the upside?

mmilata: probably just better human readability in trezorctl without maintaning a lookup table ... but i guess it's better to have it there so that apps using trezorlib can show the human-readable name

PR #2962, opened 2023-04-14
Vendor header for prodtest

matejcik:
introduces a separate vendor header and adds the capability to build prodtest from build-docker


mmilata (approved):
Don't see any problem in changes. Not sure if there's anything to check fingerprints of.

PR #2969, opened 2023-04-18
Debuglink improvements

grdddj:
Draft of sort-of working debuglink changes, which should make it more reliable and model-independent (eventually).

The idea is to keep count of which events were sent via `DebuglinkDecision` and matching the resulting layouts with them. Only when we call `wait=True` or `wait_layout()` will we consume all the layout changes and wait for all the previously sent events to finish.

There are some caveats and edge cases - for example:
- there are layout changes that did not originate from `DebuglinkDecision`
- when the `Layout` changes, we need to account for it
- when the session/test-case changes, we need to account for it. Overall it is quite fragile.

It was tested on PIN and passphrase click tests, which had to be modified. Probably majority of other tests would have to be modified for them to run properly.

It definitely needs more improvements if we want to use it like this.


grdddj:
I am already satisfied with the current version. Little more cleanup will be maybe needed, but it is ready for review.

Changing from draft to a regular `PR`

PR #2976, opened 2023-04-21
Firmware combining tweaks

matejcik:
Improved the `combine_firmware` script to make it easier to manually pass the parts from outside.
Added a "decombine" snippet that can take apart the result, so that, e.g., we can pull out the boardloader from a previous version instead of having to compile it from the current code version.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -321,10 +321,10 @@ sizecheck: ## check sizes of binary files
 
 combine: ## combine boardloader + bootloader + prodtest into one combined image
 	./tools/combine_firmware \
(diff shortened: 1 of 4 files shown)


TychoVrahe (approved):
The combine script is calling the third part of the binary "firmware", but the decombine calls it prodtest. Firmware is probably better technically, but since it is in snippets it doesn't matter much.

I am not very happy about how the `BOOTLOADER_START` etc constants are repeated across several files (makefile, linkerscripts, combine script, implicitly decombine snippet and sort of also embed/trezorhal/flash.c), it doesn't look very sustainable. We don't have to deal with it now, but it might get messy if we opt for some other MCU in the future.

Don't forget to deal with the changelog, but this looks fine otherwise

PR #2989, opened 2023-04-28
support STM32F429I-DISC1

TychoVrahe:
Adds support for STM32F429I-DISC1 board.

Some adjustments are done so that GUI is rectangular, and homescreen format support different dimensions. 

The resistive touch screen is quite annoying to use, and one issue that i didn't solve yet is that it sometimes does not initialize properly, and power off / on of the discovery board is needed.


TychoVrahe:
should this have a changelog entry?

matejcik:
overall style comment: don't add trailing commas to tuples (as in `if TREZOR_MODEL in ('T', 'R', )`)
the reason we write `in ('T',)` is to indicate that it's a 1-tuple (and not just the string 'T' in parens) and it's actually required in the 1 element case. not so in other cases. `(item,)` is a Python idiom, putting trailing comma elsewhere is not.

common/protob/messages-management.proto
```diff
@@ -121,8 +121,11 @@ message Features {
     optional uint32 display_rotation = 39;      // in degrees from North
     optional bool experimental_features = 40;   // are experimental message types enabled?
     optional bool busy = 41;                    // is the device busy, showing "Do not disconnect"?
-    optional HomescreenFormat homescreen_format = 42;   // format of the homescreen, 1 = TOIf 144x144, 2 = jpg 240x240
+    optional HomescreenFormat homescreen_format = 42;   // format of the homescreen, 1 = TOIf, 2 = jpg
     optional bool hide_passphrase_from_host = 43;   // should we hide the passphrase when it comes from host?
+    optional uint32 homescreen_width = 44;         // homescreen width in pixels
+    optional uint32 homescreen_height = 45;         // homescreen height in pixels
```
mmilata, line 127: Not sure about the arbitrary size screen thing. On one hand it's nice to see that the UI works reasonably well on display it was not developed for. On the other hand it would make sense to have this build as similar to model T as possible, and probably shouldn't add features to trezorlib that are not supported by anything we currently sell.

My proposal is to keep the 320x240 resolution but only support 240x240 homescreens for now, and revert the protobuf changes. Thoughts @matejcik?

core/SConscript.boardloader
```diff
@@ -124,6 +127,10 @@ if TREZOR_MODEL in ('T', 'R'):
     CPU_ASFLAGS = '-mthumb -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16'
     CPU_CCFLAGS = '-mthumb -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -mtune=cortex-m4 '
     CPU_MODEL   = 'STM32F427xx'
+elif TREZOR_MODEL in ('DISC1',):
+    CPU_ASFLAGS = '-mthumb -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16'
+    CPU_CCFLAGS = '-mthumb -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -mtune=cortex-m4 '
+    CPU_MODEL   = 'STM32F429xx'
```
mmilata, line 133: Would it still work if we merged the two branches, i.e. used STM32F427xx for disco1?

core/SConscript.bootloader
```diff
@@ -173,16 +168,16 @@ env.Replace(
 env.Replace(
     TREZOR_MODEL=TREZOR_MODEL, )
 
-if TREZOR_MODEL in ('T', 'R'):
+if TREZOR_MODEL in ('T', 'R', ):
```
mmilata, line 171: Nit: trailing comma.

core/SConscript.firmware
```diff
@@ -832,7 +831,7 @@ BINARY_NAME += "-dirty" if tools.get_git_modified() else ""
 BINARY_NAME += ".bin"
 
 
-if TREZOR_MODEL in ('T', 'R'):
+if TREZOR_MODEL in ('T', 'R', 'DISC1', ):
```
mmilata, line 834: Nit: trailing comma.

core/embed/boardloader/main.c
```diff
@@ -215,6 +218,10 @@ int main(void) {
 
   clear_otg_hs_memory();
 
+#ifdef USE_SDRAM
+  sdram_init();
```
mmilata, line 222: Not sure I understand how is the SDRAM used, is it in addition to the on-chip RAM or instead of it? IOW, do we have more memory than on model T?

mmilata (commented):
Nice work, runs fine on my board, even passed bunch of random device tests i ran on it. Couple comments inline, also please:
* add changelog entry &
* add nightly-only CI job so that we notice if it fails to build for some reason.

mmilata:
I've added instructions for building and flashing in 742090523764cfd6e50822ec9ecd0b980efdf277, hope that's ok.

core/SConscript.boardloader
TychoVrahe: The main thing here is that Disco uses LTDC that is not available on F427. STM HAL needs to be configured accordingly. It would most likely work if we set STM32F429xx for model T and R, but it doesn't feel right

core/embed/boardloader/main.c
TychoVrahe: In addition to on chip RAM, this board has external RAM - 8MB more. Its used here for holding frame buffer, which wouldn't fit on chip RAM. (frame buffer is required for display with RGB interface)

common/protob/messages-management.proto
TychoVrahe: This is not strictly needed here and we can revert the changes now, but i expect that we will run into this sooner or later- e.g. T3W1 will need different homescreen size too and we will have to differentiate somehow, and adding Jpeg320x240 seemed like worse option to me

core/SConscript.firmware
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/0b8f237c069cc390c6adaa7c36454fcb8c4e0e14

TychoVrahe:
added changelog entries: https://github.com/trezor/trezor-firmware/commit/f19dc8a916eea72b17ab7bf61af601c406e8e9e3

though the original commit still contains `[no changelog]` as i dont want to force push right now, will try to remember to remove it

mmilata (approved):

core/embed/trezorhal/flash.h
```diff
@@ -26,7 +26,8 @@
 
 // see docs/memory.md for more information
 
-#if defined TREZOR_MODEL_T || defined TREZOR_MODEL_R
+#if defined TREZOR_MODEL_T || defined TREZOR_MODEL_R || \
+    defined TREZOR_MODEL_DISC1
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/4b35f3b1c69c42fad8ef5fc7d5339bebccd127dd

few lines below where is `FLASH_SECTOR_STORAGE_1` defined i left it based on model, as its arbitrary selection which could, in theory at least, differ between models running on same MCU. Though the rest of the flash layout is fixed now but in theory should also be defined per model. Anyway, we will probably have to refactor this part more if we start using another MCU line where the flash is somewhat different.

core/embed/trezorhal/i2c.c
```diff
@@ -1,39 +1,49 @@
...
 #include "i2c.h"
 #include "common.h"
 
 static I2C_HandleTypeDef i2c_handle;
 
 void HAL_I2C_MspInit(I2C_HandleTypeDef *hi2c) {
   // enable I2C clock
-  __HAL_RCC_I2C1_CLK_ENABLE();
-  // GPIO have already been initialised by touch_init
+  I2C_INSTANCE_CLK_EN();
```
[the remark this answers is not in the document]
TychoVrahe: Disco board uses I2C3, instead of I2C1 as model T. So this difference was outsourced to `trezorhal/boards/stm32f429i-disc1.h `and `trezorhal/boards/trezor_t.h`

matejcik (approved):

TychoVrahe:
using D001 also in CI: https://github.com/trezor/trezor-firmware/commit/ecef0a0ee09917bab4715dd2c7d6df700c20bffe

TychoVrahe:
rebased. fixed some issues on the way, like building core T1 variant and homescreen handling needed some adjustments after TR merge. 

will wait with merging this until vendor headers and internal_model PRs get merged too - there are some modifications with impact on this too.

hiviah (approved):

TychoVrahe:
rebased.

added vendor header for DISC1, added model definition to trezorlib, in micropython this is now also recognized as DISC1 and reports as such to host. downside is that suite doesn't work with this, but i think its reasonable for this only to be supported in trezorctl for now?

also removed a condition that said which models supports passphrase on device, as in core everything now does.

TychoVrahe:
and displaying vendor image is fixed here too.

hiviah:
Reviewed the added commits, LGTM.

PR #3003, opened 2023-05-09
fix bootloader emulator compilation

TychoVrahe:
the https://github.com/trezor/trezor-firmware/pull/2912 refactoring moved some files and didn't address bootloader emu which was introduced in meantime


matejcik (approved):
utACK

Issue #3009, opened 2023-05-12
Revert changes from model T


Hannsek:
https://github.com/trezor/trezor-firmware/pull/2803#issuecomment-1542293887

grdddj:
Done in https://github.com/trezor/trezor-firmware/pull/3011

PR #3011, opened 2023-05-12
fix(core): revert some UI changes for TT

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3009#event-9229052320

as agreed with @Hannsek


mmilata (approved):
LGTM, CI failures are unrelated

PR #3027, opened 2023-05-22
fix(core): do not handle click signal in non-debug builds (fixes #3026)

### LINKED ISSUE #3026: Non-debug builds do not finish some layouts
I am observing a strange thing on both `TT` (on master) and `TR` (on its branch) when doing `non-debug-frozen-emulator-build`, e.g. `make build_unix_frozen` or `TREZOR_MODEL=R make build_unix_frozen`

The issue looks like this:
- after a command, e.g. `trezorctl device wipe`, the `UI` does not go back to the `homescreen` (the screen is frozen at the end of the loader), but there seems to be an "infinite loop" spawning new `Homescreen` objects (which is the case also right after spawning the emulator).

On `TR`, I am facing it even on the hardware device.

Everything is fine in a debug-build, so there is probably some difference coming from `if __debug__:`, but I was not able to find something relevant.
 
image

matejcik:
@grdddj please check if this solves your problem

### DIFF
--- a/core/src/trezor/ui/layouts/tt_v2/homescreen.py
+++ b/core/src/trezor/ui/layouts/tt_v2/homescreen.py
@@ -29,13 +29,14 @@ class HomescreenBase(RustLayout):
         else:
             self._paint()
 
-    # In __debug__ mode, ignore {confirm,swipe,input}_signal.
-    def create_tasks(self) -> tuple[loop.AwaitableTask, ...]:
-        return (
-            self.handle_timers(),
-            self.handle_input_and_rendering(),
-            self.handle_click_signal(),  # so we can receive debug events
-        )
+    if __debug__:
+        # In __debug__ mode, ignore {confirm,swipe,input}_signal.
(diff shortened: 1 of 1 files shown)


grdddj (approved):
Excellent! Tried both on `master` and `TR` branch and the UI is not freezing anymore. Thanks for the fix :) 

I just wonder if we cannot simplify it by just deleting the `create_tasks` implementation from `Homescreen` completely - it is already inheriting it from `RustLayout`. Not sure what are the consequences of "not ignoring" the `handle_swipe_signal` and `handle_result_signal` in the current version. Will try it.

matejcik:
i think it's something to do with tests, where the homescreen would eat signals intended for other components if they arrive too early.

merging as soon as the CI pipeline goes green

grdddj:
> i think it's something to do with tests, where the homescreen would eat signals intended for other components if they arrive too early.
> 
> merging as soon as the CI pipeline goes green

I have pushed the complete deletion to `TR` branch, so curious what `CI` will tell.

Agree with merging this.

Issue #3025, opened 2023-05-22
T2B1 - pin entry


Hannsek:
After selecting first 4 digits, jump onto the `enter pin` button.

grdddj:
Done in https://github.com/trezor/trezor-firmware/pull/3042

PR #3042, opened 2023-05-29
feat(core,rust): TR PIN dialogue going to ENTER when user enters 4th digit

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3025

As soon as the user enters the 4th PIN digit, it will bring him to the `ENTER` choice

### DIFF
--- a/core/embed/rust/src/ui/model_tr/component/input_methods/pin.rs
+++ b/core/embed/rust/src/ui/model_tr/component/input_methods/pin.rs
@@ -27,9 +27,12 @@ enum PinAction {
 }
 
 const MAX_PIN_LENGTH: usize = 50;
+/// After how many digits the user will be brought to the "ENTER" choice.
+const NUM_DIGITS_SWITCH_TO_ENTER: usize = 4;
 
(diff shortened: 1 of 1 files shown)


Hannsek:
Are we ok with this? As If user has 6 digits pin and someone listening to clicks, attacker will know the 5th number. @andrewkozlik

andrewkozlik:
> If user has 6 digits pin and someone listening to clicks, attacker will know the 5th number.

I don't like that. But there is a much bigger problem. An attacker can determine the entire PIN with high probability just by listening to the correct PIN entry enough times. I didn't do the exact math, but we are talking something like ~20 times.

Hannsek:
How come? We shuffle the position after every entry.

Hannsek:
> As soon as the user enters the 4th PIN digit, it will bring him to the ENTER choice
Maybe if we shift it on 6th number, it would be enough as only few users would have pin longer then 6 digits imo.

Hannsek:
Closing this in favour of https://github.com/trezor/trezor-firmware/issues/3061.

PR #3045, opened 2023-05-30
use the full ethereum rlp package

kristian1108:
I'm not sure why it was decided to use the `simple-rlp` library: https://github.com/SamuelHaidu/simple-rlp

This makes the `trezor` package incompatible with `web3py` if it's installed first due to the fact that the developer of `simple-rlp` chose a conflicting directory name in their release.

We should just be using the full `pyrlp` dependency. It's much better supported & actually compatible with standard ethereum tooling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ __pycache__/
 proto.gv*
 .DS_Store
 crypto/tests/libtrezor-crypto.so.dSYM/
+python/venv
+venv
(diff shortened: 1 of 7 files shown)


prusnak:
> At the very least, we should change the dependency of simple-rlp to a fork that has the appropriate directory name to alleviate the `rlp` directory conflict. I'm happy to submit that PR instead of this if it's preferable.

If this fork is on PyPI and properly maintained we can consider it.

Other option is to include local copy of simple-rlp to trezorlib and use that instead.

PR #3046, opened 2023-05-30
fix the directory collision with simple-rlp

kristian1108:
See https://github.com/trezor/trezor-firmware/pull/3045#issuecomment-1569117660

Now uses: https://pypi.org/project/simpler-rlp/0.0.1/


prusnak:
@matejcik I am still wondering whether we shouldn't just copy https://github.com/SamuelHaidu/simple-rlp/blob/master/rlp/rlp.py as `_rlp.py` into our source tree and drop the dependency completely.

matejcik:
good idea @prusnak 

@kristian1108 please submit that, thank you

prusnak:
superseded by https://github.com/trezor/trezor-firmware/pull/3305

PR #3048, opened 2023-05-31
Add model identifier to vendor headers

### LINKED ISSUE #3028: Add model field to vendor header
We need different vendor headers for different models because we intend to use different signing keys. Also, different model VHs are themselves going to be signed by different keys baked in different bootloaders.

While there is no _technical_ requirement to have the model field in VH, it costs very little to put it there and greatly simplifies management of VHs. There are 14 unused bytes in the VH, so we can fit the four-byte `hw_model` field easily.

I don't think we need `hw_revision` - the VHs should be common for all revisions of the same model, no?

Unsure about `monotonic`. I don't entirely understand its role in bootloader / firmware headers. But probably not necessary for VH either.

cc @TychoVrahe @prusnak 

---

I believe firmware can completely ignore this field in the VH and only check the firmware field. This saves us the headache of having to special-case T2T1 to be equal to 0000000. All treatment of `hw_model` would only happen in our tooling.

---

alternatives:

1. use same signing keys for T2T1 and T2B1 (and possible future T2 platform models at least)
    * seems undesirable from the security hygiene standpoint
2. use different signing keys but keep VHs as-is and differentiate based on which keys they have / how they are signed / file names / etc
    * doable but impractical, esp. in signing tools where we want to be extra sure about what we're signing

matejcik:
resolves #3028 

this means that we currently can't build production versions of T2B1 because we don't have the signed headers


core/SConscript.firmware
```diff
@@ -794,15 +794,17 @@ cmake_gen = env.Command(
...
+BOOTLOADER_SUFFIX = MODEL_IDENTIFIER
 if BOOTLOADER_QA:
-    VENDORHEADER = 'embed/vendorheader/vendorheader_qa_DO_NOT_SIGN_signed_dev.bin'
+    VENDORHEADER = f'embed/vendorheader/{MODEL_IDENTIFIER}/vendorheader_qa_DO_NOT_SIGN_signed_dev.bin'
     BOOTLOADER_SUFFIX = MODEL_IDENTIFIER + '_qa'
 elif PRODUCTION:
-    VENDORHEADER = 'embed/vendorheader/vendorheader_satoshilabs_signed_prod.bin'
-    BOOTLOADER_SUFFIX = MODEL_IDENTIFIER
+    VENDORHEADER = f'embed/vendorheader/{MODEL_IDENTIFIER}/vendorheader_satoshilabs_signed_prod.bin'
+elif TREZOR_MODEL == 'R':
```
TychoVrahe, line 803: wrong order of elifs? workaround for production build of TR does not work

TychoVrahe (commented):
Prodtest VH for T2B1 is missing - is that because we don't have pubkeys yet? 

It might be better to replace satoshilabs vendor image for T2B1 with something that T2B1 can display?  Applies to unsafe VH too. It should save some flash space too.

Can't verify that pubkeys for T2B1 satoshilabs VH are correct.

matejcik:
> Prodtest VH for T2B1 is missing - is that because we don't have pubkeys yet?

yes

> It might be better to replace satoshilabs vendor image for T2B1 with something that T2B1 can display? Applies to unsafe VH too. It should save some flash space too.

also yes

> Can't verify that pubkeys for T2B1 satoshilabs VH are correct.

[notion](https://www.notion.so/satoshilabs/Trezor-signing-derivation-paths-b3b6e733d3434bd3a834ce5fc9bd557b)

core/SConscript.firmware
TychoVrahe: ah ok

matejcik:
vendor images 64x64 b&w

vendor_unsafe_new
vendor_satoshilabs_new

i would like to test them with the bootloader emulator (some changes might be needed)

TychoVrahe:
or we can just not show it. can't really fit "UNSAFE, DO NOT USE!" to right half of the screen and to split lines in the bootloader seems excessive.

matejcik:
the layout engine can split lines already :woman_shrugging: 

if we go with your design, i think the version number isn't supposed to be there?

TychoVrahe (approved):

matejcik:
noting here for posterity that we're keeping the 64x64 images for now, but will most likely change them for the final version of the bootloader

PR #3050, opened 2023-06-01
add internal model field to features

### LINKED ISSUE #3006: Add internal_model attribute to fw
related: https://github.com/trezor/trezor-suite/issues/7908

TychoVrahe:
closes #3006 

offloaded the model names to separate files `embed/models/model_xx.h` as these ifdefs will just keep growing and its cleaner to have all the names defined in one place


TychoVrahe:
rebased.

moved pubkeys to `model_xx`.h too. renamed boardloader capabilities tags to avoid name collision

mmilata (approved):
Looks straightforward. Any plans to support the field in legacy?

TychoVrahe:
probably might be useful to add to legacy too, but leaving it to separate PR - don't have the means right now to test it in legacy bootloader and since the messages there are composed as raw bytes its necessary

PR #3055, opened 2023-06-02
improve touch driver robustness

TychoVrahe:
Do i2c cycle also if one of the initialization commands fail. 3 tries before fatal error.


prusnak (approved):
utACK

hiviah (approved):
Tested again, just to make sure.

Issue #3053, opened 2023-06-02
Font change on T2B1


Hannsek:
Change body and button font to `Pixel Operator 8, regular, size 8`

Addresses/xpubs/amounts should be kept as monospace.

PR #3062, opened 2023-06-05
Font changes for T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3053

Creates a new font `FONT_BIG`, so it can accommodate one extra font into `TR`.

There may be still some mismatched fonts - a careful review of the screen difference will be needed.


Hannsek:
I would merge this. Even with "bugs". Most of the "bugs" will be solved by different issues focused on major flows which I will gradually put on Github. For the rest I will create different issue.

matejcik (approved):
LGTM code-wise.

The `FONT_XYZ` stopped being nice a little while ago, but no point stopping this PR on it. I might create an issue for an improvement.

Any breakage in TR UI tests is OK at this point IMHO, as @Hannsek says, we'll discover any actual problems soon enough.

Issue #3061, opened 2023-06-05
PIN and word entry


Hannsek:
Fully randomize the starting position and the position after each entry of PIN and word (recovery) entry 

Make last input number visible for the user so he knows that he input the correct number.

grdddj:
What about passphrase, will we randomize it as well?

PR #3064, opened 2023-06-06
Shuffle input methods for T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3061:
- shuffling the choice order in some input methods
  - PIN
    - always going to a random digit, even at the beginning
    - showing the last inputted digit in plain-text on the top
  - recovery word entry
    -  going to a random letter
    - for words, starting at the very left, but randomizing the order of words


Hannsek:
cc @andrewkozlik

Hannsek:
Please add also passphrase shuffling.

grdddj:
> Please add also passphrase shuffling.

Included now, it can be ready for testing.

One thing I was considering is that now when you press `SPACE`, it will also jump to some character category - not sure if we want this.

Hannsek:
I will try it this week. We'll see.

Hannsek:
Passphrase shuffling doesn't work on emulator.

grdddj:
> Passphrase shuffling doesn't work on emulator.

Works for me and for CI, e.g. https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4470267822/artifacts/test_ui_report/passed/TR-click_tests-test_passphrase_tr.py%3A%3Atest_passphrase_input%5Babc123ABC_%3C%3E-mtHHfh6uHtJiACwp7kzJZ97yueT6sEdQiG%5D.html

Did you pull the latest changes on this branch?

Hannsek:
After deleting the local branch and download it again, it worked. LGTM

core/embed/rust/src/trezorhal/random.rs
```diff
@@ -12,7 +12,7 @@ pub fn shuffle<T>(slice: &mut [T]) {
 
 /// Returns a random number in the range [min, max].
 pub fn uniform_between(min: u32, max: u32) -> u32 {
-    assert!(max > min);
+    assert!(max >= min);
```
matejcik, line 15: exactly as Andrew has predicted :)) good thing there's the other assert in `uniform_between_except`

tests/click_tests/test_autolock.py
```diff
@@ -195,21 +195,22 @@ def test_autolock_passphrase_keyboard(device_handler: "BackgroundDeviceHandler")
...
     elif debug.model == "R":
-        debug.press_left()  # go to BACK
-        debug.press_middle()  # PRESS back
-        debug.press_left()  # go to ENTER
-        debug.press_middle()  # press ENTER
+        # NOTE: because of passphrase randomization we just input the passphrase
+        # this way instead of "manually" clicking the buttons.
+        # (It would be a pain to go back to MENU from random position and
+        # navigating to CONFIRM - which is not in scope for this test.)
+        debug.input("j" * 8)
```
matejcik, line 211: this test is useless without actually clicking.

instead, skip the test for model R

matejcik: no, sorry, I misunderstood what's going on here. using `input` this way is fine, but then please move the `assert result` to top level (bc it's now the same for both models) and modify the comment accordingly

matejcik (approved):

Hannsek:
> > Please add also passphrase shuffling.
> 
> Included now, it can be ready for testing.
> 
> One thing I was considering is that now when you press `SPACE`, it will also jump to some character category - not sure if we want this.

As `SPACE` is from passphrase pov same character as other symbol, it should shuffle after clicking it as well.

PR #3065, opened 2023-06-06
fix(python): fix T2B1 keys in trezorlib

TychoVrahe:
we managed to put wrong keys into trezorlib  - bootloader keys were same as boardloader keys, the vendor header had what was meant to be bootloader keys. 

i am guessing that different terminology used in notion vs trezorlib does not help here: what is called "vendor header signing keys" in notion is called "bootloader keys" in trezorlib and "bootloader signing keys" in notion are "boardloader keys" in trezorlib".  

https://www.notion.so/satoshilabs/Trezor-signing-derivation-paths-b3b6e733d3434bd3a834ce5fc9bd557b


mmilata (approved):
Ah right, it's _what is signed_ vs _where verification happens_. Confirming that keys match notion.

Issue #3067, opened 2023-06-08
T2B1 - recovery fix


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/7319e858-2107-4cdf-86be-b0e32c2059bb)

Center the word between the header and the middle button.

PR #3071, opened 2023-06-13
chore(core/rust): center TR's input method vertically

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3067:
- center the choices vertically in the dedicated area

In case of PIN/wipe-code, it does not look like centered, as there is a second (invisible) prompt that is only showing e.g. `15 attempts left` at the start when previously bad PIN was inputted. We could account for this if wanted

image

image


Hannsek:
Passphrase and PIN characters should be centered between bottom button and asterixis. 👍🏻

Hannsek (approved):
Looks fine to me. Someone should probably look at the code.

Issue #3070, opened 2023-06-13
Master diff is not visible for T2B1


grdddj:
Artifacts from `unix ui changes` do not show `master diff` for `T2B1`

It seems like the `T2B1` test screens/hashes are not being uploaded properly, as the CI jobs, e.g.

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/4462195106

show a lot of these logs
```
WARNING: missing screens for TR-device_tests-test_session_id_and_passphrase.py::test_multiple_passphrases. Did the test run?
WARNING: No such recorded collection was found on 'https://data.trezor.io/dev/firmware/ui_tests/97df1d41e48484b8372bf540ccc580b9532ef50e95101876777809029207deca.zip'.
```

Issue #3066, opened 2023-06-06
CI tests sometimes miss UI report


grdddj:
Lately, it is quite common that some CI tests - be it `click` or `device` tests - do not have the `UI report` (`test_ui_report` directory) exported in the artifacts

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/4422578927/artifacts/browse

Hannsek:
It seems that the problem is only with `core device r tests` and this is happening with all jobs.

Or can it be related to pipeline #891615270?

PR #3072, opened 2023-06-13
T2B1 CI fixes

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3066
- add UI report into `TR` device tests

Fixes https://github.com/trezor/trezor-firmware/issues/3070
- add `TR` into `master-diff`

We should merge this before all other `TR` PRs, because this will upload all the missing screens (click tests) from `master`, so other PRs will have full `master-diff`


matejcik (approved):
utACK

Issue #2695, opened 2022-12-19
T2B1 design - U2F, FIDO2


Hannsek:
<img width="781" alt="image" src="https://user-images.githubusercontent.com/57008159/208421389-e3979b1d-2e35-44b9-8a41-b707ef0e2fff.png">

grdddj:
Done in [4a1be36](https://github.com/trezor/trezor-firmware/pull/2610/commits/4a1be3618d501b2f266c194cdf79b61c43bd5bf3)

![image](https://user-images.githubusercontent.com/42543243/210355754-3613813a-27ed-4606-8911-26819e52178f.png)

Hannsek:
https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?type=design&node-id=8825-68720&t=kyXA4olzpDvXBNZr-4

PR #3075, opened 2023-06-14
fix(core): add dynamic title to the FIDO dialogue for model R

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2695:
- forwarding header/title of a specific FIDO screen instead of hardcoding it in layout


matejcik (approved):
utACK

PR #3079, opened 2023-06-15
Signed vendor headers and versioned boardloader for T2B1

matejcik:
boardloader and bootloader builds are tagged to specific commits in this branch

@prusnak please merge manually as a merge-commit so that we don't lose them


prusnak:
Merge with a merge commit to protect f04aff1 commit hash

Issue #2299, opened 2022-05-26
UI2: Global layout object


matejcik:
We have already faced several issues, mostly in automated tests, where a race condition arose between two layouts competing for screen time.

The current UI code handles the competition via a "cancellation channel", but that is not 100% stable 

With Rust UI, we need a clean way to register and deregister timer events. The way the current code works, all timer events go through the same `loop.Timer` instance, which means that the same instance can be resumed multiple times, producing highly non-deterministic results unless being handled very carefully.

As an alternative, instead of spawning a `handle_timers` loop that awaits the timer object, we could schedule timer objects directly on the 

(summary tbd)

PR #3083, opened 2023-06-15
Unify touch and button handling, enable usage of both in one model

TychoVrahe:
We will need this in a future model. Only layout for TT is modifed to be able to serve both, with hope that we will unify the layout API soon.


matejcik:
why not keep the button and touch interfaces separate and just allow both to be available at the same time?

---

in general we don't really want to keep the button code active in TT and vice-versa for TR. to facilitate that, we should do to `USE_BUTTON` and `USE_TOUCH` the same thing we do to `BITCOIN_ONLY`, namely:
(a) always use `util.USE_XYZ` instead of `from util import USE_XYZ`
(b) in `site_tools/micropython/__init__.py`, statically replace `util.USE_XYZ` with True / False as appropriate

TychoVrahe:
> why not keep the button and touch interfaces separate and just allow both to be available at the same time?

not sure how to do that actually, could that be done with `loop.race`? as loop.wait() only accepts a single interface. and if using loop.race is possible for this, i kinda don't like doing the added complexity of polling in parallel

TychoVrahe:
added the replacement: https://github.com/trezor/trezor-firmware/commit/5925829ef467e7381737a59ccb358adcb82d7132

matejcik:
> not sure how to do that actually, could that be done with `loop.race`? as loop.wait() only accepts a single interface. and if using loop.race is possible for this, i kinda don't like doing the added complexity of polling in parallel

`loop.race` would work, yes, but the way you'd do it is define `handle_touch` and `handle_button` and specify both in `create_tasks()` (conditional on whether you're on touch- or button-enabled model of course)

TychoVrahe:
ok trying it that way here: https://github.com/trezor/trezor-firmware/commit/72b816af52b7c7fa31085313f83c134f5fd2b912

a bit annoying is needing to add the variants in several layout subclasses (for debug)

But mainly i am realizing here that this PR does not yet handle debug both interfaces properly, will need some more work. I don't like how button debug signal is implemented too: the middle button synthesis in TR layout seems wrong there, shouldn't it be synthesized rather in concrete tests? although it would probably require of sending PRESS and RELEASE separately, or maybe allow sending arbitrary button sequence. I would say that in this layout class we shouldn't care much about concrete buttons and their names (LEFT, RIGHT, etc), just receive a button code and send it to rust for interpretation. Thinking how to make the layout compatible with future models, where we won't have a LEFT and RIGHT button, but for example POWER button. Also not sure what 'swipe' is for in TR/button based model, seems odd.

Hannsek:
What is the state of this one?

matejcik:
it's in progress as part of #2299

matejcik:
closing in favor of #2335

Issue #3080, opened 2023-06-15
T2B1 tutorial


Hannsek:
https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?type=design&node-id=8825-71196&t=kkbHmTnhtxYhd0gk-4

Issue #3016, opened 2023-05-15
T2B1: regenerate icons with odd height


grdddj:
In the past, there was a limitation that the icons need to have even height. Because of that, some icons have now an empty bottom row, to satisfy this limitation. We can regenerate those to make them odd-heighted

Hannsek:
Is this done also for bootloader?

grdddj:
> Is this done also for bootloader?

Bootloader is using the majority of icons from firmware, so it benefits from this as well.

There are some extra icons like these

![image](https://github.com/trezor/trezor-firmware/assets/42543243/2be18349-a17f-45b6-a1b8-ad3d0ba86236)

that I cannot find as `png` icons (in `core/assetrs/model_r`). These work fine however, so no necessary action there

Hannsek:
Ok, eventually, we can put there pngs of these icons.

PR #3084, opened 2023-06-16
T2B1 tutorial and buttons

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3080:
- modified tutorial copy and other smaller changes (buttons, titles)

To make the buttons  - Fixes https://github.com/trezor/trezor-firmware/issues/3016:
- regenerating icons to match new designs and not to have an empty bottom row

One screen is not exact as in `Figma` - the `...doesn't fit` - it is not centered vertically as that would require a special layout for this screen. Same thing with the left button - it is the last page of the two-page flow, so it does not have "fat" up arrow on the left, but just "slim"

image


Hannsek:
> utACK for code

this was the fastest review I ever saw. 🧙🏻‍♂️

grdddj:
> I belive we can _fix_ the vertical centering by addind header: image

Unfortunately, this is not possible currently, there is no concept of "header" in this flow - it is just hardcoded `bold` text at the top and then the rest in `normal` font 

> Btw, I don't thing the body text is centered between the header and pressed button: image


Right, it might not be precisely centered - I have just put some pixel margin on top "from the eye" as I was not sure it is precisely centered in `Figma`. Do we want it or is this good enough?

Hannsek:
> Unfortunately, this is not possible currently, there is no concept of "header" in this flow - it is just hardcoded bold text at the top and then the rest in normal font

Can we introduce the concept of header in this flow? We have headers in other flows so why not in this one?

> Right, it might not be precisely centered - I have just put some pixel margin on top "from the eye" as I was not sure it is precisely centered in Figma. Do we want it or is this good enough?

Ok, but yeah, we want this everywhere. I thought that we center the _body_ text on every screen, at least on TT (@mmilata)

grdddj:
> > Unfortunately, this is not possible currently, there is no concept of "header" in this flow - it is just hardcoded bold text at the top and then the rest in normal font
> 
> Can we introduce the concept of header in this flow? We have headers in other flows so why not in this one?
> 
> > Right, it might not be precisely centered - I have just put some pixel margin on top "from the eye" as I was not sure it is precisely centered in Figma. Do we want it or is this good enough?
> 
> Ok, but yeah, we want this everywhere. I thought that we center the _body_ text on every screen, at least on TT (@mmilata)

Both the header and centering introduced in recent commits

image

grdddj:
> Nice, perfect.👌🏻 Just please add a line to the second _screen scroll_ screen to look like this: image

Done during the rebase and force-push. I can already merge this if you agree.

(There is this difference in line count, it will fit on Trezor on one line - hope it is not a problem)

image

Hannsek:
Btw how big is the gap between individual letters in fonts on TR?

grdddj:
> Btw how big is the gap between individual letters in fonts on TR?

Seems that one pixel - at least here in regular and bold - I guess same everywhere else (but is attribute of the font)

image

Issue #3090, opened 2023-06-18
PIN design


Hannsek:
https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?type=design&node-id=8825-71199&t=4gMBHfDOelVRGYl9-4

grdddj:
![image](https://github.com/trezor/trezor-firmware/assets/42543243/88d0afb0-ed9f-4ddb-968e-823b79dd7ef2)

Here, we would currently show the last inputted digit - `**3` - is it OK, or the inputted `3` should only be visible right after choosing it, and every subsequent interaction should hide it?

Hannsek:
I think it is better tho hide it with every subsequent interaction. What do you think?

grdddj:
> I think it is better tho hide it with every subsequent interaction. What do you think?

Agreed, it is safer and I hope will be enough confirmation for the user that the inputted digit is there. Implementation-wise it looks doable

grdddj:
I am missing updated screens/copy for PIN disabling and PIN change.

Also, wipe code specification is not there.

grdddj:
> Can you please provide me with desired screens that should be changed?

Sure, these are current texts and buttons for PIN and wipe-code that do not have Figma designs:

```
description="Do you want to disable PIN protection?",
verb="DISABLE",

description="Do you want to change your PIN?",
verb="CHANGE",

description="Do you want to disable wipe code protection?",
verb="DISABLE",

description="Do you want to change the wipe code?",
verb="CHANGE",

description="Do you want to enable wipe code?",
verb="ENABLE",
```

grdddj:
Merged in https://github.com/trezor/trezor-firmware/pull/3135

PR #3096, opened 2023-06-20
T2B1 PIN dialogs

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3090
- improve the designs of PIN screens

Missing designs:
- [ ] remove + update PIN dialog
- [ ] wipe code screens

This time I did not record any UI diffs, so all the changes so far are visible in `differing_screens.html`


Hannsek:
image
"turn on" instead of "enable:

Please add vspace between first 2 paragraphs. 

it should look like this: 
image


instead of this: image


vertical space between upper border and header: 
image
Please delete it.

It seems that this copy is not centered: 
image

"turned" on the next line please: 
image

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -748,6 +748,37 @@ extern "C" fn tutorial(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_centered_popup_middle_button(
```
matejcik, line 751: shouldn't this be `show_error` or whichever the equivalent on TT side is?

(in general please try to use semantic names. when we change the layout of "show error", we don't want to be renaming this to `leftaligned_popup_with_right_icon`)

grdddj: Right, renamed to `show_error` in [f813823](https://github.com/trezor/trezor-firmware/pull/3096/commits/f813823b3cfaaf6c6425e630335d01f32267b808)

I am implementing very similar functionality on other current PRs, so after all is merged, we might want to do some review of the whole Rust API

grdddj:
@Hannsek  Done all the changes in [230f59c](https://github.com/trezor/trezor-firmware/pull/3096/commits/230f59c7e871163008e5b1f626eeb3f680b8e861)

Regarding the not-centered wipe-code warning, we do center it vertically, between the top and buttons. It does not look like centered here, because the buttons take up to 13px normally (in height), but here only 9px height is visible

grdddj:
> image -> image
> 
> Header should be uppercase image
> 
> When there is more characters/numbers thank can fit on the screen, currently, there are three dots. Please remove it and let there only asterisks.

Done these in [00e7034](https://github.com/trezor/trezor-firmware/pull/3096/commits/00e703433d0542c96dc2a16c0b7991fa5086a9b2), together with removing the second set of screen (4-50 digits etc.) from wipecode setting

Hannsek:
The "success" screen should look like this: 
image
not like this
image

This also applies for other "success" screen. Should I create a separate issue for that?

image
The dash ("_") or the center number is not horizontally centered.
image

The new line for "wipe code" is not necessary (this also applies for model T), but I will put here the designs for wipe code later.

matejcik:
persistence test for TT is broken:
image
(fwiw this image doesn't look right, there should not be a line break before "wipe code", right?

Hannsek:
RIght, it was like this in model R before for some reason. It is addressed in the comment above.

matejcik:
ACK on code side.

I'm still going to be obstinate about the "enable PIN" screens. Four is _not_ an acceptable number of screens before getting to do the thing I'm doing.

grdddj:
> The "success" screen should look like this: image not like this image
> 
> This also applies for other "success" screen. Should I create a separate issue for that?

This layout - gray colors instead of green - should be used in "success screens for turning security things off"? If so, please create a new issue, it is its own thing.

grdddj:
I have added the newline because of this screen, because it cannot fit one line completely:

image

Deleted the newline now

Hannsek:
> This layout - gray colors instead of green - should be used in "success screens for turning security things off"?

All success screens should look like this.

grdddj:
> 

So

> > This layout - gray colors instead of green - should be used in "success screens for turning security things off"?
> 
> All success screens should look like this.

So that really deserves a separate issue for "TT success screens"

Hannsek:
https://github.com/trezor/trezor-firmware/issues/3133

grdddj:

Issue #3091, opened 2023-06-18
Passphrase design


Hannsek:
https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?type=design&node-id=8825-68715&t=4gMBHfDOelVRGYl9-4

matejcik:
![image](https://github.com/trezor/trezor-firmware/assets/596073/a16bdb8f-f4e2-4f90-bb98-89f0235ec61a)
can't be done.
the decision where the passphrase is entered comes from the host.
(or you can preconfigure it by "force on device")

matejcik:
Only one side can decide, the other has to wait for decision.

the conversation goes like this:
Host -> Trezor: `GetAddress`
Trezor -> Host: `PassphraseRequest` (and shows "please enter passphrase on host")
(Trezor waits for response)
Host -> Trezor: `PassphraseAck("helloworld")` **or** `PassphraseAck(on_device=true)`
Trezor -> Host: `Address("tb1foobarbaz")`

an alternative is what we had originally:
Host -> Trezor: `GetAddress`
Trezor -> Host: `ButtonRequest` ( = wait for me to ask the user)
Host -> Trezor: `ButtonAck` ( = i'm waiting for your response)
(Host waits)
Trezor -> Host: `PassphraseRequest` (user decided to enter on host)
Host -> Trezor: `PassphraseAck("helloworld")`
Trezor -> Host: `Address("tb1foobarbaz")`

problem is, the conversation **must** be request/response. Trezor can't up and interject "actually I changed my mind, here's a different response". Technically we could do something about it but it would be a significant change to the whole protocol and Connect would hate us.

(we can consider it for the encrypted session protocol, while we're breaking things anyway. but even then it does make things more difficult for both sides)

mayyyyybe the host could ping the Trezor repeatedly, like, instead of sending the passphrase or "on_device", every 200 ms it sends `PassphraseAck(did the user click the button?)` and the Trezor can either respond `PassphraseRequest` (no they didn't) or `ButtonRequest` (actually they did and they're entering the passphrase right now).
but that's also a relatively big change

matejcik:
also then we have the nasty cornercase of the user entering half the passphrase on host, and then deciding to press the button on Trezor

grdddj:
![image](https://github.com/trezor/trezor-firmware/assets/42543243/0968db79-b215-41af-90e9-a9f5b78350c1)

[...]

PR #3098, opened 2023-06-21
T2B1 passphrase flow

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3091:
- small design changes on screens around passphrase dialog - e.g. `turn on` button, centering the "input passphrase" text

---

EDIT:
Differing screens:
- [click tests R](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4512703306/artifacts/test_ui_report/differing_screens.html)
- [click tests](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4512703303/artifacts/test_ui_report/differing_screens.html)
- [device tests](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4512703282/artifacts/test_ui_report/differing_screens.html)
- [device tests R](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4512703285/artifacts/test_ui_report/differing_screens.html) - there are bugs, to be fixes soon

So that the above-links do not need to be generated manually (and are not outdated after a new commit), a story was created - https://github.com/trezor/trezor-firmware/issues/3099

---

EDIT2:
first try of the auto-generated and auto-updated links: http://grdddj.eu:8002/gitlab/grdddj/tr_passphrase

### DIFF
--- a/core/mocks/generated/trezorui2.pyi
+++ b/core/mocks/generated/trezorui2.pyi
@@ -185,6 +185,11 @@ def show_info(
 
 
 # rust/src/ui/model_tr/layout.rs
+def show_passphrase() -> object:
+    """Show passphrase on host dialog."""
+
+
+# rust/src/ui/model_tr/layout.rs
 def show_mismatch() -> object:
     """Warning modal, receiving address mismatch."""
 
--- a/core/embed/rust/librust_qstr.h
+++ b/core/embed/rust/librust_qstr.h
@@ -108,6 +108,7 @@ static void _librust_qstrs(void) {
   MP_QSTR_show_info;
   MP_QSTR_show_lockscreen;
   MP_QSTR_show_mismatch;
+  MP_QSTR_show_passphrase;
   MP_QSTR_show_progress;
   MP_QSTR_show_progress_coinjoin;
   MP_QSTR_show_remaining_shares;
--- a/core/embed/rust/src/ui/model_tr/layout.rs
+++ b/core/embed/rust/src/ui/model_tr/layout.rs
@@ -900,6 +900,17 @@ extern "C" fn new_show_info(n_args: usize, args: *const Obj, kwargs: *mut Map) -
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_show_passphrase() -> Obj {
+    let block = move || {
+        let text: StrBuffer = "Please enter your passphrase.".into();
+        let paragraph = Paragraph::new(&theme::TEXT_NORMAL, text).centered();
+        let content = Paragraphs::new([paragraph]);
+        let obj = LayoutObj::new(content)?;
+        Ok(obj.into())
+    };
+    unsafe { util::try_or_raise(block) }
+}
+
 extern "C" fn new_show_mismatch() -> Obj {
     let block = move || {
         let get_page = move |page_index| {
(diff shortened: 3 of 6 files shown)


Hannsek (approved):

Hannsek:
UI wise ok. 👍🏻

core/embed/rust/src/ui/model_tr/component/frame.rs
```diff
@@ -66,8 +66,14 @@ where
     fn place(&mut self, bounds: Rect) -> Rect {
         const TITLE_SPACE: i16 = 2;
 
-        let (title_area, content_area) = bounds.split_top(theme::FONT_HEADER.line_height());
-        let content_area = content_area.inset(Insets::top(TITLE_SPACE));
+        // Title might be empty, in that case allocating whole area for the content
+        let (title_area, content_area) = if self.title.get_text().is_empty() {
```
matejcik, line 70: could we instead convert title to `Option`?

core/mocks/generated/trezorui2.pyi
```diff
@@ -178,6 +178,7 @@ def show_info(
     title: str,
     description: str = "",
     time_ms: int = 0,
+    centered: bool = False,
```
matejcik, line 181: when do we want this to be false and when true? can you show examples?

grdddj: Normally it is false, putting that to true in this screen:
image

core/embed/rust/src/ui/model_tr/component/frame.rs
grdddj: Theoretically yes, but I would rather leave it like this - it is just one corner-case when the title is empty.

I would say it is not worth complicating the code because of it.

matejcik: but this way there is hidden behavior (that you need to explicitly sidestep elsewhere)

given the "centered" option, i'm thinking maybe the "show passphrase" screen should be its own thing instead of hacking "show_info" to support it

grdddj: Right, will go with new `show_passphrase` for now. After merging all the  new`T2B1` designs, we might want to take a closer look at the API and maybe unify it more with `TT` - we use `show_simple` in that case (but also getting optional title).

grdddj: Done in [48bf66b](https://github.com/trezor/trezor-firmware/pull/3098/commits/48bf66b51097fc92c1fcfc2963c4552a9a236250)

core/embed/rust/src/ui/model_tr/component/address_details.rs
```diff
@@ -58,7 +58,9 @@ where
             Paragraphs::new(para)
         };
         let xpub_view = Frame::new(
-            "".into(),
+            // Need to supply a non-empty temporary title, otherwise it would not allocate space
+            // for it
+            "FILL ME".into(),
```
matejcik, line 63: you can also revert this, right?

matejcik (approved):
one small nit otherwise ACK

core/embed/rust/src/ui/model_tr/component/address_details.rs
grdddj: Correct, reverted in the force-push

grdddj:

PR #3100, opened 2023-06-21
Iter changes

matejcik:
Fixes test failure in nightly ASAN jobs by disabling Micropython's built-in stack checking in the Rust test suite (where we don't actually need it for anything).

That is commit 81051d1368855e7cb88453d92e79c8ce9108bc9d, the rest is just gravy.

Namely:

17e25c60cab78b0b11a4e35fdfa4f54d0707fab1 allows easier examination of arbitrary `Obj`s, in particular in debug builds where you can retrieve the name of the type.

3f228c368e61e392830e2d31e7c6d7cef5c72f81 moves the iteration method from `Iter` to `IterBuf`, streamlining from:
```
let mut iter_buf = IterBuf::new();
let iter = Iter::try_from_obj_with_buf(obj, &mut iter_buf)?;
for x in iter { ... }
```
to:
```
for x in IterBuf::new().try_iterate(obj)? { ... }
```

0cc15629e853c3c1646312c27cc9b4a332d9d1b9 fixes an oversight that prevented our print!/println! macros from working in tests

2806b36a90fbdfe7bb17736a56ea4f3cbbab87e1 silences a clippy complaint

### DIFF
--- a/core/embed/rust/src/micropython/testutil.rs
+++ b/core/embed/rust/src/micropython/testutil.rs
@@ -1,6 +1,6 @@
 extern "C" {
     fn gc_init(start: *mut cty::c_void, end: *mut cty::c_void);
-    fn mp_stack_ctrl_init();
+    fn mp_stack_set_top(top: *mut cty::c_void);
     fn mp_stack_set_limit(limit: usize);
     fn mp_init();
 }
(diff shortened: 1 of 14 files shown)


core/embed/rust/src/micropython/testutil.rs
```diff
@@ -11,8 +11,8 @@ static mut MPY_INITIALIZED: bool = false;
 /// Initialize the MicroPython environment.
 ///
 /// This is very hacky, in no way safe, and should not be used in production.
-/// The stack is configured on a best-effort basis and depending on from where
-/// this is called, you might get errorneous "recursion exceeded" problems.
+/// The stack is configured to span all of memory, effectively disabling the
+/// stack guard. I have no idea what can happen.
```
mmilata, line 15: Curiously, it doesn't seem to affect scanning for GC roots, if I'm reading the code correctly. Doesn't really matter for unit for tests though.

mmilata (approved):
Nice!

Issue #3089, opened 2023-06-18
Wallet creation, seed backup design


Hannsek:
https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Trezor-T%2C-Trezor-One?type=design&node-id=8825-71197&t=4gMBHfDOelVRGYl9-4

Hannsek:
Thanks for that. 

As pagination on the `STANDARD BACKUP` page would "ruin" the header, let's not put it there.
![image](https://github.com/trezor/trezor-firmware/assets/57008159/db8e8fb5-1b8c-42e1-abbe-7f78d1f11b38)

> Also, with the left-arrow, just want to confirm (as I see it in other designs as well) that we have changed its appearance - it is always a "slim" arrow instead of the "fat" arrow as before?

Yes, with the one-fat-arrow design, the user is more guided in which direction he/she should continue.

grdddj:
> But user will be able to go up to the `STANDARD BACKUP` page by the up arrow. Is it ok from the code pov? @grdddj

I think it is fine implementation-wise.

To which point would the user be redirecting when selecting `CHECK AGAIN` after an unsuccessful "quiz" attempt?

Hannsek:
I'd say to this one below as the user wants to see the words. 
![image](https://github.com/trezor/trezor-firmware/assets/57008159/43da2eff-58d2-40c3-b77c-8d1ec5ee6447)

grdddj:
> I'd say to this one below as the user wants to see the words. ![image](https://user-images.githubusercontent.com/57008159/246887752-43da2eff-58d2-40c3-b77c-8d1ec5ee6447.png)

For simplicity purposes, currently, it moves to the previous page (`Write down all XX words`) - it is an edge case (caused by user error) and it would unnecessarily complicate the code to move in the middle of one flow

Hannsek:
Wallet creation - Shamir backup design: https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?type=design&node-id=8825-71198&mode=design&t=S14phM8hFc2KN54k-4

grdddj:
Merged in https://github.com/trezor/trezor-firmware/pull/3135

PR #3107, opened 2023-06-26
T2B1 wallet creation and seed backup

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3089:
- improve the designs of wallet creation and seed backup flows

### DIFF
--- a/core/embed/rust/src/ui/component/text/op.rs
+++ b/core/embed/rust/src/ui/component/text/op.rs
@@ -97,15 +97,13 @@ impl<'a, T: StringType + Clone + 'a> OpTextLayout<T> {
                     };
                 }
                 // Drawing text
-                Op::Text((text, is_complete)) => {
+                Op::Text(text, continued) => {
                     // Try to fit text on the current page and if they do not fit,
                     // return the appropriate OutOfBounds message
 
                     // Inserting the ellipsis at the very beginning of the text if needed
-                    // (just once for the first Op::Text on the non-first page).
-                    // Also just for incomplete texts that were separated.
-                    self.layout.continues_from_prev_page =
-                        skip_bytes > 0 && total_processed_chars == 0 && !is_complete;
+                    // (just for incomplete texts that were separated)
+                    self.layout.continues_from_prev_page = continued;
 
                     let fit = self.layout.layout_text(text.as_ref(), cursor, sink);
 
@@ -147,17 +145,17 @@ impl<'a, T: StringType + Clone + 'a> OpTextLayout<T> {
         let mut skipped = 0;
         ops_iter.filter_map(move |op| {
             match op {
-                Op::Text((text, _is_complete)) if skipped < skip_bytes => {
+                Op::Text(text, _continued) if skipped < skip_bytes => {
                     let skip_text_bytes_if_fits_partially = skip_bytes - skipped;
                     skipped = skipped.saturating_add(text.as_ref().len());
                     if skipped > skip_bytes {
                         // Fits partially
                         // Skipping some bytes at the beginning, leaving rest
-                        // Signifying that the text is not complete
-                        Some(Op::Text((
+                        // Signifying that the text continues from previous page
+                        Some(Op::Text(
                             text.skip_prefix(skip_text_bytes_if_fits_partially),
-                            false,
-                        )))
+                            true,
+                        ))
                     } else {
                         // Does not fit at all
                         None
@@ -187,7 +185,7 @@ impl<T: StringType + Clone> OpTextLayout<T> {
     }
 
     pub fn text(self, text: T) -> Self {
-        self.with_new_item(Op::Text((text, true)))
+        self.with_new_item(Op::Text(text, false))
     }
 
     pub fn newline(self) -> Self {
@@ -241,9 +239,9 @@ impl<T: StringType + Clone> OpTextLayout<T> {
 #[derive(Clone)]
 pub enum Op<T: StringType> {
     /// Render text with current color and font.
-    /// Bool signifies whether the text is complete or not
+    /// Bool signifies whether the text continues or not
     /// (so we know whether to show ellipsis or not).
-    Text((T, bool)),
+    Text(T, bool),
     /// Set current text color.
     Color(Color),
     /// Set currently used font.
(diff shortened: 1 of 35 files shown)


Hannsek:
I've found following inaccuracies: 

image
image
image
image

not centered

align numbers to left side: 
image

Can we get rid of the "processing" text above? 
image

Nitpick: 
It seems to me that the space between the center and right number is bigger then between center and left number.

The carousel is not centerd between the pressed button and header: 
image
image

Hannsek:
image
This screen should look differently.
image

grdddj:
TT screen and number allignment solved in [7772222](https://github.com/trezor/trezor-firmware/pull/3107/commits/7772222e400f3bcb3b5b734eed1ab3a22a66e749)

As for the general centering, I have created a special branch for it - https://github.com/trezor/trezor-firmware/tree/grdddj/tr_content_centering

We are currently having 2px margin below the header, so it is visually delimited from the content in all times. I am deleting this margin there, so some screens change. Let's go through the changes together and decide whether we want it

Hannsek:
image
@andrewkozlik from security pov, should these numbers be shuffled as well when user is choosing the number of shamir shares?

core/embed/rust/src/ui/component/text/op.rs
```diff
@@ -238,7 +241,9 @@ impl<T: StringType + Clone> OpTextLayout<T> {
 #[derive(Clone)]
 pub enum Op<T: StringType> {
     /// Render text with current color and font.
-    Text(T),
+    /// Bool signifies whether the text is complete or not
+    /// (so we know whether to show ellipsis or not).
+    Text((T, bool)),
```
matejcik, line 246: why not make a separate op for ellipsis instead?

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -761,11 +792,45 @@ extern "C" fn tutorial(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj
...
-                .with_scrollbar(false)
-                .with_common_title("HELLO".into()),
-        )?;
+        let obj = LayoutObj::new(Flow::new(pages).with_scrollbar(false))?;
+        Ok(obj.into())
+    };
+    unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
+}
+
+extern "C" fn new_popup_middle_button(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
```
matejcik, line 801: again, shouldn't "popup middle button" be something like "show_info"?

core/embed/rust/src/ui/component/text/op.rs
grdddj: How would it then look like? As we cannot know beforehand at which position the ellipsis will be?

To make sure, the use-case of this change is the following:
- 1. we want to show both ellipses (forward and backward one) when we are splitting coherent text (single `Op::Text`, basically one sentence) into more pages

image

- 2. we do not want to show any ellipsis when one coherent text ends at the end of one page and second `Op::Text` (sentence) starts at the beginning of the second page

image

andrewkozlik:
> should these numbers be shuffled as well when user is choosing the number of shamir shares?

I think we don't necessarily need to shuffle these numbers.

core/embed/rust/src/ui/component/text/op.rs
matejcik: i have a _slightly_ cleaner solution in mind, but it seems to require larger modifications to the Op processing pipeline, for which this is not a good place

for now, just modify the `Text` to not use the tuple. Rename `is_complete` to `continued` and update the comments accordingly, to make it clear that this only happens when it's a `Text` item that is continuing from previous page.

core/embed/rust/src/ui/component/text/op.rs
```diff
@@ -97,14 +97,15 @@ impl<'a, T: StringType + Clone + 'a> OpTextLayout<T> {
...
                 // Drawing text
-                Op::Text(text) => {
+                Op::Text((text, is_complete)) => {
                     // Try to fit text on the current page and if they do not fit,
                     // return the appropriate OutOfBounds message
 
                     // Inserting the ellipsis at the very beginning of the text if needed
                     // (just once for the first Op::Text on the non-first page).
+                    // Also just for incomplete texts that were separated.
                     self.layout.continues_from_prev_page =
```
matejcik, line 107: we can directly set `self.layout.continues_from_prev_page = continued` (or `!is_complete` in the current version of this code), right? there is no way to get `Text(t, continued=true)` when we already processed some chars

grdddj:
> image should look like this: image
> 
> Why this screen is there? image it should jump right on this screen image
> 
> image -> image
> 
> image -> image this should be the "done" screen image
> 
> image
> 
> BACKUP CHECK instead of SEED CHECK
> 
> image Let's change the font to regular Uppercase "continue".
> 
> image Uppercase "check again".
> 
> image -> image

TR-related things sorted out in https://github.com/trezor/trezor-firmware/pull/3119

For the TT-related things, I think this also deserves its own issue, like "TT recovery flow"

core/embed/rust/src/ui/component/text/op.rs
grdddj: Done in [3006d28](https://github.com/trezor/trezor-firmware/pull/3107/commits/3006d285ae8e2eba439680480ffd48cc3b176dae)

Hannsek:
image
This looks strange as I suppose the "new amount" value is on the next screen.

grdddj:
> image This looks strange as I suppose the "new amount" value is on the next screen.

Added an extra newline in https://github.com/trezor/trezor-firmware/pull/3135

image

core/embed/rust/src/ui/component/text/op.rs
```diff
@@ -238,7 +239,9 @@ impl<T: StringType + Clone> OpTextLayout<T> {
 #[derive(Clone)]
 pub enum Op<T: StringType> {
     /// Render text with current color and font.
-    Text(T),
+    /// Bool signifies whether the text continues or not
+    /// (so we know whether to show ellipsis or not).
```
matejcik, line 243:
```suggestion
    /// Bool signifies whether this is a split Text op continued from previous page.
    /// If true, a leading ellipsis will be rendered.
```

grdddj:

PR #3108, opened 2023-06-26
trezorhal refactoring

TychoVrahe:
- restructures trezorhal to allow easy adding different platforms than stm32f4
- moves unix HW implementation to trezorhal/unix
- moves platform selection to site_scons/boards

(also jpg_to_h is moved to tools and arguement is added)


core/embed/rust/build.rs
```diff
@@ -77,24 +84,28 @@ fn prepare_bindings() -> bindgen::Builder {
...
-            "-I../../build/firmware",
-            "-I../../vendor/micropython/lib/stm32lib/STM32F4xx_HAL_Driver/Inc",
-            "-I../../vendor/micropython/lib/stm32lib/CMSIS/STM32F4xx/Include",
-            "-I../../vendor/micropython/lib/cmsis/inc",
-            "-DUSE_HAL_DRIVER",
-            "-DSTM32_HAL_H=<stm32f4xx.h>",
-        ]);
+        let mut clang_args: Vec<&str> = Vec::new();
+
+        let includes = env::var("RUST_INCLUDES").unwrap();
```
mmilata, line 96: This will probably fail when running some `cargo ...` commands directly. I don't have better solution though, we can leave it for now and solve it when needed.

mmilata (commented):
Generally looking good. Firmware/bootloader digest doesn't match but that can be explained by reordering of the source files.

However the bootloader image became 1536 bytes larger - perhaps there's something extra (statics?) in `SOURCE_HAL` that wasn't there before?

TychoVrahe:
Reordering files is indeed cause of differing binaries. 

The 1.5 kB in bootloader was due to compiled DMA handler, removed that here: https://github.com/trezor/trezor-firmware/commit/eea072759c6c2f1663e26b4a14d966e738355069

There will still be some small increases in flash size: the `ensure` macro produces error screen with filename a line number and due to the additional `stm32f4` folder the paths are a bit longer, resulting in larger binaries

mmilata (approved):

Issue #3099, opened 2023-06-21
Autogenerate links for all UI changes connected with a PR


grdddj:
It would be super nice if we would automate the process of getting the whole UI diff report for a specific PR/branch, including all the changes it made.

Currently, the reviewer needs to spend quite some effort on opening the latest CI jobs and go through at least 4 reports (T+R x click+device). Yes, `master_diff` should unite all these reports, but only in case all (4) tests succeeded - otherwise the report is not generated (maybe this is low-hanging fruit, generate it anyway, if possible?).

Preferably, it would be just one link for the whole branch, always resolving to the latest commit on the branch.

In the past, something like this (referencing the latest CI run of a branch) could be done (I think?) only for `master` - but might be already available to all branches (would need to look into `Gitlab` API).

If we would have that static link, it could be even pre-filled in the PR description (like we have prefilled stuff in the issue description).

grdddj:
Notes from brainstorming about this problem:
- it would be nice to have these links in the `PR` itself, most preferably on the bottom combined with other `GH Actions`
  - OR some `GH bot` writing a comment with the differences between `master::fixtures.json` and `branch::fixtures.json`
- there is an issue with getting _all_ the master-differing-screens from every `PR`, as some tests may fail, some may have already recorded fixtures, etc.
  - to combat this, we might want to add `master-diff` report on every UI test job and then link to this report (to 4 of them)

PR #3109, opened 2023-06-27
feat(tests): running master-diff report after each UI test

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3099:
- mostly code reorganization (moving things to `common.py`) to allow creating master-diff report after each individual UI test
  - each UI test will have `master_diff.html` in the main report dir
- added new `--do-master-diff` `pytest` option to do this (`False` by default)
- master screens are being cached locally in `tests/ui_tests/reporting/master_cache` to allow for quick results and less traffic

### DIFF
--- a/ci/prepare_ui_artifacts.py
+++ b/ci/prepare_ui_artifacts.py
@@ -5,10 +5,11 @@ from pathlib import Path
 ROOT = Path(__file__).resolve().parent.parent
 sys.path.insert(0, str(ROOT))
 # Needed for setup purposes, filling the FILE_HASHES dict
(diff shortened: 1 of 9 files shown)


matejcik:
what i'm missing here is the "normal" master diff that lists the tests that differed
otherwise ACK

grdddj:
> what i'm missing here is the "normal" master diff that lists the tests that differed otherwise ACK

Done in [e289db2](https://github.com/trezor/trezor-firmware/pull/3109/commits/e289db2bfb0aa28f2c8133852efd4a87b18fbb31)

matejcik (approved):

Issue #3077, opened 2023-06-14
Unstable device tests


grdddj:
`tests/device_tests/bitcoin/test_signtx.py::test_information_replacement` for `TT` is timing out/freezing sometimes - happens to me locally and at least two times in recent CI jobs:

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/4472147414

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/4465296836

EDIT: also `tests/device_tests/test_busy_state.py::test_busy_expiry` seems to be unstable sometimes:

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/4473590546

EDIT2: also `test_lots_of_inputs` is timing out sometimes

PR #3110, opened 2023-06-27
Introduce a pseudo-global context

matejcik:
Related to #2299.

This PR introduces `wire.context` module that exposes functions `call`, `call_any`, `wait`, and `maybe_call` (which does not care whether there is a context to send the call to, doing what `DUMMY_CONTEXT` was doing previously).

The nice thing about this is that we no longer need to pass `Context` around, because whenever someone needs it, `wire.context.call()` will automagically pick the right context object.

This is now ready for review, but not yet ready to merge. I still need to add documentation for `wire.context`. Also some UI tests are failing, not sure if the problem is intermittent or caused by this.


core/src/trezor/ui/layouts/common.py
```diff
@@ -26,21 +26,20 @@ async def button_request(
         log.debug(__name__, "ButtonRequest.type=%s", br_type)
     workflow.close_others()
     if pages is not None:
-        await ctx.call(ButtonRequest(code=code, pages=pages), ButtonAck)
+        await context.maybe_call(ButtonRequest(code=code, pages=pages), ButtonAck)
     else:
```
grdddj, line 30: Instead of this if-else, just single `await context.maybe_call(ButtonRequest(code=code, pages=pages), ButtonAck)` should be enough

core/src/trezor/ui/layouts/common.py
```diff
@@ -26,21 +26,20 @@ async def button_request(
...
     br_type: str,
     br_code: ButtonRequestType = ButtonRequestType.Other,
 ) -> Any:
     if hasattr(layout, "page_count") and layout.page_count() > 1:  # type: ignore [Cannot access member "page_count" for type "LayoutType"]
         # We know for certain how many pages the layout will have
-        await button_request(ctx, br_type, br_code, pages=layout.page_count())  # type: ignore [Cannot access member "page_count" for type "LayoutType"]
-        return await ctx.wait(layout)
+        await button_request(br_type, br_code, pages=layout.page_count())  # type: ignore [Cannot access member "page_count" for type "LayoutType"]
+        return await context.wait(layout)
     else:
```
grdddj, line 43: The code we call is almost the same in both if-else branches, we might simplify it a little, e.g.:

```python
if hasattr(layout, "page_count") and layout.page_count() > 1:
    page_count = layout.page_count()
else:
    page_count = None
await button_request(br_type, br_code, pages=page_count)
return await context.wait(layout)
```

core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -1168,12 +1100,10 @@ def request_passphrase_on_host() -> None:
...
 
-async def request_passphrase_on_device(ctx: GenericContext, max_len: int) -> str:
-    await button_request(
-        ctx, "passphrase_device", code=ButtonRequestType.PassphraseEntry
-    )
+async def request_passphrase_on_device(max_len: int) -> str:
+    await button_request("passphrase_device", code=ButtonRequestType.PassphraseEntry)
 
-    result = await ctx.wait(
+    result = await ctx_wait(
```
grdddj, line 1106: Not sure why we do not call `interact` here (instead of manually calling `button_request` and then `ctx_wait`).

Seems like my fault, probably we were calling the result many times and then changed it to just once.

core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -1217,21 +1146,19 @@ async def request_pin_on_device(
         )
     )
 
-    result = await ctx.wait(dialog)
+    result = await ctx_wait(dialog)
```
grdddj, line 1149: Here we may also use `interact` as above

core/src/trezor/ui/layouts/tr/recovery.py
```diff
@@ -7,15 +7,10 @@
...
-if TYPE_CHECKING:
-    from trezor.wire import GenericContext
-    from typing import Iterable, Callable
 
-
-async def request_word_count(ctx: GenericContext, dry_run: bool) -> int:
-    await button_request(ctx, "word_count", code=ButtonRequestType.MnemonicWordCount)
+async def request_word_count(dry_run: bool) -> int:
+    await button_request("word_count", code=ButtonRequestType.MnemonicWordCount)
     count = await interact(
```
grdddj, line 13: Do we send two duplicate button requests here - one in `button_request` and second in `interact`?

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -1202,15 +1136,13 @@ def request_passphrase_on_host() -> None:
...
-        ctx, "passphrase_device", code=ButtonRequestType.PassphraseEntry
-    )
+async def request_passphrase_on_device(max_len: int) -> str:
+    await button_request("passphrase_device", code=ButtonRequestType.PassphraseEntry)
 
     keyboard = RustLayout(
         trezorui2.request_passphrase(prompt="Enter passphrase", max_len=max_len)
     )
-    result = await ctx.wait(keyboard)
+    result = await ctx_wait(keyboard)
```
grdddj, line 1145: May use `interact` here

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -1244,26 +1175,24 @@ async def request_pin_on_device(
             wrong_pin=wrong_pin,
         )
     )
-    result = await ctx.wait(dialog)
+    result = await ctx_wait(dialog)
```
grdddj, line 1178: `interact` here

grdddj (commented):
Very nice from both the readability and code-size perspective (decreased binary size by 1024 bytes)!

I have just some small comments, that are maybe not directly connected to these changes, but which were uncovered by it (mostly using `interact` instead of manually doing `button_request` and `ctx_wait`).

As for the failing `CI` UI tests, it is happening quite often lately - I guess it increases the priority of https://github.com/trezor/trezor-firmware/issues/3077

mmilata (commented):
PRs that remove more than they add are the best

core/mocks/generated/trezorproto.pyi
```diff
@@ -1,34 +1,47 @@
...
+# Note that MessageType "subclasses" are not true subclasses, but instead instances
+# of the built-in metaclass MsgDef. MessageType instances are in fact instances of
+# the built-in type Msg. That is why isinstance checks do not work, and instead the
+# MessageTypeSubclass.is_type_of() method must be used.
+
+
+# rust/src/protobuf/obj.rs
+class MessageType:
+    MESSAGE_NAME: ClassVar[str] = "MessageType"
+    MESSAGE_WIRE_TYPE: ClassVar[int | None] = None
```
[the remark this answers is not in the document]
matejcik: https://github.com/trezor/trezor-firmware/pull/3131 draft of solution

core/src/trezor/ui/layouts/tr/recovery.py
matejcik: right, kind of an entrenched assumption in the tests too. kinda surprising that we didn't notice this before.

n.b., what is this about? https://github.com/trezor/trezor-firmware/blob/matejcik/get-context/tests/input_flows.py#L1810C34-L1810C34
looks like the warning about wrong word is just up and missing

matejcik:
documentation added in https://github.com/trezor/trezor-firmware/pull/3110/commits/d6955874f1427c69a647add6d73eafdf2ecff1a3

core/src/trezor/ui/layouts/tr/recovery.py
matejcik: 81305badef3cbadaa7de0b696ce83953dae4089f

note that UI tests are failing on the above. I didn't want to just record over the test before we have an idea of what's going on there

core/src/trezor/ui/layouts/tr/__init__.py
matejcik: 81305badef3cbadaa7de0b696ce83953dae4089f

core/src/trezor/ui/layouts/common.py
matejcik: 81305badef3cbadaa7de0b696ce83953dae4089f

core/src/trezor/ui/layouts/common.py
matejcik: 81305badef3cbadaa7de0b696ce83953dae4089f

core/src/trezor/wire/context.py
```diff
@@ -0,0 +1,205 @@
...
+
+    Msg = TypeVar("Msg", bound=protobuf.MessageType)
+    HandlerTask = Coroutine[Any, Any, protobuf.MessageType]
+    Handler = Callable[["Context", Msg], HandlerTask]
+
+    LoadedMessageType = TypeVar("LoadedMessageType", bound=protobuf.MessageType)
+
+
+# If set to False protobuf messages marked with "experimental_message" option are rejected.
+EXPERIMENTAL_ENABLED = False
```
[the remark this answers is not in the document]
matejcik: https://github.com/trezor/trezor-firmware/pull/3110/commits/2b191c9ce860922b0af747cfa2e91f43a4825d22

core/src/trezor/ui/layouts/tr/recovery.py
grdddj: Yes, I am missing the warning screen locally as well. It is connected with `TR` recovery, I would fix it in the appropriate PR - https://github.com/trezor/trezor-firmware/pull/3119 not to slow down this PR (feel free to revert these changes)

grdddj (approved):
Looks good, thanks for the `interact` changes. Let's do the `TR` recovery fixes in a different PR not to slow this one with unrelated stuff.

mmilata (approved):

PR #3116, opened 2023-06-29
T2B1 content centering

grdddj:
WIP


grdddj:
> What needs to be done here in order to be ready for review?

This is an experimentation change, that makes the screen fit one more text iine (by shrinking the header margin). I will force-push it to include the latest UI and we can then together analyze the UI diff

grdddj:
Correction - it seems to increase the line-count just for `confirm properties` - key/value pairs. I am not sure it looks good, we might maybe decrease the margin between **key** and _value_.

https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4770078123/artifacts/test_ui_report/diff/TR-device_tests-bitcoin-test_nonstandard_paths.py::test_signmessage[m-4-255-script_types0].html

image

Hannsek:
Maybe it's unrelated to this pr, but some screens look as the content on it is not centered: 
image
image

Decreasing the space between the key and value seem like a good idea, as some screens look a bit weird. 
image
image

Hannsek:
It also changed other screens, which is not wanted:
image
image

obrusvit:
Obsolete PR

PR #3120, opened 2023-06-30
refactor: add abstraction over flash memory layout

TychoVrahe:
As a preparation for using different MCU, embedded flash driver is modified so that the firmware can handle differently organized flash.

`flash_area_t` is defined to describe where something is located on the flash, possibly on multiple, possible discrete, sectors.

firmware should ideally access the flash only through this area interface.

this also affects storage, which is naturally working with flash.

in legacy, the new interface is only used for storage which has common implementation, the rest in legacy stays MCU-bound, which is probably fine

builds on #3108


core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -104,48 +110,220 @@ const void *flash_get_address(uint8_t sector, uint32_t offset, uint32_t size) {
...
+    const void *area_end = flash_get_address(sector + 1, 0, 0);
+    if ((uint32_t)addr + size > (uint32_t)area_end) {
+      return NULL;
+    }
+    return (const void *)addr;
+  }
+  return NULL;
+}
+
+secbool flash_area_erase(const flash_area_t *area,
```
mmilata, line 229: Can this function just call `flash_area_erase_bulk` with `count = 1`?

core/embed/bootloader/messages.c
```diff
@@ -680,29 +692,13 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
-      // 3,  // skip because of MPU protection
-      FLASH_SECTOR_FIRMWARE_START,
-      7,
-      8,
-      9,
-      10,
-      FLASH_SECTOR_FIRMWARE_END,
-      FLASH_SECTOR_UNUSED_START,
-      13,
-      14,
```
mmilata, line 695: The new code does not wipe these.

mmilata (commented):
Concept ACK. Since this touches pretty much everything another pair of eyes would be nice.

core/embed/trezorhal/flash.h
```diff
@@ -24,96 +24,50 @@
...
-// OTP blocks allocation
-#define FLASH_OTP_BLOCK_BATCH 0
-#define FLASH_OTP_BLOCK_BOOTLOADER_VERSION 1
-#define FLASH_OTP_BLOCK_VENDOR_HEADER_LOCK 2
-#define FLASH_OTP_BLOCK_RANDOMNESS 3
-#define FLASH_OTP_BLOCK_DEVICE_VARIANT 4
+const void *flash_get_address(uint16_t sector, uint32_t offset, uint32_t size);
+uint32_t flash_sector_size(uint16_t sector);
+uint16_t flash_total_sectors(const flash_area_t *area);
+int flash_get_sector_num(const flash_area_t *area, uint32_t sector_inner_num);
```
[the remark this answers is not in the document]
TychoVrahe: changed https://github.com/trezor/trezor-firmware/commit/55785b2157d3e49913a1c4140bca60d582d5f63a

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -104,48 +110,220 @@ const void *flash_get_address(uint8_t sector, uint32_t offset, uint32_t size) {
...
+      return sector;
+    } else {
+      remaining -= area->subarea[i].num_sectors;
+    }
+  }
+
+  return -1;
+}
+
+int32_t flash_get_sector(const flash_area_t *area, uint32_t offset) {
```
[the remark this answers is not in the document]
TychoVrahe: removed: https://github.com/trezor/trezor-firmware/commit/55785b2157d3e49913a1c4140bca60d582d5f63a

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -104,48 +110,220 @@ const void *flash_get_address(uint8_t sector, uint32_t offset, uint32_t size) {
...
+        }
+        continue;
+      }
+      // in correct sector
+      break;
+    }
+
+    const void *addr = flash_get_address(sector, tmp_offset, size);
+    const void *area_end = flash_get_address(sector + 1, 0, 0);
+    if ((uint32_t)addr + size > (uint32_t)area_end) {
```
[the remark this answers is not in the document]
TychoVrahe: Assumption is here that when getting flash address for writing/reading data, only one sector should be modified/accessed at time. On F4 (and U5) there are no gaps in address space so in this context, i think it is safe assumption. However the check indeed seems redundant, removed: https://github.com/trezor/trezor-firmware/commit/55785b2157d3e49913a1c4140bca60d582d5f63a

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -104,48 +110,220 @@ const void *flash_get_address(uint8_t sector, uint32_t offset, uint32_t size) {
...
+uint32_t flash_sector_size(uint16_t sector) {
   if (sector >= FLASH_SECTOR_COUNT) {
     return 0;
   }
   return FLASH_SECTOR_TABLE[sector + 1] - FLASH_SECTOR_TABLE[sector];
 }
 
-secbool flash_erase_sectors(const uint8_t *sectors, int len,
-                            void (*progress)(int pos, int len)) {
+uint32_t flash_subarea_get_size(const flash_subarea_t *subarea) {
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/55785b2157d3e49913a1c4140bca60d582d5f63a

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -104,48 +110,220 @@ const void *flash_get_address(uint8_t sector, uint32_t offset, uint32_t size) {
...
+    }
+    return (const void *)addr;
+  }
+  return NULL;
+}
+
+secbool flash_area_erase(const flash_area_t *area,
+                         void (*progress)(int pos, int len)) {
   ensure(flash_unlock_write(), NULL);
   FLASH_EraseInitTypeDef EraseInitStruct;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/55785b2157d3e49913a1c4140bca60d582d5f63a

core/embed/trezorhal/stm32f4/flash.c
TychoVrahe: yeah https://github.com/trezor/trezor-firmware/commit/55785b2157d3e49913a1c4140bca60d582d5f63a

core/embed/bootloader/messages.c
TychoVrahe: right, added a WIPE_AREA to precisely define what should be wiped https://github.com/trezor/trezor-firmware/commit/55785b2157d3e49913a1c4140bca60d582d5f63a

mmilata (approved):

mmilata:
Can you please take a brief look at this @matejcik or @andrewkozlik?

mmilata:
rebased on top of master since #3108 is already in

core/embed/models/model_T2T1_layout.c
```diff
@@ -0,0 +1,81 @@
...
+    .subarea[0] =
+        {
+            .first_sector = 4,
+            .num_sectors = 1,
+        },
+    .subarea[1] =
+        {
+            .first_sector = 6,
+            .num_sectors = 9,
+        },
```
andrewkozlik, line 66: Why is sector 15 skipped from the wipe area?

core/embed/models/model_T1B1_layout.c
andrewkozlik, line 1: Shouldn't this file define all the variables that are advertised as `extern` in core/embed/models/layout_common.h? Those being `BOARDLOADER_AREA`, `WIPE_AREA` and `ALL_WIPE_AREA`.

core/embed/models/model_T1B1_layout.c
```diff
@@ -0,0 +1,39 @@
...
+            .num_sectors = 1,
+        },
+};
+
+const flash_area_t FIRMWARE_AREA = {
+    .num_subareas = 1,
+    .subarea[0] =
+        {
+            .first_sector = 6,
+            .num_sectors = 6,
```
andrewkozlik, line 37: `.num_sectors = 7`?

core/embed/trezorhal/flash.h
```diff
@@ -24,71 +24,41 @@
...
+ * it may be composed of multiple non-contiguous subareas.
+ *
+ * flash_subarea_t represents a contiguous area in flash memory, specified by
+ * first_sector and num_sectors.
+ */
+
+typedef struct {
+  uint16_t first_sector;
+  uint16_t num_sectors;
+} flash_subarea_t;
```
andrewkozlik, line 56: I can see that `flash_subarea_t` and `flash_area_t` are defined in three places (here, `legacy/flash.h` and in `storage/tests/c/flash.h`). Also the functions that are further down below seem to be declared three times. Is there a good reason for that? I'd expect there to be something like `storage/flash_common.h` which would declare these.

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -104,48 +110,165 @@ const void *flash_get_address(uint8_t sector, uint32_t offset, uint32_t size) {
...
+
+      if (s == subarea->num_sectors - 1) {
+        // last sector in subarea
+        return secfalse;
+      }
+      continue;
+    }
+    *sector_out = sector;
+    *offset_out = tmp_offset;
+    return sectrue;
```
andrewkozlik, line 182: This code seems more complicated than it needs to be. I think we can make it easier to read, or am I overlooking something?
```suggestion
    if (tmp_offset < sector_size) {
      *sector_out = sector;
      *offset_out = tmp_offset;
      return sectrue;
    }
    tmp_offset -= sector_size;
    sector++;
```

andrewkozlik (commented):
core reviewed, I haven't looked at legacy yet.

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -182,6 +305,37 @@ secbool flash_write_word(uint8_t sector, uint32_t offset, uint32_t data) {
...
+    return secfalse;
+  }
+  return flash_write_byte(sector, sector_offset, data);
+}
+
+secbool flash_area_write_word(const flash_area_t *area, uint32_t offset,
+                              uint32_t data) {
+  uint16_t sector;
+  uint32_t sector_offset;
+  if (get_sector_and_offset(area, offset, &sector, &sector_offset) != sectrue) {
```
andrewkozlik, line 322: I don't want to get into premature optimization, but I feel I should mention that when writing a whole sector, e.g. during firmware installation, `get_sector_and_offset()` will get called 32768 times, even though it needs to be called only once. I recommend to measure the firmware installation time before and after this PR. If it introduces a noticeable slowdown, then I recommend defining a function which writes a sequence of words, like
```c
secbool flash_area_write_words(const flash_area_t *area, uint32_t offset,
                               uint32_t *words, size_t word_count);
```
We can debate whether it should be able to traverse sectors. FTR, if we get into optimization, then we can also get rid of multiple calls to `flash_get_address()` and alignment checks.

core/embed/trezorhal/unix/flash.c
```diff
@@ -28,12 +28,81 @@
...
 #include "profile.h"
 
 #ifndef FLASH_FILE
 #define FLASH_FILE profile_flash_path()
 #endif
 
+#if defined STM32F427xx || defined STM32F429xx
+#define FLASH_SECTOR_COUNT 24
+#elif defined STM32F405x
+#define FLASH_SECTOR_COUNT 12
```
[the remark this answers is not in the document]
mmilata: I think this is meant for model 1 which uses STM32F405x headers even though it's actually F2. Probably best to change the ifdef to use TREZOR_MODEL.

core/embed/trezorhal/flash.h
```diff
@@ -24,71 +24,41 @@
...
-    defined TREZOR_MODEL_DISC1
-#define FLASH_SECTOR_STORAGE_1 4
-#define FLASH_SECTOR_STORAGE_2 16
-#elif defined TREZOR_MODEL_1
-#define FLASH_SECTOR_STORAGE_1 2
-#define FLASH_SECTOR_STORAGE_2 3
-#else
-#error Unknown Trezor model
+#ifndef STM32U5A9xx
+#define FLASH_BYTE_ACCESS 1
```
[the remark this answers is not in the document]
andrewkozlik: FTR, when writing this comment, I didn't realized that `FLASH_BYTE_ACCESS` is boolean defined/undefined, I thought it represented the number of bytes that have to be written in one access. Nevertheless, I think the comment remains more or less valid. When adding a new MCU it seems better to avoid defaults.

storage/tests/c/flash.c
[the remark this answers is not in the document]
andrewkozlik: `legacy/flash.c` also seems to duplicate this code. Again, I haven't checked carefully, so maybe there is a good reason that I am not seeing right now.

core/embed/models/model_T2T1_layout.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/def5983ae4355ad4cbc4bb7dd0d4065baa4b356e

core/embed/models/model_T1B1_layout.c
TychoVrahe: hmm i adjusted what was in flash.h, seemed that it wasn't set correctly there for T1 at all: https://github.com/trezor/trezor-firmware/commit/df115b63de11c2f18e4c4e9219ebfd9feebaf028

core/embed/trezorhal/unix/flash.c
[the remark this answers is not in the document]
TychoVrahe: hm do you see benefit in keeping the unix implementation of flash driver mimicking STM32F4/F2? this will get even more messier when we add U5 models as the flash is quite different and with the abstraction implemented we might not care anymore about keeping the sectors 1:1 with the MCU used. thinking about removing the ifdefs all together and just keeping single implementation/layout for unix.

core/embed/trezorhal/flash.h
TychoVrahe: well this was the case all along, this PR adds the new functions/definitions in same manner. i agree that there is a lot of repetition, but not sure what would be the best place for the common file - maybe trezor-common? (don't really like putting the common file into storage as the definition of flash interface is not really what would you expect from that)

core/embed/trezorhal/flash.h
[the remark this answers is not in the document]
TychoVrahe: instead of adding the ifdef here, i think it might be better to move this inside the concrete trezorhal - e.g. `trezorhal/stm32f4/platform.h` would have this defined, whereas `trezorhal/stm32u5/platform.h` would not. (or we can add there a new header `trezorhal/stm32f4/flash_specifics.h` or something). this wont change when changing inside MCU line i.e. to STM32U599xx, and when adding some other family like STM32L5 you need to check everything anyway

core/embed/reflash/main.c
```diff
@@ -90,12 +87,8 @@ int main(void) {
...
 #define BOOTLOADER_TOTAL_SIZE (128 * 1024)
 
-  flash_from_sdcard(FLASH_SECTOR_BOARDLOADER_START, 0 * BOARDLOADER_CHUNK_SIZE,
-                    BOARDLOADER_CHUNK_SIZE);
-  flash_from_sdcard(1, 1 * BOARDLOADER_CHUNK_SIZE, BOARDLOADER_CHUNK_SIZE);
-  flash_from_sdcard(FLASH_SECTOR_BOARDLOADER_END, 2 * BOARDLOADER_CHUNK_SIZE,
-                    BOARDLOADER_CHUNK_SIZE);
-  flash_from_sdcard(FLASH_SECTOR_BOOTLOADER, BOARDLOADER_TOTAL_SIZE,
+  flash_from_sdcard(&BOARDLOADER_AREA, 0, BOARDLOADER_TOTAL_SIZE);
+  flash_from_sdcard(&BOARDLOADER_AREA, BOARDLOADER_TOTAL_SIZE,
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/ae3d33dcc61985ce17c50fa3e06240bc27b83978

core/embed/bootloader_ci/messages.c
```diff
@@ -616,9 +617,11 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
   const uint32_t *const src = (const uint32_t *const)chunk_buffer;
   for (int i = 0; i < chunk_size / sizeof(uint32_t); i++) {
-    ensure(flash_write_word(FIRMWARE_SECTORS[firmware_block],
-                            i * sizeof(uint32_t), src[i]),
-           NULL);
+    ensure(
+        flash_area_write_word(
+            &FIRMWARE_AREA,
+            firmware_block * IMAGE_CHUNK_SIZE + i * sizeof(uint32_t), src[i]),
+        NULL);
```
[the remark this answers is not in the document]
TychoVrahe: this PR does not really intend to add support for flash without byte access, e.g. storage does not handle this also. if this is a problem i will rather remove the `FLASH_BYTE_ACCESS` from this PR entirely and reintroduce it when its all ready.

core/embed/bootloader/messages.c
```diff
@@ -651,11 +650,24 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
+  }
+#else
+  for (int i = 0; i < chunk_size / (sizeof(uint32_t) * 4); i++) {
+    ensure(flash_area_write_quadword(
+               &FIRMWARE_AREA,
+               firmware_block * IMAGE_CHUNK_SIZE + i * 4 * sizeof(uint32_t),
+               &src[4 * i]),
            NULL);
   }
+#endif
```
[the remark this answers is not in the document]
TychoVrahe: agree https://github.com/trezor/trezor-firmware/commit/16d8bd5b524becb72091507f8f3b5b6ce3aaaa00

core/embed/lib/image.c
```diff
@@ -215,39 +217,41 @@ secbool check_single_hash(const uint8_t *const hash, const uint8_t *const data,
...
-                             const uint8_t *sectors, int blocks) {
-  if (0 == sectors || blocks < 1) {
+                             const flash_area_t *area) {
+  if (0 == area) {
     return secfalse;
   }
-  const void *data =
-      flash_get_address(sectors[0], firstskip, IMAGE_CHUNK_SIZE - firstskip);
+
+  uint32_t size = IMAGE_CHUNK_SIZE;
```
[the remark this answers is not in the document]
TychoVrahe: reverted to use just `IMAGE_CHUNK_SIZE` https://github.com/trezor/trezor-firmware/commit/9c7c6eb1726f99e881b9b41b22d341f96d54c475

core/embed/trezorhal/stm32f4/flash.c
TychoVrahe: yeah this is better https://github.com/trezor/trezor-firmware/commit/e9e84cc14873a636045edb6d4c2a6623f333d355

storage/tests/c/flash.c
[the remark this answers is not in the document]
TychoVrahe: similarly to the header files, i don't think that tying flash driver implementation to storage is the correct solution. maybe we could move some of the functions to trezor-common, but with the implementation it might a bit more complicated, for example the U5 have uniform sector size 8kB, which allows some optimizations when looking up the sectors for area write, and byte/word writes should not be implemented at all.

core/embed/models/model_T1B1_layout.c
TychoVrahe: boardloader does not exist for T1, and `WIPE_AREA` and `ALL_WIPE_AREA` are used by bootloader (and boardloader), which is not implemented in `core` for T1. So i consider it a bit useless to have it here

core/embed/models/model_T2T1_layout.c
andrewkozlik: So there is no reason to split it into two subareas, is there?

core/embed/models/model_T1B1_layout.c
andrewkozlik: I see, I hadn't realized that. :+1:

core/embed/trezorhal/stm32f4/flash.c
andrewkozlik: That's about a 5 % increase in time. I agree it's negligible.

core/embed/bootloader_ci/messages.c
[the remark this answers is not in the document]
andrewkozlik: No, that's not my point. This part of the code should be essentially the same as in `core/embed/bootloader/messages.c`. So based on our discussion in https://github.com/trezor/trezor-firmware/pull/3120#discussion_r1266417000, I think we should write by quadwords here.

core/embed/trezorhal/unix/flash.c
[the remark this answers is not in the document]
andrewkozlik: Firstly, just so we are on the same page, my comment was basically saying that `FLASH_SECTOR_COUNT` for `STM32F405x` is incompatible with the definitions of `STORAGE_AREAS`, `FIRMWARE_AREA` etc. in this file. So these need to be defined differently for `STM32F405x`. I would try to avoid redefining them and instead reuse the definitions from `core/embed/models/model_T1T1_layout.c` and `core/embed/models/model_T2T1_layout.c` so that the layout in the hardware build is always the same as that in the unix build and we don't have to update it in two places.

As for the higher-level question of whether we should keep the sectors 1:1 with the MCU used, I don't know. In principle we should do our best to keep the emulator as close to the hardware implementation as possible, so that we detect bugs as early as possible. On the other hand I am not sure whether the sector layout is significant in that regard. I can think of two things that are relevant:
- The differences in storage size for different models. I believe we should keep the storage sector sizes 1:1 with hardware, so that our device tests detect if some new feature starts to clutter the storage too much.
- The firmware hashing functionality that reads the firmware sectors. That would have to be somehow hacked in the emulator, so it wouldn't get tested very well by the device tests.

Finally the question of keeping the unix implementation of the flash driver that mimicks STM32F4/F2. Are we ever going to run `core` firmware on T1? I guess it's nice to keep, since we already invested some resources into this effort, but not sure if it has any future that would be worth the maintenance. Nevertheless, the problems that we are discovering here are going to come up sooner or later with newer MCU models, so we might as well try and solve them early, while the code is fresh in our minds.

storage/tests/c/flash.c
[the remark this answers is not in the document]
andrewkozlik: storage/tests/c/flash.c

core/embed/models/model_T2T1_layout.c
TychoVrahe: actually, the reason for skipping sector 15 was MPU settings of bootloader, which makes the sector 15 not accessible. reverted the change and added a comment about this here: https://github.com/trezor/trezor-firmware/commit/e626bc52baf9179ee9bdc3be5716825c0ed8ee90

why is the MPU set this way, i dont know.

core/embed/trezorhal/flash.h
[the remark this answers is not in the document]
TychoVrahe: moved  it here: https://github.com/trezor/trezor-firmware/commit/d0032acb4cafb4d044d3c920ecc832614e909c78

core/embed/bootloader_ci/messages.c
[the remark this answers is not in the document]
TychoVrahe: done here https://github.com/trezor/trezor-firmware/commit/a3138befaf2d7b609f65345f4fe4c602bea82afb

TychoVrahe:
fixed bootloader emulator build: https://github.com/trezor/trezor-firmware/commit/5b51f95f8ff9cfcb25752daa5cf3a2022d2a1606

storage/tests/c/flash.c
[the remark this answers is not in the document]
andrewkozlik: We have these four files (and their header files) which contain essentially the same new code, or at least at first glance, I don't see a real difference. Actually, they are already diverging, but the only reason for that seems to be that some of them were further refactored while others were not. Each adds about ~200 lines of code. I only reviewed the first one, because it just seems inefficient to do 4 times and to maintain it in 4 different places.
- `core/embed/trezorhal/stm32f4/flash.c`
- `core/embed/trezorhal/unix/flash.c`
- `legacy/flash.c`
- `storage/tests/c/flash.c`

From https://github.com/trezor/trezor-firmware/pull/3120#discussion_r1267297083:
> well this was the case all along, this PR adds the new functions/definitions in same manner

Yes and no. There were differences between how `legacy` code, `core` code and emulator code accessed the flash memory, e.g. `svc_flash_program()` vs. `HAL_FLASH_Program()` vs. direct access. These will still need to be implemented individually. The amount of common code on the other hand is now significantly increased. Also, there was a time when the flash sector layouts were different for `legacy`, `core` and `storage`. Now it's becoming all shared.

> it might a bit more complicated, for example the U5 have uniform sector size 8kB, which allows some optimizations when looking up the sectors for area write, and byte/word writes should not be implemented at all.

I think that's a problem which we will have to deal with anyway in the core implementation and I don't think it makes any difference where the code is placed. In fact I think factoring it out will make things easier in the long run. When we add U5 support, the storage will behave a bit differently for that MCU and we will probably want to run a separate `storage` test to ensure that new firmware releases don't break the storage format, i.e. they comply with the reference implementation. This will be easier and especially more reliable to do if the implementation is factored out and doesn't have to be copied/edited in two places.

I am unsure about moving the code to `trezor-common`. This has never contained actual firmware code, it contains protobuf definitions, FIDO definitions, maybe some script to process the definitions etc. We also export it to it's own repository, because the protobuf definitions are used by connect I guess. The point is it seems more higher level. 

Another option is creating a `flash` directory in the root of the monorepo and I am totally OK with that, but it just makes more sense to me to add it to the `storage` directory (we could also rename `storage` -> `flash`).

core/embed/bootloader_ci/messages.c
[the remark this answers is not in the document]
andrewkozlik: Oh good, I hadn't even noticed that. But that's a change in `core/embed/reflash/main.c`. This thread is about `core/embed/bootloader_ci/messages.c`.

storage/tests/c/flash.c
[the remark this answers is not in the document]
prusnak: `trezor-common` contains stuff that is supposed to be consumed by second parties (i.e. coin definitions on trezor.io and suite, protob by connect and suite, etc.) and third parties (protob, etc). It doesn't sound like a good place for adding low-level code.

core/embed/models/model_T2T1_layout.c
andrewkozlik: Alright, now I have a different question. Shouldn't sector 15 be also skipped in `ALL_WIPE_AREA` here in this file a few lines below and also in `core/embed/trezorhal/unix/flash.c`?

(On a side note, I dislike the fact that the layout is defined twice in both these files. But this topic is already under discussion in another thread.)

andrewkozlik: Now I realized it doesn't need to be skipped in `ALL_WIPE_AREA`, because `ALL_WIPE_AREA` is used by the boardloader only.

storage/tests/c/flash.c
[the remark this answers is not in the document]
TychoVrahe: moved the implementation/interface to storage/flash_common. https://github.com/trezor/trezor-firmware/commit/8b47b76a7737e6ad39e12955455d8df976a72572

note that the implementation won't be (at least partly) usable for U5, so i expect splitting it into something like flash_common_f4 + flash_common_u5 in near future.

core/embed/trezorhal/unix/flash.c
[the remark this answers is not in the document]
TychoVrahe: reusing the layout files: https://github.com/trezor/trezor-firmware/commit/557f0b871716334520a19b8a64898794663ddd64

this most likely wont be possible to keep this simple when adding U5 support. if we want to keep this close to real hw implementation there will be some larger ifdefs.

i don't think that T1 in core is relevant anymore, i would be ok with dropping it. it doesn't work anyway for several reasons. but that is for @matejcik to decide and out of scope here.

core/embed/trezorhal/flash.h
```diff
@@ -22,101 +22,47 @@
...
+secbool flash_write_byte(uint16_t sector, uint32_t offset, uint8_t data);
 
-#define FLASH_STATUS_ALL_FLAGS                                            \
-  (FLASH_SR_RDERR | FLASH_SR_PGSERR | FLASH_SR_PGPERR | FLASH_SR_PGAERR | \
-   FLASH_SR_WRPERR | FLASH_SR_SOP | FLASH_SR_EOP)
+secbool flash_write_word(uint16_t sector, uint32_t offset, uint32_t data);
 
-void flash_init(void);
+secbool flash_area_write_byte(const flash_area_t *area, uint32_t offset,
+                              uint8_t data);
```
andrewkozlik, line 59: Please remove `flash_area_write_byte()`. Already declared in `flash_common.h`.

legacy/flash.h
```diff
@@ -45,4 +46,27 @@ secbool __wur flash_erase(uint8_t sector);
...
+} flash_subarea_t;
+
+typedef struct {
+  flash_subarea_t subarea[4];
+  uint8_t num_subareas;
+} flash_area_t;
+
+const void *flash_area_get_address(const flash_area_t *area, uint32_t offset,
+                                   uint32_t size);
+uint32_t flash_area_get_size(const flash_area_t *area);
```
andrewkozlik, line 61: Please remove `flash_area_get_size()`. Already declared in `storage/flash_common.h`.

legacy/flash.h
```diff
@@ -45,4 +46,27 @@ secbool __wur flash_erase(uint8_t sector);
 secbool __wur flash_write_byte(uint8_t sector, uint32_t offset, uint8_t data);
```
andrewkozlik, line 46: Please remove `flash_erase()`. Dead code.

legacy/flash.h
```diff
@@ -45,4 +46,27 @@ secbool __wur flash_erase(uint8_t sector);
...
+uint32_t flash_area_get_size(const flash_area_t *area);
+
+secbool __wur flash_area_erase(const flash_area_t *area,
+                               void (*progress)(int pos, int len));
+secbool __wur flash_area_erase_bulk(const flash_area_t *area, int count,
+                                    void (*progress)(int pos, int len));
+secbool __wur flash_area_write_byte(const flash_area_t *area, uint32_t offset,
+                                    uint8_t data);
+secbool __wur flash_area_write_word(const flash_area_t *area, uint32_t offset,
+                                    uint32_t data);
```
andrewkozlik, line 70: Please remove `flash_area_write_byte()` and `flash_area_write_word()`. Both already declared in `flash_common.h`.

storage/tests/c/flash.h
```diff
@@ -24,19 +24,21 @@
...
-secbool __wur flash_write_word(uint8_t sector, uint32_t offset, uint32_t data);
+secbool flash_area_erase_bulk(const flash_area_t *area, int count,
+                              void (*progress)(int pos, int len));
+
+secbool flash_write_byte(uint16_t sector, uint32_t offset, uint8_t data);
+
+secbool flash_write_word(uint16_t sector, uint32_t offset, uint32_t data);
+
+secbool flash_area_write_byte(const flash_area_t *area, uint32_t offset,
+                              uint8_t data);
```
andrewkozlik, line 42: Please remove `flash_area_write_byte()`. Already declared in `flash_common.h`.

core/embed/trezorhal/unix/flash.c
```diff
@@ -182,7 +170,45 @@ secbool flash_erase_sectors(const uint8_t *sectors, int len,
   return sectrue;
 }
```
andrewkozlik, line 172: Please remove `flash_erase_sectors()`. Dead code.

storage/tests/c/flash.c
```diff
@@ -90,7 +100,45 @@ secbool flash_erase_sectors(const uint8_t *sectors, int len,
   return sectrue;
 }
```
andrewkozlik, line 102: Please remove `flash_erase_sectors()`. Dead code.

legacy/flash.c
andrewkozlik, line 1: Please remove `flash_erase()`. Dead code.

andrewkozlik (commented):
Looks great. Some dead code to remove and we are done.

core/embed/trezorhal/flash.h
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

legacy/flash.h
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

legacy/flash.h
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

legacy/flash.h
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

included the flash_common.h and removed most of the declarations here.

storage/tests/c/flash.h
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

core/embed/trezorhal/unix/flash.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

storage/tests/c/flash.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

legacy/flash.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/626e6d0954357a79c03f7e8f80a947f9414f78f7

andrewkozlik (approved):
LGTM, utACK.

Issue #2919, opened 2023-03-29
(Almost) interaction-less firmware update


Hannsek:
https://github.com/trezor/trezor-firmware/pull/2841#issuecomment-1485345315: 
> while we're thinking about this, we could allow interaction-less upgrade, i.e., skip the "really install firmware by SatoshiLabs version 2.99" dialog, if (a) the vendor is unchanged, and (b) it is a strict **up**grade, and (c) click flag is not set in the vendor header i.e.:
> 
> * reinstall of same version still gets a confirmation
> * downgrade gets confirmation
> * vendor change gets confirmation, obviously
> * any operation with devel firmware gets confirmation
> 
> then you would, in firmware, click Yes on "proceed to installing firmware?", Trezor screen goes blue, waits for host, straight to progress bar -> reboot -> you're done

_Questions_: 
- Can we display info about to-be-installed firmware before the installation? 

The whole flow could look like this:
<img width="362" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/b57a1045-ca8a-45b8-b822-534ab0285c0a">

<img width="606" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/577b385f-7bf6-4ab2-923f-01a319beb10e">

matejcik:
can't really: the host could be lying and send one thing first and another thing after going to bootloader.
we could do that if we allow jump straight to bootloader, but that's something I'm not ready to allow just yet.

Hannsek:
@prusnak @andrewkozlik Do you have any thoughts about this?

prusnak:
I find the proposal in the original comment good. Just one note, there is no devel firmware. What you call a devel firmware is just a firmware with a different vendor (which has the click flag set at the moment).

matejcik:
interaction-less update RFC:

## 1. Firmware stage

1. host sends `FirmwareUpdate` that contains _exactly the header_, i.e., concatenated vendor header + firmware header
2. firmware verifies that:
   * the headers are valid (signatures match etc)
   * vendor matches the currently installed one
   * version is strictly higher than the current one
  
   if this fails, Failure is returned. (if the operation is actually what the user wants, they will need to go to bootloader to do it)
3. if it is OK, Trezor displays a prompt:
   "Upgrade firmware to version 2.7.9?"
   (I don't think we need to mention the vendor here, given that it is forced to be the same as current -- but we can)
4. when the user confirms, firmware sends Success and jumps to bootloader
[...]

PR #3122, opened 2023-06-30
fix erroneous shutdown when rejecting firmware upload

TychoVrahe:
this was only triggered when bootloader does not detect valid firmware, therefore shows "get started with your trezor" screen, but at least firmware header is valid so bootloader asks for confirmation.

### DIFF
--- a/core/embed/bootloader/main.c
+++ b/core/embed/bootloader/main.c
@@ -81,9 +81,9 @@ static const uint8_t * const BOOTLOADER_KEYS[] = {
 #define USB_IFACE_NUM 0
 
 typedef enum {
-  CONTINUE = 0,
-  RETURN = 1,
-  SHUTDOWN = 2,
+  SHUTDOWN = 0,
+  CONTINUE_TO_FIRMWARE = 0xAABBCCDD,
+  RETURN_TO_MENU = 0x55667788,
 } usb_result_t;
 
 static void usb_init_all(secbool usb21_landing) {
@@ -164,7 +164,7 @@ static usb_result_t bootloader_usb_loop(const vendor_header *const vhdr,
           hal_delay(100);
           usb_stop();
           usb_deinit();
-          return RETURN;
+          return RETURN_TO_MENU;
         }
         ui_screen_wipe();
         r = process_msg_WipeDevice(USB_IFACE_NUM, msg_size, buf);
@@ -200,7 +200,7 @@ static usb_result_t bootloader_usb_loop(const vendor_header *const vhdr,
           hal_delay(100);
           usb_stop();
           usb_deinit();
-          return RETURN;
+          return RETURN_TO_MENU;
         } else if (r == 0) {  // last chunk received
           ui_screen_install_progress_upload(1000);
           ui_screen_done(4, sectrue);
@@ -213,7 +213,7 @@ static usb_result_t bootloader_usb_loop(const vendor_header *const vhdr,
           usb_stop();
           usb_deinit();
           ui_screen_boot_empty(true);
-          return CONTINUE;
+          return CONTINUE_TO_FIRMWARE;
         }
         break;
       case MessageType_MessageType_GetFeatures:
@@ -227,7 +227,7 @@ static usb_result_t bootloader_usb_loop(const vendor_header *const vhdr,
           hal_delay(100);
           usb_stop();
           usb_deinit();
-          return RETURN;
+          return RETURN_TO_MENU;
         }
         process_msg_UnlockBootloader(USB_IFACE_NUM, msg_size, buf);
         screen_unlock_bootloader_success();
@@ -315,39 +315,51 @@ int bootloader_main(void) {
 
   mpu_config_bootloader();
 
+#ifdef TREZOR_EMULATOR
+  // wait a bit so that the empty lock icon is visible
+  // (on a real device, we are waiting for touch init which takes longer)
+  hal_delay(400);
+#endif
+
   const image_header *hdr = NULL;
   vendor_header vhdr;
+
   // detect whether the device contains a valid firmware
-  secbool firmware_present = sectrue;
+  volatile secbool vhdr_present = secfalse;
+  volatile secbool vhdr_keys_ok = secfalse;
+  volatile secbool vhdr_lock_ok = secfalse;
+  volatile secbool img_hdr_ok = secfalse;
+  volatile secbool model_ok = secfalse;
+  volatile secbool header_present = secfalse;
+  volatile secbool firmware_present = secfalse;
 
-  if (sectrue != read_vendor_header((const uint8_t *)FIRMWARE_START, &vhdr)) {
-    firmware_present = secfalse;
+  vhdr_present = read_vendor_header((const uint8_t *)FIRMWARE_START, &vhdr);
+
+  if (sectrue == vhdr_present) {
+    vhdr_keys_ok = check_vendor_header_keys(&vhdr);
   }
 
-  if (sectrue == firmware_present) {
-    firmware_present = check_vendor_header_keys(&vhdr);
+  if (sectrue == vhdr_keys_ok) {
+    vhdr_lock_ok = check_vendor_header_lock(&vhdr);
   }
 
-  if (sectrue == firmware_present) {
-    firmware_present = check_vendor_header_lock(&vhdr);
-  }
-
-  if (sectrue == firmware_present) {
+  if (sectrue == vhdr_lock_ok) {
     hdr = read_image_header(
         (const uint8_t *)(size_t)(FIRMWARE_START + vhdr.hdrlen),
         FIRMWARE_IMAGE_MAGIC, FIRMWARE_IMAGE_MAXSIZE);
-    if (hdr != (const image_header *)(size_t)(FIRMWARE_START + vhdr.hdrlen)) {
-      firmware_present = secfalse;
+    if (hdr == (const image_header *)(size_t)(FIRMWARE_START + vhdr.hdrlen)) {
+      img_hdr_ok = sectrue;
     }
   }
-  if (sectrue == firmware_present) {
-    firmware_present = check_image_model(hdr);
+  if (sectrue == img_hdr_ok) {
+    model_ok = check_image_model(hdr);
   }
-  if (sectrue == firmware_present) {
-    firmware_present =
+  if (sectrue == model_ok) {
+    header_present =
         check_image_header_sig(hdr, vhdr.vsig_m, vhdr.vsig_n, vhdr.vpub);
   }
-  if (sectrue == firmware_present) {
+
+  if (sectrue == header_present) {
     firmware_present = check_image_contents(
         hdr, IMAGE_HEADER_SIZE + vhdr.hdrlen, &FIRMWARE_AREA);
   }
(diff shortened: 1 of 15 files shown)


matejcik:
this doesn't look right

first off you're repeatedly erasing flash in the while-true loop, that's not great

second, we probably shouldn't infinite-loop in the first place:
* CONTINUE boots to firmware
* SHUTDOWN early-returns 
* RETURN might want to go into the normal boot menu instead of returning to "Welcome to Trezor"

matejcik:
(while we're here, maybe rename `CONTINUE` -> `CONTINUE_TO_FIRMWARE` and `RETURN` -> `RETURN_TO_MENU` ?)

TychoVrahe:
As suggested: https://github.com/trezor/trezor-firmware/commit/dfe5dedc036d123ea08df27ed4e8a6cfc8de2cb7

Note that `RETURN_TO_MENU` does not always return to menu, but sometimes to the `Intro` screen (the one you can select to install firmware or enter menu). 

Also. When user cancels the upload now, he is taken to Intro screen, where he can go to menu and 'restart' from there: which will then show fatal error with "firmware corrupted" message. This is also case when there is firmware header but corrupted fw and user chooses in the `welcome` screen to issue Wipe command through trezorctl and cancels. 

Even worse probably is that if there is no firmware header, user does the wipe command, cancels, there is crash with `BL.rs` message, as the Intro screen expects some data from the header, which are not there. This will definitely need some fixing, question is what is the correct behavior in this situation?

matejcik:
> Even worse probably is that if there is no firmware header, user does the wipe command, cancels, there is crash with `BL.rs` message, as the Intro screen expects some data from the header, which are not there. This will definitely need some fixing, question is what is the correct behavior in this situation?

maybe we should fold the "go to trezor.io/start" screen into the main thing? like, if there is a fw header, we show the blue intro, and if there isn't, we connect to host directly and show the black welcome?

TychoVrahe:
trying that in https://github.com/trezor/trezor-firmware/commit/22ca85e09a96445f0ffe816c914d161f13ca2f1b 

Seems to be working fine, one thing is that when this (corrupted fw with valid header) happens, we now go directly to bootloader menu and user has no clue why. Maybe we should in the "blue intro" screen show some warning that fw is corrupted?

Another thing, "header present" is evaluated with same logic as when deciding whether to ask user for confirmation, but at least theoretically more things can happen, like having valid header for different model will be evaluated as no-header in this context. This is probably fine, just something to be aware of.

core/embed/bootloader/main.c
```diff
@@ -388,47 +392,55 @@ int bootloader_main(void) {
...
+             NULL);
 
-    // erase storage
-    ensure(flash_erase_sectors(STORAGE_SECTORS, STORAGE_SECTORS_COUNT, NULL),
-           NULL);
-
-    // and start the usb loop
-    if (bootloader_usb_loop(NULL, NULL) != CONTINUE) {
-      return 1;
+      ui_set_initial_setup(true);
```
matejcik, line 415: please move these lines above the emulator ifdef, so that it's easier to read all that happens in the else branch

matejcik: in fact, there is probably an overall better place for this ifdef block. something like at the else-branch of the "wait for touch" sequence? or maybe even just at the top level right after showing the boot screen the first time. emulator users can stand to wait 400ms in every case, and it will make it possible for people to trigger go-to-bootloader for TR which does not have a wait-loop.

(speaking of, could we again move `screen_boot_empty` at the very start, before reading the headers?)

matejcik:
it turns out that this PR incidentally implements a variant of #2794, right? `header_present` and `firmware_present` are considered separate states.

Tagging @andrewkozlik and @prusnak for a careful review of the logic here. It _looks_ perfectly sensible to me: determine header validity separate from firmware validity, and only erase seed if the header is missing.
But we do need to be careful here never to jump into firmware unless all the checks have passed.

wrt merging, let's wait with this until #2919 is ready and rebase on top of that. When that is ready, we'll need to carefully review the resulting _code_, hopefully with notes about what to look out for from here.

core/embed/bootloader/main.c
TychoVrahe: moved in https://github.com/trezor/trezor-firmware/commit/23c3b2b8752d3772d2421b3ffe3e83478f7b81a8

core/embed/bootloader/main.c
```diff
@@ -388,47 +392,55 @@ int bootloader_main(void) {
   }
 #endif
 
-  // start the bootloader if no or broken firmware found ...
-  if (firmware_present != sectrue) {
+  // ... or if user touched the screen on start
+  // ... or we have stay_in_bootloader flag to force it
+  // ... or there is no valid firmware
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/9ed2af56f15ed3d5fd788f4a96ea27443fb7f904

TychoVrahe:
rebased on current master.

Hannsek:
What has to be done here in order to merge it?

TychoVrahe:
as said in the comment above, this is waiting for #2919 and it needs a review after rebasing.

Hannsek:
Ahh, sorry, didn't notice the comment.

core/embed/bootloader/main.c
```diff
@@ -347,6 +354,9 @@ int bootloader_main(void) {
     firmware_present =
         check_image_header_sig(hdr, vhdr.vsig_m, vhdr.vsig_n, vhdr.vpub);
   }
+
+  header_present = firmware_present;
```
matejcik, line 358: this is wrong; we need to keep `firmware_present` set to secfalse up until the next if-branch, otherwise in case of a glitch it could remain `sectrue` after the next check

in fact, isn't something like that true for all the previous checks too?

core/embed/bootloader/main.c
```diff
@@ -81,8 +81,8 @@ static const uint8_t * const BOOTLOADER_KEYS[] = {
 #define USB_IFACE_NUM 0
 
 typedef enum {
-  CONTINUE = 0,
-  RETURN = 1,
+  CONTINUE_TO_FIRMWARE = 0,
+  RETURN_TO_MENU = 1,
```
matejcik, line 85: let's use less glitchable values here

TychoVrahe:
Part of the hardening done here: https://github.com/trezor/trezor-firmware/pull/3122/commits/c5e7395c45ddd9c97bbeb39e5011265950504664

TychoVrahe:
The corrupted warnings are like this atm. In T2B1 the reboot choice is removed from the menu entirely.

Hannsek:
Ahh, this. I saw it yesterday. It seems to be ok as this should be visible only is case of something wrong happens.

TychoVrahe (commented):
Looking at the glitching protection part @hiviah , just two minor things, otherwise it looks reasonable to me

core/embed/bootloader/main.c
```diff
@@ -294,6 +299,79 @@ static void check_bootloader_version(void) {
 
 #endif
 
+void failed_jump_to_firmware(const image_header *hdr, vendor_header *vhdr_ptr) {
+  error_shutdown("INTERNAL ERROR", "(glitch)");
+}
+
+void real_jump_to_firmware(const image_header *hdr, vendor_header *vhdr_ptr) {
+  // we do not need deep copy, C doesn't have references until C20
+  vendor_header vhdr = *vhdr_ptr;
```
matejcik, line 308: what is the purpose of copying vhdr here? and what is the comment telling me? :)

matejcik: this actually causes a segfault when installing firmware in the emulator.... working on `vhdr_ptr` fixes it

matejcik:
added a thing in f286ca8afef914b285f85517a653a7df08a0a8f1, @TychoVrahe @hiviah please review

core/embed/bootloader/main.c
hiviah: Original reason was so that better diffs view that code as moved unlike removed and added with differences, but in C you can't do reference, like `vendor_header&`. In the end the `clang-format` reformatted it anyway, so just shallow copy seemed fine. BTW clang-format is very "non-deterministic in what it likes", it just makes code look "made by machine ugly".

So then changing it to pointer without the copy would be better (compiler generates memcpy there on ARM). I'll have to check what could cause segfault with shallow copy, since all calls are either `const vendor_header*` and the pointers work with the same buffers (original struct is never used again, since it's no-return, which should have been added as attribute probably).

Still weird that the shallow copy looks fine, vhdr used in const * functions, and then it segfaults. Can't catch it with watchpoints, will need to build it with less optimization since I can't see some variables.

hiviah: Apparently `vhdr.origin` points completely outside of process mapped range the moment `firmware_jump_fn` is reached:

image

Changed here:

image

And then for some reason the watchpoint was lost and dunno when changed.

matejcik: not sure what is going on, but there is actually zero reason to pass anything at all. see d27d256a6d5e3183a28f49da7c4497581c71df0f, @TychoVrahe please double-check but I believe this is how it's supposed to work

TychoVrahe: The problem in the before-fix version was that `read_vendor_header` read the vendor header into location pointed at by vhdr_ptr, but the rest of the evaluation was based on the `vhdr` variable. In case of firmware just being installed this was filled by garbage, since there wasn't any vendor header before. 

The fix makes sense to me.

TychoVrahe: Removed the comment and added initialization https://github.com/trezor/trezor-firmware/pull/3122/commits/e52f4c26efefb9c4ae5480ddebe8a284a5b0257c

core/embed/rust/src/ui/model_tr/bootloader/menu.rs
```diff
@@ -93,22 +98,40 @@ impl ChoiceFactory<&'static str> for MenuChoiceFactory {
...
+            let choice_item = MenuChoice::new(
+                Self::CHOICES[choice_index].0,
+                Self::CHOICES[choice_index].1,
+                Self::CHOICES[choice_index].2,
+            );
+            let action = match choice_index {
+                0 => MenuMsg::FactoryReset,
+                1 => MenuMsg::Close,
+                2 => MenuMsg::Reboot,
+                _ => unreachable!(),
```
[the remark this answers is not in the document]
TychoVrahe: weird but nice trick https://github.com/trezor/trezor-firmware/pull/3122/commits/1c58f2fecc59aae966c7592bd26bfa09725ed483

core/embed/bootloader/main.c
```diff
@@ -394,6 +486,8 @@ int bootloader_main(void) {
   // anyway
   uint32_t touched = 0;
 #ifdef USE_TOUCH
+  ensure(dont_optimize_out_true * (firmware_present == firmware_present_backup),
```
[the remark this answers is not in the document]
TychoVrahe: took care of it here https://github.com/trezor/trezor-firmware/pull/3122/commits/48452095d02fa9861ab30c5d5d4f8ee8cb597bbf

core/embed/bootloader/main.c
```diff
@@ -414,54 +508,63 @@ int bootloader_main(void) {
...
-    hal_delay(400);
-#endif
-    // ignore stay in bootloader
-    stay_in_bootloader = secfalse;
-    touched = false;
+  // start the bootloader ...
+  // ... if user touched the screen on start
+  // ... or we have stay_in_bootloader flag to force it
+  // ... or there is no valid firmware
+  ensure(dont_optimize_out_true * (firmware_present == firmware_present_backup),
```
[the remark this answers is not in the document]
TychoVrahe: took care of it here https://github.com/trezor/trezor-firmware/pull/3122/commits/48452095d02fa9861ab30c5d5d4f8ee8cb597bbf

core/embed/bootloader/main.c
```diff
@@ -470,15 +573,16 @@ int bootloader_main(void) {
...
+          ui_result = ui_screen_menu(firmware_present);
+          if (ui_result == 0xAABBCCDD) {  // exit menu
             screen = SCREEN_INTRO;
           }
-          if (ui_result == 2) {  // reboot
+          if (ui_result == 0x11223344) {  // reboot
             ui_screen_boot_empty(true);
-            continue_to_firmware = true;
+            continue_to_firmware = firmware_present;
+            continue_to_firmware_backup = firmware_present;
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/pull/3122/commits/60b1f960259fc224fc55e52c201e1fa8c0a870ae

TychoVrahe:
tried to handle remaining issues.

TychoVrahe:
and one more fix https://github.com/trezor/trezor-firmware/pull/3122/commits/7f93d97ec84bcfb34026d3c96b72347ebd424359

core/embed/bootloader/main.c
```diff
@@ -414,54 +508,64 @@ int bootloader_main(void) {
...
+            case CONTINUE_TO_FIRMWARE:
+              continue_to_firmware = sectrue;
+              continue_to_firmware_backup = sectrue;
+              break;
+            case RETURN_TO_MENU:
+              break;
+            default:
+            case SHUTDOWN:
+              return 1;
+              break;
```
matejcik, line 563: we probably want to call `trezor_shutdown()` here too?

TychoVrahe: Returning from `main` will ultimately have the same effect (=shutdown_privileged is called with infinite loop at the end), so i don't think that matters much. Returning from `real_jump_to_firmware` doesn't do this immediately - although it will end up same in this case as there is nothing more to be done in main. Thats until some more calls are added in between and the information is lost, so better to be explicit. Also `return 1` from `void real_jump_to_firmware(void) ` does not compile so the return code is lost - not that its used at this moment.

matejcik (approved):

bosomt:
**QA OK**

PR #3127, opened 2023-07-04
chore: update actions checkout version

vdovhanych:
was going trough some repositories and updating some of the actions, this only updates checkout as we do not have that much in here.


matejcik (approved):
utACK

PR #3129, opened 2023-07-05
Update derivation path format in the example

kandycoder:
The derivation path format of Bitcoin/0h/0/0 is out of date. Updating to m/84h/0h/0/0 format.


prusnak (changes requested):
I don't think that's the correct path since 84 is used for P2WPKH while the UTXO `a5ea715aa99ca30516f3af6f622dfe7399d883d49ad74b1fe33fdf73d201e64e:0` is of type P2PKH (path 44)

prusnak (approved):

PR #3137, opened 2023-07-12
feat(rust): update trezor-client crate metadata

wszdexdrf:
This PR updates the Manifest Data for `trezor-client` crate to reflect the recent changes in the registry.

Bumps version of trezor-client to 0.1.1.

PR #3139, opened 2023-07-12
feat(rust): update trezor-client crate metadata + run cargo build

prusnak:
Followup to https://github.com/trezor/trezor-firmware/pull/3137

When trying to publish the create I noticed the Cargo.lock file hasn't been updated - run `cargo build` to update the crate version in the lock file.

Also updated the authors section to match reality.

Once this is merged I will tag the commit and publish the trezor-client v0.1.1 crate.


matejcik (approved):
utACK

PR #3140, opened 2023-07-12
ci: allow jobs to be cancelled when redundant

vdovhanych:


matejcik (approved):
OK by me but @mmilata please also take a look.

mmilata (commented):
LGTM, can we make `interruptible: true` [the default](https://docs.gitlab.com/ee/ci/yaml/#default)?

vdovhanych:
> LGTM, can we make `interruptible: true` [the default](https://docs.gitlab.com/ee/ci/yaml/#default)?

I guess we can

mmilata:
Tried it in 4271137ff6f9c6f984eaf71d34b75fa6c24c1f8b and it seems to be working, feel free to cherrypick.

I've opted-out the deploy jobs but it probably doesn't matter all that much.

vdovhanych (commented):
Let's go like this, its a bit better, thanks @mmilata

mmilata (approved):

Issue #3133, opened 2023-07-11
TT success screens redesign


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/2cbb8a16-2ef9-41ff-8e7f-c58e312a1f0e)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/4f27bcd0-e729-4c6e-a4a1-0bfadb6b8a3b)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/35c3540d-0ffd-48de-a03a-656472a52897)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/4820d5ed-54e9-4c23-9efa-a16f6bca64d9)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/69f2fd44-4ac0-493f-8606-13803f7c588e)

PR #3143, opened 2023-07-13
TT success screen design

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3133:
- changes the colors on success screens

image

### DIFF
--- a/core/embed/rust/src/ui/component/image.rs
+++ b/core/embed/rust/src/ui/component/image.rs
@@ -104,6 +104,10 @@ impl BlendedImage {
             self.area_color,
         );
     }
+
+    pub fn height(&self) -> i16 {
+        self.bg.toif.size().y
+    }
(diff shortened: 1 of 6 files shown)


Hannsek:
Please change the tick and its position. Plus change the colour of the text to white.

I have some examples:
image
image
image
image
image

grdddj:
> Please change the tick and its position. Plus change the colour of the text to white.
> 
> I have some examples:

Right, will do - however, I would need exact requirements on how to structure all the elements on the page, e.g.:
- where to render the normal font
- where to render the bold font
- where to show the tick

Something like (guessing from the examples):
- margin from the button should be 20 px
- if only normal there, make the margin 40 px
- tick should start 15 pixels above the text

obrusvit:
Obsolete PR.

Issue #3076, opened 2023-06-14
TOS link not visible in Shamir backup wallet creation


grdddj:
For `TT`, when choosing (Super)Shamir backup, the wallet creation screen does not show `trezor.io/tos` link as it cannot fit the screen because of two-line title

Mentioned at https://github.com/trezor/trezor-firmware/pull/3074#issue-1756683383

![image](https://github.com/trezor/trezor-firmware/assets/42543243/ccb2827c-fd39-46ad-a821-318c6f69775a)

Hannsek:
With new wording, `SHAMIR` should fit on one line. 
![image](https://github.com/trezor/trezor-firmware/assets/57008159/0d3c0c53-cc45-4096-981b-ad0604b6a9cc)

However, `SUPER SHAMIR` wouldn't fit on one line. Is it possible to have different vertical space between the "paragraphs"? 
![image](https://github.com/trezor/trezor-firmware/assets/57008159/cde25e36-7370-4793-98eb-5e360a4cbc0b)

grdddj:
> With new wording, `SHAMIR` should fit on one line. ![image](https://user-images.githubusercontent.com/57008159/246651688-0d3c0c53-cc45-4096-981b-ad0604b6a9cc.png)

Right, we are currently hardcoding the line-break, we can just replace it by normal space

 
> However, `SUPER SHAMIR` wouldn't fit on one line. Is it possible to have different vertical space between the "paragraphs"? ![image](https://user-images.githubusercontent.com/57008159/246651654-cde25e36-7370-4793-98eb-5e360a4cbc0b.png)


I guess it will be possible to get rid of the vertical space when the label has two lines - we are currently putting `\r` at the end of the first paragraph to create the space

prusnak:
I think we should drop "SUPER SHAMIR" and just use "SHAMIR" for this case as well. We don't use this wording anywhere in the product.

grdddj:
> I think we should drop the word "SUPER| from "SUPER SHAMIR" and just use "SHAMIR" for this case as well. We don't use this wording anywhere else in the product.

That would make things easier, I am for it :)

Hannsek:
Fair enough.

PR #3145, opened 2023-07-13
Replace Super Shamir info just with Shamir when creating wallet

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3076:
- making the header always fit on one line


matejcik:
please fix the UI diffs

grdddj:
Seems our assumption about one line with `(SHAMIR)` was wrong - it still occupies two lines 

https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4648640149/artifacts/test_ui_report/master_diff.html

image


Our options are:
- reword the title, e.g. `RECOVER SHAMIR WALLET`
- delete the vertical space between paragraphs

Hannsek:
That is strange as I can see that in figma, there is a space. 
image

What if we get rid of the brackets?
image

grdddj:
> That is strange as I can see that in figma, there is a space. image
> 
> What if we get rid of the brackets? image

We can do this:

image


WIth dash in between, it does not fit

image

matejcik:
why _does_ it fit in figma? is the character spacing modified there or something?
or is it a 1px overflow that we just need to manually override or sth?

Hannsek:
It seems that the padding in fw is not right? I believe it should be 6 px.

grdddj:
Padding in `firmware` is really `6 px`.

I have found some differences in the text width however - e.g. in the `By continuing you agree` sentence, there are 5 (+ 6 margin) free pixels at the right in `Figma`, but only 3 (+6) in `firmware`

This difference comes most likely from one-pixel longer spaces between words in firmware, compared to FIgma (even though the spaces in FIgma are hard to measure)

image

image

Therefore it seems firmware makes all the text little bit wider, which explains rare occasions it fits in Figma and on in firmware. This does not fix the problem however - we might want to open an issue @Hannsek

matejcik:
still, if we ignore the right-side padding, _can_ we fit "Create wallet (shamir)" on firmware screen?

grdddj:
> 



> still, if we ignore the right-side padding, _can_ we fit "Create wallet (shamir)" on firmware screen?

I think we cannot ignore it just for the text, we would have to ignore it for all other components as well - it is being controlled by `theme::borders` function, being called in (probably) all our dialogs:
```rs
/// +----------+
/// |     6    |
/// |  +----+  |
/// | 6|    | 6|
/// |  +----+  |
/// |     6    |
/// +----------+
pub const fn borders() -> Insets {
    Insets::new(6, 6, 6, 6)
}
```

I will try to manually decrease the width of space by one pixel in the font file

matejcik:
> > still, if we ignore the right-side padding, _can_ we fit "Create wallet (shamir)" on firmware screen?
> 
> I think we cannot ignore it just for the text, we would have to ignore it for all other components as well

still. if we _do_ ignore the padding, _can_ we fit the text? please find out how big the overflow is

grdddj:
> still, if we ignore the right-side padding, _can_ we fit "Create wallet (shamir)" on firmware screen?

But yeah, when completely removing the padding (making it 0 on the right), it fits:

image

grdddj:
The overflow seems to be 1 (or 2?) pixels

image

grdddj:
Fits when decreasing the width of space of the title font by 1 pixel - hopefully the space is not too small
```
diff --git a/core/embed/lib/fonts/font_tthoves_bold_17.c b/core/embed/lib/fonts/font_tthoves_bold_17.c
index cf04c0a593..d4cd315dc7 100644
--- a/core/embed/lib/fonts/font_tthoves_bold_17.c
+++ b/core/embed/lib/fonts/font_tthoves_bold_17.c
@@ -6,7 +6,7 @@
 // - the third, fourth and fifth bytes are advance, bearingX and bearingY of the horizontal metrics of the glyph
 // - the rest is packed 4-bit glyph data
 
-/*   */ static const uint8_t Font_TTHoves_Bold_17_glyph_32[] = { 0, 0, 5, 0, 0 };
+/*   */ static const uint8_t Font_TTHoves_Bold_17_glyph_32[] = { 0, 0, 4, 0, 0 };
```

image

grdddj:
We might also introduce marquee to `TT`, so we do not need to hack this the next time we are missing one pixel

matejcik:
> Fits when decreasing the width of space of the title font by 1 pixel - hopefully the space is not too small

also good, but the ux designer will need to have a look at the results

grdddj:
> I would be ok with allocating one more pixel to the right of every title (that can be solved by some sort of title-specific inset/outset, right?) if it helps with this particular 1px difference

Could be done by this (just 1 pixel did not work, had to do 2)

```
--- a/core/embed/rust/src/ui/model_tt/component/frame.rs
+++ b/core/embed/rust/src/ui/model_tt/component/frame.rs
@@ -121,12 +121,13 @@ where
         const TITLE_SPACE: i16 = theme::BUTTON_SPACING;
 
         let bounds = bounds.inset(self.border);
+        let title_extra_right = 2;  // allowing for little longer titles to fit in
         if let Some(b) = &mut self.button {
             let button_side = theme::CORNER_BUTTON_SIDE;
             let (header_area, button_area) = bounds.split_right(button_side);
             let (button_area, _) = button_area.split_top(button_side);
             b.place(button_area);
-            let title_area = self.title.place(header_area);
+            let title_area = self.title.place(header_area.outset(Insets::right(title_extra_right)));
             let remaining = header_area.inset(Insets::top(title_area.height()));
             let subtitle_area = self.subtitle.place(remaining);
 
@@ -141,7 +142,7 @@ where
             let content_area = bounds.inset(Insets::top(header_height + TITLE_SPACE));
             self.content.place(content_area);
         } else {
-            let title_area = self.title.place(bounds);
+            let title_area = self.title.place(bounds.outset(Insets::right(title_extra_right)));
             let remaining = bounds.inset(Insets::top(title_area.height()));
             let subtitle_area = self.subtitle.place(remaining);
             let remaining = remaining.inset(Insets::top(subtitle_area.height()));
```

image

Hannsek:
If this adds unnecessary complexity to code and it solves just this one problem, then we can do some kind of workaround in copy for example.

core/embed/rust/src/ui/model_tt/component/frame.rs
```diff
@@ -121,12 +121,14 @@ where
         const TITLE_SPACE: i16 = theme::BUTTON_SPACING;
 
         let bounds = bounds.inset(self.border);
+        // Allowing for little longer titles to fit in
+        let title_extra_space = Insets::right(2);
```
mmilata, line 125: Nit
```suggestion
        const TITLE_EXTRA_SPACE: Insets = Insets::right(2);
```

mmilata (approved):
LGTM

core/embed/rust/src/ui/model_tt/component/frame.rs
grdddj: Changed in the force-pushed commit [1cf2c52](https://github.com/trezor/trezor-firmware/pull/3145/commits/1cf2c52b4045e522201d8a88fb9c67d296074937)

bosomt:
**QA OK**

i only see Shamir /Super was omited/ and link to tos.

**Info:**
 - Suite version: desktop 23.7.2 (900989955cb6a56111d6827091f60ca059bfcf4f)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.7.2 Chrome/112.0.5615.204 Electron/24.5.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: model T 2.6.1 Universal (revision 4a6420c5dd038c06a6641331995c7b22cf4c699c)

PR #3146, opened 2023-07-14
replace systick based wait in display SPI with loop

TychoVrahe:
replace systick based wait in display SPI with loop as systick does not work in fault handlers, hence the rendering gets stuck


matejcik:
1000 loops instead of 1ms wait seem kinda short...? but if it works :woman_shrugging: 

to confirm, this doesn't really change behavior, as the HAL_Delay call would still be blocking, right? except possibly for power consumption

matejcik (approved):
utACK, maybe @hiviah or @prusnak should have a brief look too?

prusnak:
ACK for replacing with the busy wait - it was my suggestion after all.

Not sure about 1000 cycles, though.

hiviah (approved):

PR #3149, opened 2023-07-14
fix MPU settings

TychoVrahe:
bootloader is actually 128kB.


matejcik (approved):
utACK, @hiviah please double-check

Issue #3052, opened 2023-06-02
Reintroduce fixtures.suggestions.json in CI


mmilata:
Or some other mechanism to obtain the correct fixtures.json. My use-case is avoiding a local test run when the job fails but I see in the diff that the new version is correct. For example after rebasing known-working branch on top of updated master.

PR #3150, opened 2023-07-14
Get UI fixtures from CI

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3052:
- saving `fixtures.results.json` file with the UI hashes of all the tests that were running in a specific test-pipeline
- allowing to download all the CI results and refresh the local `fixtures.json` with a simple script - `python tests/update_fixtures.py ci`

Also found out why the `fixtures.suggestion.json` is not being generated anymore - it requires that some tests are not being run - `if test_ui == "test" and check_missing and list_missing()` ... we might want to change this probably

### DIFF
--- a/tests/gitlab.py
+++ b/tests/gitlab.py
@@ -1,0 +1,136 @@
+"""
+Helper functions for communication with Gitlab.
+
+Allowing for interaction with the test results, e.g. with UI tests.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any, Iterable, Iterator
+
+import requests
+
+AnyDict = dict[Any, Any]
+
+HERE = Path(__file__).parent
+
+BRANCHES_API_TEMPLATE = "https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines.json?scope=branches&page={}"
+GRAPHQL_API = "https://gitlab.com/api/graphql"
+
+UI_JOB_NAMES = (
+    "core click R test",
+    "core device R test",
+    "core click test",
+    "core device test",
+    "core persistence test",
+    "legacy device test",
+)
+
+SAVE_GRAPHQL_RESULTS = False
+
+
+def _get_gitlab_branches(page: int) -> list[AnyDict]:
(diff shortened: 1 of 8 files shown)


tests/gitlab.py
```diff
@@ -0,0 +1,107 @@
...
...typename\n                nodes {\n                  __typename\n                  id\n                  name\n                  kind\n                  scheduledAt\n                  needs {\n                    __typename\n                    nodes {\n                      __typename\n                      id\n                      name\n                    }\n                  }\n                  previousStageJobsOrNeeds {\n                    __typename\n                    nodes {\n                      ...CiNeeds\n                    }\n                  }\n                  status: detailedStatus {\n                    __typename\n                    id\n                    icon\n                    tooltip\n                    hasDetails\n                    detailsPath\n                    group\n                    label\n                    action {\n                      __typename\n                      id\n                      buttonTitle\n                      icon\n                      path\n                      title\n                    }\n                  }\n                }\n              }\n            }\n          }\n        }\n      }\n    }\n  }\n}\n",
```
mmilata, line 47: Is there a possibility we would ever want to change the query? First time seeing graphql tbh.

mmilata (approved):
> requires that some tests are not being run - if test_ui == "test" and check_missing and list_missing() ... we might want to change this probably

That would be good enough solution of this problem for me but I don't mind having a CLI tool that does this ofc

tests/gitlab.py
grdddj: Also my first time with `graphql`. I just copy-pasted the query from somewhere and it seemed to work so I did not even look what is inside.

It is possible we would need to change it, if `Gitlab` changed the interface.

After looking at the query, half of it was rubbish, and we did not need majority of the information there, so I have stripped it to its minimal version:

```
query getJobsFromPipeline($projectPath: ID!, $iid: ID!) {
  project(fullPath: $projectPath) {
    pipeline(iid: $iid) {
      stages {
        nodes {
          groups {
            nodes {
              jobs {
                nodes {
                  id
                  name
                }
              }
            }
          }
        }
      }
    }
  }
}
```

This is getting us just all the job names and IDs, which is enough for this script.

Now, when we wanted to fetch more info, yes, we would need to change the query.

Issue #3134, opened 2023-07-11
TT recovery flow design


Hannsek:
https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?type=design&node-id=8833-74257&mode=design&t=9piUbfqOGa9z5JFr-4

PR #3163, opened 2023-07-19
TT recovery design

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3134:
- new designs for TT recovery flow and mnemonic keyboard


Hannsek:
image
image
image


Can we delete the second sentence from these screens? It is not needed here. IMO it is needed only in the first screen of the whole flow.

grdddj:
Together with some `recovery` flow changes, I tried to do https://github.com/trezor/trezor-firmware/issues/3164 and overall `input-flows` refactor for all the `recovery` tests.

I am quite satisfied with the current state - having `RecoveryFlow` class in `input_flow_helpers.py`, exposing all the possible screens there, and then just composing a sequence of each test-case in each `InputFlowXXX` in `input_flow.py`, e.g.:

```python
class InputFlowBip39RecoveryDryRun(InputFlowBase):
    def __init__(self, client: Client, mnemonic: list[str], mismatch: bool = False):
        super().__init__(client)
        self.mnemonic = mnemonic
        self.mismatch = mismatch

    def input_flow_common(self) -> BRGeneratorType:
        yield from self.REC.confirm_dry_run()
        yield from self.REC.setup_bip39_recovery(len(self.mnemonic))
        yield from self.REC.input_mnemonic(self.mnemonic)
        if self.mismatch:
            yield from self.REC.warning_bip39_dryrun_mismatch()
        else:
            yield from self.REC.success_bip39_dry_run_valid()

```

mmilata (commented):
Code changes LGTM, but please resolve/delete/create issues for the TODOs before merging. Did not look very deep into the test changes but they look ok on the surface.

grdddj:
> Code changes LGTM, but please resolve/delete/create issues for the TODOs before merging. Did not look very deep into the test changes but they look ok on the surface.

Created two issues for that. Have rebased on `master` and recorded all the tests.

Ready to merge, I just need approval - can you please do it? @mmilata  or @matejcik

Hannsek (approved):

bosomt:
**QA OK**

1b08be018603af91969d4072d0e0eafef293d5fb

bosomt:
Please consider changing wording to 2-4 letters, many words needs 4 letters to be entered.

PR #3171, opened 2023-07-24
feat(common): add model and color information (fixes #3160)

### LINKED ISSUE #3160: Add colors information to trezor-common repo
There should be a json dict `common/models/colors.json` with the following structure:
```
{
  "T2B1": {
    0: "Black",
    1: "White"
  }
}
```
Mainly for Connect/Suite consumption but we might need this too.

additionally, we might want to add internal names mapping? `common/models/names.json`
```
{
  "T1B1": "Model One",
  "T2T1": "Model T",
  "T2B1": "Model R"
}
```
how about it @Hannsek ?

matejcik:
fixes #3160

### DIFF
--- a/common/models.json
+++ b/common/models.json
@@ -1,0 +1,17 @@
+{
+    "T1B1": {
+        "name": "Trezor Model One",
+        "colors": {}
+    },
+    "T2T1": {
+        "name": "Trezor Model T",
+        "colors": {}
+    },
+    "T2B1": {
+        "name": "Trezor Model R",
+        "colors": {
+            "0": "Black",
+            "1": "White"
+        }
+    }
+}


mroz22:
what is actually the argument against defining this on protobuf level? Or mybe combination of both. I understand that protobuf does not tell what are to possible combinations of model and color. On the other hand, the more information in protobuf, the stricter generated types for me.

matejcik:
> defining this on protobuf level

what "this"?

matejcik:
argument against **model**:
We created the internal names in a systematic way so that we can fit any Trezor into a four-character string. This four-character string is encoded in firmware and we intend it to be a universal Trezor model identifier for any purpose.
It seems useless to have a lookup dictionary of number <-> internal model when I want to tell someone the four-char string that I have, only to have _them_ convert the number back to the four-char string.

argument against **color**:
Basically, bootloader knows nothing of these enums and I don't want to update bootloader code (incl. generated) every time we add a color. Bootloader just tells you the numeric value of some OTP field. We _could_ do something smarter in firmware, but there doesn't seem to be a reason for it? It would only mean that when we introduce Sparkly Glitter Trezor, we also need to ship updated firmware for it, and it's gonna be broken with old ones.

What are the arguments **for** defining this on protobuf level anyway? You talk about type safety, but you still have to handle unknown / invalid values _somehow_.

common/models.json
```diff
@@ -0,0 +1,17 @@
+{
+    "T1B1": {
+        "name": "Model One",
```
matejkriz, line 3: Current device display name is e.g. "Trezor Model One", can you please add `Trezor ` prefix for all models?

common/models.json
```diff
@@ -0,0 +1,17 @@
...
+    },
+    "T2T1": {
+        "name": "Model T",
+        "colors": {}
+    },
+    "T2B1": {
+        "name": "Model R",
+        "colors": {
+            "0": "Black",
+            "1": "White"
```
matejkriz, line 14: we would prefer lowercase, because we will primary use it for filenames
```suggestion
            "0": "black",
            "1": "white"
```

common/models.json
matejcik: fixed in force-push

common/models.json
matejkriz: I take it back, we can also use a number as a key

matejkriz (approved):

Issue #2923, opened 2023-03-30
Allowance checking in fw


Hannsek:
> we might actually want to implement allowance checking in firmware even before the general contract decoding. it looks easy, it's a call on the ERC20 ABI with address and amount or UINT_MAX

related issue: https://github.com/trezor/trezor-firmware/issues/69

PR #3172, opened 2023-07-25
feat(core): Make allowance check nicer (implements #2923)

cbergqvist:
Properly displays recipient address and allowance amount for the `approve` function, similar to ERC20 token transfers.

[no changelog]


core/src/apps/ethereum/layout.py
```diff
@@ -139,6 +165,26 @@ async def confirm_typed_data_final() -> None:
     )
 
 
+def require_confirm_approve_tx(
+    address_bytes: bytes,
+    allowance: int,
+    network: EthereumNetworkInfo,
+    token: EthereumTokenInfo,
+) -> Awaitable[None]:
+    from trezor.ui.layouts import confirm_action
```
grdddj, line 174: Small nit (you could not know) - we are already using `confirm_action` in two functions in this file, and in this case we import it on top level instead of inside the functions. (Due to flash space savings)

So please move the import on top level.

(The same is actually true for `hexlify`, please do that as well.)

grdddj (commented):
Looked at it briefly, just found one possible code improvement.

Let's wait until CI finishes to see the UI results.

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/944411639

Actually, the `style` job has already failed - there are some `python` typing inconsistencies. Try fixing that locally with the help of `make pystyle`

core/src/apps/ethereum/layout.py
cbergqvist: Thanks for the feedback, fixes pushed.

cbergqvist:
@grdddj please mirror commit ea86110 which I just added as well

grdddj:
> @grdddj please mirror commit [ea86110](https://github.com/trezor/trezor-firmware/commit/ea861107e8af94a1f4397e6e8dfa180ef4f8b9a7) which I just added as well

Done, the CI is running here
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/945769691

obrusvit:
superseded by #4958

Issue #3118, opened 2023-06-29
Create a mapping of Figma designs to UI tests


grdddj:
We want to have an easy way of comparing Figma designs with the current designs in firmware.

Something like:
- `TR recovery flow happy path` = `TR-device_tests-reset_recovery-test_recovery_bip39_t2.py::test_tt_pin_passphrase`
- `TT removing wipe code` = `TT-device_tests-test_msg_change_wipe_code_t2.py::test_set_remove_wipe_code`
- ...

Also, this will uncover designs missing in Figma and/or missing tests for some screens.

matejcik:
shouldn't this info be in figma rather than the firmware repo?

grdddj:
> shouldn't this info be in figma rather than the firmware repo?

Yes, could be easily in `Figma`/`Notion`, so it is independent of the repository and we can change it without PRs.

As for this issue, this was the natural place to open it, we can move it elsewhere, if it will not get lost/forgotten

grdddj:
Incomplete mapping created while checking TR design:

```
Tutorial
- TR-click_tests-test_tutorial.py::test_tutorial_again_and_skip

Wallet creation, Seed backup
- TR-click_tests-test_reset_bip39.py::test_reset_bip39
- TR-click_tests-test_reset_slip39_advanced.py::test_reset_slip39_advanced[2of2]
- TR-click_tests-test_reset_slip39_basic.py::test_reset_slip39_basic[1of1]

Recovery
- TR-click_tests-test_recovery.py::test_recovery_bip39
- TR-click_tests-test_recovery.py::test_recovery_slip39_basic

PIN
- TR-device_tests-test_msg_changepin_t2.py::test_set_pin
- TR-device_tests-test_msg_changepin_t2.py::test_set_failed
- TR-device_tests-test_msg_changepin_t2.py::test_change_pin
- TR-device_tests-test_msg_changepin_t2.py::test_change_invalid_current
- TR-device_tests-test_msg_changepin_t2.py::test_remove_pin
- TR-click_tests-test_pin.py::test_pin_short

Wipe code
- TR-device_tests-test_msg_change_wipe_code_t2.py::test_set_remove_wipe_code
- TR-device_tests-test_msg_change_wipe_code_t2.py::test_set_wipe_code_to_pin

Send BTC
- TR-device_tests-bitcoin-test_signtx_amount_unit.py::test_signtx_testnet[None]
[...]

PR #3175, opened 2023-07-26
feat(tests): number the screens in UI test reports, give unique IDs to images

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3118:
- prepares the UI tests for automatic retrieval of specific images and allows for linking to a given screen

The mapping itself and the application/dashboard code are currently stored at https://github.com/grdddj/figma_ui repository

The app itself is running under https://figma.grdddj.eu


Hannsek (approved):
utACK

Issue #2902, opened 2023-03-23
Buttons logic with long text


Hannsek:
![Image](https://user-images.githubusercontent.com/57008159/232702373-f8264f6e-798f-40c9-b599-0380e35dbdcd.png)

If the text is very long. We should display only the beginning of the text on the first screen together with 3 buttons: `SHOW ALL DATA`, `CONFIRM`, `˄`.

After clicking on `SHOW ALL DATA`, the whole text should be displayed on multiple screens with buttons like this:

<img width="151" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/a6b5ae2a-fad9-451d-b03b-cc833bc39f5e">

Hannsek:
related issue: #2888

Hannsek:
related issue: https://github.com/trezor/trezor-firmware/issues/3000

PR #3177, opened 2023-07-26
Display QR codes for XPUBs

### LINKED ISSUE #3047: Behaviour of the xpub displaying should be aligned with address displaying
https://github.com/trezor/trezor-suite/issues/3454#issuecomment-1569917006

xpub qr code under `i` button, missing x -> xpub mismatch?

mmilata:
Fixes #3047.

Also includes fix for Cardano to display the same hex data (including the chain code) that it sends over protobuf.

[master diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4746721205/artifacts/master_diff/master_diff.html)

1670730c0235350b17bf58f2f31f01562468f71fb957ea163fd183648b0e72a1 e75aa91d0e3ef660a63b2834d1dc18da9b160589c3e56ed7562ab99e748fa0a2 8aa32b0486b73d060db0432aa20b45ce547d8efdd82116d9fb2b24c115e02d48 16cc93ba0ab78064e85cbe93583726172df862c94fc6301a9be3e0c9702cedc9 c744383b9196672482f3b817379cf58820492f8e54cbc6d93368f8a237be3cad 79c110dfa36938ab6992e84a01a1adad1d0cbb3a000f7a28b30c7662d7b28a8d


Hannsek:
For T2B1, please redo it the same as [receiving address flow](https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?type=design&node-id=8825-68714&mode=design&t=Zf7pQpB2cW7lEG9D-4).

What is the difference between public key and xpub? 
image
Does it fit on the screen? I can see from the test that the letter "Y" is touching the "i" button.

The buttons used here are wrong: 
image
It should be like this: 
image
But it would be fixed by resolving [this](https://github.com/trezor/trezor-firmware/issues/2902) issue, right?

Hannsek:
image

Can the space between _trezor.io/support_ and the text above be decreased a bit?




Isn't the derivation path here too long? 
image

tests/input_flows.py
```diff
@@ -385,6 +385,32 @@ def input_flow_tr(self) -> GeneratorType:
...
+
+    def input_flow_tt(self) -> GeneratorType:
+        yield
+        self.debug.click(buttons.CORNER_BUTTON, wait=True)
+        # synchronize; TODO get rid of this once we have single-global-layout
+        self.debug.synchronize_at("HorizontalPage")
+
+        self.debug.swipe_left(wait=True)
+        self.debug.swipe_right(wait=True)
+        self.debug.swipe_left(wait=True)
```
grdddj, line 400: Do we want to swipe so many times to show both the QR code and account information twice? Theoretically just one `swipe_left` would be enough to show both of them

core/src/trezor/ui/layouts/tt_v2/__init__.py
```diff
@@ -457,8 +446,10 @@ def xpub_title(i: int) -> str:
             result = await ctx_wait(
                 RustLayout(
                     trezorui2.show_address_details(
+                        title=title,
                         address=address if address_qr is None else address_qr,
                         case_sensitive=case_sensitive,
+                        subtitle=subtitle,
```
grdddj, line 452: `title` and `subtitle` are little misleading names here, as they are not displayed on the same page.

`AddressDetails::new` has it as `qr_title` and `details_title`, we might want to make it so in `trezorui2.show_address_details` as well

core/src/apps/bitcoin/get_public_key.py
```diff
@@ -79,9 +79,26 @@ async def get_public_key(
...
-        from trezor.ui.layouts import show_xpub
+        from trezor.ui.layouts import show_pubkey
+        from .keychain import address_n_to_name
+        from apps.common.paths import address_n_to_str
 
-        await show_xpub(node_xpub, "XPUB")
+        path = address_n_to_str(address_n)
+        account_name = address_n_to_name(coin, address_n, script_type)
+        if account_name is None:
+            account = "Unknown path"
```
grdddj, line 89: `Unknown path` or `Unknown account`?

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -1635,7 +1636,7 @@ pub static mp_module_trezorui2: Module = obj_module! {
 
     /// def show_mismatch() -> object:
```
grdddj, line 1637: Missing `*, title: str` arguments

grdddj (commented):
Some small questions and change requests

tests/input_flows.py
Hannsek: What do you mean by "so many times"?

You should see xpub (swipe left) qr code (swipe left) account info

core/src/apps/bitcoin/get_public_key.py
Hannsek: `Unknown account` is more user friendy.

tests/input_flows.py
grdddj: These are the test-case screens, we are going twice there and back through the QR code and details, the middle screens do not show any new information. Maybe it is testing that we can successfully go back and click corner button from the second screen? It is fully valid, just wondering

image

Hannsek: 🤷🏻

mmilata: Yes, it's meant to test the code for going back. May be already covered by the receive flow testcase but it costs us almost nothing.

core/src/trezor/ui/layouts/tt_v2/__init__.py
mmilata: 9fda87ed8bbd0d362d51c4b45e1d18edfd41419b

core/src/apps/bitcoin/get_public_key.py
mmilata: 9fda87ed8bbd0d362d51c4b45e1d18edfd41419b

core/embed/rust/src/ui/model_tr/layout.rs
mmilata: 9fda87ed8bbd0d362d51c4b45e1d18edfd41419b

mmilata:
@Hannsek 

> For T2B1, please redo it the same as [receiving address flow](https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?type=design&node-id=8825-68714&mode=design&t=Zf7pQpB2cW7lEG9D-4).

Fixed in 9fda87ed8bbd0d362d51c4b45e1d18edfd41419b

> What is the difference between public key and xpub?

XPUB is an [encoding of a public key and additional data](https://en.bitcoin.it/wiki/BIP_0032#Serialization_format), notably the _chain code_ which allows for deriving child keys. Some altcoins only output bare public key - my assumption is that these are account-based coins that only use one key as they don't need to avoid address reuse, unlike UTXO-based coins.

>  Does it fit on the screen? I can see from the test that the letter "Y" is touching the "i" button.

Changed from `Confirm public key` to `Public key` in 9fda87ed8bbd0d362d51c4b45e1d18edfd41419b

> The buttons used here are wrong:
> But it would be fixed by resolving https://github.com/trezor/trezor-firmware/issues/2902 issue, right?

Yes.

> Can the space between trezor.io/support and the text above be decreased a bit?

9fda87ed8bbd0d362d51c4b45e1d18edfd41419b

> Isn't the derivation path here too long?

Good point, this is from an existing test case, normal users won't see this kind of path. There's also a more realistic test case for `m/44'/0'/10'`, should have chosen that.

grdddj:
Looks like the cancel button has disappeared from some TR's screens, e,g,

image

Generally the master-UI-diff for `TR` is quite big - even the `PIN` entry has some changes - is it some change in centering the title? https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4955951054/artifacts/test_ui_report/master_diff.html

There are also some hard failures in both device_tests

Hannsek:
> Isn't the derivation path here too long?

> Good point, this is from an existing test case, normal users won't see this kind of path. There's also a more realistic test case for m/44'/0'/10', should have chosen that.

Sure, but still. There is btc legacy #11 and the derivation path ends with /100. Shouldn’t it end with /10

mmilata:
> Looks like the cancel button has disappeared from some TR's screens, e,g,

Probably due to very old base branch, rebased to current master, device tests should be green now, please take a look.

> Sure, but still. There is btc legacy # 11 and the derivation path ends with /100. Shouldn’t it end with /10?

For the path `m/44'/0'/10'/1/100` the account number corresponds to the `10'` here, template is `m/44'/coin_type'/account'/change/address_index`. Perhaps this ought to show "Unknown account" since the XPUB is technically not for the account but for a particular address in the account, which might be confusing.

grdddj:
> > Looks like the cancel button has disappeared from some TR's screens, e,g,
> 
> Probably due to very old base branch, rebased to current master, device tests should be green now, please take a look.

Right, it helped. There is still diff in middle-button text, but it was not modified in this PR, so it will be fine after merging:
image

grdddj (approved):
LGTM!

Hannsek:
Let's put warning if non-standard derivation path for xpub is used. 
image

Also please put "non-standard" under the "account" header.

Hannsek:
But "normal" path warning says "wrong derivation path for _selected account_". Here, you did not select any account, no?

matejcik:
> "wrong derivation path for _selected account_". Here, you did not select any account, no?

The typical user presumably _did_ select an account for which to show xpub. If malware is changing their derivation path, the message is exactly right.
It also seems to work fine for a user who wants to export an xpub manually but mistypes the path? Presumably they still have a particular account in mind.

Users doing weird things on purpose either (a) know what they're doing and will not get confused, or (b) don't know what they're doing and as such are out of scope for this message :))

mmilata:
Updated to show:
f249f429d659ae32c89ed3039045924a14ce93905fd5055ce7cce169e12d5809 ... 049cb093e14909c0c9992a408085971f2e732b948d2b99532b2a8e83ef2d4aac

[e877be0](https://github.com/trezor/trezor-firmware/pull/3177/commits/e877be0ced4ad03fc2567214fde4dc84809e3f22) changes the path warning screen in entire codebase

Hannsek:
@mmilata Can you please rebase this pr as the screens are now difficult to review as there are lots of changes and regressions.

Hannsek:
Pipeline was skipped for some reason. 🤨

mmilata:
Rebased
[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/5048153199/artifacts/master_diff/index.html) | [master diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/5048153199/artifacts/master_diff/master_diff.html)

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 23.10.0 (7e7e183598efd9ee3413d74f0689322811bbc1e9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/23.10.0 Chrome/112.0.5615.204 Electron/24.5.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: Trezor T2T1 2.6.2 regular (revision b3d0fb68ffe31301af8d569f1107c57d0b4b1e77)

bosomt:
**QA OK**

tested XPUB QR codes:

TS3 - Bitcoin
TS3 - Cardano
TT - Bitcoin
TT - Cardano

**Info:**
 - OS: MacIntel
 - Screen: 1440x900
 - Device: Trezor T2B1 2.6.2 bitcoin-only (revision 6e5967af25de87bebd331d697e6ef940e64286ef)
 - Device: Trezor T2T1 2.6.3 regular (revision c937f21447966b31e930ae8d20d1093d20e9fa66)

PR #3185, opened 2023-07-27
prodtest for T2B1

TychoVrahe:
Adds a prodtest and secret handling for T2B1. 

Prodtest vendor heeader added, needs proper pubkeys.


TychoVrahe:
rebased onto #3186

andrewkozlik:
Implemented:
- The OPTIGA provisioning functions in prodtest.
- Metadata comparison, so that we can check the metadata was written correctly.
- The secure channel.

Not tested yet, but I consider it ready for review.

There are two TODO items, before we can ship it:
- Change `OPTIGA_LCS_OPERATIONAL` to `0x07`, now it's set to 0x01, so that it doesn't actually lock the OPTIGA.
- Uncomment the `secret_*()` functions in `prodtest/main.c`.

andrewkozlik:
Tested and works.

@onvej-sl have a look at the slot configuration https://github.com/trezor/trezor-firmware/pull/3185/files#diff-ac12d8a48d7d0bfd09185b086a5978facae26e5b2a7c59574e1ea37a8d816d03R732-R783 (in core/embed/prodtest/main.c) and changes in `trezor-crypto`.
@matejcik have a look at the rest.

core/embed/prodtest/main.c
```diff
@@ -516,6 +606,415 @@ static void test_otp_write_device_variant(const char *args) {
...
+  if (EXIT_SUCCESS != aes_ret) {
+    vcp_printf("ERROR: aes_decrypt_key256 error.");
+    memzero(&ctx, sizeof(ctx));
+    memzero(secret, sizeof(secret));
+    return;
+  }
+
+  // Decrypt the FIDO attestation key.
+  uint8_t fido_key[32] = {0};
+  uint8_t iv[16] = {0};
```
matejcik, line 977: you're not setting the IV

core/embed/prodtest/main.c
```diff
@@ -91,6 +105,15 @@ static void vcp_readline(char *buf, size_t len) {
   }
 }
 
+static void vcp_printf_ex(const char *fmt, ...) {
```
matejcik, line 108: `_ex` suggests to me an extended argument version, but this seems to be "no newline" version
maybe rename original to `vcp_println` and this one to `vcp_print`?

core/embed/prodtest/main.c
```diff
@@ -101,6 +124,66 @@ static void vcp_printf(const char *fmt, ...) {
   vcp_puts("\r\n", 2);
 }
 
+static void vcp_write_as_hex(uint8_t *data, uint16_t len) {
```
matejcik, line 127: `vcp_println_hex` then?

core/embed/vendorheader/T2B1/vendor_prodtest.json
```diff
@@ -0,0 +1,19 @@
...
+    "trust": {
+        "allow_run_with_secret": true,
+        "show_vendor_string": false,
+        "require_user_click": false,
+        "red_background": false,
+        "delay": 0
+    },
+    "pubkeys": [
+        "c17d221529b70da08aa42ccb1b2cef773736a7d48434209f6eb16b1b474716d2",
+        "6622ee8403464acfc2fe1e20311937b115e60771f3ee6e70f22bf2921977e534"
```
matejcik, line 17: let's not commit these pubkeys until we know the true values

matejcik (commented):
mostly LGTM, apart from the IV usage for FIDO.

@andrewkozlik if you have AES-CCM for Optiga anyway, maybe we can use that for the FIDO secret too?
(of course we have different ways of checking authenticity, e.g., we could locally verify the resulting pubkey against the certificate)

I didn't look at crypto code at all.

I will also do a cursory review of code under `optiga/`, but I don't have a good idea of what's going on in there so the best I can do is hopefully spot some C mistakes or something :woman_shrugging:

core/embed/prodtest/main.c
andrewkozlik: The IV is intentionally all-zero. I don't see any problem in that since the encryption key is unique for each ciphertext.

andrewkozlik:
> @andrewkozlik if you have AES-CCM for Optiga anyway, maybe we can use that for the FIDO secret too?

We certainly can. The reason why I decided to go with an unauthenticated mode are:
- I see no added value in the authentication. What we need is confidentiality and CBC is good enough for that. The ephemeral public key of the sender is not authenticated anyway. So if we did want authentication, then we would first need to authenticate the public key of the sender in order for the usage of an authenticated mode to add any value.
- The absence of the authentication tag makes it easier to test the provisioning process without having to actively compute the ciphertext. Irrelevant in the long term, but comes in handy at this moment.
- One advantage which we lose is that an authenticated mode would help to detect transmission errors, which CBC won't. However, the same goal can be achieved by using the `KEYFIDO READ` command. Checking with this command has an added advantage that it verifies not only that the key was transmitted correctly, but also that it was stored correctly.
- An unauthenticated mode produces shorter ciphertext.

> I will also do a cursory review of code under `optiga/`, but I don't have a good idea of what's going on in there so the best I can do is hopefully spot some C mistakes or something

That's mostly just refactoring to make the command encoding more legible, more easy to extend the commands in the future and unify the style in which the commands are encoded. I retested all the commands, so functionally it is equivalent to the earlier version. The only command that was added is the signature verification command.

core/embed/vendorheader/T2B1/vendor_prodtest.json
andrewkozlik: 945fa14d30f0ac3dbd0be3856bcddc919b1aa45b

core/embed/prodtest/main.c
andrewkozlik: e14114794714d5f8982a2c765fbe99d8d72c159e and e14114794714d5f8982a2c765fbe99d8d72c159e

core/embed/prodtest/main.c
andrewkozlik: e14114794714d5f8982a2c765fbe99d8d72c159e and e14114794714d5f8982a2c765fbe99d8d72c159e

andrewkozlik:
Now that I look at the already existing `prodtest` commands, I see that they don't put a colon `:` after the "OK" and "ERROR" in the response messages, e.g.
```
vcp_println("OK %d %d", x, y);
vcp_println("ERROR sdcard_read_blocks (0)");
vcp_println("OK %s", (const char *)data);
```
I added a fixup commit which removes the `:` from the new command responses https://github.com/trezor/trezor-firmware/pull/3185/commits/04c5d1bed7896703c2a6de5fdf25873d4b1631a0.

core/embed/prodtest/main.c
matejcik: sure but then the comments should reflect that

andrewkozlik: c77f1114a65208be5c28e667ca4addc01471c72a

Don't resolve until @onvej-sl confirms my assessment.

matejcik: what i meant is, a couple lines above a comment claims that the sha256 result is used for key _and IV_
(this comment is also good ofc)

core/embed/trezorhal/optiga/optiga_transport.c
```diff
@@ -406,8 +443,8 @@ static optiga_result optiga_receive_packet(uint8_t *packet_control_byte,
   return OPTIGA_SUCCESS;
 }
 
-optiga_result optiga_execute_command(
-    bool presentation_layer, const uint8_t *command_data, size_t command_size,
+static optiga_result optiga_transcieve(
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3185/commits/41ed809a7225e848f0343d4f841d438b9e556e5d
https://github.com/trezor/trezor-firmware/pull/3185/commits/97eb7c7ba4fdf1d57d6b50970f201f90e4c8ee4a

core/embed/prodtest/main.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3185/commits/9481fb8482a214ca14cda26974221033d14b9126

crypto/aes/aesccm.c
```diff
@@ -0,0 +1,323 @@
...
+ * OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include <string.h>
+
+#include "aesccm.h"
+#include "memzero.h"
+
```
onvej-sl, line 27: I would mention rfc3610 here or in the header file.

crypto/aes/aesccm.c
```diff
@@ -0,0 +1,323 @@
...
+  memzero(ctx, sizeof(*ctx));
+  return EXIT_SUCCESS;
+}
+
+static AES_RETURN aes_ccm_init(aes_encrypt_ctx *encrypt_ctx,
+                               const uint8_t *nonce, size_t nonce_len,
+                               const uint8_t *adata, size_t adata_len,
+                               size_t plaintext_len, size_t mac_len,
+                               cbc_mac_context *cbc_ctx, uint8_t *ctr_block) {
+  if (mac_len < 4 || mac_len > AES_BLOCK_SIZE || mac_len % 2 != 0) {
```
onvej-sl, line 151: I'm just wondering what is the use case for using a mac with the length of 32 bits. Shouldn't we be more restrictive to avoid misuse? I have no strong opinion, I'm just thinking aloud.

crypto/aes/aesccm.c
```diff
@@ -0,0 +1,323 @@
...
+
+    if (cbc_mac_update(cbc_ctx, block, block_size) != EXIT_SUCCESS ||
+        cbc_mac_update(cbc_ctx, adata, adata_len) != EXIT_SUCCESS ||
+        cbc_mac_update_zero_padding(cbc_ctx) != EXIT_SUCCESS) {
+      return EXIT_FAILURE;
+    }
+  }
+
+  // Initialize counter.
+  ctr_block[0] = q - 1;
```
onvej-sl, line 219: Although `ctr_block` was zeroed in the caller's method, I would not rely on that. I would explicitly zero it here.

crypto/aes/aesccm.c
```diff
@@ -0,0 +1,323 @@
...
+      memzero(ctx, sizeof(*ctx));
+      return EXIT_FAILURE;
+    }
+    ctx->pos = 0;
+  }
+
+  size_t block_count = size >> AES_BLOCK_SIZE_P2;
+  size %= AES_BLOCK_SIZE;
+
+  if (!ALIGN_OFFSET(data, 4)) {
```
onvej-sl, line 72: I'm strongly against such optimizations unless we know the code is too slow for our purposes.

In short, my opinion is that the increase in speed does not worth the complexity added to the code. I have some arguments both for and against and I will write them out here, in case you don't like either of these suggestions:
  * remove the optimization completely
  * factor out `void memxor (void *destination, const void *source, size_t length)` implemented as a function or as a macro that would encapsulate the optimization ([this](https://github.com/bminor/glibc/blob/master/string/memcpy.c) or [this](https://github.com/bminor/newlib/blob/master/newlib/libc/string/memcpy.c) can server as an inspiration)

crypto/tls_prf.c
```diff
@@ -0,0 +1,83 @@
...
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include <string.h>
+
+#include "hmac.h"
+#include "memzero.h"
+#include "sha2.h"
+
```
onvej-sl, line 28: I would mention rfc5246 section 5.

crypto/tls_prf.c
```diff
@@ -0,0 +1,83 @@
...
+void tls_prf_sha256(const uint8_t *secret, size_t secret_len,
```
onvej-sl, line 29: What do you think about the following alternative implementation? I expect that the memory and time complexity is about the same, but in my opinion, the readability is better.

```
void tls_prf_sha256(const uint8_t *secret, size_t secret_len,
                    const uint8_t *label, size_t label_len, const uint8_t *seed,
                    size_t seed_len, uint8_t *output, size_t out_len) {
  uint8_t a[SHA256_DIGEST_LENGTH] = {0};
  uint8_t result[SHA256_DIGEST_LENGTH] = {0};

  HMAC_SHA256_CTX hmac_ctx = {0}, hmac_ctx_backup = {0};
  hmac_sha256_Init(&hmac_ctx_backup, secret, secret_len);

  // a = HMAC(secret, label + seed)
  hmac_ctx = hmac_ctx_backup;
  hmac_sha256_Update(&hmac_ctx, label, label_len);
  hmac_sha256_Update(&hmac_ctx, seed, seed_len);
  hmac_sha256_Final(&hmac_ctx, a);

  while (1) {
    // result = HMAC(secret, a + label + seed)
    hmac_ctx = hmac_ctx_backup;
    hmac_sha256_Update(&hmac_ctx, a, SHA256_DIGEST_LENGTH);
    hmac_sha256_Update(&hmac_ctx, label, label_len);
    hmac_sha256_Update(&hmac_ctx, seed, seed_len);
    hmac_sha256_Final(&hmac_ctx, result);

    if (out_len <= SHA256_DIGEST_LENGTH) {
      break;
    }

    memcpy(output, result, SHA256_DIGEST_LENGTH);
    output += SHA256_DIGEST_LENGTH;
    out_len -= SHA256_DIGEST_LENGTH;

    // a = HMAC(secret, a)
    hmac_ctx = hmac_ctx_backup;
    hmac_sha256_Update(&hmac_ctx, a, SHA256_DIGEST_LENGTH);
    hmac_sha256_Final(&hmac_ctx, a);
  }

  memcpy(output, result, out_len);
  memzero(&hmac_ctx, sizeof(hmac_ctx));
  memzero(&hmac_ctx_backup, sizeof(hmac_ctx_backup));
  memzero(a, sizeof(a));
  memzero(result, sizeof(result));
}
```

core/embed/prodtest/main.c
onvej-sl: I agree that the initialization vector can be all zeros if every key is used only once. This has to be guaranteed by the server performing the encryption.

By the way, it seems to me that it is not difficult to introduce an arbitrary initialization vector here. This would save us from these considerations.

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+  if (!set_metadata(OID_KEY_FIDO, &metadata)) {
+    return;
+  }
+
+  // Set metadata for pairing key.
+  memzero(&metadata, sizeof(metadata));
+  metadata.lcso = OPTIGA_LCS_OPERATIONAL;
+  metadata.change = OPTIGA_ACCESS_NEVER;
+  metadata.read = OPTIGA_ACCESS_NEVER;
+  metadata.execute = OPTIGA_ACCESS_ALWAYS;
```
onvej-sl, line 780: This slot is used only in the secure channel, it is not explicitly used. Is that true?

I was unable to find in the documentation whether execute permission is required for these purposes.

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+  metadata.data_type = TYPE_PTFBIND;
+  if (!set_metadata(OID_KEY_PAIRING, &metadata)) {
+    return;
+  }
+
+  vcp_println("OK");
+}
+
+static secbool is_optiga_locked(void) {
```
onvej-sl, line 789: I'm wondering what the use case for this function is. Shouldn't this function also check that the slots are configured correctly?

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+                OID_OPTIGA_UID);
+    return;
+  }
+
+  vcp_print("OK ");
+  vcp_println_hex(optiga_id, optiga_id_size);
+}
+
+static void cert_read(uint16_t oid) {
+  uint8_t cert[1024] = {0};
```
onvej-sl, line 840: I don't know where this number came from, but it seems to me that the certificate can have up to 1727 bytes. See [this](https://github.com/Infineon/optiga-trust-m/blob/develop/documents/OPTIGA%E2%84%A2%20Trust%20M%20Solution%20Reference%20Manual.md#link24b48059_db81_40f5_8b65_7afca4918ab1).

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+  }
+
+  vcp_print("OK ");
+  vcp_println_hex(public_key, public_key_size);
+}
+
+static void keyfido_write(char *data) {
+  static const size_t EPH_PUB_KEY_SIZE = 33;
+  static const size_t PAYLOAD_SIZE = 32;
+  static const size_t CIPHERTEXT_OFFSET = EPH_PUB_KEY_SIZE;
```
onvej-sl, line 935: Is it intentionally defined as static?

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+  optiga_result ret =
+      optiga_set_data_object(0xe0e8, false, (const uint8_t *)"\0", 1);
+  if (OPTIGA_SUCCESS != ret) {
+    vcp_println("ERROR optiga_set_data error %d for 0xe0e8.", ret);
+    return;
+  }
+
+  // Set data object metadata.
+  static const optiga_metadata_item ACCESS_PAIRED = {
+      (const uint8_t *)"\x20\xE1\x40", 3};
```
onvej-sl, line 728: It would be nice if you could make use of the value defined [here](https://github.com/trezor/trezor-firmware/blob/213e9aa2b7ad193102ff4854442c4340b7846783/core/embed/prodtest/main.c#L625C28-L625C28). Perhaps a macro could be useful.

TychoVrahe:
in https://github.com/trezor/trezor-firmware/commit/d822bce2a510db028b2a6aa29d7cb8dacf558921 and https://github.com/trezor/trezor-firmware/commit/0be5e8111589f86f315dc2431c9b886015424c85

i re-added the attestation delete to booloader along with the required screens.

TychoVrahe:
more screens modifications: https://github.com/trezor/trezor-firmware/commit/1994ae1f1e489d282d0e8758a618e4e2bc9cc8f1

TychoVrahe:
dev keys usage in bootloader: https://github.com/trezor/trezor-firmware/commit/54dc3fdb8b46b5ecb65112141686f348e5687a7a

currently enabled by BOOTLOADER_DEVEL makefile/env variable

core/embed/trezorhal/optiga/optiga_commands.c
```diff
@@ -185,18 +210,44 @@ optiga_result optiga_serialize_metadata(const optiga_metadata *metadata,
   return OPTIGA_SUCCESS;
 }
 
+bool optiga_compare_metadata(const optiga_metadata *expected,
```
onvej-sl, line 213: I just want to emphasize that this function evaluates the metadata as the same even if  `stored` contains an extra tag.

onvej-sl: At least, I would mention it in a comment.

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,451 @@ static void test_otp_write_device_variant(const char *args) {
...
+  AES_RETURN aes_ret = aes_decrypt_key256(secret, &ctx);
+  if (EXIT_SUCCESS != aes_ret) {
+    vcp_println("ERROR aes_decrypt_key256 error.");
+    memzero(&ctx, sizeof(ctx));
+    memzero(secret, sizeof(secret));
+    return;
+  }
+
+  // Decrypt the FIDO attestation key.
+  uint8_t fido_key[32] = {0};
```
onvej-sl, line 1006: I would use `PAYLOAD_SIZE` here.

core/embed/trezorhal/optiga/optiga_commands.c
```diff
@@ -33,9 +33,17 @@
 #include "sha2.h"
 
 // Static buffer for commands and responses.
-static uint8_t tx_buffer[1750] = {0};
+static uint8_t tx_buffer[OPTIGA_MAX_COMMAND_SIZE] = {0};
```
onvej-sl: Not directly related to this pull request: I don't like using global static variables. I would suggest defining a structure that contains the buffer and its size, and then passing it explicitly as a parameter.

onvej-sl: The same applies to the frame status and channel status variables in `optiga_transport.c`.

core/embed/prodtest/main.c
andrewkozlik: The server generates an ephemeral key pair for each encryption, so it's guaranteed. Let's keep it simple.

crypto/aes/aesccm.c
andrewkozlik: This is the range of allowed MAC lengths specified by the NIST standard and they have a test vector for it. The choice of MAC length is a protocol-level matter. I don't see a reason to restrict it here.

crypto/aes/aesccm.c
andrewkozlik: This is just an adaptation of the existing CBC implementation in `aes_modes.c`. I left out `FAST_BUFFER_OPERATIONS`, because I don't see any portability issue, and even if there was one, other functions in trezor-crypto already rely on this feature. So this code just respects the existing convention of how AES modes are implemented.

crypto/tls_prf.c
andrewkozlik: In this case, with the current implementation of SHA and HMAC, it looks like it's OK to copy the context `HMAC_SHA256_CTX`. However, in general contexts should be treated as opaque handles whenever possible, because the underlying implementation may change. So I tried to use the existing API to produce an efficient implementation. I think the proposed code costs 2 extra SHA transforms per output block, although that could probably be fixed.

core/embed/prodtest/main.c
andrewkozlik: It is used only in the secure channel. Execute permission is required.

core/embed/prodtest/main.c
andrewkozlik: The correct configuration is checked in the `LOCK` command. This is needed for the `VARIANT` command.

crypto/aes/aesccm.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3185/commits/879086fd608dbe569a8309c330be8e562fa6650e

crypto/aes/aesccm.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3185/commits/879086fd608dbe569a8309c330be8e562fa6650e

crypto/tls_prf.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3185/commits/c24a1461bc45aa94b63f2258d0e9aa71e82610c3

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+}
+
+static void optiga_lock(void) {
+  if (!pair_optiga()) {
+    return;
+  }
+
+  // Delete trust anchor.
+  optiga_result ret =
+      optiga_set_data_object(0xe0e8, false, (const uint8_t *)"\0", 1);
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3185/commits/42e936a623ab7d7f6a9530e0112f250c69bebe6a

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,446 @@ static void test_otp_write_device_variant(const char *args) {
...
+  optiga_result ret = optiga_get_random(secret, sizeof(secret));
+  if (OPTIGA_SUCCESS != ret) {
+    vcp_println("ERROR optiga_get_random error %d,", ret);
+    return false;
+  }
+
+  // Store pairing secret.
+  ret = optiga_set_data_object(OID_KEY_PAIRING, false, secret, sizeof(secret));
+  /*
+   * TODO: Uncomment. Right now this code will render the device unusable with
```
matejcik, line 693: please uncomment this now, with the modifications in bootloader this should no longer be needed

core/embed/prodtest/main.c
matejcik: agreed, let's bump this to 2kB to be on the safe side.

(do we have stack big enough for this? maybe the big buffers should be marked `static` to go live somewhere other than the stack? or does C take care of this automagically?)

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,446 @@ static void test_otp_write_device_variant(const char *args) {
...
+  if (OPTIGA_SUCCESS != ret) {
+    vcp_println("ERROR optiga_sec_chan_handshake error %d.", ret);
+    return false;
+  }
+
+  return true;
+}
+
+static void optiga_lock(void) {
+  if (!pair_optiga()) {
```
matejcik, line 723: given that this is the only place where `pair_optiga()` is called, will it work together with `drop_privileges()` which happens before it?

or should we add `pair_optiga()` to the main flow after Optiga initialization?

andrewkozlik: The function needs to be able to call `secret_erase()`, `secret_write_header()`, `secret_write()` and `secret_read()`. Is that possible even after `drop_privileges()`? @TychoVrahe

Assuming that the answer is yes, I think that in principle `pair_optiga()` can be called in either place. But I wonder whether it makes sense to call it automatically at every startup.

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+
+  // Verify whether the secret was stored correctly in flash and OPTIGA.
+  memzero(secret, sizeof(secret));
+  if (secret_read(secret, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN) !=
+      sectrue) {
+    vcp_println("ERROR Failed to read pairing secret.");
+    return false;
+  }
+  */
+
```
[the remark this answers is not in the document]
onvej-sl: OK, then this comment should changed accordingly: `// Verify whether the secret was stored correctly in flash and OPTIGA.`

prusnak:
Don't we want to define OPTIG_LOCKED_* in a way so that the hamming distance between the values is bigger than one?

andrewkozlik:
> Don't we want to define OPTIGA_LOCKED_* in a way so that the hamming distance between the values is bigger than one?

We can, but what exactly does that defend against? An attacker in the production facility who executes a voltage-glitching attack to avoid locking the metadata? I think there are more interesting things that such a powerful attacker could do.

core/embed/prodtest/main.c
TychoVrahe: Seems like MPU is still set to allow the secret reading/writing after privileges get dropped. Do we wanna prevent this?

crypto/tls_prf.c
onvej-sl: > However, in general contexts should be treated as opaque handles whenever possible, because the underlying implementation may change.

This can be solved by implementing `hmac_sha256_Duplicate()`.

core/embed/prodtest/main.c
matejcik: Ideally, yes. The main idea about all this is that if there is a vulnerability in prodtest, it should not be possible to exfiltrate the secret through it.

core/embed/prodtest/main.c
andrewkozlik: Done in f92db56fc8319aa573536e39396597e0e99ebd46. The zero-initialization is not needed, it's just a general rule to initialize stack variables with zeros to ensure that in case some sensitive data is left in the freed stack, it gets erased.

core/embed/prodtest/main.c
andrewkozlik: f92db56fc8319aa573536e39396597e0e99ebd46

core/embed/prodtest/main.c
andrewkozlik: f92db56fc8319aa573536e39396597e0e99ebd46

core/embed/prodtest/main.c
andrewkozlik: Done in f92db56fc8319aa573536e39396597e0e99ebd46. But I am not sure how to handle the error state. Should we make sure to drop privileges? How about going to the main loop and responding with `ERROR failed to pair with Optiga` for every command?

core/embed/prodtest/main.c
andrewkozlik: Makes no difference, does it?

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,451 @@ static void test_otp_write_device_variant(const char *args) {
...
+  if (OPTIGA_SUCCESS != ret) {
+    vcp_println("ERROR optiga_set_data error %d for 0xe0e8.", ret);
+    return;
+  }
+
+  // Set data object metadata.
+  static const optiga_metadata_item ACCESS_PAIRED = {
+      (const uint8_t *)"\x20\xE1\x40", 3};
+  static const optiga_metadata_item KEY_USE_SIGN = {(const uint8_t *)"\x10", 1};
+  static const optiga_metadata_item TYPE_PTFBIND = {(const uint8_t *)"\x22", 1};
```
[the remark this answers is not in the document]
andrewkozlik: 7dbd17abb99834baaa3c1866f0a4d908b7090f09

core/embed/prodtest/main.c
andrewkozlik: 7dbd17abb99834baaa3c1866f0a4d908b7090f09

core/embed/trezorhal/optiga/optiga_commands.c
andrewkozlik: e869b104b646d4edbb93c773f4ffd02e653164b9

core/embed/trezorhal/optiga/optiga_commands.c
andrewkozlik: I don't see any use for that, especially given that there is a single Optiga. Even if there were multiple Optigas, it probably wouldn't be a good idea either.

onvej-sl: My opinion is that what I suggest is a better-structured and modular code that is easier to understand, review and maintain.

I think it is worth doing even if there is a single Optiga.

core/embed/prodtest/main.c
onvej-sl: There is no difference in terms of the result, but there can be a difference in terms of the resulting binary code. Since I rarely see such a construction, I am curious if there is any motivation behind it.

crypto/aes/aesccm.c
onvej-sl: Ok, then.

These are the arguments for which I see:
  * As you already said, a similar code is used elsewhere in the codebase. I consider this argument to be a [fallacy](https://en.wikipedia.org/wiki/Appeal_to_tradition). The fact that there is bad code in our codebase does not justify us writing bad code as well.
  * We have no signs that the code doesn't work in terms of correctness of the results.

The arguments against which I see:
  * We have no signs that the optimized code is faster.
  * Generally I think it's better to rely on compiler to optimize a low level code instead of human.
  * The code is more difficult to understand, review and maintain because it uses "non-standard" patterns.
  * It may be my lack of knowledge of the C language but these two patterns specifically raise doubts in me:
    * ```      
        ctx->state32[3] ^= ((uint32_t *)data)[3];
        if (aes_encrypt(ctx->state, ctx->state, ctx->encrypt_ctx) != EXIT_SUCCESS) {
      ```
      At least C 1999 standard says the value of a union member, other than the last one stored, is unspecified.
    * ```ctx->state32[0] ^= ((uint32_t *)data)[0];```
      I don't know whether reinterpretation of `uint8_t*` as `uint32_t*` is defined.

As I said, I'm completely okay if the problematic patterns remain in the code but are encapsulated in a `memxor` function.

core/embed/prodtest/main.c
TychoVrahe: this commit https://github.com/trezor/trezor-firmware/commit/21047b29370fdadd64f24422d9fa3d14b5529a9f disables access to the secret after privileges are dropped. untested though.

andrewkozlik: But what if the pairing fails? It seems better to return some sort of error over the VCP rather than shutting down. Any opinions about what I proposed about going to the main loop?

core/embed/prodtest/main.c
andrewkozlik: No motivation.

crypto/aes/aesccm.c
andrewkozlik: There is no fallacy in my statement. A certain level of optimization was chosen in the existing AES code. Speculation about how that choice was reached is outside of the scope of this PR. So far no proof of anything being objectively "bad" about the existing code has been given. The new code simply respects the style and the chosen level of optimization of the existing code in order to maintain consistency. That doesn't rule out the possibility of refactoring the entire AES implementation at some point in the future if we see fit, but that is also out of scope of this PR.

Regarding the rest, see https://en.wikipedia.org/wiki/Type_punning.

core/embed/prodtest/main.c
TychoVrahe: We can do the VCP response to subsequent commands, or just print an error message on display and shutdown. The former allows non-optiga related part of the test to be run, with some added complexity to the code. Not sure whether the ability to run the test further has any value for the manufacturer

andrewkozlik: I just implemented it a moment ago: https://github.com/trezor/trezor-firmware/pull/3185/commits/0382bf6dcfffd08600362e5fe40a95bf3e190f62

core/embed/prodtest/main.c
```diff
@@ -513,9 +603,443 @@ static void test_otp_write_device_variant(const char *args) {
...
+  uint32_t cpuid[3];
+  cpuid[0] = LL_GetUID_Word0();
+  cpuid[1] = LL_GetUID_Word1();
+  cpuid[2] = LL_GetUID_Word2();
+
+  vcp_print("OK ");
+  vcp_println_hex((uint8_t *)cpuid, sizeof(cpuid));
+}
+
+#ifdef USE_OPTIGA
```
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3185/commits/015133d971d17cc4386a14d6c6513ab471638fab

core/embed/prodtest/main.c
matejcik: @onvej-sl please resolve this if you're OK with the current state

core/embed/prodtest/main.c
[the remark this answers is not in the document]
onvej-sl: Perfect, thank you.

core/embed/prodtest/main.c
onvej-sl: Why do you need to check if the slots are locked in the `VARIANT` command? What informational value does it have for you if you don't know if the slots are configured correctly? Is that just a check to verify that the `LOCK` command was called before?

andrewkozlik: > Why do you need to check if the slots are locked in the `VARIANT` command?

@matejcik or @Hannsek can probably provide more insight into the production process and rationale behind this.

> What informational value does it have for you if you don't know if the slots are configured correctly? Is that just a check to verify that the `LOCK` command was called before?

Yes.

matejcik: > Why do you need to check if the slots are locked in the `VARIANT` command? What informational value does it have for you if you don't know if the slots are configured correctly? Is that just a check to verify that the `LOCK` command was called before?

`LOCK` is called in the first phase of manufacturing setup. We assume that it checks its own success, i.e., that there is no need for a separate "verify metadata" step -- there is no service the host provides just by "asking again".

`VARIANT` is called in a subsequent phase by a different person on a different station at a different time. At that point, we need to ensure that a device has already gone through the first phase, which we detect by checking that `LOCK` has already been called (as the last step of phase 1)

core/embed/prodtest/main.c
```diff
@@ -528,15 +541,28 @@ int main(void) {
...
 #endif
   usb_init_all();
 
+#ifdef USE_OPTIGA
+  optiga_init();
+  optiga_open_application();
+  pair_optiga();
+#endif
+
+  display_reinit();
```
matejcik, line 561: do we need the second `display_reinit()`, is it a leftover, or a rebase glitch? @TychoVrahe

matejcik:
force-pushed & rebased on master. 
@andrewkozlik and @TychoVrahe please do a quick pass of `main.c` and surrounding environment to make sure that the rebase didn't mess up something important. the fact that the Optiga implementations were spun off to their own file really helps a lot because there were no conflicts in those :)

core/embed/prodtest/main.c
TychoVrahe: leftover or rebase glitch, definitely not needed.

matejcik: removed & force pushed

matejcik (approved):

PR #3186, opened 2023-07-28
Integrate OPTIGA Trust M

andrewkozlik:
Further work in this PR:
- <s>I am probably doing the initialization of the I2C interface wrong. There has been some refactoring recently around I2C, so that's not reflected in this PR. @TychoVrahe </s> (_Edit: Done_)
- <s>Maybe move the code to a subdirectory in `core/embed/trezorhal` such as `se` or `optiga`.</s> (_Edit: Done_)

Further work, that can be split into separate issues/PRs:
- <s>Not sure why I can't use the full capacity of the data register. See `OPTIGA_DATA_REG_LEN` in `transport.h`.  @TychoVrahe </s> (_Edit: Solved_)
- We will need to agree on how to integrate this with the emulator. My plan is to add a new file `optiga.h`, which will implement some higher level operations, such as `optiga_get_device_cert()` or `optiga_check_pin()`. Only these would be used by the firmware directly and only they would need to be implemented for the emulator.
- Improve management of timeouts. Right now there is just a for-loop with some arbitrary maximum retry count.
- Improve error handling in general, e.g. retransmission in case of CRC failures.
- <s>Add the files to the build process only for the right firmwares. How will we determine that? Using `TREZOR_MODEL`, or something else?</s> (_Edit: Solved_)

### DIFF
--- a/core/embed/trezorhal/optiga/optiga_transport.c
+++ b/core/embed/trezorhal/optiga/optiga_transport.c
@@ -1,0 +1,506 @@
+/*
+ * This file is part of the Trezor project, https://trezor.io/
+ *
+ * Copyright (c) SatoshiLabs
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Reference manual:
+ * https://github.com/Infineon/optiga-trust-m/blob/develop/documents/Infineon_I2C_Protocol_v2.03.pdf
+ */
+
+#include "optiga_transport.h"
+#include <string.h>
+#include "common.h"
+#include "i2c.h"
+#include "optiga_hal.h"
+#include TREZOR_BOARD
+
+// Maximum possible packet size that can be transmitted.
+#define OPTIGA_MAX_PACKET_SIZE (OPTIGA_DATA_REG_LEN - 5)
+
+// Register address of the data register.
+static const uint8_t REG_DATA = 0x80;
+
+// Register address of the register holding the maximum data register length (2
+// bytes).
+static const uint8_t REG_DATA_LEN = 0x81;
+
+// Register address of the I2C state register (4 bytes).
+static const uint8_t REG_I2C_STATE = 0x82;
+
+// Register address of the register that triggers a warm device reset.
+static const uint8_t REG_SOFT_RESET = 0x88;
+
+// I2C state register - Device is busy executing a command.
+static const uint8_t I2C_STATE_BYTE1_BUSY = 0x80;
+
+// I2C state register - Device is ready to return a response.
+static const uint8_t I2C_STATE_BYTE1_RESP_RDY = 0x40;
+
+// I2C base address of Optiga.
+static const uint8_t BASE_ADDR = 0x30;
+
+// Address of Optiga use by our HAL which requires it to be shifted by one bit.
+static const uint16_t OPTIGA_ADDRESS = (BASE_ADDR << 1);
+
+// Constants for our I2C HAL.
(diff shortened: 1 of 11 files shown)


TychoVrahe:
Added a commit https://github.com/trezor/trezor-firmware/commit/3c3dec0b617c00a2dc87494b375f6a0e2f79ed0f

It handles
- the I2C initialization, which is done in the MCU-specific i2c driver, so its removed from optiga impl. We can probably simplify a bit by 
- moves the optiga code to subdirectory as proposed (trezorhal/optiga), these files should now be(or at least aim to be) independent on underlying MCU so they stand above
- the files are added to build for T2B1, revision 10, firmware for now
- adds `optiga_hal.c` file, which initializes the reset pin and exposes hw reset funciton. 

Tested the `optiga_open_application` command on T2B1 HW, returns success.

Agree with adding single interface file `optiga.h` and exposing only higher level operations if emulator support is required.

TychoVrahe:
As for the OPTIGA_DATA_REG_LEN limitation: it seems to me that the I2C is timeouting. e.g if you send 225 bytes, thats 1800 bits and with 200kHz i2c that is some 9ms. Timeout in optiga_transport is set to 10ms. So with some more bytes to send this should timeout. Try increasing the timeout. We could alternatively increase the I2C speed to 400kHz (which is the limit), just be aware that we don't want this speed for every i2c, specifically touch layer comm, and the driver is shared.

andrewkozlik:
> As for the OPTIGA_DATA_REG_LEN limitation: it seems to me that the I2C is timeouting.

Thanks! That was the problem. I increased the timeout to 15 ms and it works fine now. https://github.com/trezor/trezor-firmware/pull/3186/commits/dbca703eaf35394f3a7c787c61381c3fadd72c35

core/embed/trezorhal/optiga/optiga_transport.c
```diff
@@ -0,0 +1,497 @@
...
+  if (presentation_layer) {
+    pctr |= PCTR_PRESENTATION_LAYER;
+  }
+
+  // Transmit command packets to OPTIGA.
+  uint8_t chain = PCTR_CHAIN_NONE;
+  do {
+    size_t packet_data_size = 0;
+    if (command_size > OPTIGA_MAX_PACKET_SIZE - 1) {
+      packet_data_size = OPTIGA_MAX_PACKET_SIZE - 1;
```
hiviah, line 421: The `-1` in `OPTIGA_MAX_PACKET_SIZE - 1` seems weird, like either the max packet size is off or there was one terminator byte expected.
The `optiga_send_packet` does not use `OPTIGA_MAX_PACKET_SIZE` for size check but computes it from `packet_data_size > sizeof(frame_buffer) - 7` which makes it hard to read if the checks are equivalent and correct.

andrewkozlik: The first byte of the packet is the packet control byte `pctr`, so each packet contains at most `OPTIGA_MAX_PACKET_SIZE - 1` bytes of data. I can add a code comment to explain that or define another constant  `OPTIGA_MAX_PACKET_DATA_SIZE = OPTIGA_MAX_PACKET_SIZE - 1`. But since this is the only place where it would be used, I think the code comment would be better, so that the names don't get confused.

andrewkozlik: The reason why `optiga_send_packet()` compares the `packet_data_size` with the size of the `frame_buffer` is because it will be writing the data to the `frame_buffer`. If someone were to accidentally change the `frame_buffer` size to something smaller, then this runtime check would catch it, but comparing against `OPTIGA_MAX_PACKET_SIZE` wouldn't catch it.

hiviah: Comment would be OK. Originally I thought that the frame size is given by the protocol and can't be changed.
We could make a compile-time assert (needs changing `gnu99` to `gnu11` in `SConscript.firmware`) to make sure the size relations are correct to avoid wrong change, e.g.:

```C
#include <assert.h>
// after macro and variable definitions:
static_assert(sizeof(frame_buffer) - 7 == OPTIGA_MAX_PACKET_SIZE-1,
        "incorrect frame size vs data max size constants relation" );
```

andrewkozlik: In other parts of our codebase we use `_Static_assert()` and it seems to work fine.
https://github.com/trezor/trezor-firmware/pull/3186/commits/ac3e75470f77191a20e1b18a2694438cd708abaa https://github.com/trezor/trezor-firmware/pull/3186/commits/51ebf2b820ad99ffe4ed7c88541ff2c75776b3f8

hiviah (approved):
Awesome.

PR #3194, opened 2023-08-07
Change core builds to use GNU11 standard instead of GNU99

hiviah:
Change build to use GNU11 standard instead of very old GNU99. Allows better compile time checks.

C and Rust binaries like boardloader, bootloader, prodtest are bit-by-bit identical. Firmware has same size, but I needed to check with binary diffing and testing to make sure compiler has not generating something weird.


TychoVrahe (approved):

Hannsek:
@hiviah Can we merge this?

Issue #3178, opened 2023-07-27
Fonts in confirm_properties


Hannsek:
Change the font in confirm_properties to:

"normal" for description
**demibold** for human-readable text
`mono` for binary data

PR #3195, opened 2023-08-08
chore(core/rust): change fonts in confirm_properties for TT

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3178:
- making the key font NORMAL and the value DEMIBOLD (or MONO in case of data)


Hannsek:
I'd put monofont in all values. No?

image
image
image
image

grdddj:
> I'd put monofont in all values. No?
> image image image image

You mean all these 4 screens should use mono, or all the values in general? Both should be possible

Hannsek:
All values in general.

grdddj:
> All values in general.
> 
> I don't see any blocker to this.

Right, so I will change these and let's see the results

matejcik:
> All values in general.
> 
> I don't see any blocker to this.

do all "normal" flow screens also use mono for values? because we should be consistent here. 
three separate screens like in eth flow (max fee / max priority fee / gas limit) should look mostly the same as if the same three screens are part of `confirm_properties`

Hannsek:
I think we do not use it on many other places.

See btc send flow: 
image
or receiving flow:
image

Hannsek (approved):
LGTM :)

matejcik:
> I think we do not use it on many other places.

I distinctly recall you telling me that font in this screen is _not_ mono. I have screenshots to prove it ;)
But it looks like it _is_ mono, in which case yes, let's also put mono in `confirm_properties`.

(and if not, let's put **the same font that is used here** to `confirm_properties` also)

Hannsek:
The value field in provided screens, e.g. "Hidden wallet #3" has this font: Roboto Mono, Medium, 20px

grdddj:
List of "new" screens - https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4825080681/artifacts/test_ui_report/master_diff.html

What is the conclusion - do we want it like this? (they are all unimportant screens, so I am for it)

Hannsek:
Use mono. Please merge it. :)

bosomt:
**QA OK**

 - Device: Trezor T2T1 2.6.1 regular (revision 1b08be018603af91969d4072d0e0eafef293d5fb)

Issue #3187, opened 2023-07-31
Incorrect ShowDeviceTutorial response when cancelled 


dahaca:
**Describe the bug**
The `ShowDeviceTutorial` method responds with `success: true` when tutorial is canceled on device, which impacts Suite UI.

**Firmware version and revision**
2.6.1

**Expected behavior**
It responds with `success: false` when canceled on device

grdddj:
Quick fix done in https://github.com/trezor/trezor-firmware/pull/3198 - @dahaca please check that this solves the problem in `Suite`

PR #3198, opened 2023-08-09
fix(core): raise ActionCancelled when cancelling tutorial flow for TR

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3187:
- differentiates the case when the user cancels the tutorial, in comparison to successfully finishing it


matejcik (approved):
utACK

Issue #3176, opened 2023-07-26
Different spaces on left and right side of the highlighted letter/word/number


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/3b208d84-cd22-417b-894a-d94fd6d1c47c)

PR #3200, opened 2023-08-09
fix(core/rust): do not increase margin of right choice items

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3176:
- unify left and right distance between center and neighbouring items


Hannsek:
utACK

Hannsek (approved):

matejcik (approved):
utACK

PR #3201, opened 2023-08-09
feat(tests): notify about failed tests when updating fixtures from CI

grdddj:
Makes `tests/update_fixtures.py` more robust.

When there is a failure (not just UI error) in the test suite, do not update the `fixtures.json` with the results of these tests, notify the user and fail the update script.


tests/gitlab.py
```diff
@@ -31,6 +33,39 @@
...
+        result_line = re.search(result_pattern, raw_content)
+        if not result_line:
+            print("No results yet.")
+            return TestResult()
+        return cls.from_line(result_line.group(0))
+
+    @staticmethod
+    def _link_to_report(job_id: str) -> str:
+        return (
+            f"https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/{job_id}/raw"
```
matejcik, line 65: nit: just make it a constant and run `LINK_TO_REPORT.format(job_id)`

matejcik (approved):

tests/gitlab.py
grdddj: Right, changed in the force-push

PR #3202, opened 2023-08-09
feat(trezor-client): Add Sync and Send to Transport

wszdexdrf:
This is needed for async-hwi, where the struct Trezor is passed to async functions.


wszdexdrf:
Does this need to be in a minor update bump to 0.1.2 as well?

matejcik (approved):
utACK. this would seem to indicate that both UDP and WebUSB connections themselves are already Send + Sync, so this should be ok

prusnak:
https://crates.io/crates/trezor-client 0.1.2 released

PR #3205, opened 2023-08-11
Noninteractive jump from firmware to bootloader, copying FW image header to predefined location

hiviah:
This is linked to issue #2919 .

It implements part where firmware may call `svc_reboot_copy_image_header(const uint8_t *image_address)` and firmware image header will be copied to bootloader area, jumps directly back to bootloader where bootloader may pick up the firmware header and start installation (it is recommended to put the check of the image header somewhere at the beginning of bootloader's `main()`).


hiviah:
Hm no idea why core build fails, it works on my environment and both BTC-only should fail at the same error then, but doesn't.

matejcik:
it is boardloader that is failing. did you try that?

matejcik:
I suspect boardloader shouldn't build any of these functions anyway.

hiviah:
Ah yes, I looked at the line above and thought it was firmware it was linking.

OK that is one line fix, it's just linker symbol.

hiviah:
Adding it to linker script fixes it. Maybe I should decouple the now different versions of shutdown since it all points to a single shared `util.s` and make either `#ifdef` (but then it becomes `util.S`, without it assembler doesn't recognize macros) or `beq` on `sectrue` on parameter or something like that.

Anyway, if anyone knew how to add DWARF debug symbols to assembler functions (I need to define how deep stack is), otherwise unwind in debugger doesn't work simply because our calling condition doesn't use one register as FP, so without DWARF debugger is pretty blind when it comes to unwinding call stack if there's some assembler involved.

hiviah:
OK everything looks fine also in CI, squashing the commits and it can be finalized for preview.

TychoVrahe:
I am somewhat unhappy about the `util.s` split: there is some 90% of the code duplicated. Maybe we could keep the common functions in util.s and create new file with the different ones? or use the util.S and macros idea

hiviah:
> I am somewhat unhappy about the util.s split: there is some 90% of the code duplicated

Yes, I am unhappy about it too, `#ifdefs`  makes it too complicated at this point (without `#ifdef` you need `BEQ` on another parameter which shouldn't be in boardloader linked script). 

I tried and weighted the aspect of both. The final result, the point was to never make accidental mistake like what happened to T1 (shared this part), which was later revealed to contain exploit due to change (which was made exactly this way). We will most likely never change these for boardloader, prodtest and so on, only FW and bootloader hacks need the separate jumps and register tricks. Who ever checks boardloader? Me and you @TychoVrahe usually, otherwise almost nobody (they definitely do not disassemble it to see what is really in there).

This way it "freezes" the limited `utils` function "forever" so that we don't have to worry about changing boardloader, flashing it, write-protecting it and then finding out it is broken (if we find out soon enough, it will just be filling Vltava :smile: with a batch of bricks, otherwise we can't even recall&reflash&repack).

Any better solution would be appreciated. I weighted is as lesser evil to have few assembler functions to be copy-pasta and freezed, than to later find out that changing `util.s` is breaking the device completely in boardloader, which can't be changed (unlike T1 which has broken write protection). This solution is not pretty by far, but I think the macros would be worse.

core/embed/trezorhal/stm32f4/limited_util.s
```diff
@@ -0,0 +1,159 @@
...
+  ldr r0, =SCB_VTOR
+  str lr, [r0]
+  mov r0, r1            // zero out r0
+  // go on to the next stage
+  ldr lr, [lr, 4]       // set lr to the next stage's reset_handler
+  bx lr
+
+  .global jump_to_unprivileged
+  .type jump_to_unprivileged, STT_FUNC
+jump_to_unprivileged:
```
TychoVrahe, line 71: lets remove the `jump_to_unprivileged` from limited utils, its not used and a pointless copy here. we will only have duplicated `memset_reg` and `shutdown_privileged` which is ok-ish

core/embed/trezorhal/stm32f4/supervise.h
```diff
@@ -58,10 +60,22 @@ static inline void svc_shutdown(void) {
...
   }
 }
+
+static inline void svc_reboot_copy_image_header(const uint8_t *image_address) {
+  if (is_mode_unprivileged() && !is_mode_handler()) {
+    register const uint8_t *r0 __asm__("r0") = image_address;
+    __asm__ __volatile__("svc %0" ::"i"(SVC_REBOOT_COPY_IMAGE_HEADER), "r"(r0)
+                         : "memory");
+  } else {
+    copy_image_header_for_bootloader(image_address);
```
TychoVrahe, line 78: i know this does not originate in this PR: but perhaps we should call `ensure_compatible_settings` here too? i don't think that the else branch will be called anywhere today, but lets avoid the potential problem in future (also valid for the `svc_reboot_to_bootloader` function above). Also `mpu_config_bootloader` is missing here in comparison with the SVC call

core/embed/boardloader/.changelog.d/3205.added
```diff
@@ -0,0 +1,2 @@
+Added firmware update without interaction.
+Split builds of different parts to use simle util.s assembler, while FW+bootloader use interconnected ones.
```
TychoVrahe, line 2: typo `simple`

core/embed/bootloader/.changelog.d/3205.added
```diff
@@ -0,0 +1,2 @@
+Added firmware update without interaction.
+Split builds of different parts to use simle util.s assembler, while FW+bootloader use interconnected ones.
```
TychoVrahe, line 2: typo `simple`

core/embed/bootloader_ci/.changelog.d/3205.added
```diff
@@ -0,0 +1,2 @@
+Added firmware update without interaction.
+Split builds of different parts to use simle util.s assembler, while FW+bootloader use interconnected ones.
```
TychoVrahe, line 2: typo `simple`

core/embed/bootloader_ci/memory.ld
```diff
@@ -23,6 +23,9 @@ ccmram_end = ORIGIN(CCMRAM) + LENGTH(CCMRAM);
 sram_start = ORIGIN(SRAM);
 sram_end = ORIGIN(SRAM) + LENGTH(SRAM);
 
+/* IMAGE_HEADER_SIZE is 0x400, this is for interaction-less firmware update start */
+firmware_header_start = ccmram_end - 0x400;
```
TychoVrahe, line 27: `bootloader_ci` still cleans the header on startup. not sure we need the functionality there though

TychoVrahe (commented):
When jumping to bootloader without the image header copied, maybe we should add explicit cleaning of the header area to avoid any potential problems?

Also when jumping from bootloader to firmware, the cleaning is only done in firmware startup.s. So maybe custom firmware might remove the cleaning and see what was there, but that is probably not a problem?

core/embed/trezorhal/stm32f4/limited_util.s
hiviah: Fixed.

core/embed/trezorhal/stm32f4/supervise.h
hiviah: Fixed as we discussed.

core/embed/boardloader/.changelog.d/3205.added
hiviah: Fixed.

core/embed/bootloader/.changelog.d/3205.added
hiviah: Fixed.

core/embed/bootloader_ci/.changelog.d/3205.added
hiviah: Fixed.

core/embed/bootloader_ci/memory.ld
hiviah: Fixed as discussed - `bootloader_ci` may be used also to test this non-interactive update now.

hiviah:
So I think I fixed the stuff that needed fixing.

TychoVrahe (approved):
lgtm. please squash the commits and merge

matejcik:
please do not merge until we bump bootloader version (and let's not bump bootloader version until we are sure what is going to end up in the current one)

Issue #991, opened 2020-05-14
Replace all English strings with identifiers and collect them


tsusanka:
We need to replace all UI texts with identifiers. Firmware will then have no language and we will load the correct strings from the data blob (as defined in #988).

We need to collect all the English strings into a single file. This file will be then fed into [Crowdin](https://crowdin.com/) for the actual translating process.

We are currently considering two options:

1. Extract the strings using libcst/gettext, see @matejcik's comment [here](https://github.com/trezor/trezor-firmware/issues/989#issuecomment-627883458).
2. Collect the strings using device tests. Similarly to UI tests, we would run device tests against the emulator and gather all the strings that are displayed. Each time something is printed on the display we would store the string in some local json/xml file. Later we would concatenate all those files into a single one.

Both should be done automatically on CI.

---

Part of #987.

prusnak:
I think this task is not really blocked by fully implementing #989. We can collect all localized strings to a separate file (e.g. `i18n.py`) before the blob mechanism is in place.

matejcik:
Agreed. We need i18n.py anyway, the emulator-running-from-sources must be able to load strings without using a blob.

tsusanka:
> We can collect all localized strings to a separate file (e.g. i18n.py) before the blob mechanism is in place.

Let's do this first!

tsusanka:
I have merged #989 into this one.

PR #3206, opened 2023-08-11
Firmware translations

grdddj:
Creates a proof of concept of translating firmware into various languages.

Currently includes `english` and `czech` translations (done by `ChatGPT)`, mainly for `model R`.

Build process
- `TREZOR_LANG=cs` will set the language for Czech both in `micropython` and `Rust`

Translations infrastructure:
- translations are defined in `cs.json`, in individual sections connected to their context
- `cs.{py,rs}.mako` takes data from `JSON` and creates a {py,rs} file with all the translation symbols as constant strings variables/attributes (so we have a static check that all strings are defined)
- layouts import the translation module/object and access some attributes on it

TODO:
- make sure the strings are not stored in memory but in flash
- check the flash size consumption (experiment with `micropython::const` and `rust::trait` approaches)
- decide whether store all the languages inside one binary (and have messages to change the language), or not allowing the change
- include font characters for czech language
- handle plural forms
- solve `micropython` vs `Rust` division - there should be only one place of defining the translations mappings
- solve differences between `TT` and `TR`
- do the translation in `C` as well (there are certainly some strings, like error messages)
- do the translation in `bootloader`?
- make sure all the languages contain the same keys (done on `Rust` side already)

Fixes https://github.com/trezor/trezor-firmware/issues/991


core/SConscript.firmware
```diff
@@ -687,6 +690,13 @@ def cargo_build():
     if PYOPT == '0':
         features.append('ui_debug')
 
+    if TREZOR_LANG == 'en':
+        features.append('lang_en')
+    elif TREZOR_LANG == 'cs':
+        features.append('lang_cs')
```
prusnak, line 696:
```suggestion
    if TREZOR_LANG in ['en', 'cs']:
        features.append(f'lang_{TREZOR_LANG}')
```

core/SConscript.unix
```diff
@@ -761,6 +765,13 @@ def cargo_build():
     if DMA2D:
         features.append('dma2d')
 
+    if TREZOR_LANG == 'en':
+        features.append('lang_en')
+    elif TREZOR_LANG == 'cs':
+        features.append('lang_cs')
```
prusnak, line 771:
```suggestion
    if TREZOR_LANG in ['en', 'cs']:
        features.append(f'lang_{TREZOR_LANG}')
```

core/embed/lib/fonts/fonts.c
```diff
@@ -18,33 +18,36 @@
  */
 
 #include "fonts.h"
+#include <stdio.h>
```
prusnak, line 21: please don't include stdio - why is it needed?

core/embed/lib/fonts/fonts.c
```diff
@@ -129,32 +132,108 @@ int font_baseline(int font) {
...
+    c_2bytes = utf8_mapping(c_2bytes);
+    if (!c_2bytes) {
+      is_printable = false;
+    }
+  }
 
-  // printable ASCII character
-  if (c >= ' ' && c < 0x7F) {
+  // printable character
+  if (is_printable && c_2bytes >= ' ') {
```
prusnak, line 216: This check is not enough - we should also check that c_2bytes <= MAX_FONT_GLYPH_INDEX. Currently MAX_FONT_GLYPH_INDEX == 156 (used to be 126). Otherwise we introduce read out of bounds below when accessing `FONT_xxx_DATA`.

prusnak (changes requested):
Good work! Left some suggestions mainly for the C part.

prusnak:
I'd prefer if `convert_char_utf8(const uint8_t c)` returned uint32_t - so we'd support full UTF-8. Of course, we will never support for 4 billions glyphs, but at least the UTF-8 support would be complete.

core/SConscript.unix
grdddj: Done in [8ad8475](https://github.com/trezor/trezor-firmware/pull/3206/commits/8ad8475eea41f7b66cbcd2b95f969ed0677a65ec)

core/embed/lib/fonts/fonts.c
grdddj: Deleted in [8ad8475](https://github.com/trezor/trezor-firmware/pull/3206/commits/8ad8475eea41f7b66cbcd2b95f969ed0677a65ec) - I used it so I can do `printf`

core/embed/lib/fonts/fonts.c
grdddj: ` && c_2bytes <= 156` added in [8ad8475](https://github.com/trezor/trezor-firmware/pull/3206/commits/8ad8475eea41f7b66cbcd2b95f969ed0677a65ec)

Good suggestion for the future, noted it as TODO

grdddj:
> I'd prefer if `convert_char_utf8(const uint8_t c)` returned uint32_t - so we'd support full UTF-8. Of course, we will never support for 4 billions glyphs, but at least the UTF-8 support would be complete.

Thanks for the feedback, I noted this down.

core/embed/rust/src/ui/display/font.rs
```diff
@@ -144,6 +144,36 @@ impl Font {
...
+    }
+
+    /// Supports UTF8 characters
+    fn get_last_glyph_from_text(self, text: &str) -> Option<Glyph> {
+        if let Some(c) = text.chars().next_back() {
+            self.get_glyph_from_char(c)
+        } else {
+            None
+        }
+    }
```
matejcik, line 175: this is completely weird. why not modify `get_glyph` to accept a `char` directly? (and then as @prusnak suggests, modify `get_char_glyph` to accept uint32)

grdddj:
Recent changes include:
- moving all the translation data into `Rust` - so we have only one `{en,cs}.json` file
- got rid of the `.py` files, which were taking quite some space - currently `micropython` requests translations via `trezortranslate::tr` function exported from `Rust`
  - we have static type-check that we only request keys that really exist
  - (I have failed to export some dict-like structure from `Rust` (did not even manage to return a static string), so went for the function approach, where we can return strings dynamically)
 
SIze considerations:
- it is currently around `70kb` bigger than the state before translations - which is very unfortunate and should be improved
- to my surprise, the current state with everything in `Rust` is about the same size as with majority of translations in `micropython`
  - it is probably because `micropython` still needs to define all the strings as keys to the translate function - and they are quite long (not sure if shortening them would take less space?)
-  `TranslationsGeneral::get_text()` takes around `20kb`, it is currently a big `match` statement, maybe it can be improved

The biggest question currently:
- how do we allow for changing the language in a built binary without rebuilding and without the binary containing all the languages?

common/protob/messages-management.proto
```diff
@@ -179,6 +179,37 @@ message ApplySettings {
...
+    required uint32 data_length = 2;   // byte length of the whole translation blob (set to 0 for default language)
+}
+
+/**
+ * Response: Device asks for more data from transaction payload.
+ * @end
+ * @next TranslationDataAck
+ */
+ message TranslationDataRequest {
+    optional uint32 data_length = 1;    // Number of bytes being requested (<= 1024)
```
matejcik, line 201:
```suggestion
    required uint32 data_length = 1;    // Number of bytes being requested (<= 1024)
    required uint32 data_offset = 2;
```

grdddj:
Transforming into regular PR, so it is ready for review.

There is still quite some work, but I think it makes sense to stop now and analyze the status and create clear requirements for what should be done.

Current high-level status:
- contains translations into czech and french, both done by ChatGPT, so the quality is rather poor
- translations are centralized in JSON files - en.json, cs.json, fr.json - in `core/embed/rust/src/ui/translations`
- english translations are embedded/hardcoded in the code, they will be there all the time, acting as a backup/default
- foreign translations are stored in two `16 kb` sectors, which is enough right now for the translations data
- fonts/glyphs are currently hardcoded into the firmware, so both czech/french fonts are there all the time
- `trezorct`l is responsible for generating the translation data blob (will be replaced by a custom tool)
- the structure of the blob is a `256-byte` header with metadata and then the translations data delimited by `0x00` byte
- the translations data has around `22kB` for each language, the extra fonts for czech and french have around `12 kB` each (on TT, TR is like `2 kB`)
- on TT hardware, it takes at most `2.5 ms` to load the translation (even with the "inefficient" delimiter storing)
- it increases the firmware size quite a lot, on TT with both the czech/french font, the firmware is `90 kB` bigger - mostly because of storing all the translations keys

Some things/considerations/questions that are missing and should be prioritized:
- we might want to include offset table into the translations blob (it would increase the size by around `1 kB`)
- the translations are not being signed/verified
- we do not yet have logic to append new translations to the existing ones (they should be always at the end)
- firmware update process is not fully accounted for, it should make sure the translations are up-to-date with firmware (tell user that they should update the translations as well)
- it might make sense to not translate altcoins at all, to save space and a lot of translating work (and confusion for users)
- we might want to load fonts dynamically with each language
- with bigger glyph sets, like Azbuka, we might have just one font, instead of four/five
- `TT` has one extra 16kb sector (occupied by secret on `TR`), which might be used specifically for the font data (fonts on TR are much smaller, they may fit together with the translations)
- in TR, how is it with the last `16 kb` sector? could we safely use it? (also on `TT`)
- it would be nice to use some data compression (for both translations and font data), but we have a limitation on the Rust side of no-std libraries only and no dynamic memory allocation ... also, it would mean some runtime CPU and RAM overhead

grdddj:
Force-pushed the latest version, rebased on `main` and with all the internal changes from @matejcik

grdddj:
> I went through the UI diff and found some screens that are not ideal:
> 
>     * unnecessary space at the beginning of the row:
> 
> 
> image


Fixed in [ba79d8d](https://github.com/trezor/trezor-firmware/pull/3206/commits/ba79d8d60e4b312d4b476a1c9a484cb892738faa)


> 
>     * why the font was changed here?
> 
> 
> image


Because of translations, previously it was possible to easily divide the english sentence into three parts and bolden just one --- with multiple foreign languages, it is harder, we do not know where to split it, so it all became just one sentence


> 
>     * obsolete "please wait" (on multiple places)
> 
> 
> image

That was actually a bug it seems, the text was there, but was covered by the loader.

Removed all the implicit `PLEASE WAIT` texts in [dca3856](https://github.com/trezor/trezor-firmware/pull/3206/commits/dca3856841fa806a02406f672fbabbdeb8f08b96) --- note that it also changed some other screens --- see the diff

> 
>     * please revert the change so the verticle space between the two strings are still there:
> 
> 
> image

That is not so easily possible, I have deleted the newline so all the other non-english languages will fit there more easily. Having newline just for english is not viable here.

> 
>     * why the copy change here?
> 
> 
> image

It was caused by unification of `TR` and `TT` - they had different copies there, so I chose the `TT`'s one in this case --- we can decide which one to use (but it should be the same for both)

> 
>     * missing questionmark
> 
> 
> image

Same as above - after unifying the text with `TT` I have picked the one without question-mark ... but we can choose which one we want

Hannsek:
> > * please revert the change so the verticle space between the two strings are still there:
> > image
> 
> That is not so easily possible, I have deleted the newline so all the other non-english languages will fit there more easily. Having newline just for english is not viable here.

The vspace really should be there as it looks really strange without one. I think that translations can be on two pages… 


> > * why the copy change here?
> > image
> 
> It was caused by unification of TR and TT - they had different copies there, so I chose the TT's one in this case --- we can decide which one to use (but it should be the same for both)

Please use TR's copy as it is the newer one.

Hannsek:
FW installation with blobs updating:
image

change language in Suite settings

image

core/embed/rust/src/translations/blob.rs
```diff
@@ -0,0 +1,381 @@
...
+            merkle_proof,
+            signature,
+            total_len: container_length + Self::BLOB_MAGIC.len() + mem::size_of::<u16>(),
+        };
+        new.verify()?;
+        Ok((new, reader))
+    }
+
+    fn verify_with_keys(&self, public_keys: &[ed25519::PublicKey]) -> Result<(), Error> {
+        let merkle_root = merkle_root(self.header_bytes, self.merkle_proof);
```
mmilata, line 363: Why do we even use a Merkle tree? Because producing a signature is expensive so we just create one for a batch of blobs?

matejcik: "expensive", yes, as in, this is also going to be signed manually so we don't want to overwhelm the signers

core/embed/rust/src/strutil.rs
```diff
@@ -23,9 +32,15 @@ impl SkipPrefix for &str {
 /// - create a new string by skipping some number of bytes (SkipPrefix) - used
 ///   when rendering continuations of long strings
 /// - create a new string from a string literal (From<&'static str>)
-pub trait StringType: AsRef<str> + From<&'static str> + SkipPrefix {}
+pub trait StringType:
```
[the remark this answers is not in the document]
matejcik: ideally yes, but there's a lot more work in that direction

core/embed/rust/src/translations/blob.rs
```diff
@@ -0,0 +1,381 @@
...
+        let magic = reader.read(Self::BLOB_MAGIC.len())?;
+        if magic != Self::BLOB_MAGIC {
+            return Err(INVALID_TRANSLATIONS_BLOB);
+        }
+
+        // read length of contained data
+        let container_length = reader.read_u16_le()? as usize;
+        // continue working on the contained data (i.e., read beyond the bounds of
+        // container_length will result in EOF).
+        let mut reader = reader.read_stream(container_length.min(reader.remaining()))?;
```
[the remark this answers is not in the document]
matejcik: yes, that's the flash case. the incoming slice is the full flash area, containing the translation blob (`container_length`) and a lot of trailing empty bytes (`reader.remaining()`)

core/embed/rust/src/translations/blob.rs
```diff
@@ -0,0 +1,381 @@
...
+    offsets: &'a [OffsetEntry],
+    data: &'a [u8],
+}
+
+fn validate_offset_table(
+    data_len: usize,
+    mut iter: impl Iterator<Item = u16>,
+) -> Result<(), Error> {
+    // every offset table must have at least the sentinel
+    let mut prev = iter.next().ok_or(INVALID_TRANSLATIONS_BLOB)?;
```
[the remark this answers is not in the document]
matejcik: good point leading to a better design: let's consider the table to be a list of _ends_, not _starts_. that way the table is trivially N items long, the last item is trivially the sentinel (and there is no special sentinel id)

core/embed/rust/src/translations/blob.rs
```diff
@@ -0,0 +1,381 @@
...
+    for next in iter {
+        // offsets must be in ascending order
+        if prev > next {
```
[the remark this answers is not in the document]
matejcik: we need to for the translation strings to work. we never remove an index (so that an older firmware looking at a newer blob sees the indices in the right places) so any removed strings will be stored as zero-length strings

grdddj:
Force-pushed after fresh rebase on `main`. I will not be doing any commits this week - feel free to completely take over the PR @matejcik 

CI hasn't finished yet but looks alright - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines/1168162478

- even the non-english tests are running now (and slowing down the entire CI), as I flagged them not only `nightly`, but also to run for all branches containing `translations`

core/embed/rust/src/translations/blob.rs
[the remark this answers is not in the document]
matejcik: actually no, I take that back. implementation-wise it's much nicer that both the start and end entries are there, and it's cheaper by one if-branch per lookup

core/embed/rust/src/translations/blob.rs
```diff
@@ -0,0 +1,381 @@
...
+        ) {
+            return Err(INVALID_TRANSLATIONS_BLOB);
+        }
+        Ok(())
+    }
+
+    pub fn get(&self, id: u16) -> Option<&'a [u8]> {
+        self.offsets
+            .iter()
+            .position(|it| it.id == id)
```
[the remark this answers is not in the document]
matejcik: 3ff3b8c6346be806db3ef88226e469e038761e8c

core/embed/rust/src/translations/blob.rs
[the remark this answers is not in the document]
matejcik: 3ff3b8c6346be806db3ef88226e469e038761e8c

grdddj:
@prusnak can you please check the requested changes? We want to merge it today or tomorrow and it is blocked by them

mmilata (approved):
rust side ACK

prusnak (approved):
changes from my earlier review were done

bosomt:
**QA OK**

tested on TT and S3 
tested reverting to en-US too

```
❯ ls -laF | grep 2.7.0
-rw-r--r--@   1 bosomt  staff     1516544 Mar  3 13:45 firmware-T2B1-2.7.0-1f67c1e.bin
-rw-r--r--@   1 bosomt  staff     1516544 Mar  7 19:22 firmware-T2B1-2.7.0-45e8a842a-signed.bin
-rw-r--r--@   1 bosomt  staff     1230336 Mar  3 13:45 firmware-T2B1-btconly-2.7.0-1f67c1e.bin
-rw-r--r--@   1 bosomt  staff     1230336 Mar  7 19:22 firmware-T2B1-btconly-2.7.0-45e8a842a-signed.bin
-rw-r--r--@   1 bosomt  staff     1608704 Mar  3 13:45 firmware-T2T1-2.7.0-1f67c1e.bin
-rw-r--r--@   1 bosomt  staff     1608704 Mar  7 19:22 firmware-T2T1-2.7.0-45e8a842a-signed.bin
-rw-r--r--@   1 bosomt  staff     1290240 Mar  3 13:45 firmware-T2T1-btconly-2.7.0-1f67c1e.bin
-rw-r--r--@   1 bosomt  staff     1289728 Mar  7 19:22 firmware-T2T1-btconly-2.7.0-45e8a842a-signed.bin
-rw-r--r--@   1 bosomt  staff       22222 Mar  7 19:39 translation-T2B1-de-DE-2.7.0.bin
-rw-r--r--@   1 bosomt  staff       23166 Mar  7 19:39 translation-T2B1-es-ES-2.7.0.bin
-rw-r--r--@   1 bosomt  staff       24476 Mar  7 19:39 translation-T2B1-fr-FR-2.7.0.bin
-rw-r--r--@   1 bosomt  staff       24742 Mar  7 19:39 translation-T2T1-de-DE-2.7.0.bin
-rw-r--r--@   1 bosomt  staff       28144 Mar  7 19:39 translation-T2T1-es-ES-2.7.0.bin
-rw-r--r--@   1 bosomt  staff       35738 Mar  7 19:39 translation-T2T1-fr-FR-2.7.0.bin
```

Issue #2820, opened 2023-02-14
Consider using ruff as the one and only linter of python code


grdddj:
If all what [ruff](https://github.com/charliermarsh/ruff) promises to do now is true, we should be able to replace all non-type checking static analysis (`isort`, `black`, `flake8` and `pylint`) just with this one tool, which should also be lightning-quick, with autofixing and much more nice features.

matejcik:
> `black`

a formatter not a linter, ruff doesn't seem to claim formatting functionality

> `isort`

not sure if there's import sorting built in?

> `pylint`

this is the big one. we essentially use a whitelist of pylint rules. if ruff can do all or most of them, we would get a neat win just from replacing pylint.

matejcik:
we also have some flake8 plugin or other, i think for checking package requirements?

grdddj:
> 



> > `black`
> 
> a formatter not a linter, ruff doesn't seem to claim formatting functionality

Right - https://github.com/charliermarsh/ruff#is-ruff-compatible-with-black. Luckily `black` is quite quick on our codebase.

> > `isort`
> 
> not sure if there's import sorting built in?

Not completely sure, https://github.com/charliermarsh/ruff#isort-i says it can do `unsorted-imports`, so it might even have a way to autofix them

> we also have some flake8 plugin or other, i think for checking package requirements?

We have `flake8-requirements`. That I did not find in the list of supported rules - https://github.com/charliermarsh/ruff#table-of-contents

---

One big problem I have is the inability to even run `ruff` on my `NixOS` machine - I can `pip install ruff` anywhere, it will go fine, is available in python shell as `import ruff`, but always getting `ruff: command not found` when trying to use it as a CLI.

On Windows, it works fine, so I guess `NixOS` will be the issue.

prusnak:
> I guess `NixOS` will be the issue.

ruff is packaged in nixpkgs, so nixos is not an issue

grdddj:
> > I guess `NixOS` will be the issue.
> 
> [ruff is packaged in nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/ruff/default.nix), so nixos is not an issue

Thanks for this info! It was possible to add it to our `shell.nix` and use it. Added as a dependency in https://github.com/trezor/trezor-firmware/pull/3207 - still with some issues, at least on my machine, I cannot use it from `Makefile`

matejcik:
[...]

PR #3207, opened 2023-08-14
chore(python): add ruff linter as a dependency

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/2820:
- adds `ruff` linter into `shell.nix`, so we can run it on python files
- usage - `ruff .` in `core`

I could not make it a `make` command or adding it into `pystyle` command, it was complaining that `ruff: No such file or directory`

### DIFF
--- a/core/embed/firmware/bootloader_hashes.py
+++ b/core/embed/firmware/bootloader_hashes.py
@@ -29,7 +29,7 @@ def aligned_digest(data: bytes, padding: bytes) -> bytes:
     (unwritten NOR-flash byte) or 0x00 (explicitly cleared byte).
     """
     if len(data) > ALIGNED_SIZE:
-        raise ValueError(fn, "too big")
+        raise ValueError("too big")
 
     assert len(padding) == 1
     digest_data = data + padding * (ALIGNED_SIZE - len(data))
@@ -55,6 +55,7 @@ def bootloader_str(file: Path) -> str:
     data = file.read_bytes()
 
     suffix = file.stem[len("bootloader_") :].upper()
+
     bytes_00 = to_uint_array(aligned_digest(data, b"\x00"))
     bytes_ff = to_uint_array(aligned_digest(data, b"\xff"))
 
@@ -84,7 +85,10 @@ def main():
 
         # write bootloader definitions
         for file in BOOTLOADERS.glob("bootloader*.bin"):
-            bl_check_new.append(bootloader_str(file))
+            try:
+                bl_check_new.append(bootloader_str(file))
+            except ValueError:
+                raise ValueError(f"Data too big in {file}")
 
         # consume up to auto-end
         for line in f:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,3 +102,11 @@ extra_standard_library = [
 ]
 known_first_party = ["trezorlib", "apps", "coin_info", "marketcap", "ui_tests", "gitlab"]
 known_third_party = ["trezor", "storage"]
+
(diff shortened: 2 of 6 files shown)


core/embed/firmware/bootloader_hashes.py
```diff
@@ -29,7 +29,7 @@ def aligned_digest(data: bytes, padding: bytes) -> bytes:
     (unwritten NOR-flash byte) or 0x00 (explicitly cleared byte).
     """
     if len(data) > ALIGNED_SIZE:
-        raise ValueError(fn, "too big")
+        raise ValueError(len(data), "too big")
```
prusnak, line 32: Let's also keep the filename, so it is obvious from the error message which file failed the assertion.
```suggestion
        raise ValueError(fn, len(data), "too big")
```

pyproject.toml
```diff
@@ -83,3 +83,12 @@ towncrier = "^23.6.0"
 [build-system]
 requires = ["poetry>=1;<1.1", "pip>=20"]
 build-backend = "poetry.masonry.api"
+
```
prusnak, line 86: nit: do we need two newlines here?

prusnak (changes requested):

core/embed/firmware/bootloader_hashes.py
grdddj: I removed the `fn`, because it is undefined here (flagged by `ruff`), we would need to pass it in as an argument

pyproject.toml
grdddj: Probably not, even though the formatter does not dislike it. Force-pushed change to just one newline

core/embed/firmware/bootloader_hashes.py
prusnak: Ah, right, nevermind. So this never worked as intended. Let's keep the change as it is.

grdddj: Theoretically the check can be moved to the higher function, there it has access to the filename:
```sh
index 4e90e0f4f0..91cc186d09 100755
--- a/core/embed/firmware/bootloader_hashes.py
+++ b/core/embed/firmware/bootloader_hashes.py
@@ -28,9 +28,6 @@ def aligned_digest(data: bytes, padding: bytes) -> bytes:
     Firmware needs to check the bootloader against a digest padded either by 0xff
     (unwritten NOR-flash byte) or 0x00 (explicitly cleared byte).
     """
-    if len(data) > ALIGNED_SIZE:
-        raise ValueError(len(data), "too big")
-
     assert len(padding) == 1
     digest_data = data + padding * (ALIGNED_SIZE - len(data))
     assert len(digest_data) == ALIGNED_SIZE
@@ -54,6 +51,9 @@ def bootloader_str(file: Path) -> str:
     """
     data = file.read_bytes()
 
+    if len(data) > ALIGNED_SIZE:
+        raise ValueError(file, len(data), "data too big")
+
     suffix = file.stem[len("bootloader_") :].upper()
     bytes_00 = to_uint_array(aligned_digest(data, b"\x00"))
     bytes_ff = to_uint_array(aligned_digest(data, b"\xff"))
```

matejcik: more like, leave this alone and wrap line 87 in a try-except adding the file info

grdddj: Done like this in the rebased force-push

obrusvit:
Obsolete PR.

Issue #3151, opened 2023-07-14
Do not enter bootloader menu if user is holding two buttons


Hannsek:
more info [here](https://satoshilabs.slack.com/archives/C016ACNF1BN/p1689170546345179?thread_ts=1689169706.886339&cid=C016ACNF1BN)

or just restrict entering the bootloader menu right away.

matejcik:
could not reproduce the problem

Issue #2756, opened 2023-01-19
Prodtest for T2B1


TychoVrahe:
Current `prodtest` binary is not adjusted for T2B1 - remove touch, sd card, add buttons etc.

PR #3209, opened 2023-08-14
Secret handling in bootloader

matejcik:
Spinned off #2756 to be merged sooner, before prodtest impl is stabilized.
@TychoVrahe do you think we could quickly fix https://github.com/trezor/trezor-firmware/issues/3151 also?


TychoVrahe:
I actually failed to reproduce #3151 . Will give it another try later today and will see.

PR #3212, opened 2023-08-15
toiftool

matejcik:
you will love it :black_heart: 

This PR adds a command `toiftool` available inside the `poetry shell` (and installable separately if you so desire).

`toiftool info something.toif` will tell you basic info
`toiftool show something.toif` will open the picture in your preferred image viewer
`toiftool convert something.jpg something.toif`, or vice-versa, will do a conversion. (this was formerly `core/tools/toif_convert.py`)


prusnak:
Why do we have both `core/tools/toiftool.py` and `python/tools/toiftool/toiftool.py`?

matejcik:
ah, that's a mistake. we don't have core/tools/toiftool anymore

PR #3216, opened 2023-08-15
chore: update common defs

szymonlesisz:
Little cleanup in trezor-common definitions:

- DOGE dust_limit should be set to 0.01 DOGE, not 1 sat below (sending amount below dust is forbidden but sending equal is OK) related to [PR in suite](https://github.com/trezor/trezor-suite/pull/9176) also described in DOGE [fee policy](https://github.com/dogecoin/dogecoin/blob/master/doc/fee-recommendation.md#dust-limits)
- removed outdated `default_fee_b` values. i believe they were there because of copy/paste of bitcoin definitions but:
   1. default values are much higher than the reality. long term average for BTG is ~12 for DGB is ~7
   2. multiple fee levels for those coins doesn't make sense since fee estimated by blockbook gives the same value for all of them
- removed dead blockbooks
  1. I've only tested coins with "support: connect"
  2. Peercoin + PeercoinTestnet is not working either even if they are alive (probably outdated blockbook build or disabled websocket support) so i didn't remove them


prusnak:
I added db36d09ebff981bef9c3a1f5efd953152ebc89be which removes MUE coin - it is dead. Please review the added commit.

mroz22 (approved):
changes make sense to me. but I am not very familiar with this repo so I can' tell if they are complete or something is missing somewhere (eg changelogs,...), so please others review too.

Issue #3183, opened 2023-07-27
Shamir info button not working correctly


grdddj:
In `TT` Shamir recovery flow, when clicking the info button on the recovery homescreen and then going back, the screen is blank

e.g. https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/4751858982/artifacts/test_ui_report/passed/TT-device_tests-reset_recovery-test_recovery_slip39_advanced.py%3A%3Atest_secret_click_info_button%5Bshares0-c2d2-850ffa77.html (below)

![image](https://github.com/trezor/trezor-firmware/assets/42543243/08254b9f-8cd6-46ef-b226-b5a73db49801)

conduition:
Replicated by doing:

```
trezorctl device setup -b advanced -t 128
```
And selecting two groups with one share each, with a two-group threshold. Then recovering and hitting the more info button here.

![1](https://github.com/trezor/trezor-firmware/assets/130806907/525c1c23-bde7-434f-935c-3a14c405c9c5)

After pressing 'more info':

![2](https://github.com/trezor/trezor-firmware/assets/130806907/461e0a14-5ed3-4b5c-838f-7c05dd59df30)

After pressing 'continue':

![3](https://github.com/trezor/trezor-firmware/assets/130806907/f6023ad0-da71-4543-a333-483da5e54755)

```
trezorctl wipe
trezorctl recover
```


I replicated the bug, but it appears slightly differently on my emulator from the screenshots in OP. 

I'm investigating the cause now

Hannsek:
I suppose you have an unreleased version of fw, 2.6.1, right?

conduition:
@Hannsek That's correct. I replicated on a41a8c5f8721f50d89a8f8b9efd9eb3a3809854c ([tree](https://github.com/trezor/trezor-firmware/tree/a41a8c5f8721f50d89a8f8b9efd9eb3a3809854c))

PR #3218, opened 2023-08-15
fix(core): more info button on shamir recovery screen

conduition:
Fixes https://github.com/trezor/trezor-firmware/issues/3183. 

It was a simple enum typo which caused the display to wipe its previous contents when the `Continue` button was pressed. Instead we should return a `Cancelled` message so that the previous screen's contents are not wiped.

## Testing

Set up a device:
```
trezorctl device setup -b advanced -t 128
```
Choose two groups with one share each, with a two-group threshold. 

Then recover with `trezorctl device recover` and hit the more info button here.

After pressing 'more info':

After pressing 'continue':

Also works if the device is power cycled.


Hannsek:
@conduition thank you for the pr. 👍🏻

conduition:
Thanks @grdddj, glad I could take a stab at it, at least.

PR #3219, opened 2023-08-16
upgrade to isort 5

matejcik:
takes advantage of and supersedes #2893 / #2908


matejcik:
not ready to merge because I built it on top of #3212 which should go in first to prevent conflicts in pyproject.toml

PR #3222, opened 2023-08-17
fixed unwanted delays in T2B1

TychoVrahe:
- delay on startup during display initialization
- delay on prodtest start because of uniitliazed spi handle
- delay when entering bootloader menu because of empty logo minimum show time


core/embed/bootloader/main.c
```diff
@@ -425,6 +425,10 @@ int bootloader_main(void) {
 
     ui_set_initial_setup(true);
 
+#ifndef USE_BACKLIGHT
+    hal_delay(2000);
```
matejcik, line 429: is this actually necessary @Hannsek ?
iiuc, this is just prolonging the time the "empty logo" is shown. If we want to prolong time of something, it should be the next `ui_screen_welcome_model` (saying "Trezor Model X") -- currently shown for 1000ms before going to "go to tio/start"

In my opinion, the role of the "empty lock" is to indicate that the device is powered up and didn't get to do anything useful yet. On TT this covers the 400ms needed to initialize touch; there is nothing like that on TR so I would expect the empty lock to basically immediately go away, no?

Hannsek: I (hope that I) remember we had the discussion about this with @TychoVrahe; that there was some other reason.

But if there is no other reason, let's do it the same as in TT as lower time can feel buggy.

TychoVrahe: removed. there is no blinking visible, visually it goes straight to the welcome screen. 
https://github.com/trezor/trezor-firmware/commit/86a5251214209609bee1cf3f63b02ac96a18f075

TychoVrahe:
https://github.com/trezor/trezor-firmware/commit/e12328049daa10e507f747db2fa5ccd230776969  here the device name is added below the empty logo. The minimum delay on startup is removed. Model T name-icon is fixed to match the font.

core/embed/bootloader/main.c
TychoVrahe: added the 1500ms delay for full logo https://github.com/trezor/trezor-firmware/commit/ad4ea54159417c747899734f51cbd556d57b1a3d

matejcik:
tweaked the offsets and prettified the code a little in 25d9531c23a79d49f81d483d4b657c97a3a2dc85

LGTM now, @TychoVrahe please review my commit and if it is OK for you then merge it

matejcik (approved):

matejcik:
no wait, one more thing actually - when the device is empty, we should be using the empty logo, not the full one, right @Hannsek ?

Hannsek:
Yes, as agreed yesterday [on Slack](https://satoshilabs.slack.com/archives/C02BY51RD8R/p1692283678331799?thread_ts=1692281568.420909&cid=C02BY51RD8R).

matejcik:
now the code should reflect that, @TychoVrahe please check and merge if it looks OK to you

TychoVrahe:
the usage of empty logo on empty device affects TT bootloader, do we want a changelog entry for that change?

TychoVrahe:
dtto for the device name under logo.

PR #3227, opened 2023-08-18
print correct model in print_firmware_version

TychoVrahe:
print correct model in print_firmware_version

### DIFF
--- a/python/.changelog.d/3227.fixed
+++ b/python/.changelog.d/3227.fixed
@@ -1,0 +1,1 @@
+Fixed printing Trezor model when validating firmware image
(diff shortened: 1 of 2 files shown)


matejcik (commented):
please add a changelog entry

(use `+.fixed` to create an entry not linked to an issue)

prusnak:
> (use `+.fixed` to create an entry not linked to an issue)

You still can use the PR number instead, no?

matejcik:
or that, yes. not sure if it adds any value in this case

matejcik (approved):

Issue #3191, opened 2023-08-07
"Show all" instead of "view full" 


Hannsek:
Replace on both, TT and T2B1, `view full` with `show all`. E.g.

![image](https://github.com/trezor/trezor-firmware/assets/57008159/e05d4390-34ac-4f14-abf3-625d6581eef9)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/dd0bdef4-0c94-4693-9735-31d8b1108edd)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/18ac6311-6f32-4d25-a888-83a659d88b1b)

grdddj:
Is it just highlighting the text in Figma, or do we want the text background to be yellow on TT?

Hannsek:
It was highlighted in Figma. Replaced the screens.

PR #3229, opened 2023-08-23
chore(core): replace `View full` button text with `Show all`

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3191:
- change button text to `Show all`


Hannsek:
lgtm

there is some problem in `core device test` as I cannot see the screens from master?
image

also 6 failed tests in `core device R test`?
image

Hannsek (approved):

grdddj:
> there is some problem in `core device test` as I cannot see the screens from master? image

No idea here, the pipeline is green otherwise and nothing changed in this flow

> also 6 errors in `core device R test`? what does it mean? image

Error means that UI comparison failed - that is expected, as we changed the button text

bosomt:
**QA OK**

 - Device: Trezor T2T1 2.6.1 regular (revision a8a97276a8d57c75f97a904bfec0119be6452cf9)

Issue #3168, opened 2023-07-21
T2B1 fido, u2f design


Hannsek:
https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?type=design&node-id=8825-68720&mode=design&t=LNvdUCDk0z9VwL5J-4

Hannsek:
Edited in Figma. Waiting for copy approval form @b-irving.

Hannsek:
@grdddj this is done?

grdddj:
> @grdddj isn't this done already?

I think so - https://figma.grdddj.eu/flow/FIDO

We can close it then

Hannsek:
Need change the error screen + change the numbering of the screens [here](https://figma.grdddj.eu/flow/FIDO).

PR #3230, opened 2023-08-23
chore(core): change error screens according to design

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3168:
- change the design of (not only) Fido error message


Hannsek:
image

1. row: change `close` to `try again` please
2. row: from which flow this screen comes? When users would see this?
3. row: 👌🏻

grdddj:
> * row: change `close` to `try again` please
> 
> * row: from which flow this screen comes? When users would see this?
> 
> * row: 👌🏻

Changed all the button texts to `TRY AGAIN` in this error screen.

The second one comes when you are asked for PIN and you give an incorrect one. So `TRY AGAIN` makes sense there as well

Hannsek:
> The second one comes when you are asked for PIN and you give an incorrect one. So TRY AGAIN makes sense there as well

There shouldn't be a separate screen. There should be this: 
image
and after the first move it should change
image

grdddj:
> > The second one comes when you are asked for PIN and you give an incorrect one. So TRY AGAIN makes sense there as well
> 
> There shouldn't be a separate screen. There should be this: image and after the first move it should change image

Right, let's rather create a separate issues for this? I will record the test changes here and will be ready for merge

mmilata (approved):
LGTM! Please fix style before merging.

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1440x900

PR #3232, opened 2023-08-24
setup single-place layout selection

TychoVrahe:
setup single-place layout selection

this allows easier addition of next models and easier maintenance.


core/SConscript.firmware
```diff
@@ -378,6 +378,19 @@ env = Environment(ENV=os.environ, CFLAGS=f"{ARGUMENTS.get('CFLAGS', '')} -DPRODU
 
 FEATURES_AVAILABLE = tools.configure_board(TREZOR_MODEL, FEATURES_WANTED, env, CPPDEFINES_HAL, SOURCE_HAL, PATH_HAL)
 
+
+if TREZOR_MODEL in ('T', 'DISC1'):
+    LAYOUT = 'LAYOUT_TTV2'
+elif TREZOR_MODEL in ('1', 'R'):
+    LAYOUT = 'LAYOUT_TR'
```
mmilata, line 385: I don't like `ttv2` and `tr`, let's use `touch` and `button` (or `oled` or something like that)?

This might also be a good time to rename:
* `core/src/trezor/ui/layouts/tt_v2` -> `core/src/trezor/ui/touch`
* `core/src/trezor/ui/layouts/tr` -> `core/src/trezor/ui/button`
* `core/embed/rust/src/ui/model_tt` -> `core/embed/rust/src/ui/touch`
* `core/embed/rust/src/ui/model_tr` -> `core/embed/rust/src/ui/button`

Thoughts?

matejcik: I actually think the "ttv2" and "tr" are appropriate names here, because the UI layout is more tied to a specific model / form factor than to the control scheme.

e.g., consider a future Model E with 128x128 oled. I would expect it to have different layouts. Same thing for a bigger screen touch model: the layouts are probably going to look different.

TychoVrahe: i agree with matejcik here. As a sidenote, if/when we introduce model with 128x128 oled or something like that, the layout will be different but many of the UI components will be shared with "tr" - so we will still need to introduce some kind of hierarchy in the rust UI codebase.

mmilata: Fair enough, however what do you think about dropping `v2` suffix?

core/SConscript.firmware
```diff
@@ -428,6 +441,7 @@ env.Replace(
         'FIRMWARE',
         'TREZOR_MODEL_'+TREZOR_MODEL,
         'USE_HAL_DRIVER',
+        LAYOUT,
```
mmilata, line 444: Nit: not sure layout is the right word here. What do you think about using `UI` instead?

core/SConscript.firmware
TychoVrahe: i don't have a preference, the suffix hints that there was some evolution, not sure that it can help in some way. on the other hand, the names are kinda arbitrary anyway (`tr` will mean nothing to future devs as we are abandoning the name and models with same screen sizes like T3T1 will share the layouts, so `tt` is not exact either)

core/SConscript.firmware
TychoVrahe: i can see the problem with `LAYOUT`, but `UI` seems a bit too general. how about `UI_LAYOUTS`, `UI_TYPE` or something like that?

mmilata: Both sound fine. Either `UI_LAYOUTS` or `UI_LAYOUT`?

mmilata (approved):

core/SConscript.firmware
TychoVrahe: Decided for dropping the suffix, because in rust we don't have it and its bound together by the interface definition. So its better to have the name closer. 

Also the rust's `model_feature` is selected in same place as UI_LAYOUT now, for the same reason: its bound together by interface.

https://github.com/trezor/trezor-firmware/pull/3232/commits/a55e18ef1d698c5d78e3242ab755bd41abef33d4

core/SConscript.firmware
TychoVrahe: `UI_LAYOUT` it is

PR #3234, opened 2023-08-24
move model specific norcow config to model header

TychoVrahe:
move model specific norcow config to model header


matejcik (approved):
utACK

PR #3235, opened 2023-08-24
add vendor headers for devices with dev bootloader

TychoVrahe:
so that on dev devices we can run dev firmware with locked bootloader and without waiting.

added as a new vendor headers, but these (at this moment) same as QA headers, it might also make sense to keep only one - probably the dev as it seems more general.


matejcik:
shouldn't we just be using `satoshilabs_signed_dev` for this?

matejcik:
...of course not, because these use the SatoshiLabs keys.

> added as a new vendor headers, but these (at this moment) same as QA headers, it might also make sense to keep only one - probably the dev as it seems more general.

yes, please do that. I would have expected to just use the "qa" one, but calling it "dev" is more readable

TychoVrahe:
done here https://github.com/trezor/trezor-firmware/pull/3235/commits/3301f2c42c5d41413575413f8d3a442b9c1c5863

matejcik (approved):

PR #3236, opened 2023-08-24
add install restricted screen on installation with locked bootloader

TychoVrahe:
Somehow we didn't add it with the last T2B1 bootloader changes. 

Also fixes residual "attestation present" wording. Especially the return message when failing to install might be a reason to add this to earliest production bootloader version, as its something a user see in trezorctl and suite can react to it too, so having two version might be annoying.

### DIFF
--- a/core/embed/bootloader/main.c
+++ b/core/embed/bootloader/main.c
@@ -188,7 +188,11 @@ static usb_result_t bootloader_usb_loop(const vendor_header *const vhdr,
       case MessageType_MessageType_FirmwareUpload:
         r = process_msg_FirmwareUpload(USB_IFACE_NUM, msg_size, buf);
         if (r < 0 && r != UPLOAD_ERR_USER_ABORT) {  // error, but not user abort
-          ui_screen_fail();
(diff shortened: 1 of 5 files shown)


core/embed/bootloader/main.c
```diff
@@ -188,7 +188,15 @@ static usb_result_t bootloader_usb_loop(const vendor_header *const vhdr,
...
         if (r < 0 && r != UPLOAD_ERR_USER_ABORT) {  // error, but not user abort
+#ifdef USE_OPTIGA
+          if (r == UPLOAD_ERR_BOOTLOADER_LOCKED) {
+            ui_screen_install_restricted();
+          } else {
+            ui_screen_fail();
+          }
+#else
           ui_screen_fail();
+#endif
```
matejcik, line 199: let's avoid the ifdef here, and instead redefine `ui_screen_install_restricted` to `ui_screen_fail` if there is no optiga.

TychoVrahe: https://github.com/trezor/trezor-firmware/commit/cc79e566602abcc4bb1a044dc6fb8d445a6051f3

matejcik (approved):

Issue #3210, opened 2023-08-15
Chunked addresses


Hannsek:
From Suite repo:

> As the user would be able to change the view of the addresses (chunked/not-chunked), Suite should remember the last state. So if the user chose to have it chunked last time, next time he/she receives/sends, he/she should see the address chunked, and vice versa.
> 
> Each view should have its own message which would be sent to Trezor. According to this message, Trezor would display chunked/not-chunked address.

Is this possible? Receiving messages from suite and change the view appropriately? @matejcik @grdddj @mmilata 

[T2B1](https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?type=design&node-id=3790-11545&mode=design&t=SCxX0NadL7vcFa04-0) 

[T2T1](https://www.figma.com/file/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?type=design&node-id=478-3&mode=design&t=GVbDeO39c8DE3E39-0)

grdddj:
> Is this possible? Receiving messages from suite and change the view appropriately?

I guess technically possible, but probably not easy to do. The details of how to display something are very far away from receiving a protobuf message from `Suite`.

Seems we would need to have a `single global layout` so we can check it is the address screen and force it to recreate/update with changed parameters?

prusnak:
Easiest is to expand the `GetAddress` protobuf message with field `optional chunked bool` and show chunked address only if it is set to true.

matejcik:
> Is this possible? Receiving messages from suite and change the view appropriately? @matejcik @grdddj @mmilata

it should be, with a tiny amount of trickery.
`ctx.wait(layout)` is shortcut for `loop.race(layout, ctx.read_any())`

if instead of `ctx.wait(layout)`, we use explicit `loop.race(layout, ctx.read(GetAddress))`, we will be able to intercept a subsequent GetAddress without ever leaving the running workflow.

the only difficulty is that the `ctx.wait` call is hidden deep enough in the stack that we'll need a way to propagate it out. we might want to move `show_address` out of layouts and into `apps.common` and implement the logic there, take an argument specifying what message to wait for...

Issue #690, opened 2019-11-10
Describe protobuf communication


tsusanka:
Part of #685.

We should extend [communication.md](https://github.com/trezor/trezor-firmware/blob/f4385ede2aed4e74a5fc51ba33a07d2150cb6c8d/core/docs/src/communication.md), add an example and document protobuf messages, which might not be clear right away. I'm thinking PassphraseReq/PassphraseAck - after #1, ButtonRequests and may some other?

PR #3237, opened 2023-08-25
Chunkifying the receiving address

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3210:
- allowing for showing the address in chunks of 4 characters

image

image


grdddj:
Interesting edge-case in TT and Cardano - when there are more pages, the text area is smaller because of the scrollbar, and the last character does not fit. Might be possible to decrease the `x_offset` in case there are more pages

image

prusnak:
> Interesting edge-case in TT

Do we still want to show the dots after adding the new hints? Especially since you can't go back anyway ...

common/protob/messages-bitcoin.proto
```diff
@@ -110,6 +110,7 @@ message GetAddress {
     optional MultisigRedeemScriptType multisig = 4;                     // filled if we are showing a multisig address
     optional InputScriptType script_type = 5 [default=SPENDADDRESS];    // used to distinguish between various address formats (non-segwit, segwit, etc.)
     optional bool ignore_xpub_magic = 6;                                // ignore SLIP-0132 XPUB magic, use xpub/tpub prefix for all account types
+    optional bool chunkify = 7 [default=true];                          // display the address in chunks of 4 characters
```
prusnak, line 113: We don't want this to default to true on the protobuf level, do we? 

Or in another word, we would use `default=true` only if the old behaviour was chunkifying by default and we are introducing an option to turn it off. But it's the other way around - the old behaviour did not chunkify be default.

```suggestion
    optional bool chunkify = 7;                                         // display the address in chunks of 4 characters
```

OTOH it's fine to default this value on higher levels - for example trezorlib cli.

prusnak (changes requested):

common/protob/messages-bitcoin.proto
grdddj: Right, in the end, I will probably default it to `false` here, but for the development process, it is nice to have `true` here so I do not have to change tests and see the chunkifying on all the possible places.

grdddj:
> 
> @grdddj We should display max 16 chunks per page, not more. See the image below.

Is it worth doing this complexity, when the only addresses longer than 16 chunks are `Cardano`? Others are fine even without this limit

> What is the vertical space between rows/chunks? Is it currently variable?

Vertical space between coherent text (inside one paragraph) is constant, as in all other text

> Also, can we align the "next page" arrow to fit the last chunk (or block of text)?

Yes, we can somehow, but same as above, it is only valid for `Cardano`, others do not need it

Hannsek:
Is it worth doing this complexity, when the only addresses longer than 16 chunks are Cardano? Others are fine even without this limit?

Hmm, probably not. And we can set the cardano address to appear as a standard string. 
However, we should do it for (multisig) xpubs as well in the future.

> Vertical space between coherent text (inside one paragraph) is constant, as in all other text

We should make it larger in both, TT and TR.

grdddj:
> > Vertical space between coherent text (inside one paragraph) is constant, as in all other text
> 
> We should make it larger in both, TT and TR.

Do you want to increase the vertical margin in this special case of chunked address, or in all the screens in general?

Hannsek:
Only in this special case of chunked address.

Hannsek:
Also please increase the horizontal space between chunks on TT by 1 px.

grdddj:
Last commit introduces chunkifying also for the address in sign flow. In `TR`, there is one edge-case where the word `CoinJoin` (as an `address_label`) is also shown at that screen, and it is chunkified as well. It would be quite complicated to render the label without chunks and the address with chunks.

@Hannsek Please consider if this looks good enough - and please look at all the other changed screens in signing flow.

image

Hannsek:
We should render the label not chunked.

grdddj:
Did also the `account` addition, and some of the labels do not fit properly:
image

mmilata:
Not sure if it's an improvement on model T:
0d60046844f9f4e9dd1713962f13dd7acfc30830f97531ff0246f5c12b4c595a

mmilata:
Is the font really fixed width? Sometimes the alignment looks very wrong:
9edaffd47cb893106d4586c603663c85c90fb1cbb24fcf0e071cdf4f0e5c094b

grdddj:
> Is the font really fixed width? Sometimes the alignment looks very wrong: 9edaffd47cb893106d4586c603663c85c90fb1cbb24fcf0e071cdf4f0e5c094b

Right, the font is an `almost-mono`. Some letters like `m`, `M`, `w` or `W` are one pixel wider than others (8 vs 7 pixels - including margin) - can be seen in https://github.com/trezor/trezor-firmware/blob/master/core/embed/lib/fonts/font_pixeloperatormono_regular_8.c

Seems there are no easy fixes, as the font is so small (narrow) that these four letters just cannot be made as narrow as other characters.

Possible measures:
- manually increasing the width of _all_ the letters from 7 to 8 pixels ... at the cost of making the majority of the text wastefully wider
- accounting for it in the rendering, remembering the number or "extra pixels" in the chunk and decreasing the `x-offset` - this way at least the chunks would be starting at the same vertical line (however, not applicable when there are no chunks)

Do you see some other possibilities?

Hannsek:
Please delete "account" in receive flow.
We need to figure out the wording in send on TT. So probably keep the "account" only for send flow on T2B1.

> accounting for it in the rendering, remembering the number or "extra pixels" in the chunk and decreasing the x-offset - this way at least the chunks would be starting at the same vertical line (however, not applicable when there are no chunks)

Please try it

Hannsek:
I can see some regression with text:
image
image
image
image

Hannsek:
The bigger problem is this, wrong positioning of the left arrow. 
image

grdddj:
High-level question - does somebody really care about `xpub` on `Trezor` screen that they will appreciate the chunkification in this case? We might not do it for `xpub`, only for `address` - would save us this trouble with paginated content.

Or second thought - will somebody care that it is chunkified but "wrongly"?

I will however spend a little while to make the previous-page-ellipsis take the same space as the left-most chunk

Hannsek:
> Or second thought - will somebody care that it is chunkified but "wrongly"?

yes, I'd remove it from xpubs for now

common/protob/messages-bitcoin.proto
grdddj: Now I do not have any default value on the protobuf side

grdddj:
CI is green, @Hannsek please go over the screens briefly and then we are ready to merge

grdddj:
Thanks for the review!

The tests are now a mixture of chunkified and non-chunkified screens, so we can see/test both.

There are a lot of new tests because of this, and therefore some screens have "missing" on the left - so nothing had to be changed there, they just were not there before.

I did not yet fix the next/previous page arrows, as that is only for Cardano. I can theoretically still do it before the release, but preferably in a different PR - so please create a new issue for that, if we want to do it.

Hannsek:
It is ready for merge from my POV as I will open new issue.

The approval is still on @mmilata.

grdddj:
> It is ready for merge from my POV as I will open new issue.
> 
> The approval is still on @mmilata.

Right, I will now squash the commits, resolve conflicts and add a changelog entries

core/src/apps/tezos/sign_tx.py
```diff
@@ -71,12 +71,16 @@ async def sign_tx(msg: TezosSignTx, keychain: Keychain) -> TezosSignedTx:
             # operation to transfer tokens from a smart contract to an implicit account or a smart contract
             elif transfer is not None:
                 to = _get_address_from_contract(transfer.destination)
-                await layout.require_confirm_tx(to, transfer.amount)
+                await layout.require_confirm_tx(
+                    to, transfer.amount, chunkify=bool(msg.chunkify)
```
matejcik, line 75: small nit: we could get rid of the `bool()` noise by declaring the field `[default=false]` in the protobuf

grdddj: Right, I kind of did not want to do this, as that would make the protobuf definition line longer and all the comments would need to be aligned - causing quite a big protobuf diff.

Poor reason, what do you think about it?

matejcik: let's keep as-is now, and maybe create a separate issue for it?
we want to move the protobuf comment to the line before anyway for #690

grdddj: Agreed, issue created - https://github.com/trezor/trezor-firmware/issues/3274

core/embed/rust/src/ui/component/text/op.rs
```diff
@@ -258,4 +271,6 @@ pub enum Op<T: StringType> {
     CursorOffset(Offset),
     /// Force continuing on the next page.
     NextPage,
+    /// Render the following text in a chunkified way. None will disable that.
+    Chunkify(Option<(Chunks, i16)>),
```
mmilata, line 275: Should probably have separate operation for changing the line spacing as these are really a different thing.

core/src/apps/nem/transfer/__init__.py
```diff
@@ -12,11 +12,12 @@ async def transfer(
     common: NEMTransactionCommon,
     transfer: NEMTransfer,
     node: bip32.HDNode,
+    chunkify: bool = False,
```
matejcik, line 15: argument here and in layout.ask_transfer does not need a default value

core/src/apps/monero/layout.py
```diff
@@ -143,6 +145,7 @@ async def _require_confirm_output(
     dst: MoneroTransactionDestinationEntry,
     network_type: MoneroNetworkType,
     payment_id: bytes | None,
+    chunkify: bool = False,
```
matejcik, line 148: remove default value

core/src/apps/cardano/layout.py
```diff
@@ -210,6 +210,7 @@ async def confirm_sending(
     to: str,
     output_type: Literal["address", "change", "collateral-return"],
     network_id: int,
+    chunkify: bool = False,
```
matejcik, line 213: remove default value

core/src/apps/bitcoin/sign_tx/layout.py
```diff
@@ -62,6 +62,7 @@ async def confirm_output(
     coin: CoinInfo,
     amount_unit: AmountUnit,
     output_index: int,
+    chunkify: bool = False,
```
matejcik, line 65: remove default value

core/src/apps/binance/layout.py
```diff
@@ -34,21 +34,18 @@ def make_input_output_pages(msg: BinanceInputOutput, direction: str) -> None:
     for txoutput in msg.outputs:
         make_input_output_pages(txoutput, "Confirm output")
 
-    await _confirm_transfer(items)
+    await _confirm_transfer(items, chunkify=bool(msg.chunkify))
 
 
-async def _confirm_transfer(inputs_outputs: Sequence[tuple[str, str, str]]) -> None:
+async def _confirm_transfer(
+    inputs_outputs: Sequence[tuple[str, str, str]], chunkify: bool = False
```
matejcik, line 41: remove default value

core/embed/rust/src/ui/component/text/layout.rs
```diff
@@ -52,6 +52,40 @@ pub struct TextLayout {
...
 
+/// Configuration for chunkifying the text into smaller parts.
+#[derive(Copy, Clone)]
+pub struct Chunks {
+    /// How many characters will be grouped in one chunk.
+    pub chunk_size: usize,
+    /// How big will be the space between chunks (in pixels).
+    pub x_offset: i16,
+    /// Optional characters that are wider and should be accounted for
+    pub wider_chars: Option<&'static str>,
```
mmilata, line 63: Kinda hate this tbh (as well as the whole idea of trying to render variable-width font as a fixed-width font). Trying to think how to make this more general. Can we instead remember the width of a normal (or smallest) character and keep track of the actual chunk size when processing chunk?

core/embed/rust/src/ui/component/text/layout.rs
```diff
@@ -537,13 +600,30 @@ impl Span {
...
             let char_width = text_font.char_width(ch);
 
+            // When there is a set chunk size and we reach it,
+            // adjust the line advances and return the line.
+            if let Some(chunkify_config) = chunks {
+                if i == chunkify_config.chunk_size {
+                    line.advance.y = 0;
+                    // Decreasing the offset for each wider character in the chunk
+                    line.advance.x += chunkify_config.x_offset - chunks_wider_chars;
+                    return line;
```
mmilata, line 618: I wonder if this can't break line-counting somewhere?

matejcik (commented):
non-Rust part LGTM, just some superfluous default arguments

core/src/apps/bitcoin/keychain.py
```diff
@@ -342,6 +342,7 @@ def __init__(
         require_bech32: bool,
         require_taproot: bool,
         account_level: bool = False,
+        show_account_str: bool = False,
```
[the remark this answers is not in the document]
grdddj: Good idea, done in [185e6dc](https://github.com/trezor/trezor-firmware/pull/3237/commits/185e6dc852c62f2d745d7c88176e33cdc5302fb3)

core/embed/rust/src/ui/component/text/op.rs
grdddj: Done in [7f4ea77](https://github.com/trezor/trezor-firmware/pull/3237/commits/7f4ea773963e7e86969c25065f6428c46c663929)

core/embed/rust/src/ui/component/text/layout.rs
```diff
@@ -76,6 +110,14 @@ pub struct TextStyle {
...
     /// Specifies what to do at the end of the page.
     pub page_breaking: PageBreaking,
+
+    /// Optionally chunkify all the text with a specified chunk
+    /// size and pixel offset for the next chunk.
+    pub chunks: Option<Chunks>,
+
+    /// Optionally increase the vertical space between text lines
+    /// (can be even negative, in which case it will decrease it).
+    pub extra_y_line_advance: Option<i16>,
```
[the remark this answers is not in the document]
grdddj: Done in [eb3cab9](https://github.com/trezor/trezor-firmware/pull/3237/commits/eb3cab9f34925b8b4f2896ceeab1f3b2ef926a39)

core/embed/rust/src/ui/component/text/op.rs
```diff
@@ -90,6 +90,15 @@ impl<'a, T: StringType + Clone + 'a> OpTextLayout<T> {
...
                 }
+                Op::Chunkify(chunk) => {
+                    if let Some((chunk_config, extra_y_offset)) = chunk {
+                        self.layout.style.chunks = Some(chunk_config);
+                        self.layout.style.extra_y_line_advance = Some(extra_y_offset);
+                    } else {
+                        self.layout.style.chunks = None;
+                        self.layout.style.extra_y_line_advance = None;
+                    }
+                }
```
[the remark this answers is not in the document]
grdddj: I have simplified it in [eb3cab9](https://github.com/trezor/trezor-firmware/pull/3237/commits/eb3cab9f34925b8b4f2896ceeab1f3b2ef926a39) by separating to two `Op` enums

core/src/apps/binance/layout.py
grdddj: Done in [7e03759](https://github.com/trezor/trezor-firmware/pull/3237/commits/7e03759ea7c3f0e2e56218ee1bd5cb18bef9c134)

core/src/apps/bitcoin/sign_tx/layout.py
grdddj: Done in [7e03759](https://github.com/trezor/trezor-firmware/pull/3237/commits/7e03759ea7c3f0e2e56218ee1bd5cb18bef9c134)

core/src/apps/cardano/layout.py
grdddj: Done in [7e03759](https://github.com/trezor/trezor-firmware/pull/3237/commits/7e03759ea7c3f0e2e56218ee1bd5cb18bef9c134)

core/src/apps/monero/layout.py
grdddj: Done in [7e03759](https://github.com/trezor/trezor-firmware/pull/3237/commits/7e03759ea7c3f0e2e56218ee1bd5cb18bef9c134)

core/src/apps/nem/transfer/__init__.py
grdddj: Done in [7e03759](https://github.com/trezor/trezor-firmware/pull/3237/commits/7e03759ea7c3f0e2e56218ee1bd5cb18bef9c134)

core/embed/rust/src/ui/component/text/layout.rs
grdddj: The tests seem to be alright, it works even with longer addresses that paginate (Cardano ones)

core/embed/rust/src/ui/component/text/layout.rs
grdddj: Nor am I happy with this, but this "almost-mono-font" can hardly be fixed to be "completely-mono", because of its small width, as I am describing [here](https://github.com/trezor/trezor-firmware/pull/3237#issuecomment-1715136060)

Not sure I follow your suggestion - can you prototype some code for it?

Merging this does not need to be final, we can improve it in the future

common/protob/messages-bitcoin.proto
```diff
@@ -110,6 +110,7 @@ message GetAddress {
     optional MultisigRedeemScriptType multisig = 4;                     // filled if we are showing a multisig address
     optional InputScriptType script_type = 5 [default=SPENDADDRESS];    // used to distinguish between various address formats (non-segwit, segwit, etc.)
     optional bool ignore_xpub_magic = 6;                                // ignore SLIP-0132 XPUB magic, use xpub/tpub prefix for all account types
+    optional bool chunkify = 7;                                         // display the address in chunks of 4 characters
```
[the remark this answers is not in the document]
grdddj: Can change it to anything, if we agree. What is your preference? @matejcik 

`show_chunks` sounds good to me

mmilata (approved):
I'm OK with merging this, rename the parameter to `show_chunks` if you want.

grdddj:
@prusnak please review the change request, it is blocking the merge.

prusnak (approved):
I reviewed only protobuf def changes

PR #3243, opened 2023-08-31
ci: build bootloader emulators

### LINKED ISSUE #3193: Add bootloader emulator builds into CI
They are not being used very often, so it can happen that we break them without noticing - and then they do not work when we need them.

mmilata:
Fixes #3193.

### DIFF
--- a/core/SConscript.bootloader_emu
+++ b/core/SConscript.bootloader_emu
@@ -128,8 +128,12 @@ SOURCE_TREZORHAL = [
     'embed/trezorhal/unix/rng.c',
     'embed/trezorhal/unix/usb.c',
     'embed/trezorhal/unix/random_delays.c',
-    'embed/trezorhal/unix/secret.c',
 ]
+
+if TREZOR_MODEL in ('R', ):
+    SOURCE_TREZORHAL += [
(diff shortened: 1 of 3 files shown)


core/SConscript.bootloader_emu
```diff
@@ -128,7 +128,6 @@ SOURCE_TREZORHAL = [
     'embed/trezorhal/unix/rng.c',
     'embed/trezorhal/unix/usb.c',
     'embed/trezorhal/unix/random_delays.c',
-    'embed/trezorhal/unix/secret.c',
```
matejcik, line 131: isn't this a problem for model R ?

mmilata: right, it's not added by the board configuration function since this is emulator ...

TychoVrahe: yeah we need to add this conditionally in sconscript. i was thinking about something like "emulator boards", so far seems a bit like overkill but when adding new models these sconscripts might get messy, so we might eventually need that

mmilata: 194f517b49df31d3f2176a1cfcb16de02fa356f8

TychoVrahe (approved):

Issue #3226, opened 2023-08-18
Use internal model names in tests


TychoVrahe:
E.g. fixtures file has this:
```
"TR": {
"click_tests": {
"TR_test_autolock.py::test_autolock_does_not_interrupt_preauthorized": "757b981b1f41c948785e339e9f678b08925b2c310a1baad2ccb909689ac42e2b",
```
shoud us T2B1 instead of TR (goes for other models too)

Also some tests have stuff like `if client.debug.model in ("T", "R"):` where we should be using internal model name as well.

mmilata:
#3583 fixes conditionals in test cases but not fixtures.json
fixing that might be an opportunity to start using the internal names in build system as well

PR #3247, opened 2023-09-04
Internal model name in tests

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3226
- using internal model names wherever possible in tests


python/src/trezorlib/cli/cosi.py
```diff
@@ -34,7 +34,7 @@ def cli() -> None:
 
 
 def cosi_warn(client: "TrezorClient") -> None:
-    if client.features.model == "1" and client.version < (1, 11, 2):
+    if client.features.internal_model == "T1B1" and client.version < (1, 11, 2):
```
matejcik, line 37: perhaps we should probably be using `client.model is models.TREZOR_ONE` and similar everywhere

matejcik: to make the code shorter, we could introduce `T1B1 = TREZOR_ONE` in models.py

TychoVrahe:
mostly superseded by https://github.com/trezor/trezor-firmware/pull/4477

Issue #242, opened 2019-06-12
Further integration of HMAC DRBG


andrewkozlik:
We would like to use the HMAC DRBG as a source of random numbers in more places than just generating random delays. Eventually we would like to use it for everything, assuming the DRBG would be reseeded before generating security-critical values.
1. Determine where the DRBG can be used safely as is without reseeding, e.g. generating nonces, and make the replacements.
2. Ensure periodic reseeding of the DRBG from the TRNG.
3. Add more sources of randomness. These could be the value of the system clock at certain moments, e.g. user interaction.

As for 1., below is some preliminary research on where we could replace the TRNG with the DRBG right away:

random32():
  - storage mask NO
  - u2f keypath YES

random_buffer():
  - storage salt YES
  - storage keys NO
  - storage IVs YES
  - crypto bip39 NO
  - crypto monero random_scalar ?
  - modtrezorcrypto generating private keys in secp256k1 etc. NO
  - legacy UUID YES
  - legacy fsm_msgGetEntropy NO
  - legacy nem salt and IV YES
  - legacy reset int_entropy NO
  - legacy OTP hardware entropy NO

random_uniform():
  - storage guard key NO 
  - legacy recovery YES

random_permute():
  - legacy pinmatrix YES
  - legacy recovery YES

random.uniform():
  - u2f channel ID YES
  - u2f keypath YES
  - reset_device check YES

random.bytes():
  - reset_device internal_ent NO
  - slip39 master secret and shares NO
  - slip39 identifier YES
  - get_entropy NO
  - passphrase cache salt YES
  - storage.py device_id YES
  - nem/monero salt, IV and nonce YES

keepkeyjon:
Why is it preferable to use the DRBG over using the TRNG in as many places as possible?

andrewkozlik:
There are two reasons:
1. There is the recent freezing bug, which was apparently caused by reading from the TRNG before USB operations. It has something to do with I2C, I am not familiar with the details. To avoid this reocurring, it would be best to limit use of the TRNG especially when generating things like IVs, which do not require secrecy.
2. The advantage of the DRBG is that you can feed it multiple sources of entropy including the random data from the TRNG. This is a precaution in case the TRNG produces random numbers with lower entropy than it should or if it becomes biased as the hardware ages. It also makes things more difficult for attackers who are aiming to control the random numbers generated by the device. If entropy is taken from a single source such as the TRNG, then the attacker needs to disable only that one source.

[...]

PR #3255, opened 2023-09-07
Device authentication for Model R

andrewkozlik:

PR #3256, opened 2023-09-08
Use Optiga as a source of randomness in seed generation

andrewkozlik:
This PR adds a new boolean parameter `strong` to `random.bytes()`. When set to true, randomness from Optiga's TRNG is XOR-ed with the MCU's TRNG to produce the output. This is used in the most security-critical operations which are seed generation and Shamir backup to generate the random shares. In the future I'd like to see the randomness from Optiga being used for the generation of all random numbers. For that we'd need to test the device extensively to make sure that it doesn't unnecessarily slow down some operations on Trezor. It seems better to do that as part of https://github.com/trezor/trezor-firmware/issues/242 and the `strong` parameter can then be used to enforce reseeding of the DRBG.


andrewkozlik:
Converted to draft until https://github.com/trezor/trezor-firmware/pull/3255 is merged.

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-random.h
```diff
@@ -40,22 +44,52 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_1(mod_trezorcrypto_random_uniform_obj,
...
-  random_buffer((uint8_t *)vstr.buf, l);
+  vstr_init_len(&vstr, len);
+#if USE_OPTIGA
+  if (n_args > 1 && args[1] == mp_const_true) {
+    uint8_t *dest = (uint8_t *)vstr.buf;
+    if (!optiga_random_buffer(dest, len)) {
+      vstr_clear(&vstr);
+      mp_raise_msg(&mp_type_RuntimeError,
+                   "Failed to get randomness from Optiga.");
+    }
```
andrewkozlik, line 67: Should we fail to generate a seed if Optiga fails to generate randomness or should we silently ignore? I think it's better to err on the side of caution and fail. We could also retry, but if something is wrong with the communication it's probably best for the user to replug the device.

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-random.h
```diff
@@ -40,22 +44,52 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_1(mod_trezorcrypto_random_uniform_obj,
...
+  uint32_t len = trezor_obj_get_uint(args[0]);
+  if (len > 1024) {
     mp_raise_ValueError("Maximum requested size is 1024");
   }
   vstr_t vstr = {0};
-  vstr_init_len(&vstr, l);
-  random_buffer((uint8_t *)vstr.buf, l);
+  vstr_init_len(&vstr, len);
+#if USE_OPTIGA
+  if (n_args > 1 && args[1] == mp_const_true) {
```
matejcik, line 61:
```suggestion
  if (n_args > 1 && mp_obj_is_true(args[1])) {
```

matejcik (commented):
wondering if we could test that this works somehow.

one idea is, unit test will reseed to 0, grab random buffer, reseed again, grab strong random buffer, check that they come out different.
but unit tests only run in emulator so usefulness of the result is limited

andrewkozlik:
> wondering if we could test that this works somehow.

I can't think of an easy way to test this. We could expose both sources of randomness, the CPU TRNG and Optiga's TRNG, via `GetEntropy` and test that in the device tests. If we want to unit test the `random.bytes()` function properly, then maybe we should have an independent PRNG for the Optiga in the emulator and check that the two sources are getting XOR-ed. But then that will get more complicated once the DRBG is introduced. Also, the unit test will work only with the Trezor R emulator, because the `strong` option is ignored for other Trezor models.

matejcik:
do we want to get this into the current release?

andrewkozlik:
Rebased and updated fixtures.

matejcik (approved):
utACK if CI passes

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -89,3 +90,22 @@ bool optiga_read_cert(uint8_t index, uint8_t *cert, size_t max_cert_size,
...
+    dest += OPTIGA_RANDOM_MAX_SIZE;
+    size -= OPTIGA_RANDOM_MAX_SIZE;
+  }
+
+  if (size < OPTIGA_RANDOM_MIN_SIZE) {
+    static uint8_t buffer[OPTIGA_RANDOM_MIN_SIZE] = {0};
+    optiga_result ret = optiga_get_random(buffer, OPTIGA_RANDOM_MIN_SIZE);
+    memcpy(dest, buffer, size);
+    return ret == OPTIGA_SUCCESS;
+  }
```
andrewkozlik, line 108: Rereading the code, I realized these two blocks need to be swapped, otherwise requesting for example between 257-263 bytes fails. Done in 8f20a0177f86eadb2cab64ba9cc7045ce0dbbca0.

prusnak (approved):
utACK

Issue #3265, opened 2023-09-13
Fix button icons


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/4f3d1308-af01-41b3-8853-358a9c746ba8)
<img width="214" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/efd039c2-204d-4a0d-95db-e971aa2bdeb4">
<img width="123" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/369296d2-81d2-49d4-b3c9-7b170fbd0f64">
<img width="147" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/da982e25-ceec-4777-8a6d-f35e5dcda964">

Hannsek:
Maybe only the left upper arrow is off…

PR #3269, opened 2023-09-14
Fix up arrow on T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3265:
- moves the up=arrow one pixel up and to the right


Hannsek (approved):
lgtm

Issue #3268, opened 2023-09-13
Report unlocked bootloader in `Features`


matejcik:
ideally from the bootloader too?

PR #3270, opened 2023-09-14
Add bootloader_locked to Features

andrewkozlik:
Resolves https://github.com/trezor/trezor-firmware/issues/3268, as discussed [here](https://github.com/trezor/trezor-firmware/pull/3255#discussion_r1324385328).

Adds `bootloader_locked` to `Features` message for Trezor Model R in both firmware and bootloader.

If the bootloader is unlocked, then `AuthenticateDevice` will return `ProcessError("Bootloader unlocked.")` before asking the user to confirm device authenticity check.

### DIFF
--- a/core/embed/bootloader/messages.c
+++ b/core/embed/bootloader/messages.c
@@ -44,6 +44,10 @@
 #include "emulator.h"
 #endif
 
+#if USE_OPTIGA
+#include "secret.h"
+#endif
+
 #define MSG_HEADER1_LEN 9
 #define MSG_HEADER2_LEN 1
 
@@ -311,6 +315,11 @@ static void send_msg_features(uint8_t iface_num,
     MSG_SEND_ASSIGN_VALUE(unit_color, unit_variant_get_color());
     MSG_SEND_ASSIGN_VALUE(unit_btconly, unit_variant_get_btconly());
   }
+
+#if USE_OPTIGA
+  MSG_SEND_ASSIGN_VALUE(bootloader_locked,
+                        (secret_bootloader_locked() == sectrue));
+#endif
   MSG_SEND(Features);
 }
 
--- a/core/src/apps/management/authenticate_device.py
+++ b/core/src/apps/management/authenticate_device.py
@@ -11,9 +11,12 @@ async def authenticate_device(msg: AuthenticateDevice) -> AuthenticityProof:
     from trezor.crypto.hashlib import sha256
     from trezor.messages import AuthenticityProof
     from trezor.ui.layouts import confirm_action
-    from trezor.utils import BufferReader
+    from trezor.utils import BufferReader, bootloader_locked
 
     from apps.common.writers import write_compact_size
+
+    if not bootloader_locked():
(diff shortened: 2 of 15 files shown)


core/embed/bootloader/messages.c
```diff
@@ -44,6 +44,10 @@
 #include "emulator.h"
 #endif
 
+#if defined(TREZOR_MODEL_R) && !defined(TREZOR_EMULATOR)
```
matejcik, line 47: why is emulator excluded here? it is supposed to work and recognize whether it is unlocked

core/src/apps/management/authenticate_device.py
```diff
@@ -11,10 +11,13 @@ async def authenticate_device(msg: AuthenticateDevice) -> AuthenticityProof:
     from trezor.crypto.hashlib import sha256
     from trezor.messages import AuthenticityProof
     from trezor.ui.layouts import confirm_action
-    from trezor.utils import BufferReader
+    from trezor.utils import BufferReader, bootloader_locked
 
     from apps.common.writers import write_compact_size
 
+    if not bootloader_locked():
+        raise wire.ProcessError("Bootloader unlocked.")
```
matejcik, line 19:
```suggestion
        raise wire.ProcessError("Cannot authenticate when bootloader is unlocked.")
```
(otherwise it sounds like this action caused the bootloader to unlock)

core/embed/bootloader/messages.c
andrewkozlik: I assumed that the emulator should always return true, but I hadn't realized that it sets up the memory to indicate that there is a secret stored in the corresponding sector. However, the firmware emulator doesn't seem to be doing that.

matejcik: Right, that might be a missing piece. I think I intended to add a cmdline option to the bootloader-emulator that would write a secret header.

For firmware emulator, i would leave it to always-true for now, but it seems worth making it work for the bootloader emulator.

andrewkozlik: Correction, neither of them sets up the memory to indicate that there is a secret stored in the corresponding sector. So both will always return false and won't firmware allow authenticity checks if we use secret.c in the emulator. Of course we can include secret.h without using it, if that's what you mean in this comment.

andrewkozlik: 55238596fa265acc17e68d573bd4ac37d1cee290

core/src/apps/management/authenticate_device.py
andrewkozlik: 55238596fa265acc17e68d573bd4ac37d1cee290

andrewkozlik:
After defining `USE_OPTIGA` in 55238596fa265acc17e68d573bd4ac37d1cee290, `trezorctl device unlock-bootloader` works with the emulator.

matejcik:
(sorry about the forcepush, restored to your commit now)

matejcik (approved):

Issue #3271, opened 2023-09-14
Improvements on chunkifying addresses - spaces


Hannsek:
https://github.com/trezor/trezor-firmware/pull/3237#issuecomment-1719145612

- [ ] Fix the _next screen_ arrow
![image](https://github.com/trezor/trezor-firmware/assets/57008159/ba63cd33-614a-47d2-88d1-14007b4dff40)

- [ ] Fix chunks on the second screen. The first chunk of the second screen should be placed above the second chunk on the second row
![image](https://github.com/trezor/trezor-firmware/assets/57008159/f08186c5-09e0-44d5-849b-5e37a6a46c9b)

Issue #3272, opened 2023-09-14
Improvements on chunkifying addresses - alignment


Hannsek:
https://github.com/trezor/trezor-firmware/pull/3237#issuecomment-1719145612

We should be able to determine the widest row and make the spaces between chunks according to that row. So that this won't happen.
![image](https://github.com/trezor/trezor-firmware/assets/57008159/f7034ea5-86f3-4ff7-bfbf-2ce51a58f471)

PR #3273, opened 2023-09-14
Fix chunkify arrows on T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3271
- makes sure the arrow will always occupy the whole last/first chunk when paginating


core/embed/rust/src/ui/component/text/layout.rs
```diff
@@ -264,7 +264,15 @@ impl TextLayout {
         {
             sink.prev_page_ellipsis(*cursor, self);
             // Move the cursor to the right, always the same distance
-            cursor.x += self.style.prev_page_ellipsis_width();
+            // Special case in chunkifying text - move the cursor so that we
+            // start with the second chunk.
+            if let Some(chunk_config) = self.style.chunks {
+                // Assuming mono-font, so all the letters have the same width
```
Hannsek, line 270: That is not right, right? Regarding **w** and **m**.

grdddj:
Looks like this:

image

Noticed just one place where the next page arrow is not completely visible - not sure why, would need to investigate this special case:

image

core/embed/rust/src/ui/component/text/layout.rs
grdddj: It does not matter here, we just want to offset the first chunk and the space (which combined should always be constant). 

(m, M, w and W are accounted for elsewhere, we are decreasing the space between chunks when we encounter them)

Hannsek:
Cannot we put the "next-page" arrow to the left side of the last chunk and not the right? The same with the "previous-page" arrow. Put it on the right side instead of left.

core/embed/rust/src/ui/component/text/layout.rs
Hannsek: Wouldn't it change with [this](https://github.com/trezor/trezor-firmware/issues/3272)?

grdddj: I am little confused with this second issue you mention here - what is the goal there?

grdddj:
> Cannot we put the "next-page" arrow to the left side of the last chunk and not the right? The same with the "previous-page" arrow. Put it on the right side instead of left.
> 
> image

Right, corrected in the current state:

image

Hannsek:
It does look ok. 👍

Hannsek:
@matejcik review pls

matejcik (approved):
utACK

bosomt:
**QA OK**

**Info:**
 - Suite version: web 24.1.0 (50fe764c31b889a043eb335c4f7421bcb8d8541b)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:120.0) Gecko/20100101 Firefox/120.0
 - OS: MacIntel
 - Screen: 1915x1244
 - Device: Trezor T2B1 2.6.4 regular (revision 1f693d3561c375661563970e066f678590abc19c)

PR #3275, opened 2023-09-14
disable access to secret in flash after boot

TychoVrahe:
disable access to secret in flash after boot.


hiviah:
ACK from me as well.

PR #3276, opened 2023-09-14
T2B1 sending same button requests as T2T1

grdddj:
Fixing the issue that in `confirm_output`, `TT` was sending two button requests, while `TR` only one.

It was caused by `TR` using only one `Rust` layout for both address and amount, while `TT` is using two calls from `micropython`.

Unified by splitting `Rust::TR::confirm_output` into `confirm_output_address` and `confirm_output_amount` and sending appropriate `button_requests` when spawning these layouts.

A lot of tests changed because of this, and there is UI diff for `TR`, as we can't anymore show the scrollbar, and also in `device_tests`, it does not paginate, so the pressed-button-state is not recorded.


mmilata (approved):
LGTM

PR #3277, opened 2023-09-14
add optiga to emulator for T2B1

TychoVrahe:
fixes at least non functioning unlock-bootloader in bootloader emulator.


matejcik:
there's also https://github.com/trezor/trezor-firmware/pull/3270, pick which approach is better (likely this one?)

matejcik (approved):

PR #3285, opened 2023-09-18
skip signature check for make_upload commands

TychoVrahe:
Since the recent vendor header changes, these commands were not usable with dev bootloader. I think that this is mostly used by devs so the check is not important, and the shortcut is handy.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -349,10 +349,10 @@ combine: ## combine boardloader + bootloader + prodtest into one combined image
 		$(PRODTEST_BUILD_DIR)/combined.bin
 
 upload: ## upload firmware using trezorctl
-	trezorctl firmware_update -f $(FIRMWARE_BUILD_DIR)/firmware.bin
+	trezorctl firmware_update -s -f $(FIRMWARE_BUILD_DIR)/firmware.bin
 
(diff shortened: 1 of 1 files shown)


Hannsek:
What does this shortcut do exactly?

TychoVrahe:
uploads the latest local build of firmware/prodtest to the device via trezorctl

Hannsek:
Sure, but I meant the `-s`

TychoVrahe:
thats for skipping the signature validation

Hannsek:
Ok, then I am confused, because I sometimes use `trezorctl firmware_update -f "path/to/the/firmware.bin" -s`. 
What this pr adds?

TychoVrahe:
This PR adds the `-s` flag to the `make upload` command.

If you want to upload the latest local build, you can type `trezorctl firmware_update -f "./build/firmware/firmware.bin" -s` or you can type `make upload` which is faster to type. But currently the `make upload` does not work with dev bootloader because it fails the signature check, so this PR makes it work again.

matejcik (approved):

Hannsek:
> This PR adds the -s flag to the make upload command.

Ahh, sure, thanks for explaining.

Issue #3260, opened 2023-09-12
Change homescreen text on T2B1


Hannsek:
Change homescreen screen on T2B1 should look like this. Same logic as on TT; preview of the homescreen on the screen.

![Image](https://github.com/trezor/trezor-firmware/assets/57008159/2e60c9ea-d740-4dad-8e42-9ccd68d6f00e)

PR #3287, opened 2023-09-19
feat(core): show new homescreen preview for T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3260:
- shows preview of the uploaded image

image


Hannsek:
From the images above it looks ok. Thanks! 👍

grdddj:
@mmilata please look at the new `Rust` code, it should be quick

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -392,6 +412,19 @@ extern "C" fn new_confirm_properties(n_args: usize, args: *const Obj, kwargs: *m
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_confirm_homescreen(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let title: StrBuffer = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
+        let data: Obj = kwargs.get(Qstr::MP_QSTR_image)?;
+
+        let buffer = unsafe { unwrap!(get_buffer(data)) };
```
mmilata, line 420: You can't keep a reference to the buffer and return to micropython. This can be worked around with closures, see `new_confirm_homescreen` in `ui/model_tt/layout.rs`.

mmilata (changes requested):
LGTM except the buffer handling.

core/embed/rust/src/ui/model_tr/layout.rs
grdddj: Right, changed in [06a3e8b](https://github.com/trezor/trezor-firmware/pull/3287/commits/06a3e8bdb4e93732942ef533e4e68fbda42e5ce9)

What things could happen when we would not do this?

mmilata: Due to how buffers work in micropython, the garbage collector would not understand that we have a reference to the buffer and might deallocate it while in use by rust code.

Also `(b)` is violated here because we cannot guarantee that python code does not touch the buffer once we return to micropython:
> SAFETY: The caller is responsible for ensuring immutability of the returned buffer, in particular that: ... (b) the buffer is not modified in MicroPython while the reference to it is being held.

mmilata:
Got rid of some of the lifetimes in 7c73feb8ce74798efe263697538e7c10aea81637, PTAL

mmilata (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 23.10.1 (0b3ccfe0b8e9bc2af00609d640bc1be31ab8edaa)
 - OS: MacIntel
 - Screen: 1440x900

Issue #3264, opened 2023-09-12
T2B1 Hold "delete" to delete everything


Hannsek:
In flows where user types something (PIN, recovery, …) there is "delete" button. 
If user long-presses the "delete" button, it should delete everything what has been typed. (The same behaviour as with TT PIN)

PR #3290, opened 2023-09-19
feat(core/rust): differentiate between short and long button press

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3264:
- allows for longer `DELETE` press to delete the whole inputted content, not just one character - in `PIN` or `passphrase` inputs

### DIFF
--- a/core/embed/rust/src/ui/model_tr/component/button_controller.rs
+++ b/core/embed/rust/src/ui/model_tr/component/button_controller.rs
@@ -3,7 +3,7 @@ use super::{
 };
 use crate::{
     strutil::StringType,
-    time::Duration,
+    time::{Duration, Instant},
     ui::{
         component::{base::Event, Component, EventCtx, Pad, TimerToken},
         event::{ButtonEvent, PhysicalButton},
@@ -40,7 +40,8 @@ enum ButtonState {
 
 pub enum ButtonControllerMsg {
     Pressed(ButtonPos),
-    Triggered(ButtonPos),
+    /// Which button was triggered, and whether it was pressed for a longer time
+    Triggered(ButtonPos, bool),
 }
 
 /// Defines what kind of button should be currently used.
@@ -105,6 +106,11 @@ where
 {
     pos: ButtonPos,
     button_type: ButtonType<T>,
+    /// Holds the timestamp of when the button was pressed.
+    pressed_since: Option<Instant>,
+    /// How long the button should be pressed to send `long_press=true` in
+    /// `Triggered`
+    long_press_ms: u32,
 }
 
 impl<T> ButtonContainer<T>
@@ -114,9 +120,12 @@ where
     /// Supplying `None` as `btn_details`  marks the button inactive
     /// (it can be later activated in `set()`).
     pub fn new(pos: ButtonPos, btn_details: Option<ButtonDetails<T>>) -> Self {
+        const DEFAULT_LONG_PRESS_MS: u32 = 1000;
         Self {
             pos,
             button_type: ButtonType::from_button_details(pos, btn_details),
+            pressed_since: None,
+            long_press_ms: DEFAULT_LONG_PRESS_MS,
         }
     }
 
@@ -151,7 +160,14 @@ where
     /// hold.
     pub fn maybe_trigger(&mut self, ctx: &mut EventCtx) -> Option<ButtonControllerMsg> {
         match self.button_type {
-            ButtonType::Button(_) => Some(ButtonControllerMsg::Triggered(self.pos)),
+            ButtonType::Button(_) => {
+                // Finding out whether the button was long-pressed
+                let long_press = self.pressed_since.map_or(false, |since| {
+                    Instant::now().saturating_duration_since(since).to_millis() > self.long_press_ms
+                });
+                self.pressed_since = None;
+                Some(ButtonControllerMsg::Triggered(self.pos, long_press))
+            }
             _ => {
                 self.hold_ended(ctx);
                 None
@@ -167,6 +183,11 @@ where
             }
         }
         false
+    }
+
+    /// Saving the timestamp of when the button was pressed.
+    pub fn got_pressed(&mut self) {
+        self.pressed_since = Some(Instant::now());
     }
 
     /// Registering hold event.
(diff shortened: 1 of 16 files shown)


core/embed/rust/src/ui/model_tr/component/button_controller.rs
```diff
@@ -40,7 +40,8 @@ enum ButtonState {
 
 pub enum ButtonControllerMsg {
     Pressed(ButtonPos),
-    Triggered(ButtonPos),
+    /// Which button was triggered, and whether it was pressed for a longer time
+    Triggered((ButtonPos, bool)),
```
mmilata, line 44:
```suggestion
    Triggered(ButtonPos, bool),
```

core/embed/rust/src/ui/model_tr/component/button_controller.rs
```diff
@@ -117,6 +123,8 @@ where
         Self {
             pos,
             button_type: ButtonType::from_button_details(pos, btn_details),
+            pressed_since: None,
+            long_press_ms: 1000,
```
mmilata, line 127: Constant pls?

mmilata (changes requested):
LGTM, got 2 nits.

core/embed/rust/src/ui/model_tr/component/button_controller.rs
grdddj: Done in [8d9c7cb](https://github.com/trezor/trezor-firmware/pull/3290/commits/8d9c7cbe5b182269857a71fb39862f6aca9004e5)

core/embed/rust/src/ui/model_tr/component/button_controller.rs
grdddj: Done in [38a17f4](https://github.com/trezor/trezor-firmware/pull/3290/commits/38a17f47b34cb458ea6784b284da0d09192d61fc)

Hannsek:
Can the row with symbols be deleted without releasing the buttons? So that the rows of asterisks will be deleted when hodling "delete" button. So after some time, the buttons "would be released" even without physically releasing the buttons. Does it make sense?

grdddj:
> Can the row with symbols be deleted without releasing the buttons? So that the rows of asterisks will be deleted when hodling "delete" button. So after some time, the buttons "would be released" even without physically releasing the buttons. Does it make sense?

It can, but I am afraid that it would complicate things quite a bit - as currently we are always waiting for the physical release. I suppose we would this behavior only here in `DELETE` button example and not anywhere else - e.g. when just scrolling the paginated page (holding the right for longer should not paginate there).

Will think about the implementation

Hannsek:
Yes, that is right, only here with "delete".

Aren't we using it with "hold to move" feature?

mmilata (approved):
LGTM code-wise

grdddj:
> Yes, that is right, only here with "delete".
> 
> Aren't we using it with "hold to move" feature?

Hold-to-move is implemented little differently.

Done in [112c6bb](https://github.com/trezor/trezor-firmware/pull/3290/commits/112c6bb21374e2cc9984fd35ffc20a744c12a552) - the functionality to trigger the event without releasing and [112c6bb](https://github.com/trezor/trezor-firmware/pull/3290/commits/112c6bb21374e2cc9984fd35ffc20a744c12a552) - configuring the `ChoiceItems` so that this behavior is only valid for chosen (`DELETE`) items

@mmilata please re-review. I am aware that in the current status, we theoretically do not need the `bool` (long-pressed) in  `Triggered(ButtonPos, bool)` - but I would still keep it for some future requirements, where we would want to wait for the release and still evaluate the long-press (this information is almost free).

core/embed/rust/src/ui/model_tr/component/button_controller.rs
```diff
@@ -105,6 +110,13 @@ where
 {
     pos: ButtonPos,
     button_type: ButtonType<T>,
+    /// Holds the timestamp of when the button was pressed.
+    pressed_since: Option<Instant>,
+    /// How long the button should be pressed to send `long_press=true` in
+    /// `ButtonControllerMsg::Triggered`
+    long_press_ms: u32,
+    /// Timer for sending `ButtonControllerMsg::LongPressed`
+    long_pressed_timer: Option<TimerToken>,
```
mmilata, line 119: Nit: it would kinda make sense to keep both in a single option, i.e. `Option<(Instant, TimerToken)>` since they're not really independent.

mmilata (approved):

Hannsek:
It now works strange. After the long press, sometimes it deletes all characters and stays on the delete button, sometimes it deletes all characters and jump elsewhere and sometimes it does delete all characters but only after release of the buttons.

grdddj:
> It now behaves inconsistently. After the long press, sometimes it deletes all characters and stays on the delete button, sometimes it deletes all characters and jump elsewhere and sometimes it does delete all characters but only after release of the buttons.

I cannot replicate any of these on the emulator. Where do you try it, and did you notice some pattern of these 3 possibilities?

Hannsek:
I tried it on locally built emulator. No pattern. Will try it again today.

grdddj:
The new approach broke click tests, the reason was overflowing the `loop::_queue` with `Timer` events when a lot of buttons were pressed in a short time frame (each press is spawning a new `TimerEvent`).

I have "fixed" it for testing emulator in [157ac3e](https://github.com/trezor/trezor-firmware/pull/3290/commits/157ac3e976124fe246eacd04e78efe516e86c7f2), but very theoretically it might also be an issue on the hardware device.

Further, considering the instabilities @Hannsek is seeing, I am tempted to be rather defensive and remove the "trigger without release" functionality - unless we really test it well and make sure it does not have any dangerous side-effects. The benefit of it seems marginal (most of the users will never have this use-case IMO).

One possibility is merging the state of deleting with release and then test/troubleshoot/postpone the deleting without release as a separate issue/PR

Hannsek:
I agree with merging the state of deleting with release and postpone the other functionality to later. Please create another PR.🙏🏻

mmilata:
> overflowing the loop::_queue with Timer events when a lot of buttons were pressed in a short time frame

These are mostly generated by buttons that don't even react to long press, right? We probably should opt-in to the long press when constructing the button instead of adding timers for every button.

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1440x900

PR #3292, opened 2023-09-20
add basic support for efficient frame buffer graphics

TychoVrahe:
Adjusts the dma2d operations to work with internal framebufffer, some basic rendering functions are modified to be more efficient with FB: bar/rect, rect_rounded, icon, text. 

Bar/Rect rendering including the rounded variants are rewritten to rust. (bar implementation in C stays too, as its used by prodtest which does not use Rust yet)


TychoVrahe:
@mmilata perhaps?

core/embed/rust/src/trezorhal/display.rs
```diff
@@ -74,51 +80,51 @@ pub fn text_baseline(font: i32) -> i16 {
...
-            bgcolor,
-            radius,
-        )
+        ffi::DISPLAY_DATA_ADDRESS.write_volatile(c);
     }
 }
 
-pub fn bar_radius_buffer(x: i16, y: i16, w: i16, h: i16, radius: u8, buffer: &mut BufferText) {
+#[cfg(feature = "framebuffer")]
+pub fn get_fb_addr() -> u32 {
```
mmilata, line 92: Returning integer seems wrong. Let's use something like
```
struct FrameBuffer(*mut u32)
```

Is there a situation where we'd want to work with multiple framebuffers? If not perhaps we can keep the pointer private to this module?

core/embed/rust/src/trezorhal/display.rs
```diff
@@ -74,51 +80,51 @@ pub fn text_baseline(font: i32) -> i16 {
...
+#[cfg(all(feature = "framebuffer", not(feature = "framebuffer32bit")))]
+pub fn pixel(fb: u32, x: i16, y: i16, c: u16) {
     unsafe {
-        ffi::DISPLAY_DATA_ADDRESS.write_volatile(c);
+        let addr = fb
+            + ((y as u32 + DISPLAY_FRAMEBUFFER_OFFSET_Y) * DISPLAY_FRAMEBUFFER_WIDTH
+                + (x as u32 + DISPLAY_FRAMEBUFFER_OFFSET_X))
+                * 2;
+        let ptr = addr as *mut u16;
+        ptr.write_volatile(c);
```
mmilata, line 114: I think the unsafe block only needs to be around this line?

core/embed/rust/src/trezorhal/display.rs
```diff
@@ -74,51 +80,51 @@ pub fn text_baseline(font: i32) -> i16 {
...
+pub fn pixel(fb: u32, x: i16, y: i16, c: u32) {
     unsafe {
-        ffi::DISPLAY_DATA_ADDRESS.write_volatile((c & 0xff) as u8);
-        ffi::DISPLAY_DATA_ADDRESS.write_volatile((c >> 8) as u8);
+        let addr = fb
+            + ((y as u32 + DISPLAY_FRAMEBUFFER_OFFSET_Y) * DISPLAY_FRAMEBUFFER_WIDTH
+                + (x as u32 + DISPLAY_FRAMEBUFFER_OFFSET_X))
+                * 4;
+        let ptr = addr as *mut u32;
+        ptr.write_volatile(c);
```
mmilata, line 127: I think the unsafe block only needs to be around this line?

core/embed/rust/src/ui/display/mod.rs
```diff
@@ -78,34 +86,218 @@ pub fn fade_backlight(_: u16) {}
 #[cfg(not(feature = "backlight"))]
 pub fn fade_backlight_duration(_: u16, _: u32) {}
 
+#[cfg(not(feature = "framebuffer"))]
 /// Fill a whole rectangle with a specific color.
 pub fn rect_fill(r: Rect, fg_color: Color) {
-    display::bar(r.x0, r.y0, r.width(), r.height(), fg_color.into());
+    let r = r.translate(get_offset());
+    let r = r.translate(get_offset());
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/a7d30a27369d7db238a55ed346b7388da4251619

core/embed/rust/src/ui/display/mod.rs
```diff
@@ -78,34 +86,218 @@ pub fn fade_backlight(_: u16) {}
...
+    15, 15, 15, 15, 15, 15, 15, 15, 0, 0, 0, 0, 3, 14, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 0,
+    0, 0, 3, 14, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 0, 0, 0, 12, 15, 15, 15, 15, 15, 15,
+    15, 15, 15, 15, 15, 15, 0, 0, 8, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 0, 3, 15,
+    15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 0, 9, 15, 15, 15, 15, 15, 15, 15, 15, 15,
+    15, 15, 15, 15, 15, 1, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 5, 15, 15,
+    15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 9, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15,
+    15, 15, 15, 15, 15, 12, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 14, 15, 15,
+    15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15,
+    15, 15, 15, 15, 15,
+];
```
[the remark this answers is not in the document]
TychoVrahe: this is cool, now it actually looks like the corner if you look at the table from a distance :)

https://github.com/trezor/trezor-firmware/commit/4cc9dca08ce68e991ea89f9e5cd0efb3043e9d12

core/embed/trezorhal/stm32f4/displays/ltdc.c
```diff
@@ -371,3 +370,47 @@ void display_sync(void) {}
...
+  // start of next line
+  pixels = pixels - line_rem;
+  cursor_x = window_x0;
+  cursor_y++;
+
+  // add the rest of pixels
+  cursor_y += pixels / w;
+  cursor_x += pixels % w;
+
+  // todo check if we are out of bounds height wise
```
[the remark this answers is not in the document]
TychoVrahe: wrapped around https://github.com/trezor/trezor-firmware/commit/822837f2b917f906c0716bf1e0a971efbd585b0d

core/embed/lib/display_interface.h
```diff
@@ -24,11 +24,28 @@
...
+
 #ifndef PIXELDATA
 #define PIXELDATA(c) display_pixeldata(c)
 #endif
 
 void display_pixeldata(uint16_t c);
+void display_pixeldata_dirty(void);
 
 #ifndef PIXELDATA_DIRTY
 #define PIXELDATA_DIRTY()
```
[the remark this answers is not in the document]
TychoVrahe: actually i don't think we need the macro. its called after some rendering, so not performance critical and its a bit cleaner without it: https://github.com/trezor/trezor-firmware/commit/6569e787752fda3f595087a14d12f8ee4fcecd96

core/embed/rust/src/trezorhal/display.rs
TychoVrahe: we will most likely want to use 2 framebuffers, but probably not simultaneously. 

Problem here is that need to use the address in the inlined `pixel` function, and there i don't want to call the ffi as it would decrease performance.

added the struct here: https://github.com/trezor/trezor-firmware/commit/c37f89d7113ecd773327d2840223e44e7edd829c

core/embed/rust/src/trezorhal/display.rs
TychoVrahe: using offset function now on the pointer, so that needs unsafe too: https://github.com/trezor/trezor-firmware/commit/c37f89d7113ecd773327d2840223e44e7edd829c

core/embed/rust/src/trezorhal/display.rs
TychoVrahe: using offset function now on the pointer, so that needs unsafe too: https://github.com/trezor/trezor-firmware/commit/c37f89d7113ecd773327d2840223e44e7edd829c

mmilata (approved):

Issue #3291, opened 2023-09-20
Do not land on the same letter when doing recovery


Hannsek:
There are instances where, if a user selects a specific letter, the cursor position may remain unchanged. This feels buggy. Therefore, restrict consecutive selections of the same letter.

PR #3295, opened 2023-09-21
Recovery shuffle letters in T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3291:
- makes sure the letter always changes after selection in recovery flow

### DIFF
--- a/core/embed/rust/src/ui/model_tr/component/input_methods/wordlist.rs
+++ b/core/embed/rust/src/ui/model_tr/component/input_methods/wordlist.rs
@@ -157,19 +157,49 @@ where
         ChoiceFactoryWordlist::new(self.wordlist_type, self.textbox.content())
     }
 
+    fn get_last_textbox_letter(&self) -> Option<char> {
+        self.textbox.content().chars().last()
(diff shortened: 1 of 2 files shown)


Hannsek:
I briefly tried it and it seems to work. 👍🏻

core/embed/rust/src/ui/model_tr/component/input_methods/wordlist.rs
```diff
@@ -167,7 +171,32 @@ where
...
+                        if let Some(last_letter) = self.get_last_textbox_letter() {
+                            if current_letter == last_letter {
+                                // Randomly trying again when the last and current letter match
+                                continue;
+                            }
+                        }
+                    }
+                    break random_position;
+                }
+            }
```
mmilata, line 199: The function's getting quite long, would it be possible to factor out some of the logic into a function? One possibility would be something like `get_random_position_except(choices_count, current_index)` though i don't know if it's easy to obtain the index here ...

mmilata (approved):

PR #3296, opened 2023-09-21
PIN verification using Optiga

andrewkozlik:
Creating this draft PR, so we can start discussion about the cryptographic aspects of the PIN verification process using Optiga. @onvej-sl 

I did some refactoring around the `ui_progress`, which as it turns out is not really necessary, since we decided to set `PIN_STRETCH_ITERATIONS` to 1, so it takes no noticeable time (unless the security monitor kicks in). But I think the refactor still has some value, so I decided to keep it. It might be used in optiga.c in the future if we have our own configuration of the security monitor which will allow us to increase `PIN_STRETCH_ITERATIONS`. For now I removed some of the unused code in a [separate commit](https://github.com/trezor/trezor-firmware/pull/3296/commits/b5d178e1e14436d002fb740dcfdc1cceafa4426c), so that it can be reverted when the time comes.


core/embed/trezorhal/optiga/optiga.c
```diff
@@ -109,3 +158,432 @@ bool optiga_random_buffer(uint8_t *dest, size_t size) {
...
+  metadata.change = ACCESS_PIN_SECRET;
+  metadata.read = OPTIGA_META_ACCESS_NEVER;
+  metadata.execute = ACCESS_PIN_COUNTER;
+  metadata.data_type = TYPE_AUTOREF;
+  if (!optiga_set_metadata(OID_SALTED_PIN, &metadata)) {
+    return false;
+  }
+
+  // TODO consider adding LUC execute access condition for the stretching
+  // secrets.
```
andrewkozlik, line 264: I am not sure this is worth doing, considering that Optiga already has the security monitor in place. This could be more trouble than it's worth.

storage/storage.c
```diff
@@ -472,79 +465,155 @@ static secbool is_not_wipe_code(const uint8_t *pin, size_t pin_len) {
...
+  if (ext_salt != NULL) {
+    memcpy(salt + salt_len, ext_salt, EXTERNAL_SALT_SIZE);
+    salt_len += EXTERNAL_SALT_SIZE;
+  }
+
+  PBKDF2_HMAC_SHA256_CTX ctx = {0};
+  pbkdf2_hmac_sha256_Init(&ctx, optiga_secret, OPTIGA_PIN_SECRET_SIZE, salt,
+                          salt_len, 1);
+  // TODO Do we even need this one iteration or can we use 0 or completely leave
+  // out pbkdf2_hmac_sha256_Update()?
```
andrewkozlik, line 547: I haven't had time to look into this yet. Opinions @onvej-sl?

storage/storage.c
```diff
@@ -472,79 +465,155 @@ static secbool is_not_wipe_code(const uint8_t *pin, size_t pin_len) {
...
+static void derive_kek_set(const uint8_t *pin, size_t pin_len,
+                           const uint8_t *storage_salt, const uint8_t *ext_salt,
+                           uint8_t kek[SHA256_DIGEST_LENGTH],
+                           uint8_t keiv[SHA256_DIGEST_LENGTH]) {
+#if USE_OPTIGA
+  uint8_t optiga_secret[OPTIGA_PIN_SECRET_SIZE] = {0};
+  // TODO consider completely skipping optiga_pin_set() and optiga_pin_verify()
+  // when pin_len == 0. It adds no security in this case. If the Optiga goes
+  // bust, then the user can at least use their Trezor without PIN to transfer
+  // their funds elsewhere for example.
```
andrewkozlik, line 570: I would like to do this. Opinions @onvej-sl?

core/embed/trezorhal/optiga/optiga_transport.c
```diff
@@ -137,11 +137,32 @@ static size_t sec_chan_size = 0;
...
+  if (log_hex != NULL) {                                                     \
+    static uint8_t prev_data[4];                                             \
+    static size_t prev_size = 0;                                             \
+    static int repeated = 0;                                                 \
+    if (prev_size == data_size && memcmp(data, prev_data, data_size) == 0) { \
+      repeated += 1;                                                         \
+    } else {                                                                 \
+      if (repeated > 0) {                                                    \
+        char ext_prefix[25];                                                 \
+        snprintf(ext_prefix, sizeof(ext_prefix), "%sREPEATED (%d) ", prefix, \
```
matejcik, line 153: fwiw we have a `mini_snprintf` that would be a better choice here if we want this code to go master repo

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -19,8 +19,57 @@
...
+#define OID_PIN_CMAC OPTIGA_OID_SYM_KEY
+
+// Key for ECDH PIN derivation step (OID 0xE0F3).
+#define OID_PIN_ECDH (OPTIGA_OID_ECC_KEY + 3)
+
+// The number of times optiga_pin_stretch_secret() is called in PIN processing.
+#define PIN_STRETCH_PHASE_COUNT 2
+
+// The number of times the stretching is repeated in each PIN processing phase.
+#define PIN_STRETCH_ITERATIONS 5
```
andrewkozlik, line 48: We need to decide on the exact number of iterations. Ideally this would be set so that it takes about one second to complete the PIN verification. The problem is that one PIN verification causes `2 + 6 * PIN_STRETCH_ITERATIONS` security events. That translates to `12 * PIN_STRETCH_ITERATIONS - 6` writes to the security event counter (SEC), assuming the default configuration and that SEC credit is at its maximum when the PIN verification begins. The endurance of the SEC is 1600000 writes. So if we set this constant to `PIN_STRETCH_ITERATIONS=2`, then that allows for a maximum of 88888 PIN verifications over the lifetime of the device. That seems perfectly acceptable. In terms of lifetime, this is equivalent to setting the LUC counter limit [here](https://github.com/trezor/trezor-firmware/pull/3296/files#r1333205454) to `6 * PIN_STRETCH_ITERATIONS * 88888 = 1066656`.

Here are the results for more `PIN_STRETCH_ITERATIONS` values:

`PIN_STRETCH_ITERATIONS` | Max. PIN verifications | Equivalent LUC limit
-- | -- | --
1 | 266,667 | 1,600,000
2 | 88,889 | 1,066,667
3 | 53,333 | 960,000
4 | 38,095 | 914,286
5 | 29,630 | 888,889
6 | 24,242 | 872,727
7 | 20,513 | 861,538
8 | 17,778 | 853,333
9 | 15,686 | 847,059
10 | 14,035 | 842,105

As the table demonstrates, the endurance of the monotonic counter (600,000 writes) is actually not the limiting factor, so we might as well implement the LUC execute access condition for the stretching secrets.

andrewkozlik: Two more things to add:

- Note that there is a natural limit on the number of PIN verifications determined by the `OID_PIN_COUNTER` (600,000 writes). That's at most 300,000 successful unlock operations. Again the limiting factor is the SEC.
- There are situations where we currently unnecessarily involve Optiga in the PIN verification process, e.g. when changing the PIN we first verify and then set the PIN even if the storage is unlocked. I think the only part where it's necessary to involve the Optiga in setting the PIN, because the Optiga adds no security to the PIN verification when the storage is already unlocked. We just have to remember to increment `OID_PIN_COUNTER` if PIN verification fails. This should be handled in a separate issue.

storage/storage.c
andrewkozlik: Let's leave the code as is. One iteration takes place automatically in the Init function and `pbkdf2_hmac_sha256_Update(&ctx, 1);` is a actually a nop, but it expresses that one iteration has taken place, which is useful for documentation purposes.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -109,3 +158,432 @@ bool optiga_random_buffer(uint8_t *dest, size_t size) {
...
+  if (!optiga_pin_init_metadata() || !optiga_pin_init_stretch()) {
+    return false;
+  }
+
+  // Combine the PIN with salt from the flash storage before sending it to the
+  // Optiga. This ensures that in the unlikely case of an attacker recording
+  // communication between the MCU and Optiga, they will not gain knowledge of
+  // the PIN without at least having to conduct a brute-force search.
+  uint8_t salted_pin[OPTIGA_PIN_SECRET_SIZE] = {0};
+  hmac_sha256(pin, pin_size, storage_salt, STORAGE_SALT_SIZE, salted_pin);
```
onvej-sl, line 452: The time complexity of the attack is `2^(8 * STORAGE_SALT_SIZE + pin_entropy) = 2^(32 + pin_entropy)`, measured in the number of `HMAC-SHA256` computations.

The purpose of the secure element is to allow a user to safely use a short pin. However, since the salt has only 32 bits, the brute-force attack is feasible for short pins.

storage/storage.c
onvej-sl: This is an attack scenario in which the attacker benefits from the fact that the optiga is not used: The attacker has a flash dump but nothing else. In particular, the attacker has no physical access to the device.

What I don't know is how realistic this attack is.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: My assumption here is that an attacker that has physical access to the device knows the storage salt.

core/embed/trezorhal/optiga/optiga_transport.c
andrewkozlik: I suppose we don't need to show the exact count. Also it's hard to write it out clearly. So I simplified it https://github.com/trezor/trezor-firmware/pull/3296/commits/286acf0e7b9eb936074d154e4f83d98af4f59abf.

storage/storage.c
andrewkozlik: Interesting point, but the scenario seems unrealistic to me.

core/embed/trezorhal/optiga/optiga.c
onvej-sl: I don't understand what you mean. If you assume that the attacker knows the storage salt, then there is no reason to mix the salt with the pin.

onvej-sl: This thought occurred to me: If we want a Trezor with a malicious Optiga to be as secure as a Trezor without Optiga, then we have to replace the [HMAC](https://github.com/trezor/trezor-firmware/blob/a19732de622892fe5ff359f934203731d57b1af7/core/embed/trezorhal/optiga/optiga.c#L452) with [PBKDF](https://github.com/trezor/trezor-firmware/blob/a19732de622892fe5ff359f934203731d57b1af7/storage/storage.c#L489-L523). Otherwise, it is easier for the attacker to bruteforce the salted pin sent to the optiga than to bruteforce the pin verification code in the storage.

onvej-sl: The same applies to [this](https://github.com/trezor/trezor-firmware/blob/a19732de622892fe5ff359f934203731d57b1af7/core/embed/trezorhal/optiga/optiga.c#L520) HMAC.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: See https://github.com/trezor/trezor-firmware/pull/3296#discussion_r1334085131.

andrewkozlik:
@onvej-sl has reviewed the cryptographic solution and will do a final check tomorrow. I would like another set of eyes to go over the code, @matejcik please. I am hoping to get this merged tomorrow. The reason why it is in draft form is because it's branched off of the hashing-to-curve PR, which is not merged yet.

storage/storage.c
```diff
@@ -469,16 +476,10 @@ static secbool is_not_wipe_code(const uint8_t *pin, size_t pin_len) {
...
-  // elapsed_ms = number of elapsed milliseconds in the ongoing substep.
+  ui_rem -= elapsed_ms;
   if (ui_callback && ui_message) {
-    uint32_t progress = 0;
-    if (ui_total > 1000000) {
-      // Avoid overflow, precise enough.
-      progress = (ui_total - ui_rem) / (ui_total / 1000);
-    } else {
-      progress = ((ui_total - ui_rem) * 1000 + elapsed_ms) / ui_total;
-    }
```
andrewkozlik: I am not sure why we were ever worried about overflow here, since the largest possible time in milliseconds is 2^16 * 1000 which fits well within uint32. We'd have to exceed 2^22 seconds for this to be a problem.

andrewkozlik:
The last two commits https://github.com/trezor/trezor-firmware/pull/3296/commits/4419899eec453c4998401a0b52f7ffecc4a83360
https://github.com/trezor/trezor-firmware/pull/3296/commits/4ed316bd44effa2831ec1995c00d14276a39452c that I added today reintroduce `ui_progress()` into the Optiga code. They don't change anything about the functionality. The execution of the Optiga part takes about 1 second to complete, so this smooths out the progress a bit better.

storage/storage.c
andrewkozlik: It can overflow. I got lost in the change from measuring in seconds to milliseconds and the fact that progress is measured in units per mille. Fixed 5d7913436570052ef247a94c50f044964206d399.

onvej-sl (approved):

matejcik (approved):
LGTM code-wise, except the `progess(0)` thing you discovered

andrewkozlik:
> LGTM code-wise, except the `progess(0)` thing you discovered

Should be fixed by https://github.com/trezor/trezor-firmware/pull/3296/commits/38b331dba0fa9650e2b43fae99365c19024948f0

Issue #3282, opened 2023-09-18
"Account" key missing in altcoins receive flow


Hannsek:
T2B1 only?

<img width="263" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/d947a90a-b18d-49c9-9ff2-967df4e27b27">

grdddj:
Currently `account` for altcoins like `bnb` is empty - I think there is no concept like "segwit" or "taproot" in these altcoins - what do you want to display there?

matejcik:
"Account #1" ?

Hannsek:
Ethereum #1, when it is on ETH#1 account?
XRP #1
and so on.
Does it make sense?

grdddj:
> Ethereum #1, when it is on ETH#1 account? XRP #1 and so on. Does it make sense?

Right, so the name of the currency and ` #{num}`, that is alright

PR #3299, opened 2023-09-22
Add account number into address details for altcoins

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3282:
- adds the account number into detailed information about address or public key


Hannsek:
Nice, thanks!

core/src/apps/common/paths.py
```diff
@@ -372,3 +372,42 @@ def _path_item(i: int) -> str:
...
+
+    # Trying all possible patterns - at least one should match
+    for patt in pattern:
+        try:
+            return _get_account_num_single(address_n, patt)
+        except ValueError:
+            pass
+
+    # This function should not raise
+    return 0
```
mmilata, line 389: We should return None instead and not display account number at all in this case.

core/src/apps/common/paths.py
```diff
@@ -372,3 +372,42 @@ def _path_item(i: int) -> str:
...
+    # Validating address_n is compatible with pattern
+    # (need to have same length and same hardening)
+    address_str = address_n_to_str(address_n)
+    addr_parts = address_str.split("/")
+    patt_parts = pattern.split("/")
+    if len(addr_parts) != len(patt_parts):
+        raise ValueError
+    for add, pat in zip(addr_parts, patt_parts):
+        if add.count("'") != pat.count("'"):
+            raise ValueError
```
mmilata, line 402: Why not just call `PathSchema.parse(pattern).match(address_n)` here instead of reimplementing subset of it?

core/src/apps/nem/get_address.py
```diff
@@ -24,16 +24,19 @@ async def get_address(msg: NEMGetAddress, keychain: Keychain) -> NEMAddress:
...
 
     if msg.show_display:
+        from . import PATTERNS
+
         await show_address(
             address,
             case_sensitive=False,
-            path=address_n_to_str(address_n),
+            path=paths.address_n_to_str(address_n),
+            account=f"NEM #{paths.get_account_num(msg.address_n, list(PATTERNS))}",
```
mmilata, line 39: Nit: if the function accepted `Sequence[...]` instead of `list[...]` there would be no need to convert to list.

mmilata (changes requested):

core/src/apps/nem/get_address.py
grdddj: Right, changed in [b558b39](https://github.com/trezor/trezor-firmware/pull/3299/commits/b558b39a7128223d023b95140d0257bf93e1261f)

core/src/apps/common/paths.py
grdddj: Done in [b558b39](https://github.com/trezor/trezor-firmware/pull/3299/commits/b558b39a7128223d023b95140d0257bf93e1261f) - a whole new helper function doing this logic

core/src/apps/common/paths.py
grdddj: Right, done in [b558b39](https://github.com/trezor/trezor-firmware/pull/3299/commits/b558b39a7128223d023b95140d0257bf93e1261f) - I did not like this much as it needs `slip44_id` to be forwarded, and in `ETH` it is not even defined (using `address_n[1]`), but I agree it is better

core/src/apps/common/paths.py
```diff
@@ -374,32 +374,38 @@ def unharden(item: int) -> int:
     return item ^ (item & HARDENED)
 
 
-def get_account_num(address_n: list[int], pattern: str | list[str]) -> int:
+def get_account_name(
+    coin: str, address_n: Sequence[int], pattern: str | Sequence[str], slip44_id: int
```
mmilata, line 378: Nit: use `Bip32Path` for `address_n`.

core/src/apps/common/paths.py
```diff
@@ -374,32 +374,38 @@ def unharden(item: int) -> int:
...
+) -> str | None:
+    account_num = _get_account_num(address_n, pattern, slip44_id)
+    if account_num is None:
+        return None
+    return f"{coin} #{account_num}"
+
+
+def _get_account_num(
+    address_n: Sequence[int], pattern: str | Sequence[str], slip44_id: int
```
mmilata, line 387: Nit: use `Bip32Path` for `address_n`.

mmilata (approved):

core/src/apps/common/paths.py
grdddj: Done in the force-push

core/src/apps/common/paths.py
grdddj: Done in the force-push

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 23.10.0 (7e7e183598efd9ee3413d74f0689322811bbc1e9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/23.10.0 Chrome/112.0.5615.204 Electron/24.5.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1440x900
 - Device: Trezor T2B1 2.6.1 regular (revision 364c1ce123ead6e4dffab8cd3abb208e19158bce)

Issue #3204, opened 2023-08-10
Marquees in T2B1 behave inconsistently 


Hannsek:
Can be seen in this test, that the two marquees on multisig screens are not as smooth as the one on the main screen.

`INTERACT=1 pytest -s ../tests/device_tests/bitcoin/test_getaddress_show.py -k 'multisig_xpubs and xpubs3-True'`

PR #3300, opened 2023-09-22
Fix marquee in T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3204:
- makes marquee to always take 2 seconds to animate from end to end (pause of 1 second persists)
- @Hannsek not sure what are the exact requirements for "behaving consistently" - but this way it works fine with the multisig xpubs titles
- It might now work so well with much shorter or much longer texts, but we should not use these extensively

### DIFF
--- a/core/embed/rust/src/ui/component/marquee.rs
+++ b/core/embed/rust/src/ui/component/marquee.rs
@@ -10,7 +10,6 @@ use crate::{
     },
 };
 
-const MILLIS_PER_LETTER_M: u32 = 300;
 const ANIMATION_DURATION_MS: u32 = 2000;
 const PAUSE_DURATION_MS: u32 = 1000;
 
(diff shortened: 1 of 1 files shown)


Hannsek:
> makes marquee to always take 2 seconds to animate from end to end (pause of 1 second persists)

I believe this is the problem. It would go slowly in short titles and very quickly in long, right?

Hannsek:
Will take a look later today.

grdddj:
> > makes marquee to always take 2 seconds to animate from end to end (pause of 1 second persists)
> 
> I believe this is the problem. It would go slowly in short titles and very quickly in long, right?

Currently yes - constant time for all the lengths, therefore speed differs. Previously it was doing some calculation, which it seems did not account for non-mono font.

> Will take a look later today.

Thanks, let's figure out what logic we should use (something like constant speed for one letter)

mmilata (approved):
utACK

PR #3303, opened 2023-09-26
improve optiga communication reliability

TychoVrahe:
- adds delays to physical layer
- resets the optiga via pin during initialization

### DIFF
--- a/core/embed/firmware/main.c
+++ b/core/embed/firmware/main.c
@@ -300,6 +300,9 @@ void SVC_C_Handler(uint32_t *stack) {
       // raising privileges.
       stack[6] = (uintptr_t)reboot_to_bootloader;
       return;
+    case SVC_GET_SYSTICK_VAL: {
+      systick_val_copy = SysTick->VAL;
+    } break;
     default:
       stack[0] = 0xffffffff;
       break;
--- a/core/embed/trezorhal/stm32f4/i2c.c
+++ b/core/embed/trezorhal/stm32f4/i2c.c
@@ -171,9 +171,19 @@ HAL_StatusTypeDef i2c_transmit(uint16_t idx, uint8_t addr, uint8_t *data,
                                uint16_t len, uint32_t timeout) {
   return HAL_I2C_Master_Transmit(&i2c_handle[idx], addr, data, len, timeout);
 }
+
 HAL_StatusTypeDef i2c_receive(uint16_t idx, uint8_t addr, uint8_t *data,
                               uint16_t len, uint32_t timeout) {
-  return HAL_I2C_Master_Receive(&i2c_handle[idx], addr, data, len, timeout);
+  HAL_StatusTypeDef ret =
+      HAL_I2C_Master_Receive(&i2c_handle[idx], addr, data, len, timeout);
+#ifdef USE_OPTIGA
+  if (idx == OPTIGA_I2C_INSTANCE) {
+    // apply GUARD_TIME as specified by the OPTIGA datasheet
+    // (only applies to the I2C bus to which the OPTIGA is connected)
+    hal_delay_us(50);
+  }
+#endif
+  return ret;
 }
 
 HAL_StatusTypeDef i2c_mem_write(uint16_t idx, uint8_t addr, uint16_t mem_addr,
--- a/core/SConscript.bootloader_emu
+++ b/core/SConscript.bootloader_emu
@@ -136,6 +136,7 @@ if TREZOR_MODEL in ('R', ):
     ]
     SOURCE_TREZORHAL += [
         'embed/trezorhal/unix/secret.c',
+        'embed/trezorhal/unix/optiga_hal.c',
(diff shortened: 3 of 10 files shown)


core/embed/trezorhal/stm32f4/i2c.c
```diff
@@ -169,11 +169,25 @@ void i2c_cycle(uint16_t idx) {
 
 HAL_StatusTypeDef i2c_transmit(uint16_t idx, uint8_t addr, uint8_t *data,
                                uint16_t len, uint32_t timeout) {
-  return HAL_I2C_Master_Transmit(&i2c_handle[idx], addr, data, len, timeout);
+  HAL_StatusTypeDef ret =
+      HAL_I2C_Master_Transmit(&i2c_handle[idx], addr, data, len, timeout);
+#ifdef USE_OPTIGA
```
matejcik, line 174: seems like a bad place to put this. instead into `optiga_i2c_transmit` / receive?

TychoVrahe: Maybe. The receive 50us delay was added intentionally to i2c driver because datasheet says (or at least i interpret it this way) that the delay needs to be there even when communicating with other devices on the i2c bus. As the transmit delay is more experiment based, it might be ok to have it only after talking to optiga, but i don't know that for sure. Does not matter much for T2B1 as there is only optiga on the bus

matejcik: if that's the case, why the optiga guards?

TychoVrahe: The "GUARD_TIME" delay only needs to be there if optiga is on the bus. From datasheet: 
```
Note 1: For two consecutive accesses on the same device
GUARD_TIME re-specifies the value of tBUF as specified by [I2Cbus].
```
Basically what i think they are saying that i2c standard is not good enough when optiga is involved.

matejcik: that's in conflict with "the delay needs to be there even when communicating with other devices on the i2c bus"

either let's add guard time for all calls under `#ifdef USE_OPTIGA` (and add a comment explaining the above) but omit the device id check

or don't add it here but in `optiga_i2_transmit`

TychoVrahe: moved the transmission delay to optiga transport: https://github.com/trezor/trezor-firmware/pull/3303/commits/ff3e21121dd4616c098616c1a010cf3250fa0750

leaving 50us guard_time on the i2c driver level.

core/embed/firmware/main.c
```diff
@@ -300,6 +300,10 @@ void SVC_C_Handler(uint32_t *stack) {
       // raising privileges.
       stack[6] = (uintptr_t)reboot_to_bootloader;
       return;
+    case SVC_GET_SYSTICK_VAL: {
+      uint32_t *ret_ptr = (uint32_t *)stack[0];
```
matejcik, line 304: this looks insecure? allows unprivileged code to write to any address.
maybe we could return the value in some register or something?

core/embed/trezorhal/stm32f4/i2c.c
```diff
@@ -171,9 +171,17 @@ HAL_StatusTypeDef i2c_transmit(uint16_t idx, uint8_t addr, uint8_t *data,
...
   return HAL_I2C_Master_Transmit(&i2c_handle[idx], addr, data, len, timeout);
 }
+
 HAL_StatusTypeDef i2c_receive(uint16_t idx, uint8_t addr, uint8_t *data,
                               uint16_t len, uint32_t timeout) {
-  return HAL_I2C_Master_Receive(&i2c_handle[idx], addr, data, len, timeout);
+  HAL_StatusTypeDef ret =
+      HAL_I2C_Master_Receive(&i2c_handle[idx], addr, data, len, timeout);
+#ifdef USE_OPTIGA
+  if (idx == OPTIGA_I2C_INSTANCE) {
```
matejcik, line 180: oh, wait, `idx` is not a _device_ on a bus, but a _bus_ out of theoretical multiple ones, right?
please add a comment mentioning this

TychoVrahe: https://github.com/trezor/trezor-firmware/pull/3303/commits/52ce05eecebe1b1c6872e051008bd65b4ada94ab

core/embed/firmware/main.c
matejcik: (alternately just reserve a `static uint32_t` at a known location and the caller will need to take the systick val out of there?)

TychoVrahe: going with the static idea: https://github.com/trezor/trezor-firmware/pull/3303/commits/8f7ae5380721fe586edadfc4e0272bbe5d8b1fb3

matejcik (approved):
utACK

PR #3305, opened 2023-09-26
fix(python): drop simple-rlp dependency and use internal copy

prusnak:
supersedes https://github.com/trezor/trezor-firmware/pull/3046

not tested, please test before merging

### DIFF
--- a/python/src/trezorlib/_rlp.py
+++ b/python/src/trezorlib/_rlp.py
@@ -1,0 +1,55 @@
+# inspired by core/src/trezor/crypto/rlp.py
+
+import typing as t
+from collections.abc import Sequence
+
+if t.TYPE_CHECKING:
+    RLPItem = t.Union[t.Sequence["RLPItem"], bytes, int]
+
+
+def _byte_size(x: int) -> int:
+    if x < 0:
+        raise ValueError("only unsigned ints are supported")
+    return (x.bit_length() + 7) // 8
+
+
+def _int_to_bytes(n: int) -> bytes:
+    """Convert to a correctly sized bytes object."""
+    return n.to_bytes(_byte_size(n), "big")
+
+
+def _encode_with_length(value: bytes, header_byte: int) -> bytes:
+    length = len(value)
+    if length == 1 and value[0] <= 0x7F:
+        return value
+    elif length <= 55:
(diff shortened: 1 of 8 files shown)


matejcik:
simplified to just the encode method, and further simplified using code that we already have in core :)
added some tests

python/src/trezorlib/_rlp.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def _byte_size(x: int) -> int:
+    if x < 0:
+        raise ValueError("only unsigned ints are supported")
+    return (x.bit_length() + 7) // 8
+
+
+def _int_to_bytes(n: int) -> bytes:
+    """Convert python objects to bytes can be deserialize with converters"""
```
mmilata, line 17: nit: kinda hard to parse

python/tests/test_rlp.py
```diff
@@ -0,0 +1,24 @@
...
+    (b"A" * 1024, b"\xb9\x04\x00" + b"A" * 1024),
+    ([b"dog", b"cat", [b"spy"]], b"\xcd\x83dog\x83cat\xc4\x83spy"),
+    ([b"A" * 1024], b"\xf9\x04\x03\xb9\x04\x00" + b"A" * 1024),
+    ([], b"\xc0"),
+    (b"", b"\x80"),
+    (1, b"\x01"),
+    (0x7F, b"\x7f"),
+    (0x80, b"\x81\x80"),
+    (0x1_0000_0001, b"\x85\x01\x00\x00\x00\x01"),
+    ([0x1234, 0x5678], b"\xc6\x82\x12\x34\x82\x56\x78"),
```
mmilata, line 17: nit: testcase for the `len > 55` branch would be nice

mmilata (approved):

python/src/trezorlib/_rlp.py
matejcik: yeah that would be because it comes from a different function
eba3577832a788ab7cc56008551410f88b71224b

python/tests/test_rlp.py
matejcik: eba3577832a788ab7cc56008551410f88b71224b

mmilata (approved):

PR #3310, opened 2023-09-28
TT: fix low contrast for BIP-39 recovery

### LINKED ISSUE #3259: Very low contrast in recovery words
image

On real device it is almost impossible to see the last two letters of the word error

version 2.6.1

mmilata:
Fixes #3259.

emu00000001 emu00000000


Hannsek:
Working nice. Also please change the colour of the pressed button to GREEN_DARK + whole text on pressed button WHITE.

mmilata:
Changed, please try now.

Hannsek:
LGTM, thank you!

grdddj (approved):
LGTM!

Just one consideration - does not having so many styles increase the flash usage? There might be some common `ButtonStyle` objects, that are duplicated, maybe only with one different field. Or does it not increase the flash size when it is just a `&` reference? Definitely not a blocker for merging - moreover, the size tool did not find any difference against `master` - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/5198215487 (which may be strange?]

mmilata:
Flash size is important consideration for sure. According to `mem::size_of`, `ButtonStyleSheet` has 12 bytes and `ButtonStyle` also 12, so in total 48 per stylesheet. I figured 96B is not so much compared to adding additional logic.

There might be some duplication of the `ButtonStyle` objects but the savings are probably not worth the effort.

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1440x900
 - Device: Trezor T2T1 2.6.3 regular (revision c937f21447966b31e930ae8d20d1093d20e9fa66)

PR #3315, opened 2023-09-29
adjust T2B1 display refresh rate and contrast when displaying qr code

TychoVrahe:
adjust T2B1 display refresh rate and contrast when displaying qr code.

also SPI baudrate is adjusted - was slightly outside of tolerance.

Not very sure about the API of the functionality, its a bit single purpose but i also didnt want to get into exposing the contrast and frequency settings of the display.

[no changelog]


matejcik:
as discussed on slack, we won't do the different rate for QR code.
do we still want to keep the SPI change?

TychoVrahe:
yes. made a new PR for the SPI change: https://github.com/trezor/trezor-firmware/pull/3317

PR #3323, opened 2023-10-03
refactor(vendor): updated and replaced stm32 libs

cepetr:
With this pull request, we are discontinuing the use of the micropython/stm32lib submodule and relocating all related components into the vendor directory.

*Motivation: Currently, we are utilizing STM32 HAL/CMSIS libraries from the micropython repository as a submodule called stm32lib. This submodule includes a copy of the STM32CubeXX firmware libraries. However, the repository occasionally receives updates and is somewhat outdated. Additionally, it lacks STM32U5 firmware libraries.*

We have introduced four new submodules containing CMSIS and HAL drivers for STM32F4 and STM32U5. These submodules can be updated individually as needed.

The STM32U5 CMSIS/HAL drivers are up to date, while the STM32F4 CMSIS/HAL drivers have been aligned as closely as possible with the original stm32lib versions. Further updates for STM32F4 are possible but require additional testing.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,15 +21,12 @@
 [submodule "vendor/fido2-tests"]
 	path = vendor/fido2-tests
 	url = https://github.com/trezor/fido2-tests.git
-[submodule "vendor/stm32f4_cmsis"]
-	path = vendor/stm32f4-cmsis
-	url = https://github.com/STMicroelectronics/cmsis_device_f4.git
-[submodule "vendor/stm32u5_hal"]
-	path = vendor/stm32u5-drivers
+[submodule "vendor/stm32u5-hal"]
+	path = vendor/stm32u5-hal-drivers
 	url = https://github.com/STMicroelectronics/stm32u5xx_hal_driver.git
-[submodule "vendor/stm32f4_hal"]
-	path = vendor/stm32f4-drivers
-	url = https://github.com/STMicroelectronics/stm32f4xx_hal_driver.git
-[submodule "vendor/stm32u5_cmsis"]
-	path = vendor/stm32u5-cmsis
+[submodule "vendor/stm32u5-cmsis"]
+	path = vendor/stm32u5-cmsis-device
 	url = https://github.com/STMicroelectronics/cmsis_device_u5.git
+[submodule "vendor/stm32xx-lib"]
+	path = vendor/stm32xx-lib
+	url = https://github.com/micropython/stm32lib.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,3 +21,15 @@
 [submodule "vendor/fido2-tests"]
 	path = vendor/fido2-tests
 	url = https://github.com/trezor/fido2-tests.git
+[submodule "vendor/stm32f4_cmsis"]
+	path = vendor/stm32f4-cmsis
+	url = https://github.com/STMicroelectronics/cmsis_device_f4.git
+[submodule "vendor/stm32u5_hal"]
+	path = vendor/stm32u5-drivers
+	url = https://github.com/STMicroelectronics/stm32u5xx_hal_driver.git
+[submodule "vendor/stm32f4_hal"]
+	path = vendor/stm32f4-drivers
+	url = https://github.com/STMicroelectronics/stm32f4xx_hal_driver.git
+[submodule "vendor/stm32u5_cmsis"]
+	path = vendor/stm32u5-cmsis
+	url = https://github.com/STMicroelectronics/cmsis_device_u5.git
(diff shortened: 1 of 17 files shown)


prusnak:
> while the STM32F4 CMSIS/HAL drivers have been aligned as closely as possible with the original stm32lib versions.

@cepetr Please elaborate more about how did you figure out which version of the library to use. Did you diff the original code with the new code to see you are not introducing any regressions?

cepetr:
The version of stm32lib we are currently using is based on STM32CubeF4 1.16.0, which was last updated on April 14, 2017. This package includes STM32 HAL drivers (version 1.7.1) and CMSIS device (version 2.6.1). It's worth noting that these packages are quite outdated, and they are not incorporated in any ST GitHub repository nor available for download from ST's official pages. You can review the release notes here: https://htmlpreview.github.io/?https://github.com/STMicroelectronics/STM32CubeF4/blob/master/Release_Notes.html

As for the CMSIS device package, the closest available version is 2.6.3, which includes minor changes primarily addressing compilation warnings and adding new bit definitions. However, there is a notable exception in the definition of the ErrorStatus enum, which is heavily used in HAL_ll_ functions, making this change potentially risky. You can find the release notes here: https://htmlpreview.github.io/?https://github.com/STMicroelectronics/cmsis_device_f4/blob/master/Release_Notes.html

Regarding the HAL drivers, the closest available version is 1.7.6, and the list of changes is considerably longer. You can review the release notes here: https://htmlpreview.github.io/?https://github.com/STMicroelectronics/stm32f4xx_hal_driver/blob/master/Release_Notes.html

With this combination, we effectively have something akin to STM32CubeF4 1.24.1 (dated April 12, 2019).

**After discussions with matejcik, we have decided rather not to update the STM32F4 libraries as it can potentially introduce problems and still requires a significant amount of testing. Instead, we will focus on adding drivers for STM32U5 only. Nevertheless, we will move stm32lib from its current location in micropython/lib/stm32lib to the vendor folder.**

The vendor folder will contain the following subfolders:

```
vendor
   stm32u5-cmsis-device
   stm32u5-hal-drivers
   stm32f4-lib 
```

prusnak:
Yes. This approach makes sense.

In the next iteration we can try to diff the current stm32f4 sources with the latest from STM and perform the update and testing when we feel that's right thing to do.

prusnak (approved):
I confirmed that the original submodule `vendor/micropython/lib/stm32lib` and the new submodule `vendor/stm32xx-lib` point to the same repository (https://github.com/micropython/stm32lib) and to the same commit: [`eb80f0126e50687aac966f4c39a2b5a5deffbe78`](https://github.com/micropython/stm32lib/tree/eb80f0126e50687aac966f4c39a2b5a5deffbe78)

cepetr:
It appears that we need to revisit our current solution, which originated from an attempt to add STM32U5 drivers and create a unified approach for adopting MCU drivers and definitions across all MCUs. Here's a summary of what I have found and at the end there's are mye solution proposals.

#### STM32F2 (Model T1)
For MCU drivers and definitions, we rely on libopencm3, which serves as an open-source alternative to ARM CMSIS Core/Device and ST's HAL Drivers libraries. This library is located in the /vendor/libopencm3 directory.

#### STM32F4 (Model TT)
MCU drivers and definitions are sourced from the MicroPython project's codebase. The ARM CMSIS Core is found in `/vendor/micropython/lib/src/cmsis` (though it's very outdated, and its origin is unclear). Meanwhile, the ARM CMSIS Device and ST's HAL drivers libraries are located in the submodule `/vendor/micropython/lib/stm32lib` (also very outdated). It's important to note that we do not rely on MicroPython hardware modules, as we have our own hardware abstraction layer. Therefore, there is no dependency of MicroPython on src/cmsis and `src/stm32lib`.

#### STM32U5
To support this MCU, we need at least the following three components:
1) ARM CMSIS Core
2) ST'S CMSIS Device
3) ST'S HAL Drivers

These components are available in their respective GitHub repositories (~360MB) or can be obtained as a STM32CubeU5 bundle, which includes various other components (~1GB). While the STM32CubeU5 bundle is larger and usually receives slower updates from individual component repositories, it ensures that the component versions are properly aligned.

Here are the solution options:

There are three distinctly different approaches that are not easy to unify. Since we are not planning to update STM32F4 or STM32F2 drivers at the moment, I would prefer not to make any changes to them. We may revisit them later if we decide to do so. For now, our focus will be solely on STM32U5.

1) The easiest and the least intrusive solution is to just and STM32CubeU5 repo as a submodule to `/vendor folder`. Then we'l have everything we need for U5 and other MCU drivers will be intact. Unfortunatelly this adds ~1GB to our repo. If this is not the problem I would prefer this solution.

2) The second option is to add three submodules for CMSIS core, CMSIS device, and ST's HAL drivers. This option offers a size advantage but comes with some disadvantages related to version management.

3) We can create a new repository that contains only the components we need, essentially a pruned STM32CubeU5 repository. This option resolves the size issue but adds some maintenance overhead.

cepetr:
We have finally chosen to implement Solution #1. With this choice, there is no need for any modifications to the master branch, and we will proceed with the work in the tychovrahe/u5 branch. As a result, I will close this pull request and delete the cepetr/new-stm32-lib branch.

PR #3332, opened 2023-10-05
ci(nix): use ST fork of OpenOCD

mmilata:


cepetr (approved):
stm32u5 flashing/debugging is working now

prusnak (approved):
utACK

PR #3333, opened 2023-10-06
Prodtest versioning

matejcik:

### DIFF
--- a/core/embed/prodtest/towncrier.toml
+++ b/core/embed/prodtest/towncrier.toml
@@ -1,0 +1,1 @@
+../../../tools/towncrier.toml
--- a/core/embed/prodtest/version.h
+++ b/core/embed/prodtest/version.h
@@ -1,6 +1,6 @@
 #define VERSION_MAJOR 0
 #define VERSION_MINOR 2
-#define VERSION_PATCH 3
+#define VERSION_PATCH 4
 #define VERSION_BUILD 0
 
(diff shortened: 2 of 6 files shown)


core/embed/prodtest/towncrier.toml
```diff
@@ -0,0 +1 @@
+../../../tools/towncrier.toml
```
mmilata, line 1: Dead link?

core/embed/prodtest/version.h
```diff
@@ -1,6 +1,6 @@
 #define VERSION_MAJOR 0
 #define VERSION_MINOR 2
-#define VERSION_PATCH 2
+#define VERSION_PATCH 4
```
mmilata, line 3: Why skip 3?

core/embed/prodtest/towncrier.toml
matejcik: it's not dead locally :woman_shrugging: i think this is github bug. `core/ ( embed/ ( prodtest / .. ) /.. ) /.. / tools/towncrier.toml` exists

core/embed/prodtest/version.h
matejcik: it's there in the commits actually :) first go to 0.2.3 for a commit that will be tagged, and bump immediately after in "bump version after release"

mmilata (approved):

PR #3335, opened 2023-10-09
Optiga improvements

andrewkozlik:
Let's discuss the merits of the individual commits in this PR. See also:
- https://www.notion.so/satoshilabs/Implement-PIN-protection-using-OPTIGA-in-firmware-8688fbc787354b848bb89ccda54a8982
- https://satoshilabs.slack.com/archives/C04TYLWUK0Q/p1696596731865469?thread_ts=1696431191.030249&cid=C04TYLWUK0Q

### DIFF
--- a/core/embed/trezorhal/optiga/optiga_commands.c
+++ b/core/embed/trezorhal/optiga/optiga_commands.c
@@ -75,7 +75,7 @@ static optiga_result process_output(uint8_t **out_data, size_t *out_size) {
 
   *out_data = tx_buffer + 4;
   *out_size = tx_size - 4;
-  OPTIGA_LOG("SUCCESS ", *out_data, *out_size)
+  OPTIGA_LOG("SUCCESS", *out_data, *out_size)
   return OPTIGA_SUCCESS;
 }
 
--- a/core/embed/trezorhal/optiga/optiga_transport.c
+++ b/core/embed/trezorhal/optiga/optiga_transport.c
@@ -184,7 +184,7 @@ optiga_result optiga_init(void) {
 }
 
 static optiga_result optiga_i2c_write(const uint8_t *data, uint16_t data_size) {
-  OPTIGA_LOG(">>> ", data, data_size)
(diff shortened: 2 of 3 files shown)


storage/storage.c
```diff
@@ -1149,9 +1179,11 @@ static secbool unlock(const uint8_t *pin, size_t pin_len,
   // Now we can check for wipe code.
   ensure_not_wipe_code(unlock_pin, unlock_pin_len);
 
-  // Get the pin failure counter
+  // Get the pin failure counter. To ensure correct exponential backoff,
+  // synchronize counters in case they diverged. However, we are unaware of any
+  // way how the Optiga counter could be ahead of the MCU counter.
```
matejcik, line 1281: i thought the obvious answer is "user manages to unplug device at exactly the right moment"?
(more like, the MCU counter shouldn't be ahead of Optiga unless Optiga is bad)

matejcik:
I rebased the PR on the current state, there were some small conflicts so @andrewkozlik please check that everything is as it should be.

I also did some additional changes:
c49a05be9bbf6f0420dc81d31c8a921d149eee47 changes NDEBUG to PRODUCTION and tweaks naming
6ed3bcfaa48b15e7d0855231e99dfbc4cd9143fe is the current version of the provisioning script
f2c34e9e2efbdce4a90c68e5722d97bd6d4e52a2 enables Optiga logging in debug builds
93b14046b2fd3225293cd172540e4d175fb724fe prevents locking metadata in non-production prodtest

core/embed/firmware/main.c
```diff
@@ -92,6 +92,19 @@
 // from util.s
 extern void shutdown_privileged(void);
 
+#ifdef USE_OPTIGA
+#if !PYOPT
+static void optiga_log_hex(const char *prefix, const uint8_t *data,
+                           size_t data_size) {
+  printf("%s: ", prefix);
```
andrewkozlik, line 99: Should we add some extra prefix here, like: `"OPTIGA %s: "`?

PR #3336, opened 2023-10-10
Update Python and Pillow

matejcik:
This resolves dependabot errors about vulnerable libwebp. We should not be affected by the vulnerability, but better safe than sorry.

It also adds Python 3.11 to the tox matrix for trezorlib.
I tried adding Python 3.12 but it fails because something something distutils -- they were removed from Python in 3.12, but tooling still depends on them. My guess is that we'd need to upgrade pip / setuptools / poetry, but I'm not sure which of them and how exactly. @mmilata if you could take a brief look and see if you can fix it easily, that would be nice; if not, let's live without Py3.12 for now.


mmilata:
I don't see anything obvious. Perhaps we need to bump nixpkgs for 3.12 to work, there might be some bugfixes in more recent package.

mmilata:
3.11 seems to fail on aio-libs/aiohttp#6600

matejcik:
the too-old stellar sdk is holding us back, I'll see if I can fix it easily

matejcik:
rebased on top of `drop-simple-rlp` because it was causing trouble. 
e1ff3cae8d4bad38eca09b8a8595f037f61c9ce7 should resolve the py3.11 problem

mmilata (approved):

PR #3339, opened 2023-10-11
Remove redundant display initialization in the intermediate firmware on T1

### LINKED ISSUE #3308: T1 display is black or glitches
**Describe the bug**
When upgrading the bootloader on the T1 model using intermediate firmware, the device's display sometimes remains black or glitches, and the device becomes unusually hot. It appears that the intermediate firmware, unlike the release firmware, performs unnecessary display initialization in the main() function, leading to this issue. Removing this initialization step could potentially resolve the problem.

- Firmware 1.21.1
- Intermediate firmware v3 (https://github.com/trezor/data/blob/master/firmware/1/trezor-inter-v3.bin)

**To Reproduce**
Reproducing the problem may not be straightforward. The following sequence has worked for me::

1. Power on the device while holding the left button (this puts the device into bootloader mode)
2. Wipe the device using following command
    `trezorctl device wipe --bootloader`
3. Upload standard firmware
    `trezorctl fw update`
4. Power off and on the device again while holding the left button
5. Upload intermediate firmware
    `trezorctl fw update -f trezor-inter-v3.bin`
6. Repeat step 5 (or steps 4 and 5)  (sometimes it may require 10 or more attempts)  until a glitch or black screen appears. 

**Screenshots**
The display is black or appears as shown on this link:
https://forum.trezor.io/t/trezor-one-bug-on-the-screen-while-downloading-wallet-software/13587

**Additional context**
The same bug can be reproduced if the oledInit() function is called in the standard firmware at the beginning of the main() function. This seems to be the key difference between the standard and intermediate firmware versions, which explains why this behavior is not typically encountered. The cause of this phenomenon is currently unknown and requires further investigation. It appears that when the device enters this state, the only solution is to power it off and then power it back on.

cepetr:
I have removed redundant display initialization in the intermediate firmware on T1. This change should reduce the probability of display glitches described in issue #3308. However, please note that this is just a workaround, as the root cause of the problem is still unknown.

fixes #3308

### DIFF
--- a/legacy/intermediate_fw/trezor.c
+++ b/legacy/intermediate_fw/trezor.c
@@ -107,7 +107,7 @@ int main(void) {
   setupApp();
   __stack_chk_guard = random32();  // this supports compiler provided
                                    // unpredictable stack protection checks
-  oledInit();
+
   if (is_mode_unprivileged()) {
     layoutDialog(&bmp_icon_warning, NULL, NULL, NULL, "Cannot update", NULL,
                  NULL, "Unprivileged mode", "Unsigned firmware", NULL);


matejcik (approved):
LGTM, @hiviah please have a quick look too

PR #3342, opened 2023-10-13
chore: update branch name to main where its used

vdovhanych:

### DIFF
--- a/ci/posttest.yml
+++ b/ci/posttest.yml
@@ -33,7 +33,7 @@ core unix coverage posttest:
 unix ui changes:
   stage: posttest
   except:
-    - master
+    - main
   <<: *gitlab_caching
   needs:
     - core click test
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -352,7 +352,7 @@ core unix frozen debug build arm:
   <<: *gitlab_caching
   needs: []
   only:
-    - master
+    - main
     - tags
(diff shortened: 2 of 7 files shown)


matejcik (approved):
utACK but let's see if CI is happy first

ci/posttest.yml
```diff
@@ -43,10 +43,10 @@ unix ui changes:
     - core device R test
     - legacy device test
   script:
-    - $NIX_SHELL --run "poetry run python -m tests.ui_tests.reporting master-diff"
-    - mv tests/ui_tests/reports/master_diff .
+    - $NIX_SHELL --run "poetry run python -m tests.ui_tests.reporting main-diff"
+    - mv tests/ui_tests/reports/main_diff .
```
matejcik, line 47: the command is still called `master-diff` and the path is hardcoded to `master_diff` it seems.
honestly i'd maybe keep those names? it seems more readable than "main_diff", which sounds more like "the main one of all the various diffs" than "diff from main branch".
(or rename it to "diff-from-main")

vdovhanych: O, I'll check it.

matejcik: f93d9a636d93370dfa68ecb49d1eb7dd8b236627

vdovhanych (commented):
lgtm then :)

PR #3345, opened 2023-10-16
prepare fw for differently sized fw chunks

TychoVrahe:
part of preparations for U5.


python/src/trezorlib/firmware/core.py
```diff
@@ -129,19 +130,26 @@ def code_hashes(self) -> t.List[bytes]:
...
         hashes = []
+
+        if self.HASH_PARAMS.chunk_size is None:
+            internal_name = Model.from_hw_model(self.header.hw_model).value.decode()
+            trezor_model = by_internal_name(internal_name)
+            if trezor_model is None:
+                return hashes
+            chunk_size = trezor_model.chunk_size
+        else:
+            chunk_size = self.HASH_PARAMS.chunk_size
```
matejcik, line 141: ...and then this can just be
```
params = Model.from_hw_model(self.header.hw_model).hash_params()
```

python/src/trezorlib/models.py
```diff
@@ -32,6 +32,7 @@ class TrezorModel:
     vendors: Collection[str]
     usb_ids: Collection[UsbId]
     default_mapping: mapping.ProtobufMapping
+    chunk_size: int
```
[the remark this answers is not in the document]
TychoVrahe: done https://github.com/trezor/trezor-firmware/commit/9544f53afd8ba2e38b7b8c7ede6acee73cbd0d74

python/src/trezorlib/firmware/core.py
TychoVrahe: i still need to distinguish legacy and core, because (at least in tests) both have `hw_model` field zeroed and thus would default to T in `from_hw_model`. Added a `get_hash_params` function for that purpose. It is overridden in `LegacyV2Firmware`
https://github.com/trezor/trezor-firmware/commit/9544f53afd8ba2e38b7b8c7ede6acee73cbd0d74

core/embed/lib/image.c
```diff
@@ -224,12 +224,13 @@ secbool check_image_contents(const image_header *const hdr, uint32_t firstskip,
     return secfalse;
   }
 
-  const void *data =
-      flash_area_get_address(area, firstskip, IMAGE_CHUNK_SIZE - firstskip);
+  int remaining = hdr->codelen;
+
+  const void *data = flash_area_get_address(
```
matejcik, line 229: why is `flash_area_get_address` affected by `remaining`? shouldn't `IMAGE_CHUNK_SIZE` always match?

TychoVrahe: the area can be smaller than the chunk:
- on u5, we will (maybe) have 128kB bootloader and 256kb chunks
- also the fw area is not divisble by 256kb, so the last part will always by smaller

matejcik (approved):
ACK on code, unfortunately #3349 introduced conflicts in `models.py` so please fix those

PR #3346, opened 2023-10-16
remove residual usage of old flash api

TychoVrahe:
should only be using `flash_area_xx` functions now - at least in core

### DIFF
--- a/core/embed/extmod/modtrezorutils/modtrezorutils.c
+++ b/core/embed/extmod/modtrezorutils/modtrezorutils.c
@@ -48,6 +48,8 @@
 #include "secret.h"
 #endif
 
+#define FW_HASHING_CHUNK_SIZE 1024
+
 static void ui_progress(mp_obj_t ui_wait_callback, uint32_t current,
                         uint32_t total) {
   if (mp_obj_is_callable(ui_wait_callback)) {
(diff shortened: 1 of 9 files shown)


core/embed/extmod/modtrezorutils/modtrezorutils.c
```diff
@@ -171,20 +173,24 @@ STATIC mp_obj_t mod_trezorutils_firmware_hash(size_t n_args,
     ui_wait_callback = args[1];
   }
 
-  uint16_t firmware_sectors = flash_total_sectors(&FIRMWARE_AREA);
+  uint32_t firmware_size = flash_area_get_size(&FIRMWARE_AREA);
+  uint32_t chunks = firmware_size / FW_HASHING_CHUNK_SIZE;
```
mmilata, line 177: The function will return a wrong hash if `firmware_size % FW_HASHING_CHUNK_SIZE != 0`, is there any chance a imaginary future hardware would have some weird sector size that would cause this? If yes I guess we can just put an assert here.

mmilata (approved):

mmilata (approved):

core/embed/extmod/modtrezorutils/modtrezorutils.c
TychoVrahe: page sizes of e.g. 512 bytes or smaller are probably possible, though we are not very likely to encounter them soon, so adding ensure seems reasonable, see https://github.com/trezor/trezor-firmware/pull/3346/commits/efb80270cca8271ba7d4761654f9c1148804e1e3

PR #3347, opened 2023-10-16
CLI: force bitcoin only fw on bitcoin only units

TychoVrahe:
installation of universal fw is still possible via custom url.


matejcik:
closing in favor of #3349

PR #3348, opened 2023-10-17
add boundaries checks for icon rendering (framebuffer variant)

TychoVrahe:
So that it correctly renders on the edge of display and such.

Also display_refresh calls are added in TT UI, as a preparations for models with displays that actually needs this. Shouldn't affect TT.


mmilata (approved):
LGTM

PR #3349, opened 2023-10-17
Trezorlib features for TS3

matejcik:
mainly concerned with the firmware command, plus some internal rearrangements

supersedes https://github.com/trezor/trezor-firmware/pull/3347

### DIFF
--- a/python/.changelog.d/+13d8ef2d.changed
+++ b/python/.changelog.d/+13d8ef2d.changed
@@ -1,0 +1,1 @@
+Internal names are used consistently in constants and names. Original model-based names are kept as aliases for backwards compatibility.
--- a/python/src/trezorlib/cli/firmware.py
+++ b/python/src/trezorlib/cli/firmware.py
@@ -32,18 +32,30 @@ from urllib.parse import urlparse
 import click
 import requests
 
-from .. import exceptions, firmware, messages
+from .. import exceptions, firmware, messages, models
 from ..firmware import models as fw_models
-from . import with_client
+from ..models import TrezorModel
+from . import ChoiceType, with_client
 
 if TYPE_CHECKING:
     from ..client import TrezorClient
     from . import TrezorConnection
 
-ALLOWED_FIRMWARE_FORMATS = {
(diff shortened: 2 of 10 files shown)


python/.changelog.d/+13d8ef2d.changed
```diff
@@ -0,0 +1 @@
+Internal names are used consistently in constants and names. Original model-based names are kept as aliases for backwards compatibility.
```
grdddj, line 1: Why does the file have such a strange name? (`+13d8ef2d.changed`)

I would expect `3349.changed`.

Does not apply when you plan to rename it after having PR number :)

python/src/trezorlib/cli/firmware.py
```diff
@@ -58,6 +81,24 @@ def _get_file_name_from_url(url: str) -> str:
     return os.path.basename(full_path)
 
 
+def _print_firmware_model(hw_model: Union[bytes, fw_models.Model]) -> None:
+    try:
+        model_name = fw_models.Model.from_hw_model(hw_model).name
+        click.echo(f"{model_name} firmware image.")
+    except ValueError:
+        pass
+    else:
```
grdddj, line 90: Could be more readable to `return` after `click.echo` 3 lines above - the same as is done 5 lines below

grdddj (approved):
Code-wise ACK with small comments

Have not tried to run it

python/.changelog.d/+13d8ef2d.changed
matejcik: when you run `towncrier create +.changed`, it will generate a name like this. then there is no linked issue in the changelog. i can rename it, it doesn't seem to useful though

grdddj: Not insisting on the renaming, was just curious

python/src/trezorlib/cli/firmware.py
matejcik: 40ac5672d33e78ffc2330f3abe2b6989a855d1df

i thought it's more readable with the else, but i admit you're right; try-else is rather uncommon construction

Issue #3203, opened 2023-08-10
trezorctl: better notification about passphrase entry


matejcik:
when entering passphrase on TT with "always on device" mode, the following prompt is shown in console:
`Please confirm action on your Trezor device.`
at the same time, TT screen is showing a keyboard with no accompanying text.

in this situation, trezorctl should emit `Please enter passphrase on your Trezor device.`

Similarly, a PIN prompt should say "Please enter PIN" etc.

The default implementation only shows "please confirm action" once per multiple actions, this should be changed so that if the current prompt is the same as last time, it doesn't get printed, but if it's different, it gets printed.

Issue #3136, opened 2023-07-12
trezorctl eth web3 error send-tx


nolash:
**Describe the bug**

when trying eth send-tx

AttributeError: 'Web3' object has no attribute 'isConnected'. Did you mean: 'is_connected'?

**Firmware version and revision**

$ trezorctl get-features
Features (166 bytes) {
    bootloader_hash: 32 bytes 0xfa12a44fa05fd1d20539358b54f301cee4c3219c9f1bb3a5772ffd609af9e8e2,
    capabilities: [<Capability.Bitcoin: 1>, <Capability.Bitcoin_like: 2>, <Capability.Crypto: 5>, <Capability.Ethereum: 7>, <Capability.NEM: 10>, <Capability.Stellar: 12>, <Capability.U2F: 14>],
    fw_vendor: 'SatoshiLabs',
    initialized: True,
    language: 'en-US',
    major_version: 1,
    minor_version: 11,
    model: '1',
    needs_backup: False,
    no_backup: False,
    passphrase_protection: False,
    patch_version: 2,
    pin_protection: True,
    revision: 20 bytes 0x0d87b55ba4fed7eecc72bf2a94ee473830b095e9,
    safety_checks: <SafetyCheckLevel.Strict: 0>,
    unfinished_backup: False,
    unlocked: False,
    vendor: 'trezor.io',
}

**Desktop/smartphone setup (please complete the following information):**
 - Wallet software: [e.g. Trezor Suite, Electrum]

trezorctl

 - OS: [e.g. Windows, macOS, Linux, Android]

linux arch

 - Browser [e.g. Chrome, Firefox, Safari]

na

 - Version [e.g. 90.0.2]

$ pip freeze
aiohttp==3.8.4
aiosignal==1.3.1
async-timeout==4.0.2
attrs==23.1.0
bitarray==2.7.6
certifi==2023.5.7
charset-normalizer==3.2.0
click==8.1.4
construct==2.10.68
construct-classes==0.1.2
cytoolz==0.12.1
ecdsa==0.18.0
eth-abi==4.1.0
eth-account==0.9.0
eth-hash==0.5.2
eth-keyfile==0.6.1
eth-keys==0.4.0
eth-rlp==0.3.0
eth-typing==3.4.0
eth-utils==2.2.0
frozenlist==1.3.3
hexbytes==0.3.1
idna==3.4
jsonschema==4.18.0
jsonschema-specifications==2023.6.1
libusb1==3.0.0
[...]

Issue #3037, opened 2023-05-26
Use m/44h/128h/0h instead of m/44'/128'/0' for hardened paths everywhere


prusnak:
People have trouble quoting paths in shell (see https://github.com/trezor/trezor-firmware/issues/3036 for an example). Using `h` instead of `'` helps with this.

matejcik:
is it ok to change this in descriptors too? (it changes the checksum)

prusnak:
> is it ok to change this in descriptors too? (it changes the checksum)

Yes, descriptors too, please

PR #3360, opened 2023-10-19
trezorlib 0.13.8

matejcik:
Implemented fixes for a couple low-hanging fruits: #3136 , #3037 , #3203


matejcik:
1f45aeb70a3fe86da1b2b9d11703654ad52e0f92 should fix py3.6
but otherwise yes, it's time. I won't do it for this release, but let's plan 0.14 to drop py3.6 and py3.7

PR #3363, opened 2023-10-20
Interaction-less update

### LINKED ISSUE #2794: Interrupting firmware update at the wrong time erases seed

### LINKED ISSUE #2919: (Almost) interaction-less firmware update

cepetr:
This pull request primarily involves a significant update, commonly referred to as an "interaction-less update,".

resolves #2919 
resolves #2794

The interaction-less update can be initiated via the command line using `trezorctl fw update`. However, several conditions must be met for this to work:

1. The uploaded firmware must come from the same vendor, and its version must be higher.
2. The current firmware version on the device must be equal to or higher than 2.6.3.
3. The device must not be in bootloader mode.

~If any of these conditions are not met, the standard firmware upload via the bootloader is selected. Additionally, users can manually enable or disable the interaction-less update by specifying the `--ilu` or `--no-ilu` options.~

The interaction-less update process begins on the `trezorctl` side by sending a `RebootToBootloader` message. This message has been extended to include two optional parameters: `bootCommand` and `firmwareHeader`. In this context, `bootCommand` is set to `INSTALL UPGRADE`, and `firmwareHeader` contain the new firmware headers (the first ~5.5KB of the binary).

When the firmware receives the `RebootToBootloader` message with the `INSTALL UPGRADE` command, it parses and checks the provided firmware header. It compares the vendor and version fields with the current firmware. If the vendor is the same, and the version is higher, it displays a confirmation dialog. Once the user confirms the firmware upgrade, the firmware proceeds to jump to the bootloader.

A new interface between the firmware and the bootloader has been implemented. The firmware can now jump to the bootloader and provide a command (enum) and arguments (up to 256 bytes). In this case, the firmware jumps to the bootloader with the `INSTALL_UPGRADE` command, and the arguments are set to the hash of the vendor and image header received in the `RebootToUploader` message.

When the bootloader is initiated and detects the `INSTALL_UPGRADE` command, it skips the initial screens and immediately starts waiting for host commands.

Following this, `trezorctl` continues with the standard firmware update process, which includes uploading the new firmware. The bootloader checks whether the uploaded firmware header has the same hash as the firmware confirmed by the user. It also performs the same checks as the firmware (comparing vendor and version). If the firmware is the same and newer, it proceeds with flashing the firmware without further confirmation.


python/src/trezorlib/cli/firmware.py
```diff
@@ -635,15 +663,57 @@ def update(
...
+                click.echo(
+                    "Interaction-less update can not be started from bootloader mode."
+                )
+                exit(1)
+            elif not ilu_supported:
+                click.echo(
+                    "Interaction-less upgrade is not supported by the current firmware version."
+                )
+                exit(1)
+            elif not strict_upgrade:
```
TychoVrahe, line 693: maybe we could ignore this if skip_check is set

core/embed/bootloader/memory.ld
```diff
@@ -23,8 +23,9 @@ ccmram_end = ORIGIN(CCMRAM) + LENGTH(CCMRAM);
 sram_start = ORIGIN(SRAM);
 sram_end = ORIGIN(SRAM) + LENGTH(SRAM);
 
-/* IMAGE_HEADER_SIZE is 0x400, this is for interaction-less firmware update start */
-firmware_header_start = ccmram_end - 0x400;
+/* reserve 256 bytes for bootloader arguments */
+boot_args_end = ccmram_end;
+boot_args_start = boot_args_end - 0x100;
```
TychoVrahe, line 28: if no more elegant solution can be found, would prefer the memory to be split as previously discussed, to ensure no overlap:

CCMRAM (wal) : ORIGIN = 0x10000000, LENGTH = 64K - 0x100
BOOT_ARGS (wal) : ORIGIN = 0x1000FF00, LENGTH = 0x100

and then

/* reserve 256 bytes for bootloader arguments */
boot_args_start = ORIGIN(BOOT_ARGS);
boot_args_end = ORIGIN(BOOT_ARGS) + LENGTH(BOOT_ARGS);

TychoVrahe (changes requested):
Good work. Leaving few comments in addition to mentioned todos and things flagged by CI.

core/embed/bootloader/messages.c
```diff
@@ -577,9 +581,42 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
+
+      if (g_boot_command == BOOT_COMMAND_INSTALL_UPGRADE) {
+        BLAKE2S_CTX ctx;
+        uint8_t hash[BLAKE2S_DIGEST_LENGTH];
+        blake2s_Init(&ctx, BLAKE2S_DIGEST_LENGTH);
+        blake2s_Update(&ctx, CHUNK_BUFFER_PTR,
+                       vhdr.hdrlen + received_hdr->hdrlen);
+        blake2s_Final(&ctx, hash, BLAKE2S_DIGEST_LENGTH);
+
+        // firmware must have be the same as confirmed by the user
```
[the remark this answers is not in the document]
cepetr: fixed

core/embed/bootloader/messages.c
```diff
@@ -577,9 +581,42 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
+        extern uint8_t boot_args_start;
```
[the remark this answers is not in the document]
cepetr: fixed

core/embed/bootloader/messages.c
```diff
@@ -577,9 +581,42 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
+        if (is_upgrade != sectrue || is_newvendor != secfalse) {
+          MSG_SEND_INIT(Failure);
+          MSG_SEND_ASSIGN_VALUE(code, FailureType_Failure_ProcessError);
+          MSG_SEND_ASSIGN_STRING(message, "Not a firmware upgrade");
+          MSG_SEND(Failure);
+          return UPLOAD_ERR_NOT_FIRMWARE_UPGRADE;
+        }
+
+        // upload firmware without confirmation
+        is_new = sectrue;
```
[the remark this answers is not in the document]
cepetr: I've rather introduced a new flag is_ilu (iteraction-less update).

python/src/trezorlib/cli/firmware.py
```diff
@@ -635,15 +663,57 @@ def update(
...
+            )
 
     if dry_run:
         click.echo("Dry run. Not uploading firmware to device.")
-    else:
+        return
+
+    print(ilu)
+
+    if ilu is False:
```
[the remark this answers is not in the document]
cepetr: fixed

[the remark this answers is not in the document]
cepetr: yes, there should be `if ilu is not False`

python/src/trezorlib/cli/firmware.py
```diff
@@ -460,6 +461,36 @@ def upload_firmware_into_device(
         sys.exit(3)
 
 
+def _is_ilu_supported(client: "TrezorClient") -> bool:
+    """Check if the firmware supports interaction-less update."""
+    f = client.features
+    version = (f.major_version, f.minor_version, f.patch_version)
+    return version >= (2, 6, 3)
```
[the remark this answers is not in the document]
TychoVrahe: @matejcik take a look at my argument here, as i was told that this was your suggestion.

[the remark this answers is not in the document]
matejcik: bootloader itself can't really support ILU, can it? it doesn't make sense at that point. "ilu" here means "i can send a special message before the actual fw upload to improve the UX flow". if you're in bootloader already, you just blast the firmware contents and if the bootloader wants to do something smart about it, there is nothing else needed from host side.

so we need to check for bootloader here and return false. other than that, version-gating this seems fine.
i mean, i'm not _against_ adding a capability for this, but like, in what situation would a future Trezor _not_ have this capability?

python/src/trezorlib/device.py
```diff
@@ -238,8 +238,14 @@ def unlock_path(client: "TrezorClient", n: "Address") -> "MessageType":
 
 @session
 @expect(messages.Success, field="message", ret_type=str)
-def reboot_to_bootloader(client: "TrezorClient") -> "MessageType":
-    return client.call(messages.RebootToBootloader())
+def reboot_to_bootloader(
+    client: "TrezorClient",
+    bootCommand: messages.BootCommand = None,
+    bootArgs: bytes = None,
```
matejcik, line 244:
```suggestion
    boot_command: messages.BootCommand = messages.BootCommand.STOP_AND_WAIT,
    firmware_header: Optional[bytes] = None,
```

python/src/trezorlib/cli/firmware.py
```diff
@@ -460,6 +461,36 @@ def upload_firmware_into_device(
...
+    """Check if the firmware is from the same vendor and the
+    firmware is newer than the currently installed firmware.
+    """
+    try:
+        fw = firmware.parse(firmware_data)
+    except Exception as e:
+        click.echo(e)
+        sys.exit(2)
+
+    if isinstance(fw, firmware.VendorFirmware):
```
matejcik, line 481: flip this: if not isinstance-> return false, otherwise continue the checks

python/src/trezorlib/cli/firmware.py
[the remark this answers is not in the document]
TychoVrahe: Hm the bootloader itself has to have support for ILU for it to work correctly, but right now i can't think of situation where this would be needed to know, certainly not needed in current implementation.

I can imagine a future trezor model that does prioritize code simplicity over UX where this would be disabled, or for even more of cost-sensitive model where we would cut features to bare minimum, or if there is a security exploit of this feature that would only be solvable by disabling it. Whether these are likely to happen is another thing, but definitely in reach of my imagination. 

Anyway, worst case we will force trezorctl/suite update and adjust this later, so not a big deal.

core/embed/trezorhal/stm32f4/supervise.c
```diff
@@ -0,0 +1,106 @@
+#include STM32_HAL_H
+
+#include <model.h>
+
+#include "../mpu.h"
+#include "common.h"
+#include "supervise.h"
+
+#ifdef FIRMWARE
```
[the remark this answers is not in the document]
cepetr: new define ARM_USER_MODE introduced, dabac40433754fef56245f7705b7048642fb3107

python/src/trezorlib/cli/firmware.py
```diff
@@ -635,15 +663,57 @@ def update(
...
+
+        if not client.features.bootloader_mode and ilu_supported and strict_upgrade:
+
+            with obj.client_context() as client:
+                device.reboot_to_bootloader(
+                    client,
+                    bootCommand=messages.BootCommand.INSTALL_UPGRADE,
+                    bootArgs=firmware_data[
+                        : 6 * 1024
+                    ],  # TODO!@#: How many bytes should we send ?
```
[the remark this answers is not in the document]
cepetr: resolved e860ad3b41ab3e15ab4cc32223e6b44f0b9a16a2

python/src/trezorlib/cli/firmware.py
```diff
@@ -635,15 +663,57 @@ def update(
...
+                )
+                time.sleep(3)  # TODO!@#: How to wait properly ?
```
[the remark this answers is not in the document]
cepetr: @matejcik Could you suggest a proper solution?

python/src/trezorlib/cli/firmware.py
```diff
@@ -593,11 +639,10 @@ def update(
     dry_run: bool,
     beta: bool,
     bitcoin_only: Optional[bool],
```
matejcik, line 641: the flow is super confusing and I thing the whole function should be rethought and refactored.

initial idea, high-level steps (probably worth making a separate function each)

1. figure out what firmware to install, return already validated firmware bytes
2. jump to bootloader -- if not already there, invoke ILU. Thinking deeper about this, I'm honestly not sure if the whole ILU detection makes any sense? We can _always_ invoke the ILU call if we're in firmware, and the older ones will simply not recognize the new options and drop you into BL as normal. So let's just do that.
3. upload the firmware that we got from step (1)

(1) and (2) can share the same client instance. (3) needs a new one, so that's two `with client_context` blocks

the wait loop between (2) and (3) can be written like this:

```
click.echo("Waiting for bootloader...")
while True:
    time.sleep(0.5)
    try:
        obj.get_transport()
        break
    except Exception:
        pass

with obj.client_context() as client:
    if not client.bootloader_mode:
        click.echo("Failed to enter bootloader mode, please try again.")
        sys.exit(1)
```

core/embed/trezorhal/common.h
```diff
@@ -51,10 +51,19 @@
   })
 #endif
 
-#define STAY_IN_BOOTLOADER_FLAG 0x0FC35A96
+// Defines special boot action for reboot_to_bootloader() function
+typedef enum {
+  // Normal flow, like after power on
+  BOOT_COMMAND_NONE = 0x00000000,
```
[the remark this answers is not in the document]
cepetr: It represents "any other/random" value. You can use it as an argument when calling svc_reboot_to_bootloader() to just reboot. If it's confusing, I can remove it.

common/protob/messages-management.proto
```diff
@@ -486,6 +486,13 @@ message CancelAuthorization {
  * @next Success
  */
 message RebootToBootloader {
+    optional BootCommand bootCommand = 1 [default=STOP_AND_WAIT];
+    optional bytes bootArgs = 2;
```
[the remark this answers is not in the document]
cepetr: refactored eca45ab231e40cd9103233275aae2eb04f325e9f, 1e9b943ce656e3dd51928be456cb37f90be3ced9, ce485b74549c26dfde6604485551d174074cd193

python/src/trezorlib/device.py
cepetr: refactored ce485b74549c26dfde6604485551d174074cd193

python/src/trezorlib/cli/firmware.py
cepetr: refactored 0c7a1f94ff272509877c88331208e6a8dfacef34

python/src/trezorlib/cli/firmware.py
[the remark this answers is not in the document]
cepetr: resolved 90871359a6118a3677078e475f99585fbbf4d8d9

python/src/trezorlib/cli/firmware.py
cepetr: Not relevant already. As @matejcik suggested, I've removed the --ilu/--no-ilu arguments.

python/src/trezorlib/cli/firmware.py
```diff
@@ -607,43 +650,68 @@ def update(
...
+                )
+
+        if dry_run:
+            click.echo("Dry run. Not uploading firmware to device.")
+            return
+
+        strict_upgrade = _is_strict_update(client, firmware_data)
+        ilu_supported = _is_ilu_supported(client)
+
+        if not client.features.bootloader_mode and ilu_supported and strict_upgrade:
```
matejcik, line 691: just this:
```suggestion
        if not client.features.bootloader_mode:
```

python/src/trezorlib/device.py
```diff
@@ -238,8 +238,16 @@ def unlock_path(client: "TrezorClient", n: "Address") -> "MessageType":
 
 @session
 @expect(messages.Success, field="message", ret_type=str)
-def reboot_to_bootloader(client: "TrezorClient") -> "MessageType":
-    return client.call(messages.RebootToBootloader())
+def reboot_to_bootloader(
+    client: "TrezorClient",
+    boot_command: messages.BootCommand = None,
```
matejcik, line 243:
```suggestion
    boot_command: messages.BootCommand = messages.BootCommand.STOP_AND_WAIT,
```

cepetr: changed 66d891aa495d5f2b3506ec3a291e7970fb8ef8a7

core/embed/trezorhal/common.h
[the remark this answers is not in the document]
matejcik: i think it's just the comment that is confusing to me. how about this?
```
typedef enum {
  // Normal boot sequence
  BOOT_COMMAND_NONE = 0x00000000,
  // Stay in bootloader menu
  BOOT_COMMAND_STOP_AND_WAIT = 0x0FC35A96,
  // Proceed to firmware installation
  BOOT_COMMAND_INSTALL_UPGRADE = 0xFA4A5C8D,
```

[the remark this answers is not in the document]
cepetr: ok, I understand. Fixed 1a197541f270963ef370283c57d165b53c7665a3

python/src/trezorlib/cli/firmware.py
cepetr: I've made a change so that trezorctl fw update always switches to the bootloader, using INSTALL_UPGRADE if the firmware is newer, or using STOP_AND_WAIT in other cases. 3139f02150230bdfd0bddd881ce1c7cfdd67e406

matejcik:
@TychoVrahe please take a look at the latest changes
@hiviah please review the bootloader part too

matejcik:
@mmilata please take over the outstanding UI issues, namely:

* the "connect to host" screen needs white-on-black variant (similar to the bootloader loader)
* the "firmware details" screen, incl. fingerprint, should be shared between firmware and bootloader, and used in the firmware part of this PR

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -167,6 +167,50 @@ secbool flash_area_erase_bulk(const flash_area_t *area, int count,
...
+  uint32_t sector_offset = 0;
+  *_bytes_erased = 0;
+
+  for (int s = 0; s < area->num_subareas; s++) {
+    for (int i = 0; i < area->subarea[s].num_sectors; i++) {
+      uint32_t sector_index = area->subarea[s].first_sector + i;
+      uint32_t sector_size = FLASH_SECTOR_TABLE[sector_index + 1] -
+                             FLASH_SECTOR_TABLE[sector_index];
+
+      if (offset == sector_offset) {
```
TychoVrahe, line 181: what is the reason not to accept any offset within the sector? its not necessarily wrong but its also not super-clear from the function parameter names nor the comment, to me at least.

core/embed/boardloader/memory.ld
```diff
@@ -3,9 +3,10 @@
 ENTRY(reset_handler)
 
 MEMORY {
-  FLASH  (rx)  : ORIGIN = 0x08000000, LENGTH = 48K
-  CCMRAM (wal) : ORIGIN = 0x10000000, LENGTH = 64K
-  SRAM   (wal) : ORIGIN = 0x20000000, LENGTH = 192K
+  FLASH  (rx)     : ORIGIN = 0x08000000, LENGTH = 48K
+  CCMRAM (wal)    : ORIGIN = 0x10000000, LENGTH = 64K - 0x100
+  BOOT_ARGS (wal) : ORIGIN = 0x1000FF00, LENGTH = 0x100
```
TychoVrahe, line 8: i don't think we want this for f4 boardloader

core/embed/trezorhal/stm32f4/flash.c
```diff
@@ -167,6 +167,50 @@ secbool flash_area_erase_bulk(const flash_area_t *area, int count,
   return sectrue;
 }
 
+secbool flash_area_erase_partial(const flash_area_t *area, uint32_t offset,
```
TychoVrahe, line 170: this needs also unix implementation for bootloader emulator to work. (or alternatively flash_common implementation and only implement flash_erase_sector for both f4 and unix, in flash common there is also get_sector_and_offset function that can can be used for getting the sector index)

core/embed/firmware/memory_T.ld
```diff
@@ -22,8 +23,9 @@ data_size = SIZEOF(.data);
 ccmram_start = ORIGIN(CCMRAM);
 ccmram_end = ORIGIN(CCMRAM) + LENGTH(CCMRAM);
 
-/* IMAGE_HEADER_SIZE is 0x400, this is for interaction-less firmware update start */
-firmware_header_start = ccmram_end - 0x400;
+/* reserve 256 bytes for bootloader arguments */
+boot_args_start = ORIGIN(BOOT_ARGS);
```
TychoVrahe, line 27: i think we should also take care and clear boot args at the start of firmware/prodtest/reflash in respective startup.s files.

core/embed/lib/image.c
```diff
@@ -224,35 +243,102 @@ secbool check_image_contents(const image_header *const hdr, uint32_t firstskip,
...
+
+    offset += bytes_to_check;
+  }
+
+  // Check the padding to the end of the area
+  end_offset = flash_area_get_size(area);
+
+  if (offset < end_offset) {
+    // Use the first byte in the checked area as the expected padding byte
+    // Firmware is padded with 0xFF, while the bootloader itself is padded with
```
TychoVrahe, line 273: to be super clear on this: bootloader is padded with zeros if it's installed via combined bin or bootloader update from firmware. When flashing bootloader separately via makefile for example, it will be padded with 0xFF too. 

We do allow here padding with any byte, but it only make sense for the images to be padded with 0x00 or 0xFF. Should we add a constraint?

TychoVrahe (changes requested):
The CI issues are still pending:
- boot args not defined in unix/emulator
- also the new flash_area_erase_partial funciton is not defined for unix, mentioned below too
- legacy does not compile due to the changes in reboot to bootloader message (nanopb is used there unlike in core firmware)

core/embed/boardloader/memory.ld
cepetr: I've left it intentionally here, but yes, it's not necessary and maybe confusing. Removed  66381328fab284b792145d8f3d733fb036cb4921

core/embed/trezorhal/stm32f4/flash.c
cepetr: unix implementation added 0636c0f50b2199f6f3f80782c82c3252a83f75ae.

core/embed/lib/image.c
cepetr: I've updated comment and added a check. see 1729331c329c80047115cd01d5d5c84116443a10

core/embed/boardloader/memory.ld
TychoVrahe: that commit removes it from bootloader, not boardloader. 

(i don't want it in boardloader because it again prevents the end of ccmram to be cleared)

core/embed/trezorhal/stm32f4/flash.c
cepetr: I slightly updated the comment to make it more clear.  It's not possible to erase just part of the sector, so function fails if the caller is not precise in specifying what to erase.  see fd93c86278676cf67b186afe7bae15548652a462

core/embed/lib/image.c
TychoVrahe: the ensure is wrong here, just return secfalse.

i.e. this function is called at the start of bootloader, if sore some reason the padding is not correct, the bootloader shutdowns and the user has no way of erasing the firmware and uploading something valid

core/embed/trezorhal/stm32f4/flash.c
TychoVrahe: btw are you using the underscore in parameter name to indicate it is output? because in rust its used to indicate that the parameter is unused, and it might be confusing to mix those meanings

core/embed/firmware/memory_T.ld
cepetr: I've added clearing of boot_arg, see 5eb2365acd075c7d95cb4b3a7da6ba29c5ce1f64

core/embed/trezorhal/stm32f4/flash.c
cepetr: Yes, it might be confusing. I've changed it. 8feb089e11a8b24c0df4775682fc3b4e237a0b11.

core/embed/lib/image.c
cepetr: yes, that's true. Fixed 1f6ac8c06319d379bc13a2e8ae1834627811ac7a

core/embed/boardloader/memory.ld
cepetr: Yeah, I was too fast. fixed d52365f2dc54c582266b47b9f154c6e893684663

core/embed/trezorhal/stm32f4/flash.c
TychoVrahe: it remained in the header file and the comment.

cepetr: fixed 1ff1235d27d1790071e05c3f1d49df41d998bc7e

cepetr: fixed b217e3c144650b89cfae5fe1a1929b6b7e015e3d

core/embed/bootloader/messages.c
```diff
@@ -667,22 +707,49 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
+      // write quad word (16 bytes) to the flash
+      ensure(
+          flash_area_write_quadword(&FIRMWARE_AREA, write_offset, quadword_ptr),
+          NULL);
+      write_offset += 4 * sizeof(uint32_t);
+      quadword_ptr += 4;
+    }
+    ensure(flash_lock_write(), NULL);
+    bytes_remaining -= bytes_to_write;
   }
```
matejcik, line 738: this piece of code is rather opaque and needs significantly more comments

core/src/apps/management/reboot_to_bootloader.py
```diff
@@ -7,20 +7,65 @@
...
+        await ctx.write(Success(message="Rebooting"))
+        # make sure the outgoing USB buffer is flushed
+        await loop.wait(ctx.iface.iface_num() | io.POLL_WRITE)
+        utils.reboot_to_bootloader(BootCommand.STOP_AND_WAIT)
+        raise RuntimeError
+
+    elif (
+        msg.boot_command == BootCommand.INSTALL_UPGRADE
+        and msg.firmware_header is not None
+    ):
```
matejcik, line 33: for slightly better forward compatibility, let's make this the main `if` branch, and the other version an `else` branch (so that an unrecognized argument goes to bootloader the usual way)

core/src/apps/management/reboot_to_bootloader.py
```diff
@@ -7,20 +7,65 @@
...
+                verb="INSTALL",
+            )
+
+            ctx = get_context()
+            await ctx.write(Success(message="Rebooting"))
+            # make sure the outgoing USB buffer is flushed
+            await loop.wait(ctx.iface.iface_num() | io.POLL_WRITE)
+            # reboot to the bootloader, pass the firmware header hash
+            utils.reboot_to_bootloader(BootCommand.INSTALL_UPGRADE, hdr["hash"])
+            raise RuntimeError
```
matejcik, line 68: this chunk of code should be shared between the two branches.
(so probably each branch sets `boot_command = X` and `boot_args = Y`, and then the common tail invokes `reboot_to_bootloader(boot_command, boot_args)`?)

core/SConscript.prodtest
```diff
@@ -109,7 +109,7 @@ tools.add_font('DEMIBOLD', FONT_DEMIBOLD, CPPDEFINES_MOD, SOURCE_MOD)
 tools.add_font('MONO', FONT_MONO, CPPDEFINES_MOD, SOURCE_MOD)
 tools.add_font('BIG', FONT_BIG, CPPDEFINES_MOD, SOURCE_MOD)
 
-env = Environment(ENV=os.environ, CFLAGS='%s -DPRODUCTION=%s' % (ARGUMENTS.get('CFLAGS', ''), ARGUMENTS.get('PRODUCTION', '0')), CONSTRAINTS=["limited_util_s"])
+env = Environment(ENV=os.environ, CFLAGS='%s -DPRODUCTION=%s' % (ARGUMENTS.get('CFLAGS', ''), ARGUMENTS.get('PRODUCTION', '0')))
```
matejcik, line 112: what's this about?

matejcik (changes requested):

matejcik (changes requested):
some more notes

(@mmilata maybe resolve the Python side while you're on the UI?)

core/SConscript.prodtest
cepetr: It was a bug, I've fixed. Limited_utils.s is intended for boardloader. Other apps have to use utils.s with the full functionality.

core/embed/trezorhal/stm32f4/flash.c
cepetr: I've fixed emulator build, see 9bb8f1df6934a387f5025c99e797305f7d7c1594. As we discussed, it's just temporary solution.

mmilata:
nevermind, i think i understand

core/embed/bootloader/messages.c
cepetr: I've added some comments, see  d18d6e46ac3bb225ba483ecec148a552216d441e

core/embed/extmod/modtrezorutils/modtrezorutils.c
```diff
@@ -245,18 +246,89 @@ STATIC mp_obj_t mod_trezorutils_unit_btconly(void) {
 STATIC MP_DEFINE_CONST_FUN_OBJ_0(mod_trezorutils_unit_btconly_obj,
                                  mod_trezorutils_unit_btconly);
 
-/// def reboot_to_bootloader() -> None:
+/// def reboot_to_bootloader(
+///     boot_command : BootCommand | None = None,
```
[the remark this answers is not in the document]
cepetr: fixed 14aa2246f50ed547feff1a051e58a20fdb5f6f2c

core/embed/extmod/modtrezorutils/modtrezorutils.c
```diff
@@ -245,18 +246,89 @@ STATIC mp_obj_t mod_trezorutils_unit_btconly(void) {
...
+        mp_obj_new_str_copy(&mp_type_str, info.vstr, info.vstr_len));
+    mp_obj_dict_store(
+        result, MP_ROM_QSTR(MP_QSTR_fingerprint),
+        mp_obj_new_bytes(info.fingerprint, sizeof(info.fingerprint)));
+    mp_obj_dict_store(result, MP_ROM_QSTR(MP_QSTR_hash),
+                      mp_obj_new_bytes(info.hash, sizeof(info.hash)));
+
+    return result;
+  } else
+    return mp_const_none;
```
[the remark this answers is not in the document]
cepetr: fixed 14aa2246f50ed547feff1a051e58a20fdb5f6f2c

core/src/apps/management/reboot_to_bootloader.py
cepetr: changed ProcessError to DataError, see 1bdf75b75a876a748fe92c1d7d01dfd62301efd0

matejcik:
> sorry, can you clarify which screen is that?

sorry, what I meant is "waiting for host"

core/embed/lib/image.c
```diff
@@ -224,35 +243,106 @@ secbool check_image_contents(const image_header *const hdr, uint32_t firstskip,
...
+    memcpy(info->vstr, vhdr.vstr, info->vstr_len);
   }
 
+  // copy firmware version
+  info->ver_major = ihdr->version & 0xFF;
+  info->ver_minor = (ihdr->version >> 8) & 0xFF;
+  info->ver_patch = (ihdr->version >> 16) & 0xFF;
+
+  // copy fingerprint from image header
+  memcpy(&info->fingerprint, ihdr->sig, sizeof(info->fingerprint));
```
mmilata, line 339: This copies the signature instead. I think you need to call `get_image_fingerprint` to obtain the fingerprint.

core/embed/bootloader/messages.c
```diff
@@ -410,8 +415,8 @@ static bool _read_payload(pb_istream_t *stream, const pb_field_t *field,
     // update loader but skip first block
     if (firmware_block > 0) {
       ui_screen_install_progress_upload(
-          250 + 750 * (firmware_block * IMAGE_CHUNK_SIZE + chunk_written) /
-                    (firmware_block * IMAGE_CHUNK_SIZE + firmware_remaining));
+          1000 * (firmware_block * IMAGE_CHUNK_SIZE + chunk_written) /
```
obrusvit, line 418: Is this intentional?

mmilata:
The UI + python changes are in #3371, feel free to merge or cherry-pick.

core/embed/bootloader/messages.c
cepetr: There is no longer an erase phase before writing the firmware into the flash memory, as there was in the past. So yes, this change is intentional.

core/embed/lib/image.c
cepetr: The input to this function consists of just the vendor and image header received in the RebootToBootloader message. We do not have the complete firmware binary at this point.

core/src/apps/management/reboot_to_bootloader.py
cepetr: @mmilata fixed the rest, see f62afd3947755a0645b92be72ca3208b175cdf2e

core/src/apps/management/reboot_to_bootloader.py
cepetr: refactored by @mmilata commit, see f62afd3947755a0645b92be72ca3208b175cdf2e

core/embed/lib/image.c
TychoVrahe: probably better to leave the field empty or even delete the field, since we can't know the fingerprint from headers only

matejcik: of course we can, the fingerprint is a hash of the header excluding the signature fields. the argument to `get_image_fingerprint` is just `hdr`

cepetr: fixed, now passing proper fingerprint, see 8a88f2c8bab88a161635fe1bdac5f9bdd0a099c0

matejcik (approved):
LGTM, now just waiting for @hiviah's review

core/embed/bootloader/main.c
```diff
@@ -413,6 +394,7 @@ int bootloader_main(void) {
   volatile secbool header_present = secfalse;
   volatile secbool firmware_present = secfalse;
   volatile secbool firmware_present_backup = secfalse;
+  volatile secbool auto_upgrade = secfalse;
```
hiviah, line 397: I think we should make this with the `_backup` with `ensure` version to prevent glitching. Not entirely sure if it's 100% required, but making upgrade harder to glitch is better. OTOH user could get to upgrade by pressing buttons, but maybe there is some weird combination of somewhat-glitched state that could make it useful.

matejcik:
minor nit: we should also check `auto_update` on line 489 to skip the 100ms "wait for touch" period.

TychoVrahe (approved):

matejcik (changes requested):
important feature that we forgot: ILU must only support full-trust images (so `(vhdr.vtrust & VTRUST_ALL) == VTRUST_ALL`)
it should not be possible to use ILU to update from unsigned to unsigned.

matejcik:
... firmware (python) doesn't need to check that (for this PR). It doesn't seem practical to export the vtrust flags to python just for this purpose, so maybe just "if vendor name is UNSAFE DO NOT USE" then reject or something dumb like that. going forward this might be a compile time flag, something coming from modtrezorutils or somesuch.

core/embed/bootloader_ci/messages.c
```diff
@@ -402,8 +402,6 @@ static bool _read_payload(pb_istream_t *stream, const pb_field_t *field,
   return true;
 }
 
-secbool check_vendor_header_keys(const vendor_header *const vhdr);
-
 static int version_compare(uint32_t vera, uint32_t verb) {
```
hiviah, line 405: I'm not entirely sure if this can or can't overflow/underflow due to integer promotion rules and so on. I.e. if you have highest bit set in `vera` or `verb`, is it going to behave correctly?
I see warning that narrowing `uint32_t` into `int` this way is implementation-defined, so compiler can theoretically change behavior even if it behaves correctly now in some optimization in later version. Someone should double-check that we are not depending on undefined/implementation-specific behavior.
I suspect that this might promote the `uint32_t` into `int`, and shift-right on `int` and right shift on negative integer is implementation-defined. Probably a check that makes it error if either `vera` or `verb` have highest bit set.

cepetr:
@matejcik I've added checking of vtrust field of vendor header, see 3ebf3d24be241040a43f0ceaa07fbb68e6e3a2a3 and a3119cb3b2c6f3023f2af81f26a2cd70130e6e39

hiviah (approved):

matejcik (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 23.11.1 (9a466944f069ad7aba27ac5aeef84dce12aa7e64)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.11.1 Chrome/116.0.5845.228 Electron/26.4.2 Safari/537.36
 - OS: MacIntel
 - Screen: 2560x1440

Issue #3355, opened 2023-10-18
Allow middle-click for unlocking T2B1


Hannsek:
<img width="169" alt="image" src="https://github.com/trezor/trezor-firmware/assets/57008159/bdbf5189-08d7-4b6c-b058-7109db20266d">

PR #3367, opened 2023-10-23
feat(core): make middle-click unlock the device for T2B1

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3355:
- middle-click also unlocks the device from lock-screen (in addition to single left or right clicks)


Hannsek:
If it works then ACK.

matejcik (approved):
utACK

bosomt:
**QA OK**

PR #3368, opened 2023-10-23
T2T1: update sign/verify message flow

### LINKED ISSUE #2831: UI2: Sign message
image

mmilata:
Fixes #2831. Introduces `chunkify` field to `SignMessage`, `VerifyMessage`, `EthereumSignMessage`, and `EthereumVerifyMessage`.

[UI diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/5357342803/artifacts/master_diff/index.html) | [differing screens](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/5357342803/artifacts/master_diff/index.html)

I haven't implemented the screen with _Show All_ button because I'm not convinced it solves a real problem. We've been telling our users that blind signing is bad yet this feature does it by default. Do we have an example of a real-world service which makes its users sign very long data?

0c1ecdf60d7043fbed68f7a97272ada053d531e2a599a2d078c1e5f81fb12f41  6caed08b89966b31a1308fb3ad4741c1588d38c0de54cb832a63396f9efb0a44 0b8c1ed4413a27a06a9c5ad2db3d1c68758cd5ae2afe3088f053dd3ebf76759f 9ce2a599b19e4bba87c145a04697863e57da54415e883febcf6c11cc2232f47b


Hannsek:
Thanks. 

I get your point, however when signing some message on ETH there are just a lot of random bytes, right? So user just has to go through all of them. @tomasklim do you have any example of a long eth message?

Plus maybe put the "Message sign" on the next screen. We will put there derivation path key in the future.

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -1150,32 +1153,80 @@ async def confirm_sign_identity(
...
     )
 
-    await confirm_blob(
-        br_type,
-        title,
-        message,
-        "Confirm message:",
-        hold=not verify,
-        br_code=BR_TYPE_OTHER,
+    items = []
```
grdddj, line 1181: Could be worth to type this ... `items: list[tuple[str, str]]`, so we have an extra check

core/src/apps/bitcoin/sign_message.py
```diff
@@ -35,11 +35,21 @@ async def sign_message(
 
     node = keychain.derive(address_n)
     address = get_address(script_type, coin, node)
+    path = address_n_to_str(address_n)
```
grdddj, line 38: Isn't this whole operation of getting `account` to display done already on some other place (it seems familiar to me)? Could be maybe refactored into a function

core/embed/rust/src/ui/model_tt/component/scroll.rs
```diff
@@ -21,7 +21,8 @@ impl ScrollBar {
     /// Center to center.
     const DOT_INTERVAL: i16 = 18;
 
-    fn new(layout: LinearPlacement) -> Self {
+    pub fn new(axis: Axis) -> Self {
```
grdddj, line 24: This `new` seems to not be used outside - has the `pub` some other idea? Seems little clearer to hide it when not used

core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -1135,7 +1135,12 @@ async def confirm_sign_identity(
 
 
 async def confirm_signverify(
-    coin: str, message: str, address: str, verify: bool
+    message: str,
+    address: str,
+    verify: bool,
+    path: str | None = None,
+    account: str | None = None,
+    chunkify: bool = False,
```
grdddj, line 1143: Will we implement the same `info` (and `chunkify`) functionality for `TR` as well? @Hannsek

(so we would actually use these function parameters)

grdddj (approved):
Code-wise LGTM, I have just some smaller points/questions that are not blockers for merge.

Also something from looking at the UI diff screens:

image

- theoretically, to account for the missing line, we could add a newline between `MUST` and `NOT`


image

- it is definitely an edge-case, but we might consider to show something like `<message empty>`? Or not, as it would only make our life harder with translations of this string?

core/src/trezor/ui/layouts/tr/__init__.py
Hannsek: It is already implemented, no? Except of the "SHOW ALL" screen.

Hannsek:
If you would sign [some contract on ETH](https://etherscan.io/tx/0x835623ebe44a42ee125365b74ba44bf1907343098993da331452365099f5a4d5), it looks like this:

on T1: 
image
user cannot even check all transaction data

on TT:
image
image
similar to currently proposed UX.


So please implement also the `Show all` button together with other buttons. 🙏🏻

mmilata:
Eth contracts are handled in different flow, and they are always bunch of hexadecimal nonsense. This is not the case for SignMessage.

> Plus maybe put the "Message sign" on the next screen. We will put there derivation path key in the future.

What do you mean? Derivation path is already in the info.

core/embed/rust/src/ui/model_tt/component/scroll.rs
mmilata: It's used in [simple_page.rs](https://github.com/trezor/trezor-firmware/blob/fc4f8bfa70aaeaba943306318c592df95a05ead3/core/embed/rust/src/ui/model_tt/component/simple_page.rs#L32).

core/src/trezor/ui/layouts/tt/__init__.py
mmilata: 0165c99fa6e8204325e576ac3d3b9da13f04dba7

mmilata:
> theoretically, to account for the missing line, we could add a newline between MUST and NOT

Done in 0165c99fa6e8204325e576ac3d3b9da13f04dba7.

> it is definitely an edge-case, but we might consider to show something like <message empty>? Or not, as it would only make our life harder with translations of this string?

But how do we distinguish empty message from nonempty message that says `<message empty>`?

grdddj:
> But how do we distinguish empty message from nonempty message that says `<message empty>`?

Yes, that is another reason why not to do it, so forget my proposal.

core/src/trezor/ui/layouts/tr/__init__.py
Hannsek: on TR, it should be like this:
image

But we can change it to:
image
🤔

bosomt:
not yet implemented in Trezor Suite
checked with TT and S3 new 2.64 FW

**Info:**
 - Suite version: desktop 23.12.1 (9ec6ac5438d7a952abcbbdc049bf729a47d4e039)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.12.1 Chrome/118.0.5993.129 Electron/27.0.4 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.6.4 regular (revision 42e9ed0e09033d474dee1a560fe5870646fa440e)

Hannsek:
Can you please post here screenshots from Suite and fw side by side on Both, T2T1 and T2B1? 🙏🏻

bosomt:
23.12.1 + 2.6.4 FW 

telegram-cloud-photo-size-4-6033095966502010744-y
telegram-cloud-photo-size-4-6033095966502010745-y
telegram-cloud-photo-size-4-6033095966502010746-y
telegram-cloud-photo-size-4-6033095966502010747-y

Hannsek:
Ahh, ok, then it is not the same on both devices. I'll create another issue regarding this. @mmilata

PR #3371, opened 2023-10-25
Interaction-less update: UI

mmilata:


matejcik:
all cherry picked into #3363 so closing in favor of that

reviewed and LGTM too

PR #3373, opened 2023-10-26
docs(core): differentiate changelog entries between models

### LINKED ISSUE #3159: Differentiate changelog entries between core models
We should be able to produce changelog for T2T1 independent of changelog for T2B1, and not include T2B1-specific items for T2T1 and vice versa.

Probably as simple as creating a convention for changelog entries, e.g., `[T2B1] Changelog goes here` for model-specific?

mmilata:
Fixes #3159. Not sure if we actually want the `CHANGELOG.XXXX.md` files but we should at least update the docs.

Rendered:
* [CHANGELOG.md](https://github.com/trezor/trezor-firmware/blob/mmilata/changelog-models/core/CHANGELOG.md)
* [CHANGELOG.T2B1.md](https://github.com/trezor/trezor-firmware/blob/mmilata/changelog-models/core/CHANGELOG.T2B1.md)
* [CHANGELOG.T2T1.md](https://github.com/trezor/trezor-firmware/blob/mmilata/changelog-models/core/CHANGELOG.T2T1.md)

### DIFF
--- a/core/CHANGELOG.T2B1.md
+++ b/core/CHANGELOG.T2B1.md
@@ -1,0 +1,798 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [2.6.2] (internal release)
+
+### Added
+- Integrate Optiga into PIN verification.  [#3296]
+
+
+## [2.6.1] (internal release)
+
+### Added
+- Implement UI.  [#2610]
+- QR code display when exporting XPUBs.  [#3047]
+- Added hw model field to all vendor headers.  [#3048]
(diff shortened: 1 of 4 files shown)


matejcik:
@matejkriz @Hermez-cz does this look good to you?

matejcik (approved):
LGTM codewise but let's wait for Suite guys to see the results

matejkriz:
Technically, I would expect T2B1 changelog starting with 2.6.1 and filtering out versions with no changes related to the model like this https://github.com/trezor/trezor-firmware/blob/mmilata/changelog-models/core/CHANGELOG.T2T1.md#262-internal-release. But the product priority seems to be low, so it's up to @Hermez-cz if and when we want to deal with this...

matejcik:
let's keep it as-is then

PR #3374, opened 2023-10-26
T2T1: nudge page_prev_ellipsis to the right

mmilata:
the bug and the fix
Screenshot 2023-10-27 at 01-01-40 TT-device_tests-cardano-test_address_public_key py test_cardano_get_address True-parameters24-result24
[UI diff](tbd) | [differing screens](tbd)


matejcik (approved):
ACK but how in the world does it solve the stray ellipsis in the bottom left picture??

mmilata:
I think it just moves the non-black pixels to the area that's cleared when changing page?

There must be some kind of placement bug somewhere (did not see negative offsets used), this is just a workaround that also looks better IMO.

Hannsek:
Nice! Ideal state is still https://github.com/trezor/trezor-firmware/pull/3273.

PR #3383, opened 2023-11-01
Bootloader release 2.1.4

matejcik:
Just generate a changelog and bump version. Commit 22e48717a67171b14de0bec0921cf9389bb3247f -- or what it ends up after merging -- will be tagged `core/bl2.1.4` for signing.

### DIFF
--- a/core/embed/bootloader/CHANGELOG.md
+++ b/core/embed/bootloader/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 2.1.4 [November 2023]
+
+### Added
+- Minimize risk of losing seed when upgrading firmware.  [#2794]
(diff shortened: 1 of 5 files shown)


mmilata (approved):

core/embed/bootloader/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 2.1.4 [November 2023]
+
+### Added
+- Minimize risk of losing seed when upgrading firmware  [#2794]
+- Support interaction-less upgrade  [#2919]
```
mmilata, line 11: nit
```suggestion
- Minimize risk of losing seed when upgrading firmware.  [#2794]
- Support interaction-less upgrade.  [#2919]
```

matejcik: fixed in rebase. we should specify this as policy and ideally auto-check or auto-add the periods

PR #3384, opened 2023-11-01
New vendor headers for TS3

matejcik:
plus slightly improved vendor header selection logic in SCons

Please review the generated binaries too. Don't forget that in the poetry shell you can use `toiftool info` and `toiftool show`. Use `./core/tools/headertool.py` to examine generated headers.

This should be merged only after the new vendor headers are signed and the appropriate `_signed_prod.bin` files are added. In the current state builds will fail because these files are missing.


mmilata (approved):

mmilata:
`make build_firmware TREZOR_MODEL=R PRODUCTION=0` now fails because of missing `vendorheader_unsafe_signed_dev.bin`.

I'm gonna go ahead, merge this, and solve later as it doesn't affect release.

PR #3386, opened 2023-11-02
Embed signed bootloaders 2.1.4

mmilata:
For T2B1 and T2T1.


matejcik (approved):
utACK

is there a sort call missing in bootloader_hashes?

mmilata:
yep #3387

PR #3390, opened 2023-11-02
Update TS3 vendor headers

mmilata:
* add correct btconly VH
* add signed "unsafe" VH


matejcik (approved):
utACK

mmilata:
Added signed VH in 30da630c813f05dcbf51a46d6631f8674d995920

Issue #3392, opened 2023-11-06
Align send flow button request on T2B1 with TT


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/36012c26-ef0f-4941-acee-57babef084e8)

Currently, when user clicks CONTINUE, it won't send button request to Suite.

Hannsek:
@grdddj Can you please check the differences between TS3 and TT button requests on different flow? Cardano, btc, altcoins, …

grdddj:
> @grdddj Can you please check the differences between TS3 and TT button requests on different flow? Cardano, btc, altcoins, …

Right, I have noticed at least `modify-amount` flow to be different - fixed as well in https://github.com/trezor/trezor-firmware/pull/3393

Hannsek:
@grdddj is this done?

grdddj:
> @grdddj is this done?

Will be done after merging https://github.com/trezor/trezor-firmware/pull/3393

PR #3393, opened 2023-11-07
ETH send flow - unify button requests

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3392:
- send the same button requests in ETH send flow for T2B1 as for T2T1


core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -1066,21 +1080,46 @@ async def confirm_modify_output(
...
+        await raise_if_not_confirmed(ctx_wait(address_layout))
+
+        if send_button_request:
+            send_button_request = False
+            await button_request(
+                "modify_output",
+                ButtonRequestType.ConfirmOutput,
+                modify_layout.page_count(),
+            )
+        result = await ctx_wait(modify_layout)
```
mmilata, line 1118: I'm afraid you'll have to call `.request_complete_repaint()` on the two layouts otherwise they'll paint over each other.

mmilata (commented):
Looking good, would like to see UI diffs before approving.

tools/check-bitcoin-only
```diff
@@ -10,7 +10,7 @@ EXCEPTIONS+=( "mnemonic" )  # has NEM in it
 EXCEPTIONS+=( "workflow" "overflow" )  # has Flo in it
 EXCEPTIONS+=( "SyntaxError" )  # has Axe in it
 EXCEPTIONS+=( "DKDNEM" )  # has NEM in it, some sort of weird coincidence
-EXCEPTIONS+=( "confirm_ethereum_tx" )  # is model-specific, so is in layout/__init__.py instead of ethereum/layout.py
+EXCEPTIONS+=( "confirm_ethereum_tx ethereum_tx_summary" )  # is model-specific, so is in layout/__init__.py instead of ethereum/layout.py
```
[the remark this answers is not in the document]
grdddj: Done in [a5d3e2f](https://github.com/trezor/trezor-firmware/pull/3393/commits/a5d3e2f3273b2782bf3fa37744f8075aa218ac4c)

grdddj:
> Looking good, would like to see UI diffs before approving.

Rebased on `main` so we see the correct UI diff

core/src/trezor/ui/layouts/tr/__init__.py
grdddj: Is this situation different in `TR` than in `TT`? This is the same exact code as in `confirm_modify_output`'s code in `core/src/trezor/ui/layouts/tt/__init__.py`

mmilata: Seems to be the same & in both cases there is garbage on screen when you go back from the MODIFY AMOUNT screen in `test_p2wpkh_in_p2sh_fee_bump_from_external`.

TBH this should be handled automagically somewhere in `class Layout`, not sure if it's worth spending time on fixing it or rewriting to rust ...

Hannsek:
vertical alignment has changed for some reason:

core/src/trezor/ui/layouts/tr/__init__.py
grdddj: Good catch. I did the quick fix in [4970a6b](https://github.com/trezor/trezor-firmware/pull/3393/commits/4970a6bdce3c90b58cef13a3edca60dc9c643dfa) --- requesting repaint in all `while True:` cases where we define the layout in front of the loop

grdddj:
> vertical alignment has changed for some reason: image
> 
> diff on TT is not available.

Yes, the alignment might have changes a little, because we are using a different layout for this screen. Hopefully, it is not an issue.

I have pushed new commits, so the `TT` diff should be visible in a moment

Hannsek:
> Yes, the alignment might have changes a little, because we are using a different layout for this screen. Hopefully, it is not an issue.

Can we somehow easily fix it?

matejcik:
is this ready to merge (after fixing the conflict)?

grdddj:
> is this ready to merge (after fixing the conflict)?

From my side yes. Feel free to take over the PR

matejcik (approved):

bosomt:
**QA OK**

already released via 2.6.4 firmware version

PR #3395, opened 2023-11-08
chore(core): replace master branch comparison with main branch

grdddj:
Changing the comparison branch (for CI artifacts) from outdated `master` to current `main`


prusnak:
Do we want to also rename the script or it's just not worth the hassle?

grdddj:
> Do we want to also rename the script or it's just not worth the hassle?

I think it is not worth it, we already decided to keep the `master-diff` naming for UI comparison, this is a similar situation

prusnak (approved):
utACK

Issue #3280, opened 2023-09-17
Device recovery tests get stuck on T2B1


mmilata:
**Describe the bug**
Some tests in `test_recovery_bip39_dryrun.py`, `test_recovery_slip39_basic.py`, `test_recovery_slip39_advanced.py` get stuck when running on T2B1 hardware. Serial console logs an exception `RuntimeError: Waiting for event that already happened - 21 > 1` (different numbers not just 21).

**Firmware version and revision**
master @ 21959996bda33c6ca3e2e6f6fe89b7072cb767d3

**To Reproduce**
Only seems to happen after several tests have run. E.g. `pytest -v ../tests/device_tests/ -k test_recovery_bip39_dryrun` is sufficent to trigger this here. Client gets stuck waiting and firmware throws exception:
```
23327000 trezor.loop ERROR exception:
Traceback (most recent call last):
  File "trezor/loop.py", line 162, in _step
  File "apps/debug/__init__.py", line 127, in return_layout_change
  File "apps/debug/__init__.py", line 117, in get_layout_change_content
RuntimeError: Waiting for event that already happened - 21 > 1
```

Issue #2724, opened 2023-01-09
T2B1: make sure all tests are updated for the new model


matejcik:

grdddj:
`Click` and `device` tests are supported fully. Not sure what other kinds of tests (`persistence`, `upgrade`) do we want to run for `TR`

Hannsek:
Are there any additional tests we want?

swageeth:
Is model R estimate to be released on September 18?

mmilata:
I think we need to run a T2B1 version of *every* existing T2T1 test job. Notably also `u2f` and `fido2`.

Hannsek:
Can we close this issue or persistence and upgrade tests are still missing?

Not a high prio, right?

mmilata:
FIDO2 tests only need the T2B1 equivalent of trezor/fido2-tests#4

matejcik:
persistence tests enabled in #3686

PR #3396, opened 2023-11-08
ci: convert build and test jobs to github actions

mmilata:
Initial port of core+legacy build+test jobs from GitLab CI. What's missing:
* deploy jobs
  * credentials required
* HTML reports from UI jobs
  * GHActions only make it possible to download artifact zip file, we might have to push the HTMLs elsewhere for reviewer convenience
* caching
  * the runtime is still comparable with GitLab CI however, and if I understand this correctly we don't pay for compute time in the public repo
  * pipeline eats about 3h compute time, which means <10 pipelines for free in the private repo
* arm/macos builds
* connect test
* hardware tests

There are also bunch of new tests for T2B1 that we didn't have before, but there should still be more - #2724. The T2B1 device tests are sometimes failing due to #3280, no idea why it doesn't manifest on GitLab.

My proposal is to review & merge this PR and let both systems exist for a while until the issues are resolved.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions


vdovhanych:
Deploy jobs could be done by uploading it to data.trezor.io/dev/firmware or better we would create a separate domain for that. Nonetheless we could do it by uploading files via aws cli like [here](https://github.com/trezor/trezor-suite/pull/9406/files#diff-b96b8b0583ecd1ae2eec82983a96fd7bd68aa3f8f99dc2246f40987f7ceecf26R57) also with setup credentials step. 

arm build will most likely be done with docker buildx cross platform build same way as its now using nix docker image.

.github/workflows/core.yml
```diff
@@ -0,0 +1,549 @@
...
+      - run: nix-shell --arg fullDeps true --run "unset _PYTHON_SYSCONFIGDATA_NAME && cd HWI && poetry install && poetry run ./test/test_trezor.py --model_t ../core/build/unix/trezor-emu-core bitcoind"
+      - uses: actions/upload-artifact@v3
+        with:
+          name: core-test-hwi-${{ matrix.model }}
+          path: HWI/trezor-t-emulator.stdout
+          retention-days: 7
+
+  core_memory_profile:
+    name: Memory allocation report
+    if: false  # NOTE manual job, comment out to run
```
vdovhanych, line 328: how would this work, you would uncomment it every time you want it to run or is this just for testing purpose for now? I think we cant have manual jobs in one workflow, it would need to be a separate one.

vdovhanych (approved):
Let's merge this and improve it later by adding all the to-do stuff.

.github/workflows/core.yml
mmilata: Yes, needs to be uncommented before every run, as it's very rarely used. I may turn it into a manual workflow later.

PR #3399, opened 2023-11-13
Fix CI bootloader build for T2B1

mmilata:
It builds now but I'm not able to test it. I'd be very grateful if somebody with a devkit can confirm that it always waits for firmware update after reboot and installs it without any user interaction.


TychoVrahe (approved):
added a commit fixing an old bug causing failures while uploading fw, otherwise it works as described. No GUI is shown, i assume its ok.

PR #3407, opened 2023-11-17
improve T2B1 homescreen checks

### LINKED ISSUE #3263: T2B1 homescreen checks
The T2B1 homescreen code does not seem to check the format correctly, it just checks that the image is grayscale but does not care about the TOIg/TOIG difference - we should only accept TOIG as only that we render correctly. 

Also there is unwrap [here](https://github.com/trezor/trezor-firmware/blob/a30c59ccde4759b45bc9d2776e61fadba7b21587/core/embed/rust/src/ui/model_tr/component/homescreen.rs#L58) that will crash the FW if toif cannot be constructed from the stored data, which is probably undesirable - we should just show the default image.

TychoVrahe:
closes #3263 

- Only TOIG is accepted as homescreen format.
- Default image is painted if something else is stored in storage.

### DIFF
--- a/core/embed/rust/Cargo.lock
+++ b/core/embed/rust/Cargo.lock
@@ -73,9 +73,9 @@ dependencies = [
 
 [[package]]
 name = "cstr_core"
-version = "0.2.4"
+version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "917ba9efe9e1e736671d5a03f006afc4e7e3f32503e2077e0bcaf519c0c8c1d3"
+checksum = "dd98742e4fdca832d40cab219dc2e3048de17d873248f83f17df47c1bea70956"
 dependencies = [
  "cty",
  "memchr",
--- a/core/embed/rust/Cargo.toml
+++ b/core/embed/rust/Cargo.toml
@@ -90,8 +90,9 @@ default_features = false
 version = "0.3.3"
 
 [dependencies.cstr_core]
-version = "0.2.4"
-default_features = false
+version = "0.2.6"
+default-features = false
+features = ["nightly"]
 
 # Build dependencies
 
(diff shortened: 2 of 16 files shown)


TychoVrahe:
On second thought, i decided to move the homescreen format check to the homescreen module in rust. The format is dependent on the homescreen implementation, the check code is reused there and it eliminates the need for model-specific code in python. I didn't want to reimplement detailed error messages (format check failing due to dimensions, not being grayscale etc), i don't think its needed as it is also checked in CLI or suite, so unified message should be enough.

matejcik:
I wanted to suggest a refactor of the TR version of homescreen check, but it involved changes that are somewhat more spread out. Please check my commits.

TychoVrahe:
ACK for the TR refactor and related changes.

matejcik (approved):

matejcik:
@TychoVrahe can you provide a `TOIg` homescreen image for QA to test with?

TychoVrahe:
[homescreen_legacy.zip](https://github.com/trezor/trezor-firmware/files/13685062/homescreen_legacy.zip)

sorry for the delay.

bosomt:
```
trezorctl set homescreen homescreen_legacy.toif
Error: TOIF file is corrupted
```

**Info:**
 - Suite version: desktop 23.12.3 (8fd6a446f2f03d4c8407e1f5f97a804f315a0253)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.12.3 Chrome/118.0.5993.129 Electron/27.0.4 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.6.5 regular (revision f4d3a7456afaa850f9651d31ddc9d7e4b68ab7b8)

bosomt:
can i approve this ?
T2B1 FW works with Suite homescreens as expected

matejcik:
:yes:

Issue #3403, opened 2023-11-16
Fix chunkifying issue in cardano send flow


Hannsek:
Chunks in receive flow display 4 4 4 4 symbols.

In send flows it however displays 4 4 4 3 symbols.
This should be fixed.

Receive flow:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/4b57a82d-aa7b-4168-95fd-67c35ba36cf3)
Send flow:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/e409b961-5945-4c97-a54b-7b5772ba2eba)

PR #3409, opened 2023-11-21
fix(core): fix the chunkification of longer Cardano addresses in send flow

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3403:
- decrease the horizontal space in chunkified view when there is a scrollbar on the right

image

Diff: https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/5576547749/artifacts/test_ui_report/differing_screens.html


Hannsek:
So the horizontal space is the same as in receive flow now?

grdddj:
> So the horizontal space is the same as in receive flow now?

Yes, they share the same chunkified properties

mmilata (approved):
LGTM

Hannsek:
@grdddj I oversaw one mistake. In cardano send flow we currently display 5 rows instead od 4 as with any other coins. Suite team is not happy because of that. Can we please fix it? 🙏🏻

grdddj:
> @grdddj I oversaw one mistake. In cardano send flow we currently display 5 rows instead od 4 as with any other coins. Suite team is not happy because of that. Can we please fix it? 🙏🏻

Right, done in https://github.com/trezor/trezor-firmware/pull/3413

bosomt:
**QA OK**

checked via
https://connect.trezor.io/9/#/method/cardanoGetAddress

PR #3412, opened 2023-11-22
Update to latest nixpkgs-unstable, rustc

### LINKED ISSUE #3337: Update nixpkgs
- [ ] see if python tests pass with 3.12 #3336

mmilata:
Fixes #3337.


ci/shell.nix
```diff
@@ -75,22 +75,23 @@ let
   }));
   llvmPackages = nixpkgs.llvmPackages_14;
   # see pyright/README.md for update procedure
-  pyright = nixpkgs.callPackage ./pyright {};
+  # XXX why are we even building it from source? # pyright = nixpkgs.callPackage ./pyright {};
```
prusnak, line 78: I think that it was not packaged in nixpkgs before?

mmilata: i vaguely recall that we needed one particular version but looks like there's no reason anymore (re #3767)?

matejcik: i'm afraid we'll need to keep the current system, pyright is still very much in flux and new breaking changes appear in every release. i had to wait for two bugs to get fixed in 1.1.360 before being able to do #3767 
we can revisit this in a year again, see if things have calmed down

PR #3413, opened 2023-11-22
feat(core): set maximum number of chunkified lines to 4

grdddj:
Extends https://github.com/trezor/trezor-firmware/pull/3409:
- make sure the chunkified address does not exceed 4 rows per page

image


Hannsek:
Looks good.

One more thing, I noticed there is different header then expected. The whole flow should be aligned with btc send flow. So it should display "RECIPIENT". But we'll cover that in different issue where I want to edit the whole Cardano send/receive flow.

matejcik (approved):

Issue #3246, opened 2023-09-04
ETH fees units


Hannsek:
Please change ETH fees units in ETH send flow as follows:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/ab25827b-8c64-40dd-be45-18d9abe7b51a)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/de0fc492-6c07-49a1-a586-08bd3f0c4284)

Maximum fee: wei ETH -> ETH
Gas price: ETH -> Gwei

Add (if not already satisfied) priority fee key.

Hannsek:
@grdddj please take a look, it should be low hanging.

grdddj:
> @grdddj please take a look, it should be low hanging.

Right, PR done - https://github.com/trezor/trezor-firmware/pull/3414

It was quite easy to change, but I am not sure we want it like this --- please check the results/diffs

PR #3414, opened 2023-11-22
chore(core): hardcode ETH and Gwei units in ETH send summary

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3246:
- enforces usage of certain units in ETH summary - `ETH` for amount and total fee, `Gwei ETH` for the rest.

Not sure, however, that we want to force it, as there are these cases where a very small value gets displayed in a very big unit. Previously, the code was checking this --- either showing `ETH` if the value was bigger than `10 ^ 9` and `wei ETH` otherwise.

image


Hannsek:
Ahhh, ok, it cannot be like that :(
For _Maximum fee_ and _Amount_ please revert the changes.

> Gwei ETH for the rest.

We should use only Gwei/Wei (without ETH) for gas limit a and priority fee.

matejcik:
i think we just need to touch the "10^9" number -- so that we get at most, say, 10 decimals?

grdddj:
> Ahhh, ok, it cannot be like that :( For _Maximum fee_ and _Amount_ please revert the changes.
> 
> > Gwei ETH for the rest.
> 
> We should use only Gwei/Wei (without ETH) for gas limit a and priority fee.

Reverted and changed in [b428d53](https://github.com/trezor/trezor-firmware/pull/3414/commits/b428d533c735c9b8d43161198e4b01f2943e75fa)

grdddj:
> i think we just need to touch the "10^9" number -- so that we get at most, say, 10 decimals?

Not sure what are the requirements ... but we have good unittests for this function, feel free to suggest some new behavior

https://github.com/trezor/trezor-firmware/blob/main/core/tests/test_apps.ethereum.layout.py

There are some edge-cases like this, where there will be much more than 10 valid digits:
```python
text = format_ethereum_amount(1000000000000000001, None, ETH)
self.assertEqual(text, "1.000000000000000001 ETH")
```

@Hannsek  - a general question, who is wanting these changes?

Hannsek:
It's me. Our fee units are not standard.

It should look like this: 
image

grdddj:
> It's me. Our fee units are not standard.
> 
> It should look like this: image

Should be the case now. Priority fee is only shown in `eip1559`, which looks like we have test-screens only for `TR`

image

image

Hannsek:
Why do we have `max gas price` on T2B1 and `gas price` on TT. We should have it the same on both devices. Please use only `gas price`.

grdddj:
> Why do we have `max gas price` on T2B1 and `gas price` on TT? We should have it the same on both devices. Please use only `gas price`.

That is not the difference between models, but between `1559` and `normal` signing. All these texts are the same for both devices

mmilata (approved):

PR #3415, opened 2023-11-22
feat: update Rust client

DaniPopes:
Updates dependencies to latest version and regenerates bindings.

Would be nice if a release could be made after this PR, since we're planning to use this crate in [Alloy](https://github.com/alloy-rs/core).


matejcik (approved):
utACK

PR #3417, opened 2023-11-24
Fix invalid encoding of signatures from Optiga

### LINKED ISSUE #3411: Fix invalid encoding of Optiga signatures
The Optiga chip sometimes seem to produce invalid DER encoding of ECDSA signatures: where DER specifies that INTEGERs must be prefixed by 0x00 if and only if the leading byte is >= 0x80, Optiga seems to always(?) add the 0x00 prefix. The resulting DER encoding is invalid -- but can be easily fixed by stripping the wrong 0x00 bytes.

We need to fix this in two places:
* `optiga_calc_sign()` when producing the signature, so that all results (incl. e.g. in `AuthenticateDevice`) already come out correct, and
* when providing stored device certificate, which might have been signed by a buggy optiga too

andrewkozlik:
Fixes #3411. Tested on development device using 8 `AuthenticateDevice` calls. There is no guarantee that any of the signatures were corrupt, but there is a unit test in `crypto`, which should check that integers get reencoded correctly.


crypto/buffer.c
andrewkozlik, line 1: Adapted from io.c from trezor-provisioning.

crypto/buffer.h
```diff
@@ -0,0 +1,59 @@
...
+typedef struct {
+  const uint8_t *data;
+  size_t size;
+  size_t pos;
+} BUFFER_READER;
+
+// Buffer writer struct, wraps a pointer to non-const data.
+// Always should have the same layout as BUFFER_READER. We consider it legal to
+// cast an BUFFER_WRITER pointer to an IOreader pointer. However, NOT the other
+// way around.
```
andrewkozlik, line 40: I am not actually convinced that it makes sense to cast `BUFFER_WRITER` to `BUFFER_READER`.

Note to self: `IOreader` -> `BUFFER_READER`.

matejcik: the only reason for that in io.c was to reuse something like io_remaining which was called from everywhere. in this adaptation that is no longer the case, and you could do buffer_writer_remaining() and buffer_reader_remaining() separately either way

matejcik:
shouldn't most of the functions in `der.h` and `buffer.h` be tagged `__wur__` ?

matejcik:
code LGTM, but let's add the `__wur__`s if there is not a problem with it that i'm not seeing

matejcik (approved):

PR #3420, opened 2023-11-25
feat(rust/trezor-client): add `get_ecdh_session_key` method

romanz:
Please let me know if there are any unit/device tests that can be added.


matejcik:
the CI has an emulator running with the all-all-all seed so you could add a smoke test at least, see the tests in `lib.rs`

Hannsek:
@romanz Will you add unit tests please?

romanz:
Yes, will do on the weekend.

matejcik:
crap, i didn't realize that we don't have the debuglink autoresponder in Rust yet

let's keep the test but mark it `#[ignore]` until we get to making debuglink work

romanz:
Reformatted and marked the new test with `#[ignore]` -> 9fea5194b.

matejcik (approved):

PR #3423, opened 2023-11-28
fix(core): restore filled-lock display time

matejcik:
in #3222 the `_WELCOME_SCREEN_MS` code got lost, which means that in 2.6.3 the user is staring at the empty lock for several seconds and on the filled lock for only a fraction of a second. which looks a bit like a bug.

i believe the removal is a result of a bad rebase rather than intention, right @TychoVrahe ?

### DIFF
--- a/core/embed/firmware/main.c
+++ b/core/embed/firmware/main.c
@@ -47,6 +47,7 @@
 #include "model.h"
 #include "mpu.h"
 #include "random_delays.h"
+#include "rust_ui.h"
 
 #include TREZOR_BOARD
 
@@ -110,6 +111,8 @@ int main(void) {
 #endif
 
   display_reinit();
+
+  screen_boot_full();
 
 #if !defined TREZOR_MODEL_1
   parse_boardloader_capabilities();
--- a/core/embed/rust/librust_qstr.h
+++ b/core/embed/rust/librust_qstr.h
@@ -67,7 +67,6 @@ static void _librust_qstrs(void) {
   MP_QSTR_description;
   MP_QSTR_details_title;
   MP_QSTR_disable_animation;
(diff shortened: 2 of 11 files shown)


TychoVrahe:
Not sure anymore, might as well have been intentional as i hate the delay and recall lobbying against it. 

Anyway, if this is a must, could we move the full lock drawing to C right after display initialization and shorten the `_WELCOME_SCREEN_MS` proportionally?

Or maybe we could revisit #2521 where the move to C is handled out of necessity: https://github.com/trezor/trezor-firmware/pull/2521/files#diff-353348c881a5795cffe70dc7c7f7d917150de906ea40873b62e08c4c98c1098e

matejcik:
drawing the full lock in C would probably work OK, and given how long micropython initialization takes, maybe we don't need to enforce _any_ delay in Python.

#2521 is of course the ultimate solution, but we need something for the upcoming 2.6.4 release :)

matejcik (commented):
utACK, can't approve my own PR so feel free to go ahead

TychoVrahe (approved):

bosomt:
**QA OK**

Info:
 - Suite version: desktop 23.11.5 (4566edae060459a5645f0ddabde574b998d53c53)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/23.11.5 Chrome/118.0.5993.129 Electron/27.0.4 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.6.4 regular (revision 1f693d3561c375661563970e066f678590abc19c)

PR #3424, opened 2023-11-28
Fix missing confirm_address

matejcik:
Due to an innocent looking mistake in code, all calls to `confirm_address` have been silently skipped on TS3. This resulted in several skipped confirmation screens all over the place.

This PR fixes the problem and adds a lint to check for it in the future.


grdddj (commented):
NIce job!

The `pylint` worked fine for me locally, even though the speed is disastrous. Did you consider or try creating `flake8` plugin? It seems to be much quicker. Unfortunately, `ruff` still seems not supporting user plugins. (Sure, `pylint` runs anyway, but it would be nice to get rid of it, so for possible future plugins)

At the first glance, I am not very happy about the different signatures of `confirm_address` between `TR` and `TT` layout code --- was there any reason to not use the second option `async def confirm_address() -> None` (which should be possible with `return await confirm_blob()`)?

matejcik:
> was there any reason to not use the second option `async def confirm_address() -> None` (which should be possible with `return await confirm_blob()`)?

the sync option (`def foo() -> Awaitable[x]`) is always preferable because calling an async function allocates a new awaitable, while calling a synchronous function is free.

> am not very happy about the different signatures of `confirm_address`

yes, the TT functions should be converted to what TR does, but I didn't want to do that here. (and it might be moot anyway, as we'll soon be able to get rid of those functions)

> the speed is disastrous.

does it seem to you that it's slower than it was before? pylint is super slow generally but I didn't notice this making it worse

grdddj:
Thanks for the clarifications.

> > the speed is disastrous.
> 
> does it seem to you that it's slower than it was before? pylint is super slow generally but I didn't notice this making it worse

I did not notice any worsening either, so it was a complaint about general `pylint` - this one quick plugin should not make it any slower.

grdddj (approved):

mmilata (approved):
Awesome, looking forward to writing some more checks.

PR #3428, opened 2023-11-29
fix(core/ui): horizontal scrolling in transaction summary details

mmilata:
Needs solana PR merged.


Hannsek:
@mmilata what is the status of this? Wasn't this in the solana pr?

mmilata:
Do we still want it?

> We should probably still merge it because it affects all coins and according to figma the scrolling should be horizontal everywhere (transaction summary details), right? Probably didn't catch this because most coins only have one screen worth of info

Hannsek:
We won't have this behaviour in T3T1, there will be if needed vertical swipe.

Hannsek:
What does this pr do?

PR #3435, opened 2023-12-01
Elligator 2

onvej-sl:
This pull request implements [elligator 2](https://www.rfc-editor.org/rfc/rfc9380.html#ell2-opt).


crypto/elligator2.c
```diff
@@ -0,0 +1,230 @@
...
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES
+ * OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
```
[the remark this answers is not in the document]
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/3435/commits/9acdaed06b7b6598b9e432644359801f55198379.

onvej-sl:
In https://github.com/trezor/trezor-firmware/pull/3435/commits/db58a0d0f98a3f0e726b76abc29e9580ebad7dc9, I changed the api from
```c
bool map_to_curve_elligator2_curve25519(const bignum25519 input, curve25519_key output)
```
to
```c
bool map_to_curve_elligator2_curve25519(const uint8_t input[32], curve25519_key output)
```
.

onvej-sl:
@M1nd3r Please review `crypto/**`
@matejcik Please review `core/**` yourself or assign some else to do it.

core/src/trezor/crypto/__init__.py
```diff
@@ -9,6 +9,11 @@
     random,
 )
 
+try:
+    from trezorcrypto import elligator2  # noqa: F401
+except Exception:
+    pass
+
```
matejcik, line 16: this code would run on production firmware
maybe hide the try-except inside an `if __debug__:` at least? and add a comment that this is for an upcoming feature?

onvej-sl: For now, it will be only used in the new THP. So, it might give more sense to hide it this way:
```
if utils.THP_VERSION == 3:
    from trezorcrypto import elligator2
```
What do you think?

matejcik: It's cleaner but it is not doing the same thing .)

We need to statically exclude the branch, so that it does not end up compiled into production firmwares. `if __debug__` will do that, because the bytecode compiler can statically evaluate it to `if False` and completely omit the branch.

(we also have custom code that does search-and-replace of `if utils.BITCOIN_ONLY` and a couple other things, but `THP_VERSION` is not one of those things.)

I consider `if __debug__` to be a hack that we'll drop at some point -- but it doesn't seem worth more work into a clean solution.

onvej-sl: The goal wasn't to do the same thing. 

Either what you're suggesting doesn't make sense to me, or I don't understand the problem you are trying to solve. If you remove `if __debug__`, the branch will still be included in firmwares that don't use the new THP.

matejcik: the problem is, I want to statically exclude the import statement from compiled firmwares (for the time being)
`if THP_VERSION` is _dynamically_ excluding it, i.e., the if-condition (and the import statement and the reference to the module) becomes part of compiled bytecode.

(i mean, it's not as if this has any practical effect either way, but, well. it's cleaner)

`if __debug__` is a fancier version of `if False` that still lets us run the code when we need it -- with the expectation that once we enable THP, we enable it for all models at once.
in other words, the import will become unconditional

do you expect that we will want the import to be conditional at some point in the future? like, say, TS3 with the old protocol but T3W1 with THP?

if so, how about:
```suggestion
if __debug__:  # exclude from built firmwares until THP is ready
    if utils.THP_VERSION == 3:
        from trezorcrypto import elligator2
```
why version 3 btw?

(FWIW, the conditional import will cause problems with the typechecker, but we'll cross _that_ bridge when we come to it)

crypto/elligator2.c
```diff
@@ -0,0 +1,215 @@
...
+  curve25519_add_reduce(tv1, tv1, tv1);
+
+  // xd = tv1 + 1
+  bignum25519 xd = {0};
+  curve25519_add_reduce(xd, tv1, one);
+  memzero(one, sizeof(one));
+
+  // x1n = -J
+  bignum25519 x1n = {0};
+  curve25519_neg(x1n, j);
```
M1nd3r, line 86: x1n is a constant as long as J is a constant. I suppose we could have it pre-computed. See file ed25519-donna-impl-base.c, lines 19-20 (A=J)

crypto/elligator2.c
```diff
@@ -0,0 +1,195 @@
...
+  bignum25519 input_bignum = {0};
+  curve25519_expand(input_bignum, (unsigned char*)input);
+
+  // c3 = sqrt(-1)
+  bignum25519 c3 = {0};
+  curve25519_set_sqrtneg1(c3);
+
+  // c2 = 2^c1 = 2^((q + 3) / 8) = c3 + 1
+  bignum25519 one = {0};
+  bignum25519 c2 = {0};
```
M1nd3r, line 64: c2 is not needed, please, remove it

onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/3435/commits/2052ce286aa69dcb7edb279d10b6456bccc43205.

M1nd3r (commented):
LGTM, but someone else should check the build scripts

onvej-sl:
@matejcik Can you please review https://github.com/trezor/trezor-firmware/pull/3435/commits/e92595bff754a852b52cafb73bfedf8ab04c2e01 or assign someone else to do it?

core/src/trezor/crypto/__init__.py
onvej-sl: See https://github.com/trezor/trezor-firmware/pull/3435/commits/e92595bff754a852b52cafb73bfedf8ab04c2e01.

onvej-sl:
The squashed and rebased version of this pull request can be found here: https://github.com/trezor/trezor-firmware/pull/3818

matejcik (approved):

M1nd3r (approved):

ci/build.yml
```diff
@@ -197,6 +197,8 @@ core unix regular build:
   stage: build
   <<: *gitlab_caching
   needs: []
+  variables:
```
M1nd3r, line 200: @onvej-sl I am curious - will the thp switch be available in "frozen" build too, when it is defined here in core unix regular?

Issue #3416, opened 2023-11-24
Change go to bootloader screens text


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/931470e2-be12-4e2f-8747-c850714870f3)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/3c09b18f-f6ce-480d-abe9-f9f283c0668d)

PR #3438, opened 2023-12-05
chore(core): change `Go to bootloader` screen text

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/3416:
- change the text before going to bootloader


Hannsek (approved):

Hannsek:
Don't we have to regenerate fixtures?

grdddj:
> Don't we have to regenerate fixtures?

There was no change - meaning we do not have UI tests for this screen.

I was also surprised that all the tests passed, and was considering adding this test. Decided that it is not worth the effort

PR #3439, opened 2023-12-05
fix(core): correct action of left button in Ethereum summary

grdddj:
Fixes the following bug:
- when users would click left button in the summary, it would crash, as it was attempting to go to previous page, which was not defined

Also, new UI test with input flow was made to check it now


matejcik:
fixture for the new test is missing for TT

grdddj:
> fixture for the new test is missing for TT

Right, I am doing the TT changes now. WIll also probably structure the commits little differently, with fix, tests and fixtures in separate commits.

grdddj:
Force-pushed with adding `TT` tests and some refactors.

Some takeaways and possible improvements from this:
- we should have tests for all these "edge-cases" of going between layouts when we support it (generally all the `while True`) layout concepts
- even with this, tests would not catch this bug if we were using `debug.press_no() / press_yes()`, because that sends the signal right away and is disconnected from the real user's action of pressing the button
  - therefore, it might make sense to replace _all_  the occurrences of `press_no/yes` in input flows with `press_left/right` for `TR` and with `debug.click` for `TT`

grdddj:
Tests in `Gitlab` failed because of some glitch screen, otherwise are fine:

image

mmilata (approved):
LGTM

mmilata:
> we should have tests for all these "edge-cases" of going between layouts when we support it (generally all the while True) layout concepts

While we'll be rewriting the rest of the layout handling in rust, we might be able to generate such testcases, if we choose suitable representation. Meanwhile :+1: for manually writing testcases that go back.

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 24.1.0 (e24c80afa0c49184f185c16112b545aab9bf0271)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.1.0 Chrome/118.0.5993.129 Electron/27.0.4 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.6.4 regular (revision 42e9ed0e09033d474dee1a560fe5870646fa440e)

PR #3445, opened 2023-12-06
fix(solana): fix Memo, multisig and polish instructions UI

gabrielKerekes:
This PR fixes two important issues with Solana, one medium issue and minor issues with the way some instructions are displayed:

1.  Memo instruction is currently treated as an uknown instruction i.e. it's basically being blind signed.
2. Multisig token transfer instruction is also treated as a predefined instruction and multisig signers are not shown. The only warning the user getting that it's a multisig instruction is the `Transaction requires X signers which increases the fee.` warning.

3. The `The following instruction is a multisig instruction.` message is shown after the instruction parameters are shown but before the accounts - it should be displayed even before the parameters.

4. The order in which some properties are displayed has been changed and the wording of some of the properties has been updated.

_Note: I would have regenerated the UI fixtures but I get different results for all of them although definitely not all of them should've changed. Not sure what's going on here. This happens for both Trezor T and R, I run the T emulator with the `-a` switch to stop animation._


matejcik (commented):
code LGTM

Hannsek:
@gabrielKerekes Please post here the UI changes.

gabrielKerekes:
Hopefully this will help you - the `failed` folder contains the `html` files from the tests report.

[failed:changed ui tests.zip](https://github.com/trezor/trezor-firmware/files/13598950/failed.changed.ui.tests.zip)

matejcik:
all seems good in CI, please add changelog (since this is going out in the next release so it's a change)

matejcik (approved):

gabrielKerekes:
👋 Will there be a FW release this month or when can we expect this to be released? 🙏

PR #3454, opened 2023-12-08
ci: HTML UI reports for github actions

mmilata:
Github Actions only exposes artifacts as .zip files with a bunch of limitations, which makes them unsuitable for UI reports HTML. This PR implements copying the reports to an external server. Links to the reports are then posted to the pull request discussion. To minimize noise, comments are edited if they already exist, and are not posted at all when there are no changes.

Example comment with working links: https://github.com/trezor/trezor-firmware/pull/3436#issuecomment-1841808188

@vdovhanych can you please take over the PR and replace the `rsync` invocations with `aws s3 sync` or anything else that is needed to get the reports to data.trezor.io? Currently they are copied to my [throwaway VPS](https://trezorui.martinmilata.cz/) that will be deleted 31.12. You may also want to set up some kind of periodic deletion of old reports.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -2,6 +2,18 @@ name: Core
 
 on: [pull_request]
 
+permissions:
+  id-token: write      # for fetching the OIDC token
+  contents: read       # for actions/checkout
+  pull-requests: write # For dflook comments on PR
+
+env:
+  PULL_COMMENT: |
+    |core UI changes|[master diff](https://data.trezor.io/dev/firmware/master_diff/${{ github.run_id }}/)|    |    |
+    |---------------|------------------------------------------------------------------------------------|----|----|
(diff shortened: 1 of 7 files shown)


.github/workflows/core.yml
```diff
@@ -545,13 +570,73 @@ jobs:
...
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+      - run: |
+          git fetch origin main
+          git diff --quiet origin/main...HEAD -- tests/ui_tests/fixtures.json || echo "FIXTURES_CHANGED=$?" >> $GITHUB_OUTPUT
+        id: check-fixtures-changed
+      - uses: ./.github/actions/ui-comment
+        if: ${{ steps.check-fixtures-changed.outputs.FIXTURES_CHANGED == '1' }}
+
+  core_ui_deploy:
```
vdovhanych, line 614: This doesnt work on github yet, is this something that we want to have or is it enough what you did with the ui tests in this pr?

mmilata: Is there no deploy runner? I think this is needed for master diffs to work, though as long as we keep the GitLab pipelines running we can convert it later.

vdovhanych (approved):

Issue #3448, opened 2023-12-07
fix excessive stack usage in blurred homescreen


TychoVrahe:
The blurring data are stored on stack, and the size of this data is dependent on screen width. With larger screens this can cause stack overflows, so another method needs to be implemented.

We can do it similarly like other graphical buffers: store it statically in some designated RAM.

TychoVrahe:
closed via https://github.com/trezor/trezor-firmware/pull/3456

PR #3456, opened 2023-12-11
chore(core): optimize stack usage during blurring

cepetr:
The PR addresses the optimization of stack usage during image blurring on the homescreen. I replaced the stack-allocated buffer with a statically allocated buffer, saving 1440B on the stack for model T. This resolves issue #3448.

I am not happy with such a solution at all, but it properly utilizes the already existing buffer framework. Just look at it, and let's disccuss it.


TychoVrahe:
This is what i imagined as a fix. 

What i dislike about this and the whole buffers concept in general is that we statically allocate memory that is used for short caclulations. It is even worse with the buffers for specific, single case calculations like the blurring here. But i currently dont know a better way to do this, so ideas are welcome.

To list the limitations we deal with:
- some the buffers are too large for the stack, especially on higher resolution displays
- even with the smaller buffers, stack might be in CCMRAM unusable by DMA so we cant use it (currently case for TT bootloader, though we might as well move the stack, the reasons for it being there are not very strong)
- we dont have any dynamic allocation, we cant even use micropython heap as this needs to work in bootloader too

cepetr:
I see two potential solutions that can be combined:

1) We can increase the stack size and allow larger allocations (for buffers not requiring DMA access) on the stack. I don't like it much, but I think it's reasonable if we don't have a global heap.

2) Instead of using statically allocated buffers for a specific purpose, we can establish a designated scratch memory (let's say 40KB) with our custom allocator. This allocator would replace the use of `BufferXXX::get_cleared()`, allowing us to use `vec!` or `Box<>` allocations instead. Such a solution does not suffer from the heap fragmentation we are afraid of.

matejcik:
> 2\. Instead of using statically allocated buffers for a specific purpose, we can establish a designated scratch memory (let's say 40KB) with our custom allocator. This allocator would replace the use of `BufferXXX::get_cleared()`, allowing us to use `vec!` or `Box<>` allocations instead. Such a solution does not suffer from the heap fragmentation we are afraid of.

The issue to solve with this is when to deallocate: once you store a pointer to this custom arena on the GC heap, it can get dropped without notice. The GC allows "allocate with destructor" with some limitations, so we'd need to investigate that and properly clean up after ourselves, plus also think through the implications of the GC on Rust's memory guarantees. (i suspect we'd need to refcount everything).

One option is, if we use this scratchpad arena only for buffers (DMA or otherwise), make all the functions touching it `unsafe` and control allocation and deallocation explicitly. Other option is carefully think through the above and make it available in safe code. It would be very nice to have, but so far it seemed too difficult.

TychoVrahe (approved):
LGTM, lets merge this fix now and deal with allocation later.

Issue #286, opened 2019-06-28
Recovery: Add a return button to go back when entering words


tsusanka:
Child of #269 
There is currently no back button when entering words. If I confirm a word and then realize it was wrong, I have no way to go back.

A simple solution is to add a simple backspace button when there is no input in the keyboard, which will take the user one screen back:

![Screenshot from 2019-06-25 09-08-11](https://user-images.githubusercontent.com/1835345/60332243-b017ba80-9996-11e9-8827-1aaa5df15ecf.png)

We might need to shorten the text. @bentonoliver could make up the wording? Consider doing https://github.com/trezor/trezor-firmware/issues/285 along with that.

tsusanka:
@ZdenekSL I'm not sure if we want do this as part of Shamir 2.0 so I'm leaving this up to you (or up to the planning ceremony).

prusnak:
Let's do this as part of Shamir 2.0, please.

We can use the backspace button to discard the latest word entered and go one word back.

ZdenekSL:
With UX designer, specify the text for a screen if we add backspace.

tsusanka:
Discussed with Petr and Matej that this is quite minor issue, which will affect all other users. So we think we should not implement this.

Reasons:
- Users enter the words carefully, so mistakes should occur rarely.
- Persistence will make this easier - only the one specific share needs to be entered again.
- It adds a risk that user will return by mistake.
- It is non-destructive operation if I make a mistake. The worst case scenario you have to enter 20 (or 33) words again (3 min? 5 min?).
- Most of the users will fill in the words correctly with no need for backspace button.

Example: 
Out of 1000 users will do recovery. Out of them 10 people will make a mistake. Because of those 10 people we'd modify the UI (show backspace and shorten the text) for the remaining 990. Those numbers are though a guesstimate.

ZdenekSL:
Ok, just for clarification, if I add share, that has a wrong word in it, I can add the same share, with the correct word and it will pass. Correct? If so, please close this issues, we can reopen it if we have user feedback about the need of backspace.

tsusanka:
> Ok, just for clarification, if I add share, that has a wrong word in it, I can add the same share, with the correct word and it will pass. Correct?

If you enter a share, which is incorrect (some word wrong / missing etc.) we'll let you know and allow you to enter the share again. Does that answer your question?

ZdenekSL:
Yes, thank you.

Hannsek:
[...]

PR #3458, opened 2023-12-12
Recovery go to previous word

grdddj:
Fixes https://github.com/trezor/trezor-firmware/issues/286:
- allowing to go back to previous word in recovery process

I am creating a bunch of click tests to try all the possible scenarios


mmilata (commented):
Looking good.

Are you planning to add more tests as part of this PR?

I noticed that in model T you can go and change any previous word, but on TS3 you can only go to the last one, is that intentional?

Hannsek:
I can change any previous word on TS3.

If someone press `previous`, it goes to the screen where there is delete button and the word:
image

Instead, can it be the screen with multiple options? 
image
This is because if I go back, I have to delete at least 2 letters every time I want to change something

On TT, instead this screen
image
go to this
image

(Instead 
image
go to 
image)

In general, go to screen that has multiple options without deleting single letter.

grdddj:
> I can change any previous word on TS3.
> 
> If someone press `previous`, it goes to the screen where there is delete button and the word: image
> 
> Instead, can it be the screen with multiple options? image This is because if I go back, I have to delete at least 2 letters every time I want to change something
> 
> On TT, instead this screen image go to this image
> 
> (Instead image go to image)
> 
> In general, go to screen that has multiple options without deleting single letter.
> 
> Ideally, if this screen suggests the same word. See the word _machine_, there it is right. But instead of _bind_ you can see the suggested bicycle, that's not right.

These things are technically possible, but certainly not easily/cleanly. I would vote for simplicity --- the cost of users having to click delete a couple of times is minimal, considering how very few users will ever encounter this situation.

grdddj:
> Are you planning to add more tests as part of this PR?

Yes, I will add more tests, once we finalize the wanted requirements.

core/src/apps/management/recovery_device/layout.py
```diff
@@ -48,12 +48,38 @@ async def request_mnemonic(
 
     await button_request("mnemonic", code=ButtonRequestType.MnemonicInput)
 
+    # Allowing to go back to previous words, therefore cannot use just loop over range(word_count)
     words: list[str] = []
```
matejcik, line 52: `words = [""] * word_count` simplifies a lot of the special cases below

core/src/apps/management/recovery_device/layout.py
```diff
@@ -48,12 +48,38 @@ async def request_mnemonic(
...
-    for i in range(word_count):
+    i = 0
+    while True:
+        # All the words have been entered
+        if i >= word_count:
+            break
+
+        # Not allowing negative indexes
+        if i < 0:
```
matejcik, line 60: ...and this should be folded into `if not word` below so that we never start with a negative index in the first place

core/src/apps/management/recovery_device/layout.py
grdddj: Right, thanks for the idea. Done in [33ce5da](https://github.com/trezor/trezor-firmware/pull/3458/commits/33ce5da629fc0aeb862c95a9ebe04027ef058fa2)

core/src/apps/management/recovery_device/layout.py
grdddj: Done in [33ce5da](https://github.com/trezor/trezor-firmware/pull/3458/commits/33ce5da629fc0aeb862c95a9ebe04027ef058fa2)

grdddj:
> These things are technically possible, but certainly not easily/cleanly. I would vote for simplicity --- the cost of users having to click delete a couple of times is minimal, considering how very few users will ever encounter this situation.

@Hannsek Want to ask about the current status - are you OK with this, or do you want to do the changes above? I would estimate it at couple hours to 1 day of work.

If we are OK with the current state, I will do some more tests to freeze the functionality and then we are ready to merge.

Hannsek:
Im ok with this for now.
But please change the word `delete` to `previous` on TS3 when no letter is input, if not yet changed.

Hannsek:
@grdddj What is missing here?

grdddj:
> @grdddj What is missing here?

I think that nothing really important. The tests I wanted to add do not have a big priority

grdddj:
> So we can merge it?

I guess so, after the CI will be green and the PR is approved

mmilata (commented):
I think it's ready but some tests are failing. Also please remove the `TODO` lines and/or turn them into GH issue.

grdddj:
> I think it's ready but some tests are failing. Also please remove the `TODO` lines and/or turn them into GH issue.

Right, sorry about that. Should be rebased and green now, TODOs deleted

grdddj:
CI is green, ready to merge @mmilata

mmilata (approved):

bosomt:
Testing on dry run.
`Previous` button is available even before i enter first word /nothing happens when pressed/
Is that expected ?

**Info:**
 - Suite version: desktop 24.2.2 (6544eb4cdb6f8982f3f391307e2bff5784db2bcf)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.2.2 Chrome/118.0.5993.129 Electron/27.0.4 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.6.5 regular (revision e1f696b4dd8c024a9ceca276d7336b4d93ae7ff2)

Hannsek:
It is not. Good catch. @grdddj

grdddj:
> It is not. Good catch. @grdddj

That is OK in my perspective, if the button does nothing. Yeah, it would be clearer when not showing it at all, but at the cost of more code.

How do we want it to look/behave like?

Hannsek:
Ideally, it should not be there if no word is present. I'll create an issue.

bosomt:
**QA OK**

PR #3459, opened 2023-12-12
fix large area marquee rendering

TychoVrahe:
Fixing integer overflow and subsequent panic when rendering to large enough area.


matejcik (approved):
utACK

PR #3461, opened 2023-12-14
Add ETC common assets

chris-mercer:
+ Add ETC's Mordor Testnet
+ Add ETC's WETC for DeFi ecosystem forming now that its the largest PoW smart contract blockchain.
+ Add ETC logo


matejcik:
No reason whatsoever to have those built-in.

METC is already listed in https://github.com/trezor/definitions/blob/main/definitions-latest.json
WETC-on-ETC will be pulled in when we next update the definitions ... speaking of, i don't understand why you need wetc on the etc network. wetc on ethereum is already supported

chris-mercer:
> No reason whatsoever to have those built-in.
> 
> METC is already listed in https://github.com/trezor/definitions/blob/main/definitions-latest.json WETC-on-ETC will be pulled in when we next update the definitions ... speaking of, i don't understand why you need wetc on the etc network. wetc on ethereum is already supported

Apologies. Fixed Upstream in ethereum-lists

PR #3464, opened 2023-12-15
implement full certificate verification in trezorctl

### LINKED ISSUE #3364: Improve device authentication via trezorctl
Currently, `trezorctl device authenticate` will just spit out some hashes and raw certificates at you.

Ideally, it should also validate that everything goes well together with everything else. We even have code to do it: https://github.com/trezor/trezor-firmware/commit/1db2a424b8918289b921b8c31aa4c4c165b76896
the problem that needs to be solved here is what to do about the `cryptography` dependency, which is rather heavy for just this one feature. Maybe we can detect it and ask the user to install manually if missing?

as a start, we might want to implement this as a separate script in `tools/` directory, so that power users can run it.

Another issue we should resolve is synchronisation with Suite side whitelist/blacklist for known public keys. With the naive implementation, if we revoke a signing key, the device will still pass authentication with Trezor.

matejcik:
fixes #3364

### DIFF
--- a/python/src/trezorlib/cli/device.py
+++ b/python/src/trezorlib/cli/device.py
@@ -14,9 +14,12 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
+import io
+import logging
 import secrets
 import sys
-from typing import TYPE_CHECKING, Optional, Sequence
+from logging import DEBUG, ERROR, INFO, WARNING
+from typing import TYPE_CHECKING, BinaryIO, Optional, Sequence, Tuple
 
 import click
 
@@ -44,6 +47,8 @@ SD_PROTECT_OPERATIONS = {
     "off": messages.SdProtectOperationType.DISABLE,
     "refresh": messages.SdProtectOperationType.REFRESH,
 }
+
+LOG = logging.getLogger(__name__)
 
 
 @click.group(name="device")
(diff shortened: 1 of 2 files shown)


python/src/trezorlib/cli/device.py
```diff
@@ -334,17 +339,278 @@ def set_busy(
...
+            continue
+
+        if skip_whitelist:
+            print_step(INFO, "Skipping public key whitelist check.")
+        else:
+            if cert.public_key() not in TS3_CA_WHITELIST:
+                print_step(WARNING, f"CA certificate not in whitelist: {cert}")
+
+        try:
+            first_cert.verify_issued_by(cert)
```
andrewkozlik, line 583: Naming is slightly confusing. How about `cert` -> `ca_cert` and `first_cert` -> `cert`?

python/src/trezorlib/cli/device.py
```diff
@@ -334,17 +339,278 @@ def set_busy(
...
+        except exceptions.InvalidSignature:
+            print_step(ERROR, f"Certificate verification failed: {cert}")
+            continue
+        else:
+            print_step(DEBUG, f"Certificate verified successfully: {cert}")
```
andrewkozlik, line 588: In both cases above, semantically it would make more sense to me to see the certificate subject name that was verified successfully or failed verification, not the issuer name. In other words who was verified successfully, not by whom it was verified.

python/src/trezorlib/cli/device.py
```diff
@@ -334,17 +339,278 @@ def set_busy(
...
+    bytes.fromhex(x)
+    for x in (
+        "04b12efa295ad825a534b7c0bf276e93ad116434426763fa87bfa8a2f12e726906dcf566813f62eba8f8795f94dba0391c53682809cbbd7e4ba01d960b4f1c68f1",
+        "04cb87d4c5d0fd5854e829f4c1b666e49a86c25c88a904c0feb66f1338faed0d7760010d7ea1a6474cbcfe1143bd4b5397a4e8b7fe86899113caecf42a984b0c0f",
+        "0450c45878b2c6403a5a16e97a8957dc3ea36919bce9321b357f6e7ebe6257ee54102a2c2fa5cefed1dabc498fc76dc0bcf3c3a8a415eac7cc32e7c18185f25b0d",
+        "0454d310d88d55d3044d80fcdbce9a63bf3118545fae71f6eca303272dcc4d25cf775ae3c18ae9f41b2cf29377bc4696fc79c8824a6fd6b9ca5fb6805ed6557aab",
+        "04e94bf05586a8e7a3e9aba32662a439be5f378da372219c8ee7cf8b4684dbfbd7ba88ed920c06f9f26deab9077654647738df8cf70898fea1c3aaf2ef086fc578",
+        "048c6c104bd7cc59cd5c5717533786a72ab59685bd13937f5542820e90f6ac6945e520e19d1d627a8e81ef5a94ef87de7a6a0d778e7dc9d389db877a5f9b629dd8",
+    )
+]
```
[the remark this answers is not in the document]
matejcik: I'm hoping that this will get replaced by a whitelist URL (the docstring hints at it too) before the next version of trezorlib is released :)

matejcik:
superseded by #3944

Issue #3429, opened 2023-11-30
Bootloader: calculate image hashes while waiting for touch to come up


matejcik:
Hashing the firmware image takes considerable time, 450 ms per pass on the STM32F4.
Touch controller has a 400 ms startup time.

We can initialize the touch controller, start hashing, and when we come back to touch, it's already booted up.
(if the hashing is faster on the U5, we'll just be done sooner and wait only up to the 400ms total elapsed time mark)

PR #3466, opened 2023-12-18
Optimize device boot time

TychoVrahe:
This PR optimizes boot times with few improvements:
- on T3T1/STM32U5, hash processor is used to calculate image hash, the hash algorithm is switched to SHA256
- checking of empty area after fw image is optimized by comparing words rather than bytes
- I2C initialization on U5 is sped up by using static value for timing register
- touch is initialized at the start of bootloader, so that the waiting time is utilized to check fw image. (see https://github.com/trezor/trezor-firmware/issues/3429 )
- unnecessary delays are removed from touch driver initialization
- the period for checking touch for bootloader entry is shortened from 200ms to 10ms, which is sufficient as long as the user is swiping before this check, which i think is the point anyway. It might change the UX of entering the bootloader a bit though. Users should not need to enter bootloader that often now that we have interaction less fw update, but for TT we might want to keep the original delay not to mess with users muscle memory


core/embed/bootloader/main.c
```diff
@@ -541,15 +549,16 @@ int bootloader_main(void) {
...
+    for (int i = 0; i < 10; i++) {
       touched = touch_is_detected() | touch_read();
       if (touched) {
         break;
       }
 #ifdef TREZOR_EMULATOR
       hal_delay(25);
 #else
-      hal_delay(1);
+      hal_delay_us(1000);
```
matejcik, line 561: why?

core/embed/lib/image_hash_conf.h
```diff
@@ -0,0 +1,50 @@
...
+}
+static inline void _blake2s_Final(BLAKE2S_CTX *ctx, void *out) {
+  blake2s_Final(ctx, out, BLAKE2S_DIGEST_LENGTH);
+}
+
+#define IMAGE_HASH_CTX BLAKE2S_CTX
+#define IMAGE_HASH_INIT _blake2s_Init
+#define IMAGE_HASH_UPDATE blake2s_Update
+#define IMAGE_HASH_FINAL _blake2s_Final
+#define IMAGE_HASH_CALC _blake2s_Calc
```
matejcik, line 44: this is kind of weird style, i would expect something like
```suggestion
#define IMAGE_HASH_CALC(data, len, hash) blake2s(data, len, hash, BLAKE2S_DIGEST_LENGTH)
```
etc.
esp. given that `blake2s_Update` is used raw

is your version better somehow? or does it not matter?

core/embed/bootloader/main.c
TychoVrahe: The hal_delay function does delay _at least_ that amount of milliseconds. In reality, it waits for two n+1 systicks, as the time needed to reach the first systick is unknown. But when done in loop, the first systick (on second and next pass throughs) will be almost the whole millisecond, therefore the total delay will be n+1 ms.  Resulting here in double of the intended delay.

core/embed/lib/image.c
```diff
@@ -279,23 +281,42 @@ secbool check_image_contents(const image_header *const hdr, uint32_t firstskip,
...
+          return secfalse;
+        }
+      }
+
+      offset += single_bytes_to_check;
 
-      const uint8_t *data =
-          (const uint8_t *)flash_area_get_address(area, offset, bytes_to_check);
+      const uint32_t *data = (const uint32_t *)flash_area_get_address(
+          area, offset, words_to_check);
```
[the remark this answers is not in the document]
TychoVrahe: right https://github.com/trezor/trezor-firmware/pull/3466/commits/bce6d50eb1d0041560b8e5352d4e8f890f51b9bb

core/embed/lib/image_hash_conf.h
TychoVrahe: my version just sucks https://github.com/trezor/trezor-firmware/pull/3466/commits/3b9962c3baa42201ff7b08c1abcc7eb86af01ac7

matejcik (commented):
not sure what the vendor header changes are doing here

otherwise ACK code-wise but not approving yet, I would like more people to play with the shortened bootloader delay

also @cepetr please review too

TychoVrahe:
The vendor headers are regenerated for U5 models because we now use SHA256 to calculate their hash (which is then signed). Unsafe prod signed vendor headers are removed because they are no longer valid after the algo change.

core/embed/bootloader/main.c
```diff
@@ -404,6 +408,7 @@ int bootloader_main(void) {
   random_delays_init();
 
 #ifdef STM32U5
+
```
cepetr, line 415: Unnecessary nested #ifdef STM32U5

core/embed/lib/image.c
```diff
@@ -19,13 +19,13 @@
...
 #include <string.h>
 
-#include "blake2s.h"
 #include "ed25519-donna/ed25519.h"
 
 #include "common.h"
 #include "flash.h"
 #include "image.h"
 #include "model.h"
+#include TREZOR_BOARD
```
cepetr, line 28: This explicit #include is probably unnecessary (already included in image.h)

core/embed/trezorhal/stm32u5/hash_processor.c
```diff
@@ -0,0 +1,127 @@
...
+
+    __HAL_HASH_SET_MDMAT();
+
+    HAL_HASHEx_SHA256_Start_DMA(&hhash, (uint8_t *)data, chunk);
+    while (HAL_HASH_GetState(&hhash) != HAL_HASH_STATE_READY)
+      ;
+
+    if (last) {
+      HASH->STR |= HASH_STR_DCAL;
+      HAL_HASHEx_SHA256_Finish(&hhash, hash, 1000);
```
cepetr, line 61: HAL_HASHex_SHA2256_Finish() expects the output buffer to be aligned to 32 bits, but it's not ensured here.

core/embed/trezorhal/stm32u5/hash_processor.c
```diff
@@ -0,0 +1,127 @@
...
+
+static void hash_processor_sha256_calc_dma(const uint8_t *data, uint32_t len,
+                                           uint8_t *hash) {
+  while (len > 0) {
+    uint32_t chunk = len > 0x8000 ? 0x8000 : len;
+    bool last = (len - chunk) <= 0;
+
+    __HAL_HASH_SET_MDMAT();
+
+    HAL_HASHEx_SHA256_Start_DMA(&hhash, (uint8_t *)data, chunk);
```
cepetr, line 55: Not sure, but HAL_HASHex_SHA256_Start_DMA() expects the input buffer to be aligned to 32 bits, but it's not ensured here.

core/embed/bootloader/main.c
```diff
@@ -550,15 +549,16 @@ int bootloader_main(void) {
...
+    for (int i = 0; i < 10; i++) {
       touched = touch_is_detected() | touch_read();
       if (touched) {
         break;
       }
 #ifdef TREZOR_EMULATOR
       hal_delay(25);
 #else
-      hal_delay(1);
+      hal_delay_us(1000);
```
cepetr, line 561: Isn't it too long? 10 * 1000ms?

core/embed/bootloader/main.c
TychoVrahe: yes. i will be moving the sec area configuration to boardloader anyway in subsequent PR, so this code will be somewhat simplified. I would keep this as is until then as this was not modified here (apart for adding a blank line which is removed in this fixup: https://github.com/trezor/trezor-firmware/commit/e419986e6222fcf7b01fd59b68cd6d8f7e82465b )

core/embed/lib/image.c
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/da2b3a551e37a2b097db42a71251cbe41fb3b93d

core/embed/trezorhal/stm32u5/hash_processor.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/a7afd0f8600c62cc972c63ca7f2c0e7473f1fe12

core/embed/trezorhal/stm32u5/hash_processor.c
TychoVrahe: its ensured in caller `hash_processor_sha256_calc`, since this is static function i think its sufficient

core/embed/bootloader/main.c
TychoVrahe: this is delay in microseconds

TychoVrahe:
added a fix with dma interrupt priority setting, and dma channel security configuration (now allowed in unprivileged mode)

https://github.com/trezor/trezor-firmware/pull/3466/commits/c2beaa456311dfbc60af48fffacec88022728ba8

core/embed/trezorhal/stm32u5/hash_processor.c
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+void hash_processor_sha256_calc(const uint8_t *data, uint32_t len,
+                                uint8_t *hash) {
+  if (((uint32_t)data & 0x3) == 0) {
+    hash_processor_sha256_calc_dma(data, len, hash);
+  } else {
+    HAL_HASHEx_SHA256_Start(&hhash, (uint8_t *)data, len, hash, 1000);
+  }
```
cepetr, line 76: HAL_HASHex_SHA256_Start() input data must be aligned to 32 bits as well, see implementation in stm32f4xx_hal_hash_ex.c. The STM32 API is very bad in this...

core/embed/trezorhal/stm32u5/hash_processor.c
cepetr: You are right,but please see the new comment below. I think the problem with alignment goes much deeper.

cepetr: I close this issue: since cortex-m allows unaligned accesses, this is not a problem at all.

cepetr (approved):

PR #3474, opened 2023-12-29
T3T1: use internal framebuffer 

TychoVrahe:
Implements framebuffer in internal RAM for T3T1, using ST7789V display driver.

Two framebuffers are used so that one can be used for drawing while the other is being transferred to display with DMA.

For maximum efficiency interrupts are used for starting the DMA transfers after tearing effect line signal. 

Note: while this PR allows seamless switching of layouts without fading out-in, this is not changed in this PR and is considered out of scope here.

Builds on previous U5/T3T1 pull requests: #3370 #3422 #3466 

This also fixes a bug in signing (bitcoin-like) tx flow, where after confirming the TX with hold to confirm, "loading transaction" loader is briefly shown and then fade-in/fade-out into "signing transaction" loader. This is specific to TT and HW only, because of originally missing `display_refresh` call in the progress init (TT does not need it to show the loader on display, while emulator and TS3 do)

The ST7789V driver is getting quite complex with all the variations, maybe it would be better to split the implementation to F4 without framebuffer and U5 with framebuffer, the U5 variant wouldn't need to carry all the model T compatibility ifdefs so it would be somewhat cleaner


cepetr (approved):
It looks ok and functional. But as we discussed, this will be refactored as soon as we integrate the new drawing library.

PR #3480, opened 2024-01-04
add makefile commands for erasing bootloader and firmware

TychoVrahe:
Since we are checking the empty sectors of the bootloader/firmware are in flash, flashing the firmware/bootloader with debugger is no longer very reliable: sometimes it happens that new build produces binary occupying less sectors than previous one and then flashing leaves the additional sector unerased and integrity check fails with corruped firmware message. 

This PR adds a command that deletes all the bootloader/firmware sectors, so that developer can manually fix that in case it happens. We could make the erasure automatic, but that increases the time to flash every time.

To make this work for all models, unfortunately some constants definitions are further duplicated between makefile, layouts and linker scripts. We could maybe avoid this by generating the files from some common truth source, but i am leaving that for future.

The makefile commands for erasing and reading storage are also fixed in this PR


prusnak:
Typo: 6x 'BOARDLODAER' in the Makefile

core/Makefile
```diff
@@ -314,11 +356,22 @@ flash_combine: $(PRODTEST_BUILD_DIR)/combined.bin ## flash combined using OpenOC
 flash_erase: ## erase all sectors in flash bank 0
 	$(OPENOCD) -c "init; reset halt; flash info 0; flash erase_sector 0 0 last; flash erase_check 0; exit"
 
+flash_erase_bootloader: ## erase bootlaoder
```
hiviah, line 359: Small typo in comment - `bootlaoder` instead of `bootloader`

core/Makefile
```diff
@@ -314,11 +356,22 @@ flash_combine: $(PRODTEST_BUILD_DIR)/combined.bin ## flash combined using OpenOC
...
+	$(OPENOCD) -c "init; reset halt; flash info 0; flash erase_sector 0 $(BOOTLOADER_SECTOR_START) $(BOOTLOADER_SECTOR_END); exit"
+
+flash_erase_firmware: ## erase bootlaoder
```
hiviah, line 362: Small typo in comment - `bootlaoder` instead of `bootloader`

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/3480/commits/a21292edcb44ba5daa116be4c109830febf1a144

core/Makefile
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/3480/commits/a21292edcb44ba5daa116be4c109830febf1a144

hiviah (approved):

PR #3483, opened 2024-01-08
POC: add user adjustable brightness setting

TychoVrahe:
This PR adds user adjustable brightness setting via trezorctl (`trezorctl set brightness`)

The backlight driver is fixed: when going from zero backlight there needs to be a longer pulse to wake up the IC, with user adjustable this wasn't guaranteed to happen.

POC state:
- needs proper UI design
- backlight for some screens are defined in rust, which currently does not respect the user setting. Needs to be dealt with somehow.
- This brightness is selected with slider on device. Created a new component for that, `NumberInputSliderDialog`, the current `NumberInputDialog` seems a bit too specific with the INFO&CONTINUE buttons which i don't need here, but with some effort these to might be probably merged


Hannsek:
We can probably close this, no?

TychoVrahe:
closing in favor of https://github.com/trezor/trezor-firmware/pull/3791

PR #3484, opened 2024-01-09
fix rust warnings and clippy errors

TychoVrahe:
just to have clean slate. 

Not entirely sure about the `#![allow(internal_features)]` which is needed to suppress warning about `#![feature(lang_items)]`


matejcik (approved):
utACK

matejcik:
> Not entirely sure about the `#![allow(internal_features)]`

i think it's correct to have it there

PR #3487, opened 2024-01-10
ci: update table with links to html reports

mmilata:


vdovhanych (approved):
Looks good, some of the links don't work but that is because that job failed or why?

PR #3441, opened 2023-12-06
SD card backup

obrusvit:
This is a draft PR for SD card backup and recovery feature. Test coverage limited for now.

PR #3490, opened 2024-01-12
allow SD card hotswap based on production date for Model T

TychoVrahe:
Allows SD card hotswap based on production date for Model T, the 2018 and earlier dont have the ability.

Pointing this to T3T1/U5 branch because there is some flash_otp refactoring done and want to avoid conflicts.

### DIFF
--- a/core/embed/lib/unit_variant.c
+++ b/core/embed/lib/unit_variant.c
@@ -1,10 +1,15 @@
-#include "unit_variant.h"
+#include <string.h>
+
 #include "flash_otp.h"
 #include "model.h"
+#include "unit_variant.h"
+#include TREZOR_BOARD
 
 static uint8_t unit_variant_color = 0;
 static bool unit_variant_btconly = false;
 static bool unit_variant_ok = false;
+
+static int16_t unit_variant_build_year = -1;
 
 static void unit_variant_0x01(const uint8_t *data) {
   unit_variant_color = data[1];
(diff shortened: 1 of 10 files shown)


core/embed/lib/unit_variant.c
```diff
@@ -34,3 +65,20 @@ uint8_t unit_variant_get_color(void) { return unit_variant_color; }
 bool unit_variant_get_btconly(void) { return unit_variant_btconly; }
 
 bool unit_variant_present(void) { return unit_variant_ok; }
+
+bool unit_variant_is_sd_hotswap_enabled(void) {
```
obrusvit, line 69: Just a comment: unix OTP (mocked for emu) is currently programmed to contain all `1`, resulting in `unit_variant_get_build_year()` returning `-1` (no dash found)  for emulator, i.e. emulator won't be "hot-swappable". 

I will probably mock different OTP for emu in #3441 to simulate both cases.

core/embed/lib/unit_variant.c
```diff
@@ -1,23 +1,54 @@
...
+  secbool result =
+      flash_otp_read(FLASH_OTP_BLOCK_BATCH, 0, data, FLASH_OTP_BLOCK_SIZE);
+
+  if (sectrue != result || data[0] == 0xFF) {
+    return -1;
+  }
+
+  size_t len = strnlen((char *)data, FLASH_OTP_BLOCK_SIZE);
+
+  for (int i = 0; i < len; i++) {
```
obrusvit, line 32: I didn't find any info about how is OTP organized except this: https://docs.trezor.io/trezor-firmware/core/misc/memory.html?highlight=otp#otp 
but it's not apparent what exact data is there byte-wise.

Can we add a comment here explaining what are we parsing? Something like: `block 0 contains 'somedata-YYMMDD'`. Thanks.

core/embed/lib/unit_variant.c
hiviah: Dev boards have OTP also empty, but seem to be hot swappable in HW.
Though not sure if QA still uses some of the dev boards still, which would appear for them not swappable. I don't know how important would this detail be in QA.

core/embed/lib/unit_variant.c
prusnak: Yes, we use `{PREFIX}-YYMMDD` where `{PREFIX}` is defined here:

https://github.com/trezor/trezor-firmware/blob/633107644487480286a7b5fe9e1dea36a1efc9e2/core/embed/prodtest/main.c#L52-L56

the control read is happening here (maybe we should update the comment on line 579?):

https://github.com/trezor/trezor-firmware/blob/633107644487480286a7b5fe9e1dea36a1efc9e2/core/embed/prodtest/main.c#L578-L585

core/embed/lib/unit_variant.c
TychoVrahe: regarding the dev boards: when the OTP is empty, the idea is to not allow hot swapping as we can't say anything about the production date. we could program some of the QA boards to allow this if needed.

core/embed/lib/unit_variant.c
TychoVrahe: updated the comment and readme: https://github.com/trezor/trezor-firmware/commit/43768b56255f798baa49a026e6262a0864a54bfd

TychoVrahe: and a comment in unit_variant: https://github.com/trezor/trezor-firmware/commit/f1af347bfe5a4e9b2bbbdf77a6f88a6590934c49

matejcik (approved):
LGTM at-a-glance

obrusvit (approved):
LGTM

hiviah (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 24.7.0 (17f727ae2eb1e934aff84f8d8e5d1859f8c50efd)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.7.0 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.8.0 regular (revision 3a039c505f48b97542e6118d9fb2dc867306770a)

PR #3493, opened 2024-01-15
Flash with JLink/JTrace will keep FW integrity, but not unnecessarily erase unchanged sectors

hiviah:
Flashing FW via JLink/JTrace on STM32F42x models will not cause "firmware integrity failed" error and sectors that were unchanged won't need erase, so each FW flash will have proper padding.


TychoVrahe (approved):
utACK

PR #3495, opened 2024-01-17
add haptic to prodtest

TychoVrahe:
This PR adds ability to test haptics in prodtest.

Simple `HAPTIC duration_ms` command is added, with a limit to 6500ms.

### DIFF
--- a/core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
+++ b/core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
@@ -1,6 +1,8 @@
 #include "drv2625_lib.h"
 #include "haptic.h"
 
+#include <stdbool.h>
+
 #include STM32_HAL_H
 
 #include "i2c.h"
@@ -76,10 +78,12 @@
 #define PRESS_EFFECT_AMPLITUDE 25
 #define PRESS_EFFECT_DURATION 10
 
-static void set_reg(uint8_t addr, uint8_t value) {
+#define PRODTEST_EFFECT_AMPLITUDE 127
+
+static bool set_reg(uint8_t addr, uint8_t value) {
   uint8_t data[] = {addr, value};
-  i2c_transmit(DRV2625_I2C_INSTANCE, DRV2625_I2C_ADDRESS, data, sizeof(data),
-               1);
+  return i2c_transmit(DRV2625_I2C_INSTANCE, DRV2625_I2C_ADDRESS, data,
+                      sizeof(data), 1) == HAL_OK;
 }
 
 void haptic_calibrate(void) {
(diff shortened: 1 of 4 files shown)


prusnak:
Won't we need sub-millisecond precision? Multiplying the value by 10 seems like a waste of precision.

core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
```diff
@@ -136,19 +140,25 @@ void haptic_init(void) {
   TIM16->BDTR |= TIM_BDTR_MOE;
 }
 
-static void haptic_play_RTP(int8_t amplitude, uint16_t duration_ms) {
-  set_reg(DRV2625_REG_MODE,
-          DRV2625_REG_MODE_RTP | DRV2625_REG_MODE_TRGFUNC_ENABLE);
-  set_reg(DRV2625_REG_RTP, (uint8_t)amplitude);
+static bool haptic_play_RTP(int8_t amplitude, uint16_t duration_ms) {
+  bool success = true;
+  success &= set_reg(DRV2625_REG_MODE,
```
cepetr, line 145: As `set_reg()` returns a bool, we should avoid using the `&` operator on it (nor `&=`). I understand that it may work, but it's not considered good practice. Furthermore, if the first of `set_reg()` call fails, the routine continues and calls set_reg() again and it finally starts the timer...

In this case, I would recommend:

```cpp
if (!set_reg(...)) {
    return false;
}

if (!set_reg(...)) {
    return false;
}
```

or

```cpp
success = set_reg(...);
success = success && set_reg(...);
if (!success) {
    return false;
}
```

cepetr:
I'm not sure if anybody reads the prodtest/README.md, but it would be better to update it.

cepetr (changes requested):

core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/3495/commits/92abdb587ff27255e27fac598b4b7ddc3dad16d0

TychoVrahe:
updated the readme: https://github.com/trezor/trezor-firmware/pull/3495/commits/8507b7d2209bd1c93f214190642ef172f0a9c003

TychoVrahe:
regarding the precision: i don't see the need for sub-ms precision in haptics, we will likely use 10s of ms and longer times.

i think having interface in ms is better since its very natural to work with. (and due to the timer prescaler limitation we cant set 1ms ticks there)

cepetr (approved):

PR #3500, opened 2024-01-24
core: handle uncaught exceptions from main.py

mmilata:
Currently any exception that is not caught by an `except` block results in the hardware freezing with black screen (tested on Model T). We do define an unhandled exception callback but it doesn't seem to be used as the top-level exceptions are caught by [micropython C code](https://github.com/micropython/micropython/blob/v1.19.1/shared/runtime/pyexec.c#L129) which just does some cleanup and printing.

This PR adds a call to `error_shutdown` after returning from micropython. Other possibility would be to define `MICROPY_BOARD_AFTER_PYTHON_EXEC` micorpython hook.

### DIFF
--- a/core/embed/firmware/main.c
+++ b/core/embed/firmware/main.c
@@ -219,6 +219,9 @@ int main(void) {
   printf("CORE: Main script finished, cleaning up\n");
   mp_deinit();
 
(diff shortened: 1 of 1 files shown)


matejcik (approved):
utACK

Issue #3506, opened 2024-01-30
SDL event queue overhelming


cepetr:
If `touch_read()` function isn't called frequently enough (<10ms), the SDL event queue can
be became overhelmed by `SDL_MOUSEMOTION` events. This can result in delayed UI reactions
and unexpected behaviour.

We didn't observe this behaviour until I began experimenting with animations on the display (when call to `touch_read()` are less frequent). 

A simple solution is to merge succesive `SDL_MOUSEMOTION` events into one when the `touch_read()` function is called.

PR #3508, opened 2024-01-30
fix(core): fix mouse event processing in the simulator

cepetr:
This pull request contains a fix for #3506  issue.

The `touch_read()` function has been refactored to merge successive `SDL_MOTION`` events into one.

The behavior of the function was made more clearer and more precise:

1) The `TOUCH_START` event is generated only if `SDL_MOUSEDOWN` event
   was detected inside the simulator display area
2) When holding the mouse button and moving, `TOUCH_MOVE` is generated
3) When holding the mouse button down and leaving the display area, the `TOUCH_END` event is simulated.
   The x,y coordinates are same as in the last `TOUCH_MOVE` event

### DIFF
--- a/core/embed/trezorhal/unix/touch/touch.c
+++ b/core/embed/trezorhal/unix/touch/touch.c
@@ -32,54 +32,88 @@ extern int sdl_display_res_x, sdl_display_res_y;
 extern int sdl_touch_offset_x, sdl_touch_offset_y;
 
 static bool _touch_detected = false;
+static int _touch_x = 0;
+static int _touch_y = 0;
+
+static bool is_inside_display(int x, int y) {
+  return (x >= sdl_touch_offset_x && y >= sdl_touch_offset_y &&
+          x - sdl_touch_offset_x < sdl_display_res_x &&
+          y - sdl_touch_offset_y < sdl_display_res_y);
+}
 
 uint32_t touch_read(void) {
   emulator_poll_events();
   SDL_Event event;
   SDL_PumpEvents();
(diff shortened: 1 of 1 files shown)


core/embed/trezorhal/unix/touch/touch.c
```diff
@@ -32,54 +32,82 @@ extern int sdl_display_res_x, sdl_display_res_y;
...
+            ev_x = event.button.x - sdl_touch_offset_x;
+            ev_y = event.button.y - sdl_touch_offset_y;
+            ev_type = TOUCH_MOVE;
+          } else {
+            // use last valid coordinates and simulate TOUCH_END
+            ev_x = _touch_x;
+            ev_y = _touch_y;
+            ev_type = TOUCH_END;
+          }
         }
```
TychoVrahe, line 90: since we opted to emulate HW display behavior when leaving the display area with motion, we might also want to emulate entering the touch area from outside - on device this will generate START event, but unix will currently ignore this.

something like:
```
else {
  if (is_inside_display(event.motion.x, event.motion.y)) {
        ev_x = event.button.x - sdl_touch_offset_x;
        ev_y = event.button.y - sdl_touch_offset_y;
        ev_type = TOUCH_START;
      }
}
```

cepetr: actually very good idea, see a2dfd6e92b8cfbc811365503127a4d72cd7afb86

TychoVrahe (approved):

matejcik (approved):
please squash fixups and merge

PR #3509, opened 2024-01-31
feat(cardano): Message signing

jaskp:
Adds support for arbitrary message signing as per CIP-8.

https://cips.cardano.org/cip/CIP-0008


core/src/apps/cardano/layout.py
```diff
@@ -276,32 +281,92 @@ async def confirm_inline_datum(first_chunk: bytes, inline_datum_size: int) -> No
...
+    props: list[PropertyType] = [
         (
             f"{title} ({data_size} {bytes_optional_plural}):",
-            displayed_bytes,
+            decoder(displayed_bytes) if decoder else displayed_bytes,
         )
     ]
-    if data_size > MAX_DISPLAYED_SIZE:
+    if data_size > max_displayed_size:
         props.append(("...", None))
```
ibz, line 358: Not sure I understand what happens here, but should the `"..."` be another prop rather than be appended to the value of the truncated chunk?

PS: I see this was here before, so it is probably fine... or could/should it be improved?

jaskp:
Made some changes after feedback from software wallet devs

jaskp:
Heads-up: Rebasing again to fix conflict

matejcik:
you are adding new texts, so you will now need to convert to translations and update.

1. rebase on current main
2. find the newly added strings, convert them to `TR.cardano__...` symbols
3. add english texts for those symbols to `en.json`
4. run `python core/translations/order.py`
5. rebuild the firmware
6. regenerate fixtures
7. when that is ready, force-push here and add `translations` label
8. wait for CI results

davidmisiak:
I've rebased on top of the current `main`. (neither `update_fixtures.py ci` nor `update_fixtures.py ci --github` does anything at this moment)

davidmisiak:
Hi @matejcik, could we finish also this please?

davidmisiak:
@matejcik any update please?

gitmachtl:
We need this feature - Please merge ASAP! Thanks!

ReagulusPH:
We need this ASAP. Please merge.

thiago-tsnnst:
Hey guys, Trezor user here on Cardano, pls make this a priority!

disassembler:
We need this on Cardano. Every other hardware wallet has added this feature. It's a shame I can't use it on trezor. It's making me consider changing my hardware wallet, as much of a pain as that would be because trezor didn't follow the standard conversion of seed phrase to root key. Please trezor, add this so I don't have to go through this pain of migration.

CryptoBlocks-pro:
Yes, please add this feature so that the Cardano ecosystem can use your products. As mentioned, this will bring you to feature parity with your competitors in this regard. Thank you!

RazzorLabs:
Trezor, please consider adding this feature to enable the Cardano ecosystem to utilize your products.

Quantumplation:
Hello; I'm the CTO of Sundae Labs, a very active software development firm on Cardano. We built the second biggest DEX, a governance platform used by multiple projects, etc.

We have a strong demand within our community for this feature; It is used to sign governance votes for Sundae, Optim, Newm, and NMKR governance proposals, and without this feature, users have to either migrate their voting power into a hot wallet, or switch to using Ledger, which has recently added support for this.

Additionally, we have new DEX features in the pipeline focused on more "intent" based signatures that would be unavailable to Trezor wallet users without this support.

rcmorano:
Please get this merged! Many dApps rely on CIP8 and without this you are essentially making users runaway from your products.

pmcorcoran:
Thanks hopefully trezor integrates this soon

intertreeJK:
Please work to resolve this unfinished feature update by allowing the merging of this new code.

jterrier84:
I'm a Cardano user and Trezor customer. I need this feature on my Trezor T. Please merge this proposal!

FilipBl4gojevic:
Please merge, urgent

InputEndorsers:
It would be nice to see some progress with this integration of message signing into Trezor. I mean, do we users, have to nudge the resolution of this issue directly? Or will it move along? Other wallet providers have already integrated it. Thanks!

thenic95:
Please merge this very important feature. Thank you!

nephix:
⏱️🙄 when merged?

prusnak:
Spamming GitHub is exactly what you do not want to do as a community. It adds lots of noise into all developers' inboxes.

Locking discussion to collaborators-only.

core/src/apps/cardano/helpers/chunks.py
```diff
@@ -0,0 +1,79 @@
...
+    Generic = (object,)
+    C = Chunk = 0
+
+MAX_CHUNK_SIZE = const(1024)
+
+
+def _get_chunks_count(size: int) -> int:
+    """Integer-only version of `ceil(size / MAX_CHUNK_SIZE)`."""
+    assert size >= 0
+    return 0 if size == 0 else (size - 1) // MAX_CHUNK_SIZE + 1
```
matejcik:
```suggestion
    return (size + MAX_CHUNK_SIZE - 1) // MAX_CHUNK_SIZE
```

python/src/trezorlib/cardano.py
```diff
@@ -1019,3 +1037,47 @@ def sign_tx(
...
+    for chunk in chunks:
+        chunk_response = client.call(chunk)
+        if not isinstance(chunk_response, messages.CardanoMessageItemAck):
+            raise UNEXPECTED_RESPONSE_ERROR
+
+    final_response = client.call(messages.CardanoMessageItemHostAck())
+
+    if not isinstance(final_response, messages.CardanoSignMessageFinished):
+        raise UNEXPECTED_RESPONSE_ERROR
+
```
matejcik: code here will need to reflect `client` -> `session` change, and take advantage of `session.call(expect=...)` argument instead of raising "unexpected response" explicitly

common/protob/messages-cardano.proto
```diff
@@ -523,3 +523,49 @@ message CardanoTxBodyHash {
...
+ * @next CardanoMessageItemAck
+ */
+message CardanoMessagePayloadChunk {
+    required bytes data = 1;  // expected maximum chunk size is 1024 bytes
+}
+
+/**
+ * @next CardanoSignMessageFinished
+ */
+message CardanoMessageItemHostAck {
```
matejcik: this seems useless -- you may as well respond with `CardanoMessageSignature` directly

common/protob/messages-cardano.proto
```diff
@@ -523,3 +523,49 @@ message CardanoTxBodyHash {
...
+}
+
+/**
+ * Response: Contains signature for message and address used in signed headers
+ * @end
+ */
+message CardanoSignMessageFinished {
```
matejcik:
```suggestion
message CardanoMessageSignature {
```
(see also bitcoin `MessageSignature`, `EthereumMessageSignature`)

matejcik (commented):
Some smaller comments in the review.

The main thing is, we want to disable the codepath that doesn't show the whole message.
this one:
https://github.com/trezor/trezor-firmware/pull/3509/files#diff-ff436e870f332b283aae9a85591a15b35cbbba59bb134c7b515502cd99783d9cR50-R51

several ways to go about it:
1. Implement arbitrary-size message signing, by showing the chunks for confirmation one by one, and generate a streamed CBOR representation like Cardano signing does.
    Probably most difficult. Not required from our side.
2. Keep implementation as is, but disallow messages larger than one chunk. Possible pathway to doing (1) in the future?
3. If we go with (2), we can remove chunk size limit ... and put the payload directly in the first message, avoiding the chunking flow altogether. This simplifies protobuf to exactly two messages `CardanoSignMessage` -> `CardanoMessageSignature`.

Either way, the flag `hash_payload` could be dropped from the initial message -- or kept, if for some reason the host wants us to produce a hashed signature? But I don't think CIP-8 ever _requires_ that?

...but a possible memory utilization optimization is moving the flag to the _response_; that is, Trezor will decide whether to hash the payload before signing, and tell the host. So in the (1) case of multiple chunks, we'd display all of them to the user, but just keep a running hash instead of keeping the whole value in memory.
Or, in case (3), if the chunk is larger than say 1024 bytes, we put only its hash in CBOR to avoid producing large intermediate representations.

(or, instead of making decisions, we could hash always)

matejcik:
recording for posterity, discussed elsewhere:

We want to disable the hashing feature for now and only support non-hashed signing.

To avoid collisions with hashed signing, we want to **reject any payload** that is:
* exactly 28 bytes long (so it could be a Blake2b224 hash)
* and not fully ASCII printable

The chance of a valid payload coming out to a hash that is only composed of ASCII printable bytes is 1 in 1 trillion, which seems good enough.

jaskp:
Forced-pushed to rebase, the commits were quite old now and there were many conflicts with main.
Changes start with https://github.com/trezor/trezor-firmware/pull/3509/commits/64f97b9335ff4b02640dc4ecd9d3b8879194ef62

jaskp:
Forced-pushed to rebase, the commits were quite old now and there were many conflicts with main.
Relevant changes start with https://github.com/trezor/trezor-firmware/pull/3509/commits/fc48288ee6906fad17cb0757973754dcd42e90b0

common/protob/messages-cardano.proto
jaskp: https://github.com/trezor/trezor-firmware/pull/3509/commits/2e81b7d34a61d844f690a38f1dfbb64fc3cd4be9

common/protob/messages-cardano.proto
jaskp: https://github.com/trezor/trezor-firmware/pull/3509/commits/2e81b7d34a61d844f690a38f1dfbb64fc3cd4be9

python/src/trezorlib/cardano.py
jaskp: After replacing the chunking with offset+length, it isn't possible to use `expect=` param because it supports only one message type whereas in this case you don't know in advance whether you get a CardanoMessageDataRequest or CardanoMessageSignature.

core/src/apps/cardano/helpers/chunks.py
jaskp: https://github.com/trezor/trezor-firmware/pull/3509/commits/97619a8ee52527911bbc4d32777f94b9efe1d840

gubatron:
eyes on this please, lots of users can't claim midnight drop because they can't sign messages with their trezor ada wallets

python/src/trezorlib/cardano.py
```diff
@@ -996,3 +1015,41 @@ def sign_tx(
...
+        requested_size = response.length
+
+        if offset + requested_size > len(payload):
+            raise ValueError("Device requested data beyond payload bounds")
+
+        chunk_data = payload[offset : offset + requested_size]
+        response = session.call(m.CardanoMessageDataResponse(data=chunk_data))
+
+    if not isinstance(response, m.CardanoMessageSignature):
+        raise ValueError("Unexpected response")
```
matejcik, line 1055:
```suggestion
    response = m.CardanoMessageSignature.ensure_isinstance(response)
```

matejcik:
ACK from me, leaving the rest to @obrusvit 

whoever ends up doing the final review, please specifically confirm that we're showing the whole payload to the user; there are remnats in the code for "first chunk" display

ibz:
> whoever ends up doing the final review, please specifically confirm that we're showing the whole payload to the user

I am taking this over.

* I can confirm that we are requesting the whole data in a single chunk as per https://github.com/trezor/trezor-firmware/pull/3509/files#diff-ff436e870f332b283aae9a85591a15b35cbbba59bb134c7b515502cd99783d9cR101 - even though the chunkification mechanism stays in place, it is not actually used
* the whole data is passed to the layout function, whose parameter I am renaming from `payload_first_chunk` to `payload` as you suggested to avoid future confusion
* *HOWEVER* if the data is too long it is truncated on display and `...` is appended as per https://github.com/trezor/trezor-firmware/pull/3509/files#diff-718c35fab848e1ee6f5e2a387c71b0a0ea8d84039fedce530cb9f0584c164495R381

So your request @matejcik is both fulfilled and unfulfilled at the same time.

Do you think we should raise an error if the data is too long rather than truncating? Or simply not truncate and keep scrolling?

matejcik:
not truncate, keep scrolling.
we can employ the mechanism for skipping long messages that we came up with for Ethereum; not sure if that was ever implemented or not

obrusvit (approved):
utACK

gubatron:
Great job guys.
Now we need the same for Solana

PR #3512, opened 2024-02-02
refactor(core): shrink size of Paragraphs struct

### LINKED ISSUE #3450: Decrease model_tt::component::AddressDetail size
`AddressDetails<StrBuffer>` is currently 3772B, with 16K stack this significantly contributes to stack overflows in `_chunkify_details` tests in `PYOPT=0` fw.

This has started happening between 2.6.3 and 2.6.4, git bisect might provide some clues.

I wonder if we could write clippy lint or something that could detect huge structs being put on the stack.

cepetr:
This PR resolves #3450 by implementing a small change that reduces the size of the `AddressDetails` struct from **3308** bytes to **2028** bytes. 

To achieve this, I've introduced `TextLayoutProxy`, which acts as a proxy for the heavy `TextLayout` structure stored within the `visible` field of the `Paragraphs` component. The `TextLayoutProxy` dynamically constructs the `TextLayout` structure as needed.

**In implementing `TextLayoutProxy`, it leverages `ParagraphSource.at()` to retrieve referenced paragraphs and their layouts. This decision involves some tradeoff in terms of speed.**

### DIFF
--- a/core/embed/rust/src/ui/component/text/paragraphs.rs
+++ b/core/embed/rust/src/ui/component/text/paragraphs.rs
@@ -5,7 +5,9 @@ use crate::{
     ui::{
         component::{Component, Event, EventCtx, Never, Paginate},
         display::toif::Icon,
-        geometry::{Alignment, Alignment2D, Insets, LinearPlacement, Offset, Point, Rect},
+        geometry::{
+            Alignment, Alignment2D, Dimensions, Insets, LinearPlacement, Offset, Point, Rect,
+        },
     },
 };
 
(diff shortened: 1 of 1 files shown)


Hannsek:
What exactly does that do from UI pov?

matejcik:
> What exactly does that do from UI pov?

it should not do anything. any UI diffs are a bug

matejcik (approved):
style check is still complaining, otherwise LGTM

mmilata (approved):
Looking good :+1:

PR #3521, opened 2024-02-06
Ability to decode u32 and u64 uvarints separately.

ibz:


prusnak:
Your PR is missing the motivation for the change. What is the goal you are trying to achieve?

core/embed/rust/src/protobuf/decode.rs
```diff
@@ -289,17 +289,125 @@ impl<'a> InputStream<'a> {
...
+        let result = stream.read_uvarint_u32();
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), 0xF0000000);
+
+        // 5 bytes with more than 4 bits used in the last byte => overflow
+        let mut stream = InputStream::new(&[0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b00010000]);
+        assert!(stream.read_uvarint_u32().is_err());
+
+        // 5+ bytes => overflow
+        let mut stream = InputStream::new(&[0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b10000000]);
```
matejcik, line 373: please make sure that the encoded value is valid -- i.e., if `read_uvarint_u32` is _not_ correctly overflowing here, maybe the error is because the number is too short

maybe add a testcase that reads the same number as u32 and u64 and checks that one succeeds and the other one fails?

core/embed/rust/src/protobuf/error.rs
```diff
@@ -10,6 +10,10 @@ pub const fn unknown_field_type() -> Error {
     value_error!("Unknown field type.")
 }
 
+pub const fn overflow() -> Error {
```
matejcik, line 13: instead, use `Error::OutOfRange`. then you can match for that in tests

core/embed/rust/src/protobuf/decode.rs
```diff
@@ -289,17 +289,125 @@ impl<'a> InputStream<'a> {
...
+        let mut stream = InputStream::new(&[0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b00010000]);
+        let result = stream.read_uvarint_u64();
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), 0x100000000);
+
+        // 10 bytes with just a single bit in the last one
+        let mut stream = InputStream::new(&[0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b10000000, 0b00000001]);
+        let result = stream.read_uvarint_u64();
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), 0x8000000000000000);
```
matejcik, line 401:
```suggestion
        assert!(matches!(result, Ok(0x8000000000000000)));
```

core/embed/rust/src/protobuf/decode.rs
```diff
@@ -289,17 +289,125 @@ impl<'a> InputStream<'a> {
...
+                    return Err(error::overflow());
+                }
+            }
+            uint += payload << shift;
+            shift += 7;
+            if !continue_decoding {
+                break;
+            }
+        }
+        Ok(uint)
```
matejcik, line 313: nit: maybe you could get a tiny bit cleaner code by reversing this:
"if !continue_decoding { return Ok(uint) }" + in the failing cases break the loop to an Err(Overflow) result
(that lets you not keep explicit track of `continue_decoding`, and instead say "if byte_count == 5 && payload > 0xF { break }")

matejcik:
@prusnak : related: #2871 which also has rationale

matejcik:
next steps:

* add differentiation between (u/s)int(32/64) to `pb2py`
* refactor value decoding so that 32bit numbers are read by `read_uvarint_u32`
* add (python-side) tests for protobuf messages that declare 32bit fields and then send 64bit values

core/embed/rust/src/protobuf/decode.rs
```diff
@@ -211,30 +211,41 @@ impl Decoder {
...
                 let signed_int = zigzag::to_signed(num);
                 Ok(signed_int.try_into()?)
             }
             FieldType::Bool => {
-                let boolean = num != 0;
+                let boolean = stream.read_uvarint_u32()? != 0;
                 Ok(boolean.into())
             }
             FieldType::Bytes => {
+                let num = stream.read_uvarint_u64()?;
```
matejcik, line 234: lengths should be 32 bits

core/embed/rust/src/protobuf/decode.rs
```diff
@@ -211,30 +211,41 @@ impl Decoder {
...
                 let buf_len = num.try_into()?;
                 let buf = stream.read(buf_len)?;
                 buf.try_into()
             }
             FieldType::String => {
+                let num = stream.read_uvarint_u64()?;
```
matejcik, line 240: lengths should be 32 bits

core/embed/rust/src/protobuf/decode.rs
```diff
@@ -243,6 +254,7 @@ impl Decoder {
                 }
             }
             FieldType::Msg(msg_type) => {
+                let num = stream.read_uvarint_u64()?;
```
matejcik, line 257: lengths should be 32 bits

core/tests/test_trezor.protobuf.py
```diff
@@ -1,14 +1,14 @@
...
-from trezor.messages import WebAuthnCredential, Failure, SignMessage, DebugLinkMemoryRead
+from trezor.messages import AuthorizeCoinJoin, WebAuthnCredential, Failure, SignMessage, DebugLinkMemoryRead
 
 
 def load_uvarint32(data: bytes) -> int:
     # use known uint32 field in an all-optional message
     buffer = bytearray(len(data) + 1)
     buffer[1:] = data
-    buffer[0] = (1 << 3) | 0  # field number 1, wire type 0
+    buffer[0] = (1 << 3) | 7  # field number 1, wire type 7
```
matejcik, line 11: I don't think this change is appropriate? the "wire type" is from protobuf spec, 0 means "varint". The enum values we use internally have nothing to do with this.

core/tests/test_trezor.protobuf.py
```diff
@@ -28,13 +28,13 @@ def dump_uvarint32(value: int) -> bytearray:
     length = protobuf.encoded_length(msg)
     buffer = bytearray(length)
     protobuf.encode(buffer, msg)
-    assert buffer[0] == (1 << 3) | 0  # field number 1, wire type 0
+    assert buffer[0] == (1 << 3) | 7  # field number 1, wire type 7
```
matejcik, line 31: dtto

matejcik (changes requested):

matejcik:
superseded by #7762 

(there's still the tiny perf optimization of decoding u32 first and upshifting to u64 that we're leaving on the table)

Issue #3128, opened 2023-07-04
Monero tests can't see WebUSB device


mmilata:
Since eacd15383b6801fc09db9d51b31bdfb83da09097 hardware tests in CI are failing. Seems to be locally reproducible:
```
$ trezorctl list
webusb:001:3 - (unnamed) [Trezor T, F569E14AB37A74AC276937FA]

$ ./trezor_monero_tests --trezor-path webusb:
2023-07-04 13:45:00.752	D Device 0 Created
2023-07-04 13:45:00.752	I Test versions Fluorine Fermi' (v0.18.1.0-unknown)
2023-07-04 13:45:00.752	I Testing hardforks [15, 15], sync-test: 1
2023-07-04 13:45:00.753	D Transaction tests for HF 15
2023-07-04 13:45:00.754	D Enumerating Trezor devices...
2023-07-04 13:45:00.755	I [PARSE URI] regex not matched for uri: ^(([^:]*?)://)?(\[(.*)\](:(\d+))?)(.*)?
2023-07-04 13:45:00.755	D Reconnecting...
2023-07-04 13:45:00.757	D Some problems at connect, message: Connection refused
2023-07-04 13:45:00.757	D Failed to connect to 127.0.0.1:21325
2023-07-04 13:45:00.757	E Failed to invoke http request to  /enumerate
2023-07-04 13:45:00.757	E BridgeTransport enumeration failed:Bridge enumeration failed
2023-07-04 13:45:02.260	D Problems at cancel: Bad file descriptor
2023-07-04 13:45:02.260	D Problems at shutdown: Bad file descriptor
2023-07-04 13:45:02.260	D Enumeration yielded 0 Trezor devices
2023-07-04 13:45:02.260	E No matching Trezor device found. Device specifier: "webusb:"
2023-07-04 13:45:02.260	E Could not connect to the device Trezor:webusb:
2023-07-04 13:45:02.261	E gen_trezor_base generation failed: what=Could not connect to the device Trezor:webusb:
2023-07-04 13:45:02.262	D Hardfork added at height: 0, from 1 to 2
2023-07-04 13:45:02.263	E Chain setup failed for HF15
2023-07-04 13:45:02.264	E Exception at [main], what=Couldn't fill transaction sources
2023-07-04 13:45:02.357	D Device 0 Destroyed

```

The test binary can see the device through Trezor Bridge after I start it but it doesn't support debuglink. Emulator seems to also work fine. Any clues @ph4r05?

[...]

PR #3522, opened 2024-02-06
ci: hardware tests for github actions

mmilata:
Still failing most of the time but I'm not capable of pushing it further. Previously it was failing always so I suppose that can be considered improvement.

Right now there are 4 builders, each with one device connected:
* 1x T1
* 1x TT
* 2x TS3 (one with dead optiga, currently tests are always run on firmware with optiga disabled)

Don't forget to remove the `pull_request` trigger before merging so that these are not run on every PR but only nightly on main.


mmilata:
Note to monday self: there's something wrong with runner4, the one that had functioning Optiga. I wonder if I fried it or something.

mmilata:
The tests are failing more often than not but I think it's time to merge since fixing everything is out of scope for this PR and it's still in better shape than it was on GitLab. Plus we can gather more data by running the workflow every night.

Failures in particular:

* `runner4` doesn't see the device at all, since there's another runner with T2B1 we can leave it for now & i'll investigate when i'm in the office.
* Model 1 all-coins firmware fails to upload the firmware due to bad timings of the button sequence, btconly seems fine. Camera's been a great help here.
* Monero tests fail with `Illegal instruction`, will investigate further after #3128 is resolved.
* T2T1 & T2B1 look mostly fine except some memory-intensive tests. Will probably blacklist them after discussing with the rest of the FW team.

Want to smash the Approve button @vdovhanych? Then I'll rebase this and remove the `pull_request` trigger.

vdovhanych (approved):
Yeah, lets merge it and try to improve it later.

PR #3532, opened 2024-02-12
Fix hardware build

grdddj:
Adding missing `debug` feature, which was missing and causing development keys not being included for translations

### DIFF
--- a/core/SConscript.firmware
+++ b/core/SConscript.firmware
@@ -720,7 +720,7 @@ def cargo_build():
     features.append('ui')
     features.append('translations')
     if PYOPT == '0':
-        features.append('ui_debug')
+        features.append('debug')
 
     features.extend(FEATURES_AVAILABLE)
 


mmilata:
Vaguely recall that we made `PYOPT=0` not enable this because of:
```
/// More detailed panic handling. The difference against
/// default `panic` below is that this "debug" version
/// takes around 10 kB more space in the flash region.
```
Might want to split this or the signature verification in a separate feature eventually.

mmilata (approved):

grdddj:
> Vaguely recall that we made `PYOPT=0` not enable this because of:
> 
> ```
> /// More detailed panic handling. The difference against
> /// default `panic` below is that this "debug" version
> /// takes around 10 kB more space in the flash region.
> ```
> 
> Might want to split this or the signature verification in a separate feature eventually.

Good point, I will create an issue.

From a quick look at CI builds (`core fw regular debug build`), the results are very tiny fortunately: 

before:
```
           FLASH:        716 KB       768 KB     93.23%
          FLASH2:      848904 B       896 KB     92.52%
```

after:
```
           FLASH:        716 KB       768 KB     93.23%
          FLASH2:      849416 B       896 KB     92.58%
```

PR #3538, opened 2024-02-20
Rename trezorctl self-test to prodtest-t1.

### LINKED ISSUE #3504: rename device self-test
please rename this test thats not confusing trezorctl users

```
❯ trezorctl device self-test
Error: UnexpectedMessage: Unexpected message
```

obrusvit:
It was confusing for trezorctl users. The new name clearly signals that it's meant to work only for Model One.

Resolves #3504


mmilata (approved):
LGTM, thanks!

python/src/trezorlib/cli/device.py
```diff
@@ -53,12 +53,12 @@ def cli() -> None:
 
 @cli.command()
 @with_client
-def self_test(client: "TrezorClient") -> str:
-    """Perform a factory self-test.
+def prodtest_t1(client: "TrezorClient") -> str:
```
matejcik, line 56: while we're here, let's move the command to `debug` section

obrusvit: https://github.com/trezor/trezor-firmware/pull/3538/commits/26c17c187235fb3d0d1c6432adf2cb7a18aaba37

python/src/trezorlib/cli/debug.py
```diff
@@ -19,7 +19,11 @@
 import click
 
 from .. import mapping, messages, protobuf
-from ..debuglink import TrezorClientDebugLink, record_screen
+from ..client import TrezorClient
+from ..debuglink import TrezorClientDebugLink
+from ..debuglink import prodtest_t1 as debuglink_prodtest_t1
+from ..debuglink import record_screen
```
matejcik, line 25: i'm very surprised that isort is OK with this chunk, but if yes, let's go ahead

matejcik (approved):

python/src/trezorlib/cli/debug.py
obrusvit: yep, I also double checked when I saw it :D

PR #3544, opened 2024-02-22
Report mismatched translation blob in Features

### LINKED ISSUE #3543: Report "outdated translation blob" to host
this should probably be a new `Features` field? so that Suite can detect this and silently update to matching version.

note that with current code, it doesn't matter which direction the version goes, silent upgrade as well as downgrade is allowed. This seems right in that we ideally want a translation blob matching the firmware version. (e.g., in newer blobs, some older strings may be omitted)

mmilata:
Also returns False when blob is newer than firmware. Comparison includes build version (as in major.minor.patch.build) which we don't report in Features, not completely sure if that's right.

Fixes #3543.

### DIFF
--- a/common/protob/messages-management.proto
+++ b/common/protob/messages-management.proto
@@ -131,6 +131,7 @@ message Features {
     optional uint32 homescreen_width = 47;         // homescreen width in pixels
     optional uint32 homescreen_height = 48;         // homescreen height in pixels
     optional bool bootloader_locked = 49;       // bootloader is locked
+    optional bool language_version_matches = 50 [default=true];  // translation blob version matches firmware version
 }
 
 /**
--- a/core/src/apps/base.py
+++ b/core/src/apps/base.py
@@ -43,6 +43,20 @@ def busy_expiry_ms() -> int:
     return expiry_ms if expiry_ms > 0 else 0
 
 
+def _language_version_matches() -> bool | None:
+    """
+    Whether translation blob version matches firmware version.
+    Returns None if there is no blob.
+    """
+    from trezor import translations
+
+    header = translations.TranslationsHeader.load_from_flash()
+    if header is None:
+        return True
+
+    return header.version == utils.VERSION
+
+
 def get_features() -> Features:
     import storage.recovery as storage_recovery
     from trezor import translations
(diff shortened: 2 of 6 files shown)


matejcik:
there's a bug, tests are broken.
(i don't see the bug from the code though)

core/src/apps/base.py
```diff
@@ -43,6 +43,20 @@ def busy_expiry_ms() -> int:
...
+def _language_version_matches() -> bool | None:
+    """
+    Whether translation blob version matches firmware version.
+    Returns None if there is no blob.
+    """
+    from trezor import translations
+
+    header = translations.TranslationsHeader.load_from_flash()
+    if header is None:
+        return None
```
matejcik, line 55: I would lean towards returning `True` here
semantically, "language is en-US and it does not need update"

common/protob/messages-management.proto
```diff
@@ -131,6 +131,7 @@ message Features {
     optional uint32 homescreen_width = 47;         // homescreen width in pixels
     optional uint32 homescreen_height = 48;         // homescreen height in pixels
     optional bool bootloader_locked = 49;       // bootloader is locked
+    optional bool language_version_matches = 50;   // translation blob version matches firmware version
```
matejcik, line 134: perhaps even `default=true` to resolve old models that don't send this field?

that way Suite can just `if (!language_version_matches) { sendLanguageBlob(...) }` in exactly the situations when required?
cc @mroz22 what do you think?

core/src/apps/base.py
mmilata: Fixed in [a6dfe95](https://github.com/trezor/trezor-firmware/pull/3544/commits/a6dfe9558cdcdba2723d39a7b59b90fe1674e7b5), this was actually the reason for the failing test.

matejcik: but now you also need to update the test :)

mmilata: uhhhh yeah 3215107b2da53aba9207586677f55404f3443c19

common/protob/messages-management.proto
mmilata: added in 7abca4e5ec0e4f4072d0d6227cd94b2d6ec90fa4

matejcik (approved):

PR #3505, opened 2024-01-30
fix(core/bitcoin): display descriptors for taproot XPUBs

mmilata:
Fixes #3475. I'm assuming there's no need to add new flag to GetPublicKey since non-descriptor Taproot XPUBs are non-standard? I'm also assuming we're not displaying the descriptor checksum since it's not on the suite screenshot in #3475.

Unfortunately on T2B1 the changes [increase the number of screens to three:](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/6057776893/artifacts/master_diff/diff/TR-device_tests-bitcoin-test_descriptors.py%3A%3Atest_descriptors%5BBitcoin-1-86-InputScriptType.SPENDTAPROOT-tr%28-3a85f3dd.html). ddc50e277f04cd82e96ddc92a3dfe9a3f5641bcc3f05ed3e7e3b7b0d96f0b7a6

CI note: master diff is broken because we're skipping T2B1 tests so the screens are missing. [GitLab master diff](https://satoshilabs.gitlab.io/-/trezor/trezor-firmware/-/jobs/6057776893/artifacts/master_diff/index.html).

PR #3545, opened 2024-02-23
Send BIP-380 descriptor in GetPublicKey response

### LINKED ISSUE #3539: Send descriptor in GetPublicKey response for Taproot accounts
Currently Suite has to compute the descriptor checksum, and because the checksum is computed from pubkey representation and one pubkey has multiple legal representations this can lead to the undesirable case where Suite displays different checksum than firmware. Two possibilities:

1. Send the descriptor including checksum in `PublicKey.xpub` for Taproot accounts. Disadvantage: breaks compatibility with older firmware.
2. Add new field `PublicKey.descriptor` and put the descriptor there. Advantage is that it could work for non-Taproot accounts too. Disadvantage is that Suite needs to compute the checksum itself for older firmware versions.

Previous descriptor work: #1362 #1363 #1888 #3505.

mmilata:
Closes #3539.

Turns out whoever implemented #3505 has used fingerprint of the wrong node (leaf instead of root).

Also, are we using `h` or `'` for hardened nodes? Went with `h` for now.


prusnak:
> Also, are we using `h` or `'` for hardened nodes? Went with `h` for now.

Let's use the `h` everywhere!

python/src/trezorlib/cli/btc.py
```diff
@@ -331,6 +320,21 @@ def _get_descriptor(
         unlock_path=get_unlock_path(n),
     )
 
+    # Starting with core 2.6.5 the descriptor is included in the response.
+    if pub.descriptor is not None:
+        return pub.descriptor
+
+    if script_type == messages.InputScriptType.SPENDADDRESS:
```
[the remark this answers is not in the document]
mmilata: AFAIK we currently don't share any code between trezorlib and firmware. And it needs to exist in trezorlib in order to be compatible with older firmware versions, but at least new script types don't need to be added there. But it's possible I'm missing something.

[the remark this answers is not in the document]
peter-sanderson: Feel free to resolve my comment, would be nice to share code. But if not possible/worth it, no big deal. :v:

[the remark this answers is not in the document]
matejcik: @peter-sanderson sharing code between trezorlib and core is a no-go. we tried that before but there are a lot of subtle differences between the environments, or we need different refactors on different sides, it's better to keep copies of things

core/src/apps/bitcoin/get_public_key.py
```diff
@@ -115,24 +117,39 @@ async def get_public_key(
...
+    if script_type == InputScriptType.SPENDADDRESS:
+        fmt = "pkh({})"
+    elif script_type == InputScriptType.SPENDP2SHWITNESS:
+        fmt = "sh(wpkh({}))"
+    elif script_type == InputScriptType.SPENDWITNESS:
+        fmt = "wpkh({})"
+    elif script_type == InputScriptType.SPENDTAPROOT:
+        fmt = "tr({})"
+    else:
+        return None
```
matejcik, line 146: when does this happen?
or iow, why not raise an exception here like trezorlib did?

tests/device_tests/bitcoin/test_descriptors.py
```diff
@@ -160,7 +169,22 @@ def test_descriptors(client: Client, coin, account, purpose, script_type, descri
...
+        res = btc.get_public_node(
+            client,
+            _address_n(purpose, coin, account, script_type),
+            show_display=True,
+            coin_name=coin,
+            script_type=script_type,
+            ignore_xpub_magic=True,
+            unlock_path=btc_cli.get_unlock_path(address_n),
+        )
+        assert res.descriptor == descriptors
```
matejcik, line 190: this whole block should be under the model check, no?

also please make a separate test for this

(also please start using `client.model is models.T1B1` instead of string comparison from now on)

core/src/apps/bitcoin/get_public_key.py
mmilata: There's `InputScriptType.SPENDMULTISIG` which is not currently handled, I guess the descriptor will look very different. There's also `InputScriptType.EXTERNAL` but I think that does not make sense in `get_public_key` context.

matejcik: i don't think `SPENDMULTISIG` makes sense either though? the `GetPublicKey` message does not have a `multisig` field and a "multisig public key at path" is not a thing that makes sense either way.

OTOH I see now that SPENDMULTISIG is handled explicitly in there so i guess no reason to just break someone's misconfigured setup...

tests/device_tests/bitcoin/test_descriptors.py
mmilata: fixed in [e4f6b6b](https://github.com/trezor/trezor-firmware/pull/3545/commits/e4f6b6b6343455f445abfb021533c8ae39f66f92)
will need to update T1 fixtures still

mmilata:
okay to merge? @matejcik

matejcik (approved):

PR #3546, opened 2024-02-23
feat: Unknown instruction show program id

Arrowana:
When calling an unknown instruction, the accounts and data is meaningless without the program id.
In fact, it is better to have user rely on "trusted" program ids, which gives a better guarantee than unreadable and complex accounts and instruction data.

Unclear: How to deal with all the translations, it looks like it should come from an API.


matejcik:
yup, sounds about right.

this btw exposes a nasty surprise with translations, namely, if the format string changes, it can mess up substitution positions
cc @grdddj @Hannsek 

we should replace all `{}`s in translations with `{0}` `{1}` and so on

prusnak:
> we should replace all `{}`s in translations with `{0}` `{1}` and so on

Well, this does not change anything if contributor decides to change

foo {0} bar {1} baz

to

haha {0} foo {1} bar {2} baz

instead of the correct

haha {2} foo {0} bar {1} baz

Point being that reviewer still needs to understand what is going on no matter which method we use.

Arrowana:
I tried this in the emulator and it looks horrible 

Screenshot 2024-02-23 at 11 37 59 pm

Also without the ability to scroll further down, i'll look into the ui options to make this better

matejcik:
> Well, this does not change anything if contributor decides to change

i mean sure, there are a lot of ways contributors can mess things up :) the point here is that with explicit numbered parameters this problem is easily discoverable, and it also explicitly allows translators to reorder the parameters as they need (e.g., for languages with different word ordering rules)

Of course the larger problem with template strings in translations is that they aren't really forwards/backwards compatible. If a parameter gets added in some version, and the user downgrades the firmware but not the language blob, they will get a crash when trying to format the template.

PR #3548, opened 2024-02-23
improve flexibility of combine script

### LINKED ISSUE #3123: combine script does not combine firmware properly
TT firmwares are split into two discrete sections. When attempting to combine board+boot+firmware, combine script makes a binary, but does not account for the "hole" in flash that needs to be skipped.

TychoVrahe:
Also, combine_fw makefile target is added.

closes #3123 

[no changelog]


matejcik:
It's flexible alright but doesn't work for my usecase anymore, namely, pick _arbitrary build artifacts_ from arbitrary historic versions and/or Notion archives and combine them for the factory blob -- at least not without memorizing some ugly constants.

could we do something like, put the memory layouts into separate files and make the call look like `combine.py --layout u5.layout boardloader.bin bootloader.bin [firmware.bin [firmware.p2.bin ...]]` ?

TychoVrahe:
Actually, that specific usecase was broken already - for MCUs different than STM32F4.

So here is more complex proposal: https://github.com/trezor/trezor-firmware/commit/e22248f725968131bc4d34762c2b11dd172c1dd0

Using `core/embed/models/model_XXXX.h` as a source of truth, `makefile` and combine script are using it via new `layout_parser` script. The script is quite naive so it prevents using preprocessor features, but upside is that we duplicate these constants one less time (some of them are still duplicated in linker scripts)

TychoVrahe:
the usage requires for every passed binary to specify the name of the binary, so that the layout parser knows what to look for, e.g.
`-b BOOTLOADER ./build/bootloader/bootloader.bin`

matejcik:
the PR is now broken, it seems to include all of U5 changes

TychoVrahe:
rebased.

PR #3551, opened 2024-02-25
Remove CoSi functionality from trezorctl

### LINKED ISSUE #2975: CoSi signing from command line is broken on TT
When introducing CoSi for TT, we changed cosi_commit + cosi_sign into a single flow.

This breaks trezorctl, because it sends Initialize before every message. This kicks TT out of the flow and it can't follow up with `CosiSign`. As a result, you get "unexpected message" every time.

Options:
* firmware-side, turn it into two separate flows again. This has _some_ advantages: logically, the CoSi protocol is two-turn, with the commit being a separate thing that could plausibly be done at a different time.
   * OTOH, the nonce is kept in short-lived cache anyway and you are technically supposed to do this online
* trezorctl side, convert the CoSi signing command so that it runs interactively and prompts the user to enter the global commitments within that one session.
   * sucks for scriptability, but then again if you're scripting the CoSi flow, just write it out in python directly
* or something something allow the cosi-sign command to proceed without `Initialize`, maybe the same way `debug send-bytes` does it, without instantiating a client at all?
  * sounds unsystematic

obrusvit:
Resolves #2975


common/protob/messages.proto
```diff
@@ -168,10 +168,6 @@ enum MessageType {
     MessageType_SignedIdentity = 54 [(bitcoin_only) = true, (wire_out) = true];
     MessageType_GetECDHSessionKey = 61 [(bitcoin_only) = true, (wire_in) = true];
     MessageType_ECDHSessionKey = 62 [(bitcoin_only) = true, (wire_out) = true];
-    MessageType_CosiCommit = 71 [(bitcoin_only) = true, (wire_in) = true];
-    MessageType_CosiCommitment = 72 [(bitcoin_only) = true, (wire_out) = true];
-    MessageType_CosiSign = 73 [(bitcoin_only) = true, (wire_in) = true];
-    MessageType_CosiSignature = 74 [(bitcoin_only) = true, (wire_out) = true];
```
matejcik, line 174: replace with `reserved` values

matejcik (changes requested):

common/protob/messages.proto
obrusvit: real estate reserved https://github.com/trezor/trezor-firmware/pull/3551/commits/2081831cf3b444e2e89414bec3d54b59dfa01d19

PR #3552, opened 2024-02-26
Translations fixes/shortenings

grdddj:
Translations fixes before beta release - @Hannsek


grdddj:
> What about fixtures?

I am recording them from CI, as running all the various tests locally is a nightmare.

Are there some more changes we want to do here?

Hannsek:
@matejcik is it ready to merge and low hanging? If not, let's put it into the next release

matejcik:
it's just this one commit and the only thing needed is to run UI tests and check that it didn't break anything

grdddj:
> it's just this one commit and the only thing needed is to run UI tests and check that it didn't break anything

I have included the commit here, with `make gen` fixup and with one fix for `description_param` - commit [69743d1](https://github.com/trezor/trezor-firmware/pull/3552/commits/69743d189126f81e3cf0b19d6d823b5330878bdd) -  was broken here - 
image

matejcik (approved):

PR #3553, opened 2024-02-26
configure option bytes on STM32U5

TychoVrahe:
-fixes option bytes configuration on STM32U5
- improves robustness of BHK handling on STM32U5
- fixes missing symlinks

### DIFF
--- a/core/embed/boardloader/main.c
+++ b/core/embed/boardloader/main.c
@@ -106,7 +106,7 @@ void write_bootloader_min_version(uint8_t version) {
 }
 #endif
 
-struct BoardCapabilities capablities
+struct BoardCapabilities capabilities
     __attribute__((section(".capabilities_section"))) = {
         .header = CAPABILITIES_HEADER,
         .model_tag = TAG_MODEL_NAME,
--- a/core/embed/boardloader/main.c
+++ b/core/embed/boardloader/main.c
@@ -67,7 +67,7 @@ static const uint8_t * const BOARDLOADER_KEYS[] = {
 };
 
 #ifdef STM32U5
-void check_bootloader_version(uint8_t bld_version) {
+uint8_t get_bootloader_min_version(void) {
   const uint8_t *counter_addr =
       flash_area_get_address(&SECRET_AREA, SECRET_MONOTONIC_COUNTER_OFFSET,
                              SECRET_MONOTONIC_COUNTER_LEN);
@@ -92,10 +92,12 @@ void check_bootloader_version(uint8_t bld_version) {
     }
   }
 
-  ensure((bld_version >= counter) * sectrue, "BOOTLOADER DOWNGRADED");
+  return counter;
+}
 
-  if (bld_version > counter) {
-    for (int i = 0; i < bld_version; i++) {
+void write_bootloader_min_version(uint8_t version) {
+  if (version > get_bootloader_min_version()) {
+    for (int i = 0; i < version; i++) {
       uint32_t data[4] = {0};
       secret_write((uint8_t *)data, SECRET_MONOTONIC_COUNTER_OFFSET + i * 16,
                    16);
@@ -159,6 +161,12 @@ static uint32_t check_sdcard(void) {
                                           BOARDLOADER_KEYS)) {
       return 0;
     }
+
+#ifdef STM32U5
+    if (hdr->monotonic < get_bootloader_min_version()) {
+      return 0;
+    }
+#endif
 
     return hdr->codelen;
   }
@@ -255,6 +263,8 @@ int main(void) {
   }
 
   trustzone_init_boardloader();
+
+  secret_ensure_initialized();
 #endif
 
 #ifdef STM32F4
@@ -280,6 +290,25 @@ int main(void) {
 #if defined USE_SD_CARD
   sdcard_init();
 
+#ifdef STM32U5
+  // If the bootloader is being updated from SD card, we need to preserve the
+  // monotonic counter from the old bootloader. This is in case that the old
+  // bootloader did not have the chance yet to write its monotonic counter to
+  // the secret area - which normally happens later in the flow.
+  const image_header *old_hdr = read_image_header(
+      (const uint8_t *)BOOTLOADER_START, BOOTLOADER_IMAGE_MAGIC,
+      flash_area_get_size(&BOOTLOADER_AREA));
+
+  if ((old_hdr != NULL) &&
+      (sectrue == check_image_header_sig(old_hdr, BOARDLOADER_KEY_M,
+                                         BOARDLOADER_KEY_N,
+                                         BOARDLOADER_KEYS)) &&
+      (sectrue ==
+       check_image_contents(old_hdr, IMAGE_HEADER_SIZE, &BOOTLOADER_AREA))) {
+    write_bootloader_min_version(old_hdr->monotonic);
+  }
+#endif
+
   if (check_sdcard()) {
     return copy_sdcard() == sectrue ? 0 : 3;
   }
@@ -300,7 +329,12 @@ int main(void) {
          "invalid bootloader hash");
 
 #ifdef STM32U5
-  check_bootloader_version(hdr->monotonic);
+  uint8_t bld_min_version = get_bootloader_min_version();
+  ensure((hdr->monotonic >= bld_min_version) * sectrue,
+         "BOOTLOADER DOWNGRADED");
+  // Write the bootloader version to the secret area.
+  // This includes the version of bootloader potentially updated from SD card.
+  write_bootloader_min_version(hdr->monotonic);
 #endif
 
   ensure_compatible_settings();
--- a/core/embed/boardloader/main.c
+++ b/core/embed/boardloader/main.c
@@ -122,7 +122,7 @@ struct BoardCapabilities capablities
         .terminator_length = 0};
 
 // we use SRAM as SD card read buffer (because DMA can't access the CCMRAM)
-BUFFER_SECTION uint32_t sdcard_buf[IMAGE_HEADER_SIZE / sizeof(uint32_t)];
+BUFFER_SECTION uint32_t sdcard_buf[BOOTLOADER_IMAGE_MAXSIZE / sizeof(uint32_t)];
 
 #if defined USE_SD_CARD
 static uint32_t check_sdcard(void) {
@@ -138,8 +138,8 @@ static uint32_t check_sdcard(void) {
 
   memzero(sdcard_buf, IMAGE_HEADER_SIZE);
 
-  const secbool read_status =
-      sdcard_read_blocks(sdcard_buf, 0, IMAGE_HEADER_SIZE / SDCARD_BLOCK_SIZE);
+  const secbool read_status = sdcard_read_blocks(
+      sdcard_buf, 0, BOOTLOADER_IMAGE_MAXSIZE / SDCARD_BLOCK_SIZE);
 
   sdcard_power_off();
 
@@ -160,6 +160,21 @@ static uint32_t check_sdcard(void) {
                                           BOARDLOADER_KEY_N,
                                           BOARDLOADER_KEYS)) {
       return 0;
+    }
+
+    _Static_assert(IMAGE_CHUNK_SIZE >= BOOTLOADER_IMAGE_MAXSIZE,
+                   "BOOTLOADER IMAGE MAXSIZE too large for IMAGE_CHUNK_SIZE");
+
+    if (sectrue != (check_single_hash(
+                       hdr->hashes, ((const uint8_t *)sdcard_buf) + hdr->hdrlen,
+                       hdr->codelen))) {
+      return 0;
+    }
+
+    for (int i = IMAGE_HASH_DIGEST_LENGTH; i < sizeof(hdr->hashes); i++) {
+      if (hdr->hashes[i] != 0) {
+        return 0;
+      }
     }
 
 #ifdef STM32U5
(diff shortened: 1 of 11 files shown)


core/embed/boardloader/main.c
```diff
@@ -68,6 +68,13 @@ static const uint8_t * const BOARDLOADER_KEYS[] = {
 
 #ifdef STM32U5
 void check_bootloader_version(uint8_t bld_version) {
+  if (sectrue != secret_verify_header()) {
```
cepetr, line 71: It would be better to move this code to a dedicated function. Clearing invalid secrets does not relate to bootloader version checking.

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -69,8 +69,29 @@ secbool secret_bhk_locked(void) {
...
+      optiga_secret_empty_bytes++;
+    }
+  }
+  return sectrue * (optiga_secret_empty_bytes != len);
+}
+
 void secret_bhk_provision(void) {
   uint32_t secret[SECRET_BHK_LEN / sizeof(uint32_t)] = {0};
+
+  ensure(secret_verify_header(), "secret sector not properly initialized");
```
cepetr, line 89: If the secret sector is not properly initialized, this could potentially lead to device bricking, especially if someone else does not reinitialize the secret sector (currently handled by the board-loader). It might be a bit safer to reinitialize the secret sector here if it's invalid

cepetr (changes requested):

core/embed/trezorhal/stm32u5/lowlevel.c
```diff
@@ -129,9 +130,8 @@ secbool flash_check_option_bytes(void) {
   }
 
 #if PRODUCTION
-  // TODO error this wont work, need to add default/reset values
-#error this wont work, need to add default/reset values
-  if (FLASH->WRP1AR != (WANT_WRP_PAGE_START | (WANT_WRP_PAGE_END << 16))) {
+  if (FLASH->WRP1AR !=
+      (WRP_LOCKED_VALUE | WANT_WRP_PAGE_START | (WANT_WRP_PAGE_END << 16))) {
```
cepetr, line 134: It would be better to have two constants with a little bit different purpose:

```c
#define WRPA1R_DEFAULT_VALUE 0xFF80FFFF  -- this is MCU dependent
#define WRP1AR_LOCKED_VALUE  ((WRP1AR_DEFAULT_VALUE & ...) | ...)
```
This allows comparison and potential setting of the register to the same constant without redundancy.

TychoVrahe: https://github.com/trezor/trezor-firmware/commit/9e229edbfb823c15ce42616a626879327154ce65

core/embed/boardloader/main.c
TychoVrahe: moved https://github.com/trezor/trezor-firmware/commit/0153fba65a000ac74667dae0b7e52551bfc66d6e

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: done here: https://github.com/trezor/trezor-firmware/commit/0153fba65a000ac74667dae0b7e52551bfc66d6e

note that i was counting on boardloader doing the reinitialization, as that is the place where bootloader downgrade protection needs to be written after the secret is deleted. Which is why i am still halting the execution after this gets re-initialized, to enforce going through boardloader and write the protection.

also, this should not really happen, if the secret gets deleted at any point, the device permanently loses optiga pairing data and thus cant authenticate as genuine anymore.

TychoVrahe: i also moved the bootloader downgrade protection here: https://github.com/trezor/trezor-firmware/commit/a86b30b12bd514d967c623ece8248e6f5313b8d8

with the logic that if bootloader re-initializes the secret, the downgrade protection needs to be filled before the potential sd-card bootloader upgrades happens, otherwise it would be filled with new data already potentially allowing bootloader downgrade

TychoVrahe: (https://github.com/trezor/trezor-firmware/pull/3553/commits/2b48ab400bf417774a94b03779ea94c65fe982a9 reverts wrongly commited makefile modification)

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -9,7 +9,7 @@
 static secbool bootloader_locked_set = secfalse;
 static secbool bootloader_locked = secfalse;
 
-static secbool verify_header(void) {
+secbool secret_verify_header(void) {
   uint8_t header[SECRET_HEADER_LEN] = {0};
```
matejcik, line 13: kinda feeling that you should memzero this at the end?
or maybe just copy `sizeof(SECRET_HEADER_MAGIC)` bytes because the secret doesn't seem to be used here either

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -69,8 +80,29 @@ secbool secret_bhk_locked(void) {
...
   uint32_t secret[SECRET_BHK_LEN / sizeof(uint32_t)] = {0};
+
+  ensure(secret_ensure_initialized(), "secret reinitialized");
+
+  if (sectrue != secret_present(SECRET_BHK_OFFSET, SECRET_BHK_LEN)) {
+    secret_bhk_regenerate();
+  }
+
   secbool ok =
       secret_read((uint8_t *)secret, SECRET_BHK_OFFSET, SECRET_BHK_LEN);
```
matejcik, line 107: don't forget to memzero this buffer

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -103,17 +135,7 @@ void secret_bhk_regenerate(void) {
 }
```
matejcik, line 135: just for total hygiene, `memzero(val)` at end too?

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -103,17 +135,7 @@ void secret_bhk_regenerate(void) {
...
-  for (int i = 0; i < SECRET_OPTIGA_KEY_LEN; i++) {
-    if (optiga_secret[i] == 0xFF) {
-      optiga_secret_empty_bytes++;
-    }
-  }
-  return sectrue * (optiga_secret_empty_bytes != SECRET_OPTIGA_KEY_LEN);
+  return secret_present(SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
 }
 
 void secret_optiga_backup(void) {
```
matejcik, line 141: memzero

matejcik: dtto for the F4 version too, didn't look too closely in there

core/embed/boardloader/main.c
```diff
@@ -276,6 +276,16 @@ int main(void) {
   display_init();
   display_clear();
 
+#ifdef STM32U5
+  const image_header *old_hdr = read_image_header(
+      (const uint8_t *)BOOTLOADER_START, BOOTLOADER_IMAGE_MAGIC,
+      flash_area_get_size(&BOOTLOADER_AREA));
+  secret_ensure_initialized();
+  if (old_hdr != NULL) {
+    get_and_fill_bootloader_version(old_hdr->monotonic);
```
matejcik, line 285: I believe that `copy_sd` should not allow installing a downgrade in the first place, i.e., the check should be (a) in `copy_sd` and (b) when getting ready to jump to bootloader.

Placing the check here seems weird. What I would expect the most is splitting `get_and_fill` into `get` and `fill`, with `get` performing the check & halt, and `fill` just writing data. Then `copy_sd` would call `get` (and possibly halt), while the finalization would call both.

core/embed/boardloader/main.c
```diff
@@ -276,6 +276,16 @@ int main(void) {
   display_init();
   display_clear();
 
+#ifdef STM32U5
+  const image_header *old_hdr = read_image_header(
+      (const uint8_t *)BOOTLOADER_START, BOOTLOADER_IMAGE_MAGIC,
+      flash_area_get_size(&BOOTLOADER_AREA));
+  secret_ensure_initialized();
```
matejcik, line 283: istm a function "secret - ensure initialized" should be called once as part of the init sequence, and then not touched again (and get/fill could then assume that someone else had already called it)

is there something i'm missing? the old discussion refered to "still halting after reconfiguring the sector", i don't see that now?

matejcik:
...speaking of, while we're at this, there is TOCTOU between `check_sd` and `copy_sd` (reported in the Cure53 pentest back in December 22) so perhaps we could do something about it? probably wrap the "if check() then copy()" in a single function and keep the SD slot powered the whole time, that's gonna make any attacker have a hard time

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/5bbddd37420c278a03dd8f333913b36b26eeb49d

(the header does not contain any secret info)

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/5bbddd37420c278a03dd8f333913b36b26eeb49d

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/5bbddd37420c278a03dd8f333913b36b26eeb49d

core/embed/boardloader/main.c
TychoVrahe: This actually does not do the check, just ensuring that up-to date version if written before doing the SD card update. Which i think i needed for some unlikely scenarios the stored version gets deleted, or if for example firmware updates bootloader, trezor reboots with SD card containing old bootloader plugged in - which would otherwise result in downgrading the bootloader again.

Otherwise, divided the function in https://github.com/trezor/trezor-firmware/commit/e9f84176037ef83e7fefee99e2c20586e8dd6926 , also the check in SD update is added. `Get` function does not do the halt immediately, leaves it to the caller.

core/embed/boardloader/main.c
TychoVrahe: That referred to the check done in bootloader in `secret_bhk_provision`, which is now moved to bootloader's main: https://github.com/trezor/trezor-firmware/commit/e9f84176037ef83e7fefee99e2c20586e8dd6926

so the "secret - ensure initialized" is called in both boardloader main and bootloader main, which is kinda redundant but should not break anything.

TychoVrahe:
@matejcik ad TOCTOU: given that we have enough RAM to fit the whole bootloader image, i think a straightforward solution is to load whole image to buffer, check integrity and flash it from the buffer. This also allows to adding a checking the bootloader hash efore flashing it. https://github.com/trezor/trezor-firmware/commit/70144e94c4c53865f5986a81bd4dd52c480c98d7

core/embed/boardloader/main.c
```diff
@@ -300,7 +322,9 @@ int main(void) {
          "invalid bootloader hash");
 
 #ifdef STM32U5
-  check_bootloader_version(hdr->monotonic);
+  uint8_t bld_version = get_bootloader_min_version();
+  ensure((bld_version >= hdr->monotonic) * sectrue, "BOOTLOADER DOWNGRADED");
```
matejcik, line 326: is this check backwards?

`bld_version` is the version from storage, `hdr->version` is the current one.
the current version must be _higher_ than the stored one, otherwise it's a downgrade, right?

core/embed/boardloader/main.c
```diff
@@ -160,6 +162,18 @@ static uint32_t check_sdcard(void) {
       return 0;
     }
 
+    if (sectrue != (check_single_hash(
```
matejcik, line 165: we should static_assert something like `IMAGE_CHUNK_SIZE >= BOOTLOADER_IMAGE_MAXSIZE` so that we can be sure that the bootloader is covered by exactly one hash, right?

(also technically we should check that all the other hashes are zero. could we reuse `check_image_contents`?

core/embed/boardloader/main.c
```diff
@@ -280,6 +285,23 @@ int main(void) {
 #if defined USE_SD_CARD
   sdcard_init();
 
+#ifdef STM32U5
+  secret_ensure_initialized();
```
matejcik, line 289: This chunk is probably incorrectly nested. You initialize the secret only in case both `USE_SD_CARD` and `STM32U5` are true. This will not be the case on T3B1.

core/embed/boardloader/main.c
```diff
@@ -280,6 +285,23 @@ int main(void) {
...
+
+  const image_header *old_hdr = read_image_header(
```
matejcik, line 291: please add a comment here that this is intro for the sdcard bootloader install, and why

and below, after the "check/copy" sequence, that we're (possibly) reading the bootloader again and this time it's the changed one

matejcik (changes requested):

matejcik (changes requested):
found some problems

core/embed/boardloader/main.c
TychoVrahe: right. fixed https://github.com/trezor/trezor-firmware/commit/b4a3ad09e9a2c1306683c2a07606eea491875c6d

core/embed/boardloader/main.c
TychoVrahe: added check of other hashes and static assert: https://github.com/trezor/trezor-firmware/commit/b4a3ad09e9a2c1306683c2a07606eea491875c6d

core/embed/boardloader/main.c
TychoVrahe: moved https://github.com/trezor/trezor-firmware/commit/c8d2a20f2e4b9e9415b74a69eb4ae0d9c22ec207

core/embed/boardloader/main.c
TychoVrahe: added https://github.com/trezor/trezor-firmware/commit/c8d2a20f2e4b9e9415b74a69eb4ae0d9c22ec207

matejcik (approved):
looking good now

to be extra sure we should manually test at least some of the features
1. monotonic downgrade protection
2. bootloader with other hash chunks filled out
3. updating from sd card
4. downgrading from sd card
5. downgrading in that special case where the upgrade was already installed but did not run

TychoVrahe:
i have done the manual testing with these steps. no unexpected behavior:

Bootloader downgrade protection:
- compile bld monotonic version 2 (B1) , flash it
- compile bld monotonic version 1 (B2), embed it into QA firmware (F2), upload it
- check that bootloader was not changed
- flash bootloader B2 with debugger
- check that boardloader will detect the downgrade
- flash bootloader B1, check it runs

Booloader image check:
- compile bootloader v mono version 2, (B3), manually edit the hashes in header
- embed it into QA firmware (F3)
- run the fw, restart, boardloader should complain

SD card update:
- provide SD card with B1 bootloader
- check that the bootloader runs

SD card downgrade:
- provide SD card with B2 bootloader
- check that the bootloader is not changed
- provide SD card with B3 bootloader
- check that the bootloader is not changed

Check special case downgrade:
- erase all flash (to reset monotonic counter)
- flash boardloader + B2 bootloader
- provide SD card with B2 bootloader, put it in
- prepare FW with embedded B1 bootloader, (F1)
- upload F1
- reset, bootloader should not be downgraded and B1 bootloader should run


Notes: 
- firmware will NOT refuse to install bootloader with additional hashes, image contents are not checked
- boardloader does not report in any way why the bootloader on SD is not being installed, just continues to bootloader

PR #3556, opened 2024-02-26
use flash bursts for faster flashing

TychoVrahe:
On STM32U5  faster flashing is possible with flash burst usage.


cepetr (approved):
Looking good to me. It seems the changes does not break anything.

PR #3557, opened 2024-02-27
Remove misc crypto functions from btc-only firmware

### LINKED ISSUE #3442: Remove misc cryptographic functions from Bitcoin-only firmware
The following functions should only be in universal firmware:
* `SignIdentity`
* `GetECDHSessionKey`

the following function should be dropped from all firmwares:
* `CosiCommit`

in terms of code organization, all those and more live in `apps/misc`. It seems we should keep `apps/misc` with things that work in Bitcoin-only, and maybe move `SignIdentity` and `GetEcdhSessionKey` to something like `apps/agent` (because those are mostly for trezor-agent)?

obrusvit:
Resolves #3442


obrusvit (commented):
Additional considerations:
1. `trezorlib/misc.py` calls these messages but the functions are not exposed from `cli`, not sure if this is an omission or it's supposed to be that way.
2. we have `confirm_sign_idenity` in `trezor/ui/layouts/{tr/tt}` modules which stays in btc-only firmware. Simplifying it to a single `app/agent/layout.py` seems to produce worse results (the text is split to many screens).

PR #3561, opened 2024-02-28
Improve language tests

matejcik:
removed a bunch of tests that don't make sense

added more granular tests for silent installation (still more will be needed though, pending the ability to install an invalid or outdated blob via debuglink)

added test for extended data in the header


matejcik:
f130892cc4451f0eb941f1eb2a7a0fe4e8b1000c should prevent one of the problems from #3567 going forward

grdddj:
`test_error_version_mismatch` fails as it is not forwarding the `data["header"]["version"] = "3.5.4"` value into `prepare_blob` - it is being replaced by `client.version` ... maybe add `force_version: tuple | None = None` argument into the `build_and_sign_blob`

I see the test for checking invalid signature got removed --- was it not important?

matejcik:
> I see the test for checking invalid signature got removed --- was it not important?

i thought that it's testing the same thing as "invalid data hash", but it's not.
updated and restored in 4483f9dc1ece69ca600a6cfae9345dbad5893ffe which btw is no longer a fixup, i'll need to remember to rename it to a stand-alone commit in the rebase


> `test_error_version_mismatch` fails as it is not forwarding the `data["header"]["version"] = "3.5.4"` value into `prepare_blob`

fixed in 4f24c32629d2237cecc6d4a95aa1237fde9a3fa6

grdddj (approved):
Thanks for the fixes, LGTM if the CI will be alright

PR #3565, opened 2024-02-28
Flash area API for Python, test speedup

matejcik:
This PR implements API for direct access to flash areas from MicroPython. This is then exposed over debuglink to allow us to mess with installed firmware and test inconsistency handling.

The second part of the PR uses this capability to check that the tested Trezor is in desired state. Storage will only be wiped when necessary. This should significantly speed up hardware tests.

tagging:
@mmilata for overall review
@andrewkozlik for specific code comments
@TychoVrahe for review of the flash area interface, i.e., most of the code in `modtrezorio-flash.h`


core/embed/extmod/modtrezorutils/modtrezorutils.c
```diff
@@ -145,68 +135,6 @@ STATIC mp_obj_t mod_trezorutils_halt(size_t n_args, const mp_obj_t *args) {
...
-  if (blake2s_Final(&ctx, vstr.buf, vstr.len) != 0) {
-    vstr_clear(&vstr);
-    mp_raise_msg(&mp_type_RuntimeError, "Failed to finalize firmware hash.");
-  }
-
-  return mp_obj_new_str_from_vstr(&mp_type_bytes, &vstr);
-}
-STATIC MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mod_trezorutils_firmware_hash_obj, 0,
-                                           2, mod_trezorutils_firmware_hash);
-
```
matejcik, line 209: @andrewkozlik this code was moved...

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,302 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+
+  vstr_t vstr = {0};
+  vstr_init_len(&vstr, BLAKE2S_DIGEST_LENGTH);
+  if (blake2s_Final(&ctx, vstr.buf, vstr.len) != 0) {
+    vstr_clear(&vstr);
+    mp_raise_msg(&mp_type_RuntimeError, "Failed to finalize hash.");
+  }
+
+  return mp_obj_new_str_from_vstr(&mp_type_bytes, &vstr);
+}
```
matejcik, line 232: @andrewkozlik  ...to here. Please double-check that I didn't break something.

core/src/apps/misc/get_firmware_hash.py
matejcik, line 1: @andrewkozlik please review this change in particular.

one interesting point to note, with this API it would now be possible to request hashing of a specific firmware chunk, which would make the whole operation even faster, to the point of imperceptibility by the user.
(otoh if we did that, it might open the window for an attacker to essentially read out the firmware kilobyte by kilobyte, which might be a problem if some boundary check somewhere got messed up)

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,302 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+    {MP_ROM_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_flash_area)},
+    MP_ROM_FLASH_AREA(BOARDLOADER),
+    MP_ROM_FLASH_AREA(BOOTLOADER),
+    MP_ROM_FLASH_AREA(FIRMWARE),
+    MP_ROM_FLASH_AREA(STORAGE_A),
+    MP_ROM_FLASH_AREA(STORAGE_B),
+    MP_ROM_FLASH_AREA(TRANSLATIONS),
+#if USE_OPTIGA && !defined(TREZOR_EMULATOR)
+    MP_ROM_FLASH_AREA(SECRET),
+#endif
```
matejcik, line 415: TODO: in production mode, we should also not create objects for storage / secret

core/src/apps/misc/get_firmware_hash.py
TychoVrahe: if speed is a concern here, lets reconsider using sha256 on U5 models here too, so we can use hw accelerator.

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,302 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+
+  uint32_t blocks = data.len / FLASH_BLOCK_SIZE;
+  const flash_block_t *data_as_blocks = (const flash_block_t *)data.buf;
+
+  ensure(flash_unlock_write(), NULL);
+  for (int i = 0; i < blocks; i++) {
+    if (sectrue != flash_area_write_block(self->area,
+                                          offset + i * FLASH_BLOCK_SIZE,
+                                          data_as_blocks[i])) {
+      mp_raise_ValueError("Write failed.");
```
TychoVrahe, line 308: this returns early via nlr_jump, so we should `flash_lock_write` before raising.

core/embed/extmod/modtrezorio/modtrezorio-flash.h
TychoVrahe: on U5 models, SECRET area is inaccessible to FW even in debug, so maybe that shouldn't be created there regardless on `PRODUCTION`

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,302 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+}
+
+STATIC MP_DEFINE_CONST_FUN_OBJ_3(mod_trezorio_FlashArea_read_obj,
+                                 mod_trezorio_FlashArea_read);
+
+/// if __debug__:
+///     def write(self, offset: int, data: bytes) -> None:
+///         """
+///         Writes data to flash area.
+///         Offset and written data size must be a multiple of quadword (16
```
TychoVrahe, line 280: quadword reference obsolete, FLASH_BLOCK_SIZE is dependent on MCU.

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,302 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+STATIC mp_obj_t mod_trezorio_FlashArea_write(mp_obj_t obj_self,
+                                             mp_obj_t obj_offset,
+                                             mp_obj_t obj_data) {
+  mp_obj_FlashArea_t *self = (mp_obj_FlashArea_t *)MP_OBJ_TO_PTR(obj_self);
+  uint32_t offset = trezor_obj_get_uint(obj_offset);
+  mp_buffer_info_t data = {0};
+  mp_get_buffer_raise(obj_data, &data, MP_BUFFER_READ);
+
+  if (data.len % FLASH_BLOCK_SIZE != 0) {
+    mp_raise_ValueError("Write size must be a multiple of 16.");
```
TychoVrahe, line 292: 16 is wrong on F4

TychoVrahe (changes requested):
reviewed `modtrezorio-flash.h` only

core/src/apps/misc/get_firmware_hash.py
andrewkozlik: > request hashing of a specific firmware chunk, which would make the whole operation even faster

Why would we want to hash only a specific chunk? We have to hash the entire firmware, so doing it by chunks seems inefficient.

> read out the firmware

Reading out the firmware is not a concern, so since you mention boundary checks, I assume you are concerned about reading out the norcow storage. That might be a concern, however, since the code requires offset and length alignment to 1024 bytes, you can't even read out the firmware.

core/embed/extmod/modtrezorio/modtrezorio-flash.h
andrewkozlik: Bug: `offset` is ignored. The hash is computed from the start of the area.

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,302 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+    if (blake2s_InitKey(&ctx, BLAKE2S_DIGEST_LENGTH, challenge.buf,
+                        challenge.len) != 0) {
+      mp_raise_msg(&mp_type_ValueError, "Invalid challenge.");
+    }
+  } else {
+    blake2s_Init(&ctx, BLAKE2S_DIGEST_LENGTH);
+  }
+
+  uint32_t area_size = flash_area_get_size(self->area);
+  if (area_size - offset < length) {
```
andrewkozlik, line 204: Since we are dealing with unsigned values, I don't think this will detect `offset > area_size`, because the subtraction will wrap to a large value.

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,302 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+///     Area of the flash memory
+///     """
+
+typedef struct _mp_obj_FlashArea_t {
+  mp_obj_base_t base;
+  const flash_area_t *area;
+} mp_obj_FlashArea_t;
+
+#define FLASH_READ_CHUNK_SIZE 1024
+#define PROGRESS_CHUNK_SIZE ((1024 / FLASH_READ_CHUNK_SIZE) * 16)
```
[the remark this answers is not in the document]
matejcik: ddab6396604da4b6f062e5fc8630f6fe1bd59767

core/embed/extmod/modtrezorio/modtrezorio-flash.h
matejcik: c8db36e3fa40721edad402982a1c3f262d2b30b6

core/embed/extmod/modtrezorio/modtrezorio-flash.h
matejcik: c8db36e3fa40721edad402982a1c3f262d2b30b6

core/embed/extmod/modtrezorio/modtrezorio-flash.h
matejcik: c8db36e3fa40721edad402982a1c3f262d2b30b6

core/embed/extmod/modtrezorio/modtrezorio-flash.h
matejcik: e442983a3ca133f508964302effb3c9c2f209f85

core/embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -118,3 +123,303 @@ STATIC const mp_obj_type_t mod_trezorio_FlashOTP_type = {
...
+      mp_raise_msg(&mp_type_RuntimeError, "Failed to read flash.");
+    }
+    blake2s_Update(&ctx, data, FLASH_READ_CHUNK_SIZE);
+    if (i % CHUNKS_PER_PROGRESS_STEP ==
+        start_chunk % CHUNKS_PER_PROGRESS_STEP) {
+      ui_progress(ui_wait_callback, i * FLASH_READ_CHUNK_SIZE);
+    }
+  }
+
+  ui_progress(ui_wait_callback, length);
```
andrewkozlik, line 224: One or two of the three `ui_progress()` calls here is wrong. Now that we introduced the `offset`, it's not clear what `ui_wait_callback()`, in this case `report(byte: int)`, expects the parameter to be. Either the range of the parameter is `[offset, offset + length]` or it is `[0, length]`.

matejcik: fixed (hopefully) in d718de0ab8d46813c281e4aea7c071a7fae9fd6b
this whole piece of code is rather awkward now, but the only way around that I see is making the progress from `[offset : offset + length]` and that seems like an ugly API

matejcik:
the code here has drifted too much and we achieved a speedup via different means (#4375) so i'm going to close this

Issue #2509, opened 2022-09-12
Add screenshot diff to failed device tests in CI


TychoVrahe:
If device test fails in CI due to screens being different, it would be nice to highlight in report where the screens differ, since sometimes the differences might be too small for human eye. 

[ImageMagick](https://imagemagick.org/index.php) compare tool is one option, see example output in this comment https://github.com/trezor/trezor-firmware/pull/2471#issuecomment-1243346467

TychoVrahe:
implemented in https://github.com/trezor/trezor-firmware/pull/3569

PR #3569, opened 2024-02-29
feat(tests): show screenshot diff

cepetr:
This PR adds the third column to the table for comparing screenshots captured in UI tests 

resoves #2509 

The third column show the difference between the recorded and the current screenshot.

Screenshot from 2024-02-29 09-45-12


Hannsek:
The red pixels are the differing ones? How does those two screenshots differ?

cepetr:
> The red pixels are the differing ones? How does those two screenshots differ?

Yes, now the red pixels show the difference. We can improve it somehow in the future.

PR #3570, opened 2024-02-29
chore: upgrade style checkers

matejcik:
black 24.2.0
flake8 7.0.0

some neat improvements for if-else statements there

VSCode black plugin already has the updated version so without this PR auto-formatting will mess up your style check


mmilata (approved):
LGTM & I think the poetry.lock changes might clear up some issues I ran into in #3412.

PR #3571, opened 2024-02-29
Method signatures for Rust layout objects

### LINKED ISSUE #2672: RustLayout class and its methods lack signatures in the mock file
`LayoutObj::obj_type` should have doc comments and their output should go to `trezorui2.pyi`

matejcik:
some preliminary work for the global layout PR.
fixes #2672
also converts most functions in the layout files to `def -> Awaitable` format (instead of `async def -> T` with a single "return await").

### DIFF
--- a/core/embed/rust/src/ui/model_tr/layout.rs
+++ b/core/embed/rust/src/ui/model_tr/layout.rs
@@ -1708,13 +1708,13 @@ extern "C" fn new_show_wait_text(message: Obj) -> Obj {
 pub static mp_module_trezorui2: Module = obj_module! {
     Qstr::MP_QSTR___name__ => Qstr::MP_QSTR_trezorui2.to_obj(),
 
-    /// CONFIRMED: object
+    /// CONFIRMED: UiResult
     Qstr::MP_QSTR_CONFIRMED => CONFIRMED.as_obj(),
 
-    /// CANCELLED: object
+    /// CANCELLED: UiResult
     Qstr::MP_QSTR_CANCELLED => CANCELLED.as_obj(),
 
-    /// INFO: object
+    /// INFO: UiResult
     Qstr::MP_QSTR_INFO => INFO.as_obj(),
 
     /// def disable_animation(disable: bool) -> None:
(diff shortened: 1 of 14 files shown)


matejcik:
> What is the benefit of this change?

one less allocation per called function.
every async function instantiates a generator object which is then awaited. a plain function just gets called without this overhead.

core/src/trezor/wire/context.py
```diff
@@ -159,7 +161,7 @@ async def write(self, msg: protobuf.MessageType) -> None:
 CURRENT_CONTEXT: Context | None = None
 
 
-def wait(*tasks: Awaitable) -> Any:
+def wait(task: Awaitable[T]) -> Awaitable[T]:
     """
     Wait until one of the passed tasks finishes, and return the result, while servicing
```
mmilata, line 166: Docstring still refers to multiple tasks.

mmilata (approved):
Very nice, the upy<->rs interface was a big untyped mess ...

core/src/trezor/wire/context.py
matejcik: 9c630f56d00161e7d5209036428d5228006c87b2

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -2054,21 +2054,22 @@ pub static mp_module_trezorui2: Module = obj_module! {
...
-    /// ) -> object:
+    /// ) -> LayoutObj[UiResult]:
     ///    """Device recovery homescreen."""
     Qstr::MP_QSTR_confirm_recovery => obj_fn_kw!(0, new_confirm_recovery).as_obj(),
 
     /// def select_word_count(
     ///     *,
     ///     dry_run: bool,  # unused on TR
     /// ) -> int | str:  # TR returns str
+    /// ) -> LayoutObj[int | str]:
```
grdddj, line 2065: Seems we should delete the line above to compensate for the new line

matejcik: good catch, and how tf did this not completely break typechecking?

matejcik: b9e1bc8fa4e1ffbe1ce5c7142fd3ec417a86a7f0

PR #3578, opened 2024-03-04
U5: place sensitive variables to confidential section

TychoVrahe:
The variables are placed into SRAM2, which is tamper and ECC protected.

Done regardless on initialization, uninitialized variables and zero initialized variables for take additional flash space. This can be mitigated in future by introducing something like `confidential_noinit` section and use that for such variales.


matejcik (approved):
LGTM, I could not think of any specific confidential things that i'd want to put in there.

PR #3579, opened 2024-03-04
allow T3T1 to start without touch during testing

TychoVrahe:
Correct touch initialization is not enforced until unit variant is written, which is done later in production when display+touch is already present.

### DIFF
--- a/core/embed/bootloader/main.c
+++ b/core/embed/bootloader/main.c
@@ -419,16 +419,28 @@ int bootloader_main(void) {
   i2c_init();
 #endif
 
-#ifdef USE_TOUCH
-  touch_power_on();
-  touch_init();
-#endif
+  display_reinit();
 
 #ifdef USE_DMA2D
   dma2d_init();
 #endif
 
-  display_reinit();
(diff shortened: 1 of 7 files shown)


core/embed/bootloader/main.c
```diff
@@ -419,16 +419,28 @@ int bootloader_main(void) {
...
 #endif
 
-  display_reinit();
+  unit_variant_init();
+
+#ifdef USE_TOUCH
+  touch_power_on();
+#ifdef TREZOR_MODEL_T3T1
+  // on T3T1, tester needs to run without touch, so making an exception
+  // util unit variant is written in OTP
```
cepetr, line 434: util -> until

cepetr (commented):
It looks good to me.

cepetr (approved):

core/embed/bootloader/main.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/c61fcc571e10d48a7a7433395cf26b8133d8b40c

PR #3581, opened 2024-03-05
use U5 HW keys for additional storage encryption

TychoVrahe:
Improves security of u5 based models by including U5 HW keys (DHUK and BHK) in storage encryption and decryption procedures.

### DIFF
--- a/core/embed/prodtest/optiga_prodtest.c
+++ b/core/embed/prodtest/optiga_prodtest.c
@@ -33,12 +33,21 @@
 #include "secret.h"
 #include "sha2.h"
 
+#include TREZOR_BOARD
+
+#ifdef STM32U5
+#include "secure_aes.h"
+#endif
+
 typedef enum {
   OPTIGA_PAIRING_UNPAIRED = 0,
   OPTIGA_PAIRING_PAIRED,
   OPTIGA_PAIRING_ERR_RNG,
-  OPTIGA_PAIRING_ERR_READ,
-  OPTIGA_PAIRING_ERR_HANDSHAKE,
+  OPTIGA_PAIRING_ERR_READ_FLASH,
+  OPTIGA_PAIRING_ERR_WRITE_FLASH,
+  OPTIGA_PAIRING_ERR_WRITE_OPTIGA,
+  OPTIGA_PAIRING_ERR_HANDSHAKE1,
+  OPTIGA_PAIRING_ERR_HANDSHAKE2,
 } optiga_pairing;
 
 static optiga_pairing optiga_pairing_state = OPTIGA_PAIRING_UNPAIRED;
@@ -72,11 +81,20 @@ static bool optiga_paired(void) {
     case OPTIGA_PAIRING_ERR_RNG:
       details = "optiga_get_random error";
       break;
-    case OPTIGA_PAIRING_ERR_READ:
-      details = "failed to read pairing secret";
+    case OPTIGA_PAIRING_ERR_READ_FLASH:
+      details = "failed to read pairing secret from flash";
       break;
-    case OPTIGA_PAIRING_ERR_HANDSHAKE:
-      details = "optiga_sec_chan_handshake";
+    case OPTIGA_PAIRING_ERR_WRITE_FLASH:
+      details = "failed to write pairing secret to flash";
+      break;
+    case OPTIGA_PAIRING_ERR_WRITE_OPTIGA:
+      details = "failed to write pairing secret to Optiga";
+      break;
+    case OPTIGA_PAIRING_ERR_HANDSHAKE1:
+      details = "failed optiga_sec_chan_handshake 1";
+      break;
+    case OPTIGA_PAIRING_ERR_HANDSHAKE2:
+      details = "failed optiga_sec_chan_handshake 2";
       break;
     default:
       break;
@@ -122,50 +140,57 @@ static bool set_metadata(uint16_t oid, const optiga_metadata *metadata) {
 }
 
 void pair_optiga(void) {
-  // The pairing key may already be written and locked. The success of the
-  // pairing procedure is determined by optiga_sec_chan_handshake(). Therefore
-  // it is OK for some of the intermediate operations to fail.
-
   uint8_t secret[SECRET_OPTIGA_KEY_LEN] = {0};
-  optiga_result ret = OPTIGA_SUCCESS;
 
-  if (secret_optiga_extract(secret) != sectrue) {
+  if (secret_optiga_get(secret) != sectrue) {
+    // Generate the pairing secret.
+    if (OPTIGA_SUCCESS != optiga_get_random(secret, sizeof(secret))) {
+      optiga_pairing_state = OPTIGA_PAIRING_ERR_RNG;
+      return;
+    }
+    random_xor(secret, sizeof(secret));
+
     // Enable writing the pairing secret to OPTIGA.
     optiga_metadata metadata = {0};
     metadata.change = OPTIGA_META_ACCESS_ALWAYS;
     metadata.execute = OPTIGA_META_ACCESS_ALWAYS;
     metadata.data_type = TYPE_PTFBIND;
-    set_metadata(OID_KEY_PAIRING, &metadata);  // Ignore result.
+    (void)set_metadata(OID_KEY_PAIRING, &metadata);
 
-    // Generate pairing secret.
-    ret = optiga_get_random(secret, sizeof(secret));
-    if (OPTIGA_SUCCESS != ret) {
-      optiga_pairing_state = OPTIGA_PAIRING_ERR_RNG;
+    // Store the pairing secret in OPTIGA.
+    if (OPTIGA_SUCCESS != optiga_set_data_object(OID_KEY_PAIRING, false, secret,
+                                                 sizeof(secret))) {
+      optiga_pairing_state = OPTIGA_PAIRING_ERR_WRITE_OPTIGA;
       return;
     }
 
-    // Store pairing secret.
-    ret =
-        optiga_set_data_object(OID_KEY_PAIRING, false, secret, sizeof(secret));
-    if (OPTIGA_SUCCESS == ret) {
-      secret_erase();
-      secret_write_header();
-      secret_write(secret, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
+    // Execute the handshake to verify that the secret was stored correctly in
+    // Optiga.
+    if (OPTIGA_SUCCESS != optiga_sec_chan_handshake(secret, sizeof(secret))) {
+      optiga_pairing_state = OPTIGA_PAIRING_ERR_HANDSHAKE1;
+      return;
     }
 
-    // Verify whether the secret was stored correctly in flash and OPTIGA.
+    // Store the pairing secret in the flash memory.
+    if (sectrue != secret_optiga_set(secret)) {
+      optiga_pairing_state = OPTIGA_PAIRING_ERR_WRITE_FLASH;
+      return;
+    }
+
+    // Reload the pairing secret from the flash memory.
     memzero(secret, sizeof(secret));
-    if (secret_read(secret, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN) !=
-        sectrue) {
-      optiga_pairing_state = OPTIGA_PAIRING_ERR_READ;
+    if (sectrue != secret_optiga_get(secret)) {
+      optiga_pairing_state = OPTIGA_PAIRING_ERR_READ_FLASH;
       return;
     }
   }
 
-  ret = optiga_sec_chan_handshake(secret, sizeof(secret));
+  // Execute the handshake to verify that the secret is stored correctly in both
+  // Optiga and MCU flash.
+  optiga_result ret = optiga_sec_chan_handshake(secret, sizeof(secret));
   memzero(secret, sizeof(secret));
   if (OPTIGA_SUCCESS != ret) {
-    optiga_pairing_state = OPTIGA_PAIRING_ERR_HANDSHAKE;
+    optiga_pairing_state = OPTIGA_PAIRING_ERR_HANDSHAKE2;
     return;
   }
 
--- a/core/embed/trezorhal/secret.h
+++ b/core/embed/trezorhal/secret.h
@@ -15,7 +15,7 @@
 
 secbool secret_bootloader_locked(void);
 
-void secret_write(uint8_t* data, uint32_t offset, uint32_t len);
+void secret_write(const uint8_t* data, uint32_t offset, uint32_t len);
 
 secbool secret_read(uint8_t* data, uint32_t offset, uint32_t len);
 
@@ -31,11 +31,13 @@ void secret_hide(void);
 
 void secret_write_header(void);
 
+secbool secret_optiga_set(const uint8_t secret[SECRET_OPTIGA_KEY_LEN]);
+
+secbool secret_optiga_get(uint8_t dest[SECRET_OPTIGA_KEY_LEN]);
+
 void secret_optiga_backup(void);
 
 void secret_optiga_hide(void);
-
-secbool secret_optiga_extract(uint8_t* dest);
 
 void secret_bhk_lock(void);
 
--- a/core/embed/trezorhal/secret.h
+++ b/core/embed/trezorhal/secret.h
@@ -13,38 +13,72 @@
 #define SECRET_BHK_OFFSET (1024 * 8)
 #define SECRET_BHK_LEN 32
 
+// Checks if bootloader is locked, that is the secret storage contains optiga
+// pairing secret on platforms where access to the secret storage cannot be
+// restricted for unofficial firmware
 secbool secret_bootloader_locked(void);
 
+// Writes data to the secret storage
 void secret_write(const uint8_t* data, uint32_t offset, uint32_t len);
 
+// Reads data from the secret storage
 secbool secret_read(uint8_t* data, uint32_t offset, uint32_t len);
 
+// Checks if the secret storage has been wiped
 secbool secret_wiped(void);
 
+// Verifies that the secret storage has correct header
 secbool secret_verify_header(void);
 
+// Checks that the secret storage is initialized and initializes it if not
 secbool secret_ensure_initialized(void);
 
+// Erases the entire secret storage
 void secret_erase(void);
 
+// Disables access to the secret storage until next reset
 void secret_hide(void);
 
+// Writes the secret header to the secret storage
 void secret_write_header(void);
 
+// Writes optiga pairing secret to the secret storage
+// Encrypts the secret if encryption is available on the platform
+// Returns true if the secret was written successfully
 secbool secret_optiga_set(const uint8_t secret[SECRET_OPTIGA_KEY_LEN]);
 
+// Reads optiga pairing secret
+// Decrypts the secret if encryption is available on the platform
+// Returns true if the secret was read successfully
+// Reading can fail if optiga is not paired, the pairing secret was not
+// provisioned to the firmware (by calling secret_optiga_backup), or the secret
+// was made unavailable by calling secret_optiga_hide
 secbool secret_optiga_get(uint8_t dest[SECRET_OPTIGA_KEY_LEN]);
 
+// Backs up the optiga pairing secret from the secret storage to the backup
+// register
 void secret_optiga_backup(void);
 
+// Deletes the optiga pairing secret from the register
 void secret_optiga_hide(void);
 
+// Locks the BHK register. Once locked, the BHK register can't be accessed by
+// the software. BHK is made available to the SAES peripheral
 void secret_bhk_lock(void);
 
+// Verifies that access to the register has been disabled
 secbool secret_bhk_locked(void);
 
+// Regenerates the BHK and writes it to the secret storage
 void secret_bhk_regenerate(void);
 
+// Provision the secret BHK from the secret storage to the BHK register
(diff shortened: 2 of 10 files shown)


storage/storage.c
```diff
@@ -581,7 +598,16 @@ static secbool __wur derive_kek_set(const uint8_t *pin, size_t pin_len,
   uint8_t optiga_secret[OPTIGA_PIN_SECRET_SIZE] = {0};
   uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE] = {0};
   stretch_pin_optiga(pin, pin_len, storage_salt, ext_salt, stretched_pin);
+#ifdef STM32U5
+  uint8_t stretched_pin_saes[OPTIGA_PIN_SECRET_SIZE] = {0};
+  ensure(secure_aes_ecb_decrypt_hw(stretched_pin, OPTIGA_PIN_SECRET_SIZE,
+                                   stretched_pin_saes, SECURE_AES_KEY_XORK),
+         "secure_aes pin stretch failed");
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/c1be37bc67ef3ebb7d173ef69ae26a3736d5248a

core/embed/trezorhal/stm32u5/secure_aes.c
```diff
@@ -104,11 +160,35 @@ secbool secure_aes_decrypt(uint32_t* input, size_t size, uint32_t* output) {
...
+  if ((size_t)input % sizeof(uint32_t) != 0 ||
+      (size_t)output % sizeof(uint32_t) != 0) {
+    size_t tmp_size = size;
+    while (tmp_size >= AES_BLOCK_SIZE) {
+      uint32_t input_buffer[AES_BLOCK_SIZE / sizeof(uint32_t)];
+      uint32_t output_buffer[AES_BLOCK_SIZE / sizeof(uint32_t)];
+      memcpy(input_buffer, input, AES_BLOCK_SIZE);
+      if (HAL_CRYP_Decrypt(&hcryp, input_buffer, AES_BLOCK_SIZE, output_buffer,
+                           HAL_MAX_DELAY) != HAL_OK) {
+        return secfalse;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/c1be37bc67ef3ebb7d173ef69ae26a3736d5248a

storage/storage.c
```diff
@@ -497,7 +502,19 @@ static void derive_kek(const uint8_t *pin, size_t pin_len,
...
+  uint8_t kek_saes[SHA256_DIGEST_LENGTH] = {0};
+  ensure(secure_aes_ecb_decrypt_hw(kek_out, SHA256_DIGEST_LENGTH, kek_saes,
+                                   SECURE_AES_KEY_XORK),
+         "secure_aes derive kek failed");
+  memcpy(kek, kek_saes, SHA256_DIGEST_LENGTH);
+  memzero(kek_saes, sizeof(kek_saes));
+#else
+  memcpy(kek, kek_out, SHA256_DIGEST_LENGTH);
+#endif
+  memzero(&kek_out, sizeof(kek_out));
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/c1be37bc67ef3ebb7d173ef69ae26a3736d5248a

TychoVrahe:
one more fix addressing repeated prodtest run - double decryption of pairing secret https://github.com/trezor/trezor-firmware/commit/8890ad8a03218af1995f790323d15f200abf7f44

core/embed/prodtest/optiga_prodtest.c
```diff
@@ -127,9 +133,10 @@ void pair_optiga(void) {
   // it is OK for some of the intermediate operations to fail.
 
   uint8_t secret[SECRET_OPTIGA_KEY_LEN] = {0};
+  uint8_t secret_dec[SECRET_OPTIGA_KEY_LEN] = {0};
```
andrewkozlik, line 136: The meaning of `secret` and `secret_dec` is identical. It's confusing to have two variables for the same thing. Let's just use `secret`.

core/embed/prodtest/optiga_prodtest.c
```diff
@@ -160,17 +175,24 @@ void pair_optiga(void) {
       optiga_pairing_state = OPTIGA_PAIRING_ERR_READ;
       return;
     }
+
+#ifdef STM32U5
+    secure_aes_ecb_decrypt_hw(secret, sizeof(secret), secret_dec,
+                              SECURE_AES_KEY_DHUK);
+#else
+    memcpy(secret_dec, secret, sizeof(secret_dec));
+#endif
```
andrewkozlik, line 184: Code duplication. Let's read the secret using `secret_optiga_get()`.

andrewkozlik (changes requested):

core/embed/prodtest/optiga_prodtest.c
```diff
@@ -127,9 +133,10 @@ void pair_optiga(void) {
   // it is OK for some of the intermediate operations to fail.
```
andrewkozlik, line 133: BTW, I see that this function was refactored in another PR. The idea behind the original implementation is that if `pair_optiga()` fails, e.g. due to communication issues with Optiga, or whatever reason, then calling this function again will be able to successfully recover the device. I am not sure that is true of the refactored version.

core/embed/prodtest/optiga_prodtest.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/7d6e4595ee680b60a373b82752d2835e6ae595ce

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -163,6 +164,9 @@ void secret_optiga_backup(void) {
 }
 
 secbool secret_optiga_extract(uint8_t *dest) {
+  uint8_t secret[SECRET_OPTIGA_KEY_LEN] = {0};
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/39ea9ce51b86d6f188f3fd1f9580abaf137e03b2

`memcpy`does not respect `volatile` keyword. while i believe it would work in this case, i would rather not do that.

core/embed/trezorhal/stm32u5/secure_aes.c
```diff
@@ -50,20 +54,40 @@ static void secure_aes_load_bhk(void) {
...
+      return 0;
+  }
+}
+
+secbool secure_aes_ecb_encrypt_hw(const uint8_t* input, size_t size,
+                                  uint8_t* output, secure_aes_keysel_t key) {
   CRYP_HandleTypeDef hcryp = {0};
   uint32_t iv[] = {0, 0, 0, 0};
 
+  if (size % 16 != 0) {
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1885a969ae4c6ab595200cf62c6ebcfc0d063ff1

core/embed/prodtest/optiga_prodtest.c
```diff
@@ -150,7 +157,15 @@ void pair_optiga(void) {
     if (OPTIGA_SUCCESS == ret) {
       secret_erase();
       secret_write_header();
+#ifdef STM32U5
+      uint8_t secret_enc[SECRET_OPTIGA_KEY_LEN] = {0};
+      secure_aes_ecb_encrypt_hw(secret, sizeof(secret), secret_enc,
+                                SECURE_AES_KEY_DHUK);
+      secret_write(secret_enc, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
+      memzero(secret_enc, sizeof(secret_enc));
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/3e52219453a68f9a6b3b35ef40029aa589a9f230

core/embed/prodtest/optiga_prodtest.c
```diff
@@ -150,7 +157,15 @@ void pair_optiga(void) {
     if (OPTIGA_SUCCESS == ret) {
       secret_erase();
       secret_write_header();
+#ifdef STM32U5
+      uint8_t secret_enc[SECRET_OPTIGA_KEY_LEN] = {0};
+      secure_aes_ecb_encrypt_hw(secret, sizeof(secret), secret_enc,
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/3e52219453a68f9a6b3b35ef40029aa589a9f230

core/embed/prodtest/optiga_prodtest.c
```diff
@@ -150,7 +157,15 @@ void pair_optiga(void) {
...
+                                SECURE_AES_KEY_DHUK);
+      secret_write(secret_enc, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
+      memzero(secret_enc, sizeof(secret_enc));
+#else
       secret_write(secret, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/3e52219453a68f9a6b3b35ef40029aa589a9f230

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -163,6 +164,9 @@ void secret_optiga_backup(void) {
...
+  uint8_t *secret_ptr = (uint8_t *)secret;
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/7d6e4595ee680b60a373b82752d2835e6ae595ce

core/embed/prodtest/optiga_prodtest.c
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/3e52219453a68f9a6b3b35ef40029aa589a9f230

core/embed/prodtest/optiga_prodtest.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/3e52219453a68f9a6b3b35ef40029aa589a9f230

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -173,11 +177,19 @@ secbool secret_optiga_extract(uint8_t *dest) {
...
     for (int j = 0; j < 4; j++) {
-      dest[i * 4 + j] = (val >> (j * 8)) & 0xFF;
+      secret_ptr[i * 4 + j] = (val >> (j * 8)) & 0xFF;
     }
   }
 
-  return all_zero ? secfalse : sectrue;
+  if (all_zero) {
+    return secfalse;
+  }
```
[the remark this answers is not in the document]
TychoVrahe: well these have different function

`secret_present` (`secret_optiga_present`) check whether the secret is written in flash, which used for decision whether to hide the Hide Protect area so that firmware/prodtest can't access it, modify it etc.

while 

`secret_optiga_extract` (now `secret_optiga_get`) look into the backup registers (on U5) to check whether the secret was provided by bootloader. Which will not be the case if the firmware is unofficial.

The empty values are different for backup registers and flash.

core/embed/trezorhal/secret.h
```diff
@@ -31,12 +31,14 @@ void secret_hide(void);
 
 void secret_write_header(void);
 
+secbool secret_optiga_set(uint8_t* secret);
```
prusnak, line 34: I assume that the set function does not modify the secret. We should use const in that case.

```suggestion
secbool secret_optiga_set(const uint8_t* secret);
```

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/86a5c1ac1f2ac8106e911db89b75ec029ffd874c

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -143,6 +143,17 @@ secbool secret_optiga_present(void) {
   return secret_present(SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
 }
 
+secbool secret_optiga_set(const uint8_t *secret) {
```
andrewkozlik, line 146: When passing fixed-size arrays, let's use array declarations rather than pointers in all new code, e.g. `secret_optiga_set(const uint8_t secret[SECRET_OPTIGA_KEY_LEN])`, here and in the header files as well. Same applies to `secret_optiga_get()`. This makes no difference in terms of the resulting compiled code. It is mainly a form of documentation, saving time of devs trying to figure out the expected array size. However, some compilers use it to issue warnings when a smaller array is supplied.

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/3bc37ecfbed67857165a82abd8e20f022ca05b12

core/embed/trezorhal/stm32u5/secret.c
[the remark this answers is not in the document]
andrewkozlik: I see. Let's document this in the header files, because I think this is not at all obvious from the function names. Namely:
- What are the preconditions for each function to be used:
    - What needs to be done before the function can be called (some other function needs to have written the value to the registers, for example).
    - When can the functions be meaningfully called (bootloader vs. official firmware vs. unofficial firmware)
- What does the return value of each of these functions mean.

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -143,6 +143,17 @@ secbool secret_optiga_present(void) {
...
+  uint8_t secret_enc[SECRET_OPTIGA_KEY_LEN] = {0};
+  if (sectrue != secure_aes_ecb_encrypt_hw(secret, sizeof(secret_enc),
+                                           secret_enc, SECURE_AES_KEY_DHUK)) {
+    return secfalse;
+  }
+  secret_write(secret_enc, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
```
andrewkozlik, line 152: Shouldn't we be calling  `secret_erase()` and `secret_write_header()` before `secret_write()` here?

TychoVrahe: I actually removed it intentionally, sorry for not mentioning it. The idea is that secret sector contains also monotonic counter for bootloader downgrade protection, which i would rather not delete. At this point, the optiga pairing secret should be empty and writable, so no need to erase the sector. OTOH boardloader will reinitialze the monotonic counter on next run so probably no security issue if this gets deleted here.

core/embed/trezorhal/stm32u5/secret.c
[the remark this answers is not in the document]
TychoVrahe: Added comments in https://github.com/trezor/trezor-firmware/commit/8d4fce5129e3172718e1a2cefe514d7cab5d5e97

Note: i will do a bit of refactoring of the secret.h api, to hide more platform differences inside, similarly to what you suggested with `secret_optiga_set` and `secret_optiga_get`. In order to simplify the secret handling here: https://github.com/trezor/trezor-firmware/blob/285b22e53e880fda171eb05ace67fa7a94b8e145/core/embed/bootloader/main.c#L328

I will do that in successive PR if you don't mind.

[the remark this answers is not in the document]
andrewkozlik: Also we should make the other function names more intuitive. In case of `secret_hide()` vs. `secret_optiga_hide()`, it should be more obvious that the second clears the register, whereas the first restricts access to the flash. That's totally different stuff. For example `secret_optiga_backup()` -> `secret_optiga_cache()` and `secret_optiga_hide` -> `secret_optiga_uncache()`, or `_load` vs. `_unload`. Same applies to `secret_bhk_provision()` -> `secret_bhk_cache()` or `secret_bhk_load()` depending on whatever `secret_optiga_backup()` will be renamed to.

core/embed/trezorhal/stm32u5/secret.c
andrewkozlik: The fact that the logic around `secret_write_header()` is completely different between F4 and U5 is really not good. In U5 it's done by the bootloader and also the boardloader in `secret_ensure_initialized()`? Why both? In F4 it's done by prodtest in `secret_optiga_set()`. This is confusing and difficult to review.

> boardloader will reinitialze the monotonic counter on next run

Why? I don't see that it will do that.

andrewkozlik: It seems to make sense to do this in the boardloader.
> but the F4 prodtest would still need to write the header too, to ensure compatibility with old boardloader

When we release a new version of prodtest, it will be released with a new boardloader and bundled in combined.bin. So I think we don't need to be concerned about that. I can't think of a scenario where we provisioned an F4 Trezor with an old combined.bin file, didn't run the Optiga pairing and then installed a newer prodtest version to run the Optiga pairing. Theoretically it can happen of course. On the other hand we will probably soon stop producing F4 devices, so if we are concerned about this scenario then, we can leave the code in prodtest temporarily until that moment comes.

TychoVrahe: We have relatively large amount of devices that didnt pass test, some of these require reinstalling prodtest to do the retest, with no option to update the boardloader. If some of those were provided with new prodtest, and didn't do the pairing, it would be a problem. 

Yes i agree that these should all have optiga already paired, but i am not sure we can be entirely sure this does not happen

TychoVrahe: Also. New boardloader, with old bootloader might be a problematic combination. Old bootloader, after downgrade via old firmware, will never see the optiga secret as wiped due to header being always present. This will prevent installing custom fw. Do we care though? Would need a bootloader monotonic version increment to prevent it.

core/embed/prodtest/optiga_prodtest.c
andrewkozlik: I think it is possible. Here is a suggested commit: https://github.com/trezor/trezor-firmware/pull/3581/commits/5c9a281519c26f0d63803af8bdda4ac18735d7a4 (untested)

This way if writing to Optiga fails then prodtest firmware will be able to recover from the situation by restarting.

andrewkozlik: I removed the second call to `secret_optiga_get()`, because I think it won't work, since the secret has not been loaded into the register.

andrewkozlik: Actually, perhaps simply calling `secret_optiga_backup()` in `secret_optiga_set()` will do the trick? See another suggested commit https://github.com/trezor/trezor-firmware/pull/3581/commits/fe5b4449a345e6db410582a37fa6c84632f6904c (untested). So this way prodtest will check the secret processing to the maximum extent, i.e. trying to read it after it has been stored.

andrewkozlik: Fixed typo, updated comment: https://github.com/trezor/trezor-firmware/pull/3581/commits/a58f45c10b401ef808463827bab18d4374c0030f

core/embed/trezorhal/stm32u5/secret.c
andrewkozlik: I think this is turning out to be out of scope of this PR. Let's discuss the problem IRL and resolve it in the other PR you mentioned elsewhere. For now I think we can remove the duplicate code from the bootloader.

core/embed/prodtest/optiga_prodtest.c
TychoVrahe: In case there is some low quality entropy stored in the flash, it will be used anyway. Is that a concern?

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: Removed: https://github.com/trezor/trezor-firmware/commit/c12fade159c80c2ddc6d81e5f15fe8d5d567cfdc

core/embed/prodtest/optiga_prodtest.c
andrewkozlik: I find it hard to see how that could happen, but it's worth a thought. After looking at the code one more time, I realize that your original implementation would work fine if the writing to Optiga failed. Not sure why I didn't see it before, but if it fails, then nothing will be written to the flash, so at the next reboot the procedure will try to set up the paring secret again. So that alleviates my main concern, since the writing to the flash is far less likely to fail than writing to Optiga. One problem that could happen however, is that Optiga claims success, but in reality the value is written incorrectly. That seems quite unlikely, since the communication is integrity-protected by a CRC. We can double check by executing the handshake twice.

Here is yet another fixup commit https://github.com/trezor/trezor-firmware/pull/3581/commits/9319170cea5ab77cf2afbb7fcfab9bfcd5843144, which essentially reverts what I did in terms of functionality, except that:
- it adds `random_xor()` which means thet entropy from the MCU's TRNG will be used in generating the pairing secret,
- it adds a pre-handshake, to check that the secret was store correctly in Optiga, before writing the secret to the flash,
- it maintains the change in `secret_optiga_set()`, ensuring that the backup register is loaded,
- it improves the error logging and comments.

What do you think?

andrewkozlik (approved):

core/embed/prodtest/optiga_prodtest.c
TychoVrahe: I think this is a reasonable solution, we can revisit this if we get abnormal failure rate. Tested on HW a it works fine.

PR #3586, opened 2024-03-07
docs: hint usage of trezor-user-env

obrusvit:
[no changelog]

This is a minor change in Emulator docs (and also in docs for emulator build) to guide most people to trezor-user-env. Based on Slack conversation.
2024-03-07_09-57_emu_docs

### DIFF
--- a/docs/core/build/emulator.md
+++ b/docs/core/build/emulator.md
@@ -1,5 +1,8 @@
 # Build instructions for Emulator (Unix port)
 
+> :bulb: **Hint**:
+Using emulator as described here is useful during firmware development. If you intend to use the emulator without modifying the firmware, you might be looking for [Trezor User Env](https://github.com/trezor/trezor-user-env/tree/master).
+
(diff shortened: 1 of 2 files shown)


matejcik (approved):

docs/core/build/emulator.md
```diff
@@ -1,5 +1,8 @@
 # Build instructions for Emulator (Unix port)
 
+> :bulb: **Hint**:
+Using emulator as described here is useful during firmware development. If you intend to use only released versions of firmware, you might be looking for [Trezor User Env](https://github.com/trezor/trezor-user-env/tree/master).
```
Hannsek, line 4: > If you intend to use only released versions of firmware
you can also use fw from a branch in trezor-user-env using a gitlab link, no? But maybe only internally?🤔

obrusvit: You're right. I reworded to 

`If you intend to use the emulator without modifying the firmware, you ....`

PR #3588, opened 2024-03-07
Fix check-bitcoinonly for translations

### LINKED ISSUE #3582: Remove altcoin strings from english translation blob in bitcoin-only firmware
I don't think they pose a security risk but:
* eventually somebody will complain
* we had to add `{}` and `{0}` as exceptions to `check-bitcoin-only` which makes the script less effective

matejcik:
disable "show whitespace" to see that there are no code changes in layouts, just an added indent

fixes #3582


mmilata (approved):
LGTM, will need a rebase on top of #3592 (`{}`, `{0}`) no longer needed.

PR #3590, opened 2024-03-07
add stm32u5 prepare binary

TychoVrahe:
Adds a build target that creates a small binary, which sets STM32U5 option bytes so that boardloader can properly boot.

~~Making this draft for the moment, we are waiting for updated sw for programmer that would allow setting the option bytes during production. This would eliminate the need for this.~~

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -11,6 +11,7 @@ MAKE = make -j $(JOBS)
 SCONS = scons -Q -j $(JOBS)
 
 BUILD_DIR             = build
+PREPARE_BUILD_DIR = $(BUILD_DIR)/prepare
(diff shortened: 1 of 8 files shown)


TychoVrahe:
Still no signs of better solution, marking this ready for review so that we have this ready to merge when needed.

cepetr (commented):
Looks good. Just the name of the binary doesn't sound well to me. What about `fuseflasher`, `fuseflash`, `fusesetter`, `optionsetter`, `fuseutility`...

TychoVrahe:
Closing: we will be programming the relevant option bytes via programmer.

Issue #2540, opened 2022-10-04
Fix `ticks_ms` arithmetic


andrewkozlik:
In several places we do arithmetic operations `+`, `-` and `<` with the values returned by `utime.ticks_ms()`. These values are not meant to be handled directly and doing so may produce incorrect results. Instead we should be using `utime.ticks_add()` and `utime.ticks_diff()`, see [documentation](https://github.com/trezor/micropython/blob/fce614cbe213e3f6c9e1a2a9634d2c1edae7c57b/docs/library/time.rst).

The places where we handle the values incorrectly are:
apps/common/request_pin.py
apps/homescreen/homescreen.py
apps/webauthn/fido2.py
trezor/ui/loader.py

naiyoma:
@andrewkozlik  I would like to work on this,  since the changes will involve multiple files, I'm curious to know if I should do this in a single PR or should each file change be a single PR?

matejcik:
@naiyoma make a single PR please

PR #3594, opened 2024-03-10
Fix utime operations

naiyoma:
This PR fixes issue #2540. The refactor applies to apps/webauthn/fido2.py and apps/common/request_pin.py. 
I couldn't locate the other two files (apps/homescreen/homescreen.py,  trezor/ui/loader.py)mentioned in the issue. They might have been renamed or deleted.

### DIFF
--- a/core/src/apps/webauthn/fido2.py
+++ b/core/src/apps/webauthn/fido2.py
@@ -985,7 +985,7 @@ class DialogManager:
 
     def reset_timeout(self) -> None:
         if self.state is not None:
-            self.deadline = utime.ticks_ms() + self.state.timeout_ms()
+            self.deadline = utime.ticks_add(utime.ticks_ms(), self.state.timeout_ms())
 
     def reset(self) -> None:
         if self.workflow is not None:
@@ -1002,7 +1002,7 @@ class DialogManager:
         )
 
     def is_busy(self) -> bool:
-        if utime.ticks_ms() >= self.deadline:
+        if utime.ticks_diff(utime.ticks_ms(), self.deadline) >= 0:
             self.reset()
 
         if not self._workflow_is_running():
@@ -1015,7 +1015,10 @@ class DialogManager:
         return True
 
     def set_state(self, state: State) -> bool:
-        if self.state == state and utime.ticks_ms() < self.deadline:
+        if (
+            self.state == state
+            and utime.ticks_diff(utime.ticks_ms(), self.deadline) < 0
+        ):
             self.reset_timeout()
             return True
 
@@ -1036,7 +1039,7 @@ class DialogManager:
         try:
             if not state:
                 return
-            while utime.ticks_ms() < self.deadline:
+            while utime.ticks_diff(utime.ticks_ms(), self.deadline) < 0:
                 if state.keepalive_status() != _KEEPALIVE_STATUS_NONE:
                     cmd = cmd_keepalive(state.cid, state.keepalive_status())
                     await send_cmd(cmd, self.iface)
--- a/core/src/apps/common/request_pin.py
+++ b/core/src/apps/common/request_pin.py
@@ -97,7 +97,7 @@ async def verify_user_pin(
     if (
         cache_time_ms
         and last_unlock
-        and utime.ticks_ms() - last_unlock <= cache_time_ms
+        and utime.ticks_diff(utime.ticks_ms(), last_unlock) <= cache_time_ms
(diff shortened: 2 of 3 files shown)


core/src/apps/webauthn/fido2.py
```diff
@@ -1036,7 +1036,7 @@ async def keepalive_loop(self) -> None:
         try:
             if not state:
                 return
-            while utime.ticks_ms() < self.deadline:
+            while utime.ticks_diff(utime.ticks_ms(), self.deadline) > 0:
```
prusnak, line 1039: Shouldn't this be

```suggestion
            while utime.ticks_diff(utime.ticks_ms(), self.deadline) < 0:
```
?


because:

`A < B` -> `A - B < 0`

andrewkozlik: Yeah, I agree with @prusnak, see https://docs.micropython.org/en/v1.15/library/utime.html#utime.ticks_diff.
It can also be written as:
```py
while utime.ticks_diff(self.deadline, utime.ticks_ms()) > 0
```
but I have no personal preference either way.

andrewkozlik:
> I couldn't locate the other two files (apps/homescreen/homescreen.py, trezor/ui/loader.py)mentioned in the issue. They might have been renamed or deleted.

I will try to find where the code was moved, but I can already say that `src/boot.py` needs attention.

andrewkozlik:
In `rezor/ui/loader.py` we had this code:
```py
class Loader(ui.Component):
...
    def start(self) -> None:
        if self.start_ms is not None and self.stop_ms is not None:
            self.start_ms = utime.ticks_ms() - self.elapsed_ms()  # <<< PROBLEM
```

In `apps/homescreen/homescreen.py` we had this code for "Hold to lock":
```py
    def dispatch(self, event: int, x: int, y: int) -> None:
        if (
            self.touch_ms is not None
            and self.touch_ms + _LOADER_DELAY_MS < utime.ticks_ms()  # <<< PROBLEM
        ):
```

I can't find where these two pieces of code were moved, but I guess they must have been fixed.

However, I found two more cases which need fixing in `apps/webauthn/fido2.py` on lines 1005 and 1018.

core/src/apps/webauthn/fido2.py
naiyoma: I also agree with the suggestions above

naiyoma:
> 997108d

@andrewkozlik Could you please clarify the intention behind initializing the welcome_screen_start_ms variable to the current time ? Is it intended for the difference calculated by utime.ticks_diff() to always be 0?

andrewkozlik:
> Could you please clarify the intention behind initializing the welcome_screen_start_ms variable to the current time ? Is it intended for the difference calculated by utime.ticks_diff() to always be 0?

@matejcik is in a better position to answer that, see https://github.com/trezor/trezor-firmware/commit/c5fc92e36af89501294d020b442979ad3f2a4013

matejcik:
> @andrewkozlik Could you please clarify the intention behind initializing the welcome_screen_start_ms variable to the current time ? Is it intended for the difference calculated by utime.ticks_diff() to always be 0?

how would that work? the run-time between `welcome_start_ms` and the call to `enforce_welcome_screen_duration()` is clearly not zero.
the intention is to measure time elapsed between the `welcome_start_ms` and the `enforce` call, making sure that we wait at least the specified 1200 ms before continuing

andrewkozlik (approved):
LGTM

Assuming that the device test pass, you will need to squash the fixup and add `[no changelog]` to each commit, see https://docs.trezor.io/trezor-firmware/misc/changelog.html#not-adding-changelog-entry, then force-push.

naiyoma:
@andrewkozlik I ran into an issue running the device tests.I followed the documentation(https://docs.trezor.io/trezor-firmware/tests/device-tests.html), but I'm getting this error 

`ImportError while loading conftest '/home/naiyoma/trezor-firmware/tests/conftest.py'.
../tests/conftest.py:36: in <module>
    from. import translations, ui_tests
../tests/translations.py:10: in <module>
    from. import common
../tests/common.py:307: in <module>
    debug: "DebugLink", br_code: messages.ButtonRequestType | None = None
E   TypeError: unsupported operand type(s) for |: 'EnumMeta' and 'NoneType'`


It seems there might be a configuration issue with the test suite itself, not necessarily related to my changes.

matejcik:
what is your Python version? at least 3.8 is required

PR #3604, opened 2024-03-14
introduce streaming access to storage items

TychoVrahe:
introducing access to slices of data from storage. 

as preparation for #2521 , where it is needed for accessing homescreen image in memory constraint environment.

### DIFF
--- a/storage/storage.c
+++ b/storage/storage.c
@@ -166,10 +166,8 @@ static secbool storage_upgrade_unlocked(const uint8_t *pin, size_t pin_len,
                                         const uint8_t *ext_salt);
 static secbool storage_set_encrypted(const uint16_t key, const void *val,
                                      const uint16_t len);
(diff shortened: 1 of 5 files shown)


prusnak:
Is homescreen stored as encrypted value?

If not, do we want to allow streaming for encrypted values and not only for unencrypted ones? It does not feel right to allow it for encrypted.

TychoVrahe:
Homescreen is indeed unencrypted, i wanted to keep the API universal but its not strictly needed at this point. 

just to be clear what this does: basically you can request just part of the stored data, at given offset with specified length. Encrypted items are decrypted at each separate request and just the requested part is copied into the output buffer. IMO this should be fine from security point of view, but it is some added complexity.

prusnak:
Let's not add features we have no use case for. Especially when it comes to the storage.

PR #3606, opened 2024-03-14
ci: actions: primitive icons for UI diff comments

mmilata:
Adds icons to the table with links to UI test results. Example: https://github.com/trezor/trezor-firmware/pull/3605#issuecomment-1997934587

Screenshot 2024-03-14 at 22-16-25 ci test by mmilata · Pull Request #3605 · trezor_trezor-firmware

The main limitation is that the icons are not updated automatically, you need to **reload the page** in order to see the current state. Other limitation is the broken link icon, which feels weird and doesn't allow you to distinguish queued job from job that did not run for other reasons. The icons are from model T UI and can be replaced with something nicer.


Hannsek:
We can use newer icons [Archive.zip](https://github.com/trezor/trezor-firmware/files/14612001/Archive.zip)

Let's discuss further on standup.

matejcik:
love the idea, let's try it

Hannsek (approved):
waiting for checks to finish

matejcik:
in the last comment, there is a broken icon but the test had passed

mmilata:
legacy had wrong link, fixed

PR #3608, opened 2024-03-14
refactor: extract common layout handling from bootloader

TychoVrahe:
introduces 'simplified' layout object for bootloader, common for all UI implementaitons.

originally part of #2521 , extracted into separate PR to simplify review and merging

### DIFF
--- a/core/embed/rust/src/ui/layout/simplified.rs
+++ b/core/embed/rust/src/ui/layout/simplified.rs
@@ -1,0 +1,129 @@
+#[cfg(feature = "button")]
+use crate::trezorhal::io::io_button_read;
+#[cfg(feature = "touch")]
+use crate::trezorhal::io::io_touch_read;
+#[cfg(feature = "button")]
+use crate::ui::event::ButtonEvent;
+#[cfg(feature = "touch")]
+use crate::ui::event::TouchEvent;
+use crate::ui::{
+    component::{Component, Event, EventCtx, Never},
+    constant::SCREEN,
+    display,
+};
+use num_traits::ToPrimitive;
+
+#[cfg(feature = "backlight")]
+use crate::ui::model::theme::{BACKLIGHT_DIM, BACKLIGHT_NORMAL};
+
+pub trait ReturnToC {
+    fn return_to_c(self) -> u32;
+}
+
+impl ReturnToC for Never {
+    fn return_to_c(self) -> u32 {
+        unreachable!()
+    }
+}
+
+impl<T> ReturnToC for T
+where
+    T: ToPrimitive,
+{
+    fn return_to_c(self) -> u32 {
+        self.to_u32().unwrap()
+    }
+}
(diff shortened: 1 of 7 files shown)


matejcik (approved):
utACK, looks reasonable

core/embed/rust/src/ui/mod.rs
```diff
@@ -12,10 +12,20 @@ pub mod screens;
...
+#[cfg(all(
+    feature = "model_t1",
+    not(feature = "model_tr"),
+    not(feature = "model_tt")
+))]
+pub use model_t1 as model;
+#[cfg(all(feature = "model_tr", not(feature = "model_tt")))]
+pub use model_tr as model;
+#[cfg(feature = "model_tt")]
+pub use model_tt as model;
```
matejcik, line 31: this is actually a really bad idea because the individual models don't expose even vaguely similar API. sorry I didn't catch it soojner.

core/embed/rust/src/ui/layout/simplified.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::ui::event::TouchEvent;
+use crate::ui::{
+    component::{Component, Event, EventCtx, Never},
+    constant::SCREEN,
+    display,
+};
+use num_traits::ToPrimitive;
+
+#[cfg(feature = "backlight")]
+use crate::ui::model::theme::{BACKLIGHT_DIM, BACKLIGHT_NORMAL};
```
matejcik, line 17: This breaks `make test_rust` on model R because `backlight` feature is enabled together with `model_tr`. We could reconfigure the features but relying on these constants is wrong in general -- what if on TS5 we want to fade to user-defined value instead of a built-in constant?

Short-term, we should remove this `use`, and refer to `crate::ui::model_tt::theme::XYZ` in the `fadein()` and `fadeout()` functions, probably conditional on both `backlight` _and_ `model_tt`.

Longer-term, I'm not sure yet. Something something `Model` trait that carries the constants and all these simplified functions, with implementations for each model?

core/embed/rust/src/ui/mod.rs
TychoVrahe: i understand and agree that components are a bad place to do this. What i need is a model independent (or maybe more precisely ui-type independent) way to call some specific screen, or in other words, we need an universal API for UI in rust.

I guess that means introducing a new layer.. ultimately i think this should be layouts, with (optimally single) micropython interface refactored out. But thats quite a bit of refactoring too

matejcik: how about this?

make a trait that has all you need (contents of `simplified.rs` basically, right?)
implement it for every model
make the common functions generic over an impl of the trait
expose all versions of the top-level function to C, with the different impls under different names
use C #defines to pick which one gets actually called

it's a new layer, yes, and we can start small with it and extend it to cover the micropython interface later

PR #3609, opened 2024-03-15
chore(core): fix whitespace in makefile

prusnak:
whitespace only change - see https://github.com/trezor/trezor-firmware/pull/3609/files?diff=unified&w=1 to confirm


matejcik (approved):
utACK

(github by default does a pretty good job of showing which exact parts of the line were changed)

TychoVrahe (approved):

Issue #3600, opened 2024-03-13
Unicode font uppercasing issues


grdddj:
turns out our MicroPython can't correctly uppercase unicode characters.
(no surprise really)
so essentially all calls to something.upper() in layouts are now useless and we need to make the uppercase part of the translations data

prusnak:
Aren't we just missing some compile time option that solves the issue?

obrusvit:
Related: https://github.com/trezor/trezor-firmware/issues/3224

PR #3610, opened 2024-03-15
Translate strings coming from storage

### LINKED ISSUE #3520: translations: translate "processing.." strings coming from storage
`storage.c` has a bunch of baked-in English strings for progress labels.
these need to be translated too.

(similar case may be for wipe code erasure and similar)

the way to go seems to be to return an enum instead of literal string, and let the UI code convert to a string.

this can be done after #3206 is merged, to not block it anymore

### LINKED ISSUE #3535: New string in translations requires clean build to work properly
It seems that adding new strings to the the firmware requires clean build after merging translations #3206.

Steps to reproduce:

1. Add new string to `core/translations/en.json`
2. run `python core/translations/cli.py gen`
3. run `python core/translations/order.py`
4. test a code utilizing the new string.

This results in `AttributeError` on my machine.
image

The problem might be in missing dependency of SCons and/or build.rs. Investigate and fix the issue to close this task.

matejcik:
fixes #3520 
fixes #3535


Hannsek:
Is this intended? intended

matejcik:
what is? @Hannsek

matejcik:
> Is this intended? change to the small letters.

it's a consequence of #3600 and properties of the progress layout.
I can revert it to what it was before, but that would also change the layout in TS3.

I would say this is not too important because it only happens when switching between two non-english languages?

core/SConscript.unix
```diff
@@ -833,6 +839,7 @@ rust = env.Command(
     source='',
     action=cargo_build(), )
 env.Depends(rust, protobuf_blobs)
+env.Depends(rust, TRANSLATION_DATA)
```
mmilata, line 842: Don't we need to add the same rule to `SConscript.firmware`?

mmilata (approved):
utACK, nice work unifying progress.py.

core/SConscript.unix
matejcik: it's unlikely to matter but yeah. added in rebase

matejcik:
rebased on top of #3646 

**DO NOT MERGE** until Wednesday

PR #3614, opened 2024-03-17
TStringify Paragraphs

matejcik:
This is a minimally-invasive change that uses TString as the internal storage for `Paragraph`, modifies `Paragraphs` and `ParagraphSource` accordingly. (in code, this looks mostly like replacing the `T: StringType` argument with a lifetime `'a` instead).
The invasiveness of the change is limited by using `Paragraphs<'static>` in most downstream usages. Even so, this generated a fair amount of churn as type parameters went away.

Going forward we can apply a similar conversion to components where we need it - `Label` is an obvious candidate, correct @TychoVrahe ?

I haven't tested the effect of this change on bootloader code. It _does_ apply to bootloader. In theory, in bootloader only one variant `TString::Str` should exist, and as such, the enum _should_ be completely optimized out. In practice, :woman_shrugging:

### DIFF
--- a/core/embed/rust/src/strutil.rs
+++ b/core/embed/rust/src/strutil.rs
@@ -9,38 +9,18 @@ use crate::micropython::{buffer::StrBuffer, obj::Obj};
 #[cfg(feature = "translations")]
 use crate::translations::TR;
 
-/// Trait for slicing off string prefix by a specified number of bytes.
-/// See `StringType` for deeper explanation.
-pub trait SkipPrefix {
-    fn skip_prefix(&self, bytes: usize) -> Self;
-}
-
-// XXX only implemented in bootloader, as we don't want &str to satisfy
-// StringType in the main firmware. This is because we want to avoid duplication
-// of every StringType-parametrized component.
-#[cfg(feature = "bootloader")]
-impl SkipPrefix for &str {
-    fn skip_prefix(&self, chars: usize) -> Self {
-        &self[chars..]
-    }
-}
-
-/// Trait for internal representation of strings.
-/// Exists so that we can support `StrBuffer` as well as `&str` in the UI
-/// components. Implies the following operations:
-/// - dereference into a short-lived `&str` reference (AsRef<str>)
-/// - create a new string by skipping some number of bytes (SkipPrefix) - used
-///   when rendering continuations of long strings
+/// Trait for internal representation of strings. This is a legacy crutch before
+/// we fully transition to `TString`. For now, it allows some manner of
+/// compatibility between `&str` and `StrBuffer`. Implies the following
+/// operations:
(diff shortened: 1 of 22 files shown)


core/embed/rust/src/strutil.rs
```diff
@@ -9,38 +9,16 @@ use crate::micropython::{buffer::StrBuffer, obj::Obj};
...
 /// components. Implies the following operations:
 /// - dereference into a short-lived `&str` reference (AsRef<str>)
 /// - create a new string by skipping some number of bytes (SkipPrefix) - used
 ///   when rendering continuations of long strings
 /// - create a new string from a string literal (From<&'static str>)
-pub trait StringType:
-    AsRef<str> + From<&'static str> + Into<TString<'static>> + SkipPrefix
-{
-}
+pub trait StringType: AsRef<str> + From<&'static str> + Into<TString<'static>> {}
```
obrusvit, line 19: Can we update the docs above? Namely, `SkipPrefix` part is obsolete and `Into<TString>` is not mentioned.

mmilata (approved):
Looks good to me. Was curious if it increases flash size, which it does by about 1K, so no need to concern.

TychoVrahe:
Tried to expand on this in https://github.com/trezor/trezor-firmware/pull/3649 , which should cover the needs i have for #2521 

Bootloader size does not seem affected by these changes

core/embed/rust/src/strutil.rs
matejcik: c28dfe3db1bfdb2e1610e5875ffd9a0d9db0ea37

core/embed/rust/src/ui/component/text/op.rs
```diff
@@ -147,17 +147,20 @@ impl<'a, T: StringType + Clone + 'a> OpTextLayout<T> {
...
+        I: Iterator<Item = &'b Op<'a>> + 'b,
         'a: 'b,
     {
         let mut skipped = 0;
         ops_iter.filter_map(move |op| {
             match op {
                 Op::Text(text, _continued) if skipped < skip_bytes => {
                     let skip_text_bytes_if_fits_partially = skip_bytes - skipped;
-                    skipped = skipped.saturating_add(text.as_ref().len());
+                    skipped = skipped.saturating_add(text.map(str::len));
```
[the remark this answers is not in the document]
matejcik: c28dfe3db1bfdb2e1610e5875ffd9a0d9db0ea37

core/embed/rust/src/strutil.rs
obrusvit: :+1: for crutch

matejcik:
rebased, fixed conflicts. @mmilata please look it over

mmilata (approved):

PR #3616, opened 2024-03-17
fix(core): fixed blurring algo panic in debug mode

cepetr:
This PR addresses crashes in the simulator caused by overflow in the blur algorithm.

On the right side of `+=` or `-=` operations, the value may become negative, resulting in a `u16` overflow for `RUST_PROFILE == 'dev'`.

This bug does not affect any hardware or simulators running with `RUST_PROFILE=='release'`.


TychoVrahe (approved):
lgtm

PR #3617, opened 2024-03-17
Cepetr/ts3 font fix

cepetr:
This PR adresses 3 fonts used on TS3:

- `unifont_bold_16` - Fixed font height and max_height
- `unifont_regular_16` - Fixed font height and max_height
- `pixeloperator_mono_regular_8` - Fixed underscore character

The modified font height was compensated in a few UI components by using
the real height of capital characters. As a result, only a few screens were
affected by this change.

image

The underscore character was shortened by one pixel (from 8 pixels to 7).
This alteration affected many tests, which must be modified accordingly.

image

image


core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
```diff
@@ -217,7 +217,7 @@ where
         // Getting the row area for the choices - so that displaying
         // items in the used font will show them in the middle vertically.
         let area_height_half = self.pad.area.height() / 2;
-        let font_size_half = theme::FONT_CHOICE_ITEMS.text_height() / 2;
+        let font_size_half = theme::FONT_CHOICE_ITEMS.visible_text_height("Ay") / 2;
```
grdddj, line 220: It might be worth creating some specific method, e.g. `visible_text_height_standard`, which would just call `visible_text_height("Ay")` - as it is repeated at least three times now

cepetr: You are right that it's repeating. We can introduce new function, but I am afraid it doesn't help code readability much if we do not name it properly.  I don't think there's something like "standard" height - maybe .capital_height(), uppercase_height(), lowercase_height(), allcase_height().. What do you think? @grdddj

grdddj: Right, the naming is the biggest obstacle here. From what you offer, I like `allcase` the most, as it is universal - or maybe even `universal`, as you can use it everywhere

cepetr: Finally I've choosen `allcase` (I considered also `mixedcase`). Changes here 058b5bc59cb73d8f6d8171fa9256ae47edc61a5d

grdddj (approved):
utACK

matejcik (approved):

PR #3620, opened 2024-03-18
cepetr/api improvements

cepetr:
This pull request introduces several enhancements to certain structures that the new drawing library relies on:

- `Point` - addition of the unary minus operator.
- `Rect` - introduction of methods such as `is_empty()`, `normalize()`, `intersect()`, and `has_intersection()`.
- `Color` - incorporation of a function for color blending
- `UzlibContext` - inclusion of a function for skipping bytes in an uncompressed stream without reading them

These improvements do not currently impact any functionality since these functions are not yet utilized.


core/embed/rust/src/ui/display/color.rs
```diff
@@ -86,6 +86,30 @@ impl Color {
     pub const fn black() -> Self {
         Self::rgb(0, 0, 0)
     }
+
+    /// Blends the color of `self` with the color of `fg` using specified
+    /// `alpha` value (ranging from 0 to 255).
+    ///
+    /// If `alpha` equals 0, the background color (`self`) is used.
+    /// If `alpha` equals 255, the foreground color (`fg`) is used.
+    pub fn blend(self, fg: Color, alpha: u8) -> Color {
```
TychoVrahe, line 95: we have `impl Lerp for Color` below, even if we need separate `blend` function here we can reuse that

core/embed/rust/src/ui/display/toif.rs
```diff
@@ -271,6 +271,14 @@ impl<'i> Toif<'i> {
         Offset::new(self.width(), self.height())
     }
 
+    pub fn stride(&self) -> usize {
+        if self.is_grayscale() {
+            self.width() as usize / 2
```
TychoVrahe, line 276: Not sure how this is intended to be used. The division for grayscale toif rounds down the number of pixels to even number, which might or might not be ok depending on context

core/embed/rust/src/ui/geometry.rs
```diff
@@ -241,6 +252,10 @@ impl Rect {
         }
     }
 
+    pub fn from_size(size: Offset) -> Self {
```
TychoVrahe, line 255: could this be const? also applies to bunch of functions below

core/embed/rust/src/ui/geometry.rs
```diff
@@ -329,11 +344,26 @@ impl Rect {
...
     pub const fn contains(&self, point: Point) -> bool {
         point.x >= self.x0 && point.x < self.x1 && point.y >= self.y0 && point.y < self.y1
     }
 
+    pub fn has_intersection(&self, r: Rect) -> bool {
+        self.x0 < r.x1 && self.x1 > r.x0 && self.y0 < r.y1 && self.y1 > r.y0
+    }
+
+    pub fn intersect(&self, r: Rect) -> Rect {
+        Rect::new(
```
TychoVrahe, line 361: this seems to return a 'negative' rectangle in case of no intersection. I would more expect 'zero' rectangle of even returning `Option<Rect>`

cepetr: A "negative" rectangle is actually an empty rectangle. When applying intersection or union operations on negative rectangles, they behave as if they were a "zero" rectangle. Therefore, this does not lead to undefined behavior. While we may return a "zero" rectangle, it's not without a cost; thus, I would prefer not to do so. Returning an `Option<Rect>` is highly impractical.

core/embed/rust/src/ui/display/color.rs
cepetr: Lerp is using f32, so I am a bit afraid, this change will slow down all drawing ops a bit. But I'll check it, maybe not...

core/embed/rust/src/ui/geometry.rs
cepetr: Yes, of course, it's possible, fixed 2536ed1b7195001706689a843df0cfd43b08340d

core/embed/rust/src/ui/display/toif.rs
cepetr: It's a bug, fixed here fixed 2536ed1b7195001706689a843df0cfd43b08340d

core/embed/rust/src/ui/geometry.rs
TychoVrahe: While it is true for unions and intersect operations on rectangle, its does not behave the same for other operations, for example `height()` will return different result for zero or negative rectangle. From there, we would need to be sure that every algorithm using also height is correctly handling negative results, otherwise bugs might be introduced.

Regarding the Option, i would say it can be also more practical in _some_ cases, i.e. 
```
if let Some(intersect) = x.intersect(y) {
 do something with intersect
}
```
is more convenient then
```
if x.has_intersection(y) {
 do something with x.intersect(y)
}
```

But it perhaps depends mostly on the intended usage which i don't see here, so i will leave that up to you.

If this is performance critical you can leave it as is, but consider adding comment about this.

core/embed/rust/src/ui/display/color.rs
TychoVrahe: Right, its a valid concern if this is performance critical too. 

(we also have Color::r(), g(), b() functions and Color::rgb() for return value that would simplify this, not sure whether compiler would be able to optimize that calls away so that there is no cost)

core/embed/rust/src/ui/geometry.rs
cepetr: I actually understand your point of view, but the issue is more about defining what determines an empty rectangle and how we interpret negative values for height or width. It's quite easy to create a "negative" rectangle through  new(), inset(), or with_xxx() functions. and I think it's essential for us to acknowledge that negative width or height isn't inherently abnormal and should be handled accordingly. 

So I decided to just add few comments to make it more clear, see d6a575e88629cbae10fc5df9dc6e66b6c13d435f

But I am open to further discussion:-)

matejcik: I agree with @TychoVrahe. Negative rectangles seem to be a problem -- in particular, `normalize()` will magically turn a negative rectangle into a normal one, which seems like a footgun.

...what is the use of `normalize()` anyway? it looks like a magnet for surprises

if we need super high performance for `intersect()`, at least call it `intersect_unchecked` and add an `intersect` method that does "if !self.has_intersection(other) { return Self::empty() } else self.intersect_unchecked(other)"

core/embed/rust/src/ui/display/color.rs
cepetr: After further consideration, I found that  lerp with f32 arguments leads to longer and slower code (approx 2-3x). 

So  I decided to rewrite the code to utilize existing functions, resulting in a simpler implementation, while the size and speed outcomes remain identical.

see 86be1814e7e68f3f16130898dad44ed98b01d494

core/embed/rust/src/ui/geometry.rs
cepetr: I agree that `normalize()` looks hacky. Its purpose is a bit special, and it's only used in one place. It helps create rectangles from two arbitrary points, regardless of which corner of the rectangle they are. Perhaps it would be clearer to have a function `Rect::from_arbitrary_points(Point, Point)` or to remove it altogether.

cepetr: After finding the `clamp()` function and discussing it in person, I've made the following changes:
1) removed `normalize()` as it will be replaced by something else where it's currently used
2) rermoved `intersect()` - we will continue using `clamp()` instead.
3) removed `has_intersection()` - we will use `(!intersect().is_empty)` instead.

see a04980460dd50c960648c19c6ee5c30e55cb9fd6

TychoVrahe (approved):

PR #3621, opened 2024-03-18
refactor(core): prepare for non mutable paint function

cepetr:
This PR alters the logic of certain components by eliminating the requirement for `&mut self` in `paint()` functions. The PR serves as preparation for merging a new drawing library that requires it.

Two approaches were used:

1) Interior mutability utilizing `Cell` or `RefCell`.
2) Moving the mutation of component state to event processing.

**The behavior of the device was not affected by this change.**

### DIFF
--- a/core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
+++ b/core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
@@ -294,11 +294,6 @@ where
     fn show_current_choice(&mut self, area: Rect) {
         self.get_current_item()
             .paint_center(area, self.inverse_selected_item);
-
-        // Color inversion is just one-time thing.
-        if self.inverse_selected_item {
-            self.inverse_selected_item = false;
-        }
     }
 
     /// Display all the choices fitting on the left side.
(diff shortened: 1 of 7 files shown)


mmilata:
I'd fix the FormattedText differently, please take a look at c9d959d856826acda72cf62d0727a5836f2f0634 and cherry-pick if you want. Doing a throwaway copy in `layout_ops()` might actually be more correct than carrying the state between paints.

Otherwise LGTM, hopefully we'll fix the backlight fading when it's needed on some new model.

core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
```diff
@@ -491,6 +486,13 @@ where
     }
 
     fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
+        // Cancel highlighting of the current choice.
+        // The Highlighting is started by pressing the middle button and
+        // canceled immediately when any other event is processed
+        if self.inverse_selected_item {
+            self.inverse_selected_item = false;
```
matejcik, line 493: nit: the if is not doing anything here

matejcik:
> please take a look at [c9d959d](https://github.com/trezor/trezor-firmware/commit/c9d959d856826acda72cf62d0727a5836f2f0634) and cherry-pick if you want

love it, the `mut` in `layout_ops` was very ugly

core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
cepetr: :-) aha

cepetr: fixed here c2c346fcd2da325d94509d6514e6dc0f0793d283

cepetr:
I've also cherry-picked @mmilata's changes, here 3de6493247171c4fb769abbda94047a614f81203

matejcik (approved):

mmilata (approved):

PR #3625, opened 2024-03-19
Add AES GCM to trezor-crypto

andrewkozlik:


crypto/aes/gf128mul.c
```diff
@@ -0,0 +1,471 @@
...
+#if defined( GF_MODE_LL) || defined( GF_MODE_LB )
+#  define GF_INDEX(i)  (i)
+#else
+#  define GF_INDEX(i)  (15 - (i))
+#endif
+
+/* A slow field multiplier */
+
+void gf_mul(gf_t a, const gf_t b)
+{   gf_t p[8];
```
onvej-sl, line 55: I think we require explicit initialization even in foreign code.

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-aesgcm.h
```diff
@@ -0,0 +1,205 @@
...
+///     """
+typedef struct _mp_obj_AesGcm_t {
+  mp_obj_base_t base;
+  gcm_ctx ctx;
+  enum {
+    STATE_INIT,
+    STATE_ENCRYPTING,
+    STATE_DECRYPTING,
+    STATE_FINISHED,
+  } state;
```
onvej-sl, line 39: Would it make sense to introduce `STATE_FAIL` to which the context would fall when some its method raises an exception?

crypto/aes/aes.h
```diff
@@ -131,9 +130,7 @@ AES_RETURN aes_encrypt_key192(const unsigned char *key, aes_encrypt_ctx cx[1]);
 AES_RETURN aes_encrypt_key256(const unsigned char *key, aes_encrypt_ctx cx[1]);
 #endif
 
-#if defined( AES_VAR )
```
onvej-sl, line 134: It seems to me that this, along with the missing check of the return value [here](https://github.com/trezor/trezor-firmware/blob/2446006db4d414dc00acffa18f88053e20bcd208/crypto/aes/aesgcm.c#L105), could be dangerous. If you compile it without `AES_256` and call `gcm_init_and_key(key, 32, ctx)`, the key will be ignored, but the function returns `RETURN_GOOD`.

crypto/aes/aesgcm.h
```diff
@@ -0,0 +1,233 @@
...
+Copyright (c) 1998-2010, Brian Gladman, Worcester, UK. All rights reserved.
+
+The redistribution and use of this software (with or without changes)
+is allowed without the payment of fees or royalties provided that:
+
+  source code distributions include the above copyright notice, this
+  list of conditions and the following disclaimer;
+
+  binary distributions include the above copyright notice, this list
+  of conditions and the following disclaimer in their documentation.
```
onvej-sl, line 12: This comment is not directly related to this pull request as the header was already present in our codebase.
However, does our binary distribution comply with this notice?

crypto/aes/brg_endian.h
```diff
@@ -0,0 +1,9 @@
+#ifndef _BRG_ENDIAN_H
```
onvej-sl, line 1: Missing copyright header.

crypto/aes/aesgcm.h
andrewkozlik: @matejcik?

crypto/aes/gf128mul.c
andrewkozlik: AFAICS we not. A few examples from several different files:
https://github.com/trezor/trezor-firmware/blob/866853e856299d7498699ae765757a470e253536/crypto/aes/aescrypt.c#L141
https://github.com/trezor/trezor-firmware/blob/866853e856299d7498699ae765757a470e253536/crypto/aes/aes_modes.c#L360
https://github.com/trezor/trezor-firmware/blob/866853e856299d7498699ae765757a470e253536/crypto/aes/aeskey.c#L83

crypto/aes/brg_endian.h
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3625/commits/ac5497e357a3febe65b55a43a258839c6da5adb4

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-aesgcm.h
andrewkozlik: I don't think it adds much value, but it doesn't hurt as a failsafe. https://github.com/trezor/trezor-firmware/pull/3625/commits/c9f043bd966411cdb0597b0f221ddefc46a5fc9d

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-aesgcm.h
```diff
@@ -0,0 +1,213 @@
...
+                                                 const mp_obj_t *args) {
+  mp_arg_check_num(n_args, n_kw, 2, 2, false);
+  mp_obj_AesGcm_t *o = m_new_obj_with_finaliser(mp_obj_AesGcm_t);
+  o->base.type = type;
+  o->state = STATE_INIT;
+  mp_buffer_info_t key = {0}, iv = {0};
+  mp_get_buffer_raise(args[0], &key, MP_BUFFER_READ);
+  mp_get_buffer_raise(args[1], &iv, MP_BUFFER_READ);
+  if (key.len != 16 && key.len != 24 && key.len != 32) {
+    m_del_obj(mp_obj_AesGcm_t, o);
```
onvej-sl, line 58: Is `m_del_obj()` really needed here? I'm asking because if it is, then it is missing in many other places in our codebase (for example [here](https://github.com/trezor/trezor-firmware/blob/c9f043bd966411cdb0597b0f221ddefc46a5fc9d/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-aes.h#L74))

crypto/aes/gf128mul.c
andrewkozlik: The above are just examples. There are at least ten times as many cases in the AES code. If the AES code was excluded in 2019, then I don't see any reason to override that decision today.

crypto/aes/aesgcm.h
andrewkozlik: https://satoshilabs.slack.com/archives/CKN5C51CJ/p1711375279588959?thread_ts=1711359561.629369&cid=CKN5C51CJ

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-aesgcm.h
andrewkozlik: Here are the others https://github.com/trezor/trezor-firmware/pull/3625/commits/c0f3ad49fe78a65e167ee5b161eb7f558862aad2

crypto/aes/gf128mul.c
onvej-sl: If I remember correctly, the decision was to enforce it in the foreign code as well. Therefore, my comment is compliant with the decision made at that time.

The reason why it wasn't fully accomplished is because I used an automated tool, which unfortunately failed in numerous instances with the AES code.

andrewkozlik: I see that there were some changes to the AES files as part of the 2019 PR. Personally, I am not convinced we should be doing this, not to mention that it is completely out of scope of this PR, but since we already started, let's at least be consistent. https://github.com/trezor/trezor-firmware/pull/3625/commits/10644d4f3d40114554b9d2c6c93a9de0071c43bc

PR #3631, opened 2024-03-20
slowdown i2c on u5 to 200kHz

TychoVrahe:
communication with optiga was sometimes losing data

### DIFF
--- a/core/embed/trezorhal/stm32u5/i2c.c
+++ b/core/embed/trezorhal/stm32u5/i2c.c
@@ -60,8 +60,12 @@ i2c_instance_t i2c_defs[I2C_COUNT] = {
  * Using calculation from STM32CubeMX
  * PCLKx as source, assumed 160MHz
  * Fast mode, freq = 400kHz, Rise time = 250ns, Fall time = 100ns
+ * Fast mode, freq = 200kHz, Rise time = 250ns, Fall time = 100ns
+ * SCLH and SCLL are manually modified to achieve more symmetric clock
  */
-#define I2C_TIMING_400000_Hz 0x30D2153A
+#define I2C_TIMING_400000_Hz 0x30D22728
+#define I2C_TIMING_200000_Hz 0x30D2595A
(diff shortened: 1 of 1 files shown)


core/embed/trezorhal/stm32u5/i2c.c
```diff
@@ -62,6 +62,8 @@ i2c_instance_t i2c_defs[I2C_COUNT] = {
  * Fast mode, freq = 400kHz, Rise time = 250ns, Fall time = 100ns
  */
 #define I2C_TIMING_400000_Hz 0x30D2153A
+#define I2C_TIMING_200000_Hz 0x30D2159E
+#define I2C_TIMING I2C_TIMING_200000_Hz
```
cepetr, line 66: The new timing settings lead to a 200 kHz I2C clock. However, as we only change Tscll (leaving Tsclh as it is), it may still cause problems.

The original settings were:

- Tpresc = 1 (160Mhz / (3 + 1)) = 25 ns 
- Tscdel = (13 + 1) * 25ns = 350 ns 
- Tsdadel = (2 + 1) * 25ns =  75 ns 
- Tscl = (21 + 1) * 25ns = 550 ns 
- Tscll = (58 + 1) * 25 ns = 1400 ns 

The new settings change:

- Tscll = (158 + 1) * 25ns = 3975 ns

So the clock is quite asymmetric, and Tsclh remains quite short (by the way, it was always a bit short). I would probably distribute the delay between Tscll and Tsclh.

TychoVrahe: I used CubeMX to generate these values (which is also recommended by reference manual), not sure why it generates such asymmetric values.

Anyway, i agree that symmetric clock is better. Retested this with modified settings and it seems to work, also adjusting the original 400kHz setting does not fix the optiga issue.

https://github.com/trezor/trezor-firmware/pull/3631/commits/2d54e932e117ddcf8fe6b980c0242af203491f4c

cepetr (approved):

PR #3632, opened 2024-03-21
fix order of loading BHK into SAES key register

TychoVrahe:
According to errata, BHK loading into SAES key register must be done in ascending order.

Though i have verified that encryption and decryption still works (comparing results from SAES using BHK and regular hw AES peripheral using same key), the key is loaded into key reg according to the access order to the backup register thus ends up being different from what one would expect by looking into reference manual.

I don't think this makes any difference in our case, but nevertheless, lets follow the docs.

https://www.st.com/resource/en/errata_sheet/es0499-stm32u575xx-and-stm32u585xx-device-errata-stmicroelectronics.pdf

saes_errata

saes_reference

### DIFF
--- a/core/embed/trezorhal/stm32u5/secure_aes.c
+++ b/core/embed/trezorhal/stm32u5/secure_aes.c
@@ -40,14 +40,14 @@ secbool secure_aes_init(void) {
 }
 
 static void secure_aes_load_bhk(void) {
-  TAMP->BKP7R;
-  TAMP->BKP6R;
-  TAMP->BKP5R;
-  TAMP->BKP4R;
-  TAMP->BKP3R;
-  TAMP->BKP2R;
-  TAMP->BKP1R;
   TAMP->BKP0R;
+  TAMP->BKP1R;
+  TAMP->BKP2R;
+  TAMP->BKP3R;
+  TAMP->BKP4R;
+  TAMP->BKP5R;
+  TAMP->BKP6R;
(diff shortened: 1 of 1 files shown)


core/embed/trezorhal/stm32u5/secure_aes.c
```diff
@@ -40,14 +40,14 @@ secbool secure_aes_init(void) {
...
 
 static void secure_aes_load_bhk(void) {
-  TAMP->BKP7R;
-  TAMP->BKP6R;
-  TAMP->BKP5R;
-  TAMP->BKP4R;
-  TAMP->BKP3R;
-  TAMP->BKP2R;
-  TAMP->BKP1R;
   TAMP->BKP0R;
```
prusnak, line 43: How does this work if there is no value assigned/read?

TychoVrahe: Loading BHK into SAES key register is using dedicated bus, so that the values being transferred are not accessible to software. This is the way to trigger the transfer.

prusnak: I guess there are some macros that tell the compiler to not optimize these calls. (maybe just using `volatile` is enough?)

TychoVrahe: yeah `volatile` should be enough

cepetr (approved):

PR #3634, opened 2024-03-21
Feat(solana): UI hide instructions

RostarMarek:
Until now compute budget instructions (`SetComputeUnitLimit` and `SetComputeUnitPrice`) were being displayed, adding unnecessary screens as the instructions affect only the final fee which is displayed on Trezor as well. This leads us to believe that they can be ommited as the relevant information for the user is the final fee.
Including Compute Budget instructions also broke the "Predefined transactions" logic. This wasn't a huge problem before since wallets didn't use priority fees - but they started to use them now. Priority fees support will be added to Trezor Suite as well and so not fixing the predefined transactions would break the UX on Trezor with Suite since predefined transactions would not work.
The proposal for this fix has been written [in this issue](https://github.com/trezor/trezor-suite/issues/11518#issuecomment-1991621299) and in this PR we achieve it by implementing a flag for solana instructions `is_ui_hidden` which is then used to hide instructions that that should not be shown. This flag is optional and by default will be set to `False`.


core/src/apps/solana/transaction/instructions.py.mako
```diff
@@ -190,6 +198,7 @@ def get_instruction(
                 "${program["name"]}: ${instruction["name"]}",
                 True,
                 True,
+                ${getOptionalBool(instruction, "is_ui_hidden")},
```
matejcik, line 201:
```suggestion
                ${instruction.get("is_ui_hidden", False)},
```

matejcik: same thing for `getOptionalString` btw. Mako templates are mostly just Python. that should have been just `ui_property.get("parameter")`

matejcik (approved):
small nits in template generation, overall LGTM

core/src/apps/solana/transaction/instructions.py.mako
RostarMarek: For the bool i fixed it in cfb3bba , 
but for string there is an issue that if the when we are generating and the optional field contains some string we would like to have it returned as `"${ui_property.get("parameter")}"`, but when it does not we would like to have it returned as `${ui_property.get("parameter")}`

I can provide a quick example if needed as I am not really sure If i illustrated it well enough

RostarMarek: For the same functionality the lines with `getOptionalString` could be replaced with 
```
${'"'+ui_property.get("parameter")+'"' if "parameter" in ui_property else None},
```
which I don't think is realy nicely readable

matejcik: no, that's literally what `ui_property.get("parameter")` does

```
>>> foo = { "hello": 123 }
>>> print(foo.get("hello"))
123
>>> print(foo.get("world"))
None
```

matejcik: oh, i see, the problem is with the quotes.
well, okay, let's leave it as-is.

for the record you don't have to use Mako's syntax, you can just write a python function
this one is better btw:
```
def getOptionalString(obj, string):
    if string in obj:
        return json.dumps(obj[string])
    else:
        return "None"
```
json.dumps will take care of quotes and `\n`s and the like for you

RostarMarek:
As per contributing guidelines it is rebased with autosquash, but I am not authorized to merge, so sadly someone will have to take that over from me

Hannsek:
@matejcik where are the UI tests? I would like to see the differences that this pr did.

RostarMarek:
Currently have only one on me which was when accepting the changes
Screenshot 2024-03-20 at 18-04-55 TT_en-device_tests-solana-test_sign_tx py test_solana_sign_tx transfer_with_compute_budget
and for some reasone emulator does not want to run for me right now, so if you need me to show more just let me know and I will fix it and provide more proof.

Hannsek:
Thank you very much!

- Do we need this screen? 
image

PR #3639, opened 2024-03-22
ci: actions: translations label trigger, update_fixtures.py support

mmilata:
* all-language test suite was previously run only once after adding the `translations` label, now it runs every time the label is present
* `tests/update-fixtures.py ci --github` now fetches `fixtures.results.json` from GitHub Actions (GitLab has no T3T1), followup of #3549


matejcik (approved):
utACK, seems nice

grdddj (approved):
Looks good!

And thanks for fixing the `##START_DEVICE_TESTS` lint

PR #3644, opened 2024-03-25
T3T1 Instructions component

obrusvit:
This PR adds only the instruction component definition. The actual usage will be added in another PR.

### DIFF
--- a/core/embed/rust/src/ui/model_mercury/component/instructions.rs
+++ b/core/embed/rust/src/ui/model_mercury/component/instructions.rs
@@ -1,0 +1,129 @@
+use crate::{
+    strutil::TString,
+    ui::{
+        component::{text::TextStyle, Component, Event, EventCtx, Never},
+        constant::WIDTH,
+        geometry::{Alignment, Offset, Rect},
+        model_mercury::theme,
+        shape::{Renderer, Text},
+    },
+};
+
+/// Component showing a task instruction (e.g. "Swipe up") and optionally task
+/// description (e.g. "Confirm transaction") to a user. The component
+/// is typically placed at the bottom of the screen. The height of the provided
+/// area must be 18px (only instruction) or 37px (both description and
+/// instruction). The content and style of both description and instruction is
+/// configurable separatedly.
+pub struct Instructions<'a> {
+    area: Rect,
+    text_instruction: TString<'a>,
+    text_description: Option<TString<'a>>,
+    style_instruction: &'static TextStyle,
+    style_description: &'static TextStyle,
+}
+
+impl<'a> Instructions<'a> {
+    /// height for component with only instruction [px]
+    pub const HEIGHT_SIMPLE: i16 = 18;
+    /// height for component with both description and instruction [px]
+    pub const HEIGHT_DEFAULT: i16 = 37;
+
+    pub fn new<T: Into<TString<'a>>>(instruction: T) -> Self {
+        Self {
+            area: Rect::zero(),
+            text_instruction: instruction.into(),
+            text_description: None,
+            style_instruction: &theme::TEXT_SUB,
+            style_description: &theme::TEXT_SUB,
+        }
+    }
+
+    pub fn with_description<T: Into<TString<'a>>>(self, description: T) -> Self {
+        Self {
+            text_description: Some(description.into()),
+            ..self
+        }
+    }
+
+    pub fn update_instruction<T: Into<TString<'a>>>(&mut self, ctx: &mut EventCtx, s: T) {
+        self.text_instruction = s.into();
+        ctx.request_paint();
+    }
+
+    pub fn update_description<T: Into<TString<'a>>>(&mut self, ctx: &mut EventCtx, s: T) {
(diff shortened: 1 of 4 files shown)


core/embed/rust/src/ui/model_mercury/component/instructions.rs
```diff
@@ -0,0 +1,125 @@
...
+    style_description: &'static TextStyle,
+}
+
+impl<'a> Instructions<'a> {
+    /// height for component with only instruction [px]
+    pub const HEIGHT_SIMPLE: i16 = 18;
+    /// height for component with both description and instruction [px]
+    pub const HEIGHT_DEFAULT: i16 = 37;
+
+    pub fn new<T: Into<TString<'a>>>(instruction: T) -> Self {
```
mmilata, line 32: Nit: Putting the bound in the impl block might make this code a bit more readable. Not quite sure why we usually put them there and not in the struct definition.

```rust
impl<'a, T> Instructions<'a>
where
    T: Into<TString<'a>>,
{
```

core/embed/rust/src/ui/model_mercury/component/instructions.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    pub fn with_description<T: Into<TString<'a>>>(self, description: T) -> Self {
+        Self {
+            text_description: Some(description.into()),
+            ..self
+        }
+    }
+
+    pub fn update_instruction<T: Into<TString<'a>>>(&mut self, s: T) {
```
mmilata, line 49: Nit: usually our `update_...` functions take `ctx: &mut EventCtx` as an argument so they can call `.request_paint()` on it. Perhaps it's not going to be needed with the new graphics library but I'd do it here so that we're consistent across the whole model.

core/embed/rust/src/ui/model_mercury/component/instructions.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+impl<'a> Component for Instructions<'a> {
+    type Msg = Never;
+
+    fn place(&mut self, bounds: Rect) -> Rect {
+        let h = bounds.height();
+        assert!(h == Instructions::HEIGHT_SIMPLE || h == Instructions::HEIGHT_DEFAULT);
+        assert!(bounds.width() == WIDTH);
```
mmilata, line 72: Appreciate the defensiveness here but in the past it has paid off to do the right thing for other sizes when it's not much work - e.g. model T firmware works fairly well on [STM32F429I-DISC1](https://docs.trezor.io/trezor-firmware/core/misc/disc1.html) despite its non-square screen. My suggestion is to remove the asserts and either fill the entire bounds or split off a fixed height rectangle from them.

mmilata (commented):
Looks good, got some suggestions.

core/embed/rust/src/ui/model_mercury/component/instructions.rs
obrusvit: I did try but it seems that this syntax is not supported. I tried what you suggested and also this:
```rust
impl<'a, T: Into<String<'a>>> Instructions<'a> {...}
```
but I got the same compilation error:
image

core/embed/rust/src/ui/model_mercury/component/instructions.rs
obrusvit: I actually don't see a problem here with using this component on different devices.
The `assert`s ensure that a parent component provided rectangle of certain height (18px or 37px) spanned over the whole width (`WIDTH` it taken from `model_xyz/constant.rs`). So the only problem on bigger screen is that the instruction would be unnecessarily small at the bottom.

core/embed/rust/src/ui/model_mercury/component/instructions.rs
obrusvit: fix: https://github.com/trezor/trezor-firmware/pull/3644/commits/494bb6e3dd885f6b603b2140ff6b636afbec62db

core/embed/rust/src/ui/model_mercury/component/instructions.rs
obrusvit: ChatGPT also didn't suggest a way:

> However, this syntax is not directly supported in Rust, as Rust requires specifying trait bounds for each method where they are needed, not at the impl block level.
A common workaround for this repetition is to use a type alias with the bound included, but unfortunately, Rust's type system doesn't allow specifying generic parameter bounds on type aliases in a way that would solve this problem directly.

mmilata: my bad, didn't know this doesn't work when T is just a function argument

mmilata (approved):

Hannsek:
@obrusvit please post here a screenshot of this. 🙏🏻

obrusvit:
@Hannsek This PR adds ability for screens to contain a configurable "footer" with `Instruction`  ("Swipe up", ..) and optionally `Description`. Here's a preview:
instruction_preview

PR #3646, opened 2024-03-25
Improve tooling for translation blobs

### LINKED ISSUE #3535: New string in translations requires clean build to work properly

matejcik:
fixes #3535 (separated from #3610 because this change needs to come in first, so that we can release unmodified Czech translation)

improves translations/cli so that signatures go directly into `history`, hopefully preventing a loss of signature like we did with this release merge

adds translation checking to `make gen_check` so CI should catch outdated translation hashes


grdddj (approved):
Apart from the `make gen` issues, LGTM - I have looked through the code and played a little with the `translations/cli.py` script - works fine for me

matejcik:
the `make gen` errors should be fixed in a280491cedac5869cc6d7088b0d13f38d5a09e32, please take a quick look?

grdddj:
> the `make gen` errors should be fixed in [a280491](https://github.com/trezor/trezor-firmware/commit/a280491cedac5869cc6d7088b0d13f38d5a09e32), please take a quick look?

I can run it locally fine, even in CI

 ACK the shell->python changes (haven't tested the `os.utime` magic though)

Just style check (isort) fails: https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/6478979605

Issue #3643, opened 2024-03-25
fix uninitialized structs in HAL code


TychoVrahe:
see https://github.com/trezor/trezor-firmware/pull/3630#issuecomment-2011608420

prusnak:
draft PR in https://github.com/trezor/trezor-firmware/pull/3647

PR #3647, opened 2024-03-25
fix(core): always zero-initialize structures in trezorhal

prusnak:
Untested code, please test!

Also feel free to take over and force-push to this branch.

Fixes https://github.com/trezor/trezor-firmware/issues/3643

PS: Also what is the easiest way how check how the PR increases the size of the firmware? Do we have a CI test for that?


cepetr (approved):
It's looking good.Although I haven't tested it yet, I've compiled and compared the flash size before and after these changes. The modification only increases the final binary size of the boardloader/bootloader by a few tens of bytes.

TychoVrahe:
We have CI jobs, although they only compare firmware size, not bootloader/boardloader, see 
https://github.com/trezor/trezor-firmware/actions/runs/8422330162/job/23061992299?pr=3647
https://github.com/trezor/trezor-firmware/actions/runs/8422330162/job/23061992809?pr=3647

the latter one (comparison for T2B1) seems broken, as it compares TT binary from `main` with TR build in PR. 

Current: flash=714K flash2=803K
Master:  flash=714K flash2=803K
Size_diff: 0 K

TychoVrahe:
tested on T2T1 and T3T1 hw, does not seem to cause any obvious problems. 

Noticed that we are still using copy of sd card driver from HAL, so i changed that in https://github.com/trezor/trezor-firmware/pull/3647/commits/cea992feadbee5cdef8e37b60c54858848a5b5af as it is not needed anymore and we don't need to fix the struct initializations inside STM32 HAL files

@cepetr please check the added commit.

cepetr:
I confirm that the removed file and the file at STM2 HAL are the same.

cepetr (approved):

PR #3649, opened 2024-03-26
use TString in Label and Button

TychoVrahe:
Expanding on https://github.com/trezor/trezor-firmware/pull/3614


core/embed/rust/src/ui/model_tt/layout.rs
```diff
@@ -666,10 +661,10 @@ extern "C" fn new_confirm_homescreen(n_args: usize, args: *const Obj, kwargs: *m
         };
 
         let tr_change: StrBuffer = TR::buttons__change.try_into()?;
-        let buttons = Button::cancel_confirm_text(None, Some(tr_change));
+        let buttons = Button::cancel_confirm_text(None, Some(tr_change.into()));
```
obrusvit, line 664: nit: 
```rust
// let tr_change: StrBuffer = TR::buttons__change.try_into()?;  // not needed
let buttons = Button::cancel_confirm_text(None, Some(TR::buttons__change.into()));
```

core/embed/rust/src/translations/flash.rs
```diff
@@ -64,16 +64,18 @@ pub fn init() {
     }
 }
 
-// SAFETY: Invalidates all references coming from the flash-based blob.
-// In other words, none should exist when this function is called.
+/// # Safety
+/// Invalidates all references coming from the flash-based blob.
+/// In other words, none should exist when this function is called.
```
matejcik, line 69: i'm all for doing things the way clippy likes them, but `SAFETY:` is the convention in our codebase and if we want to change it, we should change it everywhere all at once.

core/embed/rust/src/translations/translated_string.rs
```diff
@@ -4,16 +4,15 @@ use super::blob::Translations;
...
             .and_then(|s| s.translation(self as _))
             .unwrap_or(self.untranslated())
     }
 
-    pub fn map_translated<F, T>(self, fun: F) -> T
+    pub fn map_translated<'a, F, T>(self, fun: F) -> T
     where
-        F: for<'a> FnOnce(&'a str) -> T,
-        T: 'static,
+        F: FnOnce(&'a str) -> T,
```
matejcik, line 15: Come on man!
I went to all the trouble to forbid the string from escaping the closure, and yet that's exactly what you are reintroducing. `x.as_ref()` is no more safe than `x.map(|c| c)`. The lifetime of the translated string is _independent of the lifetime of its TString_ -- namely, it can become invalid any time it pleases. Hence the restrictive lifetimes.

matejcik (changes requested):

core/embed/rust/src/ui/component/label.rs
```diff
@@ -63,22 +63,25 @@ where
...
+            .fit_text(self.text.map(|c| c))
             .height()
     }
 
     pub fn text_area(&self) -> Rect {
         // XXX only works on single-line labels
         assert!(self.layout.bounds.height() <= self.font().text_max_height());
         let available_width = self.layout.bounds.width();
-        let width = self.font().text_width(self.text.as_ref());
+        let width = self.font().text_width(self.text.map(|c| c));
```
matejcik, line 84: specifically, you can't go around `s/.as_ref()/.map(|c| c)/` because that's incorrect code.
what you need here is:
```suggestion
        let width = self.text.map(|c| self.font().text_width(c)));
```

matejcik:
please cherry-pick https://github.com/trezor/trezor-firmware/pull/3671

core/embed/rust/src/ui/model_tt/layout.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/pull/3649/commits/26bdc00bde79952078e33d5bf6701a36e3c36293

core/embed/rust/src/translations/translated_string.rs
TychoVrahe: sorry, i misunderstood the whole idea. 
hopefully this will be better: https://github.com/trezor/trezor-firmware/pull/3649/commits/a045dce2ae1f54bd6d8eb5fccee051fa194de545

core/embed/rust/src/ui/component/label.rs
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/a045dce2ae1f54bd6d8eb5fccee051fa194de545

TychoVrahe:
seems to be working with the cherry-pick too. it needs rebase, ideally when the Paragraphs gets merged, @matejcik let me know when its ok with you.

mmilata:
I've squashed and rebased this on top of main, feel free to use: 0668729a2989248a2eef4eb8fb1bd328837c5b60
Also got some suggestions: bb48749ecf75af62b3ce07d6843c309efc2a0f3e

FWIW this PR saves about 2KiB of .flash2 presumably due to there being less instances of generics that depended on a string type.

TychoVrahe:
@mmilata should we close this PR in favor of https://github.com/trezor/trezor-firmware/pull/3682 ?

mmilata:
Originally created it to see CI results but sure we can continue there.

PR #3652, opened 2024-03-27
do not require sd detect pin for running SD card test on T3T1

TychoVrahe:
Due to the HW limitation and tester concept it is not possible to set the SD detect signal when testing the SD card on T3T1.

[no changelog]


cepetr (approved):

TychoVrahe:
turns out this was slightly more complicated

cepetr (approved):

PR #3654, opened 2024-03-27
Optiga code improvements

andrewkozlik:
I took several commits from https://github.com/trezor/trezor-firmware/pull/3335, which I believe to be uncontroversial and would be helpful to have merged soon. It's basically this:
- Do not lock Optiga in non-production builds. Very useful for development. I have already locked two devices accidentally.
- Logging of Optiga commands in debug builds. Very useful to expedite debugging of Optiga-related issues.
- Set trust anchor data type in prodtest. This fixes a problem that we had when switching between prodtest and hsm firmware builds I think.
- [refactor(core): Refactor Optiga macros.](https://github.com/trezor/trezor-firmware/commit/273974f8e1a9beaf13579f590a7b8d1461456518) - No significance, but makes some bits of code a little more legible.
- [chore(core): add prodtest provisioning script](https://github.com/trezor/trezor-firmware/commit/a9afaee8a93330d6c788e7f75987d8b1e8353ddf) - I didn't actually review/test this, but it's just a script for development testing, so I can see no harm.

The only actual difference in the selected commits over #3335 is that I improved the logging to show the timestamp (very useful for determining how long individual Optiga operations take) and added some color:

Screenshot from 2024-03-27 13-27-27

The remaining commits in #3335 may need a little more thought or discussion I think.


matejcik (approved):

matejcik:
rebased on main & approved.
we can simply merge this, right @andrewkozlik ?

PR #3655, opened 2024-03-27
refactor(core): introduce Simplified trait

matejcik:
to abstract over some common operations between models.

This also fixes test failure on T2B1 caused by incorrect import symbols.

related to #3608


TychoVrahe:
I like the trait idea, but i think this doesn't go far enough for my needs. 

I have expanded this concept in https://github.com/trezor/trezor-firmware/pull/3658 , see if it makes sense to you.

matejcik:
overall yes. but let's resolve that in your PR and merge this one (unless you see something wrong), so that we get the fix for the T2B1 unit tests

TychoVrahe (approved):
No problem with this as an intermediary step.

PR #3658, opened 2024-03-28
introduce UIFeatures trait

TychoVrahe:
Adds an abstraction over different UI implementations

Two UI traits are introduced:
`UIFeaturesBootloader`
`UIFeaturesCommon`

`UIFeaturesFirmware` would be next step. 

API files are introduced too, defining an unified interface between C and rust in `ui/api/bootloader_c.rs` and `ui/api/common_c.rs` . Purpose is to make all necessary type conversions here, so that specific UI implementations for different models get already pure rust arguments, and the type conversions are same for all UIs.


core/embed/rust/src/ui/model_tr/mod.rs
```diff
@@ -7,11 +7,19 @@ pub mod component;
 pub mod constant;
 #[cfg(feature = "micropython")]
 pub mod layout;
-pub mod screens;
+mod screens;
 pub mod theme;
 
-pub struct ModelTRFeatures {}
+pub struct ModelTRUIFeaturesCommon {}
```
mmilata, line 13: Why not keep one struct (per model) and implement both traits for it?

mmilata (commented):
otherwise :+1: for the C api improvements

core/embed/rust/src/ui/model_tr/mod.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/584d8d209c049bd4ad6bedd674ec5f797dd28f2e

core/embed/rust/src/ui/model_tr/bootloader/mod.rs
```diff
@@ -295,31 +256,105 @@ extern "C" fn screen_install_success(
...
 }
+
+impl UIFeaturesBootloader for ModelTRFeatures {
+    fn screen_welcome() {
+        let mut frame = Welcome::new();
+        show(&mut frame, true);
+    }
+
+    fn bld_continue_label(bg_color: Color) {
+        bld_continue_label(bg_color);
```
matejcik, line 288: any reason not to inline the functions into the trait implementation?

core/embed/rust/src/ui/model_tt/bootloader/mod.rs
```diff
@@ -320,19 +282,91 @@ extern "C" fn screen_install_success(
...
+            version_cmp,
+        )
+    }
+
+    fn screen_wipe_confirm() -> u32 {
+        screen_wipe_confirm()
+    }
+
+    fn screen_unlock_bootloader_confirm() -> u32 {
+        0
```
matejcik, line 336: i'd strongly prefer `unimplemented!()` here and below

core/embed/rust/src/ui/model_tr/bootloader/mod.rs
TychoVrahe: i was just afraid of merging with https://github.com/trezor/trezor-firmware/pull/3649 , which also edits this file heavily. But it can be handled i guess if you prefer it

matejcik: i would expect git to handle this fine, and if not, 100% sure vscode merge editor will deal with it. please do it that way.

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/6625666b53940c96d24f0384a31cf434206707fd

matejcik (approved):

PR #3659, opened 2024-03-28
chore(common): drop wallets.json

matejcik:
the wallet info is not used by firmware or consumers of trezor-common and so it moves to the definitions repo
at https://github.com/trezor/definitions


grdddj (approved):
Agree with the removal

PR #3660, opened 2024-03-30
fix(core/translations): correctly rename previous symbol

matejcik:
In #3610 I forgot that one of the symbols existed in the previous code.
This PR renames the old symbol instead of discarding and creating a new one, which is a tiny bit nicer on the size of the translation blob.


grdddj (approved):
Seems OK

PR #3664, opened 2024-03-30
ci: actions: add T3T1 firmware build

mmilata:
Skipping `build_prodtest` as it's failing with
```
scons: *** [build/prodtest/embed/prodtest/vendorheader.o] Source `embed/vendorheader/T3T1/vendorheader_unsafe_signed_prod.bin' not found, needed by target `build/prodtest/embed/prodtest/vendorheader.o'.
```

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


mmilata:
Indeed it builds fine, I probably forgot to use the parameter when building locally.

TychoVrahe (approved):

PR #3665, opened 2024-03-31
feat(tests): reduce ui error spam

matejcik:
before:
Screenshot from 2024-03-31 13-14-58

after:

Screenshot from 2024-03-31 13-14-23


matejcik:
included in #3905

PR #3670, opened 2024-04-02
T3T1 Trezorlib fix

grdddj:
Fix variable assignment and enrich error message for unknown model


python/src/trezorlib/models.py
```diff
@@ -96,7 +96,7 @@ class TrezorModel:
 TREZOR_T = T2T1
 TREZOR_R = T2B1
 TREZOR_SAFE3 = T2B1
-TREZOR_T3T1 = T2B1
+TREZOR_T3T1 = T3T1
```
matejcik, line 99: please completely drop this line.
alias "TREZOR T3T1" for just "T3T1" is not useful for anything

grdddj: Right, done in force-push with changed commit message

matejcik (approved):

PR #3671, opened 2024-04-02
refactor(core): improve safety of translation blobs

matejcik:
* the public interface to Translations is now completely safe
* it is more obvious that `map_translated` needs to work the way it does
* documentation is improved

### DIFF
--- a/core/embed/rust/src/translations/flash.rs
+++ b/core/embed/rust/src/translations/flash.rs
@@ -1,28 +1,44 @@
+use spin::{RwLock, RwLockReadGuard};
+
 use crate::{error::Error, trezorhal::translations};
 
 use super::blob::Translations;
 
-static mut TRANSLATIONS_ON_FLASH: Option<Translations> = None;
+static TRANSLATIONS_ON_FLASH: RwLock<Option<Translations>> = RwLock::new(None);
 
+/// Erase translations blob from flash.
+///
+/// The blob must be deinitialized via `deinit()` before calling this function.
 pub fn erase() -> Result<(), Error> {
-    // SAFETY: Looking is safe (in a single threaded environment).
-    if unsafe { TRANSLATIONS_ON_FLASH.is_some() } {
-        return Err(value_error!("Translations blob already set"));
-    }
+    // Write-lock is not necessary but it hints that nobody should call `erase()`
+    // while others are looking.
+    let blob = unwrap!(TRANSLATIONS_ON_FLASH.try_write());
+    {
+        if blob.is_some() {
+            return Err(value_error!("Translations blob already set"));
+        }
 
-    // SAFETY: The blob is not set, so there are no references to it.
-    unsafe { translations::erase() };
+        // SAFETY: The blob is not set, so there are no references to it.
+        unsafe { translations::erase() };
+    }
     Ok(())
 }
 
+/// Write translations blob to flash.
+///
+/// The blob must be deinitialized via `deinit()` before calling this function.
 pub fn write(data: &[u8], offset: usize) -> Result<(), Error> {
-    // SAFETY: Looking is safe (in a single threaded environment).
-    if unsafe { TRANSLATIONS_ON_FLASH.is_some() } {
-        return Err(value_error!("Translations blob already set"));
-    }
+    // Write-lock is not necessary but it hints that nobody should call `erase()`
+    // while others are looking.
+    let blob = unwrap!(TRANSLATIONS_ON_FLASH.try_write());
+    let result = {
+        if blob.is_some() {
+            return Err(value_error!("Translations blob already set"));
+        }
 
-    // SAFETY: The blob is not set, so there are no references to it.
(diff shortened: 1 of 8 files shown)


core/embed/rust/src/translations/translated_string.rs
```diff
@@ -10,16 +10,27 @@ impl TranslatedString {
...
+    ///
+    /// To guarantee that the reference does not escape the closure, we use a
+    /// HRTB of the closure's argument, which ensures that the lifetime of
+    /// the reference is too short to store it outside in any form.
     pub fn map_translated<F, T>(self, fun: F) -> T
     where
         F: for<'a> FnOnce(&'a str) -> T,
-        T: 'static,
     {
         // SAFETY: The bound on F _somehow_ ensures that the reference cannot escape
```
obrusvit, line 30: obsolete commentary

core/embed/rust/src/translations/flash.rs
```diff
@@ -1,28 +1,40 @@
...
 
+/// Erase translations blob from flash.
+///
+/// The blob must be deinitialized via `deinit()` before calling this function.
 pub fn erase() -> Result<(), Error> {
-    // SAFETY: Looking is safe (in a single threaded environment).
-    if unsafe { TRANSLATIONS_ON_FLASH.is_some() } {
-        return Err(value_error!("Translations blob already set"));
-    }
+    let blob = TRANSLATIONS_ON_FLASH.read();
```
mmilata, line 13: Conceptually, shouldn't a write-lock be taken here? Not that it really matters since there's the `is_some` check.

core/embed/rust/src/translations/flash.rs
```diff
@@ -1,28 +1,40 @@
...
 
+/// Write translations blob to flash.
+///
+/// The blob must be deinitialized via `deinit()` before calling this function.
 pub fn write(data: &[u8], offset: usize) -> Result<(), Error> {
-    // SAFETY: Looking is safe (in a single threaded environment).
-    if unsafe { TRANSLATIONS_ON_FLASH.is_some() } {
-        return Err(value_error!("Translations blob already set"));
-    }
+    let blob = TRANSLATIONS_ON_FLASH.read();
```
mmilata, line 29: Also here.

core/embed/rust/src/translations/flash.rs
```diff
@@ -1,28 +1,40 @@
+use spin::{RwLock, RwLockReadGuard};
+
 use crate::{error::Error, trezorhal::translations};
 
 use super::blob::Translations;
 
-static mut TRANSLATIONS_ON_FLASH: Option<Translations> = None;
+static TRANSLATIONS_ON_FLASH: RwLock<Option<Translations>> = RwLock::new(None);
```
mmilata, line 7: Why use a spinlock if we only have one core? Wouldn't a `RefCell` work as well?

core/embed/rust/src/translations/flash.rs
matejcik: I originally thought like, this is a write operation on flash, not on the translations object.

but if we look at the module as a whole, the lock is kind-of for "the whole API" so it would make more sense to enforce that, e.g., only one writer is allowed to write to flash at the same time 

(then again we don't have threads so 🤷‍♀️ )

core/embed/rust/src/translations/flash.rs
matejcik: I'm not sure. maybe it would?
how would you implement get() with a refcell?

mmilata: Something like this?
```rs
    TRANSLATIONS_ON_FLASH
        .try_borrow()
        .map_err(|| value_error!("Translations are in use."))
```
Maybe I'm missing something but from far enough away `RefCell` looks like a single-thread `RwLock` that panics instead of spinning forever.

matejcik: Ah, the issue is, stupidly, that `RefCell` is not `Sync`. The compiler is straight up recommending RwLock instead.
I'm gonna replace all locks with `unwrap!(try_xx)` to make it error out instead of spin

matejcik: a6bec64b71c96129382a08e8a464405ecf9bab3a

core/embed/rust/src/translations/flash.rs
matejcik: a6bec64b71c96129382a08e8a464405ecf9bab3a

core/embed/rust/src/translations/flash.rs
matejcik: a6bec64b71c96129382a08e8a464405ecf9bab3a

core/embed/rust/src/translations/flash.rs
matejcik: because statics are mandatory sync (and also maybe send)

PR #3674, opened 2024-04-03
refactor(core): add ufmt, bump heapless from 0.7.6 - 0.8.0

cepetr:
This small PR adds the [ufmt](https://docs.rs/ufmt/latest/ufmt/) crate for string formatting.

Additionally, the `heapless` crate needed to be upgraded. However, there are some breaking changes: the `String<N>` type no longer implements the `From` trait; it has been replaced by the `TryFrom` trait. Consequently, all occurrences of `from` calls have been replaced by `try_from`, and `unwrap!` has been added where necessary.

**Note: Previous version of heapless used `unwrap()` internally in `from` implementation. So the final behavior remains the same.**

Changes Made:
- Added ufmt crate for string formatting.
- Upgraded heapless crate.
- Replaced occurrences of `from` with `try_from`.
- Added `unwrap!` where necessary.

**This PR does not alter the behavior of the device from the user's perspective.**

FYI: @mmilata, @grdddj, @obrusvit

### DIFF
--- a/core/embed/rust/src/ui/macros.rs
+++ b/core/embed/rust/src/ui/macros.rs
@@ -20,6 +20,6 @@ macro_rules! build_string {
 #[allow(unused_macros)] // not used in TT UI
 macro_rules! inttostr {
     ($int:expr) => {{
-        heapless::String::<10>::from($int).as_str()
+        unwrap!(heapless::String::<10>::try_from($int)).as_str()
     }};
 }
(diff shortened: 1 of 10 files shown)


core/embed/rust/src/ui/macros.rs
```diff
@@ -20,6 +20,6 @@ macro_rules! build_string {
 #[allow(unused_macros)] // not used in TT UI
 macro_rules! inttostr {
     ($int:expr) => {{
-        heapless::String::<10>::from($int).as_str()
+        unwrap!(heapless::String::<10>::try_from($int)).as_str()
```
matejcik, line 23: note (no change requested): with ufmt in, we'll want to get rid of this macro

matejcik (approved):

PR #3679, opened 2024-04-03
Remove unused function

onvej-sl:
This pull request removes `hdnode_public_ckd_address_optimized` from `crypto/bip32.c` which has not been used in the last 5 years.


onvej-sl:
This has already been done in https://github.com/trezor/trezor-firmware/pull/4083.

PR #3683, opened 2024-04-05
refactor(core): simplify secret.h api, hide platform differences

TychoVrahe:
refactors secret.h api

- on U5, secret sector is now hidden in cases firmware has no maximum trust, even if optiga secret is not written
- the BHK unlock is moved from boardloader to boodloader, before the provisioning is made


TychoVrahe:
Consider also implementing:
https://github.com/trezor/trezor-firmware/pull/3581#discussion_r1551840927

core/embed/bootloader/main.c
```diff
@@ -325,33 +325,8 @@ void real_jump_to_firmware(void) {
...
-    secret_optiga_hide();
-  }
-#else
-  if (((vhdr.vtrust & VTRUST_SECRET) != 0) && (sectrue != secret_wiped())) {
-    ui_screen_install_restricted();
-    trezor_shutdown();
-  }
-#endif
-#endif
+  secret_prepare_fw(((vhdr.vtrust & VTRUST_SECRET) == 0) * sectrue,
```
andrewkozlik, line 328: 1. `VTRUST_SECRET` is not documented in docs/core/misc/boot.md.

2. The inversed logic is very confusing and here we inverse it yet again:
    - `(vhdr.vtrust & VTRUST_SECRET) != 0` means that the firmware is not allowed to access the Optiga pairing secret.
    - `trust_secret == sectrue` in `secret_prepare_fw()` means that the firmware is allowed to access the Optiga pairing secret.

    Either the inversed logic should be maintained throughout the code, or better yet, not inverse it in the first place. Why not call it `VTRUST_HIDE_PAIRING_SECRET`?

andrewkozlik: Hmm, it's not just `VTRUST_SECRET` but all of the flags. Like `VTRUST_RED` should be `VTRUST_NO_RED`, etc. 

That brings me to the question, why `VTRUST_SECRET` was not defined to mean that if it is set, then "firmware is allowed to access the Optiga pairing secret"?

andrewkozlik: AFAIK, the whole point behind the inverse logic is that flipping bits 1 -> 0 is easier for an attacker than the other direction. So we make sure that the values that an attacker may want to have are 1s and the the values they don't want are 0s. For example the attacker wants no red screen, hence this is indicated by the mask 0x0010. Same goes for all the other bits, except for `VTRUST_SECRET`, which the attacker wants to be 0. We can't do much about it now, I guess, other than introduce a new bit into the bitfield with the opposite meaning.

core/embed/trezorhal/stm32u5/secret.c
andrewkozlik, line 1: I had some proposals for better function naming [here](https://github.com/trezor/trezor-firmware/pull/3581#discussion_r1553498224), so please consider those here. Namely distinguishing between `secret_hide()` and `secret_optiga_hide()` etc. To make clear what is being hidden, e.g. `secret_hide()` -> `secret_storage_hide()` or `secret_area_hide()`.

TychoVrahe: https://github.com/trezor/trezor-firmware/commit/22a59eba906a2d893fdfa28eba1ebc0ba5d30da2

core/embed/bootloader/main.c
TychoVrahe: @matejcik do you remember the reason for the inverted logic? i think it was something with already signed unsafe vendor headers without the flag, which would otherwise be interpreted as that they are allowed to run with the secret, but i am not sure anymore.

so far i at least renamed the secret.h param: https://github.com/trezor/trezor-firmware/commit/ef867b6758dee3e4cb5263d25935c20e155ad97e

matejcik: Exactly right. We added the field late in the development cycle and there were already signed vendor headers with the wrong value of this bit.
(this is because full-trust image _used to be_ all ones, and low-trust image was all ones except the dropped flags. so we got a low-trust image on our hands where the respective flag was not dropped.)

core/embed/trezorhal/secret.h
```diff
@@ -55,30 +52,15 @@ secbool secret_optiga_set(const uint8_t secret[SECRET_OPTIGA_KEY_LEN]);
...
+void secret_prepare_fw(secbool allow_run_with_secret, secbool trust_all);
 
-// Checks that the optiga pairing secret is present in the secret storage.
-// This functions only works when software has access to the secret storage,
-// i.e. in bootloader. Access to secret storage is restricted by calling
-// secret_hide.
-secbool secret_optiga_present(void);
+// Shows a screen informing the user that installation of custom firmware is
+// currently restricted
+void secret_show_install_restricted_screen(void);
```
cepetr, line 66: It appears to me that this function might be better suited for inclusion in bootui.h/bootui.c, where it was originally located. 

It seems more aligned with the functionalities of the bootloader UI than with this module. Additionally, having this function implemented here leads to three identical implementations for f4, u5, and unix - I'm uncertain if they might diverge in the future.

Perhaps `secret_prepare_fw()` could return a `secbool` to indicate whether `secret_show_install_restricted_screen` needs to be called. This adjustment would help reduce the dependency of `secret.c` on `bootui.c`.

TychoVrahe: The dependency on bootui is one thing that i need to avoid, as this needs to compile in firmware and boardloader too.

The implementation on f4 and u5 is actually different, and basically this should never be shown on U5 models, so i can't really have same logic in bootloader code - the F4 needs to display this screen and shutdown, while U5 can just hides the secret - so returning a secbool does not seem like an option.

The f4 and unix implementation is same, but its also symlink. In future, we might want unix implementation to emulate the behavior of u5 or f4 depending on the model its emulating, so these might diverge too.

core/embed/bootloader/main.c
TychoVrahe: 1. Yeah this is the latest we can change this for T3T1. This is one way adding a new flag would look like, should we go for this? https://github.com/trezor/trezor-firmware/commit/f69da321baf9f3548dad61d6636e59c3e42bb7c4

2. Undefined flags are 1. The dropped flages in low-trust images are 0.

matejcik: 1. what do you think @TychoVrahe ? I can see us using different bits for T2 (F4) and T3+ (U5 and on) platforms, and keep the flipped logic isolated to T2B1 (because T2T1 doesn't care). But I'm not sure about the resulting complexity, I don't have that code in my head right now.

2. It's not (it's not going to be literally all zeroes either as there's a 1 second delay encoded as `1110`) but yes, we should make sure to generate new low-trust images with the reserved bits set to 0

andrewkozlik: 1. Regarding commit https://github.com/trezor/trezor-firmware/commit/f69da321baf9f3548dad61d6636e59c3e42bb7c4, in the json files do we need to have both `dont_provide_secret` and `allow_run_with_secret`? Why not just one of these and define the other bit in `python/src/trezorlib/firmware/vendor.py` as the complement. For older models (just TT) the bits have no meaning, so their value is not important anyway.

    How about making use of the redundancy in both bits, since we will already have them, by defining `VTRUST_SECRET_MASK 0x0180` and `VTRUST_SECRET_ALLOW 0x0100` and testing `(vhdr.vtrust & VTRUST_SECRET) == VTRUST_SECRET_ALLOW`. This should work for T2B1 as well, because the value of `vhdr.vtrust & VTRUST_SECRET` should be 0x0100 for official firmware and 0x0180 for unofficial firmware IIUC (I did not verify this).

2. I think the reserved bits should all be set to 0 regardless of the level of trust. If we want to enable some brand new vendor header feature, then having to sign a new vendor header seems quite acceptable. Meaning that we should change this to `"_reserved" / c.Default(c.BitsInteger(8), 0xff),`:
https://github.com/trezor/trezor-firmware/blob/3d1e2bc20d578c936239f44b4bed4a3aa83eeed6/python/src/trezorlib/firmware/vendor.py#L60

TychoVrahe: ad 1.) doing the inversion automatically would result in generation of different vendor header binaries for T2T1 and T2B1 - if the new VHs were signed at some point and used, this would result in wiping the storage. So i would rather avoid this.

Otherwise, implemented the suggestions in https://github.com/trezor/trezor-firmware/commit/9e423e14d87bf901c36cadd75a00bd7b1c617695 and https://github.com/trezor/trezor-firmware/commit/aad3d09002325e743c54643c44d291fa030f6874 (and reverted unwanted makefile change in https://github.com/trezor/trezor-firmware/commit/f3abaad3eb2637869852df693f730c5ae723a1f3) 

It does indeed work fine on T2B1.

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -199,14 +226,53 @@ secbool secret_optiga_get(uint8_t dest[SECRET_OPTIGA_KEY_LEN]) {
...
+  } else {
+    secret_optiga_uncache();
+  }
+#else
+  secret_disable_access();
+#endif
+
+  if (sectrue != trust_all) {
+    secret_disable_access();
+  }
```
andrewkozlik, line 277: This means that if `allow_run_with_secret` is false and `trust_all` is true, then the firmware will have access to the secret area? I think we don't have any header with this kind of configuration, but the logic seems quite clearly wrong. Also `trust_all` seems quite irrelevant here. I am thinking it should be something like this:

```suggestion
#ifdef USE_OPTIGA
  secret_optiga_uncache();
  if (sectrue == allow_run_with_secret) {
    if (secfalse != secret_optiga_present()) {
      secret_optiga_cache();
      secret_disable_access();
    }
  } else {
    secret_disable_access();
  }
#else
  secret_disable_access();
#endif
```

core/embed/bootloader/main.c
matejcik: > ad 1.) doing the inversion automatically would result in generation of different vendor header binaries for T2T1 and T2B1 - if the new VHs were signed at some point and used, this would result in wiping the storage. So i would rather avoid this.

it wouldn't. we actually can regenerate the unsigned files. the drawback is that then the unsigned and signed files in repository are different (worse auditability) unless we re-sign, and we don't want to re-sign just for this, i think

andrewkozlik: How about just defining a default, so that newer vendor headers have a clear and succinct definition (defining only `allow_run_with_secret`) and the older ones can define `dont_provide_secret` if they need. This would probably be possible by using this code in vendor.py:
 `"dont_provide_secret" / c.Default(c.Flag, lambda this: not this.allow_run_with_secret)`
Also rename `dont_provide_secret` to `_dont_provide_secret` to indicate that it's for special purposes.

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: Hm, i am a bit reluctant about the removal of `trust_all`. The `VTRUST_SECRET` was more meant for the optiga pairing secret, and here on U5 the secret sector contains also bootloader downgrade protection, which i think we never want available for firmwares without `trust_all`. Given the existing combinations of flags in vendor header this is theoretical, but i think stricter checks are better. 

The difference would be that if we ever have vendor header with `allow_run_with_secret=true` and no `trust-all`, then before the optiga secret if written to the flash, firmware with this vendor header could downgrade the bootloader.

TychoVrahe: The rest of the logic is changed here https://github.com/trezor/trezor-firmware/commit/1792b4f1d928a46e731505df40f5b4ab9ede5bfb

core/embed/bootloader/main.c
TychoVrahe: oh yeah it works https://github.com/trezor/trezor-firmware/commit/c219bebbda3779eb300e8f4d25c2146884d308d2

andrewkozlik: We don't actually need to set `_dont_provide_secret` in the files below, because the default is correct:
T2B1/vendor_dev_DO_NOT_SIGN.json
T2B1/vendor_prodtest.json
T2B1/vendor_trezor.json
T2B1/vendor_trezor_btconly.json
However, I leave it to you if you feel it should be set explicitly.

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: I dont think it would be great if prodtest had access to the secret sector always, regardles of pairing secret being present or not - the binary can leak any time thus its no better than official firmware having access.

The second argument is more interesting, not entirely sure its worth it

cepetr (approved):

matejcik (approved):

PR #3684, opened 2024-04-05
set stack protector to strong

TychoVrahe:
Set stack protector to `strong`, for bootloader and boardloader. The benefit of `all` is virtually none and this saves us a lot of flash space.


cepetr (approved):
Since we thoroughly discussed this topic offline, I simply approve it.

PR #3685, opened 2024-04-05
Initialize automatic variables to zero

onvej-sl:
GCC12 introduced [ftrivial-auto-var-init](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ftrivial-auto-var-init) which initializes all automatic variables to zero. Clang already supports it since the version 8 and despite what the [command-line option](https://github.com/trezor/trezor-firmware/pull/3685/files#diff-f27c1e4d684c213c0f1c85dcda569666f6d08d218767b7f95149eb43453f4857R42) say in clang 14, support continues in the subsequent versions. 

In `crypto/**` we follow the rule that we initialize all automatic variables to zero, see for example [this](https://github.com/trezor/trezor-firmware/blob/main/crypto/ecdsa.c#L49). This option could simplify our work since it takes care of this automatically.


onvej-sl:
It would be probably safer to use `-ftrivial-auto-var-init=zero` only for `crypto/**` files.

onvej-sl:
I discovered that emulator fails with segmentation fault if
* you introduce `-ftrivial-auto-var-init=zero` and
* you try to pass a zero-length array from python to c (which happens for example [here](https://github.com/trezor/trezor-firmware/blob/onvej-sl/initialize-variables/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-bech32.h#L48) if the input has 8 characters).

This may indicate there is a use of an uninitialized variable in MicroPython.

hiviah:
With gcc 12.3 this works and builds, but gcc 13.2 fails on micropython build (some dangling pointer error). This means only currently usable gcc is 12.x

Still can't decide if using it for `crypto` only is better, since variables are set to zero, but then it also hides those that have been forgotten.

On the plus side, it does not seem to increase code/FW size, which is a bit surprising. I have to look at the generated assembly and .bss/.data what is the difference more closely.

matejcik:
> but gcc 13.2 fails on micropython build (some dangling pointer error)

do you mean an error within gcc, or some code in micropython that should be fixed?

onvej-sl:
Rebased on top of https://github.com/trezor/trezor-firmware/pull/3693, which fixes the failing core unit test.

onvej-sl:
> With gcc 12.3 this works and builds, but gcc 13.2 fails on micropython build (some dangling pointer error). This means only currently usable gcc is 12.x

If you're referring to the following error, it doesn't depend on the presence of the `-ftrivial-auto-var-init=zero` flag.

```
vendor/micropython/py/stackctrl.c: In function 'mp_stack_ctrl_init':
vendor/micropython/py/stackctrl.c:32:32: error: storing the address of local variable 'stack_dummy' in 'mp_state_ctx.thread.stack_top' [-Werror=dangling-pointer=]
   32 |     MP_STATE_THREAD(stack_top) = (char *)&stack_dummy;
vendor/micropython/py/stackctrl.c:31:18: note: 'stack_dummy' declared here
   31 |     volatile int stack_dummy;
      |                  ^~~~~~~~~~~
In file included from vendor/micropython/py/runtime.h:29,
                 from vendor/micropython/py/stackctrl.c:27:
vendor/micropython/py/mpstate.h:323:23: note: 'mp_state_ctx' declared here
  323 | extern mp_state_ctx_t mp_state_ctx;
      |                       ^~~~~~~~~~~~
```

matejcik:
and that warning is wrong in this particular case. so we'll need to fix that for micropython+gcc13 anyway.
i say we go ahead with this PR, unless there are other problems?

onvej-sl:
I think I can see just one problem, which is somehow  related to what @hiviah mentioned. Consider this code:
```c
  int x;
  while (x < 10) {
    printf("%d", x);
    x++;
  } 
```
The code is completely fine if you compile it with `-ftrivial-auto-var-init=zero`, but it leads to undefined behavior when compiled without it. In other words, the problem arises when you forget to initialize a variable and someone compiles it without using `-ftrivial-auto-var-init=zero`. However, the compiler should issue a warning if this occurs.

Which c files besides `vendor/crypto/**` do we want to compile with the flag?
  * `vendor/micropython/**`
  * `core/embed/**`
  * `legacy/**`

matejcik:
> the problem arises when you forget to initialize a variable and someone compiles it without using `-ftrivial-auto-var-init=zero`

change it to `pattern` to force us to initialize everything explicitly? it's still defensive but is more likely to crash than to magically work?

hiviah:
After bindiffing and dissassembling both versions, while code+bss has same size, the order of symbols is different:

image

Compared the disassembly also with gdb to make sure I'm seeing the data right.

For now probably we should use `-ftrivial-auto-var-init=zero` only for `crypto/** files`

hiviah:
> change it to `pattern` to force us to initialize everything explicitly? it's still defensive but is more likely to crash than to magically work?

In some forums it was suggested to use different pattern from 0, that does not come up often, so it's easier to spot with eyes.

Then if some uninitialized var is suspected, just put `watch` on it to see if it's never written, or `rwatch` in gdb or other debugger to see that it's read before being changed.

onvej-sl:
* https://github.com/trezor/trezor-firmware/pull/3685/commits/2772f5943e3074e03f814fea9b1fbe25a33fd61d was already merged in https://github.com/trezor/trezor-firmware/pull/3693.
* https://github.com/trezor/trezor-firmware/pull/3685/commits/a2dec3ffbef7c0ba6dde6c9d9432f998b306e8bd updates to GCC 12. Though 13 s the highest version of GCC on nixos stable channel, the micropython build fails with it (see https://github.com/trezor/trezor-firmware/pull/3685#issuecomment-2045132945) Therefore, I only updated to GCC 13.
* https://github.com/trezor/trezor-firmware/pull/3685/commits/a7960d29f1ae679b291cd29154422e714e7d6bef updates to Clang 17. This is the highest version of Clang on nixos stable channel. I didn't update other packages, because I encountered some issues, especially with the updated protobuf package.
* https://github.com/trezor/trezor-firmware/pull/3685/commits/54f65c03f800d68ac367a037fe259dbf23fb469d fixes warnings/errors that are newly emitted by Clang 17.
* https://github.com/trezor/trezor-firmware/pull/3685/commits/022bbfdc4e8eab9fc1c3e6eabebcfe4ca4241a31 (and fixups https://github.com/trezor/trezor-firmware/pull/3685/commits/b31e2308b469775082eceb13bf734115fc262a0c, https://github.com/trezor/trezor-firmware/pull/3685/commits/a0ab3984b37b9a3faf1cfc91f3ae98fdfcf4fe69) initializes automatic variables to zero in crypto tests, excluding fuzzing tests. The fuzzing tests use an older version of Clang which doesn't understand the flag. It would be better to update the fuzzing machine, but I don't know how to do it.
* https://github.com/trezor/trezor-firmware/pull/3685/commits/086e8bf050aa8da4c227c8e2238d28766407e7fc, https://github.com/trezor/trezor-firmware/pull/3685/commits/7b960d16c3cf71ba54e9fff67643f9dce838eb78, https://github.com/trezor/trezor-firmware/pull/3685/commits/2ac27e22cbd6bce886b71bfd337319b2dc77744d initialize automatic variables to zero in all legacy and core builds (including boardloader and bootloader).

matejcik:
we have an open issue #3337 for nixpkgs update, so I'm blocking this until @mmilata can take a look
in general it looks nice and the CI didn't horribly break so :+1: 

with the new clang it looks like we need to reformat some C sources though

mmilata:
LGTM from nixpkgs side, it would be nicer if we didn't use multiple nixpkgs versions but I don't want to hold up the PR just for this reason. I propose to merge it and I'll deal with `newNixpkgs` as part of #3412 (which is still not ready as I keep running into new reasons the updated toolchain breaks our build/tests ...).

matejcik (approved):
perfect :100: 
@onvej-sl please resolve conflicts and merge if you can

matejcik:
oh and `make cstyle` might be needed

PR #3688, opened 2024-04-07
Bootloader UI for T3T1

### LINKED ISSUE #3681: Device startup
**Design:** [Figma](https://www.figma.com/file/n43caWJFGJlT167AQcxruu/%E3%80%94-Prod-%E3%80%95Safe-5?type=design&node-id=2128-1346&mode=design&t=xfDKG77mtBsFChev-4)

mmilata:
Fixes #3681. Any hints where to change the darker loader color?

emu-lockfull emu-welcome emu-installing emu-restarting

Other screens are unchanged, but not really because the fonts are different, I suppose we'll have to do something with the intro screen:

emu-menu emu-intro


core/embed/lib/fonts/fonts.h
```diff
@@ -110,10 +119,16 @@
...
 #endif
 
+#ifdef TREZOR_FONT_SUB_ENABLE
+#define FONT_MAX_HEIGHT_6 MAX_FONT_H(FONT_SUB_MAX_HEIGHT, FONT_MAX_HEIGHT_5)
+#else
+#define FONT_MAX_HEIGHT_6 FONT_MAX_HEIGHT_5
+#endif
+
 #ifdef TREZOR_FONT_MONO_ENABLE
 #define FONT_MAX_HEIGHT MAX_FONT_H(FONT_MONO_MAX_HEIGHT, FONT_MAX_HEIGHT_5)
```
obrusvit, line 129: There should be also `FONT_MAX_HEIGHT_6` on this line.

mmilata: nice catch, fixed in b9690a38e3f8585a230f43405acf1b3e3447c6ef

mmilata:
Rebased, I've seen some weird failures of tests with `de-DE` translations, might need to investigate if they show up again, otherwise this should be ready for review.

The fonts used are currently `Font_TTSatoshi_DemiBold_21` and `Font_TTHoves_Bold_17`, let me know if you want me to change them.

emu00000000(1)

mmilata:
Changed the stage1 logo to full lock, reimported from Figma since it's 5px higher and doesn't have to match its counterpart anymore.

core/embed/rust/src/ui/display/loader/circular_thin.rs
```diff
@@ -0,0 +1,335 @@
+use crate::ui::{
+    constant,
+    constant::{screen, LOADER_INNER, LOADER_OUTER},
+    display,
+    display::{toif::Icon, Color},
+    geometry::{Offset, Point, Rect},
+};
+
+#[cfg(feature = "dma2d")]
```
matejcik, line 9: you can drop all gates on `dma2d` if this file is only included with dma2d enabled

core/embed/rust/src/ui/display/loader/circular_thin.rs
```diff
@@ -0,0 +1,335 @@
+use crate::ui::{
```
matejcik, line 1: aside: this whole file is a copy-paste job and probably shouldn't be. but we don't want to spend that time in this PR.

does the new drawing library do something about it? @cepetr

core/embed/rust/src/ui/constant.rs
```diff
@@ -1,7 +1,13 @@
...
+pub use super::model_mercury::constant::*;
+#[cfg(all(
+    feature = "model_tr",
+    not(feature = "model_tt"),
+    not(feature = "model_mercury")
+))]
 pub use super::model_tr::constant::*;
-#[cfg(feature = "model_tt")]
+#[cfg(all(feature = "model_tt", not(feature = "model_mercury")))]
 pub use super::model_tt::constant::*;
```
matejcik, line 13: The model selector needs to be spelled the same way every time.
When all features are enabled, this selects model_mercury for constants, but model_tt for ui-features.

core/embed/rust/Cargo.toml
```diff
@@ -10,6 +10,7 @@ default = ["model_tt"]
 crypto = ["zeroize"]
 model_tt = ["jpeg"]
 model_tr = []
+model_mercury = ["jpeg"]
```
matejcik, line 13: not also `dma2d` ?

core/embed/rust/src/ui/model_mercury/bootloader/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            None,
+        )
+    }
+
+    fn screen_install_progress(progress: u16, initialize: bool, initial_setup: bool) {
+        let bg_color = if initial_setup { WELCOME_COLOR } else { BLD_BG };
+        let fg_color = if initial_setup { GREEN_LIGHT } else { BLD_FG };
+        let icon_color = BLD_FG;
+
+        ModelMercuryFeatures::screen_progress(
```
matejcik, line 285: this should be `Self::xyz` in all usages (inside `impl [X for] ModelMercuryFeatures`)

(same holds for the other models of course)

core/embed/rust/src/ui/model_mercury/component/result.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+}
+
+impl<'a> Component for ResultScreen<'a> {
+    type Msg = Never;
+
+    fn place(&mut self, _bounds: Rect) -> Rect {
+        self.bg.place(screen());
+
+        let (main_area, footer_area) = ResultFooter::<'a>::split_bounds();
```
matejcik, line 138: bump your version of `result.rs` and `error.rs` to get rid of these lifetimes

core/embed/rust/src/ui/display/loader/circular_thin.rs
cepetr: @matejcik: with the new drawing library, there will be no more "circular_xxx.rs" files; these "loaders" are implemented as just a few lines of code directly in place

core/embed/rust/src/ui/constant.rs
mmilata: uh, good catch, 9f4dc34cb76dfeb25f98d31258977bad3d35e112

core/embed/rust/src/ui/model_mercury/component/result.rs
mmilata: 9f4dc34cb76dfeb25f98d31258977bad3d35e112

core/embed/rust/src/ui/display/loader/circular_thin.rs
mmilata: 9b181470c51a8e07565e5474d9730546c064be98

mmilata:
Fonts fixed in 8eccd55e2fa6268bdebde44c5a643d0bbda3f9ac. Currently `font_ttsatoshi_demibold_21` is NORMAL, BOLD, and DEMIBOLD at the same time - the C font machinery doesn't mind but i wonder if there will be three copies in the translation blob?

`.png` files moved to `core/assets/model_mercury` in 5c175f28a868821d382dd16f94871938109cfeeb

matejcik:
> `.png` files moved to `core/assets/model_mercury` in [5c175f2](https://github.com/trezor/trezor-firmware/commit/5c175f28a868821d382dd16f94871938109cfeeb)

do we even need this, now that we have toiftool? we can extract the png at any time

matejcik:
what if we excluded the font changes for _firmware_ from this, which allows us to not mess with the "skip for german" weirdness?

core/embed/rust/src/ui/model_mercury/component/button.rs
```diff
@@ -0,0 +1,440 @@
...
+#[derive(Clone, Copy)]
+pub enum CancelInfoConfirmMsg {
+    Cancelled,
+    Info,
+    Confirmed,
+}
+
+#[derive(PartialEq, Eq)]
+pub struct IconText {
+    text: &'static str,
```
obrusvit, line 381: I would find it beneficial to have
```
pub struct IconText {
    text: TString<'static>,
    ...
```
because of menu items in `VerticalMenu`. 
We can do so in another PR though.

TychoVrahe:
Upon rebase, please incorporate changes done to model_tt ui in https://github.com/trezor/trezor-firmware/pull/3708 to mercury as well

core/embed/rust/src/ui/model_mercury/component/button.rs
mmilata: bootloader currently doesn't use `VerticalMenu`, let's do it in #3585

Hannsek:
> yes, the one in T3T1 figma is 5px higher than the one in model T firmware

So only the position is different, not the logo itself. That is fine.

matejcik (approved):

mmilata:
> So only the position is different, not the logo itself. That is fine.

No, the height of the TS5 logo is 5px smaller than the one on TT. Previously I mixed it up and said it's the other way, sorry. Left is screenshot from TT emulator, right is Figma export.

emu00000000

PR #3696, opened 2024-04-09
Unicode font uppercasing issues

obrusvit:
`some_text.upper()` calls do not work correctly in languages other than english. This PR adds additional `Font`s which are UPPERCASE only. These fonts are used where appropriate, especially in Titles and Buttons.

- For Model T, this meant only switching `FONT_BOLD` to `FONT_BOLD_UPPER`
- For Model R, this meant adding additional fonts `FONT_NORMAL_UPPER` and `FONT_BOLD_UPPER` and use them where appropriate
- For both models: some diffs emerged, listed below.

TODO:
- [x] do not duplicate glyphs for normal and `_upper` fonts
- [x] change other languages and it's fixtures
  - [x]  Spanish `es`
  - [x] German `de`
  - [x] French `fr`
  - [x] Czech `cs`


obrusvit:
One of the issues is that `altcoin_tx_summary` misuses it's `title` for `Amount:`. 
image

Hannsek:
All summary screens use that. Can it be changed back?

obrusvit:
> All summary screens use that. Can it be changed back?

The plan is to have no visual changes with this PR so yeah, it will be changed back.

obrusvit:
State:
Model R:
- `show_warning` was sometimes called with `upper()`cased title and sometimes not, so using upper font produces diffs. 
image
- (ab)use of `title` in `altcoin_tx_summary` as shown in previous comment 
image
- cases where it actually fixes a problem, e.g. here:
image
- some asserts in tests (`'Confirm' vs 'CONFIRM')

Model T:
- cases where it fixes a problem, e.g. here:
image

core/embed/lib/fonts/font_tthoves_bold_upper_17.c
```diff
@@ -0,0 +1,205 @@
...
+/* ^ */ static const uint8_t Font_TTHoves_Bold_upper_17_glyph_94[] = { 8, 5, 8, 0, 12, 0, 251, 159, 0, 16, 255, 255, 0, 112, 207, 253, 5, 192, 111, 248, 10, 242, 31, 242, 31 };
+/* _ */ static const uint8_t Font_TTHoves_Bold_upper_17_glyph_95[] = { 9, 3, 8, 0, 0, 82, 85, 85, 85, 0, 245, 255, 255, 255, 1, 245, 255, 255, 255, 1 };
+/* ` */ static const uint8_t Font_TTHoves_Bold_upper_17_glyph_96[] = { 5, 3, 5, 0, 13, 114, 23, 0, 176, 175, 0, 0, 253, 5 };
+/* a -> A */ static const uint8_t Font_TTHoves_Bold_upper_17_glyph_97[] = { 13, 12, 13, 0, 12, 0, 0, 243, 255, 31, 0, 0, 0, 0, 248, 255, 111, 0, 0, 0, 0, 253, 255, 191, 0, 0, 0, 48, 255, 168, 255, 1, 0, 0, 128, 255, 83, 255, 6, 0, 0, 224, 239, 0, 255, 12, 0, 0, 243, 159, 0, 251, 31, 0, 0, 249, 255, 255, 255, 111, 0, 0, 254, 255, 255, 255, 207, 0, 48, 255, 107, 102, 198, 255, 1, 144, 255, 5, 0, 112, 255, 7, 224, 255, 0, 0, 32, 255, 12 };
```
mmilata, line 76: Why not just generate `static const uint8_t Font_TTHoves_Bold_upper_17_glyph_97 = Font_TTHoves_Bold_upper_17_glyph_65;`. Seems like easy win of couple KiB, especially in bootloader, but maybe I'm missing some downside.

core/SConscript.firmware
```diff
@@ -42,12 +42,16 @@ if TREZOR_MODEL in ('1', 'R'):
     FONT_BOLD='Font_PixelOperator_Bold_8'
     FONT_MONO='Font_PixelOperatorMono_Regular_8'
     FONT_BIG='Font_Unifont_Regular_16'
+    FONT_NORMAL_UPPER='Font_PixelOperator_Regular_upper_8'
+    FONT_BOLD_UPPER='Font_PixelOperator_Bold_upper_8'
```
mmilata, line 46: With both normal and upper fonts included in this model the glyph data is duplicated, and I don't think the compiler can handle that for us. Both are ~8KiB which I'm not sure we can afford.

core/SConscript.bootloader
```diff
@@ -36,18 +36,24 @@ if TREZOR_MODEL in ('R', ):
     FONT_BOLD='Font_PixelOperator_Bold_8'
     FONT_MONO='Font_PixelOperator_Regular_8'
     FONT_BIG=None
+    FONT_NORMAL_UPPER=None
+    FONT_BOLD_UPPER=None
 elif TREZOR_MODEL in ('T', 'DISC1', 'DISC2'):
     FONT_NORMAL='Font_TTHoves_Regular_21'
     FONT_DEMIBOLD='Font_TTHoves_Regular_21'
-    FONT_BOLD='Font_TTHoves_Bold_17'
+    FONT_BOLD=None
```
mmilata, line 44: IIRC if such font is accidentally used it is not caught at compile time but crashes at first rendering. I wonder if changing this to `FONT_BOLD='Font_TTHoves_Bold_upper_17'` could save us some future bootloader headaches.

core/embed/lib/fonts/font_tthoves_bold_upper_17.c
obrusvit: You are right, now here only upper case letters are generated and they are referenced in the font array below
e43624f2d6f9f80ed624123d9c30d9ea325b239a

This saves some space for model T. 
Regarding model R, it helps a bit but a lot of it is still duplicated, you are right.
But what is the memory situation there? How do I know?

obrusvit: We can modify the generator further to link `_upper` fonts to the existing font though. :thinking:

core/SConscript.firmware
obrusvit: See comment above.

core/SConscript.bootloader
obrusvit: Can't we just use `FONT_BOLD_UPPER` in the bootloader? I'm not sure I follow.

core/embed/lib/fonts/font_tthoves_bold_upper_17.c
obrusvit: Is this job accurate? If it is, then we should be fine. https://github.com/trezor/trezor-firmware/actions/runs/8884320731/job/24393365753?pr=3696
image

core/SConscript.bootloader
matejcik: are we now perhaps ready to get rid of `FONT_BOLD` etc. constants and redefine them to the target symbols in Rust?

matejcik: > we might be, unless there's be a conflict with #3695

i don't think there would be -- and if it is, we'll resolve it

@obrusvit can you give it a try?

obrusvit: @matejcik yes
I will modify the generator further to reuse the same glyphs even in TS3 and then take a look at the constants.

core/embed/lib/fonts/font_tthoves_bold_upper_17.c
obrusvit: `gen_font.py` modified here 8547def0d12d63b7de9910bb9450478fd650b78f 

Now, the glyphs are stored only once and the font arrays point to the same glyphs. Nonprintable glyph is also stored only once and `_upper` version is a pointer to the glyph. The result in `.h` files look like this:

```
extern const uint8_t* const Font_PixelOperator_Regular_8[126 + 1 - 32];
extern const uint8_t Font_PixelOperator_Regular_8_glyph_nonprintable[];

extern const uint8_t* const Font_PixelOperator_Regular_8_upper[126 + 1 - 32];
extern const uint8_t Font_PixelOperator_Regular_8_upper_glyph_nonprintable[];
```

core/SConscript.bootloader
obrusvit: The generator is modified as described in a thread above.

As for this:
> are we now perhaps ready to get rid of FONT_BOLD etc. constants and redefine them to the target symbols in Rust? 

We agreed with matejcik that this is out of scope of this PR.

I would like to rebase this on `main` and do a 2nd round of review - necessary because of conflicts in translated strings.

obrusvit:
A lot of diff in fixtures due to many minor changes in languages as described in the [previous comment](https://github.com/trezor/trezor-firmware/pull/3696#issuecomment-2056401799).

Also some previous changes were not incorporated into fixtures from #3617 

Some UI fixtures were missing for TS3, e.g. some tests of wipe code.

core/tools/codegen/gen_font.py
```diff
@@ -288,25 +286,36 @@ def _h_file_name(self) -> Path:
 
     def write_files(self) -> None:
         self.write_c_files()
-        self.write_foreign_json()
+        if self.gen_normal:
+            self.write_foreign_json(upper_cased=False)
+        if self.gen_upper:
+            self.write_foreign_json(upper_cased=True)
```
mmilata, line 292: Nit: is there any advantage in generating the character twice?

core/embed/rust/src/ui/model_tt/theme/mod.rs
```diff
@@ -158,21 +158,21 @@ pub const fn label_progress() -> TextStyle {
 }
```
mmilata, line 159: Can you please also fix `theme/bootloader.rs` and `component/result.rs`?

mmilata (changes requested):
A few things but otherwise looking very good. T3T1 fixtures are missing.

core/embed/rust/src/ui/model_tt/theme/mod.rs
obrusvit: regarding `theme/bootloader.rs` is that necessary since we linked `FONT_BOLD` to `_upper` font as well?

core/tools/codegen/gen_font.py
obrusvit: I don't see an easy way how to map the same foreign glyphs from one json without changing the mechanics too much. So now if both regular and `_upper` is used, the foreign glyps are duplicated.

core/tools/codegen/gen_font.py
```diff
@@ -373,49 +405,78 @@ def _write_char_definition(self, f: TextIO, c: str, i: int) -> None:
...
+                f.write(
+                    f"extern const uint8_t* const Font_{self._name_style_size}_upper[{MAX_GLYPH} + 1 - {MIN_GLYPH}];\n"
+                )
+                f.write(
+                    f"extern const uint8_t Font_{self._name_style_size}_upper_glyph_nonprintable[];\n"
+                )
+
+    def _is_lower_case(self, i: int) -> bool:
+        """Check if the integer i corresponds to an ASCII lower-case letter."""
+        return 97 <= i <= 122  # ASCII values for 'a' to 'z'
```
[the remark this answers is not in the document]
obrusvit: https://github.com/trezor/trezor-firmware/pull/3696/commits/2fdee0fd183c32377e25c63c072b4a81b0d710e6

core/embed/rust/src/ui/model_tt/theme/mod.rs
obrusvit: Anyway, fix here eb495f2b6c6d2a72d05661132cc248ed4c045d9c

obrusvit:
T3T1 fixtures also updated @mmilata https://github.com/trezor/trezor-firmware/pull/3696/commits/f53f8ab718b8e518ca28a5c6ec4a499c6470e234

mmilata (approved):
LGTM, feel free to rebase and merge if CI passes.

bosomt:
**QA OK**

tested on Czech translation

 - Device: Trezor T2T1 2.7.2 regular (revision da75d8f4b67410b40a9cfd2954d183d81dd6e8e8)

PR #3699, opened 2024-04-10
fix(python): missing PIL.Image.ANTIALIAS

mmilata:
Image.ANTIALIAS was [removed in 10.0.0](https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#constants). 10.0.0 was released on Jul 1, 2023.

### DIFF
--- a/python/requirements-optional.txt
+++ b/python/requirements-optional.txt
@@ -1,5 +1,5 @@
 hidapi>=0.7.99.post20
 web3>=5
-Pillow
+Pillow>=10
 stellar-sdk>=6
 rlp>=1.1.0 ; python_version<'3.7'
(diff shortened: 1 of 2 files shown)


python/requirements-optional.txt
```diff
@@ -1,5 +1,5 @@
 hidapi>=0.7.99.post20
 web3>=5
-Pillow
+Pillow>=10
```
matejcik, line 3: this is also missing from setup.py, see 3c22a09f9d8c67ac68490f69925a0d471122d874

matejcik: i mean, it might still make sense to merge this separately from GFL which is presumably going to take longer to land

PR #3701, opened 2024-04-10
feat(core/prodtest): add commands TOUCH VERSION and VARIANT READ

TychoVrahe:
Adds TOUCH VERSION and VARIANT READ commands to prodtest, so that production can check the version of firmware in touch controller and stored device variant.

### DIFF
--- a/core/embed/trezorhal/stm32f4/touch/ft6x36.c
+++ b/core/embed/trezorhal/stm32f4/touch/ft6x36.c
@@ -275,3 +275,21 @@ uint32_t touch_read(void) {
 
   return 0;
 }
+
+uint8_t touch_get_version(void) {
+  uint8_t version = 0;
+  uint8_t outgoing[] = {0xA6};  // start reading from address 0xA6
+  int result = i2c_transmit(TOUCH_I2C_INSTANCE, TOUCH_ADDRESS, outgoing,
+                            sizeof(outgoing), 1);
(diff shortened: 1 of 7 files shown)


core/embed/trezorhal/stm32f4/touch/ft6x36.c
```diff
@@ -275,3 +275,21 @@ uint32_t touch_read(void) {
 
   return 0;
 }
+
+uint8_t touch_get_version(void) {
+  uint8_t version = 0;
+  uint8_t outgoing[] = {0xA6};  // start reading from address 0xA6
+  int result = i2c_transmit(TOUCH_I2C_INSTANCE, TOUCH_ADDRESS, outgoing,
+                            sizeof(outgoing), 1);
+  if (result != HAL_OK) {
```
cepetr, line 284: Calling i2c_cycle() here is quite suspicious. I think we should investigate whether it is truly necessary in this context and if it could potentially offer any benefits. However, it's not a blocker for approving this PR.

cepetr (approved):

core/embed/trezorhal/stm32f4/touch/ft6x36.c
TychoVrahe: I agree. Created a separate issue for that: https://github.com/trezor/trezor-firmware/issues/3706

PR #3702, opened 2024-04-10
Restartable timers

### LINKED ISSUE #3692: T2B1: Wrong behavior when auto-mover hits a wall
This would mostly be observed in "select number of words" where there is no wrap-around.

Hold down an arrow and let the cursor hit the other edge of the selection (e.g. start at 12 and let it go to 33).
Release the button.
Hold down the other direction button.

Expected: cursor starts moving back
Actual: nothing. When you release the button, it will move one place.

The auto-mover somehow moves "behind the selection" instead of stopping, and the opposite button press causes it to "stop" instead of starting to move.

matejcik:
fixes "Queue overflow" error in T2B1 click tests which happens if (a) your CPU is fast enough or (b) you improve performance of UI code too much (also supersedes #3363 which could not be merged due to the speed improvement triggering this problem)

### Summary of problem

Every button click in some screens will fire off a new timer event. Previously, there was no way to unschedule those timer events, so if you sent events too fast, there would be too many timers and they would overflow the queue

### Summary of solution

A same logical timer will now reuse its event if it is already scheduled.
The performance cost amortizes to zero -- you spend work upfront to unschedule the event, but you save work of processing an already scheduled event which will be ignored in the end.
This means that every button click will schedule the same timer over and over again, saving precious queue space.

### Quality of life

`Timer` is now a nicer struct member in that the user doesn't need to manually manage their `Option<TimerToken>`, and internally handles starting, stopping, and restarting.

### Other changes

* Code from #3633 is included here to massively speed up emulator tests: be8eb9c7476f3d918396a02621f04749e27eedbd + 1c87baf919d8d48624dafa289a7b7b8e5605f14f
* ~fixes #3692 519dc9f817477ceb06499973ad16d9a03ba79ce9~
* ~now that we have `ufmt`, I re-did `println!` macros to work the same way as the real thing, and introduced `uformat!` macro that works a lot like the real thing. Use this in your UI code with abandon. d5e66517c175e2501065210d091160fb54dd8652~
* ~refactor of `IgnoreButtonDelay` which was needlessly using timers, and also was confusingly mixed into the already confusing state machine of ButtonController 7f5401b82df5389953c3522b946ab61761b1568c~

(removed unrelated changes, a separate PR will follow)


matejcik:
TODO: I again forgot to add a changelog (this time for what was #3633 which is an impactful change that _could_ plausibly affect users)

Hannsek:
> impactful change that could plausibly affect users

How?

matejcik:
see the changelog:

> Improved device responsiveness by removing unnecessary screen refreshes.

if I missed a refresh somewhere, someone will see it as bad screen response

mmilata (approved):

obrusvit:
utACK

obrusvit (approved):
utACK

PR #3708, opened 2024-04-11
fix T3T1 unsafevendor header and error screen displaying

TychoVrahe:
The toif image in unsafe vendor header was faulty, resulting in inability to run fw with that VH.

Error screen had missing refresh, and when called from rust the backlight was not ensured. Along with adding the refresh i removed duplicates on places that are calling the rust screen.

Building on #3683 , because that also changes vendor headers, but we also need to modify `mercury` UI in #3688

### DIFF
--- a/core/embed/trezorhal/stm32f4/common.c
+++ b/core/embed/trezorhal/stm32f4/common.c
@@ -53,6 +53,7 @@ uint32_t systick_val_copy = 0;
 extern void shutdown_privileged(void);
 
 void __attribute__((noreturn)) trezor_shutdown(void) {
+  display_finish_actions();
 #ifdef USE_SVC_SHUTDOWN
   svc_shutdown();
 #else
@@ -69,9 +70,7 @@ error_uni(const char *label, const char *msg, const char *footer) {
   display_orientation(0);
 
 #ifdef FANCY_FATAL_ERROR
-
   screen_fatal_error_rust(label, msg, "PLEASE VISIT\nTREZOR.IO/RSOD");
-  display_refresh();
 #else
   term_set_color(COLOR_WHITE, COLOR_FATAL_ERROR);
   if (label) {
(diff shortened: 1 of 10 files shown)


matejcik:
utACK
out of horrible curiosity, how did you manage to produce the previous vendor header?

TychoVrahe:
the vendor header is generated by the script without any issues from the faulty toif. The toif had added newline at the end, i guess IDE did that when i opended the file

matejcik:
> the vendor header is generated by the script without any issues from the faulty toif. The toif had added newline at the end, i guess IDE did that when i opended the file

no, the original toif is twice as big with EF BF BD inserted liberally throughout
image
apparently EF BF BD is the replacement character encoded in utf8, so the IDE must have gone quite wild on the file :grin: 
anyway, glad to see what went wrong here

matejcik (approved):

TychoVrahe:
hmm yeah so something like i opened the file in IDE like text file, it inserted newline (this is the only difference my CLion shows when comparing the two files as text) and reencoded in utf8..

core/embed/trezorhal/stm32f4/common.c
```diff
@@ -122,6 +120,7 @@ __fatal_error(const char *expr, const char *msg, const char *file, int line,
               rev[4]);
 #endif
   term_printf("\nPlease contact Trezor support.\n");
+  display_refresh();
```
cepetr, line 123: `term_printf()` calls `display_refresh()` itself -  so it's not needed here, as well as at few other places in this file and `stm32u5/common.c`.

cepetr (changes requested):

core/embed/trezorhal/stm32f4/common.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1e3d5a49abf84dcbde52aab538f3ccc73a2ec16b

cepetr (approved):

PR #3709, opened 2024-04-11
only allow confirm-less firmware installation of full-trust images

TychoVrahe:
Applies to empty devices only.

Now even when empty, installing fw without full trust requires user confirmation.


core/embed/bootloader/messages.c
```diff
@@ -649,13 +649,20 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
...
 
       uint32_t response = INPUT_CANCEL;
-      if (sectrue == is_new || sectrue == is_ilu) {
+      if (((vhdr.vtrust & VTRUST_ALL) == VTRUST_ALL) &&
+          (sectrue == is_new || sectrue == is_ilu)) {
         // new installation or interaction less updated - auto confirm
+        // only allowed for full-trust images
         response = INPUT_CONFIRM;
       } else {
-        int version_cmp = version_compare(hdr.version, current_hdr->version);
```
cepetr, line 656: How about this simplification:
```c
      if (((vhdr.vtrust & VTRUST_ALL) == VTRUST_ALL) &&
          (sectrue == is_new || sectrue == is_ilu)) {
        // new installation or interaction less updated - auto confirm
        // only allowed for full-trust images
        response = INPUT_CONFIRM;
      } else {
          int version_cmp = version_compare(hdr.version, current_hdr->version);
          response = ui_screen_install_confirm(
              &vhdr, &hdr, should_keep_seed || is_new, is_newvendor, is_new, version_cmp);
      }
```

cepetr (changes requested):

cepetr:
I don't see any problem with this change. Just add a note on how we can simplify the code a bit.

core/embed/bootloader/messages.c
TychoVrahe: Rebased.

First, in the original version there was a problem of using `true` instead of `sectrue`.

Thus, the proposed code does `||` on `secbool` which is wrong. Also, the `current_hdr` is null when is_new is set so there is null pointer deref. 

With this taken into account, here is updated version https://github.com/trezor/trezor-firmware/pull/3709/commits/1148b43f91e88f601a9f7b1cd3f8e2cd89674869

core/embed/bootloader/bootui.c
```diff
@@ -210,7 +210,7 @@ void ui_screen_boot(const vendor_header *const vhdr,
   uint32_t vimg_len = *(uint32_t *)(vhdr->vimg + 8);
 
   screen_boot(red_screen, vendor_str, vendor_str_len, hdr->version, vhdr->vimg,
-              vimg_len, wait);
+              vimg_len + 12, wait);
```
cepetr, line 213: It's quite unfortunate that we have to parse the TOIF header here. However, if it's necessary, it would be better to be explicit about what we're doing.

What about this:
```c
#define TOIF_LENGTH(ptr)  *(uint32_t *)((ptr) + 8) + 12;

uint32_t vimg_len = TOIF_LENGTH(vhdr->vimg)
screen_boot(red_screen, vendor_str, vendor_str_len, hdr->version, vhdr->vimg,
                 vimg_len, wait);
```c

Some checks are still missing here, but if the header integrity is checked beforehand, it might be okay.

cepetr (changes requested):

core/embed/bootloader/bootui.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/5197e2b370d800ea81de572e92ea8cfaaafafb94

matejcik (approved):

cepetr (approved):
utACK

Issue #1, opened 2018-03-05
Redesign Passphrase


slush0:
Make the workflow same for client for both non-passphrase device and passphrase device. Currently, client receives state only when passphrase is enabled.

prusnak:
Related to Protocol redesign: https://github.com/trezor/trezor-firmware/issues/79

andrewkozlik:
Some good points about passphrase design were made here: https://www.reddit.com/r/Bitcoin/comments/bp0nqs/trezor_or_ledger_which_one_is_better/enof517?utm_source=share&utm_medium=web2x

To summarize:
- Make it possible to easily switch between passphrases on the fly.
- Get rid of enable/disable passphrase as a configuration option. There should be no indication of whether in the past the device was used with a (non-empty) passphrase or not.
- Use a 2nd PIN to unlock the device and automatically enter a preconfigured passphrase.

I think the first two points can be solved by adding a "Switch passphrase" button to the web wallet. As for the third point, I actually think that this is in contradiction with plausible deniability. For one thing, there is likely to be some side-channel information, which could reveal whether the primary or secondary PIN was entered. So in general I am not too keen on supporting multiple PINs. However, I would like to make it possible to configure a default passphrase which would be automatically used upon start-up. The reason is that people may want to use a passphrase for various reasons:
- Plausible deniability, i.e. to have a real wallet and a decoy wallet.
- Providing additional protection, because they do not trust the PIN and hardware to protect their seed if the wallet is stolen.
- Providing additional protection in case their paper backup is stolen.
Say you are worried about protecting your paper backup but you trust the hardware and PIN to provide sufficient protection. In that case entering a passphrase during every start-up can be a real burden, which is why you'd like to configure it.

prusnak:
I strongly agree with the first and second Reddit suggestions. Strongly Disagree with the third one.

>  I would like to make it possible to configure a default passphrase which would be automatically used upon start-up.

Where would this passphrase be stored? I think it is antipattern to store it anywhere. For various usability and security reasons.

matejcik:
agree with @prusnak. Mainly because if passphrase is stored, it becomes another piece of "static" data that you don't remember after a month of not needing it, and you need to back it up somehow. Either you add it to the seed (which defeats point (3)), or you handle it separately (which gives you a brand new point of failure).

andrewkozlik:
> Where would this passphrase be stored? I think it is antipattern to store it anywhere. For various usability and security reasons.

[...]

PR #3710, opened 2024-04-11
feat(core): add ability to request backups with any number of groups/shares.

### LINKED ISSUE #3636: Allow invocation of single-share SLIP-39 backup
The Shamir backup workflow in Trezor currently requires the user to select the number of shares and the threshold on Trezor. We want to make it possible for Suite to bypass this process and invoke 1-of-1 SLIP-39 backup on Trezor. I see a simple solution and a more generic solution.

### Solution 1

Add a new backup type `Slip39_Single`:
```proto
enum BackupType {
    Bip39 = 0;              // also called "Single Backup", see BIP-0039
    Slip39_Basic = 1;       // also called "Shamir Backup", see SLIP-0039
    Slip39_Advanced = 2;    // also called "Super Shamir" or "Shamir with Groups", see SLIP-0039#two-level-scheme
    Slip39_Single = 3;      // also called "Single Shamir", see SLIP-0039
}
```

If `Slip39_Single`, then the message handlers for `BackupDevice` and `ResetDevice` with `skip_backup==false` will go directly to the seed backup:

```py
async def backup_seed(backup_type: BackupType, mnemonic_secret: bytes) -> None:
    ...
    elif backup_type == BAK_T_SLIP39_SINGLE:
        await _backup_slip39_single(mnemonic_secret)
    ...

async def _backup_slip39_single(encrypted_master_secret: bytes) -> None:
    identifier = storage_device.get_slip39_identifier()
    iteration_exponent = storage_device.get_slip39_iteration_exponent()
    if identifier is None or iteration_exponent is None:
        raise ValueError

    mnemonics = slip39.split_ems(1, [(1, 1)], identifier, iteration_exponent, encrypted_master_secret)[0]
    await layout.slip39_basic_show_and_confirm_shares(mnemonics)
```

### Solution 2

Make it possible for Suite to specify the exact parameters of the Shamir backup. Extend `BackupDevice` with the following parameters:

```proto
message BackupDevice {
   optional uint32 group_threshold = 1;
   repeated uint32 member_thresholds = 2;
   repeated uint32 member_counts = 3;
}
```
The firmware will check that the following conditions are met:
- All three fields must be present or all three must be absent.
- The fileds must not be present if backup type is `Bip39`.
- `member_thresholds` must be the same length as `member_counts`.

There is no need to do other sanity checks like `group_threshold` being at most the number of groups etc., because that is done by `split_ems()`.

As an alternative:
```proto
message BackupDevice {
   optional uint32 group_threshold = 1;
   message Slip39Group {
       optional uint32 member_threshold = 1;
       optional uint32 member_count = 2;
   }
   repeated Slip39Group groups = 2;
}
```
Similar conditions as in the previous case need to be checked by the firmware.

The `BackupDevice` message handler goes directly to the backup process if the parameters are specified:
```py
async def backup_seed(backup_type: BackupType, mnemonic_secret: bytes, group_threshold: int, groups: list[tuple[int, int]]) -> None:
    if group_threshold:
        await _backup_slip39(mnemonic_secret, group_threshold, groups)
        return;
    ...

async def _backup_slip39(encrypted_master_secret: bytes, group_threshold: int, groups: list[tuple[int, int]]):
    identifier = storage_device.get_slip39_identifier()
    iteration_exponent = storage_device.get_slip39_iteration_exponent()
    if identifier is None or iteration_exponent is None:
        raise ValueError

    # generate the mnemonics
    mnemonics = slip39.split_ems(
        group_threshold,
        groups,
        identifier,
        iteration_exponent,
        encrypted_master_secret,
    )

    # show and confirm individual shares
    if len(groups) == 1:
        await layout.slip39_basic_show_and_confirm_shares(mnemonics)
    else
        await layout.slip39_advanced_show_and_confirm_shares(mnemonics)
```

- Should we go with the three-field representation or the `Slip39Group` sub-message?
- Should we also put the parameters in `ResetDevice`? I feel we shouldn't, because if `skip_backup==true`, then it doesn't really make sense or we'd have to store the parameters in storage. Suite defers backup anyway.
- Should we be pedantic about not allowing multiple groups when backup type was selected as `Slip39_Basic`? I don't think we should.
- Should there be a confirmation screen on Trezor repeating the numbers of shares and thresholds before backup commences? I think it's not needed.

### Further notes

- When the Shamir backup is 1-of-1, we shouldn't be showing the `Recovery share #1` and `Check share #1`, just `Recovery seed` and `Check seed` like we do in BIP-39 backup. Ideally even if the user manually configures 1-of-1, i.e. special handling in `slip39_basic_show_and_confirm_shares()` when `len(shares) == 1`, just `show_share_words(share)` and `await _share_words_confirmed(None, share_words)`.

ibz:
Fixes #3636


Hannsek:
@matejcik Do we want to merge it into main if we want to release one more fw in May?

matejcik:
> Do we want to merge it into main if we want to release one more fw in May?

no reason not to. this feature can be present in fw and inactive in Suite.

matejcik (commented):
please add new tests for the new behavior, and make sure the old tests are passing

core/src/apps/management/reset_device/__init__.py
```diff
@@ -213,8 +207,19 @@ def _compute_secret_from_entropy(
     return secret
 
 
-async def backup_seed(backup_type: BackupType, mnemonic_secret: bytes) -> None:
-    if backup_type == BAK_T_SLIP39_BASIC:
+async def backup_seed(backup_type: BackupType, mnemonic_secret: bytes, group_threshold: int | None = None, groups: list[tuple[int, int]] = None) -> None:
```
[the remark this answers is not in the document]
matejcik: make it a `Collection` then

python/src/trezorlib/cli/device.py
```diff
@@ -295,7 +302,7 @@ def unlock_bootloader(client: "TrezorClient") -> str:
 
 
 @cli.command()
-@click.argument("enable", type=ChoiceType({"on": True, "off": False}), required=False)
+@click.argument("enable", type=ChoiceType({"on": True, "off": False}))
```
matejcik, line 305: `required` defaults to `False` for _options_
not for arguments. this change is not actually a no-op

tests/input_flows.py
```diff
@@ -1375,13 +1376,59 @@ def input_flow_t3t1(self) -> BRGeneratorType:
...
         # Mnemonic phrases
-        self.mnemonics = yield from load_5_shares(self.debug)
+        self.mnemonics = yield from load_N_shares(self.debug, 5)
 
         br = yield  # safety warning
         assert br.code == B.Success
         self.debug.press_yes()
 
 
+class InputFlowSlip39SingleBackup(InputFlowBase):
```
matejcik, line 1386: is this identical to `BasicBackup` except for the 1 ? then perhaps make the `BasicBackup` parametrized instead?

(this will then let you parametrize `test_backup_single` with different configurations sent from Suite)

tests/device_tests/test_msg_backup_device.py
```diff
@@ -111,6 +112,36 @@ def test_backup_slip39_advanced(client: Client, click_info: bool):
     assert expected_ms == actual_ms
 
 
+@pytest.mark.skip_t1b1
+@pytest.mark.setup_client(needs_backup=True, mnemonic=MNEMONIC_SLIP39_ADVANCED_20)
+@pytest.mark.parametrize(
+    "click_info", [True, False], ids=["click_info", "no_click_info"]
+)
+def test_backup_slip39_single(client: Client, click_info: bool):
```
matejcik, line 120: please parametrize this by a couple combinations, e.g., 1 of 1, 2 of 2, 3 of 5

core/src/apps/management/backup_device.py
```diff
@@ -18,14 +18,19 @@ async def backup_device(msg: BackupDevice) -> Success:
...
-    mnemonic_secret, mnemonic_type = mnemonic.get()
+    mnemonic_secret, backup_type = mnemonic.get()
     if mnemonic_secret is None:
         raise RuntimeError
 
     storage_device.set_unfinished_backup(True)
     storage_device.set_backed_up()
 
-    await backup_seed(mnemonic_type, mnemonic_secret)
+    await backup_seed(
```
matejcik, line 28: ...now you should be checking, before you call `backup_seed`, the validity of arguments
(i.e., BIP39 has no groups and no group threshold; if group threshold is specified, at least that many groups are specified, etc.)

tests/input_flows.py
ibz: No, `BasicBackup` needs to get the number of shares and threshold as well.

matejcik: makes sense, but this one should be parametrized then

matejcik:
the Suite-initiated backup must skip the initial checklist
i.e., this screen must not be there:
image

in its place, a confirmation screen will be shown: "do you want to create a <M>-of-<N> Shamir backup?"
@Hannsek do we have design?

for 1-of-1 backup specifically, the title of the screen should be "Recovery share" (or just Recovery seed? @Hannsek) instead of "Share #1"

anything i'm missing @Hannsek ?
here is the current flow, starts at screen 10 https://data.trezor.io/dev/firmware/ui_report/8723408386/T2T1-en-core_click_test/passed/TT_en-click_tests-test_backup_slip39_single.py%3A%3Atest_backup_slip39_single%5B1of1%5D.html

Hannsek:
> in its place, a confirmation screen will be shown: "do you want to create a -of- Shamir backup?"
> @Hannsek do we have design?

No it won't be shown. The standard (BIP39) wallet creation flow will be shown. See [here](https://www.figma.com/file/WbzQVEk1WzP0n9lRMJUB9S/%E3%80%94-Prod-%E3%80%95Model-T?type=design&node-id=26%3A4185&mode=design&t=D4Wd3L0JcopfrDiW-1). Instead of 12 words we'll show 20 words.

Hannsek:
> for 1-of-1 backup specifically, the title of the screen should be "Recovery share" (or just Recovery seed? @Hannsek) instead of "Share https://github.com/trezor/trezor-firmware/issues/1"

Actually, it should be Wallet backup. We'll update the flows in figma so it is accurate there.

matejcik:
> > in its place, a confirmation screen will be shown: "do you want to create a -of- Shamir backup?"
> > @Hannsek do we have design?
> 
> No it won't be shown. The standard (BIP39) wallet creation flow will be shown. See [here](https://www.figma.com/file/WbzQVEk1WzP0n9lRMJUB9S/%E3%80%94-Prod-%E3%80%95Model-T?type=design&node-id=26%3A4185&mode=design&t=D4Wd3L0JcopfrDiW-1). Instead of 12 words we'll show 20 words.

I believe it must be shown in cases other than 1-of-1. It's (a) kind of a security parameter, and (b) good UX to tell the user up front that they'll be writing down 8 shares

core/src/apps/management/backup_device.py
ibz: Why here and not *inside* `backup_seed`: https://github.com/trezor/trezor-firmware/pull/3710/files#diff-35a83e5f83f9aef28cfd412567ff6f0ae1fd985bba5779cc682f0f38fb6d47ffR223

Hannsek:
Oh yes, definitely. I was thinking about 1-of-1 in my comment.

ibz:
> for 1-of-1 backup specifically, the title of the screen should be "Recovery share" instead of "Share #1"

Should we also change "Recovery share #{0} checked successfully"?

I am thinking about how this will affect translations. Maybe we should introduce new strings rather than pass "" to the format, because it's not guaranteed that would make sense in all languages?

core/src/apps/management/backup_device.py
matejcik: Separation of responsibilities? `backup_seed` just asserts that its arguments are correct and then does its thing -- which is, specifically, directing the parameters to the appropriate layout function. Adding the validation in there is (a) pushing _user-facing errors_ further away from the user-provided values, and (b) tacking on new functionality onto `backup_seed`, namely, caring for user-generated error states.

Speaking of, maybe `backup_seed` should stay unmodified and we should do "if parameters are provided then launch `_backup_slip39` directly" inline here. (the other code path for `backup_seed` doesn't care about that functionality)

matejcik:
> I am thinking about how this will affect translations. Maybe we should introduce new strings rather than pass "" to the format, because it's not guaranteed that would make sense in all languages?

more like, change & reuse BIP39 strings

Hannsek:
Please use layouts and strings from BIP39 flow. It should look the same, the only difference should be 20 words instead of 12.

tests/device_tests/test_msg_backup_device.py
```diff
@@ -111,6 +112,37 @@ def test_backup_slip39_advanced(client: Client, click_info: bool):
...
+    assert client.features.needs_backup is False
+    assert client.features.unfinished_backup is False
+    assert client.features.no_backup is False
+    assert client.features.backup_type is messages.BackupType.Slip39_Advanced
+
+    expected_ms = shamir.combine_mnemonics(MNEMONIC_SLIP39_ADVANCED_20)
+    actual_ms = shamir.combine_mnemonics(
+        IF.mnemonics[:3] + IF.mnemonics[5:8] + IF.mnemonics[10:13]
+    )
+    assert expected_ms == actual_ms
```
matejcik, line 143: you should probably be using `MNEMONIC_SLIP39_BASIC_20_3of6` because for that you also get a `MNEMONIC_SLIP39_BASIC_20_3of6_SECRET` to test against, instead of having to call `combine_mnemonics`

the code for `actual_ms` is wrong
far as I can understand, `IF.mnemonics` is a list of individual shares, right?
you should `assert len(IF.mnemonics) == share_count`
you should also assert that any(?) combination of `share_threshold` mnemonics out of `IF.mnemonics` combines into `MNEMONIC_SLIP39_BASIC_20_3of6_SECRET`

(honestly, probably just checking the last N shares is enough, no need to iterate the combinations)

core/src/apps/management/reset_device/__init__.py
```diff
@@ -155,23 +149,57 @@ async def _backup_slip39_advanced(encrypted_master_secret: bytes) -> None:
...
+            await layout.slip39_basic_show_and_confirm_shares(mnemonics[0])
+        else:
+            await layout.slip39_advanced_show_and_confirm_shares(mnemonics)
+
+
+def _get_slip39_mnemonics(
+    encrypted_master_secret: bytes,
+    group_threshold: int,
+    groups: Sequence[tuple[int, int]],
+):
```
matejcik, line 188: missing return type

core/src/apps/management/reset_device/__init__.py
```diff
@@ -111,31 +113,23 @@ async def reset_device(msg: ResetDevice) -> Success:
...
-    # generate the mnemonics
-    mnemonics = slip39.split_ems(
-        1,  # Single Group threshold
-        [(threshold, shares_count)],  # Single Group threshold/count
-        identifier,
-        iteration_exponent,
-        encrypted_master_secret,
-    )[0]
+    mnemonics = _get_slip39_mnemonics(
+        encrypted_master_secret, group_threshold, [(share_threshold, share_count)]
```
matejcik, line 127: very tiny nit:
```suggestion
        encrypted_master_secret, group_threshold, ((share_threshold, share_count),)
```
we prefer tuples to lists if we at all can
in some places it's because tuples are immutable so semantically more appropriate (e.g., when checking membership in a list of constants)
in other places, like here, because a tuple is just one allocation but a list is two.

core/src/apps/management/backup_device.py
ibz: 39e10c48c0c2978f199d4847a423bf14b36fa8e3

tests/click_tests/test_backup_slip39_single.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+pytestmark = [pytest.mark.skip_t1b1]
+
+
+@pytest.mark.parametrize(
+    "group_threshold, share_threshold, share_count",
+    [
+        pytest.param(1, 1, 1, id="1of1"),
+    ],
```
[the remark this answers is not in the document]
ibz: 39e10c48c0c2978f199d4847a423bf14b36fa8e3

tests/device_tests/test_msg_backup_device.py
ibz: 39e10c48c0c2978f199d4847a423bf14b36fa8e3

matejcik (approved):

PR #3713, opened 2024-04-12
T3T1 production keys

matejcik:
I played with GIMP for a bit and managed to squeeze the `vendor_unsafe.toif` under a limit so that the total VH size fits into 1024 bytes.
`vendor_trezor.toif` is the filled Trezor logo which was smaller from the outset.

please check the keys against expected values


core/embed/rust/src/event_loop/mod.rs
TychoVrahe, line 1: should not be included

TychoVrahe (commented):
otherwise LGTM. Checked keys against notion.

core/embed/rust/src/event_loop/mod.rs
matejcik: good catch, removed in rebase

TychoVrahe (approved):

PR #3715, opened 2024-04-12
refactor(core): generate macros for rust bindgen automatically

### LINKED ISSUE #3526: generate defines for rust bindgen automatically
Rust bindgen needs to be passed appropriate defines and include paths to work correctly. These are currently duplicated in both `build.rs` and board definitions. We should create an universal tool to pass all (necessary?) defines to rust bindgen automatically.

https://github.com/trezor/trezor-firmware/pull/3228#discussion_r1450475121

TychoVrahe:
removes most of the duplications between sconscript and bindgen build stages.

unfortunately, some duplicated declarations are still in makefile test targets, but at least we have more precise control over what is being tested. 

closes #3526

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -42,6 +42,8 @@ OPENOCD_INTERFACE ?= stlink
 # OpenOCD transport default. Alternative: jtag
 OPENOCD_TRANSPORT ?= hla_swd
 
+BINDGEN_MACROS_COMMON=-I../unix,-I../trezorhal/unix,-I../../build/unix,-I../../vendor/micropython/ports/unix,-I../../../crypto,-I../../../storage,-I../../vendor/micropython,-I../../vendor/micropython/lib/uzlib,-I../lib,-I../trezorhal,-I../trezorhal/unix,-I../models,-DTREZOR_EMULATOR,-DTREZOR_BOARD="boards/board-unix.h",
+
 ifeq ($(TREZOR_MODEL), 1)
 MCU = STM32F2
 LAYOUT_FILE = embed/models/model_T1B1.h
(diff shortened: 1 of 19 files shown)


matejcik:
i kinda hate that we have to put the bindgen macros in makefile, but can't come up with a better idea
@cepetr anything comes to mind?
could we maybe make some sort of single-source-of-truth text files that would be loaded both by `build.rs` and by sconscript, based on the model specification?
something like `include_paths.txt` and `defines.txt` per model?

(nit, if we go this way: typically you'd use `,` as the item separator, cf. gcc's `-Wl,-pie`)

TychoVrahe:
Though in scons these defines are not static even per model - they differ based on e.g. if you are emulating bootloader or firmware, if you have some flags set (for example `DISABLE_OPTIGA`). So the source of truth needs some intelligence to ensure that build.rs and scons use the same defines and includes. 

Maybe we could do some generation of the `txt` from scons for some given configuration and add it to `make gen`, and load it in makefile, or add as default for build.rs.

Separator: you mean to use comma in the `BINDGEN_MACROS`?

TychoVrahe:
``replaced the separator in: https://github.com/trezor/trezor-firmware/commit/03472fbfa1e2321059a0a6985fddbbfacbf8d86b

> but those basically match Rust build flags, correct?
not necessarily, not all C build options are passed to rust as feature - e.g. `PRODUCTION` is something user sets and isn't passed into rust. But bindgen should know about it, as it can affect what gets generated - while maybe this flag doesn't, the aim is to ensure that regardless of any future modifications bindgen see the same as gcc.

PR #3716, opened 2024-04-15
refactor: introduce emulator boards

TychoVrahe:
Introducing emulator boards concept, which allows us to handle different emutalor configurations in same manner as we handle configuration for regular HW.

Along with this, some reorganization of site_scons/models and embed/models is done, in order to keep the model specific files in as little places as possible


core/embed/trezorhal/unix/display_driver.c
```diff
@@ -27,46 +27,9 @@
...
-#define TOUCH_OFFSET_X 32
-#define TOUCH_OFFSET_Y 84
-
-#elif defined TREZOR_MODEL_T3T1
-
-#define WINDOW_WIDTH 400
-#define WINDOW_HEIGHT 600
-#define TOUCH_OFFSET_X 80
-#define TOUCH_OFFSET_Y 110
+#include TREZOR_BOARD
```
cepetr, line 30: `TREZOR_BOARD` is already included in `xdisplay.h`. However, if you prefer to keep it here, I would suggest placing it at the beginning as the first included file. Although we do not have formal rules regarding this...

core/SConscript.bootloader_emu
```diff
@@ -116,39 +117,14 @@ SOURCE_MOD += [
 
 if NEW_RENDERING:
     CPPDEFINES_MOD += ['NEW_RENDERING']
-    if TREZOR_MODEL in ('T',):
-        CPPDEFINES_MOD += ['DISPLAY_RGB565']
```
cepetr, line 120: The definition of `DISPLAY_RGB565` is inconsistent - it's only included in the emulator build, and I discovered that we don't use `DISPLAY_RGB565` anywhere in the code. However, we do use the `DISPLAY_MONO8` macro.

We have two options:
1. Eliminate `DISPLAY_RGB565` since it is not used.
2. Consistently define `DISPLAY_XXX` whenever the corresponding Rust feature `display_xxx` is defined. 

Currently, I have no preference. The second option might be better, although it offers no immediate benefit.

cepetr (changes requested):
I really like this change. It's a big step forward.

core/embed/trezorhal/unix/display-unix.h
```diff
@@ -3,39 +3,6 @@
...
-
-#ifdef TREZOR_MODEL_1
-#define MAX_DISPLAY_RESX 128
-#define MAX_DISPLAY_RESY 64
-#define DISPLAY_RESX 128
-#define DISPLAY_RESY 64
-#define TREZOR_FONT_BPP 1
-#endif
-
 extern uint8_t *const DISPLAY_DATA_ADDRESS;
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/dbff8d36031d0945ec5193f439024ef85b94396c

core/site_scons/models/T2B1/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+from typing import Optional
+
+from .emulator import configure as emul
+from .trezor_r_v3 import configure as configure_r3
+from .trezor_r_v4 import configure as configure_r4
+from .trezor_r_v6 import configure as configure_r6
+from .trezor_r_v10 import configure as configure_r10
+
+
+def configure_board(
```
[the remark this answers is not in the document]
TychoVrahe: using when there is more than one hw board: https://github.com/trezor/trezor-firmware/commit/2bb1f438a6cb7c57989ed555e6652f56ba69df6c

core/site_scons/models/T2B1/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    revision: Optional[int],
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/2965842bdc3adb7e7051abc99e94d6f6f5e5f15c

core/embed/trezorhal/unix/display_driver.c
TychoVrahe: moved https://github.com/trezor/trezor-firmware/commit/4aeb7aa83bfc4969221049687ea8839140ee20be

core/SConscript.bootloader_emu
TychoVrahe: Decided for 2, as we will soon need to distinguish RGB585 vs RGBA8888 variants in emulator driver: https://github.com/trezor/trezor-firmware/commit/847dfd5e3f5a65e95bb6b938a1fa18f24f0edb1e

Also DISC2 seem to have been missing these macros for new_rendering, so fixed it here https://github.com/trezor/trezor-firmware/commit/3efabfae14c879373add069d6a14cad9341c23bf

cepetr (approved):

PR #3717, opened 2024-04-15
add T3T1 production keys to trezorctl

TychoVrahe:
used keys from notion


python/src/trezorlib/firmware/models.py
```diff
@@ -227,18 +227,18 @@ class ModelKeys:
     boardloader_keys=[
```
matejcik, line 227: `production=True`

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/021bb1662496fa74629afc9fd843d14ab1c7548f

PR #3722, opened 2024-04-16
fix T3T1 reboot to bootloader 

TychoVrahe:
it was occasionally crashing while rebooting, because there was display update in progress and the SVC call blocked other interrupts. 

Moved the `ensure_compatible_settings` from svc handler to initial `svc_reboot_to_bootloader` call, so that inside the handler it is ensured that the display operations have finished.

The removal led to some weird hardfault during the reboot, due to some weirdness in returning from SVC. Basically it seemed that stack pointer, instead of the stack[0] value was changed, and the next stack[6] modification hardfaulted. 

The modification fixed it on U5 at least.. 

@hiviah could you try if this works on F4 also? does this modification makes sense to you?

[no changelog]

### DIFF
--- a/core/embed/bootloader/main.c
+++ b/core/embed/bootloader/main.c
@@ -353,6 +353,7 @@ void real_jump_to_firmware(void) {
     ui_screen_boot_stage_1(false);
   }
 
+  display_finish_actions();
   ensure_compatible_settings();
 
   // mpu_config_firmware();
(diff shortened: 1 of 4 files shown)


TychoVrahe:
Added the comment and moved the finish_actions out from the ensure function. https://github.com/trezor/trezor-firmware/commit/81e1cf621d8a40e7ca97246182dd92c2c1482a2c

bosomt:
**QA OK**

PR #3723, opened 2024-04-16
fix stm32u5 sha256 calculation

TychoVrahe:
The original version was failing in some scenarios - not entirely sure why, but using `HAL_HASHEx_SHA256_Accmlt_End` function with zero data to retrieve hash fixes the problem and does not require weird `HASH->STR |= HASH_STR_DCAL;` statement. Guessing that the `HAL_HASHEx_SHA256_Finish` was misused.


matejcik (approved):
utACK

cepetr (approved):

PR #3725, opened 2024-04-16
fix(core): ensure correctly programmed oem key option bytes

TychoVrahe:
This PR sets OEM key option bytes to 0xff.. value, which is interpreted as not set.

There is no possibility to read what is set, so this is only done when changing other option bytes - so we still need to check that the OEM key is not set later in fw.


core/embed/trezorhal/stm32u5/lowlevel.c
```diff
@@ -211,6 +211,11 @@ uint32_t flash_set_option_bytes(void) {
   FLASH->WRP2AR = WRP_DEFAULT_VALUE;
   FLASH->WRP2BR = WRP_DEFAULT_VALUE;
 
+  FLASH->OEM1KEYR1 = 0xFFFFFFFF;
```
cepetr, line 214: Please consider explaining in comment why we write FFs to OEMxKEY.

cepetr (approved):

PR #3728, opened 2024-04-17
add support for T3B1

### LINKED ISSUE #3785: Change manufacturer a product name on T3B1
Related issue: https://github.com/trezor/trezor-firmware/pull/3770

TychoVrahe:
Implements support for T3B1.

Refactored vendor headers a bit, basically just moved it to the common model folder.

DEVICE_CERT_CHAIN is mock for now, using values for T2B1. Also it should get same treatment as vendor headers..

Consumption mask still needs some timing adjustments.

Added missing font that affected/crashed bootloader for both T2B1 and T3B1.

fixes #3785


TychoVrahe:
@matejcik please review the production keys https://github.com/trezor/trezor-firmware/pull/3728/commits/c9bbc7185fed3a553eeb41dd7c809de7fb711039, device cert chain https://github.com/trezor/trezor-firmware/pull/3728/commits/cdc2cd797a3bc80bddf032c97e599a7718954e5c, and conceptually vendor header restructuralization  in https://github.com/trezor/trezor-firmware/pull/3728/commits/2993ad1701ac8295d49e036dc5c499f122e729fb

core/embed/models/T3B1/boards/trezor_t3b1_revB.h
```diff
@@ -0,0 +1,63 @@
...
+
+#define VDD_3V3 1
+
+#define USE_I2C 1
+#define USE_BUTTON 1
+#define USE_SBU 1
+#define USE_HASH_PROCESSOR 1
+#define USE_CONSUMPTION_MASK 1
+
+#include "displays/vg-2864ksweg01.h"
```
cepetr, line 15: To be compatible with `new_rendering`, we should use `#define DISPLAY_LEGACY_HEADER "displays/vg-2864ksweg01.h"` here. Please look at `trezor_r_v10.h`.

cepetr (changes requested):
It looks good from my side. I just found a few minor issues.

core/Makefile
```diff
@@ -30,8 +30,8 @@ PRODUCTION ?= 0
 PYOPT      ?= 1
 BITCOIN_ONLY ?= 0
 BOOTLOADER_QA ?= 0
-BOOTLOADER_DEVEL ?= 0
-TREZOR_MODEL ?= T
+BOOTLOADER_DEVEL ?= 1
```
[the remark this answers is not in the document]
TychoVrahe: no, reverted https://github.com/trezor/trezor-firmware/pull/3728/commits/d5ac620741738d39af8266eccd9b457c0b59c71a

core/embed/trezorhal/stm32u5/consumption_mask.c
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include STM32_HAL_H
+#include "rng.h"
+
+#define SAMPLES 110
+#define TIMER_PERIOD 16640  // cca 10 KHz @ 180MHz
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/3728/commits/a5ea7cb20e75f9cedc02d3095f1a9a6b59be21e3

still close enough to 10 kHz.. (note that this will undergo further tuning and i will change the constants if needed)

Also removed the special placement of the pwm_data, not needed on U5 as all ram can be accessed by DMA

core/embed/models/T3B1/boards/trezor_t3b1_revB.h
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/3728/commits/a5ea7cb20e75f9cedc02d3095f1a9a6b59be21e3

core/embed/firmware/bootloaders/bootloader_T3B1.bin
matejcik, line 1: i'm assuming that this file is not signed?
should we even include it then?

core/embed/models/T3B1/vendorheader/vendor_dev_DO_NOT_SIGN.toif
matejcik, line 1: should be a symlink to vendor_unsafe.toif

core/embed/models/T3B1/vendorheader/vendor_prodtest.toif
matejcik, line 1: it's identical to vendor_unsafe, so it should probably also be a symlink

core/embed/models/T3B1/vendorheader/vendor_prodtest.json
```diff
@@ -0,0 +1,22 @@
+{
+    "header_len": 512,
+    "text": "UNSAFE, FACTORY TEST ONLY",
+    "hw_model": "T3B1",
+    "expiry": 0,
+    "version": [0, 0],
+    "sig_m": 2,
+    "trust": {
+        "_reserved": 0,
+        "_dont_provide_secret": false,
```
matejcik, line 10: i believe we can drop those fields, right?

core/embed/models/T3B1/vendorheader/vendor_trezor.json
```diff
@@ -0,0 +1,22 @@
+{
+    "header_len": 512,
+    "text": "Trezor",
+    "hw_model": "T3B1",
+    "expiry": 0,
+    "version": [0, 0],
+    "sig_m": 2,
+    "trust": {
+        "_reserved": 0,
+        "_dont_provide_secret": false,
```
matejcik, line 10: dtto

core/embed/models/T3B1/vendorheader/vendor_trezor_btconly.json
```diff
@@ -0,0 +1,22 @@
+{
+    "header_len": 512,
+    "text": "Trezor Bitcoin-only",
+    "hw_model": "T3B1",
+    "expiry": 0,
+    "version": [0, 0],
+    "sig_m": 2,
+    "trust": {
+        "_reserved": 0,
+        "_dont_provide_secret": false,
```
matejcik, line 10: dtto

core/embed/models/T3B1/vendorheader/vendor_unsafe.json
```diff
@@ -0,0 +1,22 @@
+{
+    "header_len": 512,
+    "text": "UNSAFE, DO NOT USE!",
+    "hw_model": "T3B1",
+    "expiry": 0,
+    "version": [0, 0],
+    "sig_m": 2,
+    "trust": {
+        "_reserved": 0,
+        "_dont_provide_secret": false,
```
matejcik, line 10: dtto here? don't remember how this works for UNSAFE header, but in any case "don't provide secret" should be true.

core/embed/firmware/bootloaders/bootloader_T3B1.bin
TychoVrahe: Its probably dev signed. Its there just because FW doesn't compile without a bootloader. Maybe we should include it as QA bootloader a default to including QA bootloader when compiling for DEV bootloader event without the QA_BOOTLOADER flag set

core/embed/models/T3B1/vendorheader/vendor_dev_DO_NOT_SIGN.toif
TychoVrahe: changed https://github.com/trezor/trezor-firmware/commit/a80f055bcf5c718cb0cb69f1b6203a005f35f22b

core/embed/models/T3B1/vendorheader/vendor_prodtest.toif
TychoVrahe: changed https://github.com/trezor/trezor-firmware/commit/a80f055bcf5c718cb0cb69f1b6203a005f35f22b

core/embed/models/T3B1/vendorheader/vendor_prodtest.json
TychoVrahe: dropped https://github.com/trezor/trezor-firmware/commit/a80f055bcf5c718cb0cb69f1b6203a005f35f22b

core/embed/models/T3B1/vendorheader/vendor_trezor.json
TychoVrahe: dropped https://github.com/trezor/trezor-firmware/commit/a80f055bcf5c718cb0cb69f1b6203a005f35f22b

core/embed/models/T3B1/vendorheader/vendor_trezor_btconly.json
TychoVrahe: dropped https://github.com/trezor/trezor-firmware/commit/a80f055bcf5c718cb0cb69f1b6203a005f35f22b

core/embed/models/T3B1/vendorheader/vendor_unsafe.json
TychoVrahe: dropped https://github.com/trezor/trezor-firmware/commit/a80f055bcf5c718cb0cb69f1b6203a005f35f22b

core/embed/firmware/bootloaders/bootloader_T3B1.bin
matejcik: i'd definitely include it as a QA bootloader.
as to whether to leave it _also_  as the prod bootloader, you should still build with `BOOTLOADER_QA` so as not to brick your devkit with the official signed bootloader when we have it, no?
(if we need it to build anyhow, what about making the placeholder an empty file?)

like, perhaps the easier way here is just to keep it as a copy, but it just seems confusing. when the bootloader is not there, we know that we need to fix that. when it is there, someone could easily forget that it's not the right thing.

core/embed/models/T3B1/vendorheader/vendor_prodtest.json
matejcik: i meant both `_dont_provide_secret` and `_reserved`.

right now I can't figure out which should be in the prod header, but we definitely shouldn't have `_reserved: 0` for both UNSAFE and prod headers.
the construct file has `default: 0b11111111`. all those will end up encoded as zero in the binary header. I think we want all-zero UNSAFE header, with the only `1` being `allow_run_with_secret` -- because historically, incorrectly, this was set to `1` in the old UNSAFE header.

and _presumably_ we also want these new fields to be zeroes in the prod header, so if we add a new capability, the prod headers don't automatically gain it.

cepetr (approved):
utACK

core/embed/firmware/bootloaders/bootloader_T3B1.bin
TychoVrahe: changed the file to _qa in https://github.com/trezor/trezor-firmware/pull/3728/commits/e5618f921d0663c6543daa469d8fbf50ba6cd99a

Also changed the bootloader inclusion logic to include QA bootloader when `BOOTLOADER_DEVEL` is set.

The bootloader won't actually be updated unless the `PRODUCTION` (or `BOOTLOADER_QA`) flag is set, so there is not much danger here. 

The `BOOTLOADER_DEVEL` and `BOOTLOADER_QA` are almost redundant now, we could get rid of `BOOTLOADER_QA` flag and replace it with building with `BOOTLOADER_DEVEL=1` and `PRODUCTION=1`, not sure about that being a good idea as boardloader with PRODUCTION=1 will lock the device, so some risk is included

matejcik: let's not do that now but keep it in mind

core/embed/models/T3B1/vendorheader/vendor_prodtest.json
matejcik: (ping in case you missed this, i forgot to unresolve the comment after adding to it)

TychoVrahe: i think removing the _reserved everywhere is correct. 

see original discussion here https://github.com/trezor/trezor-firmware/pull/3683

and also this is how T3T1 VHs are setup

fixed https://github.com/trezor/trezor-firmware/commit/28a002a45723a813654108708ed5c57c04d2c8b9

matejcik (approved):

PR #3736, opened 2024-04-18
Improve error handling in crypto

onvej-sl:


crypto/zkp_ecdsa.c
```diff
@@ -364,10 +364,12 @@ int zkp_ecdsa_verify_digest(const ecdsa_curve *curve,
     }
   }
 
-  if (result == 0) {
-    secp256k1_ecdsa_signature_normalize(context_read_only, &signature,
-                                        &signature);
+  (void)secp256k1_ecdsa_signature_normalize(
+      context_read_only, &signature,
+      &signature);  // The return value inidicates whether the signature was
+                    // already normalized
```
andrewkozlik, line 370: Why should this be moved outside of the `if (result == 0)` block?

onvej-sl: It shouldn't, mistake.

Fixed in https://github.com/trezor/trezor-firmware/pull/3736/commits/8d7f815ee31f44f08c5e2f264f9f449775761e1a.

andrewkozlik (approved):

PR #3737, opened 2024-04-18
Signed artifacts for T3T1

matejcik:


TychoVrahe:
Could you also add QA bootloader from the same commit?

matejcik:
> Could you also add QA bootloader from the same commit?

6e78cce3bf0e

TychoVrahe (approved):

TychoVrahe:
Tested running the singed bootloader + prodtest on dev device running boardloader using production keys and it works fine. 

Prodtest with unsafe vendor header crashes on T3T1 HW, due to failing to write pairing secret - that will need addressing (it should not attempt to do pairing)

PR #3740, opened 2024-04-21
T3T1 - WalletCreation

obrusvit:
Draft - some pieces for discussions.


core/embed/rust/src/ui/model_mercury/component/button.rs
```diff
@@ -413,11 +413,13 @@ pub struct ButtonStyleSheet {
     pub disabled: &'static ButtonStyle,
 }
 
+// TODO: radius and borders not needed in T3T1 - sync with BL
```
mmilata, line 416: I think this should not break anything and will look more consistent BL<->FW, let's do it.

core/embed/rust/src/ui/model_mercury/component/footer.rs
```diff
@@ -64,6 +65,14 @@ impl<'a> Footer<'a> {
         self.style_description = style;
         ctx.request_paint();
     }
+
+    pub fn get_necessary_height(&self) -> i16 {
```
mmilata, line 69:
```suggestion
    pub fn height(&self) -> i16 {
```

core/embed/rust/src/ui/model_mercury/component/prompt_screen.rs
```diff
@@ -0,0 +1,141 @@
...
+            circle_inner_color: theme::ORANGE_LIGHT,
+            circle_pad_color: theme::GREY_EXTRA_DARK,
+            dismiss_type: DismissType::Tap,
+            button,
+        }
+    }
+}
+
+impl Component for PromptScreen {
+    type Msg = PageMsg<()>;
```
mmilata, line 76: I think we can return just `()` until more functionality is needed.

core/embed/rust/src/ui/model_mercury/component/status_screen.rs
```diff
@@ -0,0 +1,114 @@
...
+    type Msg = PageMsg<()>;
+
+    fn place(&mut self, bounds: Rect) -> Rect {
+        self.area = bounds;
+        match self.dismiss_type {
+            DismissType::SwipeUp(ref mut swipe) => {
+                swipe.place(bounds);
+            }
+            DismissType::Timeout(_) => {
+                // TODO: handle timeout or scrub the idea?
```
mmilata, line 69: My suggestion would be to embed a `struct Timeout` and handle it there.

core/embed/rust/src/ui/model_mercury/component/swipe_up_screen.rs
```diff
@@ -0,0 +1,69 @@
...
+            swipe: Swipe::new().up(),
+        }
+    }
+}
+
+impl<T> Component for SwipeUpScreen<T>
+where
+    T: Component,
+{
+    type Msg = PageMsg<T::Msg>;
```
mmilata, line 31: SwipeUpScreen makes sense to me, in fact with the T3T1 animations we should probably stop using `struct Swipe` in most places, but that is out of scope for this PR.

As for the message type, let's use `type Msg = ();` or an enum `{ Swiped, Content(inner) }` if needed.

core/embed/rust/src/ui/model_mercury/component/button.rs
obrusvit: https://github.com/trezor/trezor-firmware/pull/3740/commits/944a2ee6e61fd7cba9b07299672c6aabcfbdec5e

core/embed/rust/src/ui/model_mercury/component/footer.rs
obrusvit: https://github.com/trezor/trezor-firmware/pull/3740/commits/9acf5bd8d761b180387f6709b21728b8268ea6c8

core/embed/rust/src/ui/model_mercury/component/swipe_up_screen.rs
obrusvit: Ok, I did go with the enum
https://github.com/trezor/trezor-firmware/pull/3740/commits/c0f16354c5012e6804ac9ce36819ff349082f61a

core/embed/rust/src/ui/model_mercury/component/prompt_screen.rs
obrusvit: https://github.com/trezor/trezor-firmware/pull/3740/commits/807312c855cf325f4457510911a2d8581d03237b

core/embed/rust/src/ui/model_mercury/component/status_screen.rs
obrusvit: Ok, the first version is here: https://github.com/trezor/trezor-firmware/pull/3740/commits/317ed644799691ccbb6798c7b3cf5ba85eab8f78 
Very simple ctor supplied, for now the timeout is not parametrizable. Let's see what needs arise later.

PR #3750, opened 2024-04-24
Add FIRMWARE VERSION command to prodtest

andrewkozlik:
- Adds `FIRMWARE VERSION` command to prodtest.
- Improves error handling. For example the `VARIANT` and `OTP WRITE` commands used to crash when used for a second time in a row.
- Improves prodtest documentation and gives examples of how commands are actually used.

### DIFF
--- a/core/embed/prodtest/.changelog.d/+version.added
+++ b/core/embed/prodtest/.changelog.d/+version.added
@@ -1,0 +1,1 @@
+Added FIRMWARE VERSION command.
--- a/core/embed/prodtest/README.md
+++ b/core/embed/prodtest/README.md
@@ -98,9 +98,11 @@ The command requires two input parameters:
 
 If the display is not touched within the specified timeout, the command will return an `ERROR TIMEOUT`.
 
+The command does not check whether the touch point lies within the quadrant or not. It only returns the x and y coordinate of the touch point.
+
 Example (to draw a rectangle in the top-left quadrant and wait for 9 seconds for touch input):
 ```
-TOUCH 0 9
+TOUCH 09
 OK 50 90
 ```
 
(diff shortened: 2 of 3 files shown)


TychoVrahe (approved):
Looks great. While at it, could you please also fix the docs of the `TOUCH` command? There is no space expected between the two args.

For future, we might also be interested in boardloader + bootloader versions, but we can do that later.

andrewkozlik:
> fix the docs of the TOUCH command

Done and also documented the return values. https://github.com/trezor/trezor-firmware/pull/3750/commits/2ebf237575f2a2e2ed52fc4a592f6700c13f6613

> For future, we might also be interested in boardloader + bootloader versions, but we can do that later.

Good point. That's probably more complicated. We'd need to look into the headers, I guess.

I agree that we should leave it to another PR. My main goal was to improve the documentation a bit and the rest of the things I just bumped into along the way.

Are there any preferences with regard to the command name. Alternatives:
`FW VERSION`
`FW VERSION READ`

TychoVrahe:
ACK for the new commit. 

I would slightly prefer shorter version as i can't correct my typos when entering these commands manually, so `FW VERSION` would be my favorite, but i don't think it matters much

created an issue for the other version commands: https://github.com/trezor/trezor-firmware/issues/3752

andrewkozlik:
I agree it doesn't really matter, but it would be convenient to have a consistent naming, which is also easy to remember. So what should the command for bootloader and boardloader version be? `BL VERSION` and ...?

TychoVrahe:
`BR VERSION` comes to mind - we used to use BR for git tag like here https://github.com/trezor/trezor-firmware/releases/tag/core%2Fbr2.0.0 - but it was abandoned for sake of readability (in favor of [core/boardloader/v2.1.1](https://github.com/trezor/trezor-firmware/releases/tag/core%2Fboardloader%2Fv2.1.1) ) so thats one argument against

PR #3759, opened 2024-04-30
CI: fix master diff reports

matejcik:
masterdiffs have been failing due to wrongly generated names after we added languages


tests/ui_tests/common.py
```diff
@@ -262,6 +257,11 @@ def id(self) -> str:
     def fixtures_name(self) -> str:
         return f"{self.model}_{self.language}_{self.name}"
 
+    @classmethod
+    def from_fixtures(cls, fixtures_name: str, group: str) -> Self:
+        model, lang, name = fixtures_name.split("_", maxsplit=2)
+        return cls(model, group, name, lang)
```
grdddj, line 263: 20 lines above, we create `cls` using keyword-arguments, maybe it makes sense to use kwargs here as well (with 4 args it is already easy to mistake the order - and the type-check does not help here, as all are strings)

grdddj (approved):
utACK, just one nit

tests/ui_tests/common.py
matejcik: you're right, 0b99c699d880

Issue #3666, opened 2024-04-01
T3T1 UI: flows that run after final message is sent to host


Hannsek:
https://www.figma.com/file/n43caWJFGJlT167AQcxruu/%E3%80%94-Prod-%E3%80%95Safe-5?type=design&node-id=26%3A4176&mode=design&t=vdrfMMQHI03AL7c0-1

mmilata:
Seems to be complete except the part that stays on after the final protobuf message, renaming the ticket to track that. Needs #3686.

matejcik:
which flows should be covered here?
all of them?

is the design with "continue in the app" up-to-date for this feature?

Hannsek:
yes, all flows with success screen.

The designs are not changing from current designs in production.

romanz:
This issue also covers transaction signature flows.

romanz:
Re-opening since we want to keep the confirmation layout open for a few seconds (#5097), even if the client sends a message immediately after the response (currently, it closes the confirmation screen).

PR #3760, opened 2024-04-30
T3T1 - Receive flow

mmilata:
Implements #3666 but since we don't have working UI tests nor easy way to test on device I wouldn't close it after this is merged into `ui-t3t1`.

receive_0 receive_1 receive_6 receive_2 receive_3 receive_4 receive_5
multisig_warning_0 path_warning_0 path_warning_1 path_warning_2


core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -2150,9 +2110,29 @@ pub static mp_module_trezorui2: Module = obj_module! {
     ///     """Show single-line text in the middle of the screen."""
     Qstr::MP_QSTR_show_wait_text => obj_fn_1!(new_show_wait_text).as_obj(),
 
-    /// def flow_get_address() -> LayoutObj[UiResult]:
+    /// def flow_get_address(
```
obrusvit, line 2113: exposing the functions as `flow_xyz` is a good idea for now. I will use it also in flows I'm working on.

obrusvit (commented):
A final show success missing after you tap to confirm.

core/translations/en.json
```diff
@@ -10,11 +10,16 @@
...
     "addr_mismatch__xpub_mismatch": "XPUB mismatch?",
+    "address__cancel_contact_support": "If receive address doesn't match, contact Trezor Support at trezor.io/support.",
+    "address__cancel_receive": "Cancel receive",
     "address__public_key": "Public key",
+    "address__qr_code": "QR code",
     "address__title_cosigner": "COSIGNER",
     "address__title_receive_address": "RECEIVE ADDRESS",
     "address__title_yours": "YOURS",
     "address_details__derivation_path": "Derivation path:",
+    "address_details__derivation_path_no_colon": "Derivation path",
```
obrusvit, line 21: Isn't this a little bit wasteful? Can't we remove colon from `address_details__derivation_path` and append it in `tt` and `tr`. 

But I see it used in rust in `tr`, that makes it more complicated.

mmilata: Asked on slack about it, will likely change it depending on conclusion. Slight issue with appending colon is that French has a space before colon.

core/embed/rust/src/ui/model_mercury/layout.rs
mmilata: eventually we probably want them in separate python module but for now just place them here

mmilata:
> A final show success missing after you tap to confirm.

Fixed in 064c4db.

emu00000000

Hannsek:
👌🏻

small nits:
image
`Account` instead of `Address`

image
->
image

regarding the success screen - menu is missing there. How big problem would it be to add it there?
image

mmilata:
The menu is supposed to be available _after_ the last message is sent to Suite. We currently can't do that, it needs #3686 + probably bunch of extra work. Will fix the other issues later today.

obrusvit (approved):

core/translations/en.json
mmilata: Since there are only two such strings I added both variants in 3b782887634ae839ea35524831b25b97f97ce5f1 but changed the names to be consistent with `words__account_colon`.

mmilata:
> small nits:

Fixed in 3b782887634ae839ea35524831b25b97f97ce5f1

emu00000000 emu00000001

PR #3761, opened 2024-05-01
fix USB configuration for STM32U585

TychoVrahe:
Configures USB FS to use 6 endpoints - we need that when we use VCP and FIDO together.


cepetr (approved):
utACK

PR #3762, opened 2024-05-01
lockscreen for mercury_ui

TychoVrahe:
implements lockscreen for mercury UI.

Todo: 
- review backlight settings for not connected and locked states
- coinjoin authorized label/icon - new string in translations if label
- fix locked/not connected label capitalization (translations)


core/embed/rust/src/ui/model_mercury/component/homescreen/mod.rs
```diff
@@ -208,63 +192,7 @@ impl Component for Homescreen {
     }
 
     fn paint(&mut self) {
-        self.pad.paint();
```
mmilata, line 211: I think we can remove `pad` entirely since it should not be needed with new drawing library? Perhaps also `paint_notification_only`?

mmilata (approved):
LGTM, new drawing library makes it way more readable than before.

core/embed/rust/src/ui/model_mercury/component/homescreen/mod.rs
mmilata: removed both in 8079534d0e1e039b83f407e8ed62f14d0b078430

PR #3763, opened 2024-05-02
add haptic feedback to TT UI loader

TychoVrahe:
add the loader feedback to tt ui (as fallback for mercury for T3T1)

still needs to be added to mercury too, but thats a separate PR

### DIFF
--- a/core/embed/rust/src/ui/model_tt/component/button.rs
+++ b/core/embed/rust/src/ui/model_tt/component/button.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "haptic")]
-use crate::trezorhal::haptic::{play, HapticEffect};
+use crate::trezorhal::haptic::{self, HapticEffect};
 use crate::{
     strutil::TString,
     time::Duration,
@@ -30,6 +30,7 @@ pub struct Button {
     state: State,
     long_press: Option<Duration>,
     long_timer: Option<TimerToken>,
+    haptics: bool,
 }
 
 impl Button {
@@ -46,6 +47,7 @@ impl Button {
             state: State::Initial,
             long_press: None,
             long_timer: None,
+            haptics: true,
         }
     }
 
@@ -81,6 +83,11 @@ impl Button {
 
     pub fn with_long_press(mut self, duration: Duration) -> Self {
         self.long_press = Some(duration);
+        self
+    }
+
+    pub const fn without_haptics(mut self) -> Self {
+        self.haptics = false;
         self
     }
 
@@ -252,7 +259,9 @@ impl Component for Button {
                         // Touch started in our area, transform to `Pressed` state.
                         if touch_area.contains(pos) {
                             #[cfg(feature = "haptic")]
-                            play(HapticEffect::ButtonPress);
+                            if self.haptics {
+                                haptic::play(HapticEffect::ButtonPress);
+                            }
                             self.set(ctx, State::Pressed);
                             if let Some(duration) = self.long_press {
                                 self.long_timer = Some(ctx.request_timer(duration));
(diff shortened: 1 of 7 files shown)


core/embed/rust/src/ui/model_tt/component/button.rs
```diff
@@ -252,7 +259,9 @@ impl Component for Button {
                         // Touch started in our area, transform to `Pressed` state.
                         if touch_area.contains(pos) {
                             #[cfg(feature = "haptic")]
-                            play(HapticEffect::ButtonPress);
+                            if self.haptics {
+                                play(HapticEffect::ButtonPress);
```
cepetr, line 263: I would prefer to import the `haptic` module instead of using the `play` function directly. Then, we could call `haptic::play()` and eventually `haptic::play_rtp()`, which I find clearer. This can be applied to `loader.rs` as well - where these two approaches are combined now.

core/embed/trezorhal/haptic.h
```diff
@@ -23,4 +23,11 @@ bool haptic_test(uint16_t duration_ms);
 // Play haptic effect
 void haptic_play(haptic_effect_t effect);
```
cepetr, line 24: Not sure about this: but what about renaming `haptic_play()` and `haptic_play_rtp()` to something that better explains their function - like `haptic_play_effect()` and `haptic_play()` instead of `haptic_play_rtp()`?

core/embed/rust/src/ui/model_tt/component/loader.rs
```diff
@@ -173,6 +174,15 @@ impl Component for Loader {
                 } else if self.is_completely_shrunk(now) {
                     return Some(LoaderMsg::ShrunkCompletely);
                 } else {
+                    #[cfg(feature = "haptic")]
+                    {
+                        if matches!(self.state, State::Growing(_)) {
```
cepetr, line 179: The code is quite clear but using quite a lot of hard-coded constants:

Maybe we should introduce a new constant in the haptic header - `HAPTIC_MAX_AMPLITUDE`, and also use a constant instead of 1000 (if I remember correctly, we have `LOADER_MAX` somewhere), then modify the expression to use a percentage of `HAPTIC_MAX_AMPLITUDE`.

cepetr (changes requested):

core/embed/rust/src/ui/model_tt/component/button.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/f42d8f7ce5ec838d8b5021d26404dddb1a907610 and https://github.com/trezor/trezor-firmware/commit/9e492945a6e334827703b740b36876f8ef514ec1

core/embed/trezorhal/haptic.h
TychoVrahe: renamed to `haptic_play_custom` in https://github.com/trezor/trezor-firmware/commit/8f4877998e03dd176e3768f1f3a84e8a8bcaec13

(its still not ideal as the values passed might need to be different if different driver IC or actuator is used, but i don't see a very clear way to deal with this now and there is no pressing need to have it solved right now)

core/embed/rust/src/ui/model_tt/component/loader.rs
TychoVrahe: Modified the parameter to the play_custom function to accept percentage directly, which at least hide possible differences in max amplitude between different driver ICs. We lose some precision but i find it acceptable, otherwise we might switch to tenths of percents perhaps.

Also named the constants in loader module

 https://github.com/trezor/trezor-firmware/commit/021f2a41b9ad1499cf7f4fd84f6bdf98f7724df9
https://github.com/trezor/trezor-firmware/commit/34bea44a737782d325742fe0c22bbdfc01afc85f

cepetr (approved):

PR #3770, opened 2024-05-03
change USB manufacturer and product strings for new models

TychoVrahe:
`Trezor Company` and model full name is used.

For discovery kits, manufacturer is set to `Trezor DIY`


prusnak:
Let's not forget to change this for T3B1 as well in https://github.com/trezor/trezor-firmware/pull/3728

matejcik:
utACK, please add changelogs

TychoVrahe:
do we need changelogs though? it does not affect current products and previous T3T1 build are effectively internal only

For T3B1, do we want new style strings or match F4 based TS3?

prusnak:
> For T3B1, do we want new style strings or match F4 based TS3?

New

matejcik:
> do we need changelogs though? it does not affect current products and previous T3T1 build are effectively internal only

internal only yes, but "released" and tagged with a version. we need a changelog entry to say what happened between 2.1.5 and 2.1.6

TychoVrahe:
added changelogs to bootloader and prodtest: https://github.com/trezor/trezor-firmware/commit/af7d970ca01cce8f737e64a9d8a17c0f55b9f814

matejcik (approved):

PR #3775, opened 2024-05-05
add alpha capabilities to drawlib

TychoVrahe:
- adds generic mono4 blending with alpha
- adds alpha to circle, toif and text

The alpha for blending mono4 is computed by multiplying the mono4 value with the aplha, then use 8bit blend. When dma2d accelerated, the color is set to constant and the alpha channel is set to the multiple of alpha and step on the gradient.

The circle blending with alpha is not ideal when ring/outline is drawn - alpha blending will result in some of the foreground color visibility in the inner circle. This could be fixed by using native rings, or composing the ring in side layer and blending the result. However, for current use case this seems acceptable and we can implement more precise circle blending later.

### DIFF
--- a/core/embed/lib/gfx_color.h
+++ b/core/embed/lib/gfx_color.h
@@ -75,6 +75,8 @@ typedef uint32_t gfx_color32_t;
 #define gfx_color32_to_g(c) (((c) & 0x0000FF00) >> 8)
 // Extracts blue component from gfx_color32_t
 #define gfx_color32_to_b(c) (((c) & 0x000000FF) >> 0)
+// Extracts alpha component from gfx_color32_t
+#define gfx_color32_to_a(c) (((c) & 0xFF000000) >> 0)
 
 // 4-bit linear interpolation between `fg` and `bg`
 #define a4_lerp(fg, bg, alpha) (((fg) * (alpha) + ((bg) * (15 - (alpha)))) / 15)
@@ -93,6 +95,13 @@ static inline gfx_color16_t gfx_color16_rgb(uint8_t r, uint8_t g, uint8_t b) {
 // Alpha is set to 255.
 static inline gfx_color32_t gfx_color32_rgb(uint8_t r, uint8_t g, uint8_t b) {
   return (0xFFU << 24) | ((uint32_t)r << 16) | ((uint32_t)g << 8) | b;
+}
+
+// Constructs a 32-bit color from the given red (r),
+// green (g), blue (b) and alhpa (a) values in the range 0..255.
+static inline gfx_color32_t gfx_color32_rgba(uint8_t r, uint8_t g, uint8_t b,
+                                             uint8_t a) {
+  return (a << 24) | ((uint32_t)r << 16) | ((uint32_t)g << 8) | b;
 }
 
 // Converts a 16-bit color to a 32-bit color; alpha is set to 255
@@ -341,5 +350,13 @@ const gfx_color16_t* gfx_color16_gradient_a4(gfx_color_t fg, gfx_color_t bg);
 // Each element in the array represents a color, with `retval[0]` being
 // the background (`bg`) color and `retval[15]` the foreground (`fg`) color
 const gfx_color32_t* gfx_color32_gradient_a4(gfx_color_t fg, gfx_color_t bg);
+
+// Returns a color with alpha channel set
+//
+// The original color is not modified
+static inline gfx_color32_t gfx_color32_set_alpha(gfx_color32_t c,
+                                                  uint8_t alpha) {
+  return (c & 0xFFFFFF) | (alpha << 24);
+}
 
 #endif  // GFX_COLOR_H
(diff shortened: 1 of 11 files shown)


core/embed/lib/gfx_color.h
```diff
@@ -342,4 +342,9 @@ const gfx_color16_t* gfx_color16_gradient_a4(gfx_color_t fg, gfx_color_t bg);
 // the background (`bg`) color and `retval[15]` the foreground (`fg`) color
 const gfx_color32_t* gfx_color32_gradient_a4(gfx_color_t fg, gfx_color_t bg);
 
+// Returns a color with alpha channel set
+//
+// The original color is not modified
+gfx_color32_t gfx_color32_set_alpha(gfx_color32_t c, uint8_t alpha);
```
cepetr, line 348: If we improved `gfx_color32_blend_a4/a8()` so that they could interpolate the alpha channel as well, we could eliminate this function altogether.

core/embed/trezorhal/stm32f4/dma2d_bitblt.c
```diff
@@ -113,20 +116,31 @@ static void dma2d_config_clut(uint32_t layer, gfx_color_t fg, gfx_color_t bg) {
...
     cache[layer].c_fg = fg;
     cache[layer].c_bg = bg;
-
-    for (int step = 0; step < GRADIENT_STEPS; step++) {
-      clut[step] = gfx_color32_blend_a4(fg, bg, step);
+    cache[layer].alpha = alpha;
+    cache[layer].blend = blend;
+
+    if (blend) {
+      gfx_color32_t fg32 = gfx_color16_to_color32(fg);
```
cepetr, line 127: No sure if it might work, but::
- do not incorporate 8-bit alpha into the gradient
- set `AlphaMode` in layer configuration to `DMA2D_COMBINE_ALPHA`
- set `InputAlpha` in layer configuration to 8-bit alpha to `bb->src_alpha`

cepetr (changes requested):

core/embed/lib/gfx_color.h
TychoVrahe: the `gfx_color32_blend_a4/a8` with `GFX_COLOR_16BIT` setting do not have alpha channels in the input params, so this is kinda impractical

we would need to use the 32bit variant, convert the colors from 16 to 32bit format and still set alpha on one of the values

TychoVrahe: or add `start_alpha` `end_alpha` parameters to 16bit variant of `gfx_color32_blend_a4`, which also isnt very nice

cepetr: Yes, you are right, it wasn't good idea at all. Maybe, then if we need `gfx_color32_set_alpha()`, we can make it `static inline`

TychoVrahe: done here https://github.com/trezor/trezor-firmware/commit/2f741408ba9e087a6dc87a94c53c81831e7837d5

core/embed/trezorhal/stm32f4/dma2d_bitblt.c
TychoVrahe: it works like this: https://github.com/trezor/trezor-firmware/commit/2f741408ba9e087a6dc87a94c53c81831e7837d5
so setting alpha in the CLUT with alpha only from gradient, the InputAlpha is set to bb->src_alpha and combination is used. The gradient function still needs blend parameter because for blending, color is constant and only alpha changes, but for copying the alpha needs to be full and color is adjusted.

cepetr (changes requested):

core/embed/trezorhal/stm32f4/dma2d_bitblt.c
```diff
@@ -302,12 +316,16 @@ bool dma2d_rgb565_blend_mono4(const gfx_bitblt_t* params) {
...
-    dma2d_handle.LayerCfg[1].InputColorMode = DMA2D_INPUT_A4;
+    dma2d_handle.LayerCfg[1].InputColorMode = DMA2D_INPUT_L4;
     dma2d_handle.LayerCfg[1].InputOffset = bb->src_stride * 2 - bb->width;
-    dma2d_handle.LayerCfg[1].AlphaMode = 0;
-    dma2d_handle.LayerCfg[1].InputAlpha = gfx_color_to_color32(bb->src_fg);
+    dma2d_handle.LayerCfg[1].AlphaMode = DMA2D_COMBINE_ALPHA;
+    dma2d_handle.LayerCfg[1].InputAlpha = bb->src_alpha;
     HAL_DMA2D_ConfigLayer(&dma2d_handle, 1);
 
+    gfx_color_t bg = gfx_color32_set_alpha(bb->src_fg, 0);
```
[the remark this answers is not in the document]
TychoVrahe: fixed as part of cherry-pick https://github.com/trezor/trezor-firmware/commit/fb5f691d2132200128645924070d179a9f76f690

core/embed/trezorhal/stm32f4/dma2d_bitblt.c
TychoVrahe: yeah its better like that, i cherry picked your commit https://github.com/trezor/trezor-firmware/commit/fb5f691d2132200128645924070d179a9f76f690

one thing that i changed was this https://github.com/trezor/trezor-firmware/commit/0a6558c1f8cbd7daf043355e90a7fd8484433e7c - the background color should be irrelevant when blending

cepetr (approved):

Issue #3748, opened 2024-04-24
Flows enhancement


Hannsek:
```[tasklist]
### Wallet creation
- [ ] change copy according to [Figma](https://www.figma.com/file/n43caWJFGJlT167AQcxruu/%E3%80%94-Prod-%E3%80%95Safe-5?type=design&node-id=26%3A4185&mode=design&t=7ScNAc5FdDzvpuou-1)
- [ ] a bug with brightness on wallet backup words
- [ ] worse swiping on "wallet created" success screen
```

obrusvit:
Idea: all success screens can be done after we implement https://github.com/trezor/trezor-firmware/issues/3666 

Then we can refactor `core/src/apps/management/apply_settings.py` so that all layouts are pushed further according to models, i.e. no generic `confirm_action` is used. This allows to show the success screens only for mercury and not for other UIs.

PR #3791, opened 2024-05-07
add brightness setting 

TychoVrahe:
building on https://github.com/trezor/trezor-firmware/pull/3483, but adding support for `mercury`


core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -1202,6 +1216,30 @@ extern "C" fn new_request_number(n_args: usize, args: *const Obj, kwargs: *mut M
...
+        let title: TString = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
+        let min_count: u32 = kwargs.get(Qstr::MP_QSTR_min_count)?.try_into()?;
+        let max_count: u32 = kwargs.get(Qstr::MP_QSTR_max_count)?.try_into()?;
+        let count: u32 = kwargs.get(Qstr::MP_QSTR_count)?.try_into()?;
+        let value_callback: Obj = kwargs.get(Qstr::MP_QSTR_callback)?;
+        assert!(value_callback != Obj::const_none());
+
+        let callback = move |i: u32| {
+            value_callback
+                .call_with_n_args(&[i.try_into().unwrap()])
```
matejcik, line 1230:
```suggestion
                .call_with_n_args(&[unwrap!(i.try_into())])
```

matejcik: speaking of, why not reduce to `u16` and make the conversion infallible?

core/embed/trezorhal/stm32f4/backlight_pwm.c
matejcik, line 1: @cepetr please review this file

core/src/storage/device.py
matejcik, line 1: so I'm thinking, what if, instead of exposing the raw NumberSliderDialog to Python, we did the full brightness control in Rust?

we can keep the NumberSlider component pretty much as-is, just write a Rust callback that updates brightness, and another Rust call after clicking Ok that saves the value

that way we avoid most of Python, and the relevant code will just invoke something like `trezorui2.set_brightness()`

core/src/trezor/ui/style.py
matejcik, line 1: even this file could just delegate to Rust

common/protob/messages-management.proto
```diff
@@ -181,6 +182,7 @@ message ApplySettings {
     optional SafetyCheckLevel safety_checks = 9;  // Safety check level, set to Prompt to limit path namespace enforcement
     optional bool experimental_features = 10;  // enable experimental message types
     optional bool hide_passphrase_from_host = 11;  // do not show passphrase coming from host
+    optional uint32 brightness = 12;  // display brightness, 0 = select on device
```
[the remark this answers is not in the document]
Hannsek: Not right now, but will be in the future.

core/embed/trezorhal/stm32f4/backlight_pwm.c
```diff
@@ -16,6 +18,35 @@ static int pwm_period = 0;
 
 int backlight_pwm_set(int val) {
   if (BACKLIGHT != val && val >= 0 && val <= 255) {
+    // TPS61043: min 1% duty cycle
```
cepetr, line 21: The condition here should be something like `val < (255 / 100) + 1`.

In my opinion, we might not need this check. According to the datasheet:

> The CTRL pin accepts a PWM duty cycle from D = 1% to 100%. Duty cycles below 1% are also possible, with the restriction that the device is forced into shutdown if the off time of the applied PWM signal exceeds 10 ms.

core/embed/trezorhal/stm32f4/backlight_pwm.c
```diff
@@ -16,6 +18,35 @@ static int pwm_period = 0;
...
+      val = 0;
+    }
+
+    // TPS61043 goes to shutdown when duty cycle is 0 (after 32ms),
+    // so we need to set GPIO to high for at least 500us
+    // to wake it up.
+    if (BACKLIGHT_PWM_TIM->BACKLIGHT_PWM_TIM_CCR == 0) {
+      GPIO_InitTypeDef GPIO_InitStructure = {0};
+
+      HAL_GPIO_WritePin(BACKLIGHT_PWM_PORT, BACKLIGHT_PWM_PIN, GPIO_PIN_SET);
```
cepetr, line 32: Instead of reconfiguring GPIO, we **could** directly control OC1 by setting the `OC1M` bits in the `CCMR1` register. This applies to TIM1` used on the STM32F4 as well as to TIM8/17 used on the STM32U5.

Through the `OC1M` bits, OC1 can be forced high and then switched to PWM mode.

cepetr (changes requested):

core/embed/trezorhal/stm32f4/backlight_pwm.c
TychoVrahe: removed: https://github.com/trezor/trezor-firmware/commit/78a5204dd68a49dadf6dd3b73d2f65698d7ba3c8 and  https://github.com/trezor/trezor-firmware/commit/3444eb464405929e4a2b04de55d01db90656bd90

core/embed/trezorhal/stm32f4/backlight_pwm.c
TychoVrahe: there is a problem with this method actually

on TIM1 and TIM8 OC1M bits are preloaded, due to setting TIM_CR2_CCPC. 

TIM17 does not have this functionality, but we are still setting this CCPC bit (which is not breaking the rules of reference manual). The reaction of TIM17 on having this bit set is unfortunately that it doesn't update the OC1M value at all. We would need to allow setting this bit only for timers that support it, but we also already have boardloader and bootloader released, so we need to actually switch it off for some boards. 

For now, i don't want to waste more time on this so i will keep it as is, but will make an issue to fix this later.

TychoVrahe: issue https://github.com/trezor/trezor-firmware/issues/3847

core/embed/rust/src/ui/model_mercury/theme/backlight.rs
```diff
@@ -0,0 +1,41 @@
...
+// Typical backlight values.
+
+#[cfg(feature = "bootloader")]
+pub fn get_backlight_normal() -> u16 {
+    150
+}
+
+#[cfg(not(feature = "bootloader"))]
+pub fn get_backlight_normal() -> u16 {
+    storage::get_brightness().unwrap_or(150).into()
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/e9ffaae3d8d30486b01fa4b09b39536e3e375a21

core/embed/rust/src/ui/model_mercury/theme/backlight.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+#[cfg(not(feature = "bootloader"))]
+pub fn get_backlight_low() -> u16 {
+    let val = storage::get_brightness();
+    return if val.is_ok() && val.unwrap() < 45 {
+        val.unwrap().into()
+    } else {
+        45
+    };
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/e9ffaae3d8d30486b01fa4b09b39536e3e375a21

common/protob/messages-management.proto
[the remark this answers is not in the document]
TychoVrahe: so keeping it as is?

core/src/storage/device.py
TychoVrahe: yes please take over, my idea was to make this usable for different use-cases as well without having to add another function to rust-python interface, which is not very convenient thing to do. (that said, i also believe we should rethink this part and redesing the rust-python interface, along with worfklow-ui interface. but thats different story)

core/embed/trezorhal/stm32f4/backlight_pwm.c
TychoVrahe: reviewed by petr below.

common/protob/messages-management.proto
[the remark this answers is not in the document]
matejcik: i still kind of think that this should be a separate `SetBrightness` message
for all settings, you **send** the value from host, and **confirm** on Trezor

But for brightness, you may **suggest** a value from host, but you're displaying some sort of _interactive control_ on Trezor for changing or confirming that value. I don't expect there to ever be a dialog "set brightness to 15? yes/no" on Trezor, am I right @Hannsek ?

could go either way but I'd lean towards treating brightness as its own workflow separate from the "apply any number of settings from this list" workflow. It's not like it is reusing any code from `apply_settings` either. what do yall think?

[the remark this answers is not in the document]
matejcik: n.b., having a `SetBrightness` message allows you to make it
```protobuf
message SetBrightness {
    optional uint32 value = 1;  // if not specified, let the user choose
}
```
which doesn't work for settings and you have to use `0` as a special value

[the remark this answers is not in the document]
TychoVrahe: The ability to have optional value provided and no special meaning for 0 is compelling argument for me.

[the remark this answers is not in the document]
Hannsek: It might be like that. It also might be the same as threshold -> host sends a message with specified value, but you can change it on device as well. But we don't know yet.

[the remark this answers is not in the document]
TychoVrahe: From the perspective of future extensibility, having separate message is clearly better.

Done here: https://github.com/trezor/trezor-firmware/commit/bf81a08d03d32c295d44d49c147d07de0433350b

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -260,6 +261,19 @@ where
...
 
+impl<F> ComponentMsgObj for NumberInputSliderDialog<F>
+where
+    F: Fn(u32),
+{
+    fn msg_try_into_obj(&self, msg: Self::Msg) -> Result<Obj, Error> {
+        let value = self.value().try_into()?;
+        match msg {
+            NumberInputSliderDialogMsg::Confirmed => Ok((CONFIRMED.as_obj(), value).try_into()?),
+            NumberInputSliderDialogMsg::Cancelled => Ok((CANCELLED.as_obj(), value).try_into()?),
```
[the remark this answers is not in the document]
matejcik: reworked

core/embed/rust/src/ui/model_mercury/layout.rs
matejcik: reworked

matejcik:
@TychoVrahe please check my changes
you can ignore [86bc1ad](https://github.com/trezor/trezor-firmware/pull/3791/commits/86bc1ad6447a7fc25c1ba34964be042b4bdabf3f), it is a cherry-pick from main, i needed it for vscode

core/embed/rust/build.rs
```diff
@@ -14,12 +14,54 @@ fn main() {
...
+    "-DFLASH_BLOCK_WORDS=1",
+];
+#[cfg(not(feature = "model_tr"))]
+const DEFAULT_BINDGEN_MACROS_T2B1: &[&str] = &[];
+
+fn add_bindgen_macros<'a>(clang_args: &mut Vec<&'a str>, envvar: Option<&'a str>) {
+    let default_macros = DEFAULT_BINDGEN_MACROS_COMMON
+        .iter()
+        .chain(DEFAULT_BINDGEN_MACROS_T2T1)
+        .chain(DEFAULT_BINDGEN_MACROS_T2B1);
```
obrusvit, line 58: Don't we miss also `T3T1` specifics?

core/embed/rust/src/ui/model_mercury/component/number_input_slider.rs
```diff
@@ -0,0 +1,204 @@
...
+
+use super::{theme, Button, ButtonMsg};
+
+pub enum NumberInputSliderDialogMsg {
+    Changed(u16),
+    Confirmed,
+    Cancelled,
+}
+
+pub struct NumberInputSliderDialog {
```
obrusvit, line 18: Note: copy of `model_t` version. Should be reworked for `mercury` when design done.
Linking it here for now: https://github.com/trezor/trezor-firmware/issues/3748

core/embed/rust/build.rs
matejcik: ah, this is a cherrypick of https://github.com/trezor/trezor-firmware/commit/cf58fdd31306f4560b357ec9980d61026d44d878 from main. we probably should, but this commit should be removed from the PR before merging, and someone should add T3T1 features when merging the UI branch into main (or later, whenever we hit a problem caused by it, so probably when you try to run T3T1 Rust tests)

matejcik (approved):

PR #3801, opened 2024-05-07
feat(solana): add nullable tag

RostarMarek:
Fixes issue raised in [this issue](https://github.com/trezor/trezor-suite/issues/12134). The issue was in parsing the message. The question remaining is wheter any optional parameter in messages that are being parsed will be the same as here the new authority(1 byte to denote wheter the value is set and then bytes with value) or it is possible that the value bytes are not there for optional parameters. This fix works for the latter, if the former is true the parsing of `is_optional` should be changed.
This fix introduces a new flag `nullable` that for parsing purposes that reads and discards the value when the byte denotes that it is not set in the message.
One more thing to note when trying to update the fixtures of ui tests, the script moves the `T3T1` part of the fixtures to the end of the file which creates a lot of line changes and it happened on multiple machines, therefore that change is not included int this PR, let me know wheter I should add it anyways.


core/src/apps/solana/transaction/instruction.py
```diff
@@ -50,9 +50,23 @@ def parse_instruction_data(
...
+            # Property cannot be both optional and nullable as if there are multiple such properties in 1 instruction
+            # it could lead to non-deterministic parsing
+            # it is validated by mako template
+            assert not (property_template.is_nullable and property_template.is_optional)
+
             if property_template.is_optional:
                 is_included = True if reader.get() == 1 else False
 
+            if property_template.is_nullable:
+                is_included = True if reader.get() == 1 else False
```
matejcik, line 62:
```suggestion
                is_included = reader.get() == 1
```
...

core/src/apps/solana/transaction/instruction.py
```diff
@@ -50,9 +50,23 @@ def parse_instruction_data(
...
+                if not is_included:
+                    # A default (null) value is included in the parsed message for nullable properties
+                    # (e.g. new_authority in Set Authority instruction) even if the is_included flag is not set.
+                    # We can safely discard this value.
+                    if reader.remaining_count() != 0:
+                        property_template.parse(reader)
```
matejcik, line 68: if we can safely discard, we should not be checking for `remaining_count()` here, right? consider a value that is "included" but then the data ends. we _want_ the eof fail in that case

Issue #3793, opened 2024-05-07
Change copy from Shamir to multi-share


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/5fedefac-c635-422b-8517-03060b8c6704)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/37fa6eb1-6880-4e03-a501-849b6e93611a)

Replace `Shamir` with `Multi-share`

Issue #3792, opened 2024-05-07
Delete (Shamir) from Create wallet (shamir) screen on both T2B1 and T2T1


Hannsek:
![image](https://github.com/trezor/trezor-firmware/assets/57008159/05c9d270-c789-4c14-86e7-95328f926078)
![image](https://github.com/trezor/trezor-firmware/assets/57008159/9485b31a-3a62-43b1-85df-0dd20a4cb9b8)

ibz:
Closing manually. This was already fixed.

PR #3803, opened 2024-05-08
chore(core) rename Shamir to Multi-share

ibz:
#3792 & #3793


obrusvit:
Please merge after https://github.com/trezor/trezor-firmware/pull/3696

prusnak:
Don't we also want to rename the identifiers of strings?

core/mocks/trezortranslate_keys.pyi
```diff
@@ -601,15 +601,15 @@ class TR:
     reset__threshold_info: str = "= minimum number of unique word lists used for recovery."
     reset__title_backup_is_done: str = "BACKUP IS DONE"
     reset__title_create_wallet: str = "CREATE WALLET"
-    reset__title_create_wallet_shamir: str = "CREATE WALLET (SHAMIR)"
+    reset__title_create_wallet_shamir: str = "CREATE WALLET (MULTI-SHARE)"
```
Hannsek, line 604: This would be visible also when user is creating 1 of 1 right?

During the creation of 1of1 shamir, there cannot be a term multi-share...

ibz: Good point. I'll use something else in that case. What should it be? "single share"?

Hannsek: No, we now cannot distinguish which type of backup user would do (single-share or multi-share). So for now on deviceReset let's use just Create wallet as in bip39 flow.

obrusvit (changes requested):
Some of the device tests fixtures for `T2B1` are not updated as they do not currently run on CI. Compare the changed fixtures for `T2B1` vs `T2T1`.
I suggest updating those tests locally.

STew790:
**QA OK**
Both devices

**Info:**
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: Trezor T2B1 2.7.1 regular (revision 83d1ce2b8e6e4eefe09539f06b28709c8a7a89dc)

PR #3808, opened 2024-05-09
chore(core) delete "(shamir)" from "create wallet"

ibz:
#3792


ibz:
This will cover it: https://github.com/trezor/trezor-firmware/pull/3803

PR #3805, opened 2024-05-08
tap to confirm animation for mercury UI

TychoVrahe:
Implements tap-to-confirm animation for mercury UI.

PR #3580, opened 2024-03-05
T3T1 keyboards

grdddj:
Creating new keyboard design

PR #3814, opened 2024-05-09
T3T1 mercury UI - PIN keyboard and flow

obrusvit:
PR adds:
- Setting up PIN and related flow
- PIN keyboard changes, some taken from https://github.com/trezor/trezor-firmware/pull/3580
- Implementation of `confirm_action` taken from https://github.com/trezor/trezor-firmware/pull/3805 and modified slightly so that we can confirm also actions without explicit "tap to confirm" or "hold to confirm" but only with simple "swipe up"

Some limitations added here for now https://github.com/trezor/trezor-firmware/issues/3748

emu00000000emu00000001emu00000002emu00000003

emu00000004emu00000005emu00000006

emu00000008emu00000007


mmilata:
Tested on device and it works nicely:+1:
gen and style are failing though: https://github.com/trezor/trezor-firmware/actions/runs/9024319705

mmilata (commented):
Please regenerate fixtures.json (ones that don't fail) and also make sure all screens of the multi-screen flows are covered by device tests. The current copypaste from Model T records only the first screen - this doesn't need to be changed everywhere but the new screens need to be recorded in at least one test case.

Hannsek:
Please change the X/delete/check buttons according to [figma](https://www.figma.com/design/n43caWJFGJlT167AQcxruu/%E3%80%94-Prod-%E3%80%95Safe-5?node-id=2004-594&t=32mP25Xl7wAcuVbg-4).

mmilata (approved):
let's merge this if the button issue is addressed

obrusvit:
rebased, I will wait for CI to update fixtures, then merge

Issue #3781, opened 2024-05-06
Support info for TS5


matejcik:
We need to add support info for `T3T1` to the relevant files.

This support info should be a copy of `T2B1`.

prerequisite for https://github.com/trezor/definitions/issues/30

Hannsek:
Wondering if we can also prepare the info for T3B1. @matejcik

matejcik:
that's a good question because T2B1 and T3B1 should always be identical, no? which kinda means that we maybe should do the support info by not duplicating it?

VojtaStr:
@matejcik hey guys how does this looks? @matejcik

Hannsek:
As discussed, it will be done this week. :)

ibz:
Closing manually. This has been merged.

PR #3817, opened 2024-05-10
chore(core): add coin support info for T3T1

ibz:
#3781


core/site_scons/site_tools/micropython/__init__.py
```diff
@@ -51,6 +52,7 @@ def generate_frozen_module(source, target, env, for_signature):
         sed_scripts = " ".join(
             [
                 rf"-e 's/utils\.MODEL_IS_T2B1/{is_t2b1}/g'",
+                rf"-e 's/utils\.MODEL_IS_T3T1/{is_t3t1}/g'",
```
matejcik, line 55: let's not introduce more situation-specific symbols

instead change those to work like the `LAYOUT` below, iow, the replacement should be looking for `utils.MODEL == "T2B1"`

(ideally at some point in the future this sed block should instead be something more like libcst parsing the sources and modifying the symbols, so that, e.g., you can catch `utils.MODEL in ("T2B1", "T2T1")` which is basically impossible here.
but that's for the future)

core/src/apps/common/coininfo.py.mako
```diff
@@ -144,6 +144,9 @@ coins_alt_t2t1 = [c for c in supported_on("T2T1", bitcoin) if c.name not in btc_
 coins_btc_t2b1 = [c for c in supported_on("T2B1", bitcoin) if c.name in btc_names]
 coins_alt_t2b1 = [c for c in supported_on("T2B1", bitcoin) if c.name not in btc_names]
 
+coins_btc_t3t1 = [c for c in supported_on("T3T1", bitcoin) if c.name in btc_names]
+coins_alt_t3t1 = [c for c in supported_on("T3T1", bitcoin) if c.name not in btc_names]
```
matejcik, line 148: ....whiiiich you can then use to implement the TODO here, and change the code generator iterate over model names

core/src/apps/ethereum/networks.py.mako
```diff
@@ -67,6 +67,15 @@ def _networks_iterator() -> Iterator[NetworkInfoTuple]:
             "${n.shortcut}",  # symbol
             "${n.name}",  # name
         )
+% endfor
+    elif utils.MODEL_IS_T3T1:
```
matejcik, line 71: we don't like elifs in this sort of code, this should be an `if`, the below should be `if model == T2T1`, dtto in other places

matejcik (approved):

STew790:
**QA OK**

PR #3820, opened 2024-05-13
fix(doc): Add a hint for Ubuntu to fix the python-is-python3 issue

peter-sanderson:


peter-sanderson:
I was not in the poetry shell.

Issue #3640, opened 2024-03-22
Implement repeated backup


andrewkozlik:
Currently the firmware allows seed backup to be created only once and only if the seed was generated on Trezor, i.e. not recovered. We want to make it possible for the user to create a backup repeatedly, even from a recovered seed. Before being allowed to do so, the user would be required to perform dry-run recovery. In case of BIP-39 this feature does not make sense, but it does make sense in case of SLIP-39 where the user might want to change their backup solution from, say, 1-of-1 to 2-of-3.

One question to consider is, should we allow this process for BIP-39? There is no security reason to not allow it, but there is also no reason to support it, since the user obviously has the backup if they were able to complete dry-run recovery. So supporting it is just confusing, e.g. the user thinks they will be able to convert their seed to SLIP-39, which is not possible. I would disallow it.

The main implementation problem is that the device can be powered-down during dry-run recovery and disconnected from the Suite instance. So in general we cannot expect the entire process (dry-run + repeated backup) to take place in one workflow or even in one session with Suite.

We have these options to consider:
1. Do not allow powering down the device during this process.
    - Not user friendly and inconsistent with normal dry-run experience, but it can make the overall implementation easier.
    - The entire flow could simply be initiated by the `BackupDevice` message, possibly with a boolean flag `repeated_backup` set to true, but it's not necessary to have this flag, because it is implied if backup already took place (`Features.needs_backup=False`).
2. Allow powering down during the dry-run part of the process, but once dry-run is complete, the repeated backup needs to take place immediately, i.e. no other workflows are allowed in between.
    - The flow would start with `RecoveryDevice` followed by `BackupDevice`.
3. Allow powering down during the dry-run part of the process, but once dry-run is complete, the repeated backup needs to take place before powering down, i.e. other workflows are allowed in between.
    - The flow would start with `RecoveryDevice` followed by `BackupDevice`.
    - Introduce a homescreen banner "BACKUP AVAILABLE" and a global variable indicating that backup is available, which would trigger this banner.
4. Allow powering down at any time during the entire process, i.e. the repeated backup can take place at any time in the future and other workflows are allowed in between the two phases.
    - The flow would start with `RecoveryDevice` followed by `BackupDevice`.
[...]

PR #3825, opened 2024-05-15
Support extendable backup flag in SLIP-39

andrewkozlik:
Related to:
- SLIP-39 spec update: https://github.com/satoshilabs/slips/commit/8d060706b549af6443e04f55605b71f65c981663
- Reference implementation update: https://github.com/trezor/python-shamir-mnemonic/commit/1525df19df504b1f69b49179140119959f317f24
- Feature: https://github.com/trezor/trezor-firmware/issues/3640
- Feature implementation: https://github.com/trezor/trezor-firmware/pull/3743

TODO:
- [x] Update `shamir-mnemonic` package once version 0.3.0 is released.
- [x] Update screens.
- [x] Maybe add some device/upgrade tests based on the QA list below.

Notes to @trezor/qa:
- Thoroughly test Shamir backup.
- Test case 1:
    1. Install old firmware (2.7.0 or earlier).
    2. Reset device and choose Shamir backup, but do not back up.
    3. Write down some address or XPUB from passphrase-less wallet.
    4. Write down some address or XPUB from passphrased wallet.
    5. Install new firmware.
    6. Ensure that the addresses of the wallets remain the same.
    7. Create a backup and reboot.
    8. Ensure that the addresses of the wallets remain the same.
    9. Wipe the device and recover using the backup.
    10. Ensure that the addresses of the wallets remain the same.
- Test case 2:
    1. Install old firmware.
    2. Reset device and choose Shamir backup and create a backup.
    3. Write down some address or XPUB from passphrase-less wallet.
    4. Write down some address or XPUB from passphrased wallet.
    5. Install new firmware.
    6. Ensure that the addresses of the wallets remain the same.
    7. Wipe the device and recover using the backup.
    8. Ensure that the addresses of the wallets remain the same.
- TC3: Ensure that Shamir backups created with **_new_** firmware can be successfully recovered on **_new_** firmware, i.e. check that addresses remain the same.
- TC4: Ensure that Shamir backups created with **_new_** firmware are not accepted by **_old_** firmware.


andrewkozlik:
One would expect the UI tests to differ in only the second word and last three checksum words. The reason why they also differ in the share values is that the PRNG is called by the storage when writing the extendable flag, which desynchronizes the PRNG between the old and new UI tests before the shares are generated. The actual encrypted master secret is probably the same.

MiroslavProchazka:
QA test task here - https://www.notion.so/satoshilabs/Support-extendable-backup-flag-in-SLIP-39-eac94587362948c9aff60992b622312c?pvs=4

matejcik:
we talked about dropping the stored identifier for extendable backups, do you still intend to do that?

andrewkozlik:
> we talked about dropping the stored identifier for extendable backups, do you still intend to do that?

I do, and I was also hoping to writing some more upgrade test(s). Maybe I can make some time tomorrow afternoon.

docs/core/misc/slip0039.md
```diff
@@ -75,8 +75,9 @@ This process does not use passphrase.
 1. Generate the required number of random bits (128 or 256), and store as
    `_MNEMONIC_SECRET`.
 2. Generate a random identifier and store as `_SLIP39_IDENTIFIER`.
```
matejcik, line 77: no longer true

docs/core/misc/slip0039.md
```diff
@@ -101,5 +102,6 @@ This process does not use passphrase.
 2. Use `slip39.recover_ems(shares)` to combine the shares and get metadata.
 3. Store the Encrypted Master Secret as `_MNEMONIC_SECRET`.
 4. Store the identifier as `_SLIP39_IDENTIFIER`.
```
matejcik, line 104: only if not extendable

tests/upgrade_tests/test_firmware_upgrades.py
```diff
@@ -333,6 +337,41 @@ def test_upgrade_shamir_recovery(gen: str, tag: Optional[str]):
...
+            IF = InputFlowSlip39BasicBackup(emu.client, False)
+            emu.client.set_input_flow(IF.get())
+            device.backup(emu.client)
+
+        assert not emu.client.features.needs_backup
+
+        # Check result
+        groups = shamir.decode_mnemonics(IF.mnemonics[:3])
+        ems = shamir.recover_ems(groups)
+        assert ems.ciphertext == mnemonic_secret
```
matejcik, line 372: I suspect that testing this is not enough. We (also?) need to test that passphrased addresses come out the same after upgrade.

I would maybe also add a version-gated check that if you set up your device on fw \< 2.7.1, the backup will _not_ have the extendable flag despite being generated on 2.7.1 and up

matejcik (commented):
implementation LGTM

i have caught a mistake in docs, and I believe we need to add a couple checks to the upgrade-test

docs/core/misc/slip0039.md
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3825/commits/5bafd9611183e9eab044ad7632077823cdb21197

docs/core/misc/slip0039.md
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3825/commits/5bafd9611183e9eab044ad7632077823cdb21197

tests/upgrade_tests/test_firmware_upgrades.py
andrewkozlik: > I would maybe also add a version-gated check that if you set up your device on fw < 2.7.1, the backup will not have the extendable flag despite being generated on 2.7.1 and up

https://github.com/trezor/trezor-firmware/pull/3825/commits/0b480e7d3b6799644e371ddcb318207bb872c248

> I suspect that testing this is not enough. We (also?) need to test that passphrased addresses come out the same after upgrade.

I do check on [line 358](https://github.com/trezor/trezor-firmware/pull/3825/files#diff-08c67d05fcaafcc8efd599cf286f0aeda1f219e5b802f0dcae847c54416cb1d6R358) that an address of the standard wallet remains the same, which should be enough, considering that a standard wallet is a passphrased wallet with empty passphrase.

```py
assert btc.get_address(emu.client, "Bitcoin", PATH) == address
```

I am having trouble implementing the same check for a passphrased wallet. The following code returns the same address, even if I set `passphrase_protection=True` in `ResetDevice` and `_passphrase_source=2` in `ApplySettings`. Not sure whether it's worth figuring out.

```py
emu.client.use_passphrase("TREZOR")
btc.get_address(emu.client, "Bitcoin", PATH)
```

matejcik: I believe you'd need to insert an `emu.client.initialize(new_session=True)` before switching passphrase.

you are _probably_ right that the standard wallet address is a good check, but using two different passphrases is probably a better check either way

andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3825/commits/951ac292f3af662974201d69fc6f5a2278661ad1

matejcik (approved):

PR #3835, opened 2024-05-17
Drawing library - overlays for complex animations

cepetr:
This PR introduces several new functionalities to the drawing library, enhancing its capabilities for advanced drawing, particularly useful for more complex animations.

You can explore how to use these new concepts in the `screen3.rs` file on the `cepetr/drawlib-anim` temporary branch.

The PR is divided into seven commits:

1) Adds support for blending a `mono8` bitmap into an `rgb565` or `rgba8888` bitmap.
2) Improves the sector of circle drawing; it now uses `f32` angles instead of integers, which slightly improves precision for circles with larger radius.
3) The `f32` angle has been incorporated into UI components, which seems to have simplified things a bit.

4) Adds support for a temporary `ImageBuffer` (with a mutable Canvas) that UI components can use to cache images between successive calls to the `render()` routine.
5) Introduces a new `RawImage` shape for rendering `ImageBuffer` in UI components.
6) Introduces a new `render_on_canvas` function that allows for simple rendering of any shape to an `ImageBuffer`.
7) Introduces a new special shape for the unlock-screen animation (`UnlockOverlay`)

Notes:
~~1. Currently, I am not satisfied with the changes made to `render_on_canvas`. If you have any ideas for improvement, let’s discuss them.~~
~~2. `UnlockOverlay` should probably be placed within the Mercury UI, as it is not intended for general use.~~


core/embed/rust/src/ui/shape/display/direct_canvas.rs
```diff
@@ -0,0 +1,24 @@
+use crate::ui::{
+    display::Color,
+    shape::{Canvas, DirectRenderer, DrawingCache},
+};
+
+use super::bumps::run_with_bumps;
+
+/// Creates the `Renderer` object for drawing on a the specified and invokes a
```
TychoVrahe, line 8: on a the?

core/embed/rust/src/ui/shape/utils/imagebuf.rs
```diff
@@ -0,0 +1,83 @@
+use crate::ui::{
+    geometry::Offset,
+    shape::{Bitmap, BitmapView, Canvas, CanvasBuilder},
+};
+
+/// Size of image buffer in bytes
+const IMAGE_BUFFER_SIZE: usize = 64 * 1024;
```
TychoVrahe, line 7: We should make sure never to use this on F4 or MCUs with less RAM in general. Lets add conditional compilation for model_mercury for now (perhaps for this entire file), but we should figure out better way how to allocate proper amount of RAM for this (and perhaps graphics in general) according to UI requirements and MCU used

core/embed/rust/src/ui/shape/unlock_overlay.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    fn cleanup(&mut self, _cache: &DrawingCache<'a>) {
+        // TODO: inform the cache that we won't use the zlib slot anymore
+    }
+
+    fn draw(&mut self, canvas: &mut dyn Canvas, cache: &DrawingCache<'a>) {
+        let bounds = self.bounds();
+
+        let overlay_buff = &mut unwrap!(cache.image_buff(), "No image buffer");
```
TychoVrahe, line 95: maybe image_buff should be divided and we should use dedicated cache.overlay_buff for this (which would also allow cleaner disable of this feature on models with insufficient memory)

core/embed/rust/src/ui/shape/unlock_overlay.rs
```diff
@@ -0,0 +1,116 @@
+use crate::ui::{
```
TychoVrahe, line 1: yeah lets move this to mercury. we can do it in this PR, create the cshape dir and place it there.

core/embed/rust/src/ui/shape/display/direct_canvas.rs
cepetr: fixed, see 2979759b390cd2f3188e103602e5a6449a9e66ea

core/embed/rust/src/ui/shape/unlock_overlay.rs
cepetr: moved, see 04062b570ee0df6740a9d158a81c69a8e0df987e

(Unfortunately, after this change, `cepetr/drawlib-anim` is broken.)

core/embed/rust/src/ui/shape/unlock_overlay.rs
cepetr: `image_buff` in `DrawingCache` serves as a temporary buffer for shapes that can be used within its `draw()` method for any purpose. Having it as a single image is more universal than splitting it into two distinguished images.

But I absolutely agree that we need to come up with a better approach to memory management within the shape library.

TychoVrahe: I was thinking that with the divided buffers we could use the overlay along with drawing stuff that needs the image_buff - thus it would be actually more useful having it divided. 

But we can leave it as is and solve later along with the memory management if you prefer, i will leave that up to you

TychoVrahe (approved):

matejcik (changes requested):

core/embed/rust/src/ui/shape/display/bumps.rs
```diff
@@ -0,0 +1,69 @@
...
+    F: FnOnce(&'a mut Bump<[u8; BUMP_A_SIZE]>, &'a mut Bump<[u8; BUMP_B_SIZE]>),
+{
+    static mut LOCKED: bool = false;
+
+    // SAFETY:
+    // The application is single-threaded, so we can safely use a
+    // static variable as a lock against nested calls.
+
+    if unsafe { LOCKED } {
+        panic!("nested run_with_bumps!");
```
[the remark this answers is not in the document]
cepetr: see this change (calling ensure! instead of panic!) fb0604c343b5b2a3cd9d2ad2751b4ce15b2169c2,

core/embed/rust/src/ui/shape/display/bumps.rs
```diff
@@ -0,0 +1,69 @@
...
+///
+/// The function is passed two bump allocators, `bump_a` and `bump_b`, which
+/// can be used to allocate memory for temporary objects.
+///
+/// The function calls cannot be nested. The function panics if that happens.
+pub fn run_with_bumps<'a, F>(func: F)
+where
+    F: FnOnce(&'a mut Bump<[u8; BUMP_A_SIZE]>, &'a mut Bump<[u8; BUMP_B_SIZE]>),
+{
+    static mut LOCKED: bool = false;
```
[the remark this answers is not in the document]
cepetr: Finally I've decided to uses `for <'a>` trait bounds. Unfortunatelly it leads to changes in many files, see 82771429c1d9b7f0625ac8c2a76a84d87946c54a

matejcik (approved):

PR #3836, opened 2024-05-17
build(core): move default values for bindgen macros (back) into build.rs

matejcik:
this way Cargo build can run without the envvar set, e.g. in rust-analyzer

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -43,8 +43,6 @@ OPENOCD_INTERFACE ?= stlink
 # OpenOCD transport default. Alternative: jtag
 OPENOCD_TRANSPORT ?= hla_swd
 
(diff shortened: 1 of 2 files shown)


cepetr (approved):
It looks like this change has no influence on builds. It simply enables the use of Rust Analyzer for now, **and it works for me.** 

I see there's potential for improvement—not by duplicating definitions that already exist elsewhere, but that's probably not the focus at the moment.

TychoVrahe (approved):
utACK. 

@cepetr the definitions are otherwise created dynamically during scons build, so there is no obvious/simple way to pass them to stuff like rust analyzer without duplication, but conceptually you are right that this is a room for improvement

PR #3848, opened 2024-05-22
Update of en.json

obrusvit:
`en.json` according to latest figma.

Preparation for translations, not everything is used in code yet.


core/embed/rust/src/translations/generated/translated_string.rs
```diff
@@ -2523,6 +2568,51 @@ impl TranslatedString {
             Self::confirm_total__fee_rate => "Fee rate",
             Self::send__incl_transaction_fee => "incl. Transaction fee",
             Self::send__total_amount => "Total amount",
+            Self::instructions__continue_holding => "Continue\nholding",
+            Self::instructions__hold_to_continue => "Hold to continue",
+            Self::reset__cancel_create_wallet => "Cancel create wallet",
+            Self::backup__title_create_wallet_backup => "Create wallet backup",
+            Self::backup__your_wallet_backup_contains => "\"\"",
```
mmilata, line 2575: Looks suspicious, missing from en.json?

core/translations/order.json
```diff
@@ -558,7 +558,7 @@
   "556": "reset__needed_to_form_a_group",
   "557": "reset__needed_to_recover_your_wallet",
   "558": "reset__never_make_digital_copy",
-  "559": "reset__num_of_share_holders_template",
+  "559": "reset__num_of_shares_how_many",
```
mmilata, line 561: Is this a backwards compatible change? If an older firmware uses newer translations it will expect a template but won't get one.

mmilata (commented):
Looking alright, please update (at least T2T1, T2B1) fixtures.

core/translations/order.json
obrusvit: Thank you, this is an omission actually, I decided to use both for now:
- `reset__num_of_share_holders_template` for TT and TS3
- `reset__num_of_shares_how_many` for TS5

mmilata (approved):

PR #3853, opened 2024-05-23
homescreen/lockscreen for mercury

TychoVrahe:
Implements homescreen and animated lockscreen for mercury_ui.

Along with that:
- the 1s wait on startup is removed for models other than T2B1 and T2T1
- to prevent flooding with TOUCH_MOVE events, these are only allowed once per 10ms
- framebuffer switching waits for another sync signal, to ensure that the frame is shown before sending another
- ui refreshes are limited to be only done when layout is re-painted (as the refresh when performing framebuffer copy is slow)
- some other minor fixes


core/embed/rust/src/ui/component/base.rs
```diff
@@ -208,28 +208,44 @@ where
...
+        }
     }
 
     pub fn skip_paint(&mut self) {
-        self.inner.skip_paint()
+        if let Some(ref mut c) = self.inner {
+            c.skip_paint();
+        } else {
+            fatal_error!("deallocated", "Root object is deallocated")
+        }
```
matejcik, line 236: extract this into `pub fn inner_mut(&mut self)` and then call `self.inner_mut().skip_paint()` etc. in the other functions

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -291,6 +303,7 @@ impl LayoutObj {
                 Qstr::MP_QSTR_request_complete_repaint => obj_fn_1!(ui_layout_request_complete_repaint).as_obj(),
                 Qstr::MP_QSTR_trace => obj_fn_2!(ui_layout_trace).as_obj(),
                 Qstr::MP_QSTR_bounds => obj_fn_1!(ui_layout_bounds).as_obj(),
+                Qstr::MP_QSTR_delete => obj_fn_1!(ui_layout_delete).as_obj(),
```
matejcik, line 306: call it `MP_QSTR___del__`

core/embed/rust/src/ui/component/base.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/ee7f88dddc0c6a9e9708e8a30c045a4e411ce7fe

core/embed/rust/src/ui/layout/obj.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/ee7f88dddc0c6a9e9708e8a30c045a4e411ce7fe

core/embed/rust/src/ui/component/base.rs
```diff
@@ -208,28 +208,48 @@ where
...
     }
 
     pub fn skip_paint(&mut self) {
-        self.inner.skip_paint()
+        self.inner_mut().skip_paint();
     }
 
     pub fn clear_screen(&mut self) {
-        self.marked_for_clear = true;
```
matejcik, line 232: nit: i'm thinking, don't bother checking `self.inner` here if you don't need it? the requirement is "you must not corrupt memory if someone calls a function on the deleted object", we don't care if someone calls `clear()` on it and it does nothing.

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/26a1d1695336fcf6b2d7cf082cc8852292987ed3

core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
```diff
@@ -165,12 +178,14 @@ display_fb_info_t display_get_frame_buffer(void) {
 
 void display_refresh(void) {
 #ifndef BOARDLOADER
-  wait_for_fb_switch();
-  display_panel_set_window(0, 0, DISPLAY_RESX - 1, DISPLAY_RESY - 1);
 
   if (is_mode_handler()) {
+    display_panel_set_window(0, 0, DISPLAY_RESX - 1, DISPLAY_RESY - 1);
```
cepetr, line 183: If a DMA operation is running in the background, there will be two concurrent writers, and the display content will definitely be corrupted. For example, we should stop the DMA before doing this.

core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
```diff
@@ -63,45 +63,58 @@ void display_physical_fb_clear(void) {
...
-  if (current_frame_buffer == 1) {
-    bg_copy_start_const_out_8((uint8_t *)physical_frame_buffer_1,
-                              (uint8_t *)DISPLAY_DATA_ADDRESS,
-                              DISPLAY_RESX * DISPLAY_RESY * 2);
-
-  } else {
-    bg_copy_start_const_out_8((uint8_t *)physical_frame_buffer_0,
-                              (uint8_t *)DISPLAY_DATA_ADDRESS,
-                              DISPLAY_RESX * DISPLAY_RESY * 2);
+  if (pending_fb_switch == 1) {
```
cepetr, line 72: It seems to me that this new algorithm (waiting for the TE interrupt twice) can slow down `display_refresh()`, which might prevent us from achieving animation at full framerate. Am I right?

cepetr (changes requested):

core/src/session.py
```diff
@@ -27,5 +27,7 @@
 
 loop.run()
 
+workflow.kill_default()
```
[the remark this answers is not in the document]
TychoVrahe: yes this is needed, i think because the lockscreen is started again as the default task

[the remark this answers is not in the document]
matejcik: most likely not needed after we added the gc finalizer stuff
removed in https://github.com/trezor/trezor-firmware/pull/3853/commits/7de7e33002e7cf660f0dc8212bfd318a78ffa700

core/src/boot.py
```diff
@@ -20,7 +20,10 @@
 
 from apps.common.request_pin import can_lock_device, verify_user_pin
 
-_WELCOME_SCREEN_MS = 1000  # how long do we want to show welcome screen (minimum)
+if utils.INTERNAL_MODEL in ("T2T1", "T2B1"):
```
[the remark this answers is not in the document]
matejcik: https://github.com/trezor/trezor-firmware/pull/3853/commits/460c108d4aa6790e2f4e0ed4c5b82212fb9c2104

core/embed/trezorhal/stm32f4/touch/ft6x36.c
```diff
@@ -262,7 +262,9 @@ uint32_t touch_read(void) {
   if (touch_data[1] == GESTURE_NO_GESTURE) {
     xy = touch_pack_xy((X_POS_MSB << 8) | X_POS_LSB,
                        (Y_POS_MSB << 8) | Y_POS_LSB);
-    if ((number_of_touch_points == 1) && (event_flag == EVENT_PRESS_DOWN)) {
```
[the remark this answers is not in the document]
TychoVrahe: I actually cannot see a functional difference between the two variants, seems to me that both are doing the same change - adding a start event in case `EVENT_CONTACT` and not previously `touching`, or am i missing something?

core/embed/extmod/modtrezorio/modtrezorio-poll.h
```diff
@@ -84,38 +85,46 @@ STATIC mp_obj_t mod_trezorio_poll(mp_obj_t ifaces, mp_obj_t list_ref,
...
-              exr = DISPLAY_RESY - ey;
-              eyr = ex;
-              break;
-            default:
-              exr = ex;
-              eyr = ey;
-              break;
+          // ignore TOUCH_MOVE events if they are too frequent
+          if ((evt & TOUCH_MOVE) == 0 ||
+              (hal_ticks_ms() - last_touch_sample_time > 10)) {
```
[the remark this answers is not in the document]
TychoVrahe: i added an issue https://github.com/trezor/trezor-firmware/issues/3876

core/embed/firmware/memory_T3T1.ld
```diff
@@ -110,6 +105,11 @@ SECTIONS {
     . = ALIGN(4);
   } >SRAM3
 
+  .buf : ALIGN(4) {
+    *(.buf*);
+    . = ALIGN(4);
+  } >SRAM3
+
```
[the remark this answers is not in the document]
TychoVrahe: I increased the size of ImageBuffer so that it can fit the whole 240x240px jpg - which will be needed for animating homescreen. It would not fit the SRAM1, so i moved the smaller stuff in `.buf` to SRAM3, to achieve smaller effect on micropython heap size than moving the image buffer there.

core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
TychoVrahe: Probably yes. But i needed to prevent situation when two fast refreshes happened and then the first frame was not fully drawn on the display. Theoretically, i think one TE should be enough to show the frame - but observations told different story. Perhaps there is another problem and we should investigate this further when we have time.

core/embed/trezorhal/stm32f4/touch/ft6x36.c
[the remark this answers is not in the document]
TychoVrahe: using your version anyway as it seems more readable
https://github.com/trezor/trezor-firmware/commit/ddaf625fb40c21d0e5614b3c8b47c2d536f2b467

core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/45db71541d58b4f9dc8e557f914ae4a8d5bc395f

core/embed/trezorhal/stm32f4/touch/ft6x36.c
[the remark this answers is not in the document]
TychoVrahe: I mean the previous version. Lets keep the latest one for more thorough evaluation then.

Hannsek:
Please adjust the height of the background behind label to 38 px (instead of 42 px). And move the text into the (vertical) center of the background.

So that the background won't overlay the default homescreen. See [here].(https://www.figma.com/design/HKvwhavPNHg7DQoJClWxxm/%E3%80%94-Future-Prod-%E3%80%95Safe-5?node-id=4237-346&t=J0lgrdiCHBbRjsF4-4)

TychoVrahe:
@Hannsek lets do that in separate PR along with the bg image change

matejcik (approved):

STew790:
**QA OK**
The animated lock screen looks nice! 

**Info:**
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: Trezor T3T1 2.7.1 regular (revision 924492ce1b40656762296516734ee91201fb2a22)

PR #3861, opened 2024-05-24
Introduce device_secret and cred_auth_key_counter

M1nd3r:
Introduction of a `device_secret` - a 16 byte private random value independent of seed stored in persistent memory.

### DIFF
--- a/core/src/storage/device.py
+++ b/core/src/storage/device.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING
 
 import storage.cache as storage_cache
 from storage import common
+from trezor import utils
 
 if TYPE_CHECKING:
     from trezor.enums import BackupType
@@ -35,6 +36,10 @@ INITIALIZED                = const(0x13)  # bool (0x01 or empty)
 _SAFETY_CHECK_LEVEL        = const(0x14)  # int
 _EXPERIMENTAL_FEATURES     = const(0x15)  # bool (0x01 or empty)
 _HIDE_PASSPHRASE_FROM_HOST = const(0x16)  # bool (0x01 or empty)
+
+if utils.USE_THP:
+    _DEVICE_SECRET         = const(0x17)  # bytes
+    _CRED_AUTH_KEY_COUNTER = const(0x18)  # bytes
 
 SAFETY_CHECK_LEVEL_STRICT  : Literal[0] = const(0)
 SAFETY_CHECK_LEVEL_PROMPT  : Literal[1] = const(1)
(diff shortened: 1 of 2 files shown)


core/src/storage/device.py
```diff
@@ -344,3 +349,29 @@ def get_hide_passphrase_from_host() -> bool:
...
+    def get_cred_auth_key_counter() -> bytes:
+        counter = common.get(_NAMESPACE, _CRED_AUTH_KEY_COUNTER)
+        if counter is None:
+            counter = (0).to_bytes(4, "big")
+            common.set(_NAMESPACE, _CRED_AUTH_KEY_COUNTER, counter)
+        return counter
+
+    def increment_cred_auth_key_counter() -> None:
+        counter = int.from_bytes(get_cred_auth_key_counter(), "big")
+        common.set(_NAMESPACE, _CRED_AUTH_KEY_COUNTER, (counter + 1).to_bytes(4, "big"))
```
prusnak, line 377: Is int overflowing here a feature or a bug?

M1nd3r: Well, it is a bit of both.

The `increment_cred_auth_key_counter` function is called only when user wants to invalidate all credentials issued by Trezor, but not "wipe the storage". The counter has 4 bytes. I cannot imagine any "normal" scenario when this overflow is an expectable outcome. 

On the other hand, there should be a defined behavior for this unlikely scenario. I would like to know the opinions of @matejcik and @andrewkozlik . An argument against the current "silent overflow" is that it could make previously invalidated credentials valid again (until a `device_secret` is re-generated). However, it is unclear to me how to handle a potential overflow error - what should be the message to the user? If overflow happens, we could forbid user to invalidate credentials without also invalidating (re-generating) the `device_secret` - but it seems like a rather complicated UX.

andrewkozlik: Python's `to_bytes()` would raise an overflow exception, but micropython doesn't. Since this operation requires user interaction, there is no way that it is going to overflow under any kind of normal usage conditions. I'd raise the `OverflowError` "manually", or better yet, use `utils.ensure()` to protect against malicious scenarios.

prusnak: Btw I was thinking about it before ...

wouldn't it make sense to introduce a function `common.inc(namespace, variable)` that does the increment and checks for overflows?

core/src/storage/device.py
```diff
@@ -344,3 +349,29 @@ def get_hide_passphrase_from_host() -> bool:
...
+
+            device_secret = random.bytes(16, True)
+            common.set(_NAMESPACE, _DEVICE_SECRET, device_secret)
+        return device_secret
+
+    def get_cred_auth_key_counter() -> bytes:
+        counter = common.get(_NAMESPACE, _CRED_AUTH_KEY_COUNTER)
+        if counter is None:
+            counter = (0).to_bytes(4, "big")
+            common.set(_NAMESPACE, _CRED_AUTH_KEY_COUNTER, counter)
```
andrewkozlik, line 372: No need to set the 0 value in storage. Most people will probably never invalidate their pairing credentials on Trezor, so no need to waste space. The code can be simplified:
```py
return common.get(_NAMESPACE, _CRED_AUTH_KEY_COUNTER) or bytes(4)
```

M1nd3r: ACK. Fixed in [5dce7cc](https://github.com/trezor/trezor-firmware/pull/3861/commits/5dce7cc836524bdc93f590bb8915c7098e3d1d76)

core/src/storage/device.py
M1nd3r: utils.ensure added in [5dce7cc](https://github.com/trezor/trezor-firmware/pull/3861/commits/5dce7cc836524bdc93f590bb8915c7098e3d1d76)

M1nd3r: test for overflow added in [8a4b39d](https://github.com/trezor/trezor-firmware/pull/3861/commits/8a4b39da4eafc5bf69cf3caf7ab4aee5f6d04d34)

M1nd3r: @matejcik what is your opinion on @prusnak 's proposal to introduce `common.inc` function, do you think it should be a part of this PR?

matejcik: we have just this one counter though, don't we? i'd add `common.inc` only if we ever implement more

M1nd3r: Well, there is for instance `U2F_COUNTER`. However, it is used as an `int` and there is no need to access the "current counter value". It has functions `set_u2f_counter` and `next_u2f_counter` implemented in the `device.py`.

IMO I would keep it as it is now.

andrewkozlik: FTR, there is a big difference between the U2F counter and this one in that the U2F counter is not encrypted, which means it can use an optimized implementation in bitwise-writable flash storage such as that on the STM32F series. (The STM32U5 flash is not bitwise-writable.) This optimization is worthwhile only for counters that are incremented often, e.g. with every read.

andrewkozlik (approved):

PR #3862, opened 2024-05-25
Add THP credential manager

M1nd3r:
This PR provides derivation, validation, and invalidation of pairing credential authentication keys.


core/src/apps/thp/credential_manager.py
```diff
@@ -0,0 +1,108 @@
...
+
+def issue_credential(
+    host_static_pubkey: bytes,
+    credential_metadata: ThpCredentialMetadata,
+) -> bytes:
+    """
+    Issue a pairing credential binded to the provided host static public key
+    and credential metadata.
+    """
+    return _issue_credential(
```
andrewkozlik, line 53: Why aren't `_issue_credential()` and `_validate_credential()` inlined? I don't see them being used anywhere else.

python/src/trezorlib/messages.py
andrewkozlik, line 1: Technically, the new definitions shouldn't be in this file, since they are internal to Trezor. If we introduce a new MessageOption `bool internal_message` in `common/protob/messages.proto`, apply it to these messages and implement the option in `common/protob/pb2py`, then we can filer these definitions out in `python/src/trezorlib/_proto_messages.mako`. Is it worth the trouble, though?

core/src/apps/thp/credential_manager.py
M1nd3r: My idea was to have the !credential key derivation" decoupled from the "credential issuance/validation". As, you mentioned, it is not used anywhere else, so it is unnecessary. 

-> Inlined.

python/src/trezorlib/messages.py
M1nd3r: It is a relatively minor change, so I think that it should be done.

The addition of the message option is here: [a9e836b](https://github.com/trezor/trezor-firmware/pull/3862/commits/a9e836b2548dc1f03875ef21903f6589ec7f2d78)

The fixup of the thp messages is here: [9a8d8db](https://github.com/trezor/trezor-firmware/pull/3862/commits/9a8d8db86a7ee099b081df34013a418780ca79c0)

M1nd3r: Changes in pb2py and protobuf_build reverted in [f351bc3](https://github.com/trezor/trezor-firmware/pull/3862/commits/f351bc378710bb3c2bece71b7b0ed3b9ec6b5f92) and  [68126a1](https://github.com/trezor/trezor-firmware/pull/3862/commits/68126a119a32e3c53eee26407d96ea2ecef33dcc)
- @matejcik deems unnecessary to remove the proto definitions from trezorlib messages. However, we decided to keep the flag "internal-only".

Thp proto files are excluded from compilation into python and rust (results in changes in builds, not in git tracked files) here: [d58c9c6](https://github.com/trezor/trezor-firmware/pull/3862/commits/d58c9c6a72667ddabee0957eaaf4d974374485f0)

core/src/all_modules.py.mako
```diff
@@ -79,12 +80,17 @@ ${import_name}
 import ${import_name}
 % endfor
 
+if utils.USE_THP:
```
M1nd3r, line 83: @matejcik Do you want to have this change here? This hides all THP python files behind compilation flag USE_THP. I am not sure, if is is wanted/necessary.

matejcik: yeah let's keep this change

common/protob/messages-thp.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto2";
+package hw.trezor.messages.thp;
+
+// Sugar for easier handling in Java
+option java_package = "com.satoshilabs.trezor.lib.protobuf";
+option java_outer_classname = "TrezorMessageThp";
+
```
M1nd3r, line 7: @mmilata Here should possibly be 

`option (include_in_bitcoin_only) = true;`

What do you think?

mmilata: yep definitely

mmilata (commented):
Looking good, found couple of small things.

common/protob/messages.proto
```diff
@@ -54,8 +54,9 @@ extend google.protobuf.EnumOptions {
 
 /** Options for tagging message types */
 extend google.protobuf.MessageOptions {
-    optional bool experimental_message = 52001;   // indicate that a message is intended for development and beta testing only and its definition may change at any time
-    optional uint32 wire_type = 52002;   // override wire type specified in the MessageType enum
+    optional bool experimental_message = 52001;  // indicate that a message is intended for development and beta testing only and its definition may change at any time
+    optional uint32 wire_type = 52002;           // override wire type specified in the MessageType enum
+    optional bool internal_only = 52003;         // indicate that a message is intended for internal use only and should not be transmitted via the wire
```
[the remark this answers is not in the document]
M1nd3r: Ack.

Do you think it we should add the `internal_only` to `EthereumTokenInfo` and other messages without associated wire id?

core/src/apps/thp/credential_manager.py
```diff
@@ -0,0 +1,90 @@
...
+def validate_credential(
+    encoded_pairing_credential_message: bytes,
+    host_static_pubkey: bytes,
+) -> bool:
+    """
+    Validate a pairing credential binded to the provided host static public key.
+    """
+    cred_auth_key = derive_cred_auth_key()
+    expected_type = protobuf.type_for_name("ThpPairingCredential")
+    credential = wrap_protobuf_load(encoded_pairing_credential_message, expected_type)
```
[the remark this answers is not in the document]
M1nd3r: I agree with the second point that the experimental_enabled is irrelevant. However, I do not agree with the first point.
- The logging happens only on debug and basically has no effect on prod 
- It makes sense to log the received credential for testing purposes, eg. it can help detecting problems with host's implementation of credential handling
- `wrap_protobuf_load` already has exception handling and logging. The change you propose would lead to  code duplication - basically the only difference would be that the new "code" would not use flag experimental_enabled

common/protob/messages.proto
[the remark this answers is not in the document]
mmilata: Ok, I realized it's not only about wire id, even without it the message can be nested in a message that does have wire id.

Token info is not directly nested in any message but is transmitted over wire as an opaque blob, i.e. not all that internal. And the tool that generates it does so through trezorlib so the original meaning of `internal_only` doesn't apply either.

So IMO let's keep `internal_only` where it is now and perhaps improve the option description wrt when it should be used.

mmilata (approved):
LGTM, please rebase and merge if CI passes.

core/src/apps/thp/credential_manager.py
```diff
@@ -0,0 +1,90 @@
...
+    Issue a pairing credential binded to the provided host static public key
+    and credential metadata.
+    """
+    cred_auth_key = derive_cred_auth_key()
+    proto_msg = ThpAuthenticatedCredentialData(
+        host_static_pubkey=host_static_pubkey,
+        cred_metadata=credential_metadata,
+    )
+    authenticated_credential_data = _encode_message_into_new_buffer(proto_msg)
+    mac = hmac(hmac.SHA256, cred_auth_key, authenticated_credential_data).digest()
```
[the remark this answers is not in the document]
M1nd3r: added test to catch potential change in protobuf encoding in: https://github.com/trezor/trezor-firmware/pull/3862/commits/bed3414dc09c59c1e9718525a3929037d736a742

mmilata (approved):

PR #3872, opened 2024-05-29
USB driver refactoring

cepetr:
This PR focuses on refactoring the USB drivers. It does not fix any problem nor introduce new features but includes the following changes:

1. All USB-related implementation files in the trezorhal directory have been moved to the USB folder.
2. Unnecessary definitions were removed from public API headers, resulting in a cleaner API that contains only essential definitions.
3. Code implementing USB classes was moved from .h files to .c files. These files are no longer included but are compiled normally.
4. The USB classes implementation now provides a generic API to the USB driver, so the driver (implemented in `usbd.c`) is no longer aware of the implementation details.
5. The USB driver context is now stored in a single global variable `g_usb_driver`, making it clearer where the global driver context is stored and accessed.

These changes did not affect the high-level architecture of the driver or its API, as only very minor adjustments were made. They primarily create clearer boundaries between the driver components, allowing us to continue working on it safely.

The original goal was to understand and prepare the driver for power management features (e.g., suspending before entering low power modes and resuming after returning to full power).

usb-arch drawio


core/embed/trezorhal/stm32f4/usb/usbd_class_hid.c
```diff
@@ -46,12 +98,6 @@ secbool usb_hid_add(const usb_hid_info_t *info) {
     return secfalse;  // Not enough space in the configuration descriptor
   }
 
-  if ((info->ep_in & USB_EP_DIR_MASK) != USB_EP_DIR_IN) {
```
TychoVrahe, line 49: the removal of this check causes misconfiguration if the out endpoint has USB_EP_DIR_IN flag set - previously it would fail here, but now it is propagated. We should perhaps force clear of USB_EP_DIR_MASK bits

core/embed/trezorhal/stm32f4/usb/usbd_def.h
```diff
@@ -272,6 +272,8 @@ typedef struct _USBD_HandleTypeDef
 #define MAX(a, b)  (((a) > (b)) ? (a) : (b))
 #endif
 
+#define UNCONST(X) ((uint8_t *)(X))
```
TychoVrahe, line 275: Not that it is new in this PR but this whole section seems misplaced, and we have MIN, MAX macros in common.h too. In future, we might wanna consider refactoring this too

core/embed/trezorhal/stm32f4/usb/usbd_internal.h
```diff
@@ -0,0 +1,138 @@
...
+void usb_set_iface_class(uint8_t iface_num, const USBD_ClassTypeDef *class);
+
+// Returns the pointer to the next free bytes in the buffer
+// reserved for the class driver descriptors (interface, endpoint, ...)
+//
+// The function checks if the remaining space is enough and
+// returns NULL if not.
+void *usb_desc_alloc_iface(size_t desc_len);
+
+// Adds the class driver descriptors to the the configuration descriptor and
```
TychoVrahe, line 127: Some typos in this comment

core/embed/trezorhal/stm32f4/usb/usbd_internal.h
```diff
@@ -0,0 +1,138 @@
...
+// updates configuration descriptor length and number of interfaces.
+//
+// The class driver must call this function after all the its
+// descriptors have been written to the memory preallocated
+// by `usb_desc_alloc_iface`.
+void usb_desc_add_iface(size_t desc_len, uint8_t num_ifaces);
```
TychoVrahe, line 133: adding the possibility to add more interfaces would warrant name change to in my usb_desc_add_ifaces opinion. And possibly rename the desc_len to total_desc_len?

TychoVrahe (commented):
Adding some comments below, open for discussion.

Otherwise this seems like a very sensible cleanup and restructuring.

Didn't test.

core/embed/trezorhal/stm32f4/usb/usbd_class_hid.c
```diff
@@ -337,18 +386,44 @@ static int usb_hid_class_setup(USBD_HandleTypeDef *dev, usb_hid_state_t *state,
...
+    wait_random();
     // Save the report length to indicate we have read something, but don't
     // schedule next reading until user reads this one
     state->last_read_len = USBD_LL_GetRxDataSize(dev, ep_num);
   }
+
+  return USBD_OK;
 }
+
+static const USBD_ClassTypeDef usb_hid_class = {
```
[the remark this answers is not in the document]
cepetr: The definition of `usb_hid_class` depends on functions defined in this file, so it must appear after them. What is provided here is just a forward declaration because `usb_hid_class` is needed by `usb_hid_add`. We could remove this forward declaration, but it would require restructuring the file content: placing the `usb_hid_class_xxx()` functions at the beginning and `usb_hid` at the end. However, I'm not sure if we want to do this. Perhaps I don't fully understand your point.

core/embed/trezorhal/stm32f4/usb/usbd_class_hid.c
cepetr: Now, you can't pass the endpoint number with the `USB_EP_DIR_IN` flag set, as it is not available in the USB driver's public interface. You must and can only specify the endpoint number, ensuring that such a mistake can't happen.

core/embed/trezorhal/stm32f4/usb/usbd_def.h
cepetr: Yes, I agree with this, but I'm not sure if we should include `common.h` here, especially since we are gradually moving towards eliminating it altogether. I would prefer to wait with this until the fate of `common.h` is determined.

core/embed/trezorhal/stm32f4/usb/usbd_class_hid.c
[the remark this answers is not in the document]
TychoVrahe: Ah ok, makes sense. No need to restructure it.

core/embed/trezorhal/stm32f4/usb/usbd.c
```diff
@@ -79,86 +93,102 @@ static secbool __wur check_desc_str(const char *s) {
 }
 
 void usb_init(const usb_dev_info_t *dev_info) {
```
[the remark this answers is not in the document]
cepetr: Yes, I was considering this as well. There was inconsistency before in file names (some were `usb_*` and some `usbd_*`). What about naming everything with the `usbd_` prefix - both files and functions?

core/embed/trezorhal/stm32f4/usb/usbd_class_hid.c
TychoVrahe: You can just pass the endpoint out number as 0x81 for example, you don't need the macro available for that.

core/embed/trezorhal/stm32f4/usb/usbd.c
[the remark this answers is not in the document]
TychoVrahe: I am fine with either, i don't think even leaving it as is is a big problem, i just thought that renaming everything with usbd_ is much more invasive change

core/embed/trezorhal/stm32f4/usb/usbd_def.h
TychoVrahe: Fine with me. I created an issue for that https://github.com/trezor/trezor-firmware/issues/3948

core/embed/trezorhal/stm32f4/usb/usbd_class_hid.c
cepetr: Then maybe it's better to check for the maximum endpoint number. See the change 680f2de52b857e043b821eb61838b637ffbd2e23.

core/embed/trezorhal/stm32f4/usb/usbd.c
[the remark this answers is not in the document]
cepetr: Okay, I would change everything to `usbd_`, but let's do it as part of this review #3946 - it will be simpler to do later.

core/embed/trezorhal/stm32f4/usb/usbd_class_hid.c
cepetr: Finally, I found out more appropriate constant, see this change 51e6f541ce7bd526ba5e44239b102b71b4eb25a6

core/embed/trezorhal/stm32f4/usb/usbd_internal.h
cepetr: Finally, I decided to completely redo this part and get rid of this function altogether, see 12c155f6c1a9dda053535c8bbfbb6d6161fb5ace

core/embed/trezorhal/stm32f4/usb/usbd.c
```diff
@@ -38,39 +40,49 @@
...
+  usb_device_descriptor_t dev_desc;
+  // Device string descriptors
+  usb_dev_string_table_t str_table;
+  // Interfaces of registered class drivers
+  // (each class driver must add 1 or more interfaces)
+  usb_iface_t ifaces[USBD_MAX_NUM_INTERFACES];
+  // Buffer for configuration descriptor and additional descriptors
+  // (interface, endpoint, ..) added by registered class drivers
+  uint8_t desc_buffer[USB_MAX_CONFIG_DESC_SIZE] __attribute__((aligned(4)));
+  // Configuration descriptor (located at the beginning of the config_buf)
```
TychoVrahe, line 70: its desc_buffer now

core/embed/trezorhal/stm32f4/usb/usbd.c
```diff
@@ -189,50 +212,67 @@ secbool usb_configured(void) {
...
-  } else {
-    return NULL;  // Not enough space in the descriptor
+void usb_set_iface_class(uint8_t iface_num, const USBD_ClassTypeDef *class) {
+  usb_driver_t *usb = &g_usb_driver;
+
+  if (iface_num < USBD_MAX_NUM_INTERFACES) {
+    if (usb->ifaces[iface_num].class == NULL && class != NULL) {
+      usb->config_desc->bNumInterfaces++;
+    } else if (usb->ifaces[iface_num].class != NULL && class == NULL) {
+      usb->config_desc->bNumInterfaces--;
```
TychoVrahe, line 241: not sure how this unsetting works, wouldn't this break e.g. wTotalLength and expected position of the interface in the desc buffer?

cepetr: Yes, this setting was not necessary to add at all as nobody is using it. I added it just for completeness, even though no mechanism for removing classes (deallocating descriptors) has been implemented yet. I was a little bit concerned that it might lead to confusion, but I don't think it's harmful.

core/embed/trezorhal/stm32f4/usb/usbd.c
cepetr: fixed, see 61e95f3d4d270d14e36930288ba56f5ab399cfaa

core/embed/trezorhal/stm32f4/usb/usbd.c
TychoVrahe: I understand that it isn't called, so no real problem arises, but if someone calls this the descriptors will be in inconsistent state: regardless which index of interface you unset, the host will see the first few added. So unless the descriptor deallocation is properly implemented in the latter PR, which i havent seen yet, i would be for removing this functionality

cepetr: Ok, no problem with this. We may get it back in the future if needed. See d013de692c5f7358ea4d96a9c393160671daa6c8

TychoVrahe (approved):

PR #3874, opened 2024-05-31
T3T1 Rust tests

matejcik:
pointed at T3T1 branch but doesn't block its merging


matejcik:
@mmilata please double-check the layout splitting tests -- would these changes be expected?

mmilata (approved):
LGTM, thanks, the paragraph splitting looks legit

PR #3898, opened 2024-06-04
T3T1 UI: remove some leftover Model T layouts

mmilata:
Depends on #3887 but I'm setting `main` base to see the UI diff.

Most dialogs with buttons were replaced with "Swipe up" screen with cancel option in menu.

Firmware with debuglink fits in the flash (barely).


tests/device_tests/bitcoin/test_signmessage.py
```diff
@@ -316,6 +316,7 @@ def test_signmessage(
 
 @pytest.mark.skip_t1b1
 @pytest.mark.skip_t2b1
+@pytest.mark.skip_t3t1(reason="Not yet implemented in new UI")
```
matejcik, line 319: does this mean that SignMessage doesn't work now?

matejcik (approved):
always love to see file deletions :black_heart:

PR #3900, opened 2024-06-05
Touch driver refactoring

### LINKED ISSUE #3507: Touch panel events may be lost
There are few problems with implementation and processing touch panel event on T-model. 

1) When the user touches the display, touch controller starts generating events with 11ms period. The firmware
should readout those events from the touch controller registers via I2C interface. If the firmware doesn't readout the event, the controller overwrites it withby the succesive one => If the firmware is busy and is not calling touch_read() frequently enough, 
events may be lost and it's especially true for the `TOUCH_START` event.

2) The current `touch_read()` implementation discards all events if succesive calls doesn't occur within less than 50ms in order to not provide outdated events to UI components. This behaviour might lead in loosing events as well.

We didn't observe this behaviour until I began experimenting with animations on the display (when call to `touch_read()` are less frequent). 

We need to find out more suitable solution to both A) avoid loosing touch events and B) prevent providing outdated events to the UI.

Suggestions:
A) Process touch panel events within the interrupt routine. 
B) Add a timestamp to the touch event and process events only if their creating time is later than the time the target component became visible to the user.

### LINKED ISSUE #3732: Separate touch and button implementation in trezorhal/unix.
These are mangled together now but should be separated as stm hal.

cepetr:
This PR focuses on refactoring the touch screen controller drivers. It does not introduce new features but nearly fixes #3507 issue (so we can probably close it). 

1. Removed `lib/touch.c` since it's unused.
2. Polished the touch driver API (`touch.h`), resulting in fewer functions with clearer purposes. Also added the `touch_deinit()` function, allowing repeated driver initialization and deinitialization.
3. Refactored all drivers (emulator, discovery kits and FT6X36) to avoid using a many global variables, resulting in the driver state being clear and well-defined in a single location.
4. Heavily refactored the FT6X36 driver - removing duplicated code and simplifying complicated logic without affecting behavior.
5. Split the touch driver on the simulator into `touch.c` and `button.c` modules. (fixed #3732 )
6. Fixed a bug introduced in #3932 - `TOUCH VERSION` is functional again

The original goal was to prepare the touch controller driver for power management features (e.g., suspending before entering low power modes and resuming after returning to full power).


core/embed/trezorhal/stm32f4/touch/stmpe811.c
```diff
@@ -6,7 +6,7 @@
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
- * (at your option) any later version.
+ * (at your option) any later version.c
```
TychoVrahe, line 9: typo

core/embed/trezorhal/stm32f4/touch/stmpe811.c
```diff
@@ -647,62 +557,85 @@ void BSP_TS_GetState(TS_StateTypeDef *TsState) {
...
+  TS_StateTypeDef prev_state;
 
-  BSP_TS_GetState(&state);
+} touch_driver_t;
 
-  if (state.TouchDetected == 0) {
-    return 0;
-  }
+// Touch driver instance
+touch_driver_t g_touch_driver;
```
TychoVrahe, line 569: static

core/embed/trezorhal/stm32u5/touch/sitronix.c
```diff
@@ -1135,86 +1134,98 @@ static int32_t SITRONIX_Probe(uint32_t Instance) {
...
+typedef struct {
+  // Set if driver is initialized
+  secbool initialized;
+  // Last lower-level driver state
+  TS_State_t prev_state;
+
+} touch_driver_t;
+
+// Touch driver instance
+touch_driver_t g_touch_driver;
```
TychoVrahe, line 1147: static

core/embed/trezorhal/touch.h
```diff
@@ -1,32 +1,90 @@
...
+// We do not interpret the value of the version, we just print it
+// during the production test.
 uint8_t touch_get_version(void);
 
+// Sets touch controller sensitivity
+//
+// (Internally threadhsold for ????)
+secbool touch_set_sensitivity(uint8_t value);
+
+// Checks if the touch is currentl reporting any events
```
TychoVrahe, line 41: currently

core/embed/trezorhal/unix/button.c
```diff
@@ -0,0 +1,74 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include TREZOR_BOARD
+
+#ifdef USE_BUTTON
```
TychoVrahe, line 22: could we remove this ifdef in favor of only compiling this file for 'board' that support buttons? (likewise for USE_TOUCH)

core/embed/trezorhal/unix/touch.c
```diff
@@ -53,227 +51,230 @@ typedef enum {
...
          x - sdl_touch_offset_x < sdl_display_res_x &&
          y - sdl_touch_offset_y < sdl_display_res_y;
 }
 
-static bool is_button_swipe_initiated() {
-  return input_state == BUTTON_SWIPE_LEFT_INITIATED ||
-         input_state == BUTTON_SWIPE_RIGHT_INITIATED ||
-         input_state == BUTTON_SWIPE_UP_INITIATED ||
-         input_state == BUTTON_SWIPE_DOWN_INITIATED;
+static bool is_button_swipe_initiated(touch_driver_t* driver) {
```
TychoVrahe, line 76: driver can be pointer to const

core/embed/trezorhal/unix/touch.c
```diff
@@ -53,227 +51,230 @@ typedef enum {
...
+  touch_state_t state;
+  // Last valid coordinates
+  int last_x;
+  int last_y;
 
-static touch_input_state_t input_state = IDLE;
+} touch_driver_t;
+
+// Touch driver instance
+touch_driver_t g_touch_driver;
```
TychoVrahe, line 68: static

core/embed/trezorhal/unix/touch.c
```diff
@@ -53,227 +51,230 @@ typedef enum {
...
+secbool touch_ready(void) {
+  touch_driver_t* driver = &g_touch_driver;
+  return driver->initialized;
 }
 
-uint8_t touch_get_version(void) { return 0; }
-
-#endif
+secbool touch_set_sensitivity(uint8_t value) {
+  // Not implemented on the simulator
```
TychoVrahe, line 226: lets keep calling it emulator in code/comments

core/embed/trezorhal/unix/touch.c
```diff
@@ -53,227 +51,230 @@ typedef enum {
...
+  return sectrue;
+}
 
-#ifdef USE_BUTTON
+uint8_t touch_get_version(void) {
+  // Not implemented on the simulator
```
TychoVrahe, line 231: lets keep calling it emulator in code/comments

TychoVrahe (commented):
Some comments below. 

Didn't look too deep into discovery kit changes as they are not very important. Let me know if you think i should review it thoroughly

core/embed/trezorhal/stm32f4/touch/stmpe811.c
cepetr: fixed, see c04b50822941a7c6f979e8923827122438fe5749

core/embed/trezorhal/unix/touch.c
cepetr: fixed, see c04b50822941a7c6f979e8923827122438fe5749

core/embed/trezorhal/unix/touch.c
cepetr: fixed, see c04b50822941a7c6f979e8923827122438fe5749

core/embed/trezorhal/unix/touch.c
cepetr: yes, fixed, see 38441e0bc9e4a44a5a2cafff09a1118ca5142d7d

core/embed/trezorhal/stm32f4/touch/ft6x36.c
```diff
@@ -74,227 +148,278 @@ static void touch_default_pin_state(void) {
...
+  // (touch controller is powered down)
+  ft6x36_power_down();
 
-  if (sectrue != touch_set_mode()) {
+  // Power up the touch controller and perform the reset sequence
+  ft6x36_power_up();
+
+  // Read the firmware version
+  // (can be later read by `touch_get_version()`)
+  _Static_assert(sizeof(driver->fw_version) == 1);
```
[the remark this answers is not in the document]
cepetr: yes, it was weird idea. I removed it, see d641553c28edb27ab6bb424b34db6c01aa061d2f

core/embed/trezorhal/stm32f4/touch/stmpe811.c
cepetr: fixed, see a25e0d9312fd8a3398dcffab1f649eed50441c5f

core/embed/trezorhal/stm32u5/touch/sitronix.c
cepetr: fixed, see a25e0d9312fd8a3398dcffab1f649eed50441c5f

core/embed/trezorhal/unix/touch.c
cepetr: fixed, see a25e0d9312fd8a3398dcffab1f649eed50441c5f

core/embed/trezorhal/unix/button.c
cepetr: I've just removed these ifdefs, see 31b0bf98500a63a40288b8e7a441be82edd27394. Any changes in the build script are  not necessary.

core/embed/trezorhal/stm32f4/touch/ft6x36.c
```diff
@@ -74,227 +148,278 @@ static void touch_default_pin_state(void) {
...
-    } else if ((number_of_touch_points == 0) && (event_flag == EVENT_LIFT_UP)) {
-      touching = 0;
-      return TOUCH_END | xy;
+    }
+  } else if ((nb_touches == 0) && (flags == FT6X63_EVENT_LIFT_UP)) {
+    if (driver->pressed) {
+      // Finger was just lifted up
+      event = TOUCH_END | xy;
+    } else {
+      // Finger is still lifted up. Since we have already sent the
```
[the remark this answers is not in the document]
cepetr: I also considered this and have to admit I am not absolutely sure. On one side, this branch blocks sending `TOUCH_END` twice (it should not happen because of the interrupt line check, but who knows...). On the other side, it is really missed the `PRESS_DOWN` event. But if it is, then the touch duration was really short—just 10ms (with no `CONTACT` event in between).

I would rather keep it as it is since the proper solution requires simulating `TOUCH_START` and subsequent `TOUCH_END` in the next call, which is more complicated and probably brings no real value. We can of course discuss it further.

core/embed/bootloader/main.c
```diff
@@ -515,17 +514,22 @@ int bootloader_main(void) {
...
+      if (touch_activity() == sectrue) {
+        touched = 1;
         break;
       }
-#ifdef TREZOR_EMULATOR
-      hal_delay(25);
-#else
-      hal_delay_us(1000);
-#endif
+      hal_delay(5);
```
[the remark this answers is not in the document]
cepetr: In my opinion, the previous timeout might be too short to be on the safe side. According to the specs, we have to wait at least 300ms until the touch controller is ready. If somebody is touching the screen, it starts reporting events with a ~10ms period. Currently, we wait 50ms, so we have a chance to detect up to 5 events. Before, there was just a 10ms wait. I understand it worked before as well; maybe the 300ms requirement has a large margin. We can shorten it, but I am not sure that it's a good idea.

core/embed/trezorhal/stm32f4/touch/ft6x36.c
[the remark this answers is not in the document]
TychoVrahe: I am fine with that, just please adjust the comment so its clear that its intentional

core/embed/trezorhal/unix/button.c
TychoVrahe: Not necessary for sure, but should lower the compilation time both locally and also in CI, where we build a lot (though it remains very questionable whether it has some real effect), and its more consistent with how it is done in HW builds

core/embed/bootloader/main.c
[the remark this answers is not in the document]
TychoVrahe: yeah i know of zero issues with the previous setting. Otoh 40ms difference is negligible so we can stay on the safe side no prob.

core/embed/trezorhal/stm32f4/touch/ft6x36.c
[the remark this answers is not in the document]
cepetr: commented, see a96ade8a4b2b371145185d554c3ce3ff0feb8299

core/embed/trezorhal/unix/button.c
cepetr: I agree with this, but I believe such a change requires some preparation. Currently, `USE_TOUCH` and `USE_BUTTON` are defined in header files, so we cannot use them in build scripts. This situation suggests we should open a discussion about the consistent use of feature flags across C, Rust, and build scripts.

TychoVrahe: The build scripts are actually quite ready for this, see https://github.com/trezor/trezor-firmware/commit/7a8e2535c0ab930a8ea7bfcd7ff8744392f0fd5f - feel free to cherrypick. 

Yes we could also define `USE_TOUCH` and `USE_BUTTON` there or discuss some other solutions, but its not needed for this particular change.

cepetr: Nice, I didn't know about this. Cherry picked: 50f67d53db1934e2df49ddcc923e87ccbd96419e

TychoVrahe (approved):
lgtm, there are some style issues so please make sure to fix them before merging

PR #3901, opened 2024-06-05
T3T1 UI: more tap/hold prompts

mmilata:
Depends on #3887 and #3898 but setting `main` as base to see UI diff.


Hannsek:
@TychoVrahe isn't this the same what you are doing right now?

@matejcik should we put it into the new milestone?

TychoVrahe (approved):

PR #3903, opened 2024-06-05
Haptic driver refactoring

cepetr:
This PR focuses on refactoring the haptic controller driver. It does not fix any problems or introduce new features, but includes the following changes:

1. Removed `haptic_calibrate()` since it's unused, and using calibration requires more effort than just having this routine.
2. Removed the `HAPTIC_ALERT` effect since it's unused.
3. Polished the haptic driver API (`haptic.h`). Added the `haptic_deinit()` function, allowing repeated driver initialization and deinitialization, and ensured all functions that can fail return `bool`.
4. Refactored the driver to avoid using many global variables, resulting in the driver state being clear and well-defined in a single location.

The original goal was to prepare the haptic driver for power management features (e.g., suspending before entering low power modes and resuming after returning to full power).

### DIFF
--- a/core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
+++ b/core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
@@ -1,61 +1,46 @@
-#include "drv2625_lib.h"
-#include "haptic.h"
+/*
+ * This file is part of the Trezor project, https://trezor.io/
+ *
+ * Copyright (c) SatoshiLabs
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include TREZOR_BOARD
 
 #include <stdbool.h>
+#include <string.h>
+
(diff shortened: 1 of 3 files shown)


core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
```diff
@@ -75,45 +60,66 @@
...
+  // which would otherwise stop all playback.
+  bool playing_rtp;
 
-#define MAX_AMPLITUDE 127
-#define PRODTEST_EFFECT_AMPLITUDE 127
+} haptic_driver_t;
 
-bool haptic_enabled = true;
+// Haptic driver instance
+static haptic_driver_t g_haptic_driver;
```
TychoVrahe, line 77: General question. Should we explicitly zero-initialize these driver structs? they are implicitly zero-initialized but maybe would be easier to understand

TychoVrahe (approved):

core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
cepetr: That's a question. IMHO, it depends on whether such initialization leads to moving the variable into the `.data` section. If it does, I would rather leave it as it is. If not, it would be better to explicitly initialize it.

cepetr:
```c
static haptic_driver_t g_haptic_driver = {
  .initialized = false,
}
```

looks fine

cepetr: It seems that if the initialization results in zeroing the struct, GCC places it in the .bss section. So we can be explicit in these cases. See 0d4c7c481c2ca18e07cf8ea44c67e704f40b6cba

TychoVrahe: yeah that's what i would expect. 

i actually meant explicitly zeroing the whole struct like 
`static haptic_driver_t g_haptic_driver = {0}`

this also works, so feel free to go with a variant you prefer

PR #3904, opened 2024-06-05
fix(core/ui): T3T1: ethereum signtx hold

mmilata:
Add missing hold to confirm.

### DIFF
--- a/core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
+++ b/core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
@@ -63,6 +63,7 @@ impl FlowState for ConfirmAction {
                 Decision::Goto(ConfirmAction::Intro, SwipeDirection::Right)
             }
             (ConfirmAction::Menu, FlowMsg::Choice(0)) => Decision::Return(FlowMsg::Cancelled),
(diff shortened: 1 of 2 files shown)


obrusvit (approved):
utACK

Limitation: 
- ETH will have 1. Cancel, 2. More info (instead of the other way around)


If I read correctly we diverged from the original idea of `ConfirmActionSimple`. It's supposed to be:
- `ConfirmActionSimple` -> Intro, Menu

- `ConfirmAction` -> Intro, Menu, PromptScreen

mmilata:
> Limitation

For now the goal is to make both options available, we'll make it pretty in future release.

> we diverged from the original idea

I think it's just a matter of renaming `new_confirm_action_simple` to something more sensible. It still can return `ConfirmActionSimple` but the `if` construct has been reversed so it's not very visible in the diff.

bosomt:
**QA OK**

**Info:**
 - Suite version: web 24.7.0 (c00e0472d7b36073dc95265d90b83900c4f5d299)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:127.0) Gecko/20100101 Firefox/127.0
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.0 regular (revision 2e10618876be8a09a9d985098429f3016baf348e)

PR #3906, opened 2024-06-05
TEST: re-enable device tests for T2B1

matejcik:


matejcik:
seems fine. will merge in #3905

PR #3910, opened 2024-06-06
refactor(core): cleanup storage cache

### LINKED ISSUE #3865: use `storage.cache.set_bool`
#3743 introduces `storage.cache.set_bool()` / `get_bool()` pair.
in a couple place we define a 1-sized flags in cache and then check them against "true byte".
we should replace them with 0-sized flags and use `set_bool()` and `get_bool()` to manage them

ibz:
Fixes #3865


core/src/storage/cache.py
```diff
@@ -29,12 +29,11 @@
 # Keys that are valid across sessions
 APP_COMMON_SEED_WITHOUT_PASSPHRASE = const(0 | _SESSIONLESS_FLAG)
 APP_COMMON_SAFETY_CHECKS_TEMPORARY = const(1 | _SESSIONLESS_FLAG)
-STORAGE_DEVICE_EXPERIMENTAL_FEATURES = const(2 | _SESSIONLESS_FLAG)
```
andrewkozlik, line 32: What is the reason to remove the experimental features flag from the cache?

ibz: No real reason, just thought it seemed unnecessary while doing this cleanup. Do you think there was some use for it?

andrewkozlik: What about when the device is locked? I think that's the reason why it was added originally.

matejcik: a better question is, why was the flag there in the first place?

I believe the reason was that we made the storage field private for some reason, and then we couldn't see its value when the device is locked.
Do you think it was private deliberately, or did we fumble that one and had to patch it over?
This PR makes the field public (technically breaking the setting for people who touched it, but seeing as it's currently unused I don't think we care) and so can remove the cached value.

andrewkozlik: Looking at Git history leads me to https://github.com/trezor/trezor-firmware/pull/1565:
> 49447e8 makes sure that experimental_features behave as before; otherwise the value would get lost when the storage becomes locked.

matejcik: "public" and "public writable" are two separate modes for the storage, no? homescreen is also "public" but not "public writable".

i see no harm in making this field publicly readable (unencrypted), seeing as privacy implications are limited (and if we cared we can omit the info from `Features` while locked), and doing so removes a rather inscrutable piece of cache-and-storage logic, all for a feature that is not used by anyone and won't be for the foreseeable future

matejcik: > Looking at Git history leads me to #1565:
> 
> > [49447e8](https://github.com/trezor/trezor-firmware/commit/49447e8c368506f277ddfbe54c1a71147b67edfc) makes sure that experimental_features behave as before; otherwise the value would get lost when the storage becomes locked.

yeah that would lead credence to my story: we implemented experimental_features first, when there was no softlock, then we added softlock which broke it, and I plausibly didn't realize _at that time_ that publicizing the storage field is a much better fix for the same problem

andrewkozlik: > publicizing the storage field is a much better fix for the same problem

So what is the problem? Showing the information in `Features` when locked? That seems like a pretty weak reason to opt-out of authenticated encryption.

andrewkozlik: Thinking about this, the problem was probably that at that time coinjoin messages were experimental, and these messages could be processed in locked state, so we needed to know the experimental setting in locked state.

Anyway, it still seems like caching is the correct solution, because we don't need the setting to be unencrypted.

matejcik: > Thinking about this, the problem was probably that at that time coinjoin messages were experimental, and these messages could be processed in locked state, so we needed to know the experimental setting in locked state.

exactly

> Anyway, it still seems like caching is the correct solution, because we don't need the setting to be unencrypted.

We also don't need the setting to be encrypted.
Unless you have a strong objection, I prefer we remove the caching-related complexity from here.

(again, talking about developer-oriented feature which, quite honestly, we might very well just completely drop from here and reinstate it if we ever need it again, because it currently doesn't do anything)

andrewkozlik: > We also don't need the setting to be encrypted.

That _might_ be true, depending on the attacker model and depending on what features it blocks (currently none). If it is true, then the same can be said of auto lock delay and safety checks, for example.

> developer-oriented feature

Perhaps, but its purpose is to protect ordinary users from unfinished features that may be potentially harmful.

> Unless you have a strong objection, [...]

I am not vetoing the change from private to public, but I consider it unsystematic.

andrewkozlik: BTW, in case you choose to go ahead, then I think this part of the commit shouldn't be classified as a "refactor" but a "feat".

ibz: Alternatively: 1) "refactor" commit for all without touching this. 2) "feat" commit dropping the whole experimental flag (not just the cache).

We can always bring this back later, if we need it, and have it implemented according to the need at hand.

I feel like having this code around is confusing and the proof is this very discussion, and I don't see a good solution to this given that the code is not needed in the first place and all we can do is speculate about how it could be used.

matejcik: yes, that sounds reasonable actually.
we still need to keep the protobuf fields, but we can remove all the codepaths.

ibz: After all, dropping the whole feature seemed sad, because it is more complex than I thought (it includes checks for "experimental fields" when decoding protobuf, for example). So we decided to just drop the cache and thus make the storage field public. I made that part into a separate commit.

core/src/storage/__init__.py
```diff
@@ -1,8 +1,7 @@
-from storage import common, device
+from storage import cache, common, device
```
matejcik, line 1:
```suggestion
# make sure to import cache unconditionally on top level, so that it is imported (and retained) together with the storage module
from storage import cache, common, device
```

core/src/storage/__init__.py
matejcik, line 1: please make this change a separate commit, something like "fix: make sure `storage.cache` import is correctly retained"

matejcik (changes requested):

matejcik (approved):

PR #3916, opened 2024-06-06
chore(core): translate strings

ibz:
[no changelog]


core/translations/cs.json
```diff
@@ -611,6 +611,8 @@
     "recovery__title_recover": "Obnovit peněženku",
     "recovery__title_remaining_shares": "Zbývající části",
     "recovery__type_word_x_of_y_template": "Zadejte slovo {0} z {1}",
+    "recovery__unlock_repeated_backup": "Vytvořit další zálohu?",
+    "recovery__unlock_repeated_backup_verb": "Odemknout zálohu",
```
Hannsek, line 615: do we still use "unlock backup"?  We wanted to change it to "Create additional backup?"

ibz: This is the button. Here is how it looks...

image

Hannsek: Ok, we should change it to `Create backup` on the button. Can we do it?

ibz: Speaking of which, it does indeed not match your screenshot. Good catch. Let me change it now.

core/translations/cs.json
```diff
@@ -611,6 +611,8 @@
     "recovery__title_recover": "Obnovit peněženku",
     "recovery__title_remaining_shares": "Zbývající části",
     "recovery__type_word_x_of_y_template": "Zadejte slovo {0} z {1}",
+    "recovery__unlock_repeated_backup": "Vytvořit další zálohu?",
+    "recovery__unlock_repeated_backup_verb": "Vytvořte zálohu",
```
Hannsek, line 615:
```suggestion
    "recovery__unlock_repeated_backup_verb": "Vytvořit zálohu",
```

matejcik (approved):

PR #3922, opened 2024-06-07
fix bootloader lock check on STM32U5

TychoVrahe:
Fixes bootloader lock check, and consequently, device authentication check, which is refused when bootloader is unlocked.

### DIFF
--- a/core/.changelog.d/3922.fixed
+++ b/core/.changelog.d/3922.fixed
@@ -1,0 +1,1 @@
+[T3T1] Fixed device authenticity check
--- a/core/embed/trezorhal/stm32u5/secret.c
+++ b/core/embed/trezorhal/stm32u5/secret.c
@@ -38,7 +38,7 @@ secbool secret_ensure_initialized(void) {
 
 secbool secret_bootloader_locked(void) {
 #ifdef FIRMWARE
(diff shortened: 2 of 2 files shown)


matejcik (changes requested):
actually... changelog entry pls

does this affect bootloader after all? or is bootloader not using this file?

TychoVrahe:
it does not affect bootloader, since there is `#ifdef FIRMWARE `

Should we generate changlog first? there is bunch of older entries, or will we discard the previous release entirely (in which case one could argue that this falls under generic support for U5 entry)

Hannsek:
Let's discard the previous release if feasible. 🙏🏻
And let's re-release 2.7.1.

matejcik:
> Let's discard the previous release if feasible. 🙏🏻
> And let's re-release 2.7.1.

let's not.
2.7.1 is already signed, reusing that version does not have any benefits and only has drawbacks.
ideally we manage to release TS3 as 2.7.2 too

matejcik (approved):
> it does not affect bootloader, since there is `#ifdef FIRMWARE `

right, of course

cepetr (approved):
utACK

PR #3930, opened 2024-06-07
reduce animation frame requests when swiping

TychoVrahe:
Attempt to mitigate queue overflow when swiping.

The animation frame requests are useless in first swipe phase - as the animation is driver by touch move (or rather SwipeMove) events.


matejcik (approved):
utACK

PR #3932, opened 2024-06-09
prodtest - reboot command

TychoVrahe:
This PR adds Reboot command to prodtest. Device is immediately rebooted.

As part of this PR, touch version command is fixed - touch driver is powered on before polling the version, otherwise the command failed when issued after some other TOUCH test command - these power off the driver at the end. (if TOUCH VERSION is issued first it works, as touch is powered on from bootloader)


cepetr (approved):
utACK

PR #3935, opened 2024-06-10
fix(python): remove duplicated 0x prefix for publish sign-tx

r0tc:
This tiny pull request removes hardcoded `0x` prefix for sign-tx ethereum cli command with a publish flag. web3 lib adds it upfront.


matejcik (approved):
seems alright. thanks!

PR #3938, opened 2024-06-11
Improve error handling / RSOD screen

### LINKED ISSUE #744: Cleanup error_shutdown vs ensure(secfalse,msg)
It seems to me that we abuse `ensure(secfalse,msg)` where we actually want to use `error_shutdown`.

Let's discuss and clean this up.

cepetr:
This PR improves error handling in both C and Rust code:

and finally saves about 12KB of FLASH (without affecting the functionality)

1. Multiple implementations of error handling code were removed from `common.c`/`common.h` and moved to the platform-independent `lib/error_handling.h`/`.c`.
2. Duplicated code, especially in the simulator, was removed and replaced by calls to `error_shutdown()`.
3. Fixed invalid footer text in `show_pin_too_many_screen()` and `show_wipe_code_screen()`.
4. Simplified the interface of `error_shutdown()` and introduced `error_shutdown_ex()` for more complex cases.
5. Replaced calls to `__fatal_error()` with `error_shutdown()` as the former is low-level and private to `error_handling.c`.
6. Moved `show_install_restricted_screen()` to a more appropriate location (from `secret.c` to `error_handling.c`).
7. Final error handling for Bootloader/Firmware is now done in Rust in the `error_shutdown()` function, eliminating dual implementations in Rust and C.
8. Fixed the problem with nested calls to `render_on_display` when an error occurred during drawing on the T3T1 model.
9. Introduced the `dbg_trace!` macro, a `print!`-like macro for printing to the debug console.
10. Added missing error locations in console output when raising `ensure!`, `unwrap!`, and `fatal_error!`.
11. Replaced calls to `panic!` by calls `to fatal_error!`. `fatal_error!` has now just single parameter - `message`
12. Redefined `assert()` and `ensure()` macro to not print evaluated expression and filename with full path.
13. Also finally fixed #744

**IMPORTANT:** Unfortunately, (12) triggered a wave of changes in storage and legacy code, just because of the shared prototype of `__fatal_error()`. All affected code was changed appropriately, and as a result, the T1 model no longer reports the expression and function name on the fatal error screen (there's just the file and line - which is still more than on other models).


core/embed/lib/error_handling.c
```diff
@@ -0,0 +1,116 @@
...
+void __attribute__((noreturn)) error_shutdown(const char *message) {
+  error_shutdown_ex(NULL, message, NULL);
+}
+
+void __attribute__((noreturn))
+__fatal_error(const char *msg, const char *file, int line) {
+#ifdef FANCY_FATAL_ERROR
+  if (msg == NULL) {
+    char buf[128] = {0};
+    mini_snprintf(buf, sizeof(buf), "%s: %d", file, line);
```
matejcik, line 75: nit that always bothered me: should there be a space between `file: lineno`? i'm much more used to no space here

core/embed/rust/src/lib.rs
```diff
@@ -39,6 +39,9 @@ pub mod translations;
 pub mod ui;
 pub mod strutil;
 
+#[cfg(feature = "debug")]
+pub mod debug;
```
matejcik, line 43: i think you need `#[macro_use]` here to be able to use dbg_println from outside, no?

core/embed/rust/src/trezorhal/fatal_error.rs
```diff
@@ -5,44 +5,71 @@ mod ffi {
...
 
-#[cfg(not(feature = "bootloader"))]
-pub fn __fatal_error(_expr: &str, msg: &str, _file: &str, _line: u32, _func: &str) -> ! {
-    ModelUI::screen_fatal_error("INTERNAL_ERROR", msg, "PLEASE VISIT\nTREZOR.IO/RSOD");
-    ModelUI::backlight_on();
-    shutdown()
+/// Shows an error message on the screen and shuts down the device.
+/// In debug mode, also prints the error message to the console.
+#[inline(never)] // saves few kilobytes of flash
+pub fn __fatal_error(msg: &str, _file: &str, _line: u32) -> ! {
```
matejcik, line 37: given that you removed `func`, please also drop the horrible `function_name!` macro

core/embed/rust/src/ui/api/common_c.rs
```diff
@@ -22,7 +22,7 @@ extern "C" fn screen_fatal_error_rust(
     let msg = unsafe { from_c_str(msg) }.unwrap_or("");
     let footer = unsafe { from_c_str(footer) }.unwrap_or("");
 
-    ModelUI::screen_fatal_error(title, msg, footer);
+    fatal_error::error_shutdown(title, msg, footer)
```
matejcik, line 25: i'm a little surprised that this compiles. but you should probably annotate `error_shutdown_rust(...) -> !`

core/embed/rust/src/ui/shape/display/bumps.rs
```diff
@@ -32,10 +35,8 @@ pub fn run_with_bumps<F>(func: F)
 where
     F: for<'a> FnOnce(&'a mut Bump<[u8; BUMP_A_SIZE]>, &'a mut Bump<[u8; BUMP_B_SIZE]>),
 {
-    static mut LOCKED: bool = false;
-
     // SAFETY:
-    // The application is single-threaded, so we can safely use a
+    // The application is single-threaded, so we can safely use aP
```
matejcik, line 39: typo

matejcik (approved):
very nice. OK from me, someone else should also look at it from embedded POV.

while you're at this, maybe you would like to look at https://github.com/trezor/trezor-firmware/issues/744 ? (not required for this PR but it might be an opportunity to fold it in)

core/embed/rust/src/ui/api/common_c.rs
cepetr: fixed, see d44b6730f5d5aae946e51fe5c224bd4cbcccaa8e

core/embed/rust/src/lib.rs
cepetr: fixed, see abc2d798f6e08f0a443d261329420993526d09b0

core/embed/rust/src/trezorhal/fatal_error.rs
cepetr: removed, see d6f4168466119bdce402b112965816644c3cc6e9

core/embed/lib/error_handling.c
cepetr: no problem, I changed it, now it's unified with the similar code in Rust, see d6f4168466119bdce402b112965816644c3cc6e9

core/embed/rust/src/ui/shape/display/bumps.rs
cepetr: fixed, see e7a8f37a4f7f9ba80939425bbea13c6e0cb4488c

cepetr:
> while you're at this, maybe you would like to look at #744 ? (not required for this PR but it might be an opportunity to fold it in)

It looks like `error_shutdown(..)` can now perfectly replace the `ensure(secfalse, ...)` macro. The thing to consider is that we will lose file and line information in debug builds. If this is not a problem (and I think it isn't since all these `ensure(secfalse` instances contain quite unique messages), I can make this change within this PR. Should I? @matejcik

cepetr:
> > f this is not a problem (and I think it isn't since all these `ensure(secfalse` instances contain quite unique messages)
> 
> yeah, exactly. please go ahead and do it

I made the changes in commit af45de2149e34e0cc585b034a5ac8f0f7ec9018c:

1. In the storage, I've replaced `ensure(secfalse` with `handle_fault`.
2. In the storage tests, I've left it unchanged.
3. In the display/i2c drivers, I've changed it, but we should not call it at all and return with an error instead. However, this is not related to this PR.

matejcik:
> Aaaa, I didn't notice that. So I reverted this change here. In this file, it seems better to be consistent and not introduce a call to `error_shutdown`. See [8b23004](https://github.com/trezor/trezor-firmware/commit/8b23004d79ad0394558e2bfe969f9f59f940bd27)

it's helpful if you post in your own comment, instead of editing mine :)

otherwise, all good from my side

core/embed/rust/src/trezorhal/fatal_error.rs
```diff
@@ -78,19 +87,13 @@ macro_rules! unwrap {
 macro_rules! ensure {
     ($what:expr, $error:expr) => {
         if !($what) {
-            fatal_error!(stringify!($what), $error);
+            crate::trezorhal::fatal_error::__fatal_error($error, file!(), line!());
         }
     };
 }
 
 macro_rules! fatal_error {
```
TychoVrahe, line 95: i am somewhat unhappy about calling differently named function (or macro) in rust vs C code when i want to show error and shutdown. How about getting rid of this macro, create the same `error_shutdown` and `error_shutdown_ex` structure as was introduced in C and leave `__fatal_error` for internal purposes only (called from `ensure` and `unwrap`, while regular `error_shutdown` calls would lose the file and line info as C calls do).

Also, i think that we can move most of the code from this file (without the `shutdown` function that interfaces with C) to e.g. `rust/src/error_handling.rs`, possible move the `panic` implementation there too from lib.rs.

cc also @matejcik , what do you think?

core/embed/rust/src/trezorhal/fatal_error.rs
```diff
@@ -5,70 +5,79 @@ mod ffi {
...
     unsafe { ffi::trezor_shutdown() }
 }
 
-#[cfg(feature = "bootloader")]
-pub fn __fatal_error(_expr: &str, _msg: &str, _file: &str, _line: u32, _func: &str) -> ! {
-    ModelUI::screen_fatal_error("BL.rs", "BL.rs", "PLEASE VISIT\nTREZOR.IO/RSOD");
+/// Shows an error message and shuts down the device.
+pub fn error_shutdown(title: &str, msg: &str, footer: &str) -> ! {
+    // SAFETY:
+    // The is the only situation we are allowed use this function
```
TychoVrahe, line 23: typo This is

cepetr: fixed, see 640dee638a19a7ee831cfda8d92fb8cdafa5713b

core/embed/lib/error_handling.c
```diff
@@ -0,0 +1,116 @@
...
+  if (footer == NULL) {
+    footer = "PLEASE VISIT\nTREZOR.IO/RSOD";
+  }
+
+#ifdef FANCY_FATAL_ERROR
+  error_shutdown_rust(title, message, footer);
+#else
+  display_orientation(0);
+  term_set_color(COLOR_WHITE, COLOR_FATAL_ERROR);
+  if (title) {
```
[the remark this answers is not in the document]
cepetr: fixed, see 640dee638a19a7ee831cfda8d92fb8cdafa5713b

core/embed/lib/error_handling.c
```diff
@@ -0,0 +1,116 @@
...
+    term_printf("%s\n", title);
+  }
+  if (message) {
+    term_printf("%s\n", message);
+  }
+  if (footer) {
```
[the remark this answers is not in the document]
cepetr: fixed, see 640dee638a19a7ee831cfda8d92fb8cdafa5713b

core/embed/rust/src/trezorhal/fatal_error.rs
matejcik: It makes some amount of sense to me like it is now.
Our `fatal_error!` is an equivalent to `panic!`, which has just a single string reason. rust enriches it with file+line information and passes it on. This is an "internal error" with some fancy paint on it.
Or, you want to invoke an explicitly user-facing RSOD, then you call the _function_ `error_shutdown`.
These are two distinct operations. C folds them into kiiind of one-and-the-same, which I think is a small bit less semantically correct? But not enough for me to care.

In any case, I wouldn't want to introduce "error_shutdown_ex" (the "_ex" being a specifically C thing), and I don't have any other ideas that would be less confusing. We can do `error_shutdown!("just message")` and `error_shutdown("title", "message", "footer")`, or we can extend `error_shutdown!` to accept either three or one arguments, either sounds just weird.
(other options anyone?)

> Also, i think that we can move most of the code from this file (without the shutdown function that interfaces with C) to e.g. rust/src/error_handling.rs, possible move the panic implementation there too from lib.rs.

my followup PR already moves the macros. I would keep the functions (error_shutdown, __fatal_error) here, seeing as they all transitively depend on the ffi `shutdown()` call, so they are all trezorhal-y.
the traits could be moved, but perhaps do the opposite? keep them here, move the panic impls here too (is it legal to do it that way?) and call this whole module `shutdown` or something.

cepetr: Do you mind if we do it in a separate PR or in a follow-up PR #3951 to be able to merge this soon?

TychoVrahe: All right, i see your point, lets keep it as is then

TychoVrahe (approved):

PR #3942, opened 2024-06-11
T3T1 related tooling chores

mmilata:
- build with production bootloader in CI
- T3T1 changelog in `generate-changelogs.py`

### DIFF
--- a/tools/towncrier.toml
+++ b/tools/towncrier.toml
@@ -1,3 +1,6 @@
+# This is the default configuration that is used by subprojects via symlink.
+# Some subprojects have their own configuration which is not a symlink.
+
 [tool.towncrier]
 directory = ".changelog.d"
 filename = "CHANGELOG.md"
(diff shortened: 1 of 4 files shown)


tools/towncrier.toml
```diff
@@ -3,7 +3,7 @@ directory = ".changelog.d"
 filename = "CHANGELOG.md"
 template = ".towncrier.template.md"
 start_string = "The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)."
-title_format = "\n## {version} [{project_date}]"
+title_format = "\n## [{version}] ({project_date})"
```
matejcik, line 6: sure about this? last time I touched the global towncrier.toml I cried.
there are local versions in the subdirs

mmilata: i tried generating `core/CHANGELOG.md` and it looked alright, will try the others

matejcik (approved):

tools/towncrier.toml
matejcik: just checking. otherwise utACK, go ahead with the PR if this is indeed correct

mmilata: you're right, this breaks the format of the other monorepo subdirectories
f61f0a2659110835b758d69d8d7b6484a68e95a9

matejcik (approved):

Issue #3596, opened 2024-03-11
T3T1: emulate device authentication


mmilata:
Needs a certificate chain (e.g. [T2B1](https://github.com/trezor/trezor-firmware/blob/main/core/embed/trezorhal/unix/optiga.c#L27)) with leaf CN starting with `T3T1`. After this is done, remove `test_authenticate_device` from blacklist.

mmilata:
Fixed in #3944

PR #3944, opened 2024-06-12
Include a valid staging certificate for TS5 emulator

### LINKED ISSUE #3765: Replace T3T1 debug `caPubKey`
Debug value for `caPubKey` should be:
 `04829e8965018feb542e9236c9b2ce08f864a55ed9183d0259564f0e05345b04676a0bef36c59d21d3c24868b5601f0b1193a6bfcf6d814e1cfb79c2256a05e953`.
This is necessary to pass the check in Suite debug mode when emulator is used. Current value is:
`04ba6084cb9fba7c86d5d5a86108a91d55a27056da4eabbedde88a95e1cae8bce3620889167aaf7f2db166998f950984aa195e868f96e22803c3cd991be31d39e7`
It is the same key as for T2B1 - therefore validation fails [here](https://github.com/trezor/trezor-suite/blob/89c1cd5d8c2afe0bdac9d73cdf100ccaaa0b42f9/packages/connect/src/api/firmware/verifyAuthenticityProof.ts#L120).

Suite expects [these keys](https://github.com/trezor/trezor-suite/blob/a04421ba2342bddce4b5638a2508cb347b1387fc/packages/connect/src/data/deviceAuthenticityConfig.ts#L94).

matejcik:
fixes #3765 

includes certificate verification code and supersedes #3464
@andrewkozlik please have a look, I believe all your comments from #3464 are applied here


python/src/trezorlib/authentication.py
```diff
@@ -0,0 +1,303 @@
...
+
+from . import device
+from .client import TrezorClient
+
+try:
+    import cryptography  # noqa: F401
+
+    cryptography_version = [int(x) for x in cryptography.__version__.split(".")]
+    if cryptography_version[0] < 41:
+        raise ImportError("cryptography>=41 is required for this module")
```
prusnak, line 16: nit:
```suggestion
        raise ImportError(f"cryptography>=41 is required for this module, found cryptography=={cryptography.__version__}")
```

python/src/trezorlib/authentication.py
```diff
@@ -0,0 +1,303 @@
...
+        try:
+            ca_cert = Certificate(issuer_bytes)
+        except Exception:
+            LOG.error(f"Failed to parse CA certificate #{i}.")
+            failed = True
+            continue
+
+        if whitelist is None:
+            LOG.warning("Skipping public key whitelist check.")
+        else:
```
mmilata, line 226: What is the usecase for `whitelist`?

core/embed/trezorhal/unix/certs/T2B1.h
```diff
@@ -0,0 +1,79 @@
+static const unsigned char T2B1_der[] = {
+  0x30, 0x82, 0x01, 0x9f, 0x30, 0x82, 0x01, 0x45, 0xa0, 0x03, 0x02, 0x01,
```
mmilata, line 2: make cstyle changes indent on the generated files

mmilata (approved):
Did not spot any issue.

`test_authenticate_device` can now be reenabled for T3T1: f97105e40f7958409a582d384af0ac765c943030 (needs fixtures, closes #3596)

python/src/trezorlib/authentication.py
matejcik: linked to https://github.com/trezor/data/pull/79
basically you want to download a whitelist of public keys and check that your Trezor is included. this allows us to "revoke" a batch of Trezors for example

core/embed/trezorhal/unix/certs/gen.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+cd "$(dirname "$0")"
+for file in *.der
+do
+    header_file="${file%.der}.h"
+    xxd -i "$file" | sed 's/unsigned/static const unsigned/'> "$header_file"
```
andrewkozlik, line 7: Suggestion: Add a comment at the top of the header file saying that it was autogenerated by gen.sh.

matejcik: https://github.com/trezor/trezor-firmware/commit/3959a6fdff11fb409b1f3914e09c173ff17ca6f2

python/src/trezorlib/authentication.py
```diff
@@ -0,0 +1,306 @@
...
+        pubkey.verify(
+            self.fix_signature(self.cert.signature),
+            self.cert.tbs_certificate_bytes,
+            algo_params,
+        )
+
+    def verify_issued_by(self, issuer: "Certificate") -> None:
+        pubkey = issuer.cert.public_key()
+        assert isinstance(pubkey, ec.EllipticCurvePublicKey)
+        self.verify_by(pubkey)
```
andrewkozlik, line 125: I think we should be checking more in this function.

Most importantly:
- The issuer's certificate should have the cA boolean value in basic constraints set to true.

Secondly:
- The pathLenConstraint should be checked.
- The certificate issuer's name should match the subject name in the issuer's certificate.

The full process is described in https://datatracker.ietf.org/doc/html/rfc5280#section-6.1, but we don't need to be as thorough as that.

python/src/trezorlib/authentication.py
```diff
@@ -0,0 +1,306 @@
...
+        try:
+            cert.verify_issued_by(ca_cert)
+        except exceptions.InvalidSignature:
+            LOG.error(f"{cert_label} verification failed: %s", cert)
+            failed = True
+        else:
+            LOG.debug(f"{cert_label} verified successfully: %s", cert)
+        finally:
+            cert = ca_cert
+            cert_label = f"CA #{i - 1} certificate"
```
andrewkozlik, line 243: This seems to me like it should be `i`, not `i-1`. On line 223 we refer to the same certificate by index `i`, don't we?

python/src/trezorlib/authentication.py
matejcik: 5623f6d6f7f5a8a1e49b73959117d1ec499bc579

python/src/trezorlib/authentication.py
matejcik: indeed, this seems to be a leftover from some other version of this code
5623f6d6f7f5a8a1e49b73959117d1ec499bc579

matejcik:
added certificate whitelist check in https://github.com/trezor/trezor-firmware/pull/3944/commits/aab2eaf59c2d76e7022a51927902d14a10642a44

python/src/trezorlib/authentication.py
```diff
@@ -0,0 +1,345 @@
...
+        if issuer.cert.subject != self.cert.issuer:
+            LOG.error("Certificate %s is not issued by %s.", self, issuer)
+            return False
+
+        try:
+            # check issuer CA flag
+            ext = issuer.cert.extensions.get_extension_for_class(x509.BasicConstraints)
+            if ext.value.ca is not True:
+                LOG.error("Issuer is not a valid CA: %s", issuer)
+                return False
```
andrewkozlik, line 147: The error handling here seems a bit inconsistent. This function indicates errors in two ways: returning `False` or raising an exception, which is a little hard to keep track of. Is there any reason for that?

python/src/trezorlib/authentication.py
```diff
@@ -0,0 +1,345 @@
...
+            if ext.value.path_length is None:
+                LOG.error("Issuer is not a valid CA: %s", issuer)
+                return False
```
andrewkozlik, line 150: I think this is actually not an error. BTW, I am not sure right now what we have in our CAs.
> Where pathLenConstraint does not appear, no limit is imposed.

https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9

matejcik: i believe it is fine to enforce that certificates authenticating Trezors _must_ have a path limit set

this is our CA, for the record:
```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 7356626466875390824 (0x6617fccb67583b68)
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: C = CZ, O = Trezor Company s.r.o., CN = Trezor Manufacturing Root CA
        Validity
            Not Before: Apr 11 15:07:55 2024 GMT
            Not After : Apr  4 15:07:55 2054 GMT
        Subject: C = CZ, O = Trezor Company s.r.o., CN = Trezor Manufacturing CA
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:5b:78:5b:70:38:10:36:3d:ee:ad:c0:fa:a5:a7:
                    38:8d:83:85:f5:e4:f9:17:0a:3c:27:d3:d2:50:6b:
                    f9:09:0d:5b:09:04:4f:ba:d6:b9:06:44:6d:8c:db:
                    75:4b:65:86:9e:89:92:b2:6a:34:4e:01:ec:ff:75:
                    84:4f:e5:a3:d6
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Key Usage: critical
                Certificate Sign
            X509v3 Basic Constraints: critical
                CA:TRUE, pathlen:0
            X509v3 Authority Key Identifier: 
                1F:95:3C:3C:B7:93:8F:54:0E:E8:73:52:43:19:E6:06:E0:79:B6:FF
    Signature Algorithm: ecdsa-with-SHA256
    Signature Value:
        30:44:02:20:70:2f:dd:25:73:8a:45:a0:0b:a1:77:7c:6b:78:
        df:fa:be:3f:e5:3b:d3:0c:e8:26:c7:d3:91:c5:b7:0f:ea:83:
        02:20:07:52:71:71:8e:6e:02:dc:eb:9a:79:24:35:fe:16:de:
        5b:0f:08:19:09:d7:da:eb:8e:f7:d4:75:71:a3:d6:ca
```

python/src/trezorlib/authentication.py
matejcik: 27b8f2703d21f43ceb6adf1b920683e4bcf79af5
i originally intended to differentiate between "bad signature" and "bad issuer" and this seemed like the easiest way to accomplish -- but switching to just bool return value actually makes the caller code nicer so :+1:

python/src/trezorlib/authentication.py
matejcik: 2b33189bdb117d8c6684c69aabd86ca9d755b381

matejcik: ah, except this is not correct because it does not enforce _presence_ of the extensions

andrewkozlik (approved):

PR #3945, opened 2024-06-12
fix(python): increase homescreen quality when uploading via trezorctl

### LINKED ISSUE #3893: trezorctl drastically decrease quality of uploaded homescreen
Quality of jpgs uploaded through trezorctl is way worse then quality of jpgs uploaded to device through suite.

matejcik:
fixes #3893

### DIFF
--- a/python/.changelog.d/3893.added
+++ b/python/.changelog.d/3893.added
@@ -1,0 +1,1 @@
+Added `--quality` argument to `trezorctl set homescreen`.
--- a/python/.changelog.d/3893.changed
+++ b/python/.changelog.d/3893.changed
@@ -1,0 +1,1 @@
+Increased default JPEG quality for uploaded homescreen.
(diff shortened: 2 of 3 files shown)


matejcik:
@Hannsek adding you as a reviewer so that you can test it and see if it makes you happier :)

Hannsek:
Why do you introduce quality flag?
Cannot it be just set to max that would fit every time?

matejcik:
I also set the default quality to 90 which should be good for everyone.
"max quality that would fit" is not great because it would push the image quality towards the 16 kB limit every time, while in practice it's useful to have the homescreen as small as possible. (also we'd need to iterate the quality, perhaps resize multiple times before we figure out the max possible quality.

this way we set a reasonable default, while allowing tinkerers to (a) push for greater quality or (b) fit a more complex image within the limit

matejcik:
(before this, the default quality was 75 in the jpeg library)

matejcik:
superseded by #3963

PR #3946, opened 2024-06-13
USB driver refactoring II

cepetr:
This PR continues the refactoring of USB drivers and includes the following changes:

1. **Introduces the `usb_deinit()` routine and clarifies the highest-level API of the driver.** - required by entering/exiting low power modes, not used now.
2. Removes unnecessary calls to `usb_stop()` scattered throughout the codebase, as they always accompany `usb_deinit()`.
3. Removes the dependency of the class driver on the global context of the USB driver.
4. Fixes the build of the disc2 target.

These changes does not affect the device behavior.

### DIFF
--- a/core/embed/extmod/modtrezorio/modtrezorio-usb.h
+++ b/core/embed/extmod/modtrezorio/modtrezorio-usb.h
@@ -190,7 +190,9 @@ STATIC mp_obj_t mod_trezorio_USB_open(mp_obj_t self,
   mp_obj_get_array(MP_OBJ_FROM_PTR(&o->ifaces), &iface_cnt, &iface_objs);
 
   // Initialize the USB stack
-  usb_init(&o->info);
+  if (sectrue != usb_init(&o->info)) {
+    mp_raise_msg(&mp_type_RuntimeError, "failed to initialize usb driver");
+  }
 
   int vcp_iface_num = -1;
 
@@ -224,7 +226,10 @@ STATIC mp_obj_t mod_trezorio_USB_open(mp_obj_t self,
   }
 
   // Start the USB stack
-  usb_start();
+  if (sectrue != usb_start()) {
+    usb_deinit();
+    mp_raise_msg(&mp_type_RuntimeError, "failed to start usb driver");
+  }
   o->state = USB_OPENED;
 
   // If we found any VCP interfaces, use the last one for stdio,
--- a/core/embed/extmod/modtrezorio/modtrezorio-usb.h
+++ b/core/embed/extmod/modtrezorio/modtrezorio-usb.h
@@ -251,7 +251,6 @@ STATIC mp_obj_t mod_trezorio_USB_close(mp_obj_t self) {
   if (o->state != USB_OPENED) {
     mp_raise_msg(&mp_type_RuntimeError, "not initialized");
   }
-  usb_stop();
   usb_deinit();
   mp_obj_list_set_len(MP_OBJ_FROM_PTR(&o->ifaces), 0);
   mp_seq_clear(o->ifaces.items, 0, o->ifaces.alloc, sizeof(*o->ifaces.items));
@@ -271,7 +270,6 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_1(mod_trezorio_USB_close_obj,
 STATIC mp_obj_t mod_trezorio_USB___del__(mp_obj_t self) {
   mp_obj_USB_t *o = MP_OBJ_TO_PTR(self);
   if (o->state != USB_CLOSED) {
-    usb_stop();
     usb_deinit();
     o->state = USB_CLOSED;
   }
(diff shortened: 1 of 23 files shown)


core/embed/trezorhal/stm32f4/usb/usbd.c
```diff
@@ -149,6 +156,32 @@ void usb_init(const usb_dev_info_t *dev_info) {
...
+  usb_driver_t *usb = &g_usb_driver;
+
+  usb_stop();
+
+  // Set usb->dev_handle.dev_state to USBD_STATE_INITIALIZED
+  // Set usb->config_desc to NULL
+  memset(&usb, 0, sizeof(usb_driver_t));
+}
+
+void usb_start(void) {
```
TychoVrahe, line 171: should we return (sec)bool?

core/embed/trezorhal/stm32f4/usb/usbd.c
```diff
@@ -95,6 +100,8 @@ static secbool __wur check_desc_str(const char *s) {
 void usb_init(const usb_dev_info_t *dev_info) {
```
TychoVrahe, line 100: should we return (sec)bool? even though this seems to be infallible, it would be more consistent with the rest of the drivers

core/embed/trezorhal/stm32f4/usb/usbd.c
cepetr: see this change 5c362db1cd8169f39a83b593480bd5ab66e39a9f

1. `usb_start()` and `usb_stop()` are now fallible, returning a `secbool` - in my opinion, drivers like this should not use `ensure` for error handling.
2. The return value is checked in all places where these functions are called.
3. Renamed the `usb` driver variable to `drv` to be more consistent with other drivers (maybe we should agree whether `drv` or `driver` is better - currently, I use both, but I'd like to change it to `drv` everywhere).
4. Renamed some `usbd_` files to `usb_` - files with the `usbd_` prefix are part of the STM USB driver, while files with the `usb_` prefix are ours.

core/embed/trezorhal/stm32f4/usb/usbd.c
cepetr: see this change 5c362db1cd8169f39a83b593480bd5ab66e39a9f and the comment above

core/translations/signatures.json
```diff
@@ -1,8 +1,8 @@
 {
   "current": {
-    "merkle_root": "7fba6f82bbd54d09ca75a1454a0af359fd187dea5b689f4f19552b5ae1d8e2f0",
```
TychoVrahe, line 3: probably unwanted change

core/embed/extmod/modtrezorio/modtrezorio-usb.h
```diff
@@ -190,7 +190,9 @@ STATIC mp_obj_t mod_trezorio_USB_open(mp_obj_t self,
   mp_obj_get_array(MP_OBJ_FROM_PTR(&o->ifaces), &iface_cnt, &iface_objs);
 
   // Initialize the USB stack
-  usb_init(&o->info);
+  if (sectrue != usb_init(&o->info)) {
+    mp_raise_ValueError("failed to initialize usb driver");
```
TychoVrahe, line 194: `ValueError` doesn't fit well i think, how about
`mp_raise_msg(&mp_type_RuntimeError, "failed to initialize usb driver");`

core/embed/extmod/modtrezorio/modtrezorio-usb.h
```diff
@@ -224,7 +226,10 @@ STATIC mp_obj_t mod_trezorio_USB_open(mp_obj_t self,
   }
 
   // Start the USB stack
-  usb_start();
+  if (sectrue != usb_start()) {
+    usb_deinit();
+    mp_raise_ValueError("failed to start usb driver");
```
TychoVrahe, line 231: also here

core/embed/trezorhal/stm32f4/usb/usbd.c
TychoVrahe: I'd prefer `drv` too, meaning is clear and its shorter

core/translations/signatures.json
cepetr: yes, fixed f0032e5a3818a4cfffe7e5fd200c572106d882cc

core/embed/extmod/modtrezorio/modtrezorio-usb.h
cepetr: fixed, see f0032e5a3818a4cfffe7e5fd200c572106d882cc

core/embed/extmod/modtrezorio/modtrezorio-usb.h
cepetr: fixed, see f0032e5a3818a4cfffe7e5fd200c572106d882cc

TychoVrahe (approved):

PR #3949, opened 2024-06-14
Rework PIN processing

andrewkozlik:
This PR unifies the key-encryption-key (KEK) derivation process for devices with and without Optiga, as well as improving the Optiga PIN derivation process itself, making it all more modular. It is not a necessary change, but I think it will make the code cleaner in the long run, because the modularity makes it easy to add and remove steps to and from the process (PBKDF2 stretching, STM32U5 SAES, Optiga or other secure elements). Right now it complicates the code, because of all the storage upgrade stuff, but hopefully in time we will make a cut-off point for upgrades from old storage versions and get rid of all that.

Note that the process unification implies that a storage upgrade will be needed even for models without Optiga (Model T and Model One), even though it doesn't introduce any new feature to these devices. It will not be possible to downgrade to earlier firmware versions without wiping the seed.

Tests:

@trezor/qa when firmware with these changes is signed (version 2.8.0), please check that:
- upgrading TS3 from 2.7.0 to 2.8.0, and
- upgrading TS3 from 2.6.3 to 2.8.0 works smoothly.
Test with PIN configured and without PIN configured.

If you have the time, then also test the same for TT.


core/embed/trezorhal/optiga/optiga.c
```diff
@@ -376,86 +440,87 @@ static int optiga_pin_stretch_secret(OPTIGA_UI_PROGRESS ui_progress,
...
+  // OID_PIN_TOTAL_CTR.
 
-    hmac_sha256_Update(&ctx, buffer, size);
+  // Pseudocode for the stretching process:
+  // for i in range(PIN_STRETCH_ITERATIONS):
+  //   digest = HMAC(secret, "")
+  //   cmac_i = CMAC(optiga_cmac_key, digest)
+  //   hmac_i = HMAC(optiga_hmac_key, digest)
+  //   ecdh_i = ECDH(optiga_ecdh_key, digest)
+  //   secret = HMAC-SHA256(secret, cmac_i || hmac_i || ecdh_i)
```
onvej-sl, line 494: It should be `stretched_pin` instead of `secret`.

onvej-sl, line 494: What is the reason for indexing `cmac`, `hmac` and `ecdh` but not `digest` and `secret`.

I prefer not to use indices neither here nor in `optiga_pin_stretch_secret_v4`.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -363,8 +371,64 @@ static int optiga_pin_init_stretch(void) {
...
+  // Implements the functionality that is common to optiga_pin_stretch_secret()
+  // and the legacy function optiga_pin_stretch_secret_v4().
+
+  uint8_t buffer[ENCRYPT_SYM_PREFIX_SIZE + OPTIGA_PIN_SECRET_SIZE] = {0};
+  size_t size = 0;
+
+  // Combine intermediate result with OID_PIN_CMAC.
+  optiga_result res =
+      optiga_encrypt_sym(OPTIGA_SYM_MODE_CMAC, OID_PIN_CMAC, input,
+                         OPTIGA_PIN_SECRET_SIZE, buffer, sizeof(buffer), &size);
```
onvej-sl, line 386: I would either
  * check `size` here and in many other places or,
  * rewrite the `optiga_*` functions so they can accept `NULL`.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -634,6 +733,100 @@ int optiga_pin_verify(OPTIGA_UI_PROGRESS ui_progress,
...
+  // Reset the counter which limits the guesses at OID_STRETCHED_PIN.
+  res = optiga_set_data_object(OID_STRETCHED_PIN_CTR, false, COUNTER_RESET,
+                               sizeof(COUNTER_RESET));
+  optiga_clear_auto_state(OID_PIN_SECRET);
+  if (res != OPTIGA_SUCCESS) {
+    return res;
+  }
+
+  ui_progress(200);
+
```
onvej-sl, line 826: I suggest extracting this piece of of code into a new function.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -634,6 +733,100 @@ int optiga_pin_verify(OPTIGA_UI_PROGRESS ui_progress,
...
+  // HMAC the digest with the key in OID_PIN_HMAC.
+  uint8_t hmac_buffer[ENCRYPT_SYM_PREFIX_SIZE + OPTIGA_PIN_SECRET_SIZE] = {0};
+  size_t size = 0;
+  res = optiga_encrypt_sym(OPTIGA_SYM_MODE_HMAC_SHA256, OID_PIN_HMAC, digest,
+                           sizeof(digest), hmac_buffer, sizeof(hmac_buffer),
+                           &size);
+  if (res != OPTIGA_SUCCESS) {
+    memzero(digest, sizeof(digest));
+    uint8_t error_code = 0;
+    optiga_get_error_code(&error_code);
```
onvej-sl, line 758: I would check the return value of `optiga_get_error_code` here and in other 3 places.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -634,6 +733,100 @@ int optiga_pin_verify(OPTIGA_UI_PROGRESS ui_progress,
...
+    return error_code + OPTIGA_COMMAND_ERROR_OFFSET;
+  }
+
+  ui_progress(200);
+
+  // Reset the counter which limits the use of OID_PIN_HMAC.
+  res = optiga_set_data_object(OID_PIN_HMAC_CTR, false, COUNTER_RESET,
+                               sizeof(COUNTER_RESET));
+  if (res != OPTIGA_SUCCESS) {
+    return res;
```
onvej-sl, line 792: I would call `optiga_clear_auto_state(OID_STRETCHED_PIN)` here.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: d92f0efd52fe1d4fa82981c8a03ed20b114e0df9

andrewkozlik: d92f0efd52fe1d4fa82981c8a03ed20b114e0df9

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -634,6 +733,100 @@ int optiga_pin_verify(OPTIGA_UI_PROGRESS ui_progress,
...
+  }
+
+  // Read the counter-protected PIN secret from OID_PIN_SECRET.
+  uint8_t pin_secret[OPTIGA_PIN_SECRET_SIZE] = {0};
+  res = optiga_get_data_object(OID_PIN_SECRET, false, pin_secret,
+                               OPTIGA_PIN_SECRET_SIZE, &size);
+  if (res != OPTIGA_SUCCESS) {
+    optiga_clear_auto_state(OID_STRETCHED_PIN);
```
onvej-sl, line 800: I would move `optiga_clear_auto_state(OID_STRETCHED_PIN);` before the `if`.

onvej-sl, line 800: I would call `memzero(pin_secret)` here.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -634,6 +733,100 @@ int optiga_pin_verify(OPTIGA_UI_PROGRESS ui_progress,
...
+  hmac_sha256(stretched_pin, OPTIGA_PIN_SECRET_SIZE, NULL, 0, digest);
+
+  // HMAC the digest with the key in OID_PIN_HMAC.
+  uint8_t hmac_buffer[ENCRYPT_SYM_PREFIX_SIZE + OPTIGA_PIN_SECRET_SIZE] = {0};
+  size_t size = 0;
+  res = optiga_encrypt_sym(OPTIGA_SYM_MODE_HMAC_SHA256, OID_PIN_HMAC, digest,
+                           sizeof(digest), hmac_buffer, sizeof(hmac_buffer),
+                           &size);
+  if (res != OPTIGA_SUCCESS) {
+    memzero(digest, sizeof(digest));
```
onvej-sl, line 755: I would move `memzero(digest, sizeof(digest));` just before the `if`.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -465,108 +530,150 @@ int optiga_pin_set(OPTIGA_UI_PROGRESS ui_progress,
...
   if (res != OPTIGA_SUCCESS) {
-    optiga_clear_auto_state(OID_PIN_SECRET);
     return res;
   }
 
-  // Initialize the PIN counter.
-  res = optiga_set_data_object(OID_PIN_COUNTER, false, COUNTER_RESET,
+  // Initialize the counter which limits the guesses at OID_STRETCHED_PIN so
+  // that we can authorise using OID_STRETCHED_PIN.
+  res = optiga_set_data_object(OID_STRETCHED_PIN_CTR, false, COUNTER_RESET,
```
onvej-sl, line 606: Wouldn't it be enough to set the counter to one higher value instead of resetting it later again?

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -465,108 +530,150 @@ int optiga_pin_set(OPTIGA_UI_PROGRESS ui_progress,
...
-  // Authorise using OID_PIN_SECRET so that we can write to OID_PIN_COUNTER and
-  // OID_STRETCHED_PIN.
+  // Authorise using OID_PIN_SECRET so that we can write to OID_STRETCHED_PIN
+  // and OID_STRETCHED_PIN_CTR.
   res = optiga_set_auto_state(OPTIGA_OID_SESSION_CTX, OID_PIN_SECRET,
-                              out_secret, OPTIGA_PIN_SECRET_SIZE);
+                              pin_secret, sizeof(pin_secret));
   if (res != OPTIGA_SUCCESS) {
-    memzero(stretched_pin, sizeof(stretched_pin));
     return res;
```
onvej-sl, line 570: I would call `optiga_clear_auto_state(OID_PIN_SECRET)` here.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: b5d4a413d5a4890c59a6e75e6fecc7708fe9a988

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: https://github.com/trezor/trezor-firmware/commit/b5d4a413d5a4890c59a6e75e6fecc7708fe9a988

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -465,108 +530,150 @@ int optiga_pin_set(OPTIGA_UI_PROGRESS ui_progress,
...
-  // the user chooses a high-entropy PIN, then even if the Optiga and its
-  // communication link is completely compromised, it will not reduce the
-  // security of their device any more than if the Optiga was not integrated
-  // into the device in the first place.
+  // Stretch the PIN more with the counter-protected PIN secret. This method
+  // ensures that if the user chooses a high-entropy PIN, then even if the
+  // Optiga and its communication link is completely compromised, it will not
+  // reduce the security of their device any more than if the Optiga was not
+  // integrated into the device in the first place.
+  hmac_sha256(stretched_pin, OPTIGA_PIN_SECRET_SIZE, pin_secret,
```
onvej-sl, line 641: I would prefer to use `sizeof(stretched_pin)` instead of `OPTIGA_PIN_SECRET_SIZE` here and in other places.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: That would require a refactor, since we work with the number of depleted attempts, not the number of remaining attempts.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: ff74858d83e5adbfa585fa338560317433739651, but it makes no difference because the value is overwritten a few lines later anyway.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: If my knowledge of C isn't too rusty, then I'd say that `sizeof(stretched_pin)` is 4, whereas `OPTIGA_PIN_SECRET_SIZE` is 32.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: if `res != OPTIGA_SUCCESS` then `pin_secret` is not set. The `optiga_` functions modify the destination buffer only if the command succeeds.

storage/storage.c
```diff
@@ -838,6 +868,20 @@ static secbool decrypt_dek(const uint8_t *kek, const uint8_t *keiv) {
     return secfalse;
   }
 
+  uint8_t kek[SHA256_DIGEST_LENGTH] = {0};
+  uint8_t keiv[SHA256_DIGEST_LENGTH] = {0};
+  if (get_lock_version() >= 5) {
+    if (sectrue !=
+        derive_kek_unlock(pin, pin_len, storage_salt, ext_salt, kek)) {
```
onvej-sl, line 875: What is the reason for calling `derive_kek_unlock` from `decrypt_dek` instead of `unlock`? In my opinion, it reduces readability.

Is the reason to ensure that EDEK is in storage before verifying the PIN?

onvej-sl, line 875: What is the reason for calling derive_kek_unlock from decrypt_dek instead of unlock? In my opinion, it reduces readability.

andrewkozlik: It improves readability by making the function more similar to `set_pin()`. In fact, I meant to move the reading of the storage salt inside of the function too, but forgot to do it. Done in cd6c2c7079c30108b0592a9f36dddc1f9236dbef.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -465,108 +530,150 @@ int optiga_pin_set(OPTIGA_UI_PROGRESS ui_progress,
...
+              sizeof(pin_secret), stretched_pin);
+  memzero(pin_secret, sizeof(pin_secret));
+  if (res != OPTIGA_SUCCESS) {
+    return res;
+  }
```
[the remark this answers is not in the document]
onvej-sl: > Upon failure there is no need to clear them.

Yes, there is. You can have only four active authorizations. Clearing the authorizations helps prevent hard-to-find bugs in case we decide to add authorizations somewhere else in the future.

storage/storage.c
onvej-sl: So the only reason for you was to improve readability?

core/embed/trezorhal/optiga/optiga.c
onvej-sl: I'm not sure what you mean by "`&size` can't be `NULL`" or why it is relevant. What is relevant is that `size_t *output_size` can be `NULL`.

core/embed/trezorhal/optiga/optiga.c
```diff
@@ -634,6 +733,100 @@ int optiga_pin_verify(OPTIGA_UI_PROGRESS ui_progress,
...
+      return OPTIGA_ERR_COUNTER_EXCEEDED;
+    } else {
+      return error_code + OPTIGA_COMMAND_ERROR_OFFSET;
+    }
+  }
+
+  // Stretch the PIN with the result.
+  hmac_sha256(stretched_pin, OPTIGA_PIN_SECRET_SIZE, hmac_buffer, size,
+              stretched_pin);
+  memzero(hmac_buffer, sizeof(hmac_buffer));
```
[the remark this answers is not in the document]
andrewkozlik: 9d75caf0216f7bc3c08b168efeb3af50b432a7e8

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: Seems pointless, since as I mentioned IRL, I intend to create a separate workflow for changing the PIN that is less fragile than the present one and that will in all likelihood require splitting this code into functions in some way, which I am unable to specify without giving it more thought.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: Also seems fairly pointless right now, since it makes no difference to the behavior, and as I mentioned IRL the error processing will undergo a refactor in another PR. I will take it into account there.

core/embed/trezorhal/optiga/optiga.c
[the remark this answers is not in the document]
andrewkozlik: If this function fails, then we have a major problem, because it brings the device into an inconsistent state from which it will probably not be able to recover in most cases. Indeed, that is why the PIN setting process is wrapped in `ensure()`.

storage/storage.c
andrewkozlik: I think there was also something that had to do with cleaner error handling, but I am not sure exactly what it might have been.

core/embed/trezorhal/optiga/optiga.c
andrewkozlik: "`&size` can't be `NULL`", that's a fact. Thus checking that it is not `NULL` makes no sense. Furthermore, since "`&size` can't be `NULL`" and `&size` is passed here as the `output_size` argument to `optiga_encrypt_sym()`, it follows that `output_size` can't be `NULL` in this particular call.

core/embed/trezorhal/optiga/optiga.c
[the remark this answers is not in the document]
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/3949/commits/3a6dbab0d9b8f43482a2ce372794aecaab394fb1

core/embed/trezorhal/optiga/optiga.c
onvej-sl: We discussed it in person, but we did not reach a consensus.

core/embed/trezorhal/optiga/optiga.c
onvej-sl: If I understand correctly, this will be addressed in a future pull request.

core/embed/trezorhal/optiga/optiga.c
onvej-sl: Then, please cast the return value to `void` to indicate that this is intentional.

core/embed/trezorhal/optiga/optiga.c
onvej-sl: We discussed it in person, but we did not reach a consensus.

onvej-sl (approved):

bosomt:
QA Task: https://www.notion.so/satoshilabs/new-PIN-processing-f1b483bf283c44e29a3aeb0dab3bc965
will be retested with final signed firmware 🙏🏿

bosomt:
**QA OK**

    revision: 20 bytes 0xdd4671a5104952ef505d28d1f9e94d1484b4607a,

retested with both signed universal firmwares

PR #3951, opened 2024-06-14
Rust macros and types cleanup

matejcik:
building on top of #3938 for the `dbg_print!` macros

* use Rust's built-in `CStr` in favor of a stand-alone crate
* make use of `c""` string literals in favor of `cstr!` macro
* introduce `uformat!` convenience macro that will return (by default) a `ShortString` with the formatting result.
  * unwraps internally because otherwise we just write `unwrap!(uformat!())` everywhere, that's pure noise.
  * if you want to handle errors, create a String instance and invoke `uwrite!` on it manually
* unify all code to use a `ShortString` instead of picking a size ad hoc and carrying multiple copies of `String<N>`
  * this saves a meager ~1 kB of flash
* centralize `#[macro_use]` to exactly one place (the `macros` module in root). Import all other macros as needed (so things like `obj_type!` are explicitly imported from `micropython::macros`. Seems just plain nicer.)
* extend dbg_print!() functionality to places that don't have micropython, such as bootloader-emu (but note that you have to enable the `debug` feature)


obrusvit:
Feel free to include https://github.com/trezor/trezor-firmware/pull/3953

core/embed/rust/src/ui/model_mercury/component/share_words.rs
```diff
@@ -158,7 +157,7 @@ impl<'a> Component for ShareWords<'a> {
                     .text_font
                     .visible_text_height("1"),
             );
-        let ordinal = build_string!(3, inttostr!(ordinal_val), ".");
+        let ordinal = uformat!("{}.", ordinal_val);
```
obrusvit, line 160: image

obrusvit: Well, somehow the `.` is not rendered.

obrusvit:
Do you think we could achieve formatting translated template strings with `uformat!`?
Something like this would be desirable:
```rust
let text = TR::reset__select_word_x_of_y_template.map_translated(|s| {
    uformat!(s, n_choice, n_words)
});
```

`ufmt` complains with `error: expected string literal`

core/embed/rust/src/ui/model_tt/component/keyboard/slip39.rs
```diff
@@ -31,7 +31,7 @@ const MAX_LENGTH: usize = 8;
```
cepetr, line 31: `MAX_LENGTH` constant seems to be unused now.  The same applies for `MAX_LENGTH` in `bip39.rs`.

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
```diff
@@ -335,7 +337,7 @@ impl PinDots {
     }
 
     fn is_full(&self) -> bool {
-        self.digits.len() == self.digits.capacity()
+        self.digits.len() >= MAX_LENGTH
```
cepetr, line 340: What if `MAX_LENGTH`, defined at the beginning of the file is greater than the maximum length of `ShortString`? Should we check somewhere to ensure it isn't?

core/embed/rust/src/crypto/mod.rs
```diff
@@ -16,9 +18,9 @@ pub enum Error {
 impl From<Error> for crate::error::Error {
     fn from(e: Error) -> Self {
         match e {
-            Error::SignatureVerificationFailed => value_error!("Signature verification failed"),
-            Error::InvalidEncoding => value_error!("Invalid key or signature encoding"),
-            Error::InvalidParams => value_error!("Invalid cryptographic parameters"),
+            Error::SignatureVerificationFailed => value_error!(c"Signature verification failed"),
+            Error::InvalidEncoding => value_error!(c"Invalid key or signature encoding"),
+            Error::InvalidParams => value_error!(c"Invalid cryptographic parameters"),
```
cepetr, line 23: I wonder if it would be possible to avoid specifying the c-literal at every occurrence of `value_error!()`. Perhaps it could be incorporated into the macro itself, but I am not sure if it's possible at zero cost.

core/embed/rust/src/ui/model_mercury/component/share_words.rs
matejcik: i think that was before the fixup? i forgot to add the `.` in one place

core/embed/rust/src/crypto/mod.rs
matejcik: i'm pretty sure the macro system is not strong enough for it: `c"blabla"` is a single token, you can't take out (or add) the `c` at start.

Previously `value_error!` used the (3rd party) `cstr!` macro, which works by explicitly concatting a `\0` at end of the provided string, and then calling an unchecked conversion. It didn't seem worth keeping it, if the cost is "write a value_error!, see compile failure, remember to add `c` at start")

(I also looked into why we need a `CStr` here in the first place. the reason: if the incoming string is null-terminated, we don't have to copy it onto the micropython heap, we only need to allocate one block for the wrapping `str` object. This will pretty much always succeed if the GC itself is still working.)

matejcik:
> Do you think we could achieve formatting translated template strings with `uformat!`?
> Something like this would be desirable:

we can't because the format literals are parsed at compile-time.
we will need to find another library, or write our own

core/embed/rust/src/ui/model_mercury/component/share_words.rs
obrusvit: Yes, it should be green if the CI runs again.

core/embed/rust/src/ui/model_tt/component/keyboard/slip39.rs
matejcik: https://github.com/trezor/trezor-firmware/pull/3951/commits/e63cb30d53b566cbd2ddaf097bb34d6232bd5368

core/embed/rust/src/ui/model_tt/component/keyboard/pin.rs
matejcik: https://github.com/trezor/trezor-firmware/pull/3951/commits/e63cb30d53b566cbd2ddaf097bb34d6232bd5368

matejcik:
> Feel free to include #3953

very nice, cherry-picked as https://github.com/trezor/trezor-firmware/commit/d0d67805ffc0a2c91d482ccef58963cb4b4566e3

core/embed/rust/src/crypto/mod.rs
cepetr: yes, I was afraid it wouldn't be possible. I don't think it's a problem to solve.

mmilata (approved):
nice cleanup

PR #3953, opened 2024-06-14
refactor(core): remove StringType

obrusvit:
Removing the obsolete crutch `StringType` and replacing it with `TString`.


obrusvit:
Cherry-picked there: https://github.com/trezor/trezor-firmware/pull/3951

PR #3965, opened 2024-06-19
Swipe tuning

TychoVrahe:
This PR addresses several problems/UX issues with swiping on T3T1.

- when changing layout (with going through python), swipe direction is remembered and therefore the second part of the animation has right effect
- no longer blocking swipes when button is pressed, rather we selectively block swipes on hold to confirm press only. When the swipe is locked-in while, new TouchAbort event is sent so any pressed buttons are deactivated.
- swipes beginning too close to an edge (that in direction of the swipe) are ignored, as they would fail anyway
- handling of repeated touches when swiping is improved
- instant lock-trigger of swipe without TouchMove event in between Start and End is removed- this would cause some complications with the new TouchAbort mechanism and these situations are not real swipes anyway..
- hold to confirm screen when showing share words is animated
- internal swipes handling is improved - when entering/exiting screen with internal swipes, normal swipe animation is played. This mainly affects show share words flow.
- duration of canceled swipe animation is fixed

Also in this PR, a fix of two other bugs:
- touch driver sometimes issued second TouchStart/End event pair, causing double button clicks for example
- backlight handling on first paint is fixed to address now quicker timers when playing attach animation

### DIFF
--- a/core/embed/rust/src/ui/event.rs
+++ b/core/embed/rust/src/ui/event.rs
@@ -51,6 +51,8 @@ pub enum TouchEvent {
     TouchMove(Point),
     /// Touch has ended at a point on the screen.
     TouchEnd(Point),
+    /// Touch event has been suppressed by more important event - i.e. Swipe.
+    TouchAbort,
 }
 
 impl TouchEvent {
--- a/core/embed/rust/src/ui/layout/obj.rs
+++ b/core/embed/rust/src/ui/layout/obj.rs
@@ -2,6 +2,7 @@ use core::{
     cell::RefCell,
     convert::{TryFrom, TryInto},
 };
+use num_traits::{FromPrimitive, ToPrimitive};
 
 use crate::{
     error::Error,
@@ -9,7 +10,7 @@ use crate::{
     micropython::{
         buffer::StrBuffer,
         gc::Gc,
-        macros::{obj_dict, obj_fn_1, obj_fn_2, obj_fn_var, obj_map, obj_type},
+        macros::{obj_dict, obj_fn_1, obj_fn_2, obj_fn_3, obj_fn_var, obj_map, obj_type},
         map::Map,
         obj::{Obj, ObjBase},
         qstr::Qstr,
@@ -32,9 +33,33 @@ use crate::ui::{display::Color, shape::render_on_display};
 
 #[cfg(feature = "button")]
 use crate::ui::event::ButtonEvent;
-#[cfg(feature = "touch")]
-use crate::ui::event::TouchEvent;
 use crate::ui::event::USBEvent;
+#[cfg(feature = "touch")]
+use crate::ui::{component::SwipeDirection, event::TouchEvent};
+
+impl AttachType {
+    fn to_obj(self) -> Obj {
+        match self {
+            Self::Initial => Obj::const_none(),
+            #[cfg(feature = "touch")]
+            Self::Swipe(dir) => dir.to_u8().into(),
+        }
+    }
+    fn try_from_obj(obj: Obj) -> Result<Self, Error> {
+        if obj == Obj::const_none() {
+            return Ok(Self::Initial);
+        }
+        #[cfg(feature = "touch")]
+        {
+            let dir: u8 = obj.try_into()?;
+            return Ok(AttachType::Swipe(
+                SwipeDirection::from_u8(dir).ok_or(Error::TypeError)?,
+            ));
+        }
+        #[allow(unreachable_code)]
+        Err(Error::TypeError)
+    }
+}
 
 /// Conversion trait implemented by components that know how to convert their
 /// message values into MicroPython `Obj`s.
@@ -57,6 +82,7 @@ pub trait ObjComponent: MaybeTrace {
     fn obj_skip_paint(&mut self) {}
     fn obj_request_clear(&mut self) {}
     fn obj_delete(&mut self) {}
+    fn obj_get_transition_out(&self) -> Result<Obj, Error>;
 }
 
 impl<T> ObjComponent for Root<T>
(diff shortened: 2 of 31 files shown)


TychoVrahe:
@cepetr please see https://github.com/trezor/trezor-firmware/pull/3965/commits/967b81c347b5d6d8672c939b0a7b3e067d745dcb (we already discussed this)

@matejcik please review https://github.com/trezor/trezor-firmware/pull/3965/commits/48966be43adeb502758a086ed36831ced23b1585

cepetr (commented):
I looked just at 967b81c347b5d6d8672c939b0a7b3e067d745dcb and utACK.

core/src/trezor/ui/layouts/common.py
```diff
@@ -1,10 +1,13 @@
 from typing import TYPE_CHECKING
 
+import trezorui2
 from trezor import log, workflow
 from trezor.enums import ButtonRequestType
 from trezor.messages import ButtonAck, ButtonRequest
 from trezor.wire import context
 
+LAST_TRANSITION_OUT: trezorui2.AttachType | None = None
```
matejcik, line 9: move this to `ui/__init__.py`
do not implement the setter/getter, they are just a waste of flash+RAM, have the functions assign to the global directly
(you can't `from ui import LAST_TRANSITION_OUT` though, you always need to do `ui.LAST_TRANSITION_OUT = ...`)

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -6,7 +6,12 @@
 from trezor.messages import ButtonAck, ButtonRequest
 from trezor.wire import ActionCancelled, context
 
-from ..common import button_request, interact
+from ..common import (
+    button_request,
+    get_last_transition_out,
+    interact,
+    set_last_transition_out,
+)
```
matejcik, line 14: no need for those here and in `tr`

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -1315,12 +1311,15 @@ pub static mp_module_trezorui2: Module = obj_module! {
     ///
     /// T = TypeVar("T")
     ///
+    /// class AttachType:
+    ///     pass
```
matejcik, line 1315: nit:
```suggestion
    /// class AttachType:
    ///     ...
```

core/embed/rust/src/ui/event.rs
```diff
@@ -51,6 +51,8 @@ pub enum TouchEvent {
     TouchMove(Point),
     /// Touch has ended at a point on the screen.
     TouchEnd(Point),
+    /// Touch has been aborted.
+    TouchAbort(),
```
matejcik, line 55:
```suggestion
    TouchAbort,
```

core/embed/rust/src/ui/event.rs
```diff
@@ -51,6 +51,8 @@ pub enum TouchEvent {
     TouchMove(Point),
     /// Touch has ended at a point on the screen.
     TouchEnd(Point),
+    /// Touch has been aborted.
```
matejcik, line 54: i don't understand what this does

core/embed/rust/src/ui/event.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/52193a77188a1f7f3905ce948311f70085bf4578

core/embed/rust/src/ui/event.rs
TychoVrahe: tried to make it clearer https://github.com/trezor/trezor-firmware/commit/52193a77188a1f7f3905ce948311f70085bf4578

core/embed/rust/src/ui/model_mercury/layout.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/07bd5dda9eb2378e293a6e5ed216beb3c2429bbc

core/src/trezor/ui/layouts/tt/__init__.py
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/07bd5dda9eb2378e293a6e5ed216beb3c2429bbc

core/src/trezor/ui/layouts/common.py
TychoVrahe: moved https://github.com/trezor/trezor-firmware/commit/07bd5dda9eb2378e293a6e5ed216beb3c2429bbc

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -32,9 +32,42 @@ use crate::ui::{display::Color, shape::render_on_display};
...
+            #[cfg(feature = "touch")]
+            2 => AttachType::Swipe(SwipeDirection::Down),
+            #[cfg(feature = "touch")]
+            3 => AttachType::Swipe(SwipeDirection::Left),
+            #[cfg(feature = "touch")]
+            4 => AttachType::Swipe(SwipeDirection::Right),
+            _ => AttachType::Initial,
+        }
+    }
+}
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/07bd5dda9eb2378e293a6e5ed216beb3c2429bbc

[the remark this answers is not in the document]
TychoVrahe: clippy complained about the reference so https://github.com/trezor/trezor-firmware/commit/b630913483856200bfb8e87be5d1ec79d3765294

matejcik (approved):
code ACK but we're out of flash again :(

bosomt:
**QA OK**

 - Device: Trezor T3T1 2.8.0 regular (revision dd4671a5104952ef505d28d1f9e94d1484b4607a)

PR #3967, opened 2024-06-21
IRQ config cleanup and simplification

cepetr:
This PR partially prepares IRQ handling for the low power mode feature. It's mostly cleanup with a few new features:

1. Removes the unnecessary PendSV handler (now pressing Ctrl+C in the MicroPython VCP console has no effect at all).
2. Removes unnecessary synchronization between the USB stack and SDCARD driver (probably copy-pasted stuff).
3. Decreases the number of exception priorities to 3.
4. ~~SVC calls are now interruptible - we can do a lot of stuff inside (initialize/deinitialize drivers, call `hal_delay()`, etc.).~~
5. Fixes SysTick IRQ priority on STM32U5 (changed from lowest to highest). This could potentially cause problems during reading/writing from/to the SD card if the card/controller wasn't behaving correctly.
6. Fixes the procedure for setting IRQ priority. Several interrupt priorities were not set correctly, although with no harmful effect. From now we use `NVIC_xxx()` instead of `HAL_NVIC_xxx()` functions.

After this change, there are just 3 interrupt priorities:

- `IRQ_PRI_HIGHEST` `(0)` reserved for all faults except HardFault, TAMPER interrupt, SysTick and SVCall
- `IRQ_PRI_NORMAL` `(8)` used by all hardware drivers.
- `IRQ_PRI_LOWER` `(15)` reserved for PendSV.

### DIFF
--- a/core/embed/trezorhal/stm32f4/supervise.c
+++ b/core/embed/trezorhal/stm32f4/supervise.c
@@ -5,16 +5,26 @@
 #include "../mpu.h"
 #include "common.h"
 #include "display.h"
+#include "irq.h"
 #include "supervise.h"
 
 #ifdef ARM_USER_MODE
+
+void svc_init(void) {
+  NVIC_SetPriority(SVCall_IRQn, IRQ_PRI_HIGHEST);
+
+  // We need to ensure that SysTick has the expected priority.
+  // The SysTick priority is configured in the boardloader,
+  // and some early versions didn't set this properly.
+  NVIC_SetPriority(SysTick_IRQn, IRQ_PRI_HIGHEST);
+}
 
 #ifdef STM32U5
 extern uint32_t g_boot_command;
 __attribute__((noreturn)) static void _reboot_to_bootloader(
     boot_command_t boot_command) {
   g_boot_command = boot_command;
-  __disable_irq();
+  disable_irq();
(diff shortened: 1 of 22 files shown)


core/embed/trezorhal/stm32f4/supervise.c
```diff
@@ -5,16 +5,26 @@
...
 
 #ifdef ARM_USER_MODE
 
+void svc_init(void) {
+  NVIC_SetPriority(SVCall_IRQn, IRQ_PRI_LOWEST);
+
+  // We need to ensure that SysTick has the expected priority.
+  // The SysTick priority is configured in the boardloader,
+  // and some early versions didn't set this properly.
+  NVIC_SetPriority(SysTick_IRQn, IRQ_PRI_HIGHEST);
```
TychoVrahe, line 19: this may not belong here - we might also want to set this for bootloader to avoid potentially different behavior between fw and bootloader - as this affects more than SVC. how about introducing irq_init() and putting that into bootloader too?

TychoVrahe: Discussed elsewhere: lets keep this for timer implementation.

TychoVrahe (approved):

PR #3968, opened 2024-06-21
Fix memory clearing in boardloader

cepetr:
This PR fixed memory clearing during the transition from the boardloader to the bootloader.

Prior to this change, the memory was filled with potentially uninitialized values. After this change, the memory is correctly set to zero as expected. This bug did not pose any security issues since the memory was always overwritten.

Additionally, the unused `jump_to_unprivileged` function has been removed in this PR.


TychoVrahe (approved):
looks good

PR #3970, opened 2024-06-21
Fix touch driver on dev kits

cepetr:
This PR fixes the bugs in the touch driver introduced by #3900 on both discovery kits.

1) Fixes duplicated `TOUCH_MOVE` events with the same coordinates
2) Fixes `touch_activity()` detectioin on disc2 kit
3) Fixes `flash_firmware`/`flash_bootloader` make targets for disc1 kit


TychoVrahe (approved):
utACK

PR #3974, opened 2024-06-24
fix T2B1 compilation with new rendering

TychoVrahe:
fixes compilation of T2B1 / model_tr UI with new rendering.

fixes also some conditionals based on xframebuffer feature.

### DIFF
--- a/core/embed/rust/src/ui/shape/render.rs
+++ b/core/embed/rust/src/ui/shape/render.rs
@@ -1,11 +1,9 @@
 use crate::ui::{
     display::Color,
-    geometry::{Offset, Point, Rect},
+    geometry::{Offset, Rect},
 };
 
-use super::{BasicCanvas, Canvas, DrawingCache, Rgb565Canvas, Shape, ShapeClone, Viewport};
-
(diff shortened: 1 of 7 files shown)


core/embed/rust/src/ui/shape/render.rs
```diff
@@ -118,12 +123,14 @@ where
 // struct ProgressiveRenderer
 // ==========================================================================
 
+// #[cfg(not(feature = "xframebuffer"))]
```
cepetr, line 126: I would recommend placing `ProgressiveRenderer` in its own file, rather than decorating each detail in the current file with `#[cfg...]`.

cepetr (changes requested):

core/embed/rust/src/ui/shape/render.rs
TychoVrahe: moved https://github.com/trezor/trezor-firmware/pull/3974/commits/4fb6e8f9b552f3fbff6e384ec8df33edd4e49793

cepetr (approved):

PR #3975, opened 2024-06-25
Increase Optiga MAX_RETRY_READ_MS to 6 sec

andrewkozlik:
Changes Optiga's `MAX_RETRY_READ_MS` introduced in https://github.com/trezor/trezor-firmware/pull/3766. I increase it to 6 seconds in order to account for the maximum throttling down delay [1] of 5 seconds when Optiga's security event counter is high.

[1] https://github.com/Infineon/optiga-trust-m-overview/blob/main/docs/OPTIGA%E2%84%A2%20Trust%20M%20Solution%20Reference%20Manual.md#security-monitor-characteristics.

### DIFF
--- a/core/.changelog.d/+optiga.fixed
+++ b/core/.changelog.d/+optiga.fixed
@@ -1,0 +1,1 @@
+Fix Optiga-related RSODs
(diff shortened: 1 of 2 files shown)


TychoVrahe (approved):
utACK. 

From users perspective, the changelog message is quite useless - what he probably would be interested in is that this aims to fix the spurious optiga/pin related RSODs

PR #3989, opened 2024-07-02
chore(core): remove some unused assets and code

obrusvit:
Removing model_t leftovers from mercury code.

This also changes the icon used while locking the screen. The previous one was not even in assets for TS5. Cc: @Hannsek


mmilata (approved):
UI diff ack
imo `ICON_KEY` is correct since it is displayed on the lockscreen

Hannsek:
Yeah, the old one was horrible. 

I asked Luke to come up with the new lock icon, but the key is fine for now.

Hannsek:
Btw, how much space we've saved by this?

obrusvit:
@Hannsek We saved roughly "šul-nul"

PR #3995, opened 2024-07-03
add support for 32bit colors in UI

TychoVrahe:
This PR adds support for 32 bit colors in UI. This will enable doing cool stuff like gradients (on displays that support it, currently only available on DISC2).

Some minor bugs are fixed in the underlying C code of the drawing lib.


core/embed/lib/gfx_color.h
```diff
@@ -22,8 +22,11 @@
 
 #include <stdint.h>
 
+#ifdef UI_COLORS32BIT
```
cepetr, line 25: I would prefer to maintain consistency with the existing name `GFX_COLOR_32BIT` and suggest using either `UI_COLOR32BIT` (or `UI_COLOR_32BIT`) and the Rust feature `ui_color32bit` (or `ui_color_32bit`).

core/embed/rust/src/ui/display/color.rs
```diff
@@ -1,12 +1,31 @@
...
 impl Color {
+    #[cfg(not(feature = "ui_colors32bit"))]
     pub const fn from_u16(val: u16) -> Self {
         Self(val)
     }
+
+    #[cfg(feature = "ui_colors32bit")]
+    pub const fn from_u16(val: u16) -> Self {
+        Self {
+            r: ((val >> 8) & 0xF8) as u8,
```
cepetr, line 23: The code works, but it lacks precision. The value 0xFFFF is converted to 0xF8FCF800. The lower bits should replicate the highest bits value. Please look at `gfx_color16_to_r/g/b()` functions in `gfx_color.h`.

cepetr (changes requested):

core/embed/lib/gfx_color.h
TychoVrahe: renamed: https://github.com/trezor/trezor-firmware/commit/92df5bf66d3e6964ec166f0b3e8be1a7643f6c33

core/embed/rust/src/ui/display/color.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/92df5bf66d3e6964ec166f0b3e8be1a7643f6c33

PR #4008, opened 2024-07-09
Consistent error handling

cepetr:
This PR proposes a consistent approach to error handling in the C codebase. The proposal includes the following:

1. Introduction of system-wide error codes (`ts_t` type) with predefined errors prefixed with `TS_` (Trezor status) and built-in fault injection hardening.
2. A straightforward mechanism/framework for error handling within functions, utilizing `goto` statements to facilitate cleanup at the end of the function.
3. Addition of new `ensure()` macro types to simplify top-level code.
4. Simplification of STM32 HAL error handling with the `hal_status_to_ts()` mapping function and the `TS_CHECK_HAL_OK()` macro.

This PR consists of three commits. Two of them serve as examples: a refactored haptic driver and a refactored SD card driver.

Error handling pattern:

```c
ts_t module_function(int arg1, int arg2) {
  // Initialize the `__status` variable and set it to `TS_OK`.
  // This should be on the first line inside the function (rule 1).
  TS_INIT; 

  // Check arguments to ensure they are greater than 0.
  TS_CHECK_ARG(arg1 > 0);
  TS_CHECK_ARG(arg2 > 0);
  
  // Check the result of another function with a `ts_t` return value.
  // The `status` value is properly propagated.
  // Splitting it into two lines makes debugging easier (rule 2).
  ts_t status = another_function();
  TS_CHECK_OK(status);
  
  // Check the result of an STM32 HAL function returning `HAL_StatusTypeDef`.
  // The HAL_xxx error is properly mapped to TS_xxx and propagated.
  HAL_StatusTypeDef hal_status = HAL_function();
  TS_CHECK_HAL_OK(hal_status);
  
  // Check a generic condition inside the function and propagate the status code
  // if the condition is not fulfilled.
  TS_CHECK(arg1 * arg2 < 1000, TS_ERROR);

  // Alternatively, use the secure variant working with `secbool`.
  secbool ready = usb_ready();
  TS_CHECK_SEC(ready);
  
  // In some cases it's useful to return before the cleanup code, (rule 3: never use `return ...` directly)
  // TS_RETURN;
  
cleanup:
  // Potential cleanup code goes here.
  // ...

  // Return the content of the `__status` variable.
  // In case of no error, it's `TS_OK`.
  TS_RETURN;
}
``` 
We can now use the `ensure_xxx()` set of macros similar to `verify_xxx()` for handling fatal errors. The `ensure_xxx()` macros display an error message on the screen and shut down the device as before. (Note: This feature is not yet complete in this PR.)

```c
  // The generic condition must be `true`
  //  TS_CHECK(condition, ts_t) ->  
  ensure(condition, message);

  // Status must be `TS_OK`
  // TS_CHECK_OK(status) -> 
  ensure_ok(status, message);

  // The generic condition (secbool) must be `sectrue`
  // TS_CHECK_SEC(secbool, status) -> 
  ensure_sec(secbool, message);
 ```
I understand that this is a completely new approach, which is not very common and requires some time to get used to. However, it is very simple and quickly understandable. I will not merge this until we reach a consensus that this is the right approach. Please feel free to provide any comments or suggestions.

### DIFF
--- a/core/embed/lib/error_handling.h
+++ b/core/embed/lib/error_handling.h
@@ -20,6 +20,188 @@
 #ifndef LIB_ERROR_HANDLING_H
 #define LIB_ERROR_HANDLING_H
 
+#include <stdint.h>
+
+// Suppresses the intellisense error in VSCode
+#ifndef __FILE_NAME__
+#define __FILE_NAME__ __FILE__
+#endif
+
+// Status code type
+typedef struct {
+  // Do not access this field directly,
+  // use `ts_ok()` and `ts_error()` macros.
+  uint32_t code;
+} ts_t;
+
+// Build status code from any 16-bit value.
+//
+// Status codes are hardened against fault injections
+// by storing the same value in the upper 16 bits.
+#define TS_BUILD(code) ((const ts_t){(code) | ((code) << 16)})
+
+// OK status code (signalling success or no error)
+#define TS_OK TS_BUILD(0)
+
+// This offset ensures at lest 16bit hamming distance
+// between `TS_OK` and other status codes.
+#define TS_ERROR_OFFSET 0xFF00
+
+// Build error status code from any 8-bit value
+//
+// The code should be in range 0 to 255 to ensure
+// 16-bit hamming distance to `TS_OK`
+#define TS_ERROR_BUILD(code) TS_BUILD(TS_ERROR_OFFSET + (code))
+
+// Error status codes
+#define TS_ERROR TS_ERROR_BUILD(0)
+#define TS_ERROR_BUSY TS_ERROR_BUILD(1)
+#define TS_ERROR_TIMEOUT TS_ERROR_BUILD(2)
+#define TS_ERROR_NOTINIT TS_ERROR_BUILD(3)
+#define TS_ERROR_ARG TS_ERROR_BUILD(4)
+#define TS_ERROR_IO TS_ERROR_BUILD(5)
+
+// Extracts the status code integer value.
+#define ts_code(status) ((status).code)
+
+// Converts status code to 32-bit unsigned integer.
+#define ts_to_u32(status) ((status).code);
+
+// Converts 32-bit unsigned integer to status code.
+#define ts_from_u32(u32) ((const ts_t){u32})
+
+// Check status code consistency and returns its value.
+// If invalid status code is detected, it will call `__fatal_error()`.
+#define _ts_checked(status)                                          \
+  ({                                                                 \
+    ts_t _checked = (status);                                        \
+    if ((ts_code(_checked) & 0xFFFF) != (ts_code(_checked) >> 16)) { \
+      __fatal_error("ts check error", __FILE_NAME__, __LINE__);      \
+    }                                                                \
+    _checked;                                                        \
+  })
+
+// Returns `true` if status code is `TS_OK`
+#define ts_ok(status) (ts_code(_ts_checked(status)) == ts_code(TS_OK))
+
+// Returns `true` if status code is NOT `TS_OK`
+#define ts_error(status) (ts_code(_ts_checked(status)) != ts_code(TS_OK))
+
+// Returns `true` if both status codes are equal
+#define ts_eq(status1, status2) (ts_code(status1) == ts_code(status2))
+
+// Returns a string representation of the status code.
+//
+// TS_OK -> "OK"
+// TS_ERROR -> "ERROR"
+// ...
+const char *ts_string(ts_t status);
+
+// ----------------------------------------------------
+// TS_INIT, TS_RETURN and VERIFY_XXX() macros define
+// a simple error handling mechanism
+//
+// Example:
+//
+// ts_t my_function(int arg) {
+//   // initialize verify mechanism
+//   TS_INIT;
+//
+//   // check arguments
+//   TS_CHECK_ARG(arg > 0);
+//
+//   ts_t status;
+//
+//   // verify success
+//   status = some_function();
+//   TS_CHECK_OK(status);
+//
+//   // verify condition
+//   TS_CHECK(another_function() != 0, TS_ERROR_IO);
+//
+//  cleanup:
+//
+//   // clean up code comes here
+//
+//   TS_RETURN;
+// }
+
+// Declares a status variable and initializes it to `TS_OK`.
+// This variable is used to store the status
+#define TS_INIT __attribute__((unused)) ts_t __status = TS_OK;
+
+// Returns the current status.
+#define TS_RETURN return __status;
+
+// Jumps to `error` label if status is not `TS_OK`.
+#define TS_CHECK_OK(status)  \
+  do {                       \
+    ts_t _status = status;   \
+    if (ts_error(_status)) { \
+      __status = _status;    \
+      goto cleanup;          \
+    }                        \
+  } while (0)
+
+// Jumps to `error` label if the condition is not `true`.
+#define TS_CHECK(cond, status) \
+  do {                         \
+    if (!(cond)) {             \
+      __status = status;       \
+      goto cleanup;            \
+    }                          \
+  } while (0)
+
+// Jumps to `error` label if the condition is not `true`.
+// Sets the status to `TS_ERROR_ARG`.
+#define TS_CHECK_ARG(cond)     \
+  do {                         \
+    if (!(cond)) {             \
+      __status = TS_ERROR_ARG; \
+      goto cleanup;            \
+    }                          \
+  } while (0)
+
+// Jumps to `error` label if the condition is not `sectrue`.
+#define TS_CHECK_SEC(seccond, status) \
+  do {                                \
+    if ((seccond) != sectrue) {       \
+      __status = status;              \
+      goto cleanup;                   \
+    }                                 \
+  } while (0)
+
+// Do not use this function directly, use the `ensure()` macro instead.
+void __attribute__((noreturn))
+__fatal_error(const char *msg, const char *file, int line);
+
+// Ensures that status code is `TS_OK`.
+// If not, it shows an error message and shuts down the device.
+#define ensure_ok(status, msg)                     \
+  do {                                             \
+    if (!ts_ok(status)) {                          \
+      __fatal_error(msg, __FILE_NAME__, __LINE__); \
+    }                                              \
+  } while (0)
+
+// Ensures that condition is evaluated as `true`.
+// If not, it shows an error message and shuts down the device.
+#define ensure_true(cond, msg)                     \
+  do {                                             \
+    if (!(cond)) {                                 \
+      __fatal_error(msg, __FILE_NAME__, __LINE__); \
+    }                                              \
+  } while (0)
+
+// Ensures that condition is evaluated as `sectrue`.
+// If not, it shows an error message and shuts down the device.
+#define ensure(seccond, msg)                       \
+  do {                                             \
+    if ((seccond) != sectrue) {                    \
+      __fatal_error(msg, __FILE_NAME__, __LINE__); \
+    }                                              \
+  } while (0)
+
 // Shows an error message and shuts down the device.
 //
 // If the title is NULL, it will be set to "INTERNAL ERROR".
@@ -32,15 +214,6 @@ error_shutdown_ex(const char *title, const char *message, const char *footer);
 //
 // Same as `error_shutdown_ex()` but with a default header and footer.
 void __attribute__((noreturn)) error_shutdown(const char *message);
-
-// Do not use this function directly, use the `ensure()` macro instead.
-void __attribute__((noreturn))
-__fatal_error(const char *msg, const char *file, int line);
-
-// Checks for an expression and if it is false, shows an error message
-// and shuts down the device.
-#define ensure(expr, msg) \
-  (((expr) == sectrue) ? (void)0 : __fatal_error(msg, __FILE_NAME__, __LINE__))
 
 // Shows WIPE CODE ENTERED screeen and shuts down the device.
 void __attribute__((noreturn)) show_wipe_code_screen(void);
--- a/core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
+++ b/core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
@@ -82,7 +82,7 @@ static haptic_driver_t g_haptic_driver = {
     .initialized = false,
 };
 
-static bool drv2625_set_reg(i2c_bus_t *bus, uint8_t addr, uint8_t value) {
+static ts_t drv2625_set_reg(i2c_bus_t *bus, uint8_t addr, uint8_t value) {
   i2c_op_t ops[] = {
       {
           .flags = I2C_FLAG_TX | I2C_FLAG_EMBED,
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Comment by hiviah** [comment]
The weird definition and usage of `ensure` with multiplication which was made against being removed by compiler and generated better barrier assembler code. Hamming weight of `sectrue` and `secfalse` was made to be high, while Hamming distance between `TS_OK` and `TS_ERROR` is 2 bits. Not entirely sure if highest bit should not be masked to 0 in case it passes through ints and uints (e.g. as a part of some call or other macro from other library).

It might be usable to have macro shorthand for `__attribute__((cleanup(func)))` which could help with cleanup no matter how the function ends (limited in usability with arrays, but I think for already defined "deinit" functions for HW it might work). This way a cleanup for a function can be used in a sub-block of a function.
In this tiny library they seem to make it work also with arrays - https://github.com/Snaipe/libcsptr (kind of C replacement for `unique_ptr` and `shared_ptr`)

To me `goto` inside macro seems very...unexpected by someone who did not read this particular PR. And maybe better name for `error:` is `cleanup:` since code branches without error go through `error:` part without return first.

File: core/embed/lib/error_handling.h [hunk 1]
```diff
@@ -20,6 +20,151 @@
...
+// Returns the current status.
+#define verify_status() (__status)
+
+// Jumps to `error` label if status is not `TS_OK`.
+#define verify_ok(status)    \
+  do {                       \
+    ts_t _status = status;   \
+    if (ts_error(_status)) { \
+      __status = _status;    \
+      goto error;            \
```
andrewkozlik, line 105: I don't like the idea of using macros that perform jumps or return. It looks like a function, but has an unexpected side effect for someone who is not intimately familiar with it.

core/embed/lib/error_handling.h
```diff
@@ -20,6 +20,151 @@
...
 #define LIB_ERROR_HANDLING_H
 
+// Suppresses the intellisense error in VSCode
+#ifndef __FILE_NAME__
+#define __FILE_NAME__ __FILE__
+#endif
+
+// Build status code from any 16-bit value.
+// This macro is used for generic status codes.
+#define TS_BUILD(code) ((code) | (~(code) << 16))
```
andrewkozlik, line 30: It would be nice to use something that ensures a large minimum Hamming distance. This has 2. However, we don't have to worry about that at this moment.

cepetr:
> The weird definition and usage of `ensure` with multiplication which was made against being removed by compiler and generated better barrier assembler code. Hamming weight of `sectrue` and `secfalse` was made to be high, while Hamming distance between `TS_OK` and `TS_ERROR` is 2 bits. Not entirely sure if highest bit should not be masked to 0 in case it passes through ints and uints (e.g. as a part of some call or other macro from other library).

Yes, good point. Hamming distance between `TS_OK` and `TS_ERROR` is low. There are certainly several ways to improve it. I'll think about it and try to suggest something. 

> It might be usable to have macro shorthand for `__attribute__((cleanup(func)))` which could help with cleanup no matter how the function ends (limited in usability with arrays, but I think for already defined "deinit" functions for HW it might work). This way a cleanup for a function can be used in a sub-block of a function. In this tiny library they seem to make it work also with arrays - https://github.com/Snaipe/libcsptr (kind of C replacement for `unique_ptr` and `shared_ptr`)

I'm not sure that the `__attribute__((cleanup(func)))` approach is what we are looking for, although it's interesting. Cleanup code is not just about freeing resources, but also about performing some reverting actions, setting output arguments, etc. This might not be easy to do in a separate function. In some situations, you don't want to call any cleanup code if the function succeeds.

> To me `goto` inside macro seems very...unexpected by someone who did not read this particular PR. And maybe better name for `error:` is `cleanup:` since code branches without error go through `error:` part without return first.

Yes, it's a bit unusual and requires some getting used to.

`error` or `cleanup`... I have no problem with either. Both work well, but neither is perfect.

core/embed/lib/error_handling.h
```diff
@@ -20,6 +20,151 @@
...
+// by storing the negated value in the upper 16 bits.
+typedef enum {
+  TS_OK = TS_BUILD(0),
+  TS_ERROR = TS_BUILD(1),          // Generic error
+  TS_ERROR_BUSY = TS_BUILD(2),     // Busy
+  TS_ERROR_TIMEOUT = TS_BUILD(3),  // Timeout
+  TS_ERROR_NOTINIT = TS_BUILD(4),  // Not initialized
+  TS_ERROR_ARG = TS_BUILD(4),      // Invalid argument
+  TS_ERROR_IO = TS_BUILD(5),       // I/O error
+} ts_t;
```
andrewkozlik, line 44: I wonder what the set of error codes should be here. This particular set seems very specific to trezorhal. On the other hand trezor-crypto, for example, would use a completely different set of codes, like `TS_ERROR_SIGNATURE_INVALID`. Would this enum just be a large pile of error codes from various areas?

My impression is that if it is supposed to be a universal error type, then the set should be either very trivial (`sectrue`, `secfalse`) or maybe it should be something abstract like the protobuf `FailureType`.

core/embed/lib/error_handling.h
cepetr: I understand, it's a lot about getting used to it. People are always confused when they see it for the first time. However, this way of coding is very compact, and my experience is that it ends up being very easy to read.

core/embed/lib/error_handling.h
cepetr: Yes, I agree, I will come up with something more sophisticated.

core/embed/lib/error_handling.h
cepetr: This approach is similar to the errno codes used in the libc - all codes are defined in one place. If you need a new error code, define it here, but try to come up with the most generic name so that someone else can use it as well.

However, we can change it. The error type does not need to be an enum; we can use `#define` and allow modules to define their own error codes. For example, the storage or optiga driver can then have specific error codes defined elsewhere.

Not sure what's better. Both ways have their own pros and cons.

core/embed/lib/error_handling.h
cepetr: In past I've used more weird name instead of `verify_xxx() ` so the side effect was expected;-) Maybe different name could help.

core/embed/lib/error_handling.h
cepetr: Yes, I agree. That's a good point. This approach eliminates a lot of bugs.

core/embed/lib/error_handling.h
cepetr: Another perspective is that using this macro enables us to log all errors in the debug or emulator build.

andrewkozlik: I don't doubt that it is very compact and I am sure that if all devs in the team would work with it daily, it would be very easy to read and they would find nothing unexpected about the side effects. However, I think that most of the development now gets done in Python and in Rust. So if a dev needs to work with the C code only occasionally, then they have to relearn the idioms again each time. Furthermore, since we are open-source, anyone who wants to look at our code would have to familiarize themselves with this idiom before being able to understand what's going on.

One example of macros with side effects that we have in our code are those in `legacy/firmware/fsm.c`: `CHECK_PIN`, `CHECK_PIN_UNCACHED` etc. I feel it's always a real pain to touch the code that uses these macros. It would be easier for me if they were inlined.

cepetr: I understand your point of view. Unexpected side effects are almost always inappropriate - `CHECK_PIN()` and its counterparts are good examples. But I don't believe that `verify_xxx()` falls into this category since it's a cross-cutting concept used throughout the entire codebase (hopefully in the future:-), and the side effect is quite apparent and local.

Our code is open source and should be readable for anybody without special 'training'. However, I think it already contains a lot of complexity, so it's not true that it's accessible to everybody for a long time. 

But yes, I understand your concerns.

core/embed/lib/error_handling.h
```diff
@@ -20,6 +20,151 @@
...
+  } while (0)
+
+// Jumps to `error` label if the condition is not `sectrue`.
+#define verify_sec(seccond, status) \
+  do {                              \
+    if ((seccond) != sectrue) {     \
+      __status = status;            \
+      goto error;                   \
+    }                               \
+  } while (0)
```
andrewkozlik, line 135: Looking at the macros defined here, it looks like we are not getting rid of secbool. Our code currently uses `bool`, `secbool`, a custom enum (e.g. optiga code) or a custom set of integers (e.g. trezor-crypto) to indicate the return status. The most familiar, compact and easy to use is `bool`. A more secure variant is `secbool`, which should have better resilience against fault injection, but it can be cumbersome to use. The enums are also fine, but they lack type safety. Finally, the custom integers are an abomination. I think that we should be aiming to narrow down and consolidate these different approaches. That could mean for example making secbool less cumbersome to use and possibly building on top of it, but not building yet another approach alongside it.

core/embed/lib/error_handling.h
andrewkozlik: Perhaps we don't need to have a large minimum distance, but only a large distance between `TS_OK` and the error states.

cepetr: see this change cf075b5370487e7a88db7ad3295abd6196c1da4d, now the hamming distance between `TS_OK` and any `TS_ERROR_xxx` is 16. Hamming distance between error codes can be improved as well.

core/embed/lib/error_handling.h
cepetr: I've changed `ts_t` to a struct in this commit cf075b5370487e7a88db7ad3295abd6196c1da4d. This change helped me to find some bugs that were introduced during  the sdcard driver refactoring...

core/embed/lib/error_handling.h
cepetr: I've been thinking about it too. However, I still don't know how to reasonably connect it. Using two different types to represent boolean values isn't ideal. 

Maybe after the this change cf075b5370487e7a88db7ad3295abd6196c1da4d, 'secfalse' is more like `TS_OK` and `sectrue` is more like TS_BUILD(0xAAAA)... I will think about it.

core/embed/lib/error_handling.h
cepetr: By changing the enum to an integer, we have lost the debugger's ability to show error codes in a readable form. However, we can easily fix this by using [Natvis](https://code.visualstudio.com/docs/cpp/natvis).

cepetr: I've also thought about module-specific error codes. Look at the notes below. In the end, we only need to add two new codes that are not so specific and can be used elsewhere:

One common set of error codes also allows us to create a function that returns an error name string.

```c
typedef enum _optiga_pin_result {
  OPTIGA_PIN_SUCCESS = 0,       // The operation completed successfully.
  OPTIGA_PIN_INVALID,           // The PIN is invalid.
  OPTIGA_PIN_COUNTER_EXCEEDED,  // The PIN try counter limit was exceeded.
  OPTIGA_PIN_ERROR,             // Optiga processing or communication error.
} optiga_pin_result;

OPTIGA_PIN_SUCCESS => TS_OK
OPTIGA_PIN_INVALID => TS_ERROR_ARG
OPTIGA_PIN_COUNTER_EXCEEDED => new error code: TS_ERROR_TRY_LIMIT
OPTIGA_PIN_ERROR => TS_ERROR_IO


typedef enum _optiga_sign_result {
  OPTIGA_SIGN_SUCCESS = 0,   // The operation completed successfully.
  OPTIGA_SIGN_INACCESSIBLE,  // The signing key is inaccessible.
  OPTIGA_SIGN_ERROR,         // Invalid parameters or Optiga processing or
                             // communication error.
} optiga_sign_result;

OPTIGA_SIGN_SUCCESS => TS_OK
OPTIGA_SIGN_INCACESSIBLE => new error code: TS_ERROR_SIGNING_KEY
OPTIGA_SIGN_ERROR => TS_ERROR_IO

```

core/embed/lib/error_handling.h
```diff
@@ -20,6 +20,167 @@
...
+      __fatal_error("ts_check() error", __FILE_NAME__, __LINE__); \
+    }                                                             \
+    _checked;                                                     \
+  })
+
+// Returns `true` if status code is `TS_OK`
+#define ts_ok(status) (_ts_checked(status).code == TS_OK.code)
+
+// Returns `true` if status code is NOT `TS_OK`
+#define ts_error(status) (_ts_checked(status).code != TS_OK.code)
```
andrewkozlik, line 77: If there will be multiple error states, then we should also have `ts_eq()`.

core/embed/lib/error_handling.h
```diff
@@ -20,6 +20,167 @@
...
+  // Do not access this field directly,
+  // use `ts_ok()` and `ts_error()` macros.
+  uint32_t code;
+} ts_t;
+
+// Build status code from any 16-bit value.
+//
+// Status codes are hardened against fault injections
+// by storing the same value in the upper 16 bits.
+#define TS_BUILD(code) ((ts_t){(code) | ((code) << 16)})
```
andrewkozlik, line 39: Is there any reason not to use `(const ts_t)`?

core/embed/lib/error_handling.h
andrewkozlik: It might be also nice to introduce a `ts_switch(status)` macro which expands to `switch(_ts_checked(status).code)`.

core/embed/lib/error_handling.h
```diff
@@ -20,6 +20,167 @@
...
+// Check status code consistency and returns its value.
+// If invalid status code is detected, it will call `__fatal_error()`.
+#define _ts_checked(status)                                       \
+  ({                                                              \
+    ts_t _checked = (status);                                     \
+    if ((_checked.code & 0xFFFF) != (_checked.code >> 16)) {      \
+      __fatal_error("ts_check() error", __FILE_NAME__, __LINE__); \
+    }                                                             \
+    _checked;                                                     \
+  })
```
andrewkozlik, line 71: I wonder how much size and time overhead this will add throughout the entire code base and whether it's worth it. It could also be a barrier to optimization.

core/embed/lib/error_handling.h
cepetr: Probably not, but I'm not sure if adding `const` here brings any advantage.

core/embed/lib/error_handling.h
andrewkozlik: If we do decide to go ahead with this idiom, then we should at least use uppercase for the macros that have side effects to emphasize to the reader that they are macros. `verify_ok()` -> `TS_CHECK()`

core/embed/lib/error_handling.h
andrewkozlik: As a rule of thumb if it can be const, I make it const. It generally helps to detect bugs early, e.g. something like `TS_OK = verify_status()`.

core/embed/lib/error_handling.h
andrewkozlik: I think the mechanism needs to allow easy translation of error codes and the ability to suppress certain errors or treat them as success.  This should make that possible.

core/embed/lib/error_handling.h
cepetr: I haven't looked at the assembly code yet, so I can't tell what the cost is.  That's just an idea and we can omit this code if we think it's not worth it. I think this could be also optional.

core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c
```diff
@@ -78,50 +77,55 @@ static haptic_driver_t g_haptic_driver = {
...
+  ts_t status;
+
   // select library
-  if (!drv2625_set_reg(DRV2625_REG_LIBRARY,
-                       LIB_SEL | DRV2625_REG_LIBRARY_GAIN_25)) {
-    return false;
-  }
+  status = drv2625_set_reg(DRV2625_REG_LIBRARY,
+                           LIB_SEL | DRV2625_REG_LIBRARY_GAIN_25);
+  verify_ok(status);
```
andrewkozlik, line 108: What's the use of the `status` variable here?

core/embed/lib/error_handling.h
```diff
@@ -20,6 +20,167 @@
...
+//
+//   return verify_status();
+// }
+
+// Declares a status variable and initializes it to `TS_OK`.
+// This variable is used to store the status
+#define verify_init() __attribute__((unused)) ts_t __status = TS_OK;
+
+// Returns the current status.
+#define verify_status() (__status)
```
andrewkozlik, line 113: The naming of the macros above is confusing. `verify_*()` usually have a flow control side effect, but these don't. Maybe consider `#define TS_RETURN return __status` and not allow accessing `__status` directly at all.

core/embed/lib/error_handling.h
andrewkozlik: I am not a fan of mapping the codes like `OPTIGA_PIN_INVALID` -> `TS_ERROR_ARG`. That just seems error-prone, e.g. if some sub-function returns `TS_ERROR_ARG`, then we have to watch out to translate that to `OPTIGA_PIN_ERROR` or `TS_ERROR_IO`, to distinguish it from an invalid PIN. It's a pretty important distinction. If the PIN is invalid, then the user should be offered to retry, but if there is a runtime error, then we probably want the user to restart their device, rather than panic and deplete their PIN attempts.

core/embed/lib/error_handling.h
cepetr: I've added `ts_eq()` and `ts_string()`, see 8b027fbcf0760c154cfebb56e5dddf01f4903e33.

The following code cannot be compiled since `TS_xxx` is a structure. For some reason, the compiler cannot deduce that TS_OK.code is a constant expression... :-(.

```c
switch (status.code) {
  case TS_OK.code:
  case TS_ERROR.code:
}
``` 
Maybe we can give up using `case` for error codes and always use if. But I'll try to find some solution.

andrewkozlik:
```c
#define ts_case(status) case status.code

ts_switch(f()) {
  ts_case(TS_OK):
  // ...
  break;
default:
  // ...
}

File: core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c (hunk 8 above)
Comment (cepetr) [reply core/embed/trezorhal/stm32u5/haptic/drv2625/drv2625.c:108]: If you split this into two lines, you can place a breakpoint at the line containing `verify_ok()` to observe the operation status. This pattern is very useful.

File: core/embed/lib/error_handling.h (hunk 9 above)
Comment (cepetr) [reply core/embed/lib/error_handling.h:128]: Ok, what about this:

```c
ts_t module_function(int arg1, int arg2) {
  // Initialize the `__status` variable and set it to `TS_OK`.
  // This should be on the first line inside the function (rule 1).
  TS_INIT;

  // Check arguments to ensure they are greater than 0.
  TS_CHECK_ARG(arg1 > 0);
  TS_CHECK_ARG(arg2 > 0);
  
  // Check the result of another function with a `ts_t` return value.
  // The `status` value is properly propagated.
  // Splitting it into two lines makes debugging easier (rule 2).
  ts_t status = another_function();
  TS_CHECK_OK(status);
  
  // Check the result of an STM32 HAL function returning `HAL_StatusTypeDef`.
  // The HAL_xxx error is properly mapped to TS_xxx and propagated.
  HAL_StatusTypeDef hal_status = HAL_function();
  TS_CHECK_HAL_OK(hal_status);
  
  // Check a generic condition inside the function and propagate the status code
  // if the condition is not fulfilled.
  TS_CHECK(arg1 * arg2 < 1000, TS_ERROR);

  // Alternatively, use the secure variant working with `secbool`.
  secbool ready = usb_ready();
  TS_CHECK_SEC(ready);
  
  // In some cases it's useful to return before the cleanup code, (rule 3: don't return `TS_OK`)
  TS_RETURN;
  
error:
  // Potential cleanup code goes here.
  // ...

  // Return the content of the `__status` variable.
  // In case of no error, it's `TS_OK`.
  TS_RETURN;
}
```

But I personally prefer:

```c
  // In some cases it's useful to return before the cleanup code, (rule 3: don't return `TS_OK`)
  return TS_STATUS;
  
error:
  // Potential cleanup code goes here.
  // ...

  // Return the content of the `__status` variable.
  // In case of no error, it's `TS_OK`.
  return TS_STATUS;
}
```

cepetr: But I'm getting used to it. Maybe I like TS_RETURN more... :-)

andrewkozlik:
I created a document in notion summarizing my concerns.
https://www.notion.so/satoshilabs/Unify-error-handling-in-C-aebe0fe5afa1417996329350869835c4

core/embed/lib/error_handling.h
cepetr: I made these changes here 7b70f0ae4f0bcc59b6ac9637825a6b59a262bb1b

core/embed/lib/error_handling.h
cepetr: Ok, I understand. Maybe this case requires a more specific error code. We should also consider whether error codes should be used for situations like 'invalid PIN.' Is it really an error, or is it just invalid user input? Should be treated the same way as i/o error or invalid argument?

core/embed/lib/error_handling.h
andrewkozlik: You mentioned IRL some issue with the compiler not being able to treat the value as a constant. In C++ defining `TS_OK` etc. using `constexpr` would solve this, but that probably isn't an option in C. Another idea is something like this:
```
#define ts_case(status) case status##__value
#define TS_OK__value TS_BUILD(0)
#define TS_OK ((const ts_t){TS_OK__value})
```

cepetr:
Replaced with #6223

PR #4010, opened 2024-07-09
feat(core): deal with non-breaking spaces

### LINKED ISSUE #2755: Non-breaking space
Introduce non-breaking space character, which will be used to prevent breaking lines where non appropriate, like after the `a` article.

ibz:
Fixes #2755


core/.changelog.d/2755.added
```diff
@@ -0,0 +1 @@
+Deal with non-breaking spaces.
```
Hannsek, line 1: Not really ideal changelog entry 😄

ibz: > Not really ideal changelog entry 😄

Should I just remove the changelog? Does a changelog make sense for this feature?

ibz: Removed it completely...

mmilata:
Can you please also revert cca066c4768e68acb29bdccb8cf1623cffc665f6 and 1c9f9254185c64ce5bfd64f6bbe5956fb76bf6e0 as part of this PR? It will require regenerating fixtures.json but at least there'll be some data to test the code on.

core/embed/rust/src/translations/generated/translated_string.rs.mako
```diff
@@ -55,7 +56,7 @@ impl TranslatedString {
             %if any(name.startswith(prefix + "__") for prefix in ALTCOIN_PREFIXES):
             #[cfg(feature = "universal_fw")]
             %endif
-            Self::${name} => ${json.dumps(en_data.get(name, '""'))},
+            Self::${name} => ${re.sub(r'\\u([0-9a-f]{1,5})', r'\\u{\g<1>}', json.dumps(en_data.get(name, '""')))},
```
matejcik, line 59: why is this here?

ibz: Rust requires `\u{NNN}` rather than `\uNNN` which is what JSON yields.

ibz:
> Can you please also revert [cca066c](https://github.com/trezor/trezor-firmware/commit/cca066c4768e68acb29bdccb8cf1623cffc665f6) and [1c9f925](https://github.com/trezor/trezor-firmware/commit/1c9f9254185c64ce5bfd64f6bbe5956fb76bf6e0) as part of this PR? It will require regenerating fixtures.json but at least there'll be some data to test the code on.

Done. Perhaps the question now is why there are so few differences in fixtures.

The only one I see is this:

image

But I would have expected more differences, TBH.

core/embed/rust/src/translations/generated/translated_string.rs.mako
ibz: That being said, @matejcik - I found one issue with my regexp... So I decided to change it to something more conservative: https://github.com/trezor/trezor-firmware/pull/4010/commits/9ad8417c16f96240a869de880317b9823db5f7f5#diff-6b77a3cb1462becb160af081a8eb77d7c3f1c77d4365438e622b4ea44c139e6cR59

Issue was... if the word after the non-breaking space would start with a-f ... the first letter(s) of that word could be considered to be part of the code (matched by the regexp) and yield an invalid character.

So, my new solution is, rather than trying to match anything, just match `00a0` ... Anything else would lead to a Rust compilation error, in which case we could edit the regexp and add more cases... Although, most likely we don't really want any special characters anyway!

ibz:
I can confirm that the code works as expected, I tested by adding a non-breaking space in some random string, at a place where rendering would normally insert a new line, and the line did not break in that place anymore.

I just find it a little bit strange that none of the non-breaking spaces that currently exist in the translations (the reverted commit mentioned above) have any effect on any of the UI tests... Would have expected more fixtures to break.

core/embed/rust/src/translations/generated/translated_string.rs.mako
matejcik: yes, well, per json.org the escape must be exactly 4 hex chars in all cases:
https://www.json.org/img/string.png
dunno why you figured the `{1,5}` thing...? (in your defence i missed it too)

ibz: If it's always 4, it should be safe to use a regexp. Changed.

matejcik (approved):
utACK

bosomt:
**QA OK**

```
translation-T3T1-cs-CZ-2.8.1-unsigned.bin
firmware-T3T1-2.8.1-5bba2e0aa-dirty.bin
```

PR #4018, opened 2024-07-10
fix vector table alignment on STM32U5

TychoVrahe:
This PR fixes issue with vector table alignment on STM32U5.

The cortex-m33 programming manual states:
_The silicon vendor must configure the required alignment, which depends on the number of interrupts
implemented. The minimum alignment is 32 words, enough for up to 16 interrupts. For more interrupts, adjust
the alignment by rounding up to the next power of two. For example, if you require 21 interrupts, the alignment
must be on a 64-word boundary because the required table size is 37 words, and the next power of two is 64._

Given that in vectortable.s we configure > 128 interrupts, we require alignment to 256 words, or 0x400 bytes. 

Even though STM32U585 might not be affected due to real number of interrupts is 126 and only the higher end U5s exceed 128, i prefer to have the alignment unified across the whole U5 family.

The T3T1 was already aligned properly due to conveniently long vendor header, but T3B1 and D002 needed adjustments. 

Linker scripts are modified to enforce this. 

The consequence of having too low alignment was that wrong interrupt handler was called: i.e. instead of GPU2D handler (n. 132), Tamper IRQ handler was called (n. 4 = 132 mod 128)


matejcik:
not a fan of increasing vendor header size just to push an alignment of something somewhere
is that really required here?

TychoVrahe:
In bootloader we have:
`jump_to(FIRMWARE_START + vhdr.hdrlen + IMAGE_HEADER_SIZE);` 
so we'd have to add the real alignment there.

Also without the alignment in vendorheader, the headertool doesn't recognize the firmware image. I guess because the size doesn't fit anymore? would have to take that into account there too.

So probably this is doable, as the alignment for all existing models will stay zero so no incompatibility with the old bootloader

cepetr:
Maybe I am also in favor of not changing it on the U585 since there's no problem. We'll have distinct linker files for different U5 MCUs, so I would prefer to make changes on the affected models (DISC2, T3W1). I'm not sure about changing the vendor header size, but the better solution is probably to change the alignment around `jump_to()`.

matejcik (commented):
concept ACK, this is much more legible to me in terms of what we're trying to accomplish

matejcik (approved):

cepetr (approved):
utACK

PR #4020, opened 2024-07-10
Fix broken coverage reporting

mmilata:
In 5fc3c6e617619eee14429d44ab6e064ed3f5801e the termination code was modified not to raise micropython SystemExit just before exit(). Unfortunately this is needed for test coverage reports as `prof.py` relies on this exception to write out coverage data.

This has not been caught by CI because the coverage jobs were failing intermittently for long enough time for everyone to learn to ignore them. Hopefully I'll fix this next.

### DIFF
--- a/core/embed/trezorhal/unix/common.c
+++ b/core/embed/trezorhal/unix/common.c
@@ -28,6 +28,8 @@
 #include "display.h"
 #include "memzero.h"
 
+void __attribute__((noreturn)) main_clean_exit();
+
 void __attribute__((noreturn)) trezor_shutdown(void) {
   printf("SHUTDOWN\n");
 
@@ -49,7 +51,7 @@ uint32_t hal_ticks_ms() {
 static int SDLCALL emulator_event_filter(void *userdata, SDL_Event *event) {
   switch (event->type) {
     case SDL_QUIT:
-      exit(3);
+      main_clean_exit();
       return 0;
     case SDL_KEYUP:
       if (event->key.repeat) {
@@ -57,7 +59,7 @@ static int SDLCALL emulator_event_filter(void *userdata, SDL_Event *event) {
       }
       switch (event->key.keysym.sym) {
         case SDLK_ESCAPE:
-          exit(3);
+          main_clean_exit();
           return 0;
         case SDLK_p:
           display_save("emu");
--- a/core/embed/unix/main.c
+++ b/core/embed/unix/main.c
@@ -408,6 +408,19 @@ STATIC void set_sys_argv(char *argv[], int argc, int start_arg) {
   }
 }
 
+// Inject SystemExit exception. This is primarily needed by prof.py to run the
+// atexit() handler.
+void __attribute__((noreturn)) main_clean_exit() {
+  const int status = 3;
+  fflush(stdout);
+  fflush(stderr);
+  // sys.exit is disabled, so raise a SystemExit exception directly
+  nlr_raise(mp_obj_new_exception_arg1(&mp_type_SystemExit,
+                                      MP_OBJ_NEW_SMALL_INT(status)));
+  // the above shouldn't return, but make sure we exit just in case
+  exit(status);
+}
+
 #ifdef _WIN32
 #define PATHLIST_SEP_CHAR ';'
 #else
@@ -682,7 +695,9 @@ MP_NOINLINE int main_(int argc, char **argv) {
 #if !MICROPY_VFS
 
 #ifdef TREZOR_EMULATOR_FROZEN
-mp_import_stat_t mp_import_stat(const char *path) { return MP_IMPORT_STAT_NO_EXIST; }
+mp_import_stat_t mp_import_stat(const char *path) {
+  return MP_IMPORT_STAT_NO_EXIST;
+}
 #else
 mp_import_stat_t mp_import_stat(const char *path) {
   struct stat st;
--- a/core/embed/bootloader/emulator.c
+++ b/core/embed/bootloader/emulator.c
@@ -22,6 +22,9 @@ uint8_t *FIRMWARE_START = 0;
 
 void set_core_clock(int) {}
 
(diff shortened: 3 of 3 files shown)


core/embed/trezorhal/unix/common.c
```diff
@@ -28,14 +28,16 @@
...
+
 void __attribute__((noreturn)) trezor_shutdown(void) {
   printf("SHUTDOWN\n");
 
   // Wait some time to let the user see the displayed
   // message before shutting down
   hal_delay(3000);
 
-  exit(3);
+  main_clean_exit(3);
```
cepetr, line 40: If `trezor_shutdown` is called before MicroPython is initialized, calling `main_clean_exit()` leads to an exception. It's not a big problem since it's an emulator, but it is still an issue.

I wonder if it would be possible to remove the hardcoded constant (3) from here. Is it really used anywhere?

core/embed/trezorhal/unix/common.c
```diff
@@ -28,14 +28,16 @@
 #include "display.h"
 #include "memzero.h"
 
+extern void __attribute__((noreturn)) main_clean_exit(int);
```
cepetr, line 31: This `extern` indicates that there's an issue with the dependency order. However, let's leave it as is for now, since we are gradually moving towards not having `common.c`/`common.h` at all, which could resolve this issue automatically.

cepetr (changes requested):
Looks good, but there are still some minor issues that may need to be resolved.

core/embed/trezorhal/unix/common.c
matejcik: @cepetr i recall the two of us discussing whether we need to call `main_clean_exit` from here, and it causing some sort of problem (maybe recursing into trezor_shutdown, or fatal_error, or something)
do you remember if that's not a problem anymore?

matejcik: > calling `main_clean_exit()` leads to an exception

`main_clean_exit` should check whether micropython has been set up before nlr_raising?

cepetr: > @cepetr i recall the two of us discussing whether we need to call `main_clean_exit` from here, and it causing some sort of problem (maybe recursing into trezor_shutdown, or fatal_error, or something) do you remember if that's not a problem anymore?

Yes, you are right. I thought it was solved, but it wasn't. The problem occurs when calling `panic!()` or something similar from the `render()` routine. The call to `nlr_raise()` in `main_clean_exit()` triggers `trezor_shutdown()` again because the code tries to borrow `RefCell<LayoutObj>`, which is already borrowed.

cepetr: > > calling `main_clean_exit()` leads to an exception
> 
> `main_clean_exit` should check whether micropython has been set up before nlr_raising?

I think so.

matejcik: > The call to `nlr_raise()` in `main_clean_exit()` triggers `trezor_shutdown()` again because the code tries to borrow `RefCell<LayoutObj>`

right.

ISTM we should modify the `exit`s for the SDL event loop, but keep this one a raw `exit()`. it seems appropriate that `trezor_shutdown()`, which is a "hard halt" for the real device, also causes a "hard halt" in the emulator, and we should not expect to recover even for prof.py
(we'll get coverage when the emulator is shut down cleanly via `SDL_QUIT` event)

core/embed/trezorhal/unix/common.c
matejcik: `build_bootloader_emu` is failing because of this

mmilata: extern removed and bootloader_emu build fixed in 5475558ff22795b24ec380d4bc5b2ce6900251ab

core/embed/trezorhal/unix/common.c
mmilata: check added in 5475558ff22795b24ec380d4bc5b2ce6900251ab, please take a look

core/embed/unix/main.c
```diff
@@ -61,6 +61,8 @@
 STATIC bool compile_only = false;
 STATIC uint emit_opt = MP_EMIT_OPT_NONE;
 
+STATIC bool mp_initialized = false;
+
```
cepetr, line 65: To keep it as simple as possible: Since you have removed `main_clean_exit()` from `trezor_shutdown()`, this flag  is no longer needed. Additionally, I would suggest adding the static keyword to `main_clean_exit()`.

mmilata: ade1fdcc22b2ce78d8faf02864e52cebeb997ce3

Static won't work because it's called from another file. I could move it there but then I'd have to #include bunch of micropython stuff where there's currently none. At least the declaration could be moved to a header file, but which one ...

core/embed/trezorhal/unix/common.c
```diff
@@ -28,6 +28,8 @@
 #include "display.h"
 #include "memzero.h"
 
+void __attribute__((noreturn)) main_clean_exit();
+
```
cepetr, line 32: We can get rid of this now.

core/embed/unix/main.c
cepetr: Ah, okay. This would require deeper plowing... So I am satisfied.

core/embed/trezorhal/unix/common.c
mmilata:
```
embed/trezorhal/unix/common.c: In function ‘emulator_event_filter’:
embed/trezorhal/unix/common.c:52:7: error: implicit declaration of function ‘main_clean_exit’ [-Werror=implicit-function-declaration]
   52 |       main_clean_exit();
```

cepetr: Oh, sorry, I overlooked this. So everything is ok.

cepetr (approved):

PR #4021, opened 2024-07-11
add TS5 background image for emulator

TychoVrahe:


prusnak:
Can we pad the new image so it also has dimensions 400px x 600px?

No particular reason, I just like it more that way.

TychoVrahe:
like this perhaps?

Device-WithBorder

prusnak:
> like this perhaps?

yes

matejcik (approved):

prusnak (approved):

PR #4023, opened 2024-07-11
fit more info into success screen

TychoVrahe:
Introduces more sophisticated success animation to fit more info.

Some refactoring was done to frame: mostly i extracted header to separate component to not complicate the frame code with animations. As a side effect, this also safe some flash as the header does not need to be generic. I also extracted the animation from frame to save some more, overall these changes save about 7kB of flash. Some of it is claimed by the animations so net savings about 4kB.

Edit: added one more commit with frame code reshuffling, saving additional ~7kB.  https://github.com/trezor/trezor-firmware/pull/4023/commits/4c23fb8b99fded1c5071cc2d9aa7c0b3f9780936


core/translations/en.json
```diff
@@ -947,6 +947,8 @@
     "words__unknown": "Unknown",
     "words__warning": "Warning",
     "words__writable": "Writable",
-    "words__yes": "Yes"
+    "words__yes": "Yes",
+    "result__title_success": "Success",
+    "result__title_neutral": "Done"
```
mmilata, line 952: Nit: consider putting these under `words__title_`

mmilata (approved):
beautiful

core/translations/en.json
TychoVrahe: changed https://github.com/trezor/trezor-firmware/commit/14029121467f59c97ce2584c79d88c88570d72d3

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.1 regular (revision 008490bf4b8a04a085c59e217f743c9713727480)

PR #4026, opened 2024-07-13
chore(core): add signature for 2.8.0 translation blobs

matejcik:


mmilata (approved):
Fields match (but not sure what commit should be exactly).

matejcik:
> (but not sure what commit should be exactly)

not that important, honestly. it should be the commit that generates the particular Merkle tree, the tooling should ensure that this is the case

PR #4037, opened 2024-07-18
Fix mercury SLIP39 checklist more info

obrusvit:
More info during setting the number of shares and its threshold now:
- correctly shows number of words per each share (20/33)
- correctly uses the chosen number of shares in templated string
- removes incorrect "Skip backup" title
- fixes some non-`en` copy


Hannsek:
Please change the copy according to Figma 🙏🏻

obrusvit:
@Hannsek I changed the copy here. The formatting is slightly off but would be more time-consuming to get it right. https://github.com/trezor/trezor-firmware/pull/4037/commits/a8f666f8606f59fff5d40a53b8d73f39152fd8ca 
I would also like to use these new strings in wallet backup creation in model T but they didn't fit right.

I also added minor copy changes of other langs which I noticed and were easy to fix here https://github.com/trezor/trezor-firmware/pull/4037/commits/8296c7085737719a4c49c8918ab6c7298739a68b

Hannsek:
According which file did you changed the copy? Is it [this](https://www.figma.com/design/n43caWJFGJlT167AQcxruu/%E3%80%94-Prod-%E3%80%95Safe-5?node-id=2989-2329&t=YsfCQQnwF3c5jw3f-0)?

I didn't get through all of them but I found one nit:

From CI:
image

From Figma:
image

obrusvit:
@Hannsek Yes, I changed the copy according to the flow you posted. I don't see the screen you posted "From  CI" in the CI.
See full multi-share (regular Shamir) flow [here](https://data.trezor.io/dev/firmware/ui_report/10010429849/T3T1-en-core_device_test/failed/T3T1_en-device_tests-test_msg_backup_device.py%3A%3Atest_backup_slip39_basic%5Bclick_info%5D.html)

The copy is the same with these two limitations:
- the space between two paragraphs is not respected
- I use `"... you'll need 3 backup shares ..."` instead of Figma's `"... you'll need three backup shares ..."` because formatting with numbers is easier for now. We'd need the mapping for numeral -> word

Hannsek:
Strange. But great then. 👍🏻

mmilata (approved):

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.1 regular (revision 33f5d4733065ac631d61ff458a5db32c81641124)

PR #4040, opened 2024-07-18
refactor(core/mercury): remove unnecessary Child

mmilata:
[no changelog]


obrusvit (approved):
Nice cleanup. 

`Child` kept in `component/bootloader/*.rs` and `component/bl_confirm.rs` for the needs of the bootloader. This might get deleted as well in the future when we switch the bootloader to new rendering.

PR #4045, opened 2024-07-19
Improve optiga initialization error handling

onvej-sl:


andrewkozlik:
utACK

For future reference please try to summarize any relevant observations that you made while experimenting with the chip and explain the reason to move `optiga_open_application()` after the handshake. (I am not saying it's right or wrong.)

onvej-sl:
The only reason for calling `optiga_open_application` after `optiga_sec_chan_handshake` instead of before it was for code simplification. If `optiga_open_application` were called before `optiga_sec_chan_handshake` and the handshake subsequently failed, `optiga_sec_chan_handshake` would need to be called again. I tested both variants and did not observe any differences.

I discovered the following: 
  * If the `optiga_sec_chan_handshake` function fails due to a damaged pairing secret, the subsequent commands fail as well because Optiga is [not busy but not responsive](https://github.com/trezor/trezor-firmware/blob/onvej-sl/optiga-reset/core/embed/trezorhal/optiga/optiga_transport.c#L427).
  * Calling `optiga_soft_reset` resolves the issue.
  * Calling `optiga_reset` does not resolve the issue. However, if you call `optiga_read_sec` before calling `optiga_sec_chan_handshake`, `optiga_reset` resolves the issue.

andrewkozlik (approved):

PR #4047, opened 2024-07-21
core/translations: add :it: :portugal:

mmilata:
Turkish translation fails click tests and from UI reports it seems there are also some missing glyphs. PR adds this third language in the current state but does not enable tests for it.


mmilata:
Since we're not in a rush wrt blob signing I'd say let's fix these issues & then merge.

If QA or anybody wants unsigned blobs then please let me know.

Hannsek:
Do you know why it device tests are failing? Is there a reason for it?

mmilata:
The failing T2B1 click tests are mysterious, it's happening in every language including english, and this PR doesn't even touch any code that's involved, seemingly?

https://github.com/trezor/trezor-firmware/actions/runs/10066871496/job/27829340045

Hannsek:
I was reffering to `it device test still failing`

mmilata:
Got it in a state where only (it,pt,tr) T2B1 click tests are failing. There is some issue with how the test code handles translated strings. Possibly related: if `inputs__delete` translation string is not uppercase the T2B1 click tests also break apart.

ibz:
LGTM as far as this PR is concerned.

Whether the translations are good, complete, etc, is outside the scope of this.

We will probably need some way to at least dump keys that are missing in various translations, but more generally to ease keeping translations in sync.

ibz (approved):
I think we need a mechanism to detect if a char is missing (as part of the tests?). For example, you did not include some chars that are technically allowed in some languages, and I understand that it is an optimization, but that also means that if something containing an "È" suddenly appears in Italian, for example, we will probably not notice it, and it will render as a "?"...

mmilata:
Thanks for the review!

> We will probably need some way to at least dump keys that are missing in various translations, but more generally to ease keeping translations in sync.

`core/transaltions/cli.py` sounds like the right place for this.

> I think we need a mechanism to detect if a char is missing

Agreed, opened #4149.

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.3 regular (revision 7f373ae71eca855dd41b4a0fdcc7cadaa13a8281)

PR #4050, opened 2024-07-22
[draft] THP

M1nd3r:


core/src/trezor/wire/thp/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+    ):
+
+        aes_ctx = aesgcm(self.k, IV_2)
+
+        # The new value of hash `h` MUST be computed before the `encrypted_host_static_pubkey` is decrypted.
+        # However, decryption of `encrypted_host_static_pubkey` MUST use the previous value of `h` for
+        # authentication of the gcm tag.
+        aes_ctx.auth(self.h)  # Authenticate with the previous value of `h`
+        self.h = _hash_of_two(self.h, encrypted_host_static_pubkey)  # Compute new value
+        aes_ctx.decrypt_in_place(
```
onvej-sl, line 136: What is the reason for using in-place decryption here while not using in-place encryption in `handle_th1_crypto`?

core/src/trezor/wire/thp/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+        self,
+        encrypted_host_static_pubkey: utils.BufferType,
+        encrypted_payload: utils.BufferType,
+    ):
+
+        aes_ctx = aesgcm(self.k, IV_2)
+
+        # The new value of hash `h` MUST be computed before the `encrypted_host_static_pubkey` is decrypted.
+        # However, decryption of `encrypted_host_static_pubkey` MUST use the previous value of `h` for
+        # authentication of the gcm tag.
```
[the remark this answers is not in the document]
M1nd3r: Yes, I can. The new value of handshake hash h is computed using the encrypted value (line 135).  On the following line (line 136), the encrypted value is overwritten by the decrypted one, thus it cannot be used to derive the correct value of h after that.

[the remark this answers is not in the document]
onvej-sl: I see, this is a consequence of using in-place decryption. I missed that. Thank you.

core/src/trezor/wire/thp/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+        if tag != encrypted_payload[-16:]:
+            raise ThpDecryptionError()
+
+        self.h = _hash_of_two(self.h, memoryview(encrypted_payload)[:-16])
+        self.key_receive, self.key_send = _hkdf(self.ck, b"")
+        print("TREZOR_KEY_RECEIVE:", hexlify(self.key_receive))
+        print("TREZOR_KEY_SEND:   ", hexlify(self.key_send))
+
+    def get_handshake_completion_response(self, trezor_state: bytes) -> bytes:
+        aes_ctx = aesgcm(self.key_send, IV_1)
```
onvej-sl, line 165: This should be `IV_0`. This is the first usage of `key_send`.

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -0,0 +1,389 @@
...
+            + TAG_LENGTH : message_length
+            - CHECKSUM_LENGTH
+            - TAG_LENGTH
+        ],
+        0,
+        "ThpHandshakeCompletionReqNoisePayload",
+    )
+    if TYPE_CHECKING:
+        assert ThpHandshakeCompletionReqNoisePayload.is_type_of(noise_payload)
+    for i in noise_payload.pairing_methods:
```
onvej-sl, line 283: You should check that `noise_payload.pairing_methods` is a subset of `_ENABLED_PAIRING_METHODS`.

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -0,0 +1,389 @@
...
+        if i not in ctx.selected_pairing_methods:
+            ctx.selected_pairing_methods.append(i)
```
onvej-sl, line 285: Why don't you simply set `ctx.selected_pairing_methods = noise_payload.pairing_methods`?

core/src/trezor/wire/thp/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+        host_static_pubkey = memoryview(encrypted_host_static_pubkey)[:PUBKEY_LENGTH]
+        tag = aes_ctx.finish()
+        if tag != encrypted_host_static_pubkey[-16:]:
+            raise ThpDecryptionError()
+
+        self.ck, self.k = _hkdf(
+            self.ck,
+            curve25519.multiply(self.trezor_ephemeral_privkey, host_static_pubkey),
+        )
+        aes_ctx = aesgcm(self.k, IV_1)
```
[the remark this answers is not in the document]
M1nd3r: The 1 in "IV_1" stands for "first", IV_1 is defined the following way:
IV_1 = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"


Thus, I think it is correct the way it is implemented now.

core/src/trezor/wire/thp/pairing_context.py
```diff
@@ -0,0 +1,249 @@
...
+            )
+        )
+
+    def _get_code_code_entry_str(self) -> str:
+        if self.display_code_entry and self.code_code_entry is not None:
+            code_str = str(self.code_code_entry)
+            print("code_code_entry:", code_str)
+
+            return code_str[:3] + " " + code_str[3:]
+        return "NOT ALLOWED"
```
onvej-sl, line 53: I would rise an exception here.

By the the way, having both `display_code_entry` and `code_code_entry` is redundant since `display_code_entry` is `True` if and only if `code_code_entry` is not `None`.

The same applies to `_get_code_qr_code_str`.

core/src/trezor/wire/thp/crypto.py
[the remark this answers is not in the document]
onvej-sl: My mistake. I though that
```
IV_0 = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
IV_1 = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01"
```
but you use `IV_1` and `IV_2`.

core/src/trezor/wire/thp/crypto.py
M1nd3r: The in-place encryption and decryption is preferred and used when possible, as it lowers the need for allocation of memory.

To encrypt/decrypt "in place", a `bytearray` or `memoryview` object is required. The `trezor_masked_static_pubkey` is defined as an output of the function `curve25519.multiply(...)`, which returns object of type `bytes`.
Thus, it is not possible to encrypt the `trezor_masked_static_pubkey` in place.

core/src/trezor/wire/thp/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+    def handle_th1_crypto(
+        self,
+        device_properties: bytes,
+        host_ephemeral_pubkey: bytes,
+    ) -> tuple[bytes, bytes, bytes]:
+
+        trezor_static_privkey, trezor_static_pubkey = _derive_static_key_pair()
+        self.trezor_ephemeral_privkey = curve25519.generate_secret()
+        trezor_ephemeral_pubkey = curve25519.publickey(self.trezor_ephemeral_privkey)
+        self.h = _hash_of_two(PROTOCOL_NAME, device_properties)
```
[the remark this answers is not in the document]
M1nd3r: fixed here: https://github.com/trezor/trezor-firmware/pull/4050/commits/c6578aa05064628d2a221b1bf2b8a67d66b40f77

core/src/trezor/wire/thp/received_message_handler.py
M1nd3r: verification implemented here: https://github.com/trezor/trezor-firmware/pull/4050/commits/1f7ffd2edabe713963c20c527659d7d43acd77a9

core/src/trezor/wire/thp/received_message_handler.py
M1nd3r: In some previous iteration, there was a (more involved) mechanism for verification of selected pairing methods. After discussion, it was removed, as it was not MVP. This code was but a remnant of the previously functional checks.

I have now implemented replacement for the verification of methods - it does not depend on user preferences but on the used wire interface only. https://github.com/trezor/trezor-firmware/pull/4050/commits/1f7ffd2edabe713963c20c527659d7d43acd77a9

core/src/trezor/wire/thp/cpace.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def generate_keys_and_secret(self, code_code_entry: bytes) -> None:
+        """
+        Generate ephemeral key pair and a shared secret using Elligator2 with X25519.
+        """
+        sha_ctx = sha512(_PREFIX)
+        sha_ctx.update(code_code_entry)
+        sha_ctx.update(_PADDING)
+        sha_ctx.update(self.handshake_hash)
+        pregenerator = sha_ctx.digest()[:32]
```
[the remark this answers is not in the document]
M1nd3r: changes implemented here: https://github.com/trezor/trezor-firmware/pull/4050/commits/6bb0a00f64d474326ee2f06d13d7ebfa325292b5

core/src/trezor/wire/thp/channel.py
```diff
@@ -0,0 +1,317 @@
...
+        return int.from_bytes(self.channel_id, "big")
+
+    def get_channel_state(self) -> int:
+        state = int.from_bytes(self.channel_cache.state, "big")
+        if __debug__:
+            log.debug(__name__, "get_channel_state: %s", state_to_str(state))
+        return state
+
+    def get_handshake_hash(self) -> bytes:
+        return self.channel_cache.get(CHANNEL_HANDSHAKE_HASH) or b""
```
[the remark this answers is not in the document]
M1nd3r: I agree with you. This method was on my "to discuss" list. 
After the change, the method throws an Exception if the handshake hash retrieved from cache is None.


The change: https://github.com/trezor/trezor-firmware/pull/4050/commits/bbb73f760ef2a472f78622fdd6bcff6ad076fd88

core/src/trezor/wire/thp/pairing_context.py
M1nd3r: Methods now raise exceptions if the relevant variable is None. The boolean values are discarded, as they are currently redundant. There can be a scenario, where we want to toggle the visibility of the displayed methods, but now those boolean values are not MVP.

Removal of boolean variables and added exceptions:
https://github.com/trezor/trezor-firmware/pull/4050/commits/bbb73f760ef2a472f78622fdd6bcff6ad076fd88

core/src/trezor/wire/thp/channel.py
```diff
@@ -0,0 +1,319 @@
...
+            auth_data = self.get_handshake_hash()
+
+            assert key_receive is not None
+            assert nonce_receive is not None
+            assert auth_data is not None
+
+            print("Buffer before decryption:", hexlify(noise_buffer))
+            is_tag_valid = crypto.dec(
+                noise_buffer, tag, key_receive, nonce_receive, auth_data
+            )
```
onvej-sl, line 195: I think `auth_data` should be empty. The [standard](http://www.noiseprotocol.org/noise.html) says:
> Transport messages are then encrypted and decrypted by calling EncryptWithAd() and DecryptWithAd() on the relevant CipherState with zero-length associated data.

onvej-sl: I'm not sure if this is documented at all in our notion documentation.

core/src/trezor/wire/thp/cpace.py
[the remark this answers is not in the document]
onvej-sl: One more thing I forgot. The byte preceding the zero bytes changes from 0x50 to 0x6f.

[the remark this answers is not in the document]
M1nd3r: fixed here: https://github.com/trezor/trezor-firmware/pull/4050/commits/24c68bf4e98f207fe4d9cd21b1683dae4e1c36ee

core/src/trezor/wire/thp/channel.py
M1nd3r: `handshake hash` was removed from `auth_data` here: https://github.com/trezor/trezor-firmware/pull/4050/commits/0c1d6108bea6a3029ee9bc11a1d4805bcb0390cf

core/src/trezor/wire/thp/received_message_handler.py
onvej-sl: I am wondering if it might be better to raise an exception if `noise_payload.pairing_methods` is not a subset of the enabled pairing methods. Lets discuss it on Monday.

core/src/trezor/wire/thp/received_message_handler.py
onvej-sl: By the way, the `for` loop could be replaced with set intersection.

onvej-sl:
Well done, @M1nd3r !

I reviewed the usage of cryptography and compliance with our documentation of the secure channel layer.

core/src/trezor/wire/thp/received_message_handler.py
M1nd3r: exception + error message implemented here:
https://github.com/trezor/trezor-firmware/pull/4050/commits/e4029341cfae750c82ced7ffa7cf3d7410971a99

core/src/trezor/wire/thp/received_message_handler.py
M1nd3r: Note taken. Nevertheless, as an intersection will not be used, this comment is resolved (from my POV).

core/src/trezor/wire/thp/crypto.py
onvej-sl: Got it.

core/src/trezor/wire/message_handler.py
```diff
@@ -38,14 +43,33 @@ def wrap_protobuf_load(
...
         if e.args:
             raise DataError("Failed to decode message: " + " ".join(e.args))
         else:
             raise DataError("Failed to decode message")
 
 
+if utils.USE_THP:
+    from trezor.enums import ThpMessageType
+
+    def get_msg_name(msg_type: int) -> str | None:
```
mmilata, line 57: We should probably extend rust protobuf parser to handle this, can help with that.

core/src/trezor/wire/thp/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+    import usb
+
+    methods = _DEFAULT_ENABLED_PAIRING_METHODS.copy()
+    if iface is not None and iface is usb.iface_wire:
+        methods.append(ThpPairingMethod.NoMethod)
+    return methods
+
+
+def _get_device_properties(iface: WireInterface) -> ThpDeviceProperties:
+    # TODO define model variants
```
mmilata, line 152: Will these be TS3/TS5 colours or something else?

core/src/trezor/wire/thp/pairing_context.py
```diff
@@ -0,0 +1,262 @@
...
+            trezorui2.show_address_details(  # noqa
+                qr_title="Scan QR code to pair",
+                address=qr_str,
+                case_sensitive=True,
+                details_title="",
+                account="Code to rewrite:\n" + code_str,
+                path="",
+                xpubs=[],
+            )
+        )
```
mmilata, line 54: Note to self: ask for UI design and implement it.

core/src/trezor/wire/thp/pairing_context.py
```diff
@@ -0,0 +1,262 @@
...
+        super().__init__(channel_ctx.iface, channel_ctx.channel_id)
+        self.channel_ctx: Channel = channel_ctx
+        self.incoming_message = loop.mailbox()
+        self.secret: bytes = random.bytes(16)
+
+        self.display_data: PairingDisplayData = PairingDisplayData()
+        self.cpace: Cpace
+        self.host_name: str
+
+    async def handle(self, is_debug_session: bool = False) -> None:
```
mmilata, line 86: File looks similar to code in `message_handler.py`. Might help with refactoring, we'll probably need something similar for BLE management context anyway.

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -0,0 +1,458 @@
...
+
+        if s is None:
+            s = SeedlessSessionContext(ctx, session_id)
+
+        ctx.sessions[session_id] = s
+        loop.schedule(s.handle())
+
+    elif ctx.sessions[session_id].get_session_state() is SessionState.UNALLOCATED:
+        raise ThpUnallocatedSessionError(session_id)
+        # TODO what should happen here? Should I create a new seedless session?
```
mmilata, line 379: Does this happen when client creates a session with passphrase, then it's evicted from cache later? Returning an error seems like the right thing in this case.

core/src/trezor/wire/thp/session_context.py
```diff
@@ -0,0 +1,169 @@
...
+        log.debug(
+            __name__,
+            "handle - start (channel_id (bytes): %s, session_id: %d)",
+            get_bytes_as_str(self.channel_id),
+            self.session_id,
+        )
+
+    async def _handle_message(
+        self,
+        next_message: Message | None,
```
mmilata, line 71: Argument never used?

core/src/trezor/wire/thp/transmission_loop.py
```diff
@@ -0,0 +1,54 @@
...
+                self.wait_task = None
+                break
+
+    def stop_immediately(self) -> None:
+        if self.wait_task is not None:
+            self.wait_task.close()
+        self.wait_task = None
+
+    async def _wait(self, counter: int = 0) -> None:
+        timeout_ms = round(10200 - 1010000 / (counter + 100))
```
mmilata, line 50: What does this mean?

M1nd3r:
Replaced by https://github.com/trezor/trezor-firmware/pull/4976

PR #4056, opened 2024-07-23
docs(core): fix 2.8.0 changelog

mmilata:
no optiga on T2T1


Hannsek:
What about the `pin processing`? Wasn't this reworked because of optiga and therefore it didn't apply for T2T1?

mmilata:
I don't know what change this refers to but it's possible Model T code paths were also affected.

Hannsek:
@matejcik can you tell?

Also I don't see anywhere the change in Solana. Didn't we forget about that? @matejcik

PR #4060, opened 2024-07-25
prefer vertical swipes over horizontal

TychoVrahe:
Adjusting the swipe detection in order to prefer vertical swipes over horizontal, as vertical swipes are more common operation vs. special cases like entering menu. This aims to improve reliability of swipes when swiping somewhat diagonally.

### DIFF
--- a/core/embed/rust/src/ui/component/swipe_detect.rs
+++ b/core/embed/rust/src/ui/component/swipe_detect.rs
@@ -162,8 +162,10 @@ impl SwipeDetect {
     const TRIGGER_THRESHOLD: f32 = 0.3;
     const DETECT_THRESHOLD: f32 = 0.1;
 
-    const MIN_LOCK: u16 = (Self::DISTANCE as f32 * Self::DETECT_THRESHOLD) as u16;
-    const MIN_TRIGGER: u16 = (Self::DISTANCE as f32 * Self::TRIGGER_THRESHOLD) as u16;
+    const VERTICAL_PREFERENCE: f32 = 2.0;
+
+    const MIN_LOCK: f32 = Self::DISTANCE as f32 * Self::DETECT_THRESHOLD;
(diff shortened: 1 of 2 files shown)


core/embed/rust/src/ui/component/swipe_detect.rs
```diff
@@ -174,20 +176,30 @@ impl SwipeDetect {
         }
     }
 
-    const fn min_lock(&self) -> u16 {
-        Self::MIN_LOCK
+    const fn min_lock(&self, dir: SwipeDirection) -> u16 {
+        match dir {
+            SwipeDirection::Up | SwipeDirection::Down => Self::MIN_LOCK,
+            SwipeDirection::Left | SwipeDirection::Right => {
+                Self::MIN_LOCK * Self::VERTICAL_PREFERENCE as u16
```
prusnak, line 183: What is the precedence of operations here?

Is

``` rust
Self::MIN_LOCK * Self::VERTICAL_PREFERENCE as u16
```

the same as

``` rust
(Self::MIN_LOCK * Self::VERTICAL_PREFERENCE) as u16
```

TychoVrahe: the intention was to multiply floats to allow fractional preference setting if we need to tune this further. 

fixed here: https://github.com/trezor/trezor-firmware/pull/4060/commits/3c2ac84aebd094525372f20a3a6f6f087d3b9bd3

mmilata (approved):

mmilata:
style check failing

bosomt:
**QA OK**

i see improvement in my vertical/horizontal swipes ratio 


**Info:**
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.1 regular (revision 33f5d4733065ac631d61ff458a5db32c81641124)

PR #4062, opened 2024-07-26
chore(core): rename "hidden wallet" -> "passphrase wallet"

### LINKED ISSUE #4057: Rename "Hidden wallet" to "Passphrase wallet" everywhere

ibz:
Fixes #4057


obrusvit (commented):
Unfortunately it does not fit on some screens of TS3 :crying_cat_face: 
image

How about model One? If I checked correctly, there is one string containing "hidden wallet" in `legacy/firmware/config.c`, line `602`. It's for sure not a reason to make new FW version but maybe we can make the change now so that it's not forgotten in the future?

Hannsek:
Don't we have a marquee in the header here<

Hannsek:
And also not a big problem as almost no one will ever see this screen.

ibz:
> Don't we have a marquee in the header here<
> 
> Regarding T1. If it it a long hanging fruit (10 mins max), then let's do it, otherwise let's not do it.

I can confirm that the header is a marquee and the text is readable. I double checked.

I tried updating the string in Model One and checking how it looks in the emulator and in that case the text becomes unreadable - due to the same issue: it being in the header and too long - but in that case it is *not* a marquee...

image

Hannsek:
Let's not do it then.

core/translations/es.json
obrusvit, line 1: Please check the UI for spanish version.

core/translations/fr.json
obrusvit, line 1: Please check the UI for french version.

obrusvit (commented):
Please check the UI in other langs and post screenshots.

core/translations/es.json
ibz: image

It *almost* fits...

core/translations/fr.json
ibz: It fits!

image

obrusvit (approved):

core/translations/fr.json
prusnak: but the inverted question mark is a bug, no?

Hannsek: Yeah, there should be `'`.

obrusvit: We probably miss the `'` in here https://github.com/trezor/trezor-firmware/blob/33f5d4733065ac631d61ff458a5db32c81641124/core/tools/codegen/foreign_chars.py#L36 
That's why it fallbacks to non-printable char :question:

prusnak: maybe just change the translation to use normal ASCII `'`?

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor **T3T1** 2.8.1 regular (revision 33f5d4733065ac631d61ff458a5db32c81641124)
 - Device: Trezor **T2B1** 2.8.1 regular (revision 33f5d4733065ac631d61ff458a5db32c81641124)

PR #4064, opened 2024-07-29
add more thorough touch testing to prodtest

TychoVrahe:
adding several new commands to prodtest.

TOUCH_CUSTOM - to show arbitrary rectangle and report touches
TOUCH_IDLE - to test that there are no touches
(using underscore to avoid conflicts due to startswith usage)

`BOOTLOADER VERSION` and `BOARDLOADER VERSION` with obvious purpose.

### DIFF
--- a/core/embed/prodtest/README.md
+++ b/core/embed/prodtest/README.md
@@ -282,6 +282,26 @@ FIRMWARE VERSION
 OK 0.2.6
 ```
 
+### BOOTLOADER VERSION
+Returns the version of the bootlaoder.
+The command returns `OK` followed by the version in the format `<major>.<minor>.<patch>`.
+
+Example:
+```
+BOOTLOADER_VERSION
+OK 2.1.7
+```
+
+### BOARDLOADER VERSION
+Returns the version of the boardloader.
+The command returns `OK` followed by the version in the format `<major>.<minor>.<patch>`.
+
+Example:
+```
+FIRMWARE VERSION
+OK 0.2.6
+```
+
 ### WIPE
 This command invalidates the current firmware in the flash memory by erasing its beginning, including metadata.
 After performing this operation, it displays the text "WIPED" on the screen and returns the response OK.
--- a/core/embed/prodtest/README.md
+++ b/core/embed/prodtest/README.md
@@ -88,7 +88,7 @@ OK
 ```
 
 ### TOUCH
-The `TOUCH` command test the functionality of the display's touch screen.
+The `TOUCH` command tests the functionality of the display's touch screen.
 It draws a filled rectangle in one of the four display quadrants and waits for user interaction.
 
 The command requires two input parameters:
@@ -104,6 +104,48 @@ Example (to draw a rectangle in the top-left quadrant and wait for 9 seconds for
 ```
 TOUCH 09
 OK 50 90
+```
+
+
+### TOUCH_CUSTOM
+The `TOUCH_CUSTOM` command tests the functionality of the display's touch screen.
+It draws a filled rectangle on custom coordinates and waits for user interaction.
+
+The command requires five input parameters:
+
+* X position of the top-left corner of the rectangle
+* Y position of the top-left corner of the rectangle
+* Width of the rectangle
+* Height of the rectangle
+* The timeout in seconds
+
+If the display is not touched within the specified timeout, the command will return an `ERROR TIMEOUT`.
+
+The device report touch events, coordinates and timestamps (in ms), correctness of the touch point is not checked and is left to the test equipment.
+
+The test ends with first lift-up event.
+
+Example (to draw a 100x100 rectangle in the top-left (10,10) position and wait for 15 seconds for touch input):
+```
+TOUCH_CUSTOM 10 10 100 100 15
+TOUCH D 69 35 357300
+TOUCH U 69 35 357328
+OK
+```
+
+### TOUCH_IDLE
+The `TOUCH_IDLE` command tests the functionality of the display's touch screen.
+It waits for a specific time period without any touch input.
+
+The command requires one input parameter:
+* The timeout in seconds
+
+If a touch activity is detected within the specified timeout, the command will return an `ERROR TOUCH DETECTED`.
+
+Example - wait ten seconds for no touch input:
+```
+TOUCH_IDLE 10
+OK
 ```
 
 ### SENS
--- a/core/embed/prodtest/main.c
+++ b/core/embed/prodtest/main.c
@@ -24,6 +24,7 @@
 
 #include STM32_HAL_H
 
+#include "board_capabilities.h"
 #include "button.h"
 #include "common.h"
 #include "display.h"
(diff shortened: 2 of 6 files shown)


core/embed/prodtest/README.md
```diff
@@ -240,6 +282,26 @@ FIRMWARE VERSION
...
+The command returns `OK` followed by the version in the format `<major>.<minor>.<patch>`.
+
+Example:
+```
+BOOTLOADER_VERSION
+OK 2.1.7
+```
+
+### FIRMWARE VERSION
+Returns the version of the boardloader.
```
cepetr, line 296: The command name is `BOARDLOADER VERSION`

core/embed/prodtest/main.c
```diff
@@ -364,6 +393,124 @@ static void test_touch(const char *args) {
   touch_deinit();
 }
 
+static void test_touch_custom(const char *args) {
+#define NUM_PARAMS 5
```
cepetr, line 397: How about using `static const int num_params = 5;` instead of #define/#undef? If think it could work better here.

core/embed/prodtest/main.c
```diff
@@ -364,6 +393,124 @@ static void test_touch(const char *args) {
...
+
+  int params[NUM_PARAMS] = {0};
+  int num_params = 0;
+
+  extract_params(args, params, &num_params, NUM_PARAMS);
```
cepetr, line 402: Maybe it would be easier and more readable to use `sscanf` here instead of own `extract_params`.

core/embed/prodtest/README.md
```diff
@@ -106,6 +106,48 @@ TOUCH 09
...
+```
+
+### TOUCH_IDLE
+The `TOUCH_IDLE` command tests the functionality of the display's touch screen.
+It waits for a specific time period without any touch input.
+
+The command requires one input parameter:
+* The timeout in seconds
+
+If a touch event is detected within the specified timeout, the command will return an `ERROR TIMEOUT`, the touch event, its coordinates and timestamp (in ms) are reported.
```
cepetr, line 143: The behavior is different from what is described here.

core/embed/prodtest/README.md
```diff
@@ -106,6 +106,48 @@ TOUCH 09
...
+Example (to draw a 100x100 rectangle in the top-left (10,10) position and wait for 15 seconds for touch input):
+```
+TOUCH_CUSTOM 10 10 100 100 15
+TOUCH D 69 35 357300
+TOUCH U 69 35 357328
+OK
+```
+
+### TOUCH_IDLE
+The `TOUCH_IDLE` command tests the functionality of the display's touch screen.
```
cepetr, line 137: To be consistent with other commands, I would rather changed the command name to `TOUCH IDLE` (without the underscore). The same applies to `TOUCH_CUSTOM`.

core/embed/prodtest/README.md
```diff
@@ -240,6 +282,26 @@ FIRMWARE VERSION
 OK 0.2.6
 ```
 
+### BOOTLOADER_VERSION
```
cepetr, line 285: The command name is `BOOTLOADER VERSION` (without the undescore).

core/embed/prodtest/main.c
```diff
@@ -703,6 +879,12 @@ int main(void) {
     } else if (startswith(line, "TOUCH ")) {
       test_touch(line + 6);
 
+    } else if (startswith(line, "TOUCH_CUSTOM ")) {
+      test_touch_custom(line + 13);
+
```
cepetr, line 884: All other two-word commands have no underscore between the words.

core/embed/prodtest/main.c
```diff
@@ -761,7 +943,10 @@ int main(void) {
 
     } else if (startswith(line, "FIRMWARE VERSION")) {
       test_firmware_version();
-
+    } else if (startswith(line, "BOOTLOADER VERSION")) {
+      test_bootloader_version(bootloader_version);
```
cepetr, line 947: It would be better to be consistent with the other `test_` functions and have `test_bootloader_version` and `test_boardloader_version` without any arguments

core/embed/trezorhal/stm32f4/touch/ft6x36.c
```diff
@@ -33,6 +33,12 @@
 #include "panels/lx154a2422cpt23.h"
 #endif
 
+#ifdef TOUCH_DETECT_ALL
+#define TOUCH_MAX_POINTS 2
```
cepetr, line 37: What's the reason for detecting all touch points in the prodtest?

cepetr (changes requested):

core/embed/prodtest/README.md
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/aed6c9e3ff877a30cc011fd248840cd056e11cd5

core/embed/prodtest/README.md
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/b7964bb62095997f462700d08ca24db7143c35af

core/embed/prodtest/README.md
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/b7964bb62095997f462700d08ca24db7143c35af

core/embed/prodtest/main.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/f6fb2a667cbb9ad5b9a58408813623ce9cd61dee

(used `expected_params` as `num_params` is already used for real number of provided params)

core/embed/prodtest/main.c
TychoVrahe: i agree that could be an improvement from readability and maintainability POV, but it doesn't compile with our current std lib due to `undefined reference to _sbrk` and don't want to try figuring out what needs to be changed and implications of it right now

core/embed/prodtest/README.md
TychoVrahe: as mentioned in the PR description i went for the underscore to avoid conflicts with existing commands:
if you use `TOUCH CUSTOM `, it will also be recognized by` if(startswith("TOUCH ") `branch.

I could place the `TOUCH_CUSTOM` above the `TOUCH`, but still, older prodtest versions will recognize this new command as regular `TOUCH`. Using underscore prevents this.

If anything, i would go for using underscore everywhere, with space reserved as parameters delimiter. (or 0 delimiter for commands without params). This way we could avoid such conflicts systematically. But this would break compatibility with older test programs, so i am nor ready to do that yet.

core/embed/prodtest/main.c
TychoVrahe: see above

core/embed/prodtest/main.c
TychoVrahe: the problem here is that due to MPU configuration bootloader header is unreadable at the time this command is called, so i need to parse the header before the MPU is configured and store it in the variable. I could avoid the parameter by using global variable but i don't think its necessarily better. 

Boardloader version test would work without param, but at this at this point the consistency is lost anyway i don't think it matters much

core/embed/trezorhal/stm32f4/touch/ft6x36.c
TychoVrahe: For detection of "ghost" touches, which are quite often two point. Just slightly increasing probability of these getting detected during prodtest.

core/embed/prodtest/README.md
cepetr: Hmm, I understand and agree. I don't like how it's done now either. We definitely should define this protocol/language precisely to avoid ambiguity in the future.

core/embed/trezorhal/stm32f4/touch/ft6x36.c
cepetr: Then, maybe we could just use `touch_activity()` to detect unexpected behavior. With `touch_activity()` you can't report any additional information about the event, but it's not so important in the case. No sure, but I think it's better than adding such a strange configuration option to the touch driver.

TychoVrahe: This works for `TOUCH_IDLE` test, but not for `TOUCH_CUSTOM` nor `TOUCH`, where we can also catch this via likely wrong coordinates in the ghost touch with two points. 

But i like it anyway, we will detect two points ghosts in TOUCH_IDLE and one point ghosts elsewhere, which seems acceptable https://github.com/trezor/trezor-firmware/pull/4064/commits/659dc9a74d12839070548f8b420862a9e382dcb5

cepetr (approved):
utACK

PR #4066, opened 2024-07-29
refactor(core/mercury): init Paragraphs from Vec

obrusvit:
This minor commit saves `1 KiB` of debug flash. I ain't much but it's for free.


mmilata (approved):
lgtm, thanks!

PR #4067, opened 2024-07-29
Remove bias in ed25519.generate_secret()

andrewkozlik:
The function `generate_secret()` from the `ed25519` Python module generates a random 256-bit key and then "clamps" it, i.e. sets the three least significant bits to 0 and the two most significant bits to 0 and 1. I believe that the clamping should not be applied here.

The functions `ed25519_publickey()`, `ed25519_sign()` and `ed25519_cosi_sign()` first hash the secret key and then apply the clamping on the first 256 bits of the hash. In other words, the rest of the module implementation assumes that it is working with an unclamped secret key. This means that clamping does not need to be applied during key generation and it technically violates the rules for generating Ed25519 keys:
> The private key is 32 octets (256 bits, corresponding to b) of cryptographically secure random data.

https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5

In my opinion it is confusing to apply clamping in `generate_secret()` because it gives the false impression that the module somehow works with the expanded private key. It also introduces a bias into the generated keys, which is, however, cryptographically insignificant.

There is one exception in the module. The function `ed25519_sign_ext()`, which is used only in Cardano, takes as input the expanded secret key as two parameters (`secret_key: bytes` and `secret_extension: bytes`). Since the secret key is already expanded, the function doesn't even clamp it. This might be intentional, because of the way Cardano uses BIP-32 derivation. This might be worth looking into and possibly adding some comment or adding the clamping operation to be on the safe side.

Note that `ed25519.generate_secret()` isn't actually used anywhere, so this PR does not change the firmware behavior. I assume that the function is included for completeness of the module.

History: The function was added in 9c921c073f.

### DIFF
--- a/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-ed25519.h
+++ b/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-ed25519.h
@@ -100,16 +100,16 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mod_trezorcrypto_ed25519_sign_obj, 2,
 #if !BITCOIN_ONLY
 
 /// def sign_ext(
-///     secret_key: bytes, secret_extension: bytes, message: bytes
+///     secret_scalar: bytes, secret_extension: bytes, message: bytes
 /// ) -> bytes:
 ///     """
-///     Uses secret key to produce the cardano signature of message.
+///     Uses extended secret key to produce the cardano signature of message.
 ///     """
-STATIC mp_obj_t mod_trezorcrypto_ed25519_sign_ext(mp_obj_t secret_key,
+STATIC mp_obj_t mod_trezorcrypto_ed25519_sign_ext(mp_obj_t secret_scalar,
                                                   mp_obj_t secret_extension,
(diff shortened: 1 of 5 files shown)


andrewkozlik:
> This might be worth looking into and possibly adding some comment or adding the clamping operation to be on the safe side.

I tried to improve the clarity by renaming the `secret_key` parameter of `ed25519_sign_ext()` to `secret_scalar` in https://github.com/trezor/trezor-firmware/pull/4067/commits/cd3abbb686f1e60533829180ad219194e6753930, so that it's distinguished from the unexpanded `secret_key`.

As for clamping in `ed25519_sign_ext()`, I think we can add it, but we don't have to add it.
- We can add it, because the Cardano BIP-32 derivation essentially ensures that the `secret_scalar` value remains clamped by clearing the third most significant bit during master node derivation and by adding only 28-byte values multiplied by eight to the secret scalar in subsequent node derivations.
- We don't have to add it, because the function doesn't seem to rely on the `secret_scalar` being clamped, and I think the security isn't affected either.

onvej-sl (approved):

PR #4068, opened 2024-07-29
Mercury PIN keyboard exit flow

obrusvit:
**Edit: this behavior will be used only in initial PIN setup, not generally. So this PR is postponed for now.**

This PR replaces `request_pin` with `flow_request_pin`. The added benefit is that hitting 'x' button does not dismiss the
keyboard immediately but rather opens a VerticalMenu with "Cancel" option. Avoiding the abrupt closing of the PIN keyboard should improve the UX somewhat.

In order to do that, this PR also introduces `FlowMsg::Text(ShortString)` variant which enables
flows to send out strings to micropython.

Takes `2 KiB` of flash.

TODO:
- [ ] modify input_flows for cancel PIN


obrusvit:
Obsolete.

PR #4073, opened 2024-07-31
Update clang-format style

cepetr:
This PR proposes updates to the clang-format style to address certain preferences that some of us have regarding the default Google style:

Finally I decided to open this topic, even though I realize it may not be very productive. I have proposed just a few changes that I believe can bring some improvements: **Please let me know if you like it or not.**

**1) Increase indentation from 2 to 4 spaces**
This change aims to improve code readability by making it easier to locate the closing bracket of a block.

**2) Increase line limit from 80 to 100 columns**
An 80-column limit can be restrictive and cause unnecessary line breaks. Modern displays are wide enough to accommodate a 100-column limit, making this old-school rule less relevant.

**3) Modify aligning of list arguments**
The original settings often result in a lot of code on the right side of the screen, leaving blank areas on the left and making the alignment heavily dependent on the length of function names. The new settings are more compact and natural, making the code easier to write and read.

```c
uint32_t ui_screen_install_confirm(const vendor_header *const vhdr,
                                   const image_header *const hdr,
                                   secbool should_keep_seed,
                                   secbool is_newvendor, int version_cmp) {
  uint8_t fingerprint[32];
  char ver_str[64];
  get_image_fingerprint(hdr, fingerprint);
  format_ver("%d.%d.%d", hdr->version, ver_str, sizeof(ver_str));
  return screen_install_confirm(vhdr->vstr, vhdr->vstr_len, ver_str,
                                fingerprint, should_keep_seed == sectrue,
                                is_newvendor == sectrue, version_cmp);
}

-->

uint32_t ui_screen_install_confirm(
    const vendor_header *const vhdr, const image_header *const hdr, secbool should_keep_seed,
    secbool is_newvendor, int version_cmp)
{
    uint8_t fingerprint[32];
    char ver_str[64];
    get_image_fingerprint(hdr, fingerprint);
    format_ver("%d.%d.%d", hdr->version, ver_str, sizeof(ver_str));
    return screen_install_confirm(
        vhdr->vstr, vhdr->vstr_len, ver_str, fingerprint, should_keep_seed == sectrue,
        is_newvendor == sectrue, version_cmp);
}
```

```c
if (sectrue != check_image_header_sig(current_hdr, current_vhdr->vsig_m,
                                      current_vhdr->vsig_n,
                                      current_vhdr->vpub)) {
  *is_new = sectrue;
  return;
}
  
  -->
  
if (sectrue !=
    check_image_header_sig(
        current_hdr, current_vhdr->vsig_m, current_vhdr->vsig_n, current_vhdr->vpub)) {
    *is_new = sectrue;
    return;
}  
```

**4) Add break before open braces in function definition (Linux style)**
This setting helps to visually separate the function definition from the function body.

```c
static inline gfx_clip_t gfx_clip(gfx_rect_t dst, const gfx_bitmap_t* bitmap) {
  int16_t dst_x = dst.x0;
  int16_t dst_y = dst.y0;

  int16_t src_x = 0;
  int16_t src_y = 0;

->

static inline gfx_clip_t gfx_clip(gfx_rect_t dst, const gfx_bitmap_t* bitmap) 
{
    int16_t dst_x = dst.x0;
    int16_t dst_y = dst.y0;

    int16_t src_x = 0;
    int16_t src_y = 0;
```

**5) Disabled short functions on single line**

```c
void hal_delay(uint32_t ms) { usleep(1000 * ms); }

->

void hal_delay(uint32_t ms)
{
    usleep(1000 * ms);
}
```


TychoVrahe:
Regarding the proposed changes:

1) i don't have strong preference 2vs4 spaces, i am personally ok with both
2) i would prefer wider settings (i would even consider 120)
3-4) the proposed change seems somewhat preferable to me, although my personal preference would be to have every argument on separate line
5) this seems highly preferable.

In this regard, i would be ok with changing the style.

However, my worries here are mainly side-effects:
-- This breaks comparison throughout history. Some comparison tools are maybe capable of seeing through these kind of changes, but i don't think this is still a norm and generally i would worry about reliability of such functionality.
-- It will cause conflicts with any open PRs that touch C code. 

Given this, and with regard to the fact that most formatting is matter of personal preference or habits, i would be inclined to be conservative with this.

prusnak:
I don't touch the code these days too much, so I do not want to be an authority in this decision, but I think that changing the style is usually not a good idea because it makes diffing to old code hard.

If one prefers to have another style, they can apply it locally using clang-format and then reapply the original one just before commit.

Subjective takes on proposed points:

1) maybe just use a tab and decide whether it is 2, 4, 8 spaces wide? also if you do not do this, you do not have to do 2)
2) this is about being able to diff side-by-side - can your display handle 200 columns?
3) this is to make diffing easier if you add more arguments to the function
4) good idea
5) good idea

cepetr:
I have to admit this PR is more about my personal preference that rational reasons. That's just the way it is:-) I hoped to find allies in this.

**Side-effect of loosing diffs**

Yes, it's very undesirable and the strongest argument against it, but it's inevitable if we want to "improve." I understand it's a big price to pay, and if the improvements aren't worth it, this could eventually become a blocker.

**Conflicts with op PRs**

Yes, that's a problem we'll have to tackle, but only for a limited time period.
It can be routinely solved by formatting the branch before rebasing onto the main with the new style:

* Cherry-pick the commit with the clang-format change.
* Use interactive rebase to:
* * Move the cherry-picked commit to the beginning of the branch.
* * Use rebase edit and run make style on every commit in the branch.
* Rebase the branch.

**Tabs vs spaces**

If believe that using tabs is not an option when an inherent part of the formatting style involves
aligning code to columns, like this:

```c
  #define AAA     // Comment
  #define AAAA    // Comment
  #define AAAAA   // Comment
```

**Increasing line limit**
I use side-by-side comparison only occasionally, so I personally don't encounter this issue often. 
However, I agree that using 2x100 columns can lead to the need for horizontal scrolling during comparison. I can live comfortably in 80 columns limit.

**Indentation**
I consider four spaces are the industry standard and common practice. We use it in Rust and Python code as well. 
I consider it to be much better for separating different levels of nesting and it plays much better with #3, but I can live with two spaces as well.

---

For me personally, the most importand style change is also #3. 

I would also prefer having each argument on its own line if they don't fit on a single line. However, this will require trickier clang-format settings, if it’s even possible. But I'll try to find it.


Change #4 is just a supplement to #3 to avoid a code like this:

```c
  void function(
      int argument1, int argument2, int argument3) {
      int variable1;
      int variable2;
    
      ....
    }
```

Change #5 is a minor detail, but I included it because it's consistent with #4.

cepetr:
Finally, I found the `clang-format` settings that ensure function declaration and calling arguments are placed on separate lines if they do not fit on a single line. The following settings achieve points 3), 4), and 5), while I have omitted points 1) and 2).

The number and nature of the changes in this case are quite reasonable. Blame/diff will not be so problematic.


```ini
---
BasedOnStyle: Google

# Increase indentation from 2 to 4 characters
#IndentWidth: 4
#ContinuationIndentWidth: 4
# Increase column limit from 80 to 100 characters
#ColumnLimit: 100
# Sort each #include blocks separately
IncludeBlocks: Preserve
# Break before braces on function, namespace and class definition
BreakBeforeBraces: Linux
# Always break after an open bracket, if the parameters don’t fit on a single line
AlignAfterOpenBracket: AlwaysBreak
# Disable short functions on single line
AllowShortFunctionsOnASingleLine: None
# If false, a function declaration’s or function definition’s parameters will
# either all be on the same line or will have one line each.
BinPackParameters: false
# If false, a function call’s arguments will either be all on the
# same line or will have one line each.
BinPackArguments: false
# This prevents all parameters from being placed on the next line, unless necessary.
AllowAllParametersOfDeclarationOnNextLine: false
```

andrewkozlik:
I just found out that you can put a list of reformatting commits into a file like `.gitblameignore` and then run `git config blame.ignoreRevsFile .gitblameignore`. Then `git blame` will skip over the commits. I don't know how clever it is though.

cepetr:
Adding `PenaltyIndentedWhitespace: 1000` reduces alignment overall, but I'm not sure if it's an improvement:

```c
  if (sectrue !=
      check_image_header_sig(
          current_hdr,
          current_vhdr->vsig_m,
          current_vhdr->vsig_n,
          current_vhdr->vpub)) {
    *is_new = sectrue;
    return;
  }
  ```

**Comment by onvej-sl** [comment]
Just to add my two cents to the discussion:
  1. I'm slightly against it.
  2. I'm strongly for it. I would even support 120 colums.
  3. I'm for it.
  4. I have no opinion.
  5. I'm strongly for it.


  > If one prefers to have another style, they can apply it locally using clang-format and then reapply the original one just before commit.

I've just tried it myself, and it's not as straightforward as it may seem.
```
$ make cstyle
$ echo "BasedOnStyle: Mozilla" > .clang-format
$ make cstyle
$ git checkout -- .clang-format
$ make cstyle
$ git diff --shortstat
353 files changed, 7696 insertions(+), 6524 deletions(-)
```

**Comment by mmilata** [comment]
heard you're 2c short so here are mine:

1. strongly against, benefits do not outweigh breaking git blame for everything IMO & there are other ways to match the braces (lines, colors)
2. weakly in favor
3. weakly against, as in i don't care but would rather not introduce noise into history
4. weakly against, as in i don't care but would rather not introduce noise into history
5. weakly in favor

**Comment by TychoVrahe** [comment]
At this point, its clear that we are not reaching consensus about the proposed changes so we will keep status quo.

However, we might still be able to reach consensus for smaller and more limited scope:

**1.) Modify list of arguments so that each argument is on separate line**
- limited impact, easier comparison

**2.) Disabled short functions on single line (original point 5)**
- limited diff impact as these are one liners

And we would introduce the `gitblameignore`.

@cepetr could you please update this PR so we can see the resulting changes and discuss whether this would be OK for all? No rush, whenever you feel like it.

**Comment by prusnak** [comment]
> However, we might still be able to reach consensus for smaller and more limited scope:

I am OK with both changes ("each argument is on separate line" and "disable short functions")

**Comment by onvej-sl** [comment]
@cepetr , If this pull request will be ever merged, could you also include the following change in the same commit? We decided to apply formatting to some files that were unformated but we want to keep the number of reformatting commits as low as possible (see https://github.com/trezor/trezor-firmware/pull/4107#issuecomment-2312385472).
```diff
diff --git a/tools/style.c.exclude b/tools/style.c.exclude
index 5ef29b507..adeb5ca60 100644
--- a/tools/style.c.exclude
+++ b/tools/style.c.exclude
@@ -3,9 +3,7 @@
 ^\./crypto/chacha20poly1305/
 ^\./crypto/ed25519-donna/
 ^\./crypto/gui/
-^\./crypto/blake2
 ^\./crypto/check_mem
-^\./crypto/groestl
 ^\./crypto/ripemd160
 ^\./crypto/segwit_addr
 ^\./crypto/sha2
```

cepetr:
> @cepetr could you please update this PR so we can see the resulting changes and discuss whether this would be OK for all? No rush, whenever you feel like it.

Unfortunatelly I couldn't find a clang-format configuration to consistently put each parameter on a separate line when the function arguments can fit on a single line. This leads to inconsistent formatting, with some functions having arguments on one line and others spread across multiple lines, which looks messy, especially in .h files. Clang-format just isn't as flexible as it needs to be for this.

```c
const void *flash_get_address(uint8_t sector, uint32_t offset, uint32_t size);

secbool __wur flash_erase_sectors(
    const uint8_t *sectors, 
    int len, 
    void (*progress)(int pos, int len));

secbool __wur flash_write_byte(uint8_t sector, uint32_t offset, uint8_t data);
secbool __wur flash_write_word(uint8_t sector, uint32_t offset, uint32_t data);
```

cepetr:
After discussing this matter with @TychoVrahe today, I’ve decided to close it. Unfortunately, it seems there isn’t a viable way to strike a balance between what we want, the value of the change, and the limitations of clang-format.

I truly appreciate everyone’s contributions and insights on this topic. Thank you.

Issue #4058, opened 2024-07-24
Reduce size of precomputed tables for ECC


onvej-sl:
# Introduction
## Curves
There are two elliptic curves used in firmware:
  * secp256k1 (the Bitcoin curve)
  * nist256p1 (also known as secp256r1)

We also use curve25519 and ed25519 but I won't address them in this issue.

## Operations
There are five elliptic curve operations exposed to the core firmware.
  * signature genration (`sign`)
  * signature verification (`verify`)
  * recovery of public key from signature (`verify_recover`)
  * derivation of public key from private key (`publickey`)
  * Diffie-Hellman operation (`multiply`)

## Implementations
There are two implementations of elliptic curve operations used by the firmware:
  * native (`tc_*` functions in `crypto/ecdsa.c`)
  * library (`vendor/secp256k1-zkp`)

The library implementation supports only `secp256k1`, whereas the native implementation supports both curves. I noticed that `multiply` is implemented always using the native implementation, even though the library implementation supports this operation.

## Precomputed tables

The native implementation can use a precomputed table to speed up elliptic curve operations. A precomputed table for every curve takes 36 KiB.

The library implementation uses two precomputed tables. The first one is 32 KiB in size and cannot be reduced. The second one is 8 KiB in size and can be reduced to 128 B at the cost of slower `verify` and `verify_recover`.  See [this](https://github.com/trezor/trezor-firmware/issues/1534#issuecomment-817688585) for more detailed information.

The table below shows the duration of the operations.

|                | native (36KiB table) | native (no table) | library (8KiB table) | library (128B table)  |
|----------------|----------------------|-------------------|----------------------|-----------------------|
| sign           | 27 ms                | 103 ms            | 17 ms                | 17 ms                 |
| verify         | 134 ms               | 210 ms            | 8 ms                 | 11 ms                 |
| verify_recover | 134 ms               | 210 ms            | 8 ms                 | 11 ms                 |
| publickey      | 23 ms                | 99 ms             | 16 ms                | 17 ms                 |

## Usage
As far as I know, [bip32 child key derivation](https://github.com/trezor/trezor-firmware/blob/main/crypto/bip32.c#L297) is the last place in firmware where the native implementation is used for `secp256k1`.
[...]

PR #4083, opened 2024-08-02
ECC memory optimization

onvej-sl:
Solves https://github.com/trezor/trezor-firmware/issues/4058.


onvej-sl:
Flash size reduced by approximately 88 KB. :rocket:

prusnak:
I suggest we rework commit 6de25d06cd7d19b42db8660d8f7c6f4ffa957976

It performs computation of pubkey for each child derivation while the old code did it only for the last node saving lots of cpu cycles and time.

prusnak:
> I suggest we rework commit 6de25d06cd7d19b42db8660d8f7c6f4ffa957976
> 
> It performs computation of pubkey for each child derivation while the old code did it only for the last node saving lots of cpu cycles and time.
> 
> Since we are doing private CKD we do not need public key of the nodes we traverse, only the last one.

Sorry, I am wrong on this one. The commit only touches public CKD, not private. I guess this is fine then.

legacy/firmware/Makefile
```diff
@@ -19,11 +19,12 @@ ZKP_CFLAGS = \
 	-DUSE_EXTERNAL_ASM \
 	-DUSE_EXTERNAL_DEFAULT_CALLBACKS \
 	-DECMULT_GEN_PREC_BITS=4 \
-	-DECMULT_WINDOW_SIZE=8 \
+	-DECMULT_WINDOW_SIZE=2 \
```
andrewkozlik, line 22: Why not leave the window size at 8 in the legacy build? We have no issues with lack of flash memory on T1, and we probably never will, because there are no major new features being added to T1 anymore.

We can also set `USE_PRECOMPUTED_CP=1` for the legacy build.

crypto/tests/test_check.c
```diff
@@ -4115,6 +4115,48 @@ START_TEST(test_aes) {
...
+  static struct {
+    const char *priv_key;
+    const char *pub_key;
+    const char *session_key;
+  } tests[] = {
+      {"1618cc490a4a5b38d9f877759a2c312026fe9f459edb9ba49e91b6de4a237cad",
+       "0322c16706cd0d80dbb4e314799ba1323420e7ffa859a0cf0c82a444192bb6c997",
+       "047b27ebb15e8197c9a560afc1bd45e2a78b864829fc1257333a1a2d7d30d79eed17889"
+       "8ad4960f756cf155cf46d22f2e3b21df10ee5bad560c467ae0d79427a70"},
+  };
```
andrewkozlik, line 4130: It would be good to have a test case with an uncompressed public key too. Maybe also some invalid cases.

crypto/bip32.c
```diff
@@ -294,8 +294,12 @@ int hdnode_public_ckd_cp(const ecdsa_curve *curve, const curve_point *parent,
     hmac_sha512(parent_chain_code, 32, data, sizeof(data), I);
     bn_read_be(I, &c);
     if (bn_is_less(&c, &curve->order)) {  // < order
-      scalar_multiply(curve, &c, child);  // b = c * G
-      point_add(curve, parent, child);    // b = a + b
+      // b = c * G
+      uint8_t child_pubkey[65] = {};
+      ecdsa_get_public_key65(curve, I, child_pubkey);
+      ecdsa_read_pubkey(curve, child_pubkey, child);
```
andrewkozlik, line 300: This is a valid solution, but:
- nit: Return codes are ignored. I realize that the functions can't really fail unless `I` is all-zero, which is near impossible, although AFAICS `scalar_multiply()` was sure to succeed always.
- I think it would be better to call `scalar_multiply()` here and use these two lines of code inside of `scalar_multiply()` for secp256k1 when `USE_SECP256K1_ZKP_ECDSA` is defined. Otherwise, sooner or later, we are likely to accidentally use `scalar_multiply()` somewhere in the future instead of this code, which we don't want, since it's not as computationally efficient.
- Another solution to consider is to use `secp256k1_ec_seckey_tweak_add()` for the entire process (scalar multiply and point addition). I assume that's what the function does, although the documentation isn't clear to me. However, I am not sure whether it has any significant advantages. Probably just some minor savings on data conversions. (My first though was that it would be nice if we could rely solely on one library for secp256k1 operations, so that if our library has a vulnerability, such as a side-channel, then it doesn't affect the bitcoin curve operations. However, in this particular case it doesn't make a difference, because the private key operations are done by the secp256-zkp library either way.)

crypto/tests/test_check.c
```diff
@@ -4115,6 +4115,88 @@ START_TEST(test_aes) {
...
+    const char *session_key;
+  } tests[] = {
+      // Compressed public key
+      {"1618cc490a4a5b38d9f877759a2c312026fe9f459edb9ba49e91b6de4a237cad",
+       "0322c16706cd0d80dbb4e314799ba1323420e7ffa859a0cf0c82a444192bb6c997", 0,
+       "047b27ebb15e8197c9a560afc1bd45e2a78b864829fc1257333a1a2d7d30d79eed17889"
+       "8ad4960f756cf155cf46d22f2e3b21df10ee5bad560c467ae0d79427a70"},
+      // Uncompressed public key
+      {"d56e61dea5c8412292c8a645f20846109a228532b475ff33b7c32663ddb66600",
+       "0282dc03a182e2613a9d2e3f2dc1bdd9e6b541708a719f610c06d1a09be62d3aa2", 0,
```
andrewkozlik, line 4134: This doesn't look like an uncompressed public key.

crypto/bip32.c
```diff
@@ -294,8 +294,12 @@ int hdnode_public_ckd_cp(const ecdsa_curve *curve, const curve_point *parent,
     hmac_sha512(parent_chain_code, 32, data, sizeof(data), I);
     bn_read_be(I, &c);
     if (bn_is_less(&c, &curve->order)) {  // < order
-      scalar_multiply(curve, &c, child);  // b = c * G
-      point_add(curve, parent, child);    // b = a + b
+      // b = c * G
+      uint8_t child_pubkey[65] = {};
```
andrewkozlik, line 298: nit: empty initializer list is not valid until C23, while our build scripts claim to be using gnu11 and gnu99, although gcc seems to accept it without complaints

crypto/bip32.c
```diff
@@ -275,95 +275,44 @@ int hdnode_private_ckd(HDNode *inout, uint32_t i) {
...
-    case 1:  // Segwit-in-P2SH
-      ecdsa_get_address_segwit_p2sh(child_pubkey, version, hasher_pubkey,
-                                    hasher_base58, addr, addrsize);
-      break;
-    default:  // normal address
-      ecdsa_get_address(child_pubkey, version, hasher_pubkey, hasher_base58,
-                        addr, addrsize);
-      break;
-  }
+  return !result;
```
andrewkozlik, line 315: This hides a lot of assumptions, namely that `result` 0 represents success.
```suggestion
  return result == ECDSA_TWEAK_PUBKEY_SUCCESS;
```

crypto/zkp_ecdsa.c
```diff
@@ -473,3 +474,69 @@ int zkp_ecdh_multiply(const ecdsa_curve *curve,
 
   return result;
 }
+
+// ECDSA public key tweak
+// tweaked_public_key = public_key + private_tweak * G
+// curve has to be &secp256k1
+// public_key_bytes has 33 bytes
+// private_key_tweak_bytes has 32 bytes
```
andrewkozlik, line 482: `private_key_tweak_bytes` is actually called `private_tweak_bytes` in the code below. However, I think we should just call it `tweak_bytes`, because there is nothing inherently private about it. In case of BIP32 maybe, because the chain code is not entirely public information, but there are other applications where it does not need to be secret at all. (The latter would imply making the same change in zkp_ecdsa.h, ecdsa.c, ecdsa.h.)

crypto/ecdsa.h
```diff
@@ -125,4 +125,17 @@ int ecdsa_recover_pub_from_sig(const ecdsa_curve *curve, uint8_t *pub_key,
...
 int ecdsa_sig_from_der(const uint8_t *der, size_t der_len, uint8_t sig[64]);
 
+typedef enum {
+  ECDSA_TWEAK_PUBKEY_SUCCESS = 0,
+  ECDSA_TWEAK_PUBKEY_INVALID_CURVE_ERR,
+  ECDSA_TWEAK_PUBKEY_CONTEXT_ERR,
+  ECDSA_TWEAK_PUBKEY_INVALID_PUBKEY_ERR,
+  ECDSA_TWEAK_PUBKEY_INVALID_TWEAK_OR_RESULT_ERR,
+  ECDSA_TWEAK_PUBKEY_UNSPECIFIED_ERR,
+} ecdsa_tweak_pubkey_result;
```
andrewkozlik, line 135: Consider renaming `ecdsa_tweak_pubkey_result` to `ecdsa_tweak_result`, since it would probably make sense to reuse the enum if and when we implement `ecdsa_tweak_privkey()`.

crypto/ecdsa.c
```diff
@@ -30,7 +30,6 @@
 #include "address.h"
 #include "base58.h"
 #include "bignum.h"
-#include "ecdsa.h"
```
andrewkozlik, line 33: If we remove ecdsa.h, does this still compile successfully without `USE_SECP256K1_ZKP_ECDSA`? In principle code shouldn't rely on another header file including another necessary header file.

andrewkozlik (commented):
Partially reviewed.

crypto/zkp_ecdsa.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4083/commits/3bf6559a80b875348a162aa998d3d213a59f731a.

crypto/ecdsa.h
```diff
@@ -125,4 +125,17 @@ int ecdsa_recover_pub_from_sig(const ecdsa_curve *curve, uint8_t *pub_key,
...
+
+ecdsa_tweak_pubkey_result ecdsa_tweak_pubkey(const ecdsa_curve *curve,
+                                             const uint8_t *pub_key,
+                                             const uint8_t *priv_tweak,
```
andrewkozlik, line 139:
```suggestion
                                             const uint8_t *tweak,
```

crypto/ecdsa.c
```diff
@@ -1312,3 +1354,25 @@ int ecdsa_recover_pub_from_sig(const ecdsa_curve *curve, uint8_t *pub_key,
...
+  if (curve == &secp256k1) {
+    return zkp_ecdh_multiply(curve, priv_key, pub_key, session_key);
+  }
+#endif
+  return tc_ecdh_multiply(curve, priv_key, pub_key, session_key);
+}
+
+ecdsa_tweak_pubkey_result ecdsa_tweak_pubkey(const ecdsa_curve *curve,
+                                             const uint8_t *pub_key,
+                                             const uint8_t *priv_tweak,
```
andrewkozlik, line 1370:
```suggestion
                                             const uint8_t *tweak,
```

crypto/ecdsa.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4083/commits/f6675fd5ecf272faa011997889fc9ad768ebe0d7.

crypto/ecdsa.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4083/commits/f6675fd5ecf272faa011997889fc9ad768ebe0d7.

crypto/ecdsa.c
onvej-sl: > If we remove ecdsa.h, does this still compile successfully without USE_SECP256K1_ZKP_ECDSA?

Yes, it does. I think it is is included indirectly by `#include "secp256k1.h"`.

To be honest, I don't remember why I removed the include. I'm reverting it in https://github.com/trezor/trezor-firmware/pull/4083/commits/6c3e77a795ebca58b8d974d55efbd09e05f121db.

crypto/ecdsa.h
onvej-sl: I would postpone it until we implement `ecdsa_tweak_privkey`.

andrewkozlik (approved):

PR #4086, opened 2024-08-05
fix dma2d compilation with new rendering

TychoVrahe:
This PR fixes minor issues with dma2d including, compiling and initialization with respect to new rendering.

`dma2d_init` function was added to bitblt so that DMA clock can be enabled, this function is now called in boardloader for models that use dma2d always. Only one of the files is compiled based on new rendering selection.

Also i width/height swap was fixed in `gfx_clear`.


cepetr (approved):
utACK

PR #4089, opened 2024-08-06
trezorlib: WebUSB resilience

matejcik:
* converts USB errors to TransportExceptions
* retries USB calls with a short timeout instead of waiting on blocking calls -- this means that Python-level signals (such as KeyboardInterrupt) will successfully stop a USB wait, so you can easily ctrl+c out of a stuck trezorctl
* this also allows pytest-timeout to kill a stuck hardware test

(we should also leverage this to be able to re-sync USB communication after a lost packet. this is not done in this PR however.)

in addition, I noticed and fixed a race condition in debuglink on the device side, which makes hw tests more resilient too

### DIFF
--- a/core/src/apps/debug/__init__.py
+++ b/core/src/apps/debug/__init__.py
@@ -127,10 +127,13 @@ if __debug__:
 
         return content
 
-    async def return_layout_change() -> None:
+    async def return_layout_change() -> None:  # type: ignore [Return type of async generator]
         content_tokens = await get_layout_change_content()
 
-        assert isinstance(DEBUG_CONTEXT, context.Context)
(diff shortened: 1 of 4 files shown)


mmilata (approved):
nice :pray:

prusnak (approved):

bosomt:
**QA OK**

```
❯ trezorctl list
webusb:001:1 - (unnamed) [Trezor Safe 5, 88935AC5C340F69739ABB611]

❯ trezorctl btc get-address -n m/44h/0h/0h/0/0 -d
Please enter passphrase on your Trezor device.
^C^C
Aborted!
```

PR #4090, opened 2024-08-06
fix firmware update on stm32u5a

TychoVrahe:
fixes missing code alignment when checking chunk hashes. (the alignment was increased recently due to larger alignment needs on STM32U5A/G)

doesn't affect any current model (except DISC2 evaluation board)


cepetr (approved):
utACK.

Here’s a refined version of your sentence:

As we discussed in person, we should also check the empty space between the image header and the vector table in the firmware binary before writing it to flash, not just before starting the firmware. But it's not critical at all.

TychoVrahe:
@cepetr i added the padding check in https://github.com/trezor/trezor-firmware/pull/4090/commits/5416c861ea2f976008c49aad5fe65e0dfa8220b3

The padding check was also added into boardloader when installing bootloader from sd-card (although it won't apply anywhere at this point, as the bootloader is correctly aligned for all MCUs).

I also found some bugs - the padding check in check_image_contents didn't really work and the image align when checking bootloader image on was wrong too. 

So please re-review as this got a bit more complicated.

core/embed/boardloader/main.c
```diff
@@ -167,9 +167,19 @@ static uint32_t check_sdcard(void) {
...
                    "BOOTLOADER IMAGE MAXSIZE too large for IMAGE_CHUNK_SIZE");
 
-    if (sectrue != (check_single_hash(
-                       hdr->hashes, ((const uint8_t *)sdcard_buf) + hdr->hdrlen,
-                       hdr->codelen))) {
+    const uint32_t headers_end_offset = hdr->hdrlen;
+    const uint32_t code_start_offset = IMAGE_CODE_ALIGN(headers_end_offset);
+
+    for (uint32_t i = headers_end_offset; i < code_start_offset; i++) {
+      if (sdcard_buf[i] != 0) {
```
cepetr, line 174: `headers_end_offset` and `code_start_offset` are both offsets in bytes, but `sdcard_buf` is defined as an array of `uint32_t`.

core/embed/bootloader/messages.c
```diff
@@ -707,6 +709,26 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
     return UPLOAD_ERR_FIRMWARE_TOO_BIG;
   }
 
+  // check padding between headers and the code
+  for (size_t i = headers_end; i < headers_offset; i++) {
+    if (CHUNK_BUFFER_PTR[i] != 0) {
+      if (firmware_upload_chunk_retry > 0) {
```
cepetr, line 715: What's the purpose of the block beginning with `if (firmware_upload_chunk_retry > 0 ) {`? Is it really needed?

core/embed/bootloader/messages.c
```diff
@@ -707,6 +709,26 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
     return UPLOAD_ERR_FIRMWARE_TOO_BIG;
   }
 
+  // check padding between headers and the code
```
cepetr, line 712: It seems to me that it would be possible to put this check inside the block above (just after `read_offset = 0`). We could then get rid of the new global variable `headers_end` - it could be local to the block where this check is implemented.

cepetr (changes requested):

core/embed/boardloader/main.c
TychoVrahe: fixed, thanks

https://github.com/trezor/trezor-firmware/pull/4090/commits/7c61e651fccbce7cdded30cb910c85d756ab5b33

core/embed/bootloader/messages.c
TychoVrahe: the purpose is to retry uploading the chunk if the check fails, so that user doesn't have to manually retry if it fails for any reason. Its also done for hash check below. Not strictly necessary though.

cepetr: I get it. Even though USB packets have a CRC field, it's only 16-bit, so there's a small chance a packet could still arrive corrupted. Personally, I think that risk is pretty low, and I'd prefer to keep the code simpler since this FSM is already quite complex. But I’m not set on this, so it's totally up to you.

core/embed/bootloader/messages.c
cepetr: Now that I see it, how about moving the check to the beginning of the function, right after `check_image_header_sig()` and before parsing `current_vhdr`? It seems like the best place, and you wouldn’t need to reread the `vhdr`. Since the first chunk is 16KB, it should contain the headers, padding, and the beginning of the firmware altogether.

TychoVrahe: i originally didn't want to do that because we limit vendor header size to 64kB actually, in image.c.

On the other hand with vendor header larger thank `(IMAGE_INIT_CHUNK_SIZE- IMAGE_HEADER_SIZE)` this code wont work anyway so it seems to me that we should limit vendor header size to match this.

The 16kB alignment should be enough for any padding due to vector table requirements. 

So i moved the check here https://github.com/trezor/trezor-firmware/pull/4090/commits/65a7ac7e040208eb809d5923b28b855b11ada47d

And i will look into the vendor header size limit in successive PR.

cepetr (approved):
utACK

PR #4092, opened 2024-08-07
fix: clarify how to break out of build_tx loops

### LINKED ISSUE #4191: `build_tx.py` is not working
The tool `python/tools/build_tx.py` doesn't allow me to exit the loop where inputs are entered:

```
Coin name [Bitcoin]: 
Blockbook server [btc1.trezor.io]: 

Previous output to spend (txid:vout) []: 1570416eb4302cf52979afd5e6909e37d8fdd874301f7cc87e547e509cb1caa6:0
From address: 1HWDaLTpTCTtRWyWqZkzWx1wex5NKyncLW
Input amount: 100000000
BIP-32 path to derive the key: m/44h/0h/0h/0/0
Script type: SPENDADDRESS
Sequence Number to use (RBF opt-in enabled by default) [4294967293]: 

Previous output to spend (txid:vout) []: 
Error: 
Previous output to spend (txid:vout) []: 
```

It can be fixed for example by this patch:
```diff
diff --git a/python/tools/build_tx.py b/python/tools/build_tx.py
index 12aabc768..422ec37fa 100755
--- a/python/tools/build_tx.py
+++ b/python/tools/build_tx.py
@@ -62,7 +62,9 @@ def _default_script_type(address_n: Optional[List[int]], script_types: Any) -> s
     # return script_types[script_type]
 
 
-def parse_vin(s: str) -> Tuple[bytes, int]:
+def parse_vin(s: str) -> Optional[Tuple[bytes, int]]:
+    if not s:
+        return None
     txid, vout = s.split(":")
     return bytes.fromhex(txid), int(vout)
 ```

### PR DESCRIPTION
**Pull request by ibz**
Fixes #4191

### CODE REVIEW

**Comment by prusnak** [comment]
If you enter empty string, the condition above (if not x) will be true and perform break instead of continue, no?

**Comment by ibz** [comment]
> If you enter empty string, the condition above (if not x) will be true and perform break instead of continue, no?

I think that was the original intent, indeed. But it does not seem to work for me. If you enter empty string, `parse_vin` will fail to parse it, and `prompt` will keep asking you for a new value until it receives one it can parse.

**Comment by obrusvit** [comment]
related: https://github.com/trezor/trezor-firmware/pull/4144

File: python/docs/EXAMPLES.rst [hunk 1]
```diff
@@ -72,7 +72,7 @@ After authenticating, open the "Send" tab, fill-out all details, then open the "
 
 .. code::
 
-  $ python3 tools/build_tx.py | trezorctl btc sign-tx -
+  $ python3 python/tools/build_tx.py | trezorctl btc sign-tx -
```
matejcik, line 75: the "source distribution" is what you get when you download `trezor...zip` from pypi
so the paths in `python/` subdir are relative to that. please revert here & below.

(the change is appropriate in `tools/snippets`)

matejcik:
it seems that the fix proposed in https://github.com/trezor/trezor-firmware/issues/4191 is simpler, why not use that?

ibz:
> it seems that the fix proposed in #4191 is simpler, why not use that?

That also works, but there are 2 loops I got stuck in, and this PR addresses both.

python/docs/EXAMPLES.rst
ibz: Done. https://github.com/trezor/trezor-firmware/pull/4092/commits/7407bd7c16a9f83563838da03ae323714d7ba7f9

matejcik (approved):

PR #4093, opened 2024-08-07
Public key prefixes rework

onvej-sl:


crypto/bip32.c
```diff
@@ -317,6 +323,13 @@ int hdnode_public_ckd_cp(const ecdsa_curve *curve, const curve_point *parent,
 int hdnode_public_ckd(HDNode *inout, uint32_t i) {
   curve_point parent = {0}, child = {0};
 
+  if (!inout->curve->params) {
+    // SLIP-10 doesn't support public key to public key derivation for
+    // curve25519 and ed25519 Cardano BIP32-ed22519 public key to public key
```
andrewkozlik, line 328:
```suggestion
    // curve25519 and ed25519. Cardano BIP32-ed22519 public key to public key
```

crypto/bip32.c
```diff
@@ -683,6 +701,8 @@ int hdnode_get_shared_key(const HDNode *node, const uint8_t *peer_public_key,
     *result_size = 65;
     return 0;
   } else if (node->curve == &curve25519_info) {
+    // The prefix 0x04 doesn't make sence here, and may be changed or removed in
```
andrewkozlik, line 704:
```suggestion
    // The prefix 0x04 doesn't make sense here, and may be changed or removed in
```

core/src/apps/misc/get_ecdh_session_key.py
```diff
@@ -55,11 +55,13 @@ async def get_ecdh_session_key(msg: GetECDHSessionKey) -> ECDHSessionKey:
 
         if peer_public_key[0] != 0x40:
             raise DataError("Curve25519 public key should start with 0x40")
+        # The prefix 0x04 doesn't make sence here, and may be changed or removed in the future
```
andrewkozlik, line 58:
```suggestion
        # The prefix 0x04 doesn't make sense here, and may be changed or removed in the future
```

legacy/firmware/layout2.c
```diff
@@ -963,14 +963,9 @@ void layoutAddress(const char *address, const char *desc, bool qrcode,
 }
 
 void layoutPublicKey(const uint8_t *pubkey) {
-  char desc[16] = {0};
+  char desc[15] = {0};
   strlcpy(desc, "Public Key: 00", sizeof(desc));
```
andrewkozlik, line 967:
```suggestion
  char desc[] = "Public Key: 00";
```

andrewkozlik (approved):

onvej-sl:
@romanz As the author `trezor-agent`, could you please take a look, especially at these commits? 

  * https://github.com/trezor/trezor-firmware/pull/4093/commits/175987af38f318f82f4589661cb8684b187de6bd changes the prefix of the curve25519 public key returned by `get_ecdh_session_key` from `0x01` to `0x00`. Previously, the prefix `0x01` was used internally to indicate that a public key is set. There is no reason to propagate it outside Trezor. We decided to change it to `0x00` since this is what [SLIP-10](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) uses for curve25519 and ed25119. This change should be backward compatible with `trezor-agent` since `trezor-agent` [removes the prefix](https://github.com/romanz/trezor-agent/blob/master/libagent/device/trezor.py#L148) without checking it.

  * https://github.com/trezor/trezor-firmware/pull/4093/commits/1cf6c4da4b9ae50a11582002a11dc01547802bcb  informs that the prefixes of public keys returned by `get_ecdh_session_key`, `sign_identity`, and `get_public_key` may be removed as they are non-standard. The prefix will still be used internally to compute SLIP-10 nodes. Our plan is to remove the prefix, possibly when introducing a new protocol between Trezor and the computer.

romanz:
@onvej-sl Many thanks for heads up!
The changes look good - I'll be happy to adapt `trezor-agent` when the new protocol will be introduced.

Issue #4084, opened 2024-08-04
Remove 50 character limit from passphrase


iiinnniii:
I cannot write in [203](https://github.com/trezor/trezor-mcu/issues/203), because it got archived. I hope this is the correct repo for this issue.

I do not agree at all that this is no important issue.

Please let me explain.

Actually the 50 character limit is a major issue in regards to wallet recovery in case of a single letter failure within the backup (or much worse multiple letters, which I explain below). 

If you use a long passphrase, then using a 12 word bip39 seed phrase as a passphrase with a checksum actually makes sense from a wallet recovery perspective in regards to storing your crypto safe and securely if you want to use a passphrase. I read on reddit that many people are doing this it seems with other wallets, but it is not possible with Trezor. Having a checksum in the passphrase is useful for recovery purposes in case of errors within the backup. If you don't use a bip39 seed as a passphrase, but instead a 50 letter random passphrase, you essentially made your chance of wallet recovery much worse in case for example two characters are missing within the backup. Even if you are using as an example a 20 character random string as a passphrase it is also much more susceptible to errors within the backup. You do not want your passphrase become the weak point, but with the 50 character limit it becomes the weak point.

With a bip39 seed phrase as a passphrase you can use common recovery tools on an offline PC in case of errors within the backup. The checksum is very useful there. If you do not have a checksum and you miss two letters of a 20 character random string passphrase then I don't see how you would be able to recover from that without spending an immense amount of time writing code and exposing the seed on a PC connected to the internet trying to brute force and check if funds are in the hidden wallet. And if too many characters are missing that becomes unfeasible. That sounds absolutely horrible to me and a situation that nobody wants to be in.

Actually this is one thing that I really do not like about Trezor, because this is bad and there is no argument to not remove this limit. The argument made by the Trezor team was it is not important enough, but as I outlined above it is definitely important and can potentially result in the loss of funds.

By the way using only the first 4 chars of a 12 word bip39 seed also weakens the ability to recover a wallet compared to using the full words, if you would want to suggest that. It would also become the weak point if your actual recovery seed backup does use full words, which it always should with a 12 word seed.

50 random characters without checksum is much more error prone to backup failure then a longer passphrase comprised of a 12 word bip39 seed which includes a checksum. 

The 50 character limit for the passphrase decreases the ability for people to secure their crypto securely if they use a passphrase as outlined above.

This is my understanding of this issue.

I would be happy to hear that this 50 character limit finally gets removed for the good of all.

[...]

PR #4069, opened 2024-07-29
T3T1 UI: FIDO2 flow

mmilata:
emu00000000 emu00000000 emu00000001 emu00000002 emu00000001 emu00000002

Fixes: #3797, [Figma](https://www.figma.com/design/n43caWJFGJlT167AQcxruu/branch/uIdQ3oz7UJrfurU3ADXVeO/%E3%80%94-Prod-%E3%80%95Safe-5?node-id=3075-3429&t=0VU4SSSzg8bK69sg-0)

Submodule dependencies: trezor/fido2-tests#4

PR #4094, opened 2024-08-07
T3T1 UI - passphrase flow improvement

obrusvit:
This PR adds reassuring screen when entering empty passphrase.

In order to do that, this includes some necessary changes:
- aa79f02ac4d967aa74521836831a02a6d38778ef adds `FlowMsg::Text` variant which enables passing `ShortString` to uPy as `str`. Btw. `ShortString` is conveniently limited to `50` chars which needs to be revisited if we implement https://github.com/trezor/trezor-firmware/issues/4084#issuecomment-2268850876 
- ebc297f3c73dc70de08e6471ca06cd082c14a68f implements new component - a simple yes/no prompt with buttons. It will be used further in some flows from Figma
- 55633a765fc1fad71dc5149cd5906930dc5002e4 switching two buttons in passphrase keyboard, one for empty passphrase, one for non-empty
- a06c83d30600b8c585d1a9be2f5a65df83b19d64 adds dedicated `PromptMsg` instead of passing empty `Option` around. This has the benefit that the `BinarySelection` from ebc297f3c73dc70de08e6471ca06cd082c14a68f can be used as `PromptScreen::new_yes_or_no()`
- 395bcb6ccf8079661f106adca66be35d2359d2b3 implements the passphrase flow


Hannsek:
cc @lapohoda

obrusvit:
@Hannsek UI changes briefly:

The confirmation of empty passphrase:
refresh00-00000001 refresh00-00000002
refresh00-00000003 refresh00-00000004

Usage of subtitle in other screens:
refresh00-00000006 refresh00-00000007

lapohoda:
@obrusvit pls doublecheck the size of confirmation button – should be circular according to Figma

image

obrusvit:
UI diff of FIDO2 actions are coincidental due to confirm button stylesheet change and will be superseded by https://github.com/trezor/trezor-firmware/pull/4069.

matejcik (approved):
utACK.

general note: it's not great that you're adding `Text(ShortString)` in there, and especially not great that you're adding it to the passphrase Confirmed() variant. What we originally intended with ComponentMsgObj (on the model_tt side) is that you return a "user is done" message, then the ComponentMsgObj impl asks the layout to give out a `&str`, and we convert _that_ to micropython.
Here, we'd want `.map()` to pass you both the original layout and the returned message -- so that you can do this part in the `map()` closure.

on the other hand, i'd prefer getting rid of `Map` component the way we use it now. and currently doing it like this fits well with overall way of doing things in mercury. We can refactor the things when refactoring time comes, i guess.

obrusvit:
Merging amidst `T3T1` coverage failing. I'll take a look at different PR.

PR #4095, opened 2024-08-08
feat(core): improve ETH send flow.

ibz:


ibz:
Here are screenshots from the screens of the flow...

image

Note: *Account info* screen missing for now, since we don't have the account # and wallet info...

image

image

image

image

image

image

core/src/trezor/ui/layouts/mercury/__init__.py
```diff
@@ -970,48 +976,26 @@ async def confirm_ethereum_tx(
...
-            except ActionCancelled:
-                continue
+        await _confirm_summary(
+            items=(
+                (TR.words__amount, total_amount),
+                (TR.send__maximum_fee, maximum_fee),
+            ),
+            fee_items=fee_info_items,
+            cancel_text=TR.buttons__cancel,
+        )
```
obrusvit, line 998: This is a regression from previous version because we cannot go back and forth between recipient and summary, no?

core/translations/en.json
```diff
@@ -299,13 +299,13 @@
     "ethereum__amount_sent": "Amount sent:",
     "ethereum__contract": "Contract:",
     "ethereum__data_size_template": "Size: {0} bytes",
-    "ethereum__gas_limit": "Gas limit:",
```
obrusvit, line 302: Removing the `:` and adding them dynamically when constructing layouts works but brings undesirable changes to `fr` language because french uses space-colon.

We can either ignore it or duplicate the strings to have e.g.
`ethereum__gas_limit` and `ethereum__gas_limit_colon`.

tests/device_tests/ethereum/test_signtx.py
```diff
@@ -209,14 +210,21 @@ def test_data_streaming(client: Client):
...
         client.set_expected_responses(
             [
                 messages.ButtonRequest(code=messages.ButtonRequestType.SignTx),
                 (is_t1, messages.ButtonRequest(code=messages.ButtonRequestType.SignTx)),
                 (
-                    not is_t1,
+                    not is_t1 and not is_mercury,
                     messages.ButtonRequest(code=messages.ButtonRequestType.Other),
                 ),
+                (
```
obrusvit, line 222: Why was it necessary to have a different `BR`s for `mercury` in this flow?

core/embed/rust/src/ui/model_mercury/flow/confirm_summary.rs
```diff
@@ -111,37 +119,59 @@ impl ConfirmSummary {
...
+                TR::address_details__account_info.into(),
+            );
+            unwrap!(menu_items.push(MENU_ITEM_ACCOUNT_INFO));
+        }
+        menu = menu.danger(
+            theme::ICON_CANCEL,
+            cancel_text.unwrap_or(TR::send__cancel_sign.into()),
+        );
+        unwrap!(menu_items.push(MENU_ITEM_CANCEL));
+        let content_menu = Frame::left_aligned("".into(), menu)
```
obrusvit, line 164: Nit: you can use `TString::empty()` for the title arg.

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
```diff
@@ -41,6 +44,47 @@ impl FlowState for ConfirmOutput {
         *self as usize
     }
 
+    fn handle_swipe(&'static self, direction: SwipeDirection) -> StateChange {
+        match (self, direction) {
+            (Self::Address, SwipeDirection::Left) => Self::Menu.swipe(direction),
+            (Self::Address, SwipeDirection::Up) => self.return_msg(FlowMsg::Confirmed),
+            (Self::Menu, SwipeDirection::Right) => Self::Address.swipe(direction),
+            (Self::Menu, SwipeDirection::Left) => Self::AccountInfo.swipe(direction),
```
obrusvit, line 52: Not sure we want to go from `Menu` to one of the items by swiping left.

obrusvit (commented):
Some points to discuss regarding the flow.

UI part in rust seems good :+1:

core/src/trezor/ui/layouts/mercury/__init__.py
ibz: According to the design I don't see a way to go back. In the old design you would go to the Cancel screen in the 2nd step and that would let you go back to the 1st step. Now the Cancel screen is called *Cancel sign*, so it should cancel the whole flow rather than just the 2nd step. TBH, the old way was a bit clunky IMO. I do expect the Cancel screen to cancel everything, not go back one step. What we could do, maybe, is use swipe down to go back.

core/translations/en.json
ibz: Yes, I was thinking about duplication for this reason, but then I worried it would add more to the flash size and this was a decent compromise. But maybe I am wrong, maybe only strings that are actually used make it to the image, then it is not an issue.

Speaking of this, I just read about it now, it seems the space before semicolon in French should be a non-breaking space (makes sense!). Now that we support that, I should also replace all occurrences with the non-breaking space!

ibz: An even better way is introducing something like `label__with_semicolon_template = "{0} :"` and using that everywhere. I'll do that.

core/src/trezor/ui/layouts/mercury/__init__.py
Hannsek: You go back to the recipient from summary screen by swiping down.

And yes, cancel should cancel the whole flow. On TT, there is an upper arrow. See [here](https://www.figma.com/design/l0gG9XeRJ8FTDQ3cfb1wv9/Firmware?node-id=8833-74263&t=TV8t9vF9ak72q97n-4).

obrusvit:
This PR also adds Success screen with "Transaction signed" and "Continue in the app" for all models.
emu00000000


But I don't think it's desirable UX-wise e.g. when you Stake.
I think the success screens are designed for TS5 only and the idea for the future is to have them timeout after a few seconds while TrezorSuite already gets the result (`Success`). This is not yet possible, tracked here https://github.com/trezor/trezor-firmware/issues/3666 

So I would remove the Success screen for now.

ibz:
> So I would remove the Success screen for now.

OK, I'll remove the success screen from all models then, as discussed with you (re #3666).

What about the progress screen - does it even make sense given it is not a "real" progress and the signing is pretty fast? @Hannsek

core/src/trezor/ui/layouts/mercury/__init__.py
obrusvit: Still, it is a regression from model T and model R where you can browse through recipient and summary. 
Let's take model T as an example, the 2nd part (summary) has the close button `x` replaced with arrow up `^`, which is a workaround so that it seems like you go back, in fact you cancel the 2nd screen and catch `ActionCancelled`. This approach does not translate into mercury, where you have "Cancel" button. 

Possible solutions:
1. wrap it again into an infinite loop in uPy and figure out a control logic which breaks it if you cancel any window while also enabling going back and forth.
This might be difficult because we need to distinguish somehow what the user wants from the return object sent to uPy, we cannot distinguish whether `CANCELLED` comes from swipe down action of the summary or clicking the "Cancel" button. Maybe we can misuse `INFO` but I wouldn't go this way. Introducing a new object seems like an overkill but perhaps it might make sense... something like `GO_PREVIOUS` :thinking: 
2. pack all the arguments for rust and call just one flow layout. That would mean merging `confirm_output` and `confirm_summary` into something like `flow_confirm_altcoin_tx` - this might take longer to program but is the correct solution IMHO. We could handle swiping here and there.

Hannsek:
Progress screen "signing transaction" is quick for small transaction, for bigger transaction it is not that quick. Let's keep it.

core/src/trezor/ui/layouts/mercury/__init__.py
ibz: https://github.com/trezor/trezor-firmware/pull/4095/commits/acb1cdeadb4ef891c6bdcfb3e0220ca49a9ed4f6 Here is the solution.

tests/device_tests/ethereum/test_signtx.py
ibz: It was not. Thank you for pointing it: https://github.com/trezor/trezor-firmware/pull/4095/commits/ea4a00a6582e943a7a4af1448999ae7f0800b31c

core/embed/rust/src/ui/model_mercury/flow/confirm_summary.rs
ibz: https://github.com/trezor/trezor-firmware/pull/4095/commits/d2cded804019bbb5fe3d646216db582943bd5826#diff-5cb317ed79ea16b58238d2743af510c9f82a27c2764811d83f532f8d8358b6f6R161

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
ibz: https://github.com/trezor/trezor-firmware/pull/4095/commits/d2cded804019bbb5fe3d646216db582943bd5826#diff-5cb317ed79ea16b58238d2743af510c9f82a27c2764811d83f532f8d8358b6f6L95

ibz:
> So I would remove the Success screen for now.

https://github.com/trezor/trezor-firmware/pull/4095#issuecomment-2296141502

core/src/trezor/ui/layouts/mercury/__init__.py
ibz: Swipe up works now. Also, opening the menu then closing the menu brings you back to the correct screen which you were on before the menu got opened.

Clicking *Cancel sign* then canceling the *Cancel* operation however brings you back to the beginning of the flow, as opposed to the screen you were on when you entered the menu that got you to *Cancel sign*. It's a pretty trivial fix to do it the other way, if @Hannsek thinks it is better UX. I am not sure honestly which way is better... I am guessing that if you were about to cancel, then decided to not cancel... you might as well review all steps again.

BTW, this is now one flow (*confirm_output*), with optional steps for *amount* and *summary* (which in turn goes further to *sign*). It should probably be reused in other cases as well so that we can simplify some of the other use cases. I'll have a look where else it would be a fit.

obrusvit: @obrusvit generally, yes
However, the behavior of flows is not programmed at one place. You have a way to program the transitions for each flow separately, so a careful review is needed.

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
```diff
@@ -69,68 +113,302 @@ impl FlowState for ConfirmOutput {
...
+            amount.unwrap_or(Obj::const_none()),
+            None,
+        )
+        .with_subtitle(title)
+        .with_menu_button()
+        .with_footer(TR::instructions__swipe_up.into(), None)
+        .with_text_mono(text_mono)
+        .with_swipe_down()
+        .into_layout()?
+        .one_button_request(ButtonRequest::from_num(br_code, br_name));
```
obrusvit, line 296: We should use `repeated_button_request` as that's the behavior from `T` and `R`.

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
```diff
@@ -69,68 +113,302 @@ impl FlowState for ConfirmOutput {
...
+            }
+            FrameMsg::Button(_) => Some(FlowMsg::Cancelled),
+        });
+
+    // AccountInfo
+    let ad = AddressDetails::new(TR::send__send_from.into(), account, account_path)?;
+    let content_account = ad.map(|_| Some(FlowMsg::Cancelled));
+
+    // AddressMenuCancel
+    let content_address_menu_cancel = Frame::left_aligned(
```
obrusvit, line 270: All three `content_x_y_cancel` are identical and the code is copy-pasted 3x. We should write a function which constructs it.

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
```diff
@@ -69,68 +113,302 @@ impl FlowState for ConfirmOutput {
...
+                Self::SummaryMenuCancel.swipe_left()
+            }
+            (Self::SummaryMenuCancel, FlowMsg::Cancelled) => Self::SummaryMenu.swipe_right(),
+            (Self::Hold, FlowMsg::Info) => Self::HoldMenu.transit(),
+            (Self::HoldMenu, FlowMsg::Choice(MENU_ITEM_CANCEL)) => {
+                Self::HoldMenuCancel.swipe_left()
+            }
+            (Self::HoldMenuCancel, FlowMsg::Cancelled) => Self::HoldMenu.swipe_right(),
+            (
+                Self::AddressMenu | Self::SummaryMenu | Self::HoldMenu,
```
obrusvit, line 170: Probably unnecessary to specify all `Menu` if the option is only present in `Summary`, no?

Same for `AccountInfo` below.

core/embed/rust/src/ui/model_mercury/flow/confirm_summary.rs
obrusvit, line 1: This file should be reverted back now, no?

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -1758,12 +1758,18 @@ pub static mp_module_trezorui2: Module = obj_module! {
...
     ///     br_code: ButtonRequestType,
     ///     br_name: str,
+    ///     summary_items: Iterable[tuple[str, str]] | None = None,
+    ///     fee_items: Iterable[tuple[str, str]] | None = None,
+    ///     summary_title: str | None = None,
+    ///     summary_br_code: ButtonRequestType | None = None,
+    ///     summary_br_name: str | None = None,
+    ///     cancel_text: str | None = None,
     /// ) -> LayoutObj[UiResult]:
     ///     """Confirm recipient."""
```
obrusvit, line 1774: This documentation "Confirm recipient." does not seem adequate anymore :smile:

obrusvit (commented):
Sending ETH works nicely :+1: 

Apart from the individual comments, I'd like to discuss a few more points:
- the `new_confirm_summary` is very very long, perhaps we can break it down into smaller pieces or even flows
- ETH confirm stake/unstake/claim flows are broken now - they don't show the info according to Figma. The flow in python function was not changed but we changed the `confirm_summary.rs` which now disrupts it. We should check whether it can be mapped to the new `confirm_summary` but I'm afraid we'd have to write a new `flow_`. I can help.
- I didn't check other operations on ETH, e.g. with data or `unknown contract` etc.

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
ibz: https://github.com/trezor/trezor-firmware/pull/4095/commits/ec6a2a8c8d7e840f3b19e2ab09c9bb08806c1449#diff-5cb317ed79ea16b58238d2743af510c9f82a27c2764811d83f532f8d8358b6f6R188

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
ibz: https://github.com/trezor/trezor-firmware/pull/4095/commits/ec6a2a8c8d7e840f3b19e2ab09c9bb08806c1449#diff-5cb317ed79ea16b58238d2743af510c9f82a27c2764811d83f532f8d8358b6f6R169

core/embed/rust/src/ui/model_mercury/layout.rs
ibz: https://github.com/trezor/trezor-firmware/pull/4095/commits/ec6a2a8c8d7e840f3b19e2ab09c9bb08806c1449#diff-e939e2f37ae15d507723cbfa63ffa095f904d9e81ec403ddecb57c09ac09d3f4R1774

core/embed/rust/src/ui/model_mercury/flow/confirm_summary.rs
ibz: I don't think so. It is an improvement over what was before. Any reason to revert it?

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
obrusvit: I see it fixed https://github.com/trezor/trezor-firmware/pull/4095/commits/26436ef6aa930f019e85acf0d0b48e22838b91b3

core/embed/rust/src/ui/model_mercury/flow/confirm_summary.rs
obrusvit: mistake on my side

obrusvit:
There's a limitation in the ButtonRequest behavior of the `confirm_output` as it does not send the BR everytime if the user goes back and forth. We ignore it for now because Suite does not handle them repeatedly anyway.

We realized that sending `BR` from rust on every `Event::Attach` is not equivalent to sending `BR` in a `while` loop with from uPy. A layout started from uPy has two pages, it sends only one `BR`
A usage of `.repeated_button_request` from rust sends it twice for every page.

obrusvit (approved):

Hannsek:
@ibz next time please ping @lapohoda to check it. :)

bosomt:
**QA OK**

Hannsek:
@ibz Have you done also the part with [signing a contract](https://www.figma.com/design/n43caWJFGJlT167AQcxruu/%E3%80%94-Prod-%E3%80%95Safe-5?node-id=5532-3020&t=osgHsJ6kRxtaWURJ-4)? 
image

ibz:
I double checked now, seems to be a separate path that I did not touch.

PR #4099, opened 2024-08-11
feat(core): ask for 20/33 words to unlock backup

obrusvit:
Unlocking repeated backup is only possible with SLIP-39 extendable backup which contains shares of 20 or 33 words. It is UI bug to have the user choose from all options (12, 18, 20, 24, 33). This commit reduces the choices when unlocking repeated backup to (20, 33) for all models.

TODO:
- [x] click tests (9828605fee317ffc1bbcac605d2b3e8a654e45ca)
- [x] find a suitable home for `RECOVERY_TYPE_<xyz>` constants (684936f67bc663a5f72ca1d33011355969cbd5af)
- [x] is it possible to have reduced options also for regular backup check?
    - the answer is no
- [x] clippy doesn't like big size of enum variants of `SelectWordCount`
    - decided to silence this warning as I don't see a performance hit for the following reasons
       1. the larger variant predates the reduced option and it already worked with the larger one
       2. the enum is not instantiated frequently
       2. it makes the code simpler


obrusvit:
UI diff, cc @Hannsek @lapohoda:

T3T1:
image

T2B1, T3B1:
image

T2T1:
image

core/embed/rust/src/ui/model_tt/component/keyboard/word_count.rs
```diff
@@ -8,33 +8,99 @@ use crate::ui::{
...
+
+    fn render<'s>(&'s self, target: &mut impl Renderer<'s>) {
+        match self {
+            SelectWordCount::All(full_selector) => full_selector.render(target),
+            SelectWordCount::Multishare(bin_selector) => bin_selector.render(target),
+        }
+    }
+}
+
+pub struct ValueKeypad<const N: usize> {
```
mmilata, line 81: This is still the case where rustc makes copies of methods for each generic instance. Using Vec or array slices would result in 2x less code generated.

core/embed/rust/src/micropython/recovery_type.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(RecoveryType::Normal),
+            1 => Ok(RecoveryType::DryRun),
+            2 => Ok(RecoveryType::UnlockRepeatedBackup),
+            _ => Err(value_error!(c"Invalid RecoveryType")),
+        }
+    }
+}
```
mmilata, line 40: Appreciate putting the enum in its own file but it doesn't belong under `crate::micropython`. Currently the best place is under `crate::ui::layout::util`.

mmilata (changes requested):

core/embed/rust/src/micropython/recovery_type.rs
obrusvit: My thought was that this value is utilized only to convert parameters coming from uPy so why not putting it into `crate::micropython`. But I guess you're right that util is more suitable for now. Even though I can imagine moving it out of `crate::ui` in the future as it's not just UI related.
ba3332c2067f739acb727065d979da84f73ad8eb

core/embed/rust/src/ui/model_tt/component/keyboard/word_count.rs
obrusvit: I considered that the generic parameter makes it duplicated but:
- the Component is very small
- this is for Model T only, where the space is not a problem. 

Anyway, I tried using `heapless::Vec`, I tried in two arrangements:
1. `Vec` for each of these, like this
```rust
pub struct ValueKeypad {
    buttons: Vec<Button, 5>,
    numbers: Vec<u32, 5>,
    cells: Vec<(usize, usize), 5>,
}
```
2. `Vec` for a tuple of everything, something like this:
```rust
type ValueKeyPacked = (Button, u32, (usize, usize));
pub struct ValueKeypad {
    buttons: Vec<ValueKeyPacked, 5>,
}
```
Both versions result in the same binsize. Do you have other suggestions?

mmilata: 2. seems nice ... though I didn't realize the memory situation on T2T1 is less worse so feel free to leave it as it is with a comment warning against adding further instances.

core/embed/rust/src/micropython/recovery_type.rs
mmilata: I think `crate::micropython` is currently one of the tidier sub-crates, compared to e.g. UI, let's keep it that way:D.

Conceptually the enum belongs somewhere under application-specific constants, which we currently don't have. But ultimately it should be generated from protobuf along with e.g. ButtonRequestType.

core/embed/rust/src/ui/model_tt/component/keyboard/word_count.rs
obrusvit: So I measured again and `Vec` actually saves one block of flash on debug build. Not much but we might prevent bloat by other variants potentially in the future. So I implemented the version with Vec.

@mmilata ftr, the situation on `T2T1` with this change
`FLASH` - 83.72% used
`FLASH2:` - 92.75% used
d662aa1c2307182080af4ee375d7b7bed60d5bcf

mmilata (approved):

bosomt:
**QA OK**

only 20 and 33 words are available on TS3 and TS5

**Info:**
 - Suite version: desktop 24.9.0 (dea394b61290428b18d191b05391a5b365b6fa8b)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.9.0 Chrome/124.0.6367.243 Electron/30.3.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.8.2 regular (revision 9e18e72a2c7896cbcce1df4c2d908eab0f01564f)

PR #4101, opened 2024-08-13
Benchmark application

onvej-sl:
This pull request implements an application that benchmarks various cryptographic primitives.

The application is hidden behind a compilation flag:
```BENCHMARK=1 make -C core build_firmware```

Using `trezorctl`, you can:
  * List all benchmarks or only the benchmarks matching a given pattern:
```
$ trezorctl benchmark list-names
crypto/cipher/aes128-ecb/decrypt
crypto/cipher/aes128-ecb/encrypt
crypto/cipher/chacha20poly1305/decrypt
crypto/cipher/chacha20poly1305/encrypt
crypto/curve/curve25519/multiply
crypto/curve/curve25519/publickey
crypto/curve/ed25519/publickey
crypto/curve/ed25519/sign
crypto/curve/ed25519/verify
crypto/curve/nist256p1/multiply
crypto/curve/nist256p1/publickey
crypto/curve/nist256p1/sign
crypto/curve/nist256p1/verify
crypto/curve/secp256k1/multiply
crypto/curve/secp256k1/publickey
crypto/curve/secp256k1/sign
crypto/curve/secp256k1/verify
crypto/hash/blake256
crypto/hash/blake2b
crypto/hash/blake2s
crypto/hash/groestl512
crypto/hash/ripemd160
crypto/hash/sha1
crypto/hash/sha256
crypto/hash/sha3_256
crypto/hash/sha3_512
crypto/hash/sha512
```
```
$ trezorctl benchmark list-names "crypto/curve/*/sign"
crypto/curve/ed25519/sign
crypto/curve/nist256p1/sign
crypto/curve/secp256k1/sign
```
 * Run all benchmarks or only the benchmarks matching a given pattern:
```
$ trezorctl benchmark run
crypto/cipher/aes128-ecb/decrypt: 0.96 MB/s
crypto/cipher/aes128-ecb/encrypt: 0.98 MB/s
crypto/cipher/aesgcm128/decrypt: 0.31 MB/s
crypto/cipher/aesgcm128/encrypt: 0.30 MB/s
crypto/cipher/aesgcm256/decrypt: 0.29 MB/s
crypto/cipher/aesgcm256/encrypt: 0.29 MB/s
crypto/cipher/chacha20poly1305/decrypt: 2.1 MB/s
crypto/cipher/chacha20poly1305/encrypt: 2.1 MB/s
crypto/curve/curve25519/multiply: 8.8 ms
crypto/curve/curve25519/publickey: 3.9 ms
crypto/curve/ed25519/publickey: 4.1 ms
crypto/curve/ed25519/sign: 8.3 ms
crypto/curve/ed25519/verify: 11 ms
crypto/curve/nist256p1/multiply: 110 ms
crypto/curve/nist256p1/publickey: 23 ms
crypto/curve/nist256p1/sign: 27 ms
crypto/curve/nist256p1/verify: 130 ms
crypto/curve/secp256k1/multiply: 110 ms
crypto/curve/secp256k1/publickey: 17 ms
crypto/curve/secp256k1/sign: 17 ms
crypto/curve/secp256k1/verify: 8.2 ms
crypto/hash/blake256: 1.5 MB/s
crypto/hash/blake2b: 2.3 MB/s
crypto/hash/blake2s: 3.1 MB/s
crypto/hash/groestl512: 0.16 MB/s
crypto/hash/ripemd160: 5.5 MB/s
crypto/hash/sha1: 2.7 MB/s
crypto/hash/sha256: 1.9 MB/s
crypto/hash/sha3_256: 0.63 MB/s
crypto/hash/sha3_512: 0.33 MB/s
crypto/hash/sha512: 1.1 MB/s
```

### CODE REVIEW

File: core/src/apps/benchmark/__init__.py
Comment (matejcik) [inline core/src/apps/benchmark/__init__.py:1]: i'm thinking we should maybe limit the benchmark app to `__debug__` ?

```
if not __debug__:
     from trezor import utils
     utils.halt("Disabled in production mode")
```

File: core/src/apps/workflow_handlers.py [hunk 1]
```diff
@@ -204,6 +204,12 @@ def _find_message_handler_module(msg_type: int) -> str:
         if msg_type == MessageType.SolanaSignTx:
             return "apps.solana.sign_tx"
 
+        # benchmark
```
matejcik, line 207: limit to `if __debug__` ?

core/src/apps/benchmark/__init__.py
onvej-sl: What would be the advantage of limiting it to `__debug__` if it is not included in the firmware unless you compile the firmware with `BENCHMARK=1`?

matejcik: it limits our opportunity to mess up and wrongly compile it into production firmware

onvej-sl: I understand that. On the other hand, there is a higher probability that you won't measure what you expect when you use `PYOPT=0`. In fact, the results are slightly different for `PYOPT=0` and `PYOPT=1`. The biggest difference is about 10%. Nevertheless, I don't consider it a significant issue since the primary purpose is to compare results across different branches. The absolute number will never be entirely accurate.

If we restrict the benchmark application only to debug mode, I would like to add something to the sconscript to prevent the use of `PYOPT=1` and `BENCHMARK=1` together. Any preference?

  1. 
```
if BENCHMARK and PYOPT != '0':
  print("BENCHMARK=1 works only with PYOPT=0.")
  exit(1)
```
  3.
```
if BENCHMARK:
  PYOPT = '1'
```

common/protob/messages.proto
```diff
@@ -375,4 +375,10 @@ enum MessageType {
     MessageType_SolanaAddress = 903 [(wire_out) = true];
     MessageType_SolanaSignTx = 904 [(wire_in) = true];
     MessageType_SolanaTxSignature = 905 [(wire_out) = true];
+
+    // Benchmark
+    MessageType_BenchmarkListNames = 1000 [(bitcoin_only) = true, (wire_debug_in) = true];
```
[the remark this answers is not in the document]
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4101/commits/dcd7a35d88bbcfc1cbc3271f78f434cf1588c7bc and https://github.com/trezor/trezor-firmware/pull/4101/commits/636bd95c0cc05f3f3f6dfe65c1deb8c9f9b6a216.

core/src/apps/benchmark/__init__.py
matejcik: it's weird that PYOPT affects this. my guess is GC pauses, otherwise I don't see what could be the difference?
(would it make sense to stop the timer and run `gc.collect()` e.g. every 10-100 invocations?)

out of the two options above, my preference is (1)

onvej-sl: Calling `gc.collect()` didn't help. It is caused by  the use of `-O0` versus `-O1` here:
https://github.com/trezor/trezor-firmware/blob/636bd95c0cc05f3f3f6dfe65c1deb8c9f9b6a216/core/SConscript.firmware#L548

matejcik: i mean that is what PYOPT does, yes, but per [this doc](https://docs.micropython.org/en/latest/library/micropython.html#micropython.opt_level) the difference is that at -O1 (a) asserts are not compiled, and (b) __debug__ is set to False
neither of which should affect benchmark results.

in terms of effect on our code, `__debug__` implies that:
* debug statements are printed to the console -- except there should be none while a benchmark is running because the code is synchronous
* more things are built in, in particular the debuglink feature
leads me to believe that about the only thing that could affect the benchmarks is GC performance...?

core/src/apps/workflow_handlers.py
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4101/commits/f032927e94cbda9c6ef559cab98317c35afb85b8.

core/src/apps/benchmark/__init__.py
onvej-sl: Ok, I understand, it's strange. In any case, I don't hink it's necessary to find the cause.

onvej-sl: See https://github.com/trezor/trezor-firmware/pull/4101/commits/f032927e94cbda9c6ef559cab98317c35afb85b8.

core/Makefile
```diff
@@ -280,25 +281,26 @@ build_firmware: templates build_cross ## build firmware with frozen modules
 		PYOPT="$(PYOPT)" BITCOIN_ONLY="$(BITCOIN_ONLY)" \
 		BOOTLOADER_QA="$(BOOTLOADER_QA)" BOOTLOADER_DEVEL="$(BOOTLOADER_DEVEL)" \
 		DISABLE_OPTIGA="$(DISABLE_OPTIGA)" THP="$(THP)" NEW_RENDERING="$(NEW_RENDERING)" \
-		$(FIRMWARE_BUILD_DIR)/firmware.bin
+		BENCHMARK="$(BENCHMARK)" $(FIRMWARE_BUILD_DIR)/firmware.bin
```
matejcik, line 284: note that there's now a common section `SCONS_OPTS` where this will go when you're rebasing

core/SConscript.firmware
```diff
@@ -17,11 +17,17 @@ HW_REVISION = ARGUMENTS.get('HW_REVISION', None)
...
+
+if BENCHMARK and PYOPT != '0':
+  print("BENCHMARK=1 works only with PYOPT=0.")
+  exit(1)
 
 FEATURE_FLAGS = {
     "RDI": True,
     "SECP256K1_ZKP": True,  # required for trezor.crypto.curve.bip340 (BIP340/Taproot)
     "AES_GCM": False,
+    "AES_GCM": BENCHMARK,
```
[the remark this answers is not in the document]
M1nd3r: Fixed here: https://github.com/trezor/trezor-firmware/pull/4268

PR #4103, opened 2024-08-15
evaluate model before vendor header signature

TychoVrahe:
Evaluate model before vendor header signature when installing firmware. Point is to show more useful error message.


core/embed/lib/image.c
```diff
@@ -190,6 +193,21 @@ secbool __wur read_vendor_header(const uint8_t *const data,
   return sectrue;
 }
 
+secbool check_vendor_header_model(const vendor_header *const vhdr) {
+#ifdef TREZOR_MODEL_T
+  if (vhdr->hw_model == 0) {
+    // vendor headers for model T have this field set to 0
+    return sectrue;
+  }
+#else
```
matejcik, line 202: Not sure about the `#else`. If in the future we make new vendor headers for TT, we will need to remember to clear the `hw_model` field (which is also inconsistent with how firmware headers are treated, we allow both 0 and `T2T1`)

OTOH it's not that much of a problem, seeing as new vendor headers probably won't happen, and if they do, they'll be a copypaste of the current jsons

TychoVrahe: seems reasonable to allow `T2T1` here too
https://github.com/trezor/trezor-firmware/pull/4103/commits/459b8a85ba3a53850dfb02c3317adde08f56e128

matejcik (approved):

PR #4105, opened 2024-08-16
check bootloader integrity before update

TychoVrahe:
Error like this should be always catched by QA, but having this kind of check in firmware too to prevent potential device-bricking incidents seems useful.


TychoVrahe:
@matejcik pls check conceptually, @cepetr please review the code.

matejcik:
i'm wondering, in what kind of situation would this come into play?

if we build a broken bootloader image into the firmware? then maybe we should have stronger build-time checks
if the bootloader gets corrupted in flash? then the existing bootloader should not load the fw ... but perhaps if it gets corrupted at runtime...?

TychoVrahe:
Building in broken bootloader is my main worry. Yes stronger build-time checks would help but this is more ultimate solution.

I agree about flash corruption. Runtime corruption is would be more troublesome but then this check might not be ideal - as due to bootloader compression i need to do this in two phases - decompress and check, decompress and flash -> and the decompression is a process that would cause this most likely during runtime i guess..

TychoVrahe:
One more thing to consider - after kernel split this will require to do the signature check in kernel - which means some more crypto, and thus duplicated code, used there.

TychoVrahe:
All right, closing in favor of https://github.com/trezor/trezor-firmware/pull/4130

(though the `bootloader_hashes` doesn't really check the bootloader integrity so we might want to add that later)

PR #4106, opened 2024-08-16
introduce ui_overlay feature

TychoVrahe:
this fixes compilation of T3T1 bootloader with new rendering, which was overflowing RAM.


cepetr (approved):
utACK

Issue #4079, opened 2024-08-01
Reduce unrolling in hash functions


matejcik:
blake2 is the biggest offender here:
https://github.com/trezor/trezor-firmware/blob/c3161f8b18193d648ed6a8a3cf6278655114b891/crypto/blake2b.c#L189-L255
I converted the unrolling to function calls, reducing code size by 12 kB (this is also used in bootloader and possibly in boardloader too?)

@onvej-sl I would like to ask for your help evaluating this change in terms of correctness and performance impact
are there other places that you know of where we could also re-roll code, so to speak?

onvej-sl:
Done in https://github.com/trezor/trezor-firmware/pull/4107.

PR #4107, opened 2024-08-16
Memory optimization

onvej-sl:
Solves https://github.com/trezor/trezor-firmware/issues/4079.

In this pull request, I optimized the implementations of blak256, blake2s, blake2b, groestl and AES in terms of flash size.

|       Name | Size before | Size after | Throughput before | Throughput after | Usage                                                                  |
|-----------:|------------:|-----------:|------------------:|-----------------:|------------------------------------------------------------------------|
|   blake256 |      2329 B |     1938 B |          1.5 MB/s |        0.59 MB/s | decred                                                                 |
|    blake2b |     15704 B |     2336 B |          2.3 MB/s |         1.2 MB/s | tezos, zcash, cardano                                                  |
|    blake2s |      4984 B |     1804 B |          3.1 MB/s |        0.96 MB/s | [firmware hash](https://github.com/trezor/trezor-firmware/blob/main/core/embed/extmod/modtrezorutils/modtrezorutils.c#L156), [bootloader check](https://github.com/trezor/trezor-firmware/blob/main/core/embed/firmware/bl_check.c#L152), [vendor header check](https://github.com/trezor/trezor-firmware/blob/main/core/embed/lib/image.c#L193) |in firmware |
| groestl512 |      9404 B |     6088 B |         0.16 MB/s |        0.13 MB/s | groestlcoin                                                            |
|        AES |     39964 B |     9538 B |                   |                  |                                                                        |
| AES128 ECB |             |            |         0.94 MB/s |        0.80 MB/s |                                                                        |
| AES128 CCM |             |            |        0.46 MB/s  |        0.43 MB/s | optiga secure channel                                                  |
| AES128 GCM |             |            |         0.31 MB/s |        0.30 MB/s | probably in THP                                                        |
|      ripemd160  |   5788 B | 1792 B | 5.5 MB/s | 1.3 MB/s | bitcoin, eos, ripple, nem |

 It's strange that it is possible to reduce the size of AES by half with only minimal impact on its speed.

There are still two places that would make sense to optimize, but I haven't addressed them yet:
  *  groestl512 that uses precomputed tables of size 4 KB
  *  ripemd160 that has 5 KB

EDITED: I fixed the "Size before" of AES, which should have been 20480 B higher.


onvej-sl:
My suggestion is 
  * to use the memory-optimized versions of blake2b, AES and groestl512
  * keep the existing (speed-optimized) version of blake2s
 
I don't have a strong opinion on blake256, as it only saves a few hundred bytes. I'm not sure if it is worth diverging from the original implementation.

What do you think @andrewkozlik , @matejcik ?

TychoVrahe:
+1 for keeping blake2s speed optimized, at least on F4 based models.

matejcik:
> My suggestion is
> 
> * to use the memory-optimized versions of blake2b, AES and groestl512
> * keep the existing (speed-optimized) version of blake2s

agreed

> I don't have a strong opinion on blake256, as it only saves a few hundred bytes. I'm not sure if it is worth diverging from the original implementation.

let's keep the original then.

please go ahead and modify the PR accordingly.

crypto/aes/aesopt.h
```diff
@@ -295,26 +295,26 @@ Issue Date: 20/12/2007
     to be set independently for encryption and decryption
 */
 #if 1
-#  define ENC_UNROLL  FULL
+#  define ENC_UNROLL  NONE
```
prusnak, line 298: I think here it is meant to change `1` to `0` above instead of swapping the definitions. Creates smaller diff with the upstream.

prusnak (dismissed):

onvej-sl:
There are two things I'm considering:
1. In `groestl.c` and `blake2b.c`, I'm using `#ifdef` directives to switch between the original speed-optimized implementation and the new memory-optimized implementation. I'm considering whether to keep the `#ifdef` directives or to use the speed-optimized implementation directly.
2. `blake2b.c` and `groestl.c` are currently excluded from style formatting. However, I don't see a reason for this. The upstream implementations haven't changed for more than eight years. On the contrary, we occasionally make modifications to these implementations. Therefore, I'm considering including them in style formatting.

Any opinion, @andrewkozlik ?

andrewkozlik:
1. I am inclined to keeping the `#ifdef` directives, including for blake2s, e.g. say we are running out of room in the boarloader and want to use the smaller variant of blake2s, then it will come in handy.

2.
    > On the contrary, we occasionally make modifications to these implementations.
    
    Until now our modifications were only variable initialization and `memzero()`. However, the changes in this PR are significant, making it difficult to keep in sync with upstream anyway if there were changes. So I agree about applying formatting. I'd consider doing it as a single commit in https://github.com/trezor/trezor-firmware/pull/4073 to keep the number of reformatting commits low.

onvej-sl:
> Any update on ripemd160?

I didn't succeed in optimizing it myself. However a found [this](https://github.com/sipa/Coin25519/blob/master/src/crypto/ripemd160.c) implementation which is 1792 B in size and has decent speed. I added all the values to the table above.

crypto/ripemd160.c
```diff
@@ -1,343 +1,344 @@
...
+    T = self->h[1] + CL + DR;
+    self->h[1] = self->h[2] + DL + ER;
+    self->h[2] = self->h[3] + EL + AR;
+    self->h[3] = self->h[4] + AL + BR;
+    self->h[4] = self->h[0] + BL + CR;
+    self->h[0] = T;
+
+    /* Clear the buffer and wipe the temporary variables */
+    T = AL = BL = CL = DL = EL = AR = BR = CR = DR = ER = 0;
+    memset(&self->buf, 0, sizeof(self->buf));
```
andrewkozlik, line 272: Let's use `memzero()` here and anywhere else where sensitive data may be left lying around the stack.

crypto/ripemd160.c
```diff
@@ -1,343 +1,344 @@
...
 
-    left = ctx->total[0] & 0x3F;
-    fill = RIPEMD160_BLOCK_LENGTH - left;
+/* The RIPEMD160 compression function.  Operates on self->buf */
+static void ripemd160_compress(ripemd160_state *self)
+{
+    uint8_t w, round;
+    uint32_t T;
+    uint32_t AL, BL, CL, DL, EL;    /* left line */
+    uint32_t AR, BR, CR, DR, ER;    /* right line */
```
andrewkozlik, line 190: Let's initialize variables in the declaration, like we do in the rest of our code base.

crypto/ripemd160.c
```diff
@@ -1,343 +1,344 @@
...
+#endif
+    self->bufpos = 64;
+    ripemd160_compress(self);
 
-    memzero(ctx, sizeof(RIPEMD160_CTX));
+    /* Copy the final state into the output buffer */
+#ifdef PCT_BIG_ENDIAN
+    byteswap_digest(self->h);
+#endif
+    memcpy(out, &self->h, RIPEMD160_DIGEST_SIZE);
```
andrewkozlik, line 335: Let's `memzero()` the context when done, like we do in the SHA implementations. It looks like we are not doing it in blake though.

andrewkozlik:
As I mentioned earlier, I think it would be good have the `#ifdef` for blake2s as well. Firstly, it would be good to keep the blake2b and blake2s implementations consistent. Secondly, in the bootloader we may need to use the smaller blake2s version in the future.

That means we will need to add `OPTIMIZE_SIZE_BLAKE2S`, `OPTIMIZE_SIZE_BLAKE2B`, `OPTIMIZE_SIZE_ED25519`, etc., which is a good thing, because it explicitly documents what we are doing, whereas right now it is not obvious that blake2s is not size-optimized. Furthermore it documents what speed/size tradeoffs are available without having to investigate the source code. The definition in `options.h` should default to `OPTIMIZE_SIZE` when undefined to make it easy to use:
```
#ifndef OPTIMIZE_SIZE_BLAKE2S
#define OPTIMIZE_SIZE_BLAKE2S OPTIMIZE_SIZE
#endif

#ifndef OPTIMIZE_SIZE_BLAKE2B
#define OPTIMIZE_SIZE_BLAKE2B OPTIMIZE_SIZE
#endif

// ...
```

crypto/groestl.c
andrewkozlik, line 1: Also, as I mentioned, in `groestl.c` I wouldn't mind getting rid of the speed-optimized version, because it makes such little difference that it just doesn't seem worth having the extra code to maintain. WDYT?

crypto/blake2b.c
```diff
@@ -186,6 +187,30 @@ int blake2b_InitKey( blake2b_state *S, size_t outlen, const void *key, size_t ke
...
+static void G(uint64_t *m, int r, int i, uint64_t *a, uint64_t *b, uint64_t *c,
+               uint64_t *d) {
+  *a = *a + *b + m[blake2b_sigma[r][2 * i + 0]];
+  *d = rotr64(*d ^ *a, 32);
+  *c = *c + *d;
+  *b = rotr64(*b ^ *c, 24);
+  *a = *a + *b + m[blake2b_sigma[r][2 * i + 1]];
+  *d = rotr64(*d ^ *a, 16);
+  *c = *c + *d;
+  *b = rotr64(*b ^ *c, 63);
```
andrewkozlik, line 200: It would be nice if we could reuse the code from the macro, e.g. by just calling `G(r,i,*a,*b,*c,*d)` here and renaming the function to `g` so that the names don't clash. In that case `m` would need to be added as a parameter to the `G` macro. It would also be nice to do the same for the `ROUND` macro and function. Maybe introduces more complexity than I think. WDYT?

onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4107/commits/07377d792d8692b8a4fd29c36fdf562310c34a76.

crypto/groestl.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4107/commits/704bc4f1dc9e804e3e9c095e3dab1d610a7636e6.

crypto/ripemd160.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4107/commits/704bc4f1dc9e804e3e9c095e3dab1d610a7636e6.

crypto/ripemd160.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4107/commits/704bc4f1dc9e804e3e9c095e3dab1d610a7636e6.

crypto/ripemd160.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4107/commits/704bc4f1dc9e804e3e9c095e3dab1d610a7636e6.

crypto/aes/aesopt.h
onvej-sl: Done (after force-push) in https://github.com/trezor/trezor-firmware/pull/4107/commits/3d2c623d80e6b6595dcc028cfdb2c5b8dd9a1b08.

M1nd3r (approved):
I have reviewed groestl and ripemd160, looks good.

crypto/ripemd160.h
```diff
@@ -2,21 +2,23 @@
 #define __RIPEMD160_H__
 
 #include <stdint.h>
+#include <stddef.h>
 
-#define RIPEMD160_BLOCK_LENGTH 64
```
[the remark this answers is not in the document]
onvej-sl: No rationale. Fixed in https://github.com/trezor/trezor-firmware/pull/4107/commits/7b9d67661a82fe5500354c6b5ce8c2a6f622942b.

onvej-sl:
@prusnak Your suggestion was addressed in https://github.com/trezor/trezor-firmware/pull/4107/commits/d707e54333d90b514548fa8c26e375c307d548f7. I think I cannot merge until you confirm it was addressed.

PR #4110, opened 2024-08-17
Improve behavior when going to previous word during recovery

### LINKED ISSUE #3859: Persistent words in recovery
**Describe the bug**
When a word (e.g. 3rd) is deleted and user goes to the previous word (2nd), change the 2nd word and go to the next word (3rd) the word is present there.

Core

**To Reproduce**
1. type 3rd word in recovery and confirm it
2. Delete the word and go to 2nd word, change it and confirm it
3. 3rd word is be present

**Expected behavior**
Once the word is deleted it shouldn't appear again.

obrusvit:
This PR improves behavior when going to previous word during recovery process. It

- runs the check for all keys in the keyboard if their press cannot lead to a valid word if the keyboard is initialized with a pre-filled word. Typically it means that all keys are disabled when going to previous word (always for SLIP39, most of the time for BIP39)
- deletes the word if the user goes back, related discussion: https://github.com/trezor/trezor-firmware/issues/3859#issuecomment-2294877348 
- remove the option to go back by swipe movement for TS5

fixes #3859


Hannsek:
> runs the check for all keys in the keyboard if their press cannot lead to a valid word if the keyboard is initialized with a pre-filled word. Typically it means that all keys are disabled when going to previous word (always for SLIP39, most of the time for BIP39)

Does that mean that if no valid word is possible, the key is disabled?

obrusvit:
> Does that mean that if no valid word is possible, the key is disabled?

@Hannsek when you type letters in mnemonic keyboard, the keys which do not lead to a valid word are disabled. This behavior was not evaluated when you returned to a previous word. This commit fixes that.

obrusvit:
Diff when going from `3rd` word to `2nd` word on T3T1
image

mmilata (approved):
utACK

PR #4113, opened 2024-08-20
Fix prodtest build with NEW_RENDERING on

cepetr:
This pull request fixes the build of **prodtest** with the `NEW_RENDERING` option enabled.

- Fixes build scripts for prodtest.
- Adds a new function - `gfx_draw_qrcode()` - for drawing QR codes.


core/embed/lib/gfx_draw.c
```diff
@@ -241,6 +241,53 @@ void gfx_draw_text(gfx_offset_t pos, const char* text, size_t maxlen,
...
+  }
+
+  // Calculate border size (1 extra modules around the QR code)
+  int border_side = ((side + 2) * scale);
+
+  // Calculate border left-top corner
+  x -= border_side / 2;
+  y -= border_side / 2;
+
+  // Draw border around the QR code
```
TychoVrahe, line 269: IIUC this doesn't just draw the border but fills the entire area with white color, which is only then filled with black where at appropriate spots. So the comment is misleading.

core/SConscript.prodtest
```diff
@@ -43,7 +47,7 @@ if TREZOR_MODEL in ('1', 'R', 'T3B1'):
     FONT_NORMAL_UPPER=None
     FONT_BOLD_UPPER=None
     FONT_SUB=None
-elif TREZOR_MODEL in ('T',):
+elif TREZOR_MODEL in ('T','DISC2',):
```
TychoVrahe, line 50: Pointless addition as the DISC2 build of prodtest is disabled above. We might actually want to enable it as i don't think there are reasons no to do that, but currently this won't work for other reason: because prodtest is using legacy drawing API, which doesn't work with 32bit colors DISC2 uses. 

We can keep though it as it doesn't really matter, apart from making impression that its supported while its not.

core/embed/lib/gfx_draw.c
```diff
@@ -241,6 +241,53 @@ void gfx_draw_text(gfx_offset_t pos, const char* text, size_t maxlen,
...
+  gfx_rect_t border_rect = gfx_rect_wh(x, y, border_side, border_side);
+  gfx_draw_bar(border_rect, COLOR_WHITE);
```
TychoVrahe, line 271: Taking COLOR_WHITE from colors.h - where it is only defined for 16bit color depths. We should make this work for 32bit colors too.

core/embed/lib/gfx_draw.c
```diff
@@ -241,6 +241,53 @@ void gfx_draw_text(gfx_offset_t pos, const char* text, size_t maxlen,
...
+  x += scale;
+  y += scale;
+
+  // Draw black modules
+  for (int i = 0; i < side; i++) {
+    for (int j = 0; j < side; j++) {
+      if (qrcodegen_getModule(codedata, i, j)) {
+        gfx_rect_t rect =
+            gfx_rect_wh(x + i * scale, y + j * scale, scale, scale);
+        gfx_draw_bar(rect, COLOR_BLACK);
```
TychoVrahe, line 283: Also here - lets support 32 bit colors (black works even now because its all zeroes anyway)

TychoVrahe (commented):
Adding few comments mainly related to the fact that old rendering doesn't support 32bit colors.

core/embed/lib/gfx_draw.c
cepetr: fixed here c1d9151b17c3b20b9f9b508148e44f8f98503eb9

core/SConscript.prodtest
cepetr: I just tested it on DISC2 (together with npm1300 driver) and forgot to delete it. It's fixed here c1d9151b17c3b20b9f9b508148e44f8f98503eb9

core/embed/lib/gfx_draw.c
cepetr: Yes that problem, but how about leaving the `COLOR_WHITE/BLACK` issue until we remove the original display driver? Once `dma2d.h/.c` is gone, `colors.h/.c` will go with it. Then we can move those constants to `gfx_color.h` and have there both 16-bit and 32-bit versions.

TychoVrahe: Ok we can postpone this.

In case we need to support 32bit colors here sooner than we can drop old rendering we will revisit this.

TychoVrahe (approved):

PR #4117, opened 2024-08-22
fix(core/ethereum): use BIP44 variant to detect ETH account #

### LINKED ISSUE #3627: Account under "i" in receive flow is always 'ETH #1', regardless of derivation path
**Describe the bug**
Account is always 'ETH #1', regardless of derivation path

`2.7.0`

image
image

ibz:
Fixes #3627


Hannsek:
@ibz "Refs" doesn't work when linking issues. You have to use one of [these](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

obrusvit (approved):
utACK

However, there's no active UI test for Account info in ETH.

matejcik (approved):

PR #4118, opened 2024-08-22
Drop spaces before punctuation in fr

ibz:
See this: https://github.com/trezor/trezor-firmware/pull/4095/files#r1721277469


core/mocks/trezortranslate_keys.pyi
```diff
@@ -144,7 +144,7 @@ class TR:
     cardano__collateral_input_index: str = "Collateral input index:"
     cardano__collateral_output_contains_tokens: str = "The collateral return output contains tokens."
     cardano__collateral_return: str = "Collateral return"
-    cardano__confirm: str = "Confirm:"
+    cardano__confirm: str = ""
```
prusnak, line 147: why not remove these completely?

ibz: I think there were issues last time I tried to remove strings due to indexes of the other strings not matching anymore...

obrusvit: @ibz To remove a string:
- delete it from lang JSONs (`en.json`, `cs.json`, ...) 
- keep the key in `order.json`
- regenerating should keep the entry in rust enum (with empty `""`) and you'll see it removed from `mocks/trezortranslate-keys.pyi`

Can you try?

ibz: @obrusvit You are right, I remember now!

core/translations/fr.json
```diff
@@ -276,7 +275,7 @@
     "coinjoin__title_progress": "Coinjoin en cours",
     "coinjoin__waiting_for_others": "En attente des autres",
     "confirm_total__fee_rate": "Taux des frais",
-    "confirm_total__fee_rate_colon": "Taux des frais :",
+    "confirm_total__fee_rate_colon": "Taux des frais :",
```
prusnak, line 278: this should be also removed I suppose

prusnak:
This PR should remove all 4 strings, right? Because now it removes only some of these:
* address_details__derivation_path_colon
* confirm_total__fee_rate_colon
* send__total_amount_colon
* words__account_colon

core/translations/fr.json
ibz: Ideally yes, but this is used from Rust and we don't currently seem to have a way to format strings there.

prusnak:
I am wondering ... do we really need to use the template?

Shouldn't we just do `str + ":"` (probably wrapped in a nice macro) which would work both for Rust and Python?

I understand this gives us less complexity but French can live without an extra space and are right-to-left languages on the immediate roadmap?

core/translations/fr.json
ibz: We use heapless. All allocated on the stack. Very limited functionality. Not saying it is not doable, but so far I did not see any string formatting in our Rust code.

Hannsek:
Are we talking about the space before `:`? If yes, I don't think it is any issue. We just need to tell in #wg_translations,

core/translations/order.json
```diff
@@ -958,5 +958,6 @@
   "956": "words__title_done",
   "957": "reset__slip39_checklist_more_info_threshold",
   "958": "reset__slip39_checklist_more_info_threshold_example_template",
-  "959": "passphrase__continue_with_empty_passphrase"
+  "959": "passphrase__continue_with_empty_passphrase",
+  "960": "label_with_colon_template"
```
mmilata, line 962: This one doesn't seem to be used anywhere?

core/translations/fr.json
```diff
@@ -52,52 +52,52 @@
     "address__title_yours": "La vôtre",
     "address_details__account_info": "Infos sur le compte",
     "address_details__derivation_path": "Chemin de dérivation",
-    "address_details__derivation_path_colon": "Chemin de dérivation :",
+    "address_details__derivation_path_colon": "Chemin de dérivation:",
     "address_details__title_receive_address": "Adr. de récep.",
     "address_details__title_receiving_to": "Récep. sur",
-    "authenticate__confirm_template": "Autoriser l'ord. connecté à conf. que votre {0} est auth. ?",
+    "authenticate__confirm_template": "Autoriser l'ord. connecté à conf. que votre {0} est auth.?",
```
mmilata, line 58: Do we really want to do this for `?` too even if it doesn't buy us anything?

Btw [wikipedia](https://en.wikipedia.org/wiki/Punctuation#Other_languages) claims this about canadian french:
> In the [French](https://en.wikipedia.org/wiki/French_language) of [France](https://en.wikipedia.org/wiki/France) and [Belgium](https://en.wikipedia.org/wiki/Belgium), the marks ⟨:⟩, ⟨;⟩, ⟨?⟩ and ⟨!⟩ are preceded by a thin space. In [Canadian French](https://en.wikipedia.org/wiki/Canadian_French), this is only the case for ⟨:⟩.

core/translations/order.json
ibz: Yep, I removed it.

core/translations/fr.json
ibz: I think it can be argued both ways. What it buys us is consistency with other languages, which was the whole point of this PR. Although I also agree that in the case of `?` it might not matter, since we will probably not have cases of strings with or without `?` that would need to be added from code?

Anyway, your decision. If you think I should keep the space in this case, I can definitely go that way. I don't have a strong preference for either way...

mmilata: Code consistency is nice but I suppose user experience is the most important. French people are proud of their culture and we need to do everything we can to compete with Ledger :sweat_smile: 

> Although I also agree that in the case of ? it might not matter, since we will probably not have cases of strings with or without ? that would need to be added from code?

I think that's the distinction, with `:` strings are often constructed dynamically (maybe even mixed with others that already have `:` at the end) while `?` and `!` are often at the end of whole sentence. So my suggestion is to use the `:` handling from this PR but keep the spaces before `?` and `!` in fr.json.

ibz: > So my suggestion is to use the `:` handling from this PR but keep the spaces before `?` and `!` in fr.json.

https://github.com/trezor/trezor-firmware/pull/4118/commits/c4665b8a0e82f2e8761e339e9aa88de399d7fb27

core/translations/fr.json
mmilata: FWIW we have [ufmt](https://docs.rs/ufmt/latest/ufmt/) but it requires the formatting string to be known at compile-time.

This could be removed by formatting the string in python before passing it as a new parameter to the layout.

mmilata (approved):

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.3 regular (revision 7f373ae71eca855dd41b4a0fdcc7cadaa13a8281)

PR #4121, opened 2024-08-23
switch to new rendering everywhere

TychoVrahe:
This switches all targets to new rendering.

If we want to release bootloader with progressive renderer, we will need to switch that back until the  progressive renederer is battle tested in firmware. But i switch here too so it gets used more internally.


TychoVrahe:
@Hannsek pls go over the UI changes and see if there is something that you would not accept.

cepetr (approved):
utACK

cepetr (approved):
I reviewed the last commit, and everything looks good, so I'm re-approving it.

core/SConscript.firmware
```diff
@@ -15,7 +15,7 @@ PYOPT = ARGUMENTS.get('PYOPT', '1')
 DISABLE_OPTIGA = ARGUMENTS.get('DISABLE_OPTIGA', '0') == '1'
 HW_REVISION = ARGUMENTS.get('HW_REVISION', None)
 THP = ARGUMENTS.get('THP', '0') == '1' # Trezor-Host Protocol
-NEW_RENDERING = ARGUMENTS.get('NEW_RENDERING', '1') == '1' or TREZOR_MODEL in ('T3T1',)
+NEW_RENDERING = ARGUMENTS.get('NEW_RENDERING', '1') == '1'
```
matejcik, line 18: we should be dropping this as an argument, no? we want to go new rendering all the way for firmware, the only holdout is TT bootloader

core/embed/rust/src/ui/model_tr/theme/mod.rs
```diff
@@ -71,49 +71,68 @@ include_icon!(ICON_ARROW_LEFT, "model_tr/res/arrow_left.toif"); // 4*7
...
+#[cfg(not(feature = "new_rendering"))]
 include_icon!(
     ICON_ARROW_UP,
     "model_tr/res/arrow_up.toif",
     empty_right_col = true
 ); // 8*4
+#[cfg(feature = "new_rendering")]
+include_icon!(ICON_ARROW_UP, "model_tr/res/arrow_up_o.toif"); // 8*4
+#[cfg(not(feature = "new_rendering"))]
 include_icon!(
```
matejcik, line 83: ...and as such I see no reason to add _new_ icons, just replace the old ones

core/embed/rust/src/ui/model_tr/component/changing_text.rs
```diff
@@ -201,7 +220,7 @@ impl Component for ChangingTextLine {
         if self.show_content {
             // In the case text cannot fit, show ellipsis and its right part
             if !self.text_fits_completely() {
-                self.paint_long_content_with_ellipsis();
+                self.render_long_content_with_ellipsis(target);
```
matejcik, line 223: ideally we also drop the `paint_*` versions where we can?

core/SConscript.unix
```diff
@@ -10,7 +10,7 @@ TREZOR_MODEL = ARGUMENTS.get('TREZOR_MODEL', 'T')
 CMAKELISTS = int(ARGUMENTS.get('CMAKELISTS', 0))
 HW_REVISION ='emulator'
 THP = ARGUMENTS.get('THP', '0') == '1' # Trezor-Host Protocol
-NEW_RENDERING = ARGUMENTS.get('NEW_RENDERING', '1') == '1' or TREZOR_MODEL in ('T3T1',)
+NEW_RENDERING = ARGUMENTS.get('NEW_RENDERING', '1') == '1'
```
matejcik, line 13: we also don't need it here

core/SConscript.firmware
TychoVrahe: I intended this as a switch only, tried to not remove any old rendering related functionality for now. 

Obviously we should do clean-up and remove it all at some point, but i think its fine to postpone that a bit, so we still can switch for comparison when debugging etc.

Dtto for the comments below.

bosomt:
**QA OK**

**Info:**
 - Suite version: web 24.11.0 (65a8499db01e2226de44c748e725e0bf976b3ee5)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.8.4 regular (revision d71d9e9c343ad04a7f28d6b02f8d3f8945af30d6)
 - Transport: WebUsbTransport

PR #4125, opened 2024-08-26
Create an enum for valid device rotations

### LINKED ISSUE #4041: Create an enum for valid device rotations
https://github.com/trezor/trezor-suite/issues/13360

this can also make our code slightly nicer in places

bieleluk:
fixes #4041


core/src/apps/management/apply_settings.py
```diff
@@ -167,17 +167,17 @@ async def _require_confirm_change_passphrase_source(
...
         label = TR.rotation__east
-    elif rotation == 180:
+    elif rotation == DisplayRotation.South:
         label = TR.rotation__south
-    elif rotation == 270:
+    elif rotation == DisplayRotation.West:
         label = TR.rotation__west
     else:
-        raise DataError("Unsupported display rotation")
+        raise ValueError(f"Unsupported display rotation: {rotation}")
```
matejcik, line 180: use a `RuntimeError` here, this cannot happen because your if/elif is exhaustive

also don't use generic errors with texts https://docs.trezor.io/trezor-firmware/core/misc/exceptions.html

dtto below

core/src/storage/device.py
```diff
@@ -3,6 +3,7 @@
 
 from storage import common
 from trezor import utils
+from trezor.enums import DisplayRotation
```
matejcik, line 6: please move this import to (a) TYPE_CHECKING block and (b) again locally inside the function

(bla bla generated bytecode + dependencies concerns)

core/src/storage/device.py
```diff
@@ -95,16 +96,27 @@ def get_device_id() -> str:
...
+    if value == 0:
+        rotation = DisplayRotation.North
+    elif value == 90:
+        rotation = DisplayRotation.East
+    elif value == 180:
+        rotation = DisplayRotation.South
+    elif value == 270:
+        rotation = DisplayRotation.West
+    else:
+        raise ValueError(f"Unsupported display rotation: {rotation}")
```
matejcik, line 114: in this case, when loading an invalid value from storage, we don't want to crash but instead use the default

core/src/apps/management/apply_settings.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4125/commits/677d33afba60e049a84e90778b7ce6df69fcfa7b

core/src/storage/device.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4125/commits/5d07343078852d6572a53026f2a3f683e3dddc43

matejcik (approved):

bieleluk:
This PR creates and utilizes an enum for valid display rotations 

Changes include:

1. Creating protobuf enum `DisplayRotation` with all possible display rotations
2. Updating generated protobuf messages
3. Replacing integer values with `DisplayRotation` enums in display rotation functions in `core` and `trezorlib` modules

matejcik (approved):

Hannsek:
Do we have to tell Connect guys? (Because of protobuf messages)

matejcik:
this change should be 100% backwards compatible but yeah let's tell them, they will be glad :)

PR #4126, opened 2024-08-26
update display initialization sequence on T3T1

TychoVrahe:
According to instructions from display manufacturer.


cepetr (approved):
I haven't tested it, but it looks good.

PR #4130, opened 2024-08-26
add bootloader hashes to gen and gen_check

TychoVrahe:
- first, some refactoring is done to move the bootloader binaries and hashes to model-specific folders, the hashes are separated from the check code itself.
- the bootloader hashes generation is added to make gen and make gen_check targets to have better assurances that the files and hashes are in sync. 
- bootloader binaries for Model 1 are removed from core


matejcik:
LGTM, refactored bootloader_hashes a little to detect that the bootloaders are well formed and that the model matches expectations. Please review [7134399](https://github.com/trezor/trezor-firmware/pull/4130/commits/7134399fe8266584ce9ca3dc1e5c6b9f811a07c3)

TychoVrahe:
https://github.com/trezor/trezor-firmware/pull/4130/commits/7134399fe8266584ce9ca3dc1e5c6b9f811a07c3 looks good, thanks

matejcik (approved):

PR #4132, opened 2024-08-27
fix(core) Fix auto-mover hitting wall scenario

### LINKED ISSUE #3692: T2B1: Wrong behavior when auto-mover hits a wall

bieleluk:
fixes #3692

### DIFF
--- a/core/.changelog.d/3692.fixed
+++ b/core/.changelog.d/3692.fixed
@@ -1,0 +1,1 @@
+Fix auto-mover hitting wall scenario
--- a/core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
+++ b/core/embed/rust/src/ui/model_tr/component/input_methods/choice.rs
@@ -425,6 +425,15 @@ where
         &self.choices
     }
 
+    /// Check possibility of going left/right.
+    fn can_move(&self, button: ButtonPos) -> bool {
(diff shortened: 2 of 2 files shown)


matejcik (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.1.0 (0b116799334cb4a05fcb59cc480deb7c4b9a63e9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.1.0 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3B1 2.8.7 regular (revision e196413bb745443610211c325623ba7cee7309e9)

PR #4133, opened 2024-08-27
port unlock-bootloader to STM32U5 models

TychoVrahe:
This PR implements unlock-bootloader functionality for U5 models with optiga - T3B1 and T3T1.

Firmware downgrade protection is also implemented via OTP (F4) or HDP (U5). Monotonic counters are refactored to allow easier implementation of this protection.

Monotonic versions are moved to model-specific folders as they differ across models, unlike regular version. Also having single monotonic version for all firmware variants (prodtest/firmware/reflash) is desirable. 

T3T1 and T3B1 monotionic versions are inceremented.


core/embed/trezorhal/stm32f4/secret.c
```diff
@@ -98,6 +98,10 @@ secbool secret_optiga_get(uint8_t dest[SECRET_OPTIGA_KEY_LEN]) {
   return secret_read(dest, SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
 }
 
+secbool secret_optiga_present(void) { return secret_wiped(); }
```
cepetr, line 101: This function should return inverted value of `secret_wiped()`.

core/embed/bootloader/version_check.c
```diff
@@ -0,0 +1,49 @@
...
+#include <string.h>
+
+#include "version_check.h"
+
+#include "error_handling.h"
+#include "model_version.h"
+#include "monoctr.h"
+
+// protection against bootloader downgrade
+void check_bootloader_version(void) {
```
cepetr, line 29: How about renaming this function to `ensure_bootloader_version()`, `ensure_and_check_bootloader_version()` or similar ? The verb `check` implies no side effects, like writing the bootloader version to OTP/flash memory, which confuses me a bit.

cepetr (changes requested):

core/embed/boardloader/main.c
```diff
@@ -327,7 +297,6 @@ int main(void) {
        check_image_contents(old_hdr, IMAGE_HEADER_SIZE, &BOOTLOADER_AREA))) {
     write_bootloader_min_version(old_hdr->monotonic);
```
cepetr, line 298: Maybe we should avoid updating the monotonic counter here and instead do it later, just before jumping to the bootloader, when the bootloader has been fully checked. Updating the counter prematurely could potentially prevent downgrading to a previous (but functional) version.

cepetr (changes requested):

core/embed/trezorhal/stm32f4/secret.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/130162bfe30316f3605852066170800b2ad83b12

core/embed/bootloader/version_check.c
TychoVrahe: renamed as part of https://github.com/trezor/trezor-firmware/commit/905b1005146edfe3d8d42b0dfee80c032813981f

core/embed/bootloader/main.c
```diff
@@ -312,6 +286,9 @@ void real_jump_to_firmware(void) {
 
   ensure(check_image_model(hdr), "Wrong firmware model");
 
+  ensure(check_firmware_version(hdr->version, sectrue),
```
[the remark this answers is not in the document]
TychoVrahe: Splat the function in https://github.com/trezor/trezor-firmware/commit/905b1005146edfe3d8d42b0dfee80c032813981f

Regarding the placement changes in https://github.com/trezor/trezor-firmware/commit/8eb2f4f469aa53a3b47b5546cd8f85a295272a40:
I moved the update call after the signature check and only if sig is valid, but before contents check. Reason being that i want the counter updated in case user manages to upload the header correctly, but the firmware update process  is interrupted. cc @matejcik . Once the header is uploaded, user is stripped away from choice of uploading previous firmware version.

The check in main is done before signature check, as i preferred the device with "downgraded" fw to appear empty, as we do will i.e. model checks etc. (placing after the sig check would cause firmware appear as corrupted. Or we would have to add another UI logic which i am trying to avoid). The placement in real_jump_to_firmware mimics the placement in main.

core/embed/boardloader/main.c
TychoVrahe: This update is done with version of currently installed bootloader (not the one being installed from SD card. The checks (both sig and contents) are done before updating. 

I need to do that here - if bootloader has just been updated from fw, user has inserted an SD card with older version and reboots - then not updating the counter here would lead to possibility of downgrade below the currently installed version.

core/embed/models/T2B1/versions.h
```diff
@@ -0,0 +1,3 @@
+
+#define BOOTLOADER_MONOTONIC_VERSION 2
+#define FIRMWARE_MONOTONIC_VERSION 1
```
matejcik, line 3: aren't we supposed to bump T2T1 and T2B1 too?

TychoVrahe: bumped: https://github.com/trezor/trezor-firmware/commit/965280f0f55d7c562464983aaf2eb45387d9002c

core/embed/bootloader/main.c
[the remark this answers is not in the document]
cepetr: Okay, as we discussed in person, the main problem - updating the monotonic counter with a value from an unreliable source - has been fixed. So, I'm okay with it now.

core/embed/boardloader/main.c
cepetr: Okay, I see it, thanks for explanation.

matejcik (changes requested):

core/embed/bootloader/messages.c
```diff
@@ -574,6 +579,14 @@ int process_msg_FirmwareUpload(uint8_t iface_num, uint32_t msg_size,
         return UPLOAD_ERR_INVALID_IMAGE_HEADER_SIG;
       }
 
+      if (sectrue != check_firmware_min_version(received_hdr->monotonic)) {
+        MSG_SEND_INIT(Failure);
+        MSG_SEND_ASSIGN_VALUE(code, FailureType_Failure_ProcessError);
+        MSG_SEND_ASSIGN_STRING(message, "Illegal firmware downgrade");
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/bb2c8a6b94b3f2925e1a25e4c332c475f9cae0ed

core/embed/bootloader/main.c
```diff
@@ -459,12 +439,18 @@ int bootloader_main(void) {
...
   }
+
   if (sectrue == model_ok) {
+    version_ok = check_firmware_min_version(hdr->monotonic);
+  }
+
+  if (sectrue == version_ok) {
     header_present =
         check_image_header_sig(hdr, vhdr.vsig_m, vhdr.vsig_n, vhdr.vpub);
   }
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/39bad5996daeffc1b43e8dcaaded6bda6c9cb728

core/embed/bootloader/main.c
```diff
@@ -312,9 +286,14 @@ void real_jump_to_firmware(void) {
...
   ensure(check_image_model(hdr), "Wrong firmware model");
 
+  ensure(check_firmware_min_version(hdr->version),
+         "Firmware downgrade protection");
+
   ensure(check_image_header_sig(hdr, vhdr.vsig_m, vhdr.vsig_n, vhdr.vpub),
          "Firmware is corrupted");
 
+  update_firmware_min_version(hdr->version);
+
```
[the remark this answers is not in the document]
TychoVrahe: changed, no other function https://github.com/trezor/trezor-firmware/commit/39bad5996daeffc1b43e8dcaaded6bda6c9cb728

core/embed/bootloader/version_check.c
```diff
@@ -0,0 +1,52 @@
...
+}
+
+void update_firmware_min_version(uint8_t version) {
+  uint8_t min_version = 0;
+  ensure(monoctr_read(MONOCTR_FIRMWARE_VERSION, &min_version), "monoctr read");
+
+  if (version > min_version) {
+    ensure(monoctr_write(MONOCTR_FIRMWARE_VERSION, version), "monoctr write");
+  }
+}
```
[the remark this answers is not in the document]
TychoVrahe: That would not work properly on U5, where the counter is written in regular flash in secret area (via monoctr, secret and flash_area drivers), this would fail on `ensure` in the secret implementation as it would fail to write the lower version.

The `ensure_bootloader_min_version` is ok in this regard as its used exclusively on F4, where we writee in OTP that does not have these checks. But maybe we should be more explicit in this file that the bootloader function is only meant for F4.

Thus, if anything, i would change the `ensure_bootloader_min_version` to work like the fw one. But then again the original looks a bit more glitch resistant so i would go for keeping it this way.

core/embed/rust/src/ui/model_mercury/bootloader/mod.rs
```diff
@@ -300,11 +300,32 @@ impl UIFeaturesBootloader for ModelMercuryFeatures {
...
 
     fn screen_unlock_bootloader_confirm() -> u32 {
-        unimplemented!();
+        let title =
+            Label::left_aligned("UNLOCK BOOTLOADER".into(), TEXT_BOLD).vertically_centered();
+        let msg = Label::centered("This action cannot be undone!".into(), TEXT_NORMAL);
+
+        let right = Button::with_text("UNLOCK".into())
+            .styled(button_confirm())
+            .with_text_align(Alignment::Center);
```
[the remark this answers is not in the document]
TychoVrahe: Most of them should be centered, and have been fixed in https://github.com/trezor/trezor-firmware/pull/4134 . This branch just needs a rebase and everything will look beautiful :)

TychoVrahe:
Changed the welcome screen in https://github.com/trezor/trezor-firmware/pull/4140/commits/1e25014a1ce4c54c2d9559bebfa852045706ca38

TychoVrahe:
changed the welcome screen in https://github.com/trezor/trezor-firmware/pull/4140/commits/1e25014a1ce4c54c2d9559bebfa852045706ca38

core/embed/bootloader/version_check.h
```diff
@@ -23,15 +23,18 @@
...
 // required version, returns sectrue if check_version is higher or equal to the
 // stored version
 secbool check_firmware_min_version(uint8_t check_version);
 
-// Update the minimum firmware version stored in monotonic counter
-void update_firmware_min_version(uint8_t version);
\ No newline at end of file
+// Ensures firmware version is stored in monotonic counter
+// If the version cannot be written, the function will shutdown the device
+void ensure_firmware_min_version(uint8_t version);
```
matejcik, line 40: no newline at end of file -- surprised that this passes editor check...?

core/embed/bootloader/version_check.c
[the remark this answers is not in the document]
TychoVrahe: Modified the monoctr so that it returns secfalse in any case it cannot write the new value.

then, i changed the `update_firmware_min_version` to match the `ensure_bootloader_min_version` variant. I did remove the first ensure in that function, as writing would fail immediately now and later we have better error message.

see https://github.com/trezor/trezor-firmware/commit/12298d3c1e3acc9e272fef5006d60a73b4ed92eb and https://github.com/trezor/trezor-firmware/commit/e8b1dceac086565db773391d890f3916863284e4

core/embed/bootloader/version_check.h
TychoVrahe: it didn't https://github.com/trezor/trezor-firmware/pull/4133/commits/7583628226dbd6ccb186890124b400cc1ca28ed1

matejcik (approved):
lgtm now but @cepetr please go over the recent commits as well

cepetr (approved):
utACK

PR #4072, opened 2024-07-30
Add T3B1 defs, update Ethereum defs

komret:
Adding support for T3B1.

Updating Ethereum definitions and removing old testnets. This was consulted with @tomasklim.

PR #4136, opened 2024-08-28
Enable conditional branches for T3B1

matejcik:
in all cases where Python code checks for T2B1, the same should apply for T3B1.

complement to #4072


mmilata (approved):
Cool pylint plugin.

Wondering where else we should add `T3B1`. Do we want to build firmware images in CI? Emulators too?

matejcik:
let's discuss after standup

PR #4137, opened 2024-08-28
rust: Concrete renderers

matejcik:
This PR reworks the way we invoke display renderers, replacing closure-based `run_with_bumps` with a guard-based locking mechanism, and replacing `render_on_display` with a macro.

This significantly reduces the amount of trickery needed to get Rust to accept the various lifetime bounds (not to zero, unfortunately :( ), and -- perhaps more importantly -- it allows us to create a `Display` trait that is a parameter of `UIFeaturesCommon`, cleanly binding each model to a particular Display implementation and providing an obvious location to change it without relying on exclusive features.

(ideally this would also let us remove the `display_xxxx` features, because we can now _build_ all the impls at the same time, and select the one to be used via UIFeatures. @cepetr do you think this is viable?)

the drawback is that you can no longer freely build, say, model_mercury UI for the Trezor T hardware just by switching some feature flags, because ModelTTFeatures specify the NoFb implementation. But if this is desired, we can move the feature-flag selection into ModelTTFeatures (which is frankly a better place for it because model TT only works on color displays anyway)

Importantly, `Display` does not have lifetime parameters, which makes it very useful downstream -- e.g., it would be possible to keep `LayoutObj` object-safe even if parametrized by `Display`. (which is the endgame here -- get rid of all "selectors" that import different modules based on which model is selected, and so be able to e.g. run all rust tests for all models from a single binary)

---

additionally, f2baf8288a5ab3b3521ef16dcc773c9e5d84a69f fixes unsound API of `trezorhal::display::get_frame_buffer`


matejcik:
note: switch on "hide whitespace" to review `bootloader/mod.rs`

PR #4141, opened 2024-08-30
generate pipeline docs

TychoVrahe:
fixes failing gen_check.


vdovhanych (approved):

vdovhanych:
Btw we should probably get rid of this docs gen test and rewrite it, as Gitlab CI is no longer used for anything. cc @matejcik

matejcik:
yeah, sure. is anyone even looking at that document?

vdovhanych:
I don't think so, I'd say it's pretty useless in general.

PR #4142, opened 2024-08-30
fix(core/mercury): show remaining shares

obrusvit:
This commit enables showing "Remaining shares" from the context menu during recovery process. This is only shown during super-shamir.
emu00000000 emu00000001 emu00000002

TODO:
- [x] input flows and BRs misalignment - some tests might hang
- [ ] the Footer in Remaining shares window should work somewhat similar to the footer in chose credentials of FIDO2
    - Suggestion: merging this functionality without the Footer for now. Deploying the Footer later.


tests/input_flows_helpers.py
```diff
@@ -274,7 +274,7 @@ def warning_bip39_dryrun_mismatch(self) -> BRGeneratorType:
     def success_more_shares_needed(
         self, count_needed: int | None = None
     ) -> BRGeneratorType:
-        br = yield
+        br = yield  # FIXME now this yield hangs the emu
```
mmilata, line 277: still true?

obrusvit, line 277: FIXME comment should be removed.

core/src/trezor/ui/layouts/tt/recovery.py
```diff
@@ -10,6 +10,11 @@
 CONFIRMED = trezorui2.CONFIRMED  # global_import_cache
 INFO = trezorui2.INFO  # global_import_cache
 
+if TYPE_CHECKING:
+    from typing import List, Set, Tuple
```
mmilata, line 14: please use `list`, `set`, `tuple` etc ... the capitalized versions are compatible with older python versions which we don't need in firmware

mmilata (commented):
two nits but overall looking good

tests/input_flows_helpers.py
obrusvit: No, removed.
https://github.com/trezor/trezor-firmware/pull/4142/commits/3a03e8b2059ca96f7c3dacfae978ccd3ed24fb9f

core/src/trezor/ui/layouts/tt/recovery.py
obrusvit: fixed
https://github.com/trezor/trezor-firmware/pull/4142/commits/3a03e8b2059ca96f7c3dacfae978ccd3ed24fb9f

mmilata (approved):

bosomt:
**QA OK**

tedsed by recovery of super shamir /3 shares/

**Info:**
 - Suite version: desktop 24.9.1 (ce0f589a4863ec711ca3f0e5677accb2d0c06cd2)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.9.1 Chrome/124.0.6367.243 Electron/30.3.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: BridgeTransport 3.0.0

PR #4143, opened 2024-08-30
feat(core): use new ETH send flow also for staking

ibz:


core/src/trezor/ui/layouts/mercury/__init__.py
```diff
@@ -1066,36 +1074,36 @@ async def confirm_ethereum_staking_tx(
...
-            br_name=br_name,
-            br_code=br_code,
+        await raise_if_not_confirmed(
+            RustLayout(
+                trezorui2.flow_confirm_output(
+                    title=verb,
+                    subtitle=None,
+                    message=intro_question,
+                    amount=None,
+                    chunkify=False,
```
obrusvit, line 1091: Why is `chunkify` set to `False`?

core/src/trezor/ui/layouts/mercury/__init__.py
```diff
@@ -1066,36 +1074,36 @@ async def confirm_ethereum_staking_tx(
...
+                    br_code=br_code,
+                    br_name=br_name,
+                    address=address,
+                    address_title=address_title,
+                    summary_items=summary_items,
+                    fee_items=info_items,
+                    summary_title=verb,
+                    summary_br_name="confirm_total",
+                    summary_br_code=ButtonRequestType.SignTx,
+                    cancel_text=TR.buttons__cancel,  # cancel staking
```
obrusvit, line 1104: What is the comment about?

obrusvit (commented):
Tests with input flows for ETH signtx and staking are still skipped. Otherwise, it looks good. We should take a look at the parameter set of the flows, calling a function with 18 arguments does not seem good.

core/src/trezor/ui/layouts/mercury/__init__.py
obrusvit: Ok, I wasn't aware that we *always* chunkify in `mercury`.

core/src/trezor/ui/layouts/mercury/__init__.py
ibz: The title of the Cancel screen, according to Figma, should be "Cancel staking" not "Cancel" like the button, so we need to pass 2 separate parameters here, but I got stuck in trying to make the Rust compiler happy with the types, so I went for the compromise of just calling it "Cancel" all over.

ibz:
> We should take a look at the parameter set of the flows, calling a function with 18 arguments does not seem good.

I know, this flow grew too complex suddenly... It'll soon also make coffee for you after you click "Sign". I need to break it down into separate "versions".

core/src/trezor/ui/layouts/mercury/__init__.py
ibz: > Ok, I wasn't aware that we *always* chunkify in `mercury`.

I mean, we always chunkify the address on the (separate) address screen in mercury, because this is the only place we have that screen (it is the one I added now in this PR, so it is only here).

obrusvit (approved):
Please wait for CI - update `fixtures.json`
Run `make style`.

Hannsek:
I cannot see the diffs :/

ibz:
> I cannot see the diffs :/

They should be visible now. They were not before, because there were no tests enabled. After @obrusvit fixed the input flows and enabled the relevant tests, the screens appeared. But they are not really a diff, because they are new screens, so there is nothing to compare to...

Hannsek:
New screens? We had staking before. Sorry, but I am confused now. @obrusvit can you clarify please?

obrusvit:
> New screens? We had staking before. Sorry, but I am confused now. @obrusvit can you clarify please?

We didn't have UI tests for staking on TS5 since TS5 release until now. 
We had functional tests for the actual signing of the staking operations but the UI flow was tested manually only.

This PR enables the UI tests of staking for TS5 and implements them properly. That's why the screens are new.

Hannsek:
Perfect, thanks for clarifying.

bosomt:
**QA OK**

PR #4144, opened 2024-08-31
Use bitcoinlib to parse raw transaction when Blockbook server could not find it.

oren-z0:
I really like the `build_tx.py` tool, and I patched it to be able to sign a transaction that spends a utxo of a transaction that hasn't been sent to the blockchain yet.

In case the user wants to create two transactions, where the second spends the output of the first, he can do that without broadcasting them. This is useful in scenarios of relative timelocks (using nSequence).

The user can also specify "." as the blockbook host (instead of `btc1.trezor.io`) and then the script will never send http requests and instead will ask the user to provide the previous raw-transaction for every `vin`.

I couldn't find any existing code in this project that parses raw transactions, so I used: https://pypi.org/project/bitcoinlib (if `bitcoinlib` is not installed, the behavior remains the same).

### DIFF
--- a/python/tools/build_tx.py
+++ b/python/tools/build_tx.py
@@ -25,8 +25,16 @@ import requests
 
 from trezorlib import btc, messages, tools
 from trezorlib.cli import ChoiceType
-from trezorlib.cli.btc import INPUT_SCRIPTS, OUTPUT_SCRIPTS
+from trezorlib.cli.btc import messages, INPUT_SCRIPTS, OUTPUT_SCRIPTS
 from trezorlib.protobuf import to_dict
+
+try:
+    from bitcoinlib.transactions import Transaction
+    bitcoinlib_installed = True
+except ImportError:
+    Transaction = None
+    bitcoinlib_installed = False
+
 
 SESSION = requests.Session()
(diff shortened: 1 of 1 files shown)


matejcik:
we already use `python-bitcoinlib` in the firmware repository so it would be nice if you could use that too?
(yeah `bitcoinlib` and `python-bitcoinlib` are two different libs)

oren-z0:
@matejcik Fixed. The code now uses python-bitcoinlib.
I hope I got the `script_type` evaluation correctly:

```
if len(script_pub_key) == 25 and s[0] == OP_DUP and s[1] == OP_HASH160 and s[2] == 20 and s[-2] == OP_EQUALVERIFY and s[-1] == OP_CHECKSIG:
    script_type = messages.InputScriptType.SPENDADDRESS
elif script_pub_key.is_p2sh():
    script_type = messages.InputScriptType.SPENDP2SHWITNESS
elif script_pub_key.is_witness_v0_keyhash():
    script_type = messages.InputScriptType.SPENDWITNESS
elif script_pub_key.is_witness_scriptpubkey() and len(script_pub_key) == 34 and script_pub_key[0] == OP_1:
    script_type = messages.InputScriptType.SPENDTAPROOT
```

PR #4147, opened 2024-09-02
T3B1: fix coin support info and translations generation

matejcik:
also flip CI to use T3B1 everywhere that we previously used T2B1.

to be discussed: do we _also_ want to run T2B1 in some limited capacity?
(e.g., yes to UI tests, no to translations?)


matejcik:
builds on top of #4148

mmilata:
> do we also want to run T2B1 in some limited capacity?

I wonder if we should build firmware images for both? At least nightly.

UI tests don't seem very important.

mmilata (approved):
LGTM!

PR #4153, opened 2024-09-04
core: bump version to 2.8.4

mmilata:
Also bump bootloader, prodtest.


matejcik:
ideally we'll bump bootloader/prodtest after #4157

matejcik (approved):

PR #4157, opened 2024-09-05
fix bootloader unlocking issues

TychoVrahe:
This PR fixes several unreleased bootloader issues, recently introduced:
- reporting bootloader unlock state via features
- firmware downgrade protection was using incorrect version in one place
- unlocking bootloader on U5 was granting HDP access to firmware, which is unwanted as it contains other stuff than optiga pairing secret
- and finally, fixes prodtest issue when running prodtest on unlocked bootloader. Note that there are differences between F4 and U5. Unlocking bootloader on F4 will result in prodtest being able to pair optiga again (unless optiga is locked of course). On U5 this is not allowed (as HDP contains other stuff so we can't make it accessible again).

### DIFF
--- a/core/embed/prodtest/optiga_prodtest.c
+++ b/core/embed/prodtest/optiga_prodtest.c
@@ -143,6 +143,12 @@ void pair_optiga(void) {
   uint8_t secret[SECRET_OPTIGA_KEY_LEN] = {0};
 
   if (secret_optiga_get(secret) != sectrue) {
+    if (secret_optiga_writable() != sectrue) {
+      // optiga pairing secret is unwritable, so fail
+      optiga_pairing_state = OPTIGA_PAIRING_ERR_WRITE_FLASH;
+      return;
+    }
+
     // Generate the pairing secret.
     if (OPTIGA_SUCCESS != optiga_get_random(secret, sizeof(secret))) {
       optiga_pairing_state = OPTIGA_PAIRING_ERR_RNG;
--- a/core/embed/trezorhal/stm32u5/secret.c
+++ b/core/embed/trezorhal/stm32u5/secret.c
@@ -176,6 +176,29 @@ secbool secret_optiga_present(void) {
   return secret_present(SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
 }
 
+secbool secret_optiga_writable(void) {
+  const uint32_t offset = SECRET_OPTIGA_KEY_OFFSET;
+  const uint32_t len = SECRET_OPTIGA_KEY_LEN;
+
+  const uint8_t *const secret =
+      (uint8_t *)flash_area_get_address(&SECRET_AREA, offset, len);
+
+  if (secret == NULL) {
+    return secfalse;
+  }
+
+  int secret_empty_bytes = 0;
+
+  for (int i = 0; i < len; i++) {
+    // 0xFF being the default value of the flash memory (before any write)
+    // 0x00 being the value of the flash memory after manual erase
+    if (secret[i] == 0xFF) {
+      secret_empty_bytes++;
+    }
+  }
+  return sectrue * (secret_empty_bytes == len);
+}
+
 // Backs up the optiga pairing secret from the secret storage to the backup
 // register
 static void secret_optiga_cache(void) {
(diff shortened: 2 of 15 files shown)


core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -170,6 +170,29 @@ secbool secret_optiga_present(void) {
   return secret_present(SECRET_OPTIGA_KEY_OFFSET, SECRET_OPTIGA_KEY_LEN);
 }
 
+secbool secret_optiga_writable(void) {
+  uint32_t offset = SECRET_OPTIGA_KEY_OFFSET;
```
obrusvit, line 174: `offset`, `len` can be `const`.
`secret` can be `const uint8_t *`.

core/embed/prodtest/optiga_prodtest.c
```diff
@@ -143,6 +143,12 @@ void pair_optiga(void) {
   uint8_t secret[SECRET_OPTIGA_KEY_LEN] = {0};
 
   if (secret_optiga_get(secret) != sectrue) {
+    if (secret_optiga_writable() != sectrue) {
```
matejcik, line 146: due to this, the bootloader can be re-locked by prodtest on the F4 but not on the U5, correct?

shouldn't we add re-locking capability to prodtest for U5 too? we'd need a command to explicitly erase the secret area
(or maybe we just stuff `secret_erase` here?)

TychoVrahe: Correct.

I would say that re-locking bootloader is in fact unwanted. On F4 we cant really prevent it though, custom firmware can erase the sector whenever it likes and write whatever it wants there. On U5 the secret area contains other stuff we'd need to erase, namely bootloader and firmware monotonic counters, which shouldnt ideally be possible. We could attempt to re-populate the counters but that still leaves space for attacks via power disconnection in right time.

matejcik: i mean, if someone gets ahold of a leaked prodtest, they can use it to erase the secret sector, which will also lose them the pairing secret, so their device no longer registers as authentic
OTOH the factory might want the ability to re-pair a specific U5 to a different Optiga
(probably no need to solve that right now though -- we can produce an updated prodtest that does this at some later point)

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -265,7 +288,7 @@ void secret_prepare_fw(secbool allow_run_with_secret, secbool trust_all) {
 #ifdef USE_OPTIGA
   secret_optiga_uncache();
   if (sectrue == allow_run_with_secret) {
-    if (secfalse != secret_optiga_present()) {
+    if (sectrue != secret_optiga_writable()) {
```
matejcik, line 291: this flow is rather convoluted, which is probably why we made this mistake

how about flattening it like so:
```
#ifdef USE_OPTIGA
  secret_optiga_uncache();
  secbool optiga_secret_present = secret_optiga_present();
  secbool optiga_secret_writable = secret_optiga_writable();
  if (sectrue == allow_run_with_secret && sectrue == optiga_secret_writable && secfalse == optiga_secret_present) {
    // Secret is not present and the secret sector is writable.
    // This means the U5 chip is unprovisioned.
    // Allow trusted firmware (prodtest presumably) to access the secret sector,
    // early return here.
    return;
  }
  if (sectrue == allow_run_with_secret && sectrue == optiga_secret_present) {
    // Firmware is trusted and the Optiga secret is present, make it available.
    secret_optiga_cache();
  }
  // Disable access unconditionally.
  secret_disable_access();
  if (sectrue != trust_all && sectrue == optiga_secret_present) {
    // Untrusted firmware, locked bootloader. Show the restricted screen.
    show_install_restricted_screen();
  }
#else
  secret_disable_access();
#endif
```

core/embed/trezorhal/stm32u5/secret.c
```diff
@@ -37,10 +37,10 @@ secbool secret_ensure_initialized(void) {
 }
 
 secbool secret_bootloader_locked(void) {
-#ifdef FIRMWARE
-  return (TAMP->BKP8R != 0) * sectrue;
+#if defined BOOTLOADER || defined BOARDLOADER
+  return secret_optiga_present();
 #else
-  return sectrue;
+  return (TAMP->BKP8R != 0) * sectrue;
```
cepetr, line 43: To be precise, we should probably test all eight words of the secret for zeros, not just the first one.

core/embed/prodtest/optiga_prodtest.c
TychoVrahe: Yeah i would wait with adding this until there is a real need from factory.

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/4157/commits/2d8ed383f9cf4db3d2324b5254a9481a1cd5e69d

core/embed/trezorhal/stm32u5/secret.c
TychoVrahe: Agree that it improves readability. 

I added `sectrue == trust_all` as it matches the previous behavior and i think more restrictions here are good.

https://github.com/trezor/trezor-firmware/pull/4157/commits/85bc0438251f9e0d6f8b333fae8d399afb14f57c

core/embed/trezorhal/stm32u5/secret.c
obrusvit: `const`-correctness might lead to smaller binary so might be worth doing across the codebase

matejcik (approved):

PR #4158, opened 2024-09-05
ci: warn when Cargo.lock is edited, don't test :it:

mmilata:
Example comment below :arrow_down:, wording suggestions appreciated.

Also remove Italian from the language matrix for full CI pipeline (with `translations` PR label) as discussed with @Hannsek

### DIFF
--- a/.github/workflows/prebuild.yml
+++ b/.github/workflows/prebuild.yml
@@ -6,6 +6,9 @@ on:
   schedule:
     - cron: '13 23 * * *'  # every day @ 23:13
 
+env:
+  CARGOLOCK_COMMENT: Looks like you changed `Cargo.lock`. Please make sure to review the dependencies and update [internal version list](https://www.notion.so/satoshilabs/Rust-dependencies-a9cc6e8dab934def8eb27896c001e6e2).
+
 jobs:
   block-fixup:
     name: Block fixup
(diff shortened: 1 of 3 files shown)


.github/workflows/prebuild.yml
```diff
@@ -6,6 +6,9 @@ on:
   schedule:
     - cron: '13 23 * * *'  # every day @ 23:13
 
+env:
+  CARGOLOCK_COMMENT: Looks like you changed `Cargo.lock`. Please make sure to review the dependencies and update internal version list.
```
obrusvit, line 10:
```suggestion
  CARGOLOCK_COMMENT: Looks like you changed `Cargo.lock`. Please make sure to review the dependencies and update [internal version list](https://www.notion.so/satoshilabs/Rust-dependencies-a9cc6e8dab934def8eb27896c001e6e2).
```

.github/workflows/prebuild.yml
```diff
@@ -79,3 +82,40 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Ask git
+        id: git_diff
+        run: |
```
obrusvit, line 99: We can maybe simplify it to:
```
run: |
    git diff --exit-code origin/main -- core/embed/rust/Cargo.lock || echo "cargo_modified=1" >> $GITHUB_ENV
```

This should create `env.cargo_modified` variable which can be used in subsequent steps as:
`if: ${{ env.cargo_modified == '1' }}`

WDYT

obrusvit (commented):
Adding two suggestions.

.github/workflows/prebuild.yml
mmilata: Sounds good, didn't realize these persist between jobs.

mmilata:
Also added fix for broken coverage reporting due to actions/upload-artifact#598

obrusvit (approved):

Issue #1946, opened 2021-11-25
Implement multisig for Taproot


andrewkozlik:
There are a number of options how to implement multisig with Taproot [1], so first we have to decide which one(s) we will implement. Using a single `OP_CHECKSIGADD`-based script is the simplest and most similar to the way classic multisig works. Nevertheless we want to wait for Casa and Unchained Capital to see what they want to aim for.

[1] https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki#cite_note-5

prusnak:
I pinged Casa and Unchained Capital folks to chime in ...

afsheenb:
On behalf of Unchained: we've been discussing internally how best to implement multisig on Taproot and have a final meeting scheduled for mid-December. I'll report back and chime in with our plans then.

jlopp:
Given the interactivity requirements of other tapscript constructions, I expect Casa will move forward with OP_CHECKSIGADD

andrewkozlik:
> Given the interactivity requirements of other tapscript constructions, I expect Casa will move forward with OP_CHECKSIGADD

@jlopp, would that be using a single OP_CHECKSIGADD-based k-of-n script or a k-of-k script for every combination of pubkeys? The advantages and disadvantages of the two are discussed in https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki#cite_note-5. If I am not mistaken, Casa uses 2-of-3 multisig and 3-of-5 multisig. On average the single OP_CHECKSIGADD-based k-of-n script is better in both cases in terms of the size of the resulting witnesses. But note the words "on average". If you optimize the Merkle tree for the happy path / typical use case, then the k-of-k script for every combination may be more optimal in terms of witness size for the happy path.

jlopp:
Single OP_CHECKSIGADD-based k-of-n script, otherwise we'd have to change the UX to force users to either preselect which keys they're signing with, or have them sign multiple leaves of the tapscript.

afsheenb:
Chiming in again after our meeting: we're hoping to support both of the two OP_CHECKSIGADD based approaches mentioned in BIP 342, with a preference for the k-of-k script (though we anticipate making use of the k-of-n script as well).

danger89:
Any update?

mmikeww:
@coldcard @nvk

deepsimulation:
Hi @prusnak,

Any update on this?

andrewtoth:
[...]

PR #4159, opened 2024-09-05
Taproot multisig

andrewtoth:
Adds basic taproot multisig support.

Uses a single `OP_CHECKSIGADD`-based script at the top level. Uses the secp256k1 generator as the internal pubkey. If passing already generated signatures, they must be added in order and any missing signatures must be passed as a single 0 byte.

Resolves https://github.com/trezor/trezor-firmware/issues/1946


andrewtoth:
@prusnak @andrewkozlik @onvej-sl @matejcik Any feedback on this PR?

matejcik:
> any feedback on this PR?

sure!
in general, we are interested in taproot multisig functionality.
we don't have it on the roadmap at the moment, so before anything happens on the PR, you'll need to wait for us to prioritize it.
personally I don't know the status of the various taproot multisig proposals, so i can't comment on whether your solution implements the one (or one of) that we want

in the future, please open an issue first where you outline your proposed solution. that way you can avoid doing all the work in case we say it's not something we are interested in :woman_shrugging:

prusnak:
The change is not so big and honestly it is worth merging in. Of course, we really need to have the new code properly tested.

core/src/apps/bitcoin/common.py
```diff
@@ -102,6 +107,11 @@ def from_int(cls, sighash_type: int) -> "SigHashType":
     InputScriptType.SPENDMULTISIG,
 )
 
+LEAF_VERSION = const(0xC0)
+GENERATOR = const(
```
prusnak, line 111: maybe let's add a comment that the generator value is taken from https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs

prusnak (commented):
Can we also add some tests that test the construction of a Taproot multisig transaction?

andrewtoth:
Thank you for the feedback @matejcik and @prusnak 

> in the future, please open an issue first 

There is in fact an open issue, #1946, which I linked. The consensus in that issue is to implement the approach in this PR.

Also, this approach will be compatible with [BIP387](https://github.com/bitcoin/bips/blob/master/bip-0387.mediawiki) descriptors `tr(key, multi_a(k, key1, ... keyn))` and also the [BIP388](https://github.com/bitcoin/bips/blob/master/bip-0388.mediawiki) wallet policy `tr(@0/**, multi_a(k, @1/**, ..., @n/**))`.

> Can we also add some tests that test the construction of a Taproot multisig transaction?

Yes, I'd be happy to add tests and make any changes needed to have this feature merged and released. But first, I would like some assurance that this will in fact be merged. Would we be able to approve the workflows here so I can make sure I didn't miss anything else?

core/src/apps/bitcoin/common.py
andrewtoth: I will actually update this to instead use a BIP32 node with this as the pubkey and a chaincode of 32 `0` bytes, and then we can derive the unspendable internal pubkey for the multisig path. That will make this compatible with BIP388 wallet policies, which require the all keys to be a xpubs and use the same derivation path.

andrewtoth: That way other wallets can use the same construction and have the xpub for the internalPubkey be `xpub661MyMwAqRbcEYS8w7XLSVeEsBXy79zSzH1J8vCdxAZningWLdN3zgtU6QgnecKFpJFPpdzxKrwoaZoV44qAJewsc4kX9vGaCaBExuvJH57`, as described [here](https://delvingbitcoin.org/t/unspendable-keys-in-descriptors/304).

andrewtoth: @prusnak added more commentary [here](https://github.com/trezor/trezor-firmware/pull/4159/files#diff-c528ee9407682668ced569943303086cb4098504c4a3684aa85b2f766b283763R85-R105).

andrewtoth:
@matejcik @prusnak thank you - I have updated the PR to use a deterministic HDNode for the unspendable internal pubkey, and added tests for getting addresses and signing and serializing a tx.

prusnak:
> I have updated the PR to use a deterministic HDNode for the unspendable internal pubkey

Why is this preferred to using a static dummy key from previously used code?

andrewtoth:
> > I have updated the PR to use a deterministic HDNode for the unspendable internal pubkey
> 
> Why is this preferred to using a static dummy key from previously used code?

This makes it compatible with BIP388 wallet policies. I cannot make a multisig wallet using a Trezor as one key and another hardware wallet that uses wallet policies as another key with just a static dummy key.

prusnak:
> added tests for getting addresses and signing and serializing a tx.

Amazing!

prusnak:
I fixed smaller issues detected by the CI, but there is still bigger fish to fry:

```
$ cd core
$ make pyright                
python ../tools/pyright_tool.py

Ignored 88 custom-defined errors from 17 files.

ERROR: We have issues!

/Users/stick/work/trezor/trezor-firmware/core/src/apps/bitcoin/sign_tx/bitcoin.py:661: - error: Expression of type "tuple[bytes, bytes]" is incompatible with return type "bytes"
  "tuple[bytes, bytes]" is incompatible with "bytes" (reportReturnType)

/Users/stick/work/trezor/trezor-firmware/core/src/apps/bitcoin/sign_tx/bitcoin.py:677: - error: Argument of type "int" cannot be assigned to parameter "pubkey" of type "bytes" in function "multisig_pubkey_index"
  "int" is incompatible with "bytes" (reportArgumentType)

/Users/stick/work/trezor/trezor-firmware/core/src/apps/bitcoin/sign_tx/bitcoin.py:682: - error: Argument of type "int" cannot be assigned to parameter "signature" of type "bytes" in function "write_witness_multisig_taproot"
  "int" is incompatible with "bytes" (reportArgumentType)

/Users/stick/work/trezor/trezor-firmware/core/src/apps/bitcoin/sign_tx/bitcoin.py:688: - error: Argument of type "int" cannot be assigned to parameter "signature" of type "bytes" in function "write_witness_p2tr"
  "int" is incompatible with "bytes" (reportArgumentType)

/Users/stick/work/trezor/trezor-firmware/core/src/apps/bitcoin/sign_tx/bitcoin.py:713: - error: Argument of type "int | bytes" cannot be assigned to parameter "signature" of type "bytes" in function "set_serialized_signature"
  Type "int | bytes" is incompatible with type "bytes"
    "int" is incompatible with "bytes" (reportArgumentType)

/Users/stick/work/trezor/trezor-firmware/core/src/apps/bitcoin/sign_tx/decred.py:168: - error: Expression of type "DecredSigHasher" is incompatible with return type "SigHasher"
  "DecredSigHasher" is incompatible with protocol "SigHasher"
    "hash341" is an incompatible type
      Type "(i: int, tx: SignTx | PrevTx, sighash_type: SigHashType) -> bytes" is incompatible with type "(i: int, tx: SignTx | PrevTx, sighash_type: SigHashType, leaf_hash: bytes | None) -> bytes"
        Function accepts too many positional parameters; expected 3 but received 4 (reportReturnType)

/Users/stick/work/trezor/trezor-firmware/core/src/trezor/utils.py:282: - error: Expression of type "memoryview[_I@memoryview] | memoryview[int]" is incompatible with return type "memoryview[int]"
  Type "memoryview[_I@memoryview] | memoryview[int]" is incompatible with type "memoryview[int]"
    "memoryview[_I@memoryview]" is incompatible with "memoryview[int]"
      Type parameter "_I@memoryview" is invariant, but "_I@memoryview" is not the same as "int" (reportReturnType)

/Users/stick/work/trezor/trezor-firmware/core/src/trezor/utils.py:292: - error: Expression of type "_I@memoryview | int" is incompatible with return type "int"
  Type "object* | int" is incompatible with type "int"
    "object*" is incompatible with "int" (reportReturnType)

/Users/stick/work/trezor/trezor-firmware/core/src/trezor/utils.py:300: - error: Expression of type "_I@memoryview | int" is incompatible with return type "int"
  Type "object* | int" is incompatible with type "int"
    "object*" is incompatible with "int" (reportReturnType)

Found 9 issues above
make: *** [pyright] Error 1
```

andrewtoth:
For me
```
$ cd core
$ make pyright  
```
returns
```
python ../tools/pyright_tool.py
Error: [Errno 2] No such file or directory: 'pyright'
make: *** [Makefile:211: pyright] Error 1
```

I couldn't find much docs on setting up pyright. Any hints?

prusnak:
> I couldn't find much docs on setting up pyright. Any hints?

Hm. Not sure - maybe `npm install -g pyright@1.1.361`?

andrewtoth:
Thanks! After my last commit I get:
```
python ../tools/pyright_tool.py

Ignored 88 custom-defined errors from 17 files.

SUCCESS: Everything is fine!
```

andrewtoth:
I get an error that I need to add a changelog as well. Should I do that or will you?

andrewtoth:
Ah I see legacy tests are failing. I only tested this on Model T with `core`. Can this feature be disabled on legacy devices, or is legacy support required for this?

prusnak:
> Can this feature be disabled on legacy devices, or is legacy support required for this?

We can disable the tests with

```
@pytest.mark.models(skip=["legacy"])
```

andrewtoth:
For the device tests, it appears I need to add hashes for each device and each language. Could that be done in an automated step by CI?

andrewtoth:
@prusnak I'm unsure how to proceed here. I don't have any context about what the `Post comment with UI diff URLs` jobs are and why they are failing. There's also a `Changelog check` job failing that seems to be looking for a legacy changelog but there hasn't been anything touched in legacy.

Is there anything else I need to do here?

prusnak:
> @prusnak I'm unsure how to proceed here.

Let's leave that to @matejcik or someone else from the firmware team. I am sure the fix is trivial, but for me or you it would take hours to figure out.

Also legacy needs a changelog entry because we touched the code in the `legacy/` folder (added extra parameter to some functions). Again, let's have the firmware team to either add a dummy changelog entry or find a way how to suppress the change.

Again, thank you for this massive contribution and we'll take it from here.

crypto/tests/test_check.c
```diff
@@ -10587,7 +10587,7 @@ START_TEST(test_zkp_bip340_tweak) {
     ck_assert_int_eq(res, 0);
     ck_assert_mem_eq(output_priv, result, 32);
 
-    res = zkp_bip340_tweak_public_key(internal_pub, root_hash, result);
+    res = zkp_bip340_tweak_public_key(internal_pub, root_hash, result, NULL);
```
onvej-sl, line 10590: Please extend the test to also check for parity.

core/src/apps/bitcoin/multisig.py
```diff
@@ -82,6 +82,27 @@ def multisig_get_pubkey(n: HDNodeType, p: paths.Bip32Path) -> bytes:
...
+    # The following encodes this xpub into an HDNode. It is the NUMS point suggested
+    # in BIP341, with a chaincode of 32 0 bytes. Deriving a pubkey from this node
+    # results in a provably unspendable pubkey.
+    # https://delvingbitcoin.org/t/unspendable-keys-in-descriptors/304
+    # xpub661MyMwAqRbcEYS8w7XLSVeEsBXy79zSzH1J8vCdxAZningWLdN3zgtU6QgnecKFpJFPpdzxKrwoaZoV44qAJewsc4kX9vGaCaBExuvJH57
+    node = bip32.HDNode(
+        depth=0,
+        fingerprint=2084970077,
+        child_num=0,
+        chain_code=b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00",
```
andrewtoth, line 97: I think we also want to allow passing in the chain code to create a randomized xpub that still uses the NUMS point public key. That way the derived public key is provably unspendable, but cannot be fingerprinted on chain.
We will need to define a new parameter for this in the MultisigRedeemScriptType.

core/src/apps/bitcoin/multisig.py
```diff
@@ -82,6 +82,27 @@ def multisig_get_pubkey(n: HDNodeType, p: paths.Bip32Path) -> bytes:
...
+        public_key=b"\x02P\x92\x9bt\xc1\xa0IT\xb7\x8bK`5\xe9z^\x07\x8aZ\x0f(\xec\x96\xd5G\xbf\xee\x9a\xce\x80:\xc0",
+    )
+
+    for i in multisig.address_n:
+        node.derive(i, True)
+    return node.public_key()
```
onvej-sl, line 103: This is something I don't understand: The dummy pubkey is derived using `multisig.address_n`. Suppose we have two wallets. The first one uses a node `HD1` and the other a node `HD2`. Suppose we have a multisig address created from public keys `HD1/0/0` and `HD1/0/1`. Since the wallets use different paths to derive the dummy pubkey, they will generate different addresses. Consequently, if the first wallet creates a transaction with the multisig address as an output, the second wallet will create a valid witness with a probability of 1/2, because the parity of the tweaked dummy key is used in the witness. Is this correct?

andrewtoth: But `address_n` must be the same for the entire `MultisigRedeemScriptType`, so how can the address be created by `HD1/0/0` and `HD1/0/1`? That would imply an `address_n` of `[0,0]` for the first key and `[0,1]` for the second. Am I misunderstanding?

onvej-sl: I was under the impression that you could specify `address_n` for each pubkey separately in `multisig.pubkeys[*].address_n`.

https://github.com/trezor/trezor-firmware/blob/549d592e8ead44b05315391c862e3ed749d452ae/core/src/apps/bitcoin/multisig.py#L111

andrewtoth: Oh I see you can have pubkeys or nodes, and pubkeys have their own address_n individually. I haven't tested this with that. I suppose we can disable using pubkeys list for taproot multisig?
Is pubkeys list legacy or is it still widely used?

onvej-sl: By the way, I'm going to do a quick research to see what internal keys other wallets that support Taproot multisig use. If we want other wallets to be interoperable with trezor, we should follow the same approach.

andrewtoth: Hey great! I'm building a multisig wallet that works with ledger and want to support Trezor as well. I can confirm using the NUMS point as a public key and a random chaincode will still show "Dummy" on the ledger device.

onvej-sl: I'm not sure if I understand what you said. Does it mean that ledger supports taproot multisig? Does the ledger device know the random chaincode?

onvej-sl: Thank you. It is now clear to me.

I still don't understand what is wrong about using constant dummy key (that is `tr(@0,multi_a(n,@1/**, ..., @m/**))`). Is there any reason besides non-compliance with BIP 388?

In fact, BIP 388 doesn't forbid it. It states that an implementation might choose to permit additional derivation patterns; however, some considerations need to be taken into account. It also states that naked pubkeys with no `/*` suffix would result in a total loss of privacy which is not our case, which doesn't apply in our case.

>In order to allow supporting legacy derivation schemes (for example, using simply /* instead of the more common /<M;N>/* scheme most software wallets use today), or other schemes that are not covered in this document, implementations might choose to permit additional derivation patterns for the key placeholder (KP) expressions.
>
>However, care needs to be taken in view of the following considerations:
>
>    * Allowing derivation schemes with a different length or cardinality in the same wallet policy would make it difficult to guarantee that there are no repeated pubkeys for every possible address generated by the policy. For example, `@0/<0;1>/*` and `@1/*` would generate the same pubkeys if the second public key in the key information vector is one of the first two unhardened children of the first public key. This could cause malleability with potential security implications (for example, in policies containing miniscript).
>    * Allowing naked pubkeys with no `/*` suffix (for example a descriptor template like `wsh(multi(2,@0,@1/<0;1>/*)))` would cause a pubkey to be repeated in every output generated from the policy, which would result in a total loss of privacy.

andrewtoth: > BIP 388 doesn't forbid it

Ah, well I could not get it to work with the Ledger :)

> what is wrong about using constant dummy key

But, what is specifically wrong with it is a loss of privacy, albeit not "total" loss of privacy. With a constant dummy key or constant dummy xpub, an observer will know that there never was a possibility of spending with the keypath once the funds are spent.

If a constant dummy xpub is used, an observer will also know what derivation path was used for that spend.

andrewtoth: Ahh yes, this is why https://github.com/LedgerHQ/app-bitcoin-new/blob/master/doc/wallet.md#implementation-specific-restrictions

core/src/apps/bitcoin/sign_tx/tx_weight.py
```diff
@@ -72,10 +74,18 @@ def input_script_size(cls, i: TxInput) -> int:
                 pass
 
         if multisig:
+            n = len(multisig.nodes) if multisig.nodes else len(multisig.pubkeys)
             if script_type == IST.SPENDTAPROOT:
-                raise wire.ProcessError("Multisig not supported for taproot")
+                multisig_script_size = _TXSIZE_MULTISIGSCRIPT_TAPROOT + n * (
+                    1 + 1 + _TXSIZE_PUBKEY
```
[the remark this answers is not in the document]
andrewtoth: Fixed.

core/src/apps/bitcoin/multisig.py
onvej-sl: Thank you for all your input.

I did some research but I wasn't much successful. I found two wallets that claim to support multisig Taproot: [Muun](https://muun.com/) and [Nunchuk](https://nunchuk.io/). However, Muun uses key aggregation, and regarding Nunchuk, I was unable to determine from the code what they are using.

These are possible options:
  1) The internal key will be a constant unspendable key.
  2) All the public keys in the multisig will use the same derivation path. The internal key will be derived from a constant unspendable key using the same path.
  3) There will a new field in `MultisigRedeemScriptType` that specifies the path to be used to derive the internal key from a constant unspendable key.

Options (2) and (3) are compatible with Ledger.
Option (1) leaks a little privacy (see https://github.com/trezor/trezor-firmware/pull/4159#discussion_r1783090156).
Option (2) allows for fewer use cases than non-taproot multisig.

I feel I cannot decide this myself since it is not purely a technical question.  @matejcik @andrewkozlik What is you opinion?

onvej-sl: > If a constant dummy xpub is used, an observer will also know what derivation path was used for that spend.

Do they? How?

andrewkozlik: > Option (1) leaks a little privacy (see [#4159 (comment)](https://github.com/trezor/trezor-firmware/pull/4159#discussion_r1783090156)).

Don't options (2) and (3) have the same "privacy leak"? In fact option (2) leaks even more, because it reveals information about the derivation path. So AFAICS option (1) leaks the least of the three.

onvej-sl: Completely true, this also probably answers [my question](https://github.com/trezor/trezor-firmware/pull/4159#discussion_r1784356249). I suppose @andrewtoth meant what I called option (2).

andrewkozlik: > With a constant dummy key or constant dummy xpub, an observer will know that there never was a possibility of spending with the keypath once the funds are spent.

I think that's a fairly insignificant privacy issue.

> If a constant dummy xpub is used, an observer will also know what derivation path was used for that spend.

I think that this is more concerning. Firstly it reveals something about the account usage, i.e. how many addresses were handed out at a certain point in time. When the UTXOs are spent in a transaction, assuming the common input ownership heuristic, an observer can determine for example the number of payments that a business receives in a certain amount of time. Secondly, given the fact that the index is typically used monotonically, a blockchain analyst can link addresses from the same account together. If they observe two transaction outputs that was created shortly after each other, both are 2-of-3 multisig and one uses index e.g. 19265 and the other 19266, then there is a high chance they belong to the same account.

onvej-sl: > Don't options (2) and (3) have the same "privacy leak"?

This depend on how you use it.

onvej-sl: There is also an option proposed [here](https://github.com/trezor/trezor-firmware/pull/4159#discussion_r1781449326):

  4. There will a new field in `MultisigRedeemScriptType` that specifies the chain code that will be used to derive the internal key from a constant unspendable key. This option can be combined with options (2) and (3).

andrewkozlik: If I would be the first person to implement taproot multisig, then I'd go with option (1), because it's the simplest option of the ones recommended in BIP 341 [here](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs). Since this is not the first implementation, we of course want to aim for good compatibility with existing wallets, but that shouldn't be at the price of introducing unnecessary privacy issues. I am not familiar enough with BIP 388 or the Ledger implementation. Perhaps @bigspider can help us figure out the best way forward.

andrewtoth: I initially did #1, but that did not work with Ledger, so implemented #2 which is the PR as it is right now, but will implement #3 because it solves privacy issue, *but* the user needs to then store the xpub used for the internal pubkey as well, or we could default to all zero chaincode if not provided like it is now. So, optionally improve privacy.

This is the approach recommended by @bigspider in https://delvingbitcoin.org/t/unspendable-keys-in-descriptors/304 and is compatible with Ledger.

andrewtoth: Actually, 4 will specify the entire HDNode to use to derive the taproot key. Internally, we check that the root public key used by that HDNode is the NUMS point. The chaincode can be anything. If the public key is not the NUMS point we abort for now.

In the future this would allow Trezor to allow any internal public key, but a warning should be shown in the UI that the key might be spendable on its own. I think that's out of scope for this initial taproot PR.

onvej-sl: I think option (4) makes sense only when combined with option (2) or (3), since the chain code plays a role only when deriving child nodes. Which option would you combine it with?

andrewtoth: Ahh yeah that's true, the user would have to confirm. In Ledger this is confirmed with via the wallet policy.

andrewtoth: How is this different then swapping out a cosigner's xpub and using the attacker's xpub for one to generate the receive address? The user must confirm on screen that the address is a correct multisig address, and the malicious software wallet can show the same address with swapped xpubs in the software side. But, if you are not comparing addresses with all other cosigners then the same attack can happen.

andrewkozlik: > the user would have to confirm

That's awful UX, not to mention that they would need to record it in addition to their recovery seed. If we decide to go in this direction, the chaincode needs to be derived deterministically. For example a hash of all the XPUBs in the multisig.

andrewtoth: They need to record the entire descriptor along with their recovery seed already? The recover seed is not enough to prevent loss of funds for multisig.

The xpub used for the internal pubkey is simply added in the descriptor, along with all cosigner xpubs.

andrewtoth: > For example a hash of all the XPUBs in the multisig.

I like that!

andrewtoth: We can make the chaincode the sha256 hash of all xpubs sorted in lexicographical order, plus the threshold.

andrewkozlik: > The recover seed is not enough to prevent loss of funds for multisig.

Currently if I have several hardware wallets with different seeds, I can use them to create a multisig account. Even if all these wallets get destroyed, including the computer where I was doing the multisig, and my entire house burns down, except for the recovery seeds, I can still recover the multisig account only from my recovery seeds. If we add a random chain code, this will no longer be true.

andrewtoth: That's fair, but you would need to recover *all* seeds, not just the threshold which kind of defeats the purpose of the threshold.

andrewtoth: Ok, so how about this - we keep the `MultisigRedeemScriptType` the same, and the chaincode used is simply the hash as I described above? I think that's a great solution.

andrewtoth: > @darosior already implemented a deterministic derivation for the chaincode in Liana; it would be great if new implementations of the deterministic derivations are kept compatible.

I don't think it's great that it depends on the order of keys provided. If we want to sort the pubkeys beforehand and order the xpubs in the order of the derived pubkeys, that would make a different chaincode for each order.

I think we should sort the xpubs before hashing.

andrewkozlik: Well, it's great to see that someone else came to essentially the same solution. Seems like we are on the right path. To make sure that we are all on the same page, let me summarize the current plan:
- The dummy XPUB that is used to derive the internal public key will use the NUMS public key specified in BIP 341 and a chain code deterministically derived from the multisig XPUBs.
- `MultisigRedeemScriptType` doesn't change. Trezor derives the dummy XPUB internally.
- All the public keys in the multisig, including the internal public key will use the same derivation path.
- To keep things simple, I would disable `MultisigRedeemScriptType.pubkeys` for taproot multisig, i.e. enforce the use of `MultisigRedeemScriptType.nodes`. There is nothing stopping us from loosening this restriction in the future if necessary.

The question remains how to derive the chain code. This is tricky. Trezor doesn't deal with policies or descriptors directly at the moment, at least not for multisig. Right now, distinguishing between `sortedmulti_a()` and `multi_a()` would be up to the host application, and it's up to that application to fill in the `MultisigRedeemScriptType` in the correct order. So if `sortedmulti_a()` is specified, it needs to derive the public keys and sort the XPUBs accordingly in `MultisigRedeemScriptType`.

It seems to me that we can't just take the order of the XPUBs given in `MultisigRedeemScriptType`, because with `sortedmulti_a()` it would make the dummy XPUB depend on the path, since the host changes the XPUB order for each path. That would presumably be incompatible with Ledger, not to mention confusing. So unless we overhaul `MultisigRedeemScriptType` to be able to handle descriptors, we need the dummy XPUB derivation to be independent of the XPUB order. Meaning they need to be sorted somehow, as you proposed, or the derivation operation needs to be order-agnostic, e.g. dummy chain code = XOR of the multisig chain codes. The XOR is not great, because it allows a malicious participant to set the chain code to a desired value, but come to think of it, perhaps that not an issue.

All that being said, I'd very much hope for the solution to be compatible with existing implementations, and ideally with whatever will be the future specification, because sooner or later this will need to be standardized to ensure smooth interoperability amongst wallets. Hopefully your comment in delvingbitcoin.org will spark some discussion about that.

bigspider: > It seems to me that we can't just take the order of the XPUBs given in `MultisigRedeemScriptType`, because with `sortedmulti_a()` it would make the dummy XPUB depend on the path, since the host changes the XPUB order for each path. That would presumably be incompatible with Ledger, not to mention confusing.

Ledger's implementation doesn't care how the chaincode of the unspendable key is computed, as it's explicitly passed on by the client together with all the other xpubs. It _does_ recognize it as unspendable as long as the pubkey is the standard NUMS point, and that's important for security - or the client could lie and put an actual spendable pubkey.

Using the deterministic derivation only matters for UX (the device could in principle skip showing it altogether in that case); this is currently left for future improvements to the wallet policies / miniscript UX.

andrewtoth: I believe he meant incompatible with Liana, for the reasons pointed out in my earlier comment and on delving.

Of course Ledger is compatible with any xpub given in the wallet policy.

andrewkozlik: I am somewhat disappointed that our comments in delvingbitcoin.org did not spark any feedback. The way I see it, we are practically facing these options:

1. The host will supply the chaincode for the unspendable XPUB and the user will confirm it on Trezor. This has poor UX. As if multisig signing wasn't hard enough to use for laymen, we add yet another layer of complexity for them to understand.
2. Design our own scheme that derives the chaincode for the unspendable XPUB from the multisig XPUBs independently of their order. Poor interoperability, e.g. with Liana. All we can do is hope that our design is appealing enough that others adopt it.
3. Adopt the Liana scheme, which in my opinion is not the best choice.

If a standard derivation scheme arises in the future, that is not identical with our choice, we will need to maintain the old scheme somewhere in our code, which I would like to avoid if we can. We currently don't have the capacity to push the standardization process forward. @andrewtoth would you be willing to do that? Even starting a discussion on the Bitcoin mailing list could be a step forward.

andrewtoth: @andrewkozlik thanks for that summary. I agree, we should ideally have a BIP for this before merging here. #1 would be safest option in that sense since we can guarantee we will be forward compatible with any standard that emerges. But, the UX drawback is indeed unfortunate.

I do think there is already a risk with Trezor not showing all xpubs being used in a multisig address, for the same reason the dummy xpub needs to be shown. A single trezor user must take for granted all the other xpubs used in creating the multisig address shown. Ledger solves this by storing this information as a wallet policy that only has to be confirmed once.

I will be meeting with some of the other developers who have commented in that delving thread next week. I will be sure to discuss this with them.

onvej-sl: @andrewtoth In https://github.com/trezor/trezor-firmware/pull/4351, I introduced [MultisigPubkeysOrder.LEXICOGRAPHIC](https://github.com/trezor/trezor-firmware/pull/4351/commits/a1f9f89062c2a25ce70070859d9bfd7b86b47d94) that sorts pubkeys in an input or output.

I believe this would simplify the implementation of the Liana schema.

andrewtoth: Thanks @onvej-sl, that will be very useful so we don't have to sort xpubs ahead of time to be compatible with `sortedmulti_a`. I don't think Liana's scheme is what we should use here though. It would still break the property that only the backup seeds are needed to recover. Now, you would also need to know the order of the keys, although it's not as bad as remembering 32 random bytes.

I made another post [here](https://delvingbitcoin.org/t/unspendable-keys-in-descriptors/304/28) that I think covers all bases.

andrewtoth: I updated this change to compute the chaincode of the internal xpub as the sha256 hash of the concatenation of the sorted base pubkeys of the xpubs after removing duplicates. All xpubs must also have a derivation depth of 2.

I can proceed with a BIP and mailing list post if this is acceptable.

I could not update the testnet tests because testnet is currently unusable.

andrewtoth: I don't think it makes a difference in terms of security and privacy for the first 3 options.
Only the chain code and public key are used to derive the actual public key that will be used in the script though, so the last option would introduce malleability. For example, two xpubs where one lies about the depth would produce the same redeem script but would result in a different chaincode for the dummy xpub.

andrewtoth: Yes, I didn't realize but if using just the 33 bytes compressed public key like I have here it does make it compatible with Liana, if Liana has their xpubs sorted, no duplicates, and max depth of 2.

andrewkozlik: > I can proceed with a BIP and mailing list post if this is acceptable.

Thanks. Unless @onvej-sl has any objection to the proposed method, I encourage you to go ahead.

onvej-sl: I have no objections.

andrewtoth: Thanks for your patience. I have created a BIP draft specifying this behavior. Once it gets assigned a number I can rebase this and hopefully we can see it merged :crossed_fingers: 

https://github.com/bitcoin/bips/pull/1746

Asylian21:
From a broader perspective:

PSBT + Taproot script-path is becoming a foundational building block
for advanced custody models (multisig, policy wallets, coordinators/signers separation).

Without this, it’s hard to build interoperable, vendor-neutral Bitcoin tooling on top of Trezor.
Would be great to know if full PSBT/Taproot support is considered a near-term priority.

Hannsek:
We do not plan to support this in the near term.

PR #4172, opened 2024-09-11
Introduce TROPIC01 from Tropic Square

ibz:
* This PR adds [libtropic](https://github.com/tropicsquare/libtropic/) as a dependency to the build and exposes its functionality to Micropython
* On `unix` (Trezor emulator), `libtropic` requires the [Tropic model](https://github.com/tropicsquare/ts-tvl/) to be running and accepting connections over TCP, while the actual device `libtropic` will talk directly to the [TROPIC01](https://tropicsquare.com/) chip
* A smoke test is provided as part of the unit tests under `core/`

Note: A way to launch the TROPIC01 model from `emu.py` used to be provided in this PR, but was removed after we added the model as a submodule, because it was quite rudimentary. We are exploring another way to automatically launch the model, for development and for tests.


core/site_scons/models/T3T1/emulator.py
```diff
@@ -59,6 +59,26 @@ def configure(
...
+            "vendor/libtropic/src/libtropic.c",
+            "vendor/libtropic/src/lt_crc16.c",
+            "vendor/libtropic/src/lt_hkdf.c",
+            "vendor/libtropic/src/lt_l1.c",
+            "vendor/libtropic/src/lt_l1_port_wrap.c",
+            "vendor/libtropic/src/lt_l2.c",
+            "vendor/libtropic/src/lt_l2_frame_check.c",
+            "vendor/libtropic/src/lt_l3.c",
+            "vendor/libtropic/src/lt_random.c",
+            "vendor/libtropic/hal//port/unix/lt_port_unix.c",
```
prusnak, line 73:
```suggestion
            "vendor/libtropic/hal/port/unix/lt_port_unix.c",
```

core/site_scons/models/T3T1/emulator.py
```diff
@@ -59,6 +59,26 @@ def configure(
...
+            "vendor/libtropic/hal/crypto/trezor_crypto/lt_crypto_trezor_aesgcm.c",
+            "vendor/libtropic/hal/crypto/trezor_crypto/lt_crypto_trezor_ed25519.c",
+            "vendor/libtropic/hal/crypto/trezor_crypto/lt_crypto_trezor_sha256.c",
+            "vendor/libtropic/hal/crypto/trezor_crypto/lt_crypto_trezor_x25519.c",
+        ]
+        defines += ["USE_TREZOR_CRYPTO"]
```
prusnak, line 79: Maybe this should be `USE_TS_CRYPTO`, no?

```suggestion
        defines += ["USE_TS_CRYPTO"]
```

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,130 @@
...
+    if (ret != LT_OK) {
+        mp_raise_msg(&mp_type_TropicError, "Failed to read certificate.");
+    }
+
+    vstr_t vstr = {0};
+    vstr_init_len(&vstr, 1024);
+
+    bytes_to_chars(X509_cert, vstr.buf, 512);
+
+    return mp_obj_new_str_from_vstr(&mp_type_bytes, &vstr);
```
prusnak, line 111: Why convert to hex here? IMHO just return raw binary bytes and if we need hex we can call (`foo.hex()` in Python).

```suggestion
    return mp_obj_new_str_copy(&mp_type_str, X509_cert, 512);
```

prusnak (changes requested):

core/site_scons/models/T3T1/emulator.py
ibz: > Maybe this should be `USE_TS_CRYPTO`, no?

Actually, I was thinking *not*, because this is the emulator build, in which we would still use `TREZOR_CRYPTO`, because the TS chip is not there. What do you think, @pavelpolach321?

pavelpolach321: In libtropic we use USE_TREZOR_CRYPTO define if we want to use **trezor_crypto** as a source of cryptographic code which is running on host (MCU) side. 

Example: https://github.com/tropicsquare/libtropic/blob/master/hal/crypto/trezor_crypto/ts_trezor_x25519.c#L1

Therefore I think that passing USE_TREZOR_CRYPTO to compilation of libtropic files will work in your case?

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,276 @@
...
+    if (ret != LT_OK) {
+      lt_deinit(&handle);
+      mp_raise_msg(&mp_type_TropicError, "lt_ping failed.");
+    }
+  } else {
+    lt_deinit(&handle);
+    return mp_const_none;
+  }
+
+  lt_deinit(&handle);
```
pavelpolach321, line 105: I would also check return value in all lt_deinit() calls, mainly because in emulator this function internally talks to tcp server, which could in fact fail easily.

onvej-sl:
Hello @ibz, I wanted to ask if you could use fixup commits instead of force-pushing. Fixup commits make it easier to track which suggestions have already been addressed. We usually squash the fixup commits just before merging the pull request.

onvej-sl:
As one of the developers involved in integrating Optiga into Trezor, I would like to document the architecture we used for the integration. The following diagrams shows the dependencies of all relevant functions where Optiga is utilized:

```mermaid
graph TD
    subgraph optiga_transport["core∕embed∕trezorhal∕optiga∕optiga_transport.c"]
    optiga_sec_chan_handshake
    optiga_init
    optiga_execute_command
    end
    subgraph optiga_commands["core∕embed∕trezorhal∕optiga∕optiga_commands.c"]
    optiga_calc_sign --> optiga_execute_command     
    optiga_get_data_object --> optiga_execute_command    
    optiga_set_autostate --> optiga_execute_command  

    optiga_get_random --> optiga_execute_command             
    end    
    subgraph optiga["core∕embed∕trezorhal∕optiga∕optiga.c"]
    optiga_sign --> optiga_calc_sign
    optiga_read_cert --> optiga_get_data_object
    optiga_random_buffer --> optiga_get_random
    optiga_pin_verify --> optiga_set_autostate    
    end
    subgraph storage["storage∕storage.c"]
    storage_unlock --> optiga_pin_verify
    end
    subgraph modoptiga["core∕embed∕extmod∕modtrezorcrypto∕modtrezorcrypto_optiga.h"]
    sign --> optiga_sign
    get_cert --> optiga_read_cert
    end
    subgraph modconfig["core∕embed∕extmod∕modtrezorconfig∕modtrezorconfig.c"]
    unlock --> storage_unlock
    end
    subgraph modrandom["core∕embed∕extmod∕modtrezorcrypto∕modtrezorcrypto_random.c"]
    random_bytes --> optiga_random_buffer
    end    
    subgraph modmain["core∕embed∕kernel∕main.c"]
    driver_init --> optiga_sec_chan_handshake
    driver_init --> optiga_init
    end
```

Note that the functions `optiga_sec_chan_handshake` and `optiga_init` (which are perhaps analogous to `tropic_init`) are defined outside python modules (`core/embed/extmod/**`). The reason for this is that we call them from `core/embed/kernel/main.c`. However, I'm not certain if this is sufficient justification, since it seems to me that they could be called only after the micropython environment is initialized.

I am not claiming that the architecture is perfect nor that there are no reasons to integrate Tropic differently. However, it would be nice, if we could keep it consistent.

One significant difference is that we don't have an Optiga emulator. How will the code in this pull request differ if we use a physical device instead of an emulator?

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,282 @@
...
+
+/// def ping(message: str) -> str:
+///     """
+///     Test the session by pinging the chip.
+///     """
+STATIC mp_obj_t mod_trezorcrypto_tropic_ping(mp_obj_t message) {
+  lt_handle_t handle = {0};
+  lt_ret_t ret = LT_FAIL;
+
+  tropic_init(&handle);
```
onvej-sl, line 93: Why don't you call `tropic_init` once, keep the handle, and reuse it for multiple commands? It seems unnecessary to me to call it before every command.

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,282 @@
...
+  vstr_t vstr = {0};
+  vstr_init_len(&vstr, len_rand);
+
+  memcpy(vstr.buf, buff, len_rand);
+
+  return mp_obj_new_str_from_vstr(&mp_type_bytes, &vstr);
+}
+STATIC MP_DEFINE_CONST_FUN_OBJ_1(mod_trezorcrypto_tropic_random_get_obj,
+                                 mod_trezorcrypto_tropic_random_get);
+
```
onvej-sl, line 218: I'm not sure if we need this method. If we want to generate randomness in python, we would like to combine entropy from Tropic, Optiga and MCU, so we rather need to integrate this function [here](https://github.com/trezor/trezor-firmware/blob/ibz/20240805-libtropic/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-random.h#L47).

ibz:
> Hello @ibz, I wanted to ask if you could use fixup commits instead of force-pushing. Fixup commits make it easier to track which suggestions have already been addressed. We usually squash the fixup commits just before merging the pull request.

Got it, sorry for that! BTW, I just force-pushed again, because I rebased on `main`. There's no way around that, I suppose... But I will do fixup commits from now on.

ibz:
> Note that the functions `optiga_sec_chan_handshake` and `optiga_init` (which are perhaps analogous to `tropic_init`) are defined outside python modules (`core/embed/extmod/**`). The reason for this is that we call them from `core/embed/kernel/main.c`. However, I'm not certain if this is sufficient justification, since it seems to me that they could be called only after the micropython environment is initialized.

Indeed, I can move `tropic_init` out of the Python module. We should probably call it in the same place where we call `optiga_init`, which is in `main.c`.

> I am not claiming that the architecture is perfect nor that there are no reasons to integrate Tropic differently. However, it would be nice, if we could keep it consistent.

By this did you mean just the above part, about `tropic_init`? Or the other difference - that `optiga` comes under `trezorhal` whereas `tropic` does not?

The way I see it ... we have the `optiga` HAL (for the real chip) and the "unix" `optiga` HAL (for the emulator) - which are two different things. But in the case of tropic, `libtropic` is what provides the abstraction - because it can talk either to the real chip, or to the model, over TCP. So there would be no HAL layer (or - rather - `libtropic` is the HAL). This is why I went to just expose `libtropic` to Python without adding anything under `trezorhal`. But I can do it otherwise, if you think that is a good idea.

onvej-sl:
> Indeed, I can move tropic_init out of the Python module. We should probably call it in the same place where we call optiga_init, which is in main.c.

One reason for initializing optiga in main.c was that the handshake process uses a secret [stored](https://github.com/trezor/trezor-firmware/blob/main/core/embed/kernel/main.c#L116) in a memory location that becomes inaccessible once the firmware enters unprivileged mode. It seems to me that this concept applies to tropic as well. It would be beneficial to store the production handshake keys in this area.

> But in the case of tropic, libtropic is what provides the abstraction - because it can talk either to the real chip, or to the model, over TCP.

This is something I don't quite understand. Does it mean that `libtropic` itself manages the communication with the chip? How can I instruct `libtropic` to use a real chip instead of an emulator? How can I specify which interface (for example [I2C bus index](https://github.com/trezor/trezor-firmware/blob/main/core/embed/trezorhal/optiga/optiga_transport.c#L195)) it should use for communication with tropic.

pavelpolach321:
> > Indeed, I can move tropic_init out of the Python module. We should probably call it in the same place where we call optiga_init, which is in main.c.
> 
> One reason for initializing optiga in main.c was that the handshake process uses a secret [stored](https://github.com/trezor/trezor-firmware/blob/main/core/embed/kernel/main.c#L116) in a memory location that becomes inaccessible once the firmware enters unprivileged mode. It seems to me that this concept applies to tropic as well. It would be beneficial to store the production handshake keys in this area.
> 
> > But in the case of tropic, libtropic is what provides the abstraction - because it can talk either to the real chip, or to the model, over TCP.
> 
> This is something I don't quite understand. Does it mean that `libtropic` itself manages the communication with the chip? How can I instruct `libtropic` to use a real chip instead of an emulator? How can I specify which interface (for example [I2C bus index](https://github.com/trezor/trezor-firmware/blob/main/core/embed/trezorhal/optiga/optiga_transport.c#L195)) it should use for communication with tropic.

Hello, libtropic manages the communication by using its platform specific c file during compilation. For compiling for embedded target, use this file in your makefile: https://github.com/tropicsquare/libtropic/blob/develop/hal/port/stm32/lt_port_stm32.c - have a look from line 241 to the end.
For compiling for unix environment (your emulator), where you expect libtropic to talk to TROPIC01 model over tcp, then use this file 
https://github.com/tropicsquare/libtropic/blob/develop/hal/port/unix/lt_port_unix.c

onvej-sl:
Thank you for the explanation. It makes sense to me now. [This line](https://github.com/trezor/trezor-firmware/blob/7283ff608ea07fcecd512a0855e9ff166843da2e/core/site_scons/models/T3T1/emulator.py#L75) is what causes the build to use an emulator instead of a physical device.

I still don't understand whether we have to perform a handshake before every command. See [this](https://github.com/trezor/trezor-firmware/pull/4172#discussion_r1793944736) comment. Trezor with optiga performs a handshake once during startup and then reuses the handle. Is there any benefit in performing the handshake for each individual command?

ibz:
> I still don't understand whether we have to perform a handshake before every command. See [this](https://github.com/trezor/trezor-firmware/pull/4172#discussion_r1793944736) comment. Trezor with optiga performs a handshake once during startup and then reuses the handle. Is there any benefit in performing the handshake for each individual command?

We don't. That is my mistake. Will fix.

onvej-sl:
My suggestion is to:
1. Store `SHiPRIV_BYTES` in `core/embed/sec/secret/unix/secret.c`.
2. Introduce a function, `secret_tropic_get_privkey`, in `core/embed/sec/secret/unix/secret.c`. This function will return `SHiPRIV_BYTES`. 
3. Introduce a function, `tropic_handshake(const uint8_t *trezor_privkey)`, in `core/embed/sec/tropic/tropic_transport.c`. This function will compute the public key using `curve25519_scalarmult_basepoint`, call `lt_handshake`, and store the handle.
4. In `core/embed/projects/kernel/main.c`, call `tropic_handshake` and `lt_init`.

Once we want to use a physical tropic, we will implement `core/embed/sec/secret/model_with_tropic/secret.c`, and use `port_model_with_tropic.c` instead of `lt_port_unix.c`. 

The tropic commands can then be called either directly using `libtropic` and the handle in `tropic_transport.c`, or through wrappers that will be in `tropic_commands.c`.

I am still uncertain about the variable naming. Does this approach make sense to you?

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-tropic.h
ibz: https://github.com/trezor/trezor-firmware/pull/4172/commits/0c33e899df511d6d4f3535c92c9a99cde2820fb8#diff-e1a3e03160d11080fcca39f3a8a9205fde724d8cd7d61c8505f6f41c9a95b04fR46

core/embed/extmod/modtrezorcrypto/modtrezorcrypto-tropic.h
ibz: Removed it https://github.com/trezor/trezor-firmware/pull/4172/commits/0c33e899df511d6d4f3535c92c9a99cde2820fb8#diff-e1a3e03160d11080fcca39f3a8a9205fde724d8cd7d61c8505f6f41c9a95b04fL183

ibz:
> 4\. call `tropic_handshake` and `lt_init`

There is one thing I don't understand in the above. If `main.c` calls `lt_init` (or rather, some wrapper called `tropic_init`?) - which obtains the handle, where would the handle be stored? And how would `modtrezorcrypto-tropic.h` have access to that handle?

ibz:
> My suggestion is to:
> [...]

@onvej-sl OK, I did what you described (sort of). I think it is indeed much cleaner. Also it is very similar now to how Optiga works. See this: https://github.com/trezor/trezor-firmware/pull/4172/commits/cd789a3cf07024263b7959472fa799b2cbaf500c

What I don't understand now is how will we be able to test from the emulator, which is what I was trying to achieve in the first place.

I see that `core/embed/projects/unix/` also never touches `optiga` - so we are in the same position with Tropic as we are with Optiga. But still, I was thinking we want to have it somehow testable from the emulator / CI pipeline. Or am I misunderstanding - is `core/embed/projects/kernel` somehow also used by the emulator?

@matejcik do you have any opinion on this?

onvej-sl:
> I see that core/embed/projects/unix/ also never touches optiga - so we are in the same position with Tropic as we are with Optiga.

This is a difference between optiga and tropic. We don't have a optiga emulator, unfortunately, so we mock all the functions that use optiga.

> But still, I was thinking we want to have it somehow testable from the emulator / CI pipeline. Or am I misunderstanding - is core/embed/projects/kernel somehow also used by the emulator?

I believed the kernel is compiled into the emulator. However, I might be wrong as this is a new concept for me.

ibz:
> 
> I believed the kernel is compiled into the emulator. However, I might be wrong as this is a new concept for me.

OK, it seems to work. Even the smoke test works against the emulator with the Tropic model. Here are the changes you requested: https://github.com/trezor/trezor-firmware/pull/4172/commits/a0804193718a39102509020fba50d4591916af97

Please let me know how to proceed.

core/embed/sec/tropic/tropic_transport.c
```diff
@@ -0,0 +1,62 @@
...
+  lt_ret_t ret = LT_FAIL;
+
+  uint8_t X509_cert[LT_L2_GET_INFO_REQ_CERT_SIZE] = {0};
+  ret = lt_get_info_cert(&lt_handle, X509_cert, LT_L2_GET_INFO_REQ_CERT_SIZE);
+  if (ret != LT_OK) {
+    return TROPIC_ERR_GET_INFO_CERT;
+  }
+
+  uint8_t stpub[32] = {0};
+  ret = lt_cert_verify_and_parse(X509_cert, 512, stpub);
```
onvej-sl, line 48: Does the Tropic emulator use a constant public key in the certificate?

core/embed/projects/unix/main.c
```diff
@@ -498,6 +504,21 @@ static int sdl_event_filter(void *userdata, SDL_Event *event) {
...
+  secbool tropic_secret_ok = secret_tropic_get(tropic_secret);
+
+  tropic_init();
+  if (sectrue == tropic_secret_ok) {
+    if (tropic_handshake(tropic_secret) != TROPIC_SUCCESS) {
+      // ??
+    }
+  }
+  memzero(tropic_secret, sizeof(tropic_secret));
+#endif
```
onvej-sl, line 519: This should slightly differ from how it is done with optiga.

The difference is that you can communicate with an optiga without establishing a secure channel, but the set of commands you can use is restricted.

In the case of tropic, it doesn't make sense to continue if any of the functions `secret_tropic_get()`, `trocpi_init()` or `tropic_handshake` fails. So I would `ensure()` that all the functions succeed. And make sure to zero out the secret if any of the functions fails.

ibz: Perfect, makes sense. I'll do that!

core/embed/sec/tropic/tropic_transport.c
pavelpolach321: TROPIC01 model typically accepts config files during start, they consist of certificate* and STPRIV. Technically speaking those might be regenerated, but current model have them commited in repo, therefore "static". So I would say, your model is using probably still those "hardcoded" commited files.

*this will slightly change, now there is just chip's certificate but we will provide chain of certificates.

ibz: Something like this, I assume? https://github.com/trezor/trezor-firmware/pull/4172/commits/18e8049546088c05ef0c8d3dc67bebb8a51cf4a6

Basically I ran `lt_session_start` & `lt_get_info_cert` once, dumped the key, then added it as a constant which we can now use. So essentially the pairing was done once - and it will be done once for every individual Tropic chip.

PS: I also removed the Tropic related code from `kernel/main.c` for now - since it was not being built and there is no way to test it anyway! We should finalize it for the emulator first, then we can port the code to run on a physical device.

core/embed/projects/unix/main.c
onvej-sl: Now I see something that I previously overlooked:
  * The content of the function `drivers_init` will be reused in `core/embed/projects/unix/main.c` so we can move it probably to `core/embed/sec/tropic/tropic_transport.c`.
  * `tropic_secret_tropic_pubkey` is not private, so it does not need to be zeroed.

By the way,  I think `drivers_init` can be simplified by reordering the individual function. Something like this:
```c
  ensure((tropic_init() == TROPIC_SUCCESS) * sectrue, "tropic_init failed");
  ensure(secret_tropic_get_tropic_pubkey(tropic_secret_tropic_pubkey), "secret_tropic_get_tropic_pubkey failed");
  ensure(secret_tropic_get_trezor_privkey(tropic_secret_trezor_privkey), "secret_tropic_get_trezor_privkey failed");
  tropic_result result = tropic_handshake(tropic_secret_trezor_privkey, tropic_secret_tropic_pubkey);
  memzero(tropic_secret_trezor_privkey, sizeof(tropic_secret_trezor_privkey));
  ensure((result == TROPIC_SUCCESS) * sectrue, "tropic_handshake failed");
```

ibz: You are right, this looks much better! And can be used in `kernel/main.c` by simply calling `tropic_init` now.

https://github.com/trezor/trezor-firmware/pull/4172/commits/fac69b398e3051562a1a8c85d2c1ec2575137dc8

core/embed/sec/tropic/tropic_transport.c
```diff
@@ -0,0 +1,60 @@
...
+  uint8_t trezor_pubkey[SECRET_TROPIC_KEY_LEN] = {};
+  curve25519_scalarmult_basepoint(trezor_pubkey, trezor_privkey);
+
+  ret = lt_session_start(&lt_handle, tropic_pubkey, PKEY_INDEX_BYTE, trezor_privkey, trezor_pubkey);
+  if (ret != LT_OK) {
+    return TROPIC_ERR_SESSION_START;
+  }
+
+  return TROPIC_SUCCESS;
+}
```
onvej-sl, line 60: I would consider inlining this function.

andrewkozlik:
Is there anything blocking progress on this issue? I see some unresolved review comments from @pavelpolach321 and @onvej-sl. It would be nice if we could start working on integrating the TROPIC01 into the Trezor PIN verification. Presumably this PR is also a blocker for https://github.com/trezor/trezor-firmware/pull/4635.

core/tests/test_trezor.crypto.tropic.py
```diff
@@ -0,0 +1,27 @@
+from common import *  # isort:skip
+
+from slip39_vectors import vectors
+from trezor.crypto import tropic
+
+
+class TestCryptoTropic(unittest.TestCase):
+    def test_ping(self):
+        self.assertEqual(tropic.ping("HeLlO!"), "HeLlO!")
```
matejcik, line 9: does this test pass when the model is not present?

TychoVrahe:
not exactly a blocker for #4635, there is some overlap (and potentially conflicts), but the HAL PR works on its own.

core/embed/sec/tropic/tropic_transport.c
ibz: Done!

core/tests/test_trezor.crypto.tropic.py
pavelpolach321: Short answer: In python module you are calling libtropic's function, so test will NOT pass if model is not present
Longer answer: In libtropic library, switching between "model" or "physical chip" happens on lowest (transport) layer, nothing is emulated in libtropic.  If you are calling lt_ping(), libtropic must talk to either model over TCP, or chip over SPI, otherwise libtropic will return different value than 0 (probably LT_FAIL because lt_init() would fail). There are no mockups.

core/site_scons/models/T3W1/emulator.py
```diff
@@ -52,6 +52,35 @@ def configure(
...
+        ]
+        paths += ["embed/sec/tropic/inc"]
+        defines += ["USE_TREZOR_CRYPTO"]
+        features_available.append("tropic")
+
+        ########
+        # USE_TROPIC should eventually be enabled when building with the 'tropic' feature
+        # but for now we don't have the model running, which would crash the emulator on startup.
+        # defines += ["USE_TROPIC=1"]
+        ########
```
onvej-sl, line 82: If I understand it correctly, this implies that there is no test in the CI that would at least ensure this pull request compiles without errors.

What do you think about
  * introducing a new model that would be the same as T3W1 but with `USE_TROPIC=1` and 
  * adding a pipeline that will build the emulator for this model?

core/tests/test_trezor.crypto.tropic.py
onvej-sl: @matejcik, are you okay with merging this pull request even if we can't test it in CI? See also [this](https://github.com/trezor/trezor-firmware/pull/4172/files#r1975376319) comment.

core/site_scons/models/T3W1/emulator.py
ibz: Indeed, I agree this is not ideal (not building it at least).

I like your idea actually, and that could be good in the future also, besides just to have it built as part of the CI.

I assume most development locally would be done against T3W1 "plain" and T3W1_TROPIC would only be started when needed, because it needs the model running and it is not necessary for most firmware work.

matejcik: USE_TROPIC=1 is set for T3W1 hardware builds so that ensures general buildability of this code.

i don't really see a point of building this for emulator until we have the model widely available, and if the code somehow drifts between now and then, we'll fix it then. merging now, without CI involvement, is much better than letting it linger for longer (heh)

.gitmodules
```diff
@@ -27,7 +27,6 @@
 [submodule "vendor/cmsis_5"]
 	path = vendor/cmsis_5
 	url = https://github.com/ARM-software/CMSIS_5.git
-
```
matejcik, line 30: please drop this difference, it's just noise

core/embed/sec/tropic/tropic_transport.c
onvej-sl: Oh, it seems the term "to inline a function" is ambiguous. You meant to use the `inline` qualifier and I meant to replace the call with the contents of the function.

The reason I suggested this was that `tropic_handshake` makes two calls, and the rest of the function is error handling. It seems to me that inlining could make it easier to read.

core/site_scons/models/T3W1/emulator.py
onvej-sl: By the way, `SConscript.firmware` doesn't compile `libtropic.c`.

onvej-sl: And there is also tropic code that is specific for emulator.

Everything I was saying here was to support the idea of having a pipeline that tests whether the emulator build with tropic can be built and not relying on hardware build.

core/embed/sec/tropic/inc/sec/tropic_transport.h
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TREZORHAL_TROPIC_TRANSPORT_H
+#define TREZORHAL_TROPIC_TRANSPORT_H
+
+#include <trezor_types.h>
+#include "libtropic.h"
```
onvej-sl, line 24: I think this include is not necessary here and can be moved to `tropic_transport.c`.

core/embed/projects/unix/main.c
```diff
@@ -37,6 +37,7 @@
 #include <unistd.h>
 
 #include <io/display.h>
+#include <sec/secret.h>
```
onvej-sl, line 40: This is a nitpick: This include can be nested inside `#ifdef USE_TROPIC`.

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,174 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#if USE_TROPIC
+
+#include <sec/secret.h>
+#include <sec/tropic_transport.h>
+#include "libtropic.h"
+
+#define PKEY_INDEX_BYTE PAIRING_KEY_SLOT_INDEX_0
```
onvej-sl, line 26: I think this can be removed.

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,174 @@
...
+#define SIG_SIZE 64
+
+/// def ping(message: str) -> str:
+///     """
+///     Test the session by pinging the chip.
+///     """
+STATIC mp_obj_t mod_trezorcrypto_tropic_ping(mp_obj_t message) {
+  lt_ret_t ret = LT_FAIL;
+
+  uint8_t msg_in[PING_MSG_MAX_LEN] = {0};
```
onvej-sl, line 47: Allocating 4 KB buffers seems overkill to me. When I examine the implementation of `lt_ping()`, the size of `msg_in` is the same as that of `msg_out`. However, I'm not certain if this is guaranteed by the function's interface.  If not, the interface is poorly designed and should include a parameter set by the caller to indicate the size of the buffer allocated for `msg_in`, which is then set by the callee to indicate the actual size of `msg_in`. What do you think, @pavelpolach321 ?

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,174 @@
...
+  lt_ret_t ret = LT_FAIL;
+
+  uint8_t X509_cert[512] = {0};
+  ret = lt_get_info_cert(&lt_handle, X509_cert, 512);
+  if (ret != LT_OK) {
+    mp_raise_msg(&mp_type_TropicError, "lt_get_info_cert failed.");
+  }
+
+  vstr_t vstr = {0};
+  vstr_init_len(&vstr, 512);
```
onvej-sl, line 86: Is the certificate always exactly 512 bytes in size?

core/embed/sec/tropic/inc/sec/tropic_common.h
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TREZORHAL_TROPIC_COMMON_H
+#define TREZORHAL_TROPIC_COMMON_H
+
+typedef enum _tropic_result {
+  TROPIC_SUCCESS = 0,  // Operation completed successfully.
+  TROPIC_ERR_SESSION_START,
+} tropic_result;
```
onvej-sl, line 26: It seems to me that if we eliminate `tropic_handshake`, we won't need this structure anymore. However, this doesn't rule out the possibility that we might need it in the future.

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,174 @@
...
+///     Test the session by pinging the chip.
+///     """
+STATIC mp_obj_t mod_trezorcrypto_tropic_ping(mp_obj_t message) {
+  lt_ret_t ret = LT_FAIL;
+
+  uint8_t msg_in[PING_MSG_MAX_LEN] = {0};
+
+  mp_buffer_info_t message_b = {0};
+  mp_get_buffer_raise(message, &message_b, MP_BUFFER_READ);
+  if (message_b.len > 0) {
```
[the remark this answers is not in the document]
pavelpolach321: This is a good catch, we actually have a bug there, I just created issue at our repo.
lt_ping should accept message len of 0x1 - 0x1001. No zero length messages.

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,172 @@
...
+  mp_buffer_info_t message_b = {0};
+  mp_get_buffer_raise(message, &message_b, MP_BUFFER_READ);
+  if (message_b.len >= 0) {
+    ret = lt_ping(&lt_handle, (uint8_t *)message_b.buf, (uint8_t *)msg_in,
+                  message_b.len);
+    if (ret != LT_OK) {
+      mp_raise_msg(&mp_type_TropicError, "lt_ping failed.");
+    }
+  } else {
+    return mp_const_none;
```
onvej-sl, line 56: I would rather raise an exception here.

core/embed/sec/tropic/tropic_transport.c
ibz: Yeah, I don't know how I understood the one that made least amount of sense. Anyway... https://github.com/trezor/trezor-firmware/pull/4172/commits/07425d69411a20883084721064158b4c5a70e8e5#diff-2179dea73b04181bd25219fedb2e591db4a1cbd46972c416eefd1f1adf882f45R43

core/embed/sec/tropic/inc/sec/tropic_transport.h
ibz: Yep https://github.com/trezor/trezor-firmware/pull/4172/commits/07425d69411a20883084721064158b4c5a70e8e5#diff-55fbcf8fbcee5ed9d0ad61ec900e32043ee8b7f6e77f75d9921410bb55d63decL24

core/embed/projects/unix/main.c
ibz: https://github.com/trezor/trezor-firmware/pull/4172/commits/07425d69411a20883084721064158b4c5a70e8e5#diff-85c6076cdbcf5611bd53eaaaecdf7d5bdbb77966adcf50fb0f75ae7f7ff3e82fR60

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
ibz: https://github.com/trezor/trezor-firmware/pull/4172/commits/07425d69411a20883084721064158b4c5a70e8e5#diff-20d5e6900b62b56f17de07c11dd9dafc13dcefc43b3b863af9a6aa6f4775b2afL26

core/site_scons/models/T3W1/emulator.py
ibz: How if I make a new issue for that and do it separately?

matejcik: well, @onvej-sl is correct in that if we have no confirmation that the code is building _at all_, there is very little point in commiting it.
I believe that we should at minimum enable `tropic` for T3W1 hardware build.

> By the way, SConscript.firmware doesn't compile libtropic.c.

this should be fixed

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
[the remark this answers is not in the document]
onvej-sl: Maybe, this is an inconsistency between tropic emulator and physical device. @pavelpolach321 ?

.gitmodules
onvej-sl: > please drop this difference, it's just noise

I'm not sure if I understand. This is not directly related to this pull request but what is wrong about removing an extra empty line in `.gitmodules`, @matejcik ?

matejcik: i haven't looked at the file closely and assumed that there are empty lines between every entry. after your comment i see that it's the opposite. in which case sure, feel free to remove it

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
pavelpolach321: I also think that 4kB ping is not necessary, much smaller size should be sufficient here, like 32B.

Regarding your question:
Both arrays msg_in and msg_out are expected to be of the same size. Caller is responsible for preparing valid arrays for msg_out and msg_in, then size is passed to lt_ping().

Generally speaking sending and receiving L3 command goes through encryption/decryption phase.
Each l3 data frame contains `len` parameter, this parameter is placed (next to msg_out) into outgoing frame, encrypted, sent, received decrypted and then "the same" parameter is checked in incomming frame after decryption.

In lt_ping(), if `len` of outgoing packet is not the same as `len` of incomming (after reception), function returns LT_FAIL and nothing is copied into `msg_out`.

Therefore I think that we don't need parameter to tell us how many bytes were written into `msg_out`. It is either the same number as we sent, or it is LT_FAIL..?

onvej-sl: In other words, you say the interface of `lt_ping()` guarantees that the number of bytes written into `msg_out` is the same as the size of `msg_in`. Then I suggest mentioning it [here](https://github.com/tropicsquare/libtropic/blob/e9ba707e3323abefc2f67e5bed75a51adf2164d8/include/libtropic.h#L241), so that a developer using `libtropic` doesn't have to examine the implementation of `lt_ping()`.

onvej-sl: @ibz That means it is safe to replace `uint8_t msg_in[PING_MSG_MAX_LEN] = {0};` by `uint8_t msg_in[message_b.len];`.

core/site_scons/models/T3W1/emulator.py
TychoVrahe: we should enable 'tropic' in Sconscript.firmware and Sconscript.kernel, at right time:) 

I looked at this PR just briefly, but there are issues that make currently unworkable on HW. We (embedded team) can fix this in followup PR if you prefer, but generally speaking:
- tropic functionality should not be accessed directly from firmware, but rather through syscalls - firmware doesn't have access to I2C peripheral so it won't be able to communicate with the chip.
- libtropic functions should not be called directly from firmware, specialized syscalls for each function should be implemented (or we could -maybe- implement syscalls for each libtropic api function, but that seems like too wide access for firmware and too complicated)
- there is now conflicting `tropic_init` on HW, the one in tropic_transport should probably be kept and used on both hw&emu, the existing should be renamed to something like `tropic_hal_init` and called from tropic_init, and emu function tropic_hal_init should be introduced (also tropic_transport.c seems a bit weird, how about just tropic.c?)
- secret.c implementation for HW needs to be adjusted 
 
... not sure what else i missed

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
onvej-sl: I understand. Does `libtropic` know the size of the certificate? If so, it would be beneficial if `lt_et_info_cert()` returned the size of the certificate, so that the caller doesn't have to parse the certificate to determine where it ends.

core/site_scons/models/T3W1/emulator.py
onvej-sl: The way I see this pull request is that it aims to integrate the tropic emulator. The reason is that we would like to start implementing features depending on the tropic chip. It's not necessary to have a functional build for a physical tropic chip for this purpose. However, it is appropriate to have tests in CI that ensure the emulator with tropic builds successfully. These tests will help ensure that no one accidentally breaks the build when integrating the physical device.

ibz: Indeed, the goal of this PR is to integrate tropic in the *emulator*. I have been trying to add it to `Sconscript.firmware` now, but it seems there are indeed some issues. If what we want is to have it build as part of the CI pipeline, we should perhaps do it differently, not by adding it to the HW build?

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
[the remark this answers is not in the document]
pavelpolach321: Will check this

core/site_scons/models/T3W1/emulator.py
ibz: One alternative solution that I discussed with @matejcik is this:

 1. You can run the emulator locally with or without the model running - it will not crash if you build with `USE_TROPIC` and the model is not running: https://github.com/trezor/trezor-firmware/pull/4172/commits/6c08cae860b0e472d742219250d287559cb60bf3
 2. There is an extra flag that disables the tropic tests independent of USE_TROPIC: https://github.com/trezor/trezor-firmware/pull/4172/commits/281503dde506a04f4e8219fff40ac9acce4bf204 This essentially lets us build with libtropic without running the tests yet...

What do you think @onvej-sl ?

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
pavelpolach321: Libtropic does not know the size of certificate at the moment, it just gets 512B of space dedicated for certificate and then looks into certain field for STPUB key. Currently it is designed to unblock secure channel establishment, nothing more.

This 'certificate parsing part' is big TBD on our side, because we are finalizing details with production, about how certificates will exactly look. There will be the whole chain and we will have to rework this part, once specification of certificates is done.

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
onvej-sl: https://github.com/trezor/trezor-firmware/pull/4172/commits/19c5328d94864a0b880e208eb8880f5794639fc2 doesn't fix it correctly. You have to call `mp_get_buffer_raise(message, &message_b, MP_BUFFER_READ);` before `uint8_t msg_in[message_b.len];`.

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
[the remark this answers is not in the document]
onvej-sl: @ibz I think you can remove `if (message_b.len >= 0)` together with the else branch.

core/site_scons/models/T3W1/emulator.py
onvej-sl: To clarify my position: I'm unhappy that we don't have a CI pipeline to ensure that the code in this pull request compiles without errors.

Does your solution address this problem?

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
onvej-sl: Okay. Thanks.

core/site_scons/models/T3W1/emulator.py
ibz: @onvej-sl Yes, with reverting disabling of USE_TROPIC: `6cbc3f946216068c1df5f8f566a4f2739e9fa187` and setting the variable to skip the Tropic test.

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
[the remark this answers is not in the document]
ibz: Yes! https://github.com/trezor/trezor-firmware/pull/4172/commits/aab70710e251303ed7475552fcb498b6d1dc1f3d

core/site_scons/models/T3W1/emulator.py
onvej-sl: I still believe that running the build pipeline in CI once for T3W1 with tropic and once for T3W1 without tropic is a better solution. However, I'm okay with your suggestion if @matejcik approves it

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
onvej-sl: And one more thing: `result.len = strlen(result.buf);` can be removed.

core/site_scons/models/T3W1/emulator.py
ibz: So the plan is this:

 * [ ] create a `ts-tvl` GitHub action for "release" such that we can have something like the "latest" release that we can install (currently installing `ts-tvl` means navigating to the last action ran and downloading the artefact)
 * [ ] "release" `ts-tvl`
 * [ ] install the "latest" `ts-tvl` in the environment running the unit tests
 * [ ] download model config and certificates from somewhere - should we just get them from [GitHub](https://github.com/tropicsquare/ts-tvl/tree/master/tvl/server/model_config)?
 * [ ] start model as part of the unit test runner (`core/tests/run_tests.sh`) using the model config and certificates downloaded above
 * [ ] re-enable the previously disabled test and perhaps drop the venv var used to disable it
 
I'll do these steps in this order. Sounds good? @matejcik @onvej-sl

ibz: > 1. You can run the emulator locally with or without the model running - it will not crash if you build with `USE_TROPIC` and the model is not running: [6c08cae](https://github.com/trezor/trezor-firmware/commit/6c08cae860b0e472d742219250d287559cb60bf3)

I am wondering whether we should still keep this one though...

onvej-sl: > I am wondering whether we should still keep this one though...

I'm in favor of removing it.

.github/workflows/core.yml
```diff
@@ -196,6 +196,7 @@ jobs:
           submodules: recursive
       - uses: ./.github/actions/environment
       - run: nix-shell --run "poetry run make -C core build_unix"
+      - run: nix-shell --run "cd vendor/ts-tvl/tvl/server/model_config && poetry install && poetry run model_server tcp -c model_config.yml &"
```
onvej-sl, line 199: I think `poetry install` can be omitted. The python environment should already be installed at this point.

core/site_scons/models/T3W1/emulator.py
matejcik: what does `tropic_init()` actually do?

onvej-sl: Three more things:
 1. I don't like that you have to launch the the tropic emulator manually if you want to run the device tests locally.
 2. We must not forget about device tests, they will also require the tropic emulator.
 3. The tropic emulator is statefull and we want to reset the state between individual tests.

matejcik: > 3\. The tropic emulator is statefull and we want to reset the state between individual tests.

this should be part of `WipeDevice` flow, and we'll need it as a feature of the firmware

> 2\. We must not forget about device tests, they will also require the tropic emulator.

i'm thinking that this may be a good reason to introduce "T3W1 without tropic" as a separate matrix item, so that we can run the test suite in CI without involving the model. (but for now it's not super important, given that the model isn't actually doing anything)

matejcik: > Yes, but `WipeDevice` resolves the issue in the device tests but not in the unit tests.

right, so there should be a function exposed to python to do this (which will also get called in the WipeDevice flow)
and Tropic unit tests will do a `setUp()` that calls this?

.github/workflows/core.yml
ibz: > I think `poetry install` can be omitted

I tried and it doesn't seem to work.

core/site_scons/models/T3W1/emulator.py
ibz: > I'm in favor of removing it.

I removed it a7180e91da71f9bd379db8bbdef9c8b205463220

.github/workflows/core.yml
matejcik: problem is that `ts-tvl` has its own pyproject.toml so as far as `poetry run` is concerned it's a separate environment.
ideally we should avoid the `cd vendor/...` thing, because `model_server` is available to us from root -- but @ibz tells me that the model is looking for some files in cwd... personally i'd keep it simple, keep the `poetry install`, and we have a different method of launching the model in the pipeline anyway
(emulator will launch it itself somewhere around `tropic_init()`)

core/SConscript.unix
```diff
@@ -21,7 +21,7 @@ if BENCHMARK and PYOPT != '0':
   print("BENCHMARK=1 works only with PYOPT=0.")
   exit(1)
 
-FEATURES_WANTED = ["input", "sd_card", "dma2d", "optiga"]
+FEATURES_WANTED = ["input", "sd_card", "dma2d", "optiga", "tropic"]
```
matejcik, line 24: firmware and kernel should also have this, right?

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  vstr_t vstr = {0};
+  vstr_init_len(&vstr, 512);
+
+  memcpy(vstr.buf, X509_cert, 512);
+
+  return mp_obj_new_str_from_vstr(&mp_type_bytes, &vstr);
+}
+STATIC MP_DEFINE_CONST_FUN_OBJ_0(mod_trezorcrypto_tropic_get_certificate_obj,
```
matejcik, line 85: please export the 512 into a constant

legacy/firmware/Makefile
```diff
@@ -165,6 +165,7 @@ CFLAGS += -I../vendor/nanopb -Iprotob -DPB_FIELD_16BIT=1 -DPB_ENCODE_ARRAYS_UNPA
 CFLAGS += -DSCM_REVISION='"$(shell git rev-parse HEAD | sed 's:\(..\):\\x\1:g')"'
 CFLAGS += -DUSE_MONERO=0
 CFLAGS += -DUSE_OPTIGA=0
+CFLAGS += -DUSE_TROPIC=0
```
matejcik, line 168: why is this needed?

core/SConscript.unix
onvej-sl: I think this has been answered in https://github.com/trezor/trezor-firmware/pull/4172#discussion_r1977775557.

matejcik (approved):
let's goooo

PR #4175, opened 2024-09-12
feat(core): show account info in ETH send/stake flow

### LINKED ISSUE #3536: Information about account missing when sending ETH.
Sending flow of ETH and potentially other altcoins does not show the information about the _account_ as is shown in bitcoin sending flow. 

Sending BTC shows info about account:
image
Equivalent information is not found when sending e.g. ETH. 

The problem was discovered while working on #3517, where similar screens were a part of the design: https://www.figma.com/file/PNqL15hf98oIFjLc0WCrvz/Sandbox-%E2%80%A2-Model-T?type=design&node-id=47-1280&mode=design&t=ApnCXlRR331Frs2h-4

ibz:
Fixes  #3536


ibz:
This depends on [#4117](https://github.com/trezor/trezor-firmware/pull/4117) being merged, otherwise the account being shown is incorrect.

ibz:
In German the menu item is a bit long.

@Hannsek what do you suggest? Leave as it is? Or change to something like "Kontoinfo." or simply "Konto"?

image

ibz:
Same in French - in that case even the staking address is too long ...

image

Hannsek:
Do we shorten it like that somewhere else? If yes, let's do it according to your suggestion. If not, let's check with Pavel.

ibz:
In Portuguese, "Staking address" is fine, but "Claim address" is too long. Also, account info is missing. I'll fix this similarly to what we decide for the others.

image

ibz:
Of course, the more I look at these things, the more I realize there are translations not fitting in menus. Oh well... this could be a never-ending battle. I'll fix those that I found, so at least I leave the part that I touched in a better state...

ibz:
New screenshots:

image

image

ibz:
OK, I ended up changing a couple of more strings. I looked at all the UI diffs and I think all of them look better in the new version. I'll paste all changes to Pavel also in case he wants to double check.

core/src/apps/ethereum/helpers.py
```diff
@@ -195,6 +195,22 @@ def format_ethereum_amount(
     return f"{amount} {suffix}"
 
 
+def get_account_and_path(address_n: list[int]) -> tuple[str | None, str | None]:
```
obrusvit, line 198: This function has logical error. Firstly, we extract `address_n[1]` without checking for correct length. And then we extract `path` and `account_path` based on whether `address_n` is truthy or not.

The path (`address_n`) is probably checked in `paths.validate_path(...)` earlier. However, we can probably improve this function and apply some defensive programming. WDYT

```python
if not address_n or len(address_n) < 2:
     return (None, None)
```

core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -1268,6 +1270,8 @@ def confirm_solana_tx(
     async def confirm_ethereum_tx(
         recipient: str,
         total_amount: str,
+        _account: str,
```
obrusvit, line 1273: we should probably make the types the same for `TT` and `TR` as for `mercury`. I.e. `str | None`

obrusvit (commented):
A few points:

core/src/apps/ethereum/helpers.py
ibz: Indeed! https://github.com/trezor/trezor-firmware/pull/4175/commits/5b267d4e892ade8cc3908d512a62abb4b25c12b4#diff-c6fb97bf1b15b1422470f5df7175252de28b102f6e3022eb3cb797dfabe49e17R203

core/src/trezor/ui/layouts/tr/__init__.py
ibz: Yes! https://github.com/trezor/trezor-firmware/pull/4175/commits/5b267d4e892ade8cc3908d512a62abb4b25c12b4#diff-b16bb8c99dd53c14619efdcdc4d8bed9b1280510417c880aa08a2644da1451b3R1273

obrusvit (approved):
utACK

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.4 regular (revision ab96ce7954c7db384b26bc863d8752150027adbf)

Hannsek:
@ibz Did you do it only for T3T1?

PR #4177, opened 2024-09-13
chore(core): remove duplicated FlowMsg conversion

obrusvit:


mmilata (approved):
This led me to investigating how to get rid of `FlowMsg::Text` but it's not going to be easy so let's merge.

obrusvit:
> This led me to investigating how to get rid of `FlowMsg::Text` but it's not going to be easy so let's merge.

Yes..
related comment from matejcik: https://github.com/trezor/trezor-firmware/pull/4094#pullrequestreview-2245526322

PR #4180, opened 2024-09-13
Swipe detection refactors

matejcik:
a collection of simple and hopefully inoffensive refactors. (a byproduct of the upcoming GFL PR, reviewable separately)

the commits are self-explanatory -- when reviewing, please go by commits.


core/src/trezor/ui/layouts/progress.py
```diff
@@ -40,7 +40,7 @@ def __init__(
             ui.refresh()
         ui.backlight_fade(ui.BacklightLevels.NORMAL)
 
-    def set_timer(self, token: int, deadline: int) -> None:
+    def set_timer(self, token: int, duration: int) -> None:
```
obrusvit, line 43: This was the problem also before but I don't like that the unit of `duration` is not apparent until very deep in the code (instantiation of `Duration` in rust). Can we either document that `duration` is in milliseconds or rename to `duration_ms`?

obrusvit (commented):
Basically agreed to all the commits but I skipped 55c7759d61eb6508f30592d7862d6fa6c9e88ba6 which I want to understand a little bit better.

core/embed/rust/src/ui/flow/swipe.rs
```diff
@@ -234,41 +219,20 @@ impl SwipeFlow {
...
-                                state_change = self.state_unchanged();
-                                attach = true;
-                            } else if dir == SwipeDirection::Down && current_state > 0 {
-                                self.internal_state -= 1;
-                                state_change = self.state_unchanged();
-                                attach = true;
-                            }
-                        }
+                    let new_internal_state =
+                        config.paging_event(dir, self.internal_state, self.internal_pages);
```
mmilata, line 215: While at it, can we please rename `internal_state` to something else? From the current name it's not clear it's related to `internal_pages`, and also "state" everywhere else is usually an enum and not integer. Perhaps:
* `internal_state` -> `current_internal_page`
* `internal_pages` -> `internal_page_count`

mmilata (approved):
not offensive at all

core/embed/rust/src/ui/flow/swipe.rs
matejcik: i'd prefer to leave this alone here

(a) i have a draft of complete removal of internal state here
(b) after your suggestion, I actually tried the following:

```
#[derive(Copy, Clone)]
pub struct Pager {
   current: u16,
   total: u16,
}
```
which I'd like to use pervasively for paging, but that is definitely its own separate PR

core/src/trezor/ui/layouts/progress.py
matejcik: like so? https://github.com/trezor/trezor-firmware/pull/4180/commits/13dd0bad8637b748dd781fc8f4a9c5152b3e8958

core/embed/rust/src/ui/flow/swipe.rs
obrusvit: Agreed with the `Pager`, or something similar. It would be nice to unify the way we get information about number of pages and current page. We have:
- `Paginate` (this one actually has also `change_page`),
- `Swipable` provides `get_internal_page_count`,  
- `InternallySwipable` provides `current_page` and `num_pages`
- then we have `PageMap`
- I needed to add a similar function to `Paragraphs` recently
- probably more..

obrusvit (approved):

PR #4181, opened 2024-09-14
chore(core): enable compile_commands.json gen

Redblaze420:
This commit adds COMPILATION_DB flag for SCons which generates
compile_commands.json - a compilation database for C/C++ language
servers. The flag is dependent on CMAKELISTS flag.

[no changelog]


obrusvit:
:thinking: thank you for opening this PR. We were discussing internally whether we include this option or not. No decision yet.

PR #4182, opened 2024-09-15
refactor(core/mercury): separate upy args parsing

obrusvit:
This PR separates the Rust flows from micropython layer.

The idea is to do the uPy argument parsing in `ui/model_mercury/layout.rs` and then pass them as regular arguments to code in `ui/model_mercury/flow/**.rs`, i.e. they won't be passed as `args: &[Obj], kwargs: &Map`. The goal is to keep the rust flows independent from `micropython` sub-crate. This separation is done for two reasons:
- it simplifies the code in `flows` creation
- it paves the way for apps done completely in Rust and for [UiFeatures](#4322)

This PR has no impact on UI or features.

Note:
- there is a few places with the usage of `Obj` and `IterBuf` - last place is `get_address.rs`, **postponing this for now**

### DIFF
--- a/core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
+++ b/core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
@@ -2,7 +2,6 @@ use heapless::Vec;
 
 use crate::{
     error,
-    micropython::{iter::IterBuf, obj::Obj, util},
     strutil::TString,
     translations::TR,
     ui::{
@@ -213,44 +212,30 @@ fn get_cancel_page(
 
 #[allow(clippy::too_many_arguments)]
 pub fn new_confirm_output(
-    title: Option<TString<'static>>,
-    subtitle: Option<TString<'static>>,
+    main_params: ConfirmBlobParams,
     account: Option<TString<'static>>,
     account_path: Option<TString<'static>>,
     br_name: TString<'static>,
     br_code: u16,
-    message: Obj,
-    amount: Option<Obj>,
-    chunkify: bool,
-    text_mono: bool,
-    address: Option<Obj>,
-    address_title: Option<TString<'static>>,
-    summary_items: Obj,
-    fee_items: Obj,
-    summary_title: Option<TString<'static>>,
+    content_amount_params: Option<ConfirmBlobParams>,
+    address_params: Option<ConfirmBlobParams>,
+    address_title: TString<'static>,
+    summary_items_params: Option<ShowInfoParams>,
+    fee_items_params: ShowInfoParams,
     summary_br_name: Option<TString<'static>>,
     summary_br_code: Option<u16>,
     cancel_text: Option<TString<'static>>,
 ) -> Result<SwipeFlow, error::Error> {
     // Main
-    let main_content = ConfirmBlobParams::new(title.unwrap_or(TString::empty()), message, None)
-        .with_subtitle(subtitle)
-        .with_menu_button()
-        .with_footer(TR::instructions__swipe_up.into(), None)
-        .with_chunkify(chunkify)
-        .with_text_mono(text_mono)
-        .with_swipe_up()
+    let main_content = main_params
         .into_layout()?
         .one_button_request(ButtonRequest::from_num(br_code, br_name));
 
     // MainMenu
     let mut main_menu = VerticalMenu::empty();
     let mut main_menu_items = Vec::<usize, 3>::new();
-    if address.is_some() {
-        main_menu = main_menu.item(
-            theme::ICON_CHEVRON_RIGHT,
-            address_title.unwrap_or(TR::words__address.into()),
-        );
+    if address_params.is_some() {
+        main_menu = main_menu.item(theme::ICON_CHEVRON_RIGHT, address_title);
         unwrap!(main_menu_items.push(MENU_ITEM_ADDRESS_INFO));
     }
     if account.is_some() && account_path.is_some() {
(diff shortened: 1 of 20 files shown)


obrusvit:
Similar thing was done here for bootloader and "common". Need to take a look how to take the same/similar approach for firmware. https://github.com/trezor/trezor-firmware/pull/3658

obrusvit:
The only conflict with #3686 should be: `Layout::new` moved to `layout.rs` and renamed to `Layout::new_obj`

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
```diff
@@ -212,61 +210,30 @@ fn get_cancel_page(
...
-    let summary_br_name: Option<TString> = kwargs
-        .get(Qstr::MP_QSTR_summary_br_name)?
-        .try_into_option()?;
-    let summary_br_code: Option<u16> = kwargs
-        .get(Qstr::MP_QSTR_summary_br_code)?
-        .try_into_option()?;
-
-    let cancel_text: Option<TString> = kwargs.get(Qstr::MP_QSTR_cancel_text)?.try_into_option()?;
-
+pub fn new_confirm_output(
```
mmilata, line 213: Other possibility is to use `ConfirmOutput::new` at the cost of additional indent. Or module-level `fn new` if there's just one flow.

Also how is clippy not complaining about argument count? I suppose it's because clippy ignores all UI code that is not model T?

core/embed/rust/src/ui/model_mercury/flow/get_address.rs
```diff
@@ -84,153 +84,142 @@ impl FlowController for GetAddress {
...
+    title: TString<'static>,
+    description: Option<TString<'static>>,
+    extra: Option<TString<'static>>,
+    address: Obj, // TODO: get rid of Obj
+    chunkify: bool,
+    address_qr: TString<'static>,
+    case_sensitive: bool,
+    account: Option<TString<'static>>,
+    path: Option<TString<'static>>,
+    xpubs: Obj, // TODO: get rid of Obj
```
mmilata, line 97: Perhaps this can use some kind of closure?

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -1540,16 +1866,16 @@ pub static mp_module_trezorui2: Module = obj_module! {
     ///     br_code: ButtonRequestType,
     ///     br_name: str,
     /// ) -> LayoutObj[tuple[UiResult, int]]:
-    ///     """Numer input with + and - buttons, description, and context menu with cancel and
+    ///     """Numebr input with + and - buttons, description, and context menu with cancel and
```
mmilata, line 1869:
```suggestion
    ///     """Number input with + and - buttons, description, and context menu with cancel and
```

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
```diff
@@ -212,40 +211,29 @@ fn get_cancel_page(
 }
 
 pub fn new_confirm_output(
-    title: Option<TString<'static>>,
-    subtitle: Option<TString<'static>>,
+    main_params: ConfirmBlobParams,
```
mmilata, line 214: Huge struct with most members unused is also not ideal, but I suppose we can live with it if it doesn't cause stack overflows. This is probably heading in the direction of builder pattern / `ConfirmOutputParams`.

mmilata (approved):
Wrote some thoughts inline but overall this should be a harmless refactor that can be merged when CI passes.

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
obrusvit: Rationale for using standalone functions instead of putting it to `impl X` is to be consistent across all flows. Some flows have multiple enums defined (various number of states) and then it's not clear to which `impl X/Y/Z` to put it.

core/embed/rust/src/ui/model_mercury/flow/confirm_output.rs
obrusvit: Definitely, I'd like to put more thought into structuring these params. But felt this is a good first step as the flows creation is bloated even without arg parsing.

obrusvit:
Re-targeting to GFL branch #3686  and it can go right after.

core/embed/rust/src/ui/model_mercury/layout.rs
obrusvit: Making a typo while fixing a typo.
Fixed on rebase.

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -550,6 +490,216 @@ extern "C" fn new_confirm_homescreen(n_args: usize, args: *const Obj, kwargs: *m
...
+            .get(Qstr::MP_QSTR_summary_br_name)?
+            .try_into_option()?;
+        let summary_br_code: Option<u16> = kwargs
+            .get(Qstr::MP_QSTR_summary_br_code)?
+            .try_into_option()?;
+
+        let cancel_text: Option<TString> =
+            kwargs.get(Qstr::MP_QSTR_cancel_text)?.try_into_option()?;
+
+        let main_params = ConfirmBlobParams::new(
```
obrusvit, line 552: `ConfirmBlobParams::new` is used with lots of `None`s and `false` in this function. I consider removing these from `new` and instantiate them only using `with_xyz` constructs.

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -581,11 +655,33 @@ extern "C" fn new_confirm_summary(n_args: usize, args: *const Obj, kwargs: *mut
...
+            .with_footer(TR::instructions__swipe_up.into(), None)
+            .with_swipe_up();
+        for pair in IterBuf::new().try_iterate(items)? {
+            let [label, value]: [TString; 2] = util::iter_into_array(pair)?;
+            summary_params = unwrap!(summary_params.add(label, value));
+        }
+
+        let mut account_params =
+            ShowInfoParams::new(TR::send__send_from.into()).with_cancel_button();
+        for pair in IterBuf::new().try_iterate(account_items)? {
```
ibz, line 669: Can't this be extracted into a `ShowInfoParams::add_items`?

ibz (approved):

core/embed/rust/src/ui/model_mercury/layout.rs
obrusvit: Not without making `ShowInfoParams` dependent on `micropython` crate.

obrusvit:
I had to use `#[allow(clippy::too_many_arguments)]` for a few `SwipeFlow` constructors. 

This could be resolved as a part of #4322

matejcik (approved):
utACK

matejcik:
i'm thinking though, when do we bite the bullet on proc macros? :)
jpochyla was **strongly** cautioning against it, but making a `#[upy_function_call] fn confirm_action(title: TString, description: Option<TString>, ...)` is _the_ perfect usecase

PR #4194, opened 2024-09-18
refactor(core): simplify fonts handling

obrusvit:
This PR aims to further simplify the font handling in `core`. 

The main objective was to get rid of the numerous macros, which are quite cumbersome to work with. This is achieved by introducing `struct font_info_t` which is filled with needed data during font generation by `gen_font.py`. Then there is an array `font_registry` defined In the `core/embed/lib/fonts/fonts.c` which is used to access font data.

Unfortunately, I was unable to get rid of the `FONT_MAX_HEIGHT` calculation as it must be done in compile time. However, there is only one usage so getting rid of it is perhaps feasible. https://github.com/trezor/trezor-firmware/blob/02533aa27a44ed661166465648438f62cd6a0449/core/embed/lib/buffers.h#L32


obrusvit:
All UI-diffs on `TS3` seems to be caused by now usage of `BOLD` font in `Size: x bytes` which is actually correct.
image

obrusvit:
Can the change of order in translation jsons cause some problems moving forward @matejcik ? If so, I can probably revert it and just define the `enum font_id_t` starting from `1` and not change the order. I simply though that the new ordering is a little bit more logical, e.g. having `NORMAL` and `NORMAL_UPPER` in succession. 
image

matejcik:
yeah the ids are baked into the translation blobs, this would really mess up the upgrade/downgrade scenario. please revert.

aside, I believe that there was a reason for skipping 0 for font index, but perhaps it doesn't apply anymore with this refactor?

obrusvit:
I reverted the re-assignment of font IDs. The PR is nor ready for review.

For future reference: after we drop `buffers.h/.c` when getting rid of "old rendering", then we can completely remove these definitions for all fonts (also remove this from `gen_font.py` and also remove the `FONT_MAX_HEIGHT` calculation in `fonts.h`
`#define Font_PixelOperatorMono_Regular_8_MAX_HEIGHT 8`

core/embed/lib/fonts/fonts_types.h
```diff
@@ -0,0 +1,34 @@
+#ifndef _FONTS_TYPES_H
```
matejcik, line 1: any particular reason you made this a separate file, instead of putting the data into fonts.h ?

core/embed/lib/fonts/fonts.c
```diff
@@ -22,127 +22,126 @@
...
+int font_max_height(font_id_t font) {
+  const font_info_t *font_info = get_font_info(font);
+  return font_info ? font_info->max_height : 0;
+}
+
+int font_baseline(font_id_t font) {
+  const font_info_t *font_info = get_font_info(font);
+  return font_info ? font_info->baseline : 0;
+}
+
```
matejcik, line 113: kind of feels like those should be removed

matejcik (commented):
overall comment: why not get rid of `font_id_t` and instead use `const struct font_info_t*` pervasively? then it's the responsibility of the (first) caller to pass in a valid font_info_t, and everyone else can look at the attributes directly.

then instead of `FONT_NORMAL` being an enum value, you would use `&FONT_NORMAL` as a pointer to a const struct font_info_t in all places

in Rust you would change `pub enum Font` to
```
pub type FontInfo = ffi::font_info_t;
pub type Font = &'static FontInfo;
```
impl font to impl fontinfo, modify the `self` owning arguments to references because FontInfo is no longer Copy/Clone, adjust the rest...

one thing we'd need to do is keep the mapping to translation blob font ids somewhere. presumably only in Rust? because it's only Rust that interfaces with translation blobs

Hannsek:
How are we with this one? @obrusvit

core/embed/lib/fonts/fonts.c
obrusvit: all these functions are used via `ffi` from rust

core/embed/lib/fonts/fonts_types.h
obrusvit: I put these fefinitions to `fonts.h` on rebase.

core/embed/lib/fonts/fonts.c
matejcik: make rust call `get_font_info`? :woman_shrugging: that's the direction we want to be going

obrusvit: As of this commit 9f987971882f51f7b3a3afc4675d7058b7a84dc7, the bindings to rust only contain type `font_info_t` and the function `get_font_info`.

I didn't remove them completely yet because they are used from C code.

I'll now try to remove `font_id_t` completely as you suggested.

core/embed/gfx/fonts/fonts.h
```diff
@@ -21,7 +21,7 @@
 #define _FONTS_H
 
 #include <stdbool.h>
-
+#include <stdint.h>
```
cepetr, line 24: Instead of including standard headers, we now include `trezor_types.h` in API files and `trezor_rtl.h` in implementation files. Therefore, this should be `#include <trezor_types.h>`. But in general there's nothing wrong on using `stdint.h` in  generated files like `font_unifont_bold_16.h`.

core/embed/rust/src/trezorhal/display.rs
```diff
@@ -5,40 +5,19 @@ use core::ptr;
...
 }
 
-pub fn text_width(text: &str, font: i32) -> i16 {
+pub fn get_font_info(font: i32) -> Option<FontInfo> {
     unsafe {
-        ffi::font_text_width(font, text.as_ptr() as _, text.len() as _)
-            .try_into()
-            .unwrap_or(i16::MAX)
+        let font = ffi::get_font_info(font);
+        Some(*font.as_ref()?)
```
matejcik, line 17: this could use a SAFETY comment -- something along the lines of "ffi::get_font_info returns a pointer to ROM data which is 'static and immutable"

core/embed/rust/src/ui/display/font.rs
```diff
@@ -198,31 +202,28 @@ impl Font {
...
-        }
+        self.get_first_glyph_from_text(text)
+            .map_or(0, |glyph| glyph.bearing_x)
     }
 
     pub fn char_width(self, ch: char) -> i16 {
-        display::char_width(ch, self.into())
+        let mut buf = [0u8; 4];
+        let encoding = ch.encode_utf8(&mut buf);
+        self.text_width(encoding)
```
matejcik, line 212: you should be able to implement this directly as `self.get_glyph(c).adv`, and compute `text_width` by mapping to `char_width`
(btw is `char_width` used anywhere?)

core/embed/rust/src/ui/display/font.rs
```diff
@@ -198,31 +202,28 @@ impl Font {
...
     }
 
     pub fn text_height(self) -> i16 {
-        display::text_height(self.into())
+        display::get_font_info(self.into()).map_or(i16::MAX, |font| font.height.try_into().unwrap())
```
matejcik, line 216: i'm thinking, just `unwrap!()` the whole thing, asking for a text_height on an invalid font id should be an unrecoverable error, no? (dtto below)

core/embed/rust/src/ui/display/font.rs
```diff
@@ -248,14 +249,33 @@ impl Font {
         (start + end + self.visible_text_height(text)) / 2
     }
 
+    fn get_glyph_data(&self, c: u16) -> *const u8 {
```
cepetr, line 252: How about renaming `c` to `codepoint` here?

matejcik, line 252: this is very `unsafe`
but we could make it safe:
1. put a proper lifetime on `Glyph`
2. make a struct `GlyphData` that owns an `Option <RwLockReadGuard<...>>` (the same type that `flash::translations::get()` is returning) and also a `FontInfo` (or `&'static FontInfo` or `&'a FontInfo` or sth :woman_shrugging: )
3. move function `get_glyph` to that struct
4. make private `Font::get_glyph_data` that makes the GlyphData from `self` and `translations::flash::get`
5. make `pub fn Font::with_glyph_data(self, closure)` that will give a new instance of `GlyphData` to the inner closure (and add a comment that we do this to make it somewhat clearer that **GlyphData hold a lock to translations** and that if you store it long-term, you'll break things)
6. rewrite the only user of `get_glyph`, that is `shape::text::Text`, to use `with_glyph_data` in its `draw()` function

core/embed/gfx/fonts/fonts.c
```diff
@@ -24,300 +24,127 @@
 #include "librust_fonts.h"
```
cepetr, line 24: We no longer need to include `librust_fonts.h`, and we can delete the file as well.

cepetr (changes requested):
It looks good to me. I've just commented on a few minor issues.

core/embed/gfx/fonts/fonts.c
obrusvit: d42cea3818e154945ac7649bb2031822192adddb

core/embed/rust/src/ui/display/font.rs
obrusvit: 45d747fa91f2d5e8095a70708a124de6ca69186b

but I'll improve the function using matejcik points

core/embed/rust/src/trezorhal/display.rs
obrusvit: dd36660f473954787c3d6929f4d122f3f2acfb9c

core/embed/rust/src/ui/display/font.rs
obrusvit: 4660ea4d48725ac94a70b862201cc815abb63d10

core/embed/rust/src/ui/display/font.rs
matejcik: (and of course make `get_utf8_glyph` would be converted to a method of `Translation`)

core/embed/gfx/fonts/fonts.h
obrusvit: https://github.com/trezor/trezor-firmware/commit/d42cea3818e154945ac7649bb2031822192adddb

core/embed/rust/src/ui/display/font.rs
obrusvit: On it but the intermediate issue is that `Translation` cannot be included for bootloader which makes working with the `RwLock` more complicated.

matejcik: the member of `GlyphData` will be feature-gated then?

obrusvit: See here: 02aba6b8d317bef7c4a571a7c7d67064e07b51c2
Handling of translations is feature gated.

Some points to discuss:
- `fn GlypData::get_glyph_data` returns `nullptr` if `get_info` fails
- `pub fn GlyphData::get_glyph` - not sure if the `ensure!` should be there w.r.t. the point above
- `Translations::get_utf8_glyph` is now simplified, perhaps the commentary could reviewed

matejcik: left some suggestions in code. generally: we can stop mucking with pointers which makes some things much nicer now

matejcik (changes requested):

core/embed/rust/src/ui/display/font.rs
```diff
@@ -22,10 +29,10 @@ pub struct Glyph {
     pub bearing_x: i16,
     /// Top-side vertical bearing
     pub bearing_y: i16,
-    data: &'static [u8],
+    data: &'a [u8],
 }
 
-impl Glyph {
+impl<'a> Glyph<'a> {
```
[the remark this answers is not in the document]
obrusvit: mucking with pointers reduced but  `fn GlyphData::get_glyph_data` needs to `offset` into `FontInfo::glyph_data` which is `*const *const u8` from C. So I moved the calculation of total size into a separate function `GlyphData::calculate_glyph_size` and the calculation is now not done in `Glyph::load` because it takes a slice which has `len()`.

The result is two `unsafe` blocks in `get_glyph_data`

See: cadcede0b73ebecaac79d3ea3e01acbbdb77912b

core/embed/rust/src/ui/display/font.rs
```diff
@@ -106,7 +113,80 @@ impl Glyph {
...
+            c => c,
+        };
+        let gl_data = self.get_glyph_data(ch as u16);
+
+        ensure!(!gl_data.is_null(), "Failed to load glyph");
+        // SAFETY: Glyph::load is valid for data returned by get_char_glyph
+        unsafe { Glyph::load(gl_data) }
+    }
+
+    fn get_glyph_data(&self, codepoint: u16) -> *const u8 {
```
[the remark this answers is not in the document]
obrusvit: cadcede0b73ebecaac79d3ea3e01acbbdb77912b

core/embed/rust/src/translations/blob.rs
```diff
@@ -258,6 +258,23 @@ impl<'a> Translations<'a> {
...
+    /// Returns a pointer to the glyph data for the given UTF-8 codepoint.
+    ///
+    /// SAFETY: Do not mess with the lifetimes in this signature.
+    ///
+    /// The lifetimes are a useful lie that bind the lifetime of the returned
+    /// string not to the underlying data, but to the _reference_ to the
+    /// translations object. This is to facilitate safe interface to
+    /// flash-based translations. See docs for `flash::get` for details.
+    #[allow(clippy::needless_lifetimes)]
+    pub fn get_utf8_glyph<'b>(&'b self, codepoint: u16, font_index: u16) -> *const u8 {
```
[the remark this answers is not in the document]
obrusvit: cadcede0b73ebecaac79d3ea3e01acbbdb77912b

core/embed/rust/src/ui/display/font.rs
```diff
@@ -198,31 +271,27 @@ impl Font {
...
-            0
-        }
+        text.chars().next().map_or(0, |c| {
+            self.with_glyph_data(|data| data.get_glyph(c).bearing_x)
+        })
     }
 
     pub fn char_width(self, ch: char) -> i16 {
-        display::char_width(ch, self.into())
+        self.with_glyph_data(|data| data.get_glyph(ch).adv)
```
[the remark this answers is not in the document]
obrusvit: I just removed the delegation to `char_width` and instead put the whole loop in `with_glyph_data` and get the `advance` from glyph data where needed. I don't think it's too much repetition..
77b3d79e01625e077d88fcf01125ba75a559cd25

Good catch, I initially did the same mistake with other functions... I wrote `for ... with_data(|..| )` instead of `with_data(|..| for ...)`

core/embed/rust/src/ui/display/font.rs
obrusvit: I kept it the same as it is in other equivalent functions in the file. Do you insist on `unwrap!()` instead?

core/embed/rust/src/ui/display/font.rs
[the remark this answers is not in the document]
obrusvit: Made a mistake 74ad10d44127b41b970a1d7e643abe64f2bcfd11

core/embed/rust/src/translations/blob.rs
```diff
@@ -258,6 +258,19 @@ impl<'a> Translations<'a> {
     pub fn header<'b>(&'b self) -> &'b TranslationsHeader<'b> {
         &self.header
     }
+
+    /// Returns a pointer to the glyph data for the given UTF-8 codepoint.
```
matejcik, line 262: update the comment please

core/embed/rust/src/ui/display/font.rs
```diff
@@ -22,40 +29,30 @@ pub struct Glyph {
...
-    /// is pointing to a memory containing a valid glyph data, that is:
-    /// - contains valid glyph metadata
-    /// - data has appropriate size
-    /// - data must have static lifetime
-    pub unsafe fn load(data: *const u8) -> Self {
-        unsafe {
-            let width = *data.offset(0) as i16;
-            let height = *data.offset(1) as i16;
-
-            let data_bytes = match constant::FONT_BPP {
```
matejcik, line 43: you should use `calculate_glyph_size` here and `assert!` that the data length matches (or maybe debug_assert? debug_assert is fine for builtin fonts, unsure about the external ones)

core/embed/rust/src/ui/display/font.rs
```diff
@@ -85,7 +82,7 @@ impl Glyph {
         c_data >> 4
     }
 
-    pub fn bitmap(&self) -> Bitmap<'static> {
+    pub fn bitmap(&self) -> Bitmap<'a> {
```
matejcik, line 85: speaking of: it looks like `Bitmap::new` does not check that there's enough data provided here? @cepetr can you comment?

core/embed/rust/src/ui/display/font.rs
```diff
@@ -106,7 +103,106 @@ impl Glyph {
...
+
+    pub fn get_glyph(&self, ch: char) -> Glyph<'_> {
+        let ch = match ch {
+            '\u{00a0}' => '\u{0020}',
+            c => c,
+        };
+        let gl_data = self.get_glyph_data(ch as u16);
+
+        ensure!(gl_data.is_some(), "Failed to load glyph");
+        Glyph::load(gl_data.unwrap())
```
matejcik, line 148:
```suggestion
        Glyph::load(unwrap!(gl_data, "Failed to load glyph"))
```

core/embed/rust/src/ui/display/font.rs
```diff
@@ -106,7 +103,106 @@ impl Glyph {
...
+    }
+
+    fn calculate_glyph_size(header: &[u8]) -> usize {
```
matejcik, line 151: re earlier comment, this should probably be a top-level function so that `Glyph` can also use it

core/embed/rust/src/ui/display/font.rs
```diff
@@ -106,7 +103,106 @@ impl Glyph {
...
+                }
+            } else {
+                #[cfg(feature = "translations")]
+                {
+                    if codepoint >= 0x7F {
+                        // UTF8 character from embedded blob
+                        if let Some(guard) = &self.translations_guard {
+                            if let Some(translations) = guard.as_ref() {
+                                if let Some(glyph) =
+                                    translations.get_utf8_glyph(codepoint, self.font as u16)
```
matejcik, line 186: how about something like
```suggestion
    if let Some(glyph) = &self
        .translations_guard
        .map(|g| g.as_ref())
        .map(|t| t.get_utf8_glyph(codepoint, self.font as u16)) {
```

core/embed/rust/src/ui/display/font.rs
```diff
@@ -106,7 +103,106 @@ impl Glyph {
...
+                            }
+                        }
+                    }
+                }
+                unsafe {
+                    let ptr = font_info.glyph_nonprintable;
+                    let header = slice::from_raw_parts(ptr, 2);
+                    let full_size = Self::calculate_glyph_size(header);
+                    slice::from_raw_parts(ptr, full_size)
+                }
```
matejcik, line 199: this should be a its own `unsafe fn` (we could even go `Glyph::from_slice` and `unsafe fn Glyph::from_ptr`, but don't have to do that here)
AND there should be `pub fn GlyphData::nonprintable(&self) -> Glyph` that just calls this unsafe fn on `font_info.glyph_nonprintable`

core/embed/rust/src/ui/display/font.rs
matejcik: yeah, I meant in all of them, not in text_height specifically.
i think `unwrap!` is better here, it seems pointless to spend code on returning value for a nonexistent font?

matejcik (changes requested):

core/embed/rust/src/translations/blob.rs
obrusvit: 5ae0e784a26b8f4a1baa37816bf083baff6f603b

core/embed/rust/src/ui/display/font.rs
obrusvit: 80fa6c254de6a73d4742e3190d0eb8fad00985ba

core/embed/rust/src/ui/display/font.rs
obrusvit: 74a84516f9303f0b95cdd9a55e5457fb06898592

core/embed/rust/src/ui/display/font.rs
obrusvit: 74a84516f9303f0b95cdd9a55e5457fb06898592

core/embed/rust/src/ui/display/font.rs
obrusvit: 74a84516f9303f0b95cdd9a55e5457fb06898592

core/embed/rust/src/ui/display/font.rs
matejcik: ah but there's still an `unwrap()` in there. we don't want to call `fn unwrap` basically ever, always use `unwrap!`

this should be `unwrap!(display::get_font_info(self.into())).height as i16` (honestly wouldn't worry about doing a try_into ... speaking of, why is it an `int` in the struct and not `uint16_t` ?)

obrusvit: I see. `unwrap()` removed.

I kept `int` as it was before. If you don't mind, as changing it to `uint16_t` is a slightly larger change (fonts generator, generated fonts, function headers and callers), I'd postpone it to another PR not to prolong this one.

matejcik (approved):

core/embed/rust/src/ui/display/font.rs
matejcik: not stopping the PR on it anymore but you could have avoided the map() if you just unwrap!() the result of get-font-info, which would have given you an infallible non-option struct with the height attribute

obrusvit: yes, removed the `map` on rebase

obrusvit:
@matejcik this check at `Glyph::load` discovered an error in the `gen_font.py`.
```rust
ensure!(data.len() == size, "Invalid glyph data size");
```
The problem was that padding to full bytes was applied also if the glyph was already aligned. Then the size of the data read from translation blob and the expected size calculated from `width` and `height` didn't match.
For example, the czech glyph `ý` was wrongly aligned causing a crash in some `cs` tests:
```json
"ý": "060807000718ccf3cdf0de00",  // wrong
"ý": "060807000718ccf3cdf0de",    // fixed
```

Note that the problem was invisible for regular ASCII glyphs (in C arrays) because the last zero byte was simply ignored. However, I fixed it also for these cases.

Please, see c09fd18fcc4a1e55b05daa14002a21d914af4ed5

cepetr (approved):

matejcik:
good fix but this also means that this will cause a nasty crash at text render time if you run with an outdated translation blob
please change the check to `data.len() >= size`. we don't really care about trailing garbage, the font data is high-trust.

PR #4195, opened 2024-09-19
bootloader production mode fix

TychoVrahe:
Fixes stuck bootloader when in production mode and touch is not present.

(some touch related typos fixed too)


matejcik (approved):
utACK

core/embed/bootloader/main.c
```diff
@@ -376,18 +376,25 @@ int bootloader_main(void) {
 
   unit_variant_init();
 
-#ifdef USE_TOUCH
 #ifdef TREZOR_MODEL_T3T1
   // on T3T1, tester needs to run without touch, so making an exception
   // until unit variant is written in OTP
-  if (unit_variant_present()) {
-    ensure(touch_init(), "Touch screen panel was not loaded properly.");
+  secbool production_mode = unit_variant_present() ? secfalse : sectrue;
```
cepetr, line 382: How about something like this:

```c

#ifdef USE_TOUCH
secbool allow_touchless_mode = secfalse;
#ifdef TREZOR_MODE_T3T1
secbool manufacturing_mode = unit_variant_present() ? secfalse : sectrue;
allow_touchless_mode =  manufacturing_mode;
secbool touch_initialized = touch_init();
ensure((touch_initialized != secfalse) || (allow_touchless_mode != secfalse), "Touch panel...");
#endif  // TREZOR_MODE_T3T1
#endif // USE_TOUCH

..
..
..

while (allow_touchless_mode == secfalse && touch_ready() != sectrue) {
      hal_delay(1);
    }



```

cepetr (changes requested):

core/embed/bootloader/main.c
TychoVrahe: yeah its a bit better. Adjusted a bit to allow touch initialization for other models too, see https://github.com/trezor/trezor-firmware/commit/6187233b5d45a2afdef8ed300face47ea777c8e0

cepetr (approved):
utACK

PR #4202, opened 2024-09-21
Bootloader 2.1.9

matejcik:
chore commits for #4195

### DIFF
--- a/core/embed/bootloader/CHANGELOG.md
+++ b/core/embed/bootloader/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 2.1.9 [September 2024]
+
+### Changed
+
+- [T3T1] Ignore touch errors while in manufacturing.  [#4195]
+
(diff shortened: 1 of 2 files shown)


core/embed/bootloader/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 2.1.9 [September 2024]
+
+### Changed
+
+- [T3T1] Allow starting bootloader without touch controller.  [#4195]
```
TychoVrahe, line 11: Not entirely sure about this, its only allowed in manufacturing mode, it shouldn't affect regular users.

matejcik: hmmm ok so how about something like this?

> Ignore touch errors while in manufacturing.

TychoVrahe: i think its better

matejcik: force-pushed

TychoVrahe (approved):

PR #4216, opened 2024-09-24
fix(core): restore ellipsis functionality on TR

matejcik:
#4121 broke a little thing that Rust's unused variable detection told me about

for passphrase, this PR restores to pre-4121 state
the PIN confuses me a little, it might be more broken than what this single change affected?


Hannsek:
I cannot see any diffs. Is it expected?

Hannsek:
So the only change is for T3B1 in passphrase entry?

TychoVrahe (approved):
Ack.

The pin input is also affected, but a bit differently, it doesn't show the dots it just limits number of characters rendered. The difference is that `PinEntry` uses this `ChangingText` with `without_ellipsis` called. 

The `translations` fixtures need updating too probably.

obrusvit:
This can be abandoned in favor of https://github.com/trezor/trezor-firmware/pull/4328

bieleluk:
The fix has already been merged to `main` as part of the global layout implementation:
https://github.com/trezor/trezor-firmware/blob/main/core/embed/rust/src/ui/model_tr/component/changing_text.rs#L147

PR #4219, opened 2024-09-25
streamline layout definitions

TychoVrahe:
This PR is aimed to clean some mess in layout definitions, sconscripts and other minor stuff.

1) flash layout definitions in model_xxx.h are given some structure and consistency
2) this then allows to use single layout.c (per mcu family) file instead of model specific layout file.
3) sconscripts are cleaned up a bit, removed some residual model 1 mentions, got rid of exceptions for DISCovery kits, so that i.e. prodtest build is allowed for these too, reusing a recently added embed_raw_binary function to embed vendorheader
4) added bootloader binaries for discovery kits, so that we can test bootloader upgrading on these kits
5) SBU interface is removed from micropython, it was unused, since core-split it wouldn't work anyway and took some flash space

Note: RAM layout in model definitions is unchanged for now, but probably should get similar treatment at some point.


cepetr (approved):
Surprisingly, I didn't find any issues. Looks good to me.

Hannsek:
The T3T1 device tests are red. Is it intentional? 
image

PR #4221, opened 2024-09-25
fix(core): fix handover to bootloader

cepetr:
This PR fixes the firmware deadlock issue that occurred after the interaction-less update. Th bug was introduced in PR #4188, and only affects T and TS3 models. 

The fix adds a transition from handler mode to thread mode before jumping to the bootloader.

### DIFF
--- a/core/embed/trezorhal/stm32f4/bootutils.c
+++ b/core/embed/trezorhal/stm32f4/bootutils.c
@@ -82,6 +82,65 @@ static inline void __attribute__((always_inline)) delete_secrets(void) {
 }
 #endif  // STM32U5
 
+#ifdef STM32F4
+// Ensure that we are running in privileged thread mode.
+//
+// This function is used only on STM32F4, where a direct jump to the
+// bootloader is performed. It checks if we are in handler mode, and
+// if so, it switches to privileged thread mode.
+__attribute((naked, no_stack_protector)) static void ensure_thread_mode(void) {
+  __asm__ volatile(
+      // --------------------------------------------------------------
+      // Check if we are in handler mode
+      // --------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


core/embed/trezorhal/stm32f4/bootutils.c
```diff
@@ -82,6 +82,65 @@ static inline void __attribute__((always_inline)) delete_secrets(void) {
...
+      "IT       EQ                 \n"
+      "BXEQ     LR                 \n"  // return if in thread mode
+
+      // --------------------------------------------------------------
+      // Disable FP registers lazy stacking
+      // --------------------------------------------------------------
+
+      "LDR     R1, = 0xE000EF34    \n"  // FPU->FPCCR
+      "LDR     R0, [R1]            \n"
+      "BIC     R0, R0, #1          \n"  // Clear LSPACT to suppress lazy
```
TychoVrahe, line 110: from the description of `LSPACT` bit in programming manual:
`Lazy state preservation active. Indicates whether lazy preservation of the floating-point state is active.`
i would be expecting that this bit is managed by hardware and should be used by software as read only. Are you sure about this? For disabling of lazy stacking, i would expect `LSPEN` bit to be used, or by suppression do you mean that this should be only one time suppressed and this actually achieves it?

cepetr: Yes, it's a one-time suppression.

If we hadn't done it here, the first FPU instruction used in the bootloader might cause lazy stacking and an unexpected overwrite of a portion of SRAM. There are some possible alternatives: 1) use any FPU instruction here instead of writing to `FPU->FPPCR`, which causes immediate stacking of FPU registers and clears the LSPACT bit, 2) possibly returning from handler mode with a different EXC_RETURN code... 

I don't like it much either understand it looks a bit hacky.

TychoVrahe (approved):

PR #4225, opened 2024-09-27
refactor ui configuration in build scripts

TychoVrahe:
This PR refactors how UI configuration is done in sconscripts.

A single configuration file per given UI type is introduced. This file defines what fonts, build constants etc should be used in build, depending on concrete target.

Each model carries information what UI type its using.


cepetr (approved):

obrusvit (approved):
utACK

PR #4226, opened 2024-09-27
feat(core): allow two lines of text in menu items

### LINKED ISSUE #4019: Menu item text gets hidden if it is too long
**Describe the bug**
problem:
image

after a local hacky fix:
image

offending code: https://github.com/trezor/trezor-firmware/blob/700512901b822742759e677ae59fbaef04417842/core/embed/rust/src/ui/model_mercury/component/button.rs#L624-L635

**Why** do we even have this? ISTM we should default to showing both icon and text in all cases.

**To reproduce**

`trezorctl btc get-address -n m/44h/0h/0h/0/0 -d`
then click the menu

ibz:
Fixes #4019


obrusvit:
Have you considered utilizing `TextLayout`? Specifically, `TextLayout::layout_text` has some considerable machinery for `LineBreaking` strategies.

Just a suggestion, I actually don't know if it's usable in this case.

ibz:
> Have you considered utilizing `TextLayout`?

Interesting! I looked at it. It seems way too complex, because the main use case seems to be dealing with long text that can take several pages. If you tell it to render a maximum number of lines (what we want is 2 in our case), it would think that you want 2 lines on one page, which means more pages... At least that is how I understand it.

Now of course it probably could be adapted for what we need here, but this is so much simpler that it just seems overkill to use that machinery just for this.

core/embed/rust/src/ui/util.rs
```diff
@@ -131,6 +131,28 @@ pub fn char_to_string(ch: char) -> ShortString {
     s
 }
 
+/// Returns the first two lines of text to be fit on a given length.
+/// When the text is too long to fit in two lines, the rest is just kept as part of the second line.
+/// This assumes no lines are longer than 50 chars (ShortString limit)
+pub fn split_two_lines(
```
matejcik, line 137: this could easily return `(&str, &str)`  (and then at call site you'd stuff the `shape::Text` inside the tstring.map() closure)

ibz:
Some examples of how it looks now:

image

image

(in this case I should make it crash and change the string itself!)
image

image

core/embed/rust/src/ui/util.rs
ibz: > this could easily return `(&str, &str)` (and then at call site you'd stuff the `shape::Text` inside the tstring.map() closure)

Done: https://github.com/trezor/trezor-firmware/pull/4226/commits/fcfc1a2689d93d95d4e1a7385401c26a148f0844

Hannsek:
cc @lapohoda

core/embed/rust/src/ui/util.rs
```diff
@@ -131,6 +131,18 @@ pub fn char_to_string(ch: char) -> ShortString {
...
 
+/// Returns the first two lines of text to be fit on a given length.
+/// When the text is too long to fit in two lines,
+/// the rest is just kept as part of the second line.
+pub fn split_two_lines(text: &str, text_font: Font, available_width: i16) -> (&str, &str) {
+    let p = text_font.longest_prefix_words(available_width, text);
+    if p == 0 { // we did not find a space to split on...
+        (&text[0..0], &text[0..])
+    } else {
+        (&text[..p + 1], &text[p + 2..])
```
matejcik, line 142: ...and this is magically skipping a single ASCII space? that will fail if the longest prefix is the whole of the text.

(you'll probably want `str::trim` here)

core/embed/rust/src/ui/model_mercury/component/button.rs
```diff
@@ -527,44 +529,49 @@ impl IconText {
...
+                let (t1, t2) = split_two_lines(
+                    t,
+                    style.font,
+                    area.width() - Self::ICON_SPACE - Self::TEXT_MARGIN,
+                );
+                let _ = texts.push(t1);
+                let _ = texts.push(t2);
+                for i in 0..texts.len() {
+                    let mut text_pos = rects[i].left_center() + baseline_offset;
+                    text_pos = Point::new(rects[i].top_left().x + Self::ICON_SPACE, text_pos.y);
```
matejcik, line 557: style: what you want is not a `mut` variable but a rebind:
```suggestion
                    let text_pos = rects[i].left_center() + baseline_offset;
                    let text_pos = Point::new(rects[i].top_left().x + Self::ICON_SPACE, text_pos.y);
```

core/embed/rust/src/ui/model_mercury/component/button.rs
```diff
@@ -527,44 +529,49 @@ impl IconText {
...
+                let _ = rects.push(Rect::new(area.top_left(), area.right_center()));
+                let _ = rects.push(Rect::new(area.left_center(), area.bottom_right()));
+                let mut texts: Vec<&str, 2> = Vec::new();
+                let (t1, t2) = split_two_lines(
+                    t,
+                    style.font,
+                    area.width() - Self::ICON_SPACE - Self::TEXT_MARGIN,
+                );
+                let _ = texts.push(t1);
+                let _ = texts.push(t2);
```
matejcik, line 554: you are uselessly making two vecs when (a) a fixed size arrays would be much more appropriate, but (b) even better, just make a function `fn show_text(text: &str, rect: Rect)` here and call it twice with explicit arguments

also this code is not handling the situation where either t1 or t2 is empty

in fact you should probably try "split two lines" instead of the width calculation on L530:

* t1.is_empty() || t2.is_empty(), i.e., either the text fits, or it can't be split? then render `t` on single line
* else render two lines

matejcik (changes requested):

core/embed/rust/src/ui/util.rs
ibz: Indeed, https://github.com/trezor/trezor-firmware/pull/4226/commits/92302c1513c3aaf5c4f3634de07b9d3e65da183b#diff-49bfb5a6d75553b6b483d6d27558216874b122622f76dc4ebdcf835510e1bf40R145

core/embed/rust/src/ui/model_mercury/component/button.rs
ibz: Indeed, looks better with a function (a closure, actually, since I need extra context). https://github.com/trezor/trezor-firmware/pull/4226/commits/92302c1513c3aaf5c4f3634de07b9d3e65da183b#diff-70f3634936c5c39943c5bcd657c91fd5d0133e243c64994798228e192627d774R530

core/embed/rust/src/ui/model_mercury/component/button.rs
```diff
@@ -525,46 +527,39 @@ impl IconText {
...
+        self.text.map(|t| {
+            let (t1, t2) = split_two_lines(
+                t,
+                style.font,
+                area.width() - Self::ICON_SPACE - Self::TEXT_MARGIN,
+            );
 
-        let mut icon_pos = Point::new(
+            if t1.is_empty() || t2.is_empty() {
+                show_text(if t1.is_empty() { t2 } else { t1 }, area);
```
matejcik, line 548: pretty sure this will work?
```suggestion
                show_text(t, area);
```

matejcik (approved):
code LGTM

ibz:
> code LGTM

Thank you. Pushed one more commit, BTW!

core/embed/rust/src/ui/util.rs
```diff
@@ -135,15 +135,27 @@ pub fn char_to_string(ch: char) -> ShortString {
 /// the first line has to fit in the `available_width`,
 /// while the 2nd line contains the rest of the text.
 pub fn split_two_lines(text: &str, text_font: Font, available_width: i16) -> (&str, &str) {
+    let first_line;
+    let second_line;
+
     let p = text_font.longest_prefix(available_width, text);
     if p.is_empty() {
```
obrusvit, line 142: Btw. can be written like this:
```rust
let (first_line, second_line) = if p.is_empty() {
    ("", text)
} else {
    (p, text[p.len()..].trim())
}
```

ibz: Rust. Always surprising. In a good way.

bosomt:
**QA OK**

Tested on TS5 with dev build da7ddd5

PR #4228, opened 2024-09-30
docs: add Safe 3 and Safe 5 in documentation

Hannsek:

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -1,5 +1,5 @@
 # Trezor Core
 
-Firmware currently running on Model T.
+Firmware currently running on Trezor Safe 5, Safe 3 and Model T.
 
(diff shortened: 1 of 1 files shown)


core/README.md
```diff
@@ -1,5 +1,5 @@
 # Trezor Core
 
-Firmware currently running on Model T.
+Firmware currently running on Trezor Safe 5, Safe 3 and Model T.
```
matejcik, line 3:
```suggestion
Firmware currently running on Model T and the Trezor Safe family.
```
so that we don't have to touch this for a while? :)

Hannsek: better, thanks. Can someone take over? 🙏🏻

matejcik (approved):

Issue #4203, opened 2024-09-23
Syscall validation


TychoVrahe:
Followup for https://github.com/trezor/trezor-firmware/pull/4188

Syscall security is dependent on correct validation of input parameters, this is especially important when pointers are involved.

PR #4229, opened 2024-09-30
Syscall verifiers

cepetr:
This PR implements syscall argument verification, ensuring passed buffers are within memory accessible to unprivileged code. It also copies data when needed to prevent TOC/TOU attacks.

Additional changes:
1) Unified and improved argument parsing in the syscall dispatch routine.
2) Adjusted several APIs (display, system) to support argument verification.
3) Removed the `hash_processor` API.

This resolves issue #4203.


core/embed/trezorhal/stm32f4/syscall_dispatch.c
```diff
@@ -105,318 +121,419 @@ __attribute((no_stack_protector)) void syscall_handler(uint32_t *args,
...
     case SYSCALL_HAPTIC_PLAY_CUSTOM: {
-      args[0] = haptic_play_custom(args[0], args[1]);
+      int8_t amplitude_pct = (int8_t)args[1];
+      uint16_t duration_ms = (uint16_t)args[2];
+      args[0] = haptic_play_custom(amplitude_pct, duration_ms);
     } break;
 #endif
 
 #ifdef USE_OPTIGA
-      /*optiga_sign_result optiga_sign(uint8_t index, const uint8_t *digest,
```
ibz, line 350: Shouldn't we have a `SYSCALL_OPTIGA_SIGN` though?

cepetr: Yes, thanks for catching that. I've added fixed it, see 2cff1e570d4fbbf83400e8da0044dd1418a747d1

core/embed/trezorhal/stm32f4/syscall_verifiers.c
```diff
@@ -0,0 +1,667 @@
...
+
+  display_fill(bb);
+
+  return;
+
+access_violation:
+  apptask_access_violation();
+}
+
+void display_copy_rgb565__verified(const gfx_bitblt_t *bb) {
```
TychoVrahe, line 158: the bitblt structure should be copied and only after that verified, as it can be modified later be the applet. (for example by dma2d operation setup before calling this function, or maybe some callbacks)

(in the fill funciton too)

core/embed/trezorhal/stm32f4/syscall_verifiers.c
```diff
@@ -0,0 +1,667 @@
...
+  if (!probe_read_access(bb, sizeof(*bb))) {
+    goto access_violation;
+  }
+
+  uint8_t *src_ptr = (uint8_t *)bb->src_row;
+  size_t src_len = bb->src_stride * bb->height;
+
+  if (!probe_read_access(src_ptr, src_len)) {
```
TychoVrahe, line 166: I think the checks here are not sufficient: 

- the bb->width parameter is not checked which can lead to reading data beyond allowed area, regardless of src_stride.
- i don't think we can ignore the write access here. I.e. the bitblt sets the dst_y and thus can easily cause an overflow of designated area, and looking into for example st7789 implementation with framebuffer, this is not checked in the driver - nor do i think that this should be delegated only to the driver. I realize this is tricky as sometimes we write into register and sometimes to framebuffer, but we should probably do some clipping here to ensure that these functions can only ask to write into the expected display area. 

(applies to fill function too)

TychoVrahe (changes requested):
Few comments below

core/embed/trezorhal/stm32f4/syscall_verifiers.c
cepetr: You are right, see this fix b2ee71c35db10524c5d78c20d4777c3d3b99b982

core/embed/trezorhal/stm32f4/syscall_verifiers.c
cepetr: Yes, I’m aware of this – I wanted to postpone it a bit (not incorporate it into this PR) since it needs to be addressed in the implementation of the function within the driver itself. Only the driver has enough information to validate the destination parameters (for example, it currently checks the display resolution, it knows if we are using framebuffer or not...).

I wouldn’t handle clipping here (as it’s quite complex and is already managed in the Rust layer). I’d prefer to perform simple checks here to ensure we are not writing/reading beyond the unprivileged area.

Let’s solve it a bit later, and I’ll create a new issue.

core/embed/trezorhal/stm32f4/syscall_dispatch.c
```diff
@@ -70,18 +71,33 @@ __attribute((no_stack_protector)) void syscall_handler(uint32_t *args,
   switch (syscall) {
     case SYSCALL_SYSTEM_EXIT: {
       systask_t *task = systask_active();
-      systask_exit(task, (int)args[0]);
+      int exit_code = (int)args[0];
+      systask_exit(task, exit_code);
     } break;
+
     case SYSCALL_SYSTEM_EXIT_ERROR: {
       systask_t *task = systask_active();
```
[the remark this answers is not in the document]
cepetr: Yes, it might be a bit confusing. I’ve made changes – see 2d40762b9e12bcc976f292096689bca726aae59e. Still, it’s a bit of a special case.

core/embed/trezorhal/stm32f4/syscall_verifiers.c
```diff
@@ -0,0 +1,667 @@
...
+
+void systask_exit_error__verified(systask_t *task, const char *title,
+                                  size_t title_len, const char *message,
+                                  size_t message_len, const char *footer,
+                                  size_t footer_len) {
+  char title_copy[64] = {0};
+  char message_copy[64] = {0};
+  char footer_copy[64] = {0};
+
+  title_len = title ? MIN(title_len, sizeof(title_copy) - 1) : 0;
```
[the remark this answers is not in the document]
cepetr: I considered this as well and ultimately decided on this solution. In general, we shouldn’t clip the string, but I think, in this case, it’s better to display a clipped (but still relevant) message rather than a generic one. Additionally, this behavior is consistent with calling these functions directly (the string gets clipped in that case too). Maybe we could just update the comment in the header for the `system_exit_xxx` functions. How about this?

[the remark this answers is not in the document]
TychoVrahe: I still think that if an applet attempts access violation, we should show that and properly respond to that by killing the applet, potentially blacklisting it or something, rather than forcing the syscall into boundaries. In other words, access violation has higher priority than whatever message the applet wants to show.

You can't get access violation when calling these functions directly, so this comparison seems a bit irrelevant. 

You would still clip the string if the applet calls this with larger than 64, but fully accessible, string.

core/embed/trezorhal/stm32f4/syscall_verifiers.c
TychoVrahe: Added the issue https://github.com/trezor/trezor-firmware/issues/4273

core/embed/trezorhal/stm32f4/syscall_verifiers.c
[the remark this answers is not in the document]
cepetr: I finally got you point, see this fix c338491841d2643167812c79a1aa0977bd8ad02a

TychoVrahe (approved):

PR #4231, opened 2024-09-30
TS5 framerate fix

cepetr:
This PR fixes the problem of decreased display framerate on TS5.

Two issues were addressed:

1. Incorrect detection of exception/handler mode inside `display_refresh()`, which caused unnecessary waiting for the TE signal.
2. Removed unnecessary framebuffer copying in `display_get_frame_buffer` (this is not actually a bug fix, but an optimization).

The bug 1) was introduced in #4188, so no changelog entry is needed.


Hannsek:
Please fill in all the details please 🙏🏻

image

Issue #4197, opened 2024-09-19
Improve consistency of names in random_delays module 


cepetr:
This issue is a reminder that we would like to unify/consolidate the names in `random_delays.c`/`random_delays.h`:

1) Rename the files `random_delays.x` to `rdi.x`
2) Rename the `wait_random()` function to `rdi_wait()`
3) Add comments to the function declarations in `rdi.h`

PR #4233, opened 2024-10-01
refactor(core): improve random_delays api

cepetr:
This small PR slightly improves the API of the `random_delays` module by making the function names more consistent.

It reflects the fact that `random_delays` serves two purposes:

1) Provides a random generator for small delays that can be injected into any part of the code via the `wait_random()` function.
2) Implements Random Delay Interrupts (RDI).

The `wait_random()` function was not renamed to avoid introducing extensive changes, especially in legacy code, as this PR aims to be minimally intrusive.

This PR resolves issue #4197, although it addresses the problem in a different way.


core/embed/trezorhal/random_delays.h
```diff
@@ -22,17 +22,48 @@
...
-void rdi_init(void);
+// Initializes the random number generator for `wait_random()` and the RDI
+//
+// RDI is stopped by default and can be started by calling
+// `random_delays_start_rdi()`.
+void random_delays_init(void);
 
-void rdi_start(void);
-void rdi_stop(void);
+// Starts the RDI, introducing small random delays every milliscond via
```
TychoVrahe, line 51: typo milliscond

TychoVrahe (approved):

core/embed/trezorhal/random_delays.h
cepetr: fixed, see f001c4bac76eb8e206ec5a9f55bb3ddbab8c77be

PR #4242, opened 2024-10-02
Fix emulator build on MacOS 

cepetr:
This PR fixes the emulator build on macOS.

1) The `reboot()` function was renamed to `reboot_device()` to avoid a collision with the function of the same name in `unitstd.h`.

~~2) The emulator is now using the libc `snprintf` instead of the minimalistic `mini_snprintf` again (I reverted the change introduced in #4188).~~

2) Fixed `mini_printf()` implementation, so it can be compiled on MacOS.

3) Added string parameter `_Static_assert` in `image.c`. Looks like the macOS compiler has problem with single parameter version outside a function body.

4) Final testing revealed several bugs that prevented the RSOD screen from being drawn in the emulator, so I fixed them as well.


matejcik:
> 2\. The emulator is now using the libc `snprintf` instead of the minimalistic `mini_snprintf` again (I reverted the change introduced in [Core firmware split #4188](https://github.com/trezor/trezor-firmware/pull/4188)).

instead of reverting, how about this? https://github.com/trezor/trezor-firmware/commit/93e13f5480b67bb85db2a373418d6179265e48b7

cepetr:
@matejcik Yes, this is better solution.  I incorporated your proposal, see ccf0b2ba721b6133e8db92474393daa21f52f036.

core/embed/lib/terminal.c
```diff
@@ -169,29 +171,16 @@ void term_print(const char *text, int textlen) {
...
-#include "mini_printf.h"
-#endif
-
 // variadic term_print
 void term_printf(const char *fmt, ...) {
   if (!strchr(fmt, '%')) {
     term_print(fmt, strlen(fmt));
-#ifdef TREZOR_EMULATOR
-    printf("%s", fmt);
-#endif
```
matejcik, line 185: i think we want to keep these? so that the output goes to the virtual terminal and also to stdout

cepetr: This `printf` is not necessary here, as it's already in `unix/system.c`, which is a bit better location for this platform-specific code (and works event if panic/rsod handler is not registered or crashes).

PR #4245, opened 2024-10-03
Fix status screen for UI tests

obrusvit:
Show Success screens with timeout in UI diff for `mercury`.

Setting the timeout to 0 if `animations_disabled` is not necessary but speeds up the tests.
image


mmilata (approved):

obrusvit:
Not sure why I never merged this but it seems to be superseded by https://github.com/trezor/trezor-firmware/pull/5115 
cc: @romanz , thanks

PR #4247, opened 2024-10-04
SLIP-39 extend_mnemonics()

andrewkozlik:
Implements a function which extends a set of SLIP-39 mnemonics by extra shares, while maintaining the threshold.

This, for example, allows extending a 2-of-2 backup to 2-of-3, where the first two shares remain the same. It also allows reconstructing lost shares by providing any threshold number of shares and requesting the original share count.

The implementation is limited to Slip39_Basic, i.e. single group.

Slack thread: https://satoshilabs.slack.com/archives/C016ACNF1BN/p1727975559358469

### DIFF
--- a/core/src/trezor/crypto/slip39.py
+++ b/core/src/trezor/crypto/slip39.py
@@ -247,6 +247,69 @@ def split_ems(
     return mnemonics
 
 
+def extend_mnemonics(
+    share_count: int,  # The number of shares to create.
+    mnemonics: list[str],  # A threshold set of the old mnemonics.
+) -> list[str]:
+    """
+    Extends a set of mnemonics to the desired share_count, while maintaining the threshold. This,
+    for example, allows extending a 2-of-2 backup to 2-of-3, where the first two shares remain the
+    same. It also allows reconstructing lost shares by providing any threshold number of shares and
+    requesting the original share_count. The current implementation is limited to Slip39_Basic,
+    i.e. single group.
+
+    It is not possible to tell how many shares the user originally created, so if share_count is
+    less than the original number of shares, then this function will return the first share_count
+    shares.
+    """
+
+    if not mnemonics:
+        raise MnemonicError("The list of mnemonics is empty.")
+
+    (
+        identifier,
+        extendable,
+        iteration_exponent,
+        group_threshold,
+        group_count,
+        groups,
+    ) = _decode_mnemonics(mnemonics)
+
+    if group_threshold != 1 or group_count != 1 or len(groups) != 1:
+        raise MnemonicError("Extending advanced backups is not supported.")
+
+    threshold = groups[0][0]
+    shares = groups[0][1]
+    if len(shares) != threshold:
+        raise MnemonicError(
+            f"Wrong number of mnemonics. Expected {threshold} mnemonics, but {len(shares)} were provided."
+        )
+
+    if threshold == 1 and share_count > 1:
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


core/src/trezor/crypto/slip39.py
```diff
@@ -247,6 +247,65 @@ def split_ems(
...
+def extend_mnemonics(
+    share_count: int,  # The number of shares to create.
+    mnemonics: list[str],  # The old set of mnemonics.
+) -> list[str]:
+    """
+    Extends a set of mnemonics to the desired share_count, while maintaining the threshold. This,
+    for example, allows extending a 2-of-2 backup to 2-of-3, where the first two shares remain the
+    same. It also allows reconstructing lost shares by providing any threshold number of shares and
+    requesting the original share_count. The current implementation is limited to Slip39_Basic,
+    i.e. single group.
```
onvej-sl, line 259: I'm not sure if this is a feature or a bug, but this function also allows you to reduce the set of mnemonics when `share_count < len(mnemonics)`.

andrewkozlik: `_check_parameters()` should raise a `ValueError` in this case.

onvej-sl: It doesn't if `threshold <= share_count < len(mnemonics)`.

See this test:
```python
    def test_basic_sharing_reduce(self):
        identifier = slip39.generate_random_identifier()
        for extendable in (False, True):
            mnemonics = slip39.split_ems(1, [(2, 3)], identifier, extendable, 1, self.EMS)
            mnemonics = mnemonics[0]
            extended_mnemonics = slip39.extend_mnemonics(2, mnemonics[1:])
            self.assertEqual(mnemonics[:2], extended_mnemonics[:2])
            self.assertEqual(slip39.recover_ems([extended_mnemonics[1], extended_mnemonics[0]])[3], self.EMS)
```

andrewkozlik: In the test the `mnemonics` parameter of `extend_mnemonics()` has length 2, i.e. `share_count == len(mnemonics) == 2`. So AFAICS, this test does not give an example of `share_count < len(mnemonics)`. When I set `share_count=1` it does raise the exception as described above.

Nevertheless, we don't know how many shares the user originally created, so it is indeed possible for the user to generate fewer shares than they originally did. I can add this information to the comment. We can detect that they are asking for fewer shares in some cases by looking at the original share indices, but not in general. Anyway, if the user does ask for fewer shares than originally, then it doesn't make a lot of sense, but I don't see a severe problem in that.

onvej-sl: > In the test the mnemonics parameter of extend_mnemonics() has length 2, i.e. share_count == len(mnemonics) == 2. So AFAICS, this test does not give an example of share_count < len(mnemonics). When I set share_count=1 it does raise the exception as described above.

I apologize for the confusion. I used `len(mnemonics)` to mean "the total number of the shares originally created by the user."

> Nevertheless, we don't know how many shares the user originally created, so it is indeed possible for the user to generate fewer shares than they originally did. I can add this information to the comment. We can detect that they are asking for fewer shares in some cases by looking at the original share indices, but not in general. Anyway, if the user does ask for fewer shares than originally, then it doesn't make a lot of sense, but I don't see a severe problem in that.

I didn't realized that you can't detect it since the information about the total number of shares is not a part of the shares.

Feel free to resolve this conversation.

onvej-sl (approved):

PR #4250, opened 2024-10-07
add support for T3W1

TychoVrahe:
This PR introduces initial support for T3W1, including emulator build.

The HW support is currently targeted to dev board 1, running on F4 MCU. This will only be supported temporarily. To support this HW LHS200kb display panel is added. No BLE support introduced yet.

Some small preparations/adjustments done in separate commits, namely fixing some issues with 16bit display bus, removing unnecessary button syscalls and few more details.


core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_panel.c
```diff
@@ -31,7 +31,18 @@
...
 #include "panels/tf15411a.h"
 #else
+#ifdef DISPLAY_PANEL_LX154A2482
 #include "panels/lx154a2482.h"
+#define PANEL_INIT_SEQ lx154a2482_init_seq
+#define PANEL_ROTATE lx154a2482_rotate
+#elif defined(DISPLAY_PANEL_LHS200KB_IF21)
+#include "panels/lhs200kb-if21.h"
+#define PANEL_INIT_SEQ lhs200kb_if21_init_seq
+#define PANEL_ROTATE lhs200kb_if21_rotate
```
cepetr, line 41: This is starting to get a bit confusing. I think we could come up with a different approach to handle the various panels, but I don't think we need to change it just now.

cepetr (approved):
utACK

Hannsek:
@matejcik can you merge please? 🙏🏻

PR #4256, opened 2024-10-11
Cardano UI simplification for flows initiated by Suite

obrusvit:
DRAFT


obrusvit:
followed here: https://github.com/trezor/trezor-firmware/pull/4284

PR #4265, opened 2024-10-15
chore(common): separate definitions of protobuf options

M1nd3r:
Rationale: remove dependency between "options" and enum `MessageType`. `.proto` files should not depend on `messages.proto` when only "options" are needed.


common/protob/pb2py
```diff
@@ -714,7 +714,11 @@ def main(
     if verbose:
         logging.basicConfig(level=logging.DEBUG)
 
-    descriptor_proto = protoc(proto)
+    # The proto tuple is sorted to give consistent results in different environments
+    list_proto = list(proto)
+    list_proto.sort()
```
M1nd3r, line 719: This is needed, because other environments (eg. my local system) can have different default sorting of files:

- On my system messages.proto are **before** messages-stellar.proto.
- In CI environment, the messages.proto are **after** messages-stellar.proto.

This leads to a different order of generated classes in messages.py (and the order of enums in other .py file). The change of order has no impact on functionality, but it is captured as "diff" during Prebuild Gen check resulting in check fail. Sorting the file names consistently using python sort prevents the issue.

mmilata: Nit: could be:
```
descriptor_proto = protoc(sorted(proto))
```

common/protob/options.proto
```diff
@@ -0,0 +1,70 @@
...
+    optional bool wire_debug_in = 50004;        // message can be transmitted via debug wire from PC to Trezor
+    optional bool wire_debug_out = 50005;       // message can be transmitted via debug wire from Trezor to PC
+    optional bool wire_tiny = 50006;            // message is handled by Trezor when the USB stack is in tiny mode
+    optional bool wire_bootloader = 50007;      // message is only handled by Trezor Bootloader
+    optional bool wire_no_fsm = 50008;          // message is not handled by Trezor unless the USB stack is in tiny mode
+    optional bool channel_in = 50009;
+    optional bool channel_out = 50010;
+    optional bool pairing_in = 50011;
+    optional bool pairing_out = 50012;
+
```
mmilata, line 44: Please add these in a separate commit so that it's clear this one is just a refactor (bonus points for using `refactor(common):` in the commit message). Also please add docstrings.

common/protob/pb2py
M1nd3r: Fixed here: https://github.com/trezor/trezor-firmware/pull/4265/commits/adaa01552daed7348ac81c6ab430a058e9bddf1d

mmilata (approved):
LGTM, feel free to rebase & merge if CI passes

PR #4267, opened 2024-10-16
Framebuffer access control

### LINKED ISSUE #4207: Framebuffer accessibility in coreapp
Followup for https://github.com/trezor/trezor-firmware/pull/4188

When doing graphical operations, coreapp needs access to framebuffer. Currently we allow access to both framebuffers (on models that use double-buffer architecture). 

We should only allow access to the current framebuffer, so that coreapp cannot override data that is on the display without checks in kernel.

cepetr:
This small PR adds precise access control for the framebuffer (and resolves #4207).

This change is primarily aimed at devices using two framebuffers (T3T1) but also affects devices using a single framebuffer (T2B1, T3B1).

Before this change, the unprivileged app had access to both framebuffers at any time. Now, an unprivileged application has access only to the framebuffer acquired by `display_get_frame_buffer()` until it releases it by calling the `display_refresh()` function.

### DIFF
--- a/core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
+++ b/core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
@@ -177,6 +177,8 @@ bool display_get_frame_buffer(display_fb_info_t *fb) {
 
   fb->ptr = get_fb_ptr(drv->queue.wix);
   fb->stride = DISPLAY_RESX * sizeof(uint16_t);
+  // Enable access to the frame buffer from the unprivileged code
+  mpu_set_unpriv_fb(fb->ptr, PHYSICAL_FRAME_BUFFER_SIZE);
 
   return true;
 }
@@ -214,6 +216,9 @@ void display_refresh(void) {
     // the state to be copied to the display
     return;
   }
+
+  // Disable access to the frame buffer from the unprivileged code
+  mpu_set_unpriv_fb(NULL, 0);
 
 #ifndef BOARDLOADER
   if (is_mode_exception()) {
--- a/core/embed/trezorhal/stm32u5/mpu.c
+++ b/core/embed/trezorhal/stm32u5/mpu.c
@@ -192,6 +192,11 @@ typedef struct {
   bool initialized;
   // Current mode
   mpu_mode_t mode;
+  // Address of the framebuffer visible to unprivileged code
+  // (if set to 0, the framebuffer is not accessible)
+  uint32_t unpriv_fb_addr;
+  // Size of the framebuffer in bytes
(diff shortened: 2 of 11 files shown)


core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
```diff
@@ -213,6 +215,9 @@ void display_refresh(void) {
     return;
   }
 
+  // Disable access to the frame buffer from the unprivileged code
+  mpu_set_unpriv_fb(NULL, 0);
```
TychoVrahe, line 219: maybe we should be calling this in deinit functions too?

core/embed/trezorhal/stm32u5/mpu.c
```diff
@@ -307,6 +323,13 @@ mpu_mode_t mpu_reconfig(mpu_mode_t mode) {
     case MPU_MODE_SAES:
       SET_REGION( 5, PERIPH_BASE_NS,           PERIPH_SIZE,  PERIPHERAL,  YES,    YES ); // Peripherals - SAES, TAMP
       break;
+    case MPU_MODE_APP:
+      if (drv->unpriv_fb_addr != 0) {
+        SET_REGION( 5, drv->unpriv_fb_addr, drv->unpriv_fb_size,   SRAM,  YES,    YES ); // Frame buffer or display interface
```
TychoVrahe, line 328: ad the comment, we don't actually expose anything else than framebuffes here, right?

cepetr: fixed, see 929ef8e8cf0561c7aefeb339395f49264f6ebe0b

core/embed/trezorhal/stm32f4/xdisplay/st-7789/display_fb.c
cepetr: Yes, we should. Fixed, see 929ef8e8cf0561c7aefeb339395f49264f6ebe0b

TychoVrahe (approved):

Hannsek:
What is the diff here? Only the 4 pixels?

cepetr:
> What is the diff here? Only the 4 pixels?

It seems like a recurring issue with the loader. @TychoVrahe, you fixed this a few weeks ago, didn’t you?

TychoVrahe:
I created this [PR](https://github.com/trezor/trezor-firmware/pull/4150) to fix that, it hasn't been merged as another solution, via GFL, is preffered.

See also slack discussion: https://satoshilabs.slack.com/archives/C054XGU77QV/p1724926329014489

Lets ignore this here.

PR #4269, opened 2024-10-17
Update link to headertool.py in boot.md

movefasta:

### DIFF
--- a/docs/core/misc/boot.md
+++ b/docs/core/misc/boot.md
@@ -54,7 +54,7 @@ allowing a firmware update via USB.
 * Signature system is Ed25519 (allows combining signatures by multiple keys
   into one).
 * All multibyte integer values are little endian.
-* There is a tool called [headertool.py](https://github.com/trezor/trezor-firmware/blob/master/core/tools/headertool.py) which checks validity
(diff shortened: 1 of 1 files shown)


matejcik:
while you're touching this, using this link would be a much better choice: https://github.com/trezor/trezor-firmware/blob/main/core/tools/README.md

movefasta:
> while you're touching this, using this link would be a much better choice: https://github.com/trezor/trezor-firmware/blob/main/core/tools/README.md

Thank you. Yes, this is an excellent description of all the available tools that are presented in the repository. But in the documentation on the Boot Stages page, it is `headertool.py` that is specified, which corresponds to the context. I am not sure that other tools are appropriate there that are not directly related to the headers of the firmware parts.

Anyway now it's just a broken link in https://docs.trezor.io/trezor-firmware/core/misc/boot.html

matejcik:
> it is `headertool.py` that is specified, which corresponds to the context

Linking to the source code of `headertool` does not make sense anymore because it's not what you run -- instead, you use the `headertool` command present in the poetry virtual environment. That's what the general README should be telling you, which is why I suggested linking to it.

Another option is just to remove the link altogether and just say that there's `headertool` command.

obrusvit:
@coderabbitai rebase

obrusvit:
Thank you. The link was fixed in fe980ef1884a8da1c7a2d18e57285ee5fae0b23a
Apologies for not processing your PR sooner.

PR #4270, opened 2024-10-18
feat(core/ethereum): new ETH contract flow

ibz:


ibz:
Here is how the new flow looks (TS5):

image
image
image
image
image
image

ibz:
> How about TS3?

image

image

image

image

image

I know it's not 100% like Figma. Still have to do a couple of tweaks. But it's getting close...

core/src/apps/nem/multisig/layout.py
```diff
@@ -61,7 +61,7 @@ async def _require_confirm_address(action: str, address: str) -> None:
     await confirm_address(
         TR.nem__confirm_address,
         address,
-        action,
-        "confirm_multisig",
-        ButtonRequestType.ConfirmOutput,
+        description=action,
```
obrusvit, line 64: not sure why you added these keywords everywhere to `confirm_address`

core/src/apps/ethereum/sign_tx.py
```diff
@@ -130,9 +130,9 @@ async def confirm_tx_data(
         return
 
     # Handle ERC-20, currently only 'transfer' function
-    token, recipient, value = await handle_erc20_transfer(msg, defs, address_bytes)
+    token, recipient, value = await _handle_erc20_transfer(msg, defs, address_bytes)
 
-    if token is None and data_total_len > 0:
+    if data_total_len > 0:
```
obrusvit, line 135: The change here and at line 146 seems to break the basic ERC-20 `transfer` contract. Compare the behavior of your branch with `main` with e.g. this command which initiates a send (`transfer` contract) of chainLINK token:
```shell
trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d a9059cbb000000000000000000000000574bbb36871ba6b78e27f4b4dcfb76ea0091880b0000000000000000000000000000000000000000000000000000000000000123 -g 100 -G 100 -i 1 0x514910771AF9Ca656af840dff83E8264EcF986CA 0
```

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -255,14 +256,18 @@ struct ConfirmBlobParams {
     subtitle: Option<TString<'static>>,
     data: Obj,
     description: Option<TString<'static>>,
+    description_font: &'static TextStyle,
```
obrusvit, line 259: Maybe implementing a new layout specifically for confirming the unknown contract would be justified here, instead of modifying the `ConfirmBlob`.

WDYT, but I guess we can do it on the layouts cleanup later.

obrusvit (changes requested):
The flow looks nice on the example provided:
```shell
trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d a9059cbb000000000000000000000000574bbb36871ba6b78e27f4b4dcfb76ea0091880b0000000000000000000000000000000000000000000000000000000000000123 -g 100 -G 100 -i 1 0xfc6b5d6af8a13258f7cbd0d39e11b35e01a32f93 0
```

However, the simple `transfer` is now modified and I'm not sure it's what we want. See a comment below.

Also, from the point of view of a reviewer, I suggest splitting the PR into more logical commits which would ease the review.
For example, a decision on improving of the `confirm_action` by introducing `ConfirmActionMenu` and `ConfirmActionStrings` should be done by two separate commits. You can always squash later.

core/src/apps/ethereum/sign_tx.py
ibz: Good catch! I reverted that change.

core/src/apps/nem/multisig/layout.py
obrusvit: Understood! Note that using keyword arguments takes a little tiny bit more flash size, that's why they were removed on many places in the past. See here: https://github.com/grdddj/upysize#12-do-not-use-keyword-arguments 

Don't revert, just something to keep in mind.

core/src/apps/ethereum/sign_tx.py
obrusvit: `transfer` seems to be working just fine now.

obrusvit (commented):
In general, I think we should go for more specialized rust/python functions rather than extending the existing functions with more arguments but I don't want to block the PR over this. 

And as discussed in the past, updating `fixtures.json` before the review makes it difficult to review the UI changes locally :(

mmilata (approved):
Did not test but code-wise I do not see any problem :+1: 

Please don't forget to include fixtures for translations.

core/translations/en.json
```diff
@@ -959,6 +963,7 @@
     "words__title_threshold": "Threshold",
     "words__try_again": "Try again.",
     "words__unknown": "Unknown",
+    "words__view_all_data": "View all data in the menu.",
```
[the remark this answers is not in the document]
ibz: Yes, I changed to `instructions__`

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -388,7 +417,9 @@ extern "C" fn new_confirm_properties(n_args: usize, args: *const Obj, kwargs: *m
             paragraphs.into_paragraphs(),
             TR::buttons__confirm.into(),
             None,
+            false,
```
obrusvit, line 420: Rebase on `main` might cause conflict here, should be easy to resolve.

obrusvit (approved):

bosomt:
**QA OK**

tested via:
```
trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d a9059cbb000000000000000000000000574bbb36871ba6b78e27f4b4dcfb76ea0091880b0000000000000000000000000000000000000000000000000000000000000123 -g 100 -G 100 -i 1 0xfc6b5d6af8a13258f7cbd0d39e11b35e01a32f93 0
```

**Info:**
 - Suite version: desktop 24.11.0 (bef8f3c4378237d4cf63580d4ab50756f56e3ec1)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.11.0 Chrome/128.0.6613.162 Electron/32.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.4 regular (revision 3c468fa4ddb0dc28ff7476c6312721babbf71410)
 - Transport: BridgeTransport 3.0.0-bundled.24.11.0

ibz:
> Can you please run the CI in different languages? Does the text fit everywhere?

image

ibz:
`pt`

image

ibz:
`es`

image

ibz:
PS: I know that this is different from Figma, where there is a shorter version of this string in TS3 / TT. I took a shortcut and have just one string in all models. I will add the shorter version and this will be fixed. Just FYI. IMO it is not a show stopper.

Hannsek:
Right, please create an issue for the fix and link it here. 🙏🏻

ibz:
`es` "Contrato de interacción" too long...

image

PR #4271, opened 2024-10-18
feat: support Unchained p2wsh path

Shadouts:
Unchained plans to support p2wsh using the same `m/45` path we use for p2sh. This change adds the `UNCHAINED_` patterns to bitcoin `InputScriptType.SPENDP2SHWITNESS`.

### DIFF
--- a/core/src/apps/bitcoin/keychain.py
+++ b/core/src/apps/bitcoin/keychain.py
@@ -158,6 +158,9 @@ def validate_path_against_script_type(
         if slip44 == SLIP44_BITCOIN:
             append(PATTERN_GREENADDRESS_A)
             append(PATTERN_GREENADDRESS_B)
+        if coin.coin_name in BITCOIN_NAMES and multisig:
+            append(PATTERN_UNCHAINED_HARDENED)
+            append(PATTERN_UNCHAINED_UNHARDENED)
 
     elif coin.taproot and script_type == InputScriptType.SPENDTAPROOT:
         append(PATTERN_BIP86)
--- a/core/.changelog.d/4271.added
+++ b/core/.changelog.d/4271.added
@@ -1,0 +1,1 @@
+Add P2WSH support for Unchained BIP32 paths.


core/src/apps/bitcoin/keychain.py
```diff
@@ -150,6 +150,8 @@ def validate_path_against_script_type(
             append(PATTERN_GREENADDRESS_B)
         if coin.coin_name in BITCOIN_NAMES:
             append(PATTERN_CASA_UNHARDENED)
+            append(PATTERN_UNCHAINED_HARDENED)
+            append(PATTERN_UNCHAINED_UNHARDENED)
```
andrewkozlik, line 154: > Unchained plans to support p2wsh [...]

You are adding the paths under `SPENDP2SHWITNESS`. That's "legacy SegWit", i.e. [P2WPKH](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#p2wpkh-nested-in-bip16-p2sh) or [P2WSH](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#p2wsh-nested-in-bip16-p2sh) nested in BIP 16 P2SH. If you want to use the paths with P2WSH, then you should place them under `SPENDWITNESS`.

core/src/apps/bitcoin/keychain.py
```diff
@@ -150,6 +150,8 @@ def validate_path_against_script_type(
             append(PATTERN_GREENADDRESS_B)
         if coin.coin_name in BITCOIN_NAMES:
```
andrewkozlik, line 151: If these paths are only used for multisig, which they are in the already existing code, then the condition to add them should be:
```py
if coin.coin_name in BITCOIN_NAMES and multisig:
    [...]
```

andrewkozlik:
It is best practice to differentiate the BIP 32 path depending on script type, for example `m/44'/...` for P2PKH and `m/84'/...` for P2WPKH. As far as I can see, in Trezor firmware the only exception until now has been `PATTERN_GREENADDRESS_A` and `PATTERN_GREENADDRESS_B`.

There are two main reasons for this practice, which apply in this case:
- Using different paths with different script types makes account discovery simpler in terms of scanning the blockchain, because the BIP 32 path then fully determines the `scriptPubKey` and address.
- Reusing the same path with a different script type makes it more easy to commit public key reuse, because the wallet app needs to be aware of all possible script types and scan the blockchain accordingly. Older wallet apps may not be aware of the use of newer script types and may therefore inadvertently reuse public keys when generating new receive addresses or new change-output addresses.

Furthermore, I see that you are using the subtree `m / 45' / **`, but you use different paths than the one specified in [BIP 45](https://github.com/bitcoin/bips/blob/master/bip-0045.mediawiki), which as far as I can tell is in violation of [BIP 43](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki). Now is a good opportunity to fix this issue by defining a distinct path for Unchained's P2WPKH scheme.

That being said, I think that none of the above is a strict blocker for merging this PR.

core/src/apps/bitcoin/keychain.py
Shadouts: change in 62caf4c12f6eb822fbe7b24422c95b972a483301

core/src/apps/bitcoin/keychain.py
Shadouts: change in 62caf4c12f6eb822fbe7b24422c95b972a483301

matejcik:
code change ACK. @Shadouts please add [a changelog entry](https://docs.trezor.io/trezor-firmware/misc/changelog.html).

andrewkozlik:
Also squash into a single commit and force-push.

andrewkozlik:
Note that for your own benefit, it would be wise to add a device test to `tests/device_tests/bitcoin/test_nonstandard_paths.py`. This is not a requirement, just best practice.

As you are probably aware, this PR does not modify the behavior of Trezor 1 firmware. Again, it's not a requirement from our side, I am only stating this to make sure that it's clear.

Shadouts:
Changelog added, squashed, and force pushed.

matejcik (approved):

PR #4249, opened 2024-10-07
platform, common, lowlevel modules cleanup

cepetr:
This PR focuses on cleaning up tangled code in the `platform.c/h`, `common.c/h`, `lowlevel.c/h` modules, as well as the Unix emulator.

1. Removed some inappropriate dependencies (such as CPU clock setting and the `main_clean_exit` cyclic dependency) from the emulator code.
2. Most code from the `lowlevel` module has been relocated to the new `option_bytes` module.
3. Introduced a small `pvd` (Power Voltage Detector) module, which was previously split across two different places.
4. Introduced a small `reset_flags` module, which was previously located in the `lowlevel` module. Although it's currently not very useful, it might be improved in the future.
5. Most code from `platform.c` has been moved to a new `startup_init` module, which better reflects its purpose.
6. The `platform.h` header no longer exists, freeing up the name for potential future use.

Possible next steps:
1. `common.h` now only contains just utility functions.  We might consider renaming this module to better reflect its purpose.
2. The routines in `utils.s` seem slightly out of place and could be better organized between `bootutil` and `startup_init`.

PR #4278, opened 2024-10-22
fix(core): correct swapped F4/U5 reset flags

cepetr:
This PR fixes a recently introduced bug in the refactoring of the low-level module that is blocking the production build.

The reset flags for STM32F4 and STM32U5 were mistakenly swapped.


mmilata (approved):
Verified that the flags are now the same as before #4249.

PR #4283, opened 2024-10-24
fix(core): unaligned array in protobuf decoder

mmilata:
Protobuf decoder is failing some newly added Rust assertions by creating unaligned slices: https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#asserting-unsafe-preconditions


matejcik:
won't this fail again if something pushes the whole enum entry block out of alignment?

PR #4284, opened 2024-10-25
Cardano UI simplification for flows initiated by Suite #4256

obrusvit:
This PR aims to simplify Cardano flows initiated by Trezor Suite. Namely:
- send transaction with no additional features - this should behave roughly like sending Bitcoin, i.e. confirming address `i`, recipient `i` and Summary at the end. No "Choose level of details" screen.
- staking transaction with no additional features - this should behave almost exactly the same except without the initial screen asking a user to "Choose level of details"

This is achieved by:
- modifying the `ORDINARY_SIGNER` to categorize the transaction further
- based on the category, modify the layouts shown

TODO:
- [x] I need to verify that the categorization of a transaction to either `SIMPLE_SEND`, `SIMPLE_STAKE_DELEGATE`, `SIMPLE_STAKE_WITHDRAW` or `NOT_SUITE_TX` is generally valid
  - testing with Suite, I hopefully got it right 
- [x] fix discrepancies for `total_amount` calculation for the summary - e.g. change output should not be added to the final amount?
  - https://github.com/trezor/trezor-firmware/pull/4284/commits/0dc6427314bf45df46c9686af06fb5afd5456ae7
- [x] fix Recipient `i` showing
  - https://github.com/trezor/trezor-firmware/pull/4284/commits/9f71794fe8e583f7a2d2c07589be824f8b16f12d
- [x] tests - I don't think new tests are needed (there are many for ADA) but:
  - [x] checking the existing UI diff - especially `cardano/sign_tx.json` with `"simple transaction"` in the name are affected
    - there seems to be like 3 cases where a slightly more complicated transaction triggers `should_show_more = True`, shouldn't be a big deal.
  - [x] enable other UI tests with "Show more details" for `mercury`
- [x] `mercury` summary layout has a wrong button text and title


core/src/apps/cardano/sign_tx/signer.py
```diff
@@ -373,6 +385,10 @@ async def _show_output_init(self, output: CardanoTxOutput) -> None:
             assert output.address is not None  # _validate_output
             address = output.address
 
+        if self.suite_tx_type == SuiteTxType.SIMPLE_SEND:
```
obrusvit: omission: this was removed in subsequent commit

core/src/apps/cardano/sign_tx/signer.py
```diff
@@ -69,6 +73,12 @@
 _MAX_CHUNK_SIZE = const(1024)
 
 
+class SuiteTxType(IntEnum):
+    SIMPLE_SEND = 0
+    SIMPLE_STAKE = 1
+    NOT_SUITE_TX = 2
```
mmilata, line 79: Nit: maybe I missed it but I think there should be an explanation somewhere of what this enum is and how is it related to Suite.

common/tests/fixtures/cardano/sign_tx.show_details.json
```diff
@@ -4,54 +4,6 @@
...
-        "witnesses": [
-          {
-            "type": 0,
-            "pub_key": "89053545a6c254b0d9b1464e48d2b5fcf91d4e25c128afb1fcfc61d0843338ea",
-            "signature": "da07ac5246e3f20ebd1276476a4ae34a019dd4b264ffc22eea3c28cb0f1a6bb1c7764adeecf56bcb0bc6196fd1dbe080f3a7ef5b49f56980fe5b2881a4fdfa00",
-            "chain_code": "26308151516f3b0e02bb1638142747863c520273ce9bd3e5cd91e1d46fe2a635"
-          }
-        ]
-      }
-    },
```
mmilata, line 54: I understand you're removing the tests because they produce the same series of screens, but aren't they still valuable for testing the transaction serialization and signing code?

Edit: aren't, because the same transactions exist in `cardano/sign_tx.json`.

mmilata (commented):
Looking like a massive UI improvement!

There's a Safe 3 testcase where hold-to-confirm is introduced as a second-to-last action which probably isn't right:
[Screenshot 2024-10-28 at 22-19-55 T3B1_en-device_tests-cardano-test_sign_tx py test_cardano_sign_tx_failed sample_stake_pool_registration_certifi-3f8170f6](https://data.trezor.io/dev/firmware/ui_report/11553418355/T3B1-en-core_device_test/diff/T3B1_en-device_tests-cardano-test_sign_tx.py%3A%3Atest_cardano_sign_tx_failed%5Bsample_stake_pool_registration_certifi-3f8170f6.html)

Then again, T3T1 also has some weirdness which is not new?:
[Screenshot 2024-10-28 at 22-21-52 T3T1_en-device_tests-cardano-test_sign_tx py test_cardano_sign_tx_show_details plutus_transaction_with_reference_input](https://data.trezor.io/dev/firmware/ui_report/11553418355/T3B1-en-core_device_test/diff/T3B1_en-device_tests-cardano-test_sign_tx.py%3A%3Atest_cardano_sign_tx_failed%5Bsample_stake_pool_registration_certifi-3f8170f6.html)

core/embed/rust/src/ui/model_mercury/flow/confirm_with_info.rs
```diff
@@ -0,0 +1,81 @@
...
+            theme,
+        },
+    },
+};
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum ConfirmWithInfo {
+    Main,
+    Menu,
+}
```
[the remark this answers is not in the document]
obrusvit: This flow has different option in the menu "Show all" and it returns `INFO` on click of that button. It seemed the best way for now is to just create a new flow.
I'm thinking about a more universal constructor of these `SwipeFlow`s which would be supplied with menu options and what they do... No clear idea at the moment.


To answer the top comment:
> There's a Safe 3 testcase where hold-to-confirm is introduced as a second-to-last action which probably isn't right:
> Then again, T3T1 also has some weirdness which is not new?

These screens show the derivation paths and they are shown **after** the transaction is signed and only if `should_show_more is true`, i.e. when the user opted for "Show all". IIUC, Cardano TX is not initiated with derivation paths of the inputs, they are optionally provided by the host afterwards. The UX is quite strange ("Hold to confirm", then simple "Confirm") but I didn't want to change the flows not coming from Suite.
I was also perplexed at first.

obrusvit:
Tested with Suite and found out that the categorization must allow `witness_request_count` higher than `1` as this is equal to the number of inputs which can be higher even for simple transaction. The logic is similar to bitcoin -> more inputs might be needed in order to to make the TX.
 
fixed here: https://github.com/trezor/trezor-firmware/pull/4284/commits/8b5e08753748cc4ad2feae025a349b5455b133e2

core/embed/rust/src/ui/model_mercury/flow/confirm_with_info.rs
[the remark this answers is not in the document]
mmilata: > I'm thinking about a more universal constructor of these SwipeFlows which would be supplied with menu options and what they do

I think main problem is making the parameters of this constructor available to the logic in `handle_swipe`/`handle_event`, which is currently not possible without abusing RwLock/AtomicU.*

mmilata (approved):
* clippy is failing
* needs regenerated fixtures incl. translations

otherwise LGTM

obrusvit:
- ignoring changelog for python job failing as I only corrected a param
- lots of UI diff generated for T3B1 also for non-Cardano stuff because of adding a previously missing option to cancel `confirm_properties` 
image

bosomt:
**QA OK**

checked on TS5 device

**Info:**
 - Suite version: web 24.11.0 (ff5fe34e9ecaee40808f50c235c08cf8e588b0c8)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 5120x1440
 - Device: Trezor T3T1 2.8.4 regular (revision 31d228a325b5968a2f3565b803818aaf7fa29bbb)
 - Transport: WebUsbTransport

PR #4285, opened 2024-10-25
Forbid multisig to singlesig change outputs

onvej-sl:
In the main branch, the multisig fingerprint is checked only for multisig outputs. This means that a transfer from a multisig account to a singlesig account is treated as a change-output as long as all other change-output conditions are satisfied.

In this pull request, an output is treated as a change output only if
  * the output is a multisig and all the inputs are multisig or
  * the output is a singlesig and all the inputs are singlesig.

This allows users to have greater control over where their funds are sent because since https://github.com/trezor/trezor-firmware/pull/2925, we also display the path for internal outputs.

### DIFF
--- a/core/.changelog.d/4285.feat
+++ b/core/.changelog.d/4285.feat
@@ -1,0 +1,1 @@
+Forbidden multisig to singlesig change outputs.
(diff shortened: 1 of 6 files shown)


onvej-sl:
Related issue: https://github.com/trezor/trezor-firmware/issues/4319

onvej-sl:
Replaced by https://github.com/trezor/trezor-firmware/pull/4351.

PR #4287, opened 2024-10-29
make USE_xxx defines global

### LINKED ISSUE #3816: Drop obsolete Trezor boards
We have multiple old boards lying around the source code, and they impose some maintenance burden of keeping the builds up-to-date. The definitions should be dropped -- anyone who wants to run the old board will need to retrieve it from git history.

TychoVrahe:
This PR unifies how feature defs "USE_xxx" are used across board definitons.

The defs are moved to sconscript and are made global, so that platform independent source files do not have to be aware of specific board definitions.

Noteworthy changes:
- the USE_xxx is included even if the build target does not request given feature. This wasn't consistently the case, see i.e. optiga.
- the "optiga_hal" feature was removed in favor of singe "optiga" feature. In order for bootloader emulator to work without adding all the necessary cryptography, `'-Wl,--gc-sections'` was added in its scons.
- the `USE_DISP_I8080_8BIT_DW` was changed to `DISPLAY_I8080_8BIT_DW` as its not a feature but relates to specific pinout of the given board. 

Additionally, obsolete board definitions (older T2B1 and T3T1) are dropped: they are not well maintained anyway. This resolves  #3816 .


core/embed/models/D002/boards/stm32u5a9j-dk.h
```diff
@@ -4,11 +4,6 @@
 #define HSE_16MHZ
 #define VDD_1V8 1
 
-#define USE_I2C 1
-#define USE_RGB_COLORS 1
-#define USE_TOUCH 1
-// #define USE_SBU 1
-// #define USE_DISP_I8080_8BIT_DW 1
 #define USE_HASH_PROCESSOR 1
```
cepetr, line 7: It looks like `USE_HASH_PROCESSOR` is an unintended leftover here, isn’t it?

core/site_scons/models/T3T1/emulator.py
```diff
@@ -43,6 +44,7 @@ def configure(
 
     if "sd_card" in features_wanted:
         features_available.append("sd_card")
+        defines += ["USE_SD_CARD=1"]
```
cepetr, line 47: Shouldn't USE_SD_CARD=1 be one level higher?

cepetr (changes requested):

core/site_scons/models/T3T1/emulator.py
TychoVrahe: yes https://github.com/trezor/trezor-firmware/commit/ca7850be3a05da9ef650dc38117491ad9f5c0bd3

cepetr (approved):

PR #4298, opened 2024-10-31
model/board definitions cleanup

cepetr:
This PR focuses on the following code refactoring:

- Simplifies the use of `ensure_compatible_settings()` and removes it from the board definition.
- Relocates display resolution (`DISPLAY_RESX` & `DISPLAY_RESY`) to the model definition to avoid duplication.
- Removes duplicated (`TREZOR_FONT_BPP`) or unused (`MAC_DISPLAY_RESx`) constants from the emulator board definition.
- Introduces new `USE_xxx` macros, replacing `#ifdef STM32U5` in platform-independent code, which requires including STM32 HAL library headers.
 
- Fixes macOS build of bootloader_emu (uses `-dead_strip` instead of `--gcc-sections`)


core/embed/prodtest/main.c
```diff
@@ -65,8 +65,11 @@
 
 #include "memzero.h"
 
-#ifdef STM32U5
+#ifdef USE_STORAGE_HWKEY
 #include "secure_aes.h"
+#endif
+
+#ifdef STM32U5
```
TychoVrahe, line 72: residual STM32U5 define, needed for the CPU ID function, but we should get rid of it at some point, lets make an issue

core/embed/trezorhal/flash.h
```diff
@@ -26,11 +26,7 @@
 #include "flash_ll.h"
 #include "secbool.h"
 
-#ifndef TREZOR_EMULATOR
-#include STM32_HAL_H
-#endif
-
-#ifdef STM32U5
+#ifdef USE_FLASH_BURST
```
TychoVrahe, line 29: My thought was to leave this here unconditional, and use USE_FLASH_BURST in flash_area where we use #ifdef FLASH_BURST_SIZE. Or we can leave the condition here, but in any case i think it's preferable to replace the condition in flash_area with USE_FLASH_BURST too.

core/embed/models/T3W1/boards/t3w1-unix.h
```diff
@@ -1,11 +1,8 @@
 #ifndef BOARDS_T3W1_UNIX_H
 #define BOARDS_T3W1_UNIX_H
 
-#define MAX_DISPLAY_RESX 380
-#define MAX_DISPLAY_RESY 520
 #define DISPLAY_RESX 380
```
TychoVrahe, line 4: these should be removed

core/embed/prodtest/main.c
cepetr: I agree and created #4306

core/embed/models/T3W1/boards/t3w1-unix.h
cepetr: fixed, see 986f6b804e537be653745bb9758e6f44c77193bf

TychoVrahe (approved):

PR #4308, opened 2024-11-04
Add support for unrecognized Trezor models (fixes #3993)

### LINKED ISSUE #3993: trezorctl: enable support for "unknown models"
`trezorctl` should allow the user to find Trezors that (a) report themselves as valid vid/pid, (b) respond to protobuf `Initialize` with a valid structured message, but (c) cannot be parsed to a known model.

Right now, using old trezorctl with, say, TS5, will respond with "Failed to find Trezor."
Instead, `trezorctl list` should display the device labeled as "unrecognized Trezor model" or w/e, and allow talking to it on a best effort basis.

this will greatly help with prototypes of new devices, and also for customers who don't religiously keep their trezorctl up to date
(also, this will be important for T3B1 soon(tm))

matejcik:

### DIFF
--- a/python/src/trezorlib/cli/firmware.py
+++ b/python/src/trezorlib/cli/firmware.py
@@ -49,6 +49,7 @@ MODEL_CHOICE = ChoiceType(
         "T2T1": models.T2T1,
         "T2B1": models.T2B1,
         "T3T1": models.T3T1,
+        "T3B1": models.T3B1,
         # aliases
         "1": models.T1B1,
         "one": models.T1B1,
--- a/python/src/trezorlib/cli/firmware.py
+++ b/python/src/trezorlib/cli/firmware.py
@@ -95,9 +95,9 @@ def _print_firmware_model(hw_model: Union[bytes, fw_models.Model]) -> None:
         pass
 
     assert isinstance(hw_model, bytes)
-    if all(0x20 <= b < 0x80 for b in hw_model):  # isascii
+    if hw_model.isascii():
         model_name = hw_model.decode("ascii")
-        click.echo(f"Unknown hardware model: {model_name}")
+        click.echo(f"Unrecognized hardware model: {model_name}")
         return
 
     click.echo(f"Suspicious hardware model code: {hw_model.hex()} ({hw_model!r})")
@@ -404,7 +404,7 @@ def validate_firmware(
     fingerprint: Optional[str] = None,
     model: Optional[TrezorModel] = None,
     bootloader_onev2: Optional[bool] = None,
-    prompt_unsigned: bool = True,
+    verify_only: bool = False,
 ) -> None:
     """Validate the firmware through multiple tests.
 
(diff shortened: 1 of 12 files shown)


python/src/trezorlib/models.py
```diff
@@ -36,12 +39,17 @@ class TrezorModel:
 
 # ==== internal names ====
 
+USBID_TREZOR_ONE = (0x534C, 0x0001)
+USBID_TREZOR_CORE = (0x1209, 0x53C1)
```
Hannsek, line 43: don't we have different USBID for T3T1? 🤔

prusnak: Nope

python/src/trezorlib/models.py
```diff
@@ -104,9 +112,23 @@ class TrezorModel:
...
-    usb_ids=((0x1209, 0x53C1), (0x1209, 0x53C0)),
+    usb_ids=(USBID_TREZOR_CORE, USBID_TREZOR_CORE_BOOTLOADER),
+    default_mapping=mapping.DEFAULT_MAPPING,
+)
+
+# ==== unknown model ====
+
+UNKNOWN_MODEL = TrezorModel(
+    name="Unknown Trezor model",
+    internal_name="????",
```
prusnak, line 123: Just brainstorming here. Maybe this would be better from the UX perspective and more akin to other models?

```suggestion
    name="Unknown",
    internal_name="UNKN",
```

matejcik:
yeah so
when looking into [prusnak's comment](https://github.com/trezor/trezor-firmware/pull/4308#discussion_r1827642998) I realized that we use `client.model.internal_name` instead of `client.features.internal_model` in places. Or, more specifically, we don't pass `Features` into things that query model, instead we pass `TrezorModel` -- e.g., [here](https://github.com/trezor/trezor-firmware/blob/008b5df43267848449e44d22fe8fdf392c835114/python/src/trezorlib/cli/firmware.py#L211), we use `TrezorModel` to fetch the right download data.

This is a problem with this new approach ... I mean, previously, we would just fail before even getting to that place.
But now we can get there, but instead pass `????` (or `UNKN`) as the model internal name, where what we _want_ to do in cases like this is play dumb and use the value of `features.internal_model`. Which would work perfectly fine, whereas passing a fixed `UNKNOWN_MODEL` object breaks it.

(I believe this is the wackiness that tripped me the last time I played around with it)

Not sure what the right solution is here.
We can generate a new `TrezorModel` object on the fly, but that makes it difficult to identify that it's an unknown model.

python/src/trezorlib/firmware/core.py
```diff
@@ -126,7 +126,10 @@ class FirmwareImage(Struct):
 
     @staticmethod
     def calc_padding(hw_model: bytes, len: int) -> int:
-        alignment = Model.from_hw_model(hw_model).code_alignment()
+        try:
+            alignment = Model.from_hw_model(hw_model).code_alignment()
+        except ValueError:
+            alignment = Model.T3W1.code_alignment()
```
romanz, line 133: Would it be clearer to add an explicit `DEFAULT_CODE_ALIGNMENT` constant (instead of re-using T3W1 alignment)?

romanz (approved):
LGTM, with two small questions:

python/src/trezorlib/cli/firmware.py
```diff
@@ -92,7 +92,13 @@ def _print_firmware_model(hw_model: Union[bytes, fw_models.Model]) -> None:
         click.echo(f"{model_name} firmware image.")
         return
     except ValueError:
-        pass
+        assert isinstance(hw_model, bytes)
```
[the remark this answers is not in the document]
matejcik: good catch, i completely missed that :woman_facepalming: 48638e1ac7c9d378929b87f35b25491327a9dd94

python/src/trezorlib/firmware/core.py
matejcik: based on this comment, the whole function was removed :)

matejcik:
rebased on main in order to get rid of the problematic `calc_padding` function

romanz (approved):

PR #4315, opened 2024-11-04
chore(core): bump core, prodtest versions

mmilata:
[no changelog]

### DIFF
--- a/core/translations/signatures.json
+++ b/core/translations/signatures.json
@@ -1,8 +1,8 @@
 {
   "current": {
-    "merkle_root": "e11fba15b5f87f7fe23f626c73bd1b611171ba656dece9f8f510537d3556c595",
-    "datetime": "2024-11-06T12:36:28.794885",
-    "commit": "ee81befb05fa3842d4b74ccfb89082be318bf651"
+    "merkle_root": "3af4b3858cde58d36f884fb0ea592d488c8380baaed2241941bcdb38276d8d25",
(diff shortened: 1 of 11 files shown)


core/translations/signatures.json
```diff
@@ -1,8 +1,8 @@
 {
   "current": {
-    "merkle_root": "70b288d1c7cc176d116e3fb4f0737daba7c465b6ee616278396843933d69afd3",
-    "datetime": "2024-11-04T07:30:55.564229",
-    "commit": "dfdd44becba40dbdff15e8aa88edbf6cedf063ea"
+    "merkle_root": "3af4b3858cde58d36f884fb0ea592d488c8380baaed2241941bcdb38276d8d25",
+    "datetime": "2024-11-06T13:24:51.443025",
+    "commit": "9c32e76315450f636bde59a25da2b2ff11699682"
   },
   "history": [
```
matejcik, line 7: while you're at this, please add the signatures that we generated recently

mmilata: sure - what about the duplicate versions, append or overwrite?

matejcik: i think they're appended, the signatures are for a different commit.
the `translations/cli.py sign` command should do the right thing, i believe

mmilata: pushed ee3589b777498448037d476fa0247770be17fa41, please check

PR #4321, opened 2024-11-05
add power button to button driver

TychoVrahe:
preparation for T3W1


cepetr (approved):
It's look good me. There are some nits, but I fixed them in subsequent PR #4325.

core/embed/models/T3W1/boards/t3w1-unix.h
```diff
@@ -6,6 +6,8 @@
 #define TOUCH_OFFSET_X 80
 #define TOUCH_OFFSET_Y 102
 
+#define BTN_POWER_KEY SDLK_LEFT
```
prusnak, line 9: This is not intuitive because this key has another meanings for other models.

Can we use some other key? We could use `P` (currently used for screenshot/printscreen), if we changed the screenshot button to `S`.

TychoVrahe: This commit https://github.com/trezor/trezor-firmware/pull/4321/commits/04148074af99460582972ee2b9e687dda3b697ac changes the power button to 'p'

This one  https://github.com/trezor/trezor-firmware/pull/4321/commits/2229363c400e840cd215246e65825f9fbad1e7ed changes the screenshot button to 's'

PR #4327, opened 2024-11-06
translations: check if language file version matches firmware

mmilata:
Also:
* do not expand `##Safe3` when running `cli.py gen`
* make `cli.py` executable


matejcik (approved):
utACK

PR #4299, opened 2024-11-01
chore(core/translations): add translations for ETH contract flow

ibz:

PR #4329, opened 2024-11-07
core/translations: check order.json for holes

mmilata:
One more check in addition to #4327 that aims to prevent the situation where the indices in `order.json` are not contiguous, which can easily happen during merge conflict resolution (see fix in: #4299).

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -239,6 +239,7 @@ translations: ## update translations
 	python ./translations/cli.py gen
 
 translations_check: ## check that translations are up to date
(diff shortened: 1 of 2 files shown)


matejcik (approved):
utACK

(but given that 25 % of new lines is checking the argument, maybe quickly convert to `click`?)

mmilata:
Fair enough, force-pushed 4f93f6a7dce65add7cad55316524f589186ae6af

matejcik (approved):

PR #4332, opened 2024-11-08
fix firmware hashing on U5

TychoVrahe:
reconfigures the MPU to have access to all of the fw flash.


cepetr (approved):
I've tested it, and everything seems to be working fine.

PR #4335, opened 2024-11-08
ci: always post UI report comment to PR

mmilata:
Currently the comment with links to UI reports is posted when at least one of the following is satisfied:
1. there are changes in `fixtures.json` compared to `main`,
2. at least one _Device tests_ job fails.

Unfortunately there are cases when the comment would be useful but isn't posted:
* only a _Click tests_ job fails,
* in the current run everything passes, there are no changes in `fixtures.json`, comment isn't posted but in some previous run this was not the case so PR now contains outdated comment.

The first issue can be solved by extending `2.` to click tests and persistence tests, however this creates too many GitHub API requests and gets us 429 rate limited.

So until we figure out something better let's just post the comments always and hope that people who never touch the UI code don't get too annoyed ...

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -286,8 +286,6 @@ jobs:
           status: ${{ job.status }}
         if: ${{ always() && env.ACTIONS_DO_UI_TEST == 'true' }}
         continue-on-error: true
-      - uses: ./.github/actions/ui-comment
-        if: ${{ failure() && env.ACTIONS_DO_UI_TEST == 'true' }}
       - uses: ./.github/actions/upload-coverage
 
   # Click tests - UI.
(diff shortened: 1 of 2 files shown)


obrusvit (approved):
Thanks! Based on your comment I understood that posting the comment for each PR should use less requests than extending the logic. If that's the case, LGTM.

mmilata:
Currently it's up to 2 per workflow + 2 for each device test. Two because first one checks if the comment already exists (and what's its ID if it does) and the second one creates/updates it.

We can probably reduce it to constant number by uploading a file to `data.trezor.io/.../worflow_id/something` to indicate that comment has been posted and does not need to be updated for this workflow, as data.trezor.io but right now i don't really wait to spend time on it.

PR #4336, opened 2024-11-11
feat(core): Add coin definition for eCash XEC (altcoin based on bitcoin) 

Fabcien:
This PR adds the definition for eCash (XEC).

I tried to use support.py to generate the support.json but this currenly fails due to the missing release file for T3W1 so I updated manually.

`make gen` builds successfully. I also ran `make style_check` but this failed on me with 122 `reportArgumentType` errors from `pyright`. I'm probably not using the right version, but all the reports are unrelated to this change, and the other linters return no error. I expect it to pass on CI.


Fabcien:
@matejcik Can you explain why the CI is failing and if there is something I need to do to fix it ? It looks like a permission issue at first sight but not sure

matejcik:
"add to github project" -> don't worry about this one

"gen check" -> you did not run `make gen` and commit the resulting changes
"defs check" -> for some reason our checker requires `coin_name` and `coin_label` to start with an uppercase letter. i believe the right thing to do is add an exception for `e` to the regexps in `common/tools/coin_info.py`

Fabcien:
I pushed 2 new commits:
 - One to add an exception for `defs_checks` as suggested by @matejcik
 - One to include the output of `make gen`
I ran `make style_check` (thanks for the hint), `make gen_check` and `make defs_check` and they all passed.

Should I squash the commits ?

common/defs/bitcoin/ecash.json
```diff
@@ -0,0 +1,44 @@
+{
+  "coin_name": "eCash",
```
prusnak, line 2: I prefer to keep the name like this.

That way we do not need to change 2 regex-es, only one.

```suggestion
  "coin_name": "Ecash",
```

Fabcien:
@matejcik @prusnak Is there anything else I need to do ?

Fabcien:
@matejcik Thanks I updated accordingly.

Fabcien:
@matejcik @prusnak Is there anything left to do ?

Fabcien:
@matejcik Can you please re-review ? I think I addressed all the issues

matejcik (approved):

Fabcien:
Thanks @matejcik !

PR #4339, opened 2024-11-11
CI: cancel previous workflow run after pushing new code to PR

mmilata:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency


.github/workflows/cflite.yml
```diff
@@ -4,6 +4,10 @@ on:
   pull_request:
     paths:
       - '**'
+# cancel any previous runs on the same PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
vdovhanych, line 10: can you add new line after, just a nit pick.

vdovhanych (approved):
Should be good to go :)

.github/workflows/cflite.yml
mmilata: reformatted the file to have new lines between top-level keys like the other files @ 347f16f

vdovhanych (approved):

PR #4343, opened 2024-11-12
Cleanup after ETH contract flow

ibz:
This PR fixes some of the mess caused by #4270

* we now use `warning_hi_prio` instead of having that functionality duplicated into `confirm_blob`
* `confirm_blob_with_optional_pagination` is gone, and we are back to having `ask_pagination` as a parameter on `confirm_blob`

NB: This PR prevents the `Marquee` on model R from crashing, but it also makes it not scroll anymore.


Hannsek:
In TS3 we don't use these crosses.

ibz:
> In TS3 we don't use these crosses.

> Is the info about the sítě somewhere on TT?

What do you mean?

Hannsek:
edit: Is the info about the size of the data somewhere on TT?
image

Regarding the crosses, we use different cancel sign.
image
e.g. 
image

ibz:
> edit: Is the info about the size of the data somewhere on TT? 

Yes, it disappeared for a while, now it should be back.
 
> Regarding the crosses, we use different cancel sign.

Good catch! I'll look into that.

core/src/trezor/ui/layouts/mercury/__init__.py
```diff
@@ -139,6 +139,20 @@ def confirm_multisig_warning() -> Awaitable[None]:
     )
 
 
+if not utils.BITCOIN_ONLY:
```
obrusvit, line 142: Code organization wise, is there a reason you don't put `confirm_unknown_contract_warning` and `confirm_other_data` under already existing `if not BITCOIN_ONLY` branch. Now it's 3x in this file. Similar situation in `tt/__init__.py` and `tr/__init__.py`

obrusvit (approved):
LGTM.
- tested `stake`, `approve`, `transfer`, `transfer` with unknown contract and unknown contract with `model_r` and `mercury`
    - `model_r` the screen with "Warning, Unknown contract. Continue only if you know what you are doing" has only "Confirm" option and a user can cancel only on the next screen but that's probably not a part of this cleanup PR.
- notes from Hans seems to be fixed
- have one code-wise suggestion
- maybe @matejcik would like to take a look as well
- not sure if we merge with the marquee broken but I guess yes, I'd be happy to start rebasing my PRs on top of this

core/src/trezor/ui/layouts/mercury/__init__.py
ibz: I also debated with myself where it is better - under the same `if` or under a separate one, but close to the functions that are similar. I think it can be argued either way. I can move if you prefer the other way.

ibz:
>       * `model_r` the screen with "Warning, Unknown contract. Continue only if you know what you are doing" has only "Confirm" option and a user can cancel only on the next screen but that's probably not a part of this cleanup PR.

Yes, the goal of this was to fix the warning and the `confirm_blob`, not to bring the UI to the level of Figma. That part is coming in another PR I am preparing.

core/src/trezor/ui/layouts/mercury/__init__.py
obrusvit: Should be fine either way.

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -162,6 +162,16 @@ def confirm_multisig_warning() -> Awaitable[None]:
     )
 
 
+if not utils.BITCOIN_ONLY:
```
matejcik, line 165: can we move both these things to the big dedicated "if not BITCOIN_ONLY" section far below, together with the other altcoin stuff?

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -605,11 +588,30 @@ def confirm_blob(
...
+        return raise_if_not_confirmed(
+            layout,
+            br_name,
+            br_code,
+        )
+
+
+if not utils.BITCOIN_ONLY:
+
+    def confirm_other_data(data: bytes, data_total: int) -> Awaitable[None]:
```
matejcik, line 604: this is strongly tied to Ethereum and as such i'd say it should go into `apps/ethereum/layout.py` ?

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -162,6 +162,16 @@ def confirm_multisig_warning() -> Awaitable[None]:
     )
 
 
+if not utils.BITCOIN_ONLY:
+
+    def confirm_unknown_contract_warning() -> Awaitable[None]:
```
matejcik, line 167: ...this is _also_ strongly tied to Ethereum and should _also_ go to `ethereum/layout.py`, but the impl is different for mercury. could we maybe extract a building block from this? or use `show_warning` in all cases?

ibz: Well, I could actually implement `warning_hi_prio` for all models and make it use the normal warning for non-Mercury. But the other issue with this is that the copy is also different for different models (changed it in the next PR). Non-Mercury have a shorter version of the text. How could I make that work without different implementations?

core/src/trezor/ui/layouts/tt/__init__.py
matejcik: please try using "x" for confirm_blob _always_ on TT/TR and see if any use of `verb_cancel` actually pops up.

ibz: I see what you mean. Basically ignoring the `verb_cancel` parameter, if possible, on these models.

The issue I see is that sometimes on TR verb_cancel is `"<"` or `"^"`. Arguably, this misused, since these are more like "verb_back" than "verb_cancel", but then we would end up complicating it even more - how do we know whether a layout has a back or a cancel?

Maybe I can make it such that, on model TR / TT `verb_cancel` can be 1 character long - otherwise it would become `"x"`?

core/src/trezor/ui/layouts/tt/__init__.py
matejcik: yeah, that's the big copy problem. okay, leave it as-is, just please put "ethereum" in the name so that it's clearer.

core/src/trezor/ui/layouts/tt/__init__.py
matejcik: there's just one case that calls `ui.layouts.confirm_blob` with verb_cancel, and that's in impl `confirm_signverify`
where I think you could replace it with `trezorui2.confirm_blob` without loss of functionality

are there others that I missed?

matejcik: in a signverify ? just drop it

ibz: https://github.com/trezor/trezor-firmware/pull/4343/commits/7abacb3ff7d16d9a9a06e6eca0dc957171e8936c

matejcik (approved):

PR #4345, opened 2024-11-13
fix MPU configuration on F4

TychoVrahe:
This PR fixes MPU configuration on F4 - the kernel SRAM section was placed at the end of SRAM in linker script, while MPU was adjusting beggining of the SRAM.

Other unrelated fix is smuggled in here: conditional compilation of TOUCH interface exposure to python was mangled with USB events interface.


cepetr (approved):
good catch, utACK

Issue #4302, opened 2024-11-01
Improve EVM contract flow


ibz:
While implementing #4251 some shortcuts were taken. The flow does generally behave as it should, but it doesn't look exactly like Figma. This has to be fixed.

Some technical debt also needs to be taken care of:

```[tasklist]
- [x] stop using `confirm_blob` for warnings (and remove `default_cancel` on `confirm_blob`)
- [x] get rid of `confirm_blob_with_optional_pagination`
- [x] get rid of `description_font_green`
```

Tweaks to the flow:

```[tasklist]
### Tasks
- [x] TS5: show page number when paginating through input data
- [x] TS5: "X" when paginating through input data
- [x] TS5: "input data" - 1st page content is shorter
- [x] TS5: "input data" confirm footer
- [x] TS5: show data size on paginated data
- [x] TS3 & T: Unknown contract warning - different copy
- [x] TS3: paginated data - "continue" button instead of "go back"
- [x] TS3: "token contract" continue button
- [x] TS3: data size mismatch?
```

PR #4346, opened 2024-11-13
UI tweaks for ETH contract flow

ibz:
This is a continuation of #4343, addressing the UI side of #4302.


matejcik (commented):
while we're at this, can you remove `text_mono` and `info` (and possibly even `hold`) from `confirm_blob` ?

core/embed/rust/src/ui/component/text/formatted.rs
```diff
@@ -37,6 +37,16 @@ impl FormattedText {
             .layout_ops(self.char_offset, Offset::y(self.y_offset), sink)
     }
 
+    pub(crate) fn layout_content_with_offset(
```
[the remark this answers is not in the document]
ibz: Yep! 8313c30d5abe47d77ce350967138a157ac17c66d

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -21,120 +21,154 @@ use crate::{
...
 };
 
+const MENU_ITEM_CANCEL: usize = 0;
+const MENU_ITEM_INFO: usize = 1;
+
+// Extra button at the top-right corner of the Action screen
+#[derive(PartialEq)]
+pub enum ConfirmActionExtra {
+    // Opens a menu which can (optionally) lead to an extra Info screen, or cancel the action
+    Menu {
```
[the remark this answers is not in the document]
ibz: OK... 33d1def00d087e3434ddcc03014190f3f0c6a576

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -240,53 +306,69 @@ fn create_flow(
...
+        verb_info,
+    } = extra
+    {
+        // NB: The Cancel menu item is always the first,
+        // because of the MENU_ITEM_CANCEL = 0.
+        // If we want the cancel item to be somewhere else,
+        // we would need to account for that and we could not use a constant.
+        let mut menu_choices = VerticalMenu::empty().danger(
+            theme::ICON_CANCEL,
+            verb_cancel.unwrap_or(TR::buttons__cancel.into()),
```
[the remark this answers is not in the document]
ibz: Not sure what is the benefit. It would just move the check / unwrap at a higher level, most probably in `ConfirmBlobParams` - because that one *can* have it missing. Let's say I also set it to `TR::buttons__cancel ` in `ConfirmBlobParams`, I would still need to make the check in `new_confirm_action`. Unless I make it non-optional there too and set a default in uPy... Either way, the default would have to be somewhere, no?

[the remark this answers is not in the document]
ibz: OK: fc0bb630620df3c133a12e9bd437fdc6199d62d6

I think I understand your point, and it can be seen like this too, indeed.

I was seeing it from the other angle - as much as possible instantiate these "string bearing structs" like `ConfirmBlobParams` and `ConfirmActionMenuStrings` (maybe I should rename it to `ConfirmActionMenuParams` while at it...) with no parameters, have everything default to None, and do `with_XXX` only when "XXX" needs to be changed to something other than the default. Otherwise the struct would carry the None all the way to the very end, where the default would be used. I think @obrusvit 's recent changes also went in this direction - dropping as many parameters from `ConfirmBlobParams::new` and moving them to `.with_XXX`.

Of course, the two ways are not mutually exclusive, which is what I tried to do here - not sure whether you guys like it - but basically `ConfirmBlobParams::new` still means you want to go with defaults, but the default `verb_cancel` is set to `TR::buttons__cancel.into()` in the ctor rather than to `None` - so this "params bearing struct" carries the concrete default around. Maybe if this is what you meant, then I should just replace the rest of the strings in `ConfirmBlobParams` in the same way... I think I like it...

The disadvantage is of course that multiple instantiations of `ConfirmBlobParams` around multiple API functions have to `unwrap_or` *if* they want to do a `.with_verb_cancel` (in order to let the uPy code customize the verb. OOOOR, I could make `with_verb_cancel` accept an Option and have *it* set the default in case it received `None` - but I don't know why, that looks less appealing to me (even though the opposite leads to duplication of the default value).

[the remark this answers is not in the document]
matejcik: > Maybe if this is what you meant, then I should just replace the rest of the strings in `ConfirmBlobParams` in the same way... I think I like it...

yes, this is what I have in mind

> multiple instantiations of ConfirmBlobParams around multiple API functions have to unwrap_or if they want to do a .with_verb_cancel (in order to let the uPy code customize the verb. OOOOR, I could make with_verb_cancel accept an Option and have it set the default in case it received None - but I don't know why, that looks less appealing to me (even though the opposite leads to duplication of the default value).

wondering about this .. would it make sense to construct `ConfirmBlobParams` and friends by passing the micropython options -- gated by `#[feature(micropython)]` of course?
wdyt @obrusvit

ibz:
> while we're at this, can you remove `text_mono` and `info` (and possibly even `hold`) from `confirm_blob` ?

I'll do that in the *next* PR if you don't mind, together with cleanup of `warning_hi_prio`. There's already too many unrelated changes in this PR, and rebasing becomes a pain.

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -240,53 +306,69 @@ fn create_flow(
...
         );
-    }
 
-    let content_menu = Frame::left_aligned("".into(), menu_choices)
-        .with_cancel_button()
-        .with_swipe(Direction::Right, SwipeSettings::immediate());
+        if has_info {
```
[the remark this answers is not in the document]
ibz: OK, something like this I guess: 1eff1d8e0e7660386e2475c7cdf99b63b35a9523

Now `new_confirm_blob` looks strange, but once I remove the `info` parameter completely, it will look great - basically the decision on whether to show info or not is done by passing `Some` or `None` in `verb_info`.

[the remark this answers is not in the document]
ibz: Turned out slightly different actually, but anyway, we'll discuss this in the next PR, which does not cleanup on `confirm_blob`.

core/embed/rust/src/ui/model_mercury/flow/util.rs
```diff
@@ -262,9 +269,19 @@ impl ConfirmBlobParams {
         }
         .into_paragraphs();
 
+        let confirm_extra = if self.cancel {
```
obrusvit, line 277: so `cancel` param means it's *just* "Cancel" in the menu but if `cancel == false` then the menu contains <Info verb> *and* "Cancel"?

core/embed/rust/src/ui/model_mercury/component/frame.rs
```diff
@@ -347,9 +355,16 @@ fn frame_event(
...
+    header: &mut Header,
+    footer: &mut Option<Footer>,
+    bounds: Rect,
+    margin: usize,
+) -> Rect {
     let (mut header_area, mut content_area) = bounds.split_top(TITLE_HEIGHT);
-    content_area = content_area.inset(Insets::top(theme::SPACING));
+    content_area = content_area
+        .inset(Insets::top(theme::SPACING))
+        .inset(Insets::top(margin as i16));
```
obrusvit, line 367: This makes the *real* margin now `SPACING + margin`. It's quite confusing but probably fine.

core/embed/rust/src/ui/model_mercury/layout.rs
```diff
@@ -329,6 +317,44 @@ extern "C" fn new_confirm_blob(n_args: usize, args: *const Obj, kwargs: *mut Map
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_confirm_blob_intro(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
```
obrusvit, line 323: I don't love it as it creates `mercury` specific layout. But probably fine for now.

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -69,21 +69,21 @@ impl ConfirmActionStrings {
...
+    verb_cancel: TString<'static>,
     has_info: bool,
     verb_info: Option<TString<'static>>,
 }
 
 impl ConfirmActionMenuStrings {
     pub fn new() -> Self {
         Self {
-            verb_cancel: None,
+            verb_cancel: TR::buttons__cancel.into(),
```
obrusvit, line 79: maybe we can introduce:
```rust 
const VERB_CANCE_DEFAULT: TranslatedString = TR::buttions__cancel`;
```
and use it on the `unwrap_or` calls?

obrusvit (commented):
I have some points to discuss.

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -191,6 +194,22 @@ fn new_confirm_action_uni<T: Component + MaybeTrace + 'static>(
         .with_menu_button()
         .with_footer(TR::instructions__swipe_up.into(), None);
 
+    if show_page_counter {
+        fn footer_update_fn<T: Component + Paginate>(
```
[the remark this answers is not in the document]
ibz: Strongly disagree on this. If you had an `int` local variable used only inside an `if`, you would declare it inside that `if` even if it is allocated on the stack at the time the function is called, so you are not saving anything, right? The benefit is not in object creation, but in scoping and localizing code, which leads to readability. Having the function declared outside the `if` indicates that it is used in other places.

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -25,115 +25,141 @@ use super::super::{
...
+}
+
+/// The simplest form of the ConfirmAction flow:
+/// no menu, nor a separate "Tap to confirm" or "Hold to confirm".
 #[derive(Copy, Clone, PartialEq, Eq)]
-pub enum ConfirmAction {
-    Intro,
-    Menu,
-    Confirm,
+enum ConfirmAction {
```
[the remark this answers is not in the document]
ibz: If there is no menu, there *must* be a `Cancel`, according to `ConfirmActionExtra`.

This is in accordance with the design, BTW:

image

I think it generally makes sense to have just an "x" to cancel, without a menu,  if the only thing you could do from the menu was to cancel.

core/embed/rust/src/ui/model_mercury/flow/util.rs
ibz: "Cancel" is always there - whether in the menu or as an "x" (if the menu is unnecessary because it just has one option). Perhaps this should be changed, such that it is the case everywhere ... ? Basically there should be no menu with just "cancel" in it? I guess it makes no sense. I did not want to touch other flows here, but I think your concern makes sense, that this is confusing.

core/embed/rust/src/ui/model_mercury/component/frame.rs
ibz: Right. What could I do about it? Rename my parameter to `extra_margin`? Would that make more sense?

core/embed/rust/src/ui/model_tr/component/share_words.rs
```diff
@@ -150,7 +150,7 @@ impl<'a> Component for ShareWords<'a> {
 }
 
 impl<'a> Paginate for ShareWords<'a> {
-    fn page_count(&mut self) -> usize {
+    fn page_count(&self) -> usize {
         // Not defining the logic here, as we do not want it to be `&mut`.
```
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4346/commits/c01d4acaf2955df822e2a7a191378b49e86a67f1

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -318,11 +337,13 @@ pub fn new_confirm_action_simple<T: Component + Paginate + MaybeTrace + 'static>
     strings: ConfirmActionStrings,
     hold: bool,
     page_limit: Option<usize>,
+    show_page_counter: bool,
```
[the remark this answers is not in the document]
ibz: Yep! https://github.com/trezor/trezor-firmware/pull/4346/commits/1f559ff0b656f8d5f4b3cd16b0030677bcee4175

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
ibz: Perfect! https://github.com/trezor/trezor-firmware/pull/4346/commits/7218e99421ca31fffbc85b5f4921b5e1afad2af8

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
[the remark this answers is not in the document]
obrusvit: Ok. You're right that scope-wise, it makes more sense.

core/embed/rust/src/ui/model_mercury/flow/confirm_action.rs
```diff
@@ -19,120 +20,175 @@ use crate::{
...
-pub enum ConfirmAction {
-    Intro,
-    Menu,
-    Confirm,
+enum ConfirmAction {
+    Action,
 }
 
 impl FlowController for ConfirmAction {
+    #[inline]
```
matejcik, line 104: does this actually do anything?

core/src/trezor/ui/layouts/tr/__init__.py
```diff
@@ -614,9 +614,11 @@ async def _confirm_ask_pagination(
...
             return
 
-        await interact(confirm_more_layout, br_name, br_code, raise_on_cancel=None)
-
-    assert False
+        try:
+            await raise_if_not_confirmed(confirm_more_layout, br_name, br_code)
+            break
+        except ActionCancelled:
+            continue
```
matejcik, line 621: if you don't want the ActionCancelled to propagate, don't `raise_if_not_confirmed` -- instead use `interact` and choose what to do based on the return value

matejcik (changes requested):

core/src/trezor/ui/layouts/tr/__init__.py
ibz: [7c28cbff83d577f841c951816fc893f11277df54](https://github.com/trezor/trezor-firmware/pull/4346/commits/7c28cbff83d577f841c951816fc893f11277df54)

ibz:
> while we're at this, can you remove `text_mono` and `info` (and possibly even `hold`) from `confirm_blob` ?

BTW, I am centralizing these requests, and some more, here: https://github.com/trezor/trezor-firmware/issues/4386

core/embed/rust/src/ui/model_mercury/layout.rs
obrusvit: concern linked in a new issue below

core/embed/rust/src/ui/model_mercury/flow/util.rs
```diff
@@ -105,17 +107,12 @@ impl ConfirmBlobParams {
...
-        self
-    }
-
     pub const fn with_verb(mut self, verb: Option<TString<'static>>) -> Self {
         self.verb = verb;
         self
     }
 
-    pub const fn with_verb_cancel(mut self, verb_cancel: Option<TString<'static>>) -> Self {
+    pub const fn with_verb_cancel(mut self, verb_cancel: TString<'static>) -> Self {
```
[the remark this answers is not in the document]
ibz: [f5e40d976de42ce3b7c3899421a91cfca667e05d](https://github.com/trezor/trezor-firmware/commit/f5e40d976de42ce3b7c3899421a91cfca667e05d)

matejcik (approved):

core/embed/rust/src/ui/model_mercury/flow/util.rs
```diff
@@ -105,18 +106,13 @@ impl ConfirmBlobParams {
...
-    }
-
     pub const fn with_verb(mut self, verb: Option<TString<'static>>) -> Self {
         self.verb = verb;
         self
     }
 
-    pub const fn with_verb_cancel(mut self, verb_cancel: Option<TString<'static>>) -> Self {
-        self.verb_cancel = verb_cancel;
+    pub fn with_verb_cancel(mut self, verb_cancel: Option<TString<'static>>) -> Self {
```
obrusvit, line 114: could be `const`

obrusvit (approved):

core/embed/rust/src/ui/model_mercury/flow/util.rs
matejcik: not anymore due to the `unwrap_or`.
you _could_ rewrite it as a match statement but it's not worth it, nobody is going to call this in a const context, right? (or if they do they can modify it)

ibz: It has `.into()` so no...

matejcik: (`into` is a smaller problem because you can do `TString::from_translation` that is const. it's the `unwrap_or` that would need to be spelled out a different way)

core/.changelog.d/4302.added
```diff
@@ -0,0 +1,4 @@
+[T3T1] Add page counter to paginated blobs.
+[T3T1] Add cancel button to individual pages of a blob.
+[T3T1] Add margin to the first page of a blob.
+[T3B1] Add "continue" button after scrolling through the whole blob.
```
matejcik, line 4: each of those needs its own file (`4302.added`, `4302.added.1`, `4302.added.2` etc)

(check the output of `./CHANGELOG.unreleased` in core dir

obrusvit: I wonder why do we need such granularity in the changelog though. Does it mean anything for the customer?

bosomt:
**QA OK**

tested on TS5, 
i will retest on more models later

bosomt:
**QA OK**

tested on T3B1 and T2T1 /latest 2.8.7 dev version/

PR #4347, opened 2024-11-14
remove MPU_MODE_KERNEL_SRAM

TychoVrahe:
MPU_MODE_KERNEL_SRAM is removed as it doesn't have any effect.

The SRAM is always accessible by kernel, and blocked in APP mode. Making it explicitly available again doesn't provide any value.

Only placed it was used was consumption mask on F4. This still works after removal on T2B1.


cepetr (approved):
utACK

PR #4348, opened 2024-11-14
chore(core): remove loop.chan

M1nd3r:
All usage of `loop.chan` in the THP branch was replaced by the new `loop.mailbox` (introduced here: https://github.com/trezor/trezor-firmware/commit/9f01c9f896e9268142e63a134642432579890a77). The `loop.chan` is no longer used in the `main` branch thanks to the [RustLayout rework, Global layout object ](https://github.com/trezor/trezor-firmware/pull/3686) PR.

If there are no objections, the `loop.chan` can be removed from the codebase.


obrusvit:
The description of `loop.chan` from `docs/core/src/event-loop.md` could also be removed.

Ideally, we'd like a short documentation for `loop.mailbox`. Not necessarily in this PR.

M1nd3r:
> The description of `loop.chan` from `docs/core/src/event-loop.md` could also be removed.
> 
> Ideally, we'd like a short documentation for `loop.mailbox`. Not necessarily in this PR.

The description of `loop.chan` was removed and a (short) description of `loop.mailbox` added here https://github.com/trezor/trezor-firmware/pull/4348/commits/3ccc462e93c1c307e969f73048ae7f9ec31502af

mmilata (approved):

PR #4349, opened 2024-11-14
THP preparation

M1nd3r:
Minor changes from the THP branch that can be merged already.

- docs: fixes in documentation and some typo corrections
- protob: reservation of wire types for THP messages in the `messages.proto` file
- build: changes in Makefile and SConscripts affecting THP only
- protob: allow encoding of protobuf messages into memoryview
- refactor: use TYPE CHECKING instead of direct import and fix style of type ignore statement
- codec: remove unused session_id (it is/was always a constant 0)

### DIFF
--- a/core/src/apps/debug/__init__.py
+++ b/core/src/apps/debug/__init__.py
@@ -360,7 +360,7 @@ if __debug__:
 
         global DEBUG_CONTEXT
 
-        DEBUG_CONTEXT = ctx = context.Context(iface, 0, WIRE_BUFFER_DEBUG)
+        DEBUG_CONTEXT = ctx = context.Context(iface, WIRE_BUFFER_DEBUG)
 
         if storage.layout_watcher:
             try:
@@ -385,9 +385,8 @@ if __debug__:
                     msg_type = f"{msg.type} - unknown message type"
                 log.debug(
                     __name__,
-                    "%s:%x receive: <%s>",
+                    "%d receive: <%s>",
                     ctx.iface.iface_num(),
-                    ctx.sid,
                     msg_type,
                 )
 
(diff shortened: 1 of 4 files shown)


core/src/trezor/ui/layouts/common.py
```diff
@@ -64,7 +64,7 @@ def raise_if_not_confirmed(
     exc: ExceptionType = ActionCancelled,
 ) -> Awaitable[None]:
     action = interact(layout_obj, br_name, br_code, exc)
-    return action  # type: ignore ["UiResult" is incompatible with "None"]
+    return action  # type: ignore [Expression of type "Coroutine[Any, Any, UiResult]" is incompatible with return type "Awaitable[None]"]
```
mmilata, line 67: Was pyright unhappy about this, and why didn't CI catch it?

mmilata (approved):
LGTM, nothing controversial here

core/src/trezor/ui/layouts/common.py
M1nd3r: I was having some problems locally, but CI is (probably) happy with both versions. Do you want to have the old version back  (i.e. do not fix it if it is not broken) or is it ok as I have it in this PR?

common/protob/messages.proto
```diff
@@ -324,4 +324,7 @@ enum MessageType {
     MessageType_BenchmarkNames = 9101 [(bitcoin_only) = true];
     MessageType_BenchmarkRun = 9102 [(bitcoin_only) = true];
     MessageType_BenchmarkResult = 9103 [(bitcoin_only) = true];
+
+    // THP
+    reserved 1000 to 1099; // See messages-thp.proto
```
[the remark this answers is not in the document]
M1nd3r: fixed here: https://github.com/trezor/trezor-firmware/commit/a9d81b355bab5e7788ca993738473fe43277215b

core/src/trezor/ui/layouts/common.py
M1nd3r: Returned back to the previous state (as in `main`)

Issue #4319, opened 2024-11-05
Path doesn't fit line when confirming internal output on T3T1


onvej-sl:
This screen is displayed on T3T1 within the signing workflow when a user is confirming an output belonging to them. 
![image](https://github.com/user-attachments/assets/fba629f2-22d2-469f-acca-006bac496361)
The text `address path m/45'/0/1/1` does not fit the line. Instead of being wrapped, it is shortened.

This is an analogous screen on T2T1:

![image](https://github.com/user-attachments/assets/4a69026f-5b18-4b62-b9e7-1ea0dc35a7ab)

Hannsek:
Do we have to put it in the header?
What about doing it similar to this?

![image](https://github.com/user-attachments/assets/933d7853-4b69-445f-b9ba-97b522149e0a)

PR #4351, opened 2024-11-14
Multisig fix

onvej-sl:
* https://github.com/trezor/trezor-firmware/pull/4351/commits/d99fea38d791a44b4add9f2c7a0e67a84fd36318 and https://github.com/trezor/trezor-firmware/pull/4351/commits/cae49da30e5dca0c5e006088aa282a85f9f89e77 disallow per-node paths which fixes https://github.com/satoshilabs/trezor-firmware/issues/191.
* https://github.com/trezor/trezor-firmware/pull/4351/commits/d19cbfa3cbfa1252c75184ae5eb1f298f519305d introduces sorting of pubkeys in multisig which fixes https://github.com/satoshilabs/trezor-firmware/issues/193.
* https://github.com/trezor/trezor-firmware/pull/4351/commits/267c1c7aaac3fcefed0aeac994d7f58f65d23502 removes a path that is not used anymore (see https://satoshilabs.slack.com/archives/C01L1HT4B46/p1730908353546419). This was the only allowed path that doesn't end with `/change/address_index`. This caused some paths weren't properly recognized as change outputs. However, the main reason for removing this path is primarily to prevent further problems in the future.
* https://github.com/trezor/trezor-firmware/pull/4351/commits/d9b7a57e8b478fa8fa3be3463f08f242d95c4936 forbid multisig to singlesig change outputs. This allows users to have greater control over where their funds are sent because since https://github.com/trezor/trezor-firmware/pull/2925, we also display the path for internal outputs.  Nevertheless, this doesn't always work as expected, see https://github.com/trezor/trezor-firmware/issues/4319).

None of these changes were backported to legacy.


onvej-sl:
Related issue: https://github.com/trezor/trezor-firmware/issues/4356

onvej-sl:
@andrewkozlik please do a high-level review, ensuring particularly that this pull request fixes the mentioned issue.
@mmilata please do a generic review.

core/src/apps/bitcoin/sign_tx/matchcheck.py
```diff
@@ -107,6 +107,18 @@ def attribute_from_tx(self, txio: TxInput | TxOutput) -> Any:
         return multisig.multisig_fingerprint(txio.multisig)
 
 
+class PubkeysOrderChecker(MatchChecker):
+    def attribute_from_tx(self, txio: TxInput | TxOutput) -> Any:
+        if not txio.multisig:
+            return None
+        return txio.multisig is not None and txio.multisig.pubkeys_order
```
mmilata, line 114: Left hand side is always true, is it intentional?

core/src/trezor/ui/layouts/mercury/__init__.py
```diff
@@ -139,6 +139,17 @@ def confirm_multisig_warning() -> Awaitable[None]:
     )
 
 
+def confirm_multisig_different_paths_warning() -> Awaitable[None]:
+    return raise_if_not_confirmed(
+        trezorui2.flow_warning_hi_prio(
+            title=f"{TR.words__important}!",
+            description="Using different paths for different xpubs.",
```
mmilata, line 146: Nit: maybe capitalize as XPUBs?

mmilata (commented):
Looking good, found some nits, another pair of eyes on `apps/bitcoin/signing/` would be nice.

core/failing_test
```diff
@@ -0,0 +1,4535 @@
+[satoshi@ondra:/trezor-firmware/multisig/core]$ ./emu.py --headless
```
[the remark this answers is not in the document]
onvej-sl: Removed in https://github.com/trezor/trezor-firmware/pull/4351/commits/b9ca79b2c1c1559d42fa0e5cb6e4e7fcc696214f.

core/src/apps/bitcoin/sign_tx/matchcheck.py
onvej-sl: Not intentional, fixed in https://github.com/trezor/trezor-firmware/pull/4351/commits/7ed1c424dfdd7a4619c811763a28e2d78f0743dd.

legacy/firmware/transaction.c
```diff
@@ -391,6 +394,11 @@ uint32_t compile_script_multisig(const CoinInfo *coin,
 uint32_t compile_script_multisig_hash(const CoinInfo *coin,
                                       const MultisigRedeemScriptType *multisig,
                                       uint8_t *hash) {
+  if (multisig->pubkeys_order != MultisigPubkeysOrder_PRESERVED) {
+    fsm_sendFailure(FailureType_Failure_DataError,
+                    _("Sortedmulti is not supported"));
```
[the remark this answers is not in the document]
onvej-sl: Definitely. Fixed in https://github.com/trezor/trezor-firmware/pull/4351/commits/e7277e259129ff0569e84ad2ddfa27d6bac9072a.

legacy/firmware/transaction.c
```diff
@@ -436,6 +444,10 @@ uint32_t serialize_script_sig(const uint8_t *signature, uint32_t signature_len,
 uint32_t serialize_script_multisig(const CoinInfo *coin,
                                    const MultisigRedeemScriptType *multisig,
                                    uint8_t sighash, uint8_t *out) {
+  if (multisig->pubkeys_order != MultisigPubkeysOrder_PRESERVED) {
+    fsm_sendFailure(FailureType_Failure_DataError,
+                    _("Sortedmulti is not supported"));
```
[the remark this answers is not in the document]
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/4351/commits/e7277e259129ff0569e84ad2ddfa27d6bac9072a.

core/.changelog.d/4351.added
```diff
@@ -0,0 +1 @@
+Added support for sortedmulti.
```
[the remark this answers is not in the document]
onvej-sl: We can also mention [BIP383](https://github.com/bitcoin/bips/blob/master/bip-0383.mediawiki) or [BIP67](https://github.com/bitcoin/bips/blob/master/bip-0067.mediawiki). I'm hesitant to mention descriptors because I think we [do not support](https://github.com/trezor/trezor-firmware/blob/onvej-sl/multisig-fix/core/src/apps/bitcoin/get_public_key.py#L123) multisig descriptors at all. Anyway, "sortedmulti" itself might be somewhat misleading as well.

What about "Added support for lexicographic sorting of pubkeys"?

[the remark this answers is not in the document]
mmilata: Sounds good, maybe even _... multisig pubkeys_ or _... pubkeys for multisig_?

core/.changelog.d/4351.changed.1
```diff
@@ -0,0 +1 @@
+Forbidden multisig to singlesig change outputs.
```
andrewkozlik, line 1: I think the message should read "Forbid ..." or possibly "Forbade ...".

core/src/apps/bitcoin/multisig.py
```diff
@@ -28,7 +29,8 @@ def multisig_fingerprint(multisig: MultisigRedeemScriptType) -> bytes:
         if len(d.public_key) != 33 or len(d.chain_code) != 32:
             raise DataError("Invalid multisig parameters")
 
-    pubnodes = sorted(pubnodes, key=lambda n: n.public_key + n.chain_code)
+    if multisig.pubkeys_order == MultisigPubkeysOrder.LEXICOGRAPHIC:
+        pubnodes = sorted(pubnodes, key=lambda n: n.public_key + n.chain_code)
```
andrewkozlik, line 33: Add comment explaining why we sort the XPUBs.

core/src/apps/bitcoin/common.py
```diff
@@ -263,3 +264,11 @@ def _polymod(c: int, val: int) -> int:
     for j in range(0, 8):
         ret[j] = CHECKSUM_CHARSET[(c >> (5 * (7 - j))) & 31]
     return "".join(ret)
+
+
+def multisig_uses_single_path(multisig: MultisigRedeemScriptType) -> bool:
+    if multisig.pubkeys is not None:
```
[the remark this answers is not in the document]
andrewkozlik: Add comment explaining the two scenarios that may be going on here. Also we may want to get rid of `if`.

core/src/apps/bitcoin/get_address.py
```diff
@@ -107,6 +114,15 @@ async def get_address(msg: GetAddress, keychain: Keychain, coin: CoinInfo) -> Ad
 
             await confirm_multisig_warning()
 
+            # An addresss that uses different derivation paths for different xpubs could be difficult to discover
```
andrewkozlik, line 117: typo "addresss"

andrewkozlik: Also, how about adding a bit more to the rationale, along the lines that malicious software on the host computer could use this to execute a ransom attack.

core/.changelog.d/4351.changed.3
```diff
@@ -0,0 +1 @@
+Forbidden per-node paths in multisig.
```
andrewkozlik, line 1: Also "Forbid ..." or "Forbade ...", preferably the former. Also this concerns only change outputs and GetAddress, so I'd be specific about that. Maybe "Forbid different sub-paths in multisig change-outputs and receive addresses."

core/src/apps/bitcoin/sign_tx/matchcheck.py
```diff
@@ -107,6 +107,18 @@ def attribute_from_tx(self, txio: TxInput | TxOutput) -> Any:
         return multisig.multisig_fingerprint(txio.multisig)
```
andrewkozlik, line 107: Since we removed the logic which accepts singlesig change outputs even when multisig inputs are present, we can simplify the code a little bit by doing the following here:
```py
if not txio.multisig:
    return bytes()
```
What this would do is allow us to:
1. get rid of the special multisig handling in `ChangeDetector.def output_is_change()`, where we only check the fingerprint for multisig outputs, and
2. get rid of the `MultisigChecker` altogether.

core/src/apps/bitcoin/sign_tx/matchcheck.py
```diff
@@ -107,6 +107,18 @@ def attribute_from_tx(self, txio: TxInput | TxOutput) -> Any:
...
 
 
+class PubkeysOrderChecker(MatchChecker):
+    def attribute_from_tx(self, txio: TxInput | TxOutput) -> Any:
+        if not txio.multisig:
+            return None
+        return txio.multisig.pubkeys_order
```
andrewkozlik, line 114: I think it's perfectly clean to hash the `pubkeys_order` attribute into the multisig fingerprint, since the fingerprint already contains several attributes of `MultisigRedeemScriptType` (XPUBs and threshold). Then we can get rid of `PubkeysOrderChecker` while maintaining the same behavior.

core/src/apps/bitcoin/common.py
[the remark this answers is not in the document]
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4351/commits/edd0fcab3d910c606952fee325a74fa3cfff424a.

core/src/apps/bitcoin/multisig.py
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4351/commits/edd0fcab3d910c606952fee325a74fa3cfff424a and https://github.com/trezor/trezor-firmware/pull/4351/commits/c703cacdcce782dacc8b41c4302ea95ee0f0760a.

onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4351/commits/4e52009e889fca6cc7f8f2905c9b3ea2250ea429.

core/src/apps/bitcoin/sign_tx/matchcheck.py
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4351/commits/4e52009e889fca6cc7f8f2905c9b3ea2250ea429.

core/src/apps/bitcoin/sign_tx/matchcheck.py
mmilata: I'm generally fan of explicit code but this seems fine, especially if you add a comment.

core/.changelog.d/4351.changed.1
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4351/commits/54e4e27910b05c33658dfe5dd2f7d4ed7448c97d.

core/.changelog.d/4351.changed.3
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4351/commits/c5200261fbe0fa28d0c2f826f472e1336cba674c.

core/src/apps/bitcoin/sign_tx/matchcheck.py
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/4351/commits/b7d26c794b8839fb9a35208d3b7a4ae336424480.

andrewkozlik (approved):
High level this LGTM. Good job! Awaiting final approval from @mmilata.

@Hannsek we will need 3rd party multisig providers (Unchained, Casa, ...) to test that these changes do not break their setups. I believe you have the full set of contacts. Please coordinate with them before the next firmware release.

mmilata (approved):

andrewtoth:
Doesn't this break [`multisig_pubkey_index` ](https://github.com/trezor/trezor-firmware/pull/4351/files#diff-c528ee9407682668ced569943303086cb4098504c4a3684aa85b2f766b283763R60) if using lexigraphic order, since that will return the index of the pubkey if unsorted?

andrewtoth:
That's because that test does not use that function in a breaking way. That is testing unsigned txs. The `multisig_pubkey_index` function breaks when it is used for inserting the signatures.

andrewtoth:
I have it fixed in https://github.com/trezor/trezor-firmware/pull/4159/commits/d06623218f75e87a1f27a904cf3d476e5be57a37

onvej-sl:
> That's because that test does not use that function in a breaking way. That is testing unsigned txs. The multisig_pubkey_index function breaks when it is used for inserting the signatures.

I'm not sure if this is exactly what you're saying, but I discovered that the test fails to catch it because I use empty signatures in `MultisigRedeemScriptType`.

> I have it fixed in https://github.com/trezor/trezor-firmware/commit/d06623218f75e87a1f27a904cf3d476e5be57a37

The fix is not completely correct because `multisig_pubkey_index` is used [here](https://github.com/trezor/trezor-firmware/blob/748a19ac5d6e81c8ab22b00fa1ae467712f07a35/core/src/apps/bitcoin/get_address.py#L113) to detect which xpub belongs to you and the xpubs are not sorted.

PR #4354, opened 2024-11-15
Fix bitcoin sign message and fee bump of mercury UI

obrusvit:
This PR fixes 2 linked issues regarding BTC transactions, namely:
- info in sign message
- info in bump fee transaction

Considerations:
- `confirm_with_info` is used for now in both cases - in the future, we should use `confirm_action` parameterized accordingly
- bump fee has labels with colon, which is typically not used in `mercury`, however, moving the formatting to uPy would require more effort - so I decided to skip it for now
- user is unable to swipe down from "message" to "signing address" screen. Cancel is possible from both screens.

TODO:
- [x] one test gets stuck - InputFlow
    - was a problem with incorrect `pages` count of `confirm_with_info` flow. This was then incorrectly handled by `default_input_flow`


matejcik (approved):
utACK

ibz:
I am not very happy that we are extending `confirm_with_info` to make it even more like `confirm_action` when, what I believe, should be steps to move the opposite direction (removing duplication, possibly removing `confirm_with_info` completely). I'll take it onto myself to investigate this (after this PR will have been merged), since I have my hands dirty with `confirm_blob` and `confirm_action` ATM.

ibz (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.1.0 (0b116799334cb4a05fcb59cc480deb7c4b9a63e9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.1.0 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.7 regular (revision e196413bb745443610211c325623ba7cee7309e9)

PR #4357, opened 2024-11-15
Abstraction of cache and context

M1nd3r:
Abstraction of cache and context so that the current codec_v1 implementation can be switched for THP version.

### DIFF
--- a/core/SConscript.firmware
+++ b/core/SConscript.firmware
@@ -565,6 +565,7 @@ if FROZEN:
     ))
 
     SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/wire/*.py'))
+    SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/wire/codec/*.py'))
 
     SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'storage/*.py',
         exclude=[
--- a/core/src/apps/common/authorization.py
+++ b/core/src/apps/common/authorization.py
@@ -1,23 +1,20 @@
 from typing import Iterable
 
-import storage.cache as storage_cache
+from storage.cache_common import (
+    APP_COMMON_AUTHORIZATION_DATA,
+    APP_COMMON_AUTHORIZATION_TYPE,
+)
 from trezor import protobuf
 from trezor.enums import MessageType
+from trezor.wire import context
 
 WIRE_TYPES: dict[int, tuple[int, ...]] = {
     MessageType.AuthorizeCoinJoin: (MessageType.SignTx, MessageType.GetOwnershipProof),
 }
 
-APP_COMMON_AUTHORIZATION_DATA = (
-    storage_cache.APP_COMMON_AUTHORIZATION_DATA
-)  # global_import_cache
-APP_COMMON_AUTHORIZATION_TYPE = (
-    storage_cache.APP_COMMON_AUTHORIZATION_TYPE
-)  # global_import_cache
-
 
 def is_set() -> bool:
-    return storage_cache.get(APP_COMMON_AUTHORIZATION_TYPE) is not None
+    return context.cache_get(APP_COMMON_AUTHORIZATION_TYPE) is not None
 
 
 def set(auth_message: protobuf.MessageType) -> None:
@@ -29,27 +26,27 @@ def set(auth_message: protobuf.MessageType) -> None:
     # (because only wire-level messages have wire_type, which we use as identifier)
     ensure(auth_message.MESSAGE_WIRE_TYPE is not None)
     assert auth_message.MESSAGE_WIRE_TYPE is not None  # so that typechecker knows too
-    storage_cache.set_int(APP_COMMON_AUTHORIZATION_TYPE, auth_message.MESSAGE_WIRE_TYPE)
-    storage_cache.set(APP_COMMON_AUTHORIZATION_DATA, buffer)
+    context.cache_set_int(APP_COMMON_AUTHORIZATION_TYPE, auth_message.MESSAGE_WIRE_TYPE)
+    context.cache_set(APP_COMMON_AUTHORIZATION_DATA, buffer)
 
 
 def get() -> protobuf.MessageType | None:
-    stored_auth_type = storage_cache.get_int(APP_COMMON_AUTHORIZATION_TYPE)
+    stored_auth_type = context.cache_get_int(APP_COMMON_AUTHORIZATION_TYPE)
     if not stored_auth_type:
         return None
 
-    buffer = storage_cache.get(APP_COMMON_AUTHORIZATION_DATA, b"")
+    buffer = context.cache_get(APP_COMMON_AUTHORIZATION_DATA, b"")
     return protobuf.load_message_buffer(buffer, stored_auth_type)
 
 
 def is_set_any_session(auth_type: MessageType) -> bool:
-    return auth_type in storage_cache.get_int_all_sessions(
+    return auth_type in context.cache_get_int_all_sessions(
         APP_COMMON_AUTHORIZATION_TYPE
     )
 
 
 def get_wire_types() -> Iterable[int]:
-    stored_auth_type = storage_cache.get_int(APP_COMMON_AUTHORIZATION_TYPE)
+    stored_auth_type = context.cache_get_int(APP_COMMON_AUTHORIZATION_TYPE)
     if stored_auth_type is None:
         return ()
 
@@ -57,5 +54,5 @@ def get_wire_types() -> Iterable[int]:
 
 
 def clear() -> None:
-    storage_cache.delete(APP_COMMON_AUTHORIZATION_TYPE)
-    storage_cache.delete(APP_COMMON_AUTHORIZATION_DATA)
+    context.cache_delete(APP_COMMON_AUTHORIZATION_TYPE)
+    context.cache_delete(APP_COMMON_AUTHORIZATION_DATA)
--- a/core/src/apps/common/backup.py
+++ b/core/src/apps/common/backup.py
@@ -1,25 +1,27 @@
 from typing import TYPE_CHECKING
 
-import storage.cache as storage_cache
+from storage.cache_common import APP_RECOVERY_REPEATED_BACKUP_UNLOCKED
 from trezor import wire
 from trezor.enums import MessageType
+from trezor.wire import context
+from trezor.wire.message_handler import filters, remove_filter
 
 if TYPE_CHECKING:
     from trezor.wire import Handler, Msg
 
 
 def repeated_backup_enabled() -> bool:
-    return storage_cache.get_bool(storage_cache.APP_RECOVERY_REPEATED_BACKUP_UNLOCKED)
+    return context.cache_get_bool(APP_RECOVERY_REPEATED_BACKUP_UNLOCKED)
 
 
 def activate_repeated_backup() -> None:
-    storage_cache.set_bool(storage_cache.APP_RECOVERY_REPEATED_BACKUP_UNLOCKED, True)
-    wire.filters.append(_repeated_backup_filter)
+    context.cache_set_bool(APP_RECOVERY_REPEATED_BACKUP_UNLOCKED, True)
+    filters.append(_repeated_backup_filter)
 
 
 def deactivate_repeated_backup() -> None:
-    storage_cache.delete(storage_cache.APP_RECOVERY_REPEATED_BACKUP_UNLOCKED)
-    wire.remove_filter(_repeated_backup_filter)
+    context.cache_delete(APP_RECOVERY_REPEATED_BACKUP_UNLOCKED)
+    remove_filter(_repeated_backup_filter)
 
 
 _ALLOW_WHILE_REPEATED_BACKUP_UNLOCKED = (
--- a/core/src/storage/cache.py
+++ b/core/src/storage/cache.py
@@ -1,39 +1,39 @@
 import builtins
 import gc
-from micropython import const
-from typing import TYPE_CHECKING
 
-from trezor import utils
-
-if TYPE_CHECKING:
-    from typing import Sequence, TypeVar, overload
-
-    T = TypeVar("T")
+from storage import cache_codec
+from storage.cache_common import SESSIONLESS_FLAG, SessionlessCache
 
 
-_MAX_SESSIONS_COUNT = const(10)
-_SESSIONLESS_FLAG = const(128)
-_SESSION_ID_LENGTH = const(32)
+# Cache initialization
+_SESSIONLESS_CACHE = SessionlessCache()
+_PROTOCOL_CACHE = cache_codec
+_PROTOCOL_CACHE.initialize()
+_SESSIONLESS_CACHE.clear()
 
-# Traditional cache keys
-APP_COMMON_SEED = const(0)
-APP_COMMON_AUTHORIZATION_TYPE = const(1)
-APP_COMMON_AUTHORIZATION_DATA = const(2)
-APP_COMMON_NONCE = const(3)
-if not utils.BITCOIN_ONLY:
-    APP_COMMON_DERIVE_CARDANO = const(4)
-    APP_CARDANO_ICARUS_SECRET = const(5)
-    APP_CARDANO_ICARUS_TREZOR_SECRET = const(6)
-    APP_MONERO_LIVE_REFRESH = const(7)
+gc.collect()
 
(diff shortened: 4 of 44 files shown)


core/SConscript.firmware
```diff
@@ -557,6 +557,7 @@ if FROZEN:
     ))
 
     SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/wire/*.py'))
+    SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/wire/codec/*.py'))
```
M1nd3r, line 560: The codec specific files are separated so that they can be easily ommited for THP prod builds. When the THP implementation is added, it could look like this:
```
if THP:
   SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/wire/thp/*.py'))
else:
   SOURCE_PY.extend(Glob(SOURCE_PY_DIR + 'trezor/wire/codec/*.py'))
   ```

File: core/src/apps/common/backup.py [hunk 2]
```diff
@@ -1,25 +1,27 @@
 from typing import TYPE_CHECKING
 
-import storage.cache as storage_cache
```
M1nd3r, line 3: Is the change in import style from 
- `import storage.cache as storage_cache` and using `storage_cache.CONSTANT`
to
- `from storage.cache import CONSTANT` and using `CONSTANT`
ok, or is there a reason to NOT do it the way it is in the PR? @mmilata

core/SConscript.firmware
M1nd3r: Note: The condition `or PYOPT == '0'` is thee because the codec is used on DebugLink even with THP builds.

mmilata (commented):
will continue tomorrow

core/src/apps/common/authorization.py
```diff
@@ -1,23 +1,20 @@
...
     MessageType.AuthorizeCoinJoin: (MessageType.SignTx, MessageType.GetOwnershipProof),
 }
 
-APP_COMMON_AUTHORIZATION_DATA = (
-    storage_cache.APP_COMMON_AUTHORIZATION_DATA
-)  # global_import_cache
-APP_COMMON_AUTHORIZATION_TYPE = (
-    storage_cache.APP_COMMON_AUTHORIZATION_TYPE
-)  # global_import_cache
-
```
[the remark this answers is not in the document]
M1nd3r: Ok, I will keep it as it is in the PR. Any optimizations can be done later separately.

core/src/apps/common/backup.py
M1nd3r: ack

core/tests/unittest.py
```diff
@@ -236,38 +236,48 @@ def wasSuccessful(self):
...
+    try:
+        set_up_class()
+        for name in dir(o):
+            if name.startswith("test"):
+                run_test_method(o, name, set_up, tear_down, test_result)
+    finally:
+        tear_down_class()
+
+
+def run_test_method(o, name, set_up, tear_down, test_result):
```
M1nd3r, line 253: Personally, I would prefer a function with less parameters. However:
- it is not possible to pass `m` instead of `o` and `name`, as getting `name=m.__name__` will fail for test functions with decorators (e.g.` test_apps.common.keychain.py::test_verify_path`)
- getting `setUp` and `tearDown` for each test in class is redundant
- it would be possible to encapsulate the arguments into a single `test_args` object, but imo it might be an overkill for the use case
- inlining the `run_test` function will make the code less readable because of the complex nesting

With those points in mind, I think that the 5 parameters might be an ok solution.

core/src/trezor/wire/message_handler.py
```diff
@@ -0,0 +1,234 @@
...
+# If set to False protobuf messages marked with "experimental_message" option are rejected.
+EXPERIMENTAL_ENABLED = False
+
+
+def wrap_protobuf_load(
+    buffer: bytes,
+    expected_type: type[LoadedMessageType],
+) -> LoadedMessageType:
+    try:
+        if __debug__:
```
mmilata, line 29: Do you still need this? Can we add `and utils.EMULATOR`? I'm a bit worried about performance of our hardware tests.

core/src/trezor/wire/message_handler.py
```diff
@@ -0,0 +1,234 @@
...
+if TYPE_CHECKING:
+    from typing import Any, Callable, Container
+
+    from trezor.wire import Handler, LoadedMessageType, WireInterface
+
+    HandlerFinder = Callable[[Any, Any], Handler | None]
+    Filter = Callable[[int, Handler], Handler]
+
+# If set to False protobuf messages marked with "experimental_message" option are rejected.
+EXPERIMENTAL_ENABLED = False
```
mmilata, line 21: Note to self: shouldn't this also be somewhere under `storage/cache` to avoid being wiped by a restart?

core/tests/test_apps.bitcoin.keychain.py
```diff
@@ -1,17 +1,25 @@
...
 
 from apps.bitcoin.keychain import _get_coin_by_name, _get_keychain_for_coin
+from trezor.wire.codec.codec_context import CodecContext
+from storage import cache_codec
 
 
 class TestBitcoinKeychain(unittest.TestCase):
+
+    def __init__(self):
+        context.CURRENT_CONTEXT = CodecContext(None, bytearray(64))
```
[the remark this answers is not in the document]
mmilata: Thanks for implementing this!

mmilata (commented):
Did not found any major issue.

@matejcik can you please go through 8fe6fe5071363b82f94813bbab0f609681bbb97a real quick if you can spot anything that breaks the workflow restart/module unloading?

core/src/storage/cache_common.py
```diff
@@ -0,0 +1,184 @@
...
+    APP_CARDANO_ICARUS_TREZOR_SECRET = const(6)
+    APP_MONERO_LIVE_REFRESH = const(7)
+
+# Cache keys for THP channel
+if utils.USE_THP:
+    CHANNEL_HANDSHAKE_HASH = const(0)
+    CHANNEL_KEY_RECEIVE = const(1)
+    CHANNEL_KEY_SEND = const(2)
+    CHANNEL_NONCE_RECEIVE = const(3)
+    CHANNEL_NONCE_SEND = const(4)
```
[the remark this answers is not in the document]
M1nd3r: There are 2 things:

1) THP constants `CHANNEL...` should not be in this PR, as they are not used in the current codebase anywhere (yet). I will remove them.

2) My reason to put them in `cache_common` was that the rest of the codebase should/could (ideally) use only `cache.py` and `cache_common.py` files and be oblivious to the division into codec/thp cache. Also, having all the constants in one file makes it easier to find them (if you do not know their name). However, the files that will be using constants `CHANNEL_...`  must inherently use THP -> moving them to `cache_thp` makes sense from the architectural POV.

@mmilata  What do you think makes more sense - "interface-y" approach with all cache constants in cache_common or "structural" approach having thp specific (public) constants in the `cache_thp`?

core/src/storage/cache.py
```diff
@@ -156,210 +12,33 @@ def __init__(self) -> None:
 # bytearrays, then later call `clear()` on all the existing objects, which resets them
 # to zero length. This is producing some trash - `b[:]` allocates a slice.
```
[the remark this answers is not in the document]
M1nd3r: Well, I thought about it previously. The problem with this "comment" is that it explains the process of allocation and the logic behind it.
- Allocation is two-step: first allocate, then clear.
- Note that using b[:] allocates a slice.

 The allocation process actually happens just below where the commit is:
 ```

 _SESSIONLESS_CACHE = ...  # step 1
 _PROTOCOL_CACHE.initialize()    # step 1 and step 2
 _SESSIONLESS_CACHE.clear() # step 2
 
 ```
 
 The problem I see is that it is more abstract now - the "real" allocation (and use of slices) happens partly inside a _PROTOCOL_CACHE and SessionlessCache -  that are in different files now.
 
 I see a few solutions:
 1) keep it as it is now - it works as a general documentation
 2) remove the comment, move the implementation information into docs
 3) have the comment in each of the implementations (cache_codec, cache_thp)
 4) some mix of 2. and 3., divide the allocation notes and use relevant parts where necessary
 
 @mmilata WDYT?

core/src/storage/cache_common.py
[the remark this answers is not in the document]
M1nd3r: ad 1) Thp-specific constants removed here https://github.com/trezor/trezor-firmware/pull/4357/commits/8c6ed703897c697bc6610364a95481246766b1e7

core/src/storage/cache_common.py
```diff
@@ -0,0 +1,184 @@
...
+    def set(self, key: int, value: bytes) -> None:
+        utils.ensure(key < len(self.fields))
+        utils.ensure(len(value) <= self.fields[key])
+        self.data[key][0] = 1
+        self.data[key][1:] = value
+
+    def set_bool(self, key: int, value: bool) -> None:
+        utils.ensure(
+            self._get_length(key) == 0, "Field does not have zero length!"
+        )  # skipping get_length in production build
```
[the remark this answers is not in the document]
M1nd3r: fixed here: https://github.com/trezor/trezor-firmware/pull/4357/commits/f180c32bb85d70d361112143837bff9d84754e00

core/src/storage/cache_common.py
```diff
@@ -0,0 +1,184 @@
...
+        if value:
+            self.set(key, b"")
+        else:
+            self.delete(key)
+
+    def set_int(self, key: int, value: int) -> None:
+        length = self.fields[key]
```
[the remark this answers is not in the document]
M1nd3r: `_get_length` returned back here: https://github.com/trezor/trezor-firmware/pull/4357/commits/f180c32bb85d70d361112143837bff9d84754e00

core/src/trezor/wire/protocol_common.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(
+        self,
+        message_type: int,
+        message_data: bytes,
+    ) -> None:
+        self.data = message_data
+        self.type = message_type
+
+    def to_bytes(self) -> bytes:
+        return self.type.to_bytes(2, "big") + self.data
```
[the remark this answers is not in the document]
M1nd3r: Not used, removed here: https://github.com/trezor/trezor-firmware/pull/4357/commits/91d477c3c4a5ec331ce4410b11c941f94fe93cad

core/src/trezor/wire/codec/codec_context.py
```diff
@@ -0,0 +1,134 @@
...
+
+        @overload
+        async def read(
+            self, expected_types: Container[int]
+        ) -> protobuf.MessageType: ...
+
+        @overload
+        async def read(
+            self, expected_types: Container[int], expected_type: type[LoadedMessageType]
+        ) -> LoadedMessageType: ...
```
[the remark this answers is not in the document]
M1nd3r: Yes, you are right. Dropped here: https://github.com/trezor/trezor-firmware/pull/4357/commits/91d477c3c4a5ec331ce4410b11c941f94fe93cad

core/src/trezor/wire/message_handler.py
M1nd3r: Condition `and utils.EMULATOR` added here: https://github.com/trezor/trezor-firmware/pull/4357/commits/91d477c3c4a5ec331ce4410b11c941f94fe93cad

core/src/trezor/wire/codec/codec_context.py
```diff
@@ -0,0 +1,134 @@
...
+
+    reading: bool = False
```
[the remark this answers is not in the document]
M1nd3r: removed: https://github.com/trezor/trezor-firmware/pull/4357/commits/91d477c3c4a5ec331ce4410b11c941f94fe93cad

core/src/storage/cache.py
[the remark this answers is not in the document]
mmilata: I'd go with 3., some duplication won't hurt. Feel free to extend the docs if it makes sense to you.

core/src/trezor/wire/protocol_common.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class Context:
```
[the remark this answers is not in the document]
mmilata: Methods too please, ideally also the new ones.

core/src/trezor/wire/protocol_common.py
```diff
@@ -58,9 +58,6 @@ async def read(
 
     async def write(self, msg: protobuf.MessageType) -> None: ...
 
-    def write_force(self, msg: protobuf.MessageType) -> Awaitable[None]:
```
M1nd3r: Note: Removed `write_force`  as it is not used by any code in `main` or in the PR. @mmilata

core/src/trezor/wire/protocol_common.py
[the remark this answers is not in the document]
M1nd3r: Done here: https://github.com/trezor/trezor-firmware/pull/4357/commits/7f11fea37f7b9205def0a4a76289f5afefa3c9e0

core/src/trezor/wire/message_handler.py
mmilata: Been thinking about this that it will probably fill CI logs with a lot of extra text, can we perhaps enable it when THP is enabled?

M1nd3r: Sure, why not: done here https://github.com/trezor/trezor-firmware/pull/4357/commits/d303dced7de86dead01563fa848532cb36be5811

core/src/trezor/wire/__init__.py
```diff
@@ -23,15 +23,13 @@
...
-from trezor import log, loop, protobuf, utils, workflow
-from trezor.enums import FailureType
-from trezor.messages import Failure
-from trezor.wire import codec_v1, context
-from trezor.wire.errors import ActionCancelled, DataError, Error, UnexpectedMessage
+from trezor import log, loop, protobuf, utils
+from trezor.wire import message_handler, protocol_common
+from trezor.wire.codec.codec_context import CodecContext
+from trezor.wire.context import UnexpectedMessageException
+from trezor.wire.message_handler import WIRE_BUFFER, failure
```
matejcik, line 32: while we're here, all `from trezor.wire` should be relative imports here.

(that btw makes it somewhat cleaner to see what is auto-imported together with `trezor.wire`)

core/src/trezor/wire/message_handler.py
```diff
@@ -0,0 +1,234 @@
...
+            log.exception(__name__, e)
+        if e.args:
+            raise DataError("Failed to decode message: " + " ".join(e.args))
+        else:
+            raise DataError("Failed to decode message")
+
+
+_PROTOBUF_BUFFER_SIZE = const(8192)
+
+WIRE_BUFFER = bytearray(_PROTOBUF_BUFFER_SIZE)
```
matejcik, line 52: this file does not _use_ `WIRE_BUFFER` and it's super weird to have this supposedly persistent buffer stowed away in a file that doesn't use it for safe-keeping.

why not keep it in `wire.__init__` where it was before?
(moving it may make sense if you also move `handle_session`)

core/src/trezor/wire/message_handler.py
matejcik: perhaps, but what we do is re-set it manually from storage anyway, which works even on a freshly started device so :woman_shrugging:

matejcik:
> @matejcik can you please go through [8fe6fe5](https://github.com/trezor/trezor-firmware/commit/8fe6fe5071363b82f94813bbab0f609681bbb97a) real quick if you can spot anything that breaks the workflow restart/module unloading?

nothing obvious jumps out at me, we'll see when this lands

core/src/trezor/wire/__init__.py
M1nd3r: changed to relative imports here: https://github.com/trezor/trezor-firmware/pull/4357/commits/eda6521d11b2d045c1d77aaa58584f5e927d9271

core/src/trezor/wire/message_handler.py
M1nd3r: `WIRE_BUFFER` moved to `wire.__init__` here: https://github.com/trezor/trezor-firmware/pull/4357/commits/eda6521d11b2d045c1d77aaa58584f5e927d9271

matejcik (approved):
utACK

PR #4358, opened 2024-11-18
feat(ci): upload emulators from all branches to data.trezor.io

grdddj:
For `Connect` end-to-end tests, we have this requirement:
- being able to run tests with an emulator from a specified firmware feature/branch, that is still under development (not merged into `main`)

Currently, emulators are built for all PRs, but not uploaded to data.trezor.io ... the artifacts can be found manually in GHActions logs, but it is not a good and reliable way of finding the URL, e.g. https://github.com/trezor/trezor-firmware/actions/runs/11856555307/artifacts/2193027474

With this change, it will be possible to dynamically get the latest emulators for all the models from a static URL - e.g. `data.trezor.io/dev/firmware/emu-branches/trezor_host_protocol/trezor-emu-core-T2B1-universal`

Notes:
- it will need to be tested how the branches with `/` in the name behave (if they create a nested dir structure) ... maybe we will need to replace all `/` in the branch name with `_`

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -736,5 +736,37 @@ jobs:
           rm unix/trezor-emu-core
           aws s3 sync --no-progress unix s3://data.trezor.io/dev/firmware/emu-nightly
 
+  core_upload_emu_branch:
+    name: Upload emulator binaries for the current branch
+    # Not building it for nightly CI
+    if: github.event_name != 'schedule'
+    runs-on: ubuntu-latest
+    needs:
+      # Do not include ARM, they are only built on nightly
+      - core_emu
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: core-emu*debuglink-noasan
+          merge-multiple: true
+      - name: Configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::538326561891:role/gh_actions_deploy_dev_firmware_data
+          aws-region: eu-west-1
+        continue-on-error: true
+      - name: Determine branch name
+        run: |
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            branch_name=${{ github.head_ref }}
+          else
+            branch_name=${{ github.ref_name }}
+          fi
+          echo "branch_name=$branch_name" >> $GITHUB_ENV
+      - name: Upload artifacts to branch directory
+        run: |
+          rm unix/trezor-emu-core
+          aws s3 sync --no-progress unix s3://data.trezor.io/dev/firmware/emu-branches/$branch_name
+
   # Connect
   # TODO: core_connect_test


grdddj:
Force-pushed the corrected version.

Examples of the generated URLs:
https://data.trezor.io/dev/firmware/emu-branches/grdddj/emulator_upload_all_branches/trezor-emu-core-T3T1-universal
https://data.trezor.io/dev/firmware/emu-branches/grdddj/emulator_upload_all_branches/trezor-emu-core-T3W1-btconly

It took 19 seconds to upload all the emulators there, so I hope it will not slow down the CI too much - https://github.com/trezor/trezor-firmware/actions/runs/11892258767/job/33135532168?pr=4358

With regards to the branch names, I suggest leaving it like this, so not replacing the `/` in branch name ... the reason being  `directories` do not have a big meaning on `data.trezor.io` (one cannot open them), so the URL is just uniquely identified with the branch name - no matter the `/` inside

For me, it is ready for review and potential merge

mmilata:
It's not that difficult to find the URL, find you need to find the latest run id for the branch name (we do this in `tests/github.py`):
```
GET https://api.github.com/repos/trezor/trezor-firmware/actions/workflows/core.yml/runs?branch={branch}
```
Then find the [artifact URLs](https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-workflow-run-artifacts):
```
GET https://api.github.com/repos/trezor/trezor-firmware/actions/runs/{run_id}/artifacts
```

That said if @vdovhanych is OK with your approach I don't have any objections. Well, except the YAML duplication but Actions don't seem to allow expressions in `needs:` without workarounds that are even worse...

.github/workflows/core.yml
```diff
@@ -736,5 +736,34 @@ jobs:
           rm unix/trezor-emu-core
           aws s3 sync --no-progress unix s3://data.trezor.io/dev/firmware/emu-nightly
 
+  core_upload_emu_branch:
+    name: Upload emulator binaries for the current branch
```
mmilata, line 740: Want to add `if: github.event_name != 'schedule'` so that it's not run twice on scheduled builds?

vdovhanych: Yes, this should be used.

grdddj:
Right, using GH-Artifacts is possible, but:
- why do the parsing work if we can easily export it on a static URL
- it is then possible for any application, not only `tenv`, to download and work with the emulators
- GH-Artifacts currently expire after 7 days ... so we cannot be sure they will be there when we need them
- doesn't the GH-API require API tokens for downloading the artifacts? It did last time I tried. That would require somehow specifying the token in tenv - probably adding it to the CI ... possible, but extra work that can be avoided

grdddj:
Argument against CI getting slow:
- there are currently around 90 jobs, each of them takes on average 3-5 minutes ... 20 seconds job is like 1/1000 of the time, not talking about it being not-CPU-intensive, just sending files

Argument against data.trezor.io being bloated by data:
- for each branch, we will store around 50 MB of emulators data ... there will unlikely be thousands of active firmware branches, and we can always delete them ... in comparison to `UI` artifacts, this is very small amount of data

.github/workflows/core.yml
grdddj: Done in the force-push

mmilata (approved):
/shrug

Artifact expiration can be increased to 90 days btw.

vdovhanych (approved):

PR #4359, opened 2024-11-18
Simplify asserts on translations

matejcik:
with this change, instead of having to use a cumbersome custom `assert_in` `assert_in_multiple` etc. and mucking around with multiple possible translations
you instead use the normal assert capabilities, and say things like `assert TR.some__translation_key in debug.text_content()`, which any Python reader can decipher.

also pytest will point you directly to the failing assertion instead of showing you a problem in the function "assert_equals"

before:
image

after:
image

as a bonus, you're actually checking against the currently selected language, which caught a small bug in a couple tests, where language was not re-set after a wipe

---

* for templates, use `TR.regexp("translation__key").match(...)`
* if you don't know the literal TR key, e.g., caller supplies it as a string, you can use `TR.translate(key)` (or, you know, `getattr(TR, key)`, but the former is preferable)
* `assert_in_multiple` is now significantly wordier, because the operation "in multiple" does not exist in Python. personally I prefer this for clarity, but if it's a big problem we could re-introduce a helper for this situation?


tests/translations.py
```diff
@@ -66,110 +70,57 @@ def set_language(client: Client, lang: str):
...
+    @property
+    def translations(self) -> dict[str, str]:
+        return self.lang_json["translations"]
+
+    def _translate_raw(self, key: str) -> str:
+        tr = self.translations.get(key)
+        if tr is not None:
+            return tr
+        if self.lang != "en":
+            return TRANSLATIONS["en"]._translate_raw(key)
```
obrusvit, line 101: This branch returns `"en"` version in case that the `key` was not found in the set language. 

How about printing a warning here? Something like `"Warning: {key} not found in {self.lang}"`

obrusvit (approved):
Nice improvement.

I tried to make LSP somehow search for these keys on auto-completion so that you would get hints while writing tests but I had no success.

Tried:
- using `__setattr__` in `__init__`
- overriding `__dir__` 

The older approach didn't provide any auto-completion support neither though.

matejcik:
> I tried to make LSP somehow search for these keys on auto-completion so that you would get hints while writing tests but I had no success.

we'd probably have to generate a file that has all the available arguments, similar to `trezortranslate_keys.pyi`

tests/translations.py
matejcik: like this? f9218ae59ec5

PR #4363, opened 2024-11-19
Use separate implementation of secret.c for unix

onvej-sl:


cepetr:
In the emulator's `main()`, there's the following code using `secret.c`:

```c
#ifdef USE_OPTIGA
      case 'l':
        // write bootloader-lock secret
        secret_write_header();
        break;
#endif
```

This code would no longer work after this change. Do we need it? Perhaps the question should be: How do we want to emulate `secret.c`, and what do we want to test with it?

@TychoVrahe

matejcik:
> Having means to test locked bootloader behavior is useful, the separate unix implementation should still support that, even if it means just setting some variable, and then looking on it when evaluating secret presence.

this needs to be persistent across restarts of `bootloader_emu` too, otherwise it's not too useful for testing.

persisting over to the firmware emulator is also helpful (it lets us test what firmware sees), but this part is solvable by setting some flag or something

onvej-sl:
> This code would no longer work after this change.

I missed that.

So would it be okay if I copied the content of `stm32f4/secret.c` to `unix/secret.c` instead of having symlink?

To explain the motivation: In https://github.com/trezor/trezor-firmware/pull/4172#issuecomment-2485941671, I suggest how to integrate the tropic emulator. Similarly to optiga, tropic has a paring secret. In the context of emulator, the paring secret is a fixed private key. My suggestion is to introduce a function `secret_tropic_get_privkey` in `core/embed/sec/secret/unix/secret.c` that will return the secret.

matejcik:
> My suggestion is to introduce a function `secret_tropic_get_privkey` in `core/embed/sec/secret/unix/secret.c` that will return the secret.

why does it need to live in `secret.c` as opposed to e.g. `unix/tropic_secret.c` specific to emulator?

onvej-sl:
> why does it need to live in secret.c as opposed to e.g. unix/tropic_secret.c specific to emulator?

It doesn't have to but it seems consistent to me. Once we have a physical device with the tropic chip, we want to treat the tropic secret the same as the optiga secret.

cepetr:
> So would it be okay if I copied the content of `stm32f4/secret.c` to `unix/secret.c` instead of having symlink?

I'm not sure what's best in this particular case where we share the implementation between `stm32f4` and `unix`. This feels like an anomaly. It would be better to have a separate implementation specific to the emulator, but until we create that, I'm fine with using a symlink.

TychoVrahe:
if there is a need to store new data in secret that will behave differently on MCU vs Unix (i.e fixed value of priv key on unix, some provisioned per-device unique value on MCUs) than i believe that this is very legitimate reason to split the F4 and unix implementation into two and getting rid of the symlink.

matejcik:
sure but @onvej-sl seems to propose an unix-specific **API**? like, if the corresponding "get private key" function does not exist on a MCU, then I see no reason to add it to `secret.c` specifically

onvej-sl:
> sure but @onvej-sl seems to propose an unix-specific API?

Not exactly. I should have been more specific. The API will be the same for both unix and physical models but the "get private key" for emulator will return a fixed key, wheres the  "get private key" function for physical models will not be implemented at this time. However, once it is implemented, it will retrieve the secret similarly to how it is done with the Optiga secret.

matejcik:
oh, I misunderstood. then sure, what @TychoVrahe says is correct, and (for now) copy-pasting the `secret.c` for unix seems like the way to go

cepetr (approved):
I confirm that instead of a symlink,  we now have `unix/secret.c` as a copy of `stm32f4/secret.c`.

PR #4364, opened 2024-11-20
fix(core): fix invalid symbol for deflated bootloader

cepetr:
This PR fixes an invalid symbol for the deflated bootloader used in the kernel and the problem with rust unit test build.

1. The structure of the symbol for deflated has been changed, so it no longer contains the path of the original file.
2. Fixed the problem with linking unused symbol `display_wait_for_sync`.


cepetr:
@TychoVrahe  Please re-review, after adding 9aee17e57489b26a4a61ac7d4ebaa655b1c1909a. Thanks.

PR #4365, opened 2024-11-20
support powering up optiga on new T3T1 and T3B1 PCB revisions 

TychoVrahe:
Configures the power pin, adds the missing clk enable and reconfigures the output as OD.

[no changelog]


cepetr (approved):
utACK

PR #4366, opened 2024-11-20
add support for haptics on T3W1

TychoVrahe:
adds configurable pins and timer.

Note that on T3W1 we will be using DRV2624, so one of the effect we use is currently not working. We will need to fix that later.


cepetr (approved):
utACK

PR #4368, opened 2024-11-20
fix sizecheck on U5 models

TychoVrahe:
fix sizecheck on U5 models


mmilata (approved):

mmilata:
> is it even possible to produce files bigger than allowed? it shouldn't link, should it? iow, is the sizecheck doing anything useful?

it's been [added](https://github.com/trezor/trezor-firmware/commit/8b0336862f11be587f8b482d2f9d99759b860e92) in 2017 by @prusnak, perhaps he remembers? :sweat_smile:

PR #4373, opened 2024-11-21
shell.nix: embedded devtools fixes

mmilata:
* fix broken OpenOCD on Mac OS X
  * also update to openocd-cubeide-v1.13.0
* backports nixos/nixpkgs#229537
  * as this causes rebuild (from binaries) of arm-gcc-embedded we only do it with `--args devTools true` to avoid CI slowdown (1-3 minutes per workflow)

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -70,19 +70,15 @@ let
     # to use official binary, remove rustfmt from buildInputs and add it to extensions:
     extensions = [ "rust-src" "clippy" "rustfmt" ];
   };
-  openocd-stm = (nixpkgs.openocd.overrideAttrs (oldAttrs: {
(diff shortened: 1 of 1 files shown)


cepetr (approved):
I've tested it and it looks like OpenOCD as well as GDB are now working on Mac. Thanks.

cepetr:
> with these new changes, it seems the build (e.g. `make build_firmware`) is not working
> 
> ```
> error: tool 'gcc' not found
> error: tool 'g++' not found
> ```
> 
> It might be Mac specific (I've not tested it on linux yet)

I tested it on the wrong branch by mistake. Everything is okay, as I stated before.

cepetr (approved):

Issue #4253, opened 2024-10-10
Kernel fails to build with `PYOPT=0`


matejcik:
In the following code, `printf` is called when `PYOPT` is set to 0. But printf is not available in kernel context. And including `<stdio.h>` does not seem to be the way to go :woman_shrugging: 
https://github.com/trezor/trezor-firmware/blob/matejcik/global-layout-only3/core/embed/kernel/main.c#L51-L69

This does not affect current builds because `PYOPT` is never 0 in kernel, but it blocks #4234 .

The printf statements are supposed to output to the USB VCP on real device (and to console in emulator, but that's irrelevant because emulator does not use optiga)

cepetr:
Yes, that seems to be a system problem. Currently, the kernel itself is not aware of `usb_vcp`, as it is now initialized by the unprivileged app via a syscall. We need to move the USB initialization to the kernel and make `usb_vcp` available to the kernel as well. The unprivileged app would then call the print syscall instead of directly accessing the `usb_vcp` interface.

PR #4375, opened 2024-11-22
Improve HW testing for T3T1

### LINKED ISSUE #3915: Device tests don't work against T3T1 hardware
Get stuck on the first tap-to-confirm screen. Reproducible on emulator by leaving animations enabled.

matejcik:
* Fixes #3915 by disabling animations for debug builds of firmware. This will also speed up HW tests on other models.
* Fixes the `DISABLE_OPTIGA` argument that got broken in the kernel/coreapp split
* introduces "storage insecure test mode" that cuts down the number of PIN PBKDF2 iterations, and does not erase unused storage banks when wiping. This shaves off ~2 seconds **per testcase**, which should save us ~1500 s on the overall run

animations are disabled on PYOPT=0 builds by default, but you can explicitly enable them by providing `TREZOR_DISABLE_ANIMATION=0` (matching TREZOR_DISABLE_ANIMATION env var for the emulator)

STORAGE_INSECURE_TESTING_MODE forces PYOPT=0 and DISABLE_OPTIGA=1

@andrewkozlik please look over the STORAGE_INSECURE_TESTING_MODE changes, in terms of whether you're ok with this level of defence against setting that option by mistake. also, if you have other suggestions on how to speed up storage?


core/SConscript.firmware
```diff
@@ -872,6 +884,16 @@ elif 'STM32U5G9xx' in CPPDEFINES_HAL or 'STM32U585xx' in CPPDEFINES_HAL:
 else:
     raise Exception("Unknown MCU")
 
+if STORAGE_INSECURE_TESTING_MODE:
+    INSECURE_TESTING_MODE_STR = """
+#########################################################
+#                                                       #
+#   STORAGE_INSECURE_TESTING_MODE enabled, DO NOT USE   #
+#                                                       #
+#########################################################
```
mmilata, line 893: Looks like this on my system:
```
#########################################################
# #
# STORAGE_INSECURE_TESTING_MODE enabled, DO NOT USE #
# #
#########################################################
```
Maybe just use more `#` instead of spaces?

mmilata (approved):
LGTM from my side.

I wonder if we should remove `--disable-animation/--enable-animation` from emu.py to make this potentially less confusing?

core/SConscript.firmware
andrewkozlik: Shouldn't the next line be something like `action_bin.append('$ECHO $INSECURE_TESTING_MODE_STR')`?

core/SConscript.kernel
```diff
@@ -24,7 +31,7 @@ FEATURE_FLAGS = {
 }
 
 FEATURES_WANTED = ["input", "sd_card", "rgb_led", "dma2d", "consumption_mask", "usb" ,"optiga", "haptic"]
-if DISABLE_OPTIGA and PYOPT == '0':
+if DISABLE_OPTIGA and not PRODUCTION:
     FEATURES_WANTED.remove("optiga")
```
andrewkozlik, line 35: 1. The corresponding change was not made in SConscript.firmware. BTW, it sucks that there is so much code duplication between SConscript.firmware and SConscript.kernel. There must be a clean way to get rid of that.
2. Similar to line 20 where we handle `if STORAGE_INSECURE_TESTING_MODE and PRODUCTION`, wouldn't it be more appropriate to raise an error `if DISABLE_OPTIGA and PRODUCTION`, rather than silently ignore `DISABLE_OPTIGA` in production builds?
```suggestion
if DISABLE_OPTIGA:
    if PRODUCTION:
        raise RuntimeError("DISABLE_OPTIGA cannot be used in production")
    FEATURES_WANTED.remove("optiga")
```

core/SConscript.firmware
matejcik: > Shouldn't the next line be something like `action_bin.append('$ECHO $INSECURE_TESTING_MODE_STR')`?

you would think so but apparently not :) then you'd get the text in the output twice, misformatted exactly the same way, once prepended with "echo"

(`#` is a shell comment so the lines above do execute cleanly in shell. i tried a couple things with `echo` and quoting, for whatever reason I didn't stumble on the correct combination and didn't really want to spend more time on it, and i'll just remove the misformatted lines)

core/SConscript.kernel
matejcik: > 1. The corresponding change was not made in SConscript.firmware

that's because scons.firmware gets the `PYOPT` as a debug indicator, but it's fixed to 1 in kernel because of #4253. otherwise firmware's spelling is preferred -- not just "not production" but "yes debug" should be the condition for disabling optiga.

i can convert both sides to the runtime-error version though

andrewkozlik (commented):
STORAGE_INSECURE_TESTING_MODE changes LGTM.

> if you have other suggestions on how to speed up storage?

Nothing else comes to mind, except for possibly skipping the `norcow_wipe()` in `norcow_init()` when no active sector is found. But I am not sure whether it would have any impact, because `norcow_init()` is called by `storage_init()`, which is invoked in Python by `config.init()`, which is called when `core/src/boot.py` is loaded. Now I am not sure whether during device tests the module is loaded once when the device boots up or once for each device test. I assume the former, in which case it makes no difference.

Another thing I was wondering about is improving wear leveling of the flash memory, since I suspect that we tend to always pick `norcow_active_sector` to be 0 (unverified). So either choose it randomly, or switch after every wipe.

core/SConscript.firmware
matejcik: 1728f6411f85

core/SConscript.kernel
matejcik: 7e9a1f458766

matejcik:
> Another thing I was wondering about is improving wear leveling of the flash memory, since I suspect that we tend to always pick `norcow_active_sector` to be 0 (unverified). So either choose it randomly, or switch after every wipe.

i'll spin out a separate issue but i kind of suspect that it's cheaper in the long term to throw out the testing Trezor from time to time than to spend developer time on this :woman_shrugging:

PR #4234, opened 2024-10-01
Scons quality-of-life improvements

matejcik:
* `$(SCONS)` in makefile now includes all the TREZOR_FOO_BAR=1 specifiers so you don't have to copy them to 20 places when adding a new one
* `CPPDEFINES` in Sconscripts no longer need to be shell-quoted ... (but it's still a hack, read the commit description to learn more about limitations)

PR #4376, opened 2024-11-22
fix cmakelists generation

TychoVrahe:
Latest changes from https://github.com/trezor/trezor-firmware/pull/4234 broke compilation with `CMAKELISTS=1`, so this adds some needed escaping.

### DIFF
--- a/core/site_scons/tools.py
+++ b/core/site_scons/tools.py
@@ -52,7 +52,8 @@ def get_defs_for_cmake(defs: list[str | tuple[str, str]]) -> list[str]:
     result: list[str] = []
     for d in defs:
         if type(d) is tuple:
-            result.append(d[0] + "=" + d[1])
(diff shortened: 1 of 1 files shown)


core/site_scons/tools.py
```diff
@@ -52,7 +52,8 @@ def get_defs_for_cmake(defs: list[str | tuple[str, str]]) -> list[str]:
     result: list[str] = []
     for d in defs:
         if type(d) is tuple:
-            result.append(d[0] + "=" + d[1])
+            val = d[1].replace('"', '\\"').replace("(", "\\(").replace(")", "\\)")
+            result.append(d[0] + '="' + val + '"')
```
prusnak, line 56:
```suggestion
            result.append(f'{d[0]}="{val}"')
```

TychoVrahe: fixed: https://github.com/trezor/trezor-firmware/commit/92b67705b9825091decaa9896173e4f080195f3c

PR #4380, opened 2024-11-25
Fix error output in emulator

matejcik:
also ~set default~ force-set PYOPT=0 for `make build_unix` which is the previous behavior before #4234 removed it


cepetr (approved):
utACK

PR #4381, opened 2024-11-25
fix(core): use the filtered event for subsequent processing

matejcik:
That way we won't pass raw touch events to underlying component if that touch is already part of a swipe.

Fixes a crash when you would swipe right over a Cancel button in menu.


obrusvit (approved):
Tested with emu.

The crash when swiping right to go back from vertical menu performed over one of the buttons is fixed.

Issue #4387, opened 2024-11-26
Solana Associated Token Account Program with data [0]


martykan:
**Describe the bug**
When calling the Solana instruction for Create in Associated Token Account Program, FW expects `data: []`, eg. `instruction_id = None` in terms of `instructions.py`.

However it should also accept `data: [0]` (eg. `instruction_id = 0`), this is what the official Solana program JS library sends. 
Currently this is seen as an unknown instruction.

**Firmware version and revision**
2.8.3

**Additional context**
A quick way to fix this might be duplicating the instruction in `defs/solana/programs.json`, once with `id: null` and a new one with `id: 0`.

matejcik:
oh. oh Solana. never change :black_heart: 
(by which I mean, totally, do change, change as soon as you can, this is :headdesk:)

we specifically talked about this, and asked: so is instruction 0 different from empty instruction?
and the answer seems to be IT FUCKING DEPENDS everyone is just doing what they feel like at the moment

PR #4389, opened 2024-11-26
hotfix(core): duplicate Solana ATAP Create instruction info

matejcik:
so that empty data and 0x00 data do the same thing, before we figure out whether this is an always thing, or a sometimes thing, and implement a little saner solution

hotfixes #4387 before we have an actual good fix


matejcik:
this never worked as a hot-fix and it turns out the underlying issue may be invalid

PR #4390, opened 2024-11-26
fix option bytes settings for U5G models

TychoVrahe:
Adjusted option bytes for larger boardloader area, pull it automatically from flash layout settings.


cepetr (approved):
utACK

PR #4395, opened 2024-11-27
streamline U5 RAM layout

TychoVrahe:
This PR attempts to simplify RAM layout settings on U5 models. 

It doesn't handle the RAM clearing mechanism, which could be a next step.

TODO:
- aux ram -> aux1_ram
- coreapp ram -> aux2_ram
- kernel_fb -> fb1_ram
- kernel_sram -> DMABUF_RAM
- kernel_stack -> MAIN_STACK
- kernel_ram -> MAIN_RAM -> ifdef na velikost stacku pro bootloader. 
- sjednotit v likner  sciptech


core/embed/sys/linker/stm32u58/kernel.ld
```diff
@@ -4,22 +4,23 @@ ENTRY(reset_handler)
...
-  SRAM6  (wal) : ORIGIN = MCU_SRAM6, LENGTH =  0K  /* SRAM6 is not available */
-  SRAM4  (wal) : ORIGIN = MCU_SRAM4, LENGTH =  MCU_SRAM4_SIZE
+
+  MAIN_RAM (wal) : ORIGIN = MAIN_RAM_START, LENGTH = MAIN_RAM_SIZE
+  BOOT_ARGS (wal) : ORIGIN = BOOTARGS_START, LENGTH = BOOTARGS_SIZE
+  KERNEL_URAM (wal) : ORIGIN = KERNEL_URAM_START, LENGTH = KERNEL_URAM_SIZE
+  FB1_RAM (wal) : ORIGIN = FB1_RAM_START, LENGTH = FB1_RAM_SIZE
+  FB2_RAM (wal) : ORIGIN = FB2_RAM_START, LENGTH = FB2_RAM_SIZE
+
+  COREAPP_RAM (wal) : ORIGIN = AUX_RAM_START, LENGTH = AUX_RAM_SIZE
```
cepetr, line 14: What about removing `COREAPP_RAM` and `COREAPP_RAM2`, since we don't use these two memory regions (we don't place anything into) in the kernel?

core/embed/models/T3W1/model_T3W1.h
```diff
@@ -91,10 +91,25 @@
...
+#define MAIN_RAM_SIZE (64 * 1024 - 512)
+
+#define KERNEL_URAM_START 0x300CFE00
+#define KERNEL_URAM_SIZE 512
+
+#define FB2_RAM_START 0x300D0000
+#define FB2_RAM_SIZE (832 * 1024)
+
+#define AUX_RAM_START 0x301A0000
+#define AUX_RAM_SIZE (832 * 1024)
```
cepetr, line 110: Why are FB1 and FB2 not the same size? How about adding 64KB from FB2 to AUX_RAM?

cepetr (changes requested):
I thoroughly reviewed all the changes and didn't find anything wrong, just a few minor points for further discussion.

core/embed/sys/linker/stm32u58/kernel.ld
TychoVrahe: i agree that this is more in line with the new approach, so refactored further here https://github.com/trezor/trezor-firmware/commit/c4d3ed73f4f5698cd31b8ae8c0161dd519190a4a 

we used constants from linker script for initializing the applet, so this required some changes, also in F4 models which i originally didn't want to touch.

TychoVrahe:
In https://github.com/trezor/trezor-firmware/pull/4395/commits/70254382e409b3a932693f9d8600150bbbd46496, i also do this for F4 models. It can't be done as cleanly as on U5, as we have more limited RAM available and bootloader/boardloader need bigger RAM then the kernel uses, so there is an overlap between KERNEL RAM defs and MAIN_RAM, which is the one used in boot/boardloder.

@cepetr please check this too.

core/embed/sys/mpu/stm32u5/mpu.c
```diff
@@ -161,35 +155,8 @@ extern uint32_t _codelen;
...
-#define MAIN_SRAM_SIZE SRAM2_SIZE
-#define AUX_SRAM_START SRAM1_BASE
-#define AUX_SRAM_SIZE (SRAM1_SIZE - BOOTARGS_SIZE)
-#else
-#define MAIN_SRAM_START SRAM2_BASE
-#define MAIN_SRAM_SIZE SRAM2_SIZE
-#define AUX_SRAM_START SRAM5_BASE
-#define AUX_SRAM_SIZE SRAM5_SIZE
-#endif
+#define COREAPP_RAM1_START AUX_RAM_START
```
cepetr, line 158: `COREAPP_RAM1_START` and `COREAPP_RAM1_SIZE` are no longer used.

core/embed/sys/mpu/stm32u5/mpu.c
```diff
@@ -137,16 +137,10 @@ _Static_assert(NORCOW_SECTOR_SIZE == STORAGE_2_MAXSIZE, "norcow misconfigured");
...
-extern uint8_t boot_args_start;
-#define BOOTARGS_START ((uint32_t) & boot_args_start)
-
 #ifdef KERNEL
 
 extern uint8_t _uflash_start;
 extern uint8_t _uflash_end;
-#define KERNEL_RAM_U_START (KERNEL_RAM_START + KERNEL_RAM_SIZE)
-#define KERNEL_RAM_U_SIZE KERNEL_U_RAM_SIZE
 #define KERNEL_FLASH_U_START (uint32_t) & _uflash_start
```
cepetr, line 144: We can remove `KERNEL_FLASH_U_START`/`SIZE` as well, seems it's no longer used.

core/embed/projects/kernel/main.c
```diff
@@ -204,11 +204,12 @@ void drivers_init() {
 
 // defined in linker script
 extern uint32_t _codelen;
-extern uint32_t _coreapp_clear_ram_0_start;
-extern uint32_t _coreapp_clear_ram_0_size;
-extern uint32_t _coreapp_clear_ram_1_start;
-extern uint32_t _coreapp_clear_ram_1_size;
+
 #define KERNEL_SIZE (uint32_t) & _codelen
+#ifndef AUX2_RAM_START
```
cepetr, line 209: Perhaps instead of defining a fake `AUX2_RAM_START`, we could use the same approach as with `FIRMWARE_P2_START`:

```
#ifdef AUX2_RAM_START
      .data2.start = (uint32_t)AUX2_RAM_START,
      .data2.size = (uint32_t)AUX2_RAM_SIZE,
#endif
```

cepetr (changes requested):
Finally, I found a few nits; otherwise, it looks very good.

core/embed/sys/mpu/stm32u5/mpu.c
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/5e57d8d4626a6ad0dbd44ae630dd040781c64f13

core/embed/projects/kernel/main.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/5e57d8d4626a6ad0dbd44ae630dd040781c64f13

core/embed/sys/mpu/stm32u5/mpu.c
TychoVrahe: no these are still used in KERNEL_FLASH_START and COREAPP_FLASH_START calculations below - since we decided to include the U_FLASH into the coreapp mpu region along with the empty space between kernel and coreapp to fix the fw hashing.

PR #4396, opened 2024-11-27
Multisig fix in legacy

onvej-sl:
This pull request backports https://github.com/trezor/trezor-firmware/pull/4351 to legacy.


legacy/firmware/fsm_msg_coin.h
```diff
@@ -296,18 +296,44 @@ void fsm_msgGetAddress(const GetAddress *msg) {
...
+        // be exploited by an attacker who has compromised the user's computer.
+        // The attacker could randomize the address indices and then demand a
+        // ransom from the user to reveal the paths. To prevent this, we require
+        // that all xpubs use the same derivation path.
+        if (config_getSafetyCheckLevel() == SafetyCheckLevel_Strict) {
+          fsm_sendFailure(
+              FailureType_Failure_DataError,
+              _("Using different paths for different xpubs is not allowed"
+
+                ));
```
mmilata, line 317: Nit
```suggestion
              _("Using different paths for different xpubs is not allowed"));
```

legacy/firmware/fsm_msg_coin.h
```diff
@@ -296,18 +296,44 @@ void fsm_msgGetAddress(const GetAddress *msg) {
...
+          layoutHome();
+          return;
+        }
+        fsm_layoutDifferentPathsWarning();
```
mmilata, line 321: User pressing Cancel is not handled here.

onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/4396/commits/b899ff4195ec446cb7e2ca550057ebbba4d705de.

mmilata (commented):
* fixtures are missing (`tests/update_fixtures.py ci` should work)
* I'm getting some failures on hardware that I'm not seeing in emulator, e.g.:
   ```
   FAILED tests/device_tests/bitcoin/test_multisig_change.py::test_sorted_multisig_change_match_first - AssertionError: Expected responses:
   FAILED tests/device_tests/bitcoin/test_multisig_change.py::test_multisig_change_match_second - AssertionError: Expected responses:
   FAILED tests/device_tests/bitcoin/test_multisig_change.py::test_multisig_change_match_first - AssertionError: Expected responses:
   FAILED tests/device_tests/bitcoin/test_bcash.py::test_send_bch_multisig_change - AssertionError: Expected responses:
   ```
   Not 100% sure this is because of this PR, would appreciate someone else checking. Needs #4535 in order to work on HW.

onvej-sl:
> I'm getting some failures on hardware that I'm not seeing in emulator, e.g.:

I hopefully fixed it in https://github.com/trezor/trezor-firmware/pull/4396/commits/bc0ecfe61b28591067c9e2930db1fee83615e9ca.

onvej-sl:
> > I'm getting some failures on hardware that I'm not seeing in emulator, e.g.:
> 
> I hopefully fixed it in [bc0ecfe](https://github.com/trezor/trezor-firmware/commit/bc0ecfe61b28591067c9e2930db1fee83615e9ca).

My trezor one device is somewhat broken so I was only able to run `tests/device_tests/bitcoin/test_multisig_change.py` and `tests/device_tests/bitcoin/test_bcash.py` and both ran successfully.

mmilata:
Can confirm that all multisig tests now pass on my T1. But I don't understand where the problem was - shouldn't `multisig->pubkeys_order` be all zeros when not present in the message? Is this related to `-fshort-enums`?

mmilata (approved):

onvej-sl:
To be honest, I'm not certain. However, based on how the code is carefully written (see, for example, [this](https://github.com/trezor/trezor-firmware/blob/bedff6e8319dae192140212fdea4e06148b855ec/legacy/firmware/signing.c#L3320)), I thought that `property` is uninitialized if `has_property` is `false`.

prusnak:
I think the property is set to zeroes if unset by wire, but still I prefer to have the code explicit rather than implicit to be on the safe side once this behaviour changes in future versions of nanopb or something like that.

> Is this related to `-fshort-enums`?

Also this might be the culprit in some cases like we've seen recently. So explicit code is better.

mmilata:
Yep I think it's `-fshort-enums`, on hardware `sizeof(MultisigPubkeysOrder) == 1` so `sha256_Update((const uint8_t *)&(multisig->pubkeys_order), sizeof(uint32_t))` reads also the next member in the struct, and it's not 0. Emulator doesn't use that option so it works there.

legacy/firmware/crypto.c
```diff
@@ -403,21 +457,22 @@ int cryptoMultisigFingerprint(const MultisigRedeemScriptType *multisig,
...
+    // to depend on the order of the pubnodes, so we sort the pubnodes before
+    // hashing.
+    qsort(pubnodes, n, sizeof(HDNodeType *), comparePubnodesLexicographically);
   }
-  // hash sorted nodes
+
   SHA256_CTX ctx = {0};
   sha256_Init(&ctx);
   sha256_Update(&ctx, (const uint8_t *)&(multisig->m), sizeof(uint32_t));
+  sha256_Update(&ctx, (const uint8_t *)&(pubkeys_order), sizeof(uint32_t));
```
matejcik, line 475: ...that means that this code is wrong, right?
what we want is, pseudocode:
```
MultisigPubkeysOrder pubkeys_order = ...
sha256_Update(multisig->m, sizeof(multisig->m);
sha256_Update(pubkeys_order, sizeof(pubkeys_order));
...
```
etc for this whole block.

onvej-sl: I believe the code is correct now. The only difference from your pseudocode is that I cast `MultisigPubkeysOrder` to `uint32_t`.

onvej-sl:
I tested it and can confirm that the size of `MultisigPubkeysOrder` is `4` in the emulator, while it is `1` on the hardware.

legacy/firmware/crypto.c
matejcik: my point is that we're magically relying on all the things being u32-sized, with no actual proof in the code (which was the source of the previous bug)

the code is correct now, but could break again in theory

onvej-sl: I have no problem changing `uint32_t` to `MultisigPubkeysOrder`. However, I don't still understand what you mean by "but could break again in theory". My code doesn't rely on `MultisigPubkeysOrder` to be `uint32_t`.

matejcik: https://github.com/trezor/trezor-firmware/issues/4547

onvej-sl: Okay, I understand now. I initially thought you were only referring to `pubkeys_order`, but you are also talking about `m`.

PR #4399, opened 2024-11-28
fix haptic driver with reset pin initialization

TychoVrahe:
On T3W1, we use reset pin for haptic driver. Turns out, we need to bring the IC out of reset before attempting communication with it....


cepetr (approved):
utACK

PR #4404, opened 2024-11-28
Introduce show_danger

ibz:
This is a unified interface for `flow_warning_hi_prio`, which was available only on Mercury before.


core/embed/rust/src/ui/model_mercury/flow/danger.rs
```diff
@@ -22,13 +22,13 @@ use super::super::{
 };
 
 #[derive(Copy, Clone, PartialEq, Eq)]
-pub enum WarningHiPrio {
+pub enum Danger {
```
obrusvit, line 25: I's slightly prefer naming this `ShowDanger` and also the file should be `show_danger.rs`. 
For consistency.

core/src/trezor/ui/layouts/mercury/__init__.py
```diff
@@ -316,6 +314,29 @@ def show_warning(
     )
 
 
+def show_danger(
+    br_name: str,
+    content: str,
+    content_short: str | None = None,
```
obrusvit, line 320: I don't think the distinction between `content`/`content_short` is a good idea. I think we can keep the `confirm_ethereum_unknown_contract_warning` function which calls the `show_danger` with the appropriate string.

core/src/trezor/ui/layouts/tt/__init__.py
```diff
@@ -364,6 +364,25 @@ def show_warning(
...
+    br_name: str,
+    content: str,
+    content_short: str | None = None,
+    value: str | None = None,
+    title: str | None = None,
+    verb_cancel: str | None = None,
+    br_code: ButtonRequestType = ButtonRequestType.Warning,
+) -> Awaitable[None]:
+    if content_short is None:
+        content_short = content
```
obrusvit, line 377: apart from the previous comment, why showing `content_short` on `TT`?

obrusvit (commented):
A few points for discussion.

core/src/trezor/ui/layouts/tt/__init__.py
ibz: Because that's what Figma says ;)

core/src/trezor/ui/layouts/mercury/__init__.py
ibz: Fixed!

obrusvit (approved):

Hannsek:
Why did the exclamation mark disappear from the "Warning" header.

Why did the text change from "cancel and exit" to "cancel"?

ibz:
> Why did the exclamation mark disappear from the "Warning" header.

For consistency. Now all of these warnings look the same. If you think they should all have the exclamation mark, I can add that. But the ETH contract design did not have it.

> Why did the text change from "cancel and exit" to "cancel"?

Where?

Hannsek:
Strange, I was looking at the diff this morning and it was like that. Also "important" had an exclamation mark, but "Warning" did not. Now everything is fine. Thank you.

PR #4405, opened 2024-11-29
fix BOOTLOADER VERSION command in prodtest

TychoVrahe:
Fixes reading the version. Also some discrepancies in readme are fixed.


cepetr (approved):
utACK

PR #4408, opened 2024-11-29
ci: work around rate limit, no comment for legacy

mmilata:
Let's see if not posting the comment right after the job starts helps.

Also disable this for legacy firmware since two comments are twice as annoying and the changes are rare anyway. You can still see the table under the artifacts list.


obrusvit (approved):
ACK

PR #4411, opened 2024-11-29
chore(core): add missing parameter to set_brightness function

bieleluk:
This PR:
1. adds the missing brightness value parameter in `set_brightness` function
2. If the value has been passed, the brightness is set to this value
3. Parametrizes existing tests to test different scenarios (screenshots do not show the brightness correctly, but locally it works well)
4. [delizia] Fixes inconsistent brightness in `set_brightness` flow by making the flow have 1 screen only.

### DIFF
--- a/core/embed/rust/src/ui/layout_bolt/ui_firmware.rs
+++ b/core/embed/rust/src/ui/layout_bolt/ui_firmware.rs
@@ -4,6 +4,7 @@ use crate::{
     error::{value_error, Error},
     io::BinaryData,
     micropython::{buffer::StrBuffer, gc::Gc, iter::IterBuf, list::List, obj::Obj, util},
+    storage,
     strutil::TString,
     translations::TR,
     ui::{
@@ -774,12 +775,19 @@ impl FirmwareUI for UIBolt {
     }
 
     fn set_brightness(current_brightness: Option<u8>) -> Result<impl LayoutMaybeTrace, Error> {
+        let current = match current_brightness {
+            Some(value) => {
+                // Set the brightness immediately so it is applied in the `_first_paint` UI
+                // layout function
+                unwrap!(storage::set_brightness(value));
+                value
+            }
+            None => theme::backlight::get_backlight_normal(),
+        };
         let layout = RootComponent::new(Frame::centered(
             theme::label_title(),
             TR::brightness__title.into(),
-            SetBrightnessDialog::new(
-                current_brightness.unwrap_or(theme::backlight::get_backlight_normal()),
-            ),
+            SetBrightnessDialog::new(current),
         ));
 
         Ok(layout)
--- a/core/embed/rust/src/ui/layout_delizia/flow/set_brightness.rs
+++ b/core/embed/rust/src/ui/layout_delizia/flow/set_brightness.rs
@@ -18,7 +18,7 @@ use crate::{
 use super::super::{
     component::{
         number_input_slider::{NumberInputSliderDialog, NumberInputSliderDialogMsg},
-        Footer, Frame, PromptMsg, PromptScreen, StatusScreen, SwipeContent,
+        Footer, Frame,
     },
     theme,
 };
@@ -26,8 +26,6 @@ use super::super::{
 #[derive(Copy, Clone, PartialEq, Eq, ToPrimitive)]
 pub enum SetBrightness {
     Slider,
-    Confirm,
-    Confirmed,
 }
 
 impl FlowController for SetBrightness {
@@ -38,19 +36,17 @@ impl FlowController for SetBrightness {
 
     fn handle_swipe(&'static self, direction: Direction) -> Decision {
         match (self, direction) {
-            (Self::Slider, Direction::Up) => Self::Confirm.swipe(direction),
-            (Self::Confirm, Direction::Down) => Self::Slider.swipe(direction),
-            (Self::Confirmed, Direction::Up) => self.return_msg(FlowMsg::Confirmed),
+            (Self::Slider, Direction::Up) => {
+                unwrap!(storage::set_brightness(BRIGHTNESS.load(Ordering::Relaxed)));
+                self.return_msg(FlowMsg::Confirmed)
+            }
             _ => self.do_nothing(),
         }
     }
 
     fn handle_event(&'static self, msg: FlowMsg) -> Decision {
         match (self, msg) {
-            (Self::Slider, FlowMsg::Info) => self.return_msg(FlowMsg::Info),
-            (Self::Confirm, FlowMsg::Confirmed) => Self::Confirmed.swipe_up(),
-            (Self::Confirm, FlowMsg::Info) => self.return_msg(FlowMsg::Info),
-            (Self::Confirmed, FlowMsg::Confirmed) => self.return_msg(FlowMsg::Confirmed),
+            (Self::Slider, FlowMsg::Cancelled) => self.return_msg(FlowMsg::Cancelled),
             _ => self.do_nothing(),
         }
     }
@@ -67,7 +63,7 @@ fn footer_update_fn(
         footer.update_instruction(ctx, TR::instructions__swipe_horizontally);
         footer.update_description(ctx, TR::setting__adjust);
     } else {
-        footer.update_instruction(ctx, TR::instructions__tap_to_continue);
+        footer.update_instruction(ctx, TR::instructions__tap_to_confirm);
         footer.update_description(ctx, TR::setting__apply);
     }
 }
@@ -82,8 +78,7 @@ pub fn new_set_brightness(brightness: u8) -> Result<SwipeFlow, Error> {
         ),
     )
     .with_subtitle(TR::homescreen__settings_subtitle.into())
-    .with_menu_button()
-    .with_external_menu()
+    .with_cancel_button()
     .with_swipe(Direction::Up, SwipeSettings::default())
     .with_footer(
         TR::instructions__swipe_horizontally.into(),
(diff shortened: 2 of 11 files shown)


matejcik:
ack on code, please fix ui tests

obrusvit (commented):
I actually don't think this fixes the issue, does it?

If you pass the value to the `set_brightness` dialog, it merely starts the UI with the value. But the first real change happens on `NumberInputSliderDialogMsg::Changed` which happens only if you change the number.

bieleluk:
You are right, @obrusvit 
My PR just adds a missing function parameter but does not make any functional changes.
If it is not sufficient, I can implement something like the following:
- If the parameter is a valid number, then it

bieleluk:
@obrusvit 
This commit addresses your concern https://github.com/trezor/trezor-firmware/pull/4411/commits/0537f0c04a7a6969e8a44ef953ba4f29be333096

bieleluk:
Rebasing due to conflicts from the latest changes

matejcik:
when you have a little time @bieleluk please rebase this and fix the failing CI

obrusvit:
Doe this fix your recent issue @bieleluk ?

bieleluk:
> Doe this fix your recent issue @bieleluk ?

@obrusvit , [0dc258e](https://github.com/trezor/trezor-firmware/pull/4411/commits/0dc258e8fa0182a2cd02504bae546a145e5b8251) does make the difference. The `_first_print` UI layout function always takes the `brightness` value from the storage and applies it. Calling the `backlight` function directly didn't have any effect. The trick was to set the `brightness` value in the storage so it'll be ready for the `_first_print` function. The small issue is that the test screenshots do not display this correctly, but when you run it locally, it works as expected.

tests/device_tests/test_msg_applysettings.py
```diff
@@ -474,11 +478,37 @@ def test_label_too_long(client: Client):
         device.apply_settings(client, label="A" * 33)
 
 
-@pytest.mark.models(skip=["legacy", "safe3", "eckhart"])
+U8_MIN = 0
+U8_MAX = (1 << 8) - 1  # 255
```
romanz: nit:
```suggestion
U8_MAX = 0xFF  # 255
```

tests/device_tests/test_msg_applysettings.py
```diff
@@ -474,11 +478,37 @@ def test_label_too_long(client: Client):
...
+        (U8_MAX, False),
+        (U8_MAX + 1, True),
+    ],
+)
+def test_set_brightness(client: Client, value: int | None, should_raise: bool):
+    with client:
+        IF = InputFlowSetBrightness(client)
+        client.set_input_flow(IF.get())
+
+        if should_raise:
```
romanz: nit: can be simplified by https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext:
```python
with pytest.raises(exceptions.TrezorFailure) if should_raise else contextlib.nullcontext():
    device.set_brightness(client, value)
```

tests/device_tests/test_msg_applysettings.py
```diff
@@ -474,11 +478,37 @@ def test_label_too_long(client: Client):
...
+
+@pytest.mark.models(skip=["legacy", "safe3"])
 @pytest.mark.setup_client(pin=None)
-def test_set_brightness(client: Client):
-    device.set_brightness(client, None)
+@pytest.mark.parametrize(
+    "value,should_raise",
+    [
+        (None, False),
+        (U8_MIN, False),
```
romanz, line 496: will it fail for `-1`?

romanz:
Tried to lower the brightness on the emulator, but it seems that its went back to full brightness during the confirmation screen (eventually it worked, but I am not sure about the intermediate brightness "jump"):

[Screencast from 2025-07-10 22-39-09.webm](https://github.com/user-attachments/assets/9432c15d-e2f8-4763-9027-d7778f7bc290)

tests/device_tests/test_msg_applysettings.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4411/commits/9ac5e8c0087351005c909e17837a4a327436e71f

tests/device_tests/test_msg_applysettings.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4411/commits/701ce986407b53f3299950361df8f8239ce4d97e

tests/device_tests/test_msg_applysettings.py
bieleluk: It would trigger `ValueError` instead of the Trezor exception. I made a separate test for that because it would be too complicated to combine with the previous one:
https://github.com/trezor/trezor-firmware/pull/4411/commits/ea6148f05e19221267c4008c89df38224ce8e0ac

bieleluk:
> Tried to lower the brightness on the emulator, but it seems that its went back to full brightness during the confirmation screen (eventually it worked, but I am not sure about the intermediate brightness "jump"):
> Screencast.from.2025-07-10.22-39-09.webm

I can confirm this bug. I also observed this behavior in the `main` branch, so it requires further investigation.

core/embed/rust/src/ui/layout_bolt/ui_firmware.rs
```diff
@@ -764,12 +765,19 @@ impl FirmwareUI for UIBolt {
     }
 
     fn set_brightness(current_brightness: Option<u8>) -> Result<impl LayoutMaybeTrace, Error> {
+        let current = match current_brightness {
+            Some(value) => {
+                // Set the brightness immediately so it is applied in the `_first_print` UI
```
romanz: In case the user cancels the dialog, should we switch back to the original brightness setting?

romanz:
```suggestion
                // Set the brightness immediately so it is applied in the `_first_paint` UI
```
(here and below)

romanz: Maybe it would be simpler to set `ui.Layout.backlight_level` for set_brightness layout?

tests/input_flows.py
```diff
@@ -3001,6 +3001,23 @@ def input_flow_eckhart(self) -> BRGeneratorType:
             self.debug.press_yes()
 
 
+class InputFlowSetBrightness(InputFlowBase):
+
+    def input_flow_bolt(self):
+        yield
+        self.debug.press_yes()
```
romanz: Can we use `return self.client.ui.default_input_flow()` for Bolt and Delizia?

core/src/trezor/ui/layouts/delizia/__init__.py
```diff
@@ -1582,16 +1582,16 @@ def confirm_firmware_update(description: str, fingerprint: str) -> Awaitable[Non
...
+    br_name = "set_brightness"
+    await raise_if_not_confirmed(
         trezorui_api.set_brightness(current=current),
-        Menu.root(cancel=TR.buttons__cancel),
-        "set_brightness",
+        br_name,
         BR_CODE_OTHER,
     )
 
+    await show_success(br_name, TR.brightness__changed_title)
```
romanz: Maybe use a "delayed confirmation" screen with `show_continue_in_app` (allowing it to be automatically closed after a few seconds, following #3666)?
CC: @Hannsek

core/embed/rust/src/ui/layout_delizia/flow/set_brightness.rs
```diff
@@ -38,19 +36,17 @@ impl FlowController for SetBrightness {
 
     fn handle_swipe(&'static self, direction: Direction) -> Decision {
         match (self, direction) {
-            (Self::Slider, Direction::Up) => Self::Confirm.swipe(direction),
-            (Self::Confirm, Direction::Down) => Self::Slider.swipe(direction),
-            (Self::Confirmed, Direction::Up) => self.return_msg(FlowMsg::Confirmed),
+            (Self::Slider, Direction::Up) => {
+                let _ = storage::set_brightness(BRIGHTNESS.load(Ordering::Relaxed));
```
romanz: Maybe use `unwrap!(...)` here as well?
(we probably don't to ignore storage errors)

core/embed/rust/src/ui/layout_bolt/ui_firmware.rs
bieleluk: Using `backlight_level` for the default value doesn't work.
The function `_first_paint` always sets the backlight to the value from the storage. The only way to do that is to call `set_brightness` function before the brightness screen is created. It is possible to implement by adding an extra parameter to the brightness dialog, but it doesn't make much sense IMO. Nevertheless, the default value cannot be set in the suit nor `trezorctl`, so it's probably pointless either way.

bieleluk: https://github.com/trezor/trezor-firmware/pull/4411/commits/c911bc1fb8132f760e373090f08b3bfdfb7e2f1b

tests/input_flows.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4411/commits/8b63958d1868ee4bb38e0f8799a31bf78883c8da

core/embed/rust/src/ui/layout_delizia/flow/set_brightness.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/4411/commits/d5d3e5ed219eb09095577f99237edeac456e2ff9

core/src/trezor/ui/layouts/delizia/__init__.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4411/commits/2c48a13ea40d48de4821f0ebb124572f5132f94c

core/embed/rust/src/ui/layout_bolt/ui_firmware.rs
romanz: Got it, thanks!

Maybe we can `storage::set_brightness(...)` at [`new_set_brightness()`](https://github.com/trezor/trezor-firmware/blob/97da97b350ba163ab885941d29638b1da6ea6ed6/core/embed/rust/src/ui/api/firmware_micropython.rs#L808) to avoid duplocating the code above between models?

bieleluk: Currently, the `firmware_micropython.rs` file doesn't implement any functionality; it only parses the parameters from MicroPython. On one hand, we would avoid duplicating the code, but on the other hand, we would break the current status quo.

romanz (approved):
Thanks!

PR #4412, opened 2024-11-30
refactor confirm_summary

obrusvit:
This is a refactor of a transaction summary confirmation dialog: `confirm_summary`. The reason is to unify the rust function signature across models. This paves the way for easier work on #3658. uPy functions were refactored accordingly. The rust functions of all three models now have these parameters:
- `amount`
- `amount_label`
- `fee`
- `fee_label`
- `title` (Optional)
- `account_items` (Optional) - KV pairs of "send from" info (e.g. account name, derivation paths, etc.)
- `extra_items` (Optional) - KV pairs of whatever other info, typically used for other fee info (e.g. fee rate, gas price of ETH, etc.) or some other coin-specific info (e.g. `TTL` for Cardano)
- `extra_title` (Optional) - title for `extra_items`
- `verb_cancel` (Optional) - cancel button verb, can be used on `TR` and `TT` to signify whether the button is arrow up (`^`), meaning a user can go back one step in uPy flow

Changes:
- [`TT`]
  - `confirm_total` replaced by `confirm_summary`
  - drawback: info `_items` are passed to rust but are actually not used there as they are used within uPy flow of `with_info`
- [`TR`] 
  - `confirm_total` and `altcoin_tx` summary were merged to `confirm_summary`. The function now handles all Optional info placed one after another.
- [`Mercury`] 
  - `flow_confirm_summary` replaced by `confirm_summary`
  - old `confirm_total` removed - it wasn't used

UI diff:
- [`TR`]
  - info screens now use `vertically_centered()` instead of manual `.newline()`, lots of negligible diffs
image
  - :information_source: not accessible when not provided
image
  - in some cases, titles in :information_source: are fixed  
image
  - in some cases we have less pagination, and in a few cases we have more pagination.


core/embed/rust/src/ui/model_mercury/flow/confirm_summary.rs
```diff
@@ -105,34 +110,42 @@ pub fn new_confirm_summary(
...
             theme::ICON_CHEVRON_RIGHT,
             TR::address_details__account_info.into(),
         );
         unwrap!(menu_items.push(MENU_ITEM_ACCOUNT_INFO));
     }
-    menu = menu.danger(
-        theme::ICON_CANCEL,
-        cancel_text.unwrap_or(TR::send__cancel_sign.into()),
-    );
+    if content_extra.is_some() && menu_items.len() < 2 {
```
ibz, line 140: I don't really like the hardcoded "2" here. Ideally it would be somehow computed based on the screen size, but a big step forward (probably good enough) would be a constant in `vertical_menu.rs` - something like `MAX_ITEMS_ON_SCREEN`.

core/embed/rust/src/ui/model_tr/layout.rs
```diff
@@ -638,131 +639,112 @@ extern "C" fn new_confirm_output_amount(n_args: usize, args: *const Obj, kwargs:
...
+            .try_into_option()?;
+        let account_items: Option<Obj> = kwargs
+            .get(Qstr::MP_QSTR_account_items)
+            .unwrap_or_else(|_| Obj::const_none())
+            .try_into_option()?;
+        let extra_items: Option<Obj> = kwargs
+            .get(Qstr::MP_QSTR_extra_items)
+            .unwrap_or_else(|_| Obj::const_none())
+            .try_into_option()?;
+        let cancel_arrow: bool = kwargs.get_or(Qstr::MP_QSTR_cancel_arrow, false)?;
```
ibz, line 664: I think that on model R it should be `verb_cancel = "^"` versus `verb_cancel = ""` to mark the "up arrow" versus the "x" icon. I've seen that used in other places. Introducing the concept of "cancel_arrow" seems wrong, not just because it's unclear what it means, but also because it is not portable across models.

ibz (commented):
Beautiful refactor!

Some comments added.

core/embed/rust/src/ui/model_tr/layout.rs
obrusvit: Both `TT` and `TR` either uses "x" or "^" as the cancel button so that you can use "^" if the summary is a part of uPy flow.

`TT` uses `cancel_arrow` now and it used it before.
`TR` used `cancel_cross` before in `altcoin_tx_summary` which was basically the opposite of what `cancel_arrow` does. So I unified it. 
`Lincoln` will probably need it as well given current designs.
`Mercury` does not need it but maybe it will if the designs change in the future based on Lincoln.

Having `verb_cancel` there might make more sense. Let me try.

core/embed/rust/src/ui/model_mercury/flow/confirm_summary.rs
obrusvit: 29cf6d3557b4c6f877c38fe39c959148df6e4e21

matejcik:
why make `fee_items` and `extra_items` separate?

obrusvit:
> why make `fee_items` and `extra_items` separate?

@matejcik `fee_items` is used for fee specific information, whereas `extra_items` serves for whatever other info which do not suit in the other 2. Some altcoins might hide some info here (e.g. "Cardano" stuff like `TTL` etc.) 
We could make "fee informations" be a part of "extra information" but separating them makes for easy setting the titles of the info dialogs ("Fee info" vs. "More info" in `TR`) and the buttons (`VerticalMenu` of `mercury`).

core/embed/rust/src/ui/model_tr/layout.rs
obrusvit: made a mistake for `TR`, fixed here: 61448aad34b21198d00d80507fc3321b923be27a

ibz (approved):

matejcik:
code ACK, waiting for device tests before approving

matejcik (approved):

PR #4420, opened 2024-12-03
ci: add build-docker.sh jobs

mmilata:
Only for scheduled and manual pipelines so as not to slow down regular PRs.

I tried all models in a single build but ran into `No space left on device`. Job per model is faster but eats more compute time because every job builds its own docker image.

You can see all fingerprints in [workflow summary](https://github.com/trezor/trezor-firmware/actions/runs/12148645944?pr=4420#summary-33877558860).

Based on #4418.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -5,6 +5,10 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '14 23 * * *'  # every day @ 23:14
+  push:
+    branches:
+      - 'release/**'
+
 
 # cancel any previous runs on the same PR
 concurrency:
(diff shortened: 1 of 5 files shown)


mmilata:
Added another commit that changes the CI settings so that `core.yml`, `legacy.yml`, `common.yml` and `prebuild.yml` are run whenever a `release/**` branch is updated.

Previously I had the idea of having a PR-centered process but the fact that [`pull_request`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) triggers workflow on a merge commit with base branch makes it hard to work with. So I've been starting the workflows manually, might just as well do it for every push.

matejcik (approved):
utACK

PR #4423, opened 2024-12-05
fix translation area access from coreapp applet

TychoVrahe:
trustozone setting changed: assets area as unprivileged.


mmilata:
:heavy_check_mark: fixes the problem on my setup

cepetr (approved):

bosomt:
Changed translations on all models with 2.8.7 fw release with success

PR #4425, opened 2024-12-06
Add test for Slip39_Single

ibz:


obrusvit (approved):
Perhaps it would be cleaner to use `InputFlowSlip39CustomBackup` instead of `InputFlowBip39` but it's fine.

PR #4427, opened 2024-12-06
Don't swipe left if there is no menu

ibz:


obrusvit (approved):
:+1: 
Please add changelog entry for 4421.

PR #4429, opened 2024-12-09
Font generation script

bieleluk:
This PR:
- corrects destination directories of generated fonts
- collects and prints uC memory footprint of each generated font


bieleluk:
This PR is not valid anymore

PR #4435, opened 2024-12-10
resolve crashes when running without display

TychoVrahe:
This PR resolve crashes when uninitialized display returns null ptr as framebuffer and firmware attempts to render into it.

We will need to run some stages of prodtest of T3W1 running without display and this enables it, but we should probably consider allowing this explicitly via some OTP flag, and implement somewhat more elegant way to shutdown when this is not allowed. however, I consider this out of scope in this PR.

### DIFF
--- a/core/embed/rust/src/trezorhal/display.rs
+++ b/core/embed/rust/src/trezorhal/display.rs
@@ -55,13 +55,17 @@ pub fn refresh() {
 }
 
 #[cfg(feature = "framebuffer")]
-pub fn get_frame_buffer() -> (&'static mut [u8], usize) {
+pub fn get_frame_buffer() -> Option<(&'static mut [u8], usize)> {
     let mut fb_info = ffi::display_fb_info_t {
         ptr: ptr::null_mut(),
         stride: 0,
     };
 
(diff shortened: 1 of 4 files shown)


cepetr:
Am I correct in understanding that T3W1 crashes in the bootloader?

The proposed solution is simple, but I suggest a slightly different approach. It's a bit complicated:

*Detecting the Display*
- If the display is not detected or present, the `display_init()` function could return `false`. Alternatively, we could introduce a dedicated `display_detect()` function to handle this specific purpose.

*Handling Uninitialized Displays*

- If the display is not initialized, we should avoid drawing on it. We might ensure this in `bootui.c`, which serves as an intermediate layer between the bootloader and the Rust UI.

This approach ensures proper handling of cases where the display is unavailable while maintaining a clear separation of responsibilities in the code. I believe we should avoid calling any `screen_xxx` functions if the display or touch interface is not initialized. Additionally, if `display_get_frame_buffer()` returns NULL while painting on screen, it should be treated as a fatal error.

TychoVrahe:
As discussed in person, lets consider this out of scope here. I created an issue for handling this later: https://github.com/trezor/trezor-firmware/issues/4440

cepetr (approved):

PR #4436, opened 2024-12-10
Introduce Nostr

ibz:
This PR adds Nostr support to the Trezor firmware, `trezorlib` and `trezorctl`.

Note: The functionality is only available in debug mode, for now!

## NIP-06

This PR uses NIP-06 to derive the keys.

❗ We are not satisfied with using NIP-06 due to the reasons explained [here](https://github.com/nostr-protocol/nips/issues/1774).

**A new NIP using another derivation path is required before we can expose this functionality to our users!**

## Firmware

The Trezor firmware accepts the `NostrGetPubkey` and `NostrSignEvent` messages, as defined in [common/protob/messages-nostr.proto](https://github.com/trezor/trezor-firmware/blob/ibz/nostr/common/protob/messages-nostr.proto).

To avoid adding complexity to the firmware, it does not parse the event JSON. The event needs to be parsed beforehand and its individual fields passed via the protobuf message. The return value contains the pubkey, event ID and signature, which need to be added back to the JSON by the caller before forwarding the event to relays.

### UI

A simple UI exists to confirm the event being signed on your Trezor's screen. It will be improved later, but for now it just shows all the fields of the event being signed:

#### Trezor Safe 5

Main screen shows the event kind and the content:

image

From the menu you can view the `created_at` and the tags:

image
image

#### Trezor Safe 3

image
image

#### Trezor Model T

image
image

## trezorctl

`trezorctl` accepts the `nostr get-pubkey` and `nostr sign-event` commands which correspond to the protobuf messages described above.

Note: by passing `--account` (which defaults to `0` - the "first" account) to `trezorctl nostr get-pubkey` or `trezorctl nostr sign-event` you can generate a virtually infinite amount of Nostr keypairs from your Trezor - all derived from your seed phrase using the `m/44'/1237'/<account>'/0/0` derivation path, as defined by [NIP-06](https://github.com/nostr-protocol/nips/blob/master/06.md).

### get-pubkey

Example:

`ibz@localhost:~/dev/trezor-firmware$ trezorctl nostr get-pubkey`

Output:

`pubkey: 2dc0dd32b499f7bd428156334c3431893c18e6016cf1bf00187d76513368dba7`

### sign-event

`trezorctl nostr sign-event` will parse the event JSON, send the `NostrSignEvent` to the Trezor device and re-assemble the event JSON with the `id`, `pubkey` and `sig` obtained from the Trezor.

Example:

```ibz@localhost:~/dev/trezor-firmware$ trezorctl nostr sign-event "{\"kind\": 1, \"created_at\": 1733917737, \"tags\": [[\"e\", \"5c83da77af1dec6d7289834998ad7aafbd9e2191396d75ec3cc27f5a77226f36\", \"wss://nostr.example.com\"], [\"p\", \"f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca\"], [\"a\", \"30023:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd\", \"wss://nostr.example.com\"], [\"alt\", \"reply\"]], \"content\": \"Hello world\"}"```

Output:

```signed_event: {"kind": 1, "created_at": 1733917737, "tags": [["e", "5c83da77af1dec6d7289834998ad7aafbd9e2191396d75ec3cc27f5a77226f36", "wss://nostr.example.com"], ["p", "f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca"], ["a", "30023:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd", "wss://nostr.example.com"], ["alt", "reply"]], "content": "Hello world", "id": "c2ea8f08093caebf6d73bfe4edbb7e8825bac8eb4308ab634e23161db18b78f8", "pubkey": "2dc0dd32b499f7bd428156334c3431893c18e6016cf1bf00187d76513368dba7", "sig": "4176eaa45730a617f7393ba53df009278504911b3b30d35e125cb1c0f0d1e92102f468705eaed7d3ae5bb3bb4267d9411f0184a7ab6c451110553c2fca97b39c"}```

## Verifying the signatures

To ensure compatibility with the protocol, a 3rd party library, [`python-nostr`](https://github.com/jeffthibault/python-nostr) - the Python library with most GitHub stars at the time - was used to verify signatures generated by Trezor

```
>>> from nostr.event import Event
>>> from trezorlib import nostr, messages, tools
>>> from trezorlib.client import get_default_client
>>> from datetime import datetime
>>> EVENT_TO_SIGN = {'created_at': int(datetime.now().timestamp()), 'kind': 1, 'content': "Hello, world!"}
>>> sig = nostr.sign_event(get_default_client(), messages.NostrSignEvent(address_n=tools.parse_path("m/44h/1237h/0h/0/0"), **EVENT_TO_SIGN))
>>> Event(public_key=sig.pubkey.hex(), signature=sig.signature.hex(), **EVENT_TO_SIGN).verify()
True
```

### CODE REVIEW

**Comment by prusnak** [comment]
We should check that the `address_n` provided to `NostrGetPubkey ` and `NostrSignEvent`  matches `m/44'/1237'/<account>'/0/0` and fail if it does not.

File: common/protob/messages.proto [hunk 1]
```diff
@@ -105,6 +104,12 @@ enum MessageType {
     MessageType_OwnershipProof = 50 [(bitcoin_only) = true, (wire_out) = true];
     MessageType_AuthorizeCoinJoin = 51 [(bitcoin_only) = true, (wire_in) = true];
 
+    // Nostr
+    MessageType_NostrGetPubkey = 2001 [(bitcoin_only) = true, (wire_in) = true];
+    MessageType_NostrPubkey = 2002 [(bitcoin_only) = true, (wire_out) = true];
+    MessageType_NostrSignEvent = 2003 [(bitcoin_only) = true, (wire_in) = true];
+    MessageType_NostrEventSignature = 2004 [(bitcoin_only) = true, (wire_out) = true];
+
```
prusnak, line 112: These do not belong to the bitcoin only firmware.

python/src/trezorlib/cli/nostr.py
```diff
@@ -0,0 +1,82 @@
...
+    from ..client import TrezorClient
+
+
+@click.group(name="nostr")
+def cli() -> None:
+    pass
+
+
+@cli.command()
+@click.option("-n", "--address", default="m/44'/1237'/0'/0/0", help="BIP-32 path")
```
prusnak, line 35:
```suggestion
@click.option("-a", "--account", default=0, help="Account index")
```

python/src/trezorlib/cli/nostr.py
```diff
@@ -0,0 +1,82 @@
...
+@with_client
+def get_pubkey(
+    client: "TrezorClient",
+    address: str,
+) -> Dict[str, str]:
+    """Derive the pubkey from the seed."""
+
+    address_n = tools.parse_path(address)
```
prusnak, line 43:
```suggestion
    address_n = tools.parse_path("m/44h/1237h/{account}h/0/0")
```

python/src/trezorlib/cli/nostr.py
```diff
@@ -0,0 +1,82 @@
...
+        address_n,
+    )
+
+    return {
+        "pubkey": res.pubkey.hex(),
+    }
+
+
+@cli.command()
+@click.option("-n", "--address", default="m/44'/1237'/0'/0/0", help="BIP-32 path")
```
prusnak, line 56:
```suggestion
@click.option("-a", "--account", default=0, help="Account index")
```

python/src/trezorlib/cli/nostr.py
```diff
@@ -0,0 +1,82 @@
...
+def sign_event(
+    client: "TrezorClient",
+    address: str,
+    event: str,
+) -> Dict[str, str]:
+    """Sign an event using address of given path."""
+
+    event_json = json.loads(event)
+
+    address_n = tools.parse_path(address)
```
prusnak, line 68:
```suggestion
    address_n = tools.parse_path("m/44h/1237h/{account}h/0/0")
```

tests/device_tests/nostr/test_nostr.py
```diff
@@ -0,0 +1,102 @@
...
+
+@pytest.mark.setup_client(mnemonic=WHAT_BLEAK_MNEMONIC)
+def test_get_pubkey_what_bleak(client: Client):
+    _test_get_pubkey(client, WHAT_BLEAK_PK)
+
+
+def _test_get_pubkey(client, expected_pk):
+    response = nostr.get_pubkey(
+        client,
+        n=parse_path("m/44'/1237'/0'/0/0"),
```
prusnak, line 63:
```suggestion
        n=parse_path("m/44h/1237h/0h/0/0"),
```

tests/device_tests/nostr/test_nostr.py
```diff
@@ -0,0 +1,102 @@
...
+    kind = 1
+    tags = []
+    content = "Hello, world!"
+
+    response = nostr.sign_event(
+        client,
+        event=json.dumps(
+            {"created_at": created_at, "kind": kind, "tags": tags, "content": content}
+        ),
+        n=parse_path("m/44'/1237'/0'/0/0"),
```
prusnak, line 80:
```suggestion
        n=parse_path("m/44h/1237h/0h/0/0"),
```

tests/device_tests/nostr/test_nostr.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+    assert response.pubkey == bytes.fromhex(expected_pk)
+
+    expected_id = sha256(
+        json.dumps(
+            [0, expected_pk, created_at, kind, tags, content], separators=(",", ":")
+        ).encode("utf-8")
```
prusnak, line 88:
```suggestion
        ).encode()
```

core/src/apps/bitcoin/keychain.py
```diff
@@ -91,6 +91,9 @@ class MsgWithAddressScriptType(Protocol):
 # SLIP-44 coin type for all Testnet coins
 SLIP44_TESTNET = const(1)
 
+# SLIP-44 "coin type" for Nostr
+SLIP44_NOSTR = const(1237)
```
prusnak, line 95: It seems this is not used anywhere.

core/src/apps/nostr/sign_event.py
```diff
@@ -0,0 +1,35 @@
...
+    from trezor.messages import NostrEventSignature
+
+    address_n = msg.address_n
+    created_at = msg.created_at
+    kind = msg.kind
+    tags = msg.tags
+    content = msg.content
+
+    node = keychain.derive(address_n)
+    pk = node.public_key()[-32:]
```
prusnak, line 26: Can we do this? Don't we care about the public key parity?

ibz:
> We should check that the `address_n` provided to `NostrGetPubkey ` and `NostrSignEvent` matches `m/44'/1237'/<account>'/0/0` and fail if it does not.

Yup! https://github.com/trezor/trezor-firmware/pull/4436/commits/dc1a257c84fcf60ebaa5cd3a03c844542a94cf0b#diff-3dd8aea4110592e07746c9d3233736246c7083e001c9d29db7c1b03f7423f93bR19

core/src/apps/nostr/sign_event.py
ibz: > Can we do this? Don't we care about the public key parity? Or does nostr use x-only keys?

From what I understand, Nostr uses X-only keys, indeed.

common/protob/messages-nostr.proto
```diff
@@ -0,0 +1,62 @@
...
+// Sugar for easier handling in Java
+option java_package = "com.satoshilabs.trezor.lib.protobuf";
+option java_outer_classname = "TrezorMessageNostr";
+
+import "options.proto";
+
+/**
+ * Request: Ask the device for the Nostr public key
+ * @start
+ * @next NostrMessageSignature
```
obrusvit, line 13:
```suggestion
 * @next NostPubkey
```

common/protob/messages-nostr.proto
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Request: Ask device to sign event
+ * @start
+ * @next NostrEventSignature
+ * @next Failure
+ */
+message NostrSignEvent {
+    repeated uint32 address_n = 1; // used to derive the key
+    required uint32 created_at = 2;
```
obrusvit, line 48: I know it's a part of the protocol but add the comments what it means, e.g. `// UNIX timestamp set by the event creator`

core/src/apps/nostr/sign_event.py
```diff
@@ -0,0 +1,54 @@
...
+
+    await paths.validate_path(keychain, address_n)
+
+    node = keychain.derive(address_n)
+    pk = node.public_key()[-32:]
+    sk = node.private_key()
+
+    title = TR.nostr__event_kind_template.format(kind)
+
+    # confirm_value on TR only accepts one single info item
```
obrusvit, line 37: would `confirm_properties` be a better choice?

core/src/apps/nostr/sign_event.py
```diff
@@ -0,0 +1,54 @@
...
+    for t in tags:
+        tags_str += f"\n\n{t[0]}: " + (f" {' '.join(t[1:])}" if len(t) > 1 else "")
+
+    await confirm_value(
+        title, content, "", "nostr_sign_event", info_items=[("", tags_str)]
+    )
+    serialized_tags = ",".join(
+        ["[" + ",".join(f'"{t}"' for t in tag) + "]" for tag in tags]
+    )
+    serialized_event = f'[0,"{hexlify(pk).decode()}",{created_at},{kind},[{serialized_tags}],"{content}"]'
```
obrusvit, line 50: Can you comment on what is `0` here? Is that a part of the protocol or are we forcing some value?

obrusvit (commented):
Some points for discussion.

common/protob/messages-nostr.proto
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/126dc0b4a82bba4c131cc8f694a0eeece51d5f47#diff-a6b60857cdd027137830112c32a998813fd36181b11290bfa83ec573533e13f6R13

common/protob/messages-nostr.proto
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/126dc0b4a82bba4c131cc8f694a0eeece51d5f47#diff-a6b60857cdd027137830112c32a998813fd36181b11290bfa83ec573533e13f6R47

core/src/apps/nostr/sign_event.py
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/126dc0b4a82bba4c131cc8f694a0eeece51d5f47#diff-1bfc3f5ba14a1d44de5030afbb9a530d7b9c4b307ced3789640e9234244379ddR47

core/src/apps/nostr/sign_event.py
ibz: `confirm_properties` is nice but I like having it on 2 pages - one page for the actual content, and an extra page if you want to see the timestamp and tags...

The solution really is to fix `confirm_value` to accept an arbitrary number of properties in the "info" screen.

python/src/trezorlib/cli/nostr.py
```diff
@@ -0,0 +1,85 @@
...
+        event,
+    )
+
+    event_json["id"] = res.id.hex()
+    event_json["pubkey"] = res.pubkey.hex()
+    event_json["sig"] = res.signature.hex()
+
+    return {
+        "signed_event": json.dumps(event_json),
+    }
```
matejcik, line 85: any reason you're returning json-in-json?

tests/device_tests/nostr/test_nostr.py
```diff
@@ -0,0 +1,117 @@
...
+            [0, expected_pk, created_at, kind, tags, content], separators=(",", ":")
+        ).encode()
+    ).digest()
+
+    assert response.id == expected_id
+
+    vk = VerifyingKey.from_string(
+        b("\x03") + bytes.fromhex(expected_pk),
+        curve=SECP256k1,
+        # this is a pretty silly way to tell VerifyingKey
```
matejcik, line 111: i mean you could instead store the result of `json.dumps` and verify the actual message instead of the manually computed digest

tests/device_tests/nostr/test_nostr.py
```diff
@@ -0,0 +1,117 @@
...
+    "leader monkey parrot ring guide accident before fence cannon height naive bean"
+)
+LEAD_MONKEY_PK = "17162c921dc4d2518f9a101db33695df1afb56ab82f5ff3e5da6eec3ca5cd917"
+
+WHAT_BLEAK_MNEMONIC = "what bleak badge arrange retreat wolf trade produce cricket blur garlic valid proud rude strong choose busy staff weather area salt hollow arm fade"
+WHAT_BLEAK_PK = "d41b22899549e1f3d335a31002cfd382174006e166d3e658e3a5eecdb6463573"
+
+
+@pytest.mark.setup_client(mnemonic=LEAD_MONKEY_MNEMONIC)
+def test_sign_event_lead_monkey(client: Client):
```
matejcik, line 43: there's a better way to write these tests:
```
LEAD_MONKEY = pytest.mark.setup_client(...)
WHAT_BLEAK = pytest.mark.setup_client(...)

VECTORS = (  # pubkey_hex
    pytest.param("cafecacebeef", mark=LEAD_MONKEY),
    pytest.param(...)
)

@pytest.mark.parametrize("pubkey_hex", VECTORS)
def test_get_pubkey(client, pubkey_hex):
   ...
```

dtto for `sign_event` except i'd prefer _also_ hardcoding the signature bytes, in addition to verifying

tests/device_tests/nostr/test_nostr.py
```diff
@@ -0,0 +1,117 @@
...
+            "wss://nostr.example.com",
+        ],
+        ["alt", "reply"],
+    ]
+    content = "Hello, world!"
+
+    response = nostr.sign_event(
+        client,
+        event=json.dumps(
+            {"created_at": created_at, "kind": kind, "tags": tags, "content": content}
```
matejcik, line 93: instead, make a static event definition with a fixed time etc. on top level. we want the tests to be deterministic if we at all can

python/src/trezorlib/nostr.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+
+
+@expect(messages.NostrEventSignature)
+def sign_event(
+    client: "TrezorClient",
+    n: "Address",
+    event: AnyStr,
+) -> "MessageType":
+    event_json = json.loads(event)
```
matejcik, line 48: if you want a structured dict in, pass it in as a dict, not as a string. Let the caller do their own `json.loads` if they want to. for bonus points, use `TypedDict`.
(wondering if we should instead introduce a dataclass for the parameters, or perhaps use the protobuf message directly? given that we're _returning_ the corresponding protobuf instance too)

python/src/trezorlib/nostr.py
```diff
@@ -0,0 +1,62 @@
...
+from .tools import expect
+
+if TYPE_CHECKING:
+    from .client import TrezorClient
+    from .protobuf import MessageType
+    from .tools import Address
+
+
+@expect(messages.NostrPubkey)
+def get_pubkey(
```
matejcik, line 31: this function should return just `bytes` of the public key.

perhaps it's a good idea to rebase on #4490

python/src/trezorlib/cli/nostr.py
```diff
@@ -0,0 +1,85 @@
...
+@with_client
+def get_pubkey(
+    client: "TrezorClient",
+    account: int,
+) -> Dict[str, str]:
+    """Derive the pubkey from the seed."""
+
+    address_n = tools.parse_path(PATH_TEMPLATE.format(account))
+
+    res = nostr.get_pubkey(
```
matejcik, line 48: just `return nostr.get_pubkey`, which should be changed to return `bytes` per the other comment

core/src/apps/nostr/sign_event.py
prusnak: But it is not guaranteed that `keychain.derive` returns an x-only key, right?
I think we need to convert the key if a non x-only key is returned.

core/.changelog.d/4160.added
```diff
@@ -0,0 +1 @@
+Add Nostr support.
```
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/7f493ea5089762ee62535f37cc63904ebd851d6b#diff-058e8b9e68a05b1f482e5961c397d3ca53265f837ddd75e6021620d43e6d309aL1

core/src/apps/nostr/get_pubkey.py
```diff
@@ -0,0 +1,24 @@
...
+from apps.common.keychain import auto_keychain
+
+if TYPE_CHECKING:
+    from trezor.messages import NostrGetPubkey, NostrPubkey
+
+    from apps.common.keychain import Keychain
+
+
+@auto_keychain(__name__)
+async def get_pubkey(msg: NostrGetPubkey, keychain: Keychain) -> NostrPubkey:
```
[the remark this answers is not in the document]
ibz: Yes, I want to show the key, plus (very importantly) a QR code.

core/src/trezor/wire/__init__.py
```diff
@@ -95,6 +95,7 @@ async def handle_session(iface: WireInterface) -> None:
             except Exception as exc:
                 # Log and ignore. The session handler can only exit explicitly in the
                 # following finally block.
+                do_not_restart = True
```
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/7f493ea5089762ee62535f37cc63904ebd851d6b#diff-8be600ec13beb037f92f905b366e48d31a6f3f2bef7f7a0ef410cb790b2900b2L98

tests/device_tests/nostr/test_nostr.py
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/7f493ea5089762ee62535f37cc63904ebd851d6b#diff-1a61ba923388ca946c730e9cec2ca549a97e355ff7a53673617057d39f95f0f5R44

The signature cannot be hardcoded because every time an event is signed it gets a different signature. That is the reason to use `verify` rather than comparing to a known signature.

tests/device_tests/nostr/test_nostr.py
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/7f493ea5089762ee62535f37cc63904ebd851d6b#diff-1a61ba923388ca946c730e9cec2ca549a97e355ff7a53673617057d39f95f0f5R52

python/src/trezorlib/nostr.py
ibz: Actually I don't think it should be `bytes` but rather the hex-encoded bytes, which is human readable and the (most) standard way to encode Nostr keys. Another encoding commonly used is *NPUB* - and I should probably add an option to return that encoding if desired, but at the most fundamental level it is the hex encoding, which is what all code dealing with Nostr is using. `bytes` seems strange, because it looks bad, can't even be copy-pasted, and is of no use...

python/src/trezorlib/cli/nostr.py
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/7f493ea5089762ee62535f37cc63904ebd851d6b#diff-fe4291c13985e00ae63aea21ee979921f4e974707d7830066a0aa0956aa120deL48

python/src/trezorlib/cli/nostr.py
ibz: No reason, removed it. https://github.com/trezor/trezor-firmware/pull/4436/commits/7f493ea5089762ee62535f37cc63904ebd851d6b#diff-fe4291c13985e00ae63aea21ee979921f4e974707d7830066a0aa0956aa120deL84

python/src/trezorlib/mapping.py
```diff
@@ -62,7 +62,9 @@ def encode(self, msg: protobuf.MessageType) -> Tuple[int, bytes]:
         """
         wire_type = self.class_to_type_override.get(type(msg), msg.MESSAGE_WIRE_TYPE)
         if wire_type is None:
-            raise ValueError("Cannot encode class without wire type")
+            raise ValueError(
```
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/6806267e46d350c93e6649bf1eeacb8c9468c375

core/src/apps/nostr/sign_event.py
ibz: Isn't conversion to x-only basically dropping the first byte if the key is 33 bytes long? Which is the same as taking the last 32 bytes always?

tests/device_tests/nostr/test_nostr.py
matejcik: > The signature cannot be hardcoded because every time an event is signed it gets a different signature.

what is the source of this nondeterminism?

ibz: It's a feature of ECDSA - a random value is included. This is to prevent deriving of the private key if you have many signatures that you can analyze.

matejcik: that's what RFC 6979 is for -- i was under the impression that this is built into ecdsa signing on library level...?

python/src/trezorlib/nostr.py
matejcik: > `bytes` seems strange, because it looks bad, can't even be copy-pasted, and is of no use...

I mean you're deep inside a library, nobody is supposed to be copy-pasting the return value of this. 
If there is a canonical representation of the pubkey as readable string, the `NostrPubkey` message should send _that_ from Trezor. If there isn't (you apparently have two choices, npub / hex), and you're returning raw bytes, the library function should return raw bytes.

Consumers of `trezorlib.nostr.get_pubkey` are expected to do their own human readable conversion if they need to; specifically, if encoding to npub is simple, feel free to implement that in `cli/nostr.py` and add an option to the user-exposed function

matejcik: (fwiw if you return bytes from a click-exposed function, there's some sort of conversion machinery that will automagically print it as hex for the user)

python/src/trezorlib/cli/nostr.py
matejcik: follow-up question: is the `{signed_event...}` json text an expected exchange format?

python/src/trezorlib/cli/nostr.py
```diff
@@ -0,0 +1,91 @@
...
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the License along with this library.
+# If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
+
+import json
+from typing import TYPE_CHECKING, Dict
```
matejcik, line 18: instead:
```
from __future__ import annotations

# ...

import typing as t

# ...

if t.TYPE_CHECKING:
# ...

# ...
def foo() -> dict[str, str]:
```

python/src/trezorlib/nostr.py
```diff
@@ -0,0 +1,45 @@
...
+    n: "Address",
+) -> str:
+    return client.call(
+        messages.NostrGetPubkey(
+            address_n=n,
+        )
+    ).pubkey.hex()
+
+
+@expect(messages.NostrEventSignature)
```
matejcik, line 40: per #4490 please drop the `@expect` and replace with `client.call(..., expect=NostrEventSignature)`
dtto for `get_pubkey`

tests/device_tests/nostr/test_nostr.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    assert response == pubkey_hex
+
+
+@pytest.mark.parametrize("pubkey_hex", VECTORS)
+def test_sign_event(client, pubkey_hex):
+    response = nostr.sign_event(
+        client,
+        messages.NostrSignEvent(
```
matejcik, line 87: would it make sense to move this construction to top level directly?

matejcik (changes requested):

tests/device_tests/nostr/test_nostr.py
ibz: Right... I think that in theory we should be allowed to use the deterministic signatures, but from what I noticed Nostr signatures are usually of the non-deterministic kind. Do you think it should be otherwise? I guess the library can do both, if needed, but isn't the non-deterministic version better?

matejcik: > isn't the non-deterministic version better?

i'm not aware of any argument why it would be.
in our particular case, there's another issue on top: if the nonce is not deterministic, attacks like Dark Skippy can use the signatures to smuggle out secret bits without the user being able to detect it. (with rfc6979 at least you can verify the signature if you also have the seed on the host side)

so please change to deterministic signatures, unless you have a spec explicitly saying you shouldn't do that. (and if so let's see it and talk to R&D about this)

python/src/trezorlib/nostr.py
matejcik: okay, if that's the case, please modify the Trezor side to return the hex string

tests/device_tests/nostr/test_nostr.py
ibz: I don't think RFC 6979 would work actually because we need Schnorr signatures. So I used BIP-340 to sign and I got deterministic signatures now. 45d4ee0796e573dba4f1eb6bafe1f7662e3c7959

python/src/trezorlib/nostr.py
ibz: Good point. It's all `str` now: a95dc04d5f42c357a5d00b171c1d780782b55f62 !

python/src/trezorlib/cli/nostr.py
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/a95dc04d5f42c357a5d00b171c1d780782b55f62#diff-fe4291c13985e00ae63aea21ee979921f4e974707d7830066a0aa0956aa120deR17

python/src/trezorlib/nostr.py
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/a95dc04d5f42c357a5d00b171c1d780782b55f62#diff-d0a53275e15314e83ef900790cfdd6af05df702377562408a76287ae1507a853R36

tests/device_tests/nostr/test_nostr.py
ibz: https://github.com/trezor/trezor-firmware/pull/4436/commits/a95dc04d5f42c357a5d00b171c1d780782b55f62#diff-1a61ba923388ca946c730e9cec2ca549a97e355ff7a53673617057d39f95f0f5R72

core/src/apps/nostr/__init__.py
```diff
@@ -0,0 +1,10 @@
+if not __debug__:
+    from trezor import utils
+
+    utils.halt("Disabled in production mode")
+
+from apps.common.paths import PATTERN_BIP44
+
+CURVE = "secp256k1"
+SLIP44_ID = 1237
+PATTERN = PATTERN_BIP44
```
matejcik, line 10: why would Nostr use a BIP44 5-part pattern?

is there any reason for unhardened derivations at all? putting multiple pubkeys under the same "account", which still need to be "discoverable" together?

python/src/trezorlib/nostr.py
ibz: Back to `bytes` at the protocol and library level, and `str` conversion only in the cli. https://github.com/trezor/trezor-firmware/pull/4436/commits/04e42c873f7c1d900027e37e268dc4df42f72e38

core/src/apps/nostr/__init__.py
matejcik: update: we opened https://github.com/nostr-protocol/nips/issues/1774 and we could write another NIP to specify a better path, but there is no point pausing this PR any longer than it has already been paused.

please add a comment here that this path is provisional and we don't want it, then let's merge

ibz: Added comment here and at the very beginning of the PR description.

core/src/trezor/ui/layouts/bolt/__init__.py
```diff
@@ -707,9 +707,6 @@ def confirm_value(
     if description and value:
         description += ":"
 
-    if not verb and not hold:
-        raise ValueError("Either verb or hold=True must be set")
-
```
matejcik, line 712: is this a merge problem? or why is this change part of the nostr PR?

PR #4439, opened 2024-12-10
ci: create emualtor release workflow for release versions

vdovhanych:
Don't merge this yet, I want to test it on a forked repo to see if it will do what I want it to. 

It utilizes the releases.json file that adds a model list to the version in order to have emulators released for the same models we release firmware for.

### DIFF
--- a/.github/workflows/release-emu.yml
+++ b/.github/workflows/release-emu.yml
@@ -1,0 +1,203 @@
+name: "[Release] emulators for trezor-user-env"
+
+on:
+  push:
+    tags:
+      - 'core/v*'
+      - 'legacy/v*'
+  workflow_dispatch:
+
+permissions:
+  id-token: write       # for fetching the OIDC token
+  contents: read        # for actions/checkout
+
+jobs:
+  get_models:
+    name: Get models
+    runs-on: ubuntu-latest
+    outputs:
+      models: ${{ steps.get_models.outputs.models }}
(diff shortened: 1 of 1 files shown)


vdovhanych:
Should be working fine, i think we could merge it.

.github/workflows/release-emu.yml
```diff
@@ -0,0 +1,203 @@
...
+      matrix:
+        model: ${{ fromJson(needs.get_models.outputs.models) }}
+        coins: [universal]
+        type: [debuglink]
+        asan: [noasan]
+        exclude:
+          - type: normal
+            asan: asan
+    env:
+      TREZOR_MODEL: ${{ matrix.model == 'T2T1' && 'T' || matrix.model == 'T2B1' && 'R' || matrix.model }}
```
mmilata, line 51: Should be
```suggestion
      TREZOR_MODEL: ${{ matrix.model }}
```
since #4477 was merged meanwhile

mmilata (approved):
I don't like the code duplication for builds, should be probably factored out into a [composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action). But that can be done later, let's see if it works as it is.

.github/workflows/release-emu.yml
vdovhanych: removed it

mmilata (approved):

PR #4442, opened 2024-12-11
fix missing rsod info

TychoVrahe:
adds default title, footer or message if some of these are not set

### DIFF
--- a/core/embed/util/rsod/rsod.c
+++ b/core/embed/util/rsod/rsod.c
@@ -57,9 +57,15 @@ void rsod_terminal(const systask_postmortem_t* pminfo) {
       message = message_buf;
       break;
     case TASK_TERM_REASON_ERROR:
-      title = pminfo->error.title;
-      message = pminfo->error.message;
-      footer = pminfo->error.footer;
+      if (pminfo->error.title[0] != '\0') {
+        title = pminfo->error.title;
+      }
+      if (pminfo->error.message[0] != '\0') {
+        message = pminfo->error.message;
+      }
+      if (pminfo->error.footer[0] != '\0') {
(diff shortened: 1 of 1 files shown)


core/embed/util/rsod/rsod.c
```diff
@@ -57,9 +58,16 @@ void rsod_terminal(const systask_postmortem_t* pminfo) {
       message = message_buf;
       break;
     case TASK_TERM_REASON_ERROR:
-      title = pminfo->error.title;
-      message = pminfo->error.message;
-      footer = pminfo->error.footer;
+      if (strnlen(pminfo->error.title, sizeof(pminfo->error.title)) != 0) {
```
cepetr, line 61: `pminfo->error->xxx` strings are always null-terminated, so we can simplify the condition to `if (pminfo->error.title[0] != '\0') {`.

cepetr (changes requested):

core/embed/util/rsod/rsod.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/cdb0427a58d6413d64b94adb54bc2107f91cb249

TychoVrahe: changed the other half here: https://github.com/trezor/trezor-firmware/commit/53d6f1f22bb9a8549c6f84ec6e6817054cf406b7

bosomt:
**QA OK**

```
INTERNAL ERROR

no image buff

PLEASE VISIT TREZOR.IO/RSOD
```

 - Device: Trezor T3T1 2.8.7 regular (revision e196413bb745443610211c325623ba7cee7309e9)

PR #4447, opened 2024-12-13
Ensure dependency between PIN and wipe code

bieleluk:
This PR:
- ensures that the wipe code can not be enabled without without PIN being active
- ensures that the PIN cannot be turned off with the wipe code being active


bieleluk:
Rebasing on top of a big change recently added to `main`

core/src/apps/management/change_pin.py
```diff
@@ -26,6 +27,10 @@ async def change_pin(msg: ChangePin) -> Success:
     # confirm that user wants to change the pin
     await _require_confirm_change_pin(msg)
 
+    # Do not allow to remove the PIN if the wipe code is set.
+    if msg.remove and config.has_wipe_code():
+        await error_wipe_code_exists()
+
```
matejcik: pinging @andrewkozlik for this one: the deepest consideration here is whether we want to:
* show this warning, revealing the existence of wipe code,
* silently remove wipe code as well,
* or behave as before, keeping the wipe code set

andrewkozlik: > show this warning, revealing the existence of wipe code,

Features show `wipe_code_protection` if the device is unlocked. So if the device is unlocked or the entered PIN is correct, then we can show this error, without revealing anything new. The check just needs to be coded differently to ensure an unlocked device.

As for the choice between the three options, it's mainly up to product. They each make some sense. However, I am sure there was discussion with them about this already years ago, so the current behavior is not a random decision.

core/translations/en.json
```diff
@@ -719,6 +719,8 @@
     "pin__tries_left": "tries left",
     "pin__turn_off": "Are you sure you want to turn off PIN protection?",
     "pin__turn_on": "Turn on PIN protection?",
+    "pin__wipe_code_exists_description": "Wipe code must be turned off before turnig off PIN protection.",
```
mmilata:
```suggestion
    "pin__wipe_code_exists_description": "Wipe code must be turned off before turning off PIN protection.",
```

core/src/trezor/ui/layouts/bolt/__init__.py
```diff
@@ -1792,6 +1792,30 @@ def wipe_code_same_as_pin_popup() -> Awaitable[None]:
...
+async def wipe_code_pin_not_set_popup(
+    title: str, description: str, button: str
+) -> NoReturn:
+    await show_error_and_raise(
+        "warning_pin_not_set",
+        description,
+        title,
+        button,
+    )
+    assert False
```
mmilata: Appreciate the defensive coding, however:
* asserts are not included in PYOPT=1 builds so I'd change this to `raise RuntimeError  # should be unreachable` to be safe
* i'd also move this to the caller so that we don't rely on all 4 UI implementations to do the right thing

Also can you please change the two `assert False` statements in `request_pin.py` to raise instead? These are on me ...

core/src/apps/management/change_pin.py
mmilata: > The check just needs to be coded differently to ensure an unlocked device.

That should implicitly be true here as the handler is not on the `ALLOW_WHILE_LOCKED` whitelist?

mmilata (commented):
Looking good, just a few minor things

core/src/trezor/ui/layouts/bolt/__init__.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4447/commits/d5a8b3b27d038eb53f04e50e6f1d1baa05fc7171
Did everything except merging error implementations. The reason is that the `Eckhart`  has a different title from other layouts. This has already occurred several times, and the result was always the same, unfortunately.

mmilata: Sorry, I only meant moving the `raise RuntimeError` to the caller

bieleluk: https://github.com/trezor/trezor-firmware/pull/4447/commits/108ce6e58872d41fc6ceead38251ed285558f89c

PR #4448, opened 2024-12-13
introduce NRF and BLE drivers

TychoVrahe:
This PR introduces NRF driver, responsible for handling communication with NRF, and BLE driver, which provides interface for using BLE via the NRF.


core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,779 @@
...
+
+void nrf_deinit(void) {
+  nrf_driver_t *drv = &g_nrf_driver;
+  if (!drv->initialized) {
+    return;
+  }
+
+  nrf_stop();
+
+  drv->initialized = false;
```
cepetr, line 334: We should reset the UART/SPI peripherals and disable interrupts in the NVIC controller here. If we call `nrf_deinit` in `drivers_deinit()` within the bootloader, we can also remove the `if (drv->initialized)` condition from the interrupt handlers.

cepetr (changes requested):

core/embed/sys/syscall/stm32/syscall_verifiers.c
```diff
@@ -715,4 +715,56 @@ secbool firmware_get_vendor__verified(char *buff, size_t buff_size) {
...
 
+// ---------------------------------------------------------------------
+
+#ifdef USE_BLE
+void ble_get_state__verified(ble_state_t *state) {
+  if (!probe_write_access(state, sizeof(*state))) {
+    goto access_violation;
+  }
+
+  ble_get_state(state);
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

core/embed/io/ble/stm32/ble_comm_defs.h
```diff
@@ -0,0 +1,59 @@
...
+  uint32_t app_version;
+  uint32_t bld_version;
+
+} event_status_msg_t;
+
+typedef enum {
+  INTERNAL_EVENT_STATUS = 0x01,
+  INTERNAL_EVENT_PAIRING_REQUEST = 0x04,
+  INTERNAL_EVENT_PAIRING_CANCELLED = 0x05,
+} InternalEvent_t;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

core/embed/io/nrf/inc/io/nrf.h
```diff
@@ -0,0 +1,81 @@
...
+  NRF_STATUS_TIMEOUT = 1,  // Timeout occurred
+  NRF_STATUS_ERROR = 2,    // General error
+  NRF_STATUS_ABORTED = 3,  // Packet was aborted
+} nrf_status_t;
+
+typedef struct {
+  const uint8_t *data;
+  uint32_t len;
+
+} nrf_packet_t;
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

core/embed/io/nrf/inc/io/nrf.h
```diff
@@ -0,0 +1,81 @@
...
+bool nrf_is_running(void);
+
+// Register listener for a service
+// The listener will be called when a message is received for the service
+// The listener will be called from an interrupt context
+void nrf_register_listener(nrf_service_id_t service,
+                           void (*listener)(const uint8_t *data, uint32_t len));
+
+// Unregister listener for a service
+void nrf_unregister_listener(nrf_service_id_t service);
```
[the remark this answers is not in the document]
TychoVrahe: went with the second option https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

lets add the context callback when we need it

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,779 @@
...
+typedef struct {
+  UART_HandleTypeDef urt;
+  DMA_HandleTypeDef urt_tx_dma;
+
+  uint8_t urt_tx_buffers[UART_QUEUE_SIZE][sizeof(nrf_uart_tx_data_t)];
+  tsqueue_entry_t urt_tx_queue_entries[UART_QUEUE_SIZE];
+  tsqueue_t urt_tx_queue;
+  nrf_uart_tx_data_t urt_sending;
+
+  uint8_t urt_rx_buf[UART_PACKET_SIZE];
```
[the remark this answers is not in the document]
TychoVrahe: tried to do some renaming https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765. there is rx for both spi and uart, so there is some ambiguity, which is now resolved by calling the spi rx buffer 'long_rx_buffer', as it servers for sending longer messages, but i am not entirely sure that this was an overall improvement

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,779 @@
...
+}
+
+void nrf_init(void) {
+  nrf_driver_t *drv = &g_nrf_driver;
+
+  if (drv->initialized) {
+    return;
+  }
+
+  __HAL_RCC_USART3_CLK_ENABLE();
```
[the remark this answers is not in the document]
TychoVrahe: maybe, seems a bit premature though, lots of constants that we don't really need yet

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,779 @@
...
+void nrf_register_listener(nrf_service_id_t service,
+                           void (*listener)(const uint8_t *data,
+                                            uint32_t len)) {
+  nrf_driver_t *drv = &g_nrf_driver;
+  if (!drv->initialized) {
+    return;
+  }
+
+  if (service < NRF_SERVICE_CNT) {
+    drv->service_listeners[service] = listener;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,779 @@
...
+  void *context;
+} nrf_uart_tx_data_t;
+
+typedef struct {
+  UART_HandleTypeDef urt;
+  DMA_HandleTypeDef urt_tx_dma;
+
+  uint8_t urt_tx_buffers[UART_QUEUE_SIZE][sizeof(nrf_uart_tx_data_t)];
+  tsqueue_entry_t urt_tx_queue_entries[UART_QUEUE_SIZE];
+  tsqueue_t urt_tx_queue;
```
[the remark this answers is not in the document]
TychoVrahe: removed urt_tx_running https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,779 @@
...
+  uint8_t urt_rx_byte;
+  uint16_t urt_rx_idx;
+
+  SPI_HandleTypeDef spi;
+  DMA_HandleTypeDef spi_dma;
+  uint8_t spi_buffer[SPI_PACKET_SIZE];
+
+  int32_t urt_tx_msg_id;
+  bool urt_tx_running;
+  bool spi_rx_running;
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,779 @@
...
+
+  drv->urt_rx_idx = 0;
+  drv->urt_rx_len = 0;
+  drv->urt_tx_msg_id = -1;
+  drv->urt_tx_running = false;
+
+  while (tsqueue_dequeue(&drv->urt_tx_queue, (uint8_t *)&drv->urt_sending,
+                         sizeof(nrf_uart_tx_data_t), NULL,
+                         &drv->urt_tx_msg_id)) {
+    drv->urt_sending.callback(NRF_STATUS_ERROR, drv->urt_sending.context);
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/bec19dae870a86a582b25fbf032ac77ddaf72765

the gpios currently stay initialized, including the AF ones

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+
+  tsqueue_init(&drv->data_queue, drv->data_queue_entries,
+               (uint8_t *)drv->data_buffers, BLE_RX_PACKET_SIZE,
+               DATA_QUEUE_LEN);
+
+  tsqueue_init(&drv->send_queue, drv->send_queue_entries,
+               (uint8_t *)drv->send_buffer, NRF_MAX_TX_DATA_SIZE,
+               SEND_QUEUE_LEN);
+
+  systimer_t *timer = systimer_create(ble_loop, NULL);
```
mmilata, line 303: Would disabling the timer in `ble_deinit` (or `ble_stop`) help with energy consumption or is that premature optimization? Currently neither function is ever called.

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+  if (!drv->initialized) {
+    return;
+  }
+
+  ble_event_t event = {.type = BLE_PAIRING_CANCELLED, .data_len = 0};
+  tsqueue_enqueue(&drv->event_queue, (uint8_t *)&event, sizeof(event), NULL);
+}
+
+static void ble_process_rx_msg(const uint8_t *data, uint32_t len) {
+  switch (data[0]) {
```
mmilata, line 208: Is it safe to assume that `len` > 0?

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,786 @@
...
+    return false;
+  }
+
+  bool aborted = tsqueue_abort(&drv->tx_queue, id, NULL, 0, NULL);
+
+  if (aborted) {
+    return true;
+  }
+
+  if (drv->tx_msg_id == id) {
```
cepetr, line 494: It seems to me, this requires synchronization with the interrupt (UART_TxCpltCallback).

cepetr (changes requested):

core/embed/util/tsqueue/tsqueue.c
```diff
@@ -0,0 +1,192 @@
...
+
+  irq_unlock(key);
+}
+
+static int32_t get_next_id(tsqueue_t *queue) {
+  int val = 1;
+  if (queue->next_id < INT32_MAX) {
+    val = queue->next_id;
+    queue->next_id++;
+  } else {
```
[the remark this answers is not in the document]
TychoVrahe: fixed, thx https://github.com/trezor/trezor-firmware/commit/191fa1a026fb38d579b593ff0ccff0fe2f32bccd

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+  bool accept_msgs;
+  ble_event_t event_buffers[EVENT_QUEUE_LEN];
+  tsqueue_entry_t event_queue_entries[EVENT_QUEUE_LEN];
+  tsqueue_t event_queue;
+
+  uint8_t data_buffers[DATA_QUEUE_LEN][BLE_RX_PACKET_SIZE];
+  tsqueue_entry_t data_queue_entries[DATA_QUEUE_LEN];
+  tsqueue_t data_queue;
+
+  uint8_t send_buffer[NRF_MAX_TX_DATA_SIZE];
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/191fa1a026fb38d579b593ff0ccff0fe2f32bccd

core/embed/io/ble/inc/io/ble.h
```diff
@@ -0,0 +1,125 @@
...
+typedef struct {
+  ble_event_type_t type;
+  int connection_id;
+  uint8_t data_len;
+  uint8_t data[6];
+} ble_event_t;
+
+typedef struct {
+  bool connected;
+  uint8_t peer_count;
```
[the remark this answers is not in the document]
TychoVrahe: its number of bonded devices. 

currently when we advertise with whitelist we put all bonded devices there, but its not mandatory 

the intent is to use this to automatically start advertising (with whitelist) if there is at least one bonded device (but that is something that needs product decision)

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+  nrf_send_msg(NRF_SERVICE_BLE_MANAGER, data, sizeof(data), NULL, NULL);
+}
+
+static void ble_send_advertising_off(void) {
+  uint8_t cmd = INTERNAL_CMD_ADVERTISING_OFF;
+  nrf_send_msg(NRF_SERVICE_BLE_MANAGER, &cmd, sizeof(cmd), NULL, NULL);
+}
+
+static bool ble_send_erase_bonds(void) {
+  if (!nrf_is_running()) {
```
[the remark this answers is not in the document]
TychoVrahe: removed the checks here a put it into `nrf_send_msg`

core/embed/io/ble/stm32/ble.c
TychoVrahe: not sure if it affects energy consumption, but it should be stopped. https://github.com/trezor/trezor-firmware/commit/191fa1a026fb38d579b593ff0ccff0fe2f32bccd

the deinits will eventually be called when puting the device in low power mode, or when jumping between bootloader and firmware. 

but currently we don't have ble in bootloader and low power mode is waiting for https://github.com/trezor/trezor-firmware/pull/4454

core/embed/io/ble/stm32/ble.c
TychoVrahe: probably mostly safe but i added the check anyway in https://github.com/trezor/trezor-firmware/commit/191fa1a026fb38d579b593ff0ccff0fe2f32bccd

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+bool ble_get_event(ble_event_t *event) {
+  ble_driver_t *drv = &g_ble_driver;
+
+  if (!drv->initialized) {
+    return false;
+  }
+
+  ble_event_t tmp_event = {0};
+  uint16_t len = 0;
+  bool read = tsqueue_dequeue(&drv->event_queue, (uint8_t *)&tmp_event,
```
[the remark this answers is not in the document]
TychoVrahe: simplified https://github.com/trezor/trezor-firmware/commit/b23337ea0879d5a0dfcf92a24a9106479e0490a8

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+uint32_t ble_read(uint8_t *data, uint16_t max_len) {
+  ble_driver_t *drv = &g_ble_driver;
+
+  if (!drv->initialized) {
+    return 0;
+  }
+
+  tsqueue_t *queue = &drv->data_queue;
+
+  uint16_t read_len = 0;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/b23337ea0879d5a0dfcf92a24a9106479e0490a8

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+
+  memcpy(event, &tmp_event, sizeof(ble_event_t));
+
+  return true;
+}
+
+void ble_get_state(ble_state_t *state) {
+  const ble_driver_t *drv = &g_ble_driver;
+
+  if (state == NULL) {
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/b23337ea0879d5a0dfcf92a24a9106479e0490a8

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,505 @@
...
+    if (drv->ping_cntr >= PING_PERIOD / LOOP_PERIOD_MS) {
+      drv->ping_cntr = 0;
+    }
+
+    uint8_t data[NRF_MAX_TX_DATA_SIZE] = {0};
+    if (tsqueue_dequeue(&drv->send_queue, data, NRF_MAX_TX_DATA_SIZE, NULL,
+                        NULL)) {
+      if (!nrf_send_msg(NRF_SERVICE_BLE, data, NRF_MAX_TX_DATA_SIZE, NULL,
+                        NULL)) {
+        tsqueue_enqueue(&drv->send_queue, data, NRF_MAX_TX_DATA_SIZE, NULL);
```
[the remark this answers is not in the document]
TychoVrahe: yeah the intention was to always only have single message buffer here. i added comment for the TX_QUEUE_LEN constant so that no one changes it accidentally 
https://github.com/trezor/trezor-firmware/commit/191fa1a026fb38d579b593ff0ccff0fe2f32bccd

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/b23337ea0879d5a0dfcf92a24a9106479e0490a8

core/embed/io/nrf/stm32u5/nrf.c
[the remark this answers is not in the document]
TychoVrahe: see commits https://github.com/trezor/trezor-firmware/commit/c2a42399ed509ff1e2b3fca14af62c14fab152b4 for the renaming and https://github.com/trezor/trezor-firmware/commit/ea2af5e51d3b7182c279731c8b1f77fcca37687a and https://github.com/trezor/trezor-firmware/commit/8af5157bf2469250a68c33e3e60c303c950b4146 for introduction of `uart_packet_t` and `spi_packet_t` respectively.

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,770 @@
...
+  HAL_UART_AbortReceive(&drv->urt);
+  HAL_UART_AbortTransmit(&drv->urt);
+
+  if (drv->tx_request.callback != NULL) {
+    drv->tx_request.callback(NRF_STATUS_ERROR, drv->tx_request.context);
+  }
+
+  drv->rx_idx = 0;
+  drv->rx_len = 0;
+  drv->tx_request_id = -1;
```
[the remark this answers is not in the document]
TychoVrahe: fixed(we don't need to update it in the loop)  https://github.com/trezor/trezor-firmware/commit/e8a083d23efc53d89811fae328f6710ce06d5d5e

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,770 @@
...
+    return -1;
+  }
+
+  if (!nrf_is_running()) {
+    return -1;
+  }
+
+  int32_t id = 0;
+
+  nrf_tx_request_t buffer;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/e8a083d23efc53d89811fae328f6710ce06d5d5e

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,770 @@
...
+  GPIO_InitStructure.Speed = GPIO_SPEED_FREQ_LOW;
+  GPIO_InitStructure.Pin = NRF_OUT_RESET_PIN;
+  HAL_GPIO_Init(NRF_OUT_RESET_PORT, &GPIO_InitStructure);
+
+  NRF_IN_GPIO0_CLK_ENA();
+  GPIO_InitStructure.Mode = GPIO_MODE_INPUT;
+  GPIO_InitStructure.Pull = GPIO_PULLDOWN;
+  GPIO_InitStructure.Speed = GPIO_SPEED_FREQ_LOW;
+  GPIO_InitStructure.Pin = NRF_IN_GPIO0_PIN;
+  HAL_GPIO_Init(NRF_IN_GPIO0_PORT, &GPIO_InitStructure);
```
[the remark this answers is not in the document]
TychoVrahe: it will be used later, maybe during the NRF firmware update, of signaling that NRF bootloader is active, i am not sure yet

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,769 @@
...
+  spi_packet_t *packet = &drv->long_rx_buffer;
+
+  uint8_t crc = crc8((uint8_t *)&drv->long_rx_buffer, sizeof(spi_packet_t) - 1,
+                     0x07, 0x00, false);
+
+  if ((packet->service_id & 0xF0) != START_BYTE || packet->crc != crc) {
+    HAL_SPI_Abort(&drv->spi);
+    HAL_SPI_Receive_DMA(&drv->spi, (uint8_t *)&drv->long_rx_buffer,
+                        sizeof(spi_packet_t));
+    ;
```
cepetr, line 678: Unnecessary semicolon

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,769 @@
...
+#define UART_OVERHEAD_SIZE (sizeof(uart_packet_t) - NRF_MAX_TX_DATA_SIZE)
+#define UART_HEADER_SIZE (UART_OVERHEAD_SIZE - 1)
+
+#define TX_QUEUE_SIZE (8)
+
+#define START_BYTE (0xA0)
+
+typedef struct {
+  uart_packet_t packet;
+  uint8_t len;
```
cepetr, line 58: Maybe the `len` field could be omitted since it is already part of the `uart_packet_t` and serves the same purpose.

cepetr (changes requested):

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,769 @@
...
+    return -1;
+  }
+
+  if (!nrf_is_running()) {
+    return -1;
+  }
+
+  int32_t id = 0;
+
+  nrf_tx_request_t buffer = {0};
```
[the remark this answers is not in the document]
TychoVrahe: renamed https://github.com/trezor/trezor-firmware/commit/601953eaaa98ac5fe5a9f08814ada0af48bf84de

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,486 @@
...
+               TX_QUEUE_LEN);
+
+  drv->timer = systimer_create(ble_loop, NULL);
+
+  systimer_set_periodic(drv->timer, LOOP_PERIOD_MS);
+
+  nrf_init();
+  nrf_register_listener(NRF_SERVICE_BLE_MANAGER, ble_process_rx_msg);
+  nrf_register_listener(NRF_SERVICE_BLE, ble_process_data);
+  ;
```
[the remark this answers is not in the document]
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/601953eaaa98ac5fe5a9f08814ada0af48bf84de

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/601953eaaa98ac5fe5a9f08814ada0af48bf84de

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: fixed  https://github.com/trezor/trezor-firmware/commit/601953eaaa98ac5fe5a9f08814ada0af48bf84de

core/embed/io/nrf/inc/io/nrf.h
```diff
@@ -0,0 +1,77 @@
...
+// If the queue is full, the message will be dropped
+// returns ID of the message if it was successfully queued, otherwise -1
+int32_t nrf_send_msg(nrf_service_id_t service, const uint8_t *data,
+                     uint32_t len, nrf_tx_callback_t callback, void *context);
+
+// Abort a message by ID
+// If the message is already sent or the id is not found, it does nothing and
+// returns false If the message is queued, it will be removed from the queue If
+// the message is being sent, it will be sent. The callback will not be called.
+bool nrf_abort_msg(int32_t id);
```
mmilata, line 75: Aborting messages introduces a bunch of complexity into the queue code, when are we even going to use it?

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,486 @@
...
+
+  if (!drv->initialized) {
+    return;
+  }
+
+  if (len != BLE_RX_PACKET_SIZE) {
+    return;
+  }
+
+  tsqueue_enqueue(&drv->rx_queue, data, len, NULL);
```
mmilata, line 235: I'm a bit concerned about silently dropping data when a queue is full. Here it's probably OK, we haven't sent application-level ACK yet so Host will time out and resend the packet as if it got lost in the air.

`tx_queue` overflows get reported to the upper layer so that's OK too.

Not sure about `event_queue` - if events get dropped UI will show incorrect BLE state. This could be avoided by UI by ignoring the type of the event and always asking the driver for the entire BLE state, but I think we had this with USB and changed it?

Wondering if we should maintain an overflow counter for every queue in debug builds, along with a function to retrieve them.

core/embed/util/tsqueue/inc/util/tsqueue.h
```diff
@@ -0,0 +1,67 @@
...
+  int32_t id;       // ID of the entry
+  bool used;        // Used flag
+  bool aborted;     // Aborted flag
+} tsqueue_entry_t;
+
+typedef struct {
+  tsqueue_entry_t *entries;  // Array of queue entries
+  int rix;                   // Read index
+  int wix;                   // Write index
+  int qlen;                  // Queue length
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/4448/commits/bece4d87e1656373aaacc7313de75986f5504fbd

core/embed/util/tsqueue/tsqueue.c
```diff
@@ -0,0 +1,192 @@
...
+  queue->qlen = qlen;
+  queue->size = size;
+  queue->next_id = 1;
+
+  for (int i = 0; i < qlen; i++) {
+    if (buffer_mem != NULL) {
+      queue->entries[i].buffer = buffer_mem + i * size;
+      memset(queue->entries[i].buffer, 0, size);
+    }
+    queue->entries[i].len = 0;
```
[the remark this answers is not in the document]
TychoVrahe: used the whole `tsqueue_reset` https://github.com/trezor/trezor-firmware/pull/4448/commits/bece4d87e1656373aaacc7313de75986f5504fbd

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,486 @@
...
+
+  if (drv->mode_requested != BLE_MODE_PAIRING ||
+      drv->mode_current != BLE_MODE_PAIRING) {
+    ble_send_pairing_reject();
+    return;
+  }
+
+  ble_event_t event = {.type = BLE_PAIRING_REQUEST, .data_len = 6};
+  memcpy(event.data, &data[1], 6);
+  tsqueue_enqueue(&drv->event_queue, (uint8_t *)&event, sizeof(event), NULL);
```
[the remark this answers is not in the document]
TychoVrahe: added here https://github.com/trezor/trezor-firmware/pull/4448/commits/bece4d87e1656373aaacc7313de75986f5504fbd

the NRF would otherwise timeout and send the reject itself, but lets not wait as there is no hope in this situation

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,486 @@
...
+static void ble_send_advertising_on(bool whitelist) {
+  uint8_t data[2];
+  data[0] = INTERNAL_CMD_ADVERTISING_ON;
+  data[1] = whitelist ? 1 : 0;
+  nrf_send_msg(NRF_SERVICE_BLE_MANAGER, data, sizeof(data), NULL, NULL);
+}
+
+static void ble_send_advertising_off(void) {
+  uint8_t cmd = INTERNAL_CMD_ADVERTISING_OFF;
+  nrf_send_msg(NRF_SERVICE_BLE_MANAGER, &cmd, sizeof(cmd), NULL, NULL);
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/4448/commits/bece4d87e1656373aaacc7313de75986f5504fbd

Note. returning true here doesn't mean that the state was changed, its more of a signal that the driver has received the command, but it can fail during communication with NRF, or the NRF can misbehave.. So the real UI should instead rely on the events/state in most cases

core/embed/io/nrf/inc/io/nrf.h
TychoVrahe: Probably will when going to low power modes, there will also be other services that might need that. cc also @cepetr

mmilata (approved):
nothing else from me

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,515 @@
...
+
+  tsqueue_init(&drv->rx_queue, drv->rx_queue_entries,
+               (uint8_t *)drv->rx_queue_buffers, BLE_RX_PACKET_SIZE,
+               RX_QUEUE_LEN);
+
+  tsqueue_init(&drv->tx_queue, drv->ts_queue_entries,
+               (uint8_t *)drv->tx_queue_buffers, NRF_MAX_TX_DATA_SIZE,
+               TX_QUEUE_LEN);
+
+  drv->timer = systimer_create(ble_loop, NULL);
```
cepetr, line 328: There are some function calls that may fail. Maybe we could return `bool` and follow  `if (! success) goto cleanup` pattern here.

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,515 @@
...
+               EVENT_QUEUE_LEN);
+
+  tsqueue_init(&drv->rx_queue, drv->rx_queue_entries,
+               (uint8_t *)drv->rx_queue_buffers, BLE_RX_PACKET_SIZE,
+               RX_QUEUE_LEN);
+
+  tsqueue_init(&drv->tx_queue, drv->ts_queue_entries,
+               (uint8_t *)drv->tx_queue_buffers, NRF_MAX_TX_DATA_SIZE,
+               TX_QUEUE_LEN);
+
```
cepetr, line 327: How about passing `drv` as a context to `systimer_create()`. It seems to me ca bit leaner to me.

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,515 @@
...
+}
+
+static bool ble_send_disconnect(void) {
+  uint8_t cmd = INTERNAL_CMD_DISCONNECT;
+  return nrf_send_msg(NRF_SERVICE_BLE_MANAGER, &cmd, sizeof(cmd), NULL, NULL) >=
+         0;
+}
+
+static bool ble_send_pairing_reject(void) {
+  uint8_t cmd = INTERNAL_CMD_REJECT_PAIRING;
```
cepetr, line 106: All `ble_send_xxx()` functions could accept a `ble_driver_t*` argument, as they are internal to the module. This approach could help avoid accessing `g_ble_driver`, which feels somewhat inelegant and, in my opinion, should be reserved for public interface functions where it's necessary.

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,515 @@
...
+    return;
+  }
+
+  systimer_delete(drv->timer);
+
+  tsqueue_reset(&drv->event_queue);
+  tsqueue_reset(&drv->rx_queue);
+  tsqueue_reset(&drv->tx_queue);
+
+  nrf_unregister_listener(NRF_SERVICE_BLE);
```
cepetr, line 352: it seems to me that `nrf_unregister_listener()` should be called before `system_delete()` and `tsqueue_reset()`.

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,515 @@
...
+  nrf_unregister_listener(NRF_SERVICE_BLE_MANAGER);
+
+  drv->initialized = false;
```
cepetr, line 355: Since `nrf_init()` is called in `ble_init()`, `nrf_free()` should be called here.

core/embed/io/ble/stm32/ble.c
```diff
@@ -0,0 +1,515 @@
...
+}
+
+bool ble_can_write(void) {
+  ble_driver_t *drv = &g_ble_driver;
+
+  if (!drv->initialized) {
+    return false;
+  }
+
+  if (!drv->connected || !drv->accept_msgs) {
```
cepetr, line 396: There are a few callbacks accessing driver variables from the interrupt context (e.g., nrf listeners and systic timer). These variables are also accessed by different functions in the thread context, and it seems to me that some form of synchronization is required.

cepetr (changes requested):

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -0,0 +1,766 @@
...
+        drv->rx_idx = 0;
+      }
+    } else if (drv->rx_idx == 1) {
+      // received second byte: LEN
+
+      drv->rx_buffer.msg_len = drv->rx_byte;
+      drv->rx_len = drv->rx_byte;
+
+      if (drv->rx_len > sizeof(uart_packet_t)) {
+        drv->rx_len = 0;
```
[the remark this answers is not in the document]
TychoVrahe: fixed the rx_idx here: https://github.com/trezor/trezor-firmware/commit/7aecc2839812a60218c40ec2efc6b0a8dfa53c34

core/embed/io/ble/stm32/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/7aecc2839812a60218c40ec2efc6b0a8dfa53c34

core/embed/io/ble/stm32/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/7aecc2839812a60218c40ec2efc6b0a8dfa53c34

core/embed/io/ble/stm32/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/7aecc2839812a60218c40ec2efc6b0a8dfa53c34

core/embed/io/ble/stm32/ble.c
cepetr: Calling `nrf_deinit()` when `drv->initialized` is not yet set to `true` doesn't work properly. So it looks like it won't work. 

How about implementing `nrf_deinit()` in a way that allows it to be called on a partially initialized driver? In this particular case, this is very simple and then you can use just `goto cleanup`.

core/embed/io/nrf/stm32u5/nrf.c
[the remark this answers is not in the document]
TychoVrahe: Regarding the second part. The startbyte check will sync with the start message, so unless there are periodically messages that contain such byte we will be ok. This is kinda low possibility but i agree that we should probably strive to find more robust solution. However, can we agree to handle this in separate PR?

core/embed/io/ble/stm32/ble.c
TychoVrahe: I think am only calling `nrf_deinit` when its `drv->initialized` is `true`, so it should work (here we are in ble driver and its `initialized` variable is not relevant for `nrf_deinit`). 

I can still of course modify the nrf drivers deinit the way you suggested, but i want to make sure this is what you meant

core/embed/io/ble/stm32/ble.c
TychoVrahe: Do you see a place where this actually matters? all i could find is these state variables like `drv->connected`, which in thread context are only read and there is not much point of synchronization, since i.e. after i return from this `ble_can_write` the state can change anyway even if i synchronize in this function

core/embed/io/ble/stm32/ble.c
cepetr: Yes, you are right, it works—I made a decision too quickly. So, it's up to you. Personally, I would go with the approach I proposed, as it might be clearer.

core/embed/io/ble/stm32/ble.c
cepetr: You are right that there isn’t much to synchronize. The issues I see are very subtle and likely won’t cause any problems for now. However, this might change in the future if something else changes.

1. `ble_get_state()` accesses multiple variables and returns them in a structure. These variables are modified in the interrupt context, so we can’t guarantee the consistency of the structure’s content.

2. There’s also an issue (which likely won’t show until we enable LTO) with accessing variables that are modified in the background (in the interrupt context) without a memory barrier (e.g., `irq_lock`/`irq_unlock`).

core/embed/io/ble/stm32/ble.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/3075256c27894c170a4b8c611b1b4fe3ee5fc708

core/embed/io/ble/stm32/ble.c
TychoVrahe: as discussed in person, i added synchronization in https://github.com/trezor/trezor-firmware/commit/ed61578f86dbf83044411ddadbbf080229b802cb. This is done in very straightforward way and in every thread function regardless of need.

cepetr (approved):
utACK

PR #4449, opened 2024-12-16
introduce zephyr into firmware monorepo

### LINKED ISSUE #4301: Advertise colour together with name of the device over BT

TychoVrahe:
This PR introduces zephyr, or rather NCS, to firmware monorepo. Zephyr is used for implementing a BLE gateway. 

There is still work to be done, but i would like to get this into main soon, as it will allow more incremental work and ease review of things that affect both core and BLE, like communication and so. 

But this should be good enough to:
- run on both T3W1 boards, the devkit1 and revA
- should be clear from readme itself how to build and flash this.

Reviewers:
@matejcik please check the structure of the added stuff
@cepetr general architectural comments, no need to deep dive into the code, unless you want to.
@hiviah please look at the BLE specific code in `/nordic/trezor/trezor-ble`

fixes #4301


matejcik (approved):

hiviah:
I'll add notes as they come to me.

We seem to have `mcuboot overlay` missing in `nordic/trezor/west.yml`, here is a sample one, we need our own fork of mcuboot.

```
    - name: mcuboot
      url: https://github.com/hiviah/mcuboot
      path: bootloader/mcuboot
      revision: edfe1e1465dbc698bf9a195816247913490ab391
```

Also mcboot itself has several boots inside it, IIRC we are using the `boot_serial` one (need to keep in mind when one changes things and things don't appear to change).

nordic/trezor/README.md
```diff
@@ -0,0 +1,77 @@
...
+```sh
+cd west
+west init -l ./trezor
+```
+
+### Update nRF Connect SDK Modules
+
+Update the modules:
+```sh
+west update
```
hiviah, line 56: Can't yet find a way to have common SDK source so that it's not repeated by each repo clone (since now whole repo build is doubled to some 8 GB). Theoretically only build dir needs to be separate per project, but by  default they clobber over each other.

hiviah (changes requested):
Not tested on device, but otherwise LGTM.

Just make the `nordic/trezor/west.yml` to point to our mcuboot that needs to be created and check it builds (took the rev id from what we are using to build now with 2.6.2) :

    - name: mcuboot
      url: https://github.com/trezor/mcuboot
      path: bootloader/mcuboot
      revision: daf2946a0f07a14b57bd69d29ac4cdde6f810fb7

nordic/trezor/README.md
TychoVrahe: we probably could put the west.yml into firmware repo root. then it would initialize the SDK outside of the firmware repo directory, and you could share it, but potential differences between west.yml might cause headaches. 

But you get the additional SDK bloat only per clone when you actually `west init` and `west update`, which might not be that often

.gitignore
```diff
@@ -11,3 +11,13 @@ __pycache__/
 proto.gv*
 .DS_Store
 crypto/tests/libtrezor-crypto.so.dSYM/
+/nordic/.west/
```
cepetr, line 14: How about to move these into `/nordic/.gitignore` folder?

TychoVrahe:
> We seem to have `mcuboot overlay` missing in `nordic/trezor/west.yml`, here is a sample one, we need our own fork of mcuboot.

Yes, lets do it in separate PR, as we don't have proper repo yet. I created an issue: https://github.com/trezor/trezor-firmware/issues/4585

hiviah (approved):
Still not tested, separate mcuboot TODO later, but OK for now. NCS will shoot us in foot later anyway.

PR #4452, opened 2024-12-17
Zephyr/NCS mcuboot partition table and overlays

hiviah:
Build with mcuboot needs overlay to get the mcuboot there and partition table to tell it where to put it.

There are 2 options for overlays, one is without update, one with update - it's for testing purposes, e.g. you may not want to enter the recovery update until you have it all finished since then boot won't continue.

There is a pin defined in DTS which is read to check whether to enter the update or not. Check the source because I think for some weird reason it may have name you would not expect like led0.

Based on @TychoVrahe 's branch, not main.

Build must be given the overlay via absolute path e.g.:

    west -v build -b t3w1_nrf52833 testapp3 -- -Dmcuboot_OVERLAY_CONFIG=/home/ondro/work/satoshilabs/repos/ncs/testapp3/mcuboot.conf


hiviah:
Reference to older NCS build where bootloader and update was known working, along with BLE debug - https://github.com/trezor/trezor-firmware/tree/omikle/T3W1/devkit1_with_zephyr_with_bootloader - see `core/embed/ble_zephyr_fw` directory. Look at `BUILD.md` and weep.

TychoVrahe:
Cherry-picked the necessary changes into https://github.com/trezor/trezor-firmware/pull/4449 , so i am closing this.

Note that i also switched from using child image to newer sysbuild there, so the build command is eventually different (added that to readme in that branch).

PR #4454, opened 2024-12-19
Suspend to low power mode on T3W1

cepetr:
This PR improves the handling of suspending and resuming to/from low power mode on T3W1, aiming to achieve the lowest possible power consumption.

1. It enhances the initialization and deinitialization of the following drivers to enable achieving the lowest power consumption in STOP2 mode:
   - `ltd_dsi` display driver
   - `backlight_pam`
   - `drv2625` driver
   - `button` driver
   - `dma2d` driver
   - `optiga` driver

2. Adds new POWERCTL_SUSPEND  syscall 

3. Introduces a new `USE_USB` feature utilized in the boardloader code.

4. Adds a return value for the `display_init()` function, although it is not yet used anywhere.


core/embed/io/display/ltdc_dsi/panels/lx200d2406a/lx200d2406a.c
```diff
@@ -8,6 +8,8 @@
 
 #include "../../display_internal.h"
 
+#pragma GCC optimize("-O0")
```
TychoVrahe, line 11: forgotten from debug?

core/embed/io/display/ltdc_dsi/display_driver.c
```diff
@@ -1,3 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma GCC optimize("-O0")
```
TychoVrahe, line 20: debug?

core/embed/sys/powerctl/stm32u5/powerctl_suspend.c
```diff
@@ -93,6 +95,10 @@ void powerctl_suspend(void) {
       // immediately after exiting STOP2 mode.
       irq_key_t irq_key = irq_lock();
 
+      // Enable PWR peripheral clock
+      // (required by the following HAL_PWREx_EnterSTOP2Mode)
+      __HAL_RCC_PWR_CLK_ENABLE();
```
TychoVrahe, line 100: maybe we should disable after waking up? (it is enabled again in SystemInit and then disabled, so end effect zero, but it is not very clear from this function here)

core/embed/upymod/modtrezorio/modtrezorio-poll.h
```diff
@@ -150,6 +154,15 @@ STATIC mp_obj_t mod_trezorio_poll(mp_obj_t ifaces, mp_obj_t list_ref,
 #if USE_BUTTON
       else if (iface == BUTTON_IFACE) {
         const uint32_t evt = button_get_event();
+
+        // --- experimental ---
+#ifdef USE_POWERCTL
```
TychoVrahe, line 159: shouldn't go into main

core/embed/sec/optiga/stm32/optiga_hal.c
```diff
@@ -38,6 +38,24 @@ void optiga_hal_init(void) {
   hal_delay(20);
 }
 
+void optiga_hal_deinit(void) {
+  GPIO_InitTypeDef GPIO_InitStructure = {0};
+
+  GPIO_InitStructure.Mode = GPIO_MODE_ANALOG;
```
TychoVrahe, line 44: As setting pins as ANALOG when shutting down or going to low power is frequently repeating, maybe we could introduce some helper function for this, maybe in powerctl driver or introduce a gpio driver. Not very important, but it would save some typing and would be easier to understand.

cepetr: Yes, I agree that we could introduce our own GPIO driver with a simplified interface that better fits our needs. This would also resolve the duplicated initialization of GPIO clocks in the RCC and address other issues -- I created a new issue #4463.

core/embed/io/display/ltdc_dsi/panels/lx200d2406a/lx200d2406a.c
cepetr: fixed, see 2357c448899fbb123d5ef3c62dd3220450ace4eb

core/embed/io/display/ltdc_dsi/display_driver.c
cepetr: fixed see, 2357c448899fbb123d5ef3c62dd3220450ace4eb

core/embed/upymod/modtrezorio/modtrezorio-poll.h
cepetr: yes, it sits in the temporary commit, I'll remove it before merge.

core/embed/io/display/ltdc_dsi/display_driver.c
```diff
@@ -49,14 +69,22 @@ static bool display_pll_init(void) {
   return HAL_RCCEx_PeriphCLKConfig(&PLL3InitPeriph) == HAL_OK;
```
TychoVrahe, line 69: for consistency with other sub-inits, we should call deinit if this fails

TychoVrahe (commented):
LTGM - pending known issues:
- display RSOD is blue in boardloader on T3W1 A0 
- shutdown deinits display driver, making the RSOD just flash

core/embed/io/display/ltdc_dsi/display_driver.c
cepetr: fixed, see de09f858eb3470bcafa0edb83e28b2696450c698

core/embed/sys/task/stm32/system.c
```diff
@@ -79,7 +79,43 @@ void system_exit_fatal_ex(const char* message, size_t message_len,
...
+  __HAL_RCC_LTDC_RELEASE_RESET();
+#endif
+
+#ifdef __HAL_RCC_GPDMA1_FORCE_RESET
+  __HAL_RCC_GPDMA1_CLK_DISABLE();
+  __HAL_RCC_GPDMA1_FORCE_RESET();
+  __HAL_RCC_GPDMA1_RELEASE_RESET();
+#endif
+
+#ifdef __HAL_RCC_DMA2_FORCE_RESET
```
TychoVrahe, line 114: why not DMA1 too?

cepetr: > why not DMA1 too?

I haven't added it because we don't use DMA1 on the F4. However, we can reset it as well to be future-proof. Do you think we should do that?

TychoVrahe: i don't think there is much future for F4, but if anyone adds DMA2 for any reason somewhere along the line odds are that he will not know about this so it would be perhaps safer to add it, but i am also fine with omitting it for now.

TychoVrahe (approved):

PR #4457, opened 2024-12-22
re-balance distribution in flash on F4 models

TychoVrahe:
moved the crypto data to flash 1.


cepetr (approved):
**before:**
FLASH1 - 81.5%
FLASH2 - 91.6%

**after:**
FLASH1 - 84.7%
FLASH2 - 88.9%

PR #4459, opened 2024-12-27
Fix Rust Trezor-Client Timeouts

kimlwh:
The Rust client will unilaterally timeout after 100 seconds which causes connection errors when interacting with the device between protocol messages. 100 seconds is borderline for some use cases such as complicated transaction verification. This isn't seemingly a problem with the web or Python libraries so without a keep-alive mechanism this fix at least alleviates the problem with no ill effects.

Can we bump the version with this fix also?

### DIFF
--- a/rust/trezor-client/Cargo.lock
+++ b/rust/trezor-client/Cargo.lock
@@ -690,7 +690,7 @@ dependencies = [
 
 [[package]]
 name = "trezor-client"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "bitcoin",
  "byteorder",
--- a/rust/trezor-client/Cargo.toml
+++ b/rust/trezor-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trezor-client"
-version = "0.1.4"
+version = "0.1.5"
 authors = [
     "Piyush Kumar <piyushkumar2k02@kgpian.iitkgp.ac.in>",
     "joshieDo <ranriver@protonmail.com>",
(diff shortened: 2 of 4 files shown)


matejcik:
Thanks for the report. 100s is obvious nonsense.
Unfortunately, making it 300s is only 3x better nonsense ;)

For a `read` operation, we actually want to wait **indefinitely** -- either specify `READ_TIMEOUT_MS = 0`, or, if you are willing to do it, a retry-loop like [this](https://github.com/trezor/trezor-firmware/blob/main/python/src/trezorlib/transport/webusb.py#L102-L114)) is a better solution.
For a `write` operation, on the other hand, a much shorter timeout (on the order of say 1s) would be more appropriate: if the device is not reading what the host is writing, there is no point in waiting for a long time.

matejcik:
The advantage of a retry-loop is that it periodically gives control back to the app (instead of being stuck inside libusb). Other than that these changes are fine.

In order to merge this, please roll back your branch to commit [6bee83f](https://github.com/trezor/trezor-firmware/pull/4459/commits/6bee83f00e95b3a91da2957f5daf5bf1886f2c2f)

matejcik (approved):

matejcik:
perfect, thank you :100:

kimlwh:
If pushing the crate update is manual could @prusnak do as such if he approves? Thank you.

matejcik:
this broke our CI -- it appears that zero timeout is not allowed (weird, the docs said it is)

i'll try to fix it tomorrow in a separate PR

kimlwh:
Maybe only an issue with the UDP/emulator transport and not the WebUSB transport using `rusb`? The latter is where I was experiencing the issue so the UDP transport could remain unchanged, I only changed it for symmetry.

PR #4461, opened 2024-12-30
fix toif generation from grayscale

TychoVrahe:
fixed wrong width calculation


obrusvit (approved):
This solves the issue with running `toiftool convert` after doing `mogrify -colorspace gray icon.png`

PR #4467, opened 2025-01-02
Fix RSOD "no image buf"

### LINKED ISSUE #4424: TS5 no buff RSOD
**Describe the bug**
See video, before i did anything in web suite device went to RSOD.
It happens to me often recently + already happened to me with 2.8.3 production firmware/device.

**Info:**
 - Suite version: web 24.11.3 (112c22f9893956f2962972a3c2a7cf5e43a073e7)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: 
 - Transport: WebUsbTransport

**Screenshots:**

**2.8.6**

image

**2.8.3**

image

https://github.com/user-attachments/assets/d6fe66b3-517a-45d1-94ba-3dd1de3f0995

matejcik:
fixes #4424 

This is a patch over an underlying issue that causes the previous homescreen to stay in memory. It should be removed when we discover the root cause.

### DIFF
--- a/core/embed/rust/src/ui/shape/bitmap.rs
+++ b/core/embed/rust/src/ui/shape/bitmap.rs
@@ -1,4 +1,4 @@
-use crate::trezorhal::bitblt;
+use crate::{error::Error, trezorhal::bitblt};
 
 use crate::ui::{display::Color, geometry::Offset};
 
@@ -61,10 +61,8 @@ impl<'a> Bitmap<'a> {
         mut size: Offset,
         min_height: Option<i16>,
         buff: &'a [u8],
-    ) -> Option<Self> {
-        if size.x < 0 && size.y < 0 {
-            return None;
-        }
+    ) -> Result<Self, Error> {
+        assert!(size.x >= 0 && size.y >= 0);
 
         let min_stride = match format {
             BitmapFormat::MONO1 => (size.x + 7) / 8,
@@ -101,14 +99,14 @@ impl<'a> Bitmap<'a> {
                 if max_height >= min_height as usize {
                     size.y = max_height as i16;
                 } else {
-                    return None;
+                    return Err(Error::ValueError(c"Buffer too small."));
                 }
             } else {
-                return None;
+                return Err(Error::ValueError(c"Buffer too small."));
             }
         }
 
-        Some(Self {
+        Ok(Self {
             ptr: buff.as_ptr() as *mut u8,
             stride,
             size,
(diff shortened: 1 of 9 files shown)


matejcik:
@cepetr please go over my changes to the Rust part
the Python part is trivial but tagging @obrusvit to check it anyway

@bosomt please do your best to check that the RSOD isn't there anymore :)

core/embed/rust/src/ui/shape/bitmap.rs
```diff
@@ -61,10 +61,8 @@ impl<'a> Bitmap<'a> {
         mut size: Offset,
         min_height: Option<i16>,
         buff: &'a [u8],
-    ) -> Option<Self> {
-        if size.x < 0 && size.y < 0 {
-            return None;
-        }
+    ) -> Result<Self, Error> {
+        assert!(size.x >= 0 && size.y >= 0);
```
matejcik, line 65: pointing out for @cepetr: previously you were checking whether _both_ sizes are `< 0`. i changed this to an assert that disallows _either_ size `< 0`.
but now i'm looking at it and wondering whether something relied on passing `size = Offset(-1, -1)` or something like that?

obrusvit (approved):
utACK on Python part.

Proper testing on HW would be good before merging.

bosomt:
Already loading build crom GH action to device
https://github.com/trezor/trezor-firmware/actions/runs/12583218374/job/35070311323

core/embed/rust/src/ui/shape/bitmap.rs
cepetr: > pointing out for @cepetr: previously you were checking whether _both_ sizes are `< 0`. i changed this to an assert that disallows _either_ size `< 0`. but now i'm looking at it and wondering whether something relied on passing `size = Offset(-1, -1)` or something like that?

I don't recall any reason or situation where we might call it with a negative size. Actually, I believe there was a mistake in the condition - it should have been `if size.x < 0 || size.y < 0 {` - so now your assert condition is right. It would be ok to return an `Error` instead of using an `assert`; but both approaches are valid.

cepetr (approved):

bosomt:
**QA OK**

 - Device: Trezor T3T1 2.8.7 regular (revision f2aeea159f1fde98d4fc5575350a8c9fc4478367)

PR #4468, opened 2025-01-02
Randomize SCM_REVISION address in firmware

andrewkozlik:
Implements https://github.com/satoshilabs/trezor-firmware/issues/189.
Replaces https://github.com/satoshilabs/trezor-firmware/pull/190.

### Problem

When a firmware update is loaded onto a fake Trezor device, a malicious bootloader may automatically extract the revision ID of the firmware update. The malicious firmware on the device may then use the extracted revision ID to convince Suite that the firmware was successfully updated when in fact the malicious firmware is running on the device all along. This PR makes it difficult to automatically find the firmware revision ID in the firmware binary file by randomizing the address of the `SCM_REVISION` string in the firmware binary. (The address randomization is deterministic so that deterministic builds are preserved.)

### Solution

1. I defined `SCM_REVISION` as a global symbol in a separate file `core/embed/util/scm_revision/scm_revision.c`, so that it will be placed into a distinct object file separate from all other symbols and constants which could otherwise be used to easily locate it.
2. In `SConscript.firmware` I shuffle the list of object files deterministically based on the value of `SCM_REVISION` itself. Based on observation of the resulting binary files, this results in the linker shuffling the strings in the firmware binary.
    ```py
    random.Random(SCM_REVISION).shuffle(obj_program)
    ```


core/SConscript.kernel
```diff
@@ -296,6 +297,7 @@ ALLPATHS = [
         'embed/util/bl_check/inc',
         'embed/util/image/inc',
         'embed/util/rsod/inc',
+        'embed/util/scm_revision/inc',
```
TychoVrahe, line 300: not needed in kernel anymore

core/SConscript.kernel
```diff
@@ -3,6 +3,7 @@
 
 import os
 import tools, models
+import random
```
TychoVrahe, line 6: not needed in kernel anymore

core/embed/util/scm_revision/inc/util/scm_revision.h
```diff
@@ -0,0 +1,27 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef SCM_REVISION_H
+#define SCM_REVISION_H
+
+#include <stdint.h>
```
TychoVrahe, line 23: we now prefer to use `#include <trezor_types.h>` in C headers of core/embed

andrewkozlik: 883f54f

core/SConscript.kernel
andrewkozlik: https://github.com/trezor/trezor-firmware/commit/883f54f367646bb095164447746da9e1a3c9e039

core/SConscript.kernel
andrewkozlik: https://github.com/trezor/trezor-firmware/commit/883f54f367646bb095164447746da9e1a3c9e039

TychoVrahe (approved):

PR #4469, opened 2025-01-03
change resolution of TOUCH_POWER command parameter to milliseconds

### LINKED ISSUE #4407: Change timeout parameter for TOUCH_POWER prodtest command
1s resolution is too low for production. We should change the timeout parameter to accept milliseconds instead.

TychoVrahe:
As requested by testers.

fixes #4407


cepetr (approved):
utACK

PR #4471, opened 2025-01-06
Show ETH account info on model T

ibz:
This PR is a follow-up of #4175 to show the account info of ETH transactions on model T in addition to TS5.

On TS3 this feature is still missing because we did not find a good way to present it. The problem is that the "info" page currently supports just one title and one set of items, but we want the titles to differ ("Fee info" versus "Account info").

## Screenshots
image

## Test command

`trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d 6162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f706162636465666768696a6b6c6d6e6f70 --max-gas-fee 100 --max-priority-fee 100 -e 2 -g 100 -G 100 -i 1 0x1d1c328764a41bda0492b66baa30c4a339ff85ef 11
`

## Unrelated to the original issue

Two other fixes slipped into this PR as I discovered them while working on that code and they were too small to open separate issues. Made separate commits for them for easy follow-up though.

### Back button fix on TS3

image

### `br_code` consistency

`br_code` should be `SignTx` rather than `Other` while signing Ethereum transactions.

PS: Connect and Suite were notified and they can remove the hack used to support `Other`.


core/embed/rust/src/ui/component/text/op.rs
```diff
@@ -16,7 +16,7 @@ use heapless::Vec;
 
 // So that there is only one implementation, and not multiple generic ones
 // as would be via `const N: usize` generics.
-const MAX_OPS: usize = 20;
+const MAX_OPS: usize = 23;
```
obrusvit, line 19: Not sure if the previous value `20` was determined somehow before, maybe @matejcik can comment.
Anyway, we might be able to reduce it again once we get rid of `font` command from `FormattedText`

obrusvit (approved):
utACK

However, I noticed that the input flow for model T (Bolt) test does not visit all the possible info pages. TS3 Caesar input flow does that automatically which adds the screenshot with newly added info. If you think it's worth, please add it.

core/embed/rust/src/ui/component/text/op.rs
matejcik: 99 % sure it was chosen so that _some particular_ screen fits in the limit.

i'd strongly prefer getting rid of this completely but for now bumping to 23 is fine

obrusvit: > i'd strongly prefer getting rid of this completely but for now bumping to 23 is fine

As we plan to use it more often, do you have an alternative to the fixed-sized `Vec`?

matejcik: `impl OpsSource for T` :woman_shrugging:

ibz:
> TS3 Caesar input flow does that automatically

Actually it did not do it, it just visited the first 2 pages. I changed it so that they both visit all pages now. Please have a look at b9d222e87536022b862f64e580456ca9b653b59f

obrusvit (changes requested):
Let's first solve the requirements from @Hannsek

ibz:
> Let's first solve the requirements from @Hannsek

I think @Hannsek was happy with this 2nd solution - screenshot is updated, but his thumbs down is from the old solution which mixed all under pages with the same title ('Information"). Can you confirm @Hannsek ?

Hannsek:
That's great, thanks for updating it @ibz! 🙏🏻

obrusvit (approved):
Tested ACK

thanks for providing the test command

bosomt:
**QA OK**

`Send from account` and derivation path is present when accessed via (i)
tested on `trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d 6.....`

 - Device: Trezor T3T1 2.8.7 regular (revision f2aeea159f1fde98d4fc5575350a8c9fc4478367)

PR #4477, opened 2025-01-07
build(core): use internal model names everywhere

mmilata:
Replaces `TREZOR_MODEL=T` and `TREZOR_MODEL=R` with `TREZOR_MODEL=T2T1` and `TREZOR_MODEL=T2B1`.


build-docker.sh
```diff
@@ -324,6 +324,7 @@ for BITCOIN_ONLY in ${VARIANTS_legacy[@]}; do
 
   DIRSUFFIX=${BITCOIN_ONLY/1/-bitcoinonly}
   DIRSUFFIX=${DIRSUFFIX/0/}
+  DIRSUFFIX="-T1B1${DIRSUFFIX}"
```
prusnak, line 327: What is this?

mmilata: It adds the model name to the legacy build directory name, so it follows the same pattern as core. Makes the script that prints the fingerprints simpler.

prusnak: Ah yes, nevermind. I thought this was added unconditionally for all models. But now I see we have two build loops. One for core and one for legacy. 👍

prusnak (approved):

core/Makefile
```diff
@@ -50,11 +50,11 @@ OPENOCD_INTERFACE ?= stlink
 # OpenOCD transport default. Alternative: jtag
 OPENOCD_TRANSPORT ?= hla_swd
 
-ifeq ($(TREZOR_MODEL),$(filter $(TREZOR_MODEL),T))
+ifeq ($(TREZOR_MODEL),$(filter $(TREZOR_MODEL),T2T1))
```
obrusvit, line 53: Would it make sense to do something like this for a while for other people?
```makefile
ifeq ($(TREZOR_MODEL),$(filter $(TREZOR_MODEL),T))
  error("Use T2T1 instead")
else ifeq ($(TREZOR_MODEL),$(filter $(TREZOR_MODEL),R))
  error("Use T2B1 instead")
```

prusnak: I wouldn't unless there is a great reason to do so. People should fix their scripts ASAP.

core/embed/io/display/st-7789/display_panel.c
```diff
@@ -243,14 +243,14 @@ void display_panel_reinit(void) {
   } else if (id == DISPLAY_ID_ST7789V) {
     lx154a2411_gamma();
   }
-#elif defined TREZOR_MODEL_T3T1
+#elif defined TREZOR_MODEL_T2T13T1
```
cepetr, line 246: Should be only `TREZOR_MODEL_T3T1`

cepetr: created this #4496

PR #4479, opened 2025-01-08
refactor(core): consistent naming of UI models

obrusvit:
- UI layouts name changes:
  - model_tt -> layout_bolt
  - model_tr -> layout_samson
  - model_mercury -> layout_quicksilver
- rust features `model_xyz` freed for different use, now it's
`layout_xyz`
- input_flow function names are based on UI layout and not internal
model name (i.e. quicksilver instead of t3t1)
- directory names and commentary changed accordingly

[no changelog]


obrusvit:
Needed for FIDO2 tests: https://github.com/trezor/fido2-tests/pull/6

TychoVrahe:
instead of `model_xx`, can we use `ui_xx`? model means something different, and we will need `model_xx` features eventually too as a separate set.

prusnak:
> instead of `model_xx`, can we use `ui_xx`?

I guess this makes more sense. 

Also, maybe I am late to the party - but since you are renaming stuff, have you thought about different naming scheme? American car brands are OK, but maybe we could come up with something more Czech? Cities, fictional characters, etc.
But of course, if we need to do the change ASAP, let's not block the process.

mmilata:
fictional car brands https://mafiagame.fandom.com/wiki/Vehicles_in_Mafia#Vehicles

mmilata (approved):
please also update the fido2-tests submodule

obrusvit:
> instead of model_xx, can we use ui_xx? model means something different, and we will need model_xx features eventually too as a separate set.

Yes, except we have lots of features `ui_something`, e.g. `ui_debug`, `ui_antialiasing`, etc.
How about `ui_impl_xx` or `gui_xx`?

> Also, maybe I am late to the party - but since you are renaming stuff, have you thought about different naming scheme? American car brands are OK, but maybe we could come up with something more Czech? Cities, fictional characters, etc.
But of course, if we need to do the change ASAP, let's not block the process.

Hmm, I wouldn't mind something different but "mercury" and "lincoln" are a part of everyday vocabulary within the team at this point :D

TychoVrahe:
maybe `layout_xx`, which is consistent with python? but everything is fine if its not model :)

mmilata (approved):

PR #4480, opened 2025-01-08
fix(legacy): show XPUB using a QR code

### LINKED ISSUE #3043: Cannot cancel displaying public key on T1
**Describe the bug**
I cannot cancel displaying XPUB on screen of T1. Displaying it on Trezor is an [upcoming feature in Suite](https://github.com/trezor/trezor-suite/issues/3454).

1.12.1, 1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f

T1 (works fine on TT)

**To Reproduce**
1. `trezorctl btc get-descriptor -n 0 -d`
2. Click the left button - nothing happens

**Expected behavior**
Public key is not displayed anymore after clicking "Cancel"

**Screenshot**
Screenshot 2023-05-30 at 13 04 16

romanz:
Similar layout to `fsm_layoutAddress` is used.

Fixes #3043.


prusnak:
Do the all possible XPUBs (bip-44, 49, 84, 86) fit on the device screen?

romanz:
The screenshot below is for
[`xpub6BiVtCpG9fQPxnPmHXG8PhtzQdWC2Su4qWu6XW9tpWFYhxydCLJGrWBJZ5H6qTAHdPQ7pQhtpjiYZVZARo14qHiay2fvrX996oEP42u8wZy` BIP-44 XPUB](https://github.com/trezor/trezor-firmware/blob/d06382e298defc85f69d1bdfd3f7935bf2d94052/tests/device_tests/bitcoin/test_getpublickey.py#L31-L32) and due to its size, it requires two pages of text.

The QR code is rendered with [a higher resolution](https://github.com/trezor/trezor-firmware/blob/d06382e298defc85f69d1bdfd3f7935bf2d94052/legacy/firmware/layout2.c#L921-L930), and looks like it fits on the screen:

image

@prusnak I will make sure that we have tests covering all XPUB types.

prusnak:
> The QR code is rendered with [a higher resolution](https://github.com/trezor/trezor-firmware/blob/d06382e298defc85f69d1bdfd3f7935bf2d94052/legacy/firmware/layout2.c#L921-L930), and looks like it fits on the screen:

Yes, it fits. But I doubt this can be read by any camera. Even QR codes with doubled pixels are quite problematic to be read by most of the cameras from T1 display.

I am not saying we should not merge this PR, I am just saying that doing so probably does not add much value unfortunately. :-/

romanz:
I understand, thanks @prusnak!

If preferable, I think that we can fix the original issue (XPUB layout being non-cancellable) using https://github.com/trezor/trezor-firmware/commit/405a1cbf659ba7fdaf6c3c275afbdb7f5df78895, WDYT?

CC: @Hannsek

prusnak:
> If preferable,

No strong opinion. I am fine with both approaches. Maybe we can show the QR for consistency with GetAddress, but like I said, scanning it might be very problematic.

prusnak:
> Therefore I would proceed with the pr and leave the qr there for consistency.

ACK

matejcik (approved):

bosomt:
tested on my Nothing phone 1 with Android 15
i needed to zoom but QR code was successfully scanned 🎉

PR #4481, opened 2025-01-09
feat(core): Introdure driver for backlight_tps61062 + update prodtest…

kopecdav:
This PR introduce new backlight driver for T3W1 (tps61062). Instead of PWM, backlight is driven with tps buck boost with current loop which provides good stability when device switch between different power sources. Also generate much less emissions. 

To be done:
use HW timer to drive the brightness control instead of GPIO.


TychoVrahe:
superseded by https://github.com/trezor/trezor-firmware/pull/4552

PR #4486, opened 2025-01-10
fix(common): blockbook for testnet 4

karliatto:
As we want to move from Bitcoin Testnet 3 to Bitcoin Testnet 4 we need to change the blockbook default URL.
Related: https://github.com/trezor/trezor-suite/issues/14951

### DIFF
--- a/common/defs/bitcoin/bitcoin_testnet.json
+++ b/common/defs/bitcoin/bitcoin_testnet.json
@@ -12,7 +12,7 @@
   "maxfee_kb": 10000000,
   "minfee_kb": 1000,
   "signed_message_header": "Bitcoin Signed Message:\n",
-  "hash_genesis_block": "000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943",
+  "hash_genesis_block": "00000000da84f2bafbbc53dee25a72ae507ff4914b867c565be350b0da8bf043",
   "xprv_magic": 70615956,
   "xpub_magic": 70617039,
   "xpub_magic_segwit_p2sh": 71979618,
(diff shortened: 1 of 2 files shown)


prusnak:
Do we really want to nuke Testnet 3 support?

If not, it would be better to add a new coin named "Testnet4" or "Testnet 4" and keep the "Testnet" on 3?

matejcik (approved):
utACK

matejcik:
@prusnak firmware doesn't use either of the modified fields, so on firmware side Testnet3 and Testnet4 are identical, nothing is being nuked.

As such, I personally don't care which set of metadata Connect wants to store here, but I don't want to introduce a separate coin that is otherwise 100% identical to an existing one (`coin_name="Testnet"`).

(looking into the future, it would be nice if the "not-needed-by-firmware" data lived elsewhere, but currently it lives here so :woman_shrugging: )

prusnak:
> @prusnak firmware doesn't use either of the modified fields, so on firmware side Testnet3 and Testnet4 are identical, nothing is being nuked.

I understand this. My question should have been "Do we really want to nuke Testnet 3 support **in Suite**?"

matejcik:
then please discuss that at https://github.com/trezor/trezor-suite/issues/14951 :)

matejcik:
@karliatto pls let me know if you want to merge this PR as-is or if something will be done differently

karliatto:
> @karliatto pls let me know if you want to merge this PR as-is or if something will be done differently

Yes, let's merge this PR as-is. It has been decided to replace testnet3 with testnet4 backend.

karliatto:
@matejcik How is https://github.com/trezor/trezor-common getting updated from here? Automatically or does it require some manual process?

matejcik:
there's a nightly sync job

Issue #4489, opened 2025-01-10
Enable Crowdin workflow for firmware translations


obrusvit:
In order to use Crowdin effectively, we need to modify the way we work with language JSONs.

- [x] `en.json` (a source JSON) currently has the following KV pairs: `str` -> `str`. Change it to `str` -> `str | dict` where `dict` allows to use same keys but different strings for different UIs. These changes must be propagated into the generation of `translated_strings.rs`, python mocks and language blobs. This helps with:
  - setting strings of different lengths to different UIs
  - achieving de-duplication of various versions of the same string.
  - Something like this: 
```json
"title__signtx": "Sign transaction"
```
```json
"title__signtx": {"Bolt": "Sign transaction", "Caesar": "Sign tx", "Delizia": "Sign transaction"}
```
- [x] validate that [correct layout names are used](https://github.com/trezor/trezor-firmware/pull/4495#pullrequestreview-2565362115)
- [x] a script that takes the source json of each language and splits it per UI layout, i.e.
  - `en.json` -> `en_Bolt.json`, `en_Caesar.json`, `en_Delizia.json`
  - `de.json` -> `de_Bolt.json`, `de_Caesar.json`, `de_Delizia.json`
  - etc.
- [x] these JSONs are submitted to Crowdin firmware project
  - [x] it's possible to set maximal lengths of strings for individual models
- [x] after translations on Crowdin are finished, the resulting translated strings are merged together into the source JSONs
  - [ ] Optionally use Crowdin API to upload/download JSOns
  - [ ] Optionally perform sanity checks for consistency (e.g. set of keys is matching for all langs for a UI)

obrusvit:
Following up: https://github.com/trezor/trezor-firmware/issues/4867

PR #4495, opened 2025-01-13
Support per-layout translations entries

romanz:
Following #4489.

Also, use layout names for translation font keys.

### DIFF
--- a/python/src/trezorlib/_internal/translations.py
+++ b/python/src/trezorlib/_internal/translations.py
@@ -10,6 +10,7 @@ import construct as c
 from construct_classes import Struct, subcon
 from typing_extensions import Self, TypedDict
 
+from ..debuglink import LayoutType
 from ..firmware.models import Model
 from ..models import TrezorModel
 from ..tools import EnumAdapter, TupleAdapter
@@ -39,7 +40,7 @@ class JsonHeader(TypedDict):
 
 class JsonDef(TypedDict):
     header: JsonHeader
-    translations: dict[str, str]
+    translations: dict[str, str | dict[str, str]]
     fonts: dict[str, JsonFontInfo]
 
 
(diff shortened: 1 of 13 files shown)


romanz:
I will reorganize the commits above, in order to separate the text-modifying edits into a separate commit (which can be dropped - to avoid UI changes).

romanz:
i.e., there is no diff between 8952a989b and 4148e87bc, just reorged the commits' content [by the above force-push](https://github.com/trezor/trezor-firmware/pull/4495#event-15927357328).

romanz:
I have rebased this PR's content over #4516 - pushed here:
https://github.com/trezor/trezor-firmware/compare/obrusvit/ui-layouts-naming-2...romanz/translations-2%5E

Also, I have added the Crowdin split/merge functionality - pushed here:
https://github.com/trezor/trezor-firmware/compare/romanz/translations-2%5E...romanz/translations-2

python/src/trezorlib/_internal/translations.py
```diff
@@ -313,12 +313,12 @@ def blob_from_defs(
 
     translations = TranslatedStrings.from_items(translations_ordered)
 
-    if model.internal_name not in lang_data["fonts"]:
+    if layout_type.name not in lang_data["fonts"]:
         raise ValueError(
-            f"Model {model.internal_name} not found in header for {json_header['language']} v{json_header['version']}"
+            f"Model {layout_type.name} not found in header for {json_header['language']} v{json_header['version']}"
```
obrusvit, line 328: Nit: use `Layout` instead of `Model`

obrusvit (approved):
Thank you. It works as expected, we can now specify different strings per layout and per language. There's one caveat though:
- if we misspell a layout name in `en.json`, then everything is OK because the mistake is caught in compile time :+1: 
- if we misspell a layout name in other lang, e.g. `cs.json` then it fallbacks to english version :thinking: I wonder if we can do something about it now. 

Note that this problem is not new because if we miss the whole string in `cs.json` then we also get fallback to english version without any warning. I'd probably fix this issue separatedly.

python/src/trezorlib/_internal/translations.py
romanz: Thanks - fixed in c5353aec2590005624e1b5f73ede0ae7b9b32b1a.

romanz:
> - if we misspell a layout name in other lang, e.g. cs.json then it fallbacks to english version 🤔 I wonder if we can do something about it now.

Good point, thanks!
Maybe we can add a sanity check to be sure that only "allowed" layout names are used as keys for translations?
For example, we can collect all layout names used in `cs.json` "translations" section and make sure they correspond to a valid layout, using the "fonts" section keys:

image

PR #4501, opened 2025-01-14
fix(common): update solana domains

martykan:
Update Solana backend domains.
[solana1.trezor.io](https://solana1.trezor.io/) and others are dead, replaced by [sol1.trezor.io](https://sol1.trezor.io/)


matejcik:
that's more like it, i was wondering where all the CI results went :)

matejcik (approved):

PR #4502, opened 2025-01-14
minor t3w1/ble related fixes

TychoVrahe:
This PR fixes few issues with T3W1 / BLE:

- NRF driver didn't proprely disable one of the interrupts
- wrong code alignment in trezorlib
- adjustments to codec_v1 were made to accommodate the fact that our BLE implementation uses differently sized RX and TX packets

### DIFF
--- a/core/embed/upymod/modtrezorio/modtrezorio-ble.h
+++ b/core/embed/upymod/modtrezorio/modtrezorio-ble.h
@@ -90,7 +90,7 @@ STATIC mp_obj_t mod_trezorio_BLE_write(mp_obj_t msg) {
 STATIC MP_DEFINE_CONST_FUN_OBJ_1(mod_trezorio_BLE_write_obj,
                                  mod_trezorio_BLE_write);
 
-/// def read(buf: bytes, offset: int = 0) -> int
+/// def read(buf: bytearray, offset: int = 0) -> int:
 ///     """
 ///     Reads message using BLE (device).
 ///     """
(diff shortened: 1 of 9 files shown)


cepetr (approved):
All embedded-related code looks good to me

core/embed/upymod/modtrezorio/modtrezorio-ble.h
```diff
@@ -181,6 +185,12 @@ STATIC mp_obj_t mod_trezorio_BLE_disconnect(void) {
 STATIC MP_DEFINE_CONST_FUN_OBJ_0(mod_trezorio_BLE_disconnect_obj,
                                  mod_trezorio_BLE_disconnect);
 
+/// RX_PACKET_LEN: ClassVar[int]
+/// """Length of one BLE RX packet."""
+
+/// TX_PACKET_LEN: ClassVar[int]
+/// """Length of one BLE TX packet."""
+
```
matejcik, line 193:
```suggestion
/// RX_PACKET_LEN: int
/// """Length of one BLE RX packet."""
/// TX_PACKET_LEN: int
/// """Length of one BLE TX packet."""
```
these should not be `ClassVar`; the reason HID's are a ClassVar is that there's `class HID` encapsulating them.

we prooooobably don't need to introduce a class here now; perhaps in the future if we expose the management of multiple connections

matejcik (changes requested):

matejcik (approved):
ACK, you'll still need to run `make gen` to update the change in the mock files

PR #4504, opened 2025-01-15
add support for setting BLE name from Trezor firmware

TychoVrahe:
Expands the ble_issue_command to support adding data.

Allows sending the name to NRF (this will then be used for advertising and GATT)


core/embed/io/ble/stm32/ble.c
```diff
@@ -80,9 +82,16 @@ static bool ble_send_state_request(ble_driver_t *drv) {
 
 static bool ble_send_advertising_on(ble_driver_t *drv, bool whitelist) {
   (void)drv;
-  uint8_t data[2];
+
+  unit_properties_t props;
+  unit_properties_get(&props);
+
+  uint8_t data[2 + BLE_ADV_NAME_LEN] = {0};
```
cepetr, line 89: `3 + BLE_ADV_NAME_LEN` ?

core/embed/io/ble/inc/io/ble.h
```diff
@@ -37,6 +39,19 @@ typedef enum {
...
+
+typedef union {
+  uint8_t raw[32];
+  uint8_t name[BLE_ADV_NAME_LEN];
+
+} ble_command_data_t;
+
+typedef struct {
+  ble_command_type_t cmd_type;
+  int connection_id;
```
cepetr, line 52: `connection_id` seems to be unused now, isn't it?

cepetr (changes requested):

core/embed/io/ble/stm32/ble.c
TychoVrahe: yes, i added struct for this message https://github.com/trezor/trezor-firmware/commit/f6dc982dcf135ed450c382d27504a6b3b57b9748

core/embed/io/ble/inc/io/ble.h
TychoVrahe: Removed https://github.com/trezor/trezor-firmware/commit/2e5e72056ce2c3af9029a339395a8b13cea62053. 

I will add connection identification later, but it will probably be directly mac address of the central, so no point in keeping this integer.

cepetr (approved):
utACK

PR #4506, opened 2025-01-15
Disallow canceling on uncancelable pages on T3B1

ibz:
This PR fixes an issue that happens when the "middle" button is pressed on a page that should not handle middle or left button presses. If, when releasing the "middle" button (left + right), the left one is the first one to be released, this triggered a left button press which would cancel an otherwise uncancelable page, like this one:

image

To reproduce the bug (before the fix):
 * `trezorctl device setup`
 * keep confirming ever page until you arrive at the page shown in the screenshot above
 * press the left button and keep it pressed
 * while the left button is pressed, press the right button and keep both pressed
 * release the left button
 * now the process is canceled even though it should not be possible to cancel it

Note: this fix affects both the `ButtonController` and the `ButtonPage`. Normally one of the two would be enough to fix the issue, but I decided to fix it in both places that have knowledge of the missing cancel button. This duplication indicates that there is some redundancy in the logic that could be perhaps refactored.

### DIFF
--- a/core/.changelog.d/4500.fixed
+++ b/core/.changelog.d/4500.fixed
@@ -1,0 +1,1 @@
+[T3B1] Fix backup failing if middle button is pressed during confirmation.
--- a/core/embed/rust/src/ui/layout_samson/component/button_controller.rs
+++ b/core/embed/rust/src/ui/layout_samson/component/button_controller.rs
@@ -489,7 +489,11 @@ impl Component for ButtonController {
                             } else {
                                 (
                                     ButtonState::OneReleased(b),
-                                    Some(ButtonControllerMsg::Triggered(b.into(), false)),
+                                    match b {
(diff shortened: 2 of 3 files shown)


matejcik:
yeah, that would be problematic. let's leave it without a test then.

matejcik (approved):
utACK

bosomt:
**QA OK**

- i can hold LB and still navigate via RB when there is not possible to navigate via LB and possible to navigate via RB
- i can hold RB and navigate v LB when there is possible to navigate via RB
- i was not able to replicate issue anymore
- hold to .... works as expected
- middle button works as expected

**Info:**
 - Device: Trezor T3B1 2.8.7 regular (revision dd3bf253f8823fc845aa4c4a8bc2d81cd98b0a7c)

PR #4507, opened 2025-01-16
Remove optional dependency on cryptography

### LINKED ISSUE #4493: Transitive dependency - `cryptography` is no longer optional
I see that cryptography is listed as an optional dependency here: https://github.com/trezor/trezor-firmware/blob/c9b50f3cde94eb55dee400a69a7c20ec9409f720/python/requirements-optional.txt#L5

However, since this change: https://github.com/trezor/trezor-firmware/commit/5d2fa78b3f64324114b15af2d815d8121f4b29c7

`cryptography` is no longer optional because `slip10` has it as a runtime dependency so it got installed anyway.

Was that intentional or expected?

This might cause issues installing on some platforms because cryptography has no wheels for s390x and ppc64le, so users have to compile them there, which requires additional dependencies and tools.

ibz:
See #4493

### DIFF
--- a/python/requirements-optional.txt
+++ b/python/requirements-optional.txt
@@ -2,4 +2,3 @@ hidapi>=0.7.99.post20
 web3>=5
 Pillow>=10
 stellar-sdk>=6
-cryptography>=41


matejcik:
instead of dropping, we want to move it to explicit required dependencies, and adjust the command `trezorctl device authenticate` that has some logic for when cryptography is not present

matejcik:
superseded by #4616

PR #4509, opened 2025-01-16
Make button press cancel current hold

ibz:
On TS3, when holding a button to confirm, pressing the other button now cancels the "hold".

Note: we want to cancel it without an animation rather than with the "shrinking" animation as if the held button was released!

### DIFF
--- a/core/embed/rust/src/ui/layout_samson/component/button_controller.rs
+++ b/core/embed/rust/src/ui/layout_samson/component/button_controller.rs
@@ -215,17 +215,24 @@ impl ButtonContainer {
         self.long_pressed_timer.expire(event)
     }
 
-    /// Registering hold event.
+    /// Registering the hold event
     pub fn hold_started(&mut self, ctx: &mut EventCtx) {
         if let ButtonType::HoldToConfirm(htc) = &mut self.button_type {
             htc.event(ctx, Event::Button(ButtonEvent::HoldStarted));
         }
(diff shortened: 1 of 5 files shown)


core/embed/rust/src/ui/layout_samson/component/button_controller.rs
```diff
@@ -469,6 +476,20 @@ impl Component for ButtonController {
...
+                                match which_down {
+                                    PhysicalButton::Left => {
+                                        self.left_btn.hold_canceled(ctx);
+                                        return None;
+                                    }
+                                    PhysicalButton::Right => {
+                                        self.right_btn.hold_canceled(ctx);
+                                        return None;
+                                    }
+                                    _ => {}
```
bieleluk, line 490: Because the power button does not exist in t2b1 nor t3b1 models, I would suggest returning None even in this case and  remove returning of  the `ButtonState::BothDown`
```suggestion
                                    _ => {
                                        return None;
                                    }
```

bieleluk (approved):
The PR looks good to me.
I only have one suggestion for a better clarity but it does not affect functionality.

core/embed/rust/src/ui/layout_samson/component/button_controller.rs
ibz: OK, just returned None in all cases.

PR #4476, opened 2025-01-07
chore(python): bump cffi to 0.17.1

prusnak:
most probably fix for https://github.com/trezor/trezor-firmware/issues/4474

Issue #4474, opened 2025-01-07
not able to install poetry with python 3.13


bosomt:
```
❯ poetry install
The "poetry.dev-dependencies" section is deprecated and will be removed in a future version. Use "poetry.group.dev.dependencies" instead.
Installing dependencies from lock file

Package operations: 56 installs, 0 updates, 0 removals

  - Installing cffi (1.15.0): Failed

PEP517 build of a dependency failed
```

whole error https://pastebin.com/b9eVASPm

prusnak:
reopened to have the PR #4476 tested by @bosomt

prusnak:
there is now another fix in https://github.com/trezor/trezor-firmware/pull/4513

bosomt:
pillow fixed, 
not sure what to do with trezor-firmware (0.0.0) ;)
i will retest after merge

```
❯ poetry install
The "poetry.dev-dependencies" section is deprecated and will be removed in a future version. Use "poetry.group.dev.dependencies" instead.
Installing dependencies from lock file

Package operations: 35 installs, 0 updates, 0 removals

  - Installing shamir-mnemonic (0.3.0)
  - Installing slip10 (1.0.1)
  - Installing pillow (11.1.0)
  - Installing autoflake (2.3.1): Installing...
  - Installing autoflake (2.3.1)
  - Installing black (24.3.0)
  - Installing binsize (0.1.3)
  - Installing demjson3 (3.0.5)
  - Installing coverage (4.5.4)
  - Installing dominate (2.6.0)
  - Installing fido2 (0.8.1)
  - Installing graphviz (0.19.1)
  - Installing flake8-annotations (3.1.1)
  - Installing flaky (3.8.1)
  - Installing flake8-requirements (2.1.0)
  - Installing hypothesis (6.36.1)
  - Installing inotify (0.2.10)
  - Installing mako (1.1.6)
  - Installing munch (2.5.0)
  - Installing nanopb (0.4.5.post1)
  - Installing pyasn1 (0.4.8)
  - Installing pyserial (3.5)
  - Installing pytest-ordering (0.6)
  - Installing pytest-random-order (1.0.4)
  - Installing pytest-timeout (2.1.0)
  - Installing pytest-xdist (2.5.0)
  - Installing python-bitcoinlib (0.11.0)
  - Installing scan-build (2.0.20)
  - Installing scons (4.8.0)
  - Installing towncrier (23.11.0)
  - Installing tox (3.24.5)
[...]

PR #4513, opened 2025-01-17
chore(python): bump pillow dependency

prusnak:
compat with python 3.13 - see #4474 and https://github.com/trezor/trezor-firmware/pull/4476#issuecomment-2595214828

kills compat with python 3.8 which is EOL now - https://endoflife.date/python

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -973,87 +973,89 @@ files = [
 
 [[package]]
 name = "pillow"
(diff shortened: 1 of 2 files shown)


Fabcien:
I would love to see this merged, is there anything blocking the review ?

matejcik (approved):

PR #4520, opened 2025-01-20
feat(core): increase maximum passphrase length to 128 bytes

### LINKED ISSUE #4084: Remove 50 character limit from passphrase

romanz:
Fixes #4084.

Legacy is still limited to 50 bytes.

Also, deduplicate common passphrase test vectors definitions.


romanz:
Closing instead of #4531

PR #4521, opened 2025-01-20
calculate image hash including padding between header and code

TychoVrahe:
This PR alters the way image hashes are calculated.

Padding between headers and code is now included into hashed area.

In practice, this affects only U5G models, as image and vendor headers are aligned to 512B and such is also is the code alignment on F4 and U585.

T3W1, even though running on U5G, both its image headers and vendor headers are aligned to 1024B, by coincidence, so existing boardloaders and bootloaders should continue to work.

Therefore, only D002 will break and needs reflashing of board+bootloader to work.

Thanks to this change, code alignment can be omitted from trezorctl.

### DIFF
--- a/core/embed/projects/boardloader/main.c
+++ b/core/embed/projects/boardloader/main.c
@@ -181,14 +181,7 @@ static uint32_t check_sdcard(void) {
     _Static_assert(IMAGE_CHUNK_SIZE >= BOOTLOADER_MAXSIZE,
                    "BOOTLOADER IMAGE MAXSIZE too large for IMAGE_CHUNK_SIZE");
 
-    const uint32_t headers_end_offset = hdr->hdrlen;
-    const uint32_t code_start_offset = IMAGE_CODE_ALIGN(headers_end_offset);
-
-    for (uint32_t i = headers_end_offset; i < code_start_offset; i++) {
-      if (((uint8_t *)sdcard_buf)[i] != 0) {
-        return 0;
-      }
-    }
+    const uint32_t code_start_offset = hdr->hdrlen;
 
     if (sectrue !=
         (check_single_hash(hdr->hashes,
--- a/core/embed/util/image/image.c
+++ b/core/embed/util/image/image.c
@@ -263,18 +263,11 @@ secbool check_image_contents(const image_header *const hdr, uint32_t firstskip,
   }
 
   // Check the firmware integrity, calculate and compare hashes
-  size_t offset = IMAGE_CODE_ALIGN(firstskip);
-  size_t end_offset = offset + hdr->codelen;
 
-  // Check area between headers and code
-  uint32_t padding_size = offset - firstskip;
-  const uint8_t *addr =
-      (uint8_t *)flash_area_get_address(area, firstskip, padding_size);
-  for (size_t i = 0; i < padding_size; i++) {
-    if (*addr++ != 0) {
-      return secfalse;
-    }
-  }
+  // check hashes of image chunks
+  // we hash the image including the padding to the end of the area
+  size_t offset = firstskip;
+  size_t end_offset = offset + hdr->codelen;
 
   while (offset < end_offset) {
     size_t bytes_to_check = MIN(IMAGE_CHUNK_SIZE - (offset % IMAGE_CHUNK_SIZE),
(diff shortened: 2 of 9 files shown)


TychoVrahe:
it does apply to bootloader, in trezorlib bootloader uses FirmwareImage struct too, and on embedded side bootloader linker script is modified, along with image content check

core/embed/projects/boardloader/main.c
```diff
@@ -184,14 +184,16 @@ static uint32_t check_sdcard(void) {
...
 
-    for (uint32_t i = headers_end_offset; i < code_start_offset; i++) {
+    for (uint32_t i = headers_end_offset; i < padding_end; i++) {
       if (((uint8_t *)sdcard_buf)[i] != 0) {
         return 0;
       }
     }
 
+    const uint32_t code_start_offset = IMAGE_CODE_ALIGN(headers_end_offset);
+
```
matejcik, line 196: if that's the case, i don't understand this change.

it seems to me that you're (a) still checking explicitly for zero value padding (that's fine i guess) but also `check_single_hash(..., code_start_offset)` which is `IMAGE_CODE_ALIGN(headers_end_offset)`

TychoVrahe: yeah that is a bug https://github.com/trezor/trezor-firmware/pull/4521/commits/40bcbd8da2ed307cad938d43f22a62ee706beb83

core/embed/util/image/image.c
```diff
@@ -264,7 +264,6 @@ secbool check_image_contents(const image_header *const hdr, uint32_t firstskip,
 
   // Check the firmware integrity, calculate and compare hashes
   size_t offset = IMAGE_CODE_ALIGN(firstskip);
```
matejcik, line 266: this should probably be `const size_t something_else = ...` ...

core/embed/util/image/image.c
```diff
@@ -276,6 +275,11 @@ secbool check_image_contents(const image_header *const hdr, uint32_t firstskip,
     }
   }
 
+  // check hashes of image chunks
+  // we hash the image including the padding to the end of the area
+  offset = firstskip;
```
matejcik, line 280: ...and this should be `size_t offset`, otherwise it's rather confusing why the same variable is redefined this way

matejcik (commented):
looked briefly at the C code and it seems ok but I'll leave the main review to @cepetr 

note: any reason to keep the padding-is-zero check? it seems that we don't have to now (contents are signed) and it may bite us later (e.g., if we want to fill empty spaces with random data, we may not be able to)

core/embed/util/image/image.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/c4fe8423870981420e7da280af9b1d5f871fbfe8

core/embed/util/image/image.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/c4fe8423870981420e7da280af9b1d5f871fbfe8

TychoVrahe:
About the zero check - i just wanted to leave the more strict check, to be entirely clear what should be in that space. But randomizing the data would also make some sense, so for consideration, here it is removed: https://github.com/trezor/trezor-firmware/commit/f34d267adb5610717dcb5e949f3c324d8277cfbb

core/embed/projects/boardloader/main.c
```diff
@@ -184,13 +184,7 @@ static uint32_t check_sdcard(void) {
...
 
     const uint32_t headers_end_offset = hdr->hdrlen;
-    const uint32_t code_start_offset = IMAGE_CODE_ALIGN(headers_end_offset);
-
-    for (uint32_t i = headers_end_offset; i < code_start_offset; i++) {
-      if (((uint8_t *)sdcard_buf)[i] != 0) {
-        return 0;
-      }
-    }
+    const uint32_t code_start_offset = headers_end_offset;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/4521/commits/85308d10bd8c9c062dd1725f82d03b18ba21ece5

cepetr (approved):

PR #4523, opened 2025-01-21
Fix displayed batch number in prodtest

andrewkozlik:
Untested.

### DIFF
--- a/core/embed/projects/prodtest/main.c
+++ b/core/embed/projects/prodtest/main.c
@@ -1180,7 +1180,7 @@ int main(void) {
   draw_welcome_screen();
 
   char dom[32];
-  // format: {MODEL_IDENTIFIER}-YYMMDD
+  // format: {MODEL_IDENTIFIER}YYMMDD
   if (sectrue == flash_otp_read(FLASH_OTP_BLOCK_BATCH, 0, (uint8_t *)dom, 32) &&
       sectrue == startswith(dom, MODEL_IDENTIFIER) && dom[31] == 0) {
     gfx_offset_t pos;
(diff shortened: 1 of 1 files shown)


core/embed/projects/prodtest/main.c
```diff
@@ -1180,7 +1180,7 @@ int main(void) {
   draw_welcome_screen();
 
   char dom[32];
-  // format: {MODEL_IDENTIFIER}-YYMMDD
+  // format: {MODEL_IDENTIFIER}YYMMDD
```
TychoVrahe, line 1183: This change contradicts what we have in readme. I also suspect that the original `8` value was due to model T identifier being `TREZOR2`, pluse the `-` makes the shift make some sense. 
 
So where did you observe the `{MODEL_IDENTIFIER}YYMMDD` format? as it might indicate incorrect programming.

andrewkozlik: > I also suspect that the original `8` value was due to model T identifier being `TREZOR2`, pluse the `-` makes the shift make some sense.

Correct.

The comment should either say `{MODEL_IDENTIFIER}YYMMDD` or `{MODEL_INTERNAL_NAME}-YYMMDD`. I can see how it can cause confusion, but it seems to me that technically the first is more correct, since we don't use `TREZOR2` as `MODEL_INTERNAL_NAME`, or do we? For context:

```c
#ifdef TREZOR_MODEL_T2T1
#define MODEL_IDENTIFIER "TREZOR2-"
#else
#define MODEL_IDENTIFIER MODEL_INTERNAL_NAME "-"
#endif
```

TychoVrahe (approved):

PR #4526, opened 2025-01-23
Optimize Delizia-related UI structs' size

romanz:
Using [RUST_PRINT_TYPES_SIZES=1](https://github.com/trezor/trezor-firmware/pull/4526/commits/4602a1dda4ef56c803a94931e0b2e58288235bc9), a memory optimization opportunity has been found in `ui::layout_delizia::component::header::Header` (and other classes encapsulating it):

Screenshot from 2025-01-23 12-19-20

[HeaderFlowMsg.ods](https://github.com/user-attachments/files/18519091/HeaderFlowMsg.ods)


core/embed/rust/src/ui/layout_delizia/component/header.rs
```diff
@@ -61,6 +61,24 @@ impl AttachAnimation {
 
 const BUTTON_EXPAND_BORDER: i16 = 32;
 
+#[derive(Clone, Copy)]
+pub enum HeaderFlowMsg {
```
obrusvit, line 65: I suggest two minor changes:
- Let's rename it to simply `HeaderMsg` - this will be more in-line with what I plan for Eckhart. 
- Let's remove `Confirmed` variant. This one is not used anywhere. Header can only cancel or invoke menu (Info)

obrusvit (commented):
Thanks for the profiling flag :+1: 

One comment regarding the Header.

core/Makefile
```diff
@@ -44,6 +44,7 @@ TREZOR_EMULATOR_DEBUGGABLE ?= 0
 QUIET_MODE ?= 0
 TREZOR_DISABLE_ANIMATION ?= $(if $(filter 0,$(PYOPT)),1,0)
 STORAGE_INSECURE_TESTING_MODE ?= 0
+RUST_PRINT_TYPES_SIZES ?= 1
```
romanz, line 47: The default value should be 0.

core/embed/rust/src/ui/layout_delizia/component/header.rs
romanz: Thanks, fixed in b25d37c2e.

core/Makefile
romanz: Fixed in 50c03b4e8.

obrusvit (approved):

romanz:
Squashing and rebasing over `main`.

PR #4528, opened 2025-01-23
support larger homescreen images on T3W1

TychoVrahe:
This PR introduces model-dependent homescreen max sizes, as we want to allow larger files on T3W1.

Additionally, in trezorlib, jpeg upload is adjusted in order to not modify the jpeg if there is no need.


core/Makefile
```diff
@@ -30,9 +30,9 @@ PRODUCTION ?= 0
 PYOPT      ?= 1
 BITCOIN_ONLY ?= 0
 BOOTLOADER_QA ?= 0
-BOOTLOADER_DEVEL ?= 0
+BOOTLOADER_DEVEL ?= 1
 DISABLE_OPTIGA ?= 0
```
cepetr, line 34: forgotten temporary settings?

cepetr (changes requested):
Tested on T3W1. Looks good. There is still a remaining problem: if the image is resized or transcoded, the resulting image might not fit within the limit. However, this is out of the scope of this PR.

core/Makefile
TychoVrahe: yes reverted https://github.com/trezor/trezor-firmware/pull/4528/commits/eb2291acb5bdebab132dd11b7d137e98fe6530a6

cepetr (approved):

prusnak:
> if the image is resized or transcoded, the resulting image might not fit within the limit.

Yes, we should have a loop that tries do decrease the quality by 5 and re-encode until the image fits within the limit.

> However, this is out of the scope of this PR.

I agree, let's address in some future PR.

TychoVrahe:
Added a separate issue fort that: https://github.com/trezor/trezor-firmware/issues/4583

PR #4529, opened 2025-01-23
fix boot screen background color 

TychoVrahe:
applies to cases when vendor info is shown but no warning is set.

[no changelog]

### DIFF
--- a/core/embed/rust/src/ui/layout_bolt/bootloader/mod.rs
+++ b/core/embed/rust/src/ui/layout_bolt/bootloader/mod.rs
@@ -354,7 +354,11 @@ impl BootloaderUI for UIBolt {
         vendor_img: &'static [u8],
         wait: i32,
     ) {
-        let bg_color = if warning { BLD_WARN_COLOR } else { BLD_BG };
+        let bg_color = if warning {
+            BLD_WARN_COLOR
+        } else {
+            WELCOME_COLOR
+        };
 
         display::sync();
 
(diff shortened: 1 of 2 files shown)


core/embed/rust/src/ui/layout_bolt/bootloader/mod.rs
```diff
@@ -354,7 +354,11 @@ impl BootloaderUI for UIBolt {
         vendor_img: &'static [u8],
         wait: i32,
     ) {
-        let bg_color = if warning { BLD_WARN_COLOR } else { BLD_BG };
```
cepetr, line 357: Not sure about this. Shouldn't we add `initial_setup` argument and choose between `BLD_BG` and `WELCOME_COLOR`?

cepetr (changes requested):

core/embed/rust/src/ui/layout_bolt/bootloader/mod.rs
TychoVrahe: I don't think we ever want to show this screen with BLD_BG. 

The screen is shown during every boot if the vendor header requires it (which is only required for the unsafe header for now with the warning flag set, so red background). Regardless of whether the bootloader was entered or whether this was initial setup or not.

The background is more dependent on the vendor image in order to look good. The trezor image is white logo on black background, so it looks terrible on blue screen.

So maybe it would be appropriate for the vendor header to specify what color should the background be, but that is outside of scope here

cepetr (approved):
utACK

PR #4532, opened 2025-01-23
style: add `make changelog_style` for automatic changelog formatting

romanz:
Added also to `make style`.

[no changelog]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ C_FILES =  $(shell find . -type f -name '*.[ch]' | grep -f ./tools/style.c.inclu
 
 style_check: pystyle_check ruststyle_check cstyle_check changelog_check yaml_check docs_summary_check editor_check ## run all style checks
 
-style: pystyle ruststyle cstyle ## apply all code styles (C+Rust+Py)
+style: pystyle ruststyle cstyle changelog_style ## apply all code styles (C+Rust+Py+Changelog)
 
 pystyle_check: ## run code style check on application sources and tests
 	flake8 --version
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -62,6 +62,17 @@ changelog_check: ## check changelog format
...
+changelog_style: ## fix changelog format
+	./tools/generate-changelog.py --fix-style core
+	./tools/generate-changelog.py --fix-style core/embed/projects/boardloader
+	./tools/generate-changelog.py --fix-style core/embed/projects/bootloader
+	./tools/generate-changelog.py --fix-style core/embed/projects/bootloader_ci
+	./tools/generate-changelog.py --fix-style core/embed/projects/prodtest
+	./tools/generate-changelog.py --fix-style legacy/bootloader
+	./tools/generate-changelog.py --fix-style legacy/firmware
+	./tools/generate-changelog.py --fix-style legacy/intermediate_fw
+	./tools/generate-changelog.py --fix-style python
```
matejcik, line 74: this is becoming rather fragile, instead embed this list into `generate-changelog`
and make it fail if someone tries to use a non-listed path, telling them to also add the thing they want to the known path list

(that way we have a single command for "changelog check" and "changelog style" and we enforce that every changelog used by someone will be added to that list)

romanz: Sounds good, thanks for the suggestion!
WDYT about 01325265a?

romanz: (I've also added @mmilata, since he is the author of `tools/generate-changelog.py`)

tools/changelog.py
```diff
@@ -17,6 +19,23 @@
 INTERNAL_MODELS = ("T2T1", "T2B1", "T3B1", "T3T1", "D001")
 INTERNAL_MODELS_SKIP = ("D001",)
 
+ROOT = Path(__file__).parent.parent
+
+ALLOWED_PROJECTS = (
```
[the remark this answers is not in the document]
romanz: Thanks, done in ec937816d.

mmilata (commented):
Looking good, but also please update references to `generate-changelog.py` in `docs/misc/changelog.md`.

romanz:
> Looking good, but also please update references to `generate-changelog.py` in `docs/misc/changelog.md`.

Thanks! 
Updated in dcb3b2a8c.

mmilata (approved):

romanz:
Squashed and rebased over `main`.

matejcik (approved):

PR #4536, opened 2025-01-24
Refactor fonts handling to rust

obrusvit:
Following https://github.com/trezor/trezor-firmware/pull/4194#pullrequestreview-2337186061

This PR changes the font handling from `enum Font` to direct usage of `struct FontInfo`. While doing so, it became apparent that generating rust files instead of C is a better way go. The main problem was that `const` C variables are bound to `static` variables in Rust (and `static` variables cannot be used in `const` context). Each font is supplied in the following struct:

```rust
pub struct FontInfo {
    pub translation_blob_idx: u16,
    pub height: i16,
    pub max_height: i16,
    pub baseline: i16,
    pub glyph_data: &'static [&'static [u8]],
    pub glyph_nonprintable: &'static [u8],
}

// with a convenient typedef to be passed around similarly to the previous `enum Font`:
pub type Font = &'static FontInfo;
```

## The benefits are:
- getting rid of `enum Font` allows more flexible fonts handling - each layout can now specify the fonts and their operational names on their own
  - for the ease of refactoring and review, the old names were preserved
  - this will mainly benefit Eckhart, which will have more fonts not easily mapped to the `enum Font`
  - to fully use this benefit we need to get rid of using `Font` in 
    - `FormattedText` where it should be replaced by `TextStyle`
    - `Paragraphs` - usage in`Checklist`, which should be separated by layouts or implemented through `FormattedText`
- usage is safer because getting font data is conditioned by using actual variable (we got rid of mapping of `enum` value to some data which might be `null`), this results in removing some `unsafe` blocks and `unwrap`s
- the Rust code is hopefully easier to understand because we got rid of bindings to C
- the font data is now organized based on layouts

## The drawbacks are:
- we need to keep the old font handling for now for `prodtest` and `bootloader_ci`. This will be a private impl.
- we need to preserve mapping to translation blob font IDs - defined for each layout in `core/embed/rust/src/ui/layout_xyz/fonts/mod.rs`

--- 
Measurement of `FLASH` size on `T3T1` showed negligible increase by 1 block only (`512 B`).

TODO:
- [x] the `prodtest` and `bootloader_ci` build does not work yet - trying to have one or two `const font_info_t*`
- [x] figure out having "conflicting" names across layouts for `clippy` compiling with all features
- [x] previously SCons incorporated only some source files based on `stage` (either `"bootloader"` or `"firmware"`). I'm not sure that the current approach gets this right. Only some fonts are used from the bootloader so I pressume that the compiler does not include the others for bootloader binary.
  - a57cde3c2d65d23faa87e1080fe33bf3b662516d
- ~~[ ] ideally get rid of the C fonts completely~~
  - some fonts were removed at least
- [x] should review the use of `USE_RGB_COLOURS`


obrusvit:
Unfortunately, designated initializer is not possible if `FONT_NORMAL` is of type `const font_info_t*` because initializer element is not a compile-time constant.
```c
gfx_text_attr_t normal_text_attr = {
    .font = FONT_NORMAL,
    .fg_color = COLOR_BL_FG,
    .bg_color = COLOR_BL_BG,
};
```

Alternative for `bootloader_ci` and `prodtest` could be either:
- initializing them in `main()`
- reverting to `enum font_id_t` with just `NORMAL` and `BOLD` 
WDYT @TychoVrahe

TychoVrahe:
I think we can revert to `font_id_t` in C code. (and aim for moving the rest of the UI to rust sooon).

While we are structuring the fonts by layouts, maybe we can move the to the existing layout folders, i.e.:

`/ui/layout_bolt/fonts/...`

obrusvit:
Moving stuff around resulted in pretty messy fixups so I decided to rebase. Sorry for that.

core/embed/rust/src/ui/component/connect.rs
```diff
@@ -43,7 +43,7 @@ impl Component for Connect {
     }
 
     fn render<'s>(&'s self, target: &mut impl Renderer<'s>) {
-        let font = Font::NORMAL;
+        let font = FONT_NORMAL;
```
TychoVrahe, line 46: should be passed in constructor

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -645,7 +648,7 @@ where
     ) {
         let numeral = uformat!("{}.", n + 1);
         shape::Text::new(base_point, numeral.as_str())
-            .with_font(Font::SUB)
+            .with_font(FONT_SUB)
```
TychoVrahe, line 651: should be passed in constructor

core/embed/rust/src/ui/component/text/paragraphs.rs
```diff
@@ -626,7 +629,7 @@ where
             } else {
                 // current and future tasks - ordinal numbers or icon on current task
                 if self.show_numerals {
-                    let num_offset = Offset::new(4, Font::NORMAL.visible_text_height("1"));
+                    let num_offset = Offset::new(4, FONT_NORMAL.visible_text_height("1"));
```
TychoVrahe, line 632: should be passed in constructor

core/embed/rust/src/ui/shape/text.rs
```diff
@@ -34,7 +34,7 @@ impl<'a> Text<'a> {
             text,
             color: Color::white(),
             alpha: 255,
-            font: Font::NORMAL,
+            font: FONT_NORMAL,
```
TychoVrahe, line 37: should be mandatory parameter and we can get rid of the with_font below

core/embed/rust/src/ui/mod.rs
```diff
@@ -44,3 +44,16 @@ pub type ModelUI = crate::ui::layout_caesar::UICaesar;
 
 #[cfg(feature = "layout_bolt")]
 pub type ModelUI = crate::ui::layout_bolt::UIBolt;
+
+// Re-export fonts for each layout
```
TychoVrahe, line 48: shouldn't be exported outside of layouts

tdl.rs
```diff
@@ -0,0 +1,21 @@
+#[cfg(all(
```
TychoVrahe, line 1: what is this file? some remnant to be deleted?

core/tools/codegen/gen_font.py
```diff
@@ -501,14 +648,69 @@ def _write_h_file(self) -> None:
...
+    "--write-widths",
+    "-w",
+    is_flag=True,
+    default=False,
+    help="Generate character width files",
+)
+@click.option(
+    "--rust",
+    is_flag=True,
+    default=False,
```
TychoVrahe, line 697: should probably be default now

core/src/apps/cardano/layout.py
```diff
@@ -189,11 +189,8 @@ async def show_tx_init(title: str) -> bool:
     should_show_details = await layouts.should_show_more(
         TR.cardano__confirm_transaction,
         (
```
TychoVrahe, line 191: probably you need more parameters for the para array, otherwise what previously was demibold is now normal. is_data selects mono font so perhaps is_title or something

core/src/apps/ethereum/layout.py
```diff
@@ -214,9 +214,9 @@ async def should_show_domain(name: bytes, version: bytes) -> bool:
     domain_version = decode_typed_data(version, "string")
 
     para = (
-        (ui.NORMAL, TR.ethereum__name_and_version),
-        (ui.DEMIBOLD, domain_name),
-        (ui.DEMIBOLD, domain_version),
+        (TR.ethereum__name_and_version, False),
```
TychoVrahe, line 217: also here: probably you need more parameters for the para array, otherwise what previously was demibold is now normal. is_data selects mono font so perhaps is_title or something

TychoVrahe (changes requested):

TychoVrahe:
As for `USE_RGB_COLORS`: i think we still need it, but for fonts i think we should aim to include the BPP parameter into each font file, and decide how to render per font. This will allow us mixed quality font usage, i.e. we might want some 8bpp on better display for example. But its for future i think.

Regarding the stage-only fonts. I guess compiler will eliminate unused font data now, but it is fragile - once you add a font in some shared component, the code will bloat. I would aim for feature-gating the firmware-only fonts with `#[cfg(not(feature = "bootloader"))]`, but it will require resolving the shared components imports and some reorganization of theme, as it is used in bootloader too and contains references to all fonts, so probably we would need to divide it into theme::common and theme::firmware or something like that (like we already have theme::bootloader).

TychoVrahe:
And one more thought about "mapping to translation blob font IDs"
- consider putting the ID into the font generation script and putting it into `FontInfo`. Use the same value when generating transaction json/blob, and use that as source of truth when selecting a foreign glyphs. But i am not entirely sure that this will work as i am not familiar how the translation blobs work

core/embed/rust/src/ui/shape/text.rs
obrusvit: f54236d4246f412883962ed07c90e5f35e770b01

core/embed/rust/src/ui/component/text/paragraphs.rs
obrusvit: f54236d4246f412883962ed07c90e5f35e770b01

core/embed/rust/src/ui/component/text/paragraphs.rs
obrusvit: f54236d4246f412883962ed07c90e5f35e770b01

core/embed/rust/src/ui/component/connect.rs
obrusvit: f54236d4246f412883962ed07c90e5f35e770b01

tdl.rs
obrusvit: exactly, commited by mistake
e5f8ac217ea11724c0e6c3bfe6c2642283ce7666

core/src/apps/cardano/layout.py
obrusvit: I actually did it on purpose to simplify the app logic. I'll discuss the UX effect with @Hannsek . Seemed negligible to me.

obrusvit:
> consider putting the ID into the font generation script and putting it into FontInfo. Use the same value when generating transaction json/blob, and use that as source of truth when selecting a foreign glyphs. But i am not entirely sure that this will work as i am not familiar how the translation blobs work

I think that's a great idea, thanks. For now, I put it in the generator. Having some "source" for these ideas would be next step but I'd do it in another PR.
13e10be45e00200698e49d68b68cd59e718fecb3

As for `USE_RGB_COLORS` and `#[cfg(not(feature = "bootloader"))]`, I'll have some questions.

core/src/apps/cardano/layout.py
obrusvit: I got ACK from @Hannsek that the simplification justifies the UI diff.

core/tools/codegen/gen_font.py
```diff
@@ -468,22 +489,171 @@ def _write_h_file(self) -> None:
                     f"extern const font_info_t Font_{self._name_style_size}_upper_info;\n"
                 )
 
+    # --------------------------------------------------------------------
+    # Rust code generation
```
matejcik, line 493: just a sidenote: this should have been a Mako template.

perhaps it's still worth doing it that way. doesn't have to be part of the global mako generation, just place a `gen_font.mako` in this folder and invoke it explicitly from this script?

core/embed/rust/src/ui/display/font.rs
```diff
@@ -148,56 +160,46 @@ impl GlyphData {
...
-    /// Returns glyph data slice from a raw pointer by reading the header and
-    /// calculating full size.
-    unsafe fn load_glyph_from_ptr(&self, ptr: *const u8) -> &[u8] {
-        unsafe {
-            let header = slice::from_raw_parts(ptr, 2);
-            let full_size = calculate_glyph_size(header);
-            slice::from_raw_parts(ptr, full_size)
-        }
+    fn load_glyph_from_slice<'a>(&'a self, slice: &'a [u8]) -> &'a [u8] {
+        // TODO: can't we just check that `full_size == slize.len()`?
```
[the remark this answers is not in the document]
obrusvit: You're right, we don't need it. Also, the correct size of the glyph is checked in `Glyph::load(..)`
498bd51a5838481a80ac75c110ce302f5544583c

core/tools/codegen/gen_font.py
obrusvit: I will try it. Do you insist on it in this PR or later?

matejcik: it would be good to do it "soon" because if we don't this is going to fester forever

however i also don't want to hold this PR hostage for it.
up to you.

matejcik: it's something worth trying to tell chatgpt to do, perhaps?

obrusvit: Trying, `o3-mini` struggling a bit with the task :D

obrusvit: See a8b956bced9a49d41e7d22bd868fe62fc06a3dde

matejcik:
> Now we use the same `font` for the offset. Previously, it was `FONT_NORMAL` vs. `FONT_SUB`

why?

obrusvit:
> > Now we use the same `font` for the offset. Previously, it was `FONT_NORMAL` vs. `FONT_SUB`
> 
> why?

I am not exactly sure why. Note that `FONT_NORMAL` was previously only used to calculate the offset, probably out of convenience (NORMAL is the first one.). The font for numerals was chosen correctly.

For the new code, I didn't want to introduce a `Font` parameter just to compute the offset. I think the diff is negligible.

matejcik:
> nah, it's to align with the FONT_NORMAL sized text lines

speaking of, the alignment in the new version is messed up
image

obrusvit:
> speaking of, the alignment in the new version is messed up

hmm, you're right, I'll fix it

lapohoda:
The base-line of number in the list and the text should align.

image

obrusvit:
Sorry for that, the UI diff should be fixed now. 

Btw. code-wise `Checklist` calls for refactor.  It should be implemented as a separate `Component` per layout

obrusvit:
For QA @bosomt

To test this contribution, mainly report anything unusual about rendering texts in the firmware in typical scenarios. 

However, I'd ask you to pay attention to the following test. The point is to check that firmware does not have corrupted foreign glyphs (e.g. "á" in czech) when the following conditions hold:
- this PR is included in firmware (presumably version 2.8.9)
- older language blobs (e.g. 2.8.8 or older) are present in the device

In order to test:
1. use a device with previous firmware (let's say 2.8.8 or older)
2. set a foreign language by uploading the respective version language blob (e.g. cs-2.8.8)
3. update the firmware to a new version (presumably 2.8.9)
4. observe any glitches

---
If you encounter a bad glyph in the test, please:
- report the model and the language
- check that you can "fix" it by uploading a newer language blob (presumably 2.8.9)

bosomt:
**QA OK**

deployed to production
during testing unsigned/signed/released firmwares/translations/updated we havent noticed any issues with fonts

PR #4539, opened 2025-01-27
add support for T3W1 rev B

TychoVrahe:
This PR adds support for T3W1 revB PCB and makes this revision default

New RGB LED driver is introduced, as we now use LPTIM to driver the led. However, no low-power features are added at this point. The LPTIM is configured to use HSI, which should probably be turned off when entering low power mode, but this probably falls outside of the led drivers responsibility.reponsibility


core/embed/io/rgb_led/stm32u5/rgb_led_lp.c
```diff
@@ -0,0 +1,213 @@
...
+
+  HAL_LPTIM_PWM_Stop(&drv->tim_1, LPTIM_CHANNEL_1);
+  HAL_LPTIM_PWM_Stop(&drv->tim_3, LPTIM_CHANNEL_1);
+  HAL_LPTIM_PWM_Stop(&drv->tim_3, LPTIM_CHANNEL_2);
+
+  HAL_LPTIM_Counter_Stop(&drv->tim_1);
+  HAL_LPTIM_Counter_Stop(&drv->tim_3);
+
+  memset(drv, 0, sizeof(*drv));
+  drv->initialized = false;
```
cepetr, line 174: `drv->initializes = false` is not necessary here

core/embed/io/rgb_led/stm32u5/rgb_led_lp.c
```diff
@@ -0,0 +1,213 @@
+#ifdef KERNEL_MODE
```
cepetr, line 1: missing license header

cepetr (changes requested):

core/embed/io/rgb_led/stm32u5/rgb_led_lp.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1d6394cb6af528faaa36fc41a85ecdfc15858e8d

core/embed/io/rgb_led/stm32u5/rgb_led_lp.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1d6394cb6af528faaa36fc41a85ecdfc15858e8d

cepetr (approved):
utACK

PR #4540, opened 2025-01-27
fix(core): explicitly delete keyboard layout

obrusvit:
- seems that keyboard LayoutObj was not properly deallocated resulting in MemoryError. Fix by calling the dunder `del` method explicitly fixes the issue
- the problem was on Model T (Bolt), but applying it for all layouts to prevent similar problem


matejcik (approved):
utACK

bosomt:
**QA OK**

 - Device: Trezor T2T1 2.8.8 regular (revision bbcb920213b0ab847a2fedf825eac70652e33279)

PR #4543, opened 2025-01-28
feat(python): check that translations are using correct layout names

romanz:
Following https://github.com/trezor/trezor-firmware/pull/4495#pullrequestreview-2565362115.

[no changelog]

### DIFF
--- a/python/src/trezorlib/_internal/translations.py
+++ b/python/src/trezorlib/_internal/translations.py
@@ -293,6 +293,34 @@ class TranslationsBlob(Struct):
 
 # ====================
 
+# layouts with translation support
+ALL_LAYOUTS = frozenset(LayoutType) - {LayoutType.T1}
+ALL_LAYOUT_NAMES = frozenset(layout.name for layout in ALL_LAYOUTS)
+
+
+def check_blob(lang_data: JsonDef):
+    json_header: JsonHeader = lang_data["header"]
+    lang_version = f"{json_header['language']} v{json_header['version']}"
+
+    font_layout_names = set(lang_data["fonts"].keys())
+    if font_layout_names != ALL_LAYOUT_NAMES:
+        raise ValueError(
+            f"Invalid font layout names for {lang_version}: {font_layout_names}"
+        )
+
+    for key, item in lang_data["translations"].items():
+        if isinstance(item, dict):
+            item_layouts = set(item.keys())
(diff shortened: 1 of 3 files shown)


python/src/trezorlib/_internal/translations.py
```diff
@@ -293,6 +293,29 @@ def build(self) -> bytes:
...
+    json_header: JsonHeader = lang_data["header"]
+    lang_version = f"{json_header['language']} v{json_header['version']}"
+
+    font_layout_names = set(lang_data["fonts"].keys())
+    if font_layout_names != ALL_LAYOUT_NAMES:
+        raise ValueError(
+            f"Invalid font layout names for {lang_version}: {font_layout_names}"
+        )
+
+    for key, item in lang_data["translations"].items():
```
obrusvit, line 311: How about we do something like this to also `raise` when there is a layout name missing?

```python
    for key, item in lang_data["translations"].items():
        if isinstance(item, dict):
            item_layouts = set(item.keys())
            unknown_layouts = item_layouts - ALL_LAYOUT_NAMES
            missing_layouts = ALL_LAYOUT_NAMES - item_layouts

            if unknown_layouts or missing_layouts:
                raise ValueError(
                    f"Invalid translation layouts for {lang_version}: {key}"
                    f"\nUnknown layouts: {unknown_layouts}"
                    f"\nMissing layouts: {missing_layouts}"
                )
```

obrusvit (commented):
One comment, otherwise :+1:

python/src/trezorlib/_internal/translations.py
romanz: Sounds good, thanks!

I'd add a "`list` formatter", because otherwise we may get an error message with inconsistent set types:
```
ValueError: Invalid translation layouts for cs-CZ v2.8.7: address__qr_code
Unknown layouts: {'Eckha'}
Missing layouts: frozenset({'Eckhart'})
```
I think the following format is a bit better:
```
ValueError: Invalid translation layouts for cs-CZ v2.8.7: address__qr_code
Unknown layouts: ['Eckha']
Missing layouts: ['Eckhart']
```

romanz: Fixed in 71701e5d2f35c2d3971cc0ae176f8312a44fd748.

obrusvit (approved):

PR #4544, opened 2025-01-28
adjust jpeg cache size to allow work on larger displays

TychoVrahe:
respect display width.

### DIFF
--- a/core/embed/rust/src/ui/shape/cache/jpeg_cache.rs
+++ b/core/embed/rust/src/ui/shape/cache/jpeg_cache.rs
@@ -1,6 +1,7 @@
 use crate::{
     io::BinaryData,
     ui::{
+        constant::WIDTH,
         geometry::{Offset, Point, Rect},
         shape::{BasicCanvas, Bitmap, BitmapFormat, BitmapView, Canvas, Rgb565Canvas},
     },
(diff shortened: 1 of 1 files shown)


core/embed/rust/src/ui/shape/cache/jpeg_cache.rs
```diff
@@ -9,7 +11,6 @@ use crate::{
 use core::cell::UnsafeCell;
 use trezor_tjpgdec as tjpgd;
 use without_alloc::alloc::LocalAllocLeakExt;
-
```
cepetr, line 12: Deleted by mistake?

core/embed/rust/src/ui/shape/cache/jpeg_cache.rs
```diff
@@ -1,6 +1,8 @@
 use crate::{
     io::BinaryData,
+    ui,
```
cepetr, line 3: Not sure with this. Don't we use `ui::{self` instead of duplicating `ui`?

cepetr (changes requested):

core/embed/rust/src/ui/shape/cache/jpeg_cache.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/6b7e55feb271ad9337622bd5b8bfb2d7ff5c137c

core/embed/rust/src/ui/shape/cache/jpeg_cache.rs
TychoVrahe: not sure why it was even added https://github.com/trezor/trezor-firmware/commit/6b7e55feb271ad9337622bd5b8bfb2d7ff5c137c

cepetr (approved):
utACK

PR #4545, opened 2025-01-28
unify size of enums between rust and C

TychoVrahe:
By default, ARM eabi has `-fshort-enums` switched on, so we need to pass this flag to clang when building bindings for rust to account for this behavior.

For contrast, unix builds don't use this. SDL also enforces building without this flags via static asserts, so we can't use `-fshort-enums` for the whole of unix, so i opted for using different settings for unix and embedded targets.

### DIFF
--- a/core/embed/rust/build.rs
+++ b/core/embed/rust/build.rs
@@ -137,6 +137,7 @@ fn prepare_bindings() -> bindgen::Builder {
     // Pass in correct include paths and defines.
     if is_firmware() {
         clang_args.push("-nostdinc");
+        clang_args.push("-fshort-enums"); // Make sure enums use the same size as in C
 
         // Append gcc-arm-none-eabi's include paths.
         let cc_output = Command::new("arm-none-eabi-gcc")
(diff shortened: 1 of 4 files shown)


core/embed/rust/build.rs
```diff
@@ -277,6 +277,7 @@ fn generate_micropython_bindings() {
 
     // Write the bindings to a file in the OUR_DIR.
     bindings
+        .clang_arg(short_enums()) // make sure enums use the same size as in C
```
mmilata, line 280: Can we put this into `prepare_bindings`? Wouldn't have to repeat it & it already deals with some hw vs emulator issues.

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/275a284b9c1d0bc28611df183c3df47c47f65374

core/embed/rust/build.rs
```diff
@@ -158,6 +159,8 @@ fn prepare_bindings() -> bindgen::Builder {
             .map(|s| format!("-I{}", s.trim()));
 
         bindings = bindings.clang_args(include_args);
+    } else {
+        clang_args.push("fno-short-enums");
```
matejcik, line 163: is the missing dash a typo or intentional?

TychoVrahe: typo https://github.com/trezor/trezor-firmware/commit/e9407a75106124c64d6954cfcaedc7aabc989a5d

mmilata (approved):

PR #4546, opened 2025-01-28
set actuator for T3W1

TychoVrahe:
This PR configures haptic actuator for T3W1.

There still might be some tuning needed when we have final body.


cepetr (approved):
looks good. Unfortunately I don't have rev B yet, so I didn't test it.

PR #4549, opened 2025-01-28
chore(common): add 2.8.7 to releases.json

mmilata:
We need to make this part of the release process, probably either just after making the release branch, or before merging it to `main`.


obrusvit:
I didn't notice this PR before creating this one: https://github.com/trezor/trezor-firmware/pull/4619. 

Apologies.
This PR is now obsolete.

PR #4551, opened 2025-01-29
adjust T3W1 display refresh rate

TychoVrahe:
This PR adjusts T3W1 display refresh rate to be closer to 60Hz.

This is done by adjusting lane byte frequency, as opposite to adjusting porch setting, after discussing this with display supplier.


cepetr (approved):
I did few calculations and it looks good. Tested on D002 and A0 rev  of T3W1.

PR #4554, opened 2025-01-29
Release tooling improvements

mmilata:


matejcik (approved):
utACK

build-docker.sh
```diff
@@ -320,49 +303,51 @@ done
 
 # build legacy
 
-for BITCOIN_ONLY in ${VARIANTS_legacy[@]}; do
+if echo "${MODELS[@]}" | grep T1B1; then
```
prusnak, line 306: do not print the match
```suggestion
if echo "${MODELS[@]}" | grep -q T1B1 ; then
```

obrusvit (approved):

PR #1591, opened 2021-05-03
feat(core/python) Sign a Bitcoin ABC avalanche UTXO at stake

Fabcien:
As part of the development of an implementation of [the avalanche protocol](https://ipfs.io/ipfs/QmUy4jh5mGNZvLkjies1RWM4YuvJh5o2FYopNPVYwrRVGV), Bitcoin ABC requires a proof of stake to be generated in order to participate to the avalanche network.
Lot of coin holders are using a hardware wallet to securely store their coins, and this PR will make it possible for them to use their Trezor T (there is currently no plan to support the Trezor One) to generate such a proof of stake. This is achieved using [the Electrum ABC wallet](https://github.com/Bitcoin-ABC/ElectrumABC) which is compatible with the Trezor hardware wallet.

PR #4559, opened 2025-01-30
feat(crypto): add support for Schnorr crypto (eCash variant)

Fabcien:
Support Schnorr signature and verification defined in https://github.com/Bitcoin-ABC/bitcoin-abc/tree/master/doc/standards/schnorr.md.

The implementation is based on https://github.com/Bitcoin-ABC/secp256k1 as well as the test vectors.

This is a revival of https://github.com/trezor/trezor-firmware/pull/1509 that was reverted due to not being used after https://github.com/trezor/trezor-firmware/pull/1591 was rejected. The only difference is that the source files are now located in a ecash/ folder to avoid confusion with BIP340 Schnorr signatures, and a USE_ECASH flag makes it possible to disable the feature to save space.

Now that eCash is supported on Trezor, supporting this signature scheme would be very useful for eCash users.
It can be leveraged to add avalanche stake signature using the Trezor devices.

### DIFF
--- a/crypto/Makefile
+++ b/crypto/Makefile
@@ -88,6 +88,7 @@ CFLAGS += -DUSE_MONERO=1
 CFLAGS += -DUSE_NEM=1
 CFLAGS += -DUSE_CARDANO=1
 CFLAGS += -DUSE_INSECURE_PRNG=1
+CFLAGS += -DUSE_ECASH=1
 CFLAGS += -DAES_128
 CFLAGS += -DAES_192
 CFLAGS += -DAES_VAR
@@ -134,6 +135,7 @@ SRCS  += tls_prf.c
 SRCS  += hash_to_curve.c
 SRCS  += buffer.c der.c
 SRCS  += elligator2.c
+SRCS  += ecash/schnorr.c
 
 OBJS   = $(SRCS:.c=.o)
 OBJS  += secp256k1-zkp.o
--- a/crypto/README.md
+++ b/crypto/README.md
@@ -13,6 +13,7 @@ These include:
   uses RFC6979 for deterministic signatures)
 - ECDSA public key derivation
 - BIP340 Schnorr signature signing/verifying
+- eCash Schnorr variant signing/verifying
 - Base32 (RFC4648 and custom alphabets)
 - Base58 address representation
 - Ed25519 signing/verifying (also SHA3 and Keccak variants)
--- a/crypto/ecash/schnorr.c
+++ b/crypto/ecash/schnorr.c
@@ -1,0 +1,217 @@
+/**
+ * Copyright (c) 2021 The Bitcoin ABC developers
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the "Software"),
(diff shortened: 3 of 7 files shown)


prusnak:
This PR adds code which is not used anywhere in the firmware.

Fabcien:
> This PR adds code which is not used anywhere in the firmware.

Yes. I want to also revive https://github.com/trezor/trezor-firmware/pull/1591 after this PR, and I split it to make the review easier: because this is crypto it can be more involved to review.
Do you prefer if I submit the whole feature instead ?

prusnak:
I think the position from #1591 has not changed. The coin is irrelevant as ever, adding feature that is needed only for that coin makes very little sense.

BytesOfMan:
> adding feature that is needed only for that coin makes very little sense.

true, adding this support will probably only result in a few hundred trezor sales if nothing changes about XEC 

Some things that could help from trezor perspective

1. Though a less relevant coin than other utxo projects, XEC is actively maintained. For example, an XEC dev recently found and [responsibly disclosed](https://x.com/EfficiencyDOGE/status/1864357823163060316) this dogecoin exploit. Trezor would get the benefit of experienced open source dev attention for its other supported popular (but no dev team) coins, like Dogecoin

2. We have a professional team used to working open source. Because we would use this feature to run our businesses, trezor would not need to worry about maintaining this feature.

3. We are old school crypto users who have been around for 10+ years. I still have trezors from 2014. Past the point of political arguments about different projects -- understandably trezor does not really have an incentive to add a few hundred sales for a small project. But the additional apolitical dev attention of one of the few competent utxo teams outside of BTC would be a nice boost, and we are happy to help where we can.

Thanks 🙏

Kronkmeister:
I've lost $30.000 worth of XEC because there is no way for me to run a staked eCash node using a hardware wallet. XEC might at this point be ranked low but it is still objectively wrong to call the network and the work going into it as "irrelevant". I see you even use the basis of the ABC code for your BTC Schnorr sigs, proving in practice how relevant and useful their work is (among multiple other examples). One major reason for eCash being ranked lower than the leading ghost-chains is the fact that DEXs, CEXs, or services keep dismissing it, reinforcing its low rank and pushing it further down.

Either way, the additional support for XEC variant Schnorr sigs would have helped me have a much better security for my operations that relies on the Electrum wallet honeypot and gets even more justified seeing how ABC's quality code is actually useful to you handling BTC, but also potentially BCH and DOGE and other Satoshi-clients that pale in comparison to XEC in actual value-add and innovation.

Please help us secure our coins. XEC is as relevant as ever, regardless of market sentiment.

Hannsek:
Hey, thanks. We currently won't accept this. However, we are working on a solution, which will enable accepting 3rd party contributions soon. I'd like to talk to you about this. If you're interested please write me an email to jan.setina@satoshilabs.com.

Fabcien:
> Hey, thanks. We currently won't accept this. However, we are working on a solution, which will enable accepting 3rd party contributions soon. I'd like to talk to you about this. If you're interested please write me an email to [jan.setina@satoshilabs.com](mailto:jan.setina@satoshilabs.com).

I just contacted you.

PR #4564, opened 2025-01-31
reorganize boardloader code

TychoVrahe:
This PR reorganizes boardloader code: mainly SD card update functionality is separated into its own file, which allows easier reading of the main code. 

Lazy initialization of display driver is introduced, to allow faster boot times. However, this is now enabled only for some models, as some bootloaders rely on display already being initialized. This is now fixed, as bootloader should always reinitialize the driver (and bootloader should deinitiliaze it on exit, which it luckily does incompletely - otherwise such bootloaders woudn't work properly.


core/embed/projects/boardloader/sd_update.c
```diff
@@ -0,0 +1,169 @@
...
+  ensure(flash_area_write_data(&BOOTLOADER_AREA, 0, sdcard_buf,
+                               IMAGE_HEADER_SIZE + codelen),
+         NULL);
+
+  ensure(flash_lock_write(), NULL);
+
+  term_printf("\ndone\n\n");
+  term_printf("Unplug the device and remove the SD card\n");
+
+  return sectrue;
```
cepetr, line 143: We should conditionally deinitialize the display at the end of `copy_sdcard`.
```
#ifdef FIXED_HW_DEINIT
  display_deinit();
#endif
```

cepetr (changes requested):

core/embed/projects/boardloader/sd_update.c
TychoVrahe: for simplicity, i moved the init deinit to the caller, `sd_update_check_and_update`. Though we end up in `reboot_or_halt_after_rsod` in anyway.

cepetr (approved):

cepetr (approved):

Issue #4401, opened 2024-11-28
Debuglink response can get cut off if a workflow ends


matejcik:
**Describe the bug**
This does not usually happen on emulator because sending data on emulator is very fast.

However, if sending data is _not_ fast, the following race condition may occur:

1. host asks for `DebugLinkGetState`, where the response needs multiple USB packets (this is most screens actually, the layout json is long)
2. firmware starts streaming the `DebugLinkState` response, pausing on every write until the USB link is ready
3. during the pause, wirelink workflow handler is resumed
4. wirelink workflow handler exits and starts sending a Success message, which fits in just one packet
5. Success packet write loop ends (the packet is in an outgoing buffer or w/e)
6. wirelink calls `loop.clear()` to clean out the memory and "reboot" the micropython env
7. this clears out the pending debuglink write task
8. host gets stuck waiting for remaining packets of the `DebugLinkState` response which will never be sent

A hot-fix on host side is to implement a timeout for the DebugLinkState wait, and ignore failures here.

Proper fix on firmware side involves something like registering "units of work" with the workflow manager (and/or loop scheduler), and run `loop.clear()` not after a workflow but instead "when no units of work are pending".
(This approach would also solve the problems that THP is facing with not knowing very well when it's safe to call `loop.clear()`)

mmilata:
On current main + #4375 I seem to be able to reproduce it often with `PYTEST_TIMEOUT=24 pytest -v ../tests/device_tests/ -k "test_load_device_utf"`.

romanz:
Using #4568 seems to resolve the flakiness of `test_load_device_utf` test:
```
 pytest tests/device_tests/test_msg_loaddevice.py::test_load_device_utf -v --count 10
=================================================================== test session starts ====================================================================
platform linux -- Python 3.10.10, pytest-8.3.2, pluggy-1.5.0 -- /home/rzeyde/src/trezor-firmware/.venv/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/rzeyde/src/trezor-firmware/.hypothesis/examples')
Test order randomisation NOT enabled. Enable with --random-order or --random-order-bucket=<bucket_type>
rootdir: /home/rzeyde/src/trezor-firmware
configfile: setup.cfg
[...]

PR #4568, opened 2025-02-02
test(core): don't fetch full `DebugLinkState` by default

romanz:
In case the main workflow is restarting after a `DebugLinkDecision`, sending the response of `DebugLinkGetState` may get interrupted (see #4401).

Let's make the full `DebugLinkGetState` fetching explicit, to avoid the "restart" race condition.

### DIFF
--- a/python/src/trezorlib/debuglink.py
+++ b/python/src/trezorlib/debuglink.py
@@ -64,8 +64,7 @@ if TYPE_CHECKING:
         def __call__(
             self,
             hold_ms: int | None = None,
-            wait: bool | None = None,
-        ) -> "LayoutContent": ...
+        ) -> "None": ...
 
     InputFlowType = Generator[None, messages.ButtonRequest, None]
 
@@ -416,11 +415,10 @@ def _make_input_func(
     def input_func(
         self: "DebugLink",
         hold_ms: int | None = None,
-        wait: bool | None = None,
-    ) -> LayoutContent:
+    ) -> None:
         __tracebackhide__ = True  # for pytest # pylint: disable=W0612
         decision.hold_ms = hold_ms
-        return self._decision(decision, wait=wait)
+        self._decision(decision)
 
     return input_func  # type: ignore [Parameter name mismatch]
 
@@ -442,12 +440,7 @@ class DebugLink:
         self.t1_screenshot_directory: Path | None = None
         self.t1_screenshot_counter = 0
 
-        # Optional file for saving text representation of the screen
-        self.screen_text_file: Path | None = None
-        self.last_screen_content = ""
-
         self.waiting_for_layout_change = False
-        self.layout_dirty = True
 
         self.input_wait_type = DebugWaitType.IMMEDIATE
 
@@ -476,11 +469,6 @@ class DebugLink:
     def layout_type(self) -> LayoutType:
         assert self.model is not None
         return LayoutType.from_model(self.model)
-
-    def set_screen_text_file(self, file_path: Path | None) -> None:
-        if file_path is not None:
-            file_path.write_bytes(b"")
-        self.screen_text_file = file_path
 
     def open(self) -> None:
         self.transport.begin_session()
@@ -543,8 +531,19 @@ class DebugLink:
             raise TrezorFailure(result)
         return result
 
-    def read_layout(self) -> LayoutContent:
-        return LayoutContent(self.state().tokens)
+    def read_layout(self, wait: bool | None = None) -> LayoutContent:
+        """
+        Force waiting for the layout by setting `wait=True`. Force not waiting by
+        setting `wait=False` -- useful when, e.g., you are causing the next layout to be
+        deliberately delayed.
+        """
+        if wait is True:
+            wait_type = DebugWaitType.CURRENT_LAYOUT
+        elif wait is False:
+            wait_type = DebugWaitType.IMMEDIATE
+        else:
+            wait_type = None
+        return LayoutContent(self.state(wait_type=wait_type).tokens)
 
     def wait_layout(self, wait_for_external_change: bool = False) -> LayoutContent:
         # Next layout change will be caused by external event
@@ -558,18 +557,12 @@ class DebugLink:
         obj = self._call(
             messages.DebugLinkGetState(wait_layout=DebugWaitType.NEXT_LAYOUT)
         )
-        self.layout_dirty = True
         if isinstance(obj, messages.Failure):
             raise TrezorFailure(obj)
         return LayoutContent(obj.tokens)
 
     @contextmanager
-    def wait_for_layout_change(self) -> Iterator[LayoutContent]:
-        # set up a dummy layout content object to be yielded
-        layout_content = LayoutContent(
-            ["DUMMY CONTENT, WAIT UNTIL THE END OF THE BLOCK :("]
-        )
-
+    def wait_for_layout_change(self) -> Iterator[None]:
         # make sure some current layout is up by issuing a dummy GetState
         self.state()
 
@@ -579,17 +572,13 @@ class DebugLink:
         # allow the block to proceed
         self.waiting_for_layout_change = True
         try:
-            yield layout_content
+            yield
         finally:
             self.waiting_for_layout_change = False
-            self.layout_dirty = True
 
         # wait for the reply
         resp = self._read()
         assert isinstance(resp, messages.DebugLinkState)
-
-        # replace contents of the yielded object with the new thing
-        layout_content.__init__(resp.tokens)
 
     def reset_debug_events(self) -> None:
         # Only supported on TT and above certain version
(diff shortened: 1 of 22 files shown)


matejcik:
no, let's not do that again :( having to propagate `return_layout=True` everywhere is a major pain and it was super nice to be rid of it after GFL
if we go this route, I'd strongly prefer to _never_ return the layout, and forcing caller to explicitly call `read_layout()` if they need it.

on a separate note, this breaks the "read text representation of screens" feature. but i don't think anyone is actually using it?

another option is to always return the layout, but make the `LayoutContent` lazy. you'd have to track both debuglink and wirelink and invalidate the lazy-load -- raise an exception if someone tries to access it after _any other_ messages have been exchanged, because any wirelink traffic can cause the layout to change, so it's no longer the "return value of that particular decision"

romanz:
> if we go this route, I'd strongly prefer to never return the layout, and forcing caller to explicitly call read_layout() if they need it.

Sounds good, I'll change the code to separate sending `DebugLinkDecision` and fetching the `DebugLinkState` into separate methods.

> this breaks the "read text representation of screens" feature. but i don't think anyone is actually using it?

IIUC, the text representation is collected into `tests/ui_tests/reports/test/screen_text.txt` when `--record-text-layout` is specified, e.g.:

<details>
<summary>
pytest -vvv ../tests/device_tests/binance/test_get_address.py --record-text-layout
</summary>

```
T3B1_en-device_tests-binance-test_get_address.py::test_binance_get_address[m-44h-714h-0h-0-0-bnb1hgm0p7khfk85zpz-68e2cb5a

T3B1_en-device_tests-binance-test_get_address.py::test_binance_get_address[m-44h-714h-0h-0-1-bnb1egswqkszzfc2uq7-1adfb691

T3B1_en-device_tests-binance-test_get_address.py::test_binance_get_address_chunkify_details[m-44h-714h-0h-0-0-bn-59d4996f
        Account:
        BNB #1
        Derivation path:
        m/44'/714'/0'/0/0
        ********************
        -, -, -
        ////////////////////////////////////////////////////////////////////////////////
        bnb1hgm0p7khfk85zpz5v0j8wnej3a90w709vhkdfu
        ********************
        -, -, -
        ////////////////////////////////////////////////////////////////////////////////

T3B1_en-device_tests-binance-test_get_address.py::test_binance_get_address_chunkify_details[m-44h-714h-0h-0-1-bn-c9025900
        Account:
        BNB #1
        Derivation path:
        m/44'/714'/0'/0/1
        ********************
        -, -, -
        ////////////////////////////////////////////////////////////////////////////////
        bnb1egswqkszzfc2uq78zjslc6u2uky4pw46x4rstd
        ********************
        -, -, -
        ////////////////////////////////////////////////////////////////////////////////
```
</details>

Not sure if there is any other tool that depends on it...

romanz:
Manually tested on TS5: https://github.com/trezor/trezor-firmware/issues/4401#issuecomment-2635899551

tests/click_tests/test_lock.py
```diff
@@ -79,11 +80,13 @@ def hold(duration: int) -> None:
...
     if debug.layout_type is LayoutType.Caesar:
-        layout = debug.press_right()
+        debug.press_right()
     else:
-        layout = debug.click(buttons.INFO)
+        debug.click(buttons.INFO)
+    layout = debug.read_layout()
     assert "PinKeyboard" in layout.all_components()
     debug.input("1234")
+    debug.read_layout()  # TODO: is it needed?
```
matejcik, line 89: every `read_layout()` that ignores the response is suspicious.

what happens if you leave it out?

romanz: When there is no `debug.read_layout()` after `debug.input("1234")`, the test fails due to the emulator being still locked:
```
        layout = debug.read_layout()
        assert "PinKeyboard" in layout.all_components()
        debug.input("1234")
    
>       assert device_handler.features().unlocked is True
E       AssertionError: assert False is True
E        +  where False = <Features: {'vendor': 'trezor.io', 'major_version': 2, 'minor_version': 8, 'patch_version': 8, 'device_id': 'D38C19B91...n_width': 240, 'homescreen_height': 240, 'bootloader_locked': True, 'language_version_matches': True, 'optiga_sec': 0}>.unlocked
E        +    where <Features: {'vendor': 'trezor.io', 'major_version': 2, 'minor_version': 8, 'patch_version': 8, 'device_id': 'D38C19B91...n_width': 240, 'homescreen_height': 240, 'bootloader_locked': True, 'language_version_matches': True, 'optiga_sec': 0}> = features()
E        +      where features = <tests.device_handler.BackgroundDeviceHandler object at 0x723c760895a0>.features

tests/click_tests/test_lock.py:90: AssertionError
```
Maybe we can replace it with `debug.synchronize_at(...)`? 
Or just wait for `device_handler.features()` to return the expected `unlocked` value?

matejcik: Riiiiiight because sending input is non-blocking. That was also a huge advantage of doing a GetState after every single input: we'd get host-side return only after that input has already been processed. This way we can "queue" input events on the wire, which is undesirable. Crap.

It's solvable by using read_layout, sure, but _having to solve_ this is very fragile.

If only `DebugLinkDecision` returned a `Success`.......

let's discuss this later

matejcik: (for the record, `synchronize_at` is a thing that shouldn't exist anymore, and if it is needed somewhere, we need to investigate why. but we don't need to solve that here.)

romanz: abb1d59843e8d5ad061734835f066ffb6d52ba83 added a new flag to `DebugLinkGetState`, to allow client-side synchronization with the device (while keeping the `DebugLinkState` response as small as possible). Also, removed most of the `debug.read_layout()` whose return values weren't used (except for `tests/upgrade_tests/recovery_old.py`, since the upgrade tests running old emulators seem to fail otherwise).

e5b0b212964a042ac5e4eada1be5099853298d47 fixes legacy tests by restoring the [original behavior of `_snapshot_core()`](https://github.com/trezor/trezor-firmware/blob/566f82ccd80bcd9d03716753a330f680acaf164b/python/src/trezorlib/debuglink.py#L722) (not sending `DebugLinkGetState` to T1B1 after every decision).

a59d598f0e212c61abc67ff406eca48f22df72ce is a work-around for an issue in `test_pin_cancel`, causing an UI diff (the last 2 screenshots are missing):

<details>
<summary>UI diff</summary>

image

</details>

It seems like a timing-related issue, because the test passes when I run it with `-vs` flag:
```
$ core/emu.py -a -t -c pytest -vs tests/click_tests/test_pin.py::test_pin_cancel --ui=test
```
But it fails without `-vs`:
```
$ core/emu.py -a -t -c pytest tests/click_tests/test_pin.py::test_pin_cancel --ui=test
```

Attaching MicroPython logs for passing and failing runs, as well the relevant diff:
[pass.log](https://github.com/user-attachments/files/18689498/pass.log)
[fail.log](https://github.com/user-attachments/files/18689499/fail.log)
<details>
<summary>Log diff</summary>

image

</details>

matejcik: oooohhh that's a fun one!
the problem here is that `DebugLinkGetState` can survive a loop.clear but a pending `DebugLinkDecision` can't.
in this case there is no synchronisation between the returned failure and the upcoming input event -- which should not be there in the first place, the test is badly written.
...but given that it is what it is, if the input event (""intended"" for the lockscreen) arrives before the failure is sent out, it will be lost. That's an acceptable casualty IMHO and the `read_layout` is a good solution here.

a better solution would be to not send the input event (and change the expected screens), but it doesn't seem worth the hassle, the whole file is ugly and this is not making it any prettier.

just please add a comment about it, something like "the `prepare()` fixture will send a tap after we're done here, so we have to make sure the lockscreen is already up to receive it -- otherwise the input event may get lost in the loop restart"

python/src/trezorlib/debuglink.py
```diff
@@ -558,18 +557,12 @@ def wait_layout(self, wait_for_external_change: bool = False) -> LayoutContent:
         obj = self._call(
             messages.DebugLinkGetState(wait_layout=DebugWaitType.NEXT_LAYOUT)
         )
-        self.layout_dirty = True
         if isinstance(obj, messages.Failure):
             raise TrezorFailure(obj)
         return LayoutContent(obj.tokens)
 
     @contextmanager
-    def wait_for_layout_change(self) -> Iterator[LayoutContent]:
```
matejcik, line 567: i believe this removal isn't (and wasn't?) necessary -- you are asking for the default state, which will by default return the layout content, so you may as well populate it from the response, right?

(that said I'm surprised that this doesn't break any test code -- so kind of likely that nobody is using it?)

tests/click_tests/common.py
```diff
@@ -64,21 +65,23 @@ def tap_to_confirm(debug: "DebugLink") -> LayoutContent:
     elif debug.layout_type is LayoutType.Caesar:
         return debug.read_layout()
     elif debug.layout_type is LayoutType.Delizia:
-        return debug.click(buttons.TAP_TO_CONFIRM)
+        debug.click(buttons.TAP_TO_CONFIRM)
+        return debug.read_layout()
```
matejcik, line 69: please simplify this function :) it's returning `read_layout` in all codepaths, the only thing we need to do here is "if delizia then tap", maybe we don't even need to return the layout? (but i don't know the usage sites, maybe we do)

python/src/trezorlib/debuglink.py
romanz: It indeed seems that no one is using the returned layout - so I removed it, in order to simplify this method:
```
$ git grep wait_for_layout
python/src/trezorlib/debuglink.py:    def wait_for_layout_change(self) -> Iterator[None]:
tests/device_handler.py:        with self.debuglink().wait_for_layout_change():
```

WDYT?

tests/click_tests/common.py
romanz: Good catch, thanks!
It is indeed used only in `test_autolock.py`, so we can indeed simplify it :)
```
$ git grep tap_to_confirm
click_tests/common.py:def tap_to_confirm(debug: "DebugLink") -> LayoutContent:
click_tests/test_autolock.py:from .common import go_next, tap_to_confirm, unlock_gesture
click_tests/test_autolock.py:        layout = tap_to_confirm(debug)
```

romanz: Fixed in 26ab041f4.

tests/click_tests/test_lock.py
romanz: Added a comment in 26ab041f4.

python/src/trezorlib/debuglink.py
matejcik: sure, why not

matejcik (approved):

PR #4570, opened 2025-02-03
fix QR and text display in T3B1 prodtest 

TychoVrahe:
This PR fixes problem with test rendering of 1BPP fonts in C. 

Welcome screen (with QR) in prodtest for T3B1 and T3B1 is fixed to account for smaller display -> smaller QR size, different text position and colors (which are due to different background color).


cepetr (approved):
Tested on T3B1

PR #3463, opened 2023-12-14
Update monero tests binary

mmilata:
Fixes #3128. Work in progress, `gen_trezor_wallet_passphrase`.

PR #4571, opened 2025-02-03
Delizia tap-to-swipe hotfix

matejcik:
This implements tap-to-swipe functionality on Delizia, to avoid problematic swipes.

points of interest:
* [df6f791](https://github.com/trezor/trezor-firmware/pull/4571/commits/df6f79173c566789ff95c740f0a5355e2426adb3) should probably be a separate PR fixing rust-analyzer, but it's here for now :woman_shrugging: 
* https://github.com/trezor/trezor-firmware/pull/4571/commits/a929f5128d55bb66702824f3bd5792963524ac1f is a generic pager impl that we can (and ideally should) use everywhere
* [e234e25](https://github.com/trezor/trezor-firmware/pull/4571/commits/e234e25eb6fe6c5f5bc4e1e5a1f48ba6743d9ee2) introduces `PaginateFull` trait that is using `Pager` instead of just returning a total. this is used throughout Delizia, for now the original `Paginate` is kept for backwards compatibility.

the big changes in above are refactors of `Paragraphs` and `FormattedText` pagination, where we don't have to go from the beginning every time, because we now remember on which _page_ we are (until then we only knew on which _offset_ in source data we are)

[8dc1e12](https://github.com/trezor/trezor-firmware/pull/4571/commits/8dc1e120cff80b13d632b07d72beaa79eb04ee81) is the big hotfix -- introduce a virtual button in `Footer` and respond to clicks to that button by (a) replaying an artificial swipe to an internally swipable component, or (b) returning `FlowMsg::Next` up the stack, for `SwipeFlow` to pick up, which should respond by
* checking that no action has happened in response to `Next`
* checking that an action is registered for swipe-up
* if yes, triggering that action
this way all the flows written to work with swipe-up will magically also work with footer-click

the rest are some patches for some problems.

another important refactor in [cf490d0](https://github.com/trezor/trezor-firmware/pull/4571/commits/cf490d032394a16f8032b5410d92d10b76386386) is that `Frame` now has its own `map()` function that maps only the _content message_ to something. Messages from button(s), either header or footer, are always mapped to themselves.


matejcik:
Monero failure is spurious. Perhaps #3463 will help with those.

Hannsek:
TODO:

- [ ] extend the hold/tap to confirm zone as much as possible - not directly related to this PR but would be nice to tackle it here as it would help with the UX

- [ ]  extend the tap zone on all screens to the bottom half (bottom two thirds) of the diplay, including the footer
- [ ]  replace "swipe up" with "tap to continue"
- [ ]  edit tutorial according to [Figma](https://www.figma.com/design/n43caWJFGJlT167AQcxruu/%E3%80%94-Prod-%E3%80%95Safe-5?node-id=10357-361&t=GLxVhGBEQiAcSWm8-4)

translations:

DE: Zum Fortfahren tippen
FR: Toucher pour continuer
ES: Toca para continuar
PT-BR: Toque para continuar
IT: Tocca per continuare
CS: Klepnutím pokračujte

core/embed/rust/src/ui/util.rs
```diff
@@ -160,6 +160,99 @@ macro_rules! include_res {
 }
 pub(crate) use include_res;
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "debug", derive(ufmt::derive::uDebug))]
+pub struct Pager {
+    current: u16,
```
obrusvit, line 166: To clearly distinguish that `current` is 0-based index and `total` is the count of pages, I suggest either:
- renaming to `current_idx` or `curr_idx`?? Same for `current()` method, maybe `current_idx()`
- documenting it properly

```rust
pub struct Pager {
    /// Zero-indexed current page.
    current: u16,
    /// Total number of pages.
    total: u16,
}
```

core/embed/rust/src/ui/util.rs
```diff
@@ -160,6 +160,99 @@ macro_rules! include_res {
...
+    total: u16,
+}
+
+impl Pager {
+    pub fn new(total: u16) -> Self {
```
obrusvit, line 171: How about `debug_assert!(total > 0);`

obrusvit (commented):
A few comments for now, I'd like to review and test more later on.

obrusvit:
I think we can proceed with the points from @Hannsek 
- regarding the touch zone of the footer, it is now at double the size of the footer (74 px), roughly 1/3 of the screen and IMHO it's not bad.

Hannsek:
> regarding the touch zone of the footer, it is now at double the size of the footer (74 px), roughly 1/3 of the screen and IMHO it's not bad.

Can we do even more?

core/embed/rust/build.rs
```diff
@@ -56,18 +56,34 @@ const DEFAULT_BINDGEN_MACROS_COMMON: &[&str] = &[
...
+    }
+
+    let mut model_dirs: Vec<&str> = vec![];
+    // always include Bolt as the baseline
+    model_dirs.push("../models/T2T1");
+    #[cfg(feature = "model_caesar")]
+    model_dirs.push("../models/T3B1");
+    #[cfg(feature = "model_delizia")]
+    model_dirs.push("../models/T3T1");
+    #[cfg(feature = "model_eckhart")]
```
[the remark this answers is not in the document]
matejcik: fixed & merged separately in #4650

matejcik:
> replace "swipe up" with "tap to continue"

what do we think about replacing the same string id?
it would mean that the instruction is wrong when you have an outdated translation blob

obrusvit:
> what do we think about replacing the same string id?
it would mean that the instruction is wrong when you have an outdated translation blob

Having "Tap to confirm" under `instruction__swipe_up` sounds like something which might bite in the future. Even if it is only for Delizia.
And on the contrary, if we introduce a new ID, then you'd see english version of the new instruction.

I'm in favor of introducing the new one.

matejcik:
> FR: Toucher pour continuer

we translate "Tap to confirm" as "Appuyez pour confirmer", we should probably use the same word here?

> IT: Tocca per continuare

"Tocca e continua", same as "hold to continue" is "Premi e continua" ?




> * [ ]   edit tutorial according to [Figma](https://www.figma.com/design/n43caWJFGJlT167AQcxruu/branch/Fp8X7YKdNy64oO2mTBtgup/%E3%80%94-Prod-%E3%80%95Safe-5?m=auto&node-id=5220-1963&t=yhZgw5cD9AWwesJF-1)

so we're getting rid of "tap to start"?

also we need translations for the new text:
> Tap the lower half of the screen to continue, or swipe down to go back.

matejcik:
> Having "Tap to confirm" under `instruction__swipe_up` sounds like something which might bite in the future. Even if it is only for Delizia.

i mean, we can fully rename the key :woman_shrugging:
but we probably can afford one more empty key

core/embed/rust/src/ui/util.rs
matejcik: https://github.com/trezor/trezor-firmware/pull/4571/commits/88daaceef74b475018698508cb06bb0d2351f77e

matejcik:
> * [ ]  extend the hold/tap to confirm zone as much as possible - not directly related to this PR but would be nice to tackle it here as it would help with the UX
> * [ ]   extend the tap zone on all screens to the bottom half (bottom two thirds) of the diplay, including the footer

these two are the same thing. i tried to extend over the whole screen except header, please test

> * [ ]   replace "swipe up" with "tap to continue"

done

> * [ ]   edit tutorial according to [Figma](https://www.figma.com/design/n43caWJFGJlT167AQcxruu/branch/Fp8X7YKdNy64oO2mTBtgup/%E3%80%94-Prod-%E3%80%95Safe-5?m=auto&node-id=5220-1963&t=yhZgw5cD9AWwesJF-1)

done

obrusvit:
I'll do final pass including the tutorial later on.

Hannsek:
> these two are the same thing. i tried to extend over the whole screen except header, please test

Does swipe down to go back work?

Hannsek:
> extend the hold/tap to confirm zone as much as possible - not directly related to this PR but would be nice to tackle it here as it would help with the UX
>  extend the tap zone on all screens to the bottom half (bottom two thirds) of the diplay, including the footer
> these two are the same thing

They are not. The first one is about the confirmation screen where the middle confirmation button is.

matejcik:
> Does swipe down to go back work?

it seemed to do so for me :)

matejcik:
https://github.com/trezor/trezor-firmware/pull/4571/commits/2893a3f3f9ccf2662948dbbce8947398818b2e68 introduces a generally useful `stop_propagation` flag to Context. honestly, i don't know why we didn't have it before ... probably nobody needed it until now.

https://github.com/trezor/trezor-firmware/pull/4571/commits/a88a96a1b80712f4a79242d12bde0a55bfceab8c extends the tap-to-confirm / hold-to-confirm touch zone to all of "content", that is, everywhere except header & footer.
we could and probably should cover the footer too, but it's kind of messy to do it so unless we need it i'd leave as-is

obrusvit (changes requested):
We have issues:
- tapping "Menu" button sometimes result in black screen, then if you do some actions, it recovers. I noticed that in the tutorial but it's the same probably everywhere. Steps to reproduce, for example:
  - `./emu.py -es`
  - `trezorctl btc sign-message -n "m/86h/0h/0h/0/0" -t taproot "This is a message."`
  - click on the Menu button
- it is very difficult to go back (Swipe down) from Hold to Confirm screen.
  - try with the same command and go to HtC screen and try it
- waiting for translations also

matejcik:
menu is fixed

the change for enlarging tap-to-confirm is reverted (it interfered with swipe-to-go-back)

touch area for "tap to continue" is set to 2/3 screen height

bosomt:
**QA OK**

 - Device: Trezor T3T1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)

PR #4576, opened 2025-02-04
Hardware JPEG decoder 

cepetr:
This PR implements a hardware JPEG decoder for T3W1.

- Implemented a new `gfx/jpegdec` module that wraps the STM32 hardware JPEG decoder.
- Implemented a new `trezorhal/jpegdec.rs` module that provides the JpegDecoder API for Rust.
- Modified `shape::JpegImage` to support the hardware jpeg decoder.
- Additionally, two new features have been introduced - `hw_jpeg_decoder` and `ui_jpeg` - which replace `jpeg` and `ui_jpeg_decoder`.
- Implemented missing blurring for Rgba8888 canvas
- Implemented simple downscaling for Rgba8888->Rgba8888 bitblt copy operation

The simulator is still using the original JPEG decoder implemented in Rust. 
All other models (except T3W1) should not be affected by this PR.

### DIFF
--- a/core/embed/gfx/bitblt/stm32/dma2d_bitblt.c
+++ b/core/embed/gfx/bitblt/stm32/dma2d_bitblt.c
@@ -24,8 +24,27 @@
 
 #include "../dma2d_bitblt.h"
 
-static DMA2D_HandleTypeDef dma2d_handle = {
-    .Instance = (DMA2D_TypeDef*)DMA2D_BASE,
+// Number of DMA2D layers - background (0) and foreground (1)
+#define DMA2D_LAYER_COUNT 2
+
+typedef struct {
+  // Set if the driver is initialized
+  bool initialized;
+  // ST DMA2D driver handle
+  DMA2D_HandleTypeDef handle;
+  // CLUT cache
+  struct {
+    gfx_color32_t c_fg;
+    gfx_color32_t c_bg;
+  } cache[DMA2D_LAYER_COUNT];
+
+  // CLUT is configured according to the cache
+  bool clut_valid;
+
+} dma2d_driver_t;
+
+static dma2d_driver_t g_dma2d_driver = {
+    .initialized = false,
 };
 
 // Returns `true` if the specified address is accessible by DMA2D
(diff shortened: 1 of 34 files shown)


prusnak:
Can you post the results of the benchmark? I.e. how much faster is it than SW decoding on the same chip?

Also are there any limitations of the HW decoder when compared to SW encoding?

cepetr:
> Can you post the results of the benchmark? I.e. how much faster is it than SW decoding on the same chip?
> 
> Also are there any limitations of the HW decoder when compared to SW encoding?

For 380×520 color images, software decoding takes approximately 600 ms, while hardware decoding takes about 10 ms (both include YCbCr to RGB conversion and display presentation).

According to the STM datasheet, the hardware decoder should comply with the JPEG standard and parse all standard JPEG headers. Unfortunately, it cannot decode progressive JPEGs (though the same applies to the software decoder we currently use).

One current limitation of the hardware decoder, compared to the software decoder we use, is that it cannot scale JPEG images. As a result, we cannot display a small preview of the uploaded image. However, this is not a problem if we slightly change the UI.

matejcik:
> One current limitation of the hardware decoder, compared to the software decoder we use, is that it cannot scale JPEG images. As a result, we cannot display a small preview of the uploaded image. However, this is not a problem if we slightly change the UI.

we should be able to scale the image while streaming it -- unless the jpeg decoder is barfing pixels directly into the framebuffer?

cepetr:
> > One current limitation of the hardware decoder, compared to the software decoder we use, is that it cannot scale JPEG images. As a result, we cannot display a small preview of the uploaded image. However, this is not a problem if we slightly change the UI.
> 
> we should be able to scale the image while streaming it -- unless the jpeg decoder is barfing pixels directly into the framebuffer?

Yes, it would definitely be possible - and it wouldn’t be too slow if we simply skipped pixels and used (1/2, 1/4, …) scaling as before. However, it would still be much slower than just copying with DMA2D. You also have to consider that the image quality wouldn’t be perfect for some images due to aliasing.

cepetr:
> > One current limitation of the hardware decoder, compared to the software decoder we use, is that it cannot scale JPEG images. As a result, we cannot display a small preview of the uploaded image. However, this is not a problem if we slightly change the UI.
> 
> we should be able to scale the image while streaming it -- unless the jpeg decoder is barfing pixels directly into the framebuffer?

 Finally, I implemented simple downscaling so that `JpegShape` now works as it did before, without any limitations.

core/embed/gfx/bitblt/stm32/dma2d_bitblt.c
```diff
@@ -24,8 +24,27 @@
...
+    gfx_color32_t c_bg;
+  } cache[DMA2D_LAYER_COUNT];
+
+  // CLUT is configured according to the cache
+  bool clut_valid;
+
+} dma2d_driver_t;
+
+static dma2d_driver_t g_dma2d_driver = {
+    .initialized = false,
```
TychoVrahe, line 47: this actually doesn't work as expected. Most DMA2D operations are still done in firmware, where dma2d driver is now not initialized - so everything is done in software and that slows down the rendering. So we need to call `dma2d_init`/`gfx_bitblt_init` at the start of firmware main too. But that also contains RCC setup, which need to be put into #`ifdef KERNEL`, otherwise it mem manage faults.

(basically this is a mess now and we should aim to put the dma2d into syscalls soon)

TychoVrahe (changes requested):

TychoVrahe:
One more thing i noticed about blurring on T3W1. Currently the corners of the display are bit miscolored, as accessing pixels out of display area actually returns white per GFXMMU settings. I changed that in this PR https://github.com/trezor/trezor-firmware/pull/4602

cepetr:
> One more thing i noticed about blurring on T3W1. Currently the corners of the display are bit miscolored, as accessing pixels out of display area actually returns white per GFXMMU settings. I changed that in this PR #4602

Yes, I didn’t realize that. 0 as GFXMMU is much better, but still, it would be slightly better to change how we draw the blurred JPEG to eliminate these minor artifacts at the corners.

core/embed/gfx/bitblt/stm32/dma2d_bitblt.c
cepetr: Yes, you are right. I fixed it with a temporary hack - see 7410636a187b0aff679ff0f4560b12b1dda0bbf5. We will remove it as soon as we implement `gfx_bitblt` operations as syscalls.

PR #4590, opened 2025-02-07
fix(core): fix optiga pairing issue

cepetr:
This small PR fixes a problem introduced in #4499.

The Optiga secret stored in backup registers was unintentionally cleared during the transition from the bootloader to the firmware.

This fix prevents it from being cleared during all forward transitions from the boardloader → bootloader → kernel.

The bug was not present in any released version.

### DIFF
--- a/core/embed/sys/startup/stm32/sysutils.c
+++ b/core/embed/sys/startup/stm32/sysutils.c
@@ -29,7 +29,8 @@
 #endif
 
 __attribute((naked, noreturn, no_stack_protector)) void call_with_new_stack(
-    uint32_t arg1, uint32_t arg2, new_stack_callback_t callback) {
+    uint32_t arg1, uint32_t arg2, bool clear_bkpregs,
+    new_stack_callback_t callback) {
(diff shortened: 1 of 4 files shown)


core/embed/sys/startup/stm32/sysutils.c
```diff
@@ -29,7 +29,8 @@
 #endif
 
 __attribute((naked, noreturn, no_stack_protector)) void call_with_new_stack(
-    uint32_t arg1, uint32_t arg2, new_stack_callback_t callback) {
+    uint32_t arg1, uint32_t arg2, bool clear_bksram,
```
TychoVrahe, line 32: ins not actually bksram that is cleared but backup registers. bksram is different memory, 2kB block that should be intact, so the variable name and comment below are misleading

cepetr: yes, fixed 3acfd579f1f2fa0e1d8499252f9d6e0b8e4a7a58

TychoVrahe (approved):

PR #4593, opened 2025-02-07
refactor: entirely remove font handling in C 

TychoVrahe:
This PR introduces Rust in prodtest, re-implements prodtest UI in rust. 

After that, we no longer need font and bitmap drawing functionality in C so that is removed entirely.


core/SConscript.bootloader_ci
```diff
@@ -83,7 +82,7 @@ SOURCE_NANOPB = [
     'vendor/nanopb/pb_encode.c',
 ]
 
-ui.init_ui(TREZOR_MODEL, "bootloader_ci", CPPDEFINES_MOD, SOURCE_MOD, RUST_UI_FEATURES)
+ui.init_ui(TREZOR_MODEL, "bootloader_ci", RUST_UI_FEATURES)
```
obrusvit, line 85: I introduced `"bootloader_ci"` stage in order to separate it from `"bootloader"`. 

Re-consider whether we still need it.

TychoVrahe: removed in https://github.com/trezor/trezor-firmware/commit/e1511164aeee12553b96e560ed40d410709fd4a7

obrusvit:
Final blow: https://github.com/trezor/trezor-firmware/pull/4597

core/Makefile
```diff
@@ -32,7 +32,7 @@ BITCOIN_ONLY ?= 0
 BOOTLOADER_QA ?= 0
 BOOTLOADER_DEVEL ?= 0
 DISABLE_OPTIGA ?= 0
-TREZOR_MODEL ?= T2T1
+TREZOR_MODEL ?= T3T1
```
cepetr, line 35: IMHO, we can switch to T3T1 permanently, but I guess that was not the intent.

core/embed/projects/prodtest/cmd/prodtest_display.c
```diff
@@ -59,34 +49,16 @@ static void prodtest_display_bars(cli_t* cli) {
 
   cli_trace(cli, "Drawing %d vertical bars...", color_count);
 
+  prodtest_bars(colors, color_count);
+
+  const char valid_colors[] = {'R', 'G', 'B', 'W', 'r', 'g', 'b', 'w'};
```
cepetr, line 54: It might be slightly better if prodtest_bars() returned a bool indicating whether colors contained only colors. Now the list of valid colors is duplicated.

core/site_scons/ui/__init__.py
```diff
@@ -5,30 +5,33 @@
...
     ui_modules = {
         "delizia": ui_delizia,
         "caesar": ui_caesar,
         "bolt": ui_bolt,
     }
 
-    return ui_modules[models.get_model_ui(layout)]
+    layout = models.get_model_ui(model)
+
+    if layout == "delizia" and stage == "prodtest":
```
cepetr, line 17: I was wondering where the implementation for “delizia” was and finally found this. IMHO, it would be better to have a separate implementation for a few reasons:
	1.	Different QR code size
	2.	Possibly a different font
	3.	A different display border – not rectangular

core/SConscript.bootloader
```diff
@@ -155,7 +154,7 @@ env.Replace(
     '-fstack-protector-strong '
     +  env.get('ENV')["CPU_CCFLAGS"] + CCFLAGS_MOD,
     CCFLAGS_QSTR='-DNO_QSTR -DN_X64 -DN_X86 -DN_THUMB',
-    LINKFLAGS='-T build/bootloader/memory.ld -Wl,--gc-sections -Wl,-Map=build/bootloader/bootloader.map -Wl,--warn-common -Wl,--print-memory-usage',
+    LINKFLAGS=['-Tbuild/bootloader/memory.ld', '-Wl,--gc-sections', '-Wl,-Map=build/bootloader/bootloader.map', '-Wl,--warn-common', '-Wl,--print-memory-usage'],
```
cepetr, line 157: The same thing might be done in SConscript.bootloader_ci.

core/SConscript.prodtest
```diff
@@ -197,6 +202,20 @@ obj_program.extend(env.Object(source=SOURCE_MOD_CRYPTO, CCFLAGS='$CCFLAGS -ftriv
...
+#
+# Rust library
+#
+features = ['prodtest',] + FEATURES_AVAILABLE + RUST_UI_FEATURES
+
+
+rust = tools.add_rust_lib(
+    env,
+    'prodtest',
+    'release',
```
cepetr, line 214: With this change, I see no option to switch between release and development builds. The same issue applies to the firmware, bootloader, and bootloader_emu.

Maybe we could introduce a new global option that can be passed to the Makefile, such as `DEBUG_BUILD`, `C_DEBUG_BUILD`, or `RUST_DEBUG_BUILD`. We can certainly address this in a separate PR.

core/embed/rust/rust_ui_prodtest.h
```diff
@@ -0,0 +1,13 @@
+#include <trezor_types.h>
+
+void prodtest_info(char* id, uint8_t id_len, char* date, uint8_t date_len);
```
cepetr, line 3: These names might be a bit confusing, especially in the context of the `prodtest code`, where the `prodtest_` prefix is used for test functions called from the CLI.

How about unifying them with the bootloader UI functions?

`prodtest_info` -> `screen_prodtest_info`
`prodtest_welcome` -> `screen_prodtest_welcome`
`prodtest_bars` -> `screen_prodtest_bars`
`prodtest_show_text` -> `screen_prodtest_text`
`prodtest_touch` -> `screen_prodtest_touch`
`prodtest_border` -> `screen_prodtest_border`

cepetr (changes requested):

core/site_scons/ui/__init__.py
TychoVrahe: `Delizia` is UI for TS5, and `Bolt` for Model T. There is no need for different UIs in prodtest.

T3W1, which currently also uses `Bolt`, will use `Eckhart` in the end, and that will indeed need separate implementation for the reasons stated above. But lets wait with that until FW team delivers at least skeleton for `Eckhart`.

core/Makefile
TychoVrahe: not intentional, just keep forgetting to revert those: https://github.com/trezor/trezor-firmware/commit/39a33b41cc3aaaabfc307b91277c7f5bb5c34d2f

core/SConscript.bootloader
TychoVrahe: done in https://github.com/trezor/trezor-firmware/commit/54ff7c211241907256185773d8cafcea758cad90

core/SConscript.prodtest
TychoVrahe: i re-added that to bootloader-emu where it originally was: https://github.com/trezor/trezor-firmware/commit/66f5afbb7dc830dd0b748b68fdbdeae6ea3e36a8

lets handle the other build separately, see https://github.com/trezor/trezor-firmware/issues/4603

core/embed/rust/rust_ui_prodtest.h
TychoVrahe: renamed https://github.com/trezor/trezor-firmware/commit/1c37f69a570a0d81f9b83ee2b9df021f0ae0b16b

core/embed/projects/prodtest/cmd/prodtest_display.c
TychoVrahe: One thing that would justify doing that in UI if we were to validate different colors per UI, i.e. TS3 UI would allow only white and black, but even that may be better implemented as separate tests

core/site_scons/ui/__init__.py
cepetr: Yes, I mean "Eckhart" but I wrote "Delizia". Ok, we can wait.

core/embed/projects/prodtest/cmd/prodtest_display.c
cepetr: If you want to continue, you can simplify  it to this:
```
for (size_t i = 0; i < color_count; i++) {
  if (strchr("RGBWrgbw", colors[i]) == NULL) {
    break;
  }
}
```

TychoVrahe: ok lets go with that https://github.com/trezor/trezor-firmware/commit/636ac4bcb5debabc94cc8dfed62f74d7f1e132b5

core/embed/rust/src/ui/layout_caesar/prodtest/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        });
+
+        display::refresh();
+    }
+
+    fn screen_prodtest_bars(colors: &str) {
+        display::sync();
+
+        let num_colors = colors.chars().count();
+        let width = screen().width() / num_colors as i16;
```
cepetr, line 80: If the string is empty, it results in an RSOD (division by zero).

cepetr (changes requested):

core/embed/rust/src/ui/layout_caesar/prodtest/mod.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/4593/commits/5bf985431ca3f00c36c7f43db4866777a078cfb3

cepetr (approved):
Tested on T3W1

PR #4597, opened 2025-02-10
Remove C font generation

obrusvit:
[no changelog]

### DIFF
--- a/core/tools/codegen/gen_font.py
+++ b/core/tools/codegen/gen_font.py
@@ -1,15 +1,12 @@
 #!/usr/bin/env python3
 
-# script used to generate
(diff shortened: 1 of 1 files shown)


core/tools/codegen/gen_font.py
```diff
@@ -315,9 +286,6 @@ def _rs_file_name(self) -> Path:
         )
 
     def write_files(self, gen_c: bool = False) -> None:
```
TychoVrahe, line 288: the gen_c argument is now useless

obrusvit: 19921283b659babf06cba7a0d5e7e90527d8d05c

TychoVrahe (approved):

PR #4599, opened 2025-02-10
test(core): use `TESTOPTS` in `core-hw.yml` run step

romanz:
Otherwise, the problematic tests seem to be running in CI: 
https://github.com/trezor/trezor-firmware/actions/runs/13230566082/job/36927073823

[no changelog]


mmilata (approved):
ohh how did I miss this

PR #4606, opened 2025-02-11
refactor(core): simplify `fade_backlight_duration`

romanz:
- IIUC, casting to and from `i32` can be removed.
- `timer.elapsed()` can be evaluated once per loop iteration.

### DIFF
--- a/core/embed/rust/src/ui/display/mod.rs
+++ b/core/embed/rust/src/ui/display/mod.rs
@@ -41,25 +41,27 @@ pub fn fade_backlight(target: u8) {
 
 #[cfg(feature = "backlight")]
 pub fn fade_backlight_duration(target: u8, duration_ms: u32) {
-    let target = target as i32;
-    let current = backlight() as i32;
+    let current = backlight();
     let duration = Duration::from_millis(duration_ms);
 
     if animation_disabled() {
-        set_backlight(target as u8);
+        set_backlight(target);
         return;
     }
 
     let timer = Stopwatch::new_started();
 
-    while timer.elapsed() < duration {
+    loop {
         let elapsed = timer.elapsed();
-        let val = i32::lerp(current, target, elapsed / duration);
-        set_backlight(val as u8);
+        if elapsed >= duration {
+            break;
+        }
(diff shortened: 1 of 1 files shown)


core/embed/rust/src/ui/lerp.rs
```diff
@@ -66,31 +66,27 @@ macro_rules! impl_lerp_for_unsigned {
 }
 
 impl_lerp_for_signed!(i16);
-impl_lerp_for_signed!(i32);
 impl_lerp_for_unsigned!(u8);
 impl_lerp_for_unsigned!(u16);
-impl_lerp_for_unsigned!(u32);
-
-impl_lerp_for_signed!(f32);
```
matejcik, line 74: any specific reason to remove those? they won't be compiled in when unused, and it seems generally useful to have them available for all types

romanz: You're right - I will revert this change.

romanz: fb6b11b

matejcik (approved):

core/embed/rust/src/ui/display/mod.rs
```diff
@@ -41,25 +41,27 @@ pub fn fade_backlight(target: u8) {
...
 
-    while timer.elapsed() < duration {
+    loop {
         let elapsed = timer.elapsed();
-        let val = i32::lerp(current, target, elapsed / duration);
-        set_backlight(val as u8);
+        if elapsed >= duration {
+            break;
+        }
+        let val = u8::lerp(current, target, elapsed / duration);
```
prusnak, line 59: please test the fading on the device - something tells me that previous approach might have been more smooth, but i might as well be wrong

romanz: I think the fading looks similar (comparing 2.8.7 release and the firmware from this PR on T3T1):

https://github.com/user-attachments/assets/dd8e1202-77ef-40f7-879e-1a940fe4cf26


https://github.com/user-attachments/assets/45e4e1aa-353e-4047-ab67-f7929a3c0fb2

PR #4607, opened 2025-02-11
feat(core/prodtest): Allow prodtest to exit from interactive mode 

kopecdav:
Add an option to exit from prodtest interactive mode with ".+ENTER" command.

If test script or test equipment connects to prodtest serial, it may happen that somebody already switched it to interactive mode before. To make it easier for it, it may switch back to non-interactive mode and don't deal with the colored responses etc. 

Prodtest will confirm the exit command with `OK` in standard mode to be well machine readable.

### DIFF
--- a/core/embed/rtl/cli.c
+++ b/core/embed/rtl/cli.c
@@ -569,6 +569,15 @@ void cli_run_loop(cli_t* cli) {
     }
     cli->empty_lines = 0;
 
+    // Quit interactive mode on `.+ENTER`
+    if ((strcmp(cli->cmd_name, ".") == 0)) {
+      if (cli->interactive) {
+        cli->interactive = false;
+        cli_trace(cli, "Exiting interactive mode...");
+      }
+      continue;
+    }
+
     // Find the command handler
     cli->current_cmd = cli_find_command(cli, cli->cmd_name);
 
--- a/core/embed/projects/prodtest/README.md
+++ b/core/embed/projects/prodtest/README.md
@@ -12,6 +12,8 @@ The prodtest includes a simple text-based command-line interface that can be con
 
 Pressing the ENTER key twice switches the interface to interactive mode, enabling basic line editing, autocomplete functionality (using the TAB key), and text coloring.
 
+To exit from interactive mode type `.+ENTER`.
+
 ### Commands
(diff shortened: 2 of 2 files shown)


core/embed/projects/prodtest/cmd/prodtest_prodtest.c
```diff
@@ -34,6 +34,11 @@ static void prodtest_prodtest_intro(cli_t* cli) {
   cli_trace(cli, "");
 }
 
+static void prodtest_prodtest_outro(cli_t* cli) {
+  cli_trace(cli, "Quit interactive mode.");
```
cepetr, line 38: How about "Exiting interactive mode..."? I think it better matches the style of other traces.

cepetr (changes requested):

core/embed/rtl/cli.c
```diff
@@ -569,6 +569,17 @@ void cli_run_loop(cli_t* cli) {
     }
     cli->empty_lines = 0;
 
+    // Quit interactive mode on `.+ENTER`
+    if ((strcmp(cli->cmd_name, ".") == 0) && cli->interactive) {
+      cli->interactive = false;
```
cepetr, line 574: Maybe we don’t need an outro at all. The intro is useful since it provides information (that `cli.c` can't know), but the outro only states that interactive mode is exiting. A simple `cli_trace` might be enough here. Don’t you think?

cepetr (changes requested):

core/embed/rtl/cli.c
kopecdav: No problem with that, I just wanted to follow your code structure here. Only think I need is to make and cli_ok() so host script can easily recognize that the exit command propagates. If you agree, we can do that right away in `cli_run_loop()`

core/embed/projects/prodtest/cmd/prodtest_prodtest.c
kopecdav: Sure, no problem. I picked "Quit" because typically when you want to exit some console mode (like python for instance), you use quit() command. "Exiting" is fine.

cepetr (approved):

PR #4610, opened 2025-02-11
fix stwlc38 deinitialization

TychoVrahe:
This PR fixes deinitialzation of STWLC38 driver: 
- interrupt is disabled
- ~~wlc itself is disabled~~

### DIFF
--- a/core/embed/sys/powerctl/stwlc38/stwlc38.c
+++ b/core/embed/sys/powerctl/stwlc38/stwlc38.c
@@ -22,7 +22,6 @@
 
 #include <io/i2c_bus.h>
 #include <sys/irq.h>
-#include <sys/systick.h>
 #include <sys/systimer.h>
 
 #include "stwlc38.h"
@@ -75,8 +74,14 @@ static void stwlc38_fsm_continue(stwlc38_driver_t *drv);
 void stwlc38_deinit(void) {
   stwlc38_driver_t *drv = &g_stwlc38_driver;
 
-  i2c_bus_close(drv->i2c_bus);
+  NVIC_DisableIRQ(STWLC38_EXTI_INTERRUPT_NUM);
+  HAL_EXTI_ClearConfigLine(&drv->EXTI_Handle);
+
   systimer_delete(drv->timer);
+  i2c_bus_close(drv->i2c_bus);
+
+  HAL_GPIO_DeInit(STWLC38_INT_PORT, STWLC38_INT_PIN);
+  HAL_GPIO_DeInit(STWLC38_ENB_PORT, STWLC38_ENB_PIN);
   memset(drv, 0, sizeof(stwlc38_driver_t));
 }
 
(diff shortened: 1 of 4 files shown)


core/embed/sys/powerctl/stwlc38/stwlc38.c
```diff
@@ -75,6 +74,10 @@ static void stwlc38_fsm_continue(stwlc38_driver_t *drv);
 void stwlc38_deinit(void) {
   stwlc38_driver_t *drv = &g_stwlc38_driver;
 
+  NVIC_DisableIRQ(STWLC38_EXTI_INTERRUPT_NUM);
+
+  stwlc38_enable(false);
```
kopecdav, line 79: I think we should split this function into two separate enable/disable functions. This is bit confusing

core/embed/sys/powerctl/stwlc38/stwlc38.c
```diff
@@ -75,6 +74,10 @@ static void stwlc38_fsm_continue(stwlc38_driver_t *drv);
 void stwlc38_deinit(void) {
   stwlc38_driver_t *drv = &g_stwlc38_driver;
 
+  NVIC_DisableIRQ(STWLC38_EXTI_INTERRUPT_NUM);
```
kopecdav, line 77: Right after NVIC disable we may also clear the EXTI Line with `HAL_EXTI_ClearConfigLine()

core/embed/sys/powerctl/stwlc38/stwlc38.c
```diff
@@ -75,6 +74,10 @@ static void stwlc38_fsm_continue(stwlc38_driver_t *drv);
...
+
+  stwlc38_enable(false);
+
   i2c_bus_close(drv->i2c_bus);
   systimer_delete(drv->timer);
```
kopecdav, line 82: I think this is dangerous, drv->timer callback may submit to I2C bus, so we should delete the timer before the I2C bus is closed

core/embed/sys/powerctl/stwlc38/stwlc38.c
```diff
@@ -75,6 +74,10 @@ static void stwlc38_fsm_continue(stwlc38_driver_t *drv);
...
   memset(drv, 0, sizeof(stwlc38_driver_t));
```
kopecdav, line 83: Driver handler is fully cleared in the `stwlc38_init` function, so using just `drv->initialized = false` here may be more readable.

core/embed/sys/powerctl/stwlc38/stwlc38.c
kopecdav, line 82: I would `HAL_GPIO_DeInit()` all GPIOs here.

kopecdav (changes requested):

core/embed/sys/powerctl/stwlc38/stwlc38.c
cepetr: I think it’s important to clear it before enabling the interrupt. I wouldn’t clear it here, as it may give the impression that it’s necessary and could lead to others copying this pattern elsewhere.

core/embed/sys/powerctl/stwlc38/stwlc38.c
TychoVrahe: switched the order in https://github.com/trezor/trezor-firmware/commit/d763d02e8e7a616d949ee0fec15a0b5701d7ad3d

TychoVrahe: added https://github.com/trezor/trezor-firmware/commit/d763d02e8e7a616d949ee0fec15a0b5701d7ad3d

core/embed/sys/powerctl/stwlc38/stwlc38.c
TychoVrahe: i believe its better to clear it all, shouldn't matter much but lets keep cleaned what can be cleaned

core/embed/sys/powerctl/stwlc38/stwlc38.c
TychoVrahe: keeping as is

core/embed/sys/powerctl/stwlc38/stwlc38.c
TychoVrahe: you are probably right, i'd say its out of scope here and i don't want to do that right now :) 

it raises another question though, the comment for this function states that WLC is disabled by default, which makes no sense to me - so it should be like this https://github.com/trezor/trezor-firmware/commit/f7647da7c982cd8cf34b3ab99e38b7ae9839acc3 

cc @cepetr

TychoVrahe: also now i realize that disabling the WLC on deinit is not a good idea-  as we are deiniting the driver while jumping between bootloader and firmware for example - removed here https://github.com/trezor/trezor-firmware/commit/f4039c12e581b475036722c8b80cd20aba98ec05

core/embed/sys/powerctl/stwlc38/stwlc38.c
kopecdav: After another discussion with @cepetr we agreed to use `HAL_EXTI_ClearConfigLine()` here. It will properly clean all registers of the given Exti Line while the driver is deinitialized.

kopecdav (approved):

cepetr (approved):

PR #4612, opened 2025-02-12
refactor(core): separate `clear_unused_stack()` from `sys/linker` module

romanz:
Following https://github.com/trezor/trezor-firmware/pull/4595#pullrequestreview-2609833271.


cepetr (approved):
Looks good. Some tests are failing, but it doesn’t seem to be related to this.

PR #4613, opened 2025-02-12
Make Recovery Cancellable Again

matejcik:
TODO: it should be possible to write a click-test for this condition


obrusvit (approved):
MRCA

Will you add the click test in this PR?

matejcik:
had to touch a lot of things to get the tests to work, please re-review.

tests/click_tests/test_recovery.py
```diff
@@ -78,3 +78,44 @@ def test_recovery_bip39_previous_word(device_handler: "BackgroundDeviceHandler")
         bad_indexes = {1: seed_words[-1], 7: seed_words[0]}
         recovery.enter_seed_previous_correct(debug, seed_words, bad_indexes)
         recovery.finalize(debug)
+
+
+def test_recovery_cancel(device_handler: "BackgroundDeviceHandler"):
```
obrusvit, line 83: Can we make the test name more descriptive, e.g. `test_recovery_cancel_from_host`

The reason is, the test from #4640 is called `test_recovery_cancel_number_of_words` which sounds too similar to `test_recovery_cancel`. But they are quite different.

core/embed/rust/src/ui/util.rs
```diff
@@ -107,7 +107,7 @@ pub fn split_two_lines(text: &str, text_font: Font, available_width: i16) -> (&s
 
     #[cfg(feature = "ui_debug")]
     if text_font.text_width(second_line) > available_width {
-        fatal_error!(&uformat!("Text too long: {}...!", &text[..15]));
+        fatal_error!(&uformat!(len: 128, "Text too long: '{}'", text));
```
obrusvit, line 110: Nice catch.

Did you actually manage to trigger the silent panic?

matejcik: yes, it was the cause of the "mysterious" click test failures

tests/click_tests/test_recovery.py
matejcik: like this? https://github.com/trezor/trezor-firmware/pull/4613/commits/172bafe19120975982a6204f807c8df29090f9ba :)
a real "test_recovery_cancel" should be just going into the flow and exiting it. which we don't currently do at all, I think

PR #4614, opened 2025-02-12
fix(core): fix touch activity detection in bootloader

cepetr:
This PR fixes the issue introduced in #4582 that prevents touch activity detection in the bootloader on T2T1 and T3T1. 

Fortunately, it was found quickly and was not included in the bootloader release. No changelog entry is required.


TychoVrahe (approved):
utACK. I created an issue to split the two drivers, see https://github.com/trezor/trezor-firmware/issues/4624

PR #4615, opened 2025-02-12
fix(core): fix hw jpegdec issue with grayscale images

cepetr:
This PR fixes the issue with decoding grayscale JPEG images on the homescreen, which causes DMA2D to be non-functional in the core app.

No changelog entry needed.


TychoVrahe (approved):
utACK. I created an issue for adding grayscale support: https://github.com/trezor/trezor-firmware/issues/4625

PR #4626, opened 2025-02-13
Fix TX summary back button

ibz:
#4471 attempted to fix the back button on the TX summary on TS3, but I later learned that it should be "^" rather than "<". So here is the proper fix.


bosomt:
**QA OK**

i see `^` as back button during tx signing

 - Device: Trezor T3B1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)

PR #4627, opened 2025-02-13
Use middle button to confirm paginated content on TS3

### LINKED ISSUE #4622: Replace "hold to confirm" with "middle button"
We might eventually want to replace all instances of "hold to confirm" with the "middle button".

For now the most important place to do it is in `confirm_more`, which is used to confirm long messages.

It should look like this:

image

I am creating this issue so that I can work on #4601 without thinking about this change because it seems to affect many other places and it makes testing & reviewing that PR unnecessarily complicated, but eventually we can reuse this issue for other PRs dealing with TS3 "hold to confirm" throughout the code base.

ibz:
We eventually aim to replace all "hold to confirm" with "middle button" on TS3. Currently it is a mix of both and this PR, which is a follow-up of #4601, does the change for the confirmation of paginated content.


Hannsek:
Please rebase and adjust the tutorial flow so we can merge this once you have some time.

image

PR #4628, opened 2025-02-13
ci(core): increase pytest log verbosity

romanz:
It would also log packets being sent/received.


matejcik:
could we keep this in a branch? i don't think this will be relevant after the current problems are fixed

PR #4629, opened 2025-02-13
feat(core): introduce hibernate mode

cepetr:
This PR introduces the so-called hibernation mode for T3W1.

Hibernation mode utilizes PMIC ship mode, switching off CPU power while keeping only the VBAT domain powered.

- Added the `powerctl_hibernate()` function, including a syscall, to be available in coreapp.
- Added a new syscall for `powerctl_get_status()`. The function is not fully implemented yet.
- Added the `powerctl-hibernate` command to prodtest for testing purposes.
- The powerctl module is now initialized along with other drivers at prodtest startup.


core/embed/sys/powerctl/stm32u5/powerctl.c
```diff
@@ -77,16 +78,35 @@ void powerctl_deinit(void) {
...
+  }
+
+  if (!npm1300_enter_shipmode()) {
+    return false;
+  }
+
+  // Wait for the device to power off
+  systick_delay_ms(50);
+
+  return true;
```
TychoVrahe, line 109: if we return from this, shouldn't it be false, as it likely failed to hibernate?

core/embed/projects/prodtest/cmd/prodtest_powerctl.c
```diff
@@ -52,4 +69,12 @@ PRODTEST_CLI_CMD(
   .args = ""
 );
 
+PRODTEST_CLI_CMD(
+  .name = "powerctl-hibernate",
```
TychoVrahe, line 73: please add README doc

TychoVrahe (changes requested):

core/embed/sys/powerctl/stm32u5/powerctl.c
cepetr: Yes, I thought about this as well, but I don’t think it’s an error. If we return `false` here, we wouldn’t be able to distinguish between an error (due to uninitialized powerctl) and this… But I agree it’s quite strange

core/embed/projects/prodtest/cmd/prodtest_powerctl.c
cepetr: added, see 102c9108891443c77e3b6d6277582a68b11c0470

TychoVrahe (approved):

PR #4630, opened 2025-02-13
fix TS5 bootloader font

TychoVrahe:
This PR fixes recently introduced and unreleased bug that changed fonts on some places of bootloader.


obrusvit:
Sorry for that.

Unfortunately `theme::bootloader` is used in `theme/mod.rs` which creates the problem with build. Separation of those themes wasn't as straightforward when I attempted to do it.

How about you map the TT Hoves 17 to `FONT_SUB` for `feature="bootloader"` and use it in the button styles?

bosomt:
**QA OK**

```
bosomt@192 ~/D/fw> trezorctl get-features
Features (84 bytes) {
    bootloader_locked: True,
    bootloader_mode: True,
    firmware_present: True,
    fw_major: 2,
    fw_minor: 8,
    fw_patch: 9,
    fw_vendor: 'Trezor Bitcoin-only',
    internal_model: 'T3T1',
    language_version_matches: True,
    major_version: 2,
    minor_version: 1,
    model: 'Safe 5',
    patch_version: 10,
    unit_btconly: True,
    unit_color: 4,
    unit_packaging: 1,
    vendor: 'trezor.io',
```

PR #4631, opened 2025-02-14
bump T3W1 revB revision

TychoVrahe:
this PR increments T3W1 revB revision, making the builds incompatible with revA build. Prevents accidental uploading/running fw on boards with incompatible bootloader.


cepetr (approved):
utACK, but I've created #4637.

PR #4633, opened 2025-02-14
UI debug overlay

cepetr:
This PR introduces a new command-line option, `UI_DEBUG_OVERLAY`, that can be used together with `make build_firmware` / `build_bootloader` and `build_prodtest`. When enabled, it draws a small overlay with the rendering timing information in the top-right corner of the screen.

I’ve implemented this feature for the **Delizia** layout, but it can easily be added to the **Eckhart** layout as well.

The overlay displays two values:
- The first value represents the time (in milliseconds) spent in rendering routines for the current frame.
- The second value includes the rendering time (in milliseconds) as well as time spent on event processing and other tasks in Rust and MicroPython. It reflects the time between two consecutive frames being issued to the display driver.

The second value better represents the overall frame duration, while the first value is useful for optimizing rendering routines.

I hope you find this feature useful to achieve fast and smooth animations.


core/embed/rust/src/ui/shape/display/fb_rgb565.rs
```diff
@@ -53,6 +60,13 @@ where
 
         let mut target = ScopedRenderer::new(DirectRenderer::new(&mut canvas, bg_color, &cache));
 
-        func(&mut target);
+        let render_time = time::measure_us(|| func(&mut target));
```
obrusvit, line 63: Wouldn't this calculate the times even when we don't have the new features enabled? Is it negligible or wasteful?

cepetr: Yes, we could - although the measurement takes a few microseconds. I was considering it as well. If we did that, the code would be full of `#[cfg(feature = ..)` stuff. 

Actually, I think we could enable/disable this overlay from trezorctl (having a special command for it), so even the test guys can check the frame rate…

cepetr: After our offline discussion, I’ve renamed the feature to `ui_debug_overlay` and updated the related symbols for consistency. Additionally, I’ve decided to make the feature entirely conditional, meaning it is completely excluded from the release build. See 24c599e439489ca3cff3c586e7aef10a42096c8a

TychoVrahe (commented):
This looks fine to me, but expanding on this idea..

We could move the debug firmware indicator (red square in corner) rendering to rust too, and use this debug overlay to implement it. Thus, debug overlay would be used always when PYOPT=0, and 'time measurements' would be separate build flag + feature name. 

We could probably also make the red square exclusive with the time measurements, as for now the it will overlap with the time, making it harder to read. 

Obviously out of scope here, but to better prepare for this, we could name the makefile flag differently. The feature can probably stay, but it will change meaning if we implement this.

PR #4640, opened 2025-02-17
Add ability to cancel recovery on word count selector

ibz:
This PR adds a "cancel" button to the word count selection screen of the recovery process. Models: TS3, TS5, T.

This affects both the "normal" recovery and the "unlock repeated backup" process.

## Screenshots

Screenshot from 2025-02-18 17-23-08
Screenshot from 2025-02-18 17-23-28
Screenshot from 2025-02-18 17-26-30
Screenshot from 2025-02-18 17-26-49
Screenshot from 2025-02-18 17-27-33
image

## Testing

* `trezorctl device recover` on an uninitialized device
* `trezorctl device recover -b` on an already initialized device


core/src/trezor/ui/layouts/delizia/recovery.py
```diff
@@ -16,12 +17,16 @@
...
+    from apps.management.recovery_device.recover import RecoveryAborted
+    try:
+        count = await interact(
+            trezorui_api.select_word_count(recovery_type=recovery_type),
+            "recovery_word_count",
+            ButtonRequestType.MnemonicWordCount,
+        )
+        return int(count)
+    except ActionCancelled:
+        raise RecoveryAborted
```
obrusvit, line 29: Is this the correct `RecoveryAborted`? :hand_over_mouth:

core/embed/rust/src/ui/layout_delizia/component/keyboard/word_count.rs
```diff
@@ -116,16 +126,20 @@ impl Component for ValueKeypad {
     }
 
     fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
-        for (i, btn) in self.button.iter_mut().enumerate() {
+        for (i, btn) in self.buttons.iter_mut().enumerate() {
             if let Some(ButtonMsg::Clicked) = btn.event(ctx, event) {
-                return Some(SelectWordCountMsg::Selected(Self::NUMBERS[i]));
+                if Self::NUMBERS[i].is_some() {
```
obrusvit, line 131: You can do something like this which is arguably more idiomatic
```rust
            return Some(match Self::NUMBERS[i] {
                Some(num) => SelectWordCountMsg::Selected(num),
                None => SelectWordCountMsg::Cancelled,
            });
```

Also we should use `unwrap!(..)` instead of `.unwrap()` in all cases.

ibz: Sorry for force push, but this is still WIP. I have pretty much redone the Delizia word count keyboard. But now there is another issue: it is 90% the same as the Bolt word count keyboard. Perhaps I should extract it as a common component? Do you have any opinion on this @obrusvit ?

core/embed/rust/src/ui/layout_caesar/component/input_methods/choice.rs
```diff
@@ -117,6 +127,7 @@ where
             page_counter: 0,
             items_distance: DEFAULT_ITEMS_DISTANCE,
             is_carousel: false,
+            is_cancelable: false,
```
matejcik, line 130: wondering: what would happen if we used `!is_carousel` instead of `is_cancelable` ?

i'm thinking _most_ places should be cancellable, and the only place comes to mind where that is undesirable is `select_word` ... where we can cheat and make it a carousel :)

core/embed/rust/src/ui/layout_caesar/component/input_methods/passphrase.rs
```diff
@@ -398,7 +398,9 @@ impl Component for PassphraseEntry {
 
         if let Some((action, long_press)) = self.choice_page.event(ctx, event) {
             match action {
-                PassphraseAction::CancelOrDelete => {
+                CancelableChoiceAction::<PassphraseAction>::Choice(
```
matejcik, line 401:
```suggestion
                CancelableChoiceAction::Choice(
```
seems to work fine for me without it

core/embed/rust/src/ui/layout_caesar/component/input_methods/simple_choice.rs
```diff
@@ -1,3 +1,5 @@
+use crate::error::Error;
```
matejcik, line 1: unused

core/embed/rust/src/ui/layout_caesar/component/input_methods/wordlist.rs
```diff
@@ -282,12 +285,12 @@ impl Component for WordlistEntry {
     fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
         if let Some((action, long_press)) = self.choice_page.event(ctx, event) {
             match action {
-                WordlistAction::Previous => {
+                CancelableChoiceAction::<WordlistAction>::Choice(WordlistAction::Previous) => {
```
matejcik, line 288:
```suggestion
                CancelableChoiceAction::Choice(WordlistAction::Previous) => {
```

core/embed/rust/src/ui/layout_caesar/ui_firmware.rs
```diff
@@ -893,18 +894,20 @@ impl FirmwareUI for UICaesar {
...
+            WORD_COUNT_LABELS_MULTISHARE
+                .iter()
+                .map(|&label| TString::from_str(label))
+                .collect()
+        } else {
+            WORD_COUNT_LABELS_ALL
+                .iter()
+                .map(|&label| TString::from_str(label))
+                .collect()
         };
```
matejcik, line 907: istm the previous code would work, only replacing the literals with constants?

core/src/trezor/ui/layouts/delizia/recovery.py
obrusvit: So why is it still there? Please change the import from local to global for all layouts.

I tried on both the emu and the device and the bug we discussed is there.

core/embed/rust/src/ui/layout_bolt/component/keyboard/word_count.rs
```diff
@@ -62,21 +47,50 @@ impl Component for SelectWordCount {
     }
 }
 
-type ValueKeyPacked = (Button, u32, (usize, usize)); // (Button, number, cell)
+type Cell = (usize, usize);
 
-pub struct ValueKeypad {
-    buttons: Vec<ValueKeyPacked, 5>,
+struct ValueKeypad {
+    buttons: Vec<(Button, Option<u32>, Cell), 6>,
```
matejcik, line 53: you don't need to store `Cell` if you do the zip in `place()` :)

obrusvit (changes requested):
Partial review.

core/embed/rust/src/ui/layout_delizia/component/keyboard/word_count.rs
```diff
@@ -4,91 +4,110 @@ use crate::ui::{
...
+     * -----
+     * x | 4
+     */
+    const SIX_CELLS: [Cell; 6] = [
+        (0, 0, 1, 1),
+        (0, 2, 1, 1),
+        (2, 0, 1, 1),
+        (2, 2, 1, 1),
+        (4, 0, 1, 1),
+        (4, 2, 1, 1),
```
matejcik, line 70: _here_, you can leave an empty space in `SIX_CELLS` for the cell in bottom left, and make a special field for `cancel_cell` that you fill out -- either with the big one, or with the bottom left one. this again allows you to zip the buttons with `CELLS` in `place()`, except now `cancel` will be its own button, not part of the vec.

that also means that `buttons` don't need an `Option` for the value

matejcik (changes requested):

core/src/trezor/ui/layouts/delizia/recovery.py
matejcik: probably because the fix is not merged yet (waiting for your re-review @obrusvit :) )
also this has no business being in the layout code at all so :woman_shrugging:

tests/click_tests/recovery.py
```diff
@@ -63,6 +63,47 @@ def confirm_recovery(debug: "DebugLink", title: str = "recovery__title") -> None
             debug.press_right()
 
 
+def cancel_select_number_of_words(
+    debug: "DebugLink",
+    unlock_repeated_backup=False,
+) -> None:
+    def select_bolt() -> "LayoutContent":
```
obrusvit, line 70: Nit, but in this particular case, I'd probably get rid of the the individual functions (`select_bolt/caesar/..`) and just put the stuff into the `if/elif` chain.

core/embed/rust/src/ui/layout_bolt/component/keyboard/word_count.rs
```diff
@@ -14,34 +14,19 @@ use heapless::Vec;
...
 
-    pub fn new_all() -> Self {
-        Self {
-            keypad: ValueKeypad::new(&Self::NUMBERS_ALL, &Self::LABELS_ALL, &Self::CELLS_ALL),
-        }
-    }
-
-    pub fn new_multishare() -> Self {
+impl SelectWordCount {
+    pub fn new(choices: Vec<u32, 5>, labels: Vec<Label, 5>) -> Self {
```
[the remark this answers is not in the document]
ibz: > would mean that we can't share choices between Caesar and Bolt/Delizia

Yeah, that's something I tried to fix, but if you think passing a Vec is worse than duplicating the arrays, I think there's no way around it ... Unless I somehow pass a slice?

core/embed/rust/src/ui/layout_caesar/component/input_methods/choice.rs
ibz: I could try that, but then I would drop the `is_carousel` flag rather than the `is_cancelable` flag because ... it's strange to change the cancelability based on `is_carousel`, but rendering it as non-carousel whenever it is cancelable seems less strange given there is no way to show the cancel button on carousel mode...

core/src/trezor/ui/layouts/delizia/recovery.py
obrusvit: @matejcik I meant that your (unmerged) fix removed the local import form `core/src/apps/management/recovery_device/layout.py`

This code is `layouts` specific. I tested with your fix in this branch and the bug does not occur if you cancel after selecting the word count but it sure does when you cancel in the (new in this PR) cancel button in the word count selection screen.

Anyway, I read your comment here https://github.com/trezor/trezor-firmware/pull/4640#discussion_r1961788487 so it's all good 👍

core/embed/rust/src/ui/layout_delizia/component/keyboard/word_count.rs
```diff
@@ -4,91 +4,110 @@ use crate::ui::{
...
+type Label = &'static str;
 
-    pub fn new_multishare() -> Self {
-        Self::Multishare(BinarySelection::new(
-            ButtonContent::Text("20".into()),
-            ButtonContent::Text("33".into()),
-            theme::button_keyboard(),
-            theme::button_keyboard(),
-        ))
+impl SelectWordCount {
```
obrusvit, line 25: `impl SelectWordCount` was previously introduced so that we could reuse the `BinarySelection` component. This is now not an option so, IMO, `SelectWordCount` makes no sense anymore and you can simply use `impl ValueKeypad` and `impl Component for ValueKeypad`. 
It might make sense to keep the name `SelectWourdCount` though..?

obrusvit (commented):
A few comments.

core/embed/rust/src/ui/layout_bolt/component/keyboard/word_count.rs
```diff
@@ -62,21 +47,50 @@ impl Component for SelectWordCount {
...
+     * x | 3 | 4
+     */
+    const SIX_CELLS: [Cell; 6] = [(0, 0), (0, 2), (0, 4), (1, 0), (1, 2), (1, 4)];
+    const SIX_CELLS_CANCEL_POSITION: usize = 3;
+
+    /*
+     * x | 0 | 1
+     */
+    const THREE_CELLS: [Cell; 3] = [(0, 0), (0, 2), (0, 4)];
+    const THREE_CELLS_CANCEL_POSITION: usize = 0;
```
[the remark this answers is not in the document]
ibz: I don't think this would work for Delizia. What I was trying to do here is keep the same algorithm for both layouts. Is there any benefit given the extra complexity of having two ways to do the same thing?

core/embed/rust/src/ui/layout_caesar/component/input_methods/choice.rs
matejcik: how about:
```rust
enum ChoiceControls {
    Carousel,
    Cancellable,
}
```

core/.changelog.d/3503.added
```diff
@@ -0,0 +1 @@
+- Added ability to cancel recovery on word count selection screen.
```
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4640/commits/2da51e81394e6d83342d66f781c82ad3fa7dc226#diff-8b55c73d229584772ddf20531607502adcb64dc671a1f28787b73de48694d2f9R1

core/embed/rust/src/ui/layout_bolt/component/keyboard/word_count.rs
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4640/commits/2da51e81394e6d83342d66f781c82ad3fa7dc226#diff-dc9a2542693a4b3725dcb8986bff554165af00818ed72d7797e3f738af6e1f5eR62

core/src/trezor/ui/layouts/bolt/recovery.py
```diff
@@ -15,12 +16,17 @@
...
+
+    try:
+        count = await interact(
+            trezorui_api.select_word_count(recovery_type=recovery_type),
+            "recovery_word_count",
+            ButtonRequestType.MnemonicWordCount,
+        )
+        return int(count)
+    except ActionCancelled:
+        raise RecoveryAborted
```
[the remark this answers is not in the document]
ibz: Done.

core/embed/rust/src/ui/layout_caesar/component/input_methods/choice.rs
```diff
@@ -11,6 +13,11 @@ use super::super::{
 
 const DEFAULT_ITEMS_DISTANCE: i16 = 10;
 
+pub enum CancelableChoiceAction<T> {
+    Choice(T),
+    Cancel,
+}
```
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4640/commits/2da51e81394e6d83342d66f781c82ad3fa7dc226#diff-4227b0d31f81b6133eee1f517a0d9f07322ad3f4d2d6f649382c3eb31922e5e9R14

core/embed/rust/src/ui/layout_caesar/component/input_methods/choice.rs
```diff
@@ -486,7 +503,7 @@ impl<F, A> Component for ChoicePage<F, A>
 where
     F: ChoiceFactory<Action = A>,
 {
-    type Msg = (A, bool);
+    type Msg = (CancelableChoiceAction<A>, bool);
```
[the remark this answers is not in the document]
ibz: https://github.com/trezor/trezor-firmware/pull/4640/commits/2da51e81394e6d83342d66f781c82ad3fa7dc226#diff-4227b0d31f81b6133eee1f517a0d9f07322ad3f4d2d6f649382c3eb31922e5e9R504

core/embed/rust/src/ui/layout_caesar/component/input_methods/simple_choice.rs
ibz: Done

core/embed/rust/src/ui/layout_caesar/component/input_methods/wordlist.rs
ibz: https://github.com/trezor/trezor-firmware/pull/4640/commits/2da51e81394e6d83342d66f781c82ad3fa7dc226#diff-1d05f8aa2e5df83ed3fe23b11538ab0ddb4faae0b3b08af50d4559d46df3fa30R287

core/embed/rust/src/ui/layout_delizia/component/keyboard/word_count.rs
ibz: Yes, I removed that extra level of indirection...

core/embed/rust/src/ui/layout_caesar/component/input_methods/choice.rs
ibz: Still an experiment, but it might work: 812cb8d77f656e5b7b9a37c4bf9f423f6543b63d

core/embed/rust/src/ui/layout_caesar/component_msg_obj.rs
```diff
@@ -117,17 +117,25 @@ impl ComponentMsgObj for CoinJoinProgress {
 
 impl ComponentMsgObj for NumberInput {
     fn msg_try_into_obj(&self, msg: Self::Msg) -> Result<Obj, Error> {
-        (CONFIRMED.as_obj(), msg.try_into()?).try_into()
+        match msg {
+            Self::Msg::Cancel => (CANCELLED.as_obj(), 0.try_into()?).try_into(),
+            Self::Msg::Choice { item, .. } => (CONFIRMED.as_obj(), item.try_into()?).try_into(),
```
matejcik, line 122: while we're here: i forget, why do we return a tuple instead of `int / Cancelled` ?

obrusvit (approved):
My previous points were fixed. LGTM.

core/.changelog.d/3503.added
```diff
@@ -1 +1,2 @@
 Added ability to cancel recovery on word count selection screen.
+
```
matejcik, line 2: this empty line should not be here.

ibz: @obrusvit I'll remove it from the other PR also then? :)

matejcik (approved):

bosomt:
**QA OK**

tested on multiple models including dry run
firmware-T3T1-2.8.9-7614a29.bin 7614a29

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.3.0 (2a761d42998cb22ef837ae358cb9aaa39ea2123f)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.3.0 Chrome/132.0.6834.194 Electron/34.1.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T2B1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)

PR #4641, opened 2025-02-17
refactor(core): separate base and priority Solana fees

romanz:
Move fee calculation into `Transaction` class.

Also, replace floating-point division by integer division in fee calculation.

It will be used in #4639.


romanz:
I have used https://solana.com/docs/core/fees for base & priority fee definitions.

romanz:
Will be done in #4639.

PR #4644, opened 2025-02-18
feat(core): allow more paragraphs in `ShowInfoParams`

romanz:
Otherwise, it fails when showing 4 items.

Note that each item uses in 2 paragraphs. Also, items are separated by an empty paragraph.

So 3 items use 8 (2\*3+2) paragraphs and 4 items use 11 (2\*4+3) paragraphs.


romanz:
Currently, `ShowParamsInfo` supports upto 4 items:
https://github.com/trezor/trezor-firmware/blob/13c078f8af75135ac757423da3fc0c013186c32c/core/embed/rust/src/ui/layout_delizia/flow/util.rs#L305-L316

obrusvit (approved):
utACK

PR #4652, opened 2025-02-20
Add display-text command to prodtest

kopecdav:
Add simple `display-text` command to prodtest, which prints the text given in argument.

This is quite useful when the external controller needs to logs some data on tested trezor.  
We use this during battery test to log battery voltage, and actual current load.

prodtest excepts arguments splitted with space character and display-text command excepts only single argument, so to print several words use different spacing character such as "_".

### DIFF
--- a/core/embed/projects/prodtest/README.md
+++ b/core/embed/projects/prodtest/README.md
@@ -158,6 +158,15 @@ display-border
 OK
 ```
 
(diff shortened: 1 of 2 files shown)


TychoVrahe (approved):
looks good, but please squash the commits into one

PR #4655, opened 2025-02-21
Debuglink timing out fix

matejcik:

### DIFF
--- a/python/src/trezorlib/debuglink.py
+++ b/python/src/trezorlib/debuglink.py
@@ -493,8 +493,8 @@ class DebugLink:
         )
         self.transport.write(msg_type, msg_bytes)
 
-    def _read(self) -> protobuf.MessageType:
-        ret_type, ret_bytes = self.transport.read()
+    def _read(self, timeout: float | None = None) -> protobuf.MessageType:
+        ret_type, ret_bytes = self.transport.read(timeout=timeout)
         LOG.log(
             DUMP_BYTES,
             f"received type {ret_type} ({len(ret_bytes)} bytes): {ret_bytes.hex()}",
@@ -513,9 +513,9 @@ class DebugLink:
         )
         return msg
 
-    def _call(self, msg: protobuf.MessageType) -> Any:
+    def _call(self, msg: protobuf.MessageType, timeout: float | None = None) -> Any:
         self._write(msg)
-        return self._read()
+        return self._read(timeout=timeout)
 
     def state(self, wait_type: DebugWaitType | None = None) -> messages.DebugLinkState:
         if wait_type is None:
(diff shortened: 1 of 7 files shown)


python/src/trezorlib/transport/udp.py
```diff
@@ -142,14 +142,17 @@ def write_chunk(self, chunk: bytes) -> None:
...
+        start = time.time()
         while True:
             try:
                 chunk = self.socket.recv(64)
                 break
             except socket.timeout:
-                continue
+                if timeout is not None and time.time() - start > timeout:
+                    raise TransportException("Timeout reading UDP packet")
+                time.sleep(0.001)
```
romanz, line 155: Not sure why we need `time.sleep()` here.

python/src/trezorlib/debuglink.py
```diff
@@ -640,7 +640,10 @@ def _decision(self, decision: messages.DebugLinkDecision) -> None:
         if self.model is models.T1B1:
             return
         # When the call below returns, we know that `decision` has been processed in Core.
-        self._call(messages.DebugLinkGetState(return_empty_state=True))
+        # XXX Due to a bug, the reply may get lost at the end of a workflow.
+        # We assume that no single input event takes more than 5 seconds to process,
+        # and give up waiting after that.
+        self._call(messages.DebugLinkGetState(return_empty_state=True), timeout=5)
```
romanz, line 646: So just to be sure - we raise an exception here, right?
Asking, since IIRC there was a suggestion to ignore the timeout (i.e. to assume the request has been processed by the device), but then I would suggest also [logging that there was an ignored timeout](https://peps.python.org/pep-0020/#:~:text=Explicit%20is%20better%20than%20implicit.).

romanz: Maybe we can use an environment variable to set the timeout?
IIUC, it may take the device more than 5 seconds to respond, e.g. [when running `WipeDevice`](https://satoshilabs.slack.com/archives/C054XGU77QV/p1741010163181289?thread_ts=1741007974.203949&cid=C054XGU77QV).

python/src/trezorlib/transport/webusb.py
```diff
@@ -96,9 +100,10 @@ def write_chunk(self, chunk: bytes) -> None:
...
             return
 
-    def read_chunk(self) -> bytes:
+    def read_chunk(self, timeout: float | None = None) -> bytes:
         assert self.handle is not None
         endpoint = 0x80 | self.endpoint
+        start = time.time()
         while True:
             try:
                 chunk = self.handle.interruptRead(
```
romanz, line 109: It seems that `interruptRead` doesn't return `None` on timeout - added a small fix here: https://github.com/trezor/trezor-firmware/pull/4748/commits/a866ffba8fae233f4c12b9e30e28fd6e86291407

python/src/trezorlib/transport/udp.py
romanz: Resolved in https://github.com/trezor/trezor-firmware/pull/4748/commits/f20249192bb8e4b789c484ba12fda7d382e740b3.

python/src/trezorlib/debuglink.py
romanz: -> https://github.com/trezor/trezor-firmware/pull/4748/commits/27c93702e0da0eff314605204ebf756c7b0f9ecf

PR #4662, opened 2025-02-24
add touch-draw command to prodtest

TychoVrahe:
Implements simple drawing canvas where user can draw with finger or pen.

### DIFF
--- a/core/embed/projects/prodtest/cmd/prodtest_touch.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_touch.c
@@ -336,6 +336,40 @@ static void prodtest_touch_test_sensitivity(cli_t* cli) {
   screen_prodtest_welcome();
 }
 
+static void prodtest_touch_draw(cli_t* cli) {
+#define MAX_EVENTS 256
+
+  if (cli_arg_count(cli) > 0) {
+    cli_error_arg_count(cli);
+    return;
+  }
+
+  cli_trace(cli, "Starting drawing canvas...");
+  cli_trace(cli, "Press CTRL+C for exit.");
+
+  uint32_t events[MAX_EVENTS] = {0};
+  uint32_t idx = 0;
+
+  screen_prodtest_draw(events, 0);
+  display_set_backlight(150);
+
+  for (;;) {
+    uint32_t evt = touch_get_event();
+
+    if (evt != 0) {
+      events[idx] = evt;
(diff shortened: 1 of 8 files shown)


core/embed/projects/prodtest/cmd/prodtest_touch.c
```diff
@@ -336,6 +336,33 @@ static void prodtest_touch_test_sensitivity(cli_t* cli) {
...
 }
 
+static void prodtest_touch_draw(cli_t* cli) {
+  cli_trace(cli, "Starting drawing canvas...");
+  cli_trace(cli, "Press CTRL+C for exit.");
+
+  uint32_t events[256] = {0};
+  uint32_t idx = 0;
+
+  screen_prodtest_draw(events, 0);
```
cepetr, line 346: We should always verify that the correct number of arguments is provided. In this case, there should be:

```c
  if (cli_arg_count(cli) > 0) {
    cli_error_arg_count(cli);
    return;
  }
```

core/embed/rust/src/ui/layout/simplified.rs
```diff
@@ -51,6 +51,18 @@ fn button_eval() -> Option<ButtonEvent> {
...
+    }
+    let event_type = event >> 24;
+    let ex = ((event >> 12) & 0xFFF) as i16;
+    let ey = (event & 0xFFF) as i16;
+
+    TouchEvent::new(event_type, ex as _, ey as _).ok()
+}
+
 #[cfg(feature = "touch")]
 fn touch_eval() -> Option<TouchEvent> {
```
cepetr, line 67: `touch_eval()` can use the new function `touch_unpack()` to avoid repeating the same code.

cepetr (changes requested):

core/embed/projects/prodtest/cmd/prodtest_touch.c
TychoVrahe: added https://github.com/trezor/trezor-firmware/commit/df7405cf225185c8df84791b4e4df7500f75c8de

core/embed/rust/src/ui/layout/simplified.rs
TychoVrahe: removed the function entirely as there is not much left https://github.com/trezor/trezor-firmware/commit/df7405cf225185c8df84791b4e4df7500f75c8de

cepetr (approved):

PR #4663, opened 2025-02-24
Eckhart UI - Hold to Confirm animation improvement

obrusvit:
Hold to confirm animation with easing and rollback.


obrusvit:
force-pushed after some AFK discussions, PR is ready.

core/embed/rust/src/ui/layout_eckhart/component/hold_to_confirm.rs
```diff
@@ -27,17 +28,50 @@ pub struct HoldToConfirmAnim {
...
+    /// Duration after which the header overlay is shown after `start` is called
+    const HEADER_OVERLAY_DELAY: Duration = Duration::from_millis(300);
+
     pub fn new() -> Self {
+        // Assert that our ratios sum to 1.0
+        debug_assert!(
+            (Self::BOTTOM_DURATION_RATIO + Self::SIDES_DURATION_RATIO + Self::TOP_DURATION_RATIO
+                - 1.0)
+                < 0.001
+        );
```
bieleluk, line 63: I have two points here:
- assert catches only cases where the sum is greater than one
- it would be better to have the compile-time error as the sum can be computed in the compile time
My suggestion combines both these points
```suggestion
        const _: () = {
            const SUM: f32 = HoldToConfirmAnim::BOTTOM_DURATION_RATIO
                + HoldToConfirmAnim::SIDES_DURATION_RATIO
                + HoldToConfirmAnim::TOP_DURATION_RATIO;

            assert!(
                SUM >= 0.999 && SUM <= 1.001,
                "Duration ratios do not sum to 1.0"
            );
        };
```

core/embed/rust/src/ui/layout_eckhart/component/hold_to_confirm.rs
```diff
@@ -141,37 +134,158 @@ impl Component for HoldToConfirmAnim {
...
+            let rollback_elapsed = self.rollback.timer.elapsed();
+            let alpha = self.get_rollback_alpha(rollback_elapsed);
+            let rollback_duration_progressed = self
+                .rollback
+                .duration
+                .checked_add(rollback_elapsed)
+                .unwrap_or(Duration::default());
+            let (clip, top_gap) = self.get_clips(rollback_duration_progressed);
+            let top_back_rollback = self.get_top_gap_rollback(rollback_elapsed);
+            let top_gap = top_gap.union(top_back_rollback);
```
bieleluk, line 148: Simple and clever :+1:

bieleluk (changes requested):

core/embed/rust/src/ui/layout_eckhart/component/hold_to_confirm.rs
obrusvit: I wanted to use `(summation - 1.0).abs() < 0.001` but I found out that `f32::abs` is in `std` and not `core` :( 

I'm considering putting this assert into a unit instead so that we don't over-complicate that :thinking:

bieleluk: That is a good idea, let's do it that way

obrusvit: b88aec01645f9621f112e4a4205c4a28b1993fa9

bieleluk (approved):

Issue #3854, opened 2024-05-23
rounded rect rendering bug


TychoVrahe:
When the rect height is smaller than 2*radius+2, the middle portion of the rect is blended three times, resulting in very different color.

possibly, width smaller than 2*radius+2 could break this too.

cepetr:
We have few options how to solve it:
1. Clamping: If the specified corner radius is too large, we will automatically reduced it so that it does not exceed the smaller dimension of the rectangle. 
2. Fallback to rectangle: If the radius is too large, we will draw a simple rectangle with no rounded corners.
3. Raise Error: This can help developers to catch this issue and fix it.

TychoVrahe:
I was thinking about 4) keeping the radius and rendering a render a sharp corner in the middle, but 1) seems better. 

We could also enable 3) just in debug, otherwise 1)

PR #4666, opened 2025-02-24
 fix rendering of outlined Bars

TychoVrahe:
This PR fixes a problem with rendering outlines of Bar. Previous implementation drawn sides of the rectangle one over another, which, in combination with `with_alpha` usage, cause different color then the rest of the rect.


TychoVrahe:
i also added some related info to this issue: https://github.com/trezor/trezor-firmware/issues/3854

PR #4671, opened 2025-02-25
feat(core): use `Account info` instead of `Send from` in Delizia staking UI

romanz:
Following #4672.

image


bieleluk (approved):
No issues here.
The PR solves the same issues also for ETH staking.

bosomt:
**QA OK**

`Account info`  is present in info during staking dialogue

 - Device: Trezor T3T1 2.8.10 regular (revision 283561c073e750f00abd8632b6ae25df6606f834)

Issue #4672, opened 2025-02-25
Solana staking confirmation followups


romanz:
> I observed some some deviations from the Figma deign.
> 
> * Delizia:
>   
>   * [x] In the staking flow, the `Provider` has a different font.
>   * [x] In all flows, the account info menu item has title  `Send from` instead of  `Account  info`
>   * [x] In all flows, the confirm amount menu uses extra tap to cancel screen
> * Caesar:
>   
>   * [x] In the claim flow, there should be a  title `Claim` and a question below.  The question is directly in the title.
> * Bolt:
>   
>   * [ ] In the first screen of the stake flow, the fonts `Stake SOL` and `Provider` are swapped.
>   * [x] In all flows in Blockhash and Stake account menu items, there should be a bigger padding between the description and the value
> 
> They do not break the functionality so it is not critical to fix them all. If some of them are easy to fix, then let's do it.

https://github.com/trezor/trezor-firmware/pull/4639#issuecomment-2680968929

Hannsek:
@romanz Are you planning to work on this anytime soon?

romanz:
`In all flows, the confirm amount menu uses extra tap to cancel screen` marked as done - since it behaves similarly on BTC/ETH signing.

![Image](https://github.com/user-attachments/assets/d4462058-ba1c-4c20-9d6e-08aa225a33d4)
![Image](https://github.com/user-attachments/assets/c1e370a5-ca0d-4f88-ba7b-a59f6a37abd9)

romanz:
`In all flows in Blockhash and Stake account menu items, there should be a bigger padding between the description and the value` marked as done - there is an inconsistency between `stake` and `unstake`/`claim` screens in Figma.

PR #4677, opened 2025-02-26
fix(core): use correct fonts for staking confirmation on Delizia 

romanz:
Before:
 

After:

Following #4672.


romanz:
Rebasing to update UI fixtures:
```
1:  d6ca58125 = 1:  13c25ecd1 fix(core): use correct fonts for staking confirmation on Delizia
-:  --------- > 2:  644f47407 fixup! fix(core): use correct fonts for staking confirmation on Delizia
```

core/embed/rust/src/ui/layout_delizia/ui_firmware.rs
```diff
@@ -527,7 +528,10 @@ impl FirmwareUI for UIDelizia {
     ) -> Result<impl LayoutMaybeTrace, Error> {
         let confirm_main =
             ConfirmValue::new(title.unwrap_or(TString::empty()), message, description)
+                .with_description_font(&theme::TEXT_MAIN_GREY_LIGHT)
```
bieleluk, line 531: I am a bit hesitant about this change.
It is correct according to Figma, but it should have the same color as the address which also has wrong color (`GREY_EXTRA_LIGHT` instead of `GREY_LIGHT`).
In the test diff screen, there are now 3 different font colors, which does not look good. 

I would either change color of both description and address to `GREY_LIGHT` or none of them.
image

bieleluk (changes requested):
The main task of fonts differentiation looks good. I have one minor objection regarding the font color.

core/embed/rust/src/ui/layout_delizia/ui_firmware.rs
romanz: Fixed in #4733, rebasing to resolve UI fixtures' conflicts.

bieleluk (approved):

PR #4684, opened 2025-02-26
fix(core): confirm only visible Solana staking instructions

romanz:
Some existing fixtures needed to be updated since their instructions match the new confirmation logic.


matejcik (approved):
utACK.
should we merge this now @Hannsek or wait for resolution of the "partial unstake" scenarios?

bosomt:
**QA OK**

tested on 2.8.9 fad9682 firmwares

PR #4685, opened 2025-02-26
fix(core): confirm only visible Solana staking instructions

romanz:
(cherry picked from commit ee1cc4ae81eb8c89c3a5ea3d05ada9a9b3ff1dc8)


bosomt:
**QA OK**

tested on 2.8.9 fad9682 firmwares

PR #4687, opened 2025-02-26
fix(tests): don't leave emulator process running

mmilata:
When an exception is raised in the wrong place, pytest exits but leaves emulator process running in the background. This can result in confusing behavior of further pytest invocations which will connect to this old emulator. The fix attempts to keep track of running emulators in a global variable and then kill them in `atexit()` hook.

Included are two more quality of life improvements.

### DIFF
--- a/python/src/trezorlib/_internal/emulator.py
+++ b/python/src/trezorlib/_internal/emulator.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
+import atexit
 import logging
 import os
 import subprocess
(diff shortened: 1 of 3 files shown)


romanz (approved):
LGTM, thanks!

python/src/trezorlib/_internal/emulator.py
```diff
@@ -154,6 +167,8 @@ def launch_process(self) -> subprocess.Popen:
         )
 
     def start(self) -> None:
+        global _RUNNING_PIDS
```
matejcik, line 170: in none of these places `_RUNNING_PIDS` needs to be global

mmilata: 57ae7199f1

matejcik (approved):

PR #4700, opened 2025-02-28
fix jump to old bootloader

TychoVrahe:
Initialize systick before jump, as the old bootloaders may not initialize systick on start and rely on systick running from boardloader or firmware. Only relevant for F4, otherwise we are jumping through reset.

### DIFF
--- a/core/embed/sys/startup/stm32/bootutils.c
+++ b/core/embed/sys/startup/stm32/bootutils.c
@@ -168,6 +168,10 @@ static void reboot_with_args_phase_2(uint32_t arg1, uint32_t arg2) {
   if (command == BOOT_COMMAND_NONE) {
     NVIC_SystemReset();
   } else {
+#ifndef FIXED_HW_DEINIT
+    SysTick_Config(HAL_RCC_GetSysClockFreq() / 1000U);
+    NVIC_SetPriority(SysTick_IRQn, 0);
+#endif
     jump_to_vectbl(BOOTLOADER_START + IMAGE_HEADER_SIZE, command);
   }
 #else


core/embed/sys/startup/stm32/bootutils.c
```diff
@@ -168,6 +168,9 @@ static void reboot_with_args_phase_2(uint32_t arg1, uint32_t arg2) {
   if (command == BOOT_COMMAND_NONE) {
     NVIC_SystemReset();
   } else {
+#ifndef FIXED_HW_DEINIT
+    systick_init();
+#endif
```
cepetr, line 173: Not completely sure about this, but it looks like the old version used this sequence to initialize SysTick:

`main() -> periph_init() -> HAL_Init() -> HAL_InitTick(**TICK_INT_PRIORITY**)` and `TICK_INT_PRIORITY` used to be set to 0.

However, `systick_init()` sets the priority to 8 (IRQ_PRI_NORMAL).  

I don’t know if this difference could affect the old bootloader’s behavior, but I would rather fix it. Maybe instead of calling  Maybe instead of calling `systick_init()`, we could just call `SysTickConfig()` and `NVIC_SetPriority()`..

TychoVrahe: changed: https://github.com/trezor/trezor-firmware/commit/f106b8dd53ff6aefb3433d66eb7521aa7c6eda83

cepetr (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.3.0 (21292fda4b99364182a872b4eefbc9986a6da622)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.3.0 Chrome/132.0.6834.194 Electron/34.1.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T2B1 2.8.9 regular (revision 7614a29c9a9cc138fff446937b7d9489fd57821f)
 - Transport: BridgeTransport 3.0.0-bundled.25.3.0

PR #4704, opened 2025-02-28
fix drawing of 1pix bar shape 

### LINKED ISSUE #4702: Drawing of one px bars with both bg and fg color set does not work
Drawing of one px bars with both bg and fg color set does not work:
- the outline doesn't render anything due to cutting thickness on each side
- background doesn't render anything due to area being shrinked by thinckness

In this case, the expected behavior would be to just render single pixel in fg color.

Also, in each case we set both bg and fg to same color, we shouldn't draw outline+background, but rather just background with original size - less DMA2D operations with same result.

cepetr:
This PR resolves #4702 and fixes the prodtest build for T3W1 rev A.

If the background and foreground colors are the same, the bar thickness is not considered. This fixes the problem with 1-pixel bars and also optimizes drawing, as suggested in #4702.

### DIFF
--- a/core/embed/rust/src/ui/shape/bar.rs
+++ b/core/embed/rust/src/ui/shape/bar.rs
@@ -77,13 +77,22 @@ impl Shape<'_> for Bar {
 
         // TODO: fatal_error! in unsupported scenarious
 
-        let th = match self.fg_color {
-            Some(_) => self.thickness,
-            None => 0,
+        let (fg_color, bg_color, th) = if self.fg_color.is_some() {
+            let th = self.thickness;
+            if th * 2 < self.area.width() && th * 2 < self.area.height() {
+                // Draw a rectangle with a border
+                (self.fg_color, self.bg_color, th)
+            } else {
+                // Too thick border => draw a filled rectangle
+                (None, self.fg_color, 0)
+            }
+        } else {
+            // No foreground color => draw a filled rectangle
+            (None, self.bg_color, 0)
         };
 
         if self.radius == 0 {
-            if let Some(fg_color) = self.fg_color {
+            if let Some(fg_color) = fg_color {
                 // outline
                 if th > 0 {
                     let r = self.area;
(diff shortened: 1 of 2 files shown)


core/site_scons/models/T3W1/trezor_t3w1_revA.py
```diff
@@ -84,6 +84,7 @@ def configure(
         features_available.append("ble")
         defines += [("USE_BLE", "1")]
         sources += ["embed/io/nrf/stm32u5/nrf.c"]
+        sources += ["embed/io/nrf/stm32u5/nrf_test.c"]
```
TychoVrahe, line 87: this probably should be done for A0 too, but i think i will drop this revision soon anyway so we can keep it like this, just noting

core/embed/rust/src/ui/shape/bar.rs
```diff
@@ -77,9 +77,10 @@ impl Shape<'_> for Bar {
 
         // TODO: fatal_error! in unsupported scenarious
 
-        let th = match self.fg_color {
-            Some(_) => self.thickness,
-            None => 0,
+        let th = if self.fg_color.is_some() && self.fg_color != self.bg_color {
```
TychoVrahe, line 80: This actually fix the issue (drawing 1px bars when both fg_color and bg_color are set to different color will still not work - or more general problem with drawing bars with thickness larger than half of the size of he bar). That said, this is very low priority so we can merge like this and leave the issue open (with somewhat reduced scope)

TychoVrahe (approved):

core/embed/rust/src/ui/shape/bar.rs
cepetr: Yes, you are right. I think we might fix it now as well. The questing is what to draw then. What about this 31754cb814c1b90c593b1f12b6ec64315a0380be.

TychoVrahe: doesn't work as expected: the `fg_color` and `bg_color` are variables are unused as later there is `self.bg_color`/ `self.fg_color` usage, which should now be just `bg_color`/`fg_color`

cepetr: No way, the code was there! 

fixed, see 608c89fb850db240f3cb9bebb97cde7405127453

TychoVrahe (approved):

PR #4705, opened 2025-02-28
Eckhart UI - split Firmware and Bootloader components and theme

obrusvit:
Split Eckhart components into 3 modules:
1. `component` for shared components
2. `bootloader` for bootloader-specific components and screens
3. `firmware` for firmware-specific components and screens


bieleluk (approved):
LGTM

PR #4720, opened 2025-03-03
fix(core/rust): correctly extend SwipeAttach animation duration

matejcik:
so that it covers the interval from the Attach event until timer running out.

Without this change, it is possible to slip in a TouchStart event before the animation timer has had a chance to start, and lose the matching TouchEnd.

This fixes the problem where a user wildly tapping on the screen can start a hold-to-confirm and have it run without holding.


matejcik:
Note that there's multiple versions of `lazy_start` across the codebase and the same bug probably exists in those others. It's not that important because none of the others wrap a Hold-to-confirm button. But it would be nice to have a general abstraction for AttachAnimation.

PR #4721, opened 2025-03-03
Fix delizia set brightness setting

obrusvit:
[no changelog]


bosomt:
**QA OK**

tested on T3T1 fad9682201cf9289bba2adb66e6e07ed1cf78936 by swiping an taping during `trezorctl set brightness`

PR #4727, opened 2025-03-04
feat(core): double glyphs on t3w1 terminal

cepetr:
This PR improves the readability of text on the terminal on T3W1 by doubling the character width and height and adding a small Y offset.

The terminal window size on T3W1 is now 31x31 characters, which seems sufficient for RSOD purposes.

On other models, the terminal screen remains unchanged.


TychoVrahe (approved):
lgtm

PR #4729, opened 2025-03-04
fix BLE issue command return value

TychoVrahe:
This PR fixes a bug that `ble_issue_command` function always returned false for some commands.


cepetr (approved):
utACK

PR #4734, opened 2025-03-05
Collapse unchanged screenshots in UI diff

ibz:
This makes long flows easier to review in the UI diff.

### DIFF
--- a/tests/ui_tests/reporting/html.py
+++ b/tests/ui_tests/reporting/html.py
@@ -6,10 +6,11 @@ from pathlib import Path
 from typing import Iterable
 
 from dominate import document
-from dominate.tags import a, i, img, table, td, th, tr
+from dominate.tags import a, i, img, span, table, td, th, tr
 
 from ..common import UI_TESTS_DIR
 
+_COLLAPSE_UNCHANGED_THRESHOLD = 10
 _IMAGE_DIR = UI_TESTS_DIR / "images"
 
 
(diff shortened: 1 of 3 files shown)


obrusvit (changes requested):
I think it does not work as expected. I tried the following (non-sensical) diff.
image

What I then see is this:

image

pressing the "show all"
image


What I would expect is this:
- no diff in the test => show all screenshots - useful for browsing through the flow without the need to "show all"
- some diff in the test
  - if there are <= 10 screens total, show all anyway
  - if there are > 10 screens total, collapse all unchanged screenshots, show just the ones with the diff. Optionally, keep the "collapse" button in there to hide it again.

tests/ui_tests/reporting/html.py
```diff
@@ -92,13 +92,51 @@ def image_link(
...
-            image_column(right, cur_dir)
-            diff_column()
+    has_diff = False
+    rows = []
+    for l, r in diff:
+        if l != r:
+            has_diff = True
+        rows.append((l, r))
+
+    if not has_diff or len(rows) <= 10:  # show all without collapsing
```
obrusvit, line 128: Use named constant at the top of the file
`_COLLAPSE_UNCHANGED_THRESHOLD = 10`

tests/ui_tests/reporting/html.py
```diff
@@ -92,13 +92,51 @@ def image_link(
...
+):
+    with tr(bgcolor=bgcolor, _class="hidden" if hidden else ""):
+        image_column(left, cur_dir)
+        image_column(right, cur_dir)
+        diff_column()
+
+
+def collapsible_rows(rows, cur_dir):
+    collapse = len(rows) > 0
+    color = "yellow" if collapse else "white"
```
obrusvit, line 110: Let's not put `"yellow"` for the collapsed rows, it looks strange. Keep just `"white"`. That means you don't need the local var `color`

tests/ui_tests/reporting/html.py
```diff
@@ -92,13 +92,51 @@ def image_link(
...
-        with tr(bgcolor=background, onclick="createDiff(this)"):
-            image_column(left, cur_dir)
-            image_column(right, cur_dir)
-            diff_column()
+    has_diff = False
+    rows = []
+    for l, r in diff:
+        if l != r:
+            has_diff = True
+        rows.append((l, r))
```
obrusvit, line 126: perhaps more pythonic way
```suggestion
    rows = list(diff) 
    has_diff = any(l_hash != r_hash for l_hash, r_hash in rows)
```

obrusvit (commented):
I have a few nits, but otherwise it looks good and seems to work as expected. 

Tested using the same method as I commented in previous review.

obrusvit (approved):
Fix style check and merge.

PR #4735, opened 2025-03-05
add device ID to prodtest

TychoVrahe:
This PR adds device ID storing capability - OTP block is allocated and commands for writing and reading device ID are added to prodtest.

The QR code in prodtest is altered to show device ID instead of batch.


core/embed/projects/prodtest/main.c
```diff
@@ -165,13 +165,11 @@ static void usb_init_all(void) {
...
 static void show_welcome_screen(void) {
-  char dom[32] = {0};
-  // format: {MODEL_IDENTIFIER}YYMMDD
-  if ((sectrue ==
-           flash_otp_read(FLASH_OTP_BLOCK_BATCH, 0, (uint8_t *)dom, 32) &&
-       dom[31] == 0 && cstr_starts_with(dom, MODEL_IDENTIFIER))) {
-    screen_prodtest_info(dom, strlen(dom), dom + sizeof(MODEL_IDENTIFIER) - 1,
-                         strlen(dom) - sizeof(MODEL_IDENTIFIER) + 1);
+  char device_id[32] = {0};
+  if ((sectrue == flash_otp_read(FLASH_OTP_BLOCK_DEVICE_ID, 0,
```
cepetr, line 169: How about this version that uses fewer hardcoded constants and modified condition for detection whether the block is programmed or not:

```c
char device_id[FLASH_OTP_BLOCK_SIZE];

if ((sectrue == flash_otp_read(FLASH_OTP_BLOCK_DEVICE_ID, 0,
                                 (uint8_t *)device_id, sizeof(device_id)) &&
                                 (device_id[0] != 0xFF)) {
    screen_prodtest_info(device_id, strnlen(device_id, sizeof(device_id) - 1));
}
```

core/embed/projects/prodtest/main.c
```diff
@@ -165,13 +165,11 @@ static void usb_init_all(void) {
...
+                                 (uint8_t *)device_id, 32) &&
```
cepetr, line 170: Shouldn't we show batch id as before if the device id is not present?

cepetr (changes requested):

core/embed/projects/prodtest/main.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/717111864a63d8e8934290c1651efceaba6ff789

core/embed/projects/prodtest/main.c
TychoVrahe: i don't think its needed

cepetr (approved):

Issue #4623, opened 2025-02-12
Replace ellipsis icon with "..."


ibz:
In `confirm_value_intro` we should use "..." rather than the arrow icon to indicate that the text is not complete, because using the icon suggests that we are actually paginating through the text, which we are not in this case.

![Image](https://github.com/user-attachments/assets/7b2f5040-87ff-4801-9cc2-e146738343a8)
![Image](https://github.com/user-attachments/assets/ba76ffff-d4d5-42db-b33f-b8f71551a516)
![Image](https://github.com/user-attachments/assets/cb6c0140-143a-463b-8344-b2edef34e10a)

ibz:
@Hannsek Do we *actually* want this for TS5 as well? Or is it clear enough in that case - due to the footer text...

PR #4737, opened 2025-03-05
Use "..." ellipsis on "show_more"

ibz:
Separate issue #4623 as a follow-up of #4601. When confirming long messages the first page should show a classic ellipsis (...) instead of the ellipsis icon which might make the user believe that the *continue* button leads to the next page.

### Model T

image

### TS3

image

### TS5

image


ibz:
@Hannsek Please confirm the UI!

core/embed/rust/src/ui/layout_bolt/theme/mod.rs
```diff
@@ -607,6 +607,11 @@ pub const TEXT_MONO: TextStyle = TextStyle::new(fonts::FONT_MONO, FG, BG, GREY_L
     .with_page_breaking(PageBreaking::CutAndInsertEllipsisBoth)
     .with_ellipsis_icon(ICON_PAGE_NEXT, 0)
     .with_prev_page_icon(ICON_PAGE_PREV, 0);
+pub const TEXT_MONO_WITH_CLASSIC_ELLIPSIS: TextStyle =
```
bieleluk, line 610: The name `TEXT_MONO_WITH_CLASSIC_ELLIPSIS` might be a bit confusing. I would personally use something like  `TEXT_MONO_WITH_END_ELLIPSIS` because that's what makes it different from `TEXT_MONO` style.

core/embed/rust/src/ui/layout_bolt/theme/mod.rs
```diff
@@ -607,6 +607,11 @@ pub const TEXT_MONO: TextStyle = TextStyle::new(fonts::FONT_MONO, FG, BG, GREY_L
...
+    TextStyle::new(fonts::FONT_MONO, FG, BG, GREY_LIGHT, GREY_LIGHT)
+        .with_line_breaking(LineBreaking::BreakWordsNoHyphen)
+        .with_page_breaking(PageBreaking::CutAndInsertEllipsis)
+        .with_prev_page_icon(ICON_PAGE_PREV, 0);
```
bieleluk, line 614: Is this line needed if the style uses only the end ellipsis?

core/embed/rust/src/ui/layout_caesar/theme/mod.rs
```diff
@@ -42,9 +42,15 @@ pub const TEXT_MONO: TextStyle = TextStyle::new(fonts::FONT_MONO, FG, BG, FG, FG
     .with_page_breaking(PageBreaking::CutAndInsertEllipsisBoth)
     .with_ellipsis_icon(ICON_NEXT_PAGE, ELLIPSIS_ICON_MARGIN)
     .with_prev_page_icon(ICON_PREV_PAGE, PREV_PAGE_ICON_MARGIN);
+pub const TEXT_MONO_WITH_CLASSIC_ELLIPSIS: TextStyle =
+    TextStyle::new(fonts::FONT_MONO, FG, BG, FG, FG)
+        .with_page_breaking(PageBreaking::CutAndInsertEllipsisBoth)
```
bieleluk, line 47: The `bolt` text style with the same name had different page breaking: `CutAndInsertEllipsis`. Is there some reason for that?

core/embed/rust/src/ui/layout_delizia/theme/mod.rs
```diff
@@ -741,6 +741,11 @@ pub const TEXT_MONO: TextStyle = TextStyle::new(fonts::FONT_MONO, GREY_EXTRA_LIG
     .with_page_breaking(PageBreaking::CutAndInsertEllipsisBoth)
     .with_ellipsis_icon(ICON_PAGE_NEXT, 0)
     .with_prev_page_icon(ICON_PAGE_PREV, 0);
+pub const TEXT_MONO_WITH_CLASSIC_ELLIPSIS: TextStyle =
+    TextStyle::new(fonts::FONT_MONO, GREY_EXTRA_LIGHT, BG, GREY, GREY)
+        .with_line_breaking(LineBreaking::BreakWordsNoHyphen)
+        .with_page_breaking(PageBreaking::CutAndInsertEllipsisBoth)
```
bieleluk, line 747: Same point as for the `caesar` text style.

core/embed/rust/src/ui/layout_bolt/theme/mod.rs
ibz: > because that's what makes it different from TEXT_MONO style

I think that what makes it special is the fact that the ellipsis is not an icon, but a "classic" ellipsis. :) Do we have a better name for that? You are right that there is also the difference that the ellipsis is only at the end, but I think I could also change that, because it should be irrelevant - what matters is that it gets rendered as "..."

core/embed/rust/src/ui/layout_caesar/theme/mod.rs
ibz: Good catch, I think it should be this way everywhere. The point of the new style is to render the ellipsis as "...", not to not render it for previous page...

bieleluk (approved):

bosomt:
@Hannsek @ibz 

Probably not expected that ... is only used on first page and ↩️ is used on second and third page..

Screenshot 2025-04-22 at 13 40 51
Screenshot 2025-04-22 at 13 40 55
Screenshot 2025-04-22 at 13 40 59

**Info:**
 - Suite version: web 25.5.0 (388c30102f807266a9e52c2f4dbcbdcf417b561e)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/135.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 5120x1440
 - Device: Trezor T3B1 2.8.10 regular (revision 9037c398eb47aa5f5230b076e3ec3e62d8deb1dd)
 - Transport: WebUsbTransport

Hannsek:
It is. Ellipsis on the screen in the main flow indicates that the data is not fully displayed and users should check them in the "side menu flow".

But I admit it might not be clearest on TS3.

bosomt:
@Hannsek one more question, should i send follow up issue ?

see screenshot , signing long BTC message
first screenshot is when signing without showing whole message
second screenshot is when selecting `view all data` in `///` menu


image


image


Hannsek:
I think it is ok, no? What's the problem here?

PR #4740, opened 2025-03-05
WIP: chore(tests): modularize tests buttons module

bieleluk:
DO NOT MERGE


tests/buttons.py
```diff
@@ -1,85 +1,172 @@
...
+# grid coordinates
 def grid(dim: int, grid_cells: int, cell: int) -> int:
     step = dim // grid_cells
     ofs = step // 2
     return cell * step + ofs
 
 
-def grid35(x: int, y: int) -> Coords:
-    return grid(DISPLAY_WIDTH, 3, x), grid(DISPLAY_HEIGHT, 5, y)
+def grid35(x: int, y: int, layout_type: LayoutType) -> Coords:
```
romanz, line 31: Maybe it would simpler to add a "Buttons" class, that will have all the functions here as its methods, e.g.:
```python
class Buttons:
    def __init__(self, layout_type: LayoutType):
        self.width = display_height(layout_type)
        self.height = display_height(layout_type)
        
    def grid35(self, x: int, y: int) -> Coords:
        return grid(self.width, 3, x), grid(self.height, 5, y)
    
    ...
    
    def info(self) -> Coords:
        return (self.mid(), self.bottom())
        
    ...
```
So that you could do something like:
```python
   buttons = Buttons(debug.layout_type)  # can be done once
   debug.click(buttons.cancel())  # no need to pass layout type :)
```
WDYT?

romanz (changes requested):
LGTM, thanks for the refactoring!
Added one suggestion and a few nits.

tests/buttons.py
bieleluk: I have tried to address all your suggestion here:
https://github.com/trezor/trezor-firmware/pull/4740/commits/c93f9746a4df00a0edfa1158bc0ba993e1fad309

tests/buttons.py
```diff
@@ -6,167 +6,237 @@
...
+        return _grid(self.width, 3, 0)
+
+    def _mid(self) -> int:
+        return _grid(self.width, 3, 1)
+
+    def _right(self) -> int:
+        return _grid(self.width, 3, 2)
+
+    # Vertical coordinates
+    def _top(self):
```
romanz, line 56: Please add return type annotation (here and below).

tests/buttons.py
```diff
@@ -6,167 +6,237 @@
...
+
+    def number_input_plus(self) -> Coords:
+        if self.layout_type is LayoutType.Bolt:
+            return (self._right(), _grid(self.height, 5, 1))
+        elif self.layout_type is LayoutType.Delizia:
+            return (self._right(), _grid(self.height, 5, 3))
+        else:
+            raise ValueError("Wrong layout type")
+
+    def word_count_all_word(self, word_count) -> Coords:
```
romanz, line 114: nit: `coords_map.get(word_count)` may return `None` if `word_count` is not in the map.

tests/buttons.py
```diff
@@ -6,167 +6,237 @@
...
+        if self.layout_type is LayoutType.Bolt:
+            return self.grid34(0, 3)
+
+        elif self.layout_type is LayoutType.Delizia:
+            return self.grid34(0, 3)
+
+        else:
+            raise ValueError("Wrong layout type")
+
+    def word_count_rep_word(self, word_count) -> Coords:
```
romanz, line 147: Does `rep` in `word_count_rep_word` mean `repeated`?
If so, maybe it's clearer to use `repeated` instead... (here and below)

tests/click_tests/common.py
```diff
@@ -49,7 +49,8 @@ def get_char_category(char: str) -> PassphraseCategory:
 
 def go_next(debug: "DebugLink") -> LayoutContent:
     if debug.layout_type is LayoutType.Bolt:
-        debug.click(buttons.ok(debug.layout_type))
+        btns = buttons.ScreenButtons(debug.layout_type)
```
romanz, line 52: nit: consider creating a single `buttons.ScreenButtons(debug.layout_type)` object, and re-using it in the tests - e.g. (here and below):
```python
debug.click(debug.buttons.ok())
```
WDYT?

tests/buttons.py
romanz: Thanks! Reviewed below.

tests/buttons.py
romanz: Sorry, I missed the `assert`.

tests/click_tests/common.py
bieleluk: Do you mean to add the function `screen_buttons` to `debuglink.py`?

tests/buttons.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4740/commits/d7edc4ee8a01069a8a2b5db293a4ccf6a24b1400

tests/buttons.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4740/commits/25557467d1c67953b84214363cc9e221abdc7d83

tests/click_tests/common.py
romanz: Sounds good. Maybe even make it a `property`, [similar to `layout_type`](https://github.com/trezor/trezor-firmware/blob/d6c0e7e2035d967bd0cc46b5528f8eb9a9f0534b/python/src/trezorlib/debuglink.py#L469)?

bieleluk:
https://github.com/trezor/trezor-firmware/pull/4740/commits/94e10d0c0a2f184af9ce9a41f98c9a6aaffbf2a7
This commit consolidates `Bolt` and `Delizia` passphrase click tests into one file because the individual files were nearly similar.
The fixtures changes include only file names with one exception: `T3T1_en_test_passphrase_bolt_delizia.py::test_passphrase_click_same_button_many_times`, where the the test was updated according to description: clicking button `abc` instead of `def`

screenshot of old `T3T1_test_passphrase_click_same_button_many_times`:
18e2aeb333cc496f3550c38f98464d60f0ebef0f4800d05cc8edbb6b31c7ed2b
screenshot of updated `T3T1_test_passphrase_click_same_button_many_times`:
image

tests/click_tests/common.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/4740/commits/2b12c179334cbbddac3bf9dee8db787f9abe77b9

python/src/trezorlib/debuglink.py
```diff
@@ -470,6 +471,18 @@ def layout_type(self) -> LayoutType:
         assert self.model is not None
         return LayoutType.from_model(self.model)
 
+    @property
+    def screen_buttons(self) -> ScreenButtons:
```
romanz, line 475: suggestion:
```python
    @property
    def screen_buttons(self) -> ScreenButtons:
        return ScreenButtons(self.layout_type)
        
    @property
    def button_actions(self) -> ButtonActions:
        return ButtonActions(self.layout_type)        
```

bieleluk: https://github.com/trezor/trezor-firmware/pull/4740/commits/7aba2ea2da4b15604f7a05d13a9ec061574e376b

romanz (approved):

tests/click_tests/test_passphrase_bolt_delizia.py
```diff
@@ -32,27 +31,25 @@
     from ..device_handler import BackgroundDeviceHandler
 
 
-pytestmark = pytest.mark.models("delizia")
+pytestmark = pytest.mark.models("t2t1", "delizia")
```
obrusvit, line 34: Should we use "bolt" instead?

obrusvit (approved):
Thanks, looks good.

tests/click_tests/test_passphrase_bolt_delizia.py
bieleluk: according to [pytest readme](https://github.com/trezor/trezor-firmware/tree/main/tests/device_tests#readme), there is no _bolt_ shortcut available

PR #4767, opened 2025-03-11
Crash on label text overflow

ibz:
This will raise an exception when label text doesn't fit so that we can detect overflows in text rendering during tests.

Obviously the PR also comes with actual fixes to make the tests it deliberately broke pass.

Note: This is Delizia only for now, see [this comment](https://github.com/trezor/trezor-firmware/pull/4767#discussion_r2044576778). Will open a separate PR for other models.


core/embed/rust/src/ui/shape/progressive_render.rs
```diff
@@ -140,4 +140,10 @@ where
             unwrap!(self.shapes.push(holder), "Shape list full");
         }
     }
+
+    fn raise_overflow_exception(&mut self) {}
+
+    fn should_raise_overflow_exception(&self) -> bool {
+        false
+    }
```
matejcik, line 148: these should be conditional on `feature = "ui_debug"`

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -304,15 +306,18 @@ impl LayoutObjInner {
...
 
         if self.repaint != Repaint::None {
             self.repaint = Repaint::None;
-            self.root_mut().paint();
-            true
+            if self.root_mut().paint().is_err() {
+                Err(Error::OutOfRange)
+            } else {
+                Ok(true)
+            }
```
matejcik, line 318:
```suggestion
            self.root_mut().paint().map(|_| true)
```

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -594,8 +599,10 @@ extern "C" fn ui_layout_timer(this: Obj, token: Obj) -> Obj {
 extern "C" fn ui_layout_paint(this: Obj) -> Obj {
     let block = || {
         let this: Gc<LayoutObj> = this.try_into()?;
-        let painted = this.inner_mut().obj_paint_if_requested().into();
-        Ok(painted)
+        let Ok(painted) = this.inner_mut().obj_paint_if_requested() else {
+            return Err(Error::OutOfRange);
+        };
+        Ok(painted.into())
```
matejcik, line 605:
```suggestion
        let painted = this.inner_mut().obj_paint_if_requested();
        Ok(painted?.into())
```

core/embed/rust/src/ui/layout/obj.rs
```diff
@@ -146,10 +146,12 @@ where
         self.returned_value.as_ref()
     }
 
-    fn paint(&mut self) {
+    fn paint(&mut self) -> Result<(), Error> {
         render_on_display(None, Some(Color::black()), |target| {
             self.inner.render(target);
         });
+
+        Ok(())
```
matejcik, line 154: any reason the exception is not raised also here? means we're not testing pre-Delizia layouts?

(if that's the intention, please at least make a TODO comment)

matejcik (commented):
LGTM overall, some small notes and then we're ready to go

core/embed/rust/src/ui/layout/obj.rs
ibz: That was not the intention, which led to more crashes on other models, of course. :) Still WIP...

matejcik:
this uncovered some issues that should be fixed separately (please create issues for them), also cc @Hannsek :
---
I think the "address path" string doesn't belong, it should just be the derivation path
also it should not be repeated in the Amount screen
image
---
"Assign with Seed" should be the title of the screen, "System Program" is the subtitle
we can hide "1/1" in all cases. i think it's generally ok to also hide any "m/n" -- given the screen labels, I don't think how much value it brings to the user to see the instruction counter. what do you think @lapohoda @Hannsek ?
image
---
(speaking of, someone should go through Solana programs and improve the labels. they are "names of instructions" but what we want is "user-visible name of operation", which, e.g. in "Transfer Checked" is just "Transfer")

matejcik:
what i'd like to see fixed in this PR:

could we make the space for the label a couple pixels bigger, so that these strings (that previously fit) still fit?
image
image
image

this screen is apparently reusing the same string as the PIN pad.
instead, the screen should be an error with the error icon, and the string "new wipe code" should not be there
image

ibz:
> this uncovered some issues that should be fixed separately (please create issues for them)

#5061 #5062

PR #4770, opened 2025-03-11
Fx Caesar PinEntry to use Title

obrusvit:
- Title component uses Marquee internally which scrolls longer prompt texts
- this fixes e.g. czech translations not fitting the screen

### DIFF
--- a/core/embed/rust/src/ui/layout_caesar/component/input_methods/pin.rs
+++ b/core/embed/rust/src/ui/layout_caesar/component/input_methods/pin.rs
@@ -14,8 +14,8 @@ use crate::{
 };
 
 use super::super::{
-    super::fonts, theme, ButtonDetails, ButtonLayout, CancelConfirmMsg, ChangingTextLine,
-    ChoiceControls, ChoiceFactory, ChoiceItem, ChoiceMsg, ChoicePage,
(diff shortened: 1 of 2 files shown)


ibz (approved):

bosomt:
**QA NOK**

loaded latest debug version of FW and latetss translation from repo but `Nastaveni PIN kodu` does not fit into marquee, no scroll is present, probably due to only last character is partially visible ?

telegram-cloud-photo-size-4-5944817612426955220-w

 - Device: Trezor T3B1 2.9.0 regular (revision 107960ebbe787a86920d5b3d8ac2c804f87bccff)

Hannsek:
I think we can change it to "NASTAVIT PIN KÓD".

bosomt:
**QA OK**

- retested on normal and not debug version of firmware
- `firmware trezorctl set wipe-code` triggers marquee scroll even in English version
- i will retest it in `Missing horizontal scroll when setting up wipe code in CS language #4750` when translations will be available for signed firmwares

**info**
 - Device: Trezor T3B1 2.9.0 regular (revision 8fbe0051861f081809acd80033eba4a519a2bdfd)

obrusvit:
@bosomt FYI, animations (incl. the marquee movement) are not present on debug firmwares.

bosomt:
@obrusvit yeah i already know :))))

PR #4771, opened 2025-03-11
fix(core): use smaller button for upgrade cancellation

romanz:
Otherwise, the translated `Install` text may overflow.

Also, changed the `Confirm` methods signature to minimize copies and reduce stack usage.

### DIFF
--- a/core/.changelog.d/4771.fixed
+++ b/core/.changelog.d/4771.fixed
@@ -1,0 +1,1 @@
+[T2T1] Fixed upgrade confirmation text overflow.
--- a/core/src/trezor/ui/layouts/bolt/__init__.py
+++ b/core/src/trezor/ui/layouts/bolt/__init__.py
@@ -1415,22 +1415,16 @@ async def confirm_firmware_update(description: str, fingerprint: str) -> None:
         verb=TR.buttons__install,
         info=True,
     )
-    info = trezorui_api.confirm_value(
+    info = trezorui_api.show_info_with_cancel(
         title=TR.firmware_update__title_fingerprint,
-        value=fingerprint,
-        description=None,
+        items=(("", fingerprint),),
         chunkify=True,
-        verb=TR.buttons__install,
-        verb_cancel="<",
-        info=False,
-        cancel=True,
     )
     await with_info(
         main,
         info,
         br_name="firmware_update",
         br_code=BR_CODE_OTHER,
-        info_layout_can_confirm=True,
     )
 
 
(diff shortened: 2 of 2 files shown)


core/embed/rust/src/ui/layout_bolt/ui_firmware.rs
```diff
@@ -247,17 +247,18 @@ impl FirmwareUI for UIBolt {
         let title = Label::left_aligned(title_str, theme::TEXT_BOLD).vertically_centered();
         let msg = Label::left_aligned(description, theme::TEXT_NORMAL);
 
-        let left = Button::with_text(TR::buttons__cancel.into()).styled(theme::button_default());
+        let left = Button::with_icon(theme::ICON_CANCEL);
         let right = Button::with_text(TR::buttons__install.into()).styled(theme::button_confirm());
 
-        let layout = RootComponent::new(
-            Confirm::new(theme::BG, left, right, ConfirmTitle::Text(title), msg).with_info(
+        let mut confirm = Confirm::new(theme::BG, left, right, ConfirmTitle::Text(title), msg);
```
obrusvit, line 253: I thought about using a regular `ButtonPage` instead of the `Confirm` which is shared with bootloader. The reason is that the `ButtonPage` already has the `left_button_small` handling implicitly in the `place` method.

However there is that "FW fingerprint" info which does make it slightly more complicated. But how about we use `Frame<ButtonPage>` constructed `.with_info_button()` and stitch two layouts in `layouts/bolt/__init__.py::confirm_firmware_update` with `with_info`?

The benefit is that
1. we can get rid of `bl_confirm` in `FirmwareUI`
2. we don't need to put additional code to handle small left button in `bl_confirm`

obrusvit (commented):
I'd like to discuss a different approach.

core/embed/rust/src/ui/layout_bolt/ui_firmware.rs
romanz: Sounds good, thanks!

romanz: https://github.com/trezor/trezor-firmware/pull/4771/commits/b813c6ab8857fb06d8a071976125c3caff5009e3 & https://github.com/trezor/trezor-firmware/pull/4771/commits/90f4c3728c6d726d1fb09c4a0d99a82794617438

obrusvit (approved):
Looks good.

Thinking about removing the `confirm_firmware_update` from `FirmwareUI` but that's for a different task.

PR #4772, opened 2025-03-11
WIP DO NOT MERGE

bieleluk:
This PR enables usage of per-layout translations in the tests
Current translation function returns `str` for simple translations but `dict` for multi-layout translations, which causes errors in the further process. To select a proper layout translation, one parameter `layout_type` is added to all translation functions.


obrusvit:
This was my approach: https://github.com/trezor/trezor-firmware/pull/4765/commits/fe91a3a41972fd4dfc76093c08eaeff63c7eea39

Issue #3771, opened 2024-05-04
Incorporate manual modifications of font data to gen_font.py


obrusvit:
If we run `core/tools/codegen/gen_font.py` script, it creates a diff in generated font definitions. This is because of manual changes done to individual glyps for better visual appearance on certain models. The changes are documented by commentary but IMHO they should be defined in the script so that manual changes to generated code are not necessary.

Changes are done to 
- font_unifont_regular_16
- `font_pixeloperatormono_regular_8`
- `font_pixeloperator_regular_8`
- `font_pixeloperator_bold_8`

![image](https://github.com/trezor/trezor-firmware/assets/14001709/aa4c0b46-4462-4e61-a3ab-37fb8e74f8c5)

obrusvit:
Note on `font_ttsatoshi_demibold_18/21.c`. Numerals (`0` - `9` mapped to glyph indices `1298`-`1307` as of #3954 . Done manually by modifying `gen_font.py` as follows. A systematic solution would be better.

```python
def _load_char(self, c: str) -> None:
        glyph_map = {
            '0': 1298,
            '1': 1299,
            '2': 1300,
            '3': 1301,
            '4': 1302,
            '5': 1303,
            '6': 1304,
            '7': 1305,
            '8': 1306,
            '9': 1307,
        }

        if c in glyph_map:
            self.face.load_glyph(glyph_map[c], freetype.FT_LOAD_RENDER | freetype.FT_LOAD_TARGET_NORMAL)
        else:
            self.face.load_char(c, freetype.FT_LOAD_RENDER | freetype.FT_LOAD_TARGET_NORMAL)  # type: ignore
```

obrusvit:
Note: this PR touched the relevant files https://github.com/trezor/trezor-firmware/pull/6620

PR #4773, opened 2025-03-12
Use newer version of Delizia fonts

obrusvit:
The newer version of Roboto and TT Satoshi promises tighter glyphs which might ease our screen space constraints a little bit.

TODO:
- [x] alternative style for numerals as described here https://github.com/trezor/trezor-firmware/issues/3771#issuecomment-2183491289 looked completely different. Need to find the mono-spaced indices in the new data.
- [x] fixtures


obrusvit:
Not sure where to find the alternative numeral styles in the newer version of the font. @lapohoda 
image

obrusvit:
I kept the diff (fixtures) for now.

obrusvit:
The diff is mostly negligible but in many cases, it improves the layout a little bit, some examples:
en
image
image

cs
image

obrusvit:
Rebased on top of https://github.com/trezor/trezor-firmware/pull/4796

matejcik (approved):
not gonna block the PR on it, but ... why are the keyboard keys messed up _vertically_?

image

obrusvit:
@matejcik  the vertical shift by 1px comes from `Button::render_content` for the `ButtonContent::Text` variant which used `font.allcase_text_height()` to calculate the y offset. The function calls `visible_text_height("Ay")`.

Previously, the returned height was `19` (shift by `19/2 = 9px`), with the new font data, it results in `20px` (shift by `20/2 = 10px`), which is most probably because the height of `y` glyph was increased from `15` to `16` in the new data.

Anyway, I realized that we can use `text_height()` instead  which just returns a constant. This should not create any new diff because `text_height()` returns `21` (shift by `21/2 = 10px`) but should be slightly more efficient.
https://github.com/trezor/trezor-firmware/pull/4773/commits/6e64b5d5bc585fc2b7967b85c0c4ae3ffee04d6f

bosomt:
**QA OK**

- havent seen any issues with fonts 

PR #4781, opened 2025-03-16
fix(core): use action arg for PIN settings

obrusvit:
Use a better argument, useful for future Eckhart - Description there has too different font.

Diff in Bolt and Caesar -> bold font.
No diff in Delizia.

[no changelog]


obrusvit:
@Hannsek Could you confirm the diff?
Bolt:
image

Caesar:
image

obrusvit:
The correct approach was to use the description differently on Eckhart.

PR #4788, opened 2025-03-18
enable tropic in firmware/kernel on hw

TychoVrahe:
This PR unifies the way how tropic initialization is done in unix and on hw. 

Syscalls are added to enable interaction with firmware. Libtropic is no longer exposed directly to FW. 

Currently the tropic secrets are not fully supported on HW, that is left for next PR.


cepetr (approved):
LGTM and utACK

PR #4782, opened 2025-03-16
Enable GC dump on OOM

romanz:
Requires the following MicroPython patch: https://github.com/trezor/micropython/pull/14

PR #4792, opened 2025-03-18
Revert "fix(core): retry creation of homescreen layout"

romanz:
This reverts commit e0b4cab2db7c316b4d21410b235f10e77caa51c7.

~We can use https://github.com/trezor/trezor-firmware/pull/4782 in case the OOM from https://github.com/trezor/trezor-firmware/pull/4467 reproduces.~


matejcik:
> This reverts commit [e0b4cab](https://github.com/trezor/trezor-firmware/commit/e0b4cab2db7c316b4d21410b235f10e77caa51c7).
> 
> We can use #4782 in case the OOM from #4467 reproduces.

this won't work because #4467 is not a GC OOM.
but nothing stops you sticking a `mem_dump()` into the `except MemoryError` branch, achieving basically the same thing

PR #4793, opened 2025-03-18
fix(core): enable shsi clock after resuming from suspend mode

cepetr:
This PR addresses an issue causing the SAES peripheral to become non-functional after returning from suspend mode, and also introduces minor improvements to the SAES driver:

- Enables SHSI clock after returning from suspend mode
- Completely resets the SAES peripheral in `secure_aes_init()`
- Add `secure_aes_deinit()`


TychoVrahe (approved):
utACK

PR #4797, opened 2025-03-19
perf(core): simplify Button struct to save RAM

romanz:
## This branch:
```
`MnemonicKeyboard<Slip39Input>`: 1236 bytes, alignment: 4 bytes
print-type-size     field `.prompt`: 124 bytes
print-type-size     field `.swipe`: 20 bytes
print-type-size     field `.keys`: 792 bytes
print-type-size     field `.back`: 104 bytes
print-type-size     field `.input`: 192 bytes
print-type-size     field `.can_go_back`: 1 bytes
print-type-size     end padding: 3 bytes
```

## `main` branch:
```
`MnemonicKeyboard<Slip39Input>`: 1280 bytes, alignment: 4 bytes
print-type-size     field `.back`: 108 bytes
print-type-size     field `.input`: 196 bytes
print-type-size     field `.keys`: 828 bytes
print-type-size     field `.prompt`: 124 bytes
print-type-size     field `.swipe`: 20 bytes
print-type-size     field `.can_go_back`: 1 bytes
print-type-size     end padding: 3 bytes
```

### DIFF
--- a/core/embed/rust/src/ui/layout_bolt/component/button.rs
+++ b/core/embed/rust/src/ui/layout_bolt/component/button.rs
@@ -30,7 +30,7 @@ pub struct Button {
     content: ButtonContent,
     styles: ButtonStyleSheet,
     state: State,
-    long_press: Option<Duration>,
+    long_press: Duration,
     long_timer: Timer,
     haptics: bool,
 }
@@ -47,7 +47,7 @@ impl Button {
             touch_expand: None,
             styles: theme::button_default(),
             state: State::Initial,
-            long_press: None,
+            long_press: Duration::ZERO,
             long_timer: Timer::new(),
             haptics: true,
         }
@@ -84,7 +84,7 @@ impl Button {
     }
 
     pub fn with_long_press(mut self, duration: Duration) -> Self {
-        self.long_press = Some(duration);
+        self.long_press = duration;
         self
     }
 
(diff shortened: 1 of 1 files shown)


core/embed/rust/src/ui/component/base.rs
```diff
@@ -426,16 +423,16 @@ impl Timer {
     /// means that _some_ scheduled task might keep running, but this timer
     /// will not trigger when that task expires.
     pub fn stop(&mut self) {
-        self.running = false;
+        self.token = TimerToken::INVALID;
```
matejcik, line 426: can't do this, sorry. this means that stopping and rerunning the same timer in quick succession will issue new timer tokens and that will spam the event queue
(stopping a timer rust-side is invisible to python so the scheduled timer tick remains in the queue, just to "tick" later without invoking anything waiting for it. _starting_ a timer, however, schedules a new tick.
if we keep the same TimerToken, this will reschedule the previous task. if we refresh it, we create a new one.)

if you want to make Timer smaller, either (a) make TimerToken smaller -- u16 could be enough? we don't reissue them all that often.
_in theory_ we could have a long-lived UI object keeping and using an old TimerToken, and get a collision that way... but in practice we _don't_ have long-lived UI objects, right?
the other alternative is to use the highest bit of TimerToken into `running` flag
(but then i'd say `Timer` and `TimerToken` should be merged)

romanz: Thanks! I'll split the `Timer`-related optimization into a separate PR, and leave here only `Button`-related optimization.

matejcik (approved):

PR #4798, opened 2025-03-19
Don't build with Tropic by default

ibz:


matejcik (approved):

matejcik (changes requested):
actually we still want to enable tropic for CI builds

ibz:
> Since we already have `DISABLE_OPTIGA` in Makefile, wouldn't it make more sense to have also `DISABLE_TROPIC`?

Probably yes. I looked at the `THP` flag for inspiration, but you are right that `TROPIC` should be more analogous with `OPTIGA`. I'll use that.

mmilata:
Would it make sense to start (& stop) the `model_server` from emu.py? Assuming here that we want to run device tests with it enabled.

mmilata (approved):
OK let's get this merged so that we don't break trezor-suite tests. However by the time device/click/... tests are enabled we should:
* build the emulator with `DISABLE_TROPIC=0`, and
* start the model server automatically, or start it manually here + update trezor-user-env to do it as well

matejcik:
> OK let's get this merged so that we don't break trezor-suite tests. However by the time device/click/... tests are enabled we should:

we should actually run the CI test suite with DISABLE_TROPIC=1, and run with Tropic enabled only nightly or something.

PR #4800, opened 2025-03-19
fix(core): don't using `'static` lifetime for `BLEEvent`

romanz:
Not sure it is needed. Also, I don't think that the event data argument is `'static`:
https://github.com/trezor/trezor-firmware/blob/45e2a72e243f2ff88d73f7db278e03be6dffa6c2/core/embed/rust/src/ui/layout/obj.rs#L539


mmilata:
Since the array is used for the 6-digit pairing code I wonder if we should just copy it to a `[u8; 6]` (or u32 if the code is base-10) to avoid adding lifetime everywhere? In this case "everywhere" is not too bad, just thinking out loud ...

matejcik:
> Since the array is used for the 6-digit pairing code I wonder if we should just copy it to a `[u8; 6]` (or u32 if the code is base-10) to avoid adding lifetime everywhere? In this case "everywhere" is not too bad, just thinking out loud ...

great idea, and also a neat way to make the `get_buffer` usage safe.

mmilata:
Confirmed with HW that this is going to be 6 decimal digits so u32 should be fine.

romanz:
Pushed ab02b0440: by using `BLEEvent::PairingRequest(u32)`, lifetime annotation is not needed.

matejcik:
following @mmilata's comment, i'd actually suggest pushing this distinction deeper and having the BLE_EVENT result return a python type `tuple[int, int | None]` -- that is, if the data is zero length then use None, and if it is not, ensure that it fits in a small int (effectively uint29 iirc) and create that.

that way we completely avoid allocating in all cases

romanz:
Rebased over `main` (to resolve a minor confilict) and squashed into 573ef1f3ea19a9297a1e552b458cd7ba97f35665.

matejcik (approved):

Issue #4422, opened 2024-12-05
Reliable streaming of translation blobs


matejcik:
Currently the firmware needs to preallocate a chunk of memory the exact size of the translation blob. This is failing -- we're again at the point where memory fragmentation is an issue. We have to investigate that separately.

....in the meantime, there is actually no guarantee that we'll be able to allocate a contiguous chunk at any time.
_Assuming_ that we actually have enough total available memory, we should be able to, with relative ease, make a `rope` structure out of the individual chunks received from the protobuf and use this rope to verify the data integrity while in memory.

We can also go in the opposite direction: expose a function to verify the header separately, and if the header is good, allow writing data (incl. good header) into flash. The [comment](https://github.com/trezor/trezor-firmware/blob/99b2c2a820fb2d009d284eba3b351380bb484cc5/core/src/apps/management/change_language.py#L100-L105) in code reads:
```
    # Loading all the data at once, so we can verify its fingerprint
    # If we saved it gradually to the storage and only checked the fingerprint at the end
    # (with the idea of deleting the data if the fingerprint does not match),
    # attackers could still write some data into storage and then unplug the device.
```
We could relax this restriction, relying on the fact that we verify the fingerprint when booting up anyway and erase the storage if there's a mismatch, so the capability to write something unchecked for a brief period doesn't give the attacker _all that much_ to work with.

PR #4801, opened 2025-03-19
perf(core): allow GC to free confirmation-related objects

romanz:
It may help with https://github.com/trezor/trezor-firmware/issues/4422.


romanz:
Tested locally - it didn't help :(

PR #4807, opened 2025-03-20
button hibernation in prodtest

TychoVrahe:


core/embed/rtl/inc/rtl/cli.h
```diff
@@ -90,6 +90,11 @@ struct cli {
 
   // Current line buffer
   char line_buffer[CLI_LINE_BUFFER_SIZE];
+  int len;     // number of characters in the buffer (excluding '\0')
```
cepetr, line 93: These four variables previously had short names because they were used as local variable in a function. Now we should be more specific - at least with `len` - perhaps `line_line`? Also, to be consistent with the code, it might be better to place the comments before the declaration.

core/embed/rtl/cli.c
```diff
@@ -331,163 +331,170 @@ static char cli_autocomplete(cli_t* cli, const char* prefix) {
...
 
-// Reads a line from the console input and stores it in the `cli->line` buffer
+// Reads next char and processes it.
 //
-// Returns false if the input line is too long
-static bool cli_readln(cli_t* cli) {
+// Returns 1 if the input line is complete,
+// returns 0 if more characters are needed,
+// returns negative value if the input line is too long
+static int cli_process_char(cli_t* cli) {
```
cepetr, line 339: If the function is now called `cli_process_char`, it could take the argument char ch, which decouples it from reading a character.

core/embed/rtl/cli.c
```diff
@@ -331,163 +331,170 @@ static char cli_autocomplete(cli_t* cli, const char* prefix) {
...
   char* buf = cli->line_buffer;
-  int len = 0;     // number of characters in the buffer (excluding '\0')
-  int cursor = 0;  // cursor position in the buffer
 
-  int hist_idx = 0;
-  int hist_prefix = 0;
+  int ch = cli_readch(cli);
```
cepetr, line 342: I would be slightly better to make `cli_readch()` non blocking - if there are no more characters i would return 0.
Then we could process all received characters at once in `cli_run()`:
```c
int res;
do {
  int ch = cli_readch();
  if (ch == 0) { 
    return; 
  }
  res = cli_process_char(ch);
while (res == 0);
```

This would also require to use non-blocking read in vcp callback

core/embed/rtl/cli.c
```diff
@@ -526,80 +533,90 @@ static bool cli_split_args(cli_t* cli) {
   return *cstr_skip_whitespace(buf) == '\0';
 }
 
-void cli_run_loop(cli_t* cli) {
```
cepetr, line 529: Maybe `cli_run()` is not an appropriate name anymore. What about `cli_tick()` or `cli_process_io()`?

core/embed/projects/prodtest/cmd/prodtest_rgbled.c
```diff
@@ -23,6 +23,29 @@
...
+
+void prodtest_rgbled_init(void) {
+  static systimer_t* timer = NULL;
+
+  timer = systimer_create(prodtest_rgbled_timer_cb, NULL);
+
+  if (timer == NULL) {
+    return;
+  }
+  systimer_set(timer, 1000);
```
cepetr, line 45: It seems to me that similar timing logic could be implemented in the loop in prodtest main.c. It might be clearer to have everything in one place.

core/embed/projects/prodtest/main.c
```diff
@@ -248,7 +251,41 @@ int main(void) {
...
+    bool btn = button_get_event(&btn_event);
+
+    if (btn) {
+      if (btn_event.button == BTN_POWER &&
+          btn_event.event_type == BTN_EVENT_DOWN) {
+        btn_deadline = ticks_timeout(1000);
+      }
+      if (btn_event.button == BTN_POWER &&
+          btn_event.event_type == BTN_EVENT_UP) {
+        if (ticks_expired(systick_ms())) {
```
cepetr, line 274: `ticks_expire(systick_ms())` has undefined behaviour. It's mostly `true` but need not to be..

core/embed/projects/prodtest/main.c
```diff
@@ -248,7 +251,41 @@ int main(void) {
...
+          powerctl_hibernate();
+          rgb_led_set_color(0xffff00);
```
cepetr, line 276: Maybe we could introduce constants for the RGB colors.

cepetr (changes requested):

core/embed/rtl/inc/rtl/cli.h
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/e2c12cecb315382a660e13c3551cdba208e40828

core/embed/rtl/cli.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/7623646538ca07df34770513c02c63f492c280bc

core/embed/rtl/cli.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/7623646538ca07df34770513c02c63f492c280bc

core/embed/rtl/cli.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/24a5ea759d4397481200fbabcd94c287ad5fbd73

core/embed/projects/prodtest/main.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/44a8d2524b6e4366cdc80849ac0289c5e9fe4bb6

core/embed/projects/prodtest/main.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/44a8d2524b6e4366cdc80849ac0289c5e9fe4bb6

core/embed/projects/prodtest/cmd/prodtest_rgbled.c
TychoVrahe: cherry picked as https://github.com/trezor/trezor-firmware/commit/78c5dab9c73386999e2d54389cf65522d79a22cc

PR #4810, opened 2025-03-21
remove final wait from report commands

TychoVrahe:
This PR removes an unnecessary wait after last report in prodtests pmic-report and wpc-report commands

### DIFF
--- a/core/embed/projects/prodtest/cmd/prodtest_wpc.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_wpc.c
@@ -151,13 +151,14 @@ static void prodtest_wpc_report(cli_t* cli) {
                  report.opfreq, (int)report.ntc,
                  (int)abs(report.ntc * 1000) % 1000);
 
-    do {
-      if (cli_aborted(cli)) {
-        return;
-      }
-    } while (!ticks_expired(ticks + period));
-
-    ticks += period;
+    if (count > 0) {
+      do {
+        if (cli_aborted(cli)) {
+          return;
+        }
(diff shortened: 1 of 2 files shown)


core/embed/projects/prodtest/cmd/prodtest_wpc.c
```diff
@@ -151,13 +151,17 @@ static void prodtest_wpc_report(cli_t* cli) {
                  report.opfreq, (int)report.ntc,
                  (int)abs(report.ntc * 1000) % 1000);
 
-    do {
-      if (cli_aborted(cli)) {
-        return;
-      }
-    } while (!ticks_expired(ticks + period));
+    if (--count > 0) {
```
cepetr, line 154: Looks like that worked as you intended, but if the count is set to zero, it now means indefinite. It’s not necessarily wrong, but is that what you wanted?

cepetr (changes requested):

core/embed/projects/prodtest/cmd/prodtest_wpc.c
TychoVrahe: not intentional - 0 doesn't make much sense though, we can either consider this behavior a feature, or early exit. I am fine with either

cepetr: I slightly prefer to treat 0 as indicating no measurement. Perhaps you could simplify the logic by leaving the code as it was and simply adding:

```
    if (count == 0) {
      continue;
    }
```

just before waiting for the next period.

cepetr (approved):

PR #4811, opened 2025-03-21
docs: update docs

va-an:
Related to https://github.com/trezor/trezor-firmware/issues/4474#issuecomment-2621173160.

Should I replace all usages of `poetry install`?


va-an:
I've also changed the command in docs for configuring the Rust toolchain.

But I think an even more accurate solution would be to add a [rust-toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) with an explicit nightly specification, which would simplify environment setup, what do you think?

PR #4815, opened 2025-03-24
Handle layout-specific translations in tests

obrusvit:
The PR adds the current Layout (`Bolt`/`Caesar`/`Delizia`/`Eckhart`) to the thread-local context `_CURRENT_TRANSLATIONS` for test runner. This is to handle layout-specific translation assertions in test introduced in https://github.com/trezor/trezor-firmware/pull/4495.

This is `cherry-pick`ed from `ui-eckhart` branch where we encountered the issue and there it works. I thought it would be a good idea to merge separatedly.

### DIFF
--- a/tests/translations.py
+++ b/tests/translations.py
@@ -8,6 +8,7 @@ from pathlib import Path
 
 from trezorlib import cosi, device, models
 from trezorlib._internal import translations
+from trezorlib.debuglink import LayoutType
 from trezorlib.debuglink import TrezorClientDebugLink as Client
 
 from . import common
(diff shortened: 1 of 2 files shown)


tests/translations.py
```diff
@@ -72,9 +73,14 @@ def set_language(client: Client, lang: str, *, force: bool = True):
     with client:
         if not client.features.language.startswith(lang) or force:
             device.change_language(client, language_data)  # type: ignore
+    set_layout(client)
```
romanz, line 76: nit: maybe inline `set_layout()`? (since it's only used once)
```suggestion
    _CURRENT_TRANSLATION.LAYOUT = client.layout_type
```

romanz (approved):
LGTM, small suggestion below:

tests/translations.py
obrusvit: Thanks, I inlined the code on rebase.

romanz (approved):

PR #4825, opened 2025-03-25
fix bootloader build for D002

TychoVrahe:
fix bootloader build for D002 (more generally for models with HW keys but without optiga)


cepetr (approved):
utACK

PR #4826, opened 2025-03-25
Implement anti-exfiltration protocol

onvej-sl:
This pull request implements anti-exfiltration protocol in core for
  * Bitcoin-like transaction signing,
  * Ethereum transaction signing and
  * ownership proofs.


python/src/trezorlib/anti_exfil.py
```diff
@@ -0,0 +1,96 @@
...
+class AntiExfilSignature:
+    signature: bytes
+    entropy: Optional[bytes]
+    nonce_commitment: Optional[bytes]
+
+
+def generate_entropy() -> bytes:
+    import os
+
+    return os.urandom(32)
```
onvej-sl, line 18: I wanted to use `return randbytes(32)` here together with `from random import randbytes` but `make pystyle` fails with `python/src/trezorlib/anti_exfil.py:3: - error: "randbytes" is unknown import symbol (reportAttributeAccessIssue)`.

I'm not sure why. `randbytes` was added in python 3.9.

@matejcik

common/protob/messages.proto
```diff
@@ -105,6 +105,8 @@ enum MessageType {
     MessageType_GetOwnershipProof = 49 [(bitcoin_only) = true, (wire_in) = true];
     MessageType_OwnershipProof = 50 [(bitcoin_only) = true, (wire_out) = true];
     MessageType_AuthorizeCoinJoin = 51 [(bitcoin_only) = true, (wire_in) = true];
+    MessageType_OwnershipProofNonceCommitment = 606 [(bitcoin_only) = true, (wire_out) = true];
+    MessageType_OwnershipProofEntropy = 607 [(bitcoin_only) = true, (wire_in) = true];
```
onvej-sl, line 109: I used some random unoccupied numbers here. I could have probably chosen them better, but I don't fully understand the system behind it.

@matejcik

onvej-sl:
@matejcik Please
  * review introduced protobuf messages (https://github.com/trezor/trezor-firmware/pull/4826/commits/ba06f21a0ea9eccc8088239a6f6105288b86cfcc https://github.com/trezor/trezor-firmware/pull/4826/commits/98c317c34dd0e29a11c4b969383c27e5cb7deee4 https://github.com/trezor/trezor-firmware/pull/4826/commits/cb7ac84d8a1cedd09058f467b4fc41feda80df63)
  * review the interface of the introduced trezorlib functions: `ethereum.sign_tx_new`, `ethereum.sign_tx_eip1559_new` (https://github.com/trezor/trezor-firmware/pull/4826/commits/9ece9ae6c03881d1af9535915ba1dcebb6b2a4d1), `btc.sign_tx` (https://github.com/trezor/trezor-firmware/pull/4826/commits/34309b6d09cea2e13f938a820311690565d9e53b), `btc.get_ownership_proof_new` (https://github.com/trezor/trezor-firmware/pull/4826/commits/c931332b7517207e547a8845f76902f68701b97c).
  * decide what to with the aforementioned functions (keep both the old and the new / break compatibility and keep only the new)

@M1nd3r Please review the rest. Take special care to ensure that the anti-exfiltration protocol is implemented correctly in terms of security in terzorlib.

common/protob/messages-bitcoin.proto
```diff
@@ -638,6 +638,25 @@ message GetOwnershipProof {
...
+    optional bytes nonce_commitment = 1;                                // commitment to signature nonce, used in anti-exfil protocol
+}
+
+/**
+ * Request: Contains the host's entropy, used in anti-exfil protocol
+ * @next OwnershipProof
+ * @next Failure
+ */
+message OwnershipProofEntropy {
+    optional bytes entropy = 1;                                         // host's entropy, used in anti-exfil protocol
```
onvej-sl, line 659: We might want to use a more generic name for the messages that could be reused if we decide to implement an anti-exfiltration protocol for `sign_message` and `sign_identity`.

python/src/trezorlib/anti_exfil.py
matejcik: `secrets.token_bytes` ?

M1nd3r: This is caused by the fact that `pyright` uses python version 3.8 for trezorlib. 
https://github.com/trezor/trezor-firmware/blob/7248bf2a484de1611b906fb7aa915f5d6394d510/python/pyrightconfig.json#L8

matejcik suggest using secret.token_bytes()

onvej-sl: https://github.com/trezor/trezor-firmware/pull/4826/commits/5176d5950c84a1193e26cdc8fa7ded5b0bce6ee2

tests/device_tests/bitcoin/test_signtx.py
```diff
@@ -1659,3 +1660,146 @@ def test_information_replacement(client: Client):
             [out1],
             prev_txes=TX_CACHE_TESTNET,
         )
+
+
+@pytest.mark.models(skip="legacy", reason="Not implemented")
+def test_anti_exfil(client: Client):
```
M1nd3r, line 1666: @onvej-sl Please, add tests/test cases that check the following:
- anit-exfil works correctly with multiple inputs/entropy
- an error is raised (at least in trezorlib) when the length of `entropy_list` is different from the number of inputs
- possibly test with other cryptocurrencies

python/src/trezorlib/btc.py
```diff
@@ -416,7 +562,128 @@ def copy_tx_meta(tx: messages.TransactionType) -> messages.TransactionType:
...
+    the caller should provide the same list of entropies if the signing is repeated
+    due to a error to prevent the device to perform nonce-grinding attacks.
+
+    In addition to the required arguments, it is possible to specify additional
+    transaction properties (version, lock time, expiry...). Each additional argument
+    must correspond to a field in the `SignTx` data type. Note that some fields
+    (`inputs_count`, `outputs_count`, `coin_name`) will be inferred from the arguments
+    and cannot be overriden by kwargs.
+    """
+    anti_exfil_signatures, serialization = sign_tx_common(
```
M1nd3r, line 670: The `serialization` variable is unused. I would either use it or replace it by underscore.

```suggestion
    anti_exfil_signatures, _ = sign_tx_common(
```

tests/device_tests/bitcoin/test_signtx.py
onvej-sl: > * anit-exfil works correctly with multiple inputs/entropy
> * an error is raised (at least in trezorlib) when the length of entropy_list is different from the number of inputs

Done in https://github.com/trezor/trezor-firmware/pull/4826/commits/708d081a578583343ce6f3f71b730998e8b9e716.

> * possibly test with other cryptocurrencies

What other cryptocurrencies you mean?

M1nd3r: In this PR, there are similar tests in Ethereum and ZCash - those were on my mind. 
However, with ZCash, the same function would be tested - so it is not necessary there. And with Ethereum there is only one entropy - so the change does not make a sense either.

The changes in the fixup are suffucient.

python/src/trezorlib/btc.py
```diff
@@ -387,7 +525,21 @@ def copy_tx_meta(tx: messages.TransactionType) -> messages.TransactionType:
...
+            elif use_anti_exfil and res.request_type == R.TXENTROPY:
+                assert res.details.request_index is not None
+                idx = res.details.request_index
+                nonce_commitment = res.details.nonce_commitment
+                nonce_commitment_list[idx] = nonce_commitment
+                if nonce_commitment is None:
+                    # Device requests an entropy without commiting to the nonce before. This is a violation of the anti-exfil protocol.
+                    raise ValueError(f"Nonce commitment for index {idx} not provided")
+                assert entropy_list is not None
+                msg.entropy = messages.TxEntropyType(entropy=entropy_list[idx])
```
M1nd3r, line 542: I would like to know, why is another message `TxEntropyType` used - instead of having the entropy as "optional bytes" protobuf-field. What are the benefits of this approach?

M1nd3r (approved):
I have reviewed the protocol from security pov and all my objections have been resolved. Good job!

Note: A part of the security depends on the 3rd party library `secp256k1-zkp`, which was not thoroughly reviewed in the process.

common/protob/messages-bitcoin.proto
```diff
@@ -638,6 +638,25 @@ message GetOwnershipProof {
     optional bool user_confirmation = 5 [default=false];                // show a confirmation dialog and set the "user confirmation" bit in the proof
     repeated bytes ownership_ids = 6;                                   // list of ownership identifiers in case of multisig
     optional bytes commitment_data = 7 [default=''];                    // additional data to which the proof should commit
+    optional bytes entropy_commitment = 8;                              // commitment to host's entropy, non-empty value indicates anti-exfil protocol will be used
+}
+
+/**
+ * Response: Contains the commitment to the signature nonce, used in anti-exfil protocol
+ * @next OwnershipProofEntropy
+ * @next Failure
```
onvej-sl, line 647: Delete.

PR #4827, opened 2025-03-25
fix(core): allow firmware upgrade even if language change failed

romanz:

### DIFF
--- a/core/src/apps/management/change_language.py
+++ b/core/src/apps/management/change_language.py
@@ -102,6 +102,7 @@ async def do_change_language(
     # If we saved it gradually to the storage and only checked the fingerprint at the end
     # (with the idea of deleting the data if the fingerprint does not match),
     # attackers could still write some data into storage and then unplug the device.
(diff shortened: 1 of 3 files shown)


core/src/apps/management/change_language.py
```diff
@@ -13,6 +13,10 @@
 _CHUNK_SIZE = const(1024)
 
 
+class AllocationFailed(MemoryError):
```
matejcik, line 16: don't see why we need a separate exception? you may as well handle `MemoryError` in the try-except block below

romanz: I have added a separate exception just to be sure we skip an OOM originating from blob allocation (and not some other allocation done in the rest of the code), but indeed it is better to ignore any `MemoryError` originating the language change flow during firmware upgrade.

romanz: 82e68d7e0ad7514c70c41fb1a2bff1bb051c0ede

matejcik (approved):

PR #4828, opened 2025-03-26
chore(docs): update SECURITY.md

tsusanka:
We can also delete the whole thing as it will then default to https://github.com/trezor/.github/blob/master/SECURITY.md. But maybe we want to have a `SECURITY.md` file here.

This is not great, we are working on some better security page on trezor.io with better guidance and past issues to be delivered in Q2/Q3.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,15 +2,6 @@
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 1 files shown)


tsusanka:
Let's merge this please. cc @matejcik

matejcik (approved):

PR #4840, opened 2025-03-28
add BLE unpair cmd & message

TychoVrahe:
This PR adds unpair command to trezorctl, corresponding protobuf message and a simple workflow. Message has a parameter to select whether current device or all devices should be unpaired. 

Currently this only unpairs BLE, THP should be addressed separately, or we will modify this when we know more about what is rrequired.


mmilata (approved):
LGTM, maybe also cherrypick https://github.com/trezor/trezor-firmware/commit/91a697886ab0aa742120d8c61dd23e8aa5782555

PR #4842, opened 2025-03-28
chore(vendor): upgrade MicroPython to use non-reallocateble dicts

romanz:
Requires https://github.com/trezor/micropython/pull/17.


romanz:
Currently, `no_realloc` is disabled for emulator build (due to core unittests and profiling support).

matejcik:
> Currently, `no_realloc` is disabled for emulator build (due to core unittests and profiling support).

that's less than useful

could we at least enable it for the frozen build? the test suite is where we would normally discover the overflows

matejcik:
in fact it's a good idea to only enable this flag in debug builds -- in production, we'll prefer performance degradation (fragmented memory) to a full crash

core/embed/projects/firmware/main.c
```diff
@@ -76,6 +76,7 @@ int main(uint32_t cmd, void *arg) {
   // Interpreter init
   printf("CORE: Starting interpreter\n");
   mp_init();
+
```
matejcik, line 79: maybe omit this noise?

core/embed/rust/src/micropython/map.rs
```diff
@@ -19,9 +19,10 @@ impl Map {
         // .all_keys_are_qstrs = 1,
         // .is_fixed = 1,
         // .is_ordered = 1,
+        // .no_realloc = 1,
         // .used = MP_ARRAY_SIZE(table_name),
         // .alloc = MP_ARRAY_SIZE(table_name),
-        let bits = 0b111 | N << 3;
+        let bits = 0b1111 | N << 4;
```
matejcik, line 25: could you please have a look at bindgen, to see whether this limitation was lifted?

core/embed/projects/firmware/main.c
romanz: Oops, my bad... I removed the code from firmware build, instead of disabling it for non-debug builds.

romanz:
> could we at least enable it for the frozen build? the test suite is where we would normally discover the overflows

It seems that most of our emulator device tests are using profiling, so they fail when we disallow reallocations :(
image

matejcik:
does the failure come from `__main__` or `sys.modules`?

if the latter, we can just bump the limit

if `__main__`, that's a problem because it means that the profiler `__main__` is different from `main.py`, correct? so even if we tweak the prealloc size, it won't catch the actual behavior that we care about.

iirc profiling only works if you make a special build for it? if yes, let's freeze `__main__` only when not profiling. technically it would be nice to run a test suite in CI for non-profiling but with frozen `__main__`; practically it will be enough to run it locally and on HW

if there's no build flag for profiling, then uhhh let's not spend more time on it and only enable it on hw debug

romanz:
IIUC, we always build the emulator with [tracing enabled](https://github.com/trezor/trezor-firmware/blob/101d584aa4bd0902198dddde6b72f72ef5a45918/core/embed/projects/unix/mpconfigport.h#L200) so I think that we can simplify the PR by periodically checking for reallocations - as currently done by `utils.unimport_end()`.

The new code in https://github.com/trezor/trezor-firmware/pull/4842/commits/24a328403df8add1071211f0d7a96458faa5784d uses the pre-allocation constants from `mpconfigport.h`, and skips the check if [the tracing profiler](https://github.com/trezor/trezor-firmware/blob/101d584aa4bd0902198dddde6b72f72ef5a45918/core/prof/prof.py#L129) is detected in run-time.

WDYT?

core/embed/rust/src/micropython/map.rs
romanz: https://github.com/trezor/trezor-firmware/pull/4842/commits/24a328403df8add1071211f0d7a96458faa5784d doesn't require changing MicroPython code & its Rust bindings.

matejcik:
yeah, this approach makes more sense
with regard to `main.py`, you could also look up `main` in sys.modules and check the capacity of that, which should work in the profiler case

matejcik (approved):

romanz:
It seems that there is no `main` module on firmware - so this PR causes the firmware to not boot :(
Opened https://github.com/trezor/trezor-firmware/pull/4872 to fix this issue.

romanz:
It has caused also the CI hardware tests to fail.

PR #4843, opened 2025-03-28
Introduce the device menu

ibz:
This PR introduces the "device menu", which is the main menu accessed by tapping the action bar on the home screen. It also includes Rust layouts for the BT pairing (which are triggered from the said device menu) - initially it was a separate PR, but decided to do is all here to simplify building the actual BT pairing before this gets merged. 

The device menu consists of multiple "subscreens" which are either **generic submenus** or **menus specific to a connected Bluetooth device** allowing the user to disconnect, or **the about screen**. Or anything else, really.

However, the actual subscreens are constructed on-demand as the user navigates to the desired screen. Creating a new `DeviceMenuScreen` object will just allocate the memory needed to store the information that will be used to instantiate these subscreens when needed rather than allocating the subscreens themselves.

Note: we currently only support a device vector of length `1` - that is, there is just one device that the Trezor can be paired with at one time, however, the mechanism is in place such that more devices can be supported by the menu structure.

Issues not covered by this PR:
 * for each device in the device list, pass a boolean that specifies whether a device is connected or not (still to be confirmed whether we even have this information!) OR simply remove the "Connected" and "N connected devices" subtexts (probably best option for now, until we have more information?)
 * support multi-line menu items (use #4226 for inspiration!) - although the device menu looks bad without this, it is an issue with `Button`, so outside the scope of this PR
 * translations strings (all strings are currently hardcoded, as the menu structure is still subject to heavy changes)

### DIFF
--- a/core/embed/rust/src/ui/api/firmware_micropython.rs
+++ b/core/embed/rust/src/ui/api/firmware_micropython.rs
@@ -802,6 +802,19 @@ extern "C" fn new_show_homescreen(n_args: usize, args: *const Obj, kwargs: *mut 
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_show_device_menu(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let failed_backup: bool = kwargs.get(Qstr::MP_QSTR_failed_backup)?.try_into()?;
+        let battery_percentage: u8 = kwargs.get_or(Qstr::MP_QSTR_battery_percentage, 0)?;
+        let paired_devices: Obj = kwargs.get(Qstr::MP_QSTR_paired_devices)?;
+        let paired_devices: Vec<TString, 1> = util::iter_into_vec(paired_devices)?;
+        let layout = ModelUI::show_device_menu(failed_backup, battery_percentage, paired_devices)?;
+        let layout_obj = LayoutObj::new_root(layout)?;
+        Ok(layout_obj.into())
+    };
+    unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
+}
+
 extern "C" fn new_show_info(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
     let block = move |_args: &[Obj], kwargs: &Map| {
         let title: TString = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
@@ -1559,6 +1572,15 @@ pub static mp_module_trezorui_api: Module = obj_module! {
     /// ) -> LayoutObj[UiResult]:
     ///     """Idle homescreen."""
     Qstr::MP_QSTR_show_homescreen => obj_fn_kw!(0, new_show_homescreen).as_obj(),
+
+    /// def show_device_menu(
+    ///     *,
+    ///     failed_backup: bool,
+    ///     battery_percentage: int,
+    ///     paired_devices: Iterable[str],
+    /// ) -> LayoutObj[UiResult]:
+    ///     """Show the device menu."""
+    Qstr::MP_QSTR_show_device_menu => obj_fn_kw!(0, new_show_device_menu).as_obj(),
 
     /// def show_info(
     ///     *,
--- a/core/embed/rust/src/ui/api/firmware_micropython.rs
+++ b/core/embed/rust/src/ui/api/firmware_micropython.rs
@@ -815,6 +815,30 @@ extern "C" fn new_show_device_menu(n_args: usize, args: *const Obj, kwargs: *mut
     unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
 }
 
+extern "C" fn new_show_pairing_device_name(
+    n_args: usize,
+    args: *const Obj,
+    kwargs: *mut Map,
+) -> Obj {
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let device_name: TString = kwargs.get(Qstr::MP_QSTR_device_name)?.try_into()?;
+        let layout = ModelUI::show_pairing_device_name(device_name)?;
+        let layout_obj = LayoutObj::new_root(layout)?;
+        Ok(layout_obj.into())
+    };
+    unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
+}
+
+extern "C" fn new_show_pairing_code(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
+    let block = move |_args: &[Obj], kwargs: &Map| {
+        let code: TString = kwargs.get(Qstr::MP_QSTR_code)?.try_into()?;
+        let layout = ModelUI::show_pairing_code(code)?;
+        let layout_obj = LayoutObj::new_root(layout)?;
+        Ok(layout_obj.into())
+    };
+    unsafe { util::try_with_args_and_kwargs(n_args, args, kwargs, block) }
+}
+
 extern "C" fn new_show_info(n_args: usize, args: *const Obj, kwargs: *mut Map) -> Obj {
     let block = move |_args: &[Obj], kwargs: &Map| {
         let title: TString = kwargs.get(Qstr::MP_QSTR_title)?.try_into()?;
@@ -1581,6 +1605,20 @@ pub static mp_module_trezorui_api: Module = obj_module! {
     /// ) -> LayoutObj[UiResult]:
     ///     """Show the device menu."""
     Qstr::MP_QSTR_show_device_menu => obj_fn_kw!(0, new_show_device_menu).as_obj(),
+
+    /// def show_pairing_device_name(
+    ///     *,
+    ///     device_name: str,
+    /// ) -> LayoutObj[UiResult]:
+    ///     """Pairing device: first screen (device name)."""
+    Qstr::MP_QSTR_show_pairing_device_name => obj_fn_kw!(0, new_show_pairing_device_name).as_obj(),
+
+    /// def show_pairing_code(
+    ///     *,
+    ///     code: str,
+    /// ) -> LayoutObj[UiResult]:
+    ///     """Pairing device: second screen (pairing code)."""
+    Qstr::MP_QSTR_show_pairing_code => obj_fn_kw!(0, new_show_pairing_code).as_obj(),
 
     /// def show_info(
     ///     *,
--- a/core/embed/rust/src/ui/api/firmware_micropython.rs
+++ b/core/embed/rust/src/ui/api/firmware_micropython.rs
@@ -806,9 +806,17 @@ extern "C" fn new_show_device_menu(n_args: usize, args: *const Obj, kwargs: *mut
     let block = move |_args: &[Obj], kwargs: &Map| {
         let failed_backup: bool = kwargs.get(Qstr::MP_QSTR_failed_backup)?.try_into()?;
         let battery_percentage: u8 = kwargs.get_or(Qstr::MP_QSTR_battery_percentage, 0)?;
+        let firmware_version: TString = kwargs.get(Qstr::MP_QSTR_firmware_version)?.try_into()?;
+        let device_name: TString = kwargs.get(Qstr::MP_QSTR_device_name)?.try_into()?;
         let paired_devices: Obj = kwargs.get(Qstr::MP_QSTR_paired_devices)?;
         let paired_devices: Vec<TString, 1> = util::iter_into_vec(paired_devices)?;
-        let layout = ModelUI::show_device_menu(failed_backup, battery_percentage, paired_devices)?;
+        let layout = ModelUI::show_device_menu(
+            failed_backup,
+            battery_percentage,
+            firmware_version,
+            device_name,
+            paired_devices,
+        )?;
         let layout_obj = LayoutObj::new_root(layout)?;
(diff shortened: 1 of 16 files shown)


obrusvit (commented):
@bieleluk can you take a look when this PR is ready for review?

mmilata:
This is looking promising but the bad news is that on TS7 hardware pressing the homescreen button causes stack overflow. When i set `MAX_SUBMENUS = 5` and drop the menus I don't need it works. I suppose one option is to construct the VerticalMenuScreens on demand and keep only `MAX_DEPTH` vector of them. Or maybe only the currently visible one if there's not much state. Or use `GcBox` to split the large object in multiple smaller ones.

According to `rust-type-sizes.log` the 10-item menu takes some 16KiB on x86_64, on arm it will be a bit less but the file's empty for some reason on my system.

ibz:
> the bad news is that on TS7 hardware pressing the homescreen button causes stack overflow

Thanks for giving the the bad news earlier.

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
```diff
@@ -69,11 +75,33 @@ impl VerticalMenu {
         self
     }
 
-    pub fn item(mut self, button: Button) -> Self {
+    pub fn button(mut self, button: Button) -> Self {
         unwrap!(self.buttons.push(button));
         self
     }
 
+    pub fn item(
```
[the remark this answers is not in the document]
ibz: Yep, changed this!

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -0,0 +1,497 @@
...
+    CheckBackup,
+    WipeDevice,
+
+    // Device menu
+    ScreenBrightness,
+
+    Close,
+}
+
+struct MenuItem {
```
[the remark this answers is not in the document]
ibz: It would be possible, but what is the benefit? Seems like we would be storing more information in memory than the minimum necessary...

core/embed/rust/src/ui/api/firmware_micropython.rs
```diff
@@ -1560,6 +1573,15 @@ pub static mp_module_trezorui_api: Module = obj_module! {
     ///     """Idle homescreen."""
     Qstr::MP_QSTR_show_homescreen => obj_fn_kw!(0, new_show_homescreen).as_obj(),
 
+    /// def show_device_menu(
+    ///     *,
+    ///     failed_backup: bool,
+    ///     battery_percentage: bool,
```
mmilata, line 1579:
```suggestion
    ///     battery_percentage: int,
```

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -0,0 +1,515 @@
...
+    ),
+
+    // The about screen
+    AboutScreen,
+}
+
+pub struct DeviceMenuScreen<'a> {
+    bounds: Rect,
+
+    battery_percentage: usize,
```
mmilata, line 135: `usize` is for length/indexing, I'd go with u8 here

core/src/apps/homescreen/__init__.py
```diff
@@ -57,15 +57,30 @@ async def homescreen() -> None:
     try:
         res = await obj.get_result()
         if res is INFO:
```
obrusvit, line 59: Omission on my part, processing of the `res` should happen after the `try` block. This is to make sure that `Homescreen` layout is properly dropped.

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -84,8 +119,13 @@ impl Button {
...
     }
 
-    pub const fn styled(mut self, styles: ButtonStyleSheet) -> Self {
-        self.styles = styles;
+    pub const fn styled(mut self, stylesheet: ButtonStyleSheet) -> Self {
+        self.stylesheet = stylesheet;
+        self
+    }
+
+    pub const fn subtext_green(mut self) -> Self {
```
bieleluk, line 127: We should use the same approach as the we are using for the "main" stylesheet above, i.e.:
`pub const fn with_subtext_style(mut self, stylesheet: ButtonStyleSheet) -> Self`

core/embed/rust/src/ui/layout_eckhart/firmware/header.rs
```diff
@@ -138,6 +138,19 @@ impl Header {
         }
     }
 
+    pub fn with_battery(self, percentage: u8) -> Self {
+        let is_low_battery = percentage < 20; // TODO
```
bieleluk, line 142: As this function is strictly device menu-related, it should be moved there

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
     },
 };
 
-use super::super::theme;
+use super::super::{fonts, theme};
```
bieleluk, line 16: import of `fonts` is needed only for the `bootloader` config. For other config, it generates warning:
`warning: unused import: fonts`

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -0,0 +1,515 @@
...
+            Subscreen::DeviceScreen(..) => self.paired_device_screen.place(bounds),
+            Subscreen::AboutScreen => self.about_screen.place(bounds),
+        };
+
+        bounds
+    }
+
+    fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
+        // Update the menu when the screen is attached
+        if let Event::Attach(AttachType::Initial) = event {}
```
bieleluk, line 460: Does this line of code do something?

bieleluk (changes requested):

core/src/apps/homescreen/__init__.py
```diff
@@ -57,15 +57,30 @@ async def homescreen() -> None:
...
-            await raise_if_not_confirmed(
-                error_popup(
-                    "Not implemented",
-                    "DeviceMenu not yet implemented",
-                    button="OK then..",
-                ),
-                "device_menu",
-            )
+
+            ##### MOCK DATA
```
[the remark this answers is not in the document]
ibz: Yep, let's ignore this file please!

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -788,6 +791,43 @@ impl FirmwareUI for UIEckhart {
...
+        Ok(layout)
+    }
+
+    fn show_pairing_code(code: TString<'static>) -> Result<impl LayoutMaybeTrace, Error> {
+        let text: TString<'static> = "Pairing code match?".into();
+        let mut ops = OpTextLayout::new(theme::firmware::TEXT_REGULAR);
+        ops = ops.text(text, fonts::FONT_SATOSHI_REGULAR_38);
+        ops = ops.newline().newline().newline();
+        ops = ops.alignment(Alignment::Center);
+        ops = ops.text(code, fonts::FONT_SATOSHI_EXTRALIGHT_72);
```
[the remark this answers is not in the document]
ibz: There seems to be an issue in mixing chunkification with `Alignment::Center` - it basically breaks the alignment. So I decided to just go with a simpler design and convinced @lapohoda to pretend it's fine (for now). I think this will have to be re-addressed later and either:
 * make chunkification work with center alignment (actually this would still not solve the whole issue, since we really want 2 chunks of chunkified text here: "1|2|3||4|5|6" - would this even be possible)
 * inject spaces into the string - this *should* be trivial, except it is not in our heapless Rust (unless I am missing some obvious way to do it) - so it would have to be done in Python, but I was trying to avoid that so that I keep the interface clean for the caller, who should not worry about this. Alternatively, it we might maybe allocate a new string in `new_show_pairing_code`?
 * separate `.text(..)` and `.offset(..)` calls for each digit - this should work...

If all the above fails, create a new component that does the thing properly. I was trying to avoid this for now, as suggested by @obrusvit .

So, long story short ... let's keep it as it is for this iteration, and get back to it later.

core/embed/rust/src/ui/layout_eckhart/component/button.rs
ibz: The problem with this seems to be the different font sizes on the bootloader (see the constants defined at the top of `impl Button`) - basically if I let this accept *any* style, then that style it accepts has to be based on the `"bootloader"` feature anyway. Restricting this just to green/non-green allowed me to use the same mechanism with having default constants and keeping these constants private. I know it's ugly though, I also very much dislike having this `X_green` function, but I did not find another way, unfortunately.

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
ibz: Good catch! Seems it is useless...

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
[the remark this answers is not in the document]
obrusvit: I'd vote for keeping the design simplified for now.

However, I noticed in `mmilata/pairing` that there is a need to address BLE events inside pairing screens bb5adc331a4b31b872a804e88d53096530e38798 

As I think the `TextScreen` should be as simple as possible and handling bt events should not be there, this is the tipping point for a new component `PairingScreen` or `BLEPairingScreen`.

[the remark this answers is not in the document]
ibz: > The current design has larger spaces between digits than the `FONT_SATOSHI_EXTRALIGHT_72` has. We should use slightly different style, something like `FONT_SATOSHI_EXTRALIGHT_72.with_chunks(Chunks::new(1, 8))`

Yes, that's no good. I'll make a new component, once all this plus @mmilata 's code is in - which is good enough for a 1st version - and move BLE event handling in the new component.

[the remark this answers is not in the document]
mmilata: I think that for the BLE events it would probably be cleaner to make a simple wrapper component that will handle the `BLEEvent`s and passes through everything else to the internal component.

mmilata:
> for each device in the device list, pass a boolean that specifies whether a device is connected or not

Yeah, right now we only know how many paired devices there are and if any unspecified of them is connected. For the purpose of this PR I think the current API is OK, the calling code can either pass 1-element or an empty list.

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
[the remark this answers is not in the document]
mmilata: Moved BLE handling out of `TextScreen`: https://github.com/trezor/trezor-firmware/pull/4860/commits/1f8e08e270a8b010c794e7c5f2b2bc1a208b27f8

core/embed/rust/src/ui/layout_eckhart/component/button.rs
bieleluk: It is a big improvement, but I see a little little room also on `DeviceMenu` side to catch up with the `Button` change:
 If you change the `MenuItem` subtext field to
 `subtext: Option<(TString<'static>, Option<TextStyle>)>,)` 
 then you can keep only one subtext builder function:
```
 pub fn with_subtext(mut self, subtext: Option<(TString<'static>, Option<TextStyle>)>) -> Self {
        self.subtext = subtext;...
```
 and you can simplify the `build_active_subscreeen` by getting rid of one branch:
```
 let button = if let Some((subtext, subtext_style)) = item.subtext {
                        Button::new_menu_item_with_subtext(
                            item.text,
                            item.stylesheet,
                            subtext,
                            subtext_style,
                        ) ...
```
Then you just need to update few item push calls to i.e.:
`items.push(MenuItem::new....with_subtext(Some((device_name, None))))` or
`items.push(MenuItem::new....with_subtext(Some((device_name, Some(Button::SUBTEXT_STYLE_GREEN)))))`

core/embed/rust/src/ui/layout_eckhart/firmware/header.rs
bieleluk: I definitely do not insist.
My thinking was that there will definitely be changes (i.e. adding animation) to the battery icon and its implementation would then be scattered among the device menu and `Header`. If more components are going to use it, then it's obviously the right way but if not then it's debatable.

obrusvit: Per current design, the battery indication is solely in the DeviceMenu. It might change though but as for now, the better approach would probably be to construct the `Header::...::with_icon(...)` using the params coming to the `DeviceMenuScreen`

core/embed/rust/src/ui/layout_eckhart/component/button.rs
ibz: Yep, much better! fa97284c811053185185fcc016cbd7575175a0ba

bieleluk (approved):
LGTM
The failing click tests are reported in [4855](https://github.com/trezor/trezor-firmware/issues/4855) because they should not interact with the menu at all. But for now, please update the fixtures and squash all fixups.

PR #4847, opened 2025-03-30
Eckhart brightness screen

bieleluk:


obrusvit (commented):
It works nicely :+1: 

The only problem I see is that when you release the finger outside the slider, the layout is not confirmed. I'm not sure what the proper behavior should be though. Raising a question in Figma.

bieleluk:
I implemented this behavior on purpose from my experience with the mobile apps. It is the only way how not to confirm when you decide not to.

obrusvit (approved):

bosomt:
QA OK

works as described in PR
design matches Figma design
tested on https://github.com/trezor/trezor-firmware/commit/7fc30de46e070fe83cb2e047f104552bc97fecc3 EMU build

PR #4848, opened 2025-03-30
fix(core): explicitly close `default_task` generator

romanz:
Otherwise, `homescreen()` generator is not guaranteed to be closed, thus the `await` below may not return):
https://github.com/trezor/trezor-firmware/blob/84393631c883b89e468cfca148a53e234c9c320e/core/src/apps/homescreen/__init__.py#L49-L58

and the statically-allocated `ImageBuffer` won't be dropped:
https://github.com/trezor/trezor-firmware/blob/84393631c883b89e468cfca148a53e234c9c320e/core/embed/rust/src/ui/shape/utils/imagebuf.rs#L33-L44

This issue can be reproduced by running:
```
$ TREZOR_MODEL=T3T1 BITCOIN_ONLY=0 QUIET_MODE=1 PYOPT=0 make -C core build_unix_frozen
$ core/emu.py -ea --profiling -c pytest -x tests/device_tests/test_msg_applysettings.py::test_apply_homescreen_jpeg
```

It should be the fix of #4467's root cause.

### DIFF
--- a/core/src/session.py
+++ b/core/src/session.py
@@ -26,5 +26,7 @@ wire.setup(usb.iface_wire)
 # start the event loop
 loop.run()
 
+workflow.close_default()
+
 if __debug__:
     log.debug(__name__, "Restarting main loop")
--- a/core/src/trezor/ui/layouts/homescreen.py
+++ b/core/src/trezor/ui/layouts/homescreen.py
@@ -6,36 +6,9 @@ from storage.cache_common import APP_COMMON_BUSY_DEADLINE_MS
 from trezor import TR, ui, utils
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, Iterator, ParamSpec, TypeVar
+    from typing import Any, Iterator
 
     from trezor import loop
-
-    P = ParamSpec("P")
-    R = TypeVar("R")
-
-
-def _retry_with_gc(layout: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
-    """Retry creating the layout after garbage collection.
-
-    For reasons unknown, a previous homescreen layout may survive an unimport, and still
-    exists in the GC arena. At the time a new one is instantiated, the old one still
-    holds a lock on the JPEG buffer, and creating a new layout will fail with a
-    MemoryError.
-
-    It seems that the previous layout's survival is a glitch, and at a later time it is
-    still in memory but not held anymore. We assume that triggering a GC cycle will
-    correctly throw it away, and we will be able to create the new layout.
-
-    We only try this once because if it didn't help, the above assumption is wrong, so
-    no point in trying again.
-    """
-    try:
-        return layout(*args, **kwargs)
-    except MemoryError:
-        import gc
-
-        gc.collect()
-        return layout(*args, **kwargs)
 
 
 class HomescreenBase(ui.Layout):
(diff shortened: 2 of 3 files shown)


romanz:
Tested on a T3T1 device with universal and BTC-only builds:
```
=============================== 1322 passed, 167 skipped, 89 deselected, 8 xfailed in 11001.42s (3:03:21) ================================

================================ 549 passed, 942 skipped, 89 deselected, 6 xfailed in 3357.21s (0:55:57) =================================
```

matejcik:
this is nice and all but it still doesn't explain why the object is not freed (and `__del__` implicitly called) by the followup `gc.collect()`
or does `trezor.workflow` not get unimported?

romanz:
> why the object is not freed (and __del__ implicitly called) by the followup gc.collect()

I still am not sure, but it seems to be triggered by (seemingly) unrelated byte-code changes... For example, it didn't happen on 25ce4e3fc, but adding the following patch :
```diff
diff --git a/core/src/trezor/utils.py b/core/src/trezor/utils.py
index 3fcb47af3..ef19aa345 100644
--- a/core/src/trezor/utils.py
+++ b/core/src/trezor/utils.py
@@ -84,6 +84,7 @@ def unimport_end(mods: set[str], collect: bool = True) -> None:
     if collect:
         gc.collect()
 
+X = 12
 
 class unimport:
     def __init__(self) -> None:
```
will cause it to reproduce :confused: - with the following build and test:
```
$ TREZOR_MODEL=T3T1 BITCOIN_ONLY=0 QUIET_MODE=1 PYOPT=0 make -C core build_unix_frozen
$ core/emu.py -ea --profiling -o /tmp/trezor.log -c pytest -xv tests/device_tests/reset_recovery/test_recovery_bip39_t2.py
```

I also have found that adding:
```diff
diff --git a/core/prof/prof.py b/core/prof/prof.py
index 2ddf29939..3405c637f 100644
--- a/core/prof/prof.py
+++ b/core/prof/prof.py
@@ -43,7 +43,8 @@ class _Prof:
         #     print(event, frame.f_code.co_filename, frame.f_lineno)
 
         if event == "line":
-            self.__coverage.line_tick(frame.f_code.co_filename, frame.f_lineno)
+            filename = str(frame.f_code.co_filename)
+            self.__coverage.line_tick(filename, frame.f_lineno)
 
     def coverage_data(self):
         return self.__coverage.lines_execution()
```
seems to help to resolve this issue (i.e. GC works as we expect).

I will enable GC dump and analyze the heap to better understand the dependencies.

romanz:
I have added the following patch to `gc.c` to mark generators with a `'G'` in the allocation map:
```diff
diff --git a/py/gc.c b/py/gc.c
index 604334c95..0b07bea5e 100644
--- a/py/gc.c
+++ b/py/gc.c
@@ -936,6 +939,8 @@ void gc_dump_alloc_table(void) {
                     c = 'B';
                 } else if (*ptr == &mp_type_module) {
                     c = 'M';
+                } else if (*ptr == &mp_type_gen_instance) {
+                    c = 'G';
                 } else {
                     c = 'h';
                     #if 0
```
Screenshot from 2025-03-31 17-17-12
But it seems that they are not reachable when using `core/tools/analyze-memory-dump.py`:
Screenshot from 2025-03-31 17-15-42

matejcik:
upy's `dump_alloc_table` walks the whole heap, so it will show even dead but yet uncollected objects.
OTOH, our mem_dump starts with a list of known GC roots, but this list may not exactly match the true GC roots so we may want to look into that.

what you could also do is implement a "dump_unknown" function that simply walks the memory, identifies all GC pointers and lists them in the json output.
(and then call this in every case, not just in the unknown case -- to catch the possibility that our dump_* functions are missing something)

romanz:
The underlying issue should be resolved by https://github.com/trezor/trezor-firmware/pull/4871.

PR #4849, opened 2025-03-30
Eckhart UI - ActionBar pagination improvements

obrusvit:
This PR:
- fixes pagination for `ActionBar` in `Single` mode
- adds `PaginateOnly` mode for `ActionBar`
- makes sure that `TextScreen` has at least `PaginateOnly` if the text does not fit
- add a few minor confirmation improvements

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/action_bar.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/action_bar.rs
@@ -110,7 +110,7 @@ impl ActionBar {
 
     pub fn new_cancel_confirm() -> Self {
         Self::new_double(
-            Button::with_icon(theme::ICON_CROSS).styled(theme::button_cancel()),
+            Button::with_icon(theme::ICON_CROSS),
             Button::with_text(TR::buttons__confirm.into()),
         )
     }
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/action_bar.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/action_bar.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 
 use super::{
-    super::component::{Button, ButtonContent, ButtonMsg, ButtonStyleSheet},
+    super::component::{Button, ButtonMsg},
     theme, HoldToConfirmAnim,
 };
 
@@ -17,20 +17,22 @@ use super::{
 pub struct ActionBar {
     /// Behavior based on `Mode`
     mode: Mode,
-    /// Right or single button, can have text or icon
-    right_button: Button,
-    /// Optional left button, can be shorter than the right one
+    /// Right or single confirm button, can have text or icon
+    right_button: Option<Button>,
+    /// Left cancel button, can be shorter than the right one
     left_button: Option<Button>,
+    /// Area of the action bar
     area: Rect,
-    /// Whether the left button is short (default: true)
-    left_short: bool,
-    // Storage of original button content for paginated component
-    left_original: Option<(ButtonContent, ButtonStyleSheet)>,
-    right_original: Option<(ButtonContent, ButtonStyleSheet)>,
     /// Hold to confirm animation
     htc_anim: Option<HoldToConfirmAnim>,
     /// Timeout
     timeout: Option<Timeout>,
+    /// Pager for paginated content
+    pager: Pager,
+    /// Left button for paginated content
+    prev_button: Button,
+    /// Right button for paginated content
+    next_button: Button,
 }
 
 pub enum ActionBarMsg {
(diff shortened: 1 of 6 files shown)


core/embed/rust/src/ui/layout_eckhart/theme/mod.rs
```diff
@@ -155,9 +155,9 @@ pub const fn button_default() -> ButtonStyleSheet {
         },
         disabled: &ButtonStyle {
             font: fonts::FONT_SATOSHI_MEDIUM_26,
-            text_color: GREY_LIGHT,
+            text_color: GREY_EXTRA_DARK,
```
bieleluk, line 158: Is this change necessary? The disabled state of the default style was intentionally set equal to the active one because:
- it enables using Action bar timeout mode with default button style
- it was not used elsewhere in the code

core/embed/rust/src/ui/layout_eckhart/firmware/text_screen.rs
```diff
@@ -121,6 +121,11 @@ where
         };
         self.header.place(header_area);
         self.content.place(content_area.inset(Self::CONTENT_INSETS));
+        if !self.content.pager().is_single() && self.action_bar.is_none() {
+            // if the content doesn't fit, we make sure there is at least ActionBar for
+            // pagination
+            self.action_bar = Some(ActionBar::new_paginate_only());
+        }
```
bieleluk, line 128: This is an open ended question.
Here we are adding paginating action bar without explicit user's intervention because we want it like this in basically all scenarios. But what about the the paginating Hint? It probably won't be desirable in all scenarios but we should use the same approach for both of them, explicit for both or automatic for both. I would suggest that it should be user's duty to explicitly add the paginate-only action bar, similar to the paginating hint.

bieleluk (changes requested):
From the code perspective, the rework of the action bar looks good.
I have a consistency concern regarding the implicit action bar creation for pagination and one minor question regarding the change of default button style.

core/embed/rust/src/ui/layout_eckhart/theme/mod.rs
obrusvit: I think we should have the `disabled` state different from `active`. We can either
- use the ActionBar timeout with this new color (which might not be bad)
- use another style sheet (could be specific for this usecase.
- set the timeout button as "active" but not invoke `event`

But having it the same is misusing the concept.

core/embed/rust/src/ui/layout_eckhart/firmware/text_screen.rs
obrusvit: This is a fallback which should rarely happen but I envision certain "More info" screens might overflow in other languages. Explicitly constructing the `TextScreen` `.with_action_bar` is the best option. 

I deliberately excluded setting a `Hint` there because:
- while `ActionBar` is crucial to see the full text, `Hint` is just advisory
- setting `Hint` would make the layout placement more complicated because the area for action bar is always there and the area for hint is not

What I actually think we should also do is check whether `Header` with some button is there, so that the user is not stuck in there - that would be programmer's double error (not setting neither ActionBar, nor Header, but that should be caught by the most basic tests).

core/embed/rust/src/ui/layout_eckhart/theme/mod.rs
bieleluk: In that case, I would be for not calling the button event because it requires the change only in one place

core/embed/rust/src/ui/layout_eckhart/firmware/text_screen.rs
bieleluk: I agree with the check, but it can be a bit problematic. For example, the action bar in `paginate only` mode does not get user anywhere, it must be combined with the Header close or menu button. 
The problem with missing header or action bar is that it won't be caught by a device test if there is `press_yes()` in the input flow, which does not need any buttons to continue.

core/embed/rust/src/ui/layout_eckhart/theme/mod.rs
obrusvit: Looking at the designs though, not even `active` color is correct for the timeout action bar. :thinking: 

I think setting the `disabled` color to `GREY` would be the best option for now, I just checked that it looks much better that `GREY_EXTRA_DARK` which is barely visible. WDYT?

core/embed/rust/src/ui/layout_eckhart/firmware/text_screen.rs
obrusvit: Agreed, `press_yes()` would skip over but at least it would be seen in UI tests. I think the best course of action is either:
- make sure that actionable `Header` is also there if not set by a programmer
- remove this check completely

core/embed/rust/src/ui/layout_eckhart/firmware/action_bar.rs
```diff
@@ -255,22 +240,62 @@ impl ActionBar {
...
+                    // Small `prev_button` when not on first page
+                    let (left, rest) = bounds.split_left(Self::LEFT_SMALL_BUTTON_WIDTH);
+                    let (_, right) = rest.split_left(Self::SPACER_WIDTH);
+                    (left, right)
+                } else {
+                    self.next_button.set_content_offset(Offset::zero());
+                    (Rect::zero(), bounds)
+                };
+                self.right_button.place(right_area);
+                self.prev_button.place(left_area);
```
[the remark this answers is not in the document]
obrusvit: This is fine because only one button is rendered (`render`) and processed (`event`).

bieleluk (approved):
LGTM
Please squash and rebase before merging

obrusvit:
It seems that `click_tests/test_repeated_backup.py` sometimes accesses the mocked DeviceMenu, should be fixed when we drop the respective commit. 
Mentioning #4855

PR #4853, opened 2025-03-31
Add thread monitor and BLE debug monitor for nRF,  JLink flashing

hiviah:
When using NCS debug.conf overlay, build with thread monitor so you can debug threads separately (supported in Ozone, CLion, VSCode).
JLink flashing of STM32 added.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -1,4 +1,3 @@
-
 .PHONY: vendor
 
 # get the number of CPU cores in a "portable" manner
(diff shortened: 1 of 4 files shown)


core/Makefile
```diff
@@ -401,12 +400,20 @@ flash_bootloader_ci_jlink: $(BOOTLOADER_CI_BUILD_DIR)/bootloader.bin ## flash CI
...
 
 flash_firmware_jlink: $(FIRMWARE_BUILD_DIR)/firmware.bin ## flash firmware using JLink. file names must end in .bin for JLink
+ifeq ($(MCU),$(filter $(MCU),STM32F4))
 	cp -f $<.p1 $<.p1.bin
 	cp -f $<.p2 $<.p2.bin
 	## pad 2nd part so that FW integrity works after flash
 	## read&compare in flashing will avoid erasing unmodified sectors
 	truncate -s $(FIRMWARE_P2_MAXSIZE) $<.p2.bin
-	JLinkExe -nogui 1 -commanderscript embed/projects/firmware/firmware_flash.jlink
+	JLinkExe -nogui 1 -commanderscript embed/models/T2T1/firmware_flash.jlink
```
TychoVrahe, line 409: lets add the jlink file to each model, and use $(TREZOR_MODEL)

core/embed/models/T3W1/firmware_flash.jlink
```diff
@@ -0,0 +1,7 @@
+device STM32U5G9ZJ
+if swd
+speed 20000
+loadbin build/firmware/firmware.bin.padded.bin 0x0C044000
```
TychoVrahe, line 4: probably would be best to generate these files from model_xx.h, should be similar to what tools/trezor_core_tools/lsgen.py does. Or let's at least create an issue for reworking this.

TychoVrahe (changes requested):

PR #4854, opened 2025-03-31
LED for emulator

obrusvit:
This PR adds LED to the emulator. It is done via `unix/display_driver.c` which draws it right above the display to the padding area.

In addition, it adds the basic control of LED for Eckhart Homescreen.

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
@@ -3,6 +3,7 @@ use crate::{
     io::BinaryData,
     strutil::TString,
     translations::TR,
+    trezorhal::rgb_led,
     ui::{
         component::{text::TextStyle, Component, Event, EventCtx, Label, Never},
         display::{image::ImageInfo, Color},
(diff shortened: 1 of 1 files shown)


core/embed/io/display/inc/io/display.h
```diff
@@ -159,4 +159,10 @@ void display_copy_mono1p(const gfx_bitblt_t *bb);
 // The function is available only on the emulator.
 const char *display_save(const char *prefix);
 void display_clear_save(void);
+
+#ifdef USE_RGB_LED
+// Display a color on the LED.
+// The function is available only on the emulator.
+void display_rgb_led(uint32_t color);
```
TychoVrahe, line 166: I dont think this belongs here. Clean solution would be to introduce some shared SDL encapsulaion which would be used be he drivers that needs it (display, led, and potentially haptics), faster workaround would be to declare this function in unix's rgb_led.c, but this pulbic api is not the place

TychoVrahe (changes requested):
How about splitting this into two PRs, the emu-led part to main, and the homescreen to eckhart branch?

core/embed/io/display/inc/io/display.h
cepetr: Yes, I needed something similar for event polling on emulator, so I created ‘/unix/sdl_event.h’ in /embed/sys/task/inc. I don’t think it’s the best solution - maybe it’s not final - but at least it separates common and emulator specific definitions. We might do something similar here by creating ‘/unix/sdl_display.h’ in ‘/embed/io/display/inc’.

obrusvit: I created a new PR addressing your concerns. https://github.com/trezor/trezor-firmware/pull/4877

PR #4857, opened 2025-04-01
JPEG Overlay

cepetr:
This PR introduces a new shape for our drawing library - `JpegOverlay` - which allows drawing grayscale JPEGs as transparent images.

`JpegOverlay` enables fast rendering of complex backgrounds (like non-linear gradients) and can also be used for content masking. It's available only on T3W1, as it utilizes the hardware JPEG decoder.

Along with this, the hardware JPEG decoder is now emulated using `libjpeg-turbo`, which provides more accurate emulation and support for 24-bit JPEG colors. The JPEG decoder written in Rust is still used on T3T1 and T2T1.


core/embed/rust/src/trezorhal/jpegdec.rs
```diff
@@ -129,6 +129,33 @@ impl<'a> JpegDecoder<'a> {
         Ok(())
     }
 
+    /// Decodes the JPEG image and calls the output function for each slice.
+    /// The function decodes only the Y component of the image, so
+    /// the output is always grascale (BitmapFormat::MONO8).
+    /// Requires a temporary buffer of size `RGBA8888_BUFFER_SIZE`.
```
obrusvit, line 135: Is this true? Isn't `RGBA8888_BUFFER_SIZE / 4` enough?

core/embed/rust/src/trezorhal/jpegdec.rs
```diff
@@ -201,4 +228,52 @@ impl<'a> JpegDecoder<'a> {
...
+    fn write_output_mono8(
+        &self,
+        buff: &mut [u8],
+        output: &mut dyn FnMut(Rect, BitmapView) -> bool,
+    ) -> bool {
+        // SAFETY:
+        // - after aligning the buffer to u32, the we check the
+        //  length of the buffer to be at least `RGBA8888_BUFFER_SIZE`
+        let buff_u32 = unsafe { buff.align_to_mut::<u32>().1 };
+        assert!(buff_u32.len() * 4 >= RGBA8888_BUFFER_SIZE);
```
obrusvit, line 241: Similar commentary here. I think the assert does not align with the commentary. Potentially, we could introduce `MONO8_BUFFER_SIZE` but it's probably unnecessary complication.

obrusvit (commented):
A few comments regarding Rust part.

I didn't review the C part.

Tried it on both the device and emulator. https://github.com/trezor/trezor-firmware/tree/obrusvit/eckhart-homescreen-jpeg-overlay

core/embed/rust/src/trezorhal/jpegdec.rs
cepetr: Yes,  it's confusing to use the same constant for rgba8888 and mono8 functions. I add JPEGDEC_MONO8_BUFFER_SIZE and updated comments. See 64463da889fce83102b05015984199526312046b

core/embed/rust/src/trezorhal/jpegdec.rs
cepetr: see the comment below

core/embed/rust/src/trezorhal/jpegdec.rs
```diff
@@ -129,6 +129,33 @@ impl<'a> JpegDecoder<'a> {
         Ok(())
     }
 
+    /// Decodes the JPEG image and calls the output function for each slice.
+    /// The function decodes only the Y component of the image, so
+    /// the output is always grascale (BitmapFormat::MONO8).
```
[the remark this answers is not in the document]
cepetr: fixed in 64463da889fce83102b05015984199526312046b

core/embed/rust/src/ui/shape/jpeg_overlay.rs
```diff
@@ -0,0 +1,107 @@
...
+
+use super::{Canvas, DrawingCache, Renderer, Shape, ShapeClone};
+
+use crate::{trezorhal::jpegdec::JpegDecoder, ui::display::Color};
+
+use without_alloc::alloc::LocalAllocLeakExt;
+
+/// A shape for rendering grayscale JPEG images as a transparent overlay
+/// over the existing content.
+pub struct JpegOverlay<'a> {
```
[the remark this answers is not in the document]
cepetr: I might not fully understand the idea...., but I think the JpegOverlay is quite cheap to create, so there’s no need to save it for reuse.

core/embed/gfx/jpegdec/unix/jpegdec.c
```diff
@@ -0,0 +1,346 @@
...
+    // Image headers where decoded, now we can decode the image data
+    // Read scanlines until we have a full slice
+
+    for (;;) {
+      // Row in slice buffer
+      size_t row = dec->cinfo.output_scanline - dec->slice_y;
+
+      if (row >= MAX_SLICE_HEIGHT) {
+        dec->slice_y = dec->cinfo.output_scanline;
+        row = 0;
```
TychoVrahe, line 247: seems like this assignment is useless

core/embed/gfx/jpegdec/unix/jpegdec.c
```diff
@@ -0,0 +1,346 @@
...
+  return FALSE;
+}
+
+static void skip_input_data(j_decompress_ptr cinfo, long num_bytes) {
+  custom_source_mgr_t *src = (custom_source_mgr_t *)cinfo->src;
+
+  if (num_bytes > 0) {
+    while (num_bytes > (long)src->pub.bytes_in_buffer) {
+      num_bytes -= (long)src->pub.bytes_in_buffer;
+      fill_input_buffer(cinfo);
```
TychoVrahe, line 111: should we check the return value? is there anything to do if fill_input_buffer fails?

core/embed/gfx/jpegdec/unix/jpegdec.c
TychoVrahe, line 247: should the decoder state be updated here?

cepetr: `row = 0` is ok, but `break` is unecessary, fixed see e3cf6a23b66b2b7b9a3fe4319c67cba11fc1fd3e

cepetr: fixed here e3cf6a23b66b2b7b9a3fe4319c67cba11fc1fd3e, see comment above

core/embed/gfx/jpegdec/unix/jpegdec.c
cepetr: It looks a bit suspicious, but probably not. This is recommended approach from the libjpeg doc.

TychoVrahe (approved):

PR #4858, opened 2025-04-01
Fix behavior of set brightness screen

bieleluk:
- Do not confirm brightness value by releasing the finger, but by clicking on the `x` button
- Change the slider color when holding it


obrusvit (approved):
utACK

bosomt:
**QA OK**

works as described in PR
tested on 7fc30de46e070fe83cb2e047f104552bc97fecc3 EMU build

PR #4864, opened 2025-04-02
test(core): close `_raw_client` in case `client()` fixture fails

romanz:
Should help with [the following kind of hardware tests' errors](https://github.com/trezor/trezor-firmware/actions/runs/14207984894/job/39809710980) (the retry mechanism seem to get stuck).

It's easier to review with `--ignore-space-change`:
```diff
diff --git a/tests/conftest.py b/tests/conftest.py
index be4363d56..83e6512c9 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -282,6 +282,7 @@ def client(
 
     _raw_client.reset_debug_features()
     _raw_client.open()
+    try:
         try:
             _raw_client.sync_responses()
             _raw_client.init_device()
@@ -348,7 +349,7 @@ def client(
 
         with ui_tests.screen_recording(_raw_client, request):
             yield _raw_client
-
+    finally:
         _raw_client.close()
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -283,73 +283,74 @@ def client(
     _raw_client.reset_debug_features()
     _raw_client.open()
     try:
-        _raw_client.sync_responses()
-        _raw_client.init_device()
-    except Exception:
-        request.session.shouldstop = "Failed to communicate with Trezor"
-        pytest.fail("Failed to communicate with Trezor")
+        try:
+            _raw_client.sync_responses()
+            _raw_client.init_device()
+        except Exception:
+            request.session.shouldstop = "Failed to communicate with Trezor"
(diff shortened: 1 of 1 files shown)


romanz:
Rebasing over #4871.

tests/conftest.py
```diff
@@ -283,73 +283,74 @@ def client(
     _raw_client.reset_debug_features()
     _raw_client.open()
     try:
```
romanz, line 285: Maybe we should use `contextlib.closing(_raw_client)` instead?

matejcik: that would make more sense if we _already_ had a context manager, which would wrap this
but if we did we'd just close in that context manager
so i think try/finally is more readable here

matejcik (approved):

PR #4870, opened 2025-04-03
fix(core): workaround for GC-related test issue

romanz:
The underlying issue seems to be caused by the GC not freeing the old `session` since it is reachable from [the C stack](https://github.com/trezor/micropython/blob/6bdefd147256d6a72827ac6225fc5e36bd92a801/shared/runtime/gchelper_generic.c#L180), so the `ImageBuffer` [lock](https://github.com/trezor/trezor-firmware/blob/469f093b8add96d2d2451fe4f6a4ae1dba30309d/core/embed/rust/src/ui/shape/utils/imagebuf.rs#L41) is not dropped.

### DIFF
--- a/core/prof/prof.py
+++ b/core/prof/prof.py
@@ -43,7 +43,8 @@ class _Prof:
         #     print(event, frame.f_code.co_filename, frame.f_lineno)
 
         if event == "line":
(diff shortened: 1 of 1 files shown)


matejcik:
let's not do this, it's magic with no clear mechanism of effect that can break and reappear anywhere.
plus it appears that #4871 fixes the underlying problem

PR #4872, opened 2025-04-03
fix(core): correct `__main__` dict size retrieval in firmware

romanz:
Otherwise, it fails to boot (following e6f96974d).


romanz:
Tested on T3T1 and T2B1.

matejcik (approved):

PR #4875, opened 2025-04-04
Touch driver fix

cepetr:
This PR fixes several issues with the touch driver, especially on the T3W1:

. The touch controller may report out-of-range Y coordinates when touching the lower part of the display. This has been fixed by clamping the coordinates to the valid range.
. After the refactoring in #4829, we could lose `TOUCH_START` events under certain conditions. This has been addressed by treating the first `TOUCH_MOVE` as a `TOUCH_START` if no prior touch press was detected - regardless of the “starving” flag (which indicates infrequent polling of the touch controller). -- this issue affects t3t1 and t2t1 too
. The ft6x36 driver’s `touch_get_state()` can no longer return 0. While this has no effect on functionality, it makes the behavior of the function more consistent and logical.
. The ft6x36 controller almost always reports a  START followed immediately by a MOVE with the same coordinates. This previously caused a 0 to be reported after a `TOUCH_START_EVENT`, which was a bit confusing. This behavior has now been corrected.


core/embed/io/touch/ft6x36/panels/lx250a2410a.h
```diff
@@ -0,0 +1,56 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <trezor_types.h>
+
+// Performs touch coordinates correction needed for a specific panel
+// Input parameteres x, y represent original touch coordinates.
```
TychoVrahe, line 25: doubled comment and license

core/embed/io/touch/ft6x36/panels/lx250a2410a.h
```diff
@@ -0,0 +1,56 @@
...
```
TychoVrahe, line 52: typo parameteres

core/embed/io/touch/ft6x36/panels/lx250a2410a.h
cepetr: fixed, see 70e8e363fe8406dd9eff6dfb3835e97c3eac5ce1

core/embed/io/touch/ft6x36/panels/lx250a2410a.h
cepetr: fixed, see 70e8e363fe8406dd9eff6dfb3835e97c3eac5ce1

TychoVrahe (approved):
utACK

PR #4876, opened 2025-04-04
fix firmware upload in bootloader_ci

TychoVrahe:
Long-term neglected bootloader CI - only minimal change needed to make the firmware update run on TS5.


cepetr (approved):
utACK

matejcik (approved):

PR #4877, opened 2025-04-04
LED for emulator

obrusvit:
[no changelog]

This PR adds LED to the emulator. It is done via unix/display_driver.c which draws it right above the display to the padding area.

This is a 2nd PR which follows up from review concerns and changes the target branch to `main` https://github.com/trezor/trezor-firmware/pull/4854#pullrequestreview-2733908768

### DIFF
--- a/core/embed/io/display/unix/display_driver.c
+++ b/core/embed/io/display/unix/display_driver.c
@@ -24,6 +24,7 @@
 #include <trezor_rtl.h>
 
 #include <io/display.h>
+#include <io/unix/sdl_display.h>
 
 #include <SDL.h>
 #include <SDL_image.h>
@@ -75,6 +76,10 @@ typedef struct {
   uint8_t mono_framebuf[DISPLAY_RESX * DISPLAY_RESY];
 #endif
 
+#ifdef USE_RGB_LED
+  // Color of the RGB LED
+  uint32_t led_color;
+#endif
 } display_driver_t;
 
 static display_driver_t g_display_driver = {
@@ -177,6 +182,9 @@ bool display_init(display_content_mode_t mode) {
   SDL_ShowCursor(SDL_DISABLE);
 #else
   drv->orientation_angle = 0;
+#endif
+#ifdef USE_RGB_LED
+  drv->led_color = 0;
 #endif
 
   gfx_bitblt_init();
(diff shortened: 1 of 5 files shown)


core/embed/io/display/unix/display_driver.c
```diff
@@ -353,6 +398,12 @@ void display_refresh(void) {
     SDL_RenderCopyEx(drv->renderer, drv->texture, NULL, &r,
                      drv->orientation_angle, NULL, 0);
   }
+  // Show the LED
+#ifdef USE_RGB_LED
```
TychoVrahe, line 402: done like this led color is changed only when we call display refresh, which is different from how hardware behaves. can we invert the logic, so that draw_rgb_led is called from rgb_led.c?

obrusvit: I created `embed/io/display/inc/io/unix/dsl_display.h` with `display_rgb_led` function which is implemented in `unix/display_driver.c`.

The function updates the internally stored `led_color` and calls `display_refresh` there, otherwise, the LED would be overriden by clear. I'm not sure if this is desirable as we now refresh the display with the change of EMU. If we want a clear separation, we would probably have to create some SDL encapsulation as you suggested in the previous PR.

As I'm writing this, I'm thinking whether using a separate `SDL_Texture` wouldn't be better :thinking: 


5eae7404a8139bcfab733ca6dea57459f6839197

obrusvit: I didn't push the idea of using separate `SDL_Texture` and I think the current state is OK. 
@TychoVrahe WDYT?

TychoVrahe: i think its acceptable for now but i believe cleaner solution should be pursued when possible

TychoVrahe (approved):

core/embed/io/display/unix/display_driver.c
```diff
@@ -319,6 +324,63 @@ static void copy_mono_framebuf(display_driver_t *drv) {
...
+  if (drv->background) {
+    center_x += TOUCH_OFFSET_X;
+    center_y = TOUCH_OFFSET_Y / 2;
+  } else {
+    center_x += EMULATOR_BORDER;
+    center_y = EMULATOR_BORDER / 2;
+  }
+
+  // Draw the LED
+  SDL_SetRenderDrawColor(drv->renderer, r, g, b, 255);
```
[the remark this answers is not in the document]
obrusvit: The function checks whether `drv->initialized` is set at the beginning of the function.

core/embed/io/display/unix/display_driver.c
obrusvit: I created an issue: https://github.com/trezor/trezor-firmware/issues/5009

PR #4882, opened 2025-04-06
feat(core): send address & public key response before showing `StatusScreen`

romanz:
Can be tested on device with the following patch:
```diff
diff --git a/python/src/trezorlib/client.py b/python/src/trezorlib/client.py
index 529992dfb..d9dac9128 100644
--- a/python/src/trezorlib/client.py
+++ b/python/src/trezorlib/client.py
@@ -464,6 +464,7 @@ class TrezorClient(Generic[UI]):
         This is a no-op in bootloader mode, as it does not support session management.
         """
         # since: 2.3.4, 1.9.4
+        import time; time.sleep(5)
         try:
             if not self.features.bootloader_mode:
                 self.call(messages.EndSession())
```

[Screencast from 2025-04-07 11-50-38.webm](https://github.com/user-attachments/assets/63f46ee1-1b52-449b-a1d6-bacf7d62b705)


romanz:
IIUC, the UI diff is a test issue - e.g. https://data.trezor.io/dev/firmware/ui_report/14304238985/T3T1-en-core_device_test/diff/T3T1_en-device_tests-bitcoin-test_getaddress_segwit.py%3A%3Atest_show_segwit.html seems to be caused by a missing confirmation screen on `main`, which is present in this PR:

# `main`
[Screencast from 2025-04-07 20-40-06.webm](https://github.com/user-attachments/assets/f7c213ce-3e35-4ed9-94c7-7a5b916d7c83)

# this PR
[Screencast from 2025-04-07 20-44-24.webm](https://github.com/user-attachments/assets/6bb163e2-0f63-43a1-8bb2-550f0ae6b86e)

Hannsek:
Thanks. I saw a different diff here on sunday. All good now. 👍🏻

romanz:
Rebasing to resolve a merge conflict.

core/src/apps/bitcoin/get_public_key.py
```diff
@@ -116,13 +127,11 @@ async def get_public_key(
...
-
-    return PublicKey(
-        node=node_type,
-        xpub=node_xpub,
-        root_fingerprint=keychain.root_fingerprint(),
-        descriptor=descriptor,
-    )
+        await context.write(response)
+        await show_continue_in_app(TR.address__public_key_confirmed)
+        return None
```
matejcik, line 132: i'd like to wrap this in a function, something like: `return await early_response(response, show_continue_in_app(...))`

`early_response` would live in `trezor.ui.layout` (or `.common`)

type-wise, it would be cool if we could make something vaaaaguely along the lines of:
```
if TYPE_CHECKING:
    class EarlyResponse(Generic[T]):
        pass
    T = TypeVar("T", bound=MessageType)

    MaybeEarlyResponse = Union[T, EarlyResponse[T]]  # for convenience, not sure if this is legal

else:
    EarlyResponse = None

def early_response(response: T, layout: ...) -> EarlyResponse[T]:
    ...
```

then annotate the functions not with `|None` but with `MaybeEarlyResponse[ActualType]`

(and if the workflow handler requires it, we can allow handlers that return (Maybe)EarlyResponse)

core/src/trezor/wire/context.py
```diff
@@ -79,6 +79,16 @@ async def call_any(
     return await CURRENT_CONTEXT.read(expected_wire_types)
 
 
+async def write(msg: protobuf.MessageType) -> None:
```
matejcik, line 82: this should probably be left out of this PR and done a bit differently in #4919

romanz: Removed in https://github.com/trezor/trezor-firmware/commit/0b1ea106fca6cf68597987863a39f3eb8eb7ef00.

core/src/apps/bitcoin/get_public_key.py
romanz: Thanks!

I have added `early_response` and modified the return types in https://github.com/trezor/trezor-firmware/commit/0b1ea106fca6cf68597987863a39f3eb8eb7ef00.

IIUC, it would allow the type system to enforce that the only way to generate an "early response" is by using `early_response()` due to its return type - right?

Also, is it OK to that the handlers will return `ActualType | EarlyResponse` instead of `MaybeEarlyResponse[ActualType]`?

romanz: I've added `early_response` to `trezor.wire` since it is used to actually send the response back to the client...

romanz: Got it, thanks!
I'll try to make `EarlyResponse` generic :)

romanz: https://github.com/trezor/trezor-firmware/pull/4882/commits/a58f1ae2bed675b5b6bd899c5983fa0221ab9108

romanz:
Device tests passed on T3T1:
```
May 20 22:23:07 = 1386 passed, 176 skipped, 10 deselected, 8 xfailed, 2 warnings in 8753.22s (2:25:53) =
```

romanz:
Also tested using:
```python
from trezorlib.transport import get_transport
from trezorlib.client import TrezorClient, PASSPHRASE_TEST_PATH
from trezorlib.ui import ClickUI
from trezorlib import btc

c = TrezorClient(transport=get_transport(), ui=ClickUI())
print(btc.get_address(c, "Testnet", PASSPHRASE_TEST_PATH, show_display=True))
```

matejcik (approved):

romanz:
Rebasing over `main`.

PR #4890, opened 2025-04-08
Optimize device menu

ibz:
A couple of low-hanging fruit optimizations to the device menu.


bieleluk (approved):
The change itself looks good but I found two minor things from the initial implementation that cause inconsistent use of the `cross` icon.

PR #4838, opened 2025-03-27
BLE support in emulator

mmilata:
Consists of:
* expose core BLE interface as two UDP sockets, no connection state handling
* trezorlib transport that talks to these sockets - as there is no connection state handling it might be useful for writing tests for e.g. misbehaving nRF
* python script that launches dbus-daemon and a mock Linux BLE API on it - using env variable you can convince Suite/trezorlib to talk to it instead of the actual BLE devices

TODO

- [ ] connection state handling emulation
- [ ] launch DBus api from `emu.py`
- [ ] use `dbus-fast` instead of unmaintained `dbus-next`

PR #4891, opened 2025-04-08
BLE: enable advertising on firmware boot

mmilata:
Several small BLE-related changes:

* 0d55e1a: enable _directed advertising_ on firmware startup if there are > 0 bonded devices, i.e. bonded devices can connect but device does not show up in scans
* 62ea246: stub implementation of ble IO for emulator - just enough to build: the actual implementation is in #4838 and needs a bunch of work, however I'd like to start building emulator with the `ble` SCons features so that we can refer to it from rust code
* c053ef3: enables `button` support in T3W1 emulator which should be now possible thanks to #4829, did not test anything if it works just that it doesn't break `touch`
* ada8814: typo in generated type signatures, remove obsolete nrf upgrade functions
* 7ac07be: expose `ble_get_state()` to micropython as flags

### DIFF
--- a/core/site_scons/models/T3W1/emulator.py
+++ b/core/site_scons/models/T3W1/emulator.py
@@ -87,6 +87,12 @@ def configure(
         features_available.append("touch")
         defines += [("USE_TOUCH", "1")]
 
+        sources += ["embed/io/button/unix/button.c"]
+        sources += ["embed/io/button/button_fsm.c"]
+        paths += ["embed/io/button/inc"]
+        features_available.append("button")
+        defines += [("USE_BUTTON", "1")]
+
     if "ble" in features_wanted:
         sources += ["embed/io/ble/unix/ble.c"]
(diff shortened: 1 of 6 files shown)


core/site_scons/models/T3W1/emulator.py
```diff
@@ -87,6 +87,18 @@ def configure(
         features_available.append("touch")
         defines += [("USE_TOUCH", "1")]
 
+        sources += ["embed/io/button/unix/button.c"]
```
obrusvit, line 90: Nit: why don't we just `sources += ["one", "bigger", "list"]`?

obrusvit (approved):
utACK

core/site_scons/models/T3W1/emulator.py
mmilata: seemed to be predominant style around here, IDK

TychoVrahe (approved):

PR #4892, opened 2025-04-08
feat(core): check for runtime QSTR allocations

romanz:
`/usr/lib/micropython` is removed from `core/embed/projects/unix/main.c`, otherwise it is allocated as QSTR:
```
QSTR: add hash=7735 len=20 data=/usr/lib/micropython
...
AssertionError: Runtime QSTR allocation: 1 pools, 1 strings, 21 data bytes, 309 total bytes
```

The check is enabled only for debug builds.


romanz:
Unfortunately, most of our tests are running under profiler - which is failing due to the new check :(
image

romanz:
Maybe we can check that no new QSTRs were allocated within each session (instead of assert that new new QSTRs were allocated at all)?

matejcik:
the allocation is happening in `unimport_end`. why? even the profiler should not cause this.
maybe the profiler causes some new stuff in sys.modules (only thing i can come up with) that we can simply add to the already built in list of qstrs?

romanz:
I have added the fragmentation check to `unimport_end()`:
https://github.com/trezor/trezor-firmware/blob/f920ebb96a92dffd0007bf25b2820c04a528ddcb/core/src/trezor/utils.py#L69-L71
If preferable, I can move it to the end of `unimport.__exit__`:
https://github.com/trezor/trezor-firmware/blob/f920ebb96a92dffd0007bf25b2820c04a528ddcb/core/src/trezor/utils.py#L101-L106

romanz:
From https://github.com/trezor/trezor-firmware/pull/4892/commits/109f3c070a9a9b37010a8ef651195ac3b4e54170, the allocated QSTRs are:
```
     1	Q(worker_id)
     2	Q(file_name)
     3	Q(.coverage)
     4	Q(")
     5	Q(AllocCounter)
     6	Q(last_alloc_count)
     7	Q(last_line)
     8	Q(count_last_line)
     9	Q(allocs)
    10	Q(entry)
    11	Q(calls)
    12	Q(allocs_now)
    13	Q(alloc_count)
    14	Q(allocs_per_last_line)
    15	Q({}:{})
    16	Q(dump_data)
    17	Q(trace_handler)
    18	Q(/home/rzeyde/src/trezor-firmware/core/prof)
    19	Q(/home/rzeyde/src/trezor-firmware/core/prof/prof.py)
    20	Q(PATH_PREFIX)
    21	Q(TREZOR_SRC)
    22	Q(Coverage)
    23	Q(__files)
    24	Q(line_tick)
    25	Q(lineno)
    26	Q(lines_execution)
    27	Q(this_file)
    28	Q(_Prof)
    29	Q(trace_count)
    30	Q(display_flags)
    31	Q(__coverage)
    32	Q(trace_tick)
    33	Q(event)
    34	Q(arg)
    35	Q(coverage_data)
    36	Q(write_data)
    37	Q(__prof__)
```

romanz:
Added a few fixes and rebased over #4910.

core/embed/projects/unix/main.c
```diff
@@ -568,7 +568,7 @@ MP_NOINLINE int main_(int argc, char **argv) {
 #ifdef MICROPY_PY_SYS_PATH_DEFAULT
     path = MICROPY_PY_SYS_PATH_DEFAULT;
 #else
-    path = ".frozen:~/.micropython/lib:/usr/lib/micropython";
+    path = ".frozen:~/.micropython/lib";
```
mmilata, line 571: `~/.micropython/lib` also seems questionable, if nobody is using it I'd be in favor of dropping it too

mmilata (approved):
utACK

core/embed/projects/unix/main.c
romanz: 2975a07f1

PR #4893, opened 2025-04-08
refactor(core): device menu modules

mmilata:
Cleaner separation for devices that don't have device menu. Placeholders for T3W1 remain.

### DIFF
--- a/core/src/apps/homescreen/__init__.py
+++ b/core/src/apps/homescreen/__init__.py
@@ -4,9 +4,8 @@ import storage
 import storage.cache
 import storage.device
 import trezorui_api
-from trezor import config, wire
+from trezor import config, utils, wire
 from trezor.enums import MessageType
-from trezor.ui.layouts import raise_if_not_confirmed
 from trezor.ui.layouts.homescreen import Busyscreen, Homescreen, Lockscreen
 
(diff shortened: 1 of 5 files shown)


core/src/apps/homescreen/__init__.py
```diff
@@ -59,42 +58,12 @@ async def homescreen() -> None:
...
-                "pairing_code",
-            )
-    else:
-        lock_device()
+    if utils.INTERNAL_MODEL == "T3W1":
+        if res is trezorui_api.INFO:
+            from .device_menu import handle_device_menu
+
+            return await handle_device_menu()
+    lock_device()
```
obrusvit, line 66: This locks the device whenever we come back from the Device Menu.

obrusvit (changes requested):

core/src/apps/homescreen/__init__.py
mmilata: I'm not able to reproduce it - the line above should exit the function without calling `lock_device`.

obrusvit: yeah, sorry, my brain just removed the `return` for me.

obrusvit (approved):
LGTM :+1:

PR #4905, opened 2025-04-10
fix(core):  make mod_trezorio_poll return False only on timeout

cepetr:
This PR fixes a bug introduced in #4829.

`modtrezor_io_poll` returned `False` not only in case of a timeout, but also when events were signaled and then discarded due to various reasons - such as false positives or event rate throttling. That broke the task scheduling mechanism in MicroPython, which relies on it.

No changelog needed since the bug was not released yet.

### DIFF
--- a/core/embed/sys/task/inc/sys/sysevent.h
+++ b/core/embed/sys/task/inc/sys/sysevent.h
@@ -46,11 +46,11 @@ typedef struct {
 } sysevents_t;  // sys_events_t
 
 // Polls for the specified events. The function blocks until at least
-// one event is signaled or the timeout (in milliseconds) expires.
+// one event is signaled or deadline expires.
 //
 // Multiple events may be signaled simultaneously.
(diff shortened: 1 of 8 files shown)


core/embed/sys/task/inc/sys/sysevent.h
```diff
@@ -46,11 +46,11 @@ typedef struct {
 } sysevents_t;  // sys_events_t
 
 // Polls for the specified events. The function blocks until at least
-// one event is signaled or the timeout (in milliseconds) expires.
+// one event is signaled or deadline expires.
 //
 // Multiple events may be signaled simultaneously.
 //
 // Returns the events that were signaled. If the timeout expires, both
```
TychoVrahe, line 53: perhaps 'deadline passes' instead of 'timeout expires'

TychoVrahe (approved):

PR #4906, opened 2025-04-11
refactor(core): convert modtrezorio-ble to rust

mmilata:
Currently we need some of the BLE I/O functions available both in python and rust. This PR builds the micropython API on top of the rust API to avoid duplicating some of the connection handling logic.


core/embed/rust/src/trezorhal/ble/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+fn data_advname(name: &str) -> ffi::ble_command_data_t {
+    let mut data = ffi::ble_command_data_t {
+        adv_start: ffi::ble_adv_start_cmd_data_t {
+            name: [0u8; ADV_NAME_LEN],
+            static_mac: false,
+        },
+    };
+    let bytes = prefix_utf8_bytes(name, ADV_NAME_LEN);
```
romanz, line 68: Maybe it would be simpler to return an error when `name` is too long?

core/embed/rust/src/trezorhal/ble/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    state().pairing_requested
+}
+
+pub fn write(bytes: &[u8]) -> Result<(), Error> {
+    let len = bytes.len() as u16;
+    let success = unsafe { ffi::ble_write(bytes.as_ptr(), len) };
+    if success {
+        Ok(())
+    } else {
+        Err(OPERATION_FAILED)
```
romanz, line 155: nit: consider using a different error message here and in `issue_command`.

core/embed/rust/src/trezorhal/ble/micropython.rs
```diff
@@ -0,0 +1,290 @@
...
+    ///     """
+    Qstr::MP_QSTR_peer_count => obj_fn_0!(py_peer_count).as_obj(),
+
+    /// def is_connected() -> bool:
+    ///     """
+    ///     True if a host is connected to us. May or may not be paired.
+    ///     """
+    Qstr::MP_QSTR_is_connected => obj_fn_0!(py_is_connected).as_obj(),
+
+    /// def connection_strflags() -> int:
```
romanz, line 270:
```suggestion
    /// def connection_strflags() -> str:
```

core/embed/rust/src/trezorhal/ble/micropython.rs
```diff
@@ -0,0 +1,290 @@
...
+            let _ = buf.push(',');
+        }
+        buf.pop();
+        buf.as_str().try_into()
+    };
+    unsafe { util::try_or_raise(block) }
+}
+
+extern "C" fn py_iface_num(_self: Obj) -> Obj {
+    Obj::small_int(8) // FIXME SYSHANDLE_BLE_IFACE_0
```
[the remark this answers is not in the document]
TychoVrahe: for reference, syshandles will be exposed in https://github.com/trezor/trezor-firmware/pull/4833/

see: https://github.com/trezor/trezor-firmware/pull/4833/files#diff-1a2aa3602031b54883de82fe1856d39ba9c888600685551391c91a34f68ba29a

core/embed/rust/src/trezorhal/ble/micropython.rs
```diff
@@ -0,0 +1,290 @@
...
+    is_connected().into()
+}
+
+extern "C" fn py_connection_strflags() -> Obj {
+    let block = || {
+        let s = state();
+        if !s.state_known {
+            return "unknown".try_into();
+        }
+        let mut buf: ShortString = ShortString::new();
```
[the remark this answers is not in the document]
mmilata: Good idea, returning a list in fe61411760

core/embed/rust/src/trezorhal/ble/mod.rs
mmilata: Fixed in fe61411760

core/embed/rust/src/trezorhal/ble/micropython.rs
mmilata: fe61411760

core/embed/rust/librust_qstr.h
```diff
@@ -19,6 +19,7 @@ static void _librust_qstrs(void) {
   MP_QSTR_ATTACHED;
   MP_QSTR_AttachType;
   MP_QSTR_BacklightLevels;
+  MP_QSTR_BleInterface;
```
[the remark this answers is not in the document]
mmilata: That means renaming the strings so that they have a common prefix we can filter for in `librust_qstr.mako`, right?

core/embed/rust/src/trezorhal/ble/micropython.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+extern "C" fn py_iface_write(_self: Obj, msg: Obj) -> Obj {
+    let block = || {
+        // SAFETY: reference is discarded at the end of the block
+        let buf = unsafe { get_buffer(msg)? };
+        if write(buf).is_ok() {
+            Ok(buf.len().try_into()?)
+        } else {
+            Ok((-1).try_into()?)
```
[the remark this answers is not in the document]
mmilata: That would certainly be more pythonic but I'm a bit hesitant to rewrite all the other interface implementations :sweat_smile:

core/embed/rust/src/trezorhal/ble/mod.rs
mmilata: Currently we are using the device label (max 32 chars) as the advertising name, so we need to be able to handle when it's longer than `ADV_NAME_LEN` (20). We can implement this in python instead but not sure if we can do anything better than using a prefix.

core/embed/rust/librust_qstr.h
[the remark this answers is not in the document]
romanz: You're right.

Thinking about it, this should be probably done in a separate PR (e.g. `sd_card` related QSTRs can be also excluded for TS3 models).

PR #4908, opened 2025-04-11
Update VerticalMenu swiping behavior for Eckhart

bieleluk:
This PR:
- Improves event processing for the `VerticalMenu` component
- Switches swiping on/off based on the `VerticalMenu` fitting/not fitting its bounds
- Implements function `update_screen` for `VerticalMenuScreen`. It has to be called in the `DeviceMenuScreen` whenever the screen is switched.
- Adds swiping functionality for testing purposes when animations are disabled:
  - The menu is swipable until the the last item is visible on top of the menu area (even if an entire menu fits the screen).
  - This way the device tests can select `n-th` item by swiping up `n-1` times and then clicking on the 1st item area no matter what the menu item height is. The behavior can be tested by spawning the emulator with disabled animations:`./emu.py -a`

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
@@ -416,10 +416,13 @@ impl<'a> DeviceMenuScreen<'a> {
             Subscreen::Submenu(ref mut submenu_index) => {
                 match self.submenus[*submenu_index].items[idx].action {
                     Some(Action::GoTo(menu)) => {
-                        self.menu_screen.as_mut().unwrap().update_menu(ctx);
                         unwrap!(self.parent_subscreens.push(self.active_subscreen));
                         self.set_active_subscreen(menu);
                         self.place(self.bounds);
+                        if let Some(screen) = self.menu_screen.as_mut() {
+                            screen.initialize_screen(ctx);
+                        }
+                        return None;
                     }
                     Some(Action::Return(msg)) => return Some(msg),
                     None => {}
@@ -433,10 +436,13 @@ impl<'a> DeviceMenuScreen<'a> {
         None
     }
 
-    fn go_back(&mut self) -> Option<DeviceMenuMsg> {
+    fn go_back(&mut self, ctx: &mut EventCtx) -> Option<DeviceMenuMsg> {
         if let Some(parent) = self.parent_subscreens.pop() {
             self.set_active_subscreen(parent);
             self.place(self.bounds);
+            if let Some(screen) = self.menu_screen.as_mut() {
+                screen.initialize_screen(ctx);
+            }
             None
         } else {
             Some(DeviceMenuMsg::Close)
@@ -478,7 +484,7 @@ impl<'a> Component for DeviceMenuScreen<'a> {
                         }
                     }
                     Some(VerticalMenuScreenMsg::Back) => {
-                        return self.go_back();
+                        return self.go_back(ctx);
                     }
                     Some(VerticalMenuScreenMsg::Close) => {
                         return Some(DeviceMenuMsg::Close);
@@ -488,7 +494,7 @@ impl<'a> Component for DeviceMenuScreen<'a> {
             }
             Subscreen::AboutScreen => {
                 if let Some(TextScreenMsg::Cancelled) = self.about_screen.event(ctx, event) {
-                    return self.go_back();
+                    return self.go_back(ctx);
                 }
             }
         }
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
@@ -1,5 +1,6 @@
 use crate::ui::{
     component::{Component, Event, EventCtx},
+    event::TouchEvent,
     geometry::{Insets, Offset, Rect},
     shape::{Bar, Renderer},
 };
@@ -19,19 +20,19 @@ type VerticalMenuButtons = Vec<Button, MENU_MAX_ITEMS>;
 pub struct VerticalMenu {
     /// Bounds the sliding window of the menu.
     bounds: Rect,
-    /// Full bounds of the menu, including off-screen items.
-    virtual_bounds: Rect,
     /// Menu items.
     buttons: VerticalMenuButtons,
-    /// Whether to show separators between buttons.
-    separators: bool,
+    /// Full height of the menu, including overflowing items.
+    total_height: i16,
     /// Vertical offset of the current view.
     offset_y: i16,
     /// Maximum vertical offset.
-    max_offset: i16,
+    offset_y_max: i16,
     /// Adapt padding to fit entire area. If the area is too small, the padding
     /// will be reduced to min value.
-    fit_area: bool,
+    content_fit: bool,
+    /// Whether to show separators between buttons.
+    separators: bool,
 }
 
 pub enum VerticalMenuMsg {
@@ -45,13 +46,13 @@ impl VerticalMenu {
 
     fn new(buttons: VerticalMenuButtons) -> Self {
         Self {
-            virtual_bounds: Rect::zero(),
             bounds: Rect::zero(),
             buttons,
-            separators: false,
+            total_height: 0,
             offset_y: 0,
-            max_offset: 0,
-            fit_area: false,
+            offset_y_max: 0,
+            separators: false,
+            content_fit: false,
         }
     }
 
@@ -64,8 +65,8 @@ impl VerticalMenu {
         self
     }
 
-    pub fn with_fit_area(mut self) -> Self {
-        self.fit_area = true;
+    pub fn with_content_fit(mut self) -> Self {
+        self.content_fit = true;
         self
     }
 
@@ -74,18 +75,19 @@ impl VerticalMenu {
         self
     }
 
-    pub fn area(&self) -> Rect {
-        self.bounds
+    /// Check if the menu fits its area without scrolling.
+    pub fn fits_area(&self) -> bool {
+        self.total_height <= self.bounds.height()
     }
 
     /// Scroll the menu to the desired offset.
     pub fn set_offset(&mut self, offset_y: i16) {
-        self.offset_y = offset_y.max(0).min(self.max_offset);
+        self.offset_y = offset_y.max(0).min(self.offset_y_max);
     }
 
     /// Chcek if the menu is on the bottom.
     pub fn is_max_offset(&self) -> bool {
-        self.offset_y == self.max_offset
+        self.offset_y == self.offset_y_max
     }
 
     /// Get the current sliding window offset.
@@ -93,7 +95,10 @@ impl VerticalMenu {
         self.offset_y
     }
 
-    /// Update menu buttons based on the current offset.
+    /// Update state of menu buttons based on the current offset.
+    /// Enable only buttons that are fully visible in the menu area.
+    /// Meaningful only if the menu is scrollable.
+    /// If the menu fits its area, all buttons are enabled.
     pub fn update_menu(&mut self, ctx: &mut EventCtx) {
         for button in self.buttons.iter_mut() {
             let in_bounds = button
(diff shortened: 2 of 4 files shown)


core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -433,10 +436,14 @@ impl<'a> DeviceMenuScreen<'a> {
         None
     }
 
-    fn go_back(&mut self) -> Option<DeviceMenuMsg> {
+    fn go_back(&mut self, ctx: &mut EventCtx) -> Option<DeviceMenuMsg> {
         if let Some(parent) = self.parent_subscreens.pop() {
             self.set_active_subscreen(parent);
             self.place(self.bounds);
+            self.menu_screen
```
ibz, line 443: Same.

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
```diff
@@ -54,33 +57,24 @@ impl VerticalMenuScreen {
...
-                    ),
-                    _ => None, // Ignore touch events outside the bounds
-                };
-                shifted_event.map(Event::Touch)
-            }
-            _ => None, // Ignore other events
+    /// Update swipe detection and buttons state based on menu size
+    pub fn update_screen(&mut self, ctx: &mut EventCtx) {
+        if !self.menu.fits_area() {
+            // Enable swipe
```
ibz, line 63: Superfluous comment...

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
```diff
@@ -54,33 +57,24 @@ impl VerticalMenuScreen {
...
+                .with_swipe(Direction::Up, SwipeSettings::default())
+                .with_swipe(Direction::Down, SwipeSettings::default());
+            ctx.enable_swipe();
+
+            // Update the menu buttons state
+            self.menu.update_menu(ctx);
+        } else {
+            // Disable swipe
+            self.swipe_enabled = false;
+            ctx.disable_swipe();
```
ibz, line 75: Shouldn't `swipe_config` be an `Option` and we would set it to `None` here? I know it doesn't matter, but it just seems cleaner...

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
```diff
@@ -54,33 +57,24 @@ impl VerticalMenuScreen {
...
+        if !self.menu.fits_area() {
+            // Enable swipe
+            self.swipe_enabled = true;
+            self.swipe_config = SwipeConfig::new()
+                .with_swipe(Direction::Up, SwipeSettings::default())
+                .with_swipe(Direction::Down, SwipeSettings::default());
+            ctx.enable_swipe();
+
+            // Update the menu buttons state
+            self.menu.update_menu(ctx);
```
ibz, line 71: Any reason why `update_menu` is done here but not on the ``else`` branch? Wouldn't it be nicer to do it below the `if`/`else`?

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
```diff
@@ -159,15 +159,12 @@ impl Component for VerticalMenuScreen {
...
-        if !self.menu.is_max_offset() {
-            ToifImage::new(
-                self.menu.area().bottom_center(),
-                theme::ICON_CHEVRON_DOWN_MINI.toif,
-            )
-            .with_align(Alignment2D::BOTTOM_CENTER)
-            .with_fg(theme::GREY_LIGHT)
-            .render(target);
+        // Render the down arrow if the menu overflows and can be scrolled further down
+        if !self.menu.fits_area() && !self.menu.is_max_offset() {
```
ibz, line 163: I don't really like the fact that throughout the code you have two ways to tell whether we have swipe enabled or not...
 * `self.swipe_enabled`
 * `!self.menu.fits_area()`

Is there a reason for that? If not, maybe we can also use `self.swipe_enabled` here for consistency?

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
```diff
@@ -74,26 +76,31 @@ impl VerticalMenu {
         self
     }
 
-    pub fn area(&self) -> Rect {
-        self.bounds
+    /// Check if the menu fits its area without scrolling.
+    pub fn fits_area(&self) -> bool {
+        // Check if the menu fits its area without scrolling
+        self.total_height <= self.bounds.height() && self.offset_y == 0
```
ibz, line 82: The `&& self.offset_y == 0` looks strange here. What is the reason to have that? It *seems* like the condition should be sufficient without it.

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
```diff
@@ -105,21 +112,82 @@ impl VerticalMenu {
         }
     }
 
+    /// Scroll the menu by one item in given direction.
+    /// Relevant only for testing purposes when the animations are disabled.
+    pub fn scroll_item(&mut self, dir: Direction) {
```
ibz, line 117: Can this function be hidden behind a `#[cfg(feature = "ui_debug")]`?

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
```diff
@@ -105,21 +112,82 @@ impl VerticalMenu {
...
-        let menu_overflow = (self.virtual_bounds.height() - self.bounds.height()).max(0);
+        if animation_disabled() {
+            // Relevant only for testing when the animations are disabled
+            // The menu is scrollable until the last button is visible
+            self.offset_y_max = if self.buttons.len() > 0 {
+                self.total_height - unwrap!(self.buttons.last()).area().height()
+            } else {
+                0
+            }
+        } else {
```
ibz, line 159: Would it work if we did this:

* `return` at the end of the `if` block
* drop the `else` - just keep the code that is inside the `else` block
* add a `#[cfg(feature = "ui_debug")]` on the `if` block

Didn't actually try this, just thinking out loud... but if it would work, it would be much nicer!

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -416,10 +416,13 @@ impl<'a> DeviceMenuScreen<'a> {
             Subscreen::Submenu(ref mut submenu_index) => {
                 match self.submenus[*submenu_index].items[idx].action {
                     Some(Action::GoTo(menu)) => {
-                        self.menu_screen.as_mut().unwrap().update_menu(ctx);
                         unwrap!(self.parent_subscreens.push(self.active_subscreen));
                         self.set_active_subscreen(menu);
                         self.place(self.bounds);
+                        self.menu_screen
```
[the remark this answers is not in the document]
bieleluk: It is not possible to use it in `place` function, because the `ctx` is required for that. In other screens, the function can only be called in the `Attach` event but the `DeviceMenuScreen` consists of several screens and this function call is required after each screen switch.

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
```diff
@@ -54,33 +57,24 @@ impl VerticalMenuScreen {
...
-                    TouchEvent::TouchEnd(point) if self.menu.area().contains(point) => Some(
-                        TouchEvent::TouchEnd(point.ofs(Offset::y(self.menu.get_offset()))),
-                    ),
-                    _ => None, // Ignore touch events outside the bounds
-                };
-                shifted_event.map(Event::Touch)
-            }
-            _ => None, // Ignore other events
+    /// Update swipe detection and buttons state based on menu size
+    pub fn update_screen(&mut self, ctx: &mut EventCtx) {
```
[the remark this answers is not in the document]
bieleluk: The function doesn't only update the swipe status but also updates button states for partially overflowing menu items so I can't name it `update_swipe_status`. I'll try to find a better name better name for it.
The function is now public because the of the non-standard behavior of multi-screen `DeviceMenuScreen`  that I mentioned above.

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
bieleluk: the `swipe_config` cannot be an option because it is also used to implement the `Swipable` trait. But I will will try to make the swipe detector an option and thus get rid of the `swipe_enabled` bool.

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
bieleluk: When animations are disabled, the swipe works even though it completely fits the screen.
But I will try to separate the behavior for disabled animations as you suggested above.

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/4908/commits/b840a91e2bc6fb62806d78f8c95a42c782dcc003

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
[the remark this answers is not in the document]
bieleluk: https://github.com/trezor/trezor-firmware/pull/4908/commits/4cc38ca0cabb620afcbe0231eec32ea15002c241

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/4908/commits/ee374df98264a5531dcdc79c799724d1787c6b47

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
```diff
@@ -93,49 +87,58 @@ impl Component for VerticalMenuScreen {
...
+                        _ => {}
                     }
-                    _ => {}
                 }
+                _ => {}
+            };
+        // Relevant only for testing testing purposes when the animations are
+        // disabled. The menu is scrollable up/down item-wise
+        } else if animation_disabled() {
+            if let Some(SwipeEvent::End(dir)) = self.swipe.event(ctx, event, self.swipe_config) {
```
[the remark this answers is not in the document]
bieleluk: This bulky fixup separates the special amination-disabled-case the normal behavior
https://github.com/trezor/trezor-firmware/pull/4908/commits/10704bfa466e5143bd7a9b29df3fddc4a2aa6bae

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
bieleluk: Your requested change is a part of this commit
https://github.com/trezor/trezor-firmware/commit/10704bfa466e5143bd7a9b29df3fddc4a2aa6bae

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
bieleluk: Your requested change is a part of this commit
https://github.com/trezor/trezor-firmware/commit/10704bfa466e5143bd7a9b29df3fddc4a2aa6bae

bieleluk:
Rebasing due to major change in the base branch

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
```diff
@@ -64,37 +62,37 @@ impl VerticalMenu {
...
     pub fn item(mut self, button: Button) -> Self {
         unwrap!(self.buttons.push(button));
         self
     }
 
-    pub fn area(&self) -> Rect {
-        self.bounds
+    /// Check if the menu fits its area without scrolling.
+    pub fn fits_area(&self) -> bool {
+        // Check if the menu fits its area without scrolling
```
ibz, line 72: Duplicate comment

ibz (approved):

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
bieleluk: Implemented, but without additional fixup commit.

PR #4909, opened 2025-04-11
send busy response when trezor is not listening to BLE messages

TychoVrahe:


nordic/trezor/trezor-ble/src/ble/service.c
```diff
@@ -88,3 +88,20 @@ int service_send(struct bt_conn *conn, trz_packet_t *data) {
...
+
+void service_send_busy(void) {
+  struct bt_conn *conn = connection_get_current();
+
+  trz_packet_t *rx = k_malloc(sizeof(*rx));
+
+  static const uint8_t busy_packet[] = {
+      0x3f, 0x23, 0x23, 0x00, 0x03, 0x00, 0x00, 0x00, 0x19, 0x08, 0x0f, 0x12,
+      0x15, 0x44, 0x65, 0x76, 0x69, 0x63, 0x65, 0x20, 0x6c, 0x6f, 0x63, 0x6b,
+      0x65, 0x64, 0x20, 0x6f, 0x64, 0x20, 0x62, 0x75, 0x73, 0x79};
```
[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/7059847d9f0dabe4611e627b534807f5c9750586

nordic/trezor/trezor-ble/src/ble/service.c
```diff
@@ -88,3 +88,20 @@ int service_send(struct bt_conn *conn, trz_packet_t *data) {
     return -EINVAL;
   }
 }
+
+void service_send_busy(void) {
+  struct bt_conn *conn = connection_get_current();
+
+  trz_packet_t *rx = k_malloc(sizeof(*rx));
```
cepetr, line 95: Why `rx`? Wouldn't be `buff` or `data` better?

nordic/trezor/trezor-ble/src/ble/ble.c
```diff
@@ -25,21 +25,20 @@
...
-#include <signals/signals.h>
-
 #include "ble_internal.h"
 
 #define LOG_MODULE_NAME ble
 LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 
 static K_SEM_DEFINE(ble_init_ok, 0, 1);
 
+static uint8_t g_busy_flag = 0;
```
cepetr, line 35: Beacuse this variable is accessed by two threads it would be better to make it atomic (i.e use zephyr/sys/atomic.h)

nordic/trezor/trezor-ble/src/ble/ble_internal.h
```diff
@@ -46,6 +46,9 @@
 #define BT_UUID_TRZ_RX BT_UUID_DECLARE_128(BT_UUID_TRZ_RX_VAL)
 #define BT_UUID_TRZ_TX BT_UUID_DECLARE_128(BT_UUID_TRZ_TX_VAL)
 
+#define BLE_TX_PACKET_SIZE 64
+#define BLE_RX_PACKET_SIZE 244
+
```
cepetr, line 51: Perhaps we should consider having a header with common definitions shared between apps running on nrf52 and stm32. Don't think we should implement it now, but maybe we can open an issue for this.

core/embed/io/ble/stm32/ble.c
```diff
@@ -351,6 +353,24 @@ static void ble_loop(void *context) {
       }
     }
 
+    if (drv->accept_msgs && drv->busy_flag != 0) {
```
cepetr, line 356: Maybe I’m missing something, but it seems to me that with these changes we’ll send the `SET_BUSY` command periodically, regardless of whether it’s needed.

cepetr (changes requested):

nordic/trezor/trezor-ble/src/ble/ble_internal.h
TychoVrahe: agree, created the issue: https://github.com/trezor/trezor-firmware/issues/4992

nordic/trezor/trezor-ble/src/ble/service.c
TychoVrahe: renamed https://github.com/trezor/trezor-firmware/commit/1a37a9fa9cd2b6155297335de128fb271242183f

nordic/trezor/trezor-ble/src/ble/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1a37a9fa9cd2b6155297335de128fb271242183f

core/embed/io/ble/stm32/ble.c
TychoVrahe: `accept_msgs` is required state, while `busy_flag` is set when receiving current state from NRF, so the message is only send if required and current state differs

cepetr (approved):
utACK

PR #4912, opened 2025-04-12
WIP: feat(core): increase allowed lang blob sizes

obrusvit:
- this commit allows translation data and foreign font data to be larger than 65535b (u16)

### DIFF
--- a/core/embed/rust/src/translations/blob.rs
+++ b/core/embed/rust/src/translations/blob.rs
@@ -24,7 +24,7 @@ const INVALID_TRANSLATIONS_BLOB: Error = value_error!(c"Invalid translations blo
 #[repr(packed)]
 struct OffsetEntry {
     pub id: u16,
-    pub offset: u16,
+    pub offset: u32,
 }
 
 pub struct Table<'a> {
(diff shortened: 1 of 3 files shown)


core/embed/rust/src/translations/blob.rs
```diff
@@ -385,12 +389,14 @@ impl<'a> TranslationsHeader<'a> {
         );
 
         // check that there is no trailing data in the proof section
-        if proof_reader.remaining() > 0 {
-            return Err(INVALID_TRANSLATIONS_BLOB);
-        }
+        // if proof_reader.remaining() > 0 {
```
obrusvit, line 392: Temporarily commented out check. There are 2 trailing bytes because of alignment newly set to 4

core/src/apps/management/change_language.py
```diff
@@ -75,6 +75,7 @@ async def do_change_language(
 
     # Verifying header information
     if header.total_len != data_length:
+        print(f"Header total length {header.total_len} != data length {data_length}")
```
obrusvit, line 78: TODO: this check now fails and the `header.total_len` is 2 bytes shorted than `data_length`

obrusvit (commented):
Current known issues.

obrusvit:
Closing in favor of https://github.com/trezor/trezor-firmware/pull/4987

PR #4651, opened 2025-02-19
Drop GitLab config and mentions

matejcik:

PR #4915, opened 2025-04-14
ci(python): remove a workaround for Python 3.7

romanz:
It has been introduced in https://github.com/trezor/trezor-firmware/commit/88cdb72aecc9de519685a4ebbfdcae69758e94fb.

Also, remove Python 3.6 & 3.7 from `shell.nix` and upgrade `coverage` package to 5.5 (to support Python 3.10).


matejcik:
`ci/*.yml` are configs for gitlab which is not in use anymore.
we can remove them.

romanz:
> we can remove them.

IIUC #4651 should remove GitLab-related configs, right?

romanz:
Note: https://pypi.org/project/coverage/4.5.4/ supports Python <=3.8

pyproject.toml
```diff
@@ -15,7 +15,7 @@ nanopb = "^0.4.3"
 
 ## test tools
 pytest = "^8"
-coverage = "<5"  # coverage 5+ uses binary sqlite format
+coverage = "5.5"  # coverage 5+ uses binary sqlite format
```
matejcik, line 18: no need to retain the comment now that we support the newer coverage format

romanz: https://github.com/trezor/trezor-firmware/pull/4915/commits/fecf29ee55c64baa31158ca0ea2f2557b2f803de

romanz:
Rebased and fixed a few small nits.

matejcik (approved):

PR #4916, opened 2025-04-14
BLE branch

mmilata:
* Eckhart UI: yes (#4775)
* BLE pairing: yes (#4891, #4906, more)
* THP: no


core/embed/rust/src/ui/component/ble.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+impl<T> Component for BLEHandler<T>
+where
+    T: Component,
+{
+    type Msg = BLEHandlerMsg<T::Msg>;
+
+    fn place(&mut self, bounds: crate::ui::geometry::Rect) -> crate::ui::geometry::Rect {
```
obrusvit: `use` instead of `crate::full::name`?

core/embed/rust/src/ui/layout_eckhart/firmware/action_bar.rs
```diff
@@ -128,6 +129,14 @@ impl ActionBar {
         Self::new(Mode::PaginateOnly, None, None, None)
     }
 
+    pub fn new_text_only(text: TString<'static>) -> Self {
```
obrusvit, line 132: Note for me: we can use this convenient ctor where available. Not in this PR.

core/src/trezor/wire/context.py
```diff
@@ -101,9 +101,11 @@ def get_context() -> Context:
 
     Result of this function should not be stored -- the context is technically allowed
     to change inbetween any `await` statements.
+
+    Raises KeyError if there is currently no context.
     """
     if CURRENT_CONTEXT is None:
-        raise RuntimeError("No wire context")
+        raise KeyError("No wire context")
```
obrusvit: Note: Thinking if `KeyError` is the right choice :thinking:

mmilata: did not find anything better among standard exceptions, will add a custom one

mmilata: 5c6bffb135b61fa7d4f93e0b34c1472028c7b664

core/embed/rust/src/ui/component/ble.rs
mmilata: f874abf38115bcaf44893b2f432d651ced8eea30

PR #4917, opened 2025-04-15
Don't stop tests in case of transient communication failure

romanz:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ __pycache__/
 proto.gv*
 .DS_Store
 crypto/tests/libtrezor-crypto.so.dSYM/
+.hypothesis/
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import logging
 import os
+import time
 import typing as t
 from enum import IntEnum
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


matejcik:
why?

more specifically, why do you think the next test up will be able to communicate if the previous one didn't? this was a release valve to stop hammering the device after it died

romanz:
Maybe it would be better to keep only `_raw_client.sync_responses()` inside the try-except block?
```diff
diff --git a/tests/conftest.py b/tests/conftest.py
index 3977192b7..579b6ee58 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -290,11 +290,12 @@ def client(
     try:
         try:
             _raw_client.sync_responses()
-            _raw_client.init_device()
         except Exception:
             request.session.shouldstop = "Failed to communicate with Trezor"
             pytest.fail("Failed to communicate with Trezor")
 
+        _raw_client.init_device()
+
         # Resetting all the debug events to not be influenced by previous test
         _raw_client.debug.reset_debug_events()
```
This way, we still stop the suite if the device is not responsive.

romanz:
https://github.com/trezor/trezor-firmware/pull/4917/commits/14a2d98c88f5d9c1deeb1e758a976d0e079c2db8 added a retry loop, so the tests can continue if the error is transient.

matejcik (approved):

romanz:
Re-pushing seems to help...

PR #4919, opened 2025-04-15
Support workflow handlers that may or may not have context

mmilata:
There will be couple of cases where a workflow can be launched either over wire or by user interaction from homescreen. Not sure whether `CURRENT_CONTEXT` should be None like in this PR, or rather some dummy context object, or we should do something different altogether.

### DIFF
--- a/core/src/apps/management/ble/unpair.py
+++ b/core/src/apps/management/ble/unpair.py
@@ -10,16 +10,15 @@ if TYPE_CHECKING:
 async def unpair(msg: BleUnpair) -> None:
     from trezor.messages import Success
     from trezor.ui.layouts import confirm_action
-    from trezor.wire.context import get_context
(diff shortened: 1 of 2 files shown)


matejcik:
perhaps we want a `context.maybe_write()` instead, as a variant of "maybe_call" "i don't care if there is nobody listening"?

@romanz added `context.write()` in #4882 -- the reason `context.write` didn't exist before is that raw writing without waiting for an answer is usually not useful. but no problem in adding it

the alternative, of course, is `try: get_context() except: pass`

mmilata:
Yeah, the current state is probably not the way to go. Since there will be number of flows started both from host and from device menu we probably need some kind of context object for the device menu to intercept the ButtonRequests for `Poll`. Writes on such context would do nothing (except BR) and `read`/`call` would throw an exception.

Issue #4637, opened 2025-02-14
Hardware revisions in SConscript


cepetr:
We should consider separating the two different hardware revision concepts in the SConscript file.

Currently, we have:
- `hw_revision` - This ultimately translates into the `HW_REVISION` symbol used in the C/ASM source code.
- `HW_REVISION` argument – This is used in `configure_board()`

Since these two serve distinct purposes, we might rename them to make distinction clearer:

- `hw_revision`/`HW_REVISION` to `compatible_hw_revision`/`COMPATIBLE_HW_REVISION`
or
- `HW_REVISION` argument to `BOARD_REVISION`

PR #4921, opened 2025-04-15
build(core): allow passing HW_REVISION to make

mmilata:


core/Makefile
```diff
@@ -127,6 +127,7 @@ SCONS_VARS = \
 	CMAKELISTS="$(CMAKELISTS)" \
 	DISABLE_OPTIGA="$(DISABLE_OPTIGA)" \
 	DISABLE_TROPIC="$(DISABLE_TROPIC)" \
+	HW_REVISION="$(HW_REVISION)" \
```
cepetr, line 130: I’m not completely sure, but I think this won’t propagate correctly, since hw_revision is later forced to a specific value (for example, in trezor_t3W1_revA/B/C.py). The handling of hardware versions is a bit messy at the moment - see #4637.

mmilata: I see, though the variable still decides which `revA/B/C.py` gets imported in [`T3W1/__init__.py`](https://github.com/trezor/trezor-firmware/blob/main/core/site_scons/models/T3W1/__init__.py). It seemed to work on T3W1revB and T3W1revC, no idea what it builds for other models.

cepetr: Yes, you are right, it works - it confused me again.

cepetr (approved):

PR #4926, opened 2025-04-16
feat(core): allow excluding line numbers from bytecode

romanz:
Setting `MICROPY_ENABLE_SOURCE_LINE=0` helps reducing T3T1's debug firmware size by ~2%:
```
$ TREZOR_MODEL=T3T1 MICROPY_ENABLE_SOURCE_LINE=0 PYOPT=0 make -C core build_firmware
<snip>
           FLASH:     1,629,456 B      1664 KB     95.63%
```

Before this PR:
```
           FLASH:     1,662,224 B      1664 KB     97.55%
```


romanz:
Tested on T3T1 device:
```
[10:45:50.166] 45244000 trezor.loop ERROR exception:
[10:45:50.166] Traceback (most recent call last):
[10:45:50.167]   File "trezor/loop.py", in _step
[10:45:50.168]   File "trezor/wire/context.py", in with_context
[10:45:50.168]   File "apps/base.py", in handle_Cancel
[10:45:50.169] ActionCancelled: 
```

matejcik:
we should default to 0 for hardware builds (and 1 for unix builds)

romanz:
CI seems to be stuck:

romanz:
Rebased over `main`.

matejcik (approved):
utACK
what's that makefile syntax? i'm not familiar with that

romanz:
> what's that makefile syntax? 

I've used it to allow us to have different defaults for hardware and emulator builds (instead of a single global default value):
https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html

PR #4927, opened 2025-04-16
fix(core): fix USB VCP freeze on T3W1 in prodtest

cepetr:
This PR fixes a freeze in the USB driver on T3W1 when receiving packets longer than 64 bytes.

On T3W1, we use USB High Speed, whereas other products use USB Full Speed. USB HS allows packets up to 512 bytes, which we weren’t prepared for. This PR provides a hotfix for the issue in prodtest, but we need to investigate further to determine if this difference should be accounted for elsewhere.


TychoVrahe (approved):
utACK

PR #4928, opened 2025-04-16
fix(core): update EIP-1559 fee-related labels

### LINKED ISSUE #4819: EIP1559: Change fee texts
Current solution EVM EIP-1559 fees:

Rename second param to `Max fee per gas`
Rename third param to `Max priority fee`

Would be good to change in for `core` and `legacy`

romanz:
Fixes #4819.

Translations will be updated in a separate PR.


matejcik (approved):
ACK from code side, leaving to @MiroslavProchazka and @tomasklim to approve the copy

MiroslavProchazka:
@tomasklim approves, we can use the copy as it is.

PR #4929, opened 2025-04-17
fix(core): dump allocated QSTRs also in debug firmware

romanz:
Otherwise, it's hard to understand which QSTR causes the assertion error.

Previously, it was enabled only for debug emulator builds.


M1nd3r (approved):
LGTM, this change works as desired - tested on T2T1 HW

PR #4935, opened 2025-04-21
feat(core): check that GC restores free heap memory

romanz:
Enabled only for frozen debug builds.


romanz:
Tested on T3T1 device:
```
$ pytest tests/device_tests/ -k 'not authenticate and not recovery' -m 'not sd_card' --random-order-seed=1 --random-order-bucket=module --color yes -v --durations 50
Apr 22 16:55:24 ============================= test session starts ==============================
<snip>
Apr 22 19:09:35 = 1331 passed, 161 skipped, 86 deselected, 8 xfailed, 1 warning, 1 retried in 8050.94s (2:14:10) =
```

core/src/trezor/utils.py
```diff
@@ -94,17 +94,22 @@ def unimport_end(mods: set[str], collect: bool = True) -> None:
...
 
-    def __exit__(self, _exc_type: Any, _exc_value: Any, _tb: Any) -> None:
+    def __exit__(self, exc_type: Any, _exc_value: Any, _tb: Any) -> None:
         assert self.mods is not None
         unimport_end(self.mods, collect=False)
         self.mods.clear()
         self.mods = None
         gc.collect()
 
+        if __debug__ and exc_type is SystemExit:
```
mmilata, line 110: I don't understand why `SystemExit` - wouldn't that mean `check_free_heap` is only called once and always succeeds as `free_heap == 0`?

romanz: Thanks! 
It was a mistake, I have meant to use `exc_type is not SystemExit` - otherwise, the `SystemExit` exception seems to cause this check to fail when exiting the emulator via `Ctrl-C`:
```
^CGC: total: 20728576, used: 41120, free: 20687456
 No. of 1-blocks: 235, 2-blocks: 173, max blk sz: 128, max free sz: 630640
  File "main.py", line 67, in <module>
  File "trezor/utils.py", line 111, in __exit__
AssertionError: Free heap decreased by 160 bytes
```

I suspect that it's due to `SystemExit`'s attributes (which are probably allocated on the heap):
https://github.com/micropython/micropython/blob/0b729623366dab3aa2575eeab934b2629f8e52c5/py/objexcept.h#L32-L38 

Since we rely on `SystemExit` to stop the emulator in order to flush coverage data, would it be OK to skip the GC check in this case?

romanz: Fixed the incorrect `SystemExit` predicate in https://github.com/trezor/trezor-firmware/pull/4935/commits/7d4afabdc58e2212891ec8e7291257029d5e8cda

mmilata (approved):

PR #4937, opened 2025-04-21
feat(core): don't use heap for storing coverage data

romanz:
It is needed for https://github.com/trezor/trezor-firmware/pull/4935.


romanz:
Tested together with https://github.com/trezor/trezor-firmware/pull/4935#issuecomment-2821819723.

mmilata (approved):
LGTM

PR #4936, opened 2025-04-21
fix(legacy): use `Gwei` when ETH fee amount is between 1e3 and 1e9

romanz:
For example:

PR #4941, opened 2025-04-22
fix(core): don't use Gwei when fee amount is below 1000 Wei

romanz:
Similar to how it is done in #4936.


romanz:
https://github.com/trezor/trezor-firmware/pull/4936 is superseded by #4981.

PR #4944, opened 2025-04-22
change RGB LED pin configuration to input when LED is off

TychoVrahe:
[no changelog]


TychoVrahe:
we deem this not useful

PR #4946, opened 2025-04-22
perf(core): optimize `AddressDetails::xpubs` layout

romanz:
It reduces `AddressDetails` size by 128 bytes from:
```
`ui::layout_delizia::component::address_details::AddressDetails`: 2080 bytes, alignment: 4 bytes
```
to
```
`ui::layout_delizia::component::address_details::AddressDetails`: 1952 bytes, alignment: 4 bytes
```

Relevant stack sizes are also reduced to:
```
11536	trezor_lib::ui::api::firmware_micropython::new_flow_get_address
16056	trezor_lib::ui::api::firmware_micropython::new_flow_confirm_output
```
(https://github.com/trezor/trezor-firmware/actions/runs/14601983369/job/40962113990?pr=4946)

from:
```
11800	trezor_lib::ui::api::firmware_micropython::new_flow_get_address
16312	trezor_lib::ui::api::firmware_micropython::new_flow_confirm_output
```
(https://github.com/trezor/trezor-firmware/actions/runs/14591197885/job/40926681094?pr=4941)

### DIFF
--- a/core/embed/rust/src/ui/layout_caesar/component/mod.rs
+++ b/core/embed/rust/src/ui/layout_caesar/component/mod.rs
@@ -25,7 +25,7 @@ pub use loader::{Loader, LoaderMsg, LoaderStyle, LoaderStyleSheet, ProgressLoade
 pub use result::ResultScreen;
 pub use welcome_screen::WelcomeScreen;
 
-#[cfg(feature = "translations")]
+#[cfg(all(feature = "micropython", feature = "translations"))]
 mod address_details;
 mod changing_text;
 #[cfg(feature = "translations")]
@@ -44,7 +44,7 @@ mod share_words;
 mod show_more;
 mod title;
 
-#[cfg(feature = "translations")]
+#[cfg(all(feature = "micropython", feature = "translations"))]
 pub use address_details::AddressDetails;
 
 pub use changing_text::ChangingTextLine;
--- a/core/embed/rust/src/ui/layout_delizia/component/address_details.rs
+++ b/core/embed/rust/src/ui/layout_delizia/component/address_details.rs
@@ -2,6 +2,7 @@ use heapless::Vec;
 
 use crate::{
     error::Error,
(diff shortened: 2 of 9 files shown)


core/embed/rust/src/ui/layout_delizia/component/address_details.rs
```diff
@@ -2,6 +2,7 @@ use heapless::Vec;
 
 use crate::{
     error::Error,
+    micropython::buffer::StrBuffer,
```
mmilata, line 5: Please also add `feature = "micropython"` gate to component/mod.rs so that it doesn't break non-micropython build. For this reason we generally avoid StrBuffer in `src/ui`, though here it makes sense as
* it's in an array,
* in a layout that's unlikely to be used by bootloader.

romanz: Sounds good, thanks!
Fixed in aafe26b2dec3c65f286d1eae8a1d3c80ccdfcfd1.

romanz: Also applied this optimization to Bolt and Caesar in f85b2565a7.

core/embed/rust/src/ui/layout_caesar/component/mod.rs
```diff
@@ -44,7 +44,7 @@ mod share_words;
 mod show_more;
 mod title;
 
-#[cfg(feature = "translations")]
+#[cfg(all(feature = "micropython", feature = "translations"))]
```
mmilata, line 47: I think these are needed both before `mod address_details` and `pub use address_details::AddressDetails`?

romanz: 22baa1922a

bosomt:
**QA OK**

- BTC XPUB generation works as expected

PR #4947, opened 2025-04-22
perf(core): encode `Timer` using a single `u32`

romanz:
Following https://github.com/trezor/trezor-firmware/pull/4797#discussion_r2003021530.

### DIFF
--- a/core/embed/rust/src/ui/component/base.rs
+++ b/core/embed/rust/src/ui/component/base.rs
@@ -1,6 +1,7 @@
 use heapless::Vec;
 
 use crate::{
+    error::Error,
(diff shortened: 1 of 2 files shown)


matejcik:
i did say in the previous comment that we should merge Timer and TimerToken, but we actually still should keep a semantic `TimerToken` type for the value of `Event::Timer(v)` and related places. The `TimerToken` will guarantee that its value does not have the highest bit set.
The internals of `Timer` will be a `u32`, and it will have the ability to emit a `TimerToken` for `self.0 & TOKEN_BITMASK`

matejcik (approved):

PR #4949, opened 2025-04-23
fix(core): remove debug-related data from `PYOPT=1` builds

romanz:
Found by running:
```
$ strings core/build/firmware/firmware.bin | grep -i debuglink
DebugLinkLayout
DebugLinkResetDebugEvents
DebugLinkWatchLayout
DebugLinkDecision
DebugLinkEraseSdCard
DebugLinkFlashErase
DebugLinkGetState
DebugLinkLog
DebugLinkMemory
DebugLinkMemoryRead
DebugLinkMemoryWrite
DebugLinkOptigaSetSecMax
DebugLinkRecordScreen
DebugLinkReseedRandom
DebugLinkState
DebugLinkStop
```

This PR reduces T3T1 release universal firmware size by ~3.5kB:
```
           FLASH:     1576720 B      1664 KB     92.53%
```

Before this PR (@ 9037c398eb):
```
           FLASH:     1580304 B      1664 KB     92.74%
```

(Tested with `TREZOR_MODEL=T3T1 PYOPT=1 make -C core build_firmware`)


.github/workflows/core.yml
```diff
@@ -226,7 +226,7 @@ jobs:
         with:
           submodules: recursive
       - uses: ./.github/actions/environment
-      - run: nix-shell --run "poetry run make -C core build_unix_frozen"
```
romanz, line 229: `make build_unix_frozen` is not setting PYOPT=0, so `cargo clippy` fails due to missing debug-related QSTRs.

core/SConscript.firmware
```diff
@@ -506,6 +506,8 @@ PROTO_SOURCES_DIR = '../../../common/protob/'
 exclude_list = [PROTO_SOURCES_DIR + 'messages-bootloader.proto']
 if not THP:
     exclude_list.append(PROTO_SOURCES_DIR + 'messages-thp.proto')
+if PYOPT == '1':
```
mmilata, line 509: Nit: consider using `PYOPT != '0'` everywhere so as not to fail when somebody uses `PYOPT=3`.

core/embed/rust/librust_qstr.h.mako
```diff
@@ -28,6 +28,10 @@ ALTCOIN_PREFIXES = (
     "u2f",
 )
 
+DEBUG_PREFIXES = (
+  "debug",
+)
```
mmilata, line 33: Nit: Would be nice to have these prefix lists in one place, not sure where though.

Thinking if we can get rid of the lists entirely and use prefixes based on rust features. Probably too much magic and would not solve `librust_qstr.h` anyway.

core/SConscript.firmware
romanz: Thanks, fixed in a2df0f3dbb

core/embed/rust/librust_qstr.h.mako
romanz: Good idea, thanks!
Moved prefix lists to `cointool.py` in dabd98f777, which helped to exclude Nostr-related QSTR from `BITCOIN_ONLY` build :)

mmilata (approved):

PR #4952, opened 2025-04-24
Suspend and Resume BLE/NRF

TychoVrahe:
This PR addresses several stuff around BLE/NRF.

NRF is suspended to either power-off (if no connection is active), or system-off with connection active.

Uart on NFC is disabled by default, only to be turned on for tests, or when otherwise needed.

BLE packet size is unified to 244B in both directions.

Wakeup on BLE is implemented. Know issue is that it is more aggressive that we'd like to, waking up the device on stuff like connection lost so that will need addressing, but leaving that to further work.

Internal communication tests are fixed and some signal names were fixed too.

Deinitializaing/initializing comm interfaces in bootloader has been refactored. Along with that, background commnunication in menu is only active if the menu was entered after pressing 'Install firmware' button. (in next PR, this button will be renamed to 'connect to host'.


core/embed/projects/bootloader/workflow/workflow.h
```diff
@@ -80,3 +83,15 @@ workflow_result_t workflow_auto_update(const vendor_header *const vhdr,
...
+
+size_t workflow_ifaces_init(const vendor_header *const vhdr,
+                            const image_header *const hdr,
+                            protob_io_t ifaces[2]);
+
+void workflow_ifaces_deinit(protob_io_t ifaces[2]);
+
+void workflow_ifaces_pause(protob_io_t ifaces[2]);
+
+void workflow_ifaces_resume(const vendor_header *const vhdr,
```
cepetr, line 95: Maybe we could consider replacing the first two arguments of `workflow_iface_init` and `workflow_ifaces_resume` with a single `secbool usb21_landing`, and check that it is properly set in all places where these functions are called.

Not sure about this, but I think the only situation where we want to set usb21_landing is when the device is empty. So there are probably just two locations where we need to set it - the empty device workflow and when the device has just been wiped.

cepetr (changes requested):

core/embed/projects/bootloader/workflow/workflow.h
TychoVrahe: changed in http://github.com/trezor/trezor-firmware/commit/d9768eda650de86c3bb3966547e0192af3788c75

core/embed/projects/bootloader/workflow/wf_bootloader.c
```diff
@@ -75,7 +78,9 @@ workflow_result_t workflow_menu(const vendor_header* const vhdr,
       return WF_OK_REBOOT_SELECTED;
     }
     if (menu_result == MENU_WIPE) {  // wipe
+      workflow_ifaces_pause(ifaces);
       workflow_result_t r = workflow_wipe_device(NULL);
+      workflow_ifaces_resume(vhdr, hdr, ifaces);
```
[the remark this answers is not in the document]
TychoVrahe: hmm depends on result of the workflow. I actually think we only want to resume on cancel, other results will result in RSOD or WIPE screens that require rebooting: http://github.com/trezor/trezor-firmware/commit/d9768eda650de86c3bb3966547e0192af3788c75

core/embed/io/ble/stm32/ble.c
```diff
@@ -463,10 +466,58 @@ void ble_deinit(void) {
...
+      nrf_deinit();
+    } else {
+      nrf_suspend();
+    }
+
+    drv->initialized = false;
+  }
+}
+
+void ble_resume(void) {
```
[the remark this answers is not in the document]
TychoVrahe: yeah i like that : https://github.com/trezor/trezor-firmware/commit/320b0c69226dc4ef67b0d03702813420cf018792

core/embed/io/ble/stm32/ble.c
```diff
@@ -463,10 +466,58 @@ void ble_deinit(void) {
...
+  if (reboot) {
+    nrf_reboot();
+  }
+
+  if (accept_msgs) {
+    ble_start();
+  }
 
-  drv->initialized = false;
+  drv->mode_requested = mode;
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/320b0c69226dc4ef67b0d03702813420cf018792

core/embed/projects/bootloader/workflow/workflow.h
```diff
@@ -80,3 +83,15 @@ workflow_result_t workflow_auto_update(const vendor_header *const vhdr,
 workflow_result_t workflow_ble_pairing_request(const vendor_header *const vhdr,
                                                const image_header *const hdr);
 #endif
+
+size_t workflow_ifaces_init(const vendor_header *const vhdr,
```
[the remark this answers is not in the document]
TychoVrahe: can't have flexible length array inside the struct, and doing 
typedef struct {
  protob_io_t * ifaces;
  int count;
} workflow_ifaces_t;

while keeping the array of ifaces outside negates part of the benefit. While i agree that current version is somewhat ugly, i'd wait with further refactoring until it bothers us a bit more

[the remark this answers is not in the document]
cepetr:
```c 
typedef struct {
  protob_io_t ifaces[2];  <-- 2 was previously missing
  int count;
} workflow_ifaces_t; 
```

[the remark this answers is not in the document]
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/86a3ad19f0fd933ad17863f459b2aa8d77188d99

core/embed/projects/bootloader/workflow/wf_host_control.c
```diff
@@ -151,10 +132,48 @@ workflow_result_t workflow_host_control(const vendor_header *const vhdr,
...
+
+  protob_init(&ios->ifaces[1], ble_iface);
+  cnt++;
+#endif
+
+  ios->count = cnt;
+}
+
+void workflow_ifaces_deinit(protob_ios_t *ios) {
   systick_delay_ms(100);
```
cepetr, line 157: I wonder if we could get rid of this delay and wait in more deterministic way... I have no simple idea at the moment..

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -855,6 +904,29 @@ bool nrf_get_info(nrf_info_t *info) {
...
+
+  uint8_t data[1] = {MGMT_CMD_SYSTEM_OFF};
+  if (!nrf_send_msg(NRF_SERVICE_MANAGEMENT, data, 1, NULL, NULL)) {
+    return false;
+  }
+
+  uint32_t timeout = ticks_timeout(100);
+
+  while (!ticks_expired(timeout)) {
+    if (tsqueue_empty(&drv->tx_queue) && !drv->pending_spi_transaction) {
```
cepetr, line 921: `drv->pending_spi_transaction` probably requires `irq_lock`/`unlock` since it's cleared in the interrupt handler. 

We may also include `__WFI` in the loop.

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -855,6 +904,29 @@ bool nrf_get_info(nrf_info_t *info) {
...
+      break;
+    }
+  }
+  systick_delay_ms(10);
```
cepetr, line 925: Is this delay really necessary? looks a bit hacky.

cepetr (changes requested):

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/21a12d251c9fcebd35e8cad1f5a98460d369b7dc

core/embed/projects/bootloader/workflow/workflow.h
TychoVrahe: as discussed, further simplification: https://github.com/trezor/trezor-firmware/commit/cf5c584c34fb7d4fab694d28a6ec38a40f73b009

core/embed/projects/bootloader/workflow/wf_host_control.c
TychoVrahe: lets keep it as is and figure it out later if needed

cepetr (approved):
utACK

PR #4954, opened 2025-04-24
ci(common): run `ClusterFuzzLite PR fuzzing` job daily

romanz:
It seems that this job takes >1h:

image

Most of the time is spent waiting:

image


romanz:
Resolved in https://github.com/trezor/trezor-firmware/pull/4998.

PR #4955, opened 2025-04-24
Introduce backup RAM driver

kopecdav:
This PR introduce small driver which controls T3W1 backup RAM. 

Idea is that backup_ram driver will have a static union ('backup_ram_data_t')  pointing to the backup RAM memory and offer user an API to access the data storage.

Right now we will only plan to store data for fuel gauge, but data storage could be extended by putting more fields into the union  'data' section.

 ```
typedef union {
  uint8_t bytes[BACKUP_RAM_SIZE];
  struct {
    volatile fuel_gauge_backup_storage_t fg;
    // < Room for other data structures >
  } data;
} backup_ram_data_t;
```

API offers functions for full HW erase via RAMCFG register and FW erase of unallocated part

PR also add small prodtest for testing, but it will be probably removed before we merge to main.

### CODE REVIEW

**Comment by TychoVrahe** [comment]
Regarding the structure of data in backup ram.

Assuming:
- the structure is independent on current version of FW/Bootloader/prodtest
- should be securely upgradable/extensible
- we should be able to detect incompatibility if we need to introduce some at any point.
- we should be able to detect garbage in the backup sram, at least to some extent

I would suggest that
- backup_ram module would be a single source of truth when in comes to contents of the RAM, no delegation to fuel gauge via 'fuel_gauge_backup_storage_t', as if this changes for some reason, other stuff in the RAM will be shifted
- the contents should have some 'magic' header, i'd suggest four bytes, elsewhere we use stuff like TRZB, TRZS, etc. so something similar, but don't collide with existing ones
- after the header there should be a 'version' indicated, to be bumped where we add/change stuff.
- maybe add CRC at the end? no sure we really need it, but might be a bit safer. If so, recalculate on any write.

File: core/embed/sys/backup_ram/inc/sys/backup_ram.h [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+  uint32_t last_capture_timestamp;
+} fuel_gauge_backup_storage_t;
+
+typedef union {
+  uint8_t bytes[BACKUP_RAM_SIZE];
+  struct {
+    volatile fuel_gauge_backup_storage_t fg;
+    // < Room for other data structures >
+  } data;
+} backup_ram_data_t;
```
cepetr, line 43: This structure is internal to the module and does not need to be part of the public interface.

core/site_scons/models/T3W1/trezor_t3w1_revC.py
```diff
@@ -77,6 +77,10 @@ def configure(
     paths += ["embed/io/i2c_bus/inc"]
```
cepetr, line 77: I think we are very close to kicking off the T3W1 A revision, but B is still in use. We should add this at _revB as well.

core/embed/sys/linker/stm32u5g/prodtest.ld
```diff
@@ -35,6 +36,9 @@ _accessible_ram_1_end = MCU_SRAM4 + MCU_SRAM4_SIZE;
 _bootargs_ram_start = BOOTARGS_START;
 _bootargs_ram_end = BOOTARGS_START + BOOTARGS_SIZE;
 
+_backup_ram_start = ORIGIN(BACKUP_RAM);
```
cepetr, line 39: These symbols are unused now, maybe we don't need to add them.

core/embed/sys/backup_ram/backup_ram.c
```diff
@@ -0,0 +1,110 @@
...
+
+  drv->initialized = true;
+
+  return BACKUP_RAM_OK;
+}
+
+backup_ram_status_t backup_ram_deinit(void) {
+  backup_ram_driver_t *drv = &backup_ram_driver;
+
+  HAL_RAMCFG_DeInit(&drv->hramcfg);
```
cepetr, line 68: We should check whether `drv->initialized` is set. The same applies to other functions.

cepetr (changes requested):

core/embed/sys/backup_ram/inc/sys/backup_ram.h
```diff
@@ -0,0 +1,65 @@
...
+
+// Initialize backup RAM driver
+backup_ram_status_t backup_ram_init(void);
+
+// Deinitialize backup RAM driver
+backup_ram_status_t backup_ram_deinit(void);
```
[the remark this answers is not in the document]
kopecdav: Fixed in https://github.com/trezor/trezor-firmware/pull/4955/commits/294d67d756da51b519360059bad28d00f012aa3b

core/embed/sys/linker/stm32u5g/prodtest.ld
kopecdav: fixed in https://github.com/trezor/trezor-firmware/pull/4955/commits/294d67d756da51b519360059bad28d00f012aa3b

core/embed/sys/backup_ram/backup_ram.c
kopecdav: I fixed wherever I found it possible in https://github.com/trezor/trezor-firmware/pull/4955/commits/294d67d756da51b519360059bad28d00f012aa3b, but in cannot put it to deinit fucntion since I call it during intialize function cleanup where initialized variable is still false

kopecdav: Same applies for erase function, which i just fix here https://github.com/trezor/trezor-firmware/pull/4955/commits/24f3928db12851380f07b1fe4943f46c3787a17d. Erase function calls HAL driver which should be initialized to be called properly, but the HAL driver can deal with that itself.

core/site_scons/models/T3W1/trezor_t3w1_revC.py
TychoVrahe: Yes it should be added for each revision.

core/embed/sys/linker/stm32u5g/prodtest.ld
kopecdav: Backup ram removed from linker completely in https://github.com/trezor/trezor-firmware/pull/4955/commits/763566d348442bf72d37f3ef8a9bb2e4c1dc0791

core/site_scons/models/T3W1/trezor_t3w1_revC.py
kopecdav: Done in https://github.com/trezor/trezor-firmware/pull/4955/commits/85e1ac65b90af2ae3b818b3d44b0292bb3774a26

core/embed/projects/prodtest/cmd/prodtest_backup_ram.c
```diff
@@ -0,0 +1,176 @@
...
+
+  status = backup_ram_store_power_manager_data(&pm_data);
+
+  if (status != BACKUP_RAM_OK) {
+    cli_error(cli, CLI_ERROR, "Failed to write backup RAM");
+    backup_ram_deinit();
+    return;
+  }
+
+  backup_ram_deinit();
```
cepetr, line 67: We should not call `backup_ram_init()` and `backup_ram_deinit()` directly in this file. Instead, we should initialize and deinitialize the driver in `drivers_init()` and `drivers_deinit()`. Maybe I can add it to the bootloader or kernel as well.

core/embed/projects/prodtest/cmd/prodtest_backup_ram.c
```diff
@@ -0,0 +1,176 @@
...
+    cli_error_arg(cli, "Expecting soc value to store to backup RAM.");
+    return;
+  }
+
+  if (cli_arg_count(cli) > 1) {
+    cli_error_arg_count(cli);
+    return;
+  }
+
+  backup_ram_status_t status = backup_ram_init();
```
cepetr, line 44: `if (cli_arg_count(cli) == 0)` --> implies that at least one argument is needed (=> soc_percent is mandatory)
`if (cli_has_arg(cli, 'soc_percent)` --> implies that soc_percent is optional

cepetr (changes requested):

core/embed/projects/prodtest/cmd/prodtest_backup_ram.c
kopecdav: Fixed in https://github.com/trezor/trezor-firmware/pull/4955/commits/5c40ac0d88c22dc2c0f7de5881cfd437ee2ec0c7

core/embed/projects/prodtest/cmd/prodtest_backup_ram.c
kopecdav: Fixed in [dd15822](https://github.com/trezor/trezor-firmware/pull/4955/commits/dd15822f253f8356d37f866ea75d640dd4e6456c)

core/embed/projects/prodtest/cmd/prodtest_backup_ram.c
cepetr: On other place in prodtest, we use the  following pattern if there's 1 mandatory argument:

```c
static void prodtest_backup_ram_write(cli_t* cli) {
   uint32_t soc = 0;
 
   if (!cli_arg_uint32(cli, "soc_percent", &soc)) {
     cli_error_arg(cli, "Expecting soc value to store to backup RAM.");
     return;
   }
   
   if (cli_arg_count(cli) > 1) {
    cli_error_arg_count(cli);
    return;
  }
   
   ...
  ```

File: core/embed/projects/bootloader/main.c [hunk 8]
```diff
@@ -126,6 +132,9 @@ static void drivers_init(secbool *touch_initialized) {
 
 static void drivers_deinit(void) {
 #ifdef FIXED_HW_DEINIT
+#ifdef USE_BACKUP_RAM
```
cepetr, line 135: `backup_ram_deinit()` should be outside the `FIXED_HW_DEINIT` block

cepetr (changes requested):

core/embed/projects/bootloader/main.c
kopecdav: fixed in https://github.com/trezor/trezor-firmware/pull/4955/commits/f289f3baaa26db5929dbb91393cd4f5dcf489b84

core/embed/projects/prodtest/cmd/prodtest_backup_ram.c
kopecdav: fixed in https://github.com/trezor/trezor-firmware/pull/4955/commits/f289f3baaa26db5929dbb91393cd4f5dcf489b84

cepetr (approved):
utACK

kopecdav:
This PR do not deals with scenario where the device is reset with long button press. In such case the backup RAM would be erased. We will deal with this in near future.

TychoVrahe:
added issue for the button press so its not forgotten: https://github.com/trezor/trezor-firmware/issues/4979

PR #4957, opened 2025-04-24
chore(core): shorten some foreign copy

obrusvit:
- shorten some strings related to reset/recovery flows of Delizia

TODO:
- [x] `recovery__share_already_entered` should stay the same for other layouts (`Bolt`, `Caesar`, `Eckhart`) because it fits there. Need https://github.com/trezor/trezor-firmware/pull/4815 for that.
- [x] fixtures


obrusvit:
:de: 
image
image

:es: 
image
image
image
image

:fr:
image
image
image

:brazil: (Portuguese)
image
image
image

core/translations/es.json
```diff
@@ -624,7 +624,12 @@
     "recovery__num_of_words": "Elige el número de palabras de la c. seg.",
     "recovery__only_first_n_letters": "Solo tendrás que seleccionar las primeras 2-4 letras de cada palabra.",
     "recovery__progress_will_be_lost": "Se perderá todo el progreso.",
-    "recovery__share_already_entered": "Ya se ha introducido el recurso compartido",
+    "recovery__share_already_entered": {
```
ibz, line 627: "Ya se ha introducido el rec." on all models will do, since just above (line 424) you used "rec." for share already. Not sure it makes total sense, but at least we stay consistent within the same translation. *Also*, introducing this inconsistency between different models (which is surely nice that we can have, but should be for very exceptional cases or we risk complicating everything even more) is unnecessary here.

core/translations/fr.json
```diff
@@ -680,7 +680,7 @@
     "reset__num_of_share_holders_template": "{0} personnes ou emp. détiendront chacun un fragm.",
     "reset__num_of_shares_advanced_info_template": "Chaque fragm. de récup. se compose de {0} mots. Vous choisirez le nbr seuil de fragm. néc. pour former le Groupe {1}.",
     "reset__num_of_shares_basic_info_template": "Chaque fragm. de récup. se compose de {0} mots. Vous choisirez le nbr de fragm. néc. pour récup. votre portef.",
-    "reset__num_of_shares_how_many": "Combien de fragments de sauvegarde de portefeuille voulez-vous créer ?",
+    "reset__num_of_shares_how_many": "Combien de fragments de sauvegarde de voulez-vous créer ?",
```
ibz, line 683: Drop the "de" since you dropped the corresponding noun (portefeuille)...

ibz (changes requested):

core/translations/es.json
obrusvit: I kept it the same for other layouts because it fits there.

The ability to separate the strings per layout was added specifically for these cases so propose sticking with the suggested changes.

core/translations/fr.json
obrusvit: Nice catch, https://github.com/trezor/trezor-firmware/pull/4957/commits/4a6c8e568a0351c397c5bb6ce00a7424507e8a5f

core/translations/es.json
obrusvit: Additional note:
- we use the per-layout strings much more in `Eckhart`
- asserting` the translation string is `"recovery__share_already_entered"` in tests works as of [pull/4815](https://github.com/trezor/trezor-firmware/pull/4815)

ibz (approved):

core/translations/es.json
obrusvit: As per offline discussion. I changed the Delizia version to `""Ya se ha introducido el rec." `

PR #4958, opened 2025-04-25
Ethereum approve flow

ibz:
This PR introduces a separate flow for Ethereum's `approve` call.

## Test commands

`trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d "095ea7b3000000000000000000000000574bbb36871ba6b78e27f4b4dcfb76ea0091880b0000000000000000000000000000000000000000000000000000000000000064" --gas-price 14 --gas-limit 14 --nonce 0 0xfc6b5d6af8a13258f7cbd0d39e11b35e01a32f93 0`

### Approve Unlimited amount

`trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d "095ea7b3000000000000000000000000574bbb36871ba6b78e27f4b4dcfb76ea0091880bffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff" --gas-price 14 --gas-limit 14 --nonce 0 0xfc6b5d6af8a13258f7cbd0d39e11b35e01a32f93 0`

### Uniswap

`trezorctl ethereum sign-tx -n "m/44'/60'/0'/0/0" -d "095ea7b3000000000000000000000000e592427a0aece92de3edee1f18e0157c058615640000000000000000000000000000000000000000000000000000000000000064" --gas-price 14 --gas-limit 14 --nonce 0 0xfc6b5d6af8a13258f7cbd0d39e11b35e01a32f93 0 -c 200901`

PR #4960, opened 2025-04-25
Payment requests in altcoins

andrewkozlik:
Specification (branch with recent updates): https://github.com/satoshilabs/slips/blob/slip24/slip-0024.md
Notion issue: https://www.notion.so/satoshilabs/Payment-requests-1c8dc526060680feb8face4c3bae950a
Slack channel: [wg_slip24](https://satoshilabs.slack.com/archives/C08LFLVN19R)

This PR implements payment requests in Ethereum and Ripple signing, and adds the `mac` field to altcoin `*Address` messages. It also introduces the `PaymentNotification` message, which is just an informational screen for the user, but behind the scenes it is intended to verify a `CoinPurchaseMemo`.

In the messages related to payment requests I changed the required fields to optional with default values. In principle they are all required, but this allows them to be deprecated in the future. See https://protobuf.dev/best-practices/dos-donts/#add-required. This is a breaking change, but I don't see it as a problem since the `PaymentRequest` message is marked as experimental, which indicates that a message is intended for development and beta testing only and its definition may change at any time.

There are still UI changes that are missing from this PR, see below. This PR can either be merged as is (after review and updating UI test fixtures), or we can add the UI changes before merging. It doesn't make any difference, since this feature is experimental and does not work in non-debug builds, because of the missing public key, so there is no way for the user to come into contact with it.

TODO in UI (either now or in a later PR):
- Display text details memo in Bitcoin signing.
- Display text memos and text details memos in all altcoin signing, currently Ethereum and Ripple.
- Add device tests for payment notifications.
- Update UI test fixtures.

TODO after this is merged:
- Please ask @szymonlesisz or someone from the connect team to update the protobuf messages in connect. Namely to expose the `mac` field in `Address` responses for altcoins, the `payment_req` field in altcoin `SingTx` messages and the `PaymentNotification` message.

TODO before the official release of this feature:
- Remove the experimental flags from messages.
- Set the `PUBLIC_KEY` in `PaymentRequestVerifier` to Invity server's public key.

TODO in later PRs:
- Implement payment requests in Solana, Cardano, Stellar and Tezos signing.


common/protob/messages-common.proto
```diff
@@ -169,3 +169,48 @@ message HDNodeType {
...
+/**
+ * Structure representing a SLIP-0024 payment request.
+ * @next TxRequest
+ * @embed
+ */
+message PaymentRequest {
+    option (experimental_message) = true;
+
+    optional bytes nonce = 1;                        // the nonce used in the signature computation
+    optional string recipient_name = 2 [default=""]; // merchant's name
```
matejcik: should we validate the contents of the string (and the rest of the newly-optional fields) in any way? like, do we explicitly want it to be non-empty or do we not care?

e.g., it doesn't really make sense to post a TextDetailsMemo with both fields empty
if it's not breaking for any functionality, then we can of course keep it as is and not bother with validation
more like, making the field `required` indicates to the caller that they should supply some sort of value -- as opposed to "optional = you don't need to bother"

(again, i don't believe that the compatibility argument applies to Trezor wire protocol; if in the future we change the spec to differentiate between "empty" and "unspecified", older Trezors using a `required` field will fail to produce the "unspecified" result anyway)

andrewkozlik: I think I actually agree with you @matejcik about required fields. So changing the required fields to optional was not a step in the right direction and I am in favor of reverting it.

ibz:
When rebasing on `main`, here is how I solved conflicts in `ethereum/sign_tx.py` which happened due to introduction of the [`Approve` handling](https://github.com/trezor/trezor-firmware/pull/4958):

Basically on the `if SC_FUNC_SIG_APPROVE` branch I raise `DataError` and only handle the payment request on the `else` branch... @obrusvit 

```
   @@ core/src/apps/ethereum/sign_tx.py: async def confirm_tx_data(
    -     # Handle ERC-20, currently only 'transfer' function
    -     token, recipient, value = await _handle_erc20_transfer(msg, defs, address_bytes)
    - 
    -+    recipient_str = address_from_bytes(recipient, defs.network) if recipient else None
    -+    if payment_req_verifier is not None:
    -+        # If a payment_req_verifier is provided, then msg.payment_req must have been set.
    -+        assert msg.payment_req is not None
    -+        payment_req_verifier.add_output(value, recipient_str or "")
    -+        payment_req_verifier.verify()
    -+        recipient_str = msg.payment_req.recipient_name
    +     if func_sig == constants.SC_FUNC_SIG_APPROVE:
    +         assert token
    +         assert token_address
     +
    -     is_contract_interaction = token is None and data_total_len > 0
    ++        if payment_req_verifier is not None:
    ++            raise DataError("Payment Requests not supported for the APPROVE call")
    ++
    +         await require_confirm_approve(
    +             recipient,
    +             value,
    +@@ core/src/apps/ethereum/sign_tx.py: async def confirm_tx_data(
    +     else:
    +         assert func_sig == constants.SC_FUNC_SIG_TRANSFER or func_sig is None
    + 
    ++        recipient_str = address_from_bytes(recipient, defs.network) if recipient else None
    ++        if payment_req_verifier is not None:
    ++            # If a payment_req_verifier is provided, then msg.payment_req must have been set.
    ++            assert msg.payment_req is not None
    ++            payment_req_verifier.add_output(value, recipient_str or "")
    ++            payment_req_verifier.verify()
    ++            recipient_str = msg.payment_req.recipient_name
    ++
    +         is_contract_interaction = token is None and data_total_len > 0
      
    -     if is_contract_interaction:
    -         await require_confirm_other_data(msg.data_initial_chunk, data_total_len)
    +         if is_contract_interaction:
    +@@ core/src/apps/ethereum/sign_tx.py: async def confirm_tx_data(
      
    -     await require_confirm_tx(
    --        recipient,
    -+        recipient_str,
    -         value,
    -         msg.address_n,
    -         maximum_fee,
    +         assert value is not None
    +         await require_confirm_tx(
    +-            recipient,
    ++            recipient_str,
    +             value,
    +             msg.address_n,
    +             maximum_fee,
     
```

ibz (approved):

common/protob/messages.proto
```diff
@@ -118,6 +118,7 @@ enum MessageType {
     MessageType_GetECDHSessionKey = 61 [(bitcoin_only) = true, (wire_in) = true];
     MessageType_ECDHSessionKey = 62 [(bitcoin_only) = true, (wire_out) = true];
     // dropped: CosiCommit, CosiCommitment, CosiSign, CosiSignature
+    MessageType_PaymentNotification = 52 [(bitcoin_only) = true, (wire_in) = true];
```
martykan, line 121: Can we keep this at ID 37? This breaks backward compatiblity

Hannsek: cc @obrusvit @ibz @matejcik

andrewkozlik: TxAckPaymentRequest was renamed to PaymentRequest with the same ID. PaymentNotification is a new message.

martykan: Ok sorry, I looked wrong

PR #4962, opened 2025-04-28
feat(core): count per-line hits during profiling

romanz:
Unfortunately, it cannot be visualized by `coverage`, so we'll probably need another tool.

Also enabled coverage report for T3B1.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -502,6 +502,7 @@ upload_prodtest: ## upload prodtest using trezorctl
 
 coverage:  ## generate coverage report
 	./tools/coverage-report
+	./tools/coverage-annotate.py $(shell find . -name '.coverage*.json*') > htmlcov/hits.md
 
 unused:  ## find unused micropython code
 	vulture src src/_vulture_ignore.txt --exclude "messages.py,*/enums/*"
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -676,10 +676,6 @@ jobs:
       fail-fast: false
       matrix:
         model: [T2T1, T3B1, T3T1]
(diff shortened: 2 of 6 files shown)


romanz:
Added a basic annotation using a simple Markdown formatter (https://github.com/trezor/trezor-firmware/pull/4962/commits/e473a177b68574efbf3b72d29a86ef72530398f3):
image
Generated using `make -C core coverage_hits`.

Also enabled coverage collection for T3B1 model (https://github.com/trezor/trezor-firmware/pull/4962/commits/8a93cc77025be247328f6bc90d2893414fb5f421).

core/Makefile
```diff
@@ -503,6 +503,9 @@ upload_prodtest: ## upload prodtest using trezorctl
 coverage:  ## generate coverage report
 	./tools/coverage-report
 
+coverage_hits:  ## generate annotated source with per-line hits
+	./tools/coverage-annotate.py $(shell find . -name '.coverage*.json*') > htmlcov/hits.md
```
mmilata, line 507: I'd add this under the `coverage` target

mmilata (approved):
LGTM

I thought `pandoc -f markdown -t html hits.md` would produce syntax highlighted HTML but wasn't able to make it work.

core/Makefile
romanz: Fixed in https://github.com/trezor/trezor-firmware/pull/4962/commits/cceb66f45f93ed2ddb2548ced79a6b8b5c5704f0

romanz: Oops, thanks!
Should be fixed in https://github.com/trezor/trezor-firmware/pull/4962/commits/f1fc7a02c0f3570c7182f2dc8e15adbd1a5cecbf.

PR #4963, opened 2025-04-28
Add WireInterface to logging from `core/wire`

M1nd3r:
Add `wire_log` that includes `WireInterface` (eg. `WebUSB`, `BleInterface`) in output when logging.

### DIFF
--- a/core/src/trezor/wire/wire_log.py
+++ b/core/src/trezor/wire/wire_log.py
@@ -1,0 +1,61 @@
+import sys
+from typing import TYPE_CHECKING
+
+from trezor import log
+
+if TYPE_CHECKING:
+    from trezorio import WireInterface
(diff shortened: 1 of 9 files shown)


M1nd3r:
interface_logging

core/src/trezor/wire/wire_log.py
matejcik, line 1: make sure this file is excluded from non-debug builds

M1nd3r: Excluded here: https://github.com/trezor/trezor-firmware/pull/4963/commits/99304a133d737f0045b56ff9e02a10fc3c5dd427

mmilata (approved):
LGTM

romanz:
FTR, #5056 rewrites logging in Rust and changes timestamp precision to milliseconds:
image

PR #4971, opened 2025-04-29
Use better font on Delizia properties

ibz:
Cherry-picked this out of #4767 so we can release it sooner.


matejcik (approved):
ACK if the CI is green

Issue #4932, opened 2025-04-19
T1B1 ETH priority fees are displayed in Wei


Ondra-Zik-SL:
The priority fees for Ethereum should be displayed in `Gwei` instead of `Wei`, as the current format confuses users.

Another option would be to move the priority fees information to a submenu, as is the case for `Core` devices.

Example:  
https://www.reddit.com/r/TREZOR/comments/1k2vt7v/whats_is_it_after_last_trezor_suite_update_i_see/

Hannsek:
If the priority fees are very low <1e3 Wei, we should display it in Wei. cc @tomasklim

tomasklim:
I don't think so. People are used to gweis. Better to show 0.0001234 gwei instead of 1234000 wei

romanz:
For example, in the flow below we are using Gwei for `Maximum fee` but maybe Wei is more readable for the other (smaller) fee rates:

![Image](https://github.com/user-attachments/assets/9563082c-58cc-4d3e-ad6d-43f0e59d41f2)

PR #4981, opened 2025-04-30
fix(legacy): always use Gwei for ETH fee rates

romanz:
Similar to how it's done in core.

Following https://github.com/trezor/trezor-firmware/issues/4932#issuecomment-2820458343, we should use Gwei for "per gas" fee rates:

image

After this PR:

## Recent UI diff
https://data.trezor.io/dev/firmware/ui_report/16425051463/T1B1-en-legacy_device_test-index.html

### DIFF
--- a/legacy/firmware/ethereum.c
+++ b/legacy/firmware/ethereum.c
@@ -322,22 +322,18 @@ static void ethereumFormatAmount(const bignum256 *amnt,
   bignum256 bn1e9 = {0};
   bn_read_uint32(1000000000, &bn1e9);
 
-  bignum256 bn1e3 = {0};
-  bn_read_uint32(1000, &bn1e3);
-
   char suffix[50] = {' ', 0};
   int decimals = 18;
   if (token) {
     strlcpy(suffix + 1, token->symbol, sizeof(suffix) - 1);
     decimals = token->decimals;
+  } else if (use_gwei) {
+    // "per gas" fees should always use Gwei
+    strlcpy(suffix + 1, "Gwei", sizeof(suffix) - 1);
+    decimals = 9;
   } else if (bn_is_less(amnt, &bn1e9)) {
-    if (use_gwei && !bn_is_less(amnt, &bn1e3)) {
-      strlcpy(suffix + 1, "Gwei", sizeof(suffix) - 1);
-      decimals = 9;
-    } else {
-      strlcpy(suffix + 1, "Wei", sizeof(suffix) - 1);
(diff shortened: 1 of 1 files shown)


legacy/firmware/ethereum.c
```diff
@@ -446,7 +446,7 @@ static void layoutEthereumFee(const uint8_t *value, uint32_t value_len,
   bn_multiply(&val, &gas, &secp256k1.prime);
 
   ethereumFormatAmount(&gas, NULL, gas_value, sizeof(gas_value),
-                       /*use_gwei=*/true);
+                       /*use_gwei=*/false);
```
romanz, line 449: IIUC, here we are formatting the total fee amount, not "per gas" fee rate.

Hannsek:
We should squeeze this in the next release.

romanz:
Tested on a T1B1 device:

# Non-priority fee





# Priority fee

```
trezorctl eth sign-tx 0xD523794C879D9eC028960a231F866758e405bE34 100000000000000000 \
 -n "m/44'/60'/0'/0/0" --max-gas-fee 200 --max-priority-fee 300 \
 --nonce 0 -e 2 --gas-limit 100000
```

romanz:
Rebased over latest `main`.

romanz:
Rebasing over latest `main`.

legacy/firmware/ethereum.c
```diff
@@ -322,24 +322,24 @@ static void ethereumFormatAmount(const bignum256 *amnt,
...
       decimals = 9;
     } else {
-      strlcpy(suffix + 1, "Wei", sizeof(suffix) - 1);
-      decimals = 0;
+      if (bn_is_less(amnt, &bn1e9)) {
+        strlcpy(suffix + 1, "Wei", sizeof(suffix) - 1);
+        decimals = 0;
+      } else {
+        strlcpy(suffix + 1, chain_suffix, sizeof(suffix) - 1);
+      }
```
matejcik: can you flatten?
this reads confusing, and it seems it is in fact a single if/elif/else chain

romanz: b54a5b4578

matejcik: i meant even flatter, the `if (token)` seems like a sufficient level. or am i overlooking anything?

romanz: Agree - flatter is better :)
Updated in 7f174b4dc4.

PR #4986, opened 2025-04-30
core: add T3W1 Capabilities to Features

mmilata:
Thinking maybe there should be one for LED too?


matejcik (approved):

matejcik:
> Thinking maybe there should be one for LED too?

depends, is that something Suite needs to know? we don't have a feature for "touch" and "buttons" either, LED seems similar

PR #4988, opened 2025-05-02
fix(tools): use `tio --no-reconnect` instead of `--no-autoconnect`

romanz:
It was renamed in https://github.com/tio/tio/releases/tag/v3.0.

image


mmilata (approved):
good catch, thank you

PR #4993, opened 2025-05-02
fix(core): disable `check_free_heap` also in debug firmware

romanz:
Currently, it may misdetect on-heap buffers' data as valid heap pointers (resulting in `gc_mark_subtree` false-positives).


romanz:
Currently, this test fails T2B1 HW tests:
https://github.com/trezor/trezor-firmware/actions/runs/14848233040/job/41686733251#step:9:213

obrusvit (approved):

PR #5000, opened 2025-05-04
fix(python): fix `trezorctl firmware update` device restarts' handling

romanz:
Otherwise, it fails (on T3B1 and T3T1) with:
```
$ trezorctl -v firmware_update -s -f core/build/firmware/firmware.bin
...
[2025-05-04 12:29:05,881] trezorlib.client DEBUG: received message: Success
Success (11 bytes) {
    message: 'Rebooting',
}
Error: USB close failed: LIBUSB_ERROR_NO_DEVICE [-4]
```


python/src/trezorlib/transport/webusb.py
```diff
@@ -75,7 +75,7 @@ def close(self) -> None:
                 self.handle.releaseInterface(self.interface)
                 self.handle.close()
             except Exception as e:
-                raise TransportException(f"USB close failed: {e}") from e
+                raise CloseFailed(f"USB close failed: {e}") from e
```
matejcik, line 78: wondering if we even need to propagate the exception. can you see a reason for it?

romanz: Only to let the user know there was an USB error (in case the device stopped responding not because of the `RebootToBootloader` request)... so probably emitting a warning log will suffice.
WDYT?

matejcik: agreed :+1:

romanz: https://github.com/trezor/trezor-firmware/pull/5000/commits/2d354cf7b9c431f421e306400fcf9461e6396635

PR #5002, opened 2025-05-05
perf(core): use lookup tables for untranslated strings

romanz:
It seems that Rust generates non-optimal code the giant `match` statements in `translated_string.rs`.

This PR reduces flash usage by ~5kB on T3B1:

`Memory region         Used Size  Region Size  %age Used`
`           FLASH:     1378816 B      1664 KB     80.92%`  this PR @ ce14a5990c          
`           FLASH:     1383936 B      1664 KB     81.22%`  `main`  @ 53225569c8

### DIFF
--- a/core/embed/rust/src/translations/generated/translated_string.rs
+++ b/core/embed/rust/src/translations/generated/translated_string.rs
@@ -6,7 +6,7 @@
 #[cfg(feature = "micropython")]
 use crate::micropython::qstr::Qstr;
 
-#[derive(Copy, Clone, FromPrimitive, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


matejcik:
you can now move most of the impls for `TranslatedString` to `translations::translated_string`, and keep only the enum variants & the data map itself in `translations::generated::translated_string`

romanz:
Thanks! Fixed in https://github.com/trezor/trezor-firmware/pull/5002/commits/fde6f4b41de80814dd3bae4b1c8189878ca088da.

matejcik (approved):

PR #5005, opened 2025-05-05
refactor(core): remove firmware_calc_hash callback

cepetr:
This PR removes the kernel-to-app callback from the firmware hash calculation syscall.

The original `SYSCALL_FIRMWARE_CALC_HASH` with a callback has been replaced by a pair of syscalls:
- `SYSCALL_FIRMWARE_HASH_START` - initializes the firmware hash calculation
- `SYSCALL_FIRMWARE_HASH_CONTINUE` - calculates the hash of a 128 KB chunk

From the user’s perspective, the app’s behavior is unchanged.


core/embed/util/fwutils/fwutils.c
```diff
@@ -21,61 +21,88 @@
...
+    ctx->fw_offset += FW_HASHING_CHUNK_SIZE;
+    --n_chunks;
   }
 
-  if (blake2s_Final(&ctx, hash, hash_len) != 0) {
-    return secfalse;
+  if (ctx->fw_offset >= ctx->fw_size) {
+    int err = blake2s_Final(&ctx->blake, hash, hash_len);
+    if (err != 0) {
+      ctx->initialized = false;
```
TychoVrahe, line 100: should we maybe se initialized to false even on successful calculation of final hash? to indicate that we are finished

TychoVrahe (approved):
utACK, leaving one comment to consider

PR #5008, opened 2025-05-06
fix(core): fix tutorial-related translations

### LINKED ISSUE #3821: Additional symbols on Trezor display during setup in Czech firmware
In Czech firmware, during the setup - Know your Trezor/Poznej svůj Trezor on a step that says "Posunutí obrazovky, Levým tlačítkem se posunete nahoru" , there are some additional symbols on the left side of the screen around "Le" part of the word "Levým". This was replicated twice, bootloader wipe did not resolve this issue.

IMG_2651

romanz:
Fixes #3821.

### DIFF
--- a/core/.changelog.d/3821.fixed
+++ b/core/.changelog.d/3821.fixed
@@ -1,0 +1,1 @@
+Fixed tutorial-related translations.
--- a/core/embed/rust/src/ui/component/text/layout.rs
+++ b/core/embed/rust/src/ui/component/text/layout.rs
@@ -292,6 +292,8 @@ impl TextLayout {
                 sink.prev_page_ellipsis(*cursor, self);
                 cursor.x += self.style.prev_page_ellipsis_width();
             }
+            // Ignore leading '\r' character if the above ellipsis has been drawn
+            remaining_text = remaining_text.strip_prefix('\r').unwrap_or(remaining_text);
         }
 
         while !remaining_text.is_empty() {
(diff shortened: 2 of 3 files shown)


romanz:
It's strange that UI fixtures don't cover this issue :thinking:

prusnak:
can we add some test that checks that fail when a translated string contains `\r` character?

romanz:
IIUC, `\r` is currently handled here:
https://github.com/trezor/trezor-firmware/blob/50411367afaf4935aea00aa172aa7e1dde522edd/core/embed/rust/src/ui/component/text/layout.rs#L689-L693

It seems that the problem happens when the page break happens exactly before it:


It does seem to work as intended in other scenarios:
 

It was added in 2703d714c2dc1f3ba907d223825a58e2a914b173:
https://github.com/trezor/trezor-firmware/blob/2703d714c2dc1f3ba907d223825a58e2a914b173/core/embed/rust/src/ui/component/model_tt/text.rs#L578-L582

If we don't want to support it, would it be OK to replace `\n\r` by `\n`?

matejcik:
we're using `\r` to indicate "half line space", i'd prefer to get rid of it but it may not be viable

matejcik:
> It seems that the problem happens when the page break happens exactly before it:

ISTM a better fix would be to skip all whitespace after the prev-page-ellipsis, wdyt?

romanz:
Thanks! It sounds good, will modify the PR.

matejcik (approved):

bosomt:
**QA OK**

https://github.com/trezor/trezor-firmware/issues/3821#issuecomment-2984504527

PR #5014, opened 2025-05-06
style(core): simplify Rust translation conversions

romanz:

### DIFF
--- a/core/embed/rust/src/translations/translated_string.rs
+++ b/core/embed/rust/src/translations/translated_string.rs
@@ -4,7 +4,7 @@ use super::blob::Translations;
 pub use super::generated::translated_string::TranslatedString;
 
 impl TranslatedString {
-    pub(super) fn translate<'a>(self, source: Option<&'a Translations>) -> &'a str {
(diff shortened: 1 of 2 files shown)


core/embed/rust/src/translations/translated_string.rs
```diff
@@ -4,7 +4,7 @@ use super::blob::Translations;
 pub use super::generated::translated_string::TranslatedString;
 
 impl TranslatedString {
-    pub(super) fn translate<'a>(self, source: Option<&'a Translations>) -> &'a str {
+    fn translate<'a>(self, source: Option<&'a Translations>) -> &'a str {
```
matejcik, line 7: this should be a ` pub fn`, this is a perfectly safe API on top of _some arbitrary translations object_
(not sure why we made it `pub(super)` in the first place -- probably not to confuse callers about how to call it?)

romanz: 28dd3f8976

matejcik (approved):

Issue #1840, opened 2021-10-08
Limit total size of one protobuf message


matejcik:
Currently, `core` has a fixed 8kB buffer for receiving incoming messages, _and_ tries to dynamically allocate more if the incoming message is larger.

For the vast majority of workflows, 8kB is wastefully large. The only exceptions that we know of are:
- uploading a custom homescreen -- the whole homescreen (16 kB max) must fit into a single message
- downloading list of resident credentials -- all credentials (at most 100) must fit into a single message

We should resolve these separately: introduce streaming for homescreen #1120 and probably for the credential list too.

It seems that 4096 bytes per message should be Enough For Everyone (tm). We need to test if that is really the case. If yes, we should:
* decrease the buffer size
* disable the dynamic allocation feature
* clearly document the fixed message size limit

PR #5015, opened 2025-05-06
Homescreen streaming

### LINKED ISSUE #1120: Stream homescreen picture
The TOIF images might be large and not fit into the pre-allocated 8KB block (see https://github.com/trezor/trezor-firmware/issues/1114 and https://github.com/trezor/trezor-firmware/pull/1089). We should consider streaming the image.

mmilata:
Fixes #1120. T3W1 will have about 4x larger homescreen but we want to have small protobuf messages #1840.


common/protob/messages-management.proto
romanz, line 1: Maybe add a changelog entry (since the protocol is updated)?

core/src/apps/management/apply_settings.py
```diff
@@ -10,11 +10,32 @@
...
+
+    from apps.common import chunked
+
+    loader: ProgressLayout | None = None
+
+    def report(value: int) -> None:
+        nonlocal loader
+        if loader is None:
+            workflow.close_others()
+            loader = progress(TR.progress__please_wait)
```
romanz, line 31: Why not to create the progress layout before calling `get_all_chunks`, instead of launching it on-demand here?

core/src/apps/management/apply_settings.py
```diff
@@ -10,11 +10,32 @@
 if TYPE_CHECKING:
     from trezor.enums import SafetyCheckLevel
     from trezor.messages import ApplySettings, Success
+    from trezor.ui import ProgressLayout
 
 
 BRT_PROTECT_CALL = ButtonRequestType.ProtectCall  # CACHE
 
 
+async def _load_homescreen(length: int) -> bytes:
```
romanz, line 19: nit: it returns a `bytearray`, right?

romanz (approved):
Looks good, added a few questions/comments below:

core/src/apps/management/apply_settings.py
```diff
@@ -10,11 +10,32 @@
...
+    loader: ProgressLayout | None = None
+
+    def report(value: int) -> None:
+        nonlocal loader
+        if loader is None:
+            workflow.close_others()
+            loader = progress(TR.progress__please_wait)
+        loader.report(value)
+
+    buf = utils.empty_bytearray(length)
```
romanz, line 34: Should we limit `length` (to avoid OOMs)?

python/src/trezorlib/device.py
```diff
@@ -79,23 +79,34 @@ def apply_settings(
...
+            settings.homescreen_length = len(homescreen)
+
+    response = client.call(settings)
+    if settings.homescreen_length is not None:
+        response = _send_chunked_data(client, response, homescreen or b"")
+    else:
+        messages.Success.ensure_isinstance(response)
     client.refresh_features()
-    return _return_success(out)
+    return _return_success(response)
```
matejcik, line 94: instead something vaguely like this:
```
if homescreen is not None and client.version >= HOMESCREEN:
    settings.homescreen_length = len(homescreen)
    client.call(settings, expect=DataChunkRequest)
    _send_chunked_data(client, response, homescreen)
    return _return_success(messages.Success(""))
else:
   out = client.call(settings, expect=Success)
   return _return_success(out)
```
(IIUC the purpose to returning Success from `_send_chunked_data` was to propagate it here, correct? let's not do that)

core/src/apps/management/apply_settings.py
matejcik: well, it's safe to do it like this, the worst an attacker can do is get back a crash if they send an image size that's too much
but it would be nicer to pre-limit `length` to allowable homescreen lengths anyway

core/src/apps/management/apply_settings.py
mmilata: It's to handle `length == 0`, but I rewrote it to handle this case explicitly in fd1419a708062560e063c3d26ac3e20da93520ac

core/src/apps/management/apply_settings.py
mmilata: right, thanks, fd1419a708062560e063c3d26ac3e20da93520ac

python/src/trezorlib/device.py
mmilata: fd1419a708062560e063c3d26ac3e20da93520ac

core/src/apps/management/apply_settings.py
```diff
@@ -10,11 +10,32 @@
...
+    from trezor.ui.layouts.progress import progress
+
+    from apps.common import chunked
+
+    loader: ProgressLayout | None = None
+
+    def report(value: int) -> None:
+        nonlocal loader
+        if loader is None:
+            workflow.close_others()
```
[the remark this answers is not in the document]
mmilata: This is actually copypasted from `apps/management/change_language.py`. It doesn't seem to me that `ProgressLayout` does anything with tasks, however it calls `CURRENT_LAYOUT.stop()` which might be enough?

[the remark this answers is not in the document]
mmilata: should i drop it from change_language too?

python/src/trezorlib/device.py
```diff
@@ -81,7 +81,7 @@ def apply_settings(
 
     if homescreen is not None and client.version >= HOMESCREEN_STREAMING_MIN_VERSION:
         settings.homescreen_length = len(homescreen)
-        response = client.call(settings, expect=messages.DataChunkRequest)
+        response = client.call(settings)
```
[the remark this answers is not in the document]
mmilata: if `len(homescreen) == 0` then device responds with `Success`
it should but it doesn't :thinking:

core/src/apps/management/apply_settings.py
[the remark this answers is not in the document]
matejcik: sure

python/src/trezorlib/device.py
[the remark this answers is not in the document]
matejcik: > `if len(homescreen) == 0 then device responds with Success`
change the if to `if homescreen` :) the else branch will cover `homescreen == b""` case. 

> it should but it doesn't 🤔

:thinking: :thinking: :thinking: looks like a pyright bug on first glance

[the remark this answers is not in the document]
matejcik: > 🤔 🤔 🤔 looks like a pyright bug on first glance

yup, fixed in 1.1.380 so let's not worry about it now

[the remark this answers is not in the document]
mmilata: right, it uses the old API but works fine: daac678166d57bd524c7b73bacc6accd3b14e8ae

core/src/apps/management/apply_settings.py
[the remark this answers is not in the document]
mmilata: 99835e1aed554424e5d3e1a9dbc363bfe32355ff

matejcik (approved):

PR #5016, opened 2025-05-07
fix flash partitioning on NRF

TychoVrahe:
This PR fixes some inconsistencies in flash partitioning on NRF

along with that, support for old devkit is dropped.

Finally, small flash space is reserved for custom manufacturing data.


cepetr (approved):
utACK

PR #4984, opened 2025-04-30
power manager driver

kopecdav:
This PR introduces a significant upgrade to the T3W1 power management system.

The power_manager driver now provides comprehensive control over all power-related functions in battery-powered models, including:

- Complete management of the NPM1300 Power Management IC (PMIC)
- Integrated control of the STWLC38 wireless charging system
- Advanced battery management (charging/discharging cycles and state of charge calculation)
- power state monitoring and low-power mode control

The driver has been structured into several distinct modules for better organization and maintainability.

### Monitoring timer module (power_monitoring.c)

The monitoring timer is a core module that periodically runs on background and collects essential data from power components. Based on this data, it determines the current power manager state, updates the fuel gauge calculations, controls the charging system, and provides input to the internal power state machine for automated transitions.

### Internal power state machine module (power_states.c)

The power_manager driver uses its own internal state machine to manage several power states. This helps control smooth operation by preventing sudden load changes and protects discharged batteries and wireless charger from overloading.

The system handles transitions between power states in a semi-automatic way. Some transitions happen automatically when battery charge levels change, while others need the user or application to trigger them.

power_manager

### Power control module (power_control.c)

This module implements hibernate and suspend transitions, replacing the original powerctl module.

### Event loop integration (power_manager_poll.c)
Module with event-loop integration to provide access to higher-level applications.

## Power manager integration 

Power manager was integrated in both bootloader and firmware (kernel) while both components provides different function. 

### Bootloader integration

Power manager integration in bootloader:

- Control the device startup sequence (currently triggered by a short button press)
- Ensures all power conditions are met for safe power-on, protecting the battery from over-discharge and preventing wireless charger overload
- Initializes battery state of charge tracking by either:
  - Recovering data from backup RAM, or
  - Running a low-consumption battery sampling procedure for one second when no data is available
- Control the battery charging in OFF state.

The system passes both the power manager state and battery charge level to the firmware through backup RAM.

### Firmware integration

In the firmware (kernel), the power manager:

- Continuously monitors power-related variables
- Generates important events for user applications (USB/wireless charging connections, state of charge updates, etc.)
- Ensures smooth operation when running on battery power by:
  - Limiting display backlight during low battery conditions
  - Triggering device shutdown when the battery is depleted
- Control low-power modes

PR #5017, opened 2025-05-07
Secure monitor prerequisites

cepetr:
This PR includes some refactoring in preparation for introducing the secure monitor:

- new feature flags for the powerctl, display, and applet modules
- prevents incomplete builds of the MPU driver when some symbols are not properly defined in the linker scripts
- safer handling of the stack limit register during transitions to the next boot stage
- removed some unnecessary definitions and fixed incorrect usage of constants
- corrected the EXC_RETURN constant for all CPU modes
- improved the board_capabilities interface to allow relocating the module to the secure monitor


TychoVrahe (approved):
utACK. 

Note for @kopecdav : there is some 'powerctl' feature releated changes, so there will be conflicts with https://github.com/trezor/trezor-firmware/pull/4984, just to be aware of.

PR #5023, opened 2025-05-09
fix null ios handling in host control

TychoVrahe:
Since host_control can be called without communication interfaces, we need to handle this properly.


cepetr (approved):
utACK

PR #5025, opened 2025-05-09
fix nrf start/stop, service listener registration

TychoVrahe:
fixes issue with stoping and restarting communication with NRF - if there is an spi request signal on calling start, prepare the transaction. And don't prepare transaction in callback if communication is stopped.

Additionally, nrf_register_listener needs to be called after initialized flag is set.


cepetr (approved):
utACK

PR #4987, opened 2025-05-01
Support new translation blob format

romanz:
It supports multiple translations strings chunks and allows total container size to be above 64kB.

Screenshot from 2025-05-01 21-21-58

Previous format is also supported by firmware.

## Translation sizes comparison

Model | Language | Size [B] d579d87948 | Size [B] aafdfd20c8 | Delta [B] | Delta [%] | Capacity [B] | Utilization aafdfd20c8
-- | -- | -- | -- | -- | -- | -- | --
T2B1 | cs-CZ | 29258 | 29264 | 6 | 0.02% | 32768 | 89.31%
T2B1 | de-DE | 26920 | 26926 | 6 | 0.02% | 32768 | 82.17%
T2B1 | es-ES | 28568 | 28670 | 102 | 0.36% | 32768 | 87.49%
T2B1 | fr-FR | 30342 | 30348 | 6 | 0.02% | 32768 | 92.61%
T2B1 | it-IT | 25218 | 25224 | 6 | 0.02% | 32768 | 76.98%
T2B1 | pt-BR | 28092 | 28098 | 6 | 0.02% | 32768 | 85.75%
T2B1 | tr-TR | 24798 | 24804 | 6 | 0.02% | 32768 | 75.70%
T2T1 | cs-CZ | 39394 | 39272 | -122 | -0.31% | 49152 | 79.90%
T2T1 | de-DE | 29344 | 29350 | 6 | 0.02% | 49152 | 59.71%
T2T1 | es-ES | 33524 | 33530 | 6 | 0.02% | 49152 | 68.22%
T2T1 | fr-FR | 41124 | 41130 | 6 | 0.01% | 49152 | 83.68%
T2T1 | it-IT | 27274 | 27280 | 6 | 0.02% | 49152 | 55.50%
T2T1 | pt-BR | 34292 | 34298 | 6 | 0.02% | 49152 | 69.78%
T2T1 | tr-TR | 28680 | 28686 | 6 | 0.02% | 49152 | 58.36%
T3B1 | cs-CZ | 29258 | 29264 | 6 | 0.02% | 65536 | 44.65%
T3B1 | de-DE | 26920 | 26926 | 6 | 0.02% | 65536 | 41.09%
T3B1 | es-ES | 28664 | 28670 | 6 | 0.02% | 65536 | 43.75%
T3B1 | fr-FR | 30342 | 30348 | 6 | 0.02% | 65536 | 46.31%
[...]

PR #5026, opened 2025-05-09
perf(core): use a single text blob for unstranslated strings

romanz:
Flash size usage on `TREZOR_MODEL=T3T1 PYOPT=0` decreases by ~8k bytes:

|Commit|Flash usage|
|-|-|
| 8992914041 (this PR) | `1622528 B = 95.22%` |
| 1ec290d540 (main)    | `1630720 B = 95.70%` |

Otherwise, [every `&str` takes 32-bit pointer & 32-bit length](https://doc.rust-lang.org/std/primitive.str.html#representation), resulting in ~8kB = (8 bytes x ~1k strings).

### DIFF
--- a/core/embed/rust/src/translations/generated/translated_string.rs
+++ b/core/embed/rust/src/translations/generated/translated_string.rs
@@ -6,6 +6,11 @@
 #[cfg(feature = "micropython")]
 use crate::micropython::qstr::Qstr;
 
+pub struct StringsBlob {
+    pub text: &'static str,
+    pub offsets: &'static [(TranslatedString, u16)],
+}
+
 #[derive(Copy, Clone, FromPrimitive, PartialEq, Eq, PartialOrd, Ord)]
 #[cfg_attr(feature = "debug", derive(ufmt::derive::uDebug))]
 #[cfg_attr(test, derive(Debug))]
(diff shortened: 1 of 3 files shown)


matejcik:
at this point you may want to build the blob in the same way that the translated blobs are doing it, and include it via `include_bytes!`

fivemtebex:
Great implementation of this feature. I particularly like the approach to error handling.

fivemserverhostings:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

fivemeyestore:
Excellent contribution to open source! The community will benefit greatly from this.

fivemhostingserver:
The performance optimizations here are quite clever. Nice engineering work!

scriptforfivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

romanz:
It will probably conflict with https://github.com/trezor/trezor-firmware/pull/4987's changes...

romanz:
> you may want to build the blob in the same way that the translated blobs are doing it

I am not sure that we can do it using a single blob, since we want to conditionally exclude altcoin- and debug-related strings...

romanz:
Rebased to resolve a conflict.

romanz:
> at this point you may want to build the blob in the same way that the translated blobs are doing it, and include it via `include_bytes!`

In https://github.com/trezor/trezor-firmware/commit/23e87aada0cd8d30b0cc4ebfcb78f94dbfce8cb2 I have used `TranslatedStrings.from_items(...)` in order to concatenate the strings and compute the offsets. 
WDYT?

romanz:
Rebasing to resolve a conflict with #4987.

romanz:
Rebased to resolve a conflict with `main`.

romanz:
Following https://github.com/trezor/trezor-firmware/pull/5026#issuecomment-2893934233, I am not sure that reusing [translation blob lookup flow](https://github.com/trezor/trezor-firmware/blob/8992914041ac8d7a4fa677a4d6872c8b6e5bdd72/core/embed/rust/src/translations/blob.rs#L152-L160) will make this PR simpler, since we'll need to re-implement "untranslated" blob generator from `core/translations/en.json`, taking in account `TREZOR_MODEL`, `PYOPT` and `BITCOIN_ONLY` to collect the correct strings (which is currently done by `translated_string.rs.mako` templating).

Current implementation generates 3 "blobs" per model during `make templates`:
- universal strings
- debug strings
- BTC-only strings

https://github.com/trezor/trezor-firmware/blob/8992914041ac8d7a4fa677a4d6872c8b6e5bdd72/core/embed/rust/src/translations/generated/translated_string.rs

During build time, the correct blobs are combined and searched (according to the enabled features).

WDYT?

romanz:
> you may want to build the blob in the same way that the translated blobs are doing it

I found a way to use `TranslationStringsChunk` for storing English strings (saving ~10.5 kB) :point_right:  https://github.com/trezor/trezor-firmware/pull/6313

PR #5027, opened 2025-05-11
fix(core): prevent homescreen flickering on Cancel message

romanz:

### DIFF
--- a/core/src/apps/base.py
+++ b/core/src/apps/base.py
@@ -1,5 +1,6 @@
 from typing import TYPE_CHECKING
 
+import storage.cache as storage_cache
 import storage.device as storage_device
 from storage.cache_common import APP_COMMON_BUSY_DEADLINE_MS, APP_COMMON_SEED
 from trezor import TR, config, utils, wire, workflow
@@ -238,7 +239,16 @@ async def handle_GetFeatures(msg: GetFeatures) -> Features:
 
 
 async def handle_Cancel(msg: Cancel) -> Success:
-    workflow.close_others()
+    close_others = True
+    if storage_cache.homescreen_shown is storage_cache.HOMESCREEN_ON:
+        # Homescreen is shown and only this task is running
+        if len(workflow.tasks) == 1 and next(iter(workflow.tasks)).is_running():
+            # Prevent homescreen flickering on Cancel
+            close_others = False
+
+    if close_others:
+        workflow.close_others()
+
     raise wire.ActionCancelled
 
 
--- a/core/.changelog.d/5006.fixed
+++ b/core/.changelog.d/5006.fixed
@@ -1,0 +1,1 @@
+[T2T1,T3T1] Prevent homescreen flickering on `Cancel` message.


core/src/apps/base.py
```diff
@@ -238,7 +239,16 @@ async def handle_GetFeatures(msg: GetFeatures) -> Features:
...
-    workflow.close_others()
+    close_others = True
+    if storage_cache.homescreen_shown is storage_cache.HOMESCREEN_ON:
+        # Homescreen is shown and only this task is running
+        if len(workflow.tasks) == 1 and next(iter(workflow.tasks)).is_running():
+            # Prevent homescreen flickering on Cancel
+            close_others = False
+
+    if close_others:
+        workflow.close_others()
```
matejcik, line 250: how about instead you add an argument `workflow.close_others(close_default: bool = True)` and use `close_default=False` here

that way this weirdly specific logic is correctly hidden in the `workflow` module

romanz: Thanks! Fixed in 4ee8985408.

core/src/trezor/workflow.py
```diff
@@ -136,14 +136,14 @@ def kill_default() -> None:
         default_task.close()
 
 
-def close_others() -> None:
+def close_others(close_default: bool = True) -> None:
```
matejcik, line 139: please add a description of the option to the docstring

core/src/trezor/workflow.py
```diff
@@ -136,13 +136,21 @@ def kill_default() -> None:
...
     If this is called from outside a registered workflow, it is equivalent to "close
     all tasks". In that case, the default task will be restarted afterwards.
+
+    Set `close_homescreen` to `False` to keep the default task in order to avoid
+    homescreen flickering.
     """
+    close_default = True
+    if not close_homescreen:
+        # keep only homescreen-unrelated default task
+        close_default = storage_cache.homescreen_shown is None
```
matejcik, line 152: nooo! workflow has no business deciding what is "homescreen".

in fact in an ideal world `homescreen_shown` would go away completely, that's the hack here, let's not couple this hack into more logic from which we'll have to decouple it again in a year.

`close_default` is the correct argument name, asking the `workflow` module to either close (or not) the _default task_, i.e., the one that is running _in the background_
`Cancel` is supposed to close the _running workflow_. the default task is not "the running workflow", hence won't be closed by `Cancel`. this has the nice side-effect that Cancel doesn't ever stop the homescreen.

an interesting question is whether Cancel is supposed to trigger exit from the recovery homescreen --- but if yes, it will _not_ be via the `workflow.close_others()` call, but via some other more explicit mechanism.
(one salient reason: `workflow.close_others()` is synchronous but exiting the recovery homescreen will need to trigger the "are you sure" dialog)

romanz: Sounds good, thanks for the detailed explanation!
Will drop ff73235ac2299d56dc1b2ca4fa0c85e00adb16e4.

romanz: Also, changed one of the tests not to use `Cancel`.

romanz:
Unfortunately, the above approach doesn't clear the locking animation if `Cancel` message is sent while the device is being locked:
[Screencast from 2025-05-25 13-09-46.webm](https://github.com/user-attachments/assets/b9c7b0ff-cab0-4699-8f6f-556c0480ff9f)

romanz:
OTOH, maybe adding a current layout polling API would be a better approach.

PR #5032, opened 2025-05-12
refactor(storage): simplify storage api

cepetr:
**It’s likely that this PR will never be merged. Please disregard it for now.**

This PR reduces the number of arguments in the storage module’s API functions.

This refactoring is a prerequisite for the Secure Monitor API, where the number of arguments is limited (they can only be passed on the stack).

In fact, only two functions really required this refactoring, but for consistency, all places where a pin is passed to the storage module have been updated.

### DIFF
--- a/core/embed/rust/src/trezorhal/storage.rs
+++ b/core/embed/rust/src/trezorhal/storage.rs
@@ -135,7 +135,11 @@ pub fn lock() {
 /// Returns true if the PIN + salt combination is correct.
 pub fn unlock(pin: &str, salt: Option<&ExternalSalt>) -> bool {
     let salt = salt.map(|s| s.as_ptr()).unwrap_or(ptr::null());
-    ffi::sectrue == unsafe { ffi::storage_unlock(pin.as_ptr() as *const _, pin.len(), salt) }
+    let pin = ffi::storage_pin_t {
+        text: pin.as_ptr(),
+        len: pin.len(),
+    };
+    ffi::sectrue == unsafe { ffi::storage_unlock(&pin, salt) }
 }
 
 /// Change PIN and/or external salt.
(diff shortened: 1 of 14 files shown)


core/embed/rust/src/trezorhal/storage.rs
```diff
@@ -135,7 +135,11 @@ pub fn lock() {
 /// Returns true if the PIN + salt combination is correct.
 pub fn unlock(pin: &str, salt: Option<&ExternalSalt>) -> bool {
     let salt = salt.map(|s| s.as_ptr()).unwrap_or(ptr::null());
-    ffi::sectrue == unsafe { ffi::storage_unlock(pin.as_ptr() as *const _, pin.len(), salt) }
+    let pin = ffi::storage_pin_t {
```
matejcik, line 138:
```rust
impl From<&str> for ffi::storage_pin_t {
    fn from(s: &str) -> Self {
        Self {
            text: s.as_ptr(),
            len: s.len(),
        }
    }
}
```

matejcik:
do we maybe want to make this a utility type available to the whole project?

core/embed/rust/src/trezorhal/storage.rs
cepetr: yes, thanks, see 0c2c101ba7abb0c37cb107dd97131a2677fe0d88

cepetr:
> do we maybe want to make this a utility type available to the whole project?

Yes, you are right, there are a few functions where we can utilize such a type (like `system_exit_xxx`). Howver, I would rather not to complicate this commit with it.

matejcik:
> Howver, I would rather not to complicate this commit with it.

i don't mean to want to apply this change everywhere now.
just move the type to a more general space, so that we don't have to touch storage again when we do want to use it elsewhere?

cepetr:
Closing this PR as the changes are not necessary and don't need to be merged.

PR #5033, opened 2025-05-12
WIP: fiix eckhart button content alignment

bieleluk:
This PR makes the button content alignment text-specific. The all-case alignment of  "Ay" string caused a significant offset, especially for the big keypad number buttons.

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/component/button.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/component/button.rs
@@ -244,17 +244,21 @@ impl Button {
     pub fn content_height(&self) -> i16 {
         match &self.content {
             ButtonContent::Empty => 0,
-            ButtonContent::Text(_) => self.style().font.allcase_text_height(),
+            ButtonContent::Text(text) => text.map(|s| self.style().font.visible_text_height(s)),
             ButtonContent::Icon(icon) => icon.toif.height(),
             ButtonContent::IconAndText(child) => {
-                let text_height = self.style().font.allcase_text_height();
+                let text_height = child.text.map(|s| self.style().font.visible_text_height(s));
                 let icon_height = child.icon.toif.height();
                 text_height.max(icon_height)
             }
-            ButtonContent::TextAndSubtext { subtext_style, .. } => {
-                self.style().font.allcase_text_height()
+            ButtonContent::TextAndSubtext {
+                text,
+                subtext,
+                subtext_style,
+            } => {
+                text.map(|s| self.style().font.visible_text_height(s))
                     + Self::LINE_SPACING
-                    + subtext_style.text_font.allcase_text_height()
+                    + subtext.map(|s| subtext_style.text_font.visible_text_height(s))
             }
             #[cfg(feature = "micropython")]
             ButtonContent::HomeBar(_) => theme::ACTION_BAR_HEIGHT,
@@ -391,8 +395,10 @@ impl Button {
                 subtext,
                 subtext_style,
             } => {
-                let text_y_offset =
-                    Offset::y(self.content_height() / 2 - stylesheet.font.allcase_text_height());
+                let text_y_offset = Offset::y(
+                    self.content_height() / 2
+                        - text.map(|s| stylesheet.font.visible_text_height(s)),
+                );
                 let subtext_y_offset = Offset::y(self.content_height() / 2);
                 let start_of_baseline = match self.text_align {
                     Alignment::Start => self.area.left_center() + self.content_offset,


tests/click_tests/test_passphrase_bde.py
```diff
@@ -114,6 +116,22 @@ def go_to_category(debug: "DebugLink", category: PassphraseCategory) -> None:
     else:
         for _ in range(current_index - target_index):
             debug.swipe_right()
+    if check:
+        # map Rust debug string to Python enum
+        layout_map = {
+            "LettersLower": PassphraseCategory.LOWERCASE,
```
ibz: I know this is unrelated to this PR, but why not rename either one or the other enum members to have them match?

ibz (approved):

tests/click_tests/test_passphrase_bde.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5033/commits/3009285f2daa487d1bb02869092003a5a766919d

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -244,17 +244,19 @@ impl Button {
     pub fn content_height(&self) -> i16 {
         match &self.content {
             ButtonContent::Empty => 0,
-            ButtonContent::Text(_) => self.style().font.allcase_text_height(),
+            ButtonContent::Text(_) => self.style().font.height + 2 * self.style().font.baseline,
```
obrusvit: Two points regarding `content_height`
1. Why `+ 2 * baseline`. I don't think `baseline` is used correctly here. It seems that you want `self.style().font.max_height + SOME_MARGIN` but what is the margin, I'm not sure. Maybe we can add the same margin to all `ButtonContent`, not sure.
2. In case you insist on `2* baseline` then please put `elf.style().font.height + 2 * self.style().font.baseline` into a variable above the `match` case so that it's not repeated 3 times in the function

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -391,27 +395,34 @@ impl Button {
...
-                    Alignment::End => self.area.right_center() - self.content_offset,
+                let base = match self.text_align {
+                    Alignment::Start => self.area.left_center().ofs(self.content_offset),
+                    Alignment::Center => self.area.center().ofs(self.content_offset),
+                    Alignment::End => self.area.right_center().ofs(self.content_offset.neg()),
                 };
-                let text_baseline = start_of_baseline - text_y_offset;
-                let subtext_baseline = start_of_baseline + subtext_y_offset;
+
+                let text_render_origin = base.ofs(
```
obrusvit: `text_render_origin` seems quite complicated, can you divide it to multiple steps?

Also, am not sure why `self.content_height() / 2` is in there. Isn't it different from `ButtonContent::Text` variant?

tests/click_tests/test_passphrase_bde.py
```diff
@@ -124,20 +124,15 @@ def go_to_category(
...
-            "Special": PassphraseCategory.SPECIAL,
-        }
-
         layout = debug.read_layout().find_unique_value_by_key(
             "active_layout", default="", only_type=str
         )
         # do the check if Rust debug string exists
         if layout:
-            assert layout_map.get(layout) == category
+            try:
```
obrusvit: I think you want to catch `KeyError`, not `ValueError`. But generally I think this is better way:

```python
if layout:
    assert layout in PassphraseCategory, f"Unknown layout name from debug: {layout}"
    assert PassphraseCategory[layout] == category, f"Layout mismatch: expected {category}, got {PassphraseCategory[layout]}"
```

obrusvit (changes requested):
Apart from the comments below, I tested on the device and noticed one problem.

If you place your finger on the input area and proceed with moving your finger left or right, then it's registered as swipe. The following occurs:
- after lifting your finger, the input area stays visible
- after touching the display again, the input area disappears but the swipe is "finished" and you move to another keyboard layout
That's minor incosistency, I'd propose not registering swipe when the input is showing because a user might want to move the finger a little bit left/right to read the whole passphrase.

Hannsek:
Also, what I would like to see is when I press and hold the upper part because I want to check what is written there I cannot see it fully because of my finger. Would it be possible the passphrase (or PIN) stays displayed when my finger swipes out of the upper part (down)?

core/embed/rust/src/ui/layout_eckhart/component/button.rs
ibz: > text_render_origin seems quite complicated, can you divide it to multiple steps?

Actually this whole part gets changed anyway in my PR that comes on top of this, so I'd say better don't change this now, but I can fix it in my PR if needed.

bieleluk:
> Apart from the comments below, I tested on the device and noticed one problem.
> 
> If you place your finger on the input area and proceed with moving your finger left or right, then it's registered as swipe. The following occurs:
> 
>     * after lifting your finger, the input area stays visible
> 
>     * after touching the display again, the input area disappears but the swipe is "finished" and you move to another keyboard layout
>       That's minor incosistency, I'd propose not registering swipe when the input is showing because a user might want to move the finger a little bit left/right to read the whole passphrase.

The bug is solved by processing the input touch event first: 
https://github.com/trezor/trezor-firmware/pull/5033/commits/ec55bc7c1185316515fff7b5f96917556a449d16

tests/click_tests/test_passphrase_bde.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5033/commits/d07fd3d82e8356045142e69d3e39c976cc73b4c7

core/embed/rust/src/ui/layout_eckhart/component/button.rs
bieleluk: This issue is quite unusual because it involves satisfying two conflicting requirements:
- The alignment must remain constant to ensure the keyboard keys stay properly aligned.
- At the same time, large numerical constants and lowercase characters must be precisely centered, as any misalignment becomes very noticeable when the keys are pressed.

The constants I used here are a bit ad-hoc, since using `allcase_text_height` and `text_height` didn’t properly center the numerical buttons as intended.

core/embed/rust/src/ui/layout_eckhart/component/button.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5033/commits/3bf9781c1ed497ed6012aef9a7b2dea093ec8117
Simplified as much as possible

core/embed/rust/src/ui/layout_eckhart/component/button.rs
bieleluk: https://github.com/trezor/trezor-firmware/commit/3bf9781c1ed497ed6012aef9a7b2dea093ec8117
Simplified as much as possible

core/embed/rust/src/ui/layout_eckhart/component/button.rs
obrusvit: much better! thanks

obrusvit (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.6.0 (422e17f938f060f55ee711598b7e2d3487907f78)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.6.0 Chrome/136.0.7103.49 Electron/36.2.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3W1 2.9.0 regular (revision 6a2933988e23eb786ef1d4351afdaffec212a8bb)
 - Transport: NodeUsbTransport, BluetoothTransport

PR #5034, opened 2025-05-13
add bin to hex conversion to Makefile

TychoVrahe:
for convenience

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -482,6 +482,12 @@ combine_fw: ## combine boardloader + bootloader + firmware into one combined ima
 		-b FIRMWARE $(FIRMWARE_BUILD_DIR)/firmware.bin
 endif
 
+combine_to_hex: ## convert combined image to hex format
+	arm-none-eabi-objcopy \
+		-I binary -O ihex \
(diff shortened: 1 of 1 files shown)


core/Makefile
```diff
@@ -482,6 +482,12 @@ combine_fw: ## combine boardloader + bootloader + firmware into one combined ima
 		-b FIRMWARE $(FIRMWARE_BUILD_DIR)/firmware.bin
 endif
 
+combine_to_hex: ## convert combined image to hex format
+	arm-none-eabi-objcopy \
+      -I binary -O ihex \
+      --change-section-address .data=$(BOARDLOADER_START) \
+      $(PRODTEST_BUILD_DIR)/combined.bin $(PRODTEST_BUILD_DIR)/combined.hex
```
prusnak, line 489: nit: these 3 lines should be indented with 2 tabs and not spaces

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/f1a650654df485d80b4725bf6ceaf375c3373792

PR #5037, opened 2025-05-13
fix(vendor): use `MP_ERROR_TEXT` for `mp_map_presize` error message

romanz:
Requires https://github.com/trezor/micropython/pull/25.


mmilata:
this is already fixed, right?

PR #5038, opened 2025-05-14
fix booloader reboot

TychoVrahe:
Menu item reboot was not working when the menu was invoked from waiting for host. (more generally the waiting for host wasn't exited on results other then OK/CANCELLED from menu).

Size of buffer for layouts was increased to 2kB, turns out 1kB was not enough for eckhart UI.


cepetr (approved):
utACK

PR #5041, opened 2025-05-14
test(core): explicitly confirm backup success

### LINKED ISSUE #5039: `click_tests/test_repeated_backup.py::test_repeated_backup` is flaky when invoked via `make test_emu_click_ui_multicore`
Following https://github.com/trezor/trezor-firmware/pull/5028, there were several CI failures:

https://github.com/trezor/trezor-firmware/actions/runs/15014975444/job/42191021080?pr=4882
```
=========================== short test summary info ============================
FAILED ../tests/click_tests/test_repeated_backup.py::test_repeated_backup - trezorlib.exceptions.TrezorFailure: ActionCancelled: Cancelled
====== 1 failed, 51 passed, 17 skipped, 1 ui_failed in 104.57s (0:01:44) =======
```
https://github.com/trezor/trezor-firmware/actions/runs/15012996324/job/42185204774
```
=========================== short test summary info ============================
FAILED ../tests/click_tests/test_repeated_backup.py::test_repeated_backup - trezorlib.exceptions.TrezorFailure: ActionCancelled: Cancelled
====== 1 failed, 48 passed, 20 skipped, 1 ui_failed in 136.71s (0:02:16) =======
```
https://github.com/trezor/trezor-firmware/actions/runs/15016209550/job/42194974663
```
=========================== short test summary info ============================
FAILED ../tests/click_tests/test_repeated_backup.py::test_repeated_backup - trezorlib.exceptions.TrezorFailure: ActionCancelled: Cancelled
====== 1 failed, 56 passed, 12 skipped, 1 ui_failed in 120.98s (0:02:00) =======
```

romanz:
Fixes #5039.


tests/click_tests/test_repeated_backup.py
```diff
@@ -151,6 +151,9 @@ def test_repeated_backup(
 
         second_backup_2_of_3.append(" ".join(words))
 
+    # confirm backup success
+    reset.confirm_read(debug)
```
matejcik, line 155: my vague recollection says this will break on different layout models

romanz:
Fixed UI fixtures:
Screenshot from 2025-05-14 14-14-19

tests/click_tests/test_repeated_backup.py
romanz: 68f7c1138e seems to be passing... :thinking:

matejcik:
the fix looks good but it seems to uncover several issues.

(a) why is the test flaky in the first place? what instability was observed? the previous code _seems_ perfectly fine

more importantly (b) **why** does Trezor **not jump to the prompt** by itself without the message received from host

matejcik:
wrt (b) we should have two tests here: one that runs the backup process fully on-device, without host interaction beyond the initial Recovery device, and the other should be the current one

matejcik:
> ~wrt (b) we should have two tests here: one that runs the backup process fully on-device, without host interaction beyond the initial Recovery device, and the other should be the current one~

we already do! it's part of the same flow

in fact this click test is only doing the "do the backup from device" part, and there's a corresponding device-test that is doing "trigger the backup from host" part.
super confusing. wondering if we can do anything about it. we can document the testcase .... aacctually you know what

while you're touching this, please *rename the testcase* to something like "test repated backup via device", and the corresponding one "via host" 
or some other name if you can think of anything nicer

romanz:
> (a) why is the test flaky in the first place? what instability was observed? 

Added a debug log diff in https://github.com/trezor/trezor-firmware/issues/5039#issuecomment-2880172471.

romanz:
Rebased over https://github.com/trezor/trezor-firmware/pull/5042.

matejcik:
before merging, please do this ^^^

> please _rename the testcase_ to something like "test repated backup via device", and the corresponding one "via host"
> or some other name if you can think of anything nicer

PR #5049, opened 2025-05-16
External definitions modifications

matejcik:
* consolidate documentation for Ethereum and Solana definitions
* update URL scheme in docs and in trezorlib (new scheme is already live on data trezor io)
* improve behavior of `trezorctl solana sign-tx` wrt definitions

minor incompatible changes:
* methods on `DefinitionsSource` were renamed
* `ethereum.network_from_address_n` was dropped


romanz (approved):
LGTM

PR #5051, opened 2025-05-16
prodtest emulator

### LINKED ISSUE #4306: Implement `cpuid_read` in trezorhal
The production test command CPUID READ currently calls STM32 HAL directly within the `cpuid_read()` function. Moving this code to the trezorhal layer would simplify the structure and eliminate the last dependency on the STM32U5 macro in platform-independent code.

TychoVrahe:
This PR introduces basic support for prodtest emulation, closely based on bootloader emulator.

Some smaller refactoring was done along the way, mainly using portable function mini_snprintf instead of itoa, extracting cpuid reading into a module and enable emulation of this function, adding VCP support for USB emulator, extracting some common unix files from sconscrips.

only some functionality is supported in prodtest emulator, based on where we actually have unix implementation.

Mainly this enable emulated development of tropic integration.

There are still some caveats:
- emulator will look frozen, which will be solved by using sysevent polling in #5021 
- CTRL+C doesn't work for some reason
- some commands, like bootloader_version, end in segfault

I believe we can solve those later when needed.

closes #4306


cepetr (approved):
LGTM

PR #5053, opened 2025-05-18
test(core): deflake passphrase-related tests

romanz:
Currently, `PassphraseKeyboard` is shown after `request_passphrase_on_host()` layout, so in in order to avoid a race condition[^1][^2], an explicit synchronization was added.

[^1]: https://github.com/trezor/trezor-firmware/actions/runs/14999904818/job/42144246518?pr=5037
[^2]: https://github.com/trezor/trezor-firmware/actions/runs/15067093253/job/42354594590?pr=5041

### DIFF
--- a/tests/click_tests/test_autolock.py
+++ b/tests/click_tests/test_autolock.py
@@ -218,7 +218,7 @@ def test_autolock_passphrase_keyboard(device_handler: "BackgroundDeviceHandler")
     # get address
     device_handler.run(common.get_test_address)  # type: ignore
 
-    assert "PassphraseKeyboard" in debug.read_layout().all_components()
+    debug.synchronize_at("PassphraseKeyboard")
 
     if debug.layout_type is LayoutType.Caesar:
         # Going into the selected character category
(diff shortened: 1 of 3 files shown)


matejcik:
that's exactly what `wait_for_layout_change` was made for, i don't think you need to mess with passphrase callbacks...?

romanz:
I think that `Debuglink.wait_for_layout_change` does work, but IIUC it depends on the device to show `"Please enter your passphrase."` screen via:
https://github.com/trezor/trezor-firmware/blob/e98ea8323052366d37a88284c1a655e5808276b3/core/src/apps/common/passphrase.py#L38
If (due to scheduling issues) this screen is not shown yet when `wait_for_layout_change` is starting, could it detect the change from the homescreen to `request_passphrase_on_host` layout (causing the [test assertions to fail](https://github.com/trezor/trezor-firmware/pull/5053#issue-3071496805))?

romanz:
> very likely, the more appropriate solution is replacing the assertion with debug.synchronize_at("PassphraseKeyboard")

Sounds good, thanks!
Updated the PR to use `synchronize_at` @ https://github.com/trezor/trezor-firmware/commit/73ddaafdea850d169674eed420d74c48cc583b2e.

matejcik (approved):

PR #5055, opened 2025-05-18
Eckhart UI: remove label shadow on default homescreen pattern

obrusvit:
This PR removes the shadow effect from the device name on the homescreen in Eckhart UI. The reason is that it's not needed on the default pattern. However, it will be used when a user sets some custom image, the reason is that it makes the text more readable on light/dark backgrounds.

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
@@ -61,6 +61,7 @@ impl Homescreen {
         notification: Option<(TString<'static>, u8)>,
     ) -> Result<Self, Error> {
         let image = get_homescreen_image();
+        let default_hs = image.is_none();
 
         // Notification
         // TODO: better notification handling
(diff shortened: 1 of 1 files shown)


core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
```diff
@@ -241,9 +243,10 @@ impl HomeLabel {
         theme::BLACK,
     );
 
-    fn new(label: TString<'static>) -> Self {
+    fn new(label: TString<'static>, default_hs: bool) -> Self {
```
bieleluk, line 246: NIT: The parameter name `default_hs` might be a bit difficult to understand. A flipped-logic parameter, i.e. `shadow` would be a bit clearer.

bieleluk (approved):
Approved with one NIT suggestion.
After updating the fixtures, the PR can be merged.

core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
obrusvit: Done on rebase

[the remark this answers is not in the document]
obrusvit: Renamed to more descriptive `shadow`

PR #5060, opened 2025-05-19
adjust bootloader (un)locking for tropic-enabled models

TychoVrahe:
This PR adjust bootloader locking/unlocking mechanism to respect models with tropic.

When unlocking bootloader, secrets for both optiga and tropic should be erased. Therefore, for purpose of bootloader locked status detection, any secret that is not erased means that bootloader is locked and custom firmware should not be allowed to be installed and should not run.

Instead of relying on `USE_OPTIGA` for presence of bootloader lock functionality, new per-model compile flag is introduced: `LOCKABLE_BOOTLOADER`


cepetr (approved):
utACK

PR #5065, opened 2025-05-19
refactor(core): return uPy Obj from DeviceMenu

obrusvit:
[no changelog]


core/src/apps/homescreen/device_menu.py
```diff
@@ -72,21 +73,21 @@ async def handle_device_menu() -> None:
         ),
         None,
     )
-    if menu_result == "DevicePair":
+    if menu_result == DeviceMenuResult.DevicePair:
```
romanz, line 76: I think it's OK to use `is` here (since we are working with singletons, right?)

core/embed/rust/src/ui/layout/device_menu_result.rs
```diff
@@ -0,0 +1,38 @@
+use crate::micropython::{
+    macros::{obj_dict, obj_map, obj_type},
+    qstr::Qstr,
+    simple_type::SimpleTypeObj,
+    typ::Type,
+};
+
+static BACKUP_FAILED_TYPE: Type = obj_type! { name: Qstr::MP_QSTR_BackupFailed, };
```
romanz, line 8: Not sure if we need the result objects to have distinct types... can we use a single type for all of them?

core/src/apps/homescreen/device_menu.py
obrusvit: https://github.com/trezor/trezor-firmware/pull/5065/commits/385b4556fc8da04cc2ecda3ec20248f0b30c3ac4

core/embed/rust/src/ui/layout/device_menu_result.rs
obrusvit: Indeed, we can have just one base type:
https://github.com/trezor/trezor-firmware/pull/5065/commits/385b4556fc8da04cc2ecda3ec20248f0b30c3ac4

obrusvit:
@romanz I also created a preparation for returning index to disconnect device
https://github.com/trezor/trezor-firmware/pull/5065/commits/0a2e21076a7eac488a79de091bcbd0ed7c7a95ef

romanz:
Rust unittests have failed, but I am not sure it's related to this PR:
```
thread 'ui::layout_bolt::component::page::tests::paragraphs_three_long' panicked at src/ui/layout_bolt/component/page.rs:728:9:
assertion `left == right` failed
  left: Object {"active_page": Number(0), "component": String("ButtonPage"), "content": Object {"component": String("Paragraphs"), "paragraphs": Array [Array [String("This paragraph is"), String("\n"), String("using a bold font. It"), String("\n"), String("doesn't need to be all"), String("\n"), String("that long.")], Array [String("And this one is using"), String("\n"), String("MONO. Monospace is"), String("\n"), String("nice for numbers"), String("...")]]}, "hold": Bool(false), "page_count": Number(3)}
 right: Object {"active_page": Number(0), "component": String("ButtonPage"), "content": Object {"component": String("Paragraphs"), "paragraphs": Array [Array [String("This paragraph is"), String("\n"), String("using a bold font. It"), String("\n"), String("doesn't need to be all"), String("\n"), String("that long.")], Array [String("And this one is u"), String("\n"), String("sing MONO. Monosp"), String("\n"), String("ace is nice f"), String("...")]]}, "hold": Bool(false), "page_count": Number(3)}
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test ui::layout_bolt::component::page::tests::paragraphs_three_long ... FAILED
```

obrusvit:
> Rust unittests have failed, but I am not sure it's related to this PR:

That's a problem related to `format_pairing_code` test in `ui-eckhart`. I will fix it there.

PR #5066, opened 2025-05-20
WIP: Update Shamir wallet creation for Eckhart UI

bieleluk:


core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs
```diff
@@ -183,6 +192,7 @@ impl<'a> Component for ShareWordsScreen<'a> {
 
     fn render<'s>(&'s self, target: &mut impl Renderer<'s>) {
         self.header.render(target);
+        self.subtitle.render(target);
```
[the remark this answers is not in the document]
bieleluk: The `render` function is implemented for the `Option` as well so it's better to use it this way.

core/embed/rust/src/ui/layout_eckhart/firmware/updatable_info_screen.rs
```diff
@@ -14,76 +14,118 @@ use crate::{
...
+        self.action_bar.place(action_bar_area);
+
+        self.update_page(0);
+
         bounds
     }
 
     fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
+        // Update page count of the screen
+        ctx.set_page_count(self.paragraph.pager().total() as usize);
```
romanz: nit: I slightly prefer using `.into()` instead of `as usize` for casting due to https://users.rust-lang.org/t/into-vs-as-for-type-conversion/121248/4

core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs
```diff
@@ -126,20 +137,18 @@ impl<'a> Component for ShareWordsScreen<'a> {
...
-            Self::WORD_Y_OFFSET,
-        ));
-        let content_area = Rect::from_top_left_and_size(
-            top_left,
-            Offset::new(Self::WORD_AREA_WIDTH, Self::WORD_AREA_HEIGHT),
-        );
+        content_area = content_area
+            .inset(Insets::top(Self::WORD_Y_OFFSET))
+            .inset(theme::SIDE_INSETS)
+            .with_height(Self::WORD_AREA_HEIGHT);
```
romanz, line 151: Why not to add a `debug_assert!` here (instead of a separate `test_component_heights_fit_screen` unit test)?

core/embed/rust/src/ui/layout_eckhart/flow/show_share_words.rs
```diff
@@ -66,14 +66,24 @@ impl FlowController for ShowShareWords {
...
+    instructions_paragraphs: Option<ParagraphVecShort<'static>>,
     instructions_verb: Option<TString<'static>>,
     text_confirm: TString<'static>,
     text_check: TString<'static>,
 ) -> Result<SwipeFlow, error::Error> {
+    let br: ButtonRequest = ButtonRequestCode::ResetDevice.with_name("share_words");
+    // Determine whether to show the instructions or not
+    let has_intro = instructions_paragraphs.is_some();
     let nwords = words.len();
+    let initial_state = if instructions_paragraphs.is_some() {
```
romanz: nit:
```rust
    let initial_state = match instructions_paragraphs {
        Some(_) => &ShowShareWords::Instruction,
        None => &ShowShareWords::ShareWords,
    };
```

core/embed/rust/src/ui/layout_eckhart/flow/show_share_words.rs
```diff
@@ -141,10 +149,27 @@ pub fn new_show_share_words_flow(
         VerticalMenuScreenMsg::Close => Some(FlowMsg::Cancelled),
         _ => None,
     });
-    let mut res = SwipeFlow::new(&ShowShareWords::Instruction)?;
-    res.add_page(&ShowShareWords::Instruction, instruction)?
+    let mut res = SwipeFlow::new(initial_state)?;
+    // If there are no instructions, share words page sends the BR
```
romanz: Could you please explain why? I am not sure why the difference between both cases...

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -1193,11 +1193,14 @@ impl FirmwareUI for UIEckhart {
...
-            let text: TString = item.try_into()?;
-            instructions_paragraphs.add(Paragraph::new(&theme::TEXT_REGULAR, text));
-        }
+        let instructions_paragraphs = {
+            let mut vec = ParagraphVecShort::new();
+            for item in IterBuf::new().try_iterate(instructions)? {
+                let text: TString = item.try_into()?;
+                vec.add(Paragraph::new(&theme::TEXT_REGULAR, text));
+            }
+            vec.is_empty().then(|| None).unwrap_or_else(|| Some(vec))
```
romanz: nit: personally I find the following a bit more readable:
```rust
if vec.is_empty() { None } else { Some(vec) }
```

core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs
bieleluk: If we were able to get the number only in the runtime, I would agree with you. But in this particular case, all the numbers are constants and we already have the Rust unit tests set up so we can easily spot a potential miscalculation in the build time.  If we used `debug_assert!` we would still have to run something to get the runtime error.

core/embed/rust/src/ui/layout_eckhart/flow/show_share_words.rs
bieleluk: The trick is in the initial page of the flow.
Before the 1st [share](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=9156-3996&t=31CgoMucI8GkaKVL-0), the instruction must be shown. The initial page is `Instruction` and it sends the button request with remapped number of pages to `words + 2`
Before other [shares](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=9156-3945&t=9xsM25jNpJLdkBze-0), the instruction is not shown.  The initial page is `ShareWords` and it sends the button request with remapped number of pages to `words + 1`

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
bieleluk: The formatter did not allow one-liner so I used something similar
https://github.com/trezor/trezor-firmware/pull/5066/commits/5d41864dee3718b6bcf1a25604cc6a3106c44dbe

core/embed/rust/src/ui/layout_eckhart/flow/show_share_words.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/c05ec029e34746c918811f00d205d78d99391263

core/embed/rust/src/ui/layout_eckhart/firmware/updatable_info_screen.rs
bieleluk: Updated all such occurrences
https://github.com/trezor/trezor-firmware/pull/5066/commits/d62d3ca05d5e764743752c1449eab128ed67d4f6

core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs
romanz: > But in this particular case, all the numbers are constants and we already have the Rust unit tests set up so we can easily spot a potential miscalculation in the build time.

I agree, but it results in some "logical duplication" between the following parts of this module:
https://github.com/trezor/trezor-firmware/blob/deaa39e42d9125217e3c15ab54dc95fffbdf9ec7/core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs#L139-L148

https://github.com/trezor/trezor-firmware/blob/deaa39e42d9125217e3c15ab54dc95fffbdf9ec7/core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs#L344-L348

OTOH, [it seems to be happening in other modules as well](https://github.com/search?q=repo%3Atrezor%2Ftrezor-firmware%20test_component_heights_fit_screen&type=code), so it would probably be better to fix it in a separate PR.

> If we used debug_assert! we would still have to run something to get the runtime error.

Will the device tests fail in this case?

core/embed/rust/src/ui/layout_eckhart/flow/show_share_words.rs
romanz: Thanks for the explanation!

> Before other [shares](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=9156-3945&t=9xsM25jNpJLdkBze-0), the instruction is not shown. The initial page is ShareWords

If the instruction is not shown, would it be possible to remove its page?
image

romanz: IIUC, the "empty" page is not shown by default, but the user can "go back" to it:
[Screencast from 2025-06-04 16-12-52.webm](https://github.com/user-attachments/assets/00a51ad7-f193-47ff-8e68-8825b9696bb3)

romanz (commented):
Added a question here: https://github.com/trezor/trezor-firmware/pull/5066/files#r2126409049

core/embed/rust/src/ui/layout_eckhart/flow/show_share_words.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/62a47005e4a4fe66ffc339702f949cf6ab01f005
This should solve the unwanted back action to non-existing Instruction page. If you don't consider the change to be safe enough, I can create another enum i.e. `ShowShareWordsNoInstruction` that would not contain `Instruction` at all

romanz: https://github.com/trezor/trezor-firmware/commit/62a47005e4a4fe66ffc339702f949cf6ab01f005 is fine, but if possible, I personally would prefer separating the "instructions" dialog from the `show_share_words_extended` flow (so  that `ShowShareWords` won't have a `Instruction` variant), i.e. the layout below will not be generated as a part of `new_show_share_words_flow`:
image
WDYT?

tests/click_tests/reset.py
```diff
@@ -162,11 +166,11 @@ def confirm_words(debug: "DebugLink", words: list[str]) -> None:
     if debug.layout_type is LayoutType.Delizia:
         debug.swipe_up()
     elif debug.layout_type is LayoutType.Eckhart:
-        # Press ok if the select word screen is not yet present
-        if not TR.regexp("reset__select_word_template").match(
-            debug.read_layout().subtitle()
-        ):
+        # In 16-of-16 scenario, the intro screen is skipped
+        print(debug.read_layout().all_components())
```
romanz:
```suggestion
```

romanz (approved):
LGTM, thanks!

tests/click_tests/reset.py
romanz: Also, I am not sure about `16-of-16 scenario` - is it using a special layout? 
Or is it related to:
image

tests/click_tests/reset.py
```diff
@@ -116,12 +116,16 @@ def set_selection(debug: "DebugLink", diff: int) -> None:
...
     words: list[str] = []
 
+    # introductory screen
     if debug.layout_type is LayoutType.Caesar:
         debug.press_right()
     elif debug.layout_type is LayoutType.Delizia:
         debug.swipe_up()
     elif debug.layout_type is LayoutType.Eckhart:
-        debug.click(debug.screen_buttons.ok())
+        # In SLIP39, the screen is present only before the 1st share
```
romanz, line 127: nit: maybe pass a share index as an additional argument to this function?
(to explicitly skip the instructions' screen before the first share)

tests/click_tests/reset.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/83c33d2da91ce1c1eb1d200915a33d4faa5063aa
The 16 of  16 scenario would take 256 times hold to confirm actions which would cause the test timeout. For this reason, the layout is skipped using `press_yes` action.

tests/click_tests/reset.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/c628e06d9ec0a2ffeed265c243294c06274c73f1

tests/click_tests/reset.py
```diff
@@ -158,15 +164,14 @@ def read_words(debug: "DebugLink", do_htc: bool = True) -> list[str]:
     return words
 
 
-def confirm_words(debug: "DebugLink", words: list[str]) -> None:
+def confirm_words(debug: "DebugLink", words: list[str], skipt_intro=False) -> None:
```
romanz: nit (here and below):
```suggestion
def confirm_words(debug: "DebugLink", words: list[str], skip_intro=False) -> None:
```

tests/click_tests/test_repeated_backup.py
```diff
@@ -137,9 +137,11 @@ def test_repeated_backup_via_device(
     reset.confirm_read(debug, middle_r=True)
 
     second_backup_2_of_3: list[str] = []
-    for _ in range(3):
+    for share in range(3):
         # read words
-        words = reset.read_words(debug, do_htc=False)
+        is_eckahrt = debug.layout_type == LayoutType.Eckhart
```
romanz: nit:
```suggestion
        is_eckhart = debug.layout_type is LayoutType.Eckhart
```
(we usually use `is` for enums (instead of `==`).

tests/click_tests/test_repeated_backup.py
```diff
@@ -137,9 +137,11 @@ def test_repeated_backup_via_device(
...
+        skip_instruction = True if is_eckahrt and share > 0 else False
```
romanz:
```suggestion
        skip_instruction = is_eckhart and share > 0
```

tests/click_tests/reset.py
```diff
@@ -113,15 +113,21 @@ def set_selection(debug: "DebugLink", diff: int) -> None:
         raise RuntimeError("Unknown model")
 
 
-def read_words(debug: "DebugLink", do_htc: bool = True) -> list[str]:
+def read_words(
+    debug: "DebugLink", do_htc: bool = True, skip_instruction=False
```
romanz:
```suggestion
    debug: "DebugLink", do_htc: bool = True, skip_instruction: bool = False
```

tests/click_tests/reset.py
```diff
@@ -113,15 +113,21 @@ def set_selection(debug: "DebugLink", diff: int) -> None:
...
+) -> list[str]:
     words: list[str] = []
 
+    # introductory screen
     if debug.layout_type is LayoutType.Caesar:
         debug.press_right()
     elif debug.layout_type is LayoutType.Delizia:
         debug.swipe_up()
-    elif debug.layout_type is LayoutType.Eckhart:
+    elif debug.layout_type is LayoutType.Eckhart and not skip_instruction:
```
romanz: `not skip_instruction` means that there is an instruction screen which should be confirmed, right?
If so, maybe rename it to:
```suggestion
    elif debug.layout_type is LayoutType.Eckhart and confirm_instruction:
```

tests/click_tests/reset.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/d39ee0f160e8363900e175a8d4c4d592d9dd24fc

tests/click_tests/test_repeated_backup.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/c8e9c7e56bfee3ead9b5606f81df3a118314f3fd

tests/click_tests/test_repeated_backup.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/33d08abe7f83d167fe54b560dd8b6058838d02de

tests/click_tests/reset.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5066/commits/410449dce4f67551f802fecd60c334b2bcf9cd2d

PR #5067, opened 2025-05-21
chore(core): fix deprecation warning for `input_flow.throw(...)`

romanz:
https://docs.python.org/3.9/reference/expressions.html#generator.throw

Python 3.8 is no longer maintained: https://peps.python.org/pep-0569/


romanz:
It was deprecated in https://github.com/python/cpython/pull/96428.

PR #5071, opened 2025-05-21
Implement show_danger on Caesar

ibz:
On Caesar, `show_danger` was just calling `show_warning` internally which was not great because it gave the user no way to cancel the operation. This PR now properly implements `show_danger` on Caesar.

Also, having two strings `ethereum__unknown_contract_address` and `ethereum__unknown_contract_address__short` is useless since we can just use `words__know_what_your_doing`!


core/embed/rust/src/ui/layout_caesar/ui_firmware.rs
```diff
@@ -989,12 +989,34 @@ impl FirmwareUI for UICaesar {
...
     ) -> Result<impl LayoutMaybeTrace, Error> {
-        Err::<RootComponent<Empty, ModelUI>, Error>(ERROR_NOT_IMPLEMENTED)
+        let get_page = move |page_index| {
+            assert!(page_index == 0);
+
+            let btn_layout = ButtonLayout::cancel_none_text(TR::buttons__continue.into());
+            let btn_actions = ButtonActions::cancel_none_confirm();
+            let mut ops = OpTextLayout::new(theme::TEXT_NORMAL);
+            ops = ops.alignment(geometry::Alignment::Center);
+            ops = ops.alignment(geometry::Alignment::Center);
```
obrusvit, line 1004: duplicated line

obrusvit (approved):
One comment. Otherwise LGTM.

PR #5074, opened 2025-05-22
fix(core): fixed truncated strings in rsod on emulator

cepetr:
This PR fixes the issue with truncated message string shown in RSOD in the emulator.


matejcik (approved):
:)

PR #5081, opened 2025-05-23
ci(core): enable retries only for HW tests

romanz:
Following https://yaml-multiline.info, `>-` is used to split `TESTOPTS` in `core-hw.yml`.


romanz:
Thanks!
Tested in https://github.com/trezor/trezor-firmware/actions/runs/15208033070/:
image

PR #5085, opened 2025-05-25
refactor(core): convert translation strings to `str` once

romanz:
Instead of doing it on every `TranslationStringsChunk::get()` call.

Rebased over https://github.com/trezor/trezor-firmware/pull/4987.


obrusvit (approved):
LGTM

PR #5090, opened 2025-05-25
nrf update

TychoVrahe:


nordic/trezor/trezor-ble/src/management/management.c
```diff
@@ -49,8 +52,87 @@ typedef enum {
...
+ */
+static int read_image_sha256(uint8_t area_id,
+                             uint8_t out_hash[IMAGE_HASH_LEN]) {
+  const struct flash_area *fa;
+  int rc;
+
+  /* Open the flash area (slot) */
+  rc = flash_area_open(area_id, &fa);
+  if (rc) {
+    return rc;
```
cepetr, line 91: Shouldn't we use `if (rc < 0) {` to be consistent with the code below?

nordic/trezor/trezor-ble/src/management/management.c
```diff
@@ -49,8 +52,87 @@ typedef enum {
...
+  uint32_t img_size = hdr.ih_img_size;
+  uint32_t hdr_size = hdr.ih_hdr_size;
+  uint32_t tvl1_size = hdr.ih_protect_tlv_size;
+
+  /* Compute start of TLV trailer */
+  off_t off = 0 + hdr_size + img_size + tvl1_size + 4;
+
+  /* Scan TLVs until we find the SHA-256 entry */
+  while (true) {
+    uint16_t tlv_hdr[4];
```
cepetr: I'm not sure how TLV is encoded here, but do we need `tlv_hdr[4]`? Isn't `tlv_hdr[2]` enough? Maybe then we could replace the hardcoded `+ 4 ` with `+ sizeof(tlv_hdr)` ?

The same applies to the code in `embed/io/nrf/stm32u5/nrf_update.c`

core/embed/rust/Cargo.toml
```diff
@@ -32,6 +32,7 @@ ui_jpeg = []
 hw_jpeg_decoder = []
 bootloader = []
 prodtest = []
+kernel = []
```
cepetr: I couldn't find any use of this feature. Do we need it?

core/embed/io/nrf/stm32u5/nrf_update.c
```diff
@@ -0,0 +1,137 @@
...
+  uint32_t tvl1_size = hdr->ih_protect_tlv_size;
+
+  /* Compute start of TLV trailer */
+  off_t off = 0 + hdr_size + img_size + tvl1_size + 4;
+
+  /* Scan TLVs until we find the SHA-256 entry */
+  while (true) {
+    uint16_t tlv_hdr[4];
+
+    memcpy(tlv_hdr, binary_ptr + off, sizeof(tlv_hdr));
```
cepetr, line 84: The code doesn't check whether the offset lies within the area where the NRF firmware resides. If the headers are corrupted or intentionally modified, this can lead to a MemManage or SecureFault during firmware startup, causing infinite crashes or access to addresses that shouldn't be accessed. I don't think it's a major issue (since firmware integrity is always verified), but the pointer to the image comes from the unprivileged world...

core/embed/io/nrf/inc/io/nrf.h
```diff
@@ -116,3 +120,9 @@ bool nrf_test_gpio_reserved(void);
 bool nrf_system_off(void);
 
 void nrf_reboot(void);
+
+void nrf_send_urt_data(const uint8_t *data, uint32_t len);
+
+bool nrf_update_required(const uint8_t *data, size_t len);
```
cepetr: Maybe we could add some comments here. I would also rename the generic `data` and `len` to something like `image_ptr` and `image_len` (or `nrf_app` and `nrf_app_len`).

core/embed/io/nrf/inc/io/nrf.h
```diff
@@ -116,3 +120,9 @@ bool nrf_test_gpio_reserved(void);
 bool nrf_system_off(void);
 
 void nrf_reboot(void);
+
+void nrf_send_urt_data(const uint8_t *data, uint32_t len);
```
cepetr: Wouldn't be better to replace `urt` with `uart` or maybe `dfu`?

core/embed/sys/syscall/stm32/syscall_verifiers.h
```diff
@@ -207,6 +207,16 @@ secbool ble_read__verified(uint8_t *data, size_t len);
...
 #endif
 
+// ---------------------------------------------------------------------
+#ifdef USE_NRF
+
+#include <io/nrf.h>
+
+bool nrf_update_required__verified(const uint8_t *data, size_t len);
+
+bool nrf_update__veridied(const uint8_t *data, size_t len);
```
cepetr: veridied typo

core/embed/sys/syscall/stm32/syscall_dispatch.c
```diff
@@ -732,6 +736,22 @@ __attribute((no_stack_protector)) void syscall_handler(uint32_t *args,
     } break;
 #endif
 
+#ifdef USE_NRF
+
+    case SYSCALL_NRF_UPDATE_REQUIRED: {
+      const uint8_t *data = (const uint8_t *)args[0];
+      size_t len = args[1];
+      args[0] = nrf_update_required(data, len);
```
cepetr: We should call `nrf_update_required__verified` and `nrf_update__verified` here.

core/embed/sys/linker/stm32u5g/firmware.ld
```diff
@@ -50,6 +50,8 @@ SECTIONS {
     *(.text*);
     . = ALIGN(4);
     *(.rodata*);
+    KEEP(*(.nrf_app));
```
cepetr, line 50: I'm not sure if this really matters, but I would add `. = ALIGN(4)` before `KEEP(*(.nrf_app))` to ensure the data is aligned and the CPU can access it as efficiently as possible..

cepetr (changes requested):

core/embed/rust/Cargo.toml
TychoVrahe: removed in https://github.com/trezor/trezor-firmware/commit/e18e1342812b3789e765304c4db5d4d10c90ea1f and https://github.com/trezor/trezor-firmware/commit/9753cc3ab430abad7f1e9fd242ea168ddb77b6fc

( two commits as part of the stuff was in wrong commit)

nordic/trezor/trezor-ble/src/management/management.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/ac94ae092e629d83d63f42f350126b7d7a7b73f5

nordic/trezor/trezor-ble/src/management/management.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/ac94ae092e629d83d63f42f350126b7d7a7b73f5

core/embed/io/nrf/stm32u5/nrf_update.c
TychoVrahe: added checks in https://github.com/trezor/trezor-firmware/commit/688b871d11ca5211a4281da17eadc12989e3e302

core/embed/io/nrf/inc/io/nrf.h
TychoVrahe: added commits and renamed in https://github.com/trezor/trezor-firmware/commit/688b871d11ca5211a4281da17eadc12989e3e302

regenerated the rest of function comments in https://github.com/trezor/trezor-firmware/commit/81ade91291f3489e989ea2c4e71acbbcec712c20

core/embed/io/nrf/inc/io/nrf.h
TychoVrahe: renamed https://github.com/trezor/trezor-firmware/commit/688b871d11ca5211a4281da17eadc12989e3e302

core/embed/sys/syscall/stm32/syscall_verifiers.h
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/50494dfd3330c465b1bf77df3fdfb88d60696a93

core/embed/sys/syscall/stm32/syscall_dispatch.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/50494dfd3330c465b1bf77df3fdfb88d60696a93

core/embed/sys/linker/stm32u5g/firmware.ld
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/50494dfd3330c465b1bf77df3fdfb88d60696a93

core/embed/rust/src/smp/upload.rs
```diff
@@ -0,0 +1,96 @@
...
+    .is_err()
+    {
+        return false;
+    }
+
+    rem_len = rem_len.saturating_sub(CHUNK_SIZE);
+
+    let mut off: usize = CHUNK_SIZE;
+
+    while rem_len > 0 {
```
cepetr: not so important, but you could consider this:

```rust
for chunk in image_data.chunks(CHUNK_SIZE) {

}
```

in this particular case, you can try this:

```rust
let mut offset = CHUNK_SIZE;
for chunk in image_data.chunks(CHUNK_SIZE).skip(1) {
   ...
    offset += CHUNK_SIZE;
}
```

core/embed/rust/src/smp/upload.rs
```diff
@@ -0,0 +1,96 @@
...
+    unwrap!(enc.u8(0));
+    unwrap!(enc.str("hash"));
+    unwrap!(enc.bytes(image_hash));
+    unwrap!(enc.str("data"));
+    unwrap!(enc.bytes(&image_data[..CHUNK_SIZE]));
+
+    let encoded_len = writer.bytes_written();
+    receiver_lock();
+
+    send_request(&mut data, encoded_len, &mut data_encoded, 2, 1, 1);
```
cepetr: There are some hard-coded constants here, like 5 above, or 2, 1, 1 here... Wouldn't it be better to give them names or use enum? The same applies to echo and reset commands.

core/embed/rust/src/smp/upload.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub fn upload_image(image_data: &[u8], image_hash: &[u8]) -> bool {
+    if receiver_is_locked() {
+        return false;
+    }
+
+    let mut data = [0u8; MAX_PACKET_SIZE];
+    let mut data_encoded = [0u8; MAX_PACKET_SIZE];
+
+    let mut writer = BufferCounter::new(&mut data[8..]);
```
cepetr: There's a `HEADER_SIZE` constant that could be used instead of `8..`, but it would still be better to use a different approach that avoids having to specify this at all.

It seems to me that we don't need to reserve 8 bytes at all, since the data is copied into a new buffer in `send_request()`, and the 8-byte header can be easily added within that routine.

core/embed/rust/src/smp/upload.rs
```diff
@@ -0,0 +1,96 @@
...
+        unwrap!(enc.str("off"));
+        unwrap!(enc.u32(off as _));
+        unwrap!(enc.str("data"));
+        unwrap!(enc.bytes(&image_data[off..off + chunk_len]));
+
+        let encoded_len = writer.bytes_written();
+
+        receiver_lock();
+
+        send_request(&mut data, encoded_len, &mut data_encoded, 2, 1, 1);
```
cepetr: The variable names here are a bit confusing. Consider this:

`encoded_len` -> `data_len`
`data_encoded` -> `buffer`

core/embed/rust/src/smp/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+pub fn create_header(buffer: &mut [u8], op: u8, len: u16, group: u16, seq: u8, cmd_id: u8) {
+    // If buffer is too small, do nothing (mirrors C behavior of no safety check)
+    if buffer.len() < 8 {
+        return;
+    }
+
+    let len_be = len.to_be_bytes(); // [hi, lo]
+    let group_be = group.to_be_bytes(); // [hi, lo]
+
+    buffer[0] = op;
```
cepetr: Maybe we can utilize `SmpHeader` struct instead of using array indexes here...

core/embed/rust/src/smp/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+#[derive(Copy, Clone, PartialEq)]
+pub enum MsgType {
+    Echo,
+    ImageUploadResponse,
+    Unknown,
+}
+
+#[derive(Copy, Clone)]
+pub struct SmpReceiver {
+    rx_frame: [u8; BOOT_SERIAL_FRAME_MTU + 3],
```
cepetr: There are a lot of hard-coded constants here like this. They're not easy to understand and could be a source of errors in the future.

core/embed/rust/src/smp/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+
+pub fn receiver_unlock() {
+    let key = irq_lock();
+    unsafe {
+        ptr::write_volatile(&raw mut SMP_RECEIVER, None);
+    }
+    irq_unlock(key);
+}
+
+pub fn receiver_lock() {
```
cepetr: consider this refactoring:
- rename `receiver_lock/unlock` to `receiver_acquire/release`.
- instead of `receiver_is_locked()` return `Result<SmpReceiver,()>` in `receiver_lock`.

core/embed/rust/src/smp/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+    buffer[7] = cmd_id;
+}
+
+/// A simple writer that copies into a `&mut [u8]` and counts bytes written.
+pub struct BufferCounter<'a> {
+    buf: &'a mut [u8],
+    written: usize,
+}
+
+impl<'a> BufferCounter<'a> {
```
cepetr: Wouldn't `SmpBuffer` be a slightly nicer name?

core/embed/rust/src/smp/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+            (1, 1) => {
+                self.msg_type = Some(MsgType::ImageUploadResponse);
+            }
+            _ => self.msg_type = Some(MsgType::Unknown),
+        }
+    }
+}
+
+pub fn process_rx_byte(byte: u8) {
+    let opt = unsafe { ptr::read_volatile(&raw mut SMP_RECEIVER) };
```
cepetr: If I understand this code correctly, we're copying 2 × 700 bytes in every interrupt?

cepetr (changes requested):

core/embed/rust/src/smp/upload.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/63f7b76376e01a07a972bb62242f36345c600c19

core/embed/rust/src/smp/upload.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/f210c7bd4e037634c63984e63438569ac3aeb8f8

core/embed/rust/src/smp/mod.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/530f18f3ffeb6ae29b753127bd3f7faddf4bd0bb

core/embed/rust/src/smp/mod.rs
TychoVrahe: this commit refactors to use references instead of read+write: https://github.com/trezor/trezor-firmware/commit/e44c7e6c8e1ccceb328089ee212530a773352f4f

core/embed/rust/src/smp/mod.rs
TychoVrahe: partially changed here: https://github.com/trezor/trezor-firmware/commit/d5cc3a9beea95999193e3a9474e7122987f60149

i can't have a single pair of functions as i need to be acquiring and releasing and reading in thread while modification is done in interrupt

core/embed/rust/src/smp/mod.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/d5cc3a9beea95999193e3a9474e7122987f60149

core/embed/rust/src/smp/mod.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/88c2511a4ee0884602dab7b38a9f51dc55e1baf7

core/embed/rust/src/smp/upload.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/88c2511a4ee0884602dab7b38a9f51dc55e1baf7

core/embed/rust/src/smp/upload.rs
TychoVrahe: in send request the data is encoded (including header) from data to buffer when coping. So i introduced a third buffer in https://github.com/trezor/trezor-firmware/commit/88c2511a4ee0884602dab7b38a9f51dc55e1baf7 , and the header is copied outside of send_request. I think its a bit cleaner like this.

core/embed/rust/src/smp/mod.rs
TychoVrahe: some more cleaning and safety comments added in https://github.com/trezor/trezor-firmware/commit/907bc078ae0c1cbbcea7f0355b2a5c6722b4f052

cepetr (approved):
utACK

PR #5091, opened 2025-05-25
test(core): collect GC statistics via new debuglink RPC

romanz:
Also, add GC leak check for every test setup/teardown.
It will also run during HW device tests (after it was removed in https://github.com/trezor/trezor-firmware/pull/4993).

### DIFF
--- a/common/protob/messages.proto
+++ b/common/protob/messages.proto
@@ -136,6 +136,8 @@ enum MessageType {
     MessageType_DebugLinkWatchLayout = 9006 [(bitcoin_only) = true, (wire_debug_in) = true];
     MessageType_DebugLinkResetDebugEvents = 9007 [(bitcoin_only) = true, (wire_debug_in) = true];
     MessageType_DebugLinkOptigaSetSecMax = 9008 [(bitcoin_only) = true, (wire_debug_in) = true];
+    MessageType_DebugLinkGetGcInfo = 9009 [(bitcoin_only) = true, (wire_debug_in) = true];
+    MessageType_DebugLinkGcInfo = 9010 [(bitcoin_only) = true, (wire_debug_out) = true];
 
     // Ethereum
     MessageType_EthereumGetPublicKey = 450 [(wire_in) = true];
--- a/core/embed/upymod/modtrezorutils/modtrezorutils.c
+++ b/core/embed/upymod/modtrezorutils/modtrezorutils.c
@@ -40,6 +40,7 @@
 #include <util/scm_revision.h>
 #include <util/unit_properties.h>
 #include "blake2s.h"
+#include "memzero.h"
 
 #if !defined(TREZOR_EMULATOR)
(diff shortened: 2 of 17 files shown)


common/protob/messages.proto
```diff
@@ -136,6 +136,9 @@ enum MessageType {
     MessageType_DebugLinkWatchLayout = 9006 [(bitcoin_only) = true, (wire_debug_in) = true];
     MessageType_DebugLinkResetDebugEvents = 9007 [(bitcoin_only) = true, (wire_debug_in) = true];
     MessageType_DebugLinkOptigaSetSecMax = 9008 [(bitcoin_only) = true, (wire_debug_in) = true];
+    MessageType_DebugLinkGetGcInfo = 9009 [(bitcoin_only) = true, (wire_debug_in) = true];
+    MessageType_DebugLinkGcInfoItem = 9010 [(bitcoin_only) = true, (wire_debug_out) = true];
```
mmilata: Since it's not transmitted without a containing message it doesn't need an ID. The definition can be even nested inside `DebugLinkGcInfo` like `SignTx` and `CoinJoinRequest`.

core/embed/upymod/modtrezorutils/modtrezorutils.c
```diff
@@ -333,31 +334,74 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_0(mod_trezorutils_enable_oom_dump_obj,
                                  mod_trezorutils_enable_oom_dump);
 #endif  // MICROPY_OOM_CALLBACK
 
+static gc_info_t current_gc_info = {0};
+
+/// if __debug__:
+///     def clear_gc_info() -> dict[str, int]:
```
mmilata:
```suggestion
///     def clear_gc_info() -> None:
```

common/protob/messages.proto
romanz: Thanks! Fixed in b3ed1424c4.

core/embed/upymod/modtrezorutils/modtrezorutils.c
romanz: b3ed1424c4

romanz:
Pushed https://github.com/trezor/trezor-firmware/pull/5091/commits/331221c878074ec118006eb8cdd2531bbe767bb9 fixup:
 - reduce DebugLinkGcInfo response size to avoid #4401 
 - change `max_info` unit to bytes (instead of GC blocks)
 - add a device test

romanz:
Rebasing to update UI fixtures.

romanz:
Pushed a small fixup to handle non-frozen builds correctly (where `current_gc_info` is zeroed):
https://github.com/trezor/trezor-firmware/compare/2041f47b48640492ce3cc6df394e2ba1acc6586a..7e252a566d1604526fa6de3a75ae6d55fd527edd

PR #5104, opened 2025-05-27
fix bootloader entry on T3W1

TychoVrahe:
This PR fixes an issue with entering bootloader on T3W1, and changes the haptic effect on success.


kopecdav:
tested, ACK

kopecdav (approved):

PR #5105, opened 2025-05-27
prodtest backlight control fix

TychoVrahe:
This PR fixes and issue with setting backlight in prodtest. while backlight could be set, it also causes screen to go blank and was reset after any other command (showing something on display), so not very useful.

Now prodtest homescreen will be shown after the backlight command, and other commands will not override the setting.


kopecdav (approved):
Works well, thanks!

PR #5111, opened 2025-05-28
Tweak confirm_output subtitle

ibz:

### DIFF
--- a/core/embed/rust/librust_qstr.h
+++ b/core/embed/rust/librust_qstr.h
@@ -661,7 +661,6 @@ static void _librust_qstrs(void) {
   MP_QSTR_sd_card__wrong_sd_card;
   MP_QSTR_select_word;
   MP_QSTR_select_word_count;
-  MP_QSTR_send__address_path;
   MP_QSTR_send__cancel_sign;
   MP_QSTR_send__confirm_sending;
   MP_QSTR_send__from_multiple_accounts;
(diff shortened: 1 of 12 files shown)


Hannsek:
What is this? Why we are doing this?

Hannsek:
This touches more then in the issue, right?

ibz:
> This touches more then in the issue, right?

It should not. But there are still some issues currently, that's why it's still a draft.

core/src/trezor/ui/layouts/delizia/__init__.py
```diff
@@ -378,14 +378,15 @@ async def confirm_output(
...
         title = address_label
     elif title is not None:
-        pass
+        amount_subtitle = title
     elif output_index is not None:
-        title = f"{TR.words__recipient} #{output_index + 1}"
+        title = amount_subtitle = f"{TR.words__recipient} #{output_index + 1}"
     else:
-        title = TR.send__title_sending_to
+        title = amount_subtitle = TR.send__title_sending_to
```
bieleluk: This is probably out of the scope of this PR, but might be easily fixed.
The `title` variable is created but then it's assigned to the `subtitle` argument which is a bit confusing.
```suggestion
        subtitle = amount_subtitle = TR.send__title_sending_to
```

bieleluk (commented):
The PR solves the issue but it introduces 2 minor side effects:
1. On TS5, there are 2 cases, where the amount subtitle is lost.
image
image

2. On TS3, the address label is a part of the paragraph. Without the `send__address_path` prefix, it's a bit less clear IMO.
image

We should ask @Hannsek if these changes are acceptable.
Otherwise LGTM

ibz:
> 1. On TS5, there are 2 cases, where the amount subtitle is lost

Actually this is the same case. Address is shown normally, then the subtitle is missing on `amount`.

Hannsek:
image

IN which flow I can see the account type in the main flow?

PR #5113, opened 2025-05-28
fix(core): make `'.'` part of `ethereum__unknown_contract_address`

romanz:
Following https://github.com/trezor/trezor-firmware/pull/5086#discussion_r2111306385.


romanz:
If we (re)-add `'.'` to `ethereum__unknown_contract_address`, the layouts change a bit - is it OK?

image
image

@Hannsek @ibz

Hannsek:
I wouldn't put it back as it is displayed as "header" in TS3 (capital letters) and it looks weird IMO. cc @lapohoda

PR #5115, opened 2025-05-28
feat(core): show success screen after sending a response to the host

romanz:
It should implement similar workflow handling to https://github.com/trezor/trezor-firmware/commit/151eb62150 (keeping the current session until all workflows are over).

### DIFF
--- a/core/src/trezor/ui/layouts/delizia/__init__.py
+++ b/core/src/trezor/ui/layouts/delizia/__init__.py
@@ -279,6 +279,12 @@ async def show_address(
         )
         return result
 
+    title_success = (
+        TR.address__public_key_confirmed
+        if title in ("XPUB", TR.address__public_key)
+        else TR.address__confirmed
+    )
+
     await raise_if_not_confirmed(
         trezorui_api.flow_get_address(
             address=address,
@@ -291,6 +297,7 @@ async def show_address(
             account=account,
             path=path,
             xpubs=[(xpub_title(i), xpub) for i, xpub in enumerate(xpubs)],
+            title_success=title_success,
             br_name=br_name,
             br_code=br_code,
         ),
@@ -386,30 +393,19 @@ def show_danger(
 
 
 def show_success(
-    br_name: str | None,
+    br_name: str,
     content: str,
     subheader: str | None = None,
     button: str | None = None,
-    time_ms: int = 0,
 ) -> Awaitable[None]:
     return raise_if_not_confirmed(
         trezorui_api.show_success(
             title=content,
-            button=button or "",
-            description=subheader or "",
-            time_ms=time_ms,
+            button="",
+            description=subheader if subheader else "",
         ),
         br_name,
         ButtonRequestType.Success,
-    )
-
-
-def show_continue_in_app(content: str) -> Awaitable[None]:
-    return show_success(
-        content=content,
-        button=TR.instructions__continue_in_app,
-        time_ms=3200,
-        br_name=None,
     )
 
 
--- a/core/src/trezor/ui/layouts/delizia/__init__.py
+++ b/core/src/trezor/ui/layouts/delizia/__init__.py
@@ -1,14 +1,14 @@
 from typing import TYPE_CHECKING
 
 import trezorui_api
-from trezor import TR, ui, utils
+from trezor import TR, ui, utils, workflow
 from trezor.enums import ButtonRequestType, RecoveryType
 from trezor.wire import ActionCancelled
 
 from ..common import draw_simple, interact, raise_if_not_confirmed, with_info
 
 if TYPE_CHECKING:
-    from typing import Awaitable, Iterable, NoReturn, Sequence, TypeVar
+    from typing import Any, Awaitable, Coroutine, Iterable, NoReturn, Sequence, TypeVar
 
     from ..common import ExceptionType, PropertyType
 
@@ -279,12 +279,6 @@ async def show_address(
         )
         return result
 
-    title_success = (
-        TR.address__public_key_confirmed
-        if title in ("XPUB", TR.address__public_key)
-        else TR.address__confirmed
-    )
-
     await raise_if_not_confirmed(
         trezorui_api.flow_get_address(
             address=address,
@@ -297,11 +291,16 @@ async def show_address(
             account=account,
             path=path,
             xpubs=[(xpub_title(i), xpub) for i, xpub in enumerate(xpubs)],
-            title_success=title_success,
             br_name=br_name,
             br_code=br_code,
         ),
         None,
+    )
+
+    show_continue_in_app(
+        TR.address__public_key_confirmed
+        if title in ("XPUB", TR.address__public_key)
+        else TR.address__confirmed
     )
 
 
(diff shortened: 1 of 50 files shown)


core/src/trezor/ui/layouts/delizia/__init__.py
```diff
@@ -336,14 +342,14 @@ def show_danger(
...
+async def show_success(
     br_name: str | None,
     content: str,
     subheader: str | None = None,
     button: str | None = None,
     time_ms: int = 0,
-) -> Awaitable[None]:
-    return raise_if_not_confirmed(
+) -> None:
+    return await raise_if_not_confirmed(
```
matejcik, line 352: please revert this chunk, it (a) does not have any observable effect and (b) instead of `async -> return await` we prefer `def -> Awaitable`
(because calling an async function allocates an item, but passing an awaitable via a normal function doesn't)

tests/conftest.py
```diff
@@ -257,6 +259,17 @@ def _initialize_with_retries(
     pytest.fail("Failed to communicate with Trezor")
 
 
+def _skip_async_confirmation(raw_client: Client) -> None:
```
matejcik, line 262: not totally sure how this works
i assume you're trying to skip over the "show continue in app" screen, right?
this is going to be racey in some very weird cases though: because the dialog is on a timeout, you have a TOCTOU on the `press_yes`, which may end up going to the homescreen

i'd suggest `synchronize_at("Homescreen")` (maybe that will be a problem when it's RecoveryHomescreen?? dunno)
and on the Rust side, adding code that, if animations are disabled, ignore the timeout and display "continue in the app" for one frame only

matejcik (commented):
generally fine with this approach.
we still **may** end up needing units_of_work in order to make THP behave nicely -- THP runs some amount of background loops for retransmissions and some miscellaneous stuff, those must not be interrupted by the session restarts but at the same time aren't workflows

core/src/trezor/ui/layouts/delizia/__init__.py
romanz: TIL - thanks!

tests/conftest.py
romanz: > i assume you're trying to skip over the "show continue in app" screen, right?

Yes - I was mainly trying to reduce the tests' duration, following https://github.com/trezor/trezor-firmware/pull/5097#issuecomment-2911128376.

> on the Rust side, adding code that, if animations are disabled, ignore the timeout and display "continue in the app" for one frame only

Good idea, thanks! It should definitely help :)

romanz: Fixed in https://github.com/trezor/trezor-firmware/pull/5115/commits/3e04b2dc2d96f022613ce33c2713e10c7abb4c5f.

core/src/trezor/ui/layouts/delizia/__init__.py
romanz: Should be fixed in https://github.com/trezor/trezor-firmware/pull/5115/commits/e3bef9840e2d2ba698d8738916cc33368fcdc2ff.
IIUC, `workflow.spawn(...)` should accept a `Task = Union[Coroutine, Generator, "wait"]`, so I have created a local wrapper async function in `show_continue_in_app()`.

romanz:
Rebasing to resolve a conflict.

romanz:
> THP runs some amount of background loops for retransmissions and some miscellaneous stuff, those must not be interrupted by the session restarts but at the same time aren't workflows

Opened https://github.com/trezor/trezor-firmware/pull/5166, which can be useful for background loops' cancellation.

romanz:
Rebased to resolve a merge conflict.

romanz:
Pushed a small test-related fix (to handle a race in https://data.trezor.io/dev/firmware/ui_report/15725349158/T3W1-en-core_device_test-index.html).

There is no need in the last `self.debug.click(self.debug.screen_buttons.ok())` since we use "zero timeout" when animations are disabled.

core/src/trezor/ui/layouts/delizia/__init__.py
matejcik: `show_success` does return a Coroutine though, no?
like, i'm not sure what the exact hierarchy is in CPython type system, but in MicroPython an `Awaitable` is an instance of `Coroutine` so this should work without the wrapper.
is the type checker complaining?

romanz: Yes - it fails if I remove the `async` wrapper:
```
def show_continue_in_app(content: str) -> None:
    task = show_success(
        content=content,
        button=TR.instructions__continue_in_app,
        time_ms=3200,
        br_name=None,
    )
    workflow.spawn(task)
```

The error I get is:
```
./trezor-firmware/core/src/trezor/ui/layouts/delizia/__init__.py:420: - error: Argument of type "Awaitable[None]" cannot be assigned to parameter "workflow" of type "Task" in function "spawn"
  Type "Awaitable[None]" is incompatible with type "Task"
    "Awaitable[None]" is incompatible with "Coroutine[Unknown, Unknown, Unknown]"
    "Awaitable[None]" is incompatible with "Generator[Unknown, Unknown, Unknown]"
    "Awaitable[None]" is incompatible with "wait" (reportArgumentType)

Found 1 issues above
```

IIUC, `Awaitable` in CPython is something we can `__await__` on:
- https://mypy.readthedocs.io/en/stable/protocols.html#collections-abc-awaitable-t

But it doesn't seem to support `send`, `throw`, `close` - while `Generator` and `Coroutine` do support them:
- https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator
- https://docs.python.org/3/reference/datamodel.html#coroutine-objects

romanz: https://github.com/trezor/trezor-firmware/pull/5115/commits/176f0a69a888582937228f4a6dabf0b907eecc5d

romanz:
Ran TS5 device tests on 176f0a69a888582937228f4a6dabf0b907eecc5d:
```
+ pytest tests/device_tests/ -m 'not sd_card' --random-order-seed=1 --random-order-bucket=module
...
=== 1405 passed, 188 skipped, 6 deselected, 12 xfailed in 8898.82s (2:28:18) ===
```

matejcik (approved):

romanz:
Tested address, xpub and transaction signature confirmation on [Suite](https://github.com/trezor/trezor-suite/actions/runs/15726249807) with BTC Testnet4.

Hannsek:
@bosomt Please test it all various cases including altcoins, staking, swaps, … with various wallets (not only Trezor Suite)

Hannsek:
Related [Suite PR](https://github.com/trezor/trezor-suite/pull/19393).

bosomt:
**QA WIP**

- tested in Suite in send, receive, claim reward, swap including receive, ✅ 
- Exodus send and receive works as expected ✅  
- https://connect.trezor.io/ fails , needs to be fixed before releasing new 2.9.0 FW into wild ❌ 

**Info:**
 - Suite version: desktop 25.7.1 (b381d0268e9406be8b56b209967c2da19afffd81)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.7.1 Chrome/134.0.6998.178 Electron/35.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.9.0 regular (revision a09e6f04d4204607d821d9b742dcf7ffe9c4e2f5)

STew790:
**QA OK**
Tested on the new 2.9.1 and works on connect.trezor.io 🙏

**Info:**
 - Suite version: web 25.9.0 (842608cadada6a45f9f7203ac10e786f11e43494)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1920x1080
 - Device: Trezor T3T1 2.9.1 universal (revision acd56d0113a547e9b5983510b0ce9de079e30e05)
 - Transport: WebUsbTransport

PR #5118, opened 2025-05-29
Secure monitor

cepetr:
This PR introduces a new level of isolation for sensitive data in the Trezor firmware, called Secure Monitor. Secure Monitor utilizes the ARM Cortex-M33 TrustZone mechanism, along with STM32U5 GMTZ and TrustZone-aware peripherals, to isolate secure storage, access to secure elements (e.g., Optiga, Tropic), and other critical resources from the non-secure world (kernel and unprivileged apps).

From the user's perspective, there should be no visible changes. The Secure Monitor is currently built only for the T3W1.

secmon drawio

Alongside Secure Monitor, several refactorings have been made:

1) Completely refactored unprivileged SAES execution - the previous assembler-based hacks were replaced with a standard task-switching mechanism.
2) Completely refactored the callback mechanism used for storage progress - assembler-based routines were replaced with task switching as well.
3) The MPU now supports region switching for applets residing in different memory areas.
4) Modified syscall processing: syscalls can now be handled either directly in the SVCall handler (privileged handler mode, as before) or in the kernel main loop (privileged thread mode).
5) Implemented support for passing `systask_postmortem_t` in boot arguments (`USE_BOOTARGS_RSOD`) and updated the bootloader to display RSOD for the application.
6) Simplified linker scripts and memory definitions; removed unused symbols.


core/embed/projects/secmon/version.h
```diff
@@ -0,0 +1,11 @@
+#define VERSION_MAJOR 1
+#define VERSION_MINOR 0
+#define VERSION_PATCH 0
+#define VERSION_BUILD 0
+
+#define FIX_VERSION_MAJOR 1
```
TychoVrahe: we probably don't need fix version for secmon

core/embed/projects/secmon/version.h
```diff
@@ -0,0 +1,11 @@
...
+#define FIX_VERSION_MINOR 0
+#define FIX_VERSION_PATCH 0
+#define FIX_VERSION_BUILD 0
+
+#define VERSION_MONOTONIC 1
```
TychoVrahe: monotonic version should be placed along with other monotonic version in embed/models/XX/versions.h, just for T3W1 for now should be enough


or even.. do we need downgrade protection for secure monitor separately? i am thinking, since the secure monitor is tightly coupled with whole FW, maybe FW downgrade protection is enough? 

exception maybe when using custom firmware with presigned secure monitor, then you don't have to bump firmware monotonic version: but do we care?

core/site_scons/models/D002/discovery2.py
```diff
@@ -18,6 +18,7 @@ def configure(
 
     mcu = "STM32U5G9xx"
     linker_script = """embed/sys/linker/stm32u5g/{target}.ld"""
+    memory_layout = "memory.ld"
```
TychoVrahe, line 21: maybe we should enable SECMON for D002 too? to keep it at the most advanced state, and we don't need to worry about comatibility

core/embed/sys/smcall/stm32/smcall_probe.c
```diff
@@ -0,0 +1,68 @@
...
+    return false;
+  }
+
+  if (!cmse_check_address_range((void *)addr, len,
+                                CMSE_MPU_READ | CMSE_NONSECURE)) {
+    return false;
+  }
+
+  return true;
+  ;
```
TychoVrahe: extra ;

core/embed/sys/trustzone/stm32u5/trustzone.c
```diff
@@ -53,6 +118,19 @@ static void tz_configure_arm(void) {
...
   // FPU registers are cleared on exception return
   FPU->FPCCR |= FPU_FPCCR_CLRONRET_Msk;
+
+  uint32_t reg_value = SCB->AIRCR;
+  reg_value &= ~SCB_AIRCR_VECTKEY_Msk;
+  reg_value |= 0x5FAUL << SCB_AIRCR_VECTKEY_Pos;
+  // Prioritize secure world interrupts over non-secure world
+  reg_value |= SCB_AIRCR_PRIS_Msk;
+#if PRODUCTION
+  // System reset request available only in secure monitor
```
TychoVrahe: available not only in secure monitor, but all secure-mode builds

core/embed/projects/bootloader/main.c
```diff
@@ -390,6 +390,20 @@ int bootloader_main(void) {
 
   drivers_init(manufacturing_mode, &touch_initialized);
 
+#ifdef USE_BOOTARGS_RSOD
+  if (bootargs_get_command() == BOOT_COMMAND_SHOW_RSOD) {
+    // post mortem info was left in bootargs
+    boot_args_t args;
+    bootargs_get_args(&args);
+#ifdef FANCY_FATAL_ERROR
```
TychoVrahe: bootloader always has FANCY_FATAL_ERROR and gui, so this condition is not needed

core/embed/sys/syscall/inc/sys/syscall_ipc.h
```diff
@@ -0,0 +1,46 @@
...
+// Initializes IPC for syscalls
+bool syscall_ipc_init(void);
+
+// Enqueues a syscall for processing in the kernel event loop
+//
+// Queued syscalls are signalled to the kernel task by
+// asserting SYSEVENT_SYSCALL.
+//
+// The function may be called only from kernel handler mode
+// (respecively from SVCall handler).
```
TychoVrahe: typo respecively

core/embed/util/flash/stm32u5/flash.c
```diff
@@ -36,11 +36,51 @@
...
+void flash_init(void) {
+#ifdef SECMON
+  // FIRMWARE_AREA is defined here because it depends on the
+  // SECMON size which is not known at compile time.
+
+  FIRMWARE_AREA = (flash_area_t){
+      .num_subareas = 2,
+      .subarea[0] =
+          {
+              .first_sector = FIRMWARE_SECTOR_START,
```
TychoVrahe, line 57: some weird formatting difference between subarea 0 and 1, not sure why its not automatically fixed

core/embed/sys/smcall/stm32/smcall_dispatch.c
```diff
@@ -0,0 +1,380 @@
...
+
+#if PYOPT == 0
+    case SMCALL_OPTIGA_SET_SEC_MAX: {
+      optiga_set_sec_max();
+    } break;
+#endif
+#endif
+
+    case SMCALL_STORAGE_INIT: {
+      // !@# check args[0] if it is in a non-secure memory region
```
TychoVrahe: important or not?

core/embed/projects/secmon/version.h
cepetr: removed from secmon and also from kernel, see 6d12f05fef1ddefa64cdb109135a714b27f10ba2

core/embed/projects/secmon/version.h
cepetr: removed for now, see 6d12f05fef1ddefa64cdb109135a714b27f10ba2

core/embed/sys/smcall/stm32/smcall_probe.c
cepetr: fixed, see 09dedd09653c6494a3c20256b8f7c97abfdd3130

core/embed/sys/syscall/inc/sys/syscall_ipc.h
cepetr: fixed, see 6cc733c9d74ad990914287f00f7b27390fad5bb4

core/embed/util/flash/stm32u5/flash.c
cepetr: yes, that's weird, but it probably doesn't mean anything. I added a comma at the end of the array and it looks fine now. See 782791f09ba409f8d337374b4782a187a976e857

core/site_scons/models/D002/discovery2.py
cepetr: Yes, I enabled it for D002 too. See 1dfd09abbd7d3ea9f00846801d08b40fa2b63eec

core/embed/sys/smcall/stm32/smcall_dispatch.c
cepetr: I  don't think it's necessary, but to be on the safe side, I've added the check anyway. The secure monitor should use the BLXNS instruction, which checks if the address is non-secure and triggers a secure fault exception if it isn't.

see e6568cae6d3dc571858509ced01383b7dc2b3f46 and 48a047c0d75597567431ad801d6556d57fda6556

core/embed/sys/trustzone/stm32u5/trustzone.c
cepetr: i fixed the comment, see 6684370fd0405498163b4e0d33fe2176bc4fe986

core/embed/projects/bootloader/main.c
cepetr: Yes, you're right. But isn't this `#ifdef` actually correct, assuming that `FANCY_FATAL_ERROR` can theoretically be disabled in the bootloader?

TychoVrahe: i theoretically could be, i just see no point in doing that, the `FANCY_FATAL_ERROR` is defined in bootloader sconscript, so its not even model dependent or anything - and bootloader will always have GUI implementation so to me this seems entirely redundant. The ifdefs are essential for shared modules, but i don't see the point here

cepetr: Okay, I don't have a strong opinion on it. But i agree the code is simpler without the #ifdefs. So I changed it, see 44d4c05300fcd99f4eb1f85f8934ec8c98af4d6d

TychoVrahe (approved):

PR #5121, opened 2025-05-30
improved battery precharge 

kopecdav:
This PR improves prodtest pm-precharge command.

Instead of charging to fixed voltage, now you can use and argument to precharge the battery to a specific SoC.
voltage for given SoC is extracted directly from the charging battery model.

When the battery is precharged, charging is disabled so the battery charge will remain.

PR also get rid of the power manager SoC limit which should no longer be needed.


kopecdav:
Closing for now

PR #5122, opened 2025-05-30
enable multiple BLE connections

TychoVrahe:
This PR allows two BLE connections simultaneously, in order to enable easy connection switching.

After a new (already bonded) device connect, the previous connection is immediately terminated. 

Advertising while a device is connected is only allowed if 2 or more devices are bonded, otherwise no advertising while connected.


Hannsek:
> Advertising while a device is connected is only allowed if 2 or more devices are bonded, otherwise no advertising while connected.

Why is that?

TychoVrahe:
@Hannsek Why would you want that? no other device can connect as the only one bonded is connected, and it consumes more energy to advertise

Hannsek:
Oh, sorry, I thought the advertising is actually "pairing mode".

core/embed/io/ble/inc/io/ble.h
```diff
@@ -77,8 +78,10 @@ typedef enum {
   BLE_CONNECTED = 1,          // Connected to a device
   BLE_DISCONNECTED = 2,       // Disconnected from a device
   BLE_PAIRING_REQUEST = 3,    // Pairing request received
-  BLE_PAIRING_CANCELLED = 4,  // Pairing was cancelled by host
+  BLE_PAIRING_CANCELLED = 4,  // Pairing was canceled by host
   BLE_PAIRING_COMPLETED = 5,  // Pairing was completed successfully
+  BLE_CONNECTION_CHANGED =
+      6,  // Connection change (e.g. different device connected)
```
mmilata, line 84: I'm wondering if we can guarantee that nRF and STM32 agree on which data packet belongs to which connection, especially those that are in queues when this event comes? Does the application firmware need to send some kind of `BLE_CONNECTION_CHANGED_ACK` command before nRF starts handling the new connection? Or we can have address at the start of every packet.

TychoVrahe: i added the address to data packets in https://github.com/trezor/trezor-firmware/commit/93539bc2e9d0fb67c1c7481f655a8e2cab216059

nrf checks the address when it receives data from STM - but this is still not good enough, as on STM side the address is stored in BLE driver, but it should probably go to app and back to guarantee consistency. But that likely leads to larger changes in application/connection handling.

core/embed/io/ble/stm32/ble.c
```diff
@@ -172,13 +180,9 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
...
     return;
   }
-  if (len < sizeof(event_status_msg_t)) {
-    // insufficient data length
-    return;
-  }
 
   event_status_msg_t msg;
-  memcpy(&msg, data, sizeof(event_status_msg_t));
+  memcpy(&msg, data, MIN(sizeof(event_status_msg_t), len));
```
cepetr, line 185: if `len < sizeof(event_status_msg_t)`, then the `msg` structure may be partially uninitialized. What about clearing it by declaring:
```
event_status_msg_t msg = {0};
```

core/embed/io/ble/stm32/ble.c
```diff
@@ -638,11 +695,17 @@ uint32_t ble_read(uint8_t *data, uint16_t max_len) {
 
   uint16_t read_len = 0;
 
-  tsqueue_dequeue(queue, data, max_len, &read_len, NULL);
+  uint8_t rx_data[BLE_DATA_SIZE];
+
+  tsqueue_dequeue(queue, rx_data, sizeof(rx_data), &read_len, NULL);
+
+  // first 7 bytes is MAC, ignore for now
+  memcpy(data, &rx_data[BLE_DATA_HEADER_SIZE],
```
cepetr: Not sure about this, but if we receive fewer than `BLE_DATA_HEADRER_SIZE` bytes, then `MIN(read_len - BLE_DATA_HEADER_SIZE, max_len)` could evaluate to a negative value.

core/embed/io/ble/stm32/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/94080ceb76a0288e680a704e723e0730daaedc10

core/embed/io/ble/stm32/ble.c
TychoVrahe: such data shouldn't be enqueued in first place, see `ble_process_data`. That said, i still added an explicit check into the reading function: https://github.com/trezor/trezor-firmware/commit/94080ceb76a0288e680a704e723e0730daaedc10

PR #5124, opened 2025-06-01
ci(core): disable authenticity tests on HW CI

romanz:
T2B1 device seems to have a locked bootloader so #5087 is not enough.
https://github.com/trezor/trezor-firmware/actions/runs/15368486126/job/43244403997#step:8:19


romanz:
Tested at https://github.com/trezor/trezor-firmware/actions/runs/15373743145/job/43256175098.

mmilata (approved):

PR #5125, opened 2025-06-01
fix(core): Eckhart doesn't have a separate confirm page

romanz:
Found when debugging a flaky test in https://github.com/trezor/trezor-firmware/pull/5115 (due to the extra debuglink click sent).


core/embed/rust/src/ui/layout_eckhart/flow/get_address.rs
```diff
@@ -111,7 +111,7 @@ pub fn new_get_address(
             TextScreenMsg::Menu => Some(FlowMsg::Info),
         })
         .one_button_request(ButtonRequest::from_num(br_code, br_name))
-        .with_pages(|address_pages| address_pages + 1);
+        .with_pages(|address_pages| address_pages);
```
bieleluk: Is this line needed at all? Cannot we just get rid of it if we do not want to remap the page number?

romanz: Thanks! 
Removed in 497a2d2f7e.

bieleluk (approved):

PR #5132, opened 2025-06-02
fix pm_get_state syscall stub

TychoVrahe:


core/embed/sys/syscall/stm32/syscall_stubs.c
```diff
@@ -695,7 +695,7 @@ pm_status_t pm_hibernate(void) {
   return (pm_status_t)syscall_invoke0(SYSCALL_POWER_MANAGER_HIBERNATE);
 }
 
-pm_status_t pm_get_status(pm_state_t *status) {
+pm_status_t pm_get_state(pm_state_t *status) {
```
obrusvit: The parameter can be named `state`

TychoVrahe: yep https://github.com/trezor/trezor-firmware/commit/d02b13cc8bed745d35c2f35e8762aab3bf2761df

PR #5135, opened 2025-06-03
fix(core): close previous homescreen before changing it

romanz:
Otherwise, it will fail due to double `ImageBuffer` locking.
    
Also, add a test to reproduce the locking issue.

### DIFF
--- a/core/src/trezor/ui/layouts/delizia/__init__.py
+++ b/core/src/trezor/ui/layouts/delizia/__init__.py
@@ -152,6 +152,13 @@ def confirm_multisig_different_paths_warning() -> Awaitable[None]:
 def confirm_homescreen(
     image: bytes,
 ) -> Awaitable[None]:
+
+    from trezor import workflow
+
+    # Closing current homescreen workflow unlocks internal ImageBuffer,
+    # in order to display the new homescreen image.
+    workflow.close_others()
+
     return raise_if_not_confirmed(
         trezorui_api.confirm_homescreen(
             title=TR.homescreen__title_set,
--- a/tests/device_tests/test_msg_applysettings.py
+++ b/tests/device_tests/test_msg_applysettings.py
@@ -261,6 +261,10 @@ def test_apply_homescreen_jpeg_single_message(client: Client):
             client.call(messages.ApplySettings(homescreen=b""))
             client.refresh_features()
 
+            client.set_expected_responses(EXPECTED_RESPONSES_NOPIN)
+            client.call(messages.ApplySettings(homescreen=img))
+            client.refresh_features()
+
 
 @pytest.mark.models(skip=["legacy", "safe3"])
 def test_apply_homescreen_jpeg_progressive(client: Client):


romanz:
https://github.com/trezor/trezor-firmware/blob/d65b4f0e869a02006084f34f8ca7cebbb7c80202/core/.changelog.d/1120.added#L1
#5015 was not released in 2.8.10, so removing the changelog entry in https://github.com/trezor/trezor-firmware/pull/5135/commits/0107666237d6f80cf42fc711c7dc87f109f81f1e.

mmilata:
Oh, good catch. It is seems related to #5015 but it's actually caused by #5089. I asked @obrusvit to remove `close_others` because it was not clear why it was there and it ended up [copypasted to Eckhart](https://github.com/trezor/trezor-firmware/pull/4775#discussion_r2102630035).

I think image buffer locking is model-dependent implementation detail - can you please move `close_others` back under trezor.ui.layouts and add a fat warning why it is there?😅

romanz:
https://github.com/trezor/trezor-firmware/pull/5135/commits/d5c5b255ae1f686f6eefb7de8a130e8c3035b57f reverted #5089 and added a comment at `core/src/trezor/ui/layouts/delizia/__init__.py`.

obrusvit:
Thank you, should we `close_others()` also in the corresponding function for Eckhart, i.e. `core/src/trezor/ui/layouts/eckhart/__init__.py`?

romanz:
> should we `close_others()` also in the corresponding function for Eckhart?

IIUC, only Delizia uses `CachedJpeg` and `ImageBuffer` in its homescreen flows, so I think this issue doesn't happen in Eckhart:
https://data.trezor.io/dev/firmware/ui_report/15421761191/T3W1-en-core_device_test/passed/T3W1_en-device_tests-test_msg_applysettings.py%3A%3Atest_apply_homescreen_jpeg_single_message.html

obrusvit:
@romanz you're absolutely right. Although later on we may want to implement it similarly as Delizia also for Eckhart. We will know  better :)

mmilata (approved):

bosomt:
**QA OK**

- tried to load custom homescreen into device multiple times via trezorctl ✅ 
- unloaded and loaded homescreen via Suite and then changed it to custom homescreen ✅ 

**Info:**
 - Suite version: desktop 25.7.0 (4c4289cf9e7e4c3cf217f7974565ed818697bd64)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.7.0 Chrome/134.0.6998.178 Electron/35.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.9.0 regular (revision f906a91ebf561e99a9449b6f8082cfc123bbce09)

Hannsek:
Is this tested on trezorctl as well @bosomt?

bosomt:
@Hannsek yes i did, updated my comment 

```
➜  ~ history | grep moon
 1087  trezorctl set homescreen tothemoon.jpg
 1129  trezorctl set homesecreen tothemoon.jpg
 1131  trezorctl set homescreen tothemoon.jpg
 1408  trezorctl set homescreen tothemoon.jpg
➜  ~
```

PR #5154, opened 2025-06-06
Eckhart UI: remove hold to lock

obrusvit:

### DIFF
--- a/tests/click_tests/test_lock.py
+++ b/tests/click_tests/test_lock.py
@@ -49,7 +49,7 @@ def test_hold_to_lock(device_handler: "BackgroundDeviceHandler"):
         models.T3B1: 1200,
         models.T2T1: 3500,
         models.T3T1: 3500,
-        models.T3W1: 2000,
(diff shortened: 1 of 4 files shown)


tests/click_tests/test_lock.py
```diff
@@ -49,7 +49,7 @@ def test_hold_to_lock(device_handler: "BackgroundDeviceHandler"):
         models.T3B1: 1200,
         models.T2T1: 3500,
         models.T3T1: 3500,
-        models.T3W1: 2000,
+        models.T3W1: 3500,
```
bieleluk, line 52: nit: the `LOCK_HOLD_DURATION` is 300ms, not 3500ms. If the margin is intentional, it should be explicitly mentioned.

obrusvit: IIUC, the time in the test is intentionally slightly longer than the hold to lock duration. It's set that way also e.g. for `T3T1`.

PR #5155, opened 2025-06-06
bootloader: always allow reboot in menu

TychoVrahe:


core/embed/projects/bootloader/.changelog.d/5155.changed
```diff
@@ -0,0 +1 @@
+Always allow reboot from bootloader menu.
```
cepetr, line 1: Doesn't this PR modify the bootloader behavior only on T3W1? Do we need this changelog entry?

TychoVrahe: no all core models are affected

Issue #5148, opened 2025-06-06
Stellar: if `timebounds` are not set in the tx, then omit displaying it.


overcat:
Currently, for a simple send operation, it requires at least seven clicks on Model T. Can we avoid displaying it when the tx does not have timebounds set to reduce the number of clicks needed?

<img width="193" alt="Image" src="https://github.com/user-attachments/assets/c4f722be-f2b9-4ee2-81e7-bf16ec386cfa" />

overcat:
If approved, I will submit a PR. 

I haven't submitted a PR here in a long time, so I want to start with a small PR to refresh my memory of the complete process.

Hannsek:
Hello, thanks for this. Let's narrow the UX. I am not really familiar with what needs to be displayed or not in the Stelar ecosystem. The basic send flow on the device should follow this structure:

`Recipient address`
`Amount`
`Summary - total amount + fee`

<img width="97" alt="Image" src="https://github.com/user-attachments/assets/c7c833a8-7269-438b-a2ec-de7ed0247c44" />

All other details (fee info, account info, blockhash (for Solana), …) should go under the on-device `menu`:

![Image](https://github.com/user-attachments/assets/6c7b0881-bb26-4dbf-b8d6-a45f9eef06dd)

![Image](https://github.com/user-attachments/assets/ea3f540a-3acb-4e01-9557-dc2bc1f5cfce)

I am happy to discuss details of Stellar on a call so it is clear for both sides.

overcat:
Hi @Hannsek, thank you for this information. I will take some time to see if I can align Stellar's process as closely as possible with the one in the image above.

Hannsek:
Ok, thank you. Can you tell what are all the information needed to be displayed on Trezor screen?

overcat:
Hi @Hannsek 

For a Payment transaction, the message consists of the following two parts:

* https://github.com/trezor/trezor-firmware/blob/9fc0c5d82b16f205e20bec6efceb91189dcac59d/common/protob/messages-stellar.proto#L45-L72
* https://github.com/trezor/trezor-firmware/blob/9fc0c5d82b16f205e20bec6efceb91189dcac59d/common/protob/messages-stellar.proto#L45-L72

Below is my explanation for each field:

### `StellarSignTx`

* `address_n`: Used to determine the Trezor account that will sign the transaction. (This should be displayed in `account_items`.)
[...]

PR #5159, opened 2025-06-09
[Draft] refactor(core): optimize the signing process of Stellar transactions.

overcat:
For #5148 review, not yet ready to merge, and does not fully comply with contribution guidelines.


overcat:
A Few Questions
1. Do I need to manually add translation files? I'm guessing not, but when do we typically update those?

2. Do I need to manually update `./tests/ui_tests/fixtures.json`? It seems like `./tests/update_fixtures.py` might not work in my own repository. Perhaps I'll only be able to use it after the CI runs in this repository?

Hi @matejcik, could you please help me with this? I haven't contributed code here for a long time.

Hannsek:
cc @obrusvit

matejcik:
> Do I need to manually add translation files? I'm guessing not, but when do we typically update those?

you need to add English strings. you can ignore the other languages

> Do I need to manually update `./tests/ui_tests/fixtures.json`? It seems like `./tests/update_fixtures.py` might not work in my own repository. Perhaps I'll only be able to use it after the CI runs in this repository?

nobody does that manually.
if you run tests e.g. via `make test_emu_ui_multicore` in core directory, you should be able to run `tests/update_fixtures.py local` to update them. if there's a problem with that, please report what you see.

core/src/trezor/ui/layouts/bolt/__init__.py
```diff
@@ -711,10 +713,10 @@ def confirm_address(
     return confirm_value(
         title,
         address,
-        description or subtitle or "",
+        description or "",
         br_name or "confirm_address",
         br_code,
-        subtitle=None,
+        subtitle=subtitle,
```
overcat, line 775: I've made some adjustments to meet Stellar's requirements, but this has impacted certain interfaces for Ethereum and Solana.

Solana before and after:
Solana before
Solana after

Ethereum before and after:

Ethereum before
Ethereum after


I'm not sure why we're not consistent with delizia's implementation here, especially since the `subtitle` isn't ignored there.

https://github.com/trezor/trezor-firmware/blob/5194e1e10b05a215dd4b3046e213fc7687a06164/core/src/trezor/ui/layouts/delizia/__init__.py#L635-L655

core/src/apps/stellar/layout.py
```diff
@@ -85,38 +38,65 @@ async def require_confirm_memo(memo_type: StellarMemoType, memo_text: str) -> No
...
+            TR.stellar__valid_to,
+            (
+                strings.format_timestamp(timebonuds_end)
+                if timebonuds_end > 0
+                else TR.stellar__no_restriction
+            ),
+        ),
+    ]
+
+    account_name = paths.get_account_name("Stellar", address_n, PATTERN, SLIP44_ID)
```
overcat, line 90: It is worth confirming whether to use `Stellar #1` or `XLM #1`. In the `getAddress` interface, it uses `XLM #1`, while in the sign tx interface, it uses `Stellar #1`. The Trezor Suite interface also uses `Stellar #1`. I think we should keep it consistent; I prefer `Stellar #1`. If you agree, I will modify the display in `getAddress`.

- Trezor suite
    image

- get address page
    image

- sign tx page
    image




https://github.com/trezor/trezor-firmware/blob/5194e1e10b05a215dd4b3046e213fc7687a06164/core/src/apps/stellar/get_address.py#L32

core/src/trezor/ui/layouts/bolt/__init__.py
Hannsek: This is fine. Thanks for pointing this out.

core/src/apps/stellar/layout.py
Hannsek: Yes, agree, thank you.

overcat: 71b7c59bde27c2d5c27debc2bd25aa77a2492597

Renamed to `Stellar`.

obrusvit:
Hello @overcat, I rebased on top of `main` and force-pushed so that CI can run.

- I did minor fixes in https://github.com/trezor/trezor-firmware/pull/5159/commits/d79e5c62f12a16d4079d55cc65fe6a3ecd925091
- It's possible that `ui_fixtures.json` will have to be updated again, I will take care of it.

From the discussion in the issue, the last thing remaining in the UI is the cancellation from the warning screen, am I right? I'll take a look whether we can do it easily.

Thank you.

obrusvit:
One more thing @overcat, do you think you could provide `trezorctl` commands which trigger the usecases shown in your videos? Thank you.

overcat:
Hi @obrusvit
> From the discussion in the issue, the last thing remaining in the UI is the cancellation from the warning screen, am I right? I'll take a look whether we can do it easily.

This should apply to all alert pages, not just Stellar's. So, would it be better to do this in a separate PR?


> do you think you could provide trezorctl commands which trigger the usecases shown in your videos? Thank you.


Here's some test data; you can find more in https://github.com/trezor/trezor-firmware/blob/main/common/tests/fixtures/stellar/sign_tx.json (check `xdr` field).

```
trezorctl device load -m 'all all all all all all all all all all all all'
# Source not equals Trezor Account (show warning page)
trezorctl stellar sign-transaction 'AAAAAgAAAADTSPIr3MGTm9sMI5oJJ/GuXE2KFkCxLma1cUfqm1LKzQAAAGQAAAAAAAAD6AAAAAEAAAAAG4J3zQAAAABd5CqEAAAAAAAAAAEAAAAAAAAAAQAAAABdVWQkZrGFuEMVLp4hkVHbxYkgJ+xAEBpRe+1coDDC4AAAAAAAAAAAHc8WmAAAAAAAAAAA'
# Payment with non-native asset
trezorctl stellar sign-transaction 'AAAAAgAAAAAvIrnGLwi3dPPr5t1ufbk8PsLL3gJ5Vho9nFIluMMikgAAAGQAAAAAAAAD6AAAAAEAAAAAG4J3zQAAAABd5CqEAAAAAAAAAAEAAAAAAAAAAQAAAABdVWQkZrGFuEMVLp4hkVHbxYkgJ+xAEBpRe+1coDDC4AAAAAFYAAAAAAAAACmElgLCDlg2XxI8Eth6HKRVDRDNIbCuDhjUTj5W/WoVAAAAAB3PFpgAAAAAAAAAAA=='
# All Send-like operations
trezorctl stellar sign-transaction 'AAAAAgAAAAAvIrnGLwi3dPPr5t1ufbk8PsLL3gJ5Vho9nFIluMMikgAAAfQAAAAAAAAD6AAAAAEAAAAAG4J3zQAAAABd5CqEAAAAAAAAAAUAAAAAAAAAAAAAAABdVWQkZrGFuEMVLp4hkVHbxYkgJ+xAEBpRe+1coDDC4AAAAAAF9eEAAAAAAAAAAAEAAAAAXVVkJGaxhbhDFS6eIZFR28WJICfsQBAaUXvtXKAwwuAAAAABWAAAAAAAAAAphJYCwg5YNl8SPBLYehykVQ0QzSGwrg4Y1E4+Vv1qFQAAAAAdzxaYAAAAAAAAAA0AAAABWAAAAAAAAABwi6oxX35cFm2EtGS/s4/WJXj+OtJyJ+dsy7ehecRRIQAAAAAdzxaYAAAAAF1VZCRmsYW4QxUuniGRUdvFiSAn7EAQGlF77VygMMLgAAAAAkFCQ0RFRkdISUpLTAAAAAAphJYCwg5YNl8SPBLYehykVQ0QzSGwrg4Y1E4+Vv1qFQAAAAAAAeJAAAAAAAAAAAAAAAACAAAAAVgAAAAAAAAAcIuqMV9+XBZthLRkv7OP1iV4/jrScifnbMu3oXnEUSEAAAAAHc8WmAAAAABdVWQkZrGFuEMVLp4hkVHbxYkgJ+xAEBpRe+1coDDC4AAAAAJBQkNERUZHSElKS0wAAAAAKYSWAsIOWDZfEjwS2HocpFUNEM0hsK4OGNROPlb9ahUAAAAAAAHiQAAAAAAAAAAAAAAACAAAAABdVWQkZrGFuEMVLp4hkVHbxYkgJ+xAEBpRe+1coDDC4AAAAAAAAAAA'
# Text Memo
trezorctl stellar sign-transaction 'AAAAAgAAAAAvIrnGLwi3dPPr5t1ufbk8PsLL3gJ5Vho9nFIluMMikgAAAGQAAAAAAAAD6AAAAAEAAAAAG4J3zQAAAABd5CqEAAAAAQAAAA1IZWxsbywgd29ybGQhAAAAAAAAAQAAAAAAAAABAAAAAF1VZCRmsYW4QxUuniGRUdvFiSAn7EAQGlF77VygMMLgAAAAAAAAAAAdzxaYAAAAAAAAAAA='
```

Hannsek:
> This should apply to all alert pages, not just Stellar's. So, would it be better to do this in a separate PR?

I thought it is an issue only with this pr, but as you said, it is problem with all warning screens. So, let's not block this pr with this.

ibz:
Rebased on `main` and force pushed...

ibz:
BTW, I am taking this over so we can get unstuck. Thank you, @overcat for the great job!!!

After rebasing and dealing with conflicts, I am doing a final review - including a review of the UI changes.

I have to say that generally the UI looks much better than before!

In some cases it feels like a slight downgrade, for example:

image

Was replaced with:

image

I honestly liked seeing the fact that an account is created, not just a generic `Recipient #1` - what do you think @Hannsek ? Should I in this case show `Create account #1` instead of `Recipient #1` in the title?

Another question, for @overcat

image

Is this really a "maximum fee" or is it an exact fee amount that we are dealing with? Since we changed the design, I would not have expected a change in the semantics... The two are not the same!

But generally the UI feels more consistent and in line with other coins, so I think having this merged will be a plus.

PS: I will look into the warning cancellation after this gets merged, since it is a separate issue...

overcat:
Hi @ibz, it's `maximum fee`. See https://stellar.org/blog/developers/surge-pricing-on-stellar-faq

> CAP5 introduces “dutch auction” fee payments which ensure that the operation fee you submit is the maximum you will pay and the actual amount is the lowest possible.

core/src/apps/stellar/layout.py
```diff
@@ -85,38 +38,67 @@ async def require_confirm_memo(memo_type: StellarMemoType, memo_text: str) -> No
...
-        + TR.stellar__sign_tx_fee_template
-    )
-    await layouts.confirm_metadata(
-        "confirm_final",
-        TR.stellar__final_confirm,
-        text,
+async def require_confirm_final(
+    address_n: list[int],
+    fee: int,
+    timebonuds: tuple[int, int],
```
ibz: Renamed to `timebounds`

ibz (approved):
LGTM.

I removed some strings that are now unused.

One thing to note is that we used to warn when using `testnet` for a transaction, now we don't.

I suppose it is not a risk of losing funds, because at worst what happens is somebody sends testnet tokens without realizing they are testnet... not the other way around.

ibz:
I will squash, force push, and get it ready to be merged...

overcat:
> One thing to note is that we used to warn when using `testnet` for a transaction, now we don't.
> 
> I suppose it is not a risk of losing funds, because at worst what happens is somebody sends testnet tokens without realizing they are testnet... not the other way around.

Yes, this design has been clarified [here](https://github.com/trezor/trezor-firmware/issues/5148#issuecomment-2975448688 ).

ibz:
> has been clarified

Right, I missed that part at the bottom about the testnet, but seems it's all clear. It was indeed an unnecessary complication for no added benefit.

STew790:
**QA OK**
Tested in Suite https://github.com/trezor/trezor-suite/issues/21105#issuecomment-3242824932

PR #5160, opened 2025-06-09
Display additional details on RSOD screen

cepetr:
This PR adds several improvements to the RSOD screen to aid in troubleshooting firmware crashes:

1. The `systask_postmortem_info_t` structure has been extended with:
- SCM revision (8 bytes) – currently unused
- PC (program counter) – contains the address of the instruction that caused the fault

2. The RSOD message on Bolt is now aligned the same way as on Delizzia, allowing up to three lines of text to be displayed.

3. For all ARM faults, the RSOD now displays the PC value, which can be used with the `addr2line` utility to identify the corresponding source file and line number.

Screenshots for T2T1, T3B1:

```c
*((volatile uint32_t *)0) = 0;
```

5. For all fatal errors (triggered by `assert()` and `ensure()` macros), the RSOD displays the full message along with the filename and line number. This information was already embedded in the firmware, so the overall size has not increased.

Screenshots for T2T1, T3B1, T3T1:

```c
ensure(secfalse, "init failed"); // in main.c:330
```

Although I initially had bigger ambitions, I ultimately decided on a minimalistic version that doesn’t require UI changes (e.g., displaying more information on the RSOD screen) or significantly increase code size (e.g., showing the location of panics in Rust

### DIFF
--- a/core/embed/sys/trustzone/stm32u5/trustzone.c
+++ b/core/embed/sys/trustzone/stm32u5/trustzone.c
@@ -130,8 +130,8 @@ static void tz_configure_arm(void) {
   // system resets from non-secure code (e.g., during debugging).
   reg_value |= SCB_AIRCR_SYSRESETREQS_Msk;
 #endif
-  // NMI, BusFault, HardFault are handled in both secure and non-secure worlds
-  reg_value |= SCB_AIRCR_BFHFNMINS_Msk;
+  // NMI, BusFault, HardFault are handled only in secure world
+  reg_value &= ~SCB_AIRCR_BFHFNMINS_Msk;
   SCB->AIRCR = reg_value;
 }
 


Hannsek:
Can you please post here some images?

cepetr:
> Can you please post here some images?

I added images for T2T1 to the description. On T3T1, it's very similar - only the font is slightly different. On T3B1 the information is the same and fits within the reserved space.

Hannsek:
Thank you. Please post all models if you can. But no need to rush.

core/SConscript.kernel
```diff
@@ -206,6 +206,7 @@ SOURCE_MOD += [
     'embed/util/translations/translations.c',
     'embed/util/image/image.c',
     'embed/util/rsod/rsod.c',
+    'embed/util/scm_revision/scm_revision.c',
```
TychoVrahe: we probably can't do this - scm revision is part of authenticity check and its supposed to be hard to find in binary so that fake bootloader/firmware cannot pretend new firmware was installed by providing correct scm revision. in firmware, the object files are shuffled so that the location of scm revision is randomized, but no such thing is done here, even if it was we might not want to duplicate the string as might make finding it easier. Applies to secmon too.

core/embed/sys/task/stm32/systask.c
```diff
@@ -691,75 +762,80 @@ __attribute__((naked, no_stack_protector)) void MemManage_Handler(void) {
...
+  );
+}
 #endif
 
-void NMI_Handler(void) {
+__attribute__((no_stack_protector, used)) static void nmi_handler(
+    uint32_t msp, uint32_t exc_return) {
   mpu_mode_t mpu_mode = mpu_reconfig(MPU_MODE_DEFAULT);
 #ifdef STM32U5
   if ((RCC->CIFR & RCC_CIFR_CSSF) != 0) {
```
TychoVrahe: didn't we discuss we don't actually need these conditions? or did you discover a reason for it to stay?

Hannsek:
All good from my side. Thanks.

core/SConscript.kernel
cepetr: Yes, you are right. I completely removed this, since the feature it's not used now. See 50c73c730c097f88f870745bb1ba19954008a077

core/embed/sys/task/stm32/systask.c
cepetr: Yes, i forget to do it. I changed the behavior in 3ddf0711ef5a243501c17d3454b915ab662f403b and I also moved nmi/hard fault/bus fault processing to secure monitor, see 94263946537d69b974e13bebc14cead73d526745,

TychoVrahe (approved):

PR #5161, opened 2025-06-09
unify CommonUI screen definition

TychoVrahe:
This PR is minor refactoring of CommonUI screen definitions are moved to same place across different UI implementations  - now they are in layout_xxx/mod.rs like it was for bolt, and they all use simplified:show. 

This also fixes Eckhart problem of the boot stage 2 screen not being shown (as fade in parameter was ignored).


obrusvit (approved):
utACK

PR #5166, opened 2025-06-09
fix(core): gracefully close debuglink session

### LINKED ISSUE #4401: Debuglink response can get cut off if a workflow ends

romanz:
This PR should solve https://github.com/trezor/trezor-firmware/issues/4401 by introducing an event-based synchronization, following https://github.com/trezor/trezor-firmware/pull/5091#issuecomment-2954881542 and https://github.com/trezor/trezor-firmware/pull/5115#pullrequestreview-2881386515.

Before restarting the session, the wire protocol loop will "interrupt" debuglink's `read_message()` and wait for debuglink's task completion.

### DIFF
--- a/core/src/trezor/loop.py
+++ b/core/src/trezor/loop.py
@@ -571,3 +571,48 @@ class spawn(Syscall):
         is True, it would be calling close on self, which will result in a ValueError.
         """
         return self.task is this_task
+
+
+class event(Syscall):
+    """Single-shot event for synchronization across tasks."""
+
+    _EMPTY = object()
+
+    def __init__(self) -> None:
+        self.result = event._EMPTY
+        self.callback: Task | None = None
+
+    def is_set(self) -> bool:
+        """Returns `true` if `set()` was called."""
+        return self.result is not self._EMPTY
+
+    def set(self, result: Any = None) -> None:
+        """Mark this event as done."""
+        assert not self.is_set()
+        self.result = result
+
+        if self.callback is not None:
+            # invoke waiting task
+            task = self.callback
+            self.callback = None
+            _step(task, result)
+
+    def handle(self, task: Task) -> None:
+        assert self.callback is None
+        if self.is_set():
+            _step(task, self.result)
+        else:
+            # register the callback to bet invoked when `set()` is called
+            self.callback = task
+
+    def __iter__(self) -> Task:
+        assert self.callback is None
+
+        if self.is_set():
+            # exit immediately if event was set
+            return self.result
+
+        try:
+            return (yield self)
+        finally:
+            self.callback = None
--- a/core/src/apps/debug/__init__.py
+++ b/core/src/apps/debug/__init__.py
@@ -14,6 +14,7 @@ if __debug__:
     from trezor.enums import DebugWaitType, MessageType
     from trezor.messages import Success
     from trezor.ui import display
+    from trezor.wire.codec.codec_context import CodecContext
 
     if TYPE_CHECKING:
         from typing import Any, Awaitable, Callable
@@ -31,13 +32,14 @@ if __debug__:
             DebugLinkState,
         )
         from trezor.ui import Layout
-        from trezor.wire import WireInterface, context
+        from trezor.wire import WireInterface
 
         Handler = Callable[[Any], Awaitable[Any]]
 
     layout_change_box = loop.mailbox()
 
-    DEBUG_CONTEXT: context.Context | None = None
+    DEBUG_CONTEXT: CodecContext | None = None
+    DEBUG_SESSION_TASK: loop.spawn | None = None
 
     REFRESH_INDEX = 0
 
@@ -381,8 +383,7 @@ if __debug__:
 
     async def handle_session(iface: WireInterface) -> None:
         from trezor import protobuf, wire
-        from trezor.wire.codec import codec_v1
-        from trezor.wire.codec.codec_context import CodecContext
+        from trezor.wire.codec import Interrupt, codec_v1
 
         global DEBUG_CONTEXT
 
@@ -402,6 +403,9 @@ if __debug__:
                     log.exception(__name__, exc)
                     await ctx.write(wire.failure(exc))
                     continue
+                except Interrupt:
+                    log.debug(__name__, "closing debuglink session")
+                    return
 
                 req_type = None
                 try:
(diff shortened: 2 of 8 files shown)


core/src/trezor/loop.py
```diff
@@ -571,3 +571,48 @@ def is_running(self) -> bool:
         is True, it would be calling close on self, which will result in a ValueError.
         """
         return self.task is this_task
+
+
+class event(Syscall):
+    """Single-shot event for synchronization across tasks."""
```
matejcik, line 577: could we use `mailbox` instead?

romanz: We'll probably need to modify `mailbox` a bit, to allow it to become "sticky" (so that it won't be cleared after the being set, similar to [Python's `threading.Event`](https://docs.python.org/3/library/threading.html#event-objects)). WDYT?

romanz: P.S.
I was using the "stickiness" here:
https://github.com/trezor/trezor-firmware/blob/7d4de7bbf995202eb4216f738fa138d655dbb0f4/core/src/trezor/wire/codec/codec_v1.py#L41-L45
since it may happen that both `read` and `init_interrupt` are ready - and in this case we don't want to read anything from DebugLink.

matejcik: you can `interrupt.put(SOME_SPECIAL_VALUE)` and test the return value of `race` for it

come to think of it, I _think_ you can `interrupt.put(INTERRUPT_EXCEPTION)` and get the exception straight out _raised_ out of the `race` call

romanz: You are right - but I think that it may have a problem if `read` and `init_interrupt` are ready (e.g. when there is debuglink message available, and wirelink loop is done and waiting to restart).

In this case `race(read, interrupt)` may:
1. return `read` result, instead of `interrupt` and
2. cancel `init_interrupt`'s mailbox - causing the session to get stuck :(
https://github.com/trezor/trezor-firmware/blob/8d1eb777fc75bf01658963ef447047ab1efabd24/core/src/trezor/loop.py#L286-L287

matejcik: if the `read` comes first, it will schedule resumption of the `read_message` task -- and cancel the `taker` task of the `init_interrupt`
so if you later call `init_interrupt.is_empty()` in place of `is_set()` you will get the correct result. 

(note that this is IMHO a rather ugly limitation of our event loop; that `race` allows other tasks to get scheduled between the completion of one of the sub-tasks. i tried to fix it once, and i noticed you making a PR that goes to `_step` directly, which you closed for some reason... i think it's vaguely the correct move, but there may be side effects that i forgot about since last time I tried something similar)
(note also that we intend to RiiR `loop.py` soonish :crab: so this will all change anyway)

romanz: > so if you later call init_interrupt.is_empty() in place of is_set() you will get the correct result.

IIUC, the `self.clear()` call at the end of `mailbox._take` seems to prevent me from using `mailbox.is_empty` :(

```python
    def put(self, value: Any, replace: bool = False) -> None:
        """Put a value into the mailbox.

        If there is another task waiting for the value, it will be scheduled to resume.
        Otherwise, the mailbox will hold the value until someone consumes it.

        It is an error to call `put()` when there is a value already held, unless
        `replace` is set to `True`. In such case, the held value is replaced with
        the new one.
        """
        if not self.is_empty() and not replace:
            raise ValueError("mailbox already has a value")

        self.value = value
        if self.taker is not None:
            self._take(self.taker)

    def _take(self, task: Task) -> None:
        """Take a value and schedule the taker."""
        self.taker = None
        schedule(task, self.value)
        self.clear()  # <<<<
```

> i noticed you making a PR that goes to _step directly, which you closed for some reason...

I was trying to work-around this "race condition" by invoking `_step` directly in `mailbox` and in `race`, but eventually I preferred to not to change the existing code too much, and instead added a new explicit synchronization mechanism.

> (note also that we intend to RiiR loop.py soonish 🦀 so this will all change anyway)

Cool!
So maybe it's better to postpone this PR after the rewrite?

matejcik: > IIUC, the `self.clear()` call at the end of `mailbox._take` seems to prevent me from using `mailbox.is_empty` :(

what would be the sequence of tasks where `_take` gets to run after the `race` is done?
because `race` should close the `_take` task before that happens, ISTM

romanz: I think that it reproduces with https://github.com/trezor/trezor-firmware/commit/bdf9478b49fb9debda12591699a4fca41db97c40:
```
$ core/emu.py -a -c pytest -vs tests/device_tests/test_basic.py::test_ping
...
awaiting init_read
racing: (<wait object at 0x7e324d8c64e0>, <mailbox object at 0x7e324d8c6600>)
    <generator object '__iter__' at 0x7e324d8ce4e0>
    <generator object '__iter__' at 0x7e324d8cea80>
...
135453130786 trezor.wire.message_handler INFO [WebUSB] received message: Initialize
...
135453131873 trezor.wire.codec.codec_context DEBUG [WebUSB] write: Features
...
[2025-06-19 11:14:13,749] trezorlib.debuglink DEBUG: sending message: DebugLinkResetDebugEvents
135453133437 apps.debug DEBUG interrupting debuglink session
mailbox.put(None)
mailbox._take(<generator object '__iter__' at 0x7e324d8cea80>) value=None
135453133477 trezor.loop DEBUG finish: <generator object '__iter__' at 0x7e324d8ce4e0>
closing race children, except <generator object '__iter__' at 0x7e324d8ce4e0>
closing race child: <generator object '__iter__' at 0x7e324d8cea80>
awaited init_read => 64 & interrupted => False
135453133521 apps.debug DEBUG 1 receive: <9007 - unknown message type>
135453133534 trezor.wire.codec.codec_context DEBUG [WebUSB] write: Success
```

1. wire session is over, so it puts `None` into DebugLink's `mailbox`
2. meanwhile, the host sends a DebugLink message (e.g. `DebugLinkResetDebugEvents` above)
3. since DebugLink is awaiting, `mailbox._take` is invoked (clearing `mailbox.value`) and schedules its taker task.
4. `race._finish` is called, which closes `mailbox`'s taker task (losing the value)
5. `race` returns `read` result, and continues to handle DebugLink message 
6. wirelink is stuck waiting for debuglink to exit

romanz:
> So maybe it's better to postpone this PR after the rewrite?

Closing for now.

PR #5167, opened 2025-06-10
remove obsolete bootloader antiglitch protection

### LINKED ISSUE #4805: Reevaluate anti-glitch protection in bootloader code
from https://github.com/trezor/trezor-firmware/pull/4572#discussion_r2003393950:

> are we doing full reset on all models?
> 
> doesn't that mean that we can avoid the antiglitch mess? this looks like security theatre: we are trying to get the caller to not glitch something somewhere to prevent them from doing what is (hopefully) a power cycle, no?
> (if i were to glitch something here, i'd want to skip the CALL instruction for the reset)
>
> the original anti-glitch code was based on the premise that we continue to firmware after running the complicated ui / firmware update / whatever code, AND that we can't fully trust the preceding code to leave the firmware in good state.
> this is, honestly, fixed to a large degree purely by including all the validation logic in real_jump_to_firmware -- the hypothetical attacker would have to glitch out the relevant verifications in there. And if the attacker can do that, they could have done that regardless of our antiglitch protections, simply by getting the flash to bad state and then rebooting themselves for the glitch attempt.
>
> if we're always doing a full reboot though, I'd say we don't want to "continue" to firmware ever. Or more precisely: "IF is_normal_non_interactive_boot THEN jump to code that continues to firmware ELSE run the workflows and then either reboot or die"

related specifically to this code: https://github.com/trezor/trezor-firmware/blob/42fac52da078ea0ac4547524c863cbe3967e008c/core/embed/projects/bootloader/main.c#L387-L396

TychoVrahe:
As we are now always rebooting mcu when jumping to firmware, this protection is no longer needed as all the checks will be repeated on reboot. in case skipping the reboot due to glitch, there are still integrity checks right before jumping to fw.

fixes #4805


core/embed/projects/bootloader/main.c
```diff
@@ -550,15 +548,6 @@ int bootloader_main(void) {
         firmware_present = sectrue;
```
cepetr: Now, we don't need to set these flags anymore.

core/embed/projects/bootloader/main.c
```diff
@@ -550,15 +548,6 @@ int bootloader_main(void) {
...
         firmware_present_backup = sectrue;
       case WF_OK_REBOOT_SELECTED:
```
cepetr, line 555: and maybe we can also slightly modify the code at the end of `main()` to prevent jumping to the firmware without reboot:
```c
 ...

#ifdef USE_STORAGE_HWKEY
        secret_bhk_regenerate();
#endif
        ensure(erase_storage(NULL), NULL);
        error_shutdown("Bootloader fatal error");
      }
    }
  } else {   <----- this else branch
    ensure(dont_optimize_out_true * (firmware_present == firmware_present_backup),
           NULL);
    if (sectrue == firmware_present) {
      firmware_jump_fn = real_jump_to_firmware;
    }
    firmware_jump_fn();
  }
  
  return 0;
}
```

core/embed/projects/bootloader/main.c
TychoVrahe: removed: https://github.com/trezor/trezor-firmware/commit/26b4a1e442dc50cb41bc1cb3302bcd2f66f5b3d2

core/embed/projects/bootloader/main.c
TychoVrahe: this will very likely be optimized away by compiler as now all the branches in the above switch end with no-return functions, rebooting or rsods - so i dont think there is any security benefit 

i did put it it anyway in https://github.com/trezor/trezor-firmware/commit/9cb401aa1b52cb1cd4bd2883625be243c82c6dd8 as it makes the above explicit and prevents adding more branches that would violate this. 

maybe we could also consider ending the main with error_shutdown instead of return 0?

cepetr: We should never get here, but yes - perhaps it would be better.  We can now also replace `return 0` by break on line 570.

TychoVrahe: changed in https://github.com/trezor/trezor-firmware/commit/4d6473a541160960323e360d69e316ce12fd3138

for consistency, i added the break also to the default branch, but neither break is needed, strictly speaking

matejcik:
we got a review comment from @PrisionMike on the original issue but it got lost, I asked him to resubmit it.
something to do with moving one of the checks into a different place instead of removing it.

PrisionMike:
Hey! Just to add a few points from the time I reviewed the code:
1. The referred code does seem redundant at that location. There are no plausible reasons to justify protecting the reboot routine. But,
2. It is strongly recommended to add the integrity check just before the final jump instruction at l241. This is the critical jump and the spread between the last integrity check and this instruction is large.

Other less pressing but still important measures could be:

3. Moving the random delay from the start to the *real_jump* . This will ensure that no side channel analysis would leak the exact program location of the sensitive jump.
(Optional) 
4. The _ensure_ subroutine is comparing the flags with themselves. The general practise is to compare the flag with its compliment. Marginally safer and faster.

TychoVrahe:
IIUC, items 2,3 and 4 are unrelated to 1 (which is handled by this PR) and can be addressed in separate PR without affecting overall security by merging this one, right?

PrisionMike:
Agreed.

matejcik:
ok then, let's merge this one.
@PrisionMike can you please make a separate issue for 2+3+4?

PrisionMike:
Sure @matejcik

PR #5169, opened 2025-06-10
Add THP FailureTypes

M1nd3r:
Add FailureTypes specific for THP. 

Could be merged before the rest of THP to help with Suite development.

### DIFF
--- a/common/protob/messages-common.proto
+++ b/common/protob/messages-common.proto
@@ -40,6 +40,9 @@ message Failure {
         Failure_WipeCodeMismatch = 13;
         Failure_InvalidSession = 14;
         Failure_Busy = 15;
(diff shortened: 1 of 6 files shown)


common/protob/messages-common.proto
```diff
@@ -40,6 +40,9 @@ message Failure {
         Failure_WipeCodeMismatch = 13;
         Failure_InvalidSession = 14;
         Failure_Busy = 15;
+        Failure_ThpUnallocatedSession = 16;
+        Failure_InvalidProtocol = 17;
+        Failure_BufferError = 18;
```
mmilata, line 45: Can the client even do anything in this case? I'd rather use generic `FirmwareError` but we can merge it now and then deprecate (or drop if we don't release it) if it helps Suite devs.

PR #5171, opened 2025-06-10
fix(core): fix write to otp memory (U5 only)

cepetr:
This PR fixes a firmware crash that occurred while writing to the OTP memory (in the firmware and prodtest on all U5 models). The issue was introduced with the secure monitor.

Writing to the OTP memory requires access to the non-secure OTP area as well as to the non-secure FLASH controller registers that were disabled by introducing the secure monitor.

### DIFF
--- a/core/embed/sys/mpu/stm32u5/mpu.c
+++ b/core/embed/sys/mpu/stm32u5/mpu.c
@@ -131,12 +131,20 @@ static void mpu_set_attributes(void) {
 _Static_assert(NORCOW_SECTOR_SIZE == STORAGE_1_MAXSIZE, "norcow misconfigured");
 _Static_assert(NORCOW_SECTOR_SIZE == STORAGE_2_MAXSIZE, "norcow misconfigured");
 
+// PERIPH_SIZE covers secure peripherals only (+16MB of FMC1)
+// PERIPH_SIZE_EXT covers both secure and non-secure peripherals (+16MB of FMC1)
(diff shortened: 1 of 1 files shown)


core/embed/sys/mpu/stm32u5/mpu.c
```diff
@@ -135,8 +135,10 @@ _Static_assert(NORCOW_SECTOR_SIZE == STORAGE_2_MAXSIZE, "norcow misconfigured");
 // Extended peripheral block to cover FMC1 that's used for display
 // 512M of periherals + 16M for FMC1 area that follows
 #define PERIPH_SIZE (SIZE_512M + SIZE_16M)
+#define PERIPH_SIZE_EXT (SIZE_256M + SIZE_16M)
```
TychoVrahe: this is a bit confusing, in EXT variant we actually cover less area than normal variant (we still cover non-secure `FLASH` periph but not all non-secure periphs as with `U5G` and we don't cover `FMC1`). I think we should either set  `PERIPH_SIZE_EXT` to `SIZE_512M + SIZE_16M` for `U585`, or to `SIZE_256M + SIZE_16M` for `U5G`.

cepetr: You're right, the constants are swapped. I fixed it and improved comments. See db5a4ec7ab5d223fa497aa86664fa030fb099077 and c96c8f63ca8282cbf9c75ccbd1ee8aff96af56b3

TychoVrahe (approved):

PR #5173, opened 2025-06-10
Update fuel gauge config v2

kopecdav:
Updated battery model and fuel gauge config.

battery model:
More accurate dataset from device3 and device4

fuel gauge:
Increased R parameter to rely more on coulomb counter model. (fuel gauge will be less aggressive and have more inertia)

### DIFF
--- a/core/embed/sys/power_manager/stm32u5/power_manager_internal.h
+++ b/core/embed/sys/power_manager/stm32u5/power_manager_internal.h
@@ -43,10 +43,10 @@
 #define PM_WPC_CHARGE_CURR_STEP_TIMEOUT_MS 1000
 
 // Fuel gauge extended kalman filter parameters
-#define PM_FUEL_GAUGE_R 100.0f
-#define PM_FUEL_GAUGE_Q 0.0001f
-#define PM_FUEL_GAUGE_R_AGGRESSIVE 10.0f
-#define PM_FUEL_GAUGE_Q_AGGRESSIVE 0.01f
+#define PM_FUEL_GAUGE_R 2000.0f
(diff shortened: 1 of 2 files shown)


core/embed/sys/power_manager/stm32u5/power_manager_internal.h
```diff
@@ -43,10 +43,10 @@
...
 
 // Fuel gauge extended kalman filter parameters
-#define PM_FUEL_GAUGE_R 100.0f
-#define PM_FUEL_GAUGE_Q 0.0001f
-#define PM_FUEL_GAUGE_R_AGGRESSIVE 10.0f
-#define PM_FUEL_GAUGE_Q_AGGRESSIVE 0.01f
+#define PM_FUEL_GAUGE_R 2000.0f
+#define PM_FUEL_GAUGE_Q 0.001f
+#define PM_FUEL_GAUGE_R_AGGRESSIVE 1000.0f
+#define PM_FUEL_GAUGE_Q_AGGRESSIVE 0.001f
```
obrusvit, line 49: IMO, `_AGGRESSIVE` no longer emphasizes the purpose. Wouldn't something like `_LOW_BATTERY` be more descriptive?

TychoVrahe (approved):
tested a bit on device and the behavior seems much improved

core/embed/sys/power_manager/stm32u5/power_manager_internal.h
kopecdav: Not really, this settings is also used in charging mode when you less then 20% away from the charging target. In general, you want to use more aggressive setting there to correctly hit the voltage target.

PR #5174, opened 2025-06-10
Add explicit `is_data` parameters to props

bieleluk:
This PR introduces improvements to the `Props` rendering and updates all related usages:

- Makes the **3rd parameter** of the `Props` component **mandatory** — a `bool` indicating whether the value is **data** or not.
- Applies **different monospace fonts** for **data** and **non-data** props.
- Implements distinct line-breaking strategies:
  - **Non-data props**: break at white space; use **hyphens** if a word spans multiple lines.
  - **Data props**: break on a **character basis** with **no hyphens**.
- Added **explicit type-checking** and updated all `confirm_properties` calls across all UIs to include the new 3rd `bool` parameter.
- For the **Eckhart UI**, increased the size of the monospace font for props to match the **[Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=8499-6247&t=KdxCygULeyByoXnZ-0)** design.

TODO
- [x] update fixtures for all languages

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
@@ -362,8 +362,8 @@ impl FirmwareUI for UIEckhart {
 
     fn confirm_properties(
         title: TString<'static>,
-        _subtitle: Option<TString<'static>>,
         items: Obj,
+        _subtitle: Option<TString<'static>>,
         hold: bool,
     ) -> Result<impl LayoutMaybeTrace, Error> {
         let paragraphs = PropsList::new(
--- a/core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
@@ -369,8 +369,8 @@ impl FirmwareUI for UIEckhart {
         let paragraphs = PropsList::new(
             items,
             &theme::TEXT_SMALL_LIGHT,
-            &theme::TEXT_MONO_LIGHT,
-            &theme::TEXT_MONO_LIGHT,
+            &theme::TEXT_MONO_MEDIUM_LIGHT,
+            &theme::TEXT_MONO_MEDIUM_LIGHT_DATA,
         )?;
 
         let flow = flow::new_confirm_with_menu(
--- a/core/src/trezor/ui/layouts/bolt/__init__.py
+++ b/core/src/trezor/ui/layouts/bolt/__init__.py
@@ -601,11 +601,7 @@ async def _confirm_ask_pagination(
     extra_confirmation_if_not_read: bool = False,
     hold: bool = False,
 ) -> None:
-    # TODO: make should_show_more/confirm_more accept bytes directly
-    if isinstance(data, (bytes, bytearray, memoryview)):
-        from ubinascii import hexlify
-
-        data = hexlify(data).decode()
+    data = utils.hexlify_if_bytes(data)
 
     confirm_more_layout = trezorui_api.confirm_more(
         title=title,
@@ -813,8 +809,6 @@ def confirm_properties(
     hold: bool = False,
     br_code: ButtonRequestType = ButtonRequestType.ConfirmOutput,
 ) -> Awaitable[None]:
-    # Monospace flag for values that are bytes.
-    items = [(prop[0], prop[1], isinstance(prop[1], bytes)) for prop in props]
 
     if subtitle:
         title += ": " + subtitle
@@ -822,7 +816,7 @@ def confirm_properties(
     return raise_if_not_confirmed(
         trezorui_api.confirm_properties(
             title=title,
-            items=items,
+            items=list(props),
             hold=hold,
         ),
         br_name,
@@ -944,7 +938,7 @@ if not utils.BITCOIN_ONLY:
             title=TR.send__send_from,
             items=[
                 (f"{TR.words__account}:", account or ""),
-                (f"{TR.address_details__derivation_path}:", account_path or ""),
+                (TR.address_details__derivation_path_colon, account_path or ""),
             ],
         )
 
@@ -1044,18 +1038,19 @@ if not utils.BITCOIN_ONLY:
                 br_name="confirm_ethereum_approve",
             )
 
-        properties = (
-            [(TR.words__token, token_symbol)]
+        properties: list[PropertyType] = (
+            [(TR.words__token, token_symbol, True)]
             if is_revoke
             else [
                 (
                     f"{TR.ethereum__approve_amount_allowance}:",
                     total_amount or TR.words__unlimited,
+                    False,
                 )
             ]
         )
         if not is_unknown_network:
-            properties.append((f"{TR.words__chain}:", network_name))
+            properties.append((f"{TR.words__chain}:", network_name, True))
         await confirm_properties(
             "confirm_ethereum_approve",
             TR.ethereum__approve_revoke if is_revoke else TR.ethereum__approve,
@@ -1207,7 +1202,7 @@ if not utils.BITCOIN_ONLY:
 
         items = [
             (f"{TR.words__account}:", account),
-            (f"{TR.address_details__derivation_path}:", account_path),
+            (TR.address_details__derivation_path_colon, account_path),
         ]
         if stake_item is not None:
             items.append(stake_item)
@@ -1418,7 +1413,7 @@ async def confirm_signverify(
     if account is not None:
         items.append((f"{TR.words__account}:", account))
     if path is not None:
-        items.append((f"{TR.address_details__derivation_path}:", path))
+        items.append((TR.address_details__derivation_path_colon, path))
     items.append(
         (
             f"{TR.sign_message__message_size}:",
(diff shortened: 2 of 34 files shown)


core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -360,11 +360,14 @@ impl FirmwareUI for UIEckhart {
         items: Obj,
         hold: bool,
     ) -> Result<impl LayoutMaybeTrace, Error> {
+        // make sure the the hyphens are not used for line breaking
```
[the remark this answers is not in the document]
bieleluk: https://github.com/trezor/trezor-firmware/pull/5174/commits/398d676ef3ec26dc97cbfc21d576992ce6ea84aa

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -360,11 +360,14 @@ impl FirmwareUI for UIEckhart {
         items: Obj,
         hold: bool,
     ) -> Result<impl LayoutMaybeTrace, Error> {
+        // make sure the hyphens are not used for line breaking
+        const MONO_STYLE: TextStyle =
```
obrusvit: I'd just put it to the `theme` module and name it appropriately

obrusvit (changes requested):
Please take a look at ADA calls of `confirm_properties` so that the new `TextStyle` is used appropriately. Thanks.

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -363,8 +364,8 @@ impl FirmwareUI for UIEckhart {
         let paragraphs = PropsList::new(
             items,
             &theme::TEXT_SMALL_LIGHT,
-            &theme::TEXT_MONO_LIGHT,
-            &theme::TEXT_MONO_LIGHT,
+            &theme::TEXT_MONO_MEDIUM_LIGHT,
+            &TEXT_MONO_MEDIUM_LIGHT_DATA,
```
romanz: nit (for consistency):
```suggestion
            &theme::TEXT_MONO_MEDIUM_LIGHT_DATA,
```

core/src/trezor/ui/layouts/bolt/__init__.py
```diff
@@ -778,7 +778,14 @@ def confirm_properties(
     br_code: ButtonRequestType = ButtonRequestType.ConfirmOutput,
 ) -> Awaitable[None]:
     # Monospace flag for values that are bytes.
-    items = [(prop[0], prop[1], isinstance(prop[1], bytes)) for prop in props]
+    items = [
```
romanz: Consider moving this logic into `core/src/trezor/ui/layouts/common.py` to avoid duplication across models.

bieleluk: This is, unfortunately, not universal for layouts. The Caesar layout differs a lot.

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5174/commits/b9d298b1b30ca0a2f9b6bfef39cfe5ad4439835d

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
bieleluk: Reworked it completely, added dedicated MONO_DATA style for every layout.

core/src/trezor/ui/layouts/common.py
```diff
@@ -12,7 +12,7 @@
 if TYPE_CHECKING:
     from typing import Any, Awaitable, Callable, TypeVar
 
-    PropertyType = tuple[str | None, str | bytes | None]
+    PropertyType = tuple[str | None, str | bytes | None, bool | None]
```
romanz, line 15: nit: if the 3rd parameter is now mandatory, can we make it non-nullable?
```suggestion
    PropertyType = tuple[str | None, str | bytes | None, bool]
```

if yes, we can probably simplify `confirm_properties`, e.g.:
```diff
diff --git a/core/src/trezor/ui/layouts/delizia/__init__.py b/core/src/trezor/ui/layouts/delizia/__init__.py
index 767f19a0ce..894321000f 100644
--- a/core/src/trezor/ui/layouts/delizia/__init__.py
+++ b/core/src/trezor/ui/layouts/delizia/__init__.py
@@ -710,20 +710,11 @@ def confirm_properties(
     br_code: ButtonRequestType = ButtonRequestType.ConfirmOutput,
 ) -> Awaitable[None]:
     # Monospace flag for values that are bytes.
-    items: list[tuple[str | None, str | bytes | None, bool]] = [
-        (
-            prop[0],
-            prop[1],
-            prop[2] if prop[2] is not None else True,
-        )
-        for prop in props
-    ]
-
     return raise_if_not_confirmed(
         trezorui_api.confirm_properties(
             title=title,
             subtitle=subtitle,
-            items=items,
+            items=list(props),
             hold=hold,
         ),
         br_name,
```
WDYT?

core/src/trezor/ui/layouts/caesar/__init__.py
```diff
@@ -780,23 +780,26 @@ def confirm_properties(
...
-        else:
-            # When there is not space in the text, taking it as data
-            # to not include hyphens
-            is_data = value and " " not in value
-            return (key, value, bool(is_data))
+    items: list[tuple[str | None, str | bytes | None, bool]] = [
+        (
+            prop[0],
+            (
+                hexlify(prop[1]).decode()
```
romanz: This following conversion looks similar to the one here:
https://github.com/trezor/trezor-firmware/blob/1cb6ecbee33eeb838987ca4fa5348365a7deffa2/core/src/trezor/ui/layouts/caesar/__init__.py#L673-L677
Is it possible to extract the common code into a helper function?

romanz: BTW, do we need something similar for other layouts?

core/src/trezor/ui/layouts/common.py
bieleluk: The 3rd parameter is relevant only when the second is not None. Showing only the key parameter is used in many ADA cases, i.e.
```
append((address_n_to_str(key_path), None, None))
```
For this reason, I would leave there the `None` option as well.

core/src/trezor/ui/layouts/caesar/__init__.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5174/commits/205a185bb596bee5296aeda8641cbadf864449a6

core/src/trezor/ui/layouts/common.py
romanz: Understood, thanks!

romanz: For readability, maybe can we use something similar to a `namedtuple` with `None` defaults?
```
In [1]: X = namedtuple('X', ('a', 'b', 'c'), defaults=(None, None))

In [2]: X(1)
Out[2]: X(a=1, b=None, c=None)

In [3]: X(1, 2)
Out[3]: X(a=1, b=2, c=None)

In [4]: X(1, 2, 3)
Out[4]: X(a=1, b=2, c=3)
```
https://docs.python.org/3/library/collections.html#collections.namedtuple

WDYT?

core/src/trezor/utils.py
```diff
@@ -377,6 +377,17 @@ def empty_bytearray(preallocate: int) -> bytearray:
     return b
 
 
+def to_hex_string(data: str | bytes | bytearray | memoryview) -> str:
```
romanz: Thanks! 
Maybe call it `hexlify_if_bytes()`?
Can it also replace `get_bytes_as_str()` in `utils.py`?

bieleluk: https://github.com/trezor/trezor-firmware/pull/5174/commits/0d05f732c8149273e4a1451684e37c511ced1ca4

core/src/trezor/ui/layouts/common.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5174/commits/3f901680e3bfe0c75718813f76675053aa66db20
On top of your suggestion, I changed the FwUI function parameter to optional `bool` so we have 1 to 1 matching for most layouts.

Update:
This approach is probably not the right one:
- the build and style check work well but there is runtime error: "Undefined type PropertyType", because it's defined behind the `TYPE_CHECKING` switch.
- the other option was to defined a global class
```
class PropertyType(NamedTuple):
    key: str | None = None
    value: str | bytes | None = None
    is_value_data: bool | None = None
```
but for that I am getting a strange error ` '_GenericTypingObject' object isn't callable` that I wasn't able to overcome

romanz: Apologies, our MicroPython doesn't support `namedtuple` :(

romanz: We can resolve it in a later PR, e.g. by using helper functions for constructing various kinds of properties:
```python
def text_property(name: str, value: str):
    return (name, value, False)

def data_property(name: str, value: str):
    return (name, value, True)
    
...
```
([my comment above](https://github.com/trezor/trezor-firmware/pull/5174#discussion_r2161147873) was mainly about readability, trying to avoid using raw tuples...)

PR #5175, opened 2025-06-10
fix(core): fix firmware hash calculation

cepetr:
This PR fixes a bug introduced by the secure monitor PR that prevented the firmware hash from being calculated.

The firmware on T3T1 (and possibly T3B1) is crashing when retrieving the firmware hash with `trezorctl fw get-hash`. 

The crash was caused by a small gap between the kernel and the coreapp that the kernel could not access.

### DIFF
--- a/core/embed/projects/kernel/main.c
+++ b/core/embed/projects/kernel/main.c
@@ -213,11 +213,11 @@ static void kernel_loop(applet_t *coreapp) {
 
 // defined in linker script
 extern uint32_t _kernel_flash_end;
-#define KERNEL_END ((uint32_t) & _kernel_flash_end)
+#define KERNEL_END COREAPP_CODE_ALIGN((uint32_t) & _kernel_flash_end)
 
 // Initializes coreapp applet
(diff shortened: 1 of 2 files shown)


TychoVrahe (approved):

bosomt:
**QA WIP**

_T3B1_
 - Device: Trezor T3B1 2.9.0 regular (revision 107960ebbe787a86920d5b3d8ac2c804f87bccff)
```
➜  ~ trezorctl fw get-hash
4da17cdc56c127dcaf6baf125e993032e452b826844e9539ce9fa2bace7689a3
```

_T3T1_
```
➜  trezorctl get-features | grep model
    internal_model: 'T3T1',
    model: 'Safe 5',
➜  trezorctl get-features | grep rev
    revision: 20 bytes 0xf906a91ebf561e99a9449b6f8082cfc123bbce09,
➜  trezorctl fw get-hash
de7e06b31b457e957790def27af1773304d357ccc3a6efa60579dca3aa6faa95
```

PR #5176, opened 2025-06-10
test(core): don't call `clear_gc_info()` when PYOPT=1

romanz:
Fixes a `(PE)` error introduced by #5091 on non-debug builds.


cepetr (approved):
Tested on T3W1

PR #5180, opened 2025-06-11
t3w1: refactor nrf, disable uart in booloader to save flash

TychoVrahe:
This PR addresses acute flash issues with T3W1 bootloader by disabling UART, which is currently not needed in booloader. Saves about 8kB of flash space. As part of this, nrf driver is split into more files to make it a bit easier to navigate.

### DIFF
--- a/core/SConscript.prodtest
+++ b/core/SConscript.prodtest
@@ -31,6 +31,7 @@ FEATURES_WANTED = [
     "secure_domain",
     "tropic",
     "usb",
+    "smp",
     ]
 
 CCFLAGS_MOD = ''
@@ -146,6 +147,7 @@ env = Environment(
 )
 
 FEATURES_AVAILABLE = models.configure_board(TREZOR_MODEL, HW_REVISION, FEATURES_WANTED, env, CPPDEFINES_HAL, SOURCE_HAL, PATH_HAL)
+
 
 SOURCE_PRODTEST = [
     'embed/projects/prodtest/header.S',
--- a/core/embed/io/nrf/stm32u5/nrf.c
+++ b/core/embed/io/nrf/stm32u5/nrf.c
@@ -32,74 +32,10 @@
 
 #include "../crc8.h"
 #include "../nrf_internal.h"
-#include "rust_smp.h"
-
-#define MAX_SPI_DATA_SIZE (244)
-
-typedef struct {
-  uint8_t service_id;
-  uint8_t msg_len;
-  uint8_t data[MAX_SPI_DATA_SIZE];
-  uint8_t crc;
-} spi_packet_t;
-
-#define SPI_OVERHEAD_SIZE (sizeof(spi_packet_t) - MAX_SPI_DATA_SIZE)
-#define SPI_HEADER_SIZE (SPI_OVERHEAD_SIZE - 1)
-
-#define TX_QUEUE_SIZE (8)
-
-#define START_BYTE (0xA0)
 
 #define CTS_PULSE_RESEND_PERIOD_US 2000
 
-typedef struct {
-  spi_packet_t packet;
-  nrf_tx_callback_t callback;
-  void *context;
-} nrf_tx_request_t;
-
-typedef struct {
-  UART_HandleTypeDef urt;
-  DMA_HandleTypeDef urt_tx_dma;
-
-  uint8_t tx_buffers[TX_QUEUE_SIZE][sizeof(nrf_tx_request_t)];
-  tsqueue_entry_t tx_queue_entries[TX_QUEUE_SIZE];
-  tsqueue_t tx_queue;
-  nrf_tx_request_t tx_request;
-  int32_t tx_request_id;
-
-  uint8_t urt_rx_byte;
-  uint8_t urt_tx_byte;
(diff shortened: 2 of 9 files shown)


core/embed/io/nrf/stm32u5/nrf_spi.c
```diff
@@ -0,0 +1,267 @@
...
+#include <sys/mpu.h>
+#include <sys/systick.h>
+
+#include "../crc8.h"
+#include "../nrf_internal.h"
+
+#define SPI_OVERHEAD_SIZE (sizeof(spi_packet_t) - MAX_SPI_DATA_SIZE)
+#define SPI_HEADER_SIZE (SPI_OVERHEAD_SIZE - 1)
+
+#define START_BYTE (0xA0)
```
cepetr, line 36: The `START_BYTE` constant seems to be unused now, but there are still places in the code where `0xA0` is used explicitly.

core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -251,98 +186,11 @@ void nrf_init(void) {
...
-  drv->spi.Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_2;
-  drv->spi.Init.FirstBit = SPI_FIRSTBIT_MSB;
-  drv->spi.Init.TIMode = SPI_TIMODE_DISABLE;
-  drv->spi.Init.CRCCalculation = SPI_CRCCALCULATION_DISABLE;
-  drv->spi.Init.CRCPolynomial = 0;
-  drv->spi.hdmarx = &drv->spi_rx_dma;
-  drv->spi.hdmatx = &drv->spi_tx_dma;
-
-  HAL_SPI_Init(&drv->spi);
+  nrf_spi_init(drv);
```
cepetr, line 190: I guess that following two lines should be inside `#ifdef USE_SMP / #endif`:
```c
  NVIC_SetPriority(USART3_IRQn, IRQ_PRI_NORMAL);
  NVIC_EnableIRQ(USART3_IRQn);
  ```

File: core/embed/io/nrf/stm32u5/nrf.c [hunk 3]
```diff
@@ -462,37 +310,6 @@ void nrf_unregister_listener(nrf_service_id_t service) {
...
-
-  drv->pending_spi_transaction = true;
-  nrf_signal_data_ready();
-  systick_delay_us(1);
-  nrf_signal_no_data();
-  systimer_set(drv->timer, 2000);
-}
-
 /// DFU communication
 /// ----------------------------------------------------------
```
cepetr: Shouldn't we move `nrf_dfu_comm_send` and `nrf_dfu_comm_receive` into `nrf_uart.c` as well, since they rely entirely on UART presence and the `USE_SMP` macro? If you call them when `USE_SMP` is not defined, you'll see no compile error, but the they won't work.

Following the same logic, `nrf_send_msg` and `nrf_abort` could belong in `nrf_spi.c` as well.

core/SConscript.prodtest
```diff
@@ -147,6 +147,10 @@ env = Environment(
 
 FEATURES_AVAILABLE = models.configure_board(TREZOR_MODEL, HW_REVISION, FEATURES_WANTED, env, CPPDEFINES_HAL, SOURCE_HAL, PATH_HAL)
 
+if 'nrf' in FEATURES_AVAILABLE:
```
cepetr: How about adding 'smp' to the `FEATURES_WANTED` lists in `SConscript.kernel` and `SConscript.prodtest`? 

Then we could simply check the `smp` feature in `trezor_t3w1_revX.py` and include/exlude `nrf_uart.c` or define `USE_SMP` there.

core/embed/io/nrf/stm32u5/nrf_spi.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/de3cfb1b3376ebdfb6b90a4f3a47684ca2934438

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: fixed, also extracted deinit and some clock enabling: https://github.com/trezor/trezor-firmware/commit/de3cfb1b3376ebdfb6b90a4f3a47684ca2934438

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: moved: https://github.com/trezor/trezor-firmware/commit/de3cfb1b3376ebdfb6b90a4f3a47684ca2934438

TychoVrahe: and we don't need this ifdef now: https://github.com/trezor/trezor-firmware/commit/52603c5185b9e7ee5e2eb42a67e3083745279998

core/SConscript.prodtest
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/de3cfb1b3376ebdfb6b90a4f3a47684ca2934438

PR #4715, opened 2025-03-03
feat(core): use `SwipeFlow::add_page` to reduce stack usage

romanz:
Following https://github.com/trezor/trezor-firmware/pull/4696 (used to reduce `new_flow_confirm_output` stack size):

## Before
```
$ git lg -1
* 587b5e8eb tychovrahe:  (HEAD -> main, origin/main, origin/HEAD) fix(core): fix jump to old bootloader (3 days ago)

$ TREZOR_MODEL=T3T1 make -C core build_firmware; python ~/tmp/stack_sizes.py | sort -k1 -n | tail
...
7256	trezor_lib::ui::api::firmware_micropython::new_show_checklist
7440	trezor_lib::ui::api::firmware_micropython::new_flow_confirm_set_new_pin
7440	trezor_lib::ui::api::firmware_micropython::new_prompt_backup
7512	trezor_lib::ui::api::firmware_micropython::new_show_share_words_delizia
9224	trezor_lib::ui::api::firmware_micropython::new_confirm_fido
13176	trezor_lib::ui::api::firmware_micropython::new_tutorial
14096	trezor_lib::ui::api::firmware_micropython::new_continue_recovery_homepage
14144	trezor_lib::ui::api::firmware_micropython::new_flow_get_address
15792	trezor_lib::ui::api::firmware_micropython::new_confirm_summary
18096	trezor_lib::ui::api::firmware_micropython::new_flow_confirm_output
```

## After
```
$ git lg -1
* 7a0008064 Roman Zeyde:  (HEAD -> romanz/page-builder) feat(core): use `SwipeFlow::add_page` to reduce stack usage (13 minutes ago)

$ TREZOR_MODEL=T3T1 make -C core build_firmware; python ~/tmp/stack_sizes.py | sort -k1 -n | tail
...
5768	trezor_lib::ui::api::firmware_micropython::new_flow_confirm_set_new_pin
5768	trezor_lib::ui::api::firmware_micropython::new_prompt_backup
6696	trezor_lib::ui::api::firmware_micropython::new_show_share_words_delizia
7256	trezor_lib::ui::api::firmware_micropython::new_show_checklist
7616	trezor_lib::ui::api::firmware_micropython::new_confirm_fido
9112	trezor_lib::ui::api::firmware_micropython::new_tutorial
10704	trezor_lib::ui::api::firmware_micropython::new_continue_recovery_homepage
11800	trezor_lib::ui::api::firmware_micropython::new_flow_get_address
[...]

PR #5183, opened 2025-06-11
feat(core): use OpTextLayout::add_ functions to reduce stack usage

bieleluk:
Change `OpTextLayout` methods to accept `&mut self`. Similar to #4715


romanz (approved):
Thanks!

PR #5185, opened 2025-06-12
introduce IWDG driver

TychoVrahe:
This PR introduces a small IWDG driver.

Purpose is to enforce system/mcu reset after specified time elapsed, even if the MCU is in stop mode. Currently we don't start the IWDG anywhere, it will be subject to next PR - but it is expected to be started in bootloader, after checking installed firmware privileges.


core/embed/sec/iwdg/stm32/iwdg.c
```diff
@@ -0,0 +1,49 @@
...
+void iwdg_start(uint32_t time_s) {
+  if (time_s > IWDG_MAX_TIME) {
+    // Limit the maximum watchdog timeout to 4 hours
+    time_s = IWDG_MAX_TIME;
+  }
+
+  // Set the reload value based on the desired time in seconds
+  uint32_t reload_value = ((time_s * LSI_VALUE) / 1024) - 1;
+
+  IWDG_HandleTypeDef hiwdg;
```
cepetr: Even though it isn’t required here, what about `IWDG_HandleTypeDef hiwdg = {0}`

TychoVrahe: fixed: https://github.com/trezor/trezor-firmware/commit/88214a7fe1999df2d01ca8a242b611a4cff5d509

cepetr (approved):

Issue #5146, opened 2025-06-05
Change behavior when showing the passphrase/PIN


bieleluk:
## Current behavior:

When I press and hold the upper part because I want to check what is written there I cannot see it fully because of my finger. 

Expected:
The passphrase (or PIN) stays displayed when my finger swipes out of the upper part (down) and disappears after the touch is released.

_Originally posted by @Hannsek in https://github.com/trezor/trezor-firmware/issues/5033#issuecomment-2942871330_

PR #5186, opened 2025-06-12
chore(core/eckhart): extend pin/pp touch area

bieleluk:
#5146


obrusvit (approved):
It works well on the device. Tested both PIN and passphrase entry.

I would extend the `SHOWN_TOUCH_OUTSET` to `200` though. 
Also the `const` is copied there twice but that's unfortunately the case for more constants (and code) and `Pin/PassphraseInput`. In some cases it's justified so I'm not sure if it's worth abstracting.

bosomt:
**QA OK**

 - Device: Trezor T3W1 2.9.0 regular (revision c17ce0e89e26a89be55fb8dfe988b508145f31ff)

PR #5191, opened 2025-06-12
common: add THP protobuf definitions

mmilata:
Protobuf definitions extracted from #4976. There are going to be changes in the future but at this point it does make sense have the definitions in `main` for Suite to use.

Commit [49d67b2](https://github.com/trezor/trezor-firmware/pull/5191/commits/49d67b250d05db0d146ee694ad38582d11f4b094) excludes the `MessageType` enum from rust protobuf blobs. This is because the current format has limit of 255 enum variants while this PR uses 264. As it is actually not used at runtime we can drop it and save 0.5KiB of flash space. There is also a PR that increases the limit to 64k: #4283

cc @szymonlesisz @M1nd3r


common/protob/messages.proto
```diff
@@ -317,7 +319,24 @@ enum MessageType {
...
+    MessageType_ThpEndResponse = 1013 [(bitcoin_only) = true];
+    MessageType_ThpCodeEntryCommitment = 1016 [(bitcoin_only) = true];
+    MessageType_ThpCodeEntryChallenge = 1017 [(bitcoin_only) = true];
+    MessageType_ThpCodeEntryCpaceTrezor = 1018 [(bitcoin_only) = true];
+    MessageType_ThpCodeEntryCpaceHostTag = 1019 [(bitcoin_only) = true];
+    MessageType_ThpCodeEntrySecret = 1020 [(bitcoin_only) = true];
+    MessageType_ThpQrCodeTag = 1024 [(bitcoin_only) = true];
+    MessageType_ThpQrCodeSecret = 1025 [(bitcoin_only) = true];
+    MessageType_ThpNfcTagHost = 1032 [(bitcoin_only) = true];
+    MessageType_ThpNfcTagTrezor = 1033 [(bitcoin_only) = true];
```
romanz, line 343: tiny nit: if needed, it is possible to remove THP-related strings from non-THP builds (similar to https://github.com/trezor/trezor-firmware/pull/4949):
```
$ TREZOR_MODEL=T3B1 BITCOIN_ONLY=0 QUIET_MODE=1 make -C core build_firmware
$ strings core/build/firmware/firmware.bin | grep 'Thp'
ThpCodeEntryChallenge
ThpCodeEntryCommitment
ThpCodeEntryCpaceHostTag
ThpCodeEntryCpaceTrezor
ThpCodeEntrySecret
ThpCreateNewSession
ThpCredentialRequest
ThpCredentialResponse
ThpEndRequest
ThpEndResponse
ThpNfcTagHost
ThpNfcTagTrezor
ThpPairingPreparationsFinished
ThpPairingRequest
ThpPairingRequestApproved
ThpQrCodeSecret
ThpQrCodeTag
ThpSelectMethod
ThpUnallocatedSession
trezor/enums/ThpPairingMethod.py
trezor/enums/ThpPairingMethod.py
```

core/src/trezor/enums/__init__.py
```diff
@@ -574,6 +582,24 @@ class MessageType(IntEnum):
         SolanaAddress = 903
         SolanaSignTx = 904
         SolanaTxSignature = 905
+        ThpCreateNewSession = 1000
+        ThpPairingRequest = 1006
```
romanz, line 586: Should we mark the unused values (e.g. 1001-1005 here, also below) as deprecated?

common/protob/messages-debug.proto
```diff
@@ -134,6 +134,39 @@ message DebugLinkState {
...
+    optional bytes nfc_secret_trezor = 5;       // NFC secret used in NFC pairing
+}
+
+/**
+ * Request: Disable or enable the pairing dialog for a given THP channel
+ * @start
+ * @next Success
+ */
+message DebugLinkToggleThpPairingDialog{
+    optional bytes channel_id=1;
```
romanz: nit: missing spaces around `'='`
```protobuf
message DebugLinkToggleThpPairingDialog{
    optional bytes channel_id = 1;
    optional bool show_dialog = 2 [default=false];
}
```

romanz (approved):

common/protob/messages.proto
mmilata: done in [14691d1](https://github.com/trezor/trezor-firmware/pull/5191/commits/14691d1fcb3a69682c89bb714f37b80b25ac29a6)

core/src/trezor/enums/__init__.py
mmilata: Added `reserved` keyword in 20135989753b3d1b9dbde2da7e827eec74ef0495 since `MessageType` generally seems to have no intra-block holes in the ID allocation. These were never used in a released version so we can use them in the future.

common/protob/messages-debug.proto
mmilata: thanks, 20135989753b3d1b9dbde2da7e827eec74ef0495

PR #5192, opened 2025-06-12
refactor(core): wrap coverage data map with `spin::RwLock`

romanz:
Following, the upgrade to Rust 2024.

```
warning: creating a mutable reference to mutable static is discouraged
  --> src/coverage/mod.rs:70:19
   |
70 |             match COVERAGE_DATA.entry(key) {
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static
   |
   = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
   = note: mutable references to mutable statics are dangerous; it's undefined behavior if any other pointer to the static is used or if any other reference is created for the static while the mutable reference lives
   = note: `#[warn(static_mut_refs)]` on by default

warning: creating a shared reference to mutable static is discouraged
  --> src/coverage/mod.rs:88:45
   |
88 |         let list = unsafe { List::from_iter(COVERAGE_DATA.iter().map(Item))? };
   |                                             ^^^^^^^^^^^^^^^^^^^^ shared reference to mutable static
   |
   = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
   = note: shared references to mutable statics are dangerous; it's undefined behavior if the static is mutated or if a mutable reference is created for it while the shared reference lives
```

### DIFF
--- a/core/embed/rust/src/coverage/mod.rs
+++ b/core/embed/rust/src/coverage/mod.rs
@@ -1,5 +1,6 @@
 /// Off-heap data structure for collecting code coverage data.
 use heapless::{Entry, FnvIndexMap};
+use spin::RwLock;
 
 use crate::{
     error::Error,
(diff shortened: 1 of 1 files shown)


matejcik:
instead wrap it in a Mutex or a RwLock?
(from the `spin` crate that we already have)

romanz:
Thanks! 
I have used `RwLock` in https://github.com/trezor/trezor-firmware/pull/5192/commits/bae59ec11c923f26fa1c36487e288369c4df2f4e (which required https://github.com/trezor/trezor-firmware/pull/5192/commits/bbc8b06bd8949f92210a1f3e1024eacb111af697, since `StrBuffer` isn't `Send`).

matejcik (approved):

PR #5196, opened 2025-06-13
Enable BKRAM retention in vbat mode

cepetr:
This small PR correctly enables the BKRAM retention bit during the early stages of boot - before the SMPS is turned on.

This fixes the retention problem when holding power button for more than 10s.

image (1)

This setting applies only in the boardloader, because the SMPS is already enabled in later boot stages. It seems the power- and clock-initialization code in `startup_init.c` still needs a few tweaks to be completely correct.


TychoVrahe (approved):
utACK

PR #5198, opened 2025-06-13
fix(emu): use same key for screenshots

obrusvit:
- firmware emu uses `p` for power button and `s` for screenshot. So use `s` also for bootloader and prodtest emu screenshots.

[no changelog]


TychoVrahe (approved):
utACK

PR #5199, opened 2025-06-14
Suspend from application layer

obrusvit:
# Suspend, version 1
- implements suspending the device on `BTN` (power button press) from firmware application
- handles [wakeup flags](https://github.com/trezor/trezor-firmware/pull/5235) sent by [power manager](https://github.com/trezor/trezor-firmware/pull/4984).
- implements autosuspend functionality
## Implementation of suspend: 
- `BTN` event is checked in `Layout::_handle_button_events`
- on `BTN` event, the `handle_power_button_press()` function is called which locks the device. Moreover, if the device is not plugged in via USB cable, then the device enters an infinite loop of suspending the device and checking the resulting `status` and `wakeup_flag`
  - on `BTN` or `BLE` wakeup flag, a normal app fw operation is continued
  - on `PWR` wakeup flag (device starts charging), 
    - Not via USB cable => a new workflow `chargingscreen` is introduced which `races` a 4s timeout against either a `lockscreen` or `homescreen`. If the user interacts with the device, a normal app fw operation is continued. If the timeout wins, the device is suspended again.
    - Via USB cable => a normal FW operation is continued
  - other wakeup flags (NFC, RTC) are ignored for now
## Implementation of autosuspend
- `idle_timer` is used to register and `suspend_device`, similarly to #5417 
  - the precise timing is subject to tuning
## Known limitations and future work
- [ ] `bootscreen()` does not react to the wakeup events as the function basically handles it's own "event loop". Implementing the same behavior would require duplicating the logic but requiring to stay in `bootscreen`
- [ ] when `homescreen` or `lockscreen` is set and the device is plugged in the USB cable, the workflow is refreshed and the display flickers
- [ ] auto-suspend functionality working in combination to auto-lock, TBD
- [ ] setting up the time of auto-suspend by the user
- [ ] proper device tests with emulated behavior of the battery and power manager
- [ ]  we may want to implement continuing the last workflow on wakeup. This requires deeper analysis and potentially security considerations.
  - e.g. ShareWords
- [ ] when the device is suspended on inactivity, then resumed and then communicated with over BLE - after receiving the message device is put to suspend again https://github.com/trezor/trezor-firmware/pull/5199#issuecomment-3029100682
- [ ] respecting the `ALLOW_WHILE_LOCKED` workflows also for suspend


TychoVrahe:
Didn't review much code yet, but first testing showed some interesting bug:
- when the device is suspended on inactivity, then resumed and then communicated with over BLE - after receiving the message device is put to suspend again

Note: suspending and resuming BLE has its own bugs, cherrypick https://github.com/trezor/trezor-firmware/pull/5293 for testing

TychoVrahe:
> we may want to implement continuing the last workflow on wakeup

maybe that should be a default. i am thinking that for screens that want to break the flow after suspend(or on resume), we could send and Supsend/Resume events that the screen would react to.

Maybe also dimming/suspending everywhere is a bit too much: i think i've seen dimming while i was waiting for entropy check, that seems wrong, the device isn't really idle.

Do we want to reset the dim/suspend timer on communication? suspending while communication is ongoing is weird, especially if we break the flows.

TychoVrahe:
There is also some weird interaction with brightness setting screen, maybe that needs some special handling?

Hannsek:
> Do we want to reset the dim/suspend timer on communication? suspending while communication is ongoing is weird, especially if we break the flows.

I agree. How it is done now with auto-lock? Is the timer reset if host sends e.g. `get-features`?

obrusvit:
> > Do we want to reset the dim/suspend timer on communication? suspending while communication is ongoing is weird, especially if we break the flows.
> 
> I agree. How it is done now with auto-lock? Is the timer reset if host sends e.g. `get-features`?

There are workflows which can run while locked, plus also whatever is set as `default_homescreen` which is typically `Homescreen`/`LockScreen`/`RecoveryScreen`. That means 1) lock can happen while the workflow is in progress, 2) the workflow can be initiated even for locked device. The proper interaction of these workflows while suspended needs to be thought through because e.g. the USB communication is turned off in suspend.

```py
ALLOW_WHILE_LOCKED = (
    MessageType.Initialize,
    MessageType.EndSession,
    MessageType.GetFeatures,
    MessageType.Cancel,
    MessageType.LockDevice,
    MessageType.DoPreauthorized,
    MessageType.WipeDevice,
    MessageType.SetBusy,
    MessageType.Ping,
)
```

obrusvit:
Rebased the branch.

obrusvit:
Autodim functionality and `pm` module creation moved to a separate PR: https://github.com/trezor/trezor-firmware/pull/5417

obrusvit:
Superseded by #5606

PR #5202, opened 2025-06-15
secmon signing

TychoVrahe:
This PR implements separate secmon signing and signature validation. New header type for secmon is introduced.

Along with this, runtime limit is now enforced for custom FWs on T3W1. 

As part of vendor header changes, allow_provisioning flag was introduced, which allows stricter permissions setup.


core/SConscript.firmware
```diff
@@ -871,7 +872,10 @@ if BOOTLOADER_QA or BOOTLOADER_DEVEL:
 
 # select vendor header
 if BOOTLOADER_QA or BOOTLOADER_DEVEL:
-    vendor = "dev_DO_NOT_SIGN_signed_dev"
+    if UNSAFE_FW:
```
cepetr, line 875: What’s the purpose of the `unsafe_signed_dev `header? Can’t we get rid of it and use only `dev_DO_NOT_SIGN_signed_dev` instead? When I build the firmware with `BOOTLOADER_DEVEL` and `UNSAFE_FW`, I get firmware with a built-in secmon that shows a warning every time it starts.

TychoVrahe: > built-in secmon that shows a warning every time it starts.

this is the purpose, it allows testing this functionality (and different permissions) on development devices

core/embed/projects/bootloader/main.c
```diff
@@ -91,9 +94,13 @@ void failed_jump_to_firmware(void);
 CONFIDENTIAL volatile secbool dont_optimize_out_true = sectrue;
 CONFIDENTIAL void (*volatile firmware_jump_fn)(void) = failed_jump_to_firmware;
 
-static secbool is_manufacturing_mode(void) {
+static secbool is_manufacturing_mode(vendor_header *vhdr) {
   unit_properties_init();
 
+  if ((vhdr->vtrust & VTRUST_ALLOW_PROVISIONING) == VTRUST_ALLOW_PROVISIONING) {
```
matejcik: what does the `is_manufacturing_mode` function mean?

because this reads like the inverse: if provisioning is allowed, then manufacturing mode is false?

python/src/trezorlib/firmware/secmon.py
```diff
@@ -0,0 +1,127 @@
...
+    hw_revision: int
+    hash: bytes
+
+    sigmask: int
+    signature: bytes
+
+    # fmt: off
+    SUBCON = c.Struct(
+        "_start_offset" / c.Tell,
+        "magic" / EnumAdapter(c.Bytes(4), SecmonType),
```
matejcik:
```suggestion
        "magic" / c.Const(b"TSEC")
```

core/embed/projects/bootloader/main.c
TychoVrahe: yes that is a bug, thanks https://github.com/trezor/trezor-firmware/commit/fbde08d505f84415b0bf38a6657b614b7acb1fd3

TychoVrahe:
i also added a fix for reset that was in some edge cases done in boardloader - if boardloader detected BHK is locked, it rebooted device with REBOOT command, replacing the original boot command - so i made a change to  preserve the boot command.

This also means that stuff like 7/10s button press will normally lead to powered-off device waiting for another press, as there is no boot command. Also applies to stuff like resetting via debugger, where it is not really important but the button-reset is kinda unclear to me how it should behave. Maybe we should move this stuff out of boardloader so we can change it later if needed.

@cepetr check this new logic.

see https://github.com/trezor/trezor-firmware/pull/5202/commits/85c27d19f28bd36deb7ddae67bd9e0f2cdf5236d

core/embed/models/T3W1/vendorheader/vendor_unsafe.json
```diff
@@ -2,10 +2,13 @@
     "header_len": 1024,
     "text": "UNSAFE, DO NOT USE!",
     "hw_model": "T3W1",
+    "runtime_limit_min": 240,
```
matejcik: leftover?

core/embed/models/T3W1/vendorheader/vendor_dev_DO_NOT_SIGN.json
```diff
@@ -2,10 +2,13 @@
     "header_len": 1024,
     "text": "DEV ONLY, DO NOT USE!",
     "hw_model": "T3W1",
+    "runtime_limit_min": 240,
```
matejcik: leftover?

TychoVrahe:
As discussed in standup, i reverted the bootutils changes and removed the BHK unlocking from boardloader, so no need for reset and rewriting the bootargs. The BHK will be unlocked before loading in bootloader.

TychoVrahe:
bumped T3B1 and T3T1 VH versions in https://github.com/trezor/trezor-firmware/commit/1120aeffa058f10039d540f0c8a0822bd68c877b

removed reserved words from secmon header in https://github.com/trezor/trezor-firmware/commit/7a8278a9acd195fdc9adbf6ea8a95f83d4469fad and https://github.com/trezor/trezor-firmware/commit/237849eb27d38a52079fdba7a7f9ab321ac5fc50 (leaving 3 byte reserve before hash to have the hash aligned to 32bits).

python/src/trezorlib/firmware/secmon.py
```diff
@@ -0,0 +1,125 @@
...
+from .models import Model
+
+__all__ = [
+    "SecmonType",
+    "SecmonHeader",
+    "SecmonImage",
+]
+
+
+class SecmonType(Enum):
```
matejcik: I don't think we need this enum, the secmon magic is constant

python/src/trezorlib/firmware/secmon.py
```diff
@@ -0,0 +1,125 @@
...
+
+    header: SecmonHeader = subcon(SecmonHeader)
+    _header_end: int
+    _code_offset: int
+    code: bytes
+
+    SUBCON = c.Struct(
+        "header" / SecmonHeader.SUBCON,
+        "_header_end" / c.Tell,
+        "_code_offset" / c.Tell,
```
matejcik: these two are identical, and neither one seems used :)

core/embed/util/image/image.c
```diff
@@ -159,6 +172,85 @@ secbool check_image_header_sig(const image_header *const hdr, uint8_t key_m,
...
+
+  if (hdr->magic != magic) {
+    return NULL;
+  }
+  if (hdr->hdrlen != SECMON_HEADER_SIZE) {
+    return NULL;
+  }
+
+  if (hdr->codelen > (maxsize - hdr->hdrlen)) return secfalse;
+  if ((hdr->hdrlen + hdr->codelen) < 4 * 1024) return secfalse;
```
matejcik, line 187: why must this be >= 4 kB?

core/SConscript.secmon
```diff
@@ -243,7 +244,7 @@ env = Environment(
     CPPDEFSUFFIX="'",
 )
 
-if TREZOR_MODEL not in ['T3W1', 'D002']:
+if TREZOR_MODEL not in ['T3W1', 'D002'] or UNSAFE_FW or not BOOTLOADER_DEVEL:
```
matejcik: this makes it impossible to build secmon for production

if the goal was to avoid building secmon in the cases where we're using a binary artifact, that should be solved by not adding it as a dependency?

python/src/trezorlib/_internal/firmware_headers.py
```diff
@@ -183,6 +183,42 @@ def format_header(
...
+        hash_status = Status.INVALID
+    else:
+        hash_status = Status.VALID
+    #
+    # header_out["hashes"] = hashes_out
+
+    all_ok = SYM_OK if hash_status.is_ok() and sig_status.is_ok() else SYM_FAIL
+
+    output = [
+        "SECMON Header ",
```
matejcik:
```suggestion
        "SECMON Header " + _format_container(header_out),
```

core/embed/util/image/inc/util/image.h
```diff
@@ -156,3 +176,15 @@ secbool check_firmware_header(const uint8_t *header, size_t header_size,
                               firmware_header_info_t *info);
 
 secbool __wur check_bootloader_header_sig(const image_header *const hdr);
+
+const secmon_header_t *read_secmon_header(const uint8_t *const data,
+                                          const uint32_t magic,
```
matejcik: what's the point of accepting `magic` from outside?

core/embed/projects/bootloader/main.c
```diff
@@ -376,12 +412,22 @@ void real_jump_to_firmware(void) {
...
+#endif
+
   drivers_deinit();
 
   system_deinit();
 
   jump_to_next_stage(
-      IMAGE_CODE_ALIGN(FIRMWARE_START + vhdr.hdrlen + IMAGE_HEADER_SIZE));
+      IMAGE_CODE_ALIGN(FIRMWARE_START + vhdr.hdrlen + IMAGE_HEADER_SIZE) +
+      secmon_code_offset);
```
matejcik, line 430: why is `secmon_code_offset` not inside `IMAGE_CODE_ALIGN` ?

core/embed/projects/bootloader/workflow/wf_firmware_update.c
```diff
@@ -229,6 +246,42 @@ static upload_status_t process_msg_FirmwareUpload(protob_io_t *iface,
...
+      size_t secmon_start_offset =
+          (size_t)IMAGE_CODE_ALIGN(vhdr.hdrlen + IMAGE_HEADER_SIZE);
+      size_t secmon_start = (size_t)chunk_buffer + secmon_start_offset;
+      const secmon_header_t *secmon_hdr = read_secmon_header(
+          (const uint8_t *)secmon_start, SECMON_IMAGE_MAGIC, FIRMWARE_MAXSIZE);
+
+      if (secmon_hdr != NULL) {
+        ctx->secmon_code_offset =
+            IMAGE_CODE_ALIGN(vhdr.hdrlen + IMAGE_HEADER_SIZE) +
+            SECMON_HEADER_SIZE;
```
matejcik, line 264: dtto here (and perhaps some other places i missed)
isn't the secmon header start AND secmon code start also required to be aligned?

core/embed/projects/bootloader/workflow/wf_firmware_update.c
```diff
@@ -401,6 +454,46 @@ static upload_status_t process_msg_FirmwareUpload(protob_io_t *iface,
...
+    secmon_code_to_process = MIN(secmon_code_to_process, secmon_code_remaining);
+
+    sha256_Update(&ctx->secmon_hash_ctx,
+                  (uint8_t *)chunk_buffer + ctx->secmon_code_offset,
+                  secmon_code_to_process);
+
+    ctx->secmon_code_processed += secmon_code_to_process;
+    ctx->secmon_code_offset = 0;
+
+    if (ctx->secmon_code_processed >= ctx->secmon_code_size) {
```
matejcik, line 498: there's a very weird corner case where `secmon_code_size > firmware_code_size` and the hash check never triggers.

one idea to fix it, check `ctx->secmon_code_size == 0` somewhere at the end (with a comment explaining why)

python/src/trezorlib/_internal/firmware_headers.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/dc7a42eee7858d9f6927d46fb79a2d0e012d531d

python/src/trezorlib/firmware/secmon.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1965a9ec1c138cabe237fc3661fb61d670740e2f

python/src/trezorlib/firmware/secmon.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1965a9ec1c138cabe237fc3661fb61d670740e2f

python/src/trezorlib/firmware/secmon.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/1965a9ec1c138cabe237fc3661fb61d670740e2f

core/SConscript.secmon
TychoVrahe: moved to makefile : https://github.com/trezor/trezor-firmware/commit/e9796175e620fb2f4bef7704474dec020adf13eb

core/embed/util/image/inc/util/image.h
TychoVrahe: removed in https://github.com/trezor/trezor-firmware/commit/cc5847b267dd8606352c1e7faaaa2987a0aec6e9 and https://github.com/trezor/trezor-firmware/commit/2034d6d4f376acdd77e5de6cf19bde5aa6d667b0

core/embed/projects/bootloader/workflow/wf_firmware_update.c
TychoVrahe: added check for secmon size https://github.com/trezor/trezor-firmware/commit/2034d6d4f376acdd77e5de6cf19bde5aa6d667b0

core/embed/util/image/image.c
TychoVrahe: copied from image header, probably doesn't have to, dunno what the original intention was

core/embed/models/T3W1/vendorheader/vendor_unsafe.json
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/8736754a1a1beee7830aaf96012d4c1307878de4

core/embed/models/T3W1/vendorheader/vendor_dev_DO_NOT_SIGN.json
TychoVrahe: removed removed https://github.com/trezor/trezor-firmware/commit/8736754a1a1beee7830aaf96012d4c1307878de4

core/embed/projects/bootloader/main.c
TychoVrahe: so the idea was to place secmon on place, where kernel was originaly placed, that means on aligned address. even though its probably not necessary to have the header aligned, it simplifies some stuff in linker scripts. 

The secmon code itself needs to be aligned, as it contains vector table, so the alignment done few lines above if the header is present.

Thus the final jump addr is IMAGE_CODE_ALIGN(hdr+vhdr) + IMAGE_CODE_ALIGN(secmon_hdr)

core/embed/projects/bootloader/workflow/wf_firmware_update.c
TychoVrahe: for purpose of hashing and secmon_code_size calculation, the padding between header and code is included, so no alignment here is needed

matejcik (approved):

core/embed/util/image/image.c
```diff
@@ -54,6 +54,19 @@ static const uint8_t * const BOOTLOADER_KEYS[] = {
...
+const uint8_t SECMON_KEY_M = 2;
+const uint8_t SECMON_KEY_N = 3;
+static const uint8_t * const SECMON_KEYS[] = {
+#if !PRODUCTION
+  /*** DEVEL/QA KEYS  ***/
+  (const uint8_t *)"\xdb\x99\x5f\xe2\x51\x69\xd1\x41\xca\xb9\xbb\xba\x92\xba\xa0\x1f\x9f\x2e\x1e\xce\x7d\xf4\xcb\x2a\xc0\x51\x90\xf3\x7f\xcc\x1f\x9d",
+  (const uint8_t *)"\x21\x52\xf8\xd1\x9b\x79\x1d\x24\x45\x32\x42\xe1\x5f\x2e\xab\x6c\xb7\xcf\xfa\x7b\x6a\x5e\xd3\x00\x97\x96\x0e\x06\x98\x81\xdb\x12",
+  (const uint8_t *)"\x22\xfc\x29\x77\x92\xf0\xb6\xff\xc0\xbf\xcf\xdb\x7e\xdb\x0c\x0a\xa1\x4e\x02\x5a\x36\x5e\xc0\xe3\x42\xe8\x6e\x38\x29\xcb\x74\xb6",
+#else
+    MODEL_SCMON_KEYS
```
mmilata, line 66: This breaks docker builds in CI (and I'm assuming any other PRODUCTION=1 builds):

```
 embed/util/image/image.c:66:5: error: 'MODEL_SCMON_KEYS' undeclared here (not in a function); did you mean 'SECMON_KEYS'?
   66 |     MODEL_SCMON_KEYS
      |     ^~~~~~~~~~~~~~~~
      |     SECMON_KEYS
```

TychoVrahe: fix here https://github.com/trezor/trezor-firmware/pull/5406

though production=1 still doesn't build because of flash space shortage :/ which will be solved by https://github.com/trezor/trezor-firmware/pull/5320 as i didn't find any quick wins

PR #5209, opened 2025-06-16
Implement noise

onvej-sl:


crypto/noise.h
```diff
@@ -0,0 +1,88 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+// Noise protocol using KK1 handshake pattern and X25519, AES-GCM and SHA256.
+// The handshake messages and the prologue are empty.
+
+#define KEY_SIZE 32
+#define NONCE_SIZE 12
+#define TAG_SIZE 16
```
andrewkozlik: I suggest prefixing these with `NOISE_`.

crypto/noise.h
```diff
@@ -0,0 +1,88 @@
...
+typedef struct {
+  uint8_t encryption_nonce[NONCE_SIZE];
+  uint8_t decryption_nonce[NONCE_SIZE];
+  uint8_t encryption_key[KEY_SIZE];
+  uint8_t decryption_key[KEY_SIZE];
+  bool initialized;
+} noise_context_t;
+
+typedef struct {
+  uint8_t initiator_ephemeral_private_key[KEY_SIZE];
```
andrewkozlik: Wouldn't it be more appropriate to use `curve25519_key` here? It's only a coincidence that X25519 keys are the same size as AES256 keys, which I assume `KEY_SIZE` stands for.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+
+  uint8_t message[KEY_SIZE + 1] = {0};
+  message[0] = 1;
+  hmac_sha256(prk, sizeof(prk), message, 1, output1);
+
+  if (output2) {
+    memcpy(message, output1, KEY_SIZE);
+    message[KEY_SIZE] = 2;
+    hmac_sha256(prk, sizeof(prk), message, KEY_SIZE + 1, output2);
+  }
```
andrewkozlik, line 118: memzero `prk` and `message` here.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  SHA256_CTX sha256_ctx;
+  sha256_Init(&sha256_ctx);
+  sha256_Update(&sha256_ctx, hash, KEY_SIZE);
+  sha256_Update(&sha256_ctx, data, data_length);
+  sha256_Final(&sha256_ctx, hash);
+}
+
+static void hkdf(const uint8_t *salt, size_t salt_length, const uint8_t *key,
+                 size_t key_length, uint8_t output1[KEY_SIZE],
+                 uint8_t output2[KEY_SIZE]) {
```
andrewkozlik: I'd use `SHA256_DIGEST_LENGTH` instead of `KEY_SIZE` everywhere in this function, because [rfc5869](https://datatracker.ietf.org/doc/html/rfc5869) defines all of the lengths here in terms of the hash function output. Same applies to `mix_hash()`. Of course the two constants are identical, so it doesn't make a difference, but seems more semantically accurate to me.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+                                    noise_request_t *request) {
+  random_buffer(state->initiator_ephemeral_private_key, KEY_SIZE);
+  curve25519_scalarmult_basepoint(request->initiator_ephemeral_public_key,
+                                  state->initiator_ephemeral_private_key);
+  return true;
+}
+
+bool noise_handle_handshake_request(
+    noise_context_t *ctx, const uint8_t initiator_public_key[KEY_SIZE],
+    const uint8_t responder_private_key[KEY_SIZE], noise_request_t *request,
```
andrewkozlik: ISTM request can be `const`

crypto/noise.h
andrewkozlik: Same applies to multiple other places, here and in noise.c.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  encrypt(kauth, zero_nonce, handshake_hash, KEY_SIZE, NULL, 0, response->tag);
+
+  // This is unnecessary, as the handshake hash is no longer used.
+  // mix_hash(handshake_hash, NULL, 0);  // Emtpy message
+
+  memset(ctx->encryption_nonce, 0, NONCE_SIZE);
+  memset(ctx->decryption_nonce, 0, NONCE_SIZE);
+
+  ctx->initialized = true;
+
```
andrewkozlik, line 231: memzero `responder_ephemeral_private_key`, `shared_secret`, `chaining_key` and probably `kauth` too. Same applies in next function.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  mix_key(chaining_key, shared_secret, NULL);
+  curve25519_scalarmult(shared_secret, responder_private_key,
+                        request->initiator_ephemeral_public_key);
+  mix_key(chaining_key, shared_secret, kauth);
+  split(chaining_key, ctx->decryption_key, ctx->encryption_key);
+
+  memcpy(response, responder_ephemeral_public_key, KEY_SIZE);
+
+  uint8_t zero_nonce[NONCE_SIZE] = {0};
+  encrypt(kauth, zero_nonce, handshake_hash, KEY_SIZE, NULL, 0, response->tag);
```
andrewkozlik: As already mentioned earlier, here is another one of several places where I feel that `KEY_SIZE` is just strange for something that semantically should be `SHA256_DIGEST_LENGTH` or better `sizeof(handshake_hash)`.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+
+  // This is unnecessary, as the handshake hash is no longer used.
+  // mix_hash(handshake_hash, NULL, 0);  // Emtpy message
```
andrewkozlik: IIRC, `mix_hash()` should be called on the ciphertext, which includes the tag. So if someone uncomments this, it won't give the correct result. Same applies in next function.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  memset(ctx->encryption_nonce, 0, NONCE_SIZE);
+  memset(ctx->decryption_nonce, 0, NONCE_SIZE);
+
+  ctx->initialized = true;
+
+  return true;
+}
+
+bool noise_handle_handshake_response(
+    noise_handshake_state_t *state, noise_context_t *ctx,
```
andrewkozlik: `state` can be const

andrewkozlik: Opinion: it feels like `ctx` should be the first or last parameter here, because either we perceive the function as an initializer of `ctx` (argument for being first) or we perceive `ctx` as being the output of the function (argument for being last). If moved to the end, then the same should be done in `noise_handle_handshake_request()`.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  // This is unnecessary, as the handshake hash is no longer used.
+  // mix_hash(handshake_hash, NULL, 0);  // Emtpy message
+
+  memset(ctx->encryption_nonce, 0, NONCE_SIZE);
+  memset(ctx->decryption_nonce, 0, NONCE_SIZE);
+
+  ctx->initialized = true;
+
```
andrewkozlik, line 292: Actually, it would make sense to memzero the `state` here, in which case it can't be const.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+    return false;
+  }
+  if (!encrypt(ctx->encryption_key, ctx->encryption_nonce, associated_data,
+               associated_data_length, plaintext, plaintext_length,
+               ciphertext)) {
+    return false;
+  }
+  if (!increase_nonce(ctx->encryption_nonce)) {
+    // Nonce overflow
+    ctx->initialized = false;
```
andrewkozlik, line 310: I'd rather memzero the entire context. Same in next function.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+                    size_t plaintext_length, uint8_t *ciphertext) {
+  // ciphertext = AES-GCM-Encrypt(key, nonce, associated_data, plaintext)
+  const size_t ciphertext_length = plaintext_length + TAG_SIZE;
+
+  uint8_t buffer[plaintext_length + TAG_SIZE];
+  memcpy(buffer, plaintext, plaintext_length);
+
+  gcm_ctx ctx = {0};
+  if (gcm_init_and_key(key, KEY_SIZE, &ctx) != RETURN_GOOD) {
+    return false;
```
andrewkozlik, line 44: We should be memzeroing the buffer in case of error, or initializing it later, or better yet, not using it at all and using the buffer provided by the caller `ciphertext`, so that we don't have to copy stuff around and worry about cleanup.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+                    size_t associated_data_length, const uint8_t *plaintext,
+                    size_t plaintext_length, uint8_t *ciphertext) {
+  // ciphertext = AES-GCM-Encrypt(key, nonce, associated_data, plaintext)
+  const size_t ciphertext_length = plaintext_length + TAG_SIZE;
+
+  uint8_t buffer[plaintext_length + TAG_SIZE];
+  memcpy(buffer, plaintext, plaintext_length);
+
+  gcm_ctx ctx = {0};
+  if (gcm_init_and_key(key, KEY_SIZE, &ctx) != RETURN_GOOD) {
```
andrewkozlik: Why not keep the context in `noise_context_t`?

andrewkozlik: In case we keep it, the context should be memzeroed before exiting the function.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  }
+  if (gcm_decrypt_message(nonce, NONCE_SIZE, associated_data,
+                          associated_data_length, buffer, plaintext_length,
+                          buffer + plaintext_length, TAG_SIZE,
+                          &ctx) != RETURN_GOOD) {
+    return false;
+  }
+
+  memcpy(plaintext, buffer, plaintext_length);
+  return true;
```
andrewkozlik, line 90: memzero `buffer` before return or even better, just use directly the `plaintext` buffer that is already allocated.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+#include "aes/aesgcm.h"
+#include "ed25519-donna/ed25519.h"
+#include "hmac.h"
+#include "rand.h"
+#include "sha2.h"
+
+_Static_assert(KEY_SIZE == SHA256_DIGEST_LENGTH,
+               "KEY_SIZE must be equal to SHA256_DIGEST_LENGTH");
+
+static uint8_t protocol_name[32] = {'N', 'o', 'i', 's', 'e', '_',  'K',  'K',
```
M1nd3r: Nit: I would either use `SHA256_DIGEST_LENGTH` (or `KEY_SIZE`) directly in the definition of size of `protocol_name`, or `_Static_assert` that the length of `protocol_name` is equal to the length of the SHA256 digest (`HASHLEN` in noise spec). (Note: If we were to change the protocol name later to something longer than `HASHLEN`, we would need to remove that `_Static_assert` -  but I assume we do not need this flexibility in this implementation.)

Excerpt from docs for reference:
> If protocol_name is less than or equal to HASHLEN bytes in length, sets h equal to protocol_name with zero bytes appended to make HASHLEN bytes. Otherwise sets h = HASH(protocol_name).

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+                                    '1', '_', '2', '5', '5', '1',  '9',  '_',
+                                    'A', 'E', 'S', 'G', 'C', 'M',  '_',  'S',
+                                    'H', 'A', '2', '5', '6', 0x00, 0x00, 0x00};
+
+static bool encrypt(const uint8_t key[KEY_SIZE],
+                    const uint8_t nonce[NONCE_SIZE],
+                    const uint8_t *associated_data,
+                    size_t associated_data_length, const uint8_t *plaintext,
+                    size_t plaintext_length, uint8_t *ciphertext) {
```
M1nd3r, line 40: Why are there the fields `associated_data_length` and `plaintext_length`? Is there some reason why not just use `sizeof` inside the encrypt function - timing attack or similar, or is there some other reason for putting this responsibility on the caller?

Similar question for the public API in `noise.h` in `noise_send_message` and `noise_receive_message`.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  uint8_t buffer[plaintext_length + TAG_SIZE];
+  memcpy(buffer, plaintext, plaintext_length);
+
+  gcm_ctx ctx = {0};
+  if (gcm_init_and_key(key, KEY_SIZE, &ctx) != RETURN_GOOD) {
+    return false;
+  }
+  if (gcm_encrypt_message(nonce, NONCE_SIZE, associated_data,
+                          associated_data_length, buffer, plaintext_length,
+                          buffer + plaintext_length, TAG_SIZE,
```
M1nd3r: I would like to know the following. What would happen if the `caller` set the `plaintext_length` to something negative (or, in general, something other that the real plaintext length)? Would it be possible, for instance, to rewrite "unexpected" part of the memory by the TAG using the `gcm_encrypt_message` function? What is preventing it? 

I am going to test it later, but if you know the answer - or why the question is irrelevant - I would be glad to hear it.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  sha256_Init(&sha256_ctx);
+  sha256_Update(&sha256_ctx, hash, KEY_SIZE);
+  sha256_Update(&sha256_ctx, data, data_length);
+  sha256_Final(&sha256_ctx, hash);
+}
+
+static void hkdf(const uint8_t *salt, size_t salt_length, const uint8_t *key,
+                 size_t key_length, uint8_t output1[KEY_SIZE],
+                 uint8_t output2[KEY_SIZE]) {
+  // output1 || output2 = HKDF(salt, key, 2 * KEY_SIZE)
```
M1nd3r: Nit: I like the optimization - not computing the `output2` when it is not needed. IMO it deserves a better comment/documentation - because if `mix_key` is called with `output_key=NULL`, then so is the `hkdf` called with `output2=NULL`, meaning that this comment is not always True. It would be nice to have also comment describing the situation for `output2=NULL`.

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+  if (output2) {
+    memcpy(message, output1, KEY_SIZE);
+    message[KEY_SIZE] = 2;
+    hmac_sha256(prk, sizeof(prk), message, KEY_SIZE + 1, output2);
+  }
+}
+
+static void mix_key(uint8_t chaining_key[KEY_SIZE], uint8_t input_key[KEY_SIZE],
+                    uint8_t output_key[KEY_SIZE]) {
+  // chaining_key || output_key = HKDF(chaining_key, input_key, 2 * KEY_SIZE)
```
M1nd3r: Nit: What do you mean by the `2 * KEY_SIZE`? What parameter should it be? In the Noise spec, the 3rd parameter of HKDF is "number of outputs" - so why the `* KEY_SIZE`? And if it is referring to the function `hkdf` in this file, I do not understand what it should mean. Also, it seems to me that the `2 * KEY_SIZE` is not always true as the `output2` can be set to `NULL`. and the second branch is not computed. Am I missing something?

crypto/noise.c
```diff
@@ -0,0 +1,300 @@
...
+
+  uint8_t zero_nonce[NONCE_SIZE] = {0};
+  if (!decrypt(kauth, zero_nonce, handshake_hash, KEY_SIZE, response->tag,
+               TAG_SIZE, NULL)) {
+    // Wrong tag
+    return false;
+  }
+
+  // This is unnecessary, as the handshake hash is no longer used.
+  // mix_hash(handshake_hash, NULL, 0);  // Emtpy message
```
M1nd3r: Nit: typo in the word `Empty`

crypto/tests/test_check.c
```diff
@@ -11400,6 +11401,142 @@ START_TEST(test_elligator2) {
...
+
+  noise_handshake_state_t initiator_handshake_state = {0};
+  bool ret = false;
+
+  // Initiator sends request
+  ret = noise_create_handshake_request(&initiator_handshake_state, &request);
+  ck_assert_int_eq(ret, true);
+  ck_assert_mem_eq(&request, fromhex(expected_request_hex), sizeof(request));
+
+  // Responder receivesd request and sends response
```
M1nd3r: Nit: typo in the word `receives`

crypto/noise.c
M1nd3r: Nit: typo in the word `Empty`;

crypto/noise.c
andrewkozlik: This is OK. In C, the function does not know any details about how many elements the pointer points to. It's just a pointer to some memory, so you can't use `sizeof` that way. `sizeof(associated_data)` would be 4 (size of a pointer) and `sizeof(*associated_data)` would be 1 (size of one uint8_t).

crypto/noise.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/78a5726ac0d6f0c13184e873e7ecf4b5d8ec810c.

crypto/noise.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/d007a8f68cf0099813cb04188a061a6384217110.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/2e4749cda28cb775ac84184c895a22f5f767a951.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/e1a3a8e9b4f6dbe71da234ce51880ed83c57f8b8.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/aaea854d8c2903f0c3aeab8c9016972b69a05556.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/2e4749cda28cb775ac84184c895a22f5f767a951.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/e1a3a8e9b4f6dbe71da234ce51880ed83c57f8b8.

crypto/noise.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5209/commits/559982381db2aba04916a5b9ed5d6c80f10f7eab.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/cb5e008d4b26a3486909f19faedb5acd225ef0e4 by removing `state` completely.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/2e4749cda28cb775ac84184c895a22f5f767a951.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/176aff9ec847dfc9fae27e4efd08d4311fdee6bf.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/2e4749cda28cb775ac84184c895a22f5f767a951.

crypto/noise.c
onvej-sl: > Why not keep the context in noise_context_t?

I added an explaining comment in https://github.com/trezor/trezor-firmware/pull/5209/commits/1e0fcbe8bad30d368bc111cd969c53b30af707e8.



> In case we keep it, the context should be memzeroed before exiting the function.

Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/2e4749cda28cb775ac84184c895a22f5f767a951.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/commit/2e4749cda28cb775ac84184c895a22f5f767a951.

crypto/noise.c
onvej-sl: In https://github.com/trezor/trezor-firmware/pull/5209/commits/e1a3a8e9b4f6dbe71da234ce51880ed83c57f8b8, I used `SHA256_DIGEST_LENGTH`.

crypto/noise.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5209/commits/1e0fcbe8bad30d368bc111cd969c53b30af707e8.

crypto/tests/test_check.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5209/commits/1e0fcbe8bad30d368bc111cd969c53b30af707e8.

crypto/noise.c
onvej-sl: In https://github.com/trezor/trezor-firmware/pull/5209/commits/f23cb9987b626ae6e58d23843c1e4ea5ef7d497f, I removed  `buffer`.

crypto/noise.c
onvej-sl: In https://github.com/trezor/trezor-firmware/pull/5209/commits/f23cb9987b626ae6e58d23843c1e4ea5ef7d497f, I removed  `buffer`.

crypto/noise.c
onvej-sl: > Would it be possible, for instance, to rewrite "unexpected" part of the memory by the TAG using the gcm_encrypt_message function?

Yes, this could hypothetically happen if `plaintext_length + NOISE_TAG_SIZE` overflowed `size_t`. I believe this problem was fixed in https://github.com/trezor/trezor-firmware/pull/5209/commits/f23cb9987b626ae6e58d23843c1e4ea5ef7d497f by removing `buffer` completely.

crypto/noise.c
onvej-sl: Do you have a suggestion for a concise comment that could express that?

To be honest, I'm not convinced that this is worth commenting. In C, it's quite common for `NULL` to be used to "ignore" the output, similar to the use of _ in Python.

crypto/noise.c
onvej-sl: > Nit: What do you mean by the 2 * KEY_SIZE? What parameter should it be? In the Noise spec, the 3rd parameter of HKDF is "number of outputs" - so why the * KEY_SIZE? And if it is referring to the function hkdf in this file, I do not understand what it should mean.

The parameter is intended to denote the length of the output generated by HKDF. In https://github.com/trezor/trezor-firmware/pull/5209/commits/1e0fcbe8bad30d368bc111cd969c53b30af707e8, I explicitly wrote the name of the parameter.

crypto/noise.h
```diff
@@ -0,0 +1,95 @@
...
+  // There is a time-memory trade-off between storing encryption/decryption keys
+  // and storing encryption/decryption contexts, we choose to optimize for
+  // memory usage by storing the keys
+  uint8_t encryption_key[NOISE_KEY_SIZE];
+  uint8_t decryption_key[NOISE_KEY_SIZE];
+  bool initialized;
+} noise_context_t;
+
+typedef struct {
+} noise_handshake_state_t;
```
andrewkozlik: Empty and unused. Please remove.

crypto/noise.h
```diff
@@ -0,0 +1,95 @@
...
+typedef struct {
+  curve25519_key initiator_ephemeral_public_key;
+} noise_request_t;
+
+typedef struct {
+  curve25519_key responder_ephemeral_public_key;
+  uint8_t tag[NOISE_TAG_SIZE];
+} noise_response_t;
+
+// This is called by the initiator to create the handshake request
```
andrewkozlik: It would be helpful to document here that this function initializes the initiator's Noise context and similarly that `noise_handle_handshake_request()` initializes the responder's Noise context, just to make it clear that these are the first functions that are called on the context.

crypto/noise.c
```diff
@@ -0,0 +1,332 @@
...
+      return true;
+    }
+  }
+
+  // Nonce overflow
+  return false;
+}
+
+bool noise_create_handshake_request(noise_context_t *ctx,
+                                    noise_request_t *request) {
```
andrewkozlik, line 159: I recommend `memzero(ctx, sizeof(*ctx));` here.

andrewkozlik (approved):
LGTM :+1:

crypto/noise.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/3eed13b454bb02058b1f2d04762c057dad1de932.

crypto/noise.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/3eed13b454bb02058b1f2d04762c057dad1de932.

crypto/noise.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5209/commits/3eed13b454bb02058b1f2d04762c057dad1de932.

crypto/noise.c
M1nd3r: Thanks for clarification

crypto/noise.c
M1nd3r: IMO it is ok as it is.

The readability is ok and I trust your judgement with regards to C customs, so no change is required.

Just to answer - how would I do it? I would add two new lines - instead of
`  // output1 || output2 = HKDF(salt, key, output_length=2*SHA256_DIGEST_LENGTH)`,
I would have used something like
```
  // output1 || output2 = HKDF(salt, key, output_length=2*SHA256_DIGEST_LENGTH)
  // OR - when not(output2)
  // output1 = HKDF(salt, key, output_length=SHA256_DIGEST_LENGTH)
```

PR #5056, opened 2025-05-18
Rewrite `trezor.log` in Rust

romanz:

PR #5214, opened 2025-06-17
fix(core): re-add logging level filtering

romanz:
It has been removed by mistake in #5056.


M1nd3r (approved):
Well done. This looks very concise, thanks!

PR #5215, opened 2025-06-17
fix install fw confirm title issue

TychoVrahe:


obrusvit (approved):
LGTM.

Tested the screen by putting it into `Ping` handler.
Tests of interactionless update UI don't exist but maybe we can at least test the `confirm_firmware_update` function call :thinking:

PR #5217, opened 2025-06-17
chore(core): fix "trezor" -> "Trezor"

romanz:
Also, remove duplicated `coinjoin__title_do_not_disconnect` string.


obrusvit (approved):
LGTM.

I think previously there may have been two separate strings for the "Do not disconnect" screen, one was the title and one was the body. Apparently, later on only one was supposed to be used.

PR #5219, opened 2025-06-17
fix nordic gitignore.

kopecdav:


TychoVrahe (approved):
copilots suggestions seem reasonable, otherwise ack

Issue #4741, opened 2025-03-05
SignIdentity and GetECDHSessionKey


andrewkozlik:
There are several issues with the way the `SignIdentity` message ([SLIP-0013](https://github.com/satoshilabs/slips/blob/master/slip-0013.md)) and `GetECDHSessionKey` message ([SLIP-0017](https://github.com/satoshilabs/slips/blob/master/slip-0017.md)) are handled in firmware. Some of these have been identified by Mathias Herberts, others internally.

1. According to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) the `proto` part and the `host` part of a URI should be treated as case-insensitive. The SLIPs do not explain how to achieve that and the implementation treats both as case sensitive. We should amend the SLIPs to specify normalization of `proto` and `host` to lowercase in the derivation calculation and implement the conversion to lowercase in Trezor as well.

2. The `://` string, as per RFC 3986 [section 3](https://datatracker.ietf.org/doc/html/rfc3986#section-3), should only be included when the URI has an authority part. For example the simple URI `foo:bar` with a scheme of `foo` and a path of `bar` is a valid URI as per RFC 3986 but it is incorrectly reconstructed as `foo://bar` by `serialize_identity()`, turning it into a URI with scheme `foo` (correct) and host `bar` (incorrect).

3. According to the SLIPs `proto://` is mandatory in the URI, but both Core and Legacy treat is as optional, i.e. if the `proto` parameter is absent, then the `://` is not prepended.

4. According to the SLIPs when `path` is present in the identity, the URI should be constructed by appending a `/` followed by the `path`. However, neither Legacy nor Core firmware place the `/` before the `path`.

5. In Legacy firmware if `host` is absent from the identity, then `layoutDecryptIdentity()` does not display `port`, but `cryptoIdentityFingerprint()` does use `port` in the path computation.
 
6. Legacy firmware does not display `path`, while `core` firmware does.

7. The parameters for the device test that we are using [here](https://github.com/trezor/trezor-firmware/blob/45adcd963a4f992b6545f6d52ad4a9412674db57/tests/device_tests/misc/test_msg_getecdhsessionkey.py#L27-L33) are weird `host="Satoshi Nakamoto <satoshi@bitcoin.org>"`. I'd expect `user= "satoshi", host="bitcoin.org"`.

[...]

PR #5220, opened 2025-06-17
Generate Evolu keys

obrusvit:
This PR implements generating SLIP-21 node for [evolu](https://github.com/evoluhq/evolu) by Trezor device.

The following `trezorctl` command triggers the feature:
```
trezorctl evolu get-node
```
The resulting SLIP-21 node data is returned in hex format:
```
data: a81aaf51997b6ddfa33d11c038d6aba5f711754a2c823823ff8b777825cdbb32b0e71c301fa381c75081bd3bcc134b63306aa6fc9a9f52d835ad4df8cd507be6
```

Basic test against `"all all ... all"` seed included.

TODO:
- [x] update test vestors based on the new SLIP-21 path
- [x] rebase on `main` and update fixtures (new tests)
- [x] drop the DROP commit


core/src/trezor/messages.py
```diff
@@ -3936,6 +3936,30 @@ def __init__(
...
+        encryption_key: "bytes"
+
+        def __init__(
+            self,
+            *,
+            owner_id: "bytes",
+            write_key: "bytes",
+            encryption_key: "bytes",
+        ) -> None:
+            pass
```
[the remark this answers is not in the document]
obrusvit: Auto-generated file

obrusvit:
Fixups:
- use the correct messages block (starting `2100`) https://github.com/trezor/trezor-firmware/pull/5220/commits/8a1fbef78e4748810a6d5cff664e1937b0779dc7
- ignore Evolu messages on Legacy https://github.com/trezor/trezor-firmware/pull/5220/commits/cf48add2de070907662029a590aef76ddc96e45c

core/src/boot.py
```diff
@@ -115,8 +115,8 @@ async def bootscreen() -> None:
 config.init(show_pin_timeout)
 translations.init()
 
-if __debug__ and not utils.EMULATOR:
-    config.wipe()
+# if __debug__ and not utils.EMULATOR:
+#     config.wipe()
```
mmilata: leftover?

obrusvit: Thanks, the commit is marked as `DROP`, I will drop it with rebase.

obrusvit: I added a note about Evolu on proto defs and in CLI https://github.com/trezor/trezor-firmware/pull/5220/commits/3214052aa274bd9710ae5028bdfdadead664bc0d and https://github.com/trezor/trezor-firmware/pull/5220/commits/13e504a660434cbb16fbfafde2a3734e6aab072b

core/src/apps/evolu/get_keys.py
```diff
@@ -0,0 +1,41 @@
...
+        "Evolu keys",
+        action="Derive keys for Evolu?",
+        prompt_screen=True,
+    )
+    seed = await get_seed()
+    node1 = Slip21Node(seed)
+    node2 = node1.clone()
+    node3 = node1.clone()
+
+    node1.derive_path([b"Evolu", b"Owner Id"])
```
M1nd3r: Please, add "TREZOR" as the first level node .
See: https://github.com/trezor/trezor-firmware/pull/5301

obrusvit: See here: https://github.com/trezor/trezor-firmware/pull/5220/commits/0a0dc975da68549638078a7ea8c92d52dd048a28

I asked Peter Sanderson for new test vectors

core/src/apps/evolu/get_keys.py
```diff
@@ -0,0 +1,43 @@
...
+    node3 = node1.clone()
+
+    node1.derive_path(_EVOLU_KEY_PATH_PREFIX + [b"Owner Id"])
+    owner_id = node1.key()
+
+    node2.derive_path(_EVOLU_KEY_PATH_PREFIX + [b"Write Key"])
+    write_key = node2.key()
+
+    node3.derive_path(_EVOLU_KEY_PATH_PREFIX + [b"Encryption Key"])
+    encryption_key = node3.key()
```
M1nd3r: nit: To avoid using the `+` in each path derivation, the paths can be derived sequentially:
```suggestion
    node1 = Slip21Node(seed)
    node1.derive_path(_EVOLU_KEY_PATH_PREFIX)
    node2 = node1.clone()
    node3 = node1.clone()

    node1.derive_path([b"Owner Id"])
    owner_id = node1.key()

    node2.derive_path([b"Write Key"])
    write_key = node2.key()

    node3.derive_path([b"Encryption Key"])
    encryption_key = node3.key()
```

python/src/trezorlib/cli/evolu.py
```diff
@@ -0,0 +1,49 @@
...
+from .. import evolu, messages
+from . import with_client
+
+if t.TYPE_CHECKING:
+    from ..client import TrezorClient
+
+
+@click.group(name="evolu")
+def cli() -> None:
+    """Evolu commands. Evolu is a local first storage framework. See https://github.com/evoluhq/evolu"""
```
M1nd3r, line 32: 🤔 Should we let user know (at least in the docs) that evolu commands are available only on debug?

core/src/apps/evolu/get_keys.py
M1nd3r: EDIT: Never mind the following nit. The unit tests can be done even when the function has await calls to ui, so there is not a lot of added value.

nit: One think that _I would like_ is to extract the part with node/keys generation into a separate function that could be tested using python unit tests. 

```
def _derive_keys(seed: bytes)->tuple[bytes,bytes,bytes]:
   node1 = Slip21Node(seed)
   ...
   return owner_id, write_key, encryption_key
```
and use it as:

```
async def get_keys(...):
   ...
   await confirm_action(...)
   return _derive_keys(seed=await gets_seed())
```

This way, the function for deriving keys (`_derive_keys`) can be easily (and quickly) tested with multiple seeds with unit tests (`core/tests`).

@obrusvit WDYT?

tests/device_tests/evolu/test_evolu.py
```diff
@@ -0,0 +1,41 @@
...
+
+ALL_MNEMONIC = "all all all all all all all all all all all all"
+
+EXPECTED_OWNER_ID = "0940d9f3e307f3bcedbcc8361ae136b619603a686386ecd329c3ed2337cb831d"
+EXPECTED_WRITE_KEY = "2a65e7495bc4edc8ad365c5278404467fc2ec1b3e6978e2406f59329d5357f3e"
+EXPECTED_ENCRYPTION_KEY = (
+    "9ad06a43e9d4739502d59c8cafdaa6babb0481cdd0b3acb8455f080b38847642"
+)
+
+pytestmark_all_mnemonic = pytest.mark.setup_client(mnemonic=ALL_MNEMONIC)
```
M1nd3r: The `pytestmark_all_mnemonic` seems unnecessary. In `conftest.py`, the `client` is automatically set up with the `"all ..."` mnemonic, unless there is a different setup defined:

```
# in conftest.py

setup_params = dict(
        uninitialized=False,
        mnemonic=" ".join(["all"] * 12),
        pin=None,
        passphrase=False,
        needs_backup=False,
        no_backup=False,
    )
```

The `ALL_MNEMONIC` should be either removed (preferably), or moved next to the other commonly used mnemonics - into `tests/common.py`.

core/src/apps/evolu/get_keys.py
obrusvit: I think we can test via `device_tests` vectors.

tests/device_tests/evolu/test_evolu.py
obrusvit: Absolutely right, thanks.
https://github.com/trezor/trezor-firmware/pull/5220/commits/f9bfaa0e47d66cbfceb4918fcfbaeedf1f951fcb

python/src/trezorlib/cli/evolu.py
obrusvit: Why not.
https://github.com/trezor/trezor-firmware/pull/5220/commits/3829d709ed331c53df37e3d626c2f3963c6ed513

core/src/apps/evolu/get_keys.py
obrusvit: Fix for the first suggestion (removing repeated `+`). Thanks.
https://github.com/trezor/trezor-firmware/pull/5220/commits/41b8bb52bc7bddc5608e335277f9495364fdcdcb

core/src/apps/evolu/get_keys.py
obrusvit: https://github.com/trezor/trezor-firmware/pull/5220/commits/cb12cfee2b1921d4fe7d470675967527f07e58ad

core/src/apps/evolu/get_keys.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if not is_initialized():
+        raise NotInitialized("Device is not initialized")
+
+    # TODO: use translated strings when exposing this to production
+    await confirm_action(
+        "get_evolu_keys",
+        "Evolu keys",
+        action="Derive keys for Evolu?",
+        prompt_screen=True,
```
romanz: Consider allowing the user to view the derived keys (e.g. via a menu) - otherwise, I am not sure the user can verify that the returned keys in the response.

Following the comment [here](https://github.com/trezor/trezor-firmware/issues/4741#:~:text=Trezor%20doesn%27t%20have%20a%20function%20to%20display%20a%20SLIP%2D0013%20or%20SLIP%2D0017%20derived%20public%20key%20or%20public%20key%20fingerprint%20along%20with%20the%20corresponding%20identity.%20Without%20this%20feature%2C%20users%20have%20no%20secure%20means%20to%20verify%20that%20the%20BIP%2032%20path%20of%20their%20key%20was%20generated%20for%20the%20correct%20identity):

> Trezor doesn't have a function to display a SLIP-0013 or SLIP-0017 derived public key or public key fingerprint along with the corresponding identity. Without this feature, users have no secure means to verify that the BIP 32 path of their key was generated for the correct identity.

M1nd3r: @obrusvit Please, do not spend a lot time on this particular solution. In R&D, we would like to move the derivation of the `owner_id` and `keys` to the Suite. Trezor should send only  `evolu_secret` from which the id/keys will be derived. This approach allows for easier extendability in the future.

obrusvit:
I rebased on top of `main` and also made changes:
- the handler was moved to `apps/misc` because having `apps/nostr` seems like an overkill
- it's now also in production FW
- the handler was reworked as of the comment from Petr https://github.com/trezor/trezor-firmware/pull/5220#discussion_r2236559249

PR #5221, opened 2025-06-17
common/protob: THP: required fields & separate ThpMessageType

mmilata:
Followup of #5191. Related to #4976.

* some fields of THP messages are changed from optional to required, please check 
* remove THP message IDs from `MessageType` enum and put them into their own `ThpMessageType`
* add `wire_enum` message attribute so that pb2py can correctly find wire type


common/protob/messages-thp.proto
```diff
@@ -10,6 +10,38 @@ import "options.proto";
 option (include_in_bitcoin_only) = true;
 
 
+/**
+ * Mapping between Trezor wire identifier (uint) and a Thp protobuf message
+ */
+enum ThpMessageType {
+    reserved 0 to 999;  // Values reserved by other messages, see messages.proto
+
+    ThpMessageType_ThpCreateNewSession = 1000 [(bitcoin_only) = true, (wire_in) = true];
```
matejcik: this one should be part of `MessageType`, no? it's a trezor wire protocol message, analogous to `Initialize`

matejcik: you can probably omit the extensions (`bitcoin_inly`, `wire_in`, `wire_out`)

core isn't using `wire_in` and `wire_out` at all, these were for legacy

`bitcoin_only` doesn't apply unless the enum has `has_bitcoin_only_values`, which this one doesn'tn

common/protob/messages-thp.proto
```diff
@@ -10,6 +10,38 @@ import "options.proto";
 option (include_in_bitcoin_only) = true;
 
 
+/**
+ * Mapping between Trezor wire identifier (uint) and a Thp protobuf message
+ */
+enum ThpMessageType {
+    reserved 0 to 999;  // Values reserved by other messages, see messages.proto
```
matejcik: we don't have to reserve the values, we don't need `ThpMessageType` values to not conflict -- although there was some request by @szymonlesisz ?

common/protob/messages-thp.proto
```diff
@@ -10,6 +10,38 @@ import "options.proto";
...
+
+    ThpMessageType_ThpCreateNewSession = 1000 [(bitcoin_only) = true, (wire_in) = true];
+    reserved 1001 to 1005;  // never appeared in a release, reserved for future use
```
matejcik, line 26: what's the point of these reservations?

common/protob/messages-thp.proto
```diff
@@ -25,10 +57,10 @@ enum ThpPairingMethod {
  * @embed
  */
 message ThpDeviceProperties {
-    optional string internal_model = 1;               // Internal model name e.g. "T2B1".
+    required string internal_model = 1;               // Internal model name e.g. "T2B1".
     optional uint32 model_variant = 2;                // Encodes the device properties such as color.
```
matejcik: do we want to set a `default=0` here?

common/protob/messages-thp.proto
M1nd3r: Well, yes and no. It is somewhat analogous to `Initialize`, but it is available only in THP builds. AFAIK @szymonlesisz would like to have THP as a standalone/separate module (please, correct me if I am mistaken). 

To me, it makes sense to keep all THP messages in one place.

Also, I do not think that `ThpCreateNewSession` should be treated differently than `ThpCredentialRequest` - as credential request can be sent either during pairing process but also later during "standard" communication.

common/protob/messages-thp.proto
M1nd3r: Personally, I would like to avoid conflicts, as some messages can change their scope of usage. For instance, we now allow to ask for a credential not only after pairing but also during "standard" communication. I see no value in having two distinct messages that would do the same thing.

common/protob/messages-thp.proto
M1nd3r: In the original PR, values 1000-1099 are reserved for THP (in `messages.proto`) and there was no need for reserving the small gaps inside the ThpMessageType enum.

These small  gaps in the definitions are intentional and are part of the design.

common/protob/messages-thp.proto
M1nd3r: I am not sure about the `bitcoin_only`, as I recall there were some problems - that THP messages were not available on bitcoin-only builds. Ofc feel free to change it, but test it on BTC-only afterwards, please.

common/protob/messages-thp.proto
mmilata: I've added them to stay consistent with `MessageType` which seems to avoid gaps inside each "block", doesn't really help anything I guess.

matejcik: > These small gaps in the definitions are intentional and are part of the design.

what is the rationale?

common/protob/messages-thp.proto
matejcik: > as some messages can change their scope of usage

they **absolutely should not** do that.

the Trezor wire protocol and the THP negotiation are two separate realms, two separate applications on the application layer.
**There is no reason for them to share messages.** The fact that both use protobuf encoding is an **implementation detail**. They are not the same thing.

If you want to expose the same functionality via both applications, using (a) two separate messages or (b) the same definition, explicitly exposed in two different places, under either the same or two different ids, is the correct solution.

The meaning of "id" in the Trezor wire protocol and the meaning of "id" in THP negotiation are two distinct things. If you are looking up the ids in the same place, you're doing it wrong.

common/protob/messages-thp.proto
matejcik: > To me, it makes sense to keep all THP messages in one place.

that's not how any of this works. `ThpCreateNewSession` will be recognized by a different "application" on the THP application layer, and as such, it must go into the list of messages _for that other application_ (which is `MessageType`)

> as credential request can be sent either during pairing process but also later during "standard" communication.

if that's necessary, that's a new separate feature of the Trezor wire protocol, and as such the messages need to be separately exposed in the Trezor wire protocol.

common/protob/messages-thp.proto
matejcik: no, wait, sorry. maybe i'm misunderstanding the pairing phase

M1nd3r: > the Trezor wire protocol and the THP negotiation are two separate realms, two separate applications on the application layer.

I agree that the THP **negotiation** has nothing to do with the rest of the trezor wire protocol. This is definitely true for channel allocation and handshake phase, possibly a pairing as it is required for standard communication. BUT if I have a look at credential issuance, it can be considered to be standalone workflow - and it can be called during the "standard" communication.

1. It possibly fits more to the Trezor - wire (a.k.a. `MessageType`) enum as it is not exclusive to the THP -connection process.

2. It has no meaning outside the THP realm. If the device uses codec_v1 protocol, THP-credential messages should not be present in the FW.


If I understand you correctly, You would like to see:

- THP messages separated so that they can be excluded from non-THP builds.
- THP messages divided between "Trezor-wire" and "THP-negotiation" which are handled completely separately.

>  If you are looking up the ids in the same place, you're doing it wrong.


In THP PR https://github.com/trezor/trezor-firmware/pull/4976, I am using the same protobuf-handling functions for all protobuf. I do not have a separate protobuf parser for THP messages. The separation between the realm of available workflow_handlers and possible messages is done in the following way:
- The channel has a `state` - depending on it, a received messages are handled by different functions.
- There are different contexts (eg. `pairing_context`) that allow for specific `workflow_handlers` to be invoked.

If you envision a different kind of separation, please, have a look at the THP implementation in https://github.com/trezor/trezor-firmware/pull/4976/ and leave comments or change proposals there, so it can be adujsted

- `received_message_handler` is responsible for treating received messages based on the `state` of the channel the message was received on.
- `message_handler.wrap_protobuf_load` handles all protobuf messages - both wire and THP

M1nd3r: > If you are looking up the ids in the same place, you're doing it wrong.

I do not fully understand the reasoning behind this. Should we have separate enums for **all** apps? Should eg. Solana have its own `SolanaMessageType` with its own independent id numbering system, because it is a different `app`? There are also other similarities. For instance, we exclude Solana on BITCOIN_ONLY builds as we exclude THP messages (like ThpCredentialRequest) on non-THP builds. 

I am not saying that separation is not a nice thing to have, but I am not convinced that it is in line with the rest of the FW structure so far.

We possibly could (in the future) make message_type 4 bytes instead of the current 2, where the first two bytes denote the app (THP/Solana/Bitcoin/...) and the second 2 bytes denote the id in the appropriate MessageType enum. WDYT?

common/protob/messages-thp.proto
M1nd3r: There are several rationale in favor gaps, nothing particularly strong. 

- Future proofing: If a new message appears, it can be added next to the messages it is related to. For example, in case the flow for CodeEntry pairing will get new message, it can be added to the other CodeEntry messages.
- Avoid unnecessary changes during development: If there is a new message in the beginning of the pairing flow (assuming we would like to keep the messages in a meaningful order), we do not need to shift ids of all pairing messages. Eg, `ThpNfcTagHost` will keep its `id=1032`.
- Readability (not a strong reason): Different pairing methods are aligned by the 8, (1016,1024,1032,...) - if you see the binary of message id, you can see that the messages belong to the same pairing method - and as they are ordered, first CodeEntry message will end with 000b, second with 001b and so on (would be nicer with 16-bit blocks for hex). 

However, I also do not see a strong reason against this aligning to blocks by 8.

M1nd3r: > Avoid unnecessary changes during development

This is exactly how we ended up with messages with id 1006 and 1007 - as we added new messages and wanted to minimize the impact (need for changes) on other environments that already used the messages 1008-1013.

matejcik: right, i see. that was exactly what was bugging me though: (a) unaligned space between 1-5 and (b) too little space between 14-15

okay, let's keep it as is, but it would be nice if we could remember to go back and renumber the fields before we release? :) keeping the gaps for *future* extensibility

M1nd3r: Ok, lgtm. I would renumber the messages in a separate PR (after discussion with Szymon - when/how it would be best). 
There is one remaining question - should the small gaps have their own `reserved X to Y // never used, reserved for future use` blocks, or are those unnecessary?

common/protob/messages-thp.proto
M1nd3r: currently, handler of `ThpCreateNewSession` is hidden under `utils.USE_THP` flag in `base.py` - there is no dedicated `app` for this (If I understand the concept of apps correctly, which might not be fully true). 

- I thought that "exposure" to Trezor wire protocol is done by registering a workflow handler for the particular message. THP-only messages do not have registered handlers in the Trezor-wire list of handlers and thus are not accessible, where THP-message-handlers do not use the Trezor-wire list of wf_handlers and use custom list of workflow handlers (or possibly a single wf handler).
- Exceptions to this are handlers for `ThpCreateNewSession` and `ThpCredentialRequest` messages, both of those handlers are registered directly from `base.py` without any separate application (but hidden under utils.USE_THP flag)

common/protob/messages-thp.proto
matejcik: they're nice to have :woman_shrugging:

common/protob/messages-thp.proto
matejcik: > that THP messages were not available on bitcoin-only builds

that's presumably because `MessageType` enum opts into `has_bitcoin_only_values` whereas `ThpMessageType` does not

common/protob/pb2py
```diff
@@ -245,6 +245,12 @@ class ProtoMessage:
 
         if "wire_type" in extensions:
             wire_type = extensions["wire_type"]
+        elif "wire_enum" in extensions:
+            message_type_enum = find_by_name(
+                descriptor.enums, extensions["wire_enum"].bytes.decode()
+            )
+            message_type = find_by_name(message_type_enum.value, message.name)
```
matejcik:
```suggestion
            custom_message_type = find_by_name(message_type_enum.value, message.name)
            if message_type is not None and custom_message_type != message_type:
                # we are specifying two different message ids via MessageType and via the custom wire_enum
                # this is not technically a problem but we want to be nice and avoid using different ids for logically same things
                raise RuntimeError(f"Mismatch between MessageType and {message_type_enum}")
           message_type = custom_message_type
```

common/protob/messages-thp.proto
mmilata: 4b56e83

common/protob/messages-thp.proto
mmilata: a1cc761

common/protob/pb2py
mmilata: 6062055

mmilata:
With regards to THP-TWP separation, would it help if we extended `type_for_wire(wire_id)` with second argument (namespace/context/app/...) that would currently be either `MessageType` or `ThpMessageType`? With disjoint wire IDs we'd only check that the ID is in the enum. Lookup in separate tables would be possible with some pb2py changes.

matejcik:
> With regards to THP-TWP separation, would it help if we extended `type_for_wire(wire_id)` with second argument (namespace/context/app/...) that would currently be either `MessageType` or `ThpMessageType`? With disjoint wire IDs we'd only check that the ID is in the enum. Lookup in separate tables would be possible with some pb2py changes.

yes, that's kind of what I had in mind for future extensions.
for now i would just do the stupid thing with manual lookup in the enum, there's quite enough left to be done here without messing with (presumably) the binary format of the definitions

mmilata:
~Implemented the manual lookup in 5aee681. Had to extend the protobuf blobs because it wasn't possible to look up enums by name - 8223ab7, b928433. Please take a look.~

mmilata:
Force-pushed different version that allows you to tag any protobuf enum with `option (wire_enum) = true`. Wire id lookup table was extended to include 2-byte enum name Qstr which must also be passed to `type_for_wire` lookup function. I think the logic is less convoluted and now only one lookup is needed per message. PTAL

common/protob/messages-thp.proto
```diff
@@ -10,6 +10,40 @@ import "options.proto";
...
 
+/**
+ * Mapping between Trezor wire identifier (uint) and a Thp protobuf message
+ */
+enum ThpMessageType {
+    option (wire_enum) = true;
+
+    reserved 0 to 999;  // Values reserved by other messages, see messages.proto
+
+    ThpMessageType_ThpCreateNewSession = 1000;
```
M1nd3r: @mmilata Please, **move** `ThpCreateNewSession` into the `MessageType` enum. It was discussed before that this message should be part of the _trezor wire protocol_.

common/protob/messages-thp.proto
```diff
@@ -10,6 +10,40 @@ import "options.proto";
...
+    reserved 1001 to 1005;  // never appeared in a release, reserved for future use
+    ThpMessageType_ThpPairingRequest = 1006;
+    ThpMessageType_ThpPairingRequestApproved = 1007;
+    ThpMessageType_ThpSelectMethod = 1008;
+    ThpMessageType_ThpPairingPreparationsFinished = 1009;
+    ThpMessageType_ThpCredentialRequest = 1010;
+    ThpMessageType_ThpCredentialResponse = 1011;
```
M1nd3r, line 32: Please, **duplicate** the `ThpCredentialRequest` and  `ThpCredentialResponse` into the `MessageType` enum. Those messages can be used either during ThpPairingFlow or later using _trezor wire protocol_.

M1nd3r:
@mmilata  We could also **duplicate** some messages from `MessageType` to `ThpMessageType`: `Cancel` and `ButtonRequest`.

Notes: In http://github.com/trezor/trezor-firmware/pull/4976 in `pairing.py` https://github.com/trezor/trezor-firmware/pull/4976/files#diff-33acef33180f4e50401e821880709995f611242f99b211dfa10b32324130f528, there are messages that will be (in the future) checked against the `ThpMessageType` enum. -> I suspect that changes might be needed in the `context.py` functions like 
```
async def maybe_call(
    msg: protobuf.MessageType, expected_type: type[LoadedMessageType]
) -> None:
```
as they currently accept only the `MessageType`. The function `maybe_call` is used in `interact` to send `ButtonRequest`. There are also other options, eg. passing the actual context to the `interact`.

Details can and will be discussed later. Nevertheless, the messages `Cancel` and `ButtonRequest` can be part of the `ThpMessageType`.

common/protob/messages-thp.proto
mmilata: fixed in 71d0ffe1d8686654ea886ef6a84001483fcefa86

common/protob/messages-thp.proto
mmilata: fixed in 71d0ffe1d8686654ea886ef6a84001483fcefa86

mmilata:
@M1nd3r added `Cancel` and `ButtonRequest` (no ButtonAck?) in 71d0ffe1d8686654ea886ef6a84001483fcefa86

I think `protobuf.MessageType` is unfortunately named but different thing than `trezor.enums.MessageType` so no changes should be needed as long as same messages have the same wire id in all enums?

common/protob/messages-thp.proto
mmilata: FWIW 9c68e8e1eab2d3c4483f5f66a34cf85036910649 basically extends the message lookup table with 2 bytes containing build-time generated ID (Qstr index of the enum name, modtrezorproto currently doesn't expose enum objects to python)

common/protob/messages-thp.proto
```diff
@@ -44,10 +82,10 @@ message ThpHandshakeCompletionReqNoisePayload {
  * @start
  * @next Success
  */
-message ThpCreateNewSession{
+message ThpCreateNewSession {
     optional string passphrase = 1;
-    optional bool on_device = 2;       // User wants to enter passphrase on the device
-    optional bool derive_cardano = 3;  // If True, Cardano keys will be derived. Ignored with BTC-only
+    optional bool on_device = 2;                      // User wants to enter passphrase on the device
```
matejcik: please either stick a `default=false` in here or remind me why we don't want that. cc @M1nd3r 

(i vaguely recall something like, `if on_device is None and passphrase is None: seedless()` but that can apply to `not on_device` just as well?)

matejcik (approved):
LGTM, apart from that one nit about `on_device` default value (but we can change that at any later time too if necessary)

common/protob/messages-thp.proto
mmilata: added in e81747e7e8fe09546dac9d6e741204d83901bad9 
i thought we want to encourage entering passphrase on device?

M1nd3r: IMO I do not see the default as an important issue, `default=false` is ok by me.

M1nd3r:
> @M1nd3r added `Cancel` and `ButtonRequest` (no ButtonAck?) in [71d0ffe](https://github.com/trezor/trezor-firmware/commit/71d0ffe1d8686654ea886ef6a84001483fcefa86)
> 
> I think `protobuf.MessageType` is unfortunately named but different thing than `trezor.enums.MessageType` so no changes should be needed as long as same messages have the same wire id in all enums?

1) You are right about `ButtonAck`, we should probably add that too. But it does not need to happen in this PR. (We will learn about all missing messages when the separation of enums is implemented in the `thp` branch.)
2) I think you are right - as long as `ThpMessages` are part of `protobuf.MessageType`, and we will pass the correct enum into the `type_for_wire` in the right context, it should work just fine.

common/protob/messages-thp.proto
matejcik: > i thought we want to encourage entering passphrase on device?

uhh, something like, this is not the place to do that?
on the protobuf level we want the caller to be explicit about what they want -- an empty message will give you a seedless session, which either (a) works for you or (b) forces you to choose an explicit behavior of the passphrase.

mmilata:
cc @szymonlesisz - we modified `messages.proto` and `messages-thp.proto` a bit, not sure if it affects you. On binary level this should stay compatible with the previous definitions.

PR #5225, opened 2025-06-18
Refactor pairing code screen(s)

ibz:
There are two different screens showing the pairing code:

1. BLE pairing code screen
  * listens to BLE events, therefore depends on the `"ble"` feature
  * has a continue / cancel action bar
  * very similar to the `screen_confirm_pairing` in the bootloader, but it is re-implemented in FW to use the larger fonts (unabailable in BL)
2. THP pairing code screen
  * doesn't care about BLE - essentially just a static screen that can be dismissed by the user
  * uses a menu to let the user dismiss the screen, which means it has to be implemented as a "flow"
  * Note: due to it being a "flow" this would not work in the bootloader, but we don't have THP in the bootloader anyway!

### DIFF
--- a/core/embed/rust/src/ui/component/mod.rs
+++ b/core/embed/rust/src/ui/component/mod.rs
@@ -35,7 +35,7 @@ pub mod timeout;
 pub use bar::Bar;
 pub use base::{Child, Component, ComponentExt, Event, EventCtx, FlowMsg, Never, Timer};
 #[cfg(feature = "ble")]
-pub use ble::BLEHandler;
+pub use ble::{BLEHandler, BLEHandlerMsg};
 pub use border::Border;
 pub use button_request::{ButtonRequestExt, SendButtonRequest};
 #[cfg(all(
--- a/core/embed/rust/librust_qstr.h
+++ b/core/embed/rust/librust_qstr.h
@@ -705,6 +705,7 @@ static void _librust_qstrs(void) {
   MP_QSTR_share_words__words_in_order;
   MP_QSTR_share_words__wrote_down_all;
   MP_QSTR_show_address_details;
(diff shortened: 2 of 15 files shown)


core/embed/rust/src/ui/component/mod.rs
```diff
@@ -36,6 +36,8 @@ pub use bar::Bar;
 pub use base::{Child, Component, ComponentExt, Event, EventCtx, FlowMsg, Never, Timer};
 #[cfg(feature = "ble")]
 pub use ble::BLEHandler;
+#[cfg(feature = "ble")]
+pub use ble::BLEHandlerMsg;
```
mmilata: nit
```suggestion
#[cfg(feature = "ble")]
pub use ble::{BLEHandler, BLEHandlerMsg};
```

mmilata (approved):
LGTM, seems to be working, thanks!

mmilata:
after merging, rebased THP will need this
```diff
diff --git a/core/src/trezor/wire/thp/pairing_context.py b/core/src/trezor/wire/thp/pairing_context.py
index f0810fd6fa..5d586fc241 100644
--- a/core/src/trezor/wire/thp/pairing_context.py
+++ b/core/src/trezor/wire/thp/pairing_context.py
@@ -254,11 +254,10 @@ class PairingContext(Context):
         from trezor.ui.layouts.common import interact
 
         return await interact(
-            trezorui_api.show_pairing_code(
+            trezorui_api.show_thp_pairing_code(
                 title="One more step",
                 description=f"Enter this one-time security code on {self.host_name}",
                 code=self._get_code_code_entry_str(),
-                button=False,
             ),
             br_name=None,
         )
```

PR #5226, opened 2025-06-18
mvoe external tamper enable to secmon

TychoVrahe:
To prevent bricking the device entirely when tamper is active for some reason, the enable is moved to firmware. Also only alowed in PRODUCTION, as devkits don't have properly grounded tamper line.


kopecdav:
Tested, ACK

PR #5227, opened 2025-06-18
call bootloader update from coreapp

### LINKED ISSUE #4168: Restart device after upgrading the bootloader
Right now, trezor-core firmwares do not restart after upgrading the bootloader.
This might be a problem in some edge cases: the firmware assumes that it's running under the "right" bootloader, but if the bootloader was just upgraded, this assumption might not be true.

Trezor One halts after the bootloader is upgraded, asking the user to reconnect.

We should probably bring the same behavior to trezor-core. To make it user-friendly, the device can restart itself instead of asking for a reconnect ... unless the software induced restart is not "full" in some sense and could cause problems for the bootloader coming up?

TychoVrahe:
This PR changes the way bootloader update is invoked. 

Instead of embedding bootloader into secmon/kernel, its embedded in firmware(coreapp), and bootloader update is invoked via syscall + smcall. To keep the security same, additional checks are added to update smcall ( signature and hash check). 

New command is added to prodtest, which allows updating the bootloader via VPC. A simple script is added to upload the bootloader data by chunks.

This PR also fixes an issue with double-embedded bootloader when both secmon + kernel is used.

Hash processor is now made secure-only, as it is used for hashing the firmware image header in secmon. 

fixes #4168

### DIFF
--- a/core/embed/projects/firmware/main.c
+++ b/core/embed/projects/firmware/main.c
@@ -43,6 +43,8 @@
 
 #include <blake2s.h>
 
+#include "sys/bootutils.h"
+
 #ifdef USE_SECP256K1_ZKP
 #include "zkp_context.h"
 #endif
@@ -122,6 +124,12 @@ int main_func(uint32_t cmd, void *arg) {
     }
 #endif
   }
+
+#if PRODUCTION || BOOTLOADER_QA
+  if (bl_update_required) {
+    reboot_device();
+  }
+#endif
 
   screen_boot_stage_2(fading);
 


core/embed/rust/src/ui/layout_bolt/mod.rs
```diff
@@ -80,14 +80,7 @@ impl CommonUI for UIBolt {
...
 
-    fn screen_update() {
-        let mut frame = ErrorScreen::new(
-            "Update".into(),
-            "Finishing firmware update".into(),
-            "Do not turn off your Trezor".into(),
-        );
-        show(&mut frame, true);
-    }
+    fn screen_update() {}
```
matejcik, line 83: making these functions empty looks wrong at first glance
should we drop the whole function, if it is unused (as it seems, from the fact that Caesar / Delicia don't even have it)?

core/embed/util/bl_check/bl_check.c
```diff
@@ -78,10 +77,8 @@ static void uzlib_prepare(struct uzlib_uncomp *decomp, uint8_t *window,
   decomp->dest_limit = decomp->dest + destsize;
   uzlib_uncompress_init(decomp, window, window ? UZLIB_WINDOW_SIZE : 0);
 }
-#endif
 
-void check_and_replace_bootloader(void) {
-#if PRODUCTION || BOOTLOADER_QA
+bool bl_check_check(void) {
```
matejcik: the hash value(s) to check against should come from the caller
because it's the caller who has the replacement data

matejcik (commented):
concept ACK
i would like to change the `bl_check_check` signature, as commented

separately, why do we have to make hash processor secure? that seems generally impractical. can't secmon use an insecure peripheral?

core/embed/rust/src/ui/layout_bolt/mod.rs
TychoVrahe: All UIs have it and it is called (when bootloader is updating). We might want to implement it with some nice message, but since the update us fast i didnt want to waste my time with thay right now. On Bolt and Delizia this will cause quick fadeout-fadein when updating the bootloader,  no effect on Ceasar. Cc also @Hannsek if this is acceptable.

core/embed/util/bl_check/bl_check.c
TychoVrahe: changed https://github.com/trezor/trezor-firmware/commit/aa979aa54a115ec6c1d4aa9a8b219955deaeb877

TychoVrahe:
@matejcik sharing the peripheral is tricky, as we'd need to synchronize usage between secure/non-secure world, as each has its own globals which are used for this purpose. While probably possible, the usage outside of secure world is now quite limited that i don't think its worth the effort. Alternatively we could use it through smcalls, but again, i don't think its worth the effort now.

TychoVrahe:
added two more commits 
- RSOD on attempted bootloader downgrade
- rebooting device after bootloader update

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
```diff
@@ -51,6 +54,88 @@ static void prodtest_bootloader_version(cli_t *cli) {
...
+    if (bootloader_len + chunk_len > BOOTLOADER_MAXSIZE) {
+      cli_error(cli, CLI_ERROR, "BOOT: buffer overflow (have %u, %u more)",
+                (unsigned)bootloader_len, (unsigned)chunk_len);
+      return;
+    }
+
+    memcpy(&bootloader_buffer[bootloader_len], chunk_buf, chunk_len);
+    bootloader_len += chunk_len;
+
+    cli_ok(cli, "BOOT: got %u bytes (total %u)", (unsigned)chunk_len,
```
cepetr: The arguments to `cli_ok` should be values separated by spaces so that host scripts can parse them easily. For detailed logging, use `cli_trace` instead. In this situation, I would write:
```c
    cli_trace(cli, "Got %u bytes (total %u)", (unsigned)chunk_len, (unsigned)bootloader_len);
    cli_ok(cli, "%u %u", (unsigned)chunk_len, (unsigned)bootloader_len));
```
The same applies for other cli_ok occurences in this routine.

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
```diff
@@ -51,6 +54,88 @@ static void prodtest_bootloader_version(cli_t *cli) {
...
+
+  if (0 == strcmp(phase, "begin")) {
+    if (cli_arg_count(cli) != 1) {
+      cli_error_arg_count(cli);
+      return;
+    }
+
+    // Reset our state
+    bootloader_len = 0;
+    cli_ok(cli, "BOOT: begin");
```
cepetr: We should just use `cli_ok(cli, "")` and log more information with `cli_trace()`. I would also omit the `BOOT:`prefix, as the context makes it clear. Perhaps we could log something like "Bootloader update initiated."

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
```diff
@@ -51,6 +54,88 @@ static void prodtest_bootloader_version(cli_t *cli) {
...
+      cli_error_arg_count(cli);
+      return;
+    }
+
+    if (bootloader_len == 0) {
+      cli_error(cli, CLI_ERROR, "BOOT: no data received");
+      return;
+    }
+
+    parse_boardloader_capabilities();
```
cepetr: We might call `parse_boardloader_capabilities()` from `drivers_init()`, as we do in kernel/secmon. There's also another occurrence of this in prodtest_boardloader.c.

cepetr (changes requested):

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/f9696ac7239a7abd1e03f4e374558a760d0dd219

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/f9696ac7239a7abd1e03f4e374558a760d0dd219

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/f9696ac7239a7abd1e03f4e374558a760d0dd219

core/embed/projects/firmware/main.c
```diff
@@ -62,11 +84,50 @@ int main_func(uint32_t cmd, void *arg) {
...
+  const uint8_t *data = (const uint8_t *)&_deflated_bootloader_start;
+  const size_t len = (size_t)&_deflated_bootloader_size;
+
+  uint8_t hash_00[] = BOOTLOADER_00;
+  uint8_t hash_FF[] = BOOTLOADER_FF;
+
+  // Check if the boardloader is valid and replace it if not
+  bool bl_update_required =
+      bl_check_check(hash_00, hash_FF, BLAKE2S_DIGEST_LENGTH);
+  update_required |= bl_update_required;
```
cepetr: Nit: `|=` is not an appropriate operator for bool. We should is `update_required = update_required || ...`. The same applies for the assignment a few lines below.

cepetr (approved):

cepetr (approved):
I just found a nit regarding assignments in main.c, but otherwise everything seems to be working perfectly.

core/embed/projects/firmware/main.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/c4dd516d9f13c772fc102972188f52a80928ca74

Issue #957, opened 2020-04-22
Encrypt sessions when device is soft-locked


matejcik:
Soft-locked devices lock the storage, so that PIN is required to access private data.

However, derived seeds with passphrase applied are kept in RAM, so the soft-lock doesn't offer meaningful protection.

Secrets should also be encrypted in RAM when the device is locked.

tsusanka:
We should also revisit that cache.py is the only place where the caching occurs (not in trezor-crypto etc.). For example that we do not cache nodes anywhere else.

fresheneesz:
I came here to say this. Wouldn't this solve the [infamous trezor vulnerability](https://www.ledger.com/improving-the-ecosystem-disclosure-of-the-trezor-recovery-phrase-extraction-vulnerability)?

PR #5232, opened 2025-06-19
feat(core): encrypt the cached session seeds on soft-lock

M-pasta:
Regarding the issue #957:

Encrypting the derived seeds remaining in cache during soft-lock. 

ChaCha20-Poly1305 with a key derived from the DEVICE_SECRET is used for encryption and decryption. Nonce is generated randomly. The tag and nonce needed for decryption are stored as attributes of the DataCache object and deleted after decryption.


core/src/storage/seed_encryption.py
andrewkozlik, line 1: I'd call this file `cache_encryption.py`, because:
- We encrypt the seed in the flash storage, so this naming creates confusion.
- The cache may contain other secret data in the future, which we may want to encrypt.

core/src/apps/base.py
```diff
@@ -1,6 +1,7 @@
 from typing import TYPE_CHECKING
 
 import storage.device as storage_device
+from storage import cache
```
andrewkozlik: I am not sure whether it makes any difference, but I'd consider importing `encrypt_seeds()` and `decrypt_seeds()` locally in the functions where they are used, rather than importing the whole `cache` module globally.

core/src/storage/cache.py
```diff
@@ -44,6 +45,30 @@ def get_sessionless_cache() -> SessionlessCache:
     return _SESSIONLESS_CACHE
 
 
+def encrypt_seeds() -> None:
+    """
+    Encrypts seeds in all the cached sessions and the sessionless cache.
+    """
+    from storage.seed_encryption import encrypt_session_seeds
+
+    log.debug("encrypt_seeds", "Encrypting session seeds")
```
andrewkozlik: We usually follow this logging convention:
```py
if __debug__:
    log.debug(__name__, ...
```

core/src/storage/cache.py
```diff
@@ -44,6 +45,30 @@ def get_sessionless_cache() -> SessionlessCache:
...
+    for session in cache_codec._SESSIONS:
+        encrypt_session_seeds(session)
+    encrypt_session_seeds(get_sessionless_cache())
```
andrewkozlik: Use `_PROTOCOL_CACHE` and `_SESSIONLESS_CACHE`. Same below.

core/src/storage/cache_common.py
```diff
@@ -144,6 +144,9 @@ def get_int(
     ) -> int | T | None:  # noqa: F811
         return super().get_int(key & ~SESSIONLESS_FLAG, default)
 
+    def _get_length(self, key: int) -> int:
+        return super()._get_length(key & ~SESSIONLESS_FLAG)
+
```
andrewkozlik: I am curious why was this changed. Does it have influence on anything in this particular PR? It actually seem like a pretty significant bug fix, but not directly related to this PR.

core/src/storage/seed_encryption.py
andrewkozlik: Same goes for the function names here, such as `get_seed_encryption_key()` -> `get_cache_encryption_key()` and `b"Seed encryption key"` -> `b"Cache encryption key"`.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+
+from storage import cache_codec, cache_common
+from trezor import utils
+
+FIELDS_TO_ENCRYPT = [cache_common.APP_COMMON_SEED]
+if not utils.BITCOIN_ONLY:
+    FIELDS_TO_ENCRYPT += [
+        cache_common.APP_CARDANO_ICARUS_SECRET,
+        cache_common.APP_CARDANO_ICARUS_TREZOR_SECRET,
+    ]
```
andrewkozlik: I think it would be better if this was more tightly coupled with the definitions of the individual `APP_COMMON_*` elements. For example one of the following:
- Introduce `ENCRYPTED_FLAG = const(256)` and have `APP_COMMON_SEED = const(0 | ENCRYPTED_FLAG)`, etc.
- In `self.fields` specify not just the reserved length, but also a `bool` value indicating whether the entry should be encrypted when soft-locking.

My rationale is to better localize the definition of the properties of each cache item. When adding new items in the future it makes it easier for the developer to realize that they should decide on whether the cache item should or shouldn't be encrypted.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+    device_secret = get_device_secret()
+    label = b"Seed encryption key"
+    path: Slip21Path = [label, session_id]
+    key_node = Slip21Node(device_secret)
+    key_node.derive_path(path)
+    return key_node.key()
+
+
+def chain_seed_values(
+    session: cache_codec.DataCache, fileds_to_encrypt: List[int]
```
andrewkozlik: A lot of the functions in this file have `session: cache_codec.DataCache` as a parameter. Wouldn't it make more sense to make them members of `DataCache`?

andrewkozlik: Also most of these functions are not used outside of this file, so I'd prefix their names with `_`, e.g. `_chain_seed_values()`, or rather something like `_chain_encrypted_entries()`.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+def encrypt_session_seeds(session: cache_codec.DataCache) -> None:
+    """
+    Encrypts the seeds and other relevant data in the DataCache.
+    This function retrieves the session ID, derives the encryption key,
+    and encrypts the seed values using ChaCha20-Poly1305.
+    The encrypted data is then stored back in the session variables.
+    It sets the nonce and tag for decryption later as attributes of the session.
+    If there is nothing to encrypt, it does nothing.
+    """
+    from trezorcrypto import chacha20poly1305
```
andrewkozlik: Not sure whether it makes any difference, but I think we'd usually import from `trezor.crypto`.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+
+    from trezor.crypto import random
+
+    session_id, fields_to_encrypt = get_common_data(session)
+    if is_empty(session, fields_to_encrypt):
+        return
+
+    encryption_key = get_seed_encryption_key(session_id)
+    nonce = random.bytes(12, True)
```
andrewkozlik: I don't think we need to bother with a strong RNG here, that's mainly for really important secrets like generating the seed itself.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+        delattr(session, "encryption_nonce")
+        delattr(session, "encryption_tag")
+        gc.collect()  # Force garbage collection to free memory
+    except AttributeError:
+        raise ValueError("No encryption data found for the session")
+
+
+def set_decryption_data(
+    session: cache_codec.DataCache, nonce: bytes, tag: bytes
+) -> None:
```
andrewkozlik: This function and the previous one strongly indicate that we should make the functions here members of `DataCache`.

core/src/storage/seed_encryption.py
andrewkozlik: typo in `fileds_to_encrypt`

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+    if is_empty(session, fields_to_decrypt):
+        return
+
+    nonce, tag = get_decryption_data(session)
+    decryption_key = get_seed_encryption_key(session_id)
+    cipher = chacha20poly1305(decryption_key, nonce)
+    plaintext = cipher.decrypt(chain_seed_values(session, fields_to_decrypt))
+    control_tag = cipher.finish()
+    if control_tag != tag:
+        raise ValueError("Decryption failed: tag mismatch")
```
andrewkozlik: How do we actually deal with this exception? ISTM, that this would leave the cache somehow in an inconsistent state - some entries decrypted while other encrypted. Not properly locked or unlocked. One option is to try to recover from the error by deleting the session, which I don't think makes sense, so rather use `utils.ensure()` to RSOD.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+    for field in fileds_to_encrypt:
+        value = session.get(field)
+        if value:
+            plaintext += value
+    return bytes(plaintext)
+
+
+def parse_value_chain_to_cache(
+    ciphertext: bytes, session: cache_codec.DataCache, fields_to_encrypt: List[int]
+) -> None:
```
andrewkozlik: It seems to me that using `chain_seed_values()` and `parse_value_chain_to_cache()` is unnecessarily complicated. Both in terms code and memory allocations. Rather than building a large bytearray and then cutting it back up, I think you can simply call `encrypt()` and `decrypt()` on each value in succession and it works the same.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+
+    :return: the session ID and the list of fields to encrypt/decrypt
+    based on the DataCache type.
+    """
+
+    if isinstance(session, cache_codec.SessionCache):
+        session_id = session.export_session_id()
+        fields_to_encrypt = FIELDS_TO_ENCRYPT
+    elif isinstance(session, cache_common.SessionlessCache):
+        session_id = b"\x00" * cache_codec.SESSION_ID_LENGTH
```
andrewkozlik: Why not override `export_session_id()` in `SessionlessCache` to return an empty string?

core/src/storage/seed_encryption.py
andrewkozlik: The motivation for these functions is to deallocate the memory used by `encryption_nonce` and `encryption_tag` when the variables are not needed (storage is unlocked), which we discussed IRL. However, now that I look at it, I think this is not necessary, we can just have them as regular member variables and once we are done with them
```py
self.encryption_nonce = None
self.encryption_tag = None
gc.collect()
```
should simply do the trick.

andrewkozlik: BTW, I think `encryption_nonce` could be global across sessions, because each session has its own key derived, so we wouldn't actually be committing nonce reuse if we did that. Nevertheless this is more foolproof.

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+def get_seed_encryption_key(session_id: bytes) -> bytes:
+    """
+    Returns the seed encryption key for a given session ID.
+    The key is derived from the device secret and the session ID.
+    """
+    from typing import TYPE_CHECKING
+
+    from storage.device import get_device_secret
+
+    from apps.common.seed import Slip21Node
```
matejcik: we absolutely can't import from the `apps` namespace in `storage`

if we need the slip21 node impl here, I suggest moving it into `trezor.crypto`; that will likely involve a bigger refactor though, we may not want to just dump `apps.common.seed` in there as a whole.
perhaps just take out the implementation of the derivation process as a simple function?

core/src/storage/seed_encryption.py
matejcik: actually in some cases we want to import from `trezorcrypto` directly to avoid loading the `trezor.crypto` python module (`trezorcrypto` is a C module and "loading" it is essentially a free action)

not sure if it's relevant here -- in any case, both `chachapoly` and `random` should be imported from the same place

core/.changelog.d/957.changed
```diff
@@ -0,0 +1 @@
+Changed device secret to be genereated even without THP mode.
```
matejcik, line 1: changelog entries are supposed to be end-user facing; we don't need this one

core/src/storage/seed_encryption.py
```diff
@@ -0,0 +1,217 @@
...
+    :param session: The session containing the data to encrypt.
+    :param fileds_to_encrypt: The list of fields to include in the plaintext.
+    :return: A bytes object containing the concatenated values of the fields.
+    """
+    plaintext = bytearray()
+    for field in fileds_to_encrypt:
+        value = session.get(field)
+        if value:
+            plaintext += value
+    return bytes(plaintext)
```
matejcik: this is wasteful in terms of allocations
you should presize the bytearray via `trezor.utils.empty_bytearray()` and then return it without converting to `bytes`

zktex:
@M-pasta Can you please determine if this PR can be closed? You can open a new issue for the backlog if follow-up work is needed.

M-pasta:
This change is covered in [PR 6585](https://github.com/trezor/trezor-firmware/pull/6585)

PR #5234, opened 2025-06-20
Eckhart: remove boot stage 1 screen

TychoVrahe:
Removed boot stage 1 to optimize UX when booting.  

there is now larger delay between first screen is shown, we can probably optimize a bit more but it will be larger than 1 s anyway i guess. currently at about 1.4 s, which seems bordeline to me. Product confirmation needed.

Some backlight fading was removed to optimize the boot a bit.


obrusvit (approved):
Code-wise LGTM.

PR #5235, opened 2025-06-20
Enhanced suspend&wake-up logic

cepetr:
This PR improves the suspend/wake-up logic and API. There are five main improvements:

1. The device now wakes up correctly when powered on after being powered off. A new `WAKEUP_FLAG_POWER` flag indicates this condition.

2. Applications can now distinguish the wake-up reason returned by `pm_suspend()`.

3. The RTC wake-up timer can now invoke a callback instead of waking up the system. This callback can be used by the power manager to update the fuel gauge state while peripherals like the display and touch interface remain off.

4. The npm1300 and stwlc38 drivers can now complete pending operations before entering suspend mode.

5. Suspend logic has been removed from the power manager and moved to a new, independent module.

---

This PR also fixes an issue where the device would not wake up from the RTC wake-up timer when not connected to a debugger.

And finally stwlc38 i/o definitions have been moved to the board definition header.


core/embed/sys/power_manager/inc/sys/power_manager.h
```diff
@@ -154,9 +146,11 @@ pm_status_t pm_get_state(pm_state_t* state);
 
 /**
  * @brief Request device to enter suspend mode
+ * @param wakeup_reason Pointer to store wakeup flags (reason for wakeup).
+ * Can be NULL if not needed.
  * @return pm_status_t Status code indicating success or failure
  */
-pm_status_t pm_suspend(void);
+pm_status_t pm_suspend(wakeup_flags_t* wakeup_reason);
```
TychoVrahe, line 153: lets already add a parameter for allowing/disabling hibernation for the coinjoin, so that whoever needs to build on this has the final api

core/embed/sys/power_manager/npm1300/npm1300.c
```diff
@@ -94,10 +107,21 @@ typedef struct {
   // Current state of the FSM
   npm1300_fsm_state_t state;
 
+  // Set if the driver was requested to suspend background operations.
```
TychoVrahe, line 110: i would already go for introducing some driver_state enum with RESET, SUSPENDING, SUSPENDED, ACTIVE states. Many of the if (drv->initialized) checks in functions below would benefit of checking state==ACTIVE, as they don't make much sense to be called when suspended/suspending

core/embed/sys/power_manager/inc/sys/pmic.h
```diff
@@ -59,6 +59,19 @@ bool pmic_init(void);
...
+// Suspending may take some time if the driver is currently
+// performing an operation. Caller may check the status by
+// pmic_is_suspended().
+bool pmic_suspend(void);
+
+// Resumes the driver operation after it has been suspended.
+bool pmic_resume(void);
+
+// Checks whether the driver is suspended.
+bool pmic_is_suspended(void);
```
TychoVrahe, line 73: .. and then this would be get driver state or something

core/embed/sys/power_manager/stm32u5/power_manager.c
```diff
@@ -207,9 +209,31 @@ pm_status_t pm_get_state(pm_state_t* state) {
...
+// when the rtc wake-up timer expires. The callback can perform
+// measurements and update the fuel gauge state.
+// - The callback can schedule the next wake-up by calling
+// rtc_wakeup_timer_start().
+// - If the callback return with wakeup_flags set, system_suspend() returns.
+#ifdef USE_RTC
+void pm_rtc_wakeup_callback(void* context) {
+  // TODO: update fuel gauge state
+
+  if (true) {
```
TychoVrahe, line 222: some logic missing in addition to updating the fuel gauge? lets at least add another todo

core/embed/sys/smcall/stm32/smcall_dispatch.c
```diff
@@ -100,7 +100,7 @@ __attribute((no_stack_protector)) void smcall_handler(uint32_t *args,
       reboot_with_rsod__verified(pminfo);
     } break;
 
-#ifdef USE_POWER_SAVE
+#ifdef USE_SUSPEND
     case SMCALL_POWER_SAVE_SUSPEND_CPU: {
```
TychoVrahe: i think we should get rid of all the power_save names

core/embed/sys/power_manager/stm32u5/power_manager.c
cepetr: Added comment, see c68a56b5a34106683f85d46894fc6c0886738d68

core/embed/sys/suspend/inc/sys/suspend.h
```diff
@@ -0,0 +1,58 @@
...
+ * @brief Puts device into suspend mode (actually STOP2 mode on STM32U5)
+ *
+ * @return Return flags indicating the reason for wakeup
+ */
+wakeup_flags_t system_suspend(void);
+
+/**
+ * @brief Set wakeup flags
+ * @param flags Wakeup flags to set
+ * @return pm_status_t Status code indicating success or failure
```
kopecdav: left pm_status returns in API comments

core/embed/sys/suspend/inc/sys/suspend.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+void wakeup_flags_set(wakeup_flags_t flags);
+
+/**
+ * @brief Reset wakeup flags
+ * @return pm_status_t Status code indicating success or failure
```
kopecdav: return pm_status_t

core/embed/sys/suspend/inc/sys/suspend.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+void wakeup_flags_reset(void);
+
+/**
+ * @brief Get wakeup flags
+ * @param flags Pointer to store the current wakeup flags
+ * @return pm_status_t Status code indicating success or failure
```
kopecdav: return pm_status_t

core/embed/sys/smcall/stm32/smcall_dispatch.c
cepetr: Yes, i tried. What about this: 32c0fe87494b24040101b990cab38888d2104643

core/embed/sys/power_manager/npm1300/npm1300.c
cepetr: Yes, I agree. However, as we discussed in person, I would add this state by implementing a new concept for driver initialization and suspension using a registration structure. I hope we can implement it soon. Until then I would not extend the current implementation  with this change.

core/embed/sys/power_manager/inc/sys/power_manager.h
cepetr: Once we have implemented the new concept of driver initialization and deinitialization, the suspend module will only call generic suspend/resume functions and won’t need to know the specific driver APIs. I think it would be better if the BLE/NRF module could decide on its own which mode to enter during suspension. Maybe we could discuss this in person.

core/embed/sys/power_manager/stm32u5/power_manager.c
kopecdav: With single RTC timer, we will have to deal with two timing tasks. 1. short wakeup intervals to drive the fuel gauge. 2. very long time interval to decide when to switch from suspend to hibernate if required. Do you want the power manager to be responsible for it here as well?

TBH, for case 2. I do not recall from the discussion if we wanted to do hibernation right away in kernel or if we want to just wake up and let coreapp to make it.

cepetr: Yes, currently the kernel should be responsible for entering hibernation without any involvment from the app.

core/embed/sys/suspend/inc/sys/suspend.h
cepetr: fixed, see 89258a4c13fde4e867d3ab13d593f1bd869bdbc9

core/embed/sys/suspend/inc/sys/suspend.h
cepetr: fixed, see 89258a4c13fde4e867d3ab13d593f1bd869bdbc9

core/embed/sys/suspend/inc/sys/suspend.h
cepetr: fixed, see 89258a4c13fde4e867d3ab13d593f1bd869bdbc9

core/embed/sys/power_manager/stm32u5/power_manager.c
TychoVrahe: as discussed, yes PM should be responsible, meaning there are more options:
- hibernate
- wakeup (currently not needed i think)
- measure & schedule next measurement
  - if the hibernation time elapsed, hibernate

core/embed/sys/power_manager/npm1300/npm1300.c
TychoVrahe: yeah we can do that later, no prob

core/embed/sys/power_manager/inc/sys/power_manager.h
TychoVrahe: As discussed: this is not a question of BLE/NRF driver, but what application wants, thus needs PM setting, but form of the setting is not entirely clear so lets handle this later.

TychoVrahe (approved):

Issue #5189, opened 2025-06-12
Introduce multi-item menus


Hannsek:
**Design:** [Figma](https://www.figma.com/design/mZCYSG8i6s3E5xuhhiFzpQ/FW-playground?node-id=11-496&t=KZsWHlOckY15SejR-4)

**Spec**:
Implement multi-item menus across all core models.

On TS5 and TT, the user will swipe up to navigate to the next page of the menu.

On TS3, the menu will function as a carousel (as currently implemented in other flows such as recovery).

Additionally:

Remove the horizontal swipe gesture for accessing or exiting the menu on TT and TS5.

romanz:
Let's use this issue for tracking the migration to "external menu" - https://github.com/trezor/trezor-firmware/pull/7020#discussion_r3325312256

PR #5237, opened 2025-06-23
feat(core): implement hierarchical information menus for Caesar

romanz:
Following #5189.

Existing menus are still using the old layouts - it will be done by subsequent PRs.

<details>

<summary> Tested using the following snippet </summary>

```python
    from trezor.ui.layouts.menu import show_menu, Menu, Details, Item

    menu = Menu.root(
        Details(
            "Account info",
            Item("Account", "ETH #2"),
            Item("Derivation path", "m/1/2/3/4"),
        ),
        Details(
            "Block hash",
            Item.data(
                None,
                "0x1355557e447b95b89c75a5d4909701d52faf2716eb30c55a6c9fcbb735748f9e",
            ),
        ),
        Details(
            "Fee info",
            Item("Gas limit", "21000 units"),
            Item("Max fee per gas", "3.28 Gwei"),
            Item("Max priority fee", "1.7 Gwei"),
        ),
    )
    await show_menu(menu)
```

</details>

[Screencast from 2025-06-23 12-51-48.webm](https://github.com/user-attachments/assets/36859f75-ef4a-451d-9a9a-5d03d5c9ec2a)


core/embed/rust/src/ui/api/firmware_micropython.rs
```diff
@@ -1783,6 +1814,15 @@ pub static mp_module_trezorui_api: Module = obj_module! {
     ///     time_ms timeout is passed."""
     Qstr::MP_QSTR_show_progress_coinjoin => obj_fn_kw!(0, new_show_progress_coinjoin).as_obj(),
 
+    /// def show_menu_items(
```
ibz: I know we used to call these things "items" (fee items, account items...)

But "menu_items" means something else. It sounds like this is showing a menu, when in fact it is not...

Would "properties" be a better name? Since they are essentially key-value pairs... So `show_properties` perhaps? No "menu", because although this is used from the menu, it doesn't have to be that way, right?

core/src/trezor/ui/layouts/menu.py
```diff
@@ -0,0 +1,62 @@
...
+        return cls("", *children)
+
+
+class Details:
+    def __init__(self, name: str, *items: "Item") -> None:
+        self.name = name
+        self.items = items
+
+
+class Item:
```
ibz: I would call this `Property` - or something else.

core/embed/rust/src/ui/api/firmware_micropython.rs
romanz: Sounds good, changed to `properties` in 1ebf74f105.

core/src/trezor/ui/layouts/menu.py
romanz: Changed to `Property` in 1ebf74f105.

romanz:
CI seems to be stuck... re-pushing.

EDIT: will also resolve the conflicts.

ibz (approved):

romanz:
Squashed and rebased over `main` (resolving a minor conflict).

PR #5240, opened 2025-06-23
Fix fuel gauge soc limit during initial estimation

kopecdav:
This tiny PR fix an issue with out-of-bound SoC estimation during battery initialization.

The initial estimate was not bounded, so time to time it returned crazy estimates when battery was not connected and then showing SoC 255% in the core app.

This PR also add simple prodtest command `pm-new-soc-estimate` which is helpful when a user need to test battery initialization several times. Command disable the power manager, erase recovery data from backup ram an reboot the device to get a new estimate. 

Before, this was usually done by simply erasing the backup ram and manually rebooting the device, but after backup ram refactor this is no longer possible, since now it could recover from data loss.

### DIFF
--- a/core/embed/projects/prodtest/cmd/prodtest_power_manager.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_power_manager.c
@@ -27,6 +27,8 @@
 #include <rtl/mini_printf.h>
 #include <rtl/unit_test.h>
 #include <rust_ui_prodtest.h>
+#include <sys/backup_ram.h>
+#include <sys/bootutils.h>
 #include <sys/power_manager.h>
 #include <sys/rtc.h>
 #include <sys/systick.h>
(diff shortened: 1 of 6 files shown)


core/embed/projects/prodtest/cmd/prodtest_power_manager.c
```diff
@@ -381,6 +383,29 @@ void prodtest_pm_set_soc_limit(cli_t* cli) {
...
+
+  cli_trace(cli, "Erasing backup RAM and rebooting...");
+  cli_ok(cli, "");
+  systick_delay_ms(100);
+
+  // Deinitialize power manager so the monitor stop feeding the recovery data
+  // to backup RAM.
+  pm_deinit();
+
+  backup_ram_erase();
```
TychoVrahe: instead of erasing the whole backup ram, how about writing invalid PM data instead? i.e. version 0 or something. So we can keep the other data in backup ram, should we implement any.

kopecdav: I would erase only backup ram data, but right now api do not support to erase only portion of data selectively with key. Anyway, I would rather keep this way to be absolutely sure we start from scratch and both banks are cleared. 

If we decide to store more kinds of data in backup ram, I would add a selective erase.

TychoVrahe: I mean, you can write PM data with version 0, which will be found invalid at boot. Or maybe even add the missing erase key API.

Of course we could do it later, but chances are that we will not remember about this prodtest function when adding more data. Or maybe we won't sign a new prodtest after adding it - making this function broken on production devices until we do.

So i'd say lets do it now.

kopecdav: ok, how about this? https://github.com/trezor/trezor-firmware/pull/5240/commits/673c1e82fa00594387ea06fa8c258352708ffe3e (I'l let also @cepetr to review the backup ram)

TychoVrahe: yeah that looks good to me

TychoVrahe (approved):

core/embed/projects/prodtest/cmd/prodtest_power_manager.c
kopecdav: in https://github.com/trezor/trezor-firmware/pull/5240/commits/f3aad2d2d854cf32e8d1954ff317f5f5c60654dd, `backup_ram_erase_item` implementation simplified with `backup_ram_write` function according to @cepetr review.

PR #5242, opened 2025-06-23
Fix combine tool script

kopecdav:
Combine tool got broken after some changes in layout parser due to secure monitor introduction. 
This PR fixes it.


TychoVrahe (approved):
utACK

PR #5251, opened 2025-06-25
Eckhart: adjust wipe success screen in bootloader

TychoVrahe:
In eckhart UI, after wiping device from bootloader, the success screen now include a restart button, so that user doesn't have to wait until rebooot.

Same applies to unlock bootloader flow. 

Some minor adjustments to unlock bootloader screens were made too.

Also, integer overflow in wipe progress calculation is fixed.

Other models will hang in the success screen until unplugged.

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
@@ -286,9 +286,9 @@ impl BootloaderUI for UIEckhart {
     fn screen_unlock_bootloader_confirm() -> u32 {
         let msg1 =
             Label::left_aligned("Unlock bootloader".into(), TEXT_NORMAL).vertically_centered();
-        let msg2 = Label::centered("This action cannot be undone!".into(), TEXT_NORMAL);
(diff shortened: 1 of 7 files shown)


core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
```diff
@@ -369,12 +369,12 @@ impl BootloaderUI for UIEckhart {
             TEXT_NORMAL,
         ))
         .with_header(BldHeader::new("Done".into()).with_icon(theme::ICON_DONE, theme::GREY))
```
obrusvit: Can you use BldHeader::new_done() also here in `screen_wipe_success`

obrusvit (commented):
Thank you. Just one comment regarding the UI.

core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/be2a352221fa009f9a6caa8ccba2a9d043f0be9b

PR #5255, opened 2025-06-26
add timeout to nrf uart communication

TychoVrahe:
This PR introduces timeout detection for uart/dfu communication with nRF. This helps in situation where the nrf is not communicating - instead of hanging forever there will be more appropriate timeout failure.

### DIFF
--- a/core/embed/io/nrf/stm32u5/nrf_uart.c
+++ b/core/embed/io/nrf/stm32u5/nrf_uart.c
@@ -27,6 +27,7 @@
 
 #include "../nrf_internal.h"
 #include "rust_smp.h"
+#include "sys/systick.h"
 
 extern nrf_driver_t g_nrf_driver;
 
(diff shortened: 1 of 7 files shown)


core/embed/io/nrf/stm32u5/nrf_uart.c
```diff
@@ -147,7 +148,10 @@ void USART3_IRQHandler(void) {
   IRQ_LOG_EXIT();
 }
 
-void nrf_send_uart_data(const uint8_t *data, uint32_t len) {
+bool nrf_send_uart_data(const uint8_t *data, uint32_t len,
```
cepetr, line 151: You might consider the following modification. It adds a timeout to the first loop, which waits for the previous transmission to complete, and it uses variables while interrupts are disabled (this is not necessary, but it seems more correct).

```c
bool nrf_send_uart_data(const uint8_t *data, uint32_t len,
                        uint32_t timeout_ms) {
  nrf_driver_t *drv = &g_nrf_driver;

  if (!drv->initialized) {
    return false;
  }

  uint32_t deadlime = ticks_timeout(timeout_ms);
  bool result = false;

  irq_key_t key = irq_lock();

  while (drv->dfu_tx_pending && !ticks_expired(deadlime)) {
    // Wait for previous transmission to complete
    irq_unlock(key);
    key = irq_lock();
  }

  if (drv->dfu_tx_pending) {
    // If we are still pending, it means we timed out
    goto cleanup;
  }

  drv->dfu_tx_pending = true;

  HAL_UART_Transmit_IT(&drv->urt, data, len);

  while (drv->dfu_tx_pending && !ticks_expired(deadlime)) {
    // Wait for transmission to complete
    irq_unlock(key);
    irq_key_t key = irq_lock();
  }

  if (drv->dfu_tx_pending) {
    // If we are still pending, it means we timed out
    drv->dfu_tx_pending = false;
    HAL_UART_Abort_IT(&drv->urt);
    goto cleanup;
  }

  result = true;

cleanup:
  irq_unlock(key);
  return result;
}
```

TychoVrahe: yes i agree its better, see https://github.com/trezor/trezor-firmware/commit/31fc47c39bfa67032f8b00726cd9be9c0e3b4a6a along with some rust warning fixes.

cepetr (approved):
utACK

PR #5261, opened 2025-06-26
Cancel gradient for UI Eckhart

bieleluk:
Unify usage of action bar cancel buttons to have:
- orange gradient
- one button stylesheet

\+ **suppress `clippy` warnings**


romanz (approved):
LGTM

PR #5266, opened 2025-06-26
Derivation from master key in secret API

andrewkozlik:
@TychoVrahe please check that the changes to the slot layout make sense (mainly `T3W1/secret_layout.h`) as well as the code organization in general. I was thinking that it might be cleaner to put some of the code in `stm32u5/secret_keys.c` into `common/secret_keys.c`, but perhaps let's leave that for the future, since right now it would just be used in `stm32u5` anyway.

@onvej-sl please check the rest, mainly the cryptography in `stm32u5/secret_keys.c` and provisioning in `prodtest_optiga.c`. I think this way the provisioning is going to be more aligned with Tropic, but you will probably completely change this code anyway, so it's up to you how thoroughly you wish to review it.

Untested, since @onvej-sl is making changes to the same code in prodtest right now and it's going to be significantly changed anyway.


andrewkozlik:
<s>@TychoVrahe if the bootloader gets unlocked, then we call `flash_area_erase(&SECRET_AREA, NULL)`, which will erase `SECRET_TROPIC_TROPIC_PUBKEY_SLOT` won't it? I think we don't want that.</s> Similarly, I think we will want `secret_key_tropic_pairing_unprivileged()` to work after bootloader unlock, so I will probably need to create a second master key, which survives bootloader unlock.

Edit: I must have incorrectly looked at stm32f4 code. In stm32u5 the slots that are public (`SECRET_KEY_SLOT_*_PUBLIC`) don't get erased when the bootloader unlocks.

andrewkozlik:
In the fixups I added a second master key `SECRET_UNPRIVILEGED_MASTER_KEY_SLOT`, which does not get erased when the bootloader is unlocked

onvej-sl (approved):

TychoVrahe (approved):

PR #5267, opened 2025-06-26
update nrf firmware from prodtest

TychoVrahe:

### DIFF
--- a/core/embed/projects/prodtest/cmd/prodtest_nrf.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_nrf.c
@@ -69,6 +69,88 @@ static void prodtest_nrf_version(cli_t* cli) {
          info.version_patch, info.version_tweak);
 }
 
+#define NRF_UPDATE_MAXSIZE 0x50000
+__attribute__((section(".buf"))) static uint8_t nrf_buffer[NRF_UPDATE_MAXSIZE];
+static size_t nrf_len = 0;
+
+static void prodtest_nrf_update(cli_t* cli) {
+  if (cli_arg_count(cli) < 1) {
+    cli_error_arg_count(cli);
+    return;
+  }
+
+  const char* phase = cli_arg(cli, "phase");
+  if (phase == NULL) {
+    cli_error_arg(cli, "Expecting phase (begin|chunk|end).");
+    return;
+  }
+
+  if (0 == strcmp(phase, "begin")) {
+    if (cli_arg_count(cli) != 1) {
+      cli_error_arg_count(cli);
+      return;
+    }
+
+    // Reset our state
+    nrf_len = 0;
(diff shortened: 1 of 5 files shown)


core/embed/projects/prodtest/cmd/prodtest_nrf.c
```diff
@@ -69,6 +69,88 @@ static void prodtest_nrf_version(cli_t* cli) {
...
+    cli_trace(cli, "begin");
+    cli_ok(cli, "");
+
+  } else if (0 == strcmp(phase, "chunk")) {
+    if (cli_arg_count(cli) < 2) {
+      cli_error_arg_count(cli);
+      return;
+    }
+
+    // Receive next piece of the image
```
kopecdav, line 113: There is no check that the transition began, so you may just start sending chunks without `begin` command. Maybe better to create a `update_begin` flag and check it here?

core/embed/projects/prodtest/cmd/prodtest_nrf.c
```diff
@@ -69,6 +69,88 @@ static void prodtest_nrf_version(cli_t* cli) {
...
+    if (!cli_arg_hex(cli, "hex-data", chunk_buf, sizeof(chunk_buf),
+                     &chunk_len)) {
+      cli_error_arg(cli, "Expecting hex-data for chunk.");
+      return;
+    }
+
+    if (nrf_len + chunk_len > NRF_UPDATE_MAXSIZE) {
+      cli_error(cli, CLI_ERROR, "Buffer overflow (have %u, need %u)",
+                (unsigned)nrf_len, (unsigned)chunk_len);
+      return;
```
kopecdav: Also, I would then clear it on any error during the update to make sure that the transfer was consistent.

kopecdav (changes requested):
The code looks good and python scrip seems to collaborate well with the prodtest command, however the current version of the radio-test seems to be not compatible with the current version of nrf bootloader. So after flashing the radio test app into nrf, `ble-radio-test` has no response. We should fix that first or maybe merge the radio test into the main codebase, so we keep all components updated.

core/embed/projects/prodtest/cmd/prodtest_nrf.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/eafbfec489fd10a109f14521b73bed5cb873c039

core/embed/projects/prodtest/cmd/prodtest_nrf.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/eafbfec489fd10a109f14521b73bed5cb873c039

kopecdav (approved):

PR #5273, opened 2025-06-30
FIDO attestation key masking

andrewkozlik:


core/embed/projects/prodtest/cmd/prodtest_optiga.c
```diff
@@ -543,17 +544,35 @@ static void pubkey_read(cli_t* cli, uint16_t oid) {
...
 
-  cli_ok_hexdata(cli, public_key, public_key_size);
+#ifdef SECRET_KEY_MASKING
+  if (masking_key != NULL) {
+    uint8_t masked_public_key[ECDSA_PUBLIC_KEY_COMPRESSED_SIZE] = {0x02};
+    if (public_key_x_size != sizeof(public_key_x)) {
+      cli_error(cli, CLI_ERROR, "unexpected public key size");
+      return;
+    }
+    memcpy(&masked_public_key[1], public_key_x, sizeof(public_key_x));
```
andrewkozlik, line 570: Here we only have the x-coordinate of the public key, so by assigning 0x02 to the first byte we arbitrarily pick one of the two possible y-coordinates. Then later we multiply the point by the inverse of the masking scalar and read the x-coordinate of the result. This should work, but I am not sure about the justification. I think the justification is the mirror symmetry of the elliptic curve along the x-axis (which follows from the short Weierstrass form of the curve). Please check if this makes sense @onvej-sl

onvej-sl: Yes, it makes sense. To be precise, it follows from the fact that the negation of the point (x, y) is (x, -y).

Let k be a scalar, P be a point and -P be its negation, then k * -P = k * (-1) * P = (-1) * k * P = -(k * P), which means k * -P and k * P have the same x-coordinate.

Feel free to resolve this conversation.

EDITED: I swapped the x and y coordinates.

core/embed/sec/optiga/optiga.c
```diff
@@ -101,35 +104,79 @@ static const optiga_metadata_item ACCESS_PIN_HMAC_CTR =
...
-  optiga_result res =
-      optiga_calc_sign(OPTIGA_OID_ECC_KEY + index, digest, digest_size,
-                       &signature[2], max_sig_size - 2, sig_size);
+#ifdef SECRET_KEY_MASKING
+  uint8_t masking_key[ECDSA_PRIVATE_KEY_SIZE] = {0};
+  uint8_t masked_digest[SHA256_DIGEST_LENGTH] = {0};
+  bool is_masked = (index == OPTIGA_FIDO_ECC_KEY_INDEX);
+  if (is_masked) {
+    if (digest_size != SHA256_DIGEST_LENGTH ||
+        secret_key_optiga_masking(masking_key) != sectrue ||
```
onvej-sl, line 122: I dislike retrieving the masking key here in `optiga_sign()`. I would prefer if the masking key were a parameter of `optiga_sign()`, and there was a wrapper function called `optiga_sign_fido()`. This wrapper function would retrieve the masking key and call `optiga_sign()` with `index = OPTIGA_FIDO_ECC_KEY_INDEX`. What do you think?

core/embed/projects/prodtest/cmd/prodtest_optiga.c
```diff
@@ -543,17 +544,35 @@ static void pubkey_read(cli_t* cli, uint16_t oid) {
       0x4f, 0xe3, 0x42, 0xe2, 0xfe, 0x1a, 0x7f, 0x9b, 0x8e, 0xe7, 0xeb, 0x4a,
       0x7c, 0x0f, 0x9e, 0x16, 0x2b, 0xce, 0x33, 0x57, 0x6b, 0x31, 0x5e, 0xce,
       0xcb, 0xb6, 0x40, 0x68, 0x37, 0xbf, 0x51, 0xf5};
-  uint8_t public_key[32] = {0};
-  size_t public_key_size = 0;
+  uint8_t public_key_x[ECDSA_SCALAR_SIZE] = {0};
```
onvej-sl: To be pedantic, `public_key_x` is not a scalar, so there should be something like `ECDSA_COORDINATE_SIZE`. However, since both are 32 bytes, I'm not sure if it is worth it.

andrewkozlik: Yes, you are absolutely right. The two constants `ECDSA_SCALAR_SIZE` and `ECDSA_COORDINATE_SIZE` will always nearly coincide due to Hasse's theorem, I think at most off by one byte, and I am actually not aware of any curve for which they don't coincide. I can't think of a name which would nicely encompass both.

core/embed/sec/optiga/optiga.c
andrewkozlik: What exactly is it that you dislike about having the logic here?

If we move the masking key selection logic to wrapper functions, then we should make `optiga_sign()` static and also have a wrapper function `optiga_sign_device()`, and then it would also make sense to follow through with having two Python wrappers in modtrezorcrypto-optiga.h instead of just the one. I have nothing against that in terms of architecture, but it seems to me like a lot of overhead for something that can be succinctly summarized in a single line as
```c
bool is_masked = (index == OPTIGA_FIDO_ECC_KEY_INDEX);
```
If we had two or more keys in Optiga that use _different_ masking keys, then that would be a stronger reason to add the wrapper functions, but currently I see no plans to add another masked key, so I am feeling more inclined towards avoiding the overhead that this would create.

core/embed/projects/prodtest/cmd/prodtest_optiga.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/5273/commits/5fb77b8c78f387ea03c8bb9867830939d6b2ba93

core/embed/projects/prodtest/cmd/prodtest_optiga.c
andrewkozlik: added a code comment in https://github.com/trezor/trezor-firmware/pull/5273/commits/5fb77b8c78f387ea03c8bb9867830939d6b2ba93

PR #5275, opened 2025-06-30
autodim & autosuspend bootloader on inactivity

TychoVrahe:
This PR introduces auto dimming and auto suspending of bootloader on inactivity.

After suspending and resuming, the flow continues where it ended.

Timing is subject to further tuning.

### DIFF
--- a/core/embed/projects/bootloader/workflow/wf_host_control.c
+++ b/core/embed/projects/bootloader/workflow/wf_host_control.c
@@ -35,12 +35,29 @@
 #include <wire/wire_iface_ble.h>
 #endif
 
+#ifdef USE_POWER_MANAGER
+#include <io/display.h>
+#include <io/display_utils.h>
+#include <sys/power_manager.h>
+
+#define FADE_TIME_MS 10000
+#define SUSPEND_TIME_MS 15000
+
+#endif
+
 workflow_result_t workflow_host_control(const vendor_header *const vhdr,
                                         const image_header *const hdr,
                                         c_layout_t *wait_layout,
                                         uint32_t *ui_action_result,
                                         protob_ios_t *ios) {
   workflow_result_t result = WF_ERROR_FATAL;
+
+#ifdef USE_POWER_MANAGER
+  uint32_t fade_deadline = ticks_timeout(FADE_TIME_MS);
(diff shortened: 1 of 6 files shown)


core/embed/rust/src/ui/layout/simplified.rs
```diff
@@ -117,6 +142,31 @@ pub fn run(frame: &mut impl Component<Msg = impl ReturnToC>) -> u32 {
...
+            #[cfg(feature = "power_manager")]
+            {
+                if is_usb_connected() {
+                    continue;
+                }
+
+                let elapsed = Instant::now().checked_duration_since(start);
+
+                if let Some(elapsed) = elapsed {
+                    if elapsed.to_secs() >= 10 && !faded {
```
kopecdav, line 155: It seems like `10` and `15` constants are duplicit to FADE_TIME_MS and  SUSPEND_TIME_MS, can we link them together to have a single source of truth?

core/embed/projects/bootloader/workflow/wf_host_control.c
```diff
@@ -35,13 +35,30 @@
...
+
 workflow_result_t workflow_host_control(const vendor_header *const vhdr,
                                         const image_header *const hdr,
                                         c_layout_t *wait_layout,
                                         uint32_t *ui_action_result,
                                         protob_ios_t *ios) {
   workflow_result_t result = WF_ERROR_FATAL;
 
+#ifdef USE_POWER_MANAGER
+  uint32_t fade_deadline = ticks_timeout(10000);
```
kopecdav: Use FADE_TIME_MS/SUSPEND_TIME_MS macro instead of constants.

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/pull/5275/commits/47d646950d5818aeb5e0092d67cecfaaafc7d6db

core/embed/rust/src/ui/layout/simplified.rs
TychoVrahe: that would be ideal but i currently don't have a good way to share that between C and Rust, and i don't want to invent one right now

kopecdav (approved):

STew790:
**QA OK**
Dim and autosuspend works in BL.

PR #5276, opened 2025-06-30
Fuel gauge compensation in low power mode

kopecdav:
This PR introduces fuel gauge compensation in low power modes, as well as fuel gauge updates during charging in suspend mode. It builds upon the enhanced suspend logic implemented in PR https://github.com/trezor/trezor-firmware/pull/5235

The first two commits add an `rtc_get_timestamp()` function to the RTC library, which returns a timestamp representing the number of seconds since the device was powered on. Additionally, a prodtest command is introduced to retrieve and display the current timestamp in the console.

The power manager has been restructured to allow it to operate as a background task within the suspend module. As part of this, the functions `pm_driver_suspend()`, `pm_driver_is_suspended()`, and `pm_driver_resume()` were added. The power monitoring logic was also improved: instead of using the systimer to both wait for and process PMIC data, the systimer now only triggers PMIC data acquisition, and the PMIC callback handles data processing.

Finally, fuel gauge compensation was implemented to improve state-of-charge (SoC) estimation while the device is in low power mode and the typical fuel gauge estimation loop is not running.

Fuel gauge compensation has three scenarios:

1) Hibernation mode:
The RTC timestamp is now included in the power manager's recovery data stored in backup RAM. This allows the device to track how long it remained in hibernation. Upon initialization in the bootloader, the power manager compensates the state of charge (SoC) using the known hibernation current consumption and the elapsed time. Since the device temperature during hibernation is unknown, battery curves for 25 °C are used for this calculation.

2) Suspend mode (not charging):
When entering suspend mode, the RTC timestamp is recorded. Upon wake-up, the power manager compensates the SoC based on the known current consumption during suspend. Additionally, an auto-hibernation feature has been added: if there is no activity for 24 hours, the device will automatically enter hibernation to conserve power.

3) Suspend mode (charging):
In this mode, the device enters suspend but wakes up every 60 seconds to measure the charging current and update the SoC. If external power is disconnected, the device automatically transitions to the non-charging suspend scenario (2) without waking up.

### DIFF
--- a/core/embed/sys/power_manager/stm32u5/power_manager.c
+++ b/core/embed/sys/power_manager/stm32u5/power_manager.c
@@ -31,6 +31,7 @@
 #include <sys/rtc.h>
 #endif
 
+#include "../fuel_gauge/battery_model.h"
 #include "../power_manager_poll.h"
 #include "../stwlc38/stwlc38.h"
 #include "power_manager_internal.h"
@@ -44,6 +45,7 @@ pm_driver_t g_pm = {
 static void pm_monitoring_timer_handler(void* context);
 static void pm_shutdown_timer_handler(void* context);
 static bool pm_load_recovery_data(pm_recovery_data_t* recovery);
+static pm_status_t pm_wait_to_stabilize(pm_driver_t* drv, uint32_t timeout_ms);
 
 pm_status_t pm_init(bool inherit_state) {
   pm_driver_t* drv = &g_pm;
@@ -142,14 +144,12 @@ pm_status_t pm_init(bool inherit_state) {
 
   irq_unlock(irq_key);
 
-  // Poll until fuel_gauge is initialized and first PMIC & WLC measurements
-  // propagates into power_monitor.
-  bool state_machine_stabilized;
-  do {
-    irq_key_t irq_key = irq_lock();
-    state_machine_stabilized = drv->state_machine_stabilized;
-    irq_unlock(irq_key);
-  } while (!state_machine_stabilized);
+  // Wait to stabilize the state machine
+  pm_status_t status = pm_wait_to_stabilize(drv, PM_STABILIZATION_TIMEOUT_MS);
+  if (status != PM_OK) {
+    pm_deinit();
+    return status;
+  }
 
   drv->initialized = true;
 
@@ -210,22 +210,15 @@ pm_status_t pm_get_state(pm_state_t* state) {
 }
 
 // This callback is called from inside the system_suspend() function
-// when the rtc wake-up timer expires. The callback can perform
-// measurements and update the fuel gauge state.
+// when the rtc wake-up timer expires.
 // - The callback can schedule the next wake-up by calling
 // rtc_wakeup_timer_start().
 // - If the callback return with wakeup_flags set, system_suspend() returns.
 #ifdef USE_RTC
 void pm_rtc_wakeup_callback(void* context) {
-  // TODO: update fuel gauge state
-  // TODO: decide whether to reschedule the next wake-up or wake up the coreapp
-  if (true) {
-    // Reschedule the next wake-up
-    rtc_wakeup_timer_start(10, pm_rtc_wakeup_callback, NULL);
-  } else {
-    // Wake up the coreapp
-    wakeup_flags_set(WAKEUP_FLAG_RTC);
-  }
+  // No need to do anything here, but left as a placeholder for potential
+  // future use. This is an optimal place where to set a wakeup flag from RTC
+  // wakeup_flags_set(WAKEUP_FLAG_RTC);
 }
 #endif
 
@@ -247,6 +240,7 @@ pm_status_t pm_suspend(wakeup_flags_t* wakeup_reason) {
 
   // Something went wrong, suspend request was not accepted
   if (drv->request_suspend == true || drv->state != PM_STATE_SUSPEND) {
+    drv->request_suspend = false;
     irq_unlock(irq_key);
     return PM_REQUEST_REJECTED;
   }
@@ -254,15 +248,13 @@ pm_status_t pm_suspend(wakeup_flags_t* wakeup_reason) {
   irq_unlock(irq_key);
 
 #ifdef USE_RTC
-  // Automatically wakes up after specified time and call pm_rtc_wakeup_callback
-  rtc_wakeup_timer_start(10, pm_rtc_wakeup_callback, NULL);
+  // Read the current timestamp before entering suspend mode
+  if (!rtc_get_timestamp(&drv->suspend_timestamp)) {
+    return PM_ERROR;
+  }
 #endif
 
   wakeup_flags_t wakeup_flags = system_suspend();
-
-#ifdef USE_RTC
-  rtc_wakeup_timer_stop();
-#endif
 
   // TODO: Handle wake-up flags
   // UNUSED(wakeup_flags);
@@ -539,6 +531,40 @@ bool pm_driver_suspend(void) {
     return false;
   }
 
+#ifdef USE_RTC
+
+  // Capture the timestamp when device was active for the last time.
+  if (!rtc_get_timestamp(&drv->last_active_timestamp)) {
+    return false;
+  }
+
+  if (drv->usb_connected || drv->wireless_connected) {
+    drv->suspended_charging = true;
+
+    // External power source is connected, wake up early to update the fuel
+    // gauge
+    rtc_wakeup_timer_start(PM_SUSPENDED_CHARGING_TIMEOUT_S,
+                           pm_rtc_wakeup_callback, NULL);
+
+  } else {
+    if ((drv->last_active_timestamp - drv->suspend_timestamp) >=
+        PM_AUTO_HIBERNATE_TIMEOUT_S) {
+      // Device is very long time in suspend mode without external power source,
+      // hibernate it to save power.
+      pm_hibernate();
+    }
+
+    uint32_t time_to_hibernate =
+        PM_AUTO_HIBERNATE_TIMEOUT_S -
+        (drv->last_active_timestamp - drv->suspend_timestamp);
+
+    drv->suspended_charging = false;
+
+    rtc_wakeup_timer_start(time_to_hibernate, pm_rtc_wakeup_callback, NULL);
+  }
+
+#endif
+
   irq_key_t irq_key = irq_lock();
   systimer_delete(drv->monitoring_timer);
   irq_unlock(irq_key);
@@ -553,6 +579,17 @@ bool pm_driver_resume(void) {
     return false;
   }
 
+  drv->woke_up_from_suspend = true;
+
+#ifdef USE_RTC
+  rtc_wakeup_timer_stop();
+
+  uint32_t rtc_timestamp;
+  rtc_get_timestamp(&rtc_timestamp);
+  drv->time_in_suspend_s = (rtc_timestamp - drv->last_active_timestamp);
+
+#endif
+
   // Recreate the monitoring timer
   drv->monitoring_timer = systimer_create(pm_monitoring_timer_handler, NULL);
   if (drv->monitoring_timer == NULL) {
@@ -565,12 +602,46 @@ bool pm_driver_resume(void) {
   // Set the periodic sampling period
   systimer_set_periodic(drv->monitoring_timer, PM_BATTERY_SAMPLING_PERIOD_MS);
 
+  // Wait for pmic measurements to stabilize the fuel gauge estimation.
+  pm_status_t status = pm_wait_to_stabilize(drv, PM_STABILIZATION_TIMEOUT_MS);
+  if (status != PM_OK) {
+    // timeout during state machine stabilization
+    return false;
+  }
   return true;
 }
 
 bool pm_driver_is_suspended(void) {
   // No specific pending tasks, just return true
   return true;
+}
+
+void pm_compensate_fuel_gauge(float* soc, uint32_t elapsed_s,
+                              float battery_current_ma, float bat_temp_c) {
+  float compensation_mah = ((battery_current_ma)*elapsed_s) / 3600.0f;
+  bool discharging_mode = battery_current_ma >= 0.0f;
+  *soc -=
+      (compensation_mah / battery_total_capacity(bat_temp_c, discharging_mode));
+  return;
+}
+
+static pm_status_t pm_wait_to_stabilize(pm_driver_t* drv, uint32_t timeout_ms) {
+  uint32_t expire_time = ticks_timeout(timeout_ms);
+
+  // Poll until fuel_gauge is initialized and first PMIC & WLC measurements
+  // propagates into power_monitor.
+  bool state_machine_stabilized;
+  do {
+    if (ticks_expired(expire_time)) {
+      return PM_TIMEOUT;
+    }
+
+    irq_key_t irq_key = irq_lock();
+    state_machine_stabilized = drv->state_machine_stabilized;
+    irq_unlock(irq_key);
+  } while (!state_machine_stabilized);
+
+  return PM_OK;
 }
 
 #endif
--- a/core/embed/sys/power_manager/stm32u5/power_manager.c
+++ b/core/embed/sys/power_manager/stm32u5/power_manager.c
@@ -107,6 +107,27 @@ pm_status_t pm_init(bool inherit_state) {
   irq_key_t irq_key = irq_lock();
 
   if (recovery_ok) {
+#ifdef USE_RTC
+
+    // RTC compensation should happen only during initialization in bootloader
+    if (!inherit_state) {
+      // Get RTC timestamp and compare it with the timestamp from recovery data
+      // to estimate time off and compensate self-discharge of the battery.
+      uint32_t rtc_timestamp;
+      if (recovery.last_capture_timestamp != 0 &&
+          rtc_get_timestamp(&rtc_timestamp)) {
+        // If the RTC timestamp is older than the last captured timestamp,
+        // we will not use it.
+        if (rtc_timestamp >= recovery.last_capture_timestamp) {
+          pm_compensate_fuel_gauge(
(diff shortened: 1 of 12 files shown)


core/embed/sys/power_manager/stm32u5/power_manager_internal.h
```diff
@@ -40,13 +40,23 @@
 #define PM_BATTERY_CHARGING_CURRENT_MIN PMIC_CHARGING_LIMIT_MIN
 #define PM_BATTERY_SAMPLING_BUF_SIZE 10
 
+#define PM_SELF_DISG_RATE_HIBERATION_MAH 0.004f
+#define PM_SELF_DISG_RATE_SUSPEND_MAH 0.032f
```
TychoVrahe: suspending actually has variants: currently we can suspend with out without BLE connection keeping live, which has significant impact on discharge current

Hannsek:
24 hours seem like a lot, I'd shorten it to e.g. 6 hours.

core/embed/sys/time/stm32u5/rtc.c
```diff
@@ -72,6 +72,34 @@ bool rtc_init(void) {
...
+  RTC_DateTypeDef date;
+  RTC_TimeTypeDef time;
+
+  // Get current time and date, date has to be read first.
+  if (HAL_OK != HAL_RTC_GetTime(&drv->hrtc, &time, RTC_FORMAT_BIN)) {
+    return false;
+  }
+
+  // Get the current date
+  if (HAL_OK != HAL_RTC_GetDate(&drv->hrtc, &date, RTC_FORMAT_BIN)) {
```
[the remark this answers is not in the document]
kopecdav: Order is correct, fixed comment here.

core/embed/sys/power_manager/stm32u5/power_manager_internal.h
```diff
@@ -40,13 +40,23 @@
 #define PM_BATTERY_CHARGING_CURRENT_MIN PMIC_CHARGING_LIMIT_MIN
 #define PM_BATTERY_SAMPLING_BUF_SIZE 10
 
+#define PM_SELF_DISG_RATE_HIBERATION_MAH 0.004f
```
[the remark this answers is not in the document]
kopecdav: Fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/07acecb140dfc91ddf752f5dafe213386a61d350

core/embed/sys/power_manager/stm32u5/power_manager.c
```diff
@@ -532,4 +552,124 @@ static void pm_shutdown_timer_handler(void* context) {
...
+  return true;
+}
+
+void pm_compensate_fuel_gauge(float* soc, uint32_t elapsed_s,
+                              float battery_current_ma, float bat_temp_c) {
+  float compensation_mah = ((battery_current_ma)*elapsed_s) / 3600.0f;
+  bool discharging_mode = battery_current_ma >= 0.0f;
+  *soc -=
+      (compensation_mah / battery_total_capacity(bat_temp_c, discharging_mode));
+  return;
```
TychoVrahe: needless return statement. maybe we could return soc instead of having pointer in argument if we don't need to return anything else?

core/embed/sys/time/stm32u5/rtc.c
```diff
@@ -72,6 +72,34 @@ bool rtc_init(void) {
...
+    return false;
+  }
+
+  *timestamp =
+      ((date.Year * 365 * 24 * 3600) + ((date.Month - 1) * 30 * 24 * 3600) +
+       ((date.Date - 1) * 24 * 3600) + (time.Hours * 3600) +
+       (time.Minutes * 60) + time.Seconds);
```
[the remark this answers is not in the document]
TychoVrahe: i agree with AI here, lets implement proper conversion

core/embed/sys/power_manager/stm32u5/power_monitoring.c
```diff
@@ -29,119 +29,96 @@
...
+      // Just woke up from suspend, use the last known battery data to
+      // update the fuel gauge.
+      if (drv->suspended_charging) {
+        pm_compensate_fuel_gauge(&drv->fuel_gauge.soc, drv->time_in_suspend_s,
+                                 drv->pmic_data.ibat, drv->pmic_data.ntc_temp);
+
+      } else {
+        // Use known battery self-discharge rate to compensate the fuel gauge
+        // estimation during the suspend period. Since this period may be very
+        // long and the batery temperature may vary, use the average ambient
```
TychoVrahe: type battery

core/embed/sys/power_manager/stm32u5/power_monitoring.c
```diff
@@ -29,119 +29,96 @@
...
+        // temperature.
+        pm_compensate_fuel_gauge(&drv->fuel_gauge.soc, drv->time_in_suspend_s,
+                                 PM_SELF_DISG_RATE_SUSPEND_MA, 25.0f);
```
TychoVrahe, line 85: is using 25 degrees intentional? is it actually better than using current measurement?

core/embed/sys/power_manager/stm32u5/power_manager.c
```diff
@@ -532,4 +552,124 @@ static void pm_shutdown_timer_handler(void* context) {
...
+  if (status != PM_OK) {
+    // timeout during state machine stabilization
+    return false;
+  }
+  return true;
+}
+
+bool pm_driver_is_suspended(void) {
+  // No specific pending tasks, just return true
+  return true;
```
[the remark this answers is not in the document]
TychoVrahe: yes i'd also add a suspended flag set in pm_driver_suspend and unset in pm_driver_resume, with pm_driver_resume checking whether the driver is actually suspended before proceeding

TychoVrahe (changes requested):

core/embed/sys/power_manager/stm32u5/power_monitoring.c
kopecdav: Well in hibernation for sure, since we dont know where exactly the device was stored all the time, so taking actual battery temperature after wake-up might be off, I assumed that taking the typical ambient temperature might be a best guess since user will be guided to store the device in such temperature as well .. in suspend is bit questionable .. I think that it depends on the maximal suspended period. If we decide that we will go to hibernation after 6h maybe we can just use the measured temp, if it gonna be several days i would chose same strategy. Anyway, I updated the comment there https://github.com/trezor/trezor-firmware/pull/5276/commits/31f27f04f8d861a59f7f65f9aaaae228f211906b

core/embed/sys/power_manager/stm32u5/power_monitoring.c
kopecdav: fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/31f27f04f8d861a59f7f65f9aaaae228f211906b

core/embed/sys/power_manager/stm32u5/power_manager.c
kopecdav: I removed return here https://github.com/trezor/trezor-firmware/pull/5276/commits/31f27f04f8d861a59f7f65f9aaaae228f211906b. I wanted to avoid code like this in the main routines, and rather nest the variable update in the function. 

```
drv->soc = pm_compensation_fuel_gauge(drv->soc, ..)
```

core/embed/sys/power_manager/stm32u5/power_manager_internal.h
kopecdav: fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/07acecb140dfc91ddf752f5dafe213386a61d350

core/embed/sys/power_manager/stm32u5/power_manager.c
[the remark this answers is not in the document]
kopecdav: fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/31f27f04f8d861a59f7f65f9aaaae228f211906b

core/embed/sys/time/stm32u5/rtc.c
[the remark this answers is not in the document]
kopecdav: Right, fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/2214847fa146580c1cd5a24930241defc86fb261.

TychoVrahe (approved):

core/embed/sys/time/inc/sys/rtc.h
```diff
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include <trezor_bsp.h>
```
cepetr: I think this include is not needed here.  In general, we should avoid including `trezor_bsp.h` in "public" headers.

core/embed/sys/power_manager/stm32u5/power_manager.c
```diff
@@ -532,4 +552,132 @@ static void pm_shutdown_timer_handler(void* context) {
...
+
+    drv->suspended_charging = false;
+
+    rtc_wakeup_timer_start(time_to_hibernate, pm_rtc_wakeup_callback, NULL);
+  }
+
+#endif
+
+  irq_key_t irq_key = irq_lock();
+  systimer_delete(drv->monitoring_timer);
```
cepetr, line 631: We shouldn't need to wrap `systimer_delete()` with `irq_lock/irq_unlock`, as it is expected to be thread-safe.

However, it currently isn't. We should fix that in the implementation of `systimer_delete` and remove the external synchronization:

```c
void systimer_delete(systimer_t* timer) {
  systimer_driver_t* drv = &g_systimer_driver;
  if (!timer_valid(drv, timer)) {
    return;
  }

  irq_key_t irq_key = irq_lock();
  timer->callback = NULL;
  irq_unlock(irq_key);
}
```

core/embed/sys/power_manager/stm32u5/power_manager.c
```diff
@@ -532,4 +552,132 @@ static void pm_shutdown_timer_handler(void* context) {
...
+  }
+
+  // Request new pmic measurement
+  pmic_measure(pm_pmic_data_ready, NULL);
+
+  // Set the periodic sampling period
+  systimer_set_periodic(drv->monitoring_timer, PM_BATTERY_SAMPLING_PERIOD_MS);
+
+  // Wait for pmic measurements to stabilize the fuel gauge estimation.
+  pm_status_t status = pm_wait_to_stabilize(drv, PM_STABILIZATION_TIMEOUT_MS);
```
cepetr: Driver `_resume()` functions should not be blocking; however, we can wait indirectly by returning false from the `_is_suspended()` function.

core/embed/sys/power_manager/stm32u5/power_monitoring.c
```diff
@@ -29,119 +29,103 @@
...
-    }
+  // Store measurement timestamp
+  if (drv->pmic_last_update_ms == 0) {
+    drv->pmic_sampling_period_ms = PM_BATTERY_SAMPLING_PERIOD_MS;
   } else {
-    if (drv->usb_connected) {
-      drv->usb_connected = false;
-    }
+    // Calculate the time since the last PMIC update
+    drv->pmic_sampling_period_ms = systick_ms() - drv->pmic_last_update_ms;
```
cepetr: `drv->pmic_sampling_period_ms` might receive a nonsensical value if the timer wraps around. You might consider using `systick_us()`, which returns a 64-bit value and never wraps.

core/embed/sys/time/stm32u5/rtc.c
```diff
@@ -139,4 +169,51 @@ void RTC_IRQHandler(void) {
...
+    30,  // November
+    31   // December
+};
+
+static uint8_t bcd2bin(uint8_t val) { return (val & 0x0F) + ((val >> 4) * 10); }
+
+// Check for leap year
+static int is_leap_year(int year) {
+  year += 2000;  // STM32 RTC uses years 0-99 for 2000-2099
+  return ((year % 4 == 0 && year % 100 != 0) || (year % 400 == 0));
```
cepetr, line 191: It seems to me that the condition is unnecessarily complex if the year argument is always in the range 0..99, corresponding to years 2000..2099. In that case, `return (year % 4) == 0;`would be sufficient.

I'm fine with keeping the full condition for clarity or future-proofing, but in that case, you should probably remove the `+ 2000` from this function and instead call it with `is_leap_year(2000 + y)`.

core/embed/sys/time/stm32u5/rtc.c
```diff
@@ -139,4 +169,51 @@ void RTC_IRQHandler(void) {
...
+static uint32_t rtc_calendar_to_timestamp(const RTC_DateTypeDef* date,
+                                          const RTC_TimeTypeDef* time) {
+  uint8_t year = bcd2bin(date->Year);    // 0..99
+  uint8_t month = bcd2bin(date->Month);  // 1..12
+  uint8_t day = bcd2bin(date->Date);     // 1..31
+  uint8_t hour = bcd2bin(time->Hours);
+  uint8_t min = bcd2bin(time->Minutes);
+  uint8_t sec = bcd2bin(time->Seconds);
+
+  uint32_t days = 0;
```
cepetr, line 206: In general, the registers may contain values outside the expected range (for example, the month could be set to 15). We should consider detecting such cases and resetting the RTC to its default state.

core/embed/sys/time/stm32u5/rtc.c
```diff
@@ -139,4 +169,51 @@ void RTC_IRQHandler(void) {
...
+  for (int y = 0; y < year; ++y) {
```
cepetr, line 207: The code is correct, but in this particular case, we could simplify it to:
```c
    static const uint16_t cum_days[12] = {0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334}
    int days = (y * 365) + (y + 3) / 4;
    days += cum_days[m - 1] + d - 1;
    if (m > 2 && (y % 4) == 0) {
        days += 1;
    }
```

But I am ok with the current version as well.

core/embed/sys/time/stm32u5/rtc.c
TychoVrahe: i'd say lets keep the condition complete and actually change the timestamp to unix timestamp, so number of seconds since 1970 - which might be more useful. With assumption that RTC is correctly setup, which we will be doing in prodtest after https://github.com/trezor/trezor-firmware/pull/5324

core/embed/sys/time/stm32u5/rtc.c
kopecdav: This is a nice optimization, but it simplifies the leap year condition, so I would leave the current implementation for now.

core/embed/sys/time/stm32u5/rtc.c
kopecdav: RTC is reset to default during initialization if the battery dies completely and HW should wrap-around correctly. I would leave this for now since we have now backup plan what to do in the power manager if this would actually fail.

core/embed/sys/time/stm32u5/rtc.c
kopecdav: Fixed here. https://github.com/trezor/trezor-firmware/pull/5276/commits/bcf7cdc01d43a20a505120209329432800257dde

core/embed/sys/power_manager/stm32u5/power_manager.c
kopecdav: Fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/7d69013505c374d22f0dc2bf9338c9a786f6de51

core/embed/sys/time/inc/sys/rtc.h
kopecdav: Fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/bcf7cdc01d43a20a505120209329432800257dde

core/embed/sys/power_manager/stm32u5/power_monitoring.c
kopecdav: Fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/ef7db1bba2694826afb1092c163b3d506d55b857

core/embed/sys/power_manager/stm32u5/power_manager.c
kopecdav: Fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/86fdacfefeca2ca1ec71de2367ce483b37d3cfb2. Resume function is not blocking anymore, the polling for stabilization was moved to upper layer into `pm_suspend`. 

Suspend function was restructured. In case recent pmic data are available, driver do not need to wait and goes to suspend right away. If the driver just woke up from suspend to update the fuel gauge and the initial acquisition haven`t finish, driver will start suspending at background and poll for the suspended flag until its finished in `driver_is_suspended()` function.

core/embed/sys/power_manager/stm32u5/power_manager.c
```diff
@@ -532,4 +559,156 @@ static void pm_shutdown_timer_handler(void* context) {
...
+  return true;
+}
+
+bool pm_driver_is_suspended(void) {
+  pm_driver_t* drv = &g_pm;
+
+  // Poll until driver gets suspended
+  bool suspended;
+  do {
+    irq_key_t irq_key = irq_lock();
```
cepetr: I think you don't need (or maybe should not) wait here. Just return `drv->suepended` from inside `irq_lock/unlock` here. The outer loop wait for you.

cepetr (changes requested):

core/embed/sys/power_manager/stm32u5/power_manager.c
kopecdav: True! Fixed here https://github.com/trezor/trezor-firmware/pull/5276/commits/790007d2385a43b391c5e60e4da662abe4f5815e

PR #5277, opened 2025-06-30
chore(core): return choices' count in `ButtonRequest`

romanz:
[no changelog]


romanz:
Will be done in #5274.

PR #5279, opened 2025-07-01
chore(core): remove `Cargo.lock` generated by `rust-analyzer`

romanz:
Also, add `target/` to `core/embed/rust/.gitignore`.


matejcik:
whole of `embed/rust/target` should be gitignored, no? if it isn't, please also do that

romanz:
> whole of `embed/rust/target` should be gitignored, no? if it isn't, please also do that

Done in c92b60d177.

bieleluk (approved):

PR #5280, opened 2025-07-01
add ble radio test

TychoVrahe:
This PR adds radio test build.


kopecdav:
Can you please add radio test build command to the nordic/trezor/README? 
Here is what we used before.


### Building radio-test

To build the radio-test application which is used for BLE chamber test
you can use following command.

```sh
west build ./radio_test/ -b t3w1_revA_nrf52832 --sysbuild --pristine
````

TychoVrahe:
added in https://github.com/trezor/trezor-firmware/commit/6a6d0f8e6f8e654ea9afa66c5b2c946ac1ffd29f

i also added a precompiled radio test binary for T3W1: https://github.com/trezor/trezor-firmware/commit/160082d6fb8d08fbfad15279d71f6c17a17f4185 (this can be used for easy switching between regular BLE app and radio test in prodtest)

PR #5281, opened 2025-07-01
Implement prodtest `secrets-write` and `optiga-pair`

onvej-sl:


onvej-sl:
@TychoVrahe Please review `set_random_secret()` and `prodtest_secrets_write()` in https://github.com/trezor/trezor-firmware/pull/5281/commits/d2f5d0a193fe3478b9b66d70d085d66cbbcf585a.
@andrewkozlik Please review the rest.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -0,0 +1,135 @@
...
+
+#ifdef USE_OPTIGA
+  uint8_t optiga_secret[length];
+  if (OPTIGA_SUCCESS != optiga_get_random(secret, length)) {
+    return secfalse;
+  }
+  for (size_t i = 0; i < length; i++) {
+    secret[i] ^= optiga_secret[i];
+  }
+#endif
```
andrewkozlik: 1. bug: optiga_get_random overwrites secret
2. memzero optiga_secret  when done or get rid of the buffer like this:
```suggestion
#ifdef USE_OPTIGA
  if (OPTIGA_SUCCESS != optiga_get_random(secret, length)) {
    return secfalse;
  }
#endif
random_xor(secret, length);
```

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -0,0 +1,135 @@
...
+#endif
+
+  cli_ok(cli, "");
+}
+
+// clang-format off
+
+PRODTEST_CLI_CMD(
+  .name = "secrets-write",
```
andrewkozlik: Maybe `secrets-init` would be more appropriate, since `*-write` commands usually have a parameter of what to write.

onvej-sl: Renamed in https://github.com/trezor/trezor-firmware/pull/5281/commits/ec838b04cee84e0d8d2dac92ddfca0b519733364.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5281/commits/ec838b04cee84e0d8d2dac92ddfca0b519733364.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -0,0 +1,136 @@
...
+  return sectrue;
+}
+
+secbool set_random_secret(uint8_t slot, size_t length) {
+  uint8_t secret[length];
+  uint8_t secret_read[length];
+
+  secbool ret = secfalse;
+
+  if (secret_key_get(slot, secret_read, sizeof(secret_read)) != sectrue) {
```
TychoVrahe: `secret_key_writable` would be better , otherwise you will fail with RSOD when this is attempted on unlocked bootloader (`secret_key_get` returns `secfalse` as the key is no longer present, but it cannot be written as the flash is locked)

TychoVrahe (changes requested):
The new prodtest command also deserves description in https://github.com/trezor/trezor-firmware/blob/main/core/embed/projects/prodtest/README.md

core/embed/projects/prodtest/cmd/prodtest_secrets.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5281/commits/3eb801b94d0dcbdf2378d5e3dce2281d6956500c.

core/embed/projects/prodtest/README.md
```diff
@@ -620,6 +620,15 @@ secrets-init
 OK
 ```
 
+### optiga-pair
+Writes the pairing secret to the Optiga chip to pair it with the MCU. The command `optiga-pair` must be executed before calling this command.
```
TychoVrahe: You probably meant secrets-init should be called before.

onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5281/commits/cf2563194ff428191df87af053cfa62cec2308a5.

TychoVrahe (approved):

PR #5286, opened 2025-07-02
Add support for EIP-712 typed hash signatures for non-legacy devices

### LINKED ISSUE #5294: Add support for EthereumSignTypedHash to new firmware
The [Base](https://github.com/base) and [Optimism](https://github.com/ethereum-optimism) teams use security councils for contract upgrades. These security councils are made up of globally distributed individuals who maintain a hardware key as a signer to an onchain multisig contract (we use [Safe](https://github.com/safe-global/safe-smart-account)).

Generally these individual signers use Ledger for their hardware wallets. However, recently we'd like to add Trezor support to our workflow. The contract upgrade workflow is to open a PR against [ethereum-optimism/superchain-ops](https://github.com/ethereum-optimism/superchain-ops/) or [base/contract-deployments](https://github.com/base/contract-deployments/) with the upgrade script as well as validation instructions.

You can see an example of one of these validation documents [here](https://github.com/base/contract-deployments/blob/main/mainnet/2025-05-15-eip1559-denominator-reduction/VALIDATION.md). Note in particular, the most important piece of the validation is checking the EIP-712 message hash.

Here is a video of a Ledger device confirming a typed EIP-712 payload:
https://github.com/user-attachments/assets/aa4fa487-0699-4399-bda6-bf1f849ea226. Note that it doesn't support showing the individual values in the EIP-712 message struct; rather it hashes the entire message and shows the message hash.

Similarly, the Trezor Model One only supports `EthereumSignTypedHash` and not `EthereumSignTypedData` (added in #1970). You can view the confirmation of one of these messages in this video: https://github.com/user-attachments/assets/4e0519ad-fec5-4fa7-aeb4-1ecd39f7d2e9.

However, the Trezor Model T / Safe 3 / Safe 5 don't support `EthereumSignTypedHash`, they only support `EthereumSignTypedData`. This forces us to send the entire typed payload to the device, rather than just the hash. Here's a video of what that looks like: https://github.com/user-attachments/assets/dc3e42cf-0d33-4b69-96fd-555d7df96de2. Note that the user has to confirm each of the fields in the struct, including the `data` field.

Given our requirement for the user to confirm the message hash matches exactly, this doesn't work for these Trezor devices. We could totally update our process to print out the full typed data struct and get the user to validate this, however this is both laborious (there's a bunch of fields in Safe transactions that are irrelevant) and error prone (the data field can be very large as it is calldata, expecting a user to validate every byte is tricky; not to mention that a single byte change can completely change the behavior of the upgrade). This is what hashes are for: any small change in the data will cause the hash to change completely, so validating the hash is much safer.

We'd love to see newer Trezor devices support `EthereumSignTypedHash`. I believe the intent for `EthereumSignTypedData` is to validate user-readable structs (the example given in [EIP-712](https://eips.ethereum.org/EIPS/eip-712) is an email with `from/to/contents` fields), and it doesn't make sense for signing transaction calldata. Signing the hash directly is much safer for this usecase.

mdehoog:
Fixes #5294, please see that issue for a more detailed rationale for this PR.

Screenshot 2025-07-02 at 2 57 15 PM
Screenshot 2025-07-02 at 2 57 24 PM
Screenshot 2025-07-02 at 2 57 31 PM
Screenshot 2025-07-02 at 2 57 41 PM

#1835 added support for signing EIP-712 data payloads for Trezor T, but without support for signing EIP-712 data hashes.
#1970 added support for signing EIP-712 hashes for legacy Trezors, but not for newer models.

This PR adds support for signing data hashes to newer devices. This is useful for signing EIP-712 hashes directly, for example for [(Gnosis) Safe signatures](https://github.com/safe-global/safe-smart-account/blob/6c4b4fd8f160186fbc6a88a9d864578eea815bdd/contracts/Safe.sol#L389-L473).

Context: we maintain https://github.com/base/eip712sign for signing typed hashes with hardware devices. We recently started work on adding Trezor support (see base/eip712sign#19).


core/translations/en.json
```diff
@@ -372,6 +374,7 @@
     "ethereum__title_confirm_message": "Confirm message",
     "ethereum__title_confirm_struct": "Confirm struct",
     "ethereum__title_confirm_typed_data": "Confirm typed data",
+    "ethereum__title_confirm_typed_hashes": "Confirm typed hashes",
```
mdehoog, line 377: could alternatively remove this and just reuse `ethereum__title_confirm_typed_data`

matejcik:
> This is useful for signing EIP-712 hashes where the data / domain cannot be customized, like for [(Gnosis) Safe signatures](https://github.com/safe-global/safe-smart-account/blob/6c4b4fd8f160186fbc6a88a9d864578eea815bdd/contracts/Safe.sol#L389-L473).

i have no idea what that means. how can you not customize data/domain in an EIP712 payload?

> This PR adds support for signing data hashes to newer devices.

absent a **very good** rationale, we don't want to introduce blind-signing unknown EIP712 payloads

mdehoog:
@matejcik thanks very much for your reply... and apologies for not including a good rationale. I've opened https://github.com/trezor/trezor-firmware/issues/5294 which explains our use case better.

> > This is useful for signing EIP-712 hashes where the data / domain cannot be customized, like for [(Gnosis) Safe signatures](https://github.com/safe-global/safe-smart-account/blob/6c4b4fd8f160186fbc6a88a9d864578eea815bdd/contracts/Safe.sol#L389-L473).
> 
> i have no idea what that means. how can you not customize data/domain in an EIP712 payload?

I incorrectly assumed we couldn't send domain separators that were compatible with Safe's transaction hash using the existing typed data struct messages; please ignore this comment (I've removed it from the PR text).

mdehoog:
Replaced by #5344

PR #5291, opened 2025-07-02
fix(core): don't send `ButtonRequest` during menu handling

romanz:
Since existing BTC signature tests [expect specific ButtonRequest sequence](https://github.com/trezor/trezor-firmware/blob/df9567093086155aad710f827fd60eb80472ffef/tests/device_tests/bitcoin/test_signtx.py#L129-L146), use `LayoutContent.page_count()` instead of `ButtonRequest.pages`.


romanz:
Also, added a few unrelated fixups.

ibz (approved):

PR #5297, opened 2025-07-03
Use random suffix for BLE name when pairing

### LINKED ISSUE #5287: implement random BLE device name in firmware
- in pairing mode, three radnom chars should be added (uppercase and numbers), i.e. Trezor T3W1 (UG4)
- when user sets its device label, don't add random chars
- in non-pairing mode, random chars should not be used

See bootloader implementation in https://github.com/trezor/trezor-firmware/pull/5269

mmilata:
Fixes #5287.

Screenshot From 2025-07-03 14-48-09


STew790:
**QA OK**
I can see the random BLE name when pairing.

PR #5298, opened 2025-07-03
Add MockAsync and await_result_patched

M1nd3r:
This PR allows to replace some unnecessary set-ups / tear-downs in Python unit tests by patching the problematic async function (most commonly `get_seed()`).


core/tests/test_apps.ethereum.keychain.py
```diff
@@ -118,8 +103,8 @@ async def handler(msg, keychain, defs):
             self._check_keychain(keychain, 1)
             self.assertIs(defs.network, UNKNOWN_NETWORK)
 
-        await_result(handler(EthereumGetAddress(address_n=[])))
-        await_result(handler(EthereumGetAddress(address_n=[0])))
+            await_result_patched(handler(EthereumGetAddress(address_n=[])))
+            await_result_patched(handler(EthereumGetAddress(address_n=[0])))
```
mmilata: I'm confused by this indentation change, now it seems like a recursive function definition that never gets called?

mmilata (commented):
Looks reasonable to me but let's wait for matejcik a bit if he can think of better approach.

core/tests/test_apps.ethereum.keychain.py
M1nd3r: fixed here: https://github.com/trezor/trezor-firmware/pull/5298/commits/cb609ebb23eda334d97a1ff3c0789389052608ab

Note: this error happened, because in previous iteration, I was using 
```
with patch_get_seed():
   await_result(...)
```

which got replaced by the `await_result_patched`

matejcik:
ACK on MockAsync

`await_result_patched` is very weird though.
ISTM a cleaner solution is to make a subclass of `TestCase` that _already includes_ the required `setUp` / `tearDown` steps, and (e.g.) looks into a classvar for the seed value, with default being "all all all"

like so
```
# common.py

ALL_ALL_ALL = " ".join(["all"] * 12)

class TestCaseWithSeed(unittest.TestCase):
    SEED = bip39.seed(ALL_ALL_ALL, "")

    def setUpClass(self):
        context.CURRENT_CONTEXT = CodecContext(None, bytearray(64))

    def tearDownClass(self):
        context.CURRENT_CONTEXT = None

    def setUp(self):
        cache_codec.start_session()
        cache_codec.get_active_session().set(cache_common.APP_COMMON_SEED, self.SEED)
```
then
```
# test_blabla.py

class TestBitcoinKeychain(TestCaseWithSeed):
    SEED = bip39.seed("abandon abandon abandon ...", "")

    def test_foo():
        ...
```

M1nd3r:
@matejcik 
I would like to discuss this more - what the approach should be.

1) I wanted to avoid the need for setting the `context`, as with THP there (might) need to be two versions. One with `CodecContext` and one with `SessionContext` (THP). 

2) Previously, I used this:
```
with patch_get_seed():
   await_result(...)
```
 which allowed me to patch the minimal amount of code for the shortest duration possible.
 
 `await_result_patched` is a shortcut, so that I can write only one line, without the `with` block.
 
So, I would like to know your opinion - do we want to have the "small scope" and patch only the affected lines, or do we want to have some bigger scope that changes more of the code - which simplifies the code in the test function instead?

matejcik:
i believe "class" is the right scope for the effect. "test function" would perhaps be appropriate in special cases, but we can avoid that by doing multiple test classes.
no bigger than a class
no smaller than a function, _unless_ you intend to test multiple seeds with the same call params, in which case a sequence of with-statements (or a shortcut for that) would be appropriate.

remind me, why are we messing with context at all here? i would have naively expected that `get_active_session().set(...)` would do the thing we need?

M1nd3r:
Ok, I will drop the `await_result_patched` in favour of class(/function) scope solutions.

--> I will use this PR only for `MockAsync` changes, I will remove the other changes from the PR and do it differently in the class/function scope manner later, in another PR.

To answer your question:
> remind me, why are we messing with context at all here? i would have naively expected that `get_active_session().set(...)` would do the thing we need?

First, there is the cache rework - `cache` is now obtained through `context`. Example from `context.py`.

```
def cache_get(key: int, default: T | None = None) -> bytes | T | None:  # noqa: F811
    cache = _get_cache_for_key(key)
    return cache.get(key, default)
 
...
  
def _get_cache_for_key(key: int) -> DataCache:
    if key & SESSIONLESS_FLAG:
        return cache.get_sessionless_cache()
    if CURRENT_CONTEXT:
        return CURRENT_CONTEXT.cache
    raise Exception("No wire context")
```
    
function `get_seed` is wrapped by a decorator `@cache.stored_async(APP_COMMON_SEED)`.
If the `CURRENT_CONTEXT` is `None`, than the decorator invokes (after some hoops) the function `_get_cache_for_key` which fails. The old solution was to set the `CURRENT_CONTEXT` appropriately in the `test set-up`, and start a session so that a *backing cache entry* is created. The new solution in this PR was:
patch the import of `get_seed` in the file where it gets called - so that the call of `get_seed` invokes a different (mock) function that only returns the bytes (*seed*) without any decorator, thus the *cache backing* is not needed, neither is any session.

Another issue comes with THP: `get_active_session` is a concept that does not exist in THP, and the cache backing is different. Therefore, it is (sometimes) necessary to adjust the test set-ups/tear-downs for Codec vs THP because of these differences.

matejcik (approved):

matejcik:
> do it differently in the class/function scope manner later, in another PR.

idea: on class setup just initialize storage and put the right seed there?

PR #5305, opened 2025-07-07
Allow external menu on `confirm_value`

ibz:

### DIFF
--- a/core/embed/rust/src/ui/layout_delizia/flow/util.rs
+++ b/core/embed/rust/src/ui/layout_delizia/flow/util.rs
@@ -58,6 +58,7 @@ pub struct ConfirmValue {
     swipe_right: bool,
     frame_margin: usize,
     cancel: bool,
+    external_menu: bool,
 }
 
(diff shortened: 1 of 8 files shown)


core/embed/rust/src/ui/layout_delizia/flow/util.rs
```diff
@@ -170,6 +172,11 @@ impl ConfirmValue {
         self
     }
 
+    pub const fn with_external_menu(mut self, external_menu: bool) -> Self {
+        self.external_menu = external_menu;
```
romanz, line 176: Not sure if we also should enable `self.menu_button` if `external_menu == true`...
WDYT?

ibz: IDK, I guess it makes the API a little bit nicer... why would one want external menu that has no menu button? I'll do that ...

romanz: Thanks!

PR #5309, opened 2025-07-08
Menu support for confirm_with_info

ibz:

### DIFF
--- a/core/embed/rust/src/ui/api/firmware_micropython.rs
+++ b/core/embed/rust/src/ui/api/firmware_micropython.rs
@@ -439,9 +439,17 @@ extern "C" fn new_confirm_with_info(n_args: usize, args: *const Obj, kwargs: *mu
             .get(Qstr::MP_QSTR_verb_cancel)
             .unwrap_or_else(|_| Obj::const_none())
             .try_into_option()?;
+        let external_menu: bool = kwargs.get_or(Qstr::MP_QSTR_external_menu, false)?;
 
-        let layout =
-            ModelUI::confirm_with_info(title, subtitle, items, verb, verb_info, verb_cancel)?;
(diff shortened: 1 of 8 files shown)


core/embed/rust/src/ui/layout_caesar/component/frame.rs
```diff
@@ -35,6 +37,11 @@ where
         self
     }
 
+    pub fn with_menu(mut self, has_menu: bool) -> Self {
```
romanz: nit: IIUC the actual menu is not owned by the `Frame`, but by its child (e.g. `ShowMore::<Paragraphs<ParagraphVecShort>>` in TS3), right?

Maybe we should move the "menu-related" infomation in LayoutContent from top-level to its owner, and use `UnstructuredJSONReader.find_unique_value_by_key` to look it up?

core/embed/rust/src/ui/flow/swipe.rs
```diff
@@ -119,9 +120,15 @@ impl SwipeFlow {
             allow_swipe: true,
             pending_decision: None,
             returned_value: None,
+            has_menu: false,
```
romanz: Doesn't the menu below to a specific flow screen (and not to the `SwipeFlow` object, which may contain multiple screens)?

core/embed/rust/src/ui/layout_caesar/component/frame.rs
ibz: You are right, I changed this and it looks much better and seems to work.

core/embed/rust/src/ui/flow/swipe.rs
ibz: I removed the Delizia/Flow part for now.

PR #5312, opened 2025-07-08
fix(core): fix nmi handling

cepetr:
This small PR fixes a recently introduced issue in the NMI handler that affects the upcoming T3W1 boardloader and may also impact existing bootloaders on STM32U5 devices. **Specifically, it stops treating a Flash ECC double-error NMI as a fatal condition.**

On the STM32U5 an NMI can be raised under a few conditions, one of which is detection of a Flash ECC double-error. Although double-errors are normally extremely rare, they can occur if a Flash write- or erase-operation is interrupted by a power loss or reset.

If the boardloader or bootloader treats this NMI as a fatal error, it can brick the device when they attempt to read from a Flash region that was incompletely written and now contains ECC errors.

### DIFF
--- a/core/embed/sys/task/stm32/systask.c
+++ b/core/embed/sys/task/stm32/systask.c
@@ -803,16 +803,48 @@ __attribute__((naked, no_stack_protector)) void GTZC_IRQHandler(void) {
 }
 #endif
 
-__attribute__((no_stack_protector, used)) static void nmi_handler(
-    uint32_t msp, uint32_t exc_return) {
-  mpu_reconfig(MPU_MODE_DEFAULT);
-  // Clear pending Clock security interrupt flag
+__attribute__((no_stack_protector, used)) static void nmi_handler(void) {
+  mpu_mode_t mpu_mode = mpu_reconfig(MPU_MODE_DEFAULT);
 #ifdef STM32U5
-  RCC->CICR = RCC_CICR_CSSC;
+  if ((RCC->CIFR & RCC_CIFR_CSSF) != 0) {
+    RCC->CICR = RCC_CICR_CSSC;
 #else
-  RCC->CIR = RCC_CIR_CSSC;
+  if ((RCC->CIR & RCC_CIR_CSSF) != 0) {
+    RCC->CIR = RCC_CIR_CSSC;
 #endif
-  systask_exit_fault(msp, exc_return);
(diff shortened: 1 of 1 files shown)


TychoVrahe:
Could we differentiate whether the NMI was raised upon accessing flash from own-code versus next-state code/data?

Reason would be, if we get an error on own code, the device is unlikely to operate correctly and securely anyway - so bricking is kinda expected. 

If we can the error on accessing next-stage code while i.e. doing integrity checks, we don't want NMI as integrity check will fail anyway (unless both errors are the ECC bits i guess, so it might be worth handling too?) and we will try recovery.

cepetr:
> Could we differentiate whether the NMI was raised upon accessing flash from own-code versus next-state code/data?
> 
> Reason would be, if we get an error on own code, the device is unlikely to operate correctly and securely anyway - so bricking is kinda expected.
> 
> If we can the error on accessing next-stage code while i.e. doing integrity checks, we don't want NMI as integrity check will fail anyway (unless both errors are the ECC bits i guess, so it might be worth handling too?) and we will try recovery.

See this 1ed7e60c85e4270d5c69bdf986ccd6536a52ef4d. But I don't know how to easily test it.

core/embed/sys/task/stm32/systask.c
```diff
@@ -803,16 +803,50 @@ __attribute__((naked, no_stack_protector)) void GTZC_IRQHandler(void) {
...
+    // FLASH ECC double error
+    uint32_t addr = FLASH->ECCR & FLASH_ECCR_ADDR_ECC_Msk;
+    uint32_t bankid =
+        (FLASH->ECCR & FLASH_ECCR_BK_ECC_Msk) >> FLASH_ECCR_BK_ECC_Pos;
+#if defined(BOARDLOADER)
+    // In boardloader, this is a fatal error only if the address
+    // is in the bootloader code region.
+    if (bankid == 0 && addr >= BOARDLOADER_START &&
+        addr < BOARDLOADER_START + BOARDLOADER_MAXSIZE) {
+      // In boardloader this must not be treated as a
```
TychoVrahe: comment is unfinished (and obsolete). Applies to bootloader branch below too

cepetr: fixed, 1cf7d6256314c02a650b16ef5f908a9fb34d1944

PR #5314, opened 2025-07-08
Implement FB from user testing of Eckhart UI

bieleluk:
Implement changes coming from the user testing:
- checkbox instead of cross icon in set brightness screen
- enable swipe up in homescreen to open the menu
- use "Back" label in device menu instead of menu title

### DIFF
--- a/core/.changelog.d/5314.changed
+++ b/core/.changelog.d/5314.changed
@@ -1,0 +1,1 @@
+Limit swipe detection to the component's bounds.
--- a/core/embed/rust/src/ui/component/swipe.rs
+++ b/core/embed/rust/src/ui/component/swipe.rs
@@ -11,7 +11,7 @@ pub struct Swipe {
     pub allow_down: bool,
     pub allow_left: bool,
     pub allow_right: bool,
-
+    bounds: Rect,
     origin: Option<Point>,
 }
 
@@ -25,6 +25,7 @@ impl Swipe {
             allow_down: false,
             allow_left: false,
             allow_right: false,
+            bounds: Rect::zero(),
             origin: None,
         }
     }
@@ -70,6 +71,7 @@ impl Component for Swipe {
     type Msg = Direction;
 
     fn place(&mut self, bounds: Rect) -> Rect {
+        self.bounds = bounds;
         bounds
     }
 
@@ -78,7 +80,7 @@ impl Component for Swipe {
             return None;
         }
         match (event, self.origin) {
-            (Event::Touch(TouchEvent::TouchStart(pos)), _) => {
+            (Event::Touch(TouchEvent::TouchStart(pos)), _) if self.bounds.contains(pos) => {
                 // Mark the starting position of this touch.
                 self.origin.replace(pos);
             }
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/share_words.rs
@@ -150,6 +150,7 @@ impl<'a> Component for ShareWordsScreen<'a> {
             .inset(theme::SIDE_INSETS)
             .with_height(Self::WORD_AREA_HEIGHT);
 
+        self.page_swipe.place(bounds);
         self.header.place(header_area);
         self.subtitle.place(subtitle_area.inset(theme::SIDE_INSETS));
         self.content.place(content_area);


core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
```diff
@@ -176,6 +180,9 @@ impl Component for Homescreen {
         debug_assert_eq!(bounds.height(), SCREEN.height());
         debug_assert_eq!(bounds.width(), SCREEN.width());
 
+        // Enable swiping over the entire screen.
+        self.swipe.place(bounds);
```
obrusvit: The requirement was that the swipe should be registered only within the `action_bar.touch_area()`

core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
```diff
@@ -199,6 +206,15 @@ impl Component for Homescreen {
 
     fn event(&mut self, ctx: &mut EventCtx, event: Event) -> Option<Self::Msg> {
         self.event_fuel_gauge(ctx, event);
+
+        if let Some(Direction::Up) = self.swipe.event(ctx, event) {
```
obrusvit: Combine the two ways, something like
```rust
let swipe_up = matches!(self.swipe.event(ctx, event), Some(Direction::Up));
let homebar_tap = matches!(self.action_bar.event(ctx, event), Some(ActionBarMsg::Confirmed));

if swipe_up || homebar_tap {
    return if self.locked {
        Some(HomescreenMsg::Dismissed)
    } else {
        Some(HomescreenMsg::Menu)
    };
}
```

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -107,9 +108,9 @@ struct Submenu {
 }
 
 impl Submenu {
-    pub fn new(header_text: TString<'static>, items: Vec<MenuItem, SHORT_MENU_ITEMS>) -> Self {
+    pub fn new(items: Vec<MenuItem, SHORT_MENU_ITEMS>) -> Self {
         Self {
-            header_text,
+            header_text: TR::buttons__back.into(),
```
obrusvit: Remove the `header_text` member and use `TR::buttons__back` directly inplace in the actual `Header` construction

obrusvit (commented):
A few code-wise improvements.

core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5314/commits/313c8e539198a16d393bf24a970b20a323965712

core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5314/commits/3d0ce19660f42290d9d6b4b769919cb435ab4885
I have placed it correctly now, but it has not had any effect. The `Swipe` component processes [all](https://github.com/trezor/trezor-firmware/blob/b83857c0e5530d0e5e8de56eed1c345c8d2ed1e7/core/embed/rust/src/ui/component/swipe.rs#L81) touch events. I can create a separate issue, where the "swipe" could be started only within its bounds. WDYT?

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5314/commits/b15f849f7be2eff7bc29b02dafbeba0742351315

core/embed/rust/src/ui/layout_eckhart/firmware/homescreen.rs
obrusvit: :thinking: I thought that was the case.

Let's try it then and test whether we don't have some undesired consequences. We may need to `place` swipes in other components/layouts correctly.

bieleluk: Here it is: https://github.com/trezor/trezor-firmware/pull/5314/commits/f6a9244c76d8c14b95eab6c32b3e8310efcc18a3
I have limited only the swipe origin to the bounds. Doing so for an entire swipe wouldn't make sense, IMO.

obrusvit: Running all langs test to see if something breaks, you never know.

PR #5321, opened 2025-07-09
test(core): visit Delizia menu during tests

romanz:
Currently, only shows the menu (without visiting its children). Enabled by returning `has_menu=True` via LayoutContent JSON.

Also, reduce the number of `debuglink.read_layout()` calls in tests.

Starting be enabled it for `set_brightness`-related tests.


romanz:
Rebasing over `main`.

romanz:
Opened https://github.com/trezor/trezor-firmware/pull/5327 to update `main`'s UI fixtures.

PR #5331, opened 2025-07-10
test(core): fix waiting for recovery

romanz:
It seems that there is a race between starting the recovery flow and verifying its layout, causing [test failures when running persistence_tests on THP](https://github.com/trezor/trezor-firmware/actions/runs/16174356159/job/45656150594#step:6:235).

https://github.com/trezor/trezor-firmware/blob/953b0e23d1ae4ba433f512df95530154d13ace5d/tests/persistence_tests/test_shamir_persistence.py#L47-L49


M1nd3r (approved):
LGTM

PR #5340, opened 2025-07-11
fix tropic pin deinitialization order

TychoVrahe:
According to latest info, chip select needs to be deinitialized before turning off power, to prevent powering the device through chip select.


kopecdav:
Is there any documentation we can link this with?

TychoVrahe:
there should be an datasheet update including this, but unclear to me when and in what version will that be reflected

kopecdav (approved):

PR #5342, opened 2025-07-11
feat(core/eckhart): add tutorial

bieleluk:
[Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=11276-2725&p=f&t=633qu2hNRSwDh44n-0)


core/embed/rust/src/ui/layout_eckhart/firmware/tutorial_screen.rs
```diff
@@ -0,0 +1,116 @@
...
+const LOADER_SPEED: u16 = 2;
+const ANIM_DURATION_MS: u32 = 3000;
+
+pub enum TutorialScreenMsg {
+    Confirmed,
+}
+
+pub struct TutorialScreen {
+    text: Label<'static>,
+    /// Current value of the progress bar.
```
romanz: nit: is this comment related to `action_bar` or `value`?

core/embed/rust/src/ui/layout_eckhart/firmware/tutorial_screen.rs
```diff
@@ -0,0 +1,116 @@
...
+use super::super::{
+    component::Button,
+    constant::SCREEN,
+    cshape::{render_loader_indeterminate, ScreenBorder},
+    firmware::{ActionBar, ActionBarMsg},
+    theme,
+};
+
+const LOADER_SPEED: u16 = 2;
+const ANIM_DURATION_MS: u32 = 3000;
```
romanz: nit: maybe switch to `Duration`?
(can be definitely done in a separate PR)

core/embed/rust/src/ui/layout_eckhart/flow/show_tutorial.rs
```diff
@@ -0,0 +1,303 @@
...
+    )
+    .with_header(Header::new(TR::tutorial__what_is_tropic.into()).with_close_button())
+    .map(|msg| match msg {
+        TextScreenMsg::Cancelled => Some(FlowMsg::Cancelled),
+        _ => None,
+    })
+}
+
+pub fn new_show_tutorial() -> Result<SwipeFlow, error::Error> {
+    let content_step_welcome = TutorialScreen::new().map(|msg| match msg {
```
romanz: nit: maybe rename `TutorialScreen` (since IIUC it is only used for the first animated screen in the flow)?
```suggestion
    let content_step_welcome = TutorialWelcomeScreen::new().map(|msg| match msg {
```

core/embed/rust/src/ui/layout_eckhart/flow/show_tutorial.rs
```diff
@@ -0,0 +1,303 @@
...
+        .with_placement(LinearPlacement::vertical()),
+    )
+    .with_header(Header::new(TR::tutorial__what_is_tropic.into()).with_close_button())
+    .map(|msg| match msg {
+        TextScreenMsg::Cancelled => Some(FlowMsg::Cancelled),
+        _ => None,
+    })
+}
+
+pub fn new_show_tutorial() -> Result<SwipeFlow, error::Error> {
```
romanz, line 107: This function seems to use ~18.5kB of stack space:

Image

https://github.com/trezor/trezor-firmware/actions/runs/16445010107/job/46474541813#step:9:64

Please test it on device, to see that there are no stack overflows.

romanz: If needed, I would suggest adding `add_allocated_page` method to `SwipeFlow`:
```rust
    /// Add a page to the flow.
    ///
    /// Pages must be inserted in the order of the flow state index.
    pub fn add_page(
        &mut self,
        state: &'static dyn FlowController,
        page: impl FlowComponentDynTrait + 'static,
    ) -> Result<&mut Self, error::Error> {
        Ok(self.add_allocated_page(state, GcBoxFlowComponent::alloc(page)?))
    }

    pub fn add_allocated_page(
        &mut self,
        state: &'static dyn FlowController,
        alloc: GcBoxFlowComponent,
    ) -> &mut Self {
        debug_assert!(self.store.len() == state.index());
        unwrap!(self.store.push(alloc));
        self
    }
```
since it will allow us to construct individual pages on heap (so the stack will be reused after each page is moved to the heap).

core/src/apps/workflow_handlers.py
```diff
@@ -64,6 +64,7 @@ def _find_message_handler_module(msg_type: int) -> str:
         utils.INTERNAL_MODEL == "T2B1"
         or utils.INTERNAL_MODEL == "T3B1"
         or utils.INTERNAL_MODEL == "T3T1"
+        or utils.INTERNAL_MODEL == "T3W1"
```
romanz: Can we remove the `utils.INTERNAL_MODEL` check now (since it covers all core devices)?
```python
    if msg_type == MessageType.ShowDeviceTutorial:
        return "apps.management.show_tutorial"
```

tests/click_tests/test_tutorial_eckhart.py
```diff
@@ -0,0 +1,355 @@
...
+    pytest.mark.setup_client(uninitialized=True),
+]
+
+
+def _paginated_text(debug: "DebugLink") -> str:
+    """Get the text content of the current layout"""
+
+    layout = debug.read_layout()
+    pages = layout.page_count()
+    text = debug.read_layout().text_content()
```
romanz:
```suggestion
    text = layout.text_content()
```

tests/click_tests/test_tutorial_eckhart.py
```diff
@@ -0,0 +1,355 @@
...
+    assert TR.tutorial__tropic_info1 in text
+    assert TR.tutorial__tropic_info2 in text
+    assert TR.tutorial__tropic_info3 in text
+
+    debug.click(debug.screen_buttons.menu())
+
+
+def _wait_for_welcome_animation(debug: "DebugLink"):
+    """Wait for the welcome animation to finish."""
+    sleep(0.3)
```
romanz: Can we use `synchronize_at()` instead of `sleep()`?

tests/click_tests/test_tutorial_eckhart.py
```diff
@@ -0,0 +1,355 @@
...
+    assert "VerticalMenuScreen" in debug.read_layout().all_components()
+
+
+def _assert_htc_screen(debug: "DebugLink"):
+    layout = debug.read_layout()
+    assert layout.title() == TR.tutorial__last_one
+    assert TR.tutorial__title_hold in layout.text_content()
+
+
+def _assert_exit_screen(debug: "DebugLink"):
```
romanz: Seems quite similar to `_assert_htc_screen` :)

core/embed/rust/src/ui/layout_eckhart/firmware/tutorial_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5342/commits/084021b17973309ec6a6693033001fc3d0b09566

core/embed/rust/src/ui/layout_eckhart/firmware/tutorial_screen.rs
bieleluk: It doesn't work for `trezorapi_ui` but within Rust, it does:
https://github.com/trezor/trezor-firmware/pull/5342/commits/34c2cc94ef4a496bb145ed7092b47c628809c189

core/embed/rust/src/ui/layout_eckhart/flow/show_tutorial.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5342/commits/1cffcf5551d1831f5a9da3f5fe05e0c19712370c

core/src/apps/workflow_handlers.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5342/commits/cb95e673c47eec45ac06e5789d6aa34e1faf1bcf

tests/click_tests/test_tutorial_eckhart.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5342/commits/75b546fdfa4853424e605ade952b2040e90fe84e

tests/click_tests/test_tutorial_eckhart.py
bieleluk: At first, it wasn't possible because even the initial screen was `TextScreen`. But now it is: 
https://github.com/trezor/trezor-firmware/pull/5342/commits/5c7aba6aaacf12b51bc254b472d50c9d0b6f0aca

tests/click_tests/test_tutorial_eckhart.py
bieleluk: The screens differ only in the color gradient, which we cannot extract from the debuglink (TBH it's not necessary) . We can use one function for both screens:
https://github.com/trezor/trezor-firmware/pull/5342/commits/36a99d6526fe1726cd8a591bf147f0611c36125c

core/embed/rust/src/ui/layout_eckhart/flow/show_tutorial.rs
bieleluk: The biggest issue with `show_tutorial` and `continue_recovery_homepage` flows is that we need to show the same menu for several screens. Because the `SwipeFlow` doesn't implement a global state, we need to create a much bigger state machine with copy-pasted items, which significantly increases the memory footprint. For example, the _tropic info screen_ screen is used 3 times in the flow enum: `MenuTropic`, `StepHoldMenuTropic`, `HoldToExitMenuTropic`, and there is a function to generate the screen content:
`fn content_tropic() -> TropicScreenMap`
It would be great to solve this kind of use case. Adding an allocated page can be a temporary workaround, but it doesn't solve the issue.
WDYT?

romanz: > The biggest issue with show_tutorial and continue_recovery_homepage flows is that we need to show the same menu for several screens.

Good point - thanks!
I hope we can use the new menus (#5189) to optimize this use case.

romanz (approved):
Thanks!

PR #5344, opened 2025-07-12
Add support for displaying the EIP-712 message hash

mdehoog:
Fix for #5294.

This PR adds support for showing the EIP-712 message hash during the signing flow. An additional `optional bool message_hash_confirmation` field has been added to the `EthereumSignTypedData` protobuf message which, if true, will show an additional confirmation screen asking the user to confirm the message hash.

If unset, the existing flow remains unchanged. This ensures that only use-cases that require the message hash to be confirmed will use this new screen.

The flow is as follows (4th screenshot is the new screen):

Screenshot 2025-07-12 at 12 09 43 PM
Screenshot 2025-07-12 at 12 09 50 PM
Screenshot 2025-07-12 at 12 09 58 PM
Screenshot 2025-07-12 at 12 51 44 PM
Screenshot 2025-07-12 at 12 10 15 PM

### DIFF
--- a/common/tests/fixtures/ethereum/sign_typed_data.json
+++ b/common/tests/fixtures/ethereum/sign_typed_data.json
@@ -153,6 +153,84 @@
             }
         },
         {
+            "name": "basic_data_with_show_message_hash",
+            "parameters": {
+                "path": "m/44'/60'/0'/0/0",
+                "metamask_v4_compat": true,
+                "data": {
+                    "types": {
+                        "EIP712Domain": [
+                            {
+                                "name": "name",
+                                "type": "string"
+                            },
+                            {
+                                "name": "version",
+                                "type": "string"
+                            },
+                            {
+                                "name": "chainId",
+                                "type": "uint256"
+                            },
+                            {
+                                "name": "verifyingContract",
+                                "type": "address"
+                            }
+                        ],
+                        "Person": [
+                            {
+                                "name": "name",
+                                "type": "string"
+                            },
+                            {
+                                "name": "wallet",
+                                "type": "address"
+                            }
+                        ],
+                        "Mail": [
+                            {
+                                "name": "from",
+                                "type": "Person"
+                            },
+                            {
+                                "name": "to",
+                                "type": "Person"
+                            },
+                            {
+                                "name": "contents",
+                                "type": "string"
+                            }
+                        ]
+                    },
+                    "primaryType": "Mail",
+                    "domain": {
+                        "name": "Ether Mail",
+                        "version": "1",
+                        "chainId": 1,
+                        "verifyingContract": "0x1e0Ae8205e9726E6F296ab8869160A6423E2337E"
+                    },
+                    "message": {
+                        "from": {
+                            "name": "Cow",
+                            "wallet": "0xc0004B62C5A39a728e4Af5bee0c6B4a4E54b15ad"
+                        },
+                        "to": {
+                            "name": "Bob",
+                            "wallet": "0x54B0Fa66A065748C40dCA2C7Fe125A2028CF9982"
+                        },
+                        "contents": "Hello, Bob!"
+                    }
+                },
+                "message_hash": "0xea6529f0ee9eb0b207b5a8b0ebfa673d398d6a78262818da1d270bd138f81f03",
+                "show_message_hash": "0xea6529f0ee9eb0b207b5a8b0ebfa673d398d6a78262818da1d270bd138f81f03",
+                "domain_separator_hash": "0x97d6f53774b810fbda27e091c03c6a6d6815dd1270c2e62e82c6917c1eff774b"
+            },
+            "result": {
+                "address": "0x73d0385F4d8E00C5e6504C6030F47BF6212736A8",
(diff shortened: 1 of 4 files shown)


common/protob/messages-ethereum-eip712.proto
```diff
@@ -26,6 +26,7 @@ message EthereumSignTypedData {
     required string primary_type = 2;                          // name of the root message struct
     optional bool metamask_v4_compat = 3 [default=true];       // use MetaMask v4 (see https://github.com/MetaMask/eth-sig-util/issues/106)
     optional ethereum.EthereumDefinitions definitions = 4;     // network and/or token definitions
+    optional bytes message_hash = 5;                           // hash of the typed data to be signed (if set, device will show the message hash as part of signature confirmation)
```
mdehoog: Alternative could be `optional bool show_message_hash` and remove the equality check. Flow would be the same.

mdehoog: I ended up implementing this as it seems cleaner and more explicit

matejcik: this is fine, but let's call it `show_message_hash` to indicate the dual function.
(i worry that otherwise implementers will always provide it)

mdehoog: I called it `message_hash_confirmation` to match https://github.com/trezor/trezor-firmware/blob/main/common%2Fprotob%2Fmessages-bitcoin.proto... just confirming you'd prefer it be called `show_message_hash`? I don't feel strongly either way

matejcik:
hey @mdehoog did you misunderstand me there at [830efc9](https://github.com/trezor/trezor-firmware/pull/5344/commits/830efc91851746ea44cedb7f80baf13d0003c8d5) ? :)
i meant, "this is fine, to directly provide the bytes value of the hash for comparison" **but also** "let's call it a different name, so that it is clear(er) what happens when you **do** provide it"?

mdehoog:
> hey @mdehoog did you misunderstand me there at [830efc9](https://github.com/trezor/trezor-firmware/pull/5344/commits/830efc91851746ea44cedb7f80baf13d0003c8d5) ? :)
> i meant, "this is fine, to directly provide the bytes value of the hash for comparison" **but also** "let's call it a different name, so that it is clear(er) what happens when you provide it"?

Okay that's clear! Will make the change 👍

common/protob/messages-ethereum-eip712.proto
matejcik: that's not a good place to match against because `user_confirmation` is a specific name from the SLIP-19 spec. a more appropriate would be `show_display` from `GetAddress` 
_(though "show-display" is kind of like the anglo legal convention of using two terms that mean the same thing: "aiding & abetting" "assault & battery" "cease & desist" :) )_

my point is, keep it as bytes! having the hash readily available is a good property.
in this sense `show_message_hash` or `confirm_message_hash` read clearer: "show_(the following)_message_hash: 0xabcdef12465..."

matejcik (approved):
ACK from me
@obrusvit please assign someone to run the UI tests on all models and review the additional screen

obrusvit:
@mdehoog before merging, I kindly ask you to:
- [x] Provide some test vectors usable for [common/tests/fixtures/ethereum/sign_typed_data.json](https://github.com/trezor/trezor-firmware/blob/main/common/tests/fixtures/ethereum/sign_typed_data.json) which would set the new `show_message_hash` field
- [x] The test using the new fixtures will be run here https://github.com/trezor/trezor-firmware/blob/6cc9c629414e9cf8bc66b22580555c098e0fb3eb/tests/device_tests/ethereum/test_sign_typed_data.py#L31
- [x] The client side should have the new argument exposed, see here: https://github.com/trezor/trezor-firmware/blob/6cc9c629414e9cf8bc66b22580555c098e0fb3eb/python/src/trezorlib/ethereum.py#L311


After that, if you don't mind, I would take this PR over to:
- drop the merge commit (we prefer rebasing on `main` to preserve linear history)
- squash some of your commits
- reword to use conventional commit name, see https://docs.trezor.io/trezor-firmware/misc/contributing.html
- update `fixtures.json` because the new test will generate new screenshots.

Thank you.

mdehoog:
@obrusvit thanks for the assistance.
 - I've added a new case in `common/tests/fixtures/ethereum/sign_typed_data.json` that has `show_message_hash` set, and updated the client to accept this parameter
 - I added a new test case in `trezor-firmware/tests/device_tests/ethereum/test_sign_typed_data.py` that expects the correct error when the message hash doesn't match
 - I removed the merge commit and instead rebased on HEAD

Happy for you to take it from here! Let me know if I help in any way.

matejcik:
in addition to the fixtures, we also need a fail-test that passes a mismatched hash. this should be added to `tests/device_tests/ethereum/test_sign_typed_data.py` directly

mdehoog:
> in addition to the fixtures, we also need a fail-test that passes a mismatched hash. this should be added to `tests/device_tests/ethereum/test_sign_typed_data.py` directly

Agreed, I added this as part of the tests commit 👍

obrusvit:
Thank you @mdehoog.

I force pushed into the PR and approved the CI. I will finish the details.

obrusvit:
@mdehoog The functionality should be available in September release. You'll see it in the changelog.
Thank you.

PR #5345, opened 2025-07-12
erase storage on bootloader unlock

TychoVrahe:


STew790:
**QA OK**
Bootloader unlock indeed erases devices storage now.

PR #5347, opened 2025-07-14
adjust screen color according to charging state in prodtest

TychoVrahe:
Sets screen green when charging is done, red when discharging.

Applies to T3W1 prodtest homescreen only.


kopecdav:
Because of that charging state is directly linked to the charging current, screen behaves kind of funny now. If I plug the charger it goes from RED -> to GREEN -> to black() , but that is not an issue of this PR. We can merge this and i will link the charging_state to charging target instead in the temperature controller PR, that should fix it.

kopecdav (approved):

PR #5349, opened 2025-07-14
Rename  `raise_if_not_confirmed` to `raise_if_cancelled`

M1nd3r:
We do not care about the result of `interact` call, we let it raise if the action is cancelled. If the `interact` does not raise, any output (eg. `UiResult.CONFIRMED`, `UiResult.INFO`) is treated as valid - thus the renaming brings clarity to what this function actually does.

### DIFF
--- a/core/src/apps/management/ble/pair_new_device.py
+++ b/core/src/apps/management/ble/pair_new_device.py
@@ -3,7 +3,7 @@ import trezorui_api
 from storage import device as storage_device
 from trezor import utils
 from trezor.crypto import random
-from trezor.ui.layouts import CONFIRMED, raise_if_not_confirmed
+from trezor.ui.layouts import CONFIRMED, raise_if_cancelled
 from trezor.wire import ActionCancelled
 
 
@@ -35,7 +35,7 @@ async def pair_new_device() -> None:
     label = storage_device.get_label() or _default_ble_name()
     ble.start_advertising(False, label)
     try:
-        code = await raise_if_not_confirmed(
+        code = await raise_if_cancelled(
             trezorui_api.show_pairing_device_name(
                 device_name=label,
             ),
@@ -45,7 +45,7 @@ async def pair_new_device() -> None:
             raise ActionCancelled
 
         try:
-            result = await raise_if_not_confirmed(
+            result = await raise_if_cancelled(
                 trezorui_api.show_ble_pairing_code(
                     title="Bluetooth pairing",
                     description="Pairing code match?",
--- a/core/src/apps/management/ble/pair_new_device.py
+++ b/core/src/apps/management/ble/pair_new_device.py
@@ -3,7 +3,7 @@ import trezorui_api
 from storage import device as storage_device
 from trezor import utils
 from trezor.crypto import random
-from trezor.ui.layouts import CONFIRMED, raise_if_cancelled
+from trezor.ui.layouts import CONFIRMED, interact
 from trezor.wire import ActionCancelled
 
 
@@ -35,7 +35,7 @@ async def pair_new_device() -> None:
     label = storage_device.get_label() or _default_ble_name()
     ble.start_advertising(False, label)
     try:
-        code = await raise_if_cancelled(
+        code = await interact(
             trezorui_api.show_pairing_device_name(
                 device_name=label,
             ),
@@ -45,7 +45,7 @@ async def pair_new_device() -> None:
             raise ActionCancelled
 
         try:
-            result = await raise_if_cancelled(
+            result = await interact(
                 trezorui_api.show_ble_pairing_code(
                     title="Bluetooth pairing",
                     description="Pairing code match?",
--- a/core/src/trezor/ui/layouts/bolt/__init__.py
+++ b/core/src/trezor/ui/layouts/bolt/__init__.py
@@ -5,7 +5,7 @@ from trezor import TR, ui, utils
 from trezor.enums import ButtonRequestType, RecoveryType
 from trezor.wire import ActionCancelled
 
-from ..common import draw_simple, interact, raise_if_not_confirmed, with_info
+from ..common import draw_simple, interact, raise_if_cancelled, with_info
 
 if TYPE_CHECKING:
     from typing import Awaitable, Iterable, NoReturn, Sequence
@@ -42,7 +42,7 @@ def confirm_action(
     if description is not None and description_param is not None:
         description = description.format(description_param)
 
-    return raise_if_not_confirmed(
+    return raise_if_cancelled(
         trezorui_api.confirm_action(
             title=title,
             action=action,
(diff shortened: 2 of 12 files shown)


matejcik:
why?
that's certainly not the intention -- though i agree that the name suggests so
where specifically are you running into this?

M1nd3r:
We were discussing it with @romanz in the THP PR here:
https://github.com/trezor/trezor-firmware/pull/4976#discussion_r2204349536

romanz:
Since `await raise_if_not_confirmed()` returns None, it may cause us to ignore other returned values (e.g. `INFO`).

~Maybe it's better to use an `assert action is trezorui_api.CONFIRMED`?~

EDIT:
I forgot that `action` is a coroutine, so we'd need to check the resulting `value`...

romanz:
WDYT about:
```diff
diff --git a/core/src/trezor/ui/layouts/common.py b/core/src/trezor/ui/layouts/common.py
index 66224f038f..9c22b3a67f 100644
--- a/core/src/trezor/ui/layouts/common.py
+++ b/core/src/trezor/ui/layouts/common.py
@@ -40,6 +40,7 @@ async def interact(
     br_name: str | None,
     br_code: ButtonRequestType = ButtonRequestType.Other,
     raise_on_cancel: ExceptionType | None = ActionCancelled,
+    raise_if_not_confirmed: bool = False,
 ) -> T:
     # shut down other workflows to prevent them from interfering with the current one
     workflow.close_others()
@@ -54,6 +55,8 @@ async def interact(
     # raise an exception if the user cancelled the action
     if raise_on_cancel is not None and result is trezorui_api.CANCELLED:
         raise raise_on_cancel
+    if raise_if_not_confirmed and result is not trezorui_api.CONFIRMED:
+        raise ValueError(result)
     return result
 
 
@@ -63,7 +66,7 @@ def raise_if_not_confirmed(
     br_code: ButtonRequestType = ButtonRequestType.Other,
     exc: ExceptionType = ActionCancelled,
 ) -> Coroutine[Any, Any, None]:
-    action = interact(layout_obj, br_name, br_code, exc)
+    action = interact(layout_obj, br_name, br_code, exc, raise_if_not_confirmed=True)
     return action  # type: ignore ["UiResult" is incompatible with "None"]
```

matejcik:
the intention is to use `raise_if_not_confirmed` only for layouts where you don't care about the result as long as it's not `CANCELLED`

if you want to look at the value and e.g. differentiated between `CONFIRMED` and `INFO`, you should use `interact()`. you _can_ use interact's `raise_on_cancel` to handle the `CANCELLED` case for you. you should _not_ extend `interact` to handle any other cases; i don't see how that makes sense, if you care about the value, just examine the value.

so again, is there a case where you want to raise (the same exception) when the result is `CANCELLED` but also `INFO` or any other value?

(i'd be ok with renaming `raise_if_not_confirmed` to `raise_if_cancelled`, if that's what you want to do here; provided that the other UI guys agree)

M1nd3r:
> (i'd be ok with renaming `raise_if_not_confirmed` to `raise_if_cancelled`, if that's what you want to do here; provided that the other UI guys agree)

I think that if the idea is to raise **only** on `Cancel`, then there is no reason for the check as proposed in this PR. 
If that is the case, I agree that the best course of action is to rename `raise_if_not_confirmed` to `raise_if_cancelled`.

I will ask in Slack if there is someone against this change.

matejcik:
but again, the point of `raise_if_not_confirmed` is to indicate that **you do not care about return value**.

i still feel like you're not supposed to be using this function in the first place, and use `interact()` instead, if there's a risk of getting one of two different retvals

romanz:
> I agree that the best course of action is to rename `raise_if_not_confirmed` to `raise_if_cancelled`.

Sounds good.

M1nd3r:
> but again, the point of raise_if_not_confirmed is to indicate that you do not care about return value.
> 
> i still feel like you're not supposed to be using this function in the first place, and use interact() instead, if there's a risk of getting one of two different retvals

Is this an argument against the renaming, or just a general note?  I agree that you should use this **only** if you do not care about the value. However, if there are **not** two possible retvals, but the expected retval is **different** from `UiResult.CONFIRMED`, this function can be used too.

core/src/apps/management/ble/pair_new_device.py
```diff
@@ -35,7 +35,7 @@ async def pair_new_device() -> None:
     label = storage_device.get_label() or _default_ble_name()
     ble.start_advertising(False, label)
     try:
-        code = await raise_if_not_confirmed(
+        code = await raise_if_cancelled(
```
romanz: We should probably use `interact` here (since `await raise_if_cancelled` type is `None`).

core/src/apps/management/ble/pair_new_device.py
```diff
@@ -45,7 +45,7 @@ async def pair_new_device() -> None:
             raise ActionCancelled
 
         try:
```
romanz, line 47: Can it be rewritten to:
```python
        try:
            await raise_if_cancelled(
                trezorui_api.show_ble_pairing_code(
                    title="Bluetooth pairing",
                    description="Pairing code match?",
                    code=f"{code:0>6}",
                ),
                None,
            )            
            ble.allow_pairing(code)
        except ActionCancelled:
            ble.reject_pairing()
            raise
```

M1nd3r: I have two questions:

1. Can there possibly be other result than `CONFIRMED`?
2. Can other exception than `ActionCancelled` be raised there?

I do not think that changing this behavior fits in the scope of this PR. 

I have replaced the `raise_if_cancelled` by `interact` for now: https://github.com/trezor/trezor-firmware/pull/5349/commits/c9ac673ce1dc5fbc79b1b528c8c098244ebd9516

core/src/apps/management/ble/pair_new_device.py
M1nd3r: fixed here: https://github.com/trezor/trezor-firmware/pull/5349/commits/c9ac673ce1dc5fbc79b1b528c8c098244ebd9516

core/src/apps/management/ble/pair_new_device.py
romanz: > I do not think that changing this behavior fits in the scope of this PR.

Agreed.

matejcik:
> Is this an argument against the renaming, or just a general note? I agree that you should use this **only** if you do not care about the value. However, if there are **not** two possible retvals, but the expected retval is **different** from `UiResult.CONFIRMED`, this function can be used too.

it's a general note.
i am not against renaming.

what I smell as a possible problem here is, usually you should **not** care about the distinction. and you should **not** be in a situation where the distinction matters.
(and if you're not in that situation, the renaming is kind of pointless too ;) )

PR #5355, opened 2025-07-15
fixup! test(core): fix `test_emu_sanity` on THP emulators

romanz:
Don't use `trezorctl` for sanity test, since it requires pairing confirmation.

Should be squashed in fd6135445859c7c315f598bc89a3046e26d6b6bc.

image


M1nd3r (approved):
LGTM

PR #5360, opened 2025-07-15
feat(core): use menu for `confirm_value`-related info items on Caesar

romanz:
Also, move Solana `blockhash` to the first confirmation menu on all layouts.

### DIFF
--- a/core/src/trezor/ui/layouts/caesar/__init__.py
+++ b/core/src/trezor/ui/layouts/caesar/__init__.py
@@ -8,7 +8,7 @@ from trezor.wire import ActionCancelled
 from ..common import draw_simple, interact, raise_if_cancelled
 
 if TYPE_CHECKING:
-    from typing import Awaitable, Iterable, NoReturn, Sequence
+    from typing import Awaitable, Callable, Iterable, NoReturn, Sequence
 
     from ..common import ExceptionType, PropertyType
     from ..menu import Details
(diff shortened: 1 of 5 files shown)


core/src/trezor/ui/layouts/caesar/__init__.py
```diff
@@ -862,47 +862,18 @@ async def confirm_value(
...
-                await interact(
-                    trezorui_api.confirm_value(
-                        title=info_title,
-                        value=info_value,
-                        description=description,
-                        verb="",
-                        verb_cancel="<",
-                        hold=False,
-                        is_data=is_data,
-                        chunkify=chunkify_info,
```
romanz, line 898: I forgot to pass `chunkify_info` to the new menu:
image
https://data.trezor.io/dev/firmware/ui_report/16295135943/diff/T3B1_en-device_tests-ethereum-test_signtx.py%3A%3Atest_signtx_staking%5BTrue-stake_holesky%5D.html

romanz: Resolved in https://github.com/trezor/trezor-firmware/pull/5360/commits/01c0980097dc4d0fabd201ea8610b9aad1807c69.

ibz:
Actually this seems like a slight regression IMO... It's not too bad, but slightly worse UX.

Perhaps the solution to this would be (we can do it later, probably once we have menus all over the place and tests that visit all menus):

If a menu detects that it just has one item, it should not show the menu, but directly that item?

image

romanz:
> If a menu detects that it just has one item, it should not show the menu, but directly that item?

Sounds good to me :)

@Hannsek @lapohoda WDYT?

lapohoda:
It's fine. The only thing I'd change here is the initial visual cue for the user:

image

Hannsek:
IMO it should show menu even though there is just one item. Let's not unnecessary complicate things.

ibz (approved):

PR #5365, opened 2025-07-15
Battery model dependant SoC target 

kopecdav:
This PR reworks the original power manager SoC limit to use a battery model-based SoC target.

Previous implementation compared the SoC limit directly with the SoC estimate. This approach could be unreliable, especially during the device's initial startup, if the SoC estimate was inaccurate. As a result, devices could leave production with varying precharge voltages.

The new implementation translates the SoC target into a charging voltage using a temperature-dependent battery model and then it compares this voltage threshold with the actual battery voltage. Charging is stopped if the battery terminal voltage is above the threshold, and more over the SoC estimate is set to the SoC target if the threshold is crossed. This method is more precise and should improve the accuracy of initial battery SoC estimates during production.

Additionally, this PR removes the obsolete pm-precharge command from prodtest.


core/embed/sys/power_manager/stm32u5/power_manager_internal.h
```diff
@@ -35,7 +35,7 @@
 #define PM_BATTERY_UNDERVOLT_RECOVERY_THR_V 3.1f
 #define PM_BATTERY_UNDERVOLT_RECOVERY_WPC_THR_V 3.2f
 #define PM_BATTERY_LOW_THRESHOLD_SOC 15
-#define PM_SOC_LIMIT_HYSTERESIS 5
+#define PM_SOC_TARGET_HYSTERESIS 5
```
TychoVrahe: no longer needed

kopecdav: Fixed here 51b6783d4c8707757645d94d6649d387c8510c15

TychoVrahe (approved):

PR #5366, opened 2025-07-15
Update pm charging state evaluation.

kopecdav:
Power manager reported charging status directly based on the floating current measured by pmic. Since the pmic cannot change the charging current during the active charging cycle, every time the charging controller decided to change the charging current (maybe due to temperature limitations or any other reason) charging state goes from CHARGING to IDLE and then back to CHARGING. App would probably propagates this to a screen leaving impression that charging got interrupted. 

This PR links the charging status to charging_current_target_ma variable which is more relevant and should be stable even if charging controller change the charging setup.


kopecdav:
Fixed in https://github.com/trezor/trezor-firmware/pull/5692, closing

PR #5371, opened 2025-07-16
ci: add UI-specific diff markers for PR comment

romanz:
ui-diffs

Tested by https://github.com/trezor/trezor-firmware/pull/5373.


Hannsek:
So, the yellow icon will be use if there are any diffs and red icon will be used if the test crashed/didn't finish?

romanz:
Yes.

If there are both UI diffs and test failures - the red icon will be used.

Hannsek:
Hmm, then you should go through all reds all the time. Can we also display yellow, whenever there are diffs. So, if there are both UI diffs and test failures, both icons (red, yellow) would be displayed?

romanz:
I think it would be simpler to use a different icon background in case there are UI diffs - since the [icon "placeholder" is generated](https://github.com/trezor/trezor-firmware/blob/30e5f5b5dbbcf95b5aa707a5ef17f2815c42f6b6/ci/make_pull_comment.py#L48) before the tests start (so we don't know in advance if we'll need an additional "UI diff icon"):

image

(from https://github.com/trezor/trezor-firmware/pull/5373#issuecomment-3079640909)

So the results with "yellow background" are the ones with UI diffs.

Hannsek:
Perfect

romanz:
Will add a marker for `main` UI diffs as well.

romanz:
@obrusvit @Hannsek WDYT about keeping the job status square icon as-is, and adding two "UI diff" icons (the circles):

image

This way, we can quickly see if there are any UI diffs found by tests (the first round icon) or there are any UI diffs from `main` (the second one).

Hannsek:
We can do that.

romanz:
Updated this PR to use UI-specific diff markers:
https://github.com/trezor/trezor-firmware/pull/5371#issuecomment-3078008867

Tested in https://github.com/trezor/trezor-firmware/pull/5373:
image

(Please note that when the PR is not rebased over `main`, there may be `main` UI diffs detected - due to `main` UI fixtures updates)

obrusvit:
Very nice, thanks!

PR #5372, opened 2025-07-16
save flash in by not using HAL for oscillator configuration

TychoVrahe:
This saves about 3kB in bootloader.


core/embed/io/rgb_led/stm32u5/rgb_led_lp.c
```diff
@@ -64,12 +66,16 @@ void rgb_led_init(void) {
...
-  RCC_OscInitStruct.OscillatorType = RCC_OSCILLATORTYPE_HSI;
-  RCC_OscInitStruct.HSIState = RCC_HSI_ON;
-  RCC_OscInitStruct.PLL.PLLState = RCC_PLL_NONE;
-  HAL_RCC_OscConfig(&RCC_OscInitStruct);
+  uint32_t deadline = ticks_timeout(HSI_TIMEOUT_VALUE);
+
+  // enable HSI clock
+  RCC->CR |= RCC_CR_HSION;
+  // wait until the HSI is on
+  while ((RCC->CR & RCC_CR_HSION) != RCC_CR_HSION) {
```
cepetr: Should be `RCC_CR_HSIRDY`

core/embed/sec/secure_aes/stm32u5/secure_aes.c
```diff
@@ -221,13 +222,16 @@ secbool secure_aes_ecb_decrypt_hw(const uint8_t* input, size_t size,
...
-  RCC_OscInitTypeDef osc_init_def = {0};
-  osc_init_def.OscillatorType = RCC_OSCILLATORTYPE_SHSI;
-  osc_init_def.SHSIState = RCC_SHSI_ON;
-
   // Enable SHSI clock
-  if (HAL_RCC_OscConfig(&osc_init_def) != HAL_OK) {
-    goto cleanup;
+  __HAL_RCC_SHSI_ENABLE();
+
+  uint32_t deadline = ticks_timeout(HSI_TIMEOUT_VALUE);
```
cepetr, line 228: Not important, but we may introduce `SHSI_TIMEOUT_VALUE`,

cepetr (changes requested):

core/embed/io/rgb_led/stm32u5/rgb_led_lp.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/7179edaad7c9d5b6595e9619b7d1fdad07eb68d6

core/embed/sec/secure_aes/stm32u5/secure_aes.c
TychoVrahe: it actually already exists, but its private to the HAL RCC module, so i used the HSI_TIMEOUT_VALUE which has same value.

PR #5374, opened 2025-07-16
save flash by not using HAL for periph clock configuration

TychoVrahe:
Few drivers were using this - display, usb, led


cepetr (approved):
Tested on T3W1

PR #5378, opened 2025-07-17
fix(core): fix MPU configuration

cepetr:
This PR fixes an issue that causes a crash on STM32U5 devices (T3B1, T3T1, and T3W1) when the firmware is downloaded and run for the first time.

The crash results in a RSOD, but it does not leave the device in an invalid state. After a power cycle (powering the device off and on), the device operates normally.

The issue is caused by incorrect MPU configuration for a specific sequence of FLASH memory writes that occurs immediately after the device is run for the first time. For more details, see the added comment in the source code.

### DIFF
--- a/core/embed/sys/mpu/stm32u5/mpu.c
+++ b/core/embed/sys/mpu/stm32u5/mpu.c
@@ -131,20 +131,32 @@ static void mpu_set_attributes(void) {
 _Static_assert(NORCOW_SECTOR_SIZE == STORAGE_1_MAXSIZE, "norcow misconfigured");
 _Static_assert(NORCOW_SECTOR_SIZE == STORAGE_2_MAXSIZE, "norcow misconfigured");
 
-// PERIPH_SIZE covers secure peripherals only (+16MB of FMC1)
-// PERIPH_SIZE_EXT covers both secure and non-secure peripherals (+16MB of FMC1)
-// The extended size is used in a special case - MPU_MODE_OTP - when access
-// to non-secure FLASH controller registers is required.
+// PERIPH_SIZE covers both secure and non-secure peripherals
+// 0x40000000 to 0x4FFFFFFF (256M) and
+// 0x50000000 to 0x5FFFFFFF (256M).
+
+// When writing to OTP memory while running in secure mode, we have to access
+// the non-secure FLASH peripheral. Additionally, the ST HAL requires access
+// to the same non-secure peripheral during the *first* write or erase
+// operation after writing to OTP. This applies even if the following operation
+// targets a different flash region. To avoid faults, we must ensure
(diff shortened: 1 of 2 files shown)


core/embed/sys/mpu/stm32u5/mpu.c
```diff
@@ -131,20 +131,32 @@ static void mpu_set_attributes(void) {
...
-#ifdef STM32U585xx
-// On STM32U585, we need to add an additional 16M for FMC1 which
+// PERIPH_SIZE covers both secure and non-secure peripherals
+// 0x40000000 to 0x4FFFFFFF (256M) and
+// 0x50000000 to 0x5FFFFFFF (256M).
+
+// When writing to OTP memory while running in secure mode, we have to access
+// the non-secure FLASH peripheral. Additionally, the ST HAL requires access
+// to the same non-secure peripheral during the *first* write or erase
+// peration after writing to OTP. This applies even if the following operation
```
TychoVrahe: typo operation

core/embed/sys/mpu/stm32u5/mpu.c
```diff
@@ -131,20 +131,32 @@ static void mpu_set_attributes(void) {
...
+// Moreover, on STM32U585, we need to add an additional 16M for FMC1 which
 // follows the peripherals in the memory map.
+// 0x60000000 to 0x60FFFFFF (16M).
 
-#define PERIPH_SIZE (SIZE_256M + SIZE_16M)
-#define PERIPH_SIZE_EXT (SIZE_512M + SIZE_16M)
+// In the kernel, on models with a secure monitor (SECURE_MODE is not defined),
+// we can allow access *only* to the non-secure peripherals region.
+
+#ifdef STM32U585xx
```
TychoVrahe, line 152: it doesn't matter now but consider reversing order of the ifdefs: decide the MCU type insude the SECURE_MODE if, that way, it will be compatible when we enable non-secure mode on U585

TychoVrahe:
tested on T3T1

core/embed/sys/mpu/stm32u5/mpu.c
cepetr: fixed, 4b10400d3bc0cafa208ccb5a447c5765b0b93143

core/embed/sys/mpu/stm32u5/mpu.c
cepetr: I don't think we can reverse these ifdefs. On U585 we cannot have two variants like on U5G because of additional 16M of FMC1 which is adjacent to secure peripheral region.

TychoVrahe: ah ok, didn't realize that

TychoVrahe (approved):

PR #5381, opened 2025-07-18
ci(core): skip UI comment job on external PRs

obrusvit:
- the comment action fails when run on a PR created from a fork
- see https://github.com/peter-evans/create-or-update-comment/blob/9fa92b4fea95714d51d64d8e90d8817fe499e1cb/README.md

[no changelog]

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -697,7 +697,7 @@ jobs:
 
   core_ui_comment:
     name: Post comment with UI diff URLs
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     needs:
       - param


.github/workflows/core.yml
```diff
@@ -697,7 +697,7 @@ jobs:
 
   core_ui_comment:
     name: Post comment with UI diff URLs
-    if: github.event_name == 'pull_request'
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
romanz: Will it work on non-PR events (e.g. nightly/manual runs)?
```suggestion
    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
```

obrusvit: I thought that it would be implicit because the `pull_request` context would be supplied only on PRs events but you're right.

Also `pull_request` may be `null` or something so the check should prevent it.

https://github.com/trezor/trezor-firmware/pull/5381/commits/36ccf1cda2d5ab25c910756fc50b307810177111

PR #5382, opened 2025-07-18
improve Eckhart boot warning screen

TychoVrahe:
not entirely by figma design but usable

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
@@ -1,5 +1,5 @@
 use crate::ui::{
-    component::{Event, Label},
+    component::{base::Component, text::TextStyle, Event, Label},
     display::{self, toif::Toif, Color},
     geometry::{Alignment, Alignment2D, Offset, Point, Rect},
     layout::simplified::{process_frame_event, render, run, show},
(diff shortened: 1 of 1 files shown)


core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
```diff
@@ -428,10 +428,27 @@ impl BootloaderUI for UIEckhart {
         vendor_img: &'static [u8],
         wait: i32,
     ) {
-        let bg_color = if warning { RED } else { WELCOME_COLOR };
+        let bg_color = if warning {
+            Color::rgb(0xFF, 0, 0)
```
obrusvit, line 432: Based on Figma, there should be `0xFF, 0x30, 0x30`, i.e. `RED` as was before.

core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
```diff
@@ -428,10 +428,27 @@ impl BootloaderUI for UIEckhart {
...
+            WELCOME_COLOR
+        };
 
         display::sync();
 
+        let pos = Point::new(0, 200);
+
+        let label_text = vendor_str.unwrap_or("");
+
+        const TEXT_WHITE: TextStyle =
```
obrusvit, line 443: It should be `GREY_EXTRA_LIGHT`

obrusvit (commented):
Minor comments. As you wrote, it's not entirely by Figma, e.g. the icon is missing but if you want to merge it sooner, I can fix some details later.

core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
TychoVrahe: The (warning) icon is not actually an icon but vendor image, which is taken from the vendor header. And i did not want to change the vendor image because of this.

The vendor image is RGB565 toif with backgound color 0xF800, which, when converted to RGB888 is 0xFF0000, so i can't use `RED` as that results in ugly different background color different than the image red background.

core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
TychoVrahe: For similar reason, i wanted to use WHITE as it matches the color of the vendor image. Grey looks weird as its different

core/embed/rust/src/ui/layout_eckhart/ui_bootloader.rs
obrusvit: Thanks for the explanation.

obrusvit (approved):

PR #5383, opened 2025-07-18
fix(core/protobuf): broken wire_id separation by enum

mmilata:
There was a bug in pb2py blob generation that resulted in every combination of wire_id x wire_enum to be included.

Related to #5221.


M1nd3r (approved):
LGTM

PR #5384, opened 2025-07-18
nrf pairing

TychoVrahe:
This PR adds support for basic pairing of main MCU and nRF. 

Symmetric keys is exchanged in prodtest and stored on nRF. Trezor derives the key from master secret key.

### DIFF
--- a/core/embed/io/nrf/stm32u5/nrf_test.c
+++ b/core/embed/io/nrf/stm32u5/nrf_test.c
@@ -24,23 +24,30 @@
 #include <trezor_rtl.h>
 
 #include <io/nrf.h>
+#include <sec/secret_keys.h>
 #include <sys/systick.h>
 
 #include "../nrf_internal.h"
+#include "sys/irq.h"
 
 typedef enum {
   PRODTEST_CMD_SPI_DATA = 0x00,
   PRODTEST_CMD_UART_DATA = 0x01,
   PRODTEST_CMD_SET_OUTPUT = 0x02,
+  PRODTEST_CMD_PAIR = 0x03,
 } prodtest_cmd_t;
 
 typedef enum {
   PRODTEST_RESP_SPI = 0x00,
   PRODTEST_RESP_UART = 0x01,
+  PRODTEST_RESP_SUCCESS = 0x02,
+  PRODTEST_RESP_FAILURE = 0x03,
 } prodtest_resp_t;
 
 typedef struct {
   bool answered_spi;
+  bool success;
+  bool failure;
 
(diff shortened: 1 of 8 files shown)


TychoVrahe:
@andrewkozlik in successive PR i added another prodtest command for verification of the paring, which includes the HMAC challenge/response. Take a look at the HMAC usege there if you can: https://github.com/trezor/trezor-firmware/pull/5390

andrewkozlik (approved):
Approving the core/prodtest and secret_keys changes.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -59,7 +59,7 @@ secbool set_random_secret(uint8_t slot, size_t length) {
 
   secbool ret = secfalse;
 
-  if (secret_key_writable(slot) != sectrue) {
+  if (secret_key_writable(slot) == sectrue) {
```
cepetr: Why not stay consistent with the following checks and write it like this:
```c
if (secret_key_writable(slot) != sectrue) {
  ret = secfalse;  // <-- maybe we don't need this at all
  goto cleanup;
}
```

core/embed/io/nrf/stm32u5/nrf_test.c
```diff
@@ -212,4 +224,36 @@ bool nrf_test_gpio_reserved(void) {
...
+  }
+
+  if (!nrf_send_msg(NRF_SERVICE_PRODTEST, data, sizeof(data), NULL, NULL)) {
+    return false;
+  }
+
+  uint32_t timeout = ticks_timeout(100);
+
+  while (!ticks_expired(timeout)) {
+    if (g_nrf_test.success) {
```
cepetr: There's some kind of memory barrier missing - at least an irq_lock()/irq_unlock()

cepetr (changes requested):

core/embed/projects/prodtest/cmd/prodtest_secrets.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/22dd35f3fe9bfa949c70719f92608fdb0070d355

core/embed/io/nrf/stm32u5/nrf_test.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/a68dfbd98434f1d4d39015da3b1c1bb2bd6b9a62

PR #5388, opened 2025-07-19
bootloader: send attach events to layout after creation

TychoVrahe:
This PR ensures that attach events are sent to layout at the beginning of their life.

This solves an issue with using fuel gauge in bootloader, which relies on that behavior - until some change occurred, it displayed unknown battery state.

(this could also be solved by modifying the fuel gauge which may be simplified - but generally speaking sending attach events may be expected by components)

### DIFF
--- a/core/embed/rust/src/ui/api/bootloader_c.rs
+++ b/core/embed/rust/src/ui/api/bootloader_c.rs
@@ -12,6 +12,17 @@ use crate::{
 };
 
 #[no_mangle]
+extern "C" fn screen_attach(layout: *mut c_layout_t) -> u32 {
+    // SAFETY: calling code is supposed to give us exclusive access to an already
+    // initialized layout
+    unsafe {
+        let mut layout = LayoutBuffer::<<ModelUI as BootloaderUI>::CLayoutType>::new(layout);
+        let layout = layout.get_mut();
+        layout.show()
+    }
+}
+
+#[no_mangle]
 extern "C" fn screen_event(layout: *mut c_layout_t, signalled: &sysevents_t) -> u32 {
     let e = parse_event(signalled);
     // SAFETY: calling code is supposed to give us exclusive access to an already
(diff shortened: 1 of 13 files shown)


core/embed/rust/src/ui/api/bootloader_c.rs
```diff
@@ -23,6 +23,17 @@ extern "C" fn screen_event(layout: *mut c_layout_t, signalled: &sysevents_t) ->
     }
 }
 
+#[no_mangle]
+extern "C" fn screen_attach(layout: *mut c_layout_t) -> u32 {
```
obrusvit: Nit: move this function above `screen_event` so that the order is the same as in `rust_ui_bootloader.h`

core/embed/rust/src/ui/layout/simplified.rs
```diff
@@ -225,9 +235,18 @@ pub fn run(frame: &mut impl Component<Msg = impl ReturnToC>) -> u32 {
...
+pub fn show(frame: &mut impl Component<Msg = impl ReturnToC>, fading: bool) -> u32 {
     frame.place(ModelUI::SCREEN);
 
+    let e = Event::Attach(AttachType::Initial);
+    let mut ctx = EventCtx::new();
+
+    let msg = frame.event(&mut ctx, e);
+
+    if let Some(message) = msg {
+        return message.return_to_c();
```
obrusvit, line 247: Not sure if this can be triggered by any `Component` 🤔 and IIUC, it then returns from `workflow_host_control` so any further polling event is then skipped? What happens next?

obrusvit (commented):
Tested on the device, looks good. 
One question

core/embed/rust/src/ui/api/bootloader_c.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/d4c965c9ebd2dc982c08152af504f71a2a73b390

core/embed/rust/src/ui/layout/simplified.rs
TychoVrahe: intention was to keep this the same as firmware, which also returns message in `ui_layout_attach_timer_fn` after sending the attach event (if something is return by the layout).

Probably no component does that though.

In case it would return, the flow ends and nothing is rendered, execution returns to the caller flow as if we would return on normal event.

obrusvit (approved):

PR #5390, opened 2025-07-20
nrf pairing verification

TychoVrahe:
This PR implements nrf authentication/pairing key verification.

After challenge is issued by main MCU, nRF calculates HMAC and sends the output back, and the verification is done in secure monitor so that the key does not leak to less protected parts. 

So far only verification in prodtest is implemented, to be able to do this from firmware we will also need to add an SMCALL and SYSCALL.

### DIFF
--- a/core/embed/projects/prodtest/README.md
+++ b/core/embed/projects/prodtest/README.md
@@ -337,6 +337,14 @@ nrf-pair
 OK
 ```
 
+### nrf-verify-pairing
+Verifies the pairing between the main MCU and the nRF chip. The command returns `OK` if the pairing is valid, or `ERROR` if it is not.
+Example:
+```
+nrf-verify-pairing
+OK
+```
+
 
 ### touch-draw
 Starts a drawing canvas, where user can draw with finger on pen. Canvas is exited by sending CTRL+C command.
--- a/core/embed/projects/prodtest/cmd/prodtest_nrf.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_nrf.c
@@ -198,6 +198,19 @@ static void prodtest_nrf_pair(cli_t* cli) {
   }
 }
 
+static void prodtest_nrf_verify_pairing(cli_t* cli) {
+  if (cli_arg_count(cli) > 0) {
+    cli_error_arg_count(cli);
+    return;
+  }
+
+  if (nrf_authenticate()) {
+    cli_ok(cli, "");
+  } else {
+    cli_error(cli, CLI_ERROR, "Pairing verification failed.");
+  }
+}
+
 // clang-format off
 
 PRODTEST_CLI_CMD(
@@ -225,6 +238,13 @@ PRODTEST_CLI_CMD(
   .name = "nrf-pair",
   .func = prodtest_nrf_pair,
   .info = "Pair nRF chip",
+  .args = ""
+  );
+
+PRODTEST_CLI_CMD(
+  .name = "nrf-verify-pairing",
+  .func = prodtest_nrf_verify_pairing,
+  .info = "Verify nRF pairing",
   .args = ""
 );
 


core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -509,4 +515,47 @@ bool nrf_system_off(void) {
...
+  nrf_driver_t *drv = &g_nrf_driver;
+  if (!drv->initialized) {
+    return false;
+  }
+
+  drv->info_valid = false;
+
+  uint8_t data[1 + 32] = {MGMT_CMD_CHALLENGE};
+
+  uint32_t challenge[8] = {0};
```
andrewkozlik: Let's not use magic constants.
32 = SHA256_DIGEST_LENGTH
8 = SHA256_DIGEST_LENGTH/sizeof(uint32_t)
or if we switch these two lines then 32 = sizeof(challenge)

TychoVrahe: switched the lines in https://github.com/trezor/trezor-firmware/commit/7cbd40708e52b893c9e61095473b1ffa112e4f96

i could also add the `8 = SHA256_DIGEST_LENGTH/sizeof(uint32_t)` but i didn't think that challenge needs to match `SHA256_DIGEST_LENGTH`, does it? if it does not i'd leave 8 as its just arbitrary number (or we can name the constant of course)

andrewkozlik: You're right, the challenge doesn't have to be SHA256_DIGEST_LENGTH :man_facepalming:. That's exactly why magic constants are so deceptive.

core/embed/sec/secret/stm32u5/secret_keys.c
```diff
@@ -164,6 +164,32 @@ secbool secret_key_nrf_pairing(uint8_t dest[NRF_PAIRING_SECRET_SIZE]) {
...
+
+  if (mac_len != SHA256_DIGEST_LENGTH) {
+    goto cleanup;
+  }
+
+  uint8_t dest[SHA256_DIGEST_LENGTH] = {0};
+
+  hmac_sha256(key, sizeof(key), message, msg_len, dest);
+
+  if (memcmp(dest, mac, SHA256_DIGEST_LENGTH) == 0) {
```
andrewkozlik: It is preferable to compare secret values in constant time, see for example `secequal()` in storage (which also some fault injection check that we don't need here) or in Python we have `utils.consteq()`. In this case it doesn't matter that much, since the challenge/message is single-use, but this function doesn't guarantee that. If there was a way to query this function repeatedly, then the correct mac could be brute-forced in ~4096 queries.

andrewkozlik: Also, let's memzero dest, just to be on the safe side.

nordic/trezor/trezor-ble/src/prodtest/prodtest.c
```diff
@@ -66,6 +65,8 @@ static int prodtest_set(const char *key, size_t len, settings_read_cb read_cb,
 SETTINGS_STATIC_HANDLER_DEFINE(prodtest, "prodtest", NULL, prodtest_set, NULL,
                                NULL);
 
+uint8_t *prodtest_get_pairing_key(void) { return pairing_secret; }
```
andrewkozlik: return value should be const

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/e39709111b3598f6b45908aebe7d510e2af8dd13

core/embed/sec/secret/stm32u5/secret_keys.c
TychoVrahe: added copy without the FIH into secret (also including the memzero) https://github.com/trezor/trezor-firmware/commit/4082a5bd5eb748497f255aa50c5b965f26eff70e

(we should probably include such function in some shared file, but i don't see an obvious place and don't want to handle that now)

andrewkozlik: I totally agree about putting the function in a shared file.

core/embed/sec/secret/inc/sec/secret_keys.h
```diff
@@ -61,6 +61,9 @@ secbool secret_key_tropic_masking(uint8_t dest[ECDSA_PRIVATE_KEY_SIZE]);
 #define NRF_PAIRING_SECRET_SIZE 32
 secbool secret_key_nrf_pairing(uint8_t dest[NRF_PAIRING_SECRET_SIZE]);
 
+secbool secret_validate_nrf_pairing(const uint8_t *message, size_t msg_len,
+                                    const uint8_t *mac, size_t mac_len);
```
andrewkozlik, line 65: If we expect a constant size `mac`, then I think it's clearer to specify it in the API `const uint8_t mac[SHA256_DIGEST_LENGTH]` and then get rid of `mac_len`. Just my opinion, but we use this often. Unfortunately the C standard doesn't require the compiler to check the sizes, but compilers usually issue a warning when the input is an array which is smaller than specified.

andrewkozlik (approved):
`secret_keys` and `prodtest` code LGTM, leaving the nRF to @cepetr

nordic/trezor/trezor-ble/src/management/management.c
```diff
@@ -39,17 +42,21 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
...
+#define CHALLENGE_MIN_LEN 32
+
 typedef enum {
   MGMT_CMD_SYSTEM_OFF = 0x00,
   MGMT_CMD_INFO = 0x01,
   MGMT_CMD_START_UART = 0x02,
   MGMT_CMD_STOP_UART = 0x03,
   MGMT_CMD_SUSPEND = 0x04,
   MGMT_CMD_RESUME = 0x05,
+  MGMT_CMD_CHALLENGE = 0x06,
```
cepetr: Couldn't we call the pair of messages as?
`MGMT_CMD_AUTH_CHALLENGE`
`MGMT_RESP_AUTH_RESPONSE`

nordic/trezor/trezor-ble/src/management/management.c
```diff
@@ -152,6 +159,17 @@ static void send_info(void) {
   trz_comm_send_msg(NRF_SERVICE_MANAGEMENT, data, sizeof(data));
 }
 
+static void mgmt_process_challenge(uint8_t *data, uint16_t len) {
+  uint8_t response[1 + SHA256_BLOCK_LENGTH];
+  const uint8_t *key = prodtest_get_pairing_key();
```
cepetr, line 164: Maybe I am missing something, but shouldn't we read pairing key from the storage here instead of calling `prodtest_get_paring_get()`?

cepetr (changes requested):

nordic/trezor/trezor-ble/src/management/management.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/b52491a337884797530bb1740a8ad7fd3209f7a9 and https://github.com/trezor/trezor-firmware/commit/e9e2bb5decf19e40db9a3501d34b0b7778d15337

nordic/trezor/trezor-ble/src/management/management.c
TychoVrahe: Afaik there is no direct way to read from settings. 

The way settings work here is that you register a callback via SETTINGS_STATIC_HANDLER_DEFINE (which i am doing in prodtest.c) and this callback is called when settings_load is called (which is in ble.c). The callback then populate the global variable and this getter accesses it.

cepetr (approved):
utACK

PR #5394, opened 2025-07-21
Refactor storage and entropy sys/sm-calls

cepetr:
This PR includes several refactorings related to the storage and entropy modules.

- It simplifies storage initialization and completely removes entropy-related smcalls and syscalls.
- A new `sec/storage` module has been introduced, with an initialization wrapper called `storage_setup()`.
- The storage API should now be included with `#include <sec/storage.h>` (instead of the original `#include "storage"`).
- The entropy module has been relocated to `sec/storage`, renamed to `storage_salt`, and its API has been simplified.
- Circular dependencies between `storage/flash_area` and the firmware have been partially removed.
    - The flash API is now fully accessible via `#include <util/flash.h>` (instead of the previous mix of "flash_area",  <util/flash.h> and <trezor_model.h>).
- New `secure_mode` and `kernel_mode` features have been introduced (not yet propagated to Rust).
- SConscripts file have been slightly simplified -  storage files are now consolidated in one location, and a new `storage` feature has been added

This refactoring also makes it possible (and was the original goal) to initialize storage early in the kernel and access it before the core app starts - but this has not yet been enabled.


STew790:
**QA OK**
Tested update on all core devices and all went ok.

PR #5253, opened 2025-06-25
Introduce SLIP-24 UI

ibz:

PR #5395, opened 2025-07-21
Payment requests (SLIP-24) UI for Ethereum (Eckhart)

ibz:
This PR implements the TS7 UI for swaps ([SLIP-24 payment requests](https://github.com/satoshilabs/slips/blob/master/slip-0024.md)). Previously, #5253 did it for TS3 and TS5.

The PR takes advantage of the menus and back button support introduced in #5396.

Note: as discussed with @obrusvit, the design of the *trade screen* was slightly simplified, in that:
 * We don't do custom background colors for buy/sell (this can easily be improved though if we use a custom component rather than the `TextScreen`).
 * The ticker symbol is part of the same text and uses the same style as the value (again, we will do a separate component, which should be back-ported to TS5 as well).

Another simplification that was made UI wise is: if there is more than one `CoinPurchaseMemo` present in the *payment request* (ie. more than one trade happens in one transaction), there is no *back button*, but rather a *cancel button*. This should not be an issue in practice as there is no reason for the server to send multiple trades in one transaction, but it is theoretically possible according to SLIP-24 - so in that small theoretical case where it happens, the UX will be slightly downgraded by losing the *back button*. It can be improved, if needed, once we have a better application-level mechanism for dealing with the back button! If that fix is done, `FirmwareUI::confirm_trade` can be simplified by removing the `back_button` parameter and having it simply use a back button on models that support it and a cancel button on other models.


core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -405,12 +406,45 @@ impl FirmwareUI for UIEckhart {
...
+        sell_amount: TString<'static>,
+        buy_amount: TString<'static>,
+        back_button: bool,
     ) -> Result<impl LayoutMaybeTrace, Error> {
-        Err::<RootComponent<Empty, ModelUI>, Error>(ERROR_NOT_IMPLEMENTED)
+        let font = fonts::FONT_SATOSHI_REGULAR_38;
+        let text_style = theme::firmware::TEXT_REGULAR;
+        let mut ops = OpTextLayout::new(text_style);
+        ops.add_color(theme::WHITE)
+            .add_text(subtitle, fonts::FONT_SATOSHI_REGULAR_22)
```
bieleluk: Please, use the `TextScreen` builder function `with_subtitle` instead.

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -405,12 +406,45 @@ impl FirmwareUI for UIEckhart {
...
+            .add_newline()
+            .add_newline()
+            .add_color(theme::RED)
+            .add_text(sell_amount, font)
+            .add_newline()
+            .add_newline()
+            .add_color(theme::GREEN)
```
bieleluk: The proper color should be `GREEN_LIME` according to [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=11157-6501&t=RXVsHe5Lacf1bfmY-0)

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -405,12 +406,45 @@ impl FirmwareUI for UIEckhart {
...
+        let font = fonts::FONT_SATOSHI_REGULAR_38;
+        let text_style = theme::firmware::TEXT_REGULAR;
+        let mut ops = OpTextLayout::new(text_style);
+        ops.add_color(theme::WHITE)
+            .add_text(subtitle, fonts::FONT_SATOSHI_REGULAR_22)
+            .add_newline()
+            .add_newline()
+            .add_color(theme::RED)
+            .add_text(sell_amount, font)
+            .add_newline()
```
bieleluk: I would suggest using precise offset 16 + 12 + 16 according to Figma instead of 2 newlines:
`.add_offset(Offset::y(44))`

core/src/trezor/ui/layouts/eckhart/__init__.py
```diff
@@ -1048,6 +1060,160 @@ async def confirm_ethereum_approve(
...
+                        (TR.address_details__derivation_path, account_path)
+                    )
+                account_items.append((TR.ethereum__approve_chain_id, chain_id))
+
+                res = await _confirm_summary(
+                    None,
+                    None,
+                    maximum_fee,
+                    TR.words__transaction_fee,
+                    TR.words__title_summary,
```
bieleluk:
```suggestion
                    TR.words__swap,
```

core/src/trezor/ui/layouts/eckhart/__init__.py
```diff
@@ -1048,6 +1060,160 @@ async def confirm_ethereum_approve(
...
+        maximum_fee: str,
+        fee_info_items: Iterable[tuple[str, str]],
+        token_address: str,
+    ) -> None:
+        from trezor.ui.layouts.menu import Menu, confirm_with_menu
+
+        for text in texts:
+            await raise_if_cancelled(
+                trezorui_api.confirm_value(
+                    title=TR.words__confirm, value=text, description=None
```
bieleluk: Nit:
- Having a consistent title would be preferable, but it's just a feeling.
- The screen doesn't have an importance for the confirm button to have the green gradient, IMO. Setting explicit verb parameter `buttons__confirm` would make the gradient grey.

```suggestion
                    title=TR.words__swap, value=text, description=None, verb=TR.buttons__confirm
```

tests/device_tests/ethereum/test_signtx.py
```diff
@@ -533,10 +535,15 @@ def test_signtx_staking_eip1559(client: Client, parameters: dict, result: dict):
...
-    skip="t2t1,t3w1",
-    reason="T1 does not support payment requests. Payment requests not yet implemented on model T and TS7.",
+    skip="t2t1",
+    reason="T1 does not support payment requests. Payment requests not yet implemented on model T.",
+)
+@pytest.mark.parametrize(
+    "has_refund,has_text,has_multiple_purchases", list(product([True, False], repeat=3))
 )
-def test_signtx_payment_req(client: Client):
+def test_signtx_payment_req(
```
bieleluk, line 544: In `Caesar` layout in the `TradeScreen`, the test doesn't traverse the available menu. Is it intentional?
cc @romanz 
image

[test link]( https://data.trezor.io/dev/firmware/ui_report/16440970331/diff/T3B1_en-device_tests-ethereum-test_signtx.py%3A%3Atest_signtx_payment_req%5BTrue-True-True%5D.html)

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
ibz: Much better! 888e7e0444b15f81ae4fbd2c3e9db9f6bbb53ee8

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
ibz: Nice, thank you! 888e7e0444b15f81ae4fbd2c3e9db9f6bbb53ee8

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
ibz: Done: 888e7e0444b15f81ae4fbd2c3e9db9f6bbb53ee8

core/src/trezor/ui/layouts/eckhart/__init__.py
ibz: Agree with the button and changed 75cebd9e18d404d42097289e783966e74cc8b09f

Disagree with the title. Remember that this is SLIP-24, not "swaps", which is a subset of SLIP-24. Although not part of the design, it is part of the standard. The `CoinPurchaseMemo` could be entirely missing and the transaction would just confirm some text...

core/src/trezor/ui/layouts/eckhart/__init__.py
ibz: OK, I changed this to match Figma, but it will look strange if there is no `CoinPurchaseMemo` (again, should not happen, but you never know - I need to implement the standard, not "swaps").

So I did another fixup where I look at whether we have any trades or not: 7921aacf39a5cd31c985d6752d6c163240b3e4f3

Speaking of which ... this will need to change again once I implement "sell" in addition to swap (currently not supported) - so it will be either "Swap" or "Sell" or "Summary"...

core/src/trezor/ui/layouts/eckhart/__init__.py
bieleluk: I have no strong opinion about the title. The only thing that looks a bit odd is the "Confirm" in both the tile and the action bar. The "Swap" suggestion was just an example, so they are (ideally) not the same.

ibz: How about this? 46c24cc0e3a69dad735a857ec1f9243980a07140

I think they are both consistent and cover the case when there are no swaps...

tests/device_tests/ethereum/test_signtx.py
ibz: Good catch! Fixed! db6c043e9163a08a94b1e968bd60f2b2768f6d64

bieleluk (approved):

PR #5396, opened 2025-07-21
Menus for Eckhart

ibz:


core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -468,6 +471,9 @@ impl FirmwareUI for UIEckhart {
         let header = if info {
             Header::new(title)
                 .with_right_button(Button::with_icon(theme::ICON_INFO), HeaderMsg::Menu)
+        } else if external_menu {
```
romanz, line 476: nit: maybe raise an exception if both `info` and `external_menu` are set?

core/src/trezor/ui/layouts/delizia/__init__.py
```diff
@@ -39,7 +39,7 @@ def confirm_action(
     br_code: ButtonRequestType = BR_CODE_OTHER,
     prompt_screen: bool = False,
     prompt_title: str | None = None,
-) -> Awaitable[None]:
+) -> Awaitable[ui.UiResult | None]:
```
romanz: IIUC, the `| None` above is due to the use of `raise_if_cancelled()` here:
https://github.com/trezor/trezor-firmware/blob/9e23d54f47253c66fcc0994c3911cfb54f7721ae/core/src/trezor/ui/layouts/delizia/__init__.py#L66

Maybe we can use `interact()` instead - to simplify the return type of this function (to be `Awaitable[ui.UiResult]`)?

Also, we don't really return `None` in this case...

romanz: See recent discussion here: https://github.com/trezor/trezor-firmware/pull/5349 :)

ibz: Indeed, this looks much cleaner! b2d37eb9c0ffe633a4fa6b597e7eb6ba10bf01bf

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
ibz: Yep! 087bc8ece4d253503b83407e928618ca29090b08

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -431,6 +431,8 @@ impl FirmwareUI for UIEckhart {
         warning_footer: Option<TString<'static>>,
         external_menu: bool,
     ) -> Result<Gc<LayoutObj>, Error> {
+        assert!(!(info && external_menu));
```
romanz: I would prefer using [`debug_assert`](https://doc.rust-lang.org/std/macro.debug_assert.html) or just returning an `Err` (to avoid RSOD on production firmware).

PR #5400, opened 2025-07-22
feat(core): adjust bootloader suspend and dim time

obrusvit:
New values:
- 30s to dim
- 40s to suspend

### DIFF
--- a/core/embed/rust/src/ui/layout/simplified.rs
+++ b/core/embed/rust/src/ui/layout/simplified.rs
@@ -15,7 +15,7 @@ use crate::trezorhal::sysevent::{sysevents_poll, Syshandle};
 
 #[cfg(feature = "power_manager")]
 use crate::{
-    time::Instant,
+    time::{Duration, Instant},
     trezorhal::power_manager::{hibernate, is_usb_connected, suspend},
     ui::display::fade_backlight_duration,
     ui::event::PhysicalButton,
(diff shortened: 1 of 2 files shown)


core/embed/rust/src/ui/layout/simplified.rs
```diff
@@ -215,11 +215,11 @@ pub fn run(frame: &mut impl Component<Msg = impl ReturnToC>) -> u32 {
                 let elapsed = Instant::now().checked_duration_since(start);
 
                 if let Some(elapsed) = elapsed {
-                    if elapsed.to_secs() >= 10 && !faded {
+                    if elapsed.to_secs() >= 30 && !faded {
                         faded = true;
-                        fade_backlight_duration(20, 200);
+                        fade_backlight_duration(45, 200);
```
TychoVrahe: here in rust we can use ModelUI::get_backlight_low()

TychoVrahe (changes requested):

core/embed/rust/src/ui/layout/simplified.rs
obrusvit: 57e251e1aad16cd1e5d8cc4d63e12dccd12b813e

And I also introduced similar constants in Rust `FADE_TIME` and `SUSPEND_TIME` of type `Duration`

TychoVrahe (approved):

PR #5403, opened 2025-07-22
Convert `r != buf.len` to OSError in WebUSB write

M1nd3r:

### DIFF
--- a/core/embed/upymod/modtrezorio/modtrezorio-webusb.h
+++ b/core/embed/upymod/modtrezorio/modtrezorio-webusb.h
@@ -121,7 +121,17 @@ STATIC mp_obj_t mod_trezorio_WebUSB_write(mp_obj_t self, mp_obj_t msg) {
   mp_obj_WebUSB_t *o = MP_OBJ_TO_PTR(self);
   mp_buffer_info_t buf = {0};
   mp_get_buffer_raise(msg, &buf, MP_BUFFER_READ);
+
+  if (buf.len != USB_PACKET_LEN) {
+    mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Invalid buffer length"));
+  }
+
   ssize_t r = usb_webusb_write(o->info.iface_num, buf.buf, buf.len);
+
+  if (r != buf.len) {
+    mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Write failed"));
+  }
+
   return MP_OBJ_NEW_SMALL_INT(r);
 }
 STATIC MP_DEFINE_CONST_FUN_OBJ_2(mod_trezorio_WebUSB_write_obj,


core/embed/upymod/modtrezorio/modtrezorio-webusb.h
```diff
@@ -122,6 +122,9 @@ STATIC mp_obj_t mod_trezorio_WebUSB_write(mp_obj_t self, mp_obj_t msg) {
   mp_buffer_info_t buf = {0};
   mp_get_buffer_raise(msg, &buf, MP_BUFFER_READ);
   ssize_t r = usb_webusb_write(o->info.iface_num, buf.buf, buf.len);
+  if (r == -1) {
```
TychoVrahe: why just -1? there is arguably not enough (=none at all) documentation of the `usb_webusb_write` but i'd say that any negative value is error and unix implementation return 0 when bad interface number is passed so thats error too, so i'd think r <= 0 would be more precise

M1nd3r: Changed to `r <= 0` and rebased: https://github.com/trezor/trezor-firmware/pull/5403/commits/090f241df690c5b779a184b92e4ea4ffd7e57102

core/embed/upymod/modtrezorio/modtrezorio-webusb.h
```diff
@@ -122,6 +122,9 @@ STATIC mp_obj_t mod_trezorio_WebUSB_write(mp_obj_t self, mp_obj_t msg) {
   mp_buffer_info_t buf = {0};
   mp_get_buffer_raise(msg, &buf, MP_BUFFER_READ);
   ssize_t r = usb_webusb_write(o->info.iface_num, buf.buf, buf.len);
+  if (r <= 0) {
+    mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Write failed"));
```
romanz: nit: consider using `mp_type_OSError` (following https://github.com/micropython/micropython/issues/921)

core/embed/upymod/modtrezorio/modtrezorio-webusb.h
romanz: Will it fail if the user will try to write an empty buffer?
Maybe we can fail if `r != buf.len`?

core/embed/upymod/modtrezorio/modtrezorio-webusb.h
M1nd3r: `OSError` seems to be more fitting, thanks. Exception type changed here: https://github.com/trezor/trezor-firmware/pull/5403/commits/7a23357045aa1383b84753b2564f474a81d8aaf1

core/embed/upymod/modtrezorio/modtrezorio-webusb.h
M1nd3r: I do not have a strong opinion. I have changed the condition to `r != buf.len` here: https://github.com/trezor/trezor-firmware/pull/5403/commits/b375a9c1dee7db29482c7fd08c0614785cf73e47

@TychoVrahe WDYT?

TychoVrahe: i am ok with that too

TychoVrahe (approved):

core/embed/upymod/modtrezorio/modtrezorio-webusb.h
matejcik: oh cool so is `usb_webusb_write` guaranteed to block until (a) write is complete or (b) a non-retriable error occurs?

also what happens if `buf.len` is not packet-sized?

M1nd3r: Ok, so imo we have the following options for the "should raise" condition

1. `r == -1` - improvement over the current state, as it converts at least some of the errors to `OSError`
2. `r < 0` - improvement over 1), as it might catch more errors and we do not expect `len` to be negative
3. `r <=0` - will handle more errors than 2), as @TychoVrahe mentioned ( [...]unix implementation return 0 when bad interface number is passed...). But it will raise if the `buf.len` to write is also 0. (Is goes against the "least surprise" paradigm, but otherwise it does not seem to be an issue **to me** as forbidding caller to write a buffer of len==0 does not seem like a unreasonable restriction)
4. `r != buf.len` - will raise if "something" went wrong. BUT it will also NOT raise when buffer len is zero and error mentioned by @TychoVrahe occurs. And, it might not work as expected when `buf.len` is not packet sized.

_Note A]:_ We could add a second check that `buf.len` is of size `USB_PACKET_LEN`, similarly to `webusb_read`, resolving some of the problems with point 4.

My (not strong) opinion:
- Options 1 and 2 are safe and are improvements of the current state.
- Option 3 is mostly safe, except if we (for some reason) wanted to write buffers of len zero.
- Option 4 might raise for not-foreseen reasons, but when combined with _Note A]_ it guarantees that the len returned is what we expect on successful write of correct-length data input

 @matejcik What solution do you prefer?

matejcik: > We could add a second check that buf.len is of size USB_PACKET_LEN, similarly to webusb_read, resolving some of the problems with point 4.

i'd prefer to do this: passing a buffer whose len != `USB_PACKET_LEN` is a `ValueError`

other than that, my preferred solution depends on behavior of `usb_webusb_write`: if the lower level function promises to either (a) write everything or (b) fail non-retriably, then `r != buf.len` or `r != USB_PACKET_LEN` is the correct thing to check
if there's the possibility that it returns 0 for a _retriable_ error, that is, "someone jumped in the queue but try again", then it's more practical to _return_ the zero instead of raising for it.
which is it @TychoVrahe ?

(we could also fix the unix impl to return `-1` instead of `0` for a bad endpoint :) )

M1nd3r: Raising`ValueError` for _Invalid buffer length_ added here: https://github.com/trezor/trezor-firmware/pull/5403/commits/ac46feff524e174db50c3071929b940270943a2d

romanz (approved):
utACK

core/embed/upymod/modtrezorio/modtrezorio-webusb.h
TychoVrahe: i don't see a possibility of returning 0 for retriable error and the unix fix seems like a good idea too, also

>  function promises to either (a) write everything or (b) fail non-retriably,

assumption seems correct to me.

So yep, i'd go with matejciks proposal.

M1nd3r: > the unix fix seems like a good idea too

1. I would not do the fix here, as it is a independent - we would raise for 0 either way, as it will not be the buffer len
2. I have looked at the _unix_ implementation, and `return 0` is there in multiple (twelve) places. Should I change **all** of those return statements, or only some of them? Should I look in other files/places too? @TychoVrahe

M1nd3r: 3. PR for `return 0` -> `return -1` for errors in unix usb stack: 
https://github.com/trezor/trezor-firmware/pull/5437

M1nd3r: ad 2) The PR #5437  replaces `0` by `-1` in the *unix* implementation of the usb stack in all but one place - the error in `usb_emulated_read(...)` is retriable.

PR #5413, opened 2025-07-23
feat(core): derive entropy from master key

cepetr:
This PR introduces a new approach to calculating entropy (storage salt) used during storage initialization on T3W1.

- A new vendor header field (`fw_type`) has been introduced - currently used only on T3W1 and D002.
- Entropy is now derived from the unprivileged master key and diversified using the firmware type read from the vendor header.
- The entropy module API was updated to support variable-length entropy data.

**This PR introduces an incompatibility with data stored in the storage. After booting T3W1, a progress screen briefly appears with "WRONG PIN", followed by a RSOD with a runtime error. This is expected behavior, which needs to be resolved by performing a factory reset in the bootloader.**


core/embed/util/image/inc/util/image.h
```diff
@@ -88,6 +88,22 @@ typedef struct {
...
+//
+typedef enum {
+  // Reserved value (may appear in legacy vendor headers)
+  VENDOR_FW_TYPE_RESERVED = 0,
+  // Unsafe firmware (e.g. development or custom firmware)
+  VENDOR_FW_TYPE_UNSAFE = 1,
+  // Official Trezor firmware
+  VENDOR_FW_TYPE_SAFE = 2,
+  // Official Trezor firmware with BTC-only support
+  VENDOR_FW_TYPE_SAFE_BTC_ONLY = 2,
```
TychoVrahe: should be 3 and prodtest below 4

core/embed/models/T3W1/vendorheader/vendor_dev_DO_NOT_SIGN.json
```diff
@@ -2,6 +2,7 @@
     "header_len": 1024,
     "text": "DEV ONLY, DO NOT USE!",
     "hw_model": "T3W1",
+    "fw_type": 1,
```
TychoVrahe: I think that devkit headers should mimic the production headers, so that we can test any related behavior properly. Thus this one should be type 2 and `vendor_prodtest_DO_NOT_SIGN` should be 4

core/embed/util/image/inc/util/image.h
cepetr: yes, fixed, see 627e8576d2c7f8113ef42a8d68e6e7ddde6385b9

core/embed/models/T3W1/vendorheader/vendor_dev_DO_NOT_SIGN.json
cepetr: Ok, so how about to change it for D002 as well:

|                  |   T3W1      |   D002    |
|----------------------|-------------|-----------|
| dev_DO_NOT_SIGN      |   1 -> 2    |   1 -> 2  |
| prodtest_DO_NOT_SIGN |   1 -> 4    |   1 -> 4  |
| unsafe               |     1       |     1     |
| prodtest             |     4       |     –     |
| trezor_btconly       |     3       |     –     |
| trezor               |     2       |     –    |

Do we need to have `1` - unsafe then? Is anything that is not officially signed unsafe?

TychoVrahe: Unsafe is for custom firmware. On production devices its only way to upload anything not officially signed, on devkits its useful just for testing the unsafe fw differences, like the warning screen, runtime limit etc.

TychoVrahe: Otherwise i agree with the table.

cepetr: I updated values, see f71d1b6be83ec8bbfd5671846cfe7c4e1bcaa034, please double check

core/embed/models/D002/vendorheader/vendor_dev_DO_NOT_SIGN.json
```diff
@@ -2,6 +2,7 @@
     "header_len": 4608,
     "text": "DEV ONLY, DO NOT USE!",
     "hw_model": "D002",
+    "fw_type": 2,
```
andrewkozlik, line 5: I am not clear on what's the difference between UNSAFE and dev_DO_NOT_SIGN. Is UNSAFE what we normally refer to as custom firmware on normal end-user devices? What's the latter then? Is it firmware that can be run only on developer devices that have a special boardloader/bootloader?

The reason why I ask is that it seems to me that `fw_type` should be associated with the pubkeys, in which case these `fw_type`s for dev_DO_NOT_SIGN would be 1. However if these vendor headers are usable only on special developer devices, then the current numbers make sense.

TychoVrahe: correct DO_NOT_SIGN headers are for devktis/devices with non-production bootloaders only, unsafe is for custom firmware on production devices. See also discussion above.

core/embed/util/image/inc/util/image.h
```diff
@@ -88,6 +88,22 @@ typedef struct {
...
+  // Reserved value (may appear in legacy vendor headers)
+  VENDOR_FW_TYPE_RESERVED = 0,
+  // Unsafe firmware (e.g. development or custom firmware)
+  VENDOR_FW_TYPE_UNSAFE = 1,
+  // Official Trezor firmware
+  VENDOR_FW_TYPE_SAFE = 2,
+  // Official Trezor firmware with BTC-only support
+  VENDOR_FW_TYPE_SAFE_BTC_ONLY = 3,
+  // Official Trezor prodtest firmware
+  VENDOR_FW_TYPE_SAFE_PRODTEST = 4,
```
matejcik: no reason to spell out "safe", everything is "safe" except the one custom entry

```suggestion
  // Custom (unsafe) firmware
  VENDOR_FW_TYPE_UNSAFE = 1,
  // Trezor Universal firmware
  VENDOR_FW_TYPE_UNIVERSAL = 2,
  // Trezor Bitcoin-only firmware
  VENDOR_FW_TYPE_BTC_ONLY = 3,
  // Factory tester firmware
  VENDOR_FW_TYPE_PRODTEST = 4,
```

cepetr: Shouldn't we also rename  `VENDOR_FW_TYPE_UNSAFE` to `VENDOR_FW_TYPE_CUSTOM`?

core/embed/sec/entropy/inc/sec/entropy.h
andrewkozlik, line 1: I think that shortening the name to simply entropy like `entropy.h`, `entropy_get()`, `ENTROPY_MAX_SIZE` etc. is not a good idea. Entropy is a very generic term which we use to mean randomness. The previous term that we used here was "hardware entropy", which was meant to combine hardware-specific pieces of information to add randomness to the storage encryption process. The term entropy is also used for the Host's random secret during seed generation. We should either go back to the previous term, which is not very good now, because there is nothing hardware-specific about it any more, or we should use a term which expresses how the secret is used, such as "storage salt". Whatever we do, the terminology used in secret_keys.h should match.

core/embed/util/image/inc/util/image.h
cepetr: renamed, see ce27f77484dbcdb6c7b86cdac037de0702dee20b

core/embed/sec/entropy/inc/sec/entropy.h
cepetr: As we discussed in person, I'll rename the module to `storage_salt` or something similar and move it to the new `sec/storage` folder.  I’ve found it much easier to do this as part of the following PR (#5394 – not ready yet), where I am going to remove entropy api from syscall/smcall interface and add the `storage_setup()` function.

cepetr: PR #5394 where `entropy` is renamed to `storage_salt` is almost ready

Issue #5404, opened 2025-07-22
Check return values in `**/main.c`


onvej-sl:
I suggest checking the return values of the `*_init()` functions in the `**/main.c` files.

PR #5416, opened 2025-07-23
Refactor tropic inicialization

onvej-sl:


onvej-sl:
See also https://github.com/trezor/trezor-firmware/issues/5404.

core/embed/projects/secmon/main.c
```diff
@@ -86,7 +86,7 @@ static void drivers_init(void) {
 #endif
 
 #ifdef USE_TROPIC
-  tropic_init();
```
onvej-sl: I don't understand why `tropic_init()` is called in both the kernel and the secure monitor.

TychoVrahe: in kernel, `tropic_init` is called only if `SECURE_MODE` is defined, which is defined when we are not using secure monitor.

core/embed/sec/tropic/inc/sec/tropic.h
```diff
@@ -23,19 +23,13 @@
 
 #ifdef KERNEL_MODE
 
-#define TROPIC_CHIP_ID_SIZE 128
-#define TROPIC_RISCV_FW_SIZE 4
-#define TROPIC_SPECT_FW_SIZE 4
+#include "libtropic.h"
 
-bool tropic_init(void);
+bool tropic_init(bool init_secure_channel);
```
TychoVrahe: how about using the same initialization api as optiga?

there is `optiga_init`
and `optiga_init_and_configure` that also includes the secure channel handshake.

core/embed/sec/tropic/tropic.c
```diff
@@ -29,121 +29,73 @@
...
-  }
-
-  tropic_hal_deinit();
-
+  lt_deinit(&drv->handle);
   memset(drv, 0, sizeof(*drv));
 }
 
-bool tropic_get_spect_fw_version(uint8_t *version_buffer, uint16_t max_len) {
+lt_handle_t *tropic_get_handle() {
```
TychoVrahe: lt_handle_t* tropic_get_handle(void)

core/embed/sec/tropic/inc/sec/tropic.h
```diff
@@ -23,19 +23,13 @@
...
 
 void tropic_deinit(void);
 
-bool tropic_get_spect_fw_version(uint8_t* version_buffer, uint16_t max_len);
-
-bool tropic_get_riscv_fw_version(uint8_t* version_buffer, uint16_t max_len);
-
-bool tropic_get_chip_id(uint8_t* chip_id, uint16_t max_len);
+lt_handle_t* tropic_get_handle();
```
TychoVrahe: `lt_handle_t* tropic_get_handle(void)` 

i originally wanted to avoid exposing `libtropic` outside this module as it breaks encapsulation. I assume that there just too much friction to keep this approach for provisioning, but maybe we could only expose for prodtest using `#ifdef TREZOR_PRODTEST`?

core/embed/sys/suspend/stm32u5/suspend_io.c
```diff
@@ -109,7 +109,7 @@ void resume_secure_drivers() {
   optiga_init_and_configure();
 #endif
 #ifdef USE_TROPIC
-  tropic_init();
+  tropic_init(true);
```
TychoVrahe: this can also be used in prodtest, where we normally call with 'false' - is it fine in that context?

core/embed/sec/tropic/inc/sec/tropic.h
onvej-sl: This makes sense. However, it requires `optiga.c` to expose the tropic handle even in firmware so that `tropic_config.c` can use it.

core/embed/sys/suspend/stm32u5/suspend_io.c
onvej-sl: This is a problem because if this is called before invoking `tropic-pair` in prodtest, the establishing of the secure channel fails, which leads to the deinitialization of tropic. Therefore, we can either use `TREZOR_PRODTEST` or attempt to establish the secure channel always and ignore any failures. The second option is what we do with optiga but I don't like it.

core/embed/sec/tropic/inc/sec/tropic.h
TychoVrahe: Assuming you meant `tropic.c`. We could place `tropic_init_and_configure` into `tropic.c` , there is no hard requirement to have separate files, or put the exposed handle into `tropic_internal.h`. But i don't insist on this, i just think it would be easier to reason about the API.

core/embed/sys/suspend/stm32u5/suspend_io.c
TychoVrahe: I believe that we should not offload this complexity into the `suspend` module. In theory, as this is a public API we should also not assume that it is only prodtest that would call this before pairing, maybe in future we will do this in bootloader or so. So i kinda think that the optiga solution to try to establish the channel always is reasonable.

core/embed/sec/tropic/inc/sec/tropic.h
onvej-sl: > [...] but maybe we could only expose for prodtest using #ifdef TREZOR_PRODTEST?

Done in https://github.com/trezor/trezor-firmware/pull/5416/commits/5cd9e138f99f467a4655f6d477681be7e52a4283.

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5416/commits/62e55fcc1ad1b7e389914fe46a0f6e0c1a9128db.

core/embed/sys/suspend/stm32u5/suspend_io.c
onvej-sl: I decided to revert to the previous API: There is only `tropic_init(void)`, which attempts to establish the secure channel. The only significant change is that I ignore if the attempt fails.

onvej-sl: See https://github.com/trezor/trezor-firmware/pull/5416/commits/5cd9e138f99f467a4655f6d477681be7e52a4283.

core/embed/sec/tropic/inc/sec/tropic.h
onvej-sl: I decided not to use `optiga_init_and_configure` for now. It can always be refactored later, and I need this pull request to be merged soon.

core/embed/sec/tropic/inc/sec/tropic.h
TychoVrahe: `#include "libtropic.h"` should go under the `ifdef` too

onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5416/commits/9791acca2847af97a9955c02af0d40292b0c1837.

TychoVrahe (approved):

PR #5417, opened 2025-07-23
Firmware autodim

obrusvit:
This PR implements autodim functionality in firmware.

cfd9087b132b7a28c21fa2c62be11e971173e475 Firstly, a new `pm` (power manager) module is exposed to MicroPython which supports two functions:
- `suspend()` - this is not yet used in this PR
- `is_usb_connected()`

e88ee707e557d011eced3edcbb3e3d98003de3ff Similarly to bootloader, firmware sets backlight to LOW (45) after 30s of inactivity when the device is not connected via USB. This is implemented via `IdleTimer` which is newly "poked" also by `Progress::report` and also by connecting/disconnecting USB charger.


core/src/trezor/ui/__init__.py
```diff
@@ -374,34 +377,52 @@ def create_tasks(self) -> Iterator[loop.Task]:
...
+        def _handle_touch_events(self) -> Generator:
+            """Task that is waiting for the user touch input."""
+            touch = loop.wait(io.TOUCH)
+            try:
+                while True:
+                    # Using `yield` instead of `await` to avoid allocations.
+                    event = yield touch
+                    workflow.idle_timer.touch()
+                    if utils.USE_POWER_MANAGER:
+                        autodim_clear()
```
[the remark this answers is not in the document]
obrusvit: This is correct because clearing the autodim state is not needed when power management is not a part of the build.

core/embed/rust/src/ui/api/firmware_micropython.rs
```diff
@@ -1228,6 +1228,21 @@ pub extern "C" fn upy_check_homescreen_format(data: Obj) -> Obj {
...
+pub extern "C" fn upy_backlight_get() -> Obj {
+    let block = || {
+        #[cfg(feature = "backlight")]
+        {
+            let backlight_level = backlight();
+            Ok(Obj::try_from(backlight_level)?)
+        }
+        #[cfg(not(feature = "backlight"))]
+        {
+            Ok(Obj::const_none())
```
romanz: nit: maybe it's clearer to raise `mp_type_NotImplementedError`?
This way `backlight_get()` return type won't have to be `int | None`.

core/embed/upymod/modtrezorio/modtrezorio-pm.h
```diff
@@ -0,0 +1,98 @@
...
+     MP_ROM_PTR(&mod_trezorio_pm_is_usb_connected_obj)},
+
+    // Wakeup flag constants
+    {MP_ROM_QSTR(MP_QSTR_WAKEUP_FLAG_BUTTON), MP_ROM_INT(WAKEUP_FLAG_BUTTON)},
+    {MP_ROM_QSTR(MP_QSTR_WAKEUP_FLAG_POWER), MP_ROM_INT(WAKEUP_FLAG_POWER)},
+    {MP_ROM_QSTR(MP_QSTR_WAKEUP_FLAG_BLE), MP_ROM_INT(WAKEUP_FLAG_BLE)},
+    {MP_ROM_QSTR(MP_QSTR_WAKEUP_FLAG_NFC), MP_ROM_INT(WAKEUP_FLAG_NFC)},
+    {MP_ROM_QSTR(MP_QSTR_WAKEUP_FLAG_RTC), MP_ROM_INT(WAKEUP_FLAG_RTC)},
+
+    // Power manager event flags
```
romanz, line 85: Are the new event flag used in this PR?

core/embed/upymod/modtrezorio/modtrezorio-pm.h
```diff
@@ -0,0 +1,98 @@
...
+///     failure.
+///     Wakeup flags: BUTTON=1, POWER=2, BLE=4, NFC=8, RTC=16
+///     """
+STATIC mp_obj_t mod_trezorio_pm_suspend() {
+  wakeup_flags_t wakeup_flags = 0;
+  pm_status_t res = pm_suspend(&wakeup_flags);
+  if (res != PM_OK) {
+    mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Failed to suspend"));
+  }
+  return mp_obj_new_int((mp_int_t)wakeup_flags);
```
romanz: nit:

```suggestion
  return mp_obj_new_int_from_uint(wakeup_flags);
```

core/src/apps/base.py
```diff
@@ -485,6 +485,10 @@ def boot() -> None:
         workflow_handlers.register(msg_type, handler)
 
     reload_settings_from_storage()
+    if utils.USE_POWER_MANAGER:
+        from apps.management.pm.autodim import autodim_display
```
romanz, line 491: Can `pm` module be excluded from build, if `USE_POWER_MANAGER` is not set?

core/src/trezor/ui/__init__.py
```diff
@@ -374,34 +377,52 @@ def create_tasks(self) -> Iterator[loop.Task]:
...
-                event = yield touch
-                workflow.idle_timer.touch()
-                self._event(event_call, *event)
-        except Shutdown:
-            return
-        finally:
-            touch.close()
+    if utils.USE_BUTTON:
+
+        def _handle_button_events(self) -> Generator:
```
romanz, line 394: `_handle_button_events` and `_handle_touch_events` look quite similar...

Would it be possible to use `_handle_input_iface` for both of them, and rely on `if utils.USE_POWER_MANAGER:` branch being dropped in compile time for models that don't support power management?

obrusvit: This is a preparation for suspend functionality where `_handle_button_events` will also check if power button was pressed and if so, then calls `suspend_device()` function (not yet in this PR).

ISTM, that doing it all in one function seems too much because you need to check for `if utils.USE_POWER_MANAGER` and then also whether it's `io.BUTTON` or `io.TOUCH`

core/embed/upymod/modtrezorio/modtrezorio-pm.h
obrusvit: Wakeup flags are prepared for suspend functionality.

As for Event flags, only `EVENT_USB_CONNECTED_CHANGED` is used now in `src/trezor/ui/__init__.py::_handle_power_manager`

core/embed/upymod/modtrezorio/modtrezorio-pm.h
obrusvit: https://github.com/trezor/trezor-firmware/pull/5417/commits/ea72c991758f468ec113dd504a2d5b7ec174caad

core/src/apps/base.py
obrusvit: Thanks, this should do it
https://github.com/trezor/trezor-firmware/pull/5417/commits/926ab04c162b3e24a76e0c206c9f1ab296df2d09

core/embed/rust/src/ui/api/firmware_micropython.rs
obrusvit: You're right, but we do not have `NotImplementedError` exposed in `error.rs`. Should I add it now? 

Or do you think raising `RuntimeError` is fine? https://github.com/trezor/trezor-firmware/pull/5417/commits/036fc012e2a9a0797c527e0dd9df561f1455e5c4

core/src/trezor/ui/__init__.py
```diff
@@ -374,34 +377,52 @@ def create_tasks(self) -> Iterator[loop.Task]:
         """Set up background tasks for a layout.
 
         Called from `start()`. Creates and yields a list of background tasks, typically
-        event handlers for different interfaces.
+        event handlers for different interfaces. Event handlers are enabled based on build options to prevent keeping stale events in the event queue.
```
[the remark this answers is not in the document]
obrusvit: https://github.com/trezor/trezor-firmware/pull/5417/commits/f50f96043f801588f70e45f93dc062e709ce549a

core/embed/rust/src/ui/api/firmware_micropython.rs
obrusvit: I messed up the build https://github.com/trezor/trezor-firmware/pull/5417/commits/c30805912701e3f5cd0f9ae0b6f5548a38235e7a

romanz: `RuntimeError` with an appropriate error message is fine.

core/embed/rust/src/ui/api/firmware_micropython.rs
```diff
@@ -1228,6 +1228,23 @@ pub extern "C" fn upy_check_homescreen_format(data: Obj) -> Obj {
     unsafe { util::try_or_raise(block) }
 }
 
+pub extern "C" fn upy_backlight_get() -> Obj {
+    let block = || {
+        #[cfg(feature = "backlight")]
+        {
+            let backlight_level = backlight();
+            Ok(Obj::try_from(backlight_level)?)
```
romanz:
```suggestion
            Ok(Obj::from(backlight_level))
```

obrusvit: https://github.com/trezor/trezor-firmware/pull/5417/commits/29b5eab774d4f07cb2a710f937d77ef1af8db628

Thanks.

PR #5418, opened 2025-07-23
add support for direct test mode

TychoVrahe:


kopecdav:
I was able to upload the binary, but the prodtest command isn't reading DTM responses, so I can't confirm if the application is working. The nRF Connect DTM application was also confused. What's the intended usage for the prodtest command? Should it just blindly forward commands without any response handling?

TychoVrahe:
It should just forward data. nRF connect DTM should work though, but you need to first send the ble-direct-test-mode command to prodtest via tio or something, then disconnect tio and then run the DTM app.

PR #5419, opened 2025-07-24
Minor refactor

bieleluk:
- use relative path for layout-specific imports
- use fewer random numbers in paragraph spacing
- in Eckhart's `textscreen` use  page limit where needed


core/embed/rust/src/ui/layout_eckhart/flow/show_danger.rs
```diff
@@ -71,7 +71,7 @@ pub fn new_show_danger(
         Paragraph::new(&theme::TEXT_MONO_EXTRA_LIGHT, value),
     ]
     .into_paragraphs()
-    .with_placement(LinearPlacement::vertical().with_spacing(35));
+    .with_placement(LinearPlacement::vertical().with_spacing(theme::TEXT_VERTICAL_SPACING));
```
romanz, line 74: It was 35 before this commit... is it OK?

romanz (approved):
LGTM, with a small question below.

core/embed/rust/src/ui/layout_eckhart/flow/show_danger.rs
bieleluk: It was just inconsistent before. Here is 
[Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=7220-10521&t=MVV3SJRDNHYoTIah-0) example with a correct design.

PR #5420, opened 2025-07-24
WIP proplist extension

bieleluk:


core/embed/rust/src/ui/api/firmware_micropython.rs
```diff
@@ -1612,8 +1612,8 @@ pub static mp_module_trezorui_api: Module = obj_module! {
     ///     br_name: str,
     ///     address_item: (str, str) | None,
     ///     extra_item: (str, str) | None,
-    ///     summary_items: Iterable[tuple[str, str]] | None = None,
-    ///     fee_items: Iterable[tuple[str, str]] | None = None,
+    ///     summary_items: list[tuple[str | None, str | bytes | None, bool | None]] | None = None,
+    ///     fee_items: list[tuple[str | None, str | bytes | None, bool | None]] | None = None,
```
mmilata: Time to introduce a type alias maybe?

bieleluk: I'd like to do that, but I'm not sure how to introduce that in the comment of the Rust file.

mmilata: Just an asignment at the beginning:

```rs
/// PropEntry = tuple[str | None, str | bytes | None, bool | None]
```

possibly inside (not sure if needed):

```rs
/// if TYPE_CHECKING:
```

bieleluk: https://github.com/trezor/trezor-firmware/pull/5420/commits/55de9a292c48adb241df1cc9662edb108b4e4e79

core/src/trezor/ui/layouts/bolt/__init__.py
```diff
@@ -777,10 +777,12 @@ def confirm_value(
     if description and value:
         description += ":"
 
-    info_items = info_items or []
+    items: list[PropertyType] = (
+        [(k, v, True) for k, v in info_items] if info_items is not None else []
```
ibz:
```suggestion
        [(k, v, True) for k, v in info_items or []]
```

core/embed/rust/src/ui/layout/util.rs
```diff
@@ -114,6 +114,24 @@ impl PropsList {
...
     }
+
+    pub fn empty(
+        key_font: &'static TextStyle,
+        value_font: &'static TextStyle,
+        value_mono_font: &'static TextStyle,
+        key_value_padding: i16,
+        props_padding: i16,
+    ) -> Result<Self, Error> {
+        let empty_items = List::alloc(&[])?; // Create an empty GC list
```
ibz: nit: "empty list", not "empty items" :)

core/src/trezor/ui/layouts/eckhart/__init__.py
```diff
@@ -1348,9 +1359,16 @@ async def confirm_solana_staking_tx(
                 br_name=br_name,
                 address_item=stake_item,
                 extra_item=blockhash_item,
-                fee_items=fee_details,
+                fee_items=fee_items,
                 summary_title=title,
-                summary_items=(amount_item, fee_item) if amount_item else (fee_item,),
+                summary_items=(
+                    [
```
ibz: Rather than this if/else here, better construct the list beforehand and add the `amount_item` if needed?

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
```diff
@@ -344,26 +344,26 @@ impl FirmwareUI for UIEckhart {
...
         let account_paragraphs = if let Some(items) = account_items {
-            let mut paragraphs = ParagraphVecShort::new();
-            for pair in IterBuf::new().try_iterate(items)? {
-                let [label, value]: [TString; 2] = util::iter_into_array(pair)?;
-                paragraphs
-                    .add(Paragraph::new(&theme::TEXT_SMALL_LIGHT, label).no_break())
-                    .add(Paragraph::new(&theme::TEXT_MONO_LIGHT, value));
-            }
-            Some(paragraphs)
+            Some(PropsList::new(
```
ibz: Could we have something like `PropsList::items_styled(items)` instead of enumerating the styles in multiple places? Also the above `PropsList::empty` has the same problem (styles duplicated in multiple calls) so maybe we could also have `empty_styled()` and `items_styled(...)` would call that internally, so all styles would stay just in one place - `empty_styled`?

core/embed/rust/src/ui/layout_eckhart/flow/confirm_summary.rs
```diff
@@ -179,14 +180,30 @@ pub fn new_confirm_summary(
     let content_extra = content_menu_info(
         extra_title.unwrap_or(TR::buttons__more_info.into()),
         None,
-        extra_paragraphs.map_or_else(ParagraphVecShort::new, |p| p),
+        extra_paragraphs.unwrap_or_else(|| {
+            unwrap!(PropsList::empty(
```
ibz: Can we have a `empty_styled()` that is like `empty` but with the styles applied, so we don't duplicate these?

core/src/trezor/ui/layouts/bolt/__init__.py
bieleluk: The suggestion you gave is right, but it was only an intermediate commit, where the `info_items` wasn't  `PropertyType` list yet.
The current code is:
`items=list(info_items) if info_items else [],`

core/embed/rust/src/ui/layout/util.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5420/commits/9b9799adc6f6bfc4e763edd4c29ae9cd2800581e

core/src/trezor/ui/layouts/eckhart/__init__.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5420/commits/5daa860d96662c554d3e66730319f53d47de45d6

core/embed/rust/src/ui/layout_eckhart/ui_firmware.rs
bieleluk: The biggest issue here is that the `PropsList` is layout-independent. The spacing constants and the key font are mostly constant for a layout. The value and value mono fonts vary in the `Eckhart` layout (i.e., the summary screen has a bigger font compared to the properties screen), so they have to be passed each time. 
I can see 2 possible approaches to avoid repeating the constant parameters in the function call:
1. Use lazy static to define global variables before the first use. I'm not a big fan of this approach.
2. Define constants for each layout with the same name and use the conditional "import":
```
#[cfg(feature = "layout_bolt")]
use crate::ui::layout_bolt::theme::{PROPS_SPACING, PROP_INNER_SPACING};
#[cfg(feature = "layout_caesar")]
use crate::ui::layout_caesar::theme::{PROPS_SPACING, PROP_INNER_SPACING};
#[cfg(feature = "layout_delizia")]
use crate::ui::layout_delizia::theme::{PROPS_SPACING, PROP_INNER_SPACING};
#[cfg(feature = "layout_eckhart")]
use crate::ui::layout_eckhart::theme::{PROPS_SPACING, PROP_INNER_SPACING};
```

WDYT?

bieleluk: This is an example of how it can be done your way. It has advantages but also disadvantages. I don't have any strong preference for either of them.
https://github.com/trezor/trezor-firmware/pull/5420/commits/273ca4792f6d4d1d908a8635b9675438609df941

ibz: I don't have a strong opinion either, but your fix looks better IMO, since we are not repeating the styles. Nice!

PR #5421, opened 2025-07-24
Fix prodtest emulator

onvej-sl:

PR #5424, opened 2025-07-24
Update `libtropic` and `ts-tvl`

onvej-sl:
This pull request builds on of https://github.com/trezor/trezor-firmware/pull/5416 (the oldest commit) and https://github.com/trezor/trezor-firmware/pull/5421 (the second oldest commit).

The pull request updates `libtropic` and `ts-tvl` to their latest stable releases.

As side effect, it removes the `tropic.get_certificate()` function, which I found no use for in the firmware and which depends on a `libtropic` function whose API has changed.


core/site_scons/models/T3W1/emulator.py
```diff
@@ -86,6 +88,8 @@ def configure(
         paths += ["vendor/libtropic/src"]
         defines += ["USE_TREZOR_CRYPTO"]
         defines += [("LT_USE_TREZOR_CRYPTO", "1")]
+        # This is a workaround for an error caused by the redefinition of the macro `STATIC`, which is initially defined in `SConscript.unix` and later redefined in `libtropic_common.h`. As an adverse effect of defining `TEST`, all functions declared in `libtropic.c` become non-static, causing `lt_asn1_der.c` to be compiled even though it is not used.
```
onvej-sl, line 91: I don't know how to fix this issue properly. @cepetr Could you help me with that?

cepetr: `STATIC` is defined not only in `SConscript.unix`, but also in `mpconfig.h`, which is included (and heavily used) in many micropython C files, as well as in our `modtrezorxxx` files. Since we also include `libtropic` in these files, this causes a collision. Somebody needs to step aside... and it's probably not going to be MicroPython.

I believe no library should use the `STATIC` symbol in its public interface, as `libtropic` currently does. BTW, there's a similar potential issue with the `UNUSED` symbol, which need not (and should not) be defined in `libtropic_common.h`.

I think the easiest and the most correct solution is to rename `STATIC` to `LT_STATIC` (or something similar) on the `libtropic` side.

onvej-sl: > I think the easiest and the most correct solution is to rename STATIC to LT_STATIC (or something similar) on the libtropic side.

Hmm, I was hoping we could find a more elegant solution. I opened https://github.com/tropicsquare/libtropic/issues/177.

> I believe no library should use the STATIC symbol in its public interface, as libtropic currently does. BTW, there's a similar potential issue with the UNUSED symbol, which need not (and should not) be defined in libtropic_common.h.

It seems to me that this should not be a problem, as `UNUSED` is used only within `libtropic`'s `*.c` files, and these files do not include our header files.

cepetr: We include `libtropic.h`, that includes `libtropic_common.h`, so there's potential collision with our definition.

onvej-sl: You are right. I didn't realize that `UNUSED` is defined in `libtropic.h`, even though it doesn't need to be, since it is only used in `*.c` files.

onvej-sl:
Closing in favour of https://github.com/trezor/trezor-firmware/pull/5525.

PR #5437, opened 2025-07-28
chore(core): use `-1` instead of `0` for errors in unix usb stack

M1nd3r:
Follow-up to: https://github.com/trezor/trezor-firmware/pull/5403

### DIFF
--- a/core/embed/io/usb/unix/usb.c
+++ b/core/embed/io/usb/unix/usb.c
@@ -264,7 +264,7 @@ secbool usb_webusb_can_write(uint8_t iface_num) {
 int usb_hid_read(uint8_t iface_num, uint8_t *buf, uint32_t len) {
   if (iface_num >= USBD_MAX_NUM_INTERFACES ||
       usb_ifaces[iface_num].type != USB_IFACE_TYPE_HID) {
-    return 0;
+    return -1;
   }
   return usb_emulated_read(&usb_ifaces[iface_num], buf, len);
 }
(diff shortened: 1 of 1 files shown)


core/embed/io/usb/unix/usb.c
```diff
@@ -216,7 +216,7 @@ static int usb_emulated_read(usb_iface_t *iface, uint8_t *buf, uint32_t len) {
     return len;
   }
 
-  return 0;
+  return -1;
```
TychoVrahe: This one open another can of worms i believe. This zero return means that there are no data to read which is 'retriable' error as discussed in previous PR. Furthermore, iiuc stm implementation also returns 0 in some cases when there is nothing to read. STM also seems to return zero if the buffer len is shorter than amount of data to read, which seems a bit weird. Anyway, i'd skip this one for now, seems like it would need a bit more digging.

TychoVrahe, line 219: This one open another can of worms i believe. This zero return means that there are no data to read which is 'retriable' error as discussed in previous PR. Furthermore, iiuc stm implementation also returns 0 in some cases when there is nothing to read. STM also seems to return zero if the buffer len is shorter than amount of data to read, which seems a bit weird. Anyway, i'd skip this one for now, seems like it would need a bit more digging.

M1nd3r: `-1` replaced by `0` here: https://github.com/trezor/trezor-firmware/pull/5437/commits/99127dcdea30772dba31f707d4592636c91ed718

PR #5425, opened 2025-07-25
Paginate everywhere

obrusvit:
This PR makes a small cleanup by utilizing `struct Pager` and `trait PaginateFull` (which is renamed to `Paginate` in the 2nd commit) where appropriate in `layout_bolt` and `layout_caesar`.

PR #5438, opened 2025-07-28
Use u16 for page count

obrusvit:
Follow-up on https://github.com/trezor/trezor-firmware/pull/5425

Using `u16` for page count information removes unnecessary casts on most places.

### DIFF
--- a/core/embed/rust/src/ui/layout_delizia/flow/continue_recovery_homepage.rs
+++ b/core/embed/rust/src/ui/layout_delizia/flow/continue_recovery_homepage.rs
@@ -223,27 +223,7 @@ pub fn new_continue_recovery_homepage(
         res.add_page(&ContinueRecoveryBeforeShares::Main, content_main)?
             .add_page(&ContinueRecoveryBeforeShares::Menu, content_menu)?;
         res
-    } else if pages.is_none() {
-        let content_menu = Frame::left_aligned(
-            TString::empty(),
-            VerticalMenu::empty().danger(theme::ICON_CANCEL, cancel_btn.into()),
-        )
-        .with_cancel_button()
-        .map(super::util::map_to_choice);
-
-        let mut res = SwipeFlow::new(&ContinueRecoveryBetweenShares::Main)?;
-        res.add_page(&ContinueRecoveryBetweenShares::Main, content_main)?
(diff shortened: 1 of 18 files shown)


romanz:
It seems that we are close to T3W1 bootloader flash limit (on `main`):
```
$ BOOTLOADER_DEVEL=1 TREZOR_MODEL=T3W1 QUIET_MODE=1 make -C core build_bootloader
...
           FLASH:        192 KB       192 KB    100.00%
...           
```

But this PR got us over the limit :(
```
/nix/store/n6qhvpsx5ldqazhmx5lc9lyjdbmq8zj2-gcc-arm-embedded-13.3.rel1/bin/../lib/gcc/arm-none-eabi/13.3.1/../../../../arm-none-eabi/bin/ld: region `FLASH' overflowed by 512 bytes
           FLASH:      197120 B       192 KB    100.26%
```
EDIT:
At https://github.com/trezor/trezor-firmware/actions/runs/16430385640/job/46430453198, it was:
```
           FLASH:        190 KB       192 KB     98.96%
```

core/embed/rust/src/ui/layout_delizia/flow/continue_recovery_homepage.rs
```diff
@@ -280,7 +280,7 @@ pub fn new_continue_recovery_homepage(
             ButtonRequestCode::Other,
             "show_shares".into(),
         ))
-        .with_pages(move |_| n_remaining_shares);
+        .with_pages(move |_| n_remaining_shares as u16);
```
bieleluk:
```suggestion
        .with_pages(move |_| n_remaining_shares);
```
A better place for casting would be
`let n_remaining_shares = pages.as_ref().unwrap().len() as u16 / 2;`

obrusvit:
I'm surprised that this PR tipped us over the t3w1 bld flash limit. :thinking:

core/embed/rust/src/ui/layout_delizia/flow/continue_recovery_homepage.rs
obrusvit: I did this and also I realized that the `unwrap()` can panic so I reverted the `else if`/`else` logic

https://github.com/trezor/trezor-firmware/pull/5438/commits/4fd32b28fc3383f668e0f5fbfdcbdcd3260cf2b9

PR #5441, opened 2025-07-28
SLIP-24 UI bug fixes

ibz:
[no changelog]


core/src/trezor/ui/layouts/caesar/__init__.py
```diff
@@ -1163,7 +1172,7 @@ async def confirm_ethereum_payment_request(
             amount_label=None,
             fee=maximum_fee,
             fee_label=TR.words__transaction_fee,
-            title=TR.words__title_summary,
+            title=(TR.words__swap if is_swap else TR.words__title_summary),
```
bieleluk: I'm not sure if this change is desired. It is similar to one we discussed on [Slack](https://satoshilabs.slack.com/archives/C07JWCTDWEA/p1750950032595319)

core/src/trezor/ui/layouts/delizia/__init__.py
```diff
@@ -1138,7 +1157,7 @@ async def confirm_ethereum_payment_request(
             None,
             maximum_fee,
             TR.words__transaction_fee,
-            TR.words__title_summary,
+            TR.words__swap if is_swap else TR.words__title_summary,
```
bieleluk: I'm not sure if this change is desired. It is similar to one we discussed on [Slack](https://satoshilabs.slack.com/archives/C07JWCTDWEA/p1750950032595319)

core/src/trezor/ui/layouts/eckhart/__init__.py
```diff
@@ -1081,15 +1081,16 @@ async def confirm_trade(
             back_button=back_button,
         )
 
+        account_info = [(str(""), address)]
```
bieleluk, line 1084: nit: using empty string without explicit cast is probably enough:
```suggestion
        account_info = [("", address)]
```

core/src/trezor/ui/layouts/eckhart/__init__.py
```diff
@@ -1139,14 +1142,17 @@ async def confirm_ethereum_payment_request(
 
         menu_items = [create_details(TR.address__title_provider_address, recipient)]
         for r_address, r_account, r_account_path in refunds:
+            refund_account_info = [(str(""), r_address)]
```
bieleluk, line 1145: nit: using empty string without explicit cast is probably enough:
```suggestion
            refund_account_info = [("", r_address)]
```

tests/device_tests/ethereum/test_signtx.py
```diff
@@ -539,7 +559,8 @@ def test_signtx_staking_eip1559(client: Client, parameters: dict, result: dict):
     reason="T1 does not support payment requests. Payment requests not yet implemented on model T.",
 )
 @pytest.mark.parametrize(
-    "has_refund,has_text,has_multiple_purchases", list(product([True, False], repeat=3))
+    "has_refund,has_text,has_multiple_purchases",
+    list(product([True, False], repeat=3)),
 )
 def test_signtx_payment_req(
```
bieleluk, line 565: The delizia test doesn't traverse the menu in the summary screen. That's a bug. 
image
The first two screens in Delizia don't traverse the menu, even though they contain only the _Cancel_ option. I'm not sure if it's a bug or a feature in this case.
image

core/src/trezor/ui/layouts/caesar/__init__.py
ibz: Reverted a3760fd8781d1cf4321aa979b24cc2001cd0dfe0

core/src/trezor/ui/layouts/delizia/__init__.py
ibz: Reverted a3760fd8781d1cf4321aa979b24cc2001cd0dfe0

core/src/trezor/ui/layouts/eckhart/__init__.py
ibz: Unfortunately that will not be accepted by the type checker because it thinks that `""` is `Literal('')` rather than `str` which makes it think that we are dealing with a list of `tuple[Literal(''), str]` and it will complain when we add a `(str, str)` to that list :(

core/src/trezor/ui/layouts/eckhart/__init__.py
ibz: I guess I could get rid of this if I insert the other elements first, then insert this on position 0... Not sure it is worth...

tests/device_tests/ethereum/test_signtx.py
ibz: > The delizia test doesn't traverse the menu in the summary screen. That's a bug.

Indeed, the Delizia `confirm_summary` screen has not been migrated to the new menu system yet, it uses the old `account_items` and `fee_info_items` mechanism. IMO it is a separate issue that will eventually be solved...

ibz: > The first two screens in Delizia don't traverse the menu, even though they contain only the _Cancel_ option. I'm not sure if it's a bug or a feature in this case.

I would say it's both a bug and a feature, if that makes any sense...

In this case we are using `confirm_value` with no external menu, which defaults to its internal menu, which is not traversed by the tests.

I *could* if you think it makes sense to pass it an external menu which has just the **Cancel** - and then it will open that menu and take a screenshot. IDK if it's worth it TBH, although I guess that *eventually* we will get to making it traverse this menu anyway, in some way or another, it is currently in an in between state which is what makes it strange, but so are most menus in the firmware at this point...

core/src/trezor/ui/layouts/eckhart/__init__.py
bieleluk: In that case, it's probably better to explicitly mark the type:
`account_info: list[tuple[str, str]] = [("", address)]`
But I'm also fine with the current implementation.

core/src/trezor/ui/layouts/eckhart/__init__.py
bieleluk: In that case, it's probably better to explicitly mark the type:
`refund_account_info: list[tuple[str, str]] = [("", r_address)]`
But I'm also fine with the current implementation.

tests/device_tests/ethereum/test_signtx.py
bieleluk: If we know about it, it's fine. I pointed that out because the behavior was different from what I expected.

PR #5444, opened 2025-07-29
fix nrf update

TychoVrahe:
This PR introduces few nrf update related improvements and fixes. 

First, in SMP implementation receiver release is added where missing.

After updating nRF, reset is done via HW rather than via SMP command as it is more reliable.

Lastly, nrf and bootloader updating scripts are modified to exit prodtest interactive mode before proceeding with the update, as they are not able to work in that mode.

### DIFF
--- a/core/tools/bld_update.py
+++ b/core/tools/bld_update.py
@@ -16,6 +16,12 @@ def _compress(data: bytes) -> bytes:
     return compressor.compress(data) + compressor.flush()
 
 
+def exit_interactive_mode(ser):
+    ser.write(("." + "\r\n").encode())
+    time.sleep(0.1)
+    ser.reset_input_buffer()
+
+
 def send_cmd(ser, cmd, expect_ok=True):
     """Send a line, read response, and abort on CLI_ERROR."""
     ser.write((cmd + "\r\n").encode())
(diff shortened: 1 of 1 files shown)


core/tools/bld_update.py
```diff
@@ -51,6 +57,8 @@ def upload_bootloader(port, bin_path, chunk_size):
     ser.reset_input_buffer()
     ser.reset_output_buffer()
 
+    exit_interactive_mode(ser)
+
     # 1) Begin transfer
     send_cmd(ser, "bootloader-update begin")
```
kopecdav: Script is working well, but I think that we should also close the serial at the end of the script or on Exception to be sure

core/tools/nrf_update.py
```diff
@@ -36,6 +42,8 @@ def upload_nrf(port, bin_path, chunk_size):
     ser.reset_input_buffer()
     ser.reset_output_buffer()
 
+    exit_interactive_mode(ser)
```
kopecdav: Close serial at the end.

kopecdav (changes requested):

core/tools/nrf_update.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/689110417ea784b5e36952f5ed2b86875001bce9

core/tools/bld_update.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/e4c9b377609ad0eb8364644ca94be973cb06498f

kopecdav (approved):

PR #5445, opened 2025-07-29
fix prodtest bootloader-update command (T3W1)

cepetr:
This PR fixes the prodtest `bootloader-update` command and allows uploading a bootloader with the new PQC header on T3W1.

1) The `bootloader-update` command writes the new bootloader and boot header to the dedicated flash area (`BOOTUPDATE_AREA`). The update is finalized in the boardloader, so a `reboot` command is required after a successful upload.

2) The `bld_update.py` script was updated to avoid compressing the binary by default.


core/embed/sys/mpu/stm32u5/mpu.c
```diff
@@ -450,6 +450,9 @@ mpu_mode_t mpu_reconfig(mpu_mode_t mode) {
     case MPU_MODE_BOOTUCB:
       SET_REGION( 6, BOOTUCB_START,            BOOTUCB_MAXSIZE,    FLASH_DATA,  YES,    NO );
       break;
+    case MPU_MODE_BOOTUPDATE:
+      SET_REGION( 6, BOOTUPDATE_START,         BOOTUPDATE_MAXSIZE,   FLASH_DATA,  YES,    NO );
```
TychoVrahe: please fix alignment at the end of this line

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
```diff
@@ -83,6 +83,37 @@ __attribute__((
...
+
+  if (sectrue != flash_area_erase(&BOOTUPDATE_AREA, NULL)) {
+    return false;
+  }
+
+  if (sectrue != flash_unlock_write()) {
+    return false;
+  }
+
+  for (uintptr_t offset = 0; offset < size; offset += FLASH_BLOCK_SIZE) {
```
TychoVrahe: why not use flash_area_write_data?

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
```diff
@@ -145,12 +176,31 @@ static void prodtest_bootloader_update(cli_t *cli) {
       return;
     }
 
+#if USE_BOOT_UCB
+    mpu_mode_t mpu_mode = mpu_reconfig(MPU_MODE_BOOTUPDATE);
+
+    if (!write_to_bootupdate_area(bootloader_buffer, bootloader_len)) {
+      cli_error(cli, CLI_ERROR, "Failed to flash bootloader");
+      return;
```
TychoVrahe, line 181: mpu_restore before returning

core/tools/bld_update.py
```diff
@@ -83,15 +86,20 @@ def upload_bootloader(port, bin_path, chunk_size):
     show_default=True,
     help="Max bytes per chunk (in compressed form)",
 )
-def main(port, binary, chunk_size):
+@click.option(
```
TychoVrahe: I dont like using the argument for this - its specifically bound to model so one can't really choose but has to know which model requires what. We could implement a prodtest command to get model code and decide based on that. But we can also do this later.

TychoVrahe (changes requested):

core/tools/bld_update.py
cepetr: How about simply checking the first four bytes of the binary file?

TychoVrahe: right, that seems simpler

core/embed/sys/mpu/stm32u5/mpu.c
cepetr: fixed, see 5f8848a2de5c43b3ef2440b73d3a64fabb748749

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
cepetr: fixed, see 17b3051b3630c60273b2d58f40f1fb99f22427f8

core/embed/projects/prodtest/cmd/prodtest_bootloader.c
cepetr: fixed, see 9852520d3b28eaa03825b7b56f16e1d9d58bfdc9

core/tools/bld_update.py
cepetr: Ok, how about this 7061a3b3578618b4c42c4d657ad2fea716bbf011 ?

TychoVrahe (approved):

TychoVrahe (approved):

PR #5447, opened 2025-07-30
ble: stop advertising when suspended

TychoVrahe:
This PR fixes issue with advertising still running in suspend mode, if some device was connected before suspending and there is more than one device bonded.


kopecdav (approved):
utACK

PR #5449, opened 2025-07-30
External menu on Caesar `confirm_properties`

ibz:


core/embed/rust/src/ui/layout_caesar/component/page.rs
```diff
@@ -23,8 +23,10 @@ where
     pad: Pad,
     cancel_btn_details: Option<ButtonDetails>,
     confirm_btn_details: Option<ButtonDetails>,
+    info_btn_details: Option<ButtonDetails>,
```
romanz: Just curious if it needs to be an `Option`?

core/embed/rust/src/ui/layout_caesar/ui_firmware.rs
```diff
@@ -1466,13 +1477,14 @@ fn content_in_button_page<T: Component + Paginate + MaybeTrace + 'static>(
     } else {
         None
     };
-    if hold {
+    if hold && !external_menu {
```
romanz, line 1480: Can we support `hold` with an `external_menu`?

romanz (commented):
LGTM, with a few questions:

core/embed/rust/src/ui/layout_caesar/ui_firmware.rs
ibz: I don't think so, because there is no such thing as middle hold-to-confirm.

But also, every hold-to-confirm on TS3 will become middle button press, see #4622.

So `hold` will disappear...

core/embed/rust/src/ui/layout_caesar/component/page.rs
ibz: I think it didn't even need to exist... 6848ec68456e27bdf831e9982bc9eabbc6e03ad0 :)

PR #5452, opened 2025-07-30
Improve menu interaction typing

romanz:
Related to the discussion here: https://github.com/trezor/trezor-firmware/pull/5349#issuecomment-3069192726

### DIFF
--- a/core/src/trezor/ui/layouts/menu.py
+++ b/core/src/trezor/ui/layouts/menu.py
@@ -102,7 +102,7 @@ async def show_menu(
             return
 
 
-async def confirm_with_menu(
+async def interact_with_menu(
     main: trezorui_api.LayoutObj[T],
     menu: Menu,
     br_name: str | None,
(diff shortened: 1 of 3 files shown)


core/src/trezor/ui/layouts/menu.py
```diff
@@ -116,3 +116,15 @@ async def confirm_with_menu(
...
+
+async def confirm_with_menu(
+    main: trezorui_api.LayoutObj[T],
+    menu: Menu,
+    br_name: str | None,
+    br_code: ButtonRequestType = ButtonRequestType.Other,
+    raise_on_cancel: ExceptionType = ActionCancelled,
+) -> None:
+    result = await interact_with_menu(main, menu, br_name, br_code, raise_on_cancel)
+    # use this function when the layout may only return CONFIRMED on success (or raise an exception)
```
ibz, line 134: Can we edit this comment to mention that the concern is, for example **`BACK`** disrupting the `CONFIRMED` / `ActionCancelled` combo?

Also, this comment sounds more like it should be a docstring, no?

ibz (approved):

core/src/trezor/ui/layouts/menu.py
romanz: Fixed in 732e03724e

romanz:
Rebasing over `main`.

PR #5454, opened 2025-07-31
refactor debug overlay

TychoVrahe:
This PR refactors the debug overlay.

First, debug overlay, the one that shows performace stats was renamed to performance overlay, so that its less confusing.

Debug overlay now means displaying the red square in corner of the display.

Rendering debug overlay was moved to rust and made UI-dependent, so that we can have different implementation based on model needs - like shifting towards center a bit on eckhart so its not in the invisible area. Moving to rust also allows to remove some display functions from python entirely which is nice. The screenshoting mechanism is now partially in rust too, so some modifications were made to make this possible, mainly storing the screen recording data in accessible place.

Additionally, some small fixes were made to Eckhart performance overlay - fix compilation in bootloader, fix positioning.

This change also means that stuff rendered directly in rust (like fatal error) is also recorded in tests - which is causing an additional screen to appear in Wipe code activate persistence test. Seems like a good thing. Eckhart UI for this screen looks wrong but thats separate issue.


core/embed/io/display/display_utils.c
```diff
@@ -34,3 +34,72 @@ void display_fade(int start, int end, int delay) {
...
+  uint8_t target_directory[256];
+  int refresh_index;
+
+} display_recording_t;
+
+static display_recording_t g_display_recording = {0};
+
+void display_record_start(uint8_t *target_dir, size_t target_dir_len,
+                          int refresh_index) {
+  display_recording_t *drv = &g_display_recording;
```
cepetr: Maybe we should use `rec` or something similar instead of `drv` here? I was a bit confused when I saw this in the context of the display. Maybe `rec->activate` would be more appropriate .

core/embed/io/display/inc/io/display.h
```diff
@@ -157,6 +157,12 @@ void display_copy_mono1p(const gfx_bitblt_t *bb);
 #ifdef TREZOR_EMULATOR
 // Save the screen content to a file.
 // The function is available only on the emulator.
-const char *display_save(const char *prefix);
-void display_clear_save(void);
+void display_save(const char *prefix);
+
 #endif
+
+void display_record_start(uint8_t *target_dir, size_t target_dir_len,
```
cepetr: If the new functions are in `display_utils.c`, their declarations should go in `display_utils.h`.

core/embed/rust/src/ui/shape/display/fb_rgba8888.rs
```diff
@@ -73,19 +72,31 @@ where
 
         let mut target = ScopedRenderer::new(DirectRenderer::new(&mut canvas, bg_color, &cache));
 
+        #[cfg(all(feature = "ui_debug_overlay", not(feature = "ui_performance_overlay")))]
```
cepetr, line 75: I wonder if renaming `DebugOverlay` to `PerformanceOverlay` was necessary. The original idea behind `DebugOverlay` was to allow displaying various kinds of information (not just FPS-related), including things like the red dot. The assumption was that it would always be enabled in debug mode, and we could configure what it shows.

Perhaps adding a recording field to `DebugOverlay` could have been enough. The rendering function (originally `render_debug_overlay`) could then decide what to draw based on the model and settings.

With both `DebugOverlay` and `PerformanceOverlay` now in place, I’m a bit concerned that if we want to add more info in the future, we might end up with separate overlays for each case.

TychoVrahe: I think its neccessary(or rather strongly prefered) to at least rename the build feature (unless you'd want to abandon the red square entirely) - we need to differentiate what you want shown somehow, and performance overlay is way more precise for the FPS display stuff. 

We could use the same Component/rendering method for both stuff and customize within it at compile time, but that just doesn't seem correct: it limits the component usability, i.e. prevents reusing the component outside of this context, selectively or whatever.

What i don't like about current situation is actually that adding different overlays is rather ackward, and has to be done for every rendered. This would definitely be a thing to improve if we want to add more stuff. And then just specify which overlays to display and some automated way to position them on the screen...

cepetr: > I think its neccessary(or rather strongly prefered) to at least rename the build feature (unless you'd want to abandon the red square entirely) - we need to differentiate what you want shown somehow, and performance overlay is way more precise for the FPS display stuff.

Yes, I agree with this.

> We could use the same Component/rendering method for both stuff and customize within it at compile time, **but that just doesn't seem correct:** it limits the component usability, i.e. prevents reusing the component outside of this context, selectively or whatever.

That's where we see things differently - I see it as a universal debug overlay, shown when a debug version is built.

--
But I’m really not insisting on it - we can change it  if another need arises.

core/embed/io/display/display_utils.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/0906d404a3a0f9bda39c03a194afa6d68ef8adaf

core/embed/io/display/inc/io/display.h
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/0906d404a3a0f9bda39c03a194afa6d68ef8adaf

core/embed/rust/src/ui/shape/display/fb_rgba8888.rs
TychoVrahe: Yeah lets rethink this when we have more things to add to the overlays - i expect that the path forward will become clearer

cepetr (approved):
Tested on T3W1

PR #5456, opened 2025-07-31
Fix interactive test mode

bieleluk:


romanz (approved):
Thanks!

python/src/trezorlib/debuglink.py
```diff
@@ -975,7 +976,8 @@ def default_input_flow(self) -> InputFlowType:
 
     def _visit_menu_items(self) -> LayoutContent:
         layout = self.debuglink.read_layout()
-        if not layout.has_menu():
+        interact = os.environ.get("INTERACT") == "1"
```
romanz: nit: maybe extract into a helper function (since it also used in other places)?
https://github.com/search?q=repo%3Atrezor%2Ftrezor-firmware%20os.environ.get(%22INTERACT%22)&type=code

python/src/trezorlib/debuglink.py
```diff
@@ -18,6 +18,7 @@
 
 import json
 import logging
+import os
```
romanz: can be removed

PR #5458, opened 2025-07-31
shell.nix: workaround ERROR: SDL_CreateRenderer error

mmilata:
Emulator can't be started on default Ubuntu 25.04 install. Let's see if it resolves in time or if we need to fix our code (or rewrite it to use SDL3, not sure if it doesn't break our library of old emulator binaries).


M-pasta:
can confirm, it solved the problem

PR #5464, opened 2025-08-01
chore(core): use non-breakable space between value and unit

bieleluk:
This PR:
- Changes space character into a non-breakable space  _\u00A0_ in `value [unit]`  scenarios to avoid unit rendering on a new line.
- [Eckhart]: Makes hyphen and ellipsis color one shade darker for better readability (similar to other layouts)


core/src/trezor/strings.py
```diff
@@ -21,6 +21,13 @@ def format_amount(amount: int, decimals: int) -> str:
     return s
 
 
+def format_amount_unit(amount: str, unit: str) -> str:
+    """
+    Formats an amount with a unit, e.g. "1.23 BTC".
+    """
+    return f"{amount} {unit}"
```
romanz, line 28: nit: maybe add a comment that the separator is a special character, or use "\u0010"?
(currently, it seems to be rendered as a regular space in GitHub and possibly also in other editors...)
image

bieleluk: https://github.com/trezor/trezor-firmware/pull/5464/commits/26ea073a04fe2711cd2542e0928255f1158ffb83

romanz (approved):

bieleluk:
The PR didn't catch all cases so there is a follow-up : https://github.com/trezor/trezor-firmware/issues/5502

PR #5467, opened 2025-08-02
fix(python): avoid failing `emu.py` if non-debug emulator is used

romanz:
Otherwise, it fails nightly core CI:
https://github.com/trezor/trezor-firmware/actions/runs/16687051058/job/47238433930

Tested with:
```
$ TREZOR_MODEL=T2T1 PYOPT=1 QUIET_MODE=1 make -C core build_unix_frozen test_emu_sanity 
```


mmilata (approved):
Ah, my bad for introducing the ping check ...

PR #5471, opened 2025-08-03
Add and update GNU licence headers

M1nd3r:


python/setup.py
```diff
@@ -2,7 +2,7 @@
 
 # This file is part of the Trezor project.
 #
-# Copyright (C) 2012-2022 SatoshiLabs and contributors
```
romanz, line 5: Not sure we need to update those: https://hynek.me/til/copyright-years/

M1nd3r: WDYT @matejcik ?

Maybe we could/should switch to the "year of creation" instead of year ranges?

M1nd3r: After internal discussion, years were removed completely from license headers.

python/tests/test_protobuf_encoding.py
```diff
@@ -165,7 +165,7 @@ def test_simple_message():
         uvarint=12345678910,
         svarint=-12345678910,
         bool=True,
-        bytes=b"\xDE\xAD\xCA\xFE",
+        bytes=b"\xde\xad\xca\xfe",
```
romanz: Please undo

python/tests/test_protobuf_encoding.py
```diff
@@ -177,7 +177,7 @@ def test_simple_message():
     assert retr.uvarint == 12345678910
     assert retr.svarint == -12345678910
     assert retr.bool is True
-    assert retr.bytes == b"\xDE\xAD\xCA\xFE"
+    assert retr.bytes == b"\xde\xad\xca\xfe"
```
romanz: Please undo

romanz (approved):

python/tests/test_protobuf_encoding.py
M1nd3r: undone

python/tests/test_protobuf_encoding.py
M1nd3r: undone

romanz (approved):

PR #5473, opened 2025-08-04
boarloader update from prodtest

TychoVrahe:


core/embed/projects/prodtest/cmd/common.c
```diff
@@ -432,3 +454,108 @@ bool check_cert_chain(cli_t* cli, const uint8_t* chain, size_t chain_size,
...
+      binary_len = 0;
+      cli_error(cli, CLI_ERROR, "Error while finalizing the update", phase);
+      goto cleanup;
+    }
+
+    // Reset state so next begin must come before chunks
+    binary_len = 0;
+    binary_update_in_progress = false;
+
+    cli_trace(cli, "Update successful (%u bytes)", (unsigned)binary_len);
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/5f62d42238b3bda4c47fe53ee6c1b72e8e8b710d

core/embed/projects/prodtest/cmd/common.c
```diff
@@ -432,3 +454,108 @@ bool check_cert_chain(cli_t* cli, const uint8_t* chain, size_t chain_size,
...
+    }
+
+    if (binary_len == 0) {
+      cli_error(cli, CLI_ERROR, "No data received");
+      goto cleanup;
+    }
+
+    if (!finalize(binary_buffer, binary_len)) {
+      binary_len = 0;
+      cli_error(cli, CLI_ERROR, "Error while finalizing the update", phase);
```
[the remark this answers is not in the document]
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/5f62d42238b3bda4c47fe53ee6c1b72e8e8b710d

core/tools/bin_update.py
```diff
@@ -45,40 +61,47 @@ def upload_nrf(port, bin_path, chunk_size):
...
 
 
 @click.command(context_settings={"help_option_names": ["-h", "--help"]})
 @click.argument("port", metavar="<serial-port>")
 @click.argument(
-    "binary", metavar="<nrf-binary>", type=click.Path(exists=True, dir_okay=False)
+    "binary", metavar="<binary>", type=click.Path(exists=True, dir_okay=False)
 )
 @click.option(
     "--chunk-size", "-c", default=512, show_default=True, help="Max bytes per chunk"
```
kopecdav, line 87: --chunk-size should be limited to max 1024, otherwise prodtest will clip all chunks, but binary update will still finish with OK.

core/tools/bin_update.py
```diff
@@ -6,6 +6,14 @@
 import click
 import serial
 
+UPDATE_TYPES = {
+    "boardloader": {
+        "command": "boardloader-update",
```
kopecdav, line 11: Isn't bootloader missing in the list?

core/embed/projects/prodtest/cmd/common.c
```diff
@@ -432,3 +454,107 @@ bool check_cert_chain(cli_t* cli, const uint8_t* chain, size_t chain_size,
...
+      cli_error(cli, CLI_ERROR, "Update not started. Use 'begin' first.");
+      goto cleanup;
+    }
+
+    // Receive next piece of the image
+    size_t chunk_len = 0;
+    // Temporary buffer for this chunk
+    uint8_t chunk_buf[1024];
+
+    if (!cli_arg_hex(cli, "hex-data", chunk_buf, sizeof(chunk_buf),
```
kopecdav: It looks like the chunk will be clipped without notice in case if it is larger then 1024B. Not sure if there is some easy and clean way to recognize this, probably better to deal with this in the python script instead. 

But we should at least document that in README.md

cepetr: Maybe we don't need any buffer on the stack and parse incoming data directly into `binary_buffer`. Something like this:
```c 
    if (!cli_arg_hex(cli, "hex-data", &binary_buffer[binary_len], sizeof(binary_buffer) - binary_len,
                     &chunk_len)) {
      cli_error_arg(cli, "Expecting hex data for chunk.");
      goto cleanup;
    }
```

core/embed/projects/prodtest/cmd/prodtest_boardloader.c
```diff
@@ -33,6 +38,41 @@ static void prodtest_boardloader_version(cli_t* cli) {
...
+  if (res != sectrue) {
+    goto cleanup;
+  }
+
+  res = flash_area_write_data_padded(&BOARDLOADER_AREA, 0, data, len, 0xFF,
+                                     flash_area_get_size(&BOARDLOADER_AREA));
+
+cleanup:
+
+  (void)!flash_lock_write();
```
cepetr, line 62: extra `!` ?

core/embed/projects/prodtest/cmd/prodtest_boardloader.c
```diff
@@ -41,3 +81,12 @@ PRODTEST_CLI_CMD(
   .info = "Retrieve the boardloader version",
   .args = ""
 );
+
+#if !PRODUCTION && !TREZOR_MODEL_T2T1
```
cepetr, line 85: Shouldn't we add T2B1 as well, or change the condition to STM32F4 or similar?

cepetr (changes requested):

core/embed/projects/prodtest/cmd/common.c
TychoVrahe: changed in https://github.com/trezor/trezor-firmware/commit/d17ee1eeca786ce1605568d1bff0eae2bd3f6d9a

core/tools/bin_update.py
TychoVrahe: moot as of https://github.com/trezor/trezor-firmware/commit/d17ee1eeca786ce1605568d1bff0eae2bd3f6d9a

core/tools/bin_update.py
TychoVrahe: no, bootloader update has separate script - it has some additional stuff with compression and i was lazy to unify that

core/embed/projects/prodtest/cmd/prodtest_boardloader.c
TychoVrahe: `(void)!` discards the warning produced by `__wur`  marked function when result is unused. its a bit weird i admit, but its already used on few other places

core/embed/projects/prodtest/cmd/prodtest_boardloader.c
TychoVrahe: on T2B1 the buffer fits the RAM, so no need to disable this. T2T1 uses more RAM for partial framebuffers and so.

TychoVrahe:
@kopecdav i added the readme in https://github.com/trezor/trezor-firmware/commit/9d883ee71cdc51e5ec52f4ae44cda169a65375fa and updated he nrf a bootloader in https://github.com/trezor/trezor-firmware/commit/7137991b9d797065ac1e2c37a9ee501b0cd19ae7. Chunk size is moot, as these are meant as internal only i don't think we need to go into more detail

core/embed/projects/prodtest/cmd/prodtest_boardloader.c
cepetr: Ok, I thought the reason is mpu configuration. Shouldn't we reconfigure the mpu on the stm32f4 as well?

core/embed/projects/prodtest/cmd/prodtest_boardloader.c
cepetr: Interesting, I thought `(void)` does the same, but it really doesn't.

core/embed/projects/prodtest/cmd/prodtest_boardloader.c
TychoVrahe: added https://github.com/trezor/trezor-firmware/commit/510c9220a78e6710e8003f8d67243a19d75b6205

kopecdav (approved):

cepetr (approved):
utACK

PR #5474, opened 2025-08-04
Slip-24 UI for Bitcoin

ibz:


core/.changelog.d/5450.added
```diff
@@ -0,0 +1 @@
+Implement new payment requests UI for BTC.
```
bieleluk: Nit: word _implement_ sounds too technical from the user's point of view:
```suggestion
New payment requests UI for BTC.
```

core/src/apps/bitcoin/sign_tx/approvers.py
```diff
@@ -90,20 +89,23 @@ async def _add_output(self, txo: TxOutput, script_pubkey: bytes) -> None:
         self.total_out += txo.amount
 
     async def add_payment_request(
-        self, msg: PaymentRequest, keychain: Keychain
+        self,
+        payment_request: PaymentRequest,
+        keychain: Keychain,
+        tx_info: TxInfo | None,
```
bieleluk: `tx_info` and `txo` are unused. Is it intentional to maintain consistency in the API?

tests/device_tests/bitcoin/test_signtx_payreq.py
```diff
@@ -36,12 +42,89 @@
...
 
 
-def case(id, *args, altcoin: bool = False, models: str | None = None):
+@pytest.mark.models(
+    "core",
+    skip="t2t1",
+    reason="T1 does not support payment requests. Payment requests not yet implemented on model T.",
+)
+@pytest.mark.altcoin
+@pytest.mark.parametrize("has_refund,has_text", list(product([True, False], repeat=2)))
```
bieleluk: Nit: if the test name is _with_text_and_refund_, it is more logical to use the input vector in this order.
```suggestion
@pytest.mark.parametrize("has_text,has_refund", list(product([True, False], repeat=2)))
```

bieleluk (approved):
Approved with minor suggestions

core/src/apps/bitcoin/sign_tx/approvers.py
ibz: Yeah, I don't think we should change the method signature in derived classes, that's why I kept the parameters. I prefixed them with `_`: 1c319c1aee49e19317421236c13080539bf5f1f0

tests/device_tests/bitcoin/test_signtx_payreq.py
ibz: Done. I moved the code around as well! a5843cfe6720b170fccd4e341fce706f0fa7d3ac

PR #5476, opened 2025-08-04
Raise exception for invalid ThpCreateNewSession messages 

M1nd3r:
Raise exception on receiving `ThpCreateNewSession` message with `passphrase` field not equal to `None` when Trezor has `passphrase_always_on_device` set to `True`.

### DIFF
--- a/core/src/apps/common/passphrase.py
+++ b/core/src/apps/common/passphrase.py
@@ -10,16 +10,32 @@ _MAX_PASSPHRASE_LEN = const(50)
 if TYPE_CHECKING:
     from trezor.messages import ThpCreateNewSession
 
+if __debug__:
+    from trezor import log
+
 
 def is_enabled() -> bool:
     return storage_device.is_passphrase_enabled()
 
 
 async def get_passphrase(msg: ThpCreateNewSession) -> str:
-    if not is_enabled():
-        return ""
+    passphrase_always_on_device = storage_device.get_passphrase_always_on_device()
 
-    if msg.on_device or storage_device.get_passphrase_always_on_device():
+    # Device setting "disabled passphrase protection" is ignored
+    if __debug__:
+        if not is_enabled() and msg.passphrase:
+            log.warning(
+                __name__,
+                "Creating new session with passphrase, ignoring device settings.",
+            )
+
+    # When always_on_device is True, messages with passphrase raise a DataError
+    if passphrase_always_on_device and msg.passphrase is not None:
+        raise DataError(
+            "Providing passphrase in message is not allowed when PASSPHRASE_ALWAYS_ON_DEVICE is True."
+        )
+
+    if msg.on_device or passphrase_always_on_device:
         passphrase = await _get_on_device()
     else:
         passphrase = msg.passphrase or ""
--- a/core/src/trezor/wire/thp/session_manager.py
+++ b/core/src/trezor/wire/thp/session_manager.py
@@ -23,17 +23,6 @@ def get_new_session_context(
     return SessionContext(channel_ctx, session_cache)
 
 
-def get_new_seedless_session_ctx(
-    channel_ctx: Channel, session_id: int
-) -> SeedlessSessionContext:
-    """
-    Creates new `SeedlessSessionContext` that is not backed by a cache entry.
-
-    Seed cannot be derived with this type of session.
-    """
-    return SeedlessSessionContext(channel_ctx, session_id)
-
-
 def get_session_from_cache(
     channel_ctx: Channel, session_id: int
 ) -> GenericSessionContext | None:
--- a/tests/device_tests/test_session.py
+++ b/tests/device_tests/test_session.py
@@ -18,17 +18,20 @@ import pytest
 
 from trezorlib import cardano, exceptions, messages, models
 from trezorlib.client import ProtocolVersion
+from trezorlib.debuglink import SessionDebugWrapper as Session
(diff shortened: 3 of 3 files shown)


core/src/apps/common/seed.py
```diff
@@ -87,7 +87,7 @@ async def derive_and_store_roots(
             if not storage_device.is_initialized():
                 raise wire.NotInitialized("Device is not initialized")
 
-            passphrase = await get_passphrase(msg)
+            passphrase = "" if msg.passphrase is None else await get_passphrase(msg)
```
matejcik: should you even be here if `msg.passphrase is None`?

core/src/apps/common/passphrase.py
```diff
@@ -17,7 +17,7 @@ def is_enabled() -> bool:
 
 async def get_passphrase(msg: ThpCreateNewSession) -> str:
     if not is_enabled():
-        return ""
+        raise DataError("Passphrase protection is disabled")
```
matejcik: this should be `if not is_enabled() and msg.passphrase != ""`

matejcik: (or probably: `if not is_enabled -> if msg.passphrase != "" then raise; else return ""` )

M1nd3r: I am unsure why? I thought that I call this function only when I want to extract passphrase from the ThpCreateNewSession message - and the case when no passphrase is provided is handled by the
` passphrase = "" if msg.passphrase is None else await get_passphrase(msg)`
in `seed.py`.

Do you want to call the function `get_passphrase(msg: ThpCreateNewSession)` when the message in the parameter does not contain passphrase in the first place?

core/src/apps/common/seed.py
M1nd3r: I guess so? This is the case when passphrase protection is disabled 🤔

M1nd3r: My idea is that if the Host does not want session with passphrase, they just send ThpCreateNewSession without passphrase.

M1nd3r:
@matejcik 

As I see the test failing, imo it is not worth to rewrite them and change the behaviour (strictlyrequire passphrase = None when passphrase protection is disabled). So I guess I should allow passphrase="" and go with your solution mentioned in the comments above.

core/src/apps/common/seed.py
M1nd3r: changed here: https://github.com/trezor/trezor-firmware/pull/5476/commits/c9ae236ae82662b0c7fbbb4ad665ea3488be1c84

core/src/apps/common/passphrase.py
M1nd3r: adjusted here: https://github.com/trezor/trezor-firmware/pull/5476/commits/c9ae236ae82662b0c7fbbb4ad665ea3488be1c84

Note: I am handling both _passphrase==""_ and _passphrase==None_ as valid when passphrase protection is disabled.

core/src/apps/common/passphrase.py
```diff
@@ -17,6 +17,9 @@ def is_enabled() -> bool:
 
 async def get_passphrase(msg: ThpCreateNewSession) -> str:
     if not is_enabled():
+        if msg.passphrase:
+            # passphrase is not None/empty
+            raise DataError("Passphrase protection is disabled")
         return ""
 
     if msg.on_device or storage_device.get_passphrase_always_on_device():
```
matejcik, line 25: while we're here:
this should be `and msg.passphrase is None`

i recall specifying the logic of the function somewhere:
1. `if passphrase is not None and not on_device`: use contents of `passphrase` field
2. `if passphrase is None and on_device`: ask the user
3. `if passphrase is None and not on_device`: create a seedless session
4. `if passphrase is not None and on_device`: raise an error

core/src/apps/common/passphrase.py
```diff
@@ -10,16 +10,34 @@
...
+    # Device setting "disabled passphrase protection" is ignored
+    if __debug__:
+        if not is_enabled() and msg.passphrase:
+            log.warning(
+                __name__,
+                "Creating new session with passphrase, ignoring device settings.",
+            )
+
+    # When always_on_device is True, messages with passphrase raise a DataError
+    if is_always_on_device() and msg.passphrase is not None:
```
romanz: nit: call `get_passphrase_always_on_device()` once and reuse the result:
```suggestion
    is_always_on_device = storage_device.get_passphrase_always_on_device()
    if is_always_on_device and msg.passphrase is not None:
```

romanz (approved):

core/src/apps/common/passphrase.py
```diff
@@ -10,16 +10,34 @@
...
+        raise DataError(
+            "Providing passphrase in message is not allowed when PASSPHRASE_ALWAYS_ON_DEVICE is True."
```
romanz, line 35: Please add a device test for the new error.

core/src/apps/common/passphrase.py
M1nd3r: Changed here: https://github.com/trezor/trezor-firmware/pull/5476/commits/8eb2179e473dbc2955497f4f4c326b09f537e189

core/src/apps/common/passphrase.py
M1nd3r: added here: https://github.com/trezor/trezor-firmware/pull/5476/commits/ab65214906ed2b55cdac765a44d127e65d737515

PR #5479, opened 2025-08-04
fix(core): execute `race()` callback immediately

romanz:
It should avoid scheduling delay when using `loop.race()`.


romanz:
Tested separately, as it will be used in #5546 (https://github.com/trezor/trezor-firmware/pull/5546/commits/ca3212d56cac91133716513bbae6c43ceeec9e8f).

PR #5490, opened 2025-08-05
SLIP-24 UI for Ripple

ibz:
This PR implements the current payment requests UI for Ripple

### DIFF
--- a/core/.changelog.d/4951.added.2
+++ b/core/.changelog.d/4951.added.2
@@ -1,1 +1,0 @@
-[T3W1] Added SLIP-24 swaps.


core/.changelog.d/5450.added.1
```diff
@@ -0,0 +1 @@
+[T2B1,T3B1,T3T1,T3W1] New payment requests UI for Ripple.
```
bieleluk: I'm not sure about `T3W1`. As it has not been released yet, we haven't added any changelog entries so far.

core/src/apps/ripple/layout.py
```diff
@@ -28,3 +36,60 @@ async def require_confirm_tx(to: str, value: int, chunkify: bool = False) -> Non
...
+    provider_address: str,
+    verified_payment_request: PaymentRequest,
+    address_n: Bip32Path | None,
+) -> None:
+    from trezor.ui.layouts import confirm_payment_request
+
+    from apps.common.paths import address_n_to_str
+
+    assert verified_payment_request.amount is not None  # required for non-CoinJoin
+    total_amount = format_amount(verified_payment_request.amount, DECIMALS) + " XRP"
```
bieleluk: After https://github.com/trezor/trezor-firmware/pull/5464 merge, we should use a dedicated function for amount-unit formatting:
```suggestion
    from trezor.strings import format_amount_unit
    total_amount = format_amount_unit(format_amount(verified_payment_request.amount, DECIMALS) , "XRP")
```
If the tests are failing for any reason, then there is another one that should catch all forgotten occurrences: https://github.com/trezor/trezor-firmware/issues/5502

core/.changelog.d/5450.added.1
ibz: Good point! I'll remove `4951.added.2` too!! https://github.com/trezor/trezor-firmware/blob/main/core/.changelog.d/4951.added.2

core/src/apps/ripple/layout.py
ibz: Nice! I made a separate commit to deal with all the instances: 38692b5785d0f411425faf91f3b63f97aac70c2d

core/.changelog.d/5450.added.1
ibz: cbe807e6425d230e180daf39cf3faa79a75518d2 This one, was already in `main`...

bieleluk (approved):
LGTM
Please, add fixtures for all languages and re-run the T3W1 click tests because one test has randomly failed.
image

PR #5491, opened 2025-08-05
ci: do not include protocol version in artifact names

mmilata:
Fixes unavailable ARM builds in trezor-user-env. Also removes some extraneous model-protocol combinations and disables Monero tests for T3W1.

cc @HajekOndrej


romanz (approved):
Thanks!

PR #5493, opened 2025-08-05
fix(core): skip unexpected THP packet during channel allocation

romanz:
IIUC, it's possible that the device will retransmit old packets while a new test is being started.

PR #5494, opened 2025-08-05
test(core): allow wiping via DebugLink

romanz:
It allows THP-based debug builds to be wiped without unlocking.

Also, some of the entropy-related UI fixtures had to be changed.

~Rebased over #5493.~


romanz:
Rebasing to resolve a conflict.

romanz:
Also tested on T3T1 by running device tests.

PR #5497, opened 2025-08-06
reset primask in emergency rescue

TychoVrahe:
This allows RSOD to work properly in case the crash occurs in critical sections with interrupts disabled via primask


core/embed/sys/task/stm32/system.c
```diff
@@ -129,6 +129,8 @@ system_emergency_rescue_phase_2(uint32_t arg1, uint32_t arg2) {
 
   // Now we can safely enable interrupts again
   __enable_fault_irq();
+  // In case we crashed while irq_lock was active
+  __set_PRIMASK(0);
```
cepetr: Just a small nit: Maybe it would be clearer to use `__enable_irq()`, as it does the same thing.

TychoVrahe: changed https://github.com/trezor/trezor-firmware/commit/db0c027ea65c2b7be6a01009544a47780aa68f24

cepetr (approved):
utACK

PR #5501, opened 2025-08-06
test(core): add a temporary workaround for UI tests flakiness on THP

romanz:
In case of an event loop restart, it's possible that the first packet(s) of `DebugLinkRecordScreen` will be lost, resulting in `TrezorFailure: FirmwareError: Invalid magic` error responses during test setup.

This issue will be resolved as part of THP event loop restart refactoring, but till then let's wait a bit here, to reduce the packet loss probability.

### DIFF
--- a/tests/ui_tests/__init__.py
+++ b/tests/ui_tests/__init__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import shutil
+import time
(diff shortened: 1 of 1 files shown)


romanz:
Tested also in #5500 and #5494.

mmilata (approved):

PR #5509, opened 2025-08-07
Haptic support in emulator

### LINKED ISSUE #3735: Implement haptic support in emulator
We should be able to see (or hear?) what is going on with haptic when using emulator too.

bieleluk:
- [delizia, eckhart] haptic feedback support in emulator, the colored point is rendered on the touch coords: 
  - red - simple click
  - orange - intermediate HTC
  -  green - completed HTC
  - In case of tests, ~~the upper left corner pixel is rendered instead of the point~~, the haptic point is rendered out of the screen (coords [-1,-1]) because of too many new screens and their inconsistency (`press_yes` doesn't create one but clicking does, ...)
- [delizia, eckhart]: test for enabling/disabling the haptic feedback

### DIFF
--- a/core/.changelog.d/3735.added
+++ b/core/.changelog.d/3735.added
@@ -1,0 +1,1 @@
+[T3T1,T3W1] Haptic support in emulator.
--- a/core/SConscript.unix
+++ b/core/SConscript.unix
@@ -26,6 +26,7 @@ FEATURES_WANTED = [
     "ble",
     "display",
     "dma2d",
+    "haptic",
     "input",
     "kernel_mode",
     "optiga",
--- a/core/embed/io/display/unix/display_driver.c
+++ b/core/embed/io/display/unix/display_driver.c
@@ -82,6 +82,12 @@ typedef struct {
   // Color of the RGB LED
   uint32_t led_color;
 #endif
+
+#ifdef USE_HAPTIC
+  uint32_t haptic_color;
+  uint32_t haptic_expire_time;
+#endif /* USE_HAPTIC */
+
 } display_driver_t;
 
 static display_driver_t g_display_driver = {
@@ -91,6 +97,7 @@ static display_driver_t g_display_driver = {
 //!@# TODO get rid of this...
 int sdl_display_res_x = DISPLAY_RESX, sdl_display_res_y = DISPLAY_RESY;
 int sdl_touch_offset_x, sdl_touch_offset_y;
+int sdl_touch_x = -1, sdl_touch_y = -1;
 
 static void display_exit_handler(void) {
   display_deinit(DISPLAY_RESET_CONTENT);
@@ -188,6 +195,11 @@ bool display_init(display_content_mode_t mode) {
 #ifdef USE_RGB_LED
   drv->led_color = 0;
 #endif
+
+#ifdef USE_HAPTIC
+  drv->haptic_color = 0;
+  drv->haptic_expire_time = 0;
+#endif /* USE_HAPTIC */
 
   gfx_bitblt_init();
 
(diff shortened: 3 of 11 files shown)


core/embed/sys/task/sysevent.c
```diff
@@ -188,6 +189,8 @@ void sysevents_poll(const sysevents_t *awaited, sysevents_t *signalled,
 #ifdef TREZOR_EMULATOR
     // Poll SDL events and dispatch them
     sdl_events_poll();
+    // Refresh display to catch the end of the haptic feedback
+    display_refresh();
```
romanz, line 193: nit: maybe `ifdef` the refresh, so it would be done only if `USE_HAPTIC` is defined?
(here and below)

core/embed/io/display/unix/display_driver.c
```diff
@@ -386,6 +398,102 @@ void draw_rgb_led() {
...
+    return;
+  }
+
+  const uint32_t color = drv->haptic_color;
+
+  if (color == 0) {
+    return;  // No LED color set
+  }
+
+  // Extract RGB components
```
romanz, line 458: nit: refactor RGB extraction into a helper function/macro?

core/embed/sys/task/sysevent.c
bieleluk: I can do that. But I'm afraid that this change makes the test a bit more flaky, where there are extra screens captured, i.e.
https://data.trezor.io/dev/firmware/ui_report/17039020230/diff/T3W1_en-persistence_tests-test_shamir_persistence.py::test_recovery_single_reset.html
WDYT?

romanz: IIUC, the flakiness issue should be solved by #5546 -> https://github.com/trezor/trezor-firmware/actions/runs/17135342281.

core/.changelog.d/3735.added
```diff
@@ -0,0 +1 @@
+[T3T1,T3W1] Haptic support in emulator.
```
TychoVrahe, line 1: not sure about this changelog entry usefulness for users

core/SConscript.unix
```diff
@@ -26,6 +26,7 @@ FEATURES_WANTED = [
     "ble",
     "display",
     "dma2d",
+    "haptic",
```
TychoVrahe, line 29: also perhaps in bootloader emulator (and prodtest emulator maybe?)

core/embed/io/display/unix/display_driver.c
```diff
@@ -386,6 +398,102 @@ void draw_rgb_led() {
...
+    return;
+  }
+
+  switch (effect) {
+    case HAPTIC_BUTTON_PRESS:
+      drv->haptic_color = 0xFF0000;                    // Red
+      drv->haptic_expire_time = SDL_GetTicks() + 200;  // 200 ms duration
+      break;
+    case HAPTIC_HOLD_TO_CONFIRM:
+      drv->haptic_color = 0xFF00;                      // Green
```
TychoVrahe, line 415: 0x00FF00 would look nicer :)

core/embed/io/display/unix/display_driver.c
```diff
@@ -386,6 +398,102 @@ void draw_rgb_led() {
...
+      drv->haptic_expire_time = SDL_GetTicks() + 500;  // 500 ms duration
+      break;
+    case HAPTIC_BOOTLOADER_ENTRY:
```
TychoVrahe, line 418: please rebase and add the new effect

tests/device_tests/test_msg_applysettings.py
```diff
@@ -542,3 +542,43 @@ def test_set_brightness_cancel(session: Session):
         IF = InputFlowCancelBrightness(client)
         client.set_input_flow(IF.get())
         device.set_brightness(session, None)
+
+
+@pytest.mark.models(
+    "delizia",
```
TychoVrahe, line 548: i think it would be more precise to base this on model not UI

core/embed/sys/task/sysevent.c
TychoVrahe: The solution looks weird to me and we should probably look into more sustainable one, so we don't mix display/haptic/led functionality. No need to do it now under time pressure, but i just want to state that i am not very happy with that

core/.changelog.d/3735.added
obrusvit: true, you can delete the changelog entry @bieleluk

romanz:
> * In case of tests, ~the upper left corner pixel is rendered instead of the point~, the haptic point is rendered out of the screen (coords [-1,-1]) because of too many new screens and their inconsistency (`press_yes` doesn't create one but clicking does, ...)

Maybe we can use something similar to `render_debug_overlay()` (which [renders the upper right red square](https://docs.trezor.io/trezor-firmware/core/emulator/index.html#debug-and-production-mode)), which is being hidden when screen recording is in progress?
https://github.com/trezor/trezor-firmware/blob/d8cff7ac283765ed52b7b0ffcf1221a6a8bd1ea1/core/embed/rust/src/ui/shape/display/fb_rgba8888.rs#L75-L82
https://github.com/trezor/trezor-firmware/blob/d8cff7ac283765ed52b7b0ffcf1221a6a8bd1ea1/core/embed/rust/src/ui/layout_eckhart/mod.rs#L106-L116
For example, we can change the color/size of the square according to the haptic effect being used.

obrusvit:
Closing this as this PR got stale and it's not worth reviving it now.

PR #5510, opened 2025-08-07
Eckhart UI: adjust screen border

obrusvit:
- render the dispaly 1px from the edge towards center to account for minor irregularities in display production

[no changelog]


core/embed/rust/src/ui/layout_eckhart/cshape/screen_border.rs
```diff
@@ -1,51 +1,55 @@
...
     color: Color,
     side_bars: [Rect; 3],
 }
 
 impl ScreenBorder {
-    pub const WIDTH: i16 = 4;
+    pub const TOTAL_WIDTH: i16 = 5;
+    pub const LINE_WIDTH: i16 = Self::TOTAL_WIDTH - 1;
     pub const TOP_ARC_HEIGHT: i16 = ICON_BORDER_TOP.toif.height();
+    pub const SCREEN_INSET: Rect = SCREEN.inset(Insets::uniform(1));
```
bieleluk: Nit: Why is it called an `INSET` when it's a `Rect`?

core/embed/rust/src/ui/layout_eckhart/res/border/note.md
```diff
@@ -0,0 +1,2 @@
+TOP, BL, BR icons are supposed to be used completely at the edge of the display.
```
bieleluk, line 1: Are we using these icons anywhere?

bieleluk (approved):
Approved with 2 minor comments.

core/embed/rust/src/ui/layout_eckhart/res/border/note.md
obrusvit: No, I just thought that maybe we can keep them in the repo in case we want to use it the old way again.

core/embed/rust/src/ui/layout_eckhart/cshape/screen_border.rs
obrusvit: I meant `_INSET` as past participle (screen which is inset), not a noun "inset". But you're right, I renamed it to `SCREEN_SHRUNK` on rebase.

PR #5511, opened 2025-08-07
SLIP-24 for Cardano

ibz:
This PR implements SLIP-24 payment requests for Cardano.

### DIFF
--- a/core/src/apps/cardano/layout.py
+++ b/core/src/apps/cardano/layout.py
@@ -1,6 +1,6 @@
 from typing import TYPE_CHECKING
 
-from trezor import TR
+from trezor import TR, wire
 from trezor.enums import (
     ButtonRequestType,
     CardanoAddressType,
@@ -28,7 +28,10 @@ if TYPE_CHECKING:
 
     from trezor import messages
     from trezor.enums import CardanoNativeScriptHashDisplayFormat
+    from trezor.messages import PaymentRequest
     from trezor.ui.layouts import PropertyType
+
+    from apps.common.paths import Bip32Path
 
     from .helpers.credential import Credential
     from .seed import Keychain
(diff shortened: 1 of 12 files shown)


core/src/apps/cardano/layout.py
```diff
@@ -1117,3 +1120,59 @@ async def show_cardano_address(
...
+    verified_payment_request: PaymentRequest,
+    address_n: Bip32Path | None,
+    network_id: int,
+) -> None:
+    from trezor.ui.layouts import confirm_payment_request
+
+    assert verified_payment_request.amount is not None  # required for non-CoinJoin
+    total_amount = format_coin_amount(verified_payment_request.amount, network_id)
+
+    texts = []
```
bieleluk: From the testing point of view, it would be better to explicitly mark the type of the variable.
The same applies to the variables below.

core/src/apps/cardano/sign_tx/signer.py
```diff
@@ -250,6 +253,9 @@ def _validate_tx_init(self) -> None:
 
         msg = self.msg  # local_cache_attribute
 
+        if msg.payment_req and msg.outputs_count > 1:
+            raise ProcessError("Multiple outputs not supported for payment requests")
```
bieleluk, line 257: Would it make sense to have a test for this scenario?

core/src/apps/cardano/layout.py
ibz: https://github.com/trezor/trezor-firmware/pull/5511/commits/625002dd2fd64dc8b201d1842c0ea50d0b4a1a78#diff-718c35fab848e1ee6f5e2a387c71b0a0ea8d84039fedce530cb9f0584c164495R1136

core/src/apps/cardano/sign_tx/signer.py
ibz: Good idea! https://github.com/trezor/trezor-firmware/pull/5511/commits/625002dd2fd64dc8b201d1842c0ea50d0b4a1a78#diff-83abadd918ea85e304f08f37493c885eebb112d9f9367be1965caf29396dd8c1R1

PR #5514, opened 2025-08-07
feat(core): adjust RSOD

obrusvit:
- Eckhart RSOD now says "Wait for Trezor shutdown" (not restart)
- the footer now says "Please visit trezor.io/rsod" after wipe code entry or running out of PIN attempts
- default title/message/footer texts are lowercase for T3W1
- titles used in `error_handling.c` are lowercase so that it looks nicer. However we want to keep the readability on T2B1/T3B1 so the `_UPPER` font is used which renders it uppercase

### DIFF
--- a/core/embed/util/rsod/rsod.c
+++ b/core/embed/util/rsod/rsod.c
@@ -28,10 +28,10 @@
 #include <util/scm_revision.h>
 #endif
 
-#define RSOD_DEFAULT_TITLE "INTERNAL ERROR";
-#define RSOD_DEFAULT_MESSAGE "UNSPECIFIED";
-#define RSOD_DEFAULT_FOOTER "PLEASE VISIT TREZOR.IO/RSOD";
-#define RSOD_EXIT_MESSAGE "EXIT %d"
+#define RSOD_DEFAULT_TITLE "Internal error";
+#define RSOD_DEFAULT_MESSAGE "Unspecified";
+#define RSOD_DEFAULT_FOOTER "Please visit trezor.io/rsod";
+#define RSOD_EXIT_MESSAGE "Exit %d"
 
 #ifdef KERNEL_MODE
 
--- a/core/embed/rtl/error_handling.c
+++ b/core/embed/rtl/error_handling.c
@@ -27,6 +27,15 @@
 uint32_t __stack_chk_guard = 0;
 #endif
 
+#define ALL_DATA_ERASED_MESSAGE "All data has been erased from the device"
+
+#ifdef TREZOR_MODEL_T3W1
+// empty message for T3W1 so that it falls to the more appropriate default
+#define RECONNECT_DEVICE_MESSAGE ""
+#else
+#define RECONNECT_DEVICE_MESSAGE "Please reconnect\nthe device"
+#endif
+
 // Calls to this function are inserted by the compiler
 // when stack protection is enabled.
 void __attribute__((noreturn, used)) __stack_chk_fail(void) {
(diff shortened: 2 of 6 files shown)


core/embed/util/rsod/rsod.c
```diff
@@ -28,10 +28,17 @@
 #include <util/scm_revision.h>
 #endif
 
+#ifdef TREZOR_MODEL_T3W1
+#define RSOD_DEFAULT_TITLE "Internal error";
+#define RSOD_DEFAULT_MESSAGE "Unspecified";
+#define RSOD_DEFAULT_FOOTER "Please visit\ntrezor.io/rsod";
+#define RSOD_EXIT_MESSAGE "Exit %d"
+#else
 #define RSOD_DEFAULT_TITLE "INTERNAL ERROR";
```
TychoVrahe: maybbe we should let UI implementation decide on capitalization?

core/embed/rust/src/ui/layout_eckhart/mod.rs
```diff
@@ -38,6 +38,7 @@ use crate::ui::layout::simplified::show;
 use component::{ErrorScreen, UpdateScreen, WelcomeScreen};
 
 pub const WAIT_FOR_RESTART_MESSAGE: &str = "Wait for Trezor restart";
+pub const WAIT_FOR_SHUTDOWN_MESSAGE: &str = "Wait for Trezor shutdown";
```
TychoVrahe: i believe that RSOD behavior is that trezor automatically restarts, so why this message?

obrusvit: In that case, it was wrong in Figma, I pointed out the mistake and reverted the use of this message:
https://github.com/trezor/trezor-firmware/pull/5514/commits/549b65a61f7afc6848c06a86939ed108cc78fdb9

core/embed/util/rsod/rsod.c
obrusvit: I suggest keeping it lower-cased by default. Bolt/Caesar layouts use the `_UPPER` variants of their fonts for titles. However, the message and footer now changes to lower-cased text. TBH, I don't think it's a problem. WDYT

obrusvit: https://github.com/trezor/trezor-firmware/pull/5514/commits/35e677ffb480191b06a527215d94e8c17e741dd6

TychoVrahe: not a problem for me, but i am wrong person to judge that :) @Hannsek perhaps?

obrusvit: I forgot that RSOD_DEFAULT_FOOTER  T3W1 should have also the newline, i.e. `"Please visit\ntrezor.io/rsod"`. Which means we either make the `Label` for Eckhart footer less wide or we use the preprocessor macro :thinking:

TychoVrahe: right... that sucks. But that formatting is strictly speaking still UI decision so using Label would be a better fit, or maybe the messages don't belong here at all. Anyway as it gets more complicated maybe we'd be ok with the macro :) for now at least

obrusvit: I agreed with @lapohoda that we will render the footer message without the newline. Figma updated. Code-wise: removed the `\n` from the default message, adjusted the `ErrorScreen::place` method and removed unused function in `BldHeader`

https://github.com/trezor/trezor-firmware/pull/5514/commits/279aece8fc0e88222bc3d312c5de26993bae0447

TychoVrahe: ack, please rebase so we can hopefully see the RSOD changes

obrusvit: RSOD changes are visible, asked @Hannsek for review

obrusvit: @Hannsek approved

obrusvit: @TychoVrahe can you have a final check code-wise?

TychoVrahe (approved):
lgtm, i created a followup issue https://github.com/trezor/trezor-firmware/issues/5549

PR #5515, opened 2025-08-07
Eckhart UI: show pairing code in big font in bootloader

obrusvit:
- introduce new FontInfo for the largest available font (72) which contains only the numerals, this is used in bootloader for the pairing code comparison screen

[no changelog]


core/embed/rust/src/ui/layout_eckhart/fonts/font_ttsatoshi_extralight_72.rs
```diff
@@ -395,6 +395,105 @@ const Font_TTSatoshi_ExtraLight_72: [&[u8]; 95] = [
     &Font_TTSatoshi_ExtraLight_72_glyph_126,
 ];
 
+/// Array of references for 'TTSatoshi_ExtraLight_72' numeral-only ASCII glyphs
```
TychoVrahe, line 400: this is not autogenerated right? in which case, we should at least add a note here, but better would be to add an option to gen_font to generate num-only variant too - so that we can reliably reproduce the file from vector font source.

obrusvit: I added a comment for now: https://github.com/trezor/trezor-firmware/pull/5515/commits/1f131379a3c0049c1e33c40d04dbaf49b8f063af

And created an issue to adjust the generator: https://github.com/trezor/trezor-firmware/issues/5522

PR #5516, opened 2025-08-08
refactor(core): introduce `ThpContext`

romanz:
This PR moves `thp_main` code to a "context"-like class, similar to `CodecContext` for Protocol v1.

It encapsulates payload reassembly, as well as handling and responding to low-level single packet THP messages.
Also, channel cache is now using interface number for `CHANNEL_IFACE`, in order to simplify channel interface serialization.

Rebased over #5512.


core/src/trezor/wire/thp/context.py
```diff
@@ -0,0 +1,152 @@
...
+    from trezorio import WireInterface
+
+_CID_REQ_PAYLOAD_LENGTH = const(12)
+
+# Codec_v1 magic constant:
+# "?##" + Failure message type + msg_size + msg_data (code = "Failure_InvalidProtocol")
+CODEC_V1_ERROR_MESSAGE = b"\x3f\x23\x23\x00\x03\x00\x00\x00\x14\x08\x11"
+
+
+class ThpContext:
```
mmilata: Nit: wanna maybe add a sentence or two wrt. what this is and how it is different from the other contexts we have?

mmilata: Perhaps even rename the file to something like `interface_context.py`?

mmilata (approved):
Refactoring LGTM :+1:

core/src/trezor/wire/thp/context.py
romanz: https://github.com/trezor/trezor-firmware/pull/5516/commits/540e4b5d492088ac5564291d4d8a551baf20e1a8 & https://github.com/trezor/trezor-firmware/pull/5516/commits/fb03f9b86d1dd1336cd263f01eba0fca13db62a2

PR #5519, opened 2025-08-08
refactor(core): simplify THP write-related code

romanz:
- move fragmentation code into `PacketHeader.fragment_payload()`
- use `ThpContext.write_payload()` to send responses to the host
- precompute `INIT_HEADER_LENGTH` and `CONT_HEADER_LENGTH`
- remove most of `trezor.write.thp.writer` code


core/src/trezor/wire/thp/__init__.py
```diff
@@ -118,9 +122,37 @@ def pack_to_cont_buffer(self, buffer: bytearray, buffer_offset: int = 0) -> None
...
+
+        for item in items:
+            item_offset = 0
+            while item_offset < len(item):
+                n = utils.memcpy(buf, buf_offset, item, item_offset)
+                buf_offset += n
+                item_offset += n
+
+                if buf_offset == len(buf):
+                    should_pad = True
```
mmilata: Does this mean single-packet payloads do not get padded? :thinking:

mmilata (approved):

romanz:
Rebasing to resolve conflicts.

core/src/trezor/wire/thp/__init__.py
romanz: IIUC it should be OK not to zero-pad single-packet payload, since we create `packet` as a `bytearray` (which is zeroed on creation), but it's indeed not clear from the current code.

https://github.com/trezor/trezor-firmware/pull/5519/commits/2c402b18877e223b31e130f8888aaea35bd4937d moves the `bytearray` allocation into this method, to make it clearer.
Also, it uses `memzero` (added in https://github.com/trezor/trezor-firmware/pull/5519/commits/a23b6aa76fc97d7a047b7d6a937afd29138f9fe8) to avoid allocating `bytes()` only for zero padding.

PR #5520, opened 2025-08-08
Fonts cleanup

obrusvit:


bieleluk (commented):
LGTM with one exception, that after running `gen_font` script with font files from [#1](https://github.com/trezor/trezor-firmware-fonts/pull/1), the `font_unifont_bold_16.rs` has changes that are not explicitly marked as expected (unlike other files).

obrusvit:
@bieleluk I noticed that too but I cannot find the right font data..

bieleluk (approved):

PR #5523, opened 2025-08-11
fix(core): don't reuse singleton exception objects

romanz:
Otherwise, the traceback object is growing without being cleared.

Following https://github.com/trezor/trezor-firmware/issues/5472#issuecomment-3174382038.


romanz:
> I wonder if we should change this one as well:
> 
> https://github.com/trezor/trezor-firmware/blob/96afce1e0c0c5c94f0d1a74f89661863cd461cb8/core/src/apps/common/keychain.py#L37

Changed in https://github.com/trezor/trezor-firmware/pull/5523/commits/2f7c2c8ded2d0ccf9e02f48116c53f2b834cbdd8.

mmilata (approved):
Thanks!

PR #5524, opened 2025-08-11
Adjust `code_code_entry` encoding, document CPace implementation

M1nd3r:


tests/device_tests/thp/test_pairing.py
```diff
@@ -141,7 +141,7 @@ def test_pairing_code_entry(
 
     cpace = Cpace(handshake_hash=protocol.handshake_hash)
     cpace.random_bytes = os.urandom
-    cpace.generate_keys_and_secret(code.to_bytes(6, "big"), cpace_trezor_public_key)
+    cpace.generate_keys_and_secret(str(code).encode("ascii"), cpace_trezor_public_key)
```
andrewkozlik: Shouldn't the code be constant length, so that the padding is always correct? `str(code)` -> `f"{code:06}"`

M1nd3r: fixed here: https://github.com/trezor/trezor-firmware/pull/5524/commits/87784d7cef1011dde75481f340512865abd7b655

M1nd3r: Note: Variable length of `code_code_entry` in the current implementation would not only break the padding (`zpad`), but the length of the `code_code_entry` itself is a part of the `pregenerator` generation process. So it would lead to even more inconsistencies with the specification https://datatracker.ietf.org/doc/draft-irtf-cfrg-cpace/

core/src/trezor/wire/thp/cpace.py
```diff
@@ -21,11 +25,28 @@ def generate_keys(self, code_code_entry: bytes) -> None:
         """
         Generate an ephemeral key pair using Elligator2 with X25519.
         """
-        sha_ctx = sha512(_PREFIX)
-        sha_ctx.update(code_code_entry)
-        sha_ctx.update(_PADDING)
-        sha_ctx.update(self.handshake_hash)
-        sha_ctx.update(b"\x00")
+        if __debug__:
```
andrewkozlik: AFAIK assert statements are only active in debug mode anyway, so this seems redundant. https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

M1nd3r: `if __debug__:`  before assert removed.

Originally, I was unsure that using the variable `_LEN_PRS` in the statement would not cause issues - as it is available only in _debug_. It seems that `assert` is implemented better than I thought.

PR #5525, opened 2025-08-11
Implement tropic provisioning

onvej-sl:


core/embed/projects/prodtest/README.md
```diff
@@ -648,6 +648,33 @@ secrets-init
 OK
 ```
 
+### secrets-get-mcu-device-key
+Returns the device attestation key stored in MCU.
```
andrewkozlik:
```suggestion
Returns a cryptogram that encrypts and authenticates the device attestation public key stored in MCU. The commands `secrets-init` and `secure-channel-handshake-2` must be executed before calling this command.
```

core/embed/projects/prodtest/README.md
```diff
@@ -648,6 +648,33 @@ secrets-init
...
+OK 638c8a83ddc8fd84cddf5a0a4fa3d9615146cd341685dca942bab1132c2bc99b
+```
+
+### secrets-certdev-write
+Writes the X.509 device attestation certificate issued by the Trezor Company for the attestation key stored in MCU.
+
+Example:
+```
+secrets-certdev-write
+OK <hexadecimal string>
```
andrewkozlik:
```suggestion
secrets-certdev-write <hexadecimal string>
OK
```

core/embed/projects/prodtest/README.md
```diff
@@ -947,6 +974,145 @@ tropic-get-chip-id
...
+
+Example:
+```
+tropic-get-fido-masking-key
+OK dc106118a32feeef8d9211f54b9c8e9d571abe4cb104dc4ab087531cfee4574283ccf9c6f45e68be712f630d72d4999c
+```
+
+### tropic-handshake
+
+Establishes a secure channel with the Tropic chip. Expects handshake a request as input, returns a handshake response.
```
andrewkozlik:
```suggestion
Establishes a secure channel with the Tropic chip. Expects a handshake request as input, returns a handshake response.
```

core/embed/projects/prodtest/README.md
```diff
@@ -947,6 +974,145 @@ tropic-get-chip-id
...
+
+Example:
+```
+tropic-lock
+OK <hexadecimal string>
+```
+
+### secure-channel-handshake-1
+
+Returns the first handshake message for establishing a secure channel with the device.
```
andrewkozlik:
```suggestion
Returns the first handshake message for establishing a secure channel between the device and HSM.
```

core/embed/projects/prodtest/README.md
```diff
@@ -947,6 +974,145 @@ tropic-get-chip-id
...
+
+Example:
+```
+secure-channel-handshake-1
+OK 1e85285cbf805d0418be1f502a325806f68fa07c78fd63b7b960b2d0416f8b49
+```
+
+### secure-channel-handshake-2
+
+Establishes a secure channel with the device. Expects the second handshake message as input. The command `secure-channel-handshake-1` must be executed before calling this command.
```
andrewkozlik:
```suggestion
Establishes a secure channel between the device and HSM. Expects the second handshake message as input. The command `secure-channel-handshake-1` must be executed before calling this command.
```

core/embed/projects/prodtest/README.md
andrewkozlik: Since these two commands chronologically precede the `tropic-*` and `secrets-*` commands in the protocol, how about moving them somewhere before the `secrets-*` commands?

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -136,3 +204,26 @@ PRODTEST_CLI_CMD(
   .info = "Generate and write secrets to flash",
   .args = ""
 );
+
+#ifdef SECRET_MASTER_KEY_SLOT_SIZE
+PRODTEST_CLI_CMD(
+  .name = "secrets-get-mcu-device_key",
+  .func = prodtest_secrets_get_mcu_device_key,
+  .info = "Get MCU device key",
```
andrewkozlik:
```suggestion
  .name = "secrets-get-mcu-device-key",
  .func = prodtest_secrets_get_mcu_device_key,
  .info = "Get MCU device attestation public key",
```

core/embed/sys/syscall/stm32/syscall_verifiers.c
```diff
@@ -1218,33 +1218,17 @@ bool tropic_ping__verified(const uint8_t *msg_out, uint8_t *msg_in,
...
-                               uint16_t dig_len, uint8_t *sig,
-                               uint16_t sig_len) {
+                               uint16_t dig_len, uint8_t *sig) {
   if (!probe_read_access(dig, dig_len)) {
     goto access_violation;
   }
 
-  if (!probe_write_access(sig, sig_len)) {
-    goto access_violation;
-  }
```
andrewkozlik: I am not sure what the consequences of removing this are. Shouldn't we at least call it with some MAX_SIG_LEN parameter? Note that the same applies in `core/embed/sys/smcall/stm32/smcall_verifiers.c`.

core/embed/sec/tropic/stm32/tropic01.c
```diff
@@ -184,11 +184,15 @@ lt_ret_t lt_port_delay(lt_handle_t *h, uint32_t ms) {
...
+  if (len % sizeof(uint32_t) != 0) {
+    return LT_PARAM_ERR;
+  }
+
   while (len > 0) {
     uint32_t random = rng_get();
     *buff = random;
     buff++;
-    len--;
+    len -= sizeof(uint32_t);
```
andrewkozlik: This change contradicts the API documentation for `lt_port_random_bytes()`.
```c
/**
 * @brief Fill buffer with random bytes, platform defined function.
 *
 * @param buff        Buffer to be filled
 * @param len         number of 32bit numbers
 * @return lt_ret_t   S_OK if success, otherwise returns other error code.
 */
lt_ret_t lt_port_random_bytes(uint32_t *buff, uint16_t len);
```

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
```diff
@@ -127,7 +105,7 @@ STATIC mp_obj_t mod_trezorcrypto_tropic_sign(mp_obj_t key_index,
   vstr_init_len(&sig, SIG_SIZE);
```
andrewkozlik: nit: How about replacing `SIG_SIZE` with `ECDSA_RAW_SIGNATURE_SIZE` because that's what we use inside `tropic_ecc_sign()`? It appears only here and few lines below in `sig.len = SIG_SIZE`. Or if Tropic has their own constant for this then use that. (Update: Looks like they don't have one :roll_eyes:.)

core/embed/sec/tropic/stm32/tropic01.c
onvej-sl: I think you are looking at outdated documentation. See https://github.com/tropicsquare/libtropic/blob/138053406adc4d042dcf8119006fd112c050e917/include/libtropic_port.h#L97.

core/embed/sys/syscall/stm32/syscall_verifiers.c
onvej-sl: I think you're right. I'll fix it.

core/embed/sec/tropic/stm32/tropic01.c
andrewkozlik: I see. But also the `buff` parameter has changed to `void *`, meaning that the pointer doesn't have to be aligned and that the length doesn't have to be a multiple of 4. The newer ARM processors should be able to handle casting `void*` to `uint32_t*`, although I think older ones would have a problem if the pointer was unaligned. Why not just call random_buffer() from `crypto/rand.c`?

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -43,19 +63,465 @@ typedef enum {
...
 
-static bool locked = false;
+static bool tropic_is_paired = false;
+static bool tropic_is_locked = false;
+
+static uint8_t tropic_cert_chain[LT_NUM_CERTIFICATES *
+                                 LT_L2_GET_INFO_REQ_CERT_SIZE_SINGLE] = {0};
+size_t tropic_cert_chain_length = 0;
+curve25519_key tropic_public = {0};
+bool tropic_cert_chain_cached = false;
```
andrewkozlik: Looks like these can be static. BTW, `tropic_cert_chain_cached` should be equivalent to `tropic_cert_chain_length > 0`.

core/embed/sec/tropic/stm32/tropic01.c
onvej-sl: > Why not just call random_buffer() from crypto/rand.c?

I will do it.

M-pasta (approved):
`prodtest_tropic.c` checked

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -85,28 +553,67 @@ static void prodtest_tropic_get_chip_id(cli_t* cli) {
...
+  for (size_t i = 0; i < LT_NUM_CERTIFICATES; i++) {
+    memmove(&tropic_cert_chain[length], cert_store.certs[i],
+            cert_store.cert_len[i]);
+    length += cert_store.cert_len[i];
+  }
+
+  ret = lt_get_st_pub(&cert_store, tropic_public, sizeof(tropic_public));
+  if (ret != LT_OK) {
+    return false;
+  }
```
andrewkozlik: I was confused here about the compaction, because it invalidates the cert_store, but then we use the cert_store afterwards. Since `lt_get_st_pub()` looks only at the first certificate, this works, but it seems cleaner to swap the two code blocks and add a comment explicitly stating that the cert_store is invalid from that point on.
 
```suggestion
  ret = lt_get_st_pub(&cert_store, tropic_public, sizeof(tropic_public));
  if (ret != LT_OK) {
    return false;
  }

  // Compactify tropic_cert_chain for future use. This invalidates the cert_store.
  size_t length = 0;
  for (size_t i = 0; i < LT_NUM_CERTIFICATES; i++) {
    memmove(&tropic_cert_chain[length], cert_store.certs[i],
            cert_store.cert_len[i]);
    length += cert_store.cert_len[i];
  }
```

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -43,19 +63,465 @@ typedef enum {
 static tropic_handshake_state_t tropic_handshake_state =
     TROPIC_HANDSHAKE_STATE_0;
 
-static bool locked = false;
+static bool tropic_is_paired = false;
+static bool tropic_is_locked = false;
```
andrewkozlik: We don't seem to be using `tropic_is_locked` anywhere, only writing to it. What's its purpose? One can't rely on the value anyway, because Tropic can be locked before power-up, but the variable will be set to false by default until someone calls `prodtest_tropic_lock_check()`.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -117,52 +624,298 @@ static void prodtest_tropic_lock_check(cli_t* cli) {
...
+
+  return LT_OK;
+}
+
+static bool is_paired(cli_t* cli) {
+  lt_handle_t* tropic_handle = tropic_get_handle();
+  lt_ret_t ret = LT_FAIL;
+
+  // Retrieve the tropic public key.
+  curve25519_key tropic_public = {0};
```
andrewkozlik, line 721: Note that this shadows a global variable with the same name. Not a problem now, but shadowing may lead to bugs in the future. Perhaps rename the global one to `tropic_public_cached`?

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -117,52 +624,298 @@ static void prodtest_tropic_lock_check(cli_t* cli) {
...
     cli_error_arg_count(cli);
-    return;
+    goto cleanup;
   }
 
   tropic_handshake_state = TROPIC_HANDSHAKE_STATE_0;
 
+  lt_handle_t* tropic_handle = tropic_get_handle();
+
+  if (!cache_tropic_cert_chain(tropic_handle)) {
```
andrewkozlik, line 826: The reason to call this is not immediately obvious. A comment would be helpful:
```suggestion
  // Ensure that tropic_public is cached.
  if (!cache_tropic_cert_chain(tropic_handle)) {
```

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -117,52 +624,298 @@ static void prodtest_tropic_lock_check(cli_t* cli) {
...
+
+  // Try to establish a session using the factory pairing key.
+  ret = lt_session_start(tropic_handle, tropic_public, FACTORY_PAIRING_KEY_SLOT,
+                         factory_private, factory_public);
+  if (ret == LT_OK) {
+    // Write the privileged pairing key to the tropic's pairing key slot if it
+    // has not been written yet.
+    ret = pairing_key_write(tropic_handle, PRIVILEGED_PAIRING_KEYSLOT,
+                            privileged_public);
+    // If the pairing key has already been written, `write_pairing_key()`
```
andrewkozlik:
```suggestion
    // If the pairing key has already been written, `pairing_key_write()`
```
**Also in three more places below.**

core/embed/upymod/modtrezorcrypto/modtrezorcrypto-tropic.h
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5525/commits/c7c317dc0c5d9c1dc92f99719840f8c7ed01aa37.

core/embed/sec/tropic/stm32/tropic01.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5525/commits/c7c317dc0c5d9c1dc92f99719840f8c7ed01aa37.

core/embed/sys/syscall/stm32/syscall_verifiers.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5525/commits/c7c317dc0c5d9c1dc92f99719840f8c7ed01aa37.

core/embed/sec/tropic/stm32/tropic01.c
```diff
@@ -184,12 +186,7 @@ lt_ret_t lt_port_delay(lt_handle_t *h, uint32_t ms) {
 }
 
 lt_ret_t lt_port_random_bytes(uint32_t *buff, uint16_t len) {
```
andrewkozlik: According to the new function signature we should have:
```suggestion
lt_ret_t lt_port_random_bytes(void *buff, size_t count) {
```

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -228,13 +979,60 @@ static void prodtest_tropic_get_fido_masking_key(cli_t* cli) {
   memzero(fido_masking_key, sizeof(fido_masking_key));
 }
 
+static lt_ret_t l2_get_req_len(uint8_t* buffer, size_t buffer_length,
```
andrewkozlik:
```suggestion
static lt_ret_t l2_get_req_len(const uint8_t* buffer, size_t buffer_length,
```

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -228,13 +979,60 @@ static void prodtest_tropic_get_fido_masking_key(cli_t* cli) {
...
+  if (length > buffer_length) {
+    return LT_PARAM_ERR;
+  }
+
+  *req_length = length;
+
+  return LT_OK;
+}
+
+static lt_ret_t l2_get_rsp_len(uint8_t* buffer, size_t buffer_length,
```
andrewkozlik:
```suggestion
static lt_ret_t l2_get_rsp_len(const uint8_t* buffer, size_t buffer_length,
```

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -283,116 +1140,378 @@ static void prodtest_tropic_send_command(cli_t* cli) {
...
+  //   * irreversible (it cannot be undone),
+  //   * self-recovering (if the device is powered off during execution, it can
+  //   be called again to continue from where it left off).
+
   if (cli_arg_count(cli) > 0) {
     cli_error_arg_count(cli);
     return;
   }
 
   tropic_handshake_state = TROPIC_HANDSHAKE_STATE_0;
```
andrewkozlik, line 1232: How does this work? We don't need to have a handshake completed to write the config?

Let's assume that we don't need it for a blank chip, but once the config is written, then surely we need some handshake to erase and/or write the config. So how can this function be idempotent and self-recovering?

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -283,116 +1140,378 @@ static void prodtest_tropic_send_command(cli_t* cli) {
...
+    cli_error(cli, CLI_ERROR, "Unable to read certificate");
+    goto cleanup;
   }
 
-  // TODO: Implement properly
+  cli_ok_hexdata(cli, certificate, certificate_length);
 
-  systick_delay_ms(80);
+cleanup:
+  memzero(tropic_public, sizeof(tropic_public));
```
andrewkozlik:
```suggestion
  memzero(privileged_private, sizeof(privileged_private));
```

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -283,116 +1140,378 @@ static void prodtest_tropic_send_command(cli_t* cli) {
...
+  // TODO: call `check_device_cert_chain()` for FIDO certificate
 
   cli_ok(cli, "");
+
+cleanup:
+  memzero(privileged_private, sizeof(privileged_private));
 }
 
-static void prodtest_tropic_certfido_read(cli_t* cli) {
+static void read_cert(cli_t* cli, uint16_t first_slot, uint16_t slots_count) {
```
andrewkozlik: Inconsistent naming: `read_cert()` vs. `cert_write()`

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5525/commits/39a948d548fab511e729004e40bb6079e208e3d6.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5525/commits/39a948d548fab511e729004e40bb6079e208e3d6.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5525/commits/39a948d548fab511e729004e40bb6079e208e3d6.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -25,14 +25,34 @@
...
 #include "secure_channel.h"
 
+#define FACTORY_PAIRING_KEY_SLOT \
+  PAIRING_KEY_SLOT_INDEX_0  // This key is used by prodtest to inject the
+                            // privileged and unprivileged pairing keys
+#define UNPRIVILEGED_PAIRING_KEY_SLOT \
+  PAIRING_KEY_SLOT_INDEX_1  // This key is used by HSM to inject the attestation
+                            // FIDO key and generate the device key, and by an
+                            // unofficial firwmare with an unlocked bootloader
+#define PRIVILEGED_PAIRING_KEYSLOT \
```
andrewkozlik:
```suggestion
#define PRIVILEGED_PAIRING_KEY_SLOT \
```
(needs replacement in other places as well)

core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -283,116 +1140,378 @@ static void prodtest_tropic_send_command(cli_t* cli) {
...
+
+  ret =
+      lt_session_start(tropic_handle, tropic_public, PRIVILEGED_PAIRING_KEYSLOT,
+                       privileged_private, privileged_public);
+  if (ret != LT_OK) {
+    cli_error(cli, CLI_ERROR,
+              "`lt_session_start()` for privileged key failed with error %d",
+              ret);
+    goto cleanup;
+  }
```
andrewkozlik: This code is duplicated across `cert_write()`, `read_cert()` and twice in `is_paired()`. We can define a function like this:
```c
static bool tropic_handshake(int key_slot) {
  tropic_handshake_state = TROPIC_HANDSHAKE_STATE_0;

  lt_handle_t* tropic_handle = tropic_get_handle();

  curve25519_key tropic_public = {0};
  if (secret_key_tropic_public(tropic_public) != sectrue) {
    cli_error(cli, CLI_ERROR, "`secret_key_tropic_public()` failed.");
    goto cleanup;
  }

  curve25519_key private = {0};
  if (key_slot == PRIVILEGED_PAIRING_KEY_SLOT) {
    if (secret_key_tropic_pairing_privileged(private) != sectrue) {
      cli_error(cli, CLI_ERROR,
                "`secret_key_tropic_pairing_privileged()` failed.");
      goto cleanup;
    }
  } else if (key_slot == UNPRIVILEGED_PAIRING_KEY_SLOT) {
    if (secret_key_tropic_pairing_unprivileged(private) != sectrue) {
      cli_error(cli, CLI_ERROR,
                "`secret_key_tropic_pairing_unprivileged()` failed.");
      goto cleanup;
    }
  } else {
    cli_error(cli, CLI_ERROR,
              "Unknown key_slot.");
    goto cleanup;
  }

  curve25519_key public = {0};
  curve25519_scalarmult_basepoint(public, private);

  lt_ret_t ret =
      lt_session_start(tropic_handle, tropic_public, key_slot,
                       private, public);
  if (ret != LT_OK) {
    cli_error(cli, CLI_ERROR,
              "`lt_session_start()` for privileged key failed with error %d",
              ret);
    goto cleanup;
  }

cleanup:
  memzero(private, sizeof(private));
  return ret == LT_OK;  // TODO set ret correctly above
}
```
- Maybe we can reuse/replace `tropic_init()` from embed/sec/tropic/tropic.c? I didn't look closely enough. The error handling would probably complicate things.
- Probably makes sense to add `FACTORY_PAIRING_KEY_SLOT` to the if-else, so that it can be used in `prodtest_tropic_pair()`, again I didn't look close enough.

andrewkozlik: Also, I don't know how long the handshake takes, but we could optimize this, so that if the handshake is already active we don't repeat it. For example replace `TROPIC_HANDSHAKE_STATE_1` with different states for each key:
`TROPIC_HANDSHAKE_STATE_FACTORY`
`TROPIC_HANDSHAKE_STATE_PRIVILEGED`
`TROPIC_HANDSHAKE_STATE_UNPRIVILEGED`
`TROPIC_HANDSHAKE_STATE_HSM`

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -128,6 +129,73 @@ static void prodtest_secrets_init(cli_t* cli) {
...
 }
 
+#ifdef SECRET_MASTER_KEY_SLOT_SIZE
+static void prodtest_secrets_get_mcu_device_key(cli_t* cli) {
+  if (cli_arg_count(cli) > 0) {
+    cli_error_arg_count(cli);
+    return;
+  }
+
+  ed25519_secret_key mcu_private = {0};
```
andrewkozlik, line 143: Good catch with the incorrect type. Could you please also fix the type in `secret_key_mcu_device_auth(curve25519_key dest)` in secret_keys.h, stm32u5/secret_keys.c and unix/secret_keys.c?

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: I agree with you, and I wanted to refactor it exactly as you suggested, but I considered it a low priority.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5525/commits/25a48cdc97bb82c40301979ebbca16a8b8001035.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5525/commits/25a48cdc97bb82c40301979ebbca16a8b8001035.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5525/commits/25a48cdc97bb82c40301979ebbca16a8b8001035.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: I initially thought there would be some commands that would require the topic to be locked. However, I don't see any now, so I removed the variable in https://github.com/trezor/trezor-firmware/pull/5525/commits/04474e3c260c4da762d4f957076550168aa5c6b8.

core/embed/projects/prodtest/cmd/prodtest_tropic.c
onvej-sl: > How does this work? We don't need to have a handshake completed to write the config?

That is my mistake. Before writing the configuration, we need to perform a handshake even if it's a blank chip. Fixed in https://github.com/trezor/trezor-firmware/pull/5525/commits/04474e3c260c4da762d4f957076550168aa5c6b8.

onvej-sl:
I squashed all the fix-up commits, reordered the commits, and changed the commit messages. The diff between [the old branch](https://github.com/trezor/trezor-firmware/commit/07233879c3d099e1b38d6ffa046ca3cce90a2492) and [the new branch](https://github.com/trezor/trezor-firmware/commit/695cd1aa5a141b47e9389c29888b5219c8f34ac9) is empty.

onvej-sl:
I squashed 15934d0e3f7966c193c8e780636df1084613cf4b and rebased on top of main, that most notably includes https://github.com/trezor/trezor-firmware/pull/5542 and https://github.com/trezor/trezor-firmware/pull/5539.

onvej-sl:
@andrewkozlik Could you please review https://github.com/trezor/trezor-firmware/pull/5525/commits/01675e0d1417d92eb62061a5200566a16222145f?

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -128,6 +133,93 @@ static void prodtest_secrets_init(cli_t* cli) {
...
+
+  if (secret_read(prefixed_certificate, SECRET_MCU_DEVICE_CERT_OFFSET,
+                  sizeof(prefixed_certificate)) != sectrue) {
+    cli_error(cli, CLI_ERROR, "`secret_read()` failed.");
+    return;
+  }
+
+  size_t certificate_length =
+      prefixed_certificate[0] << 8 | prefixed_certificate[1];
+
```
andrewkozlik, line 217: To be on the safe side, I'd check the length here. For example, if the certificate is not stored at all, then this will probably contain 0xFF bytes, leading to a potential heartbleed-like attack. Not sure if `secret_read()` can tell that the certificate is not stored.
```suggestion

  if (certificate_length > sizeof(prefixed_certificate) - prefix_length) {
    cli_error(cli, CLI_ERROR, "Invalid certificate data.");
    return;
  }
```

PR #5528, opened 2025-08-12
SLIP-24 for Solana

ibz:
This PR implements SLIP-24 payment requests for Solana.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -111,6 +111,8 @@ def parametrize_using_common_fixtures(*paths: str) -> "MarkDecorator":
                     skiplist.append(models.T3B1)
                 if skip_model == "t3t1":
                     skiplist.append(models.T3T1)
+                if skip_model == "t3w1":
+                    skiplist.append(models.T3W1)
             if skiplist:
                 extra_marks = [pytest.mark.models(skip=skiplist)]
             else:


tests/device_tests/solana/test_sign_tx.py
```diff
@@ -73,11 +80,48 @@ def test_solana_sign_tx(session: Session, parameters, result):
                 encoded_token=encoded_token,
             )
 
+        if parameters.get("payment_req"):
```
bieleluk, line 84: In the T3T1 test, there is no menu traversing in the summary screen.
The same is also in BTC SLIP-24, so it's probably expected.

core/src/apps/solana/layout.py
```diff
@@ -525,3 +525,63 @@ async def confirm_transaction(
...
+    decimals: int,
+    token: SolanaTokenInfo,
+    fee: Fee,
+    verified_payment_request: PaymentRequest,
+) -> None:
+    from trezor.ui.layouts import confirm_payment_request
+
+    total_amount = format_amount_unit(format_amount(amount, decimals), token.symbol)
+
+    texts = []
```
bieleluk: From the testing point of view, it would be better to explicitly mark the type of the variable.
The same applies to the variables below. (similar to ADA PR)

core/src/apps/solana/predefined_transaction.py
```diff
@@ -159,16 +163,31 @@ async def try_confirm_token_transfer_transaction(
...
-        is_unknown,
-        total_token_amount,
-        transfer_token_instructions[0].decimals,
-        fee,
-        blockhash,
-    )
+
+    if verified_payment_request:
+        if len(transfer_token_instructions) > 1:
+            raise ProcessError("Multiple transfers not supported for payment requests")
```
bieleluk, line 169: Would it make sense to have a test for this scenario? (similar to ADA PR)

tests/common.py
```diff
@@ -112,9 +112,12 @@ def parametrize_using_common_fixtures(*paths: str) -> "MarkDecorator":
                 if skip_model == "t3t1":
                     skiplist.append(models.T3T1)
```
bieleluk, line 113: This is not a part of this PR, but we are missing the new T3W1 model.
Can you, please, add it here? I don't think it's worth its own issue.

bieleluk (commented):
Mostly the same comments as in ADA PR.

core/src/apps/solana/layout.py
ibz: OK e74684345e972e812cc8857f5b23d825815230eb

core/src/apps/solana/predefined_transaction.py
ibz: Yes! 7e770b5ab67c999e0e5612c5b72dd8c24ec2ceb2

tests/common.py
ibz: Sure: 19021a24426600b54cab9fe03ba451620198476c

PR #5530, opened 2025-08-12
fix(core): set model variant in THP DeviceProperties

mmilata:
will return 0 for T2T1, perhaps it should be None :thinking:


core/src/trezor/utils.py
```diff
@@ -382,6 +382,15 @@ def hexlify_if_bytes(data: str | bytes | bytearray | memoryview) -> str:
         raise TypeError("Expected str, bytes, bytearray, or memoryview")
 
 
+if USE_THP:
+
+    def unit_thp_model_variant() -> int:
```
romanz: nit: maybe inline, if it's only used once?

M1nd3r (approved):
LGTM

PR #5533, opened 2025-08-12
Eckhart UI: LED colors

obrusvit:


bieleluk (approved):

bieleluk (changes requested):
The LED doesn't work in the emulator. The device works well.

bieleluk (approved):
Let's create a separate issue for an aforementioned observation and this PR is good to go.

PR #5539, opened 2025-08-13
mcu device cert on T3W1

TychoVrahe:


core/embed/sec/secret/stm32u5/secret.c
```diff
@@ -542,6 +550,45 @@ void secret_unlock_bootloader(void) {
...
+#ifdef SECRET_LOCK_SLOT_OFFSET
+
+secbool secret_lock(void) {
+  uint8_t lock_data[SECRET_LOCK_SLOT_LEN] = {0};
+  return secret_write(lock_data, SECRET_LOCK_SLOT_OFFSET, sizeof(lock_data));
+}
+
+secbool secret_is_locked(void) {
+  uint8_t *header_data =
+      (uint8_t *)flash_area_get_address(&SECRET_AREA, 0, SECRET_HEADER_LEN);
```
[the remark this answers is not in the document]
TychoVrahe: no secret_read checks the header integrity which is specifically not wanted here, and the header len read should never bee out of bounds

cepetr (approved):
LGTM, tested on T3T1

core/embed/sec/secret/stm32f4/secret.c
```diff
@@ -84,18 +84,24 @@ void secret_unlock_bootloader(void) { secret_erase(); }
 void secret_write_header(void) {
   uint8_t header[SECRET_HEADER_LEN] = {0};
   memcpy(header, SECRET_HEADER_MAGIC, SECRET_HEADER_MAGIC_LEN);
-  secret_write(header, SECRET_HEADER_OFFSET, SECRET_HEADER_LEN);
+  ensure(secret_write(header, SECRET_HEADER_OFFSET, SECRET_HEADER_LEN),
+         "secret write header failed");
 }
 
-void secret_write(const uint8_t* data, uint32_t offset, uint32_t len) {
+secbool secret_write(const uint8_t* data, uint32_t offset, uint32_t len) {
```
onvej-sl, line 91: I just want to have it confirmed that I understand how to write the certificate. I assume that:
  * I should use `secret_write()`.
  * It's my responsibility to check that `len <= SECRET_MCU_DEVICE_CERT_SIZE`.

Is that correct? Can `len` be less than `SECRET_MCU_DEVICE_CERT_SIZE`?

TychoVrahe: - yes use secret_write
- yes currently the API is quite dumb, we might want to improve it in future
- yes it can be less, though there is a limitation that you have to use data with size being multiple of 16 bytes. I can also remove this limitation if it causes issues.

onvej-sl: I think it's okay for now. We can improve it in the future.

PR #5540, opened 2025-08-13
Extend THP credentials with app_name

mmilata:
cc @szymonlesisz


core/src/apps/thp/pairing.py
```diff
@@ -105,14 +106,17 @@ async def handle_pairing_request(
...
     if not message.host_name:
         raise DataError("Missing host_name.")
 
     ctx.host_name = message.host_name
+    ctx.app_name = message.app_name
     if __debug__ and not ctx.channel_ctx.should_show_pairing_dialog:
         await _skip_pairing_dialog(ctx)
     else:
         await ctx.show_pairing_dialog()
+        await ctx.write(ThpPairingRequestApproved())
```
romanz, line 119: nit: isn't it also done by `show_pairing_dialog`?
https://github.com/trezor/trezor-firmware/blob/8263514ce2cb7a0d07f6cb942cae4d53b1dc275a/core/src/trezor/wire/thp/pairing_context.py#L152

M1nd3r (approved):
The changes looks good in code, but please check that the UI is ok before merging.

core/src/trezor/wire/thp/ui.py
```diff
@@ -4,16 +4,23 @@
     from trezorui_api import UiResult
 
 
+def _app_on_host(app_name: str | None, host_name: str | None) -> str:
```
romanz, line 7: Can the formatting be done on the host?
Asking since I am not sure if we use (or intend to use) the `app_name`/`host_name` separately elsewhere...

core/src/apps/thp/pairing.py
mmilata: not anymore, i removed the line from `pairing_context.py`

a bit misleading to send data from `show_...` function

core/src/trezor/wire/thp/ui.py
mmilata: the current figma designs use different font/placement for app and host

core/src/apps/thp/pairing.py
M1nd3r: Yes, I was baffled by this change too when I was doing the review. Your design is better, as it reduces the code duplication and it is cleaner 👍🏼

PR #5546, opened 2025-08-13
Reimplement event loop restarts in THP

romanz:
Following #4976, we would like to refactor event loop restart logic:
- https://github.com/trezor/trezor-firmware/pull/4976#discussion_r2216285340
- https://github.com/trezor/trezor-firmware/pull/4976#discussion_r2221912370
- https://github.com/trezor/trezor-firmware/pull/4976#discussion_r2223034618

Slack canvas: https://satoshilabs.slack.com/docs/T0J8V2YBY/F097K1YK2UQ

PR #5548, opened 2025-08-14
refactor(core): move THP checksum validation to `Channel`

romanz:
Also, move channel usage update to `interface_context`.

Keep the last valid reassembled message at `Reassembler.message`. It will be used to implement `ThpChannel` read functionality in #5546.

~Rebased over https://github.com/trezor/trezor-firmware/pull/5563.~

### DIFF
--- a/core/src/trezor/wire/thp/channel.py
+++ b/core/src/trezor/wire/thp/channel.py
@@ -25,7 +25,7 @@ from trezor.wire.errors import WireBufferError
 from . import ENCRYPTED, ChannelState, PacketHeader, ThpDecryptionError, ThpError
 from . import alternating_bit_protocol as ABP
 from . import control_byte, crypto, memory_manager
-from .checksum import CHECKSUM_LENGTH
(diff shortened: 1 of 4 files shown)


core/src/trezor/wire/thp/channel.py
```diff
@@ -86,19 +89,35 @@ def get_next_message(self, packet: memoryview) -> memoryview | None:
...
-        else:
+            return False
+
+        if self.bytes_read > self.buffer_len:
             raise ThpError("read more bytes than expected")
 
+        if not verify_checksum(buffer):
+            return False
+
+        self.message = buffer
```
mmilata, line 101: Nit: Would `assert self.message is None` help here (provided it's set to none in `wire/__init__.py`)?

romanz: 71ce6a1878

romanz: Rebasing this PR over `main` (since it doesn't really depends on https://github.com/trezor/trezor-firmware/pull/5563)

PR #5551, opened 2025-08-14
Fix BootloaderV2Image verification

M1nd3r:


python/src/trezorlib/_internal/firmware_headers.py
```diff
@@ -484,8 +484,14 @@ def format(self, verbose: bool = False) -> str:
     def verify(self, dev_keys: bool = False) -> None:
         digest = self.merkle_root()
 
+        hash_params = self.get_hash_params()
+        hash_fn = hash_params.hash_function
```
obrusvit: Nit: since `hash_params` is not used elsewhere
```suggestion
        hash_fn = self.get_hash_params().hash_function
```

obrusvit (approved):
utACK

python/src/trezorlib/_internal/firmware_headers.py
M1nd3r: changed here: https://github.com/trezor/trezor-firmware/pull/5551/commits/9f2a1b7ebdb5b718fc37766b6c2173f4480e954c

PR #5552, opened 2025-08-14
boot sequence tuning

TychoVrahe:

### DIFF
--- a/core/embed/projects/bootloader/main.c
+++ b/core/embed/projects/bootloader/main.c
@@ -117,6 +117,22 @@ static secbool is_manufacturing_mode(vendor_header *vhdr) {
   return manufacturing_mode;
 }
 
+static void display_touch_init(secbool manufacturing_mode,
+                               secbool *touch_initialized) {
+  display_init(DISPLAY_RESET_CONTENT);
+
+#ifdef USE_TOUCH
+  secbool touch_init_ok = secfalse;
+  touch_init_ok = touch_init();
+  if (manufacturing_mode != sectrue) {
+    ensure(touch_init_ok, "Touch screen panel was not loaded properly.");
+  }
+  if (touch_initialized != NULL) {
+    *touch_initialized = touch_init_ok;
+  }
+#endif
+}
+
 static secbool boot_sequence(void) {
   secbool stay_in_bootloader = secfalse;
 
@@ -159,8 +175,10 @@ static secbool boot_sequence(void) {
 
   uint32_t press_start = 0;
   bool turn_on_locked = false;
-  bool haptic_played = false;
   bool bld_locked = false;
+#ifdef USE_HAPTIC
+  bool haptic_played = false;
+#endif
 
   while (!turn_on) {
     bool btn_down = button_is_down(BTN_POWER);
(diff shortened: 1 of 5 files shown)


core/embed/projects/bootloader/main.c
```diff
@@ -220,31 +234,47 @@ static secbool boot_sequence(void) {
...
-static void display_touch_init(secbool manufacturing_mode,
-                               secbool *touch_initialized) {
-  display_init(DISPLAY_RESET_CONTENT);
+    bool btn_down = button_is_down(BTN_POWER);
+    while (btn_down) {
+      uint32_t progress = systick_ms() - progress_start;
+      if (progress > 1000) {
+        progress = 1000;
+      }
+      screen_bootloader_entry_progress(progress, false);
```
kopecdav: This is just for consideration, since it seems bit strange to me.

When holding btn down for more then 2s -> bld_locked flags turns on and  the progress animation with "Starting bootloader" text starts. If you release the button before the progress animation ends, you still end up in the bootloader. 

progress animation gives you the feeling that something is going on at the background but, but actually nothing does. Maybe the progress animation is not right thing to do, or should we at least wait till the progress animation is over when user release the btn early?

core/embed/projects/bootloader/main.c
```diff
@@ -172,23 +191,18 @@ static secbool boot_sequence(void) {
...
-#ifdef USE_HAPTIC
-        haptic_play(HAPTIC_BOOTLOADER_ENTRY);
-#endif
+      if (elapsed >= 2000) {
         bld_locked = true;
+        break;
       }
 #ifdef USE_HAPTIC
-      else if (elapsed >= 500 && !haptic_played) {
+      if (elapsed >= 500 && !haptic_played) {
```
kopecdav, line 198: flag `turn_on_locked` is set `true` right after the `btn_down` is registered in the loop, but the `HAPTIC_BUTTON_PRESS` feedback is played after 500ms. So if you make a really quick short button press (maybe having the device in your pocket), you can turn on the device without any haptic feedback. 

If this was intended then code is fine.

core/embed/projects/bootloader/main.c
TychoVrahe: this is how i understand the design, tagging @Hannsek and @lapohoda for confirmation that this is how its supposed to be

core/embed/projects/bootloader/main.c
TychoVrahe: this is how i understand the design, tagging @Hannsek and @lapohoda for confirmation that this is how its supposed to be

kopecdav: ok, closing this one.

core/embed/projects/bootloader/main.c
TychoVrahe: after discussion with @Hannsek the loader will be 'full' from start: https://github.com/trezor/trezor-firmware/commit/e7f4cccf4c6adf50fc482724f0a429a986715a80

(leaving the rust implementation of the screen as loader for now as we might reconsider it later, but otherwise more simplifications are possible)

PR #5554, opened 2025-08-14
Reproducible build for nRF

obrusvit:
This PR adds `--nrf` option into the `build-docker.sh` script which triggers a reproducible build of nRF bootloader and firmware. Nix and Poetry dependencies were adjusted as needed

To build only the nRF stuff, use:
```shell
PRODUCTION=0 ./build-docker.sh --nrf --skip-normal --skip-bitcoinonly <tag>
```

TODO:
- [ ] fingerprint -> digest which will be signed
- [ ] wrap functionality from @M1nd3r  https://github.com/trezor/trezor-firmware/tree/m1nd3r/nrf-firmware
  - probably not needed
- [x] should the build happen automatically for `model T3W1` or do we want to keep it separated via `--nrf` option?
  - keeping it as a separate option `--nrf`
- [x] not sure about the environment (nix, poetry) changes and whether we don't want to separate them somehow
  - for now: `cmake`, `ninja` added under `devTools` in `nix.shell`
  - python dependencies (especially `west`) are now a part of `pyproject.toml`.


mmilata:
> should the build happen automatically for model T3W1 or do we want to keep it separated via --nrf option?

Can we instead turn the flag into another --target, and enable it by default on T3W1? 

Also job in common.yml would be nice but can be done in another PR. Looks like it kinda sorta works: https://github.com/trezor/trezor-firmware/actions/runs/17109385101/job/48526693371

shell.nix
```diff
@@ -142,6 +142,8 @@ stdenvNoCC.mkDerivation ({
     ffmpeg_7-headless
     dejavu_fonts
   ] ++ lib.optionals devTools [
+    cmake
+    ninja
```
matejcik, line 146: nit: could we move those to the main section and get rid of `devTools` ?

matejcik (approved):
ACK apart from the nit in comment, which doesn't need to be resolved in this PR

PR #5559, opened 2025-08-15
chore(core): Place props constants to FirmwareUI trait

bieleluk:
Follow-up to https://github.com/trezor/trezor-firmware/pull/5420


obrusvit (commented):
As discussed offline, let's close this one.

bieleluk:
The proposed change is not better than the current solution.

PR #5563, opened 2025-08-16
feat(core): add THP transport-level `PING`/`PONG` messages

romanz:
It will be also used to allow stopping an idle workflow due to an unresponsive host.

### DIFF
--- a/core/src/trezor/wire/thp/__init__.py
+++ b/core/src/trezor/wire/thp/__init__.py
@@ -30,6 +30,9 @@ ACK_MESSAGE = const(0x20)
 CHANNEL_ALLOCATION_REQ = const(0x40)
 _CHANNEL_ALLOCATION_RES = const(0x41)
 _ERROR = const(0x42)
+PING = const(0x43)
+_PONG = const(0x44)
+
 CONTINUATION_PACKET = const(0x80)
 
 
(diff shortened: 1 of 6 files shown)


mmilata:
Is this going to be used outside of automated tests? I guess I'll find out after reviewing #5546 ...

LGTM but would like an ACK from @M1nd3r if he has time.

cc @szymonlesisz - perhaps this might be useful to Suite?

romanz:
We are also planning to use it to detect stale workflows (e.g. due to crashed host application):
- if a new channel handshake takes place while another workflow is running, the device can `PING` the host
- if no `PONG` is received (e.g. after ~200ms), we can fail the existing workflow
- it will allow to support "stale" channel takeover (which was previously implemented using [buffer locking mechanism](https://github.com/trezor/trezor-firmware/blob/469ebc9aeaf317e66cedf7a9b2a057315a03a195/core/src/trezor/wire/thp/memory_manager.py#L119) but is being removed by #5546).

EDIT: [Slack discussion](https://satoshilabs.slack.com/archives/C078GRAK58U/p1754387732562949).

mmilata:
>  the device can PING the host

But this PR only implements the host->device direction, right?

romanz:
> >  the device can PING the host
> 
> But this PR only implements the host->device direction, right?

Yes.

M1nd3r:
LGTM - I did not have the time to check it thoroughly, but it seems that @romanz implemented the changes I suggested in our discussion.

PR #5567, opened 2025-08-16
test(core): send `Cancel` only on T1B1

romanz:
[no changelog]


core/src/trezor/wire/__init__.py
```diff
@@ -109,6 +109,10 @@ async def handle_session(iface: WireInterface) -> None:
         finally:
             # Wait for all active workflows to finish.
             await workflow.join_all()
+            if __debug__:
+                import apps.debug
+
+                await apps.debug.close_session()
```
mmilata: Since this is `if USE_THP:` block, shouldn't we also call this function when using codec_v1?

romanz: Sorry, I meant to include only 4d3b151e09715a6f65e88a1defd3641ec2ad409a in this PR... will rebase again over `main`.

romanz: > shouldn't we also call this function when using codec_v1?

You're right - but I thought to split that into a different PR :)

romanz: Rebased over `main` -> https://github.com/trezor/trezor-firmware/commit/cb599c786399e3f063ea285491f2d99501e352ed.

romanz: > Since this is `if USE_THP:` block, shouldn't we also call this function when using codec_v1?

Opened https://github.com/trezor/trezor-firmware/pull/5577.

PR #5577, opened 2025-08-18
feat(core): wait for debuglink task before restarting Codec v1 event loop

romanz:
Following #5546, but enabled for Protocol v1 (instead of v2 - which requires more work).

Note: https://github.com/trezor/trezor-firmware/pull/5577/commits/a40fb9fea0a4614a6d95bae100d777a848cff349 is required to avoid scheduling delay when processing `DebugLinkGetState` messages, due to `loop.race()` usage in https://github.com/trezor/trezor-firmware/pull/5577/commits/106ca1331de3958b6a2093d1f191bfddb90e3ed3 (which is an implicit assumption of the existing DebugLink loop implementation).

### DIFF
--- a/core/src/trezor/wire/__init__.py
+++ b/core/src/trezor/wire/__init__.py
@@ -167,9 +167,11 @@ else:
                     if not do_not_restart:
                         # Wait for all active workflows to finish.
                         await workflow.join_all()
-                        # Let the session be restarted from `main`.
                         if __debug__:
-                            log.debug(__name__, "loop.clear()", iface=iface)
+                            import apps.debug
+
+                            await apps.debug.close_session()
+                        # Let the session be restarted from `main`.
                         loop.clear()
                         return  # pylint: disable=lost-exception
 


core/src/apps/debug/__init__.py
```diff
@@ -428,6 +433,26 @@ async def dispatch_WipeDevice(msg: WipeDevice) -> None:
...
+
+        global _EXIT_FLAG
+
+        _EXIT_FLAG = True
+        _EXIT_BOX.put(None)
+        if layout_change_box.is_empty():
+            # make sure `DebugLinkGetState` won't get stuck
+            layout_change_box.put(None)
+
+        res = await loop.race(_SESSION_TASK, loop.sleep(5000))
```
mmilata: Named constant please?

core/src/trezor/loop.py
```diff
@@ -336,7 +336,7 @@ def _finish(self, task: Task, result: Any) -> None:
         if not self.finished:
             self.finished = True
             self.exit(task)
-            schedule(self.callback, result)
+            _step(self.callback, result)
```
mmilata, line 336: Looks scary, can you convince me that it does not break anything? (tests passing definitely suggest so)

romanz: IIUC, calling `schedule()` will result in the task being pushed into `loop._queue` (which may delay its actual invocation via `_step`). 
In general, it shouldn't be an issue - so many tests pass without this change.

However, when we use `DebugLinkGetState(wait_layout=NEXT_LAYOUT)` followed by another wirelink command, current implementation seems to rely on the fact `DebugLinkGetState` will be processed **before** the wirelink command (in order to correctly capture the layout being changed).
Unfortunately, it may be delayed when I add the `race()` here:
https://github.com/trezor/trezor-firmware/blob/24d73c5d5f45985d0cdeee87011c78d894901068/core/src/apps/debug/__init__.py#L476

I think that this change shouldn't affect other `race()` callers (except for maybe reducing the latency of the callback invocation).

CC: @matejcik, following the discussion in #5166 (which was more complex solution to this issue)

core/src/apps/debug/__init__.py
romanz: https://github.com/trezor/trezor-firmware/pull/5577/commits/24d73c5d5f45985d0cdeee87011c78d894901068

romanz:
[Core / Click tests (T2T1, noasan, en) (pull_request)](https://github.com/trezor/trezor-firmware/actions/runs/17090909171/job/48464630431?pr=5577) failure is fixed by https://github.com/trezor/trezor-firmware/commit/fbd36c85206971572a5ef351e6059a3714796265 (merged into `main`).

core/src/trezor/loop.py
matejcik: we're too deep in the weeds for "break anything" to be a meaningful concept. invoking `_step` directly changes some behaviors, that _hopefully_ are invisible in production firmware; and very likely shifts some raceyness from one place to another.

the main thing this changes is stack ordering; previously, a lot of things would run from the root `loop.run()`, now some of them will also be deeper under, uhh, finalizer invocation? they shouldn't be able to _escape containment_, but we may get a significantly more confusing stack trace in some situations

i say if tests pass then ship it

mmilata (approved):

romanz:
Ran device tests on T3T1 @ 24d73c5d5f45985d0cdeee87011c78d894901068:
```
== 1450 passed, 202 skipped, 6 deselected, 12 xfailed in 11178.89s (3:06:18) ===
```

romanz:
Rebasing over `main`.

romanz:
Fixing a small UI fixture diff introduced by https://github.com/trezor/trezor-firmware/pull/5577/commits/79dcd7fb11f3dcca15500d15a11199a2980bd93f.

PR #5578, opened 2025-08-19
Ignore re-transmitted THP payloads on host

romanz:


romanz:
https://github.com/trezor/trezor-firmware/actions/runs/17062307226/job/48371828632?pr=5578 seems to be a flake (unrelated to this PR).

python/src/trezorlib/transport/thp/control_byte.py
```diff
@@ -14,6 +14,8 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
+from typing import Optional
```
M1nd3r, line 17: nit: Is Optional needed?

I woud rather see int|None notation, if possible.

M1nd3r (commented):
I do not have the time to check it all now. I have one nit, otherwise the parts I checked look ok.

python/src/trezorlib/transport/thp/control_byte.py
romanz: I agree - but it seems that `int | None` is not support on Python 3.9:

```
python ./../tools/pyright_tool.py
...
thp/control_byte.py:55: - error: Alternative syntax for unions requires Python 3.10 or newer (reportGeneralTypeIssues)
```

romanz: IIUC, [Python 3.9 should be EOL soon](https://peps.python.org/pep-0596/#lifespan), so we will be able to deprecate the old-style syntax.

romanz:
> does Suite have to implement these changes too?

I think that Suite probably has implemented it (as part of [the ABP algorithm](https://www.notion.so/satoshilabs/THP-Specification-2-1-203dc5260606804192aecaa58fb961ca?source=copy_link#203dc526060680268c8cccd5a50d871b)) - right @szymonlesisz?

PR #5587, opened 2025-08-19
do not disconnect BLE when rebooting to firmware

TychoVrahe:
This PR fixes a bug that caused disconnection of BLE when deinitializing interface and thus, caused unexpected disconnection when jumping to firmware.

Along with that, an edge situation is fixed where leaving pairing mode in bootloader left the device advertising even if the paring was launched from bootloader state that did not allow connection to host. For this, new BLE driver state is introduced to keep track whether advertisement should be restarted after disconnect (end of KEEP_CONNECTED mode) , and a new command that resets this flag when needed.

### DIFF
--- a/core/embed/io/ble/stm32/ble.c
+++ b/core/embed/io/ble/stm32/ble.c
@@ -58,6 +58,7 @@ typedef struct {
   bool status_valid;
   bool accept_msgs;
   bool reboot_on_resume;
+  bool restart_adv_on_disconnect;
   uint8_t busy_flag;
   bool pairing_allowed;
   bool pairing_requested;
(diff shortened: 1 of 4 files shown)


core/embed/io/ble/stm32/ble.c
```diff
@@ -287,18 +288,16 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
...
     }
   }
 
-  if (msg.peer_count > 1 && drv->peer_count <= 1) {
-    // new bond
-    if (msg.connected && drv->mode_requested == BLE_MODE_KEEP_CONNECTION) {
-      drv->mode_requested = BLE_MODE_CONNECTABLE;
-    }
+  if (drv->mode_requested == BLE_MODE_CONNECTABLE &&
+      !drv->restart_adv_on_disconnect) {
```
kopecdav, line 299: This condition is ok for single point BLE, but for multipoint you should also check the drv->connected flag. In case there is a connected device you should go to BLE_MODE_KEEP_CONNECTION instead.

kopecdav (changes requested):

core/embed/io/ble/stm32/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/8a6b56ca23637b56624690e5044e90541b3833fd

PR #5590, opened 2025-08-19
chore(core/eckhart): enable menu separators by default

bieleluk:


core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu.rs
```diff
@@ -116,16 +116,16 @@ impl<T: MenuItems> VerticalMenu<T> {
             total_height: 0,
             offset_y: 0,
             offset_y_max: 0,
-            separators: false,
+            separators: true,
```
obrusvit: I don't see it used anywhere without the separators, so let's remove the bool completely.

Also remove the `no_separators` method. FTR, if we wanted that method, it would be called `without_separators`

bieleluk: https://github.com/trezor/trezor-firmware/pull/5590/commits/665c862512b028033b8ceb79e5124f68a4093795

PR #5592, opened 2025-08-20
remove BLE multipoint

TychoVrahe:
This PR removes the BLE multipoint behavior. Advertising should only run when no device is connected, and maximum connections is limited to 1.

To enable at least some effective switching from one device to another, 'user disconnect' flag is introduced in advertisement data. This flag indicates that previously connected host should not attempt to automatically reconnect.

BLE TX power is increased to +4dBm.

(bootloader and nRF application dev. binaries are updated)


core/embed/io/ble/stm32/ble.c
kopecdav, line 1: In `ble_start_pairing()` function it would make sense to return `false` in case anybody tries to issue pairing command and the `peer_count > 0`. Or we should remove the old peer before trying to connect the new one 

```
static bool ble_start_pairing(ble_command_t *command) {
  ble_driver_t *drv = &g_ble_driver;

  if (!drv->initialized) {
    return false;
  }

#ifndef BLE_MULTIPOINT          <---- NEW code
 if(drv->peer_count > 0){
   return false;                                
 }
#endif                                          <---- NEW code


  uint16_t retry_cnt = 0;
  irq_key_t key;

  bool connected = drv->connected;
  while (connected) {
    retry_cnt++;
    if (retry_cnt > 10) {
      // too many retries, give up
      return false;
    }

    ble_send_disconnect(drv);

    systick_delay_ms(20);  // wait for disconnect to complete

    key = irq_lock();
    connected = drv->connected;
    irq_unlock(key);
  }
```

kopecdav, line 1: This piece of code, in `ble_process_rx_msg_status()` is probably not necessary for singlepoint version and maybe within the BLE_MULTIPOINT compilation option, but also should not do any harm.. 

line 258 ~ 270
``` 
    if (memcmp(drv->connected_addr, msg.connected_addr,
               sizeof(drv->connected_addr)) != 0 ||
        drv->connected_addr_type != msg.connected_addr_type) {
      // address changed
      memcpy(drv->connected_addr, msg.connected_addr,
             sizeof(drv->connected_addr));
      drv->connected_addr_type = msg.connected_addr_type;

      ble_event_t event = {.type = BLE_CONNECTION_CHANGED};
      memcpy(event.data, drv->connected_addr, sizeof(drv->connected_addr));
      tsqueue_enqueue(&drv->event_queue, (uint8_t *)&event, sizeof(event),
                      NULL);
    }
```

core/embed/io/ble/stm32/ble.c
```diff
@@ -271,9 +281,11 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
     drv->mode_current = BLE_MODE_OFF;
```
kopecdav, line 281: This condition will probably change for SINGLEPOINT:

```
  if ((msg.advertising && !msg.advertising_whitelist) ||
      (msg.connected && drv->pairing_allowed)) {
    drv->mode_current = BLE_MODE_PAIRING;
```

msg.connected && drv->pairing_allowed should not end up in next round of pairing. 

Also drv->pairing_allowed flag assigned sooner in this function is bit fuzzy. firstly you set the drv->pairing_flag according to the `drv->mode_current` here:

```
      drv->pairing_requested = false;
      if (drv->mode_current == BLE_MODE_PAIRING) {
        drv->pairing_allowed = true;
      } else {
        drv->pairing_allowed = false;
      }
```

and then you set the drv->mode_current here based on drv->pairing_allowed

core/embed/io/ble/stm32/ble.c
```diff
@@ -300,6 +312,15 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
     drv->mode_requested = BLE_MODE_OFF;
   }
 
+#ifdef BLE_MULTIPOINT
+  if (msg.peer_count > 1 && drv->peer_count <= 1) {
```
kopecdav: This piece of code was removed in the original commit in `tychovrahe/ble/jump_disconnect` (https://github.com/trezor/trezor-firmware/commit/d936fd23c0167a2cb1444e4b082d2c1204f4784c) but returned here. 

It looks like it do not make sense to compare `msg.peer_count > 1 && drv->peer_count <= 1` cause drv->peer_count is equal to msg.peer_count since line 278.

core/embed/io/ble/stm32/ble.c
TychoVrahe: no we still do allow pairing of multiple devices, we just don't allow simultaneous connections. 

though similar condition might make sense for `BLE_SWITCH_ON` command, where we would reject if peer_count == 0, so i will consider it and potentially make a separate PR

TychoVrahe: i'd still keep this, although not entirely likely it can happen that i.e., STM is frozen for some time and misses disconnect/connect event pair and this is what re-synchronizes the state

core/embed/io/ble/stm32/ble.c
TychoVrahe: removed https://github.com/trezor/trezor-firmware/commit/2634582d8a006476c6d1b37843a23e22afc1e1ed

we will need something like that if we eveer return to multipoint but this version doesn't work like you said so we will have to figure it out differently

core/embed/io/ble/stm32/ble.c
TychoVrahe: I remember now:
- i want the mode to stay PAIRING if the device is connected - this describes the state that unpaired device connected, but hasn't started the bonding process yet.

But there might be some weirdness when bonded device connects in pairing mode. Let me come back to this in followup  PR.

core/embed/io/ble/stm32/ble.c
kopecdav: ah ok, clear now

PR #5602, opened 2025-08-21
stronger haptic effect on power on

TychoVrahe:
This PR increases the strength of haptic effect played when the device is powered on.


STew790:
**QA OK**
I thought it felt different. It is indeed stronger.

PR #5605, opened 2025-08-21
Adjust backlight normal settings to 85

kopecdav:
Update of default normal backlight settings from 150->85 according to request from design

### DIFF
--- a/core/embed/projects/bootloader/bootui.h
+++ b/core/embed/projects/bootloader/bootui.h
@@ -25,7 +25,13 @@
 
 #include "rust_ui_bootloader.h"
 
+#ifdef TREZOR_MODEL_T3W1
+#define BACKLIGHT_NORMAL 85
+#else
 #define BACKLIGHT_NORMAL 150
+#endif
+
+#define BACKLIGHT_LOW 45
 
 // Displays a warning screen before jumping to the untrusted firmware
 //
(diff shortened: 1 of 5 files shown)


core/embed/projects/bootloader/bootui.h
```diff
@@ -25,7 +25,7 @@
 
 #include "rust_ui_bootloader.h"
 
-#define BACKLIGHT_NORMAL 150
```
TychoVrahe, line 28: This (and the prodtest values) affect other models too, but i believe that this should only be done for T3W1 (or ratheer eckhart UI), so please add an ifdef. The prodtest values should probably be extracted to some single place.

kopecdav: Fixup here https://github.com/trezor/trezor-firmware/pull/5605/commits/ab6878ad7a8348b9dc5e0a4bea3be62776513db1. It will do for now, but we should probably deal with this more elegantly in the future. How about to set the default values for every model in its model_<xx>.h file? That may be a good single reference for both bootloader & prodtest.

TychoVrahe: longer term we might want to move such UI related stuff to rust where we have more clean UI separation.

Easier for short term might be to introduce some UI specific definition files in C

TychoVrahe (approved):

obrusvit (approved):

PR #5610, opened 2025-08-21
fix(core): increase `MAX_SUBSCREENS` a bit

romanz:
Otherwise, [entering the menu when connected via BLE results in a RSOD](https://drive.google.com/file/d/1RPv42auw1Ya6XWB0NJ3zy8ZD54m4AR0J).

```
[23:04:01.509] === FATAL ERROR
[23:04:01.509] Location: src/ui/layout_eckhart/firmware/device_menu_screen.rs:373
[23:04:01.509] Message: unwrap failed
```

Reproduced on 375d6e10ca.


obrusvit (approved):
utACK

PR #5612, opened 2025-08-21
A few small THP-related fixups

romanz:


core/src/apps/thp/pairing.py
```diff
@@ -112,11 +118,8 @@ async def handle_pairing_request(
 
     ctx.host_name = message.host_name
     ctx.app_name = message.app_name
-    if __debug__ and not ctx.channel_ctx.should_show_pairing_dialog:
-        await _skip_pairing_dialog(ctx)
-    else:
-        await ctx.show_pairing_dialog()
-        await ctx.write(ThpPairingRequestApproved())
+    await ctx.show_pairing_dialog()
```
M1nd3r: nit: Maybe should `show_pairing_dialog` be part of `ui`? Same as `show_connection_dialog`

tests/device_tests/thp/test_pairing.py
```diff
@@ -406,9 +406,8 @@ def test_credential_phase(client: Client) -> None:
     protocol._noise.noise_protocol.cipher_state_encrypt.n = 250
 
     protocol._send_message(ButtonAck())
-    with pytest.raises(exceptions.ThpError) as e:
+    with pytest.raises(exceptions.DecryptionFailed):
```
M1nd3r, line 409: nit: Maybe we should check for the "expected specific ThpError type" that is raised - so that we are sure we are not catching some other type.

tests/device_tests/thp/test_pairing.py
```diff
@@ -457,9 +456,8 @@ def test_credential_phase(client: Client) -> None:
     protocol._noise.noise_protocol.cipher_state_encrypt.n = 100
 
     protocol._send_message(ButtonAck())
-    with pytest.raises(exceptions.ThpError) as e:
+    with pytest.raises(exceptions.DecryptionFailed):
```
M1nd3r, line 459: nit: Same as 50 lines above...

tests/device_tests/thp/test_pairing.py
romanz: Should be handled by `pytest.raises(exceptions.DecryptionFailed)`.

tests/device_tests/thp/test_pairing.py
romanz: Should be handled by `pytest.raises(exceptions.DecryptionFailed)`.

core/src/apps/thp/pairing.py
romanz: 0900245614740fb706afb37a190f299cdabc0442

M1nd3r (approved):

PR #5613, opened 2025-08-21
fix(python): correct CLI error message on uninitialized device

romanz:
[no changelog]

### DIFF
--- a/python/src/trezorlib/cli/__init__.py
+++ b/python/src/trezorlib/cli/__init__.py
@@ -307,6 +307,9 @@ class TrezorConnection:
             sys.exit(1)
         except exceptions.FailedSessionResumption:
             sys.exit(1)
+        except exceptions.DerivationOnUninitaizedDeviceError:
+            click.echo("Device is not initialized.")
+            sys.exit(1)
         except Exception:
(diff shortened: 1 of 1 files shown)


ibz (approved):

matejcik:
we already have `Failure(NotInitialized)`
the exception name is horrible and these two should coerce to the same thing

please call the exception `NotInitializedError`, make sure it returns `"Device is not initialized."` as `__str__()`, and in `session.call()` make sure that `FailureType.NotInitialized` is raised as this type (similar to what `Cancelled` is doing)

python/src/trezorlib/cli/__init__.py
```diff
@@ -307,6 +307,9 @@ def session_context(
             sys.exit(1)
         except exceptions.FailedSessionResumption:
             sys.exit(1)
+        except exceptions.DerivationOnUninitaizedDeviceError:
+            click.echo("Device is not initialized.")
+            sys.exit(1)
         except Exception:
             click.echo("Failed to find a Trezor device.")
```
matejcik, line 314: we definitely want to emit the exception in the log
maybe it's also a good idea to say `f"(error was: {e})"` or something?

PR #5616, opened 2025-08-22
prepare embedding production nRF application

TychoVrahe:
- switch based on BOOTLOADER_DEVEL
- west is updated to include mcuboot with T3W1 production keys


TychoVrahe:
superseded by https://github.com/trezor/trezor-firmware/pull/5648

PR #5617, opened 2025-08-22
fix(core): don't raise `ThpError` on low-level errors

romanz:
`ThpError` ends up being sent over the protocol as `MessageType_Failure`, but in case of a low-level problem (e.g. retransimission timeout), it is probably not the correct thing to do.

[no changelog]


M1nd3r (approved):
utACK

Note: Originally, some of the `ThpError`s were used for flow handling - raising them should have not resulted in `Failure` message being sent. With the new flow handling, removing those errors is wanted.

PR #5618, opened 2025-08-22
fix(core): avoid blocking THP when send buffer is full

romanz:
[no changelog]


romanz:
IIUC, the previous THP buffer locking implementation solved this issue by [unlocking the THP buffers after 200ms](https://github.com/trezor/trezor-firmware/blob/12c045d4b57fe22a3de36be0591987f4cb58dc29/core/src/trezor/wire/thp/memory_manager.py#L18) - so maybe the write timeout in this PR should shorter (currently I've set it to 5 seconds).

PR #5619, opened 2025-08-22
Debug console

### LINKED ISSUE #4253: Kernel fails to build with `PYOPT=0`

cepetr:
This PR introduces a new `dbg_console` kernel module that introduces new logging layer that allows 
printing debug output via ITM SWO, SEGGER SystemView, or USB VCP.

The new `dbg_printf()` can now be called anywhere in the codebase (including interrupt routines, when not using the VCP target).

The output target can be selected with the new `DBG_CONSOLE` argument from the command line when building the firmware. The same approach is used across all layers: coreapp, kernel, secmon, bootloader, boardloader, and prodtest.

```
make build_firmware DBG_CONSOLE=<xxx>
```

where `<xxx>` can be 
- `SWO` - logging via ITM SWO (requires ST-Link or J-Link)
- `SYSTEM_VIEW` - logging via SEGGER SystemView (requires J-Link)
- `VCP` - logging via USB VCP
- empty - logging is disabled

For the `firmware` build:
- if `PYOPT=0`, `DBG_CONSOLE` defaults to `VCP`
- if `PYOPT=1`, `DBG_CONSOLE` defaults to an empty string (logging disabled)

`DBG_CONSOLE` be also used when building `boardloader`, `bootloader` and `prodtest`. For these targets, it defaults to an empty string (logging disabled).

The following table show valid combinations:

|           | bootloader | boardloader | prodtest | coreapp | kernel | secmon |
|-----------|------------|-------------|----------|---------|--------|--------|
|SWO        |  ✓         | ✓           | ✓        | ✓       |  ✓     | ✓      |
|SYSTEM_VIEW|  ✓         | ✓           | ✓        | ✓       |  ✓     | ✓      |
|VCP        |  ✓         |             |          | ✓       |  ✓     |        |

Resolves #4253 - to enable Optiga logging, `USE_OPTIGA_LOGGING` must be defined.

NOTE:  Optiga logging works only with `SWO` or `SYSTEM_VIEW` on T3W1 (since the Optiga driver is implemented in secmon on T3W1)


core/embed/sys/dbg/inc/sys/dbg_console.h
```diff
@@ -0,0 +1,76 @@
...
+
+#pragma once
+
+#include <trezor_types.h>
+
+#include <stdarg.h>
+
+#ifdef KERNEL_MODE
+
+/*
```
TychoVrahe: comment style here is not correct (use /** and @brief like elsewhere)

core/site_scons/models/stm32f4_common.py
```diff
@@ -72,7 +72,8 @@ def stm32f4_common_files(env, features_wanted, defines, sources, paths):
         "embed/sec/secret/stm32f4/secret_keys.c",
         "embed/sec/storage/stm32f4/storage_salt.c",
         "embed/sec/time_estimate/stm32/time_estimate.c",
-        "embed/sys/dbg/stm32/dbg_printf.c",
+        "embed/sys/dbg/dbg_console.c",
```
TychoVrahe: i think it would be better to use dbg console conditionally based on some USE_ feature. We could skip the relevant syscalls in produciton  builds

core/embed/sys/dbg/inc/sys/dbg_console.h
cepetr: fixed, see 143f92fe907632f7aef580c78878b18d9193f363

core/site_scons/models/stm32f4_common.py
cepetr: yes, i agree, see c0ad29edfa2e939e4c502f1894c8eb73834323ee

PR #5620, opened 2025-08-22
fix(core): handle ACK bit on THP data messages

romanz:
Both THP handshake and encrypted transport messages contain an ACK bit, which should not be ignored.


romanz:
Closing, this requires implementing THP ACK piggybacking support for both firmware and host.

PR #5621, opened 2025-08-23
Two types of auto lock delay in the Eckhart device menu

bieleluk:
This PR adds a new _Auto-lock_ submenu with  two types of auto-lock delay:
- on battery
- on USB

Design: [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=11777-2996&t=D5689VWQ3QMJnASw-0)

Prerequisities
- [x] Implement the second delay when the device runs on battery

TODO
- [x] replace mocked on-battery delay setting


obrusvit:
This PR now stages also: https://github.com/trezor/trezor-firmware/pull/5767 and https://github.com/trezor/trezor-firmware/pull/5774 into `main`

PR #5622, opened 2025-08-23
Allow cross-sign with ETH mainnet for all non-ETH networks

### LINKED ISSUE #5134: Allow EVM testnets to use standard derivation path
Relate [Slack channel](https://satoshilabs.slack.com/archives/C02V2PSDNA2/p1744131752275819)

obrusvit:
- Previously only non-standard networks (not SLIP-44 60 or 1) could cross-sign with Ethereum mainnet. Now any network that isn't Ethereum mainnet can use Ethereum derivation paths
- This is a workaround enabling access to networks like Hyperliquid that have conflicting chain IDs and can't claim official SLIP-44 registration.

closes #5134

### DIFF
--- a/common/tests/fixtures/ethereum/sign_tx.json
+++ b/common/tests/fixtures/ethereum/sign_tx.json
@@ -156,6 +156,46 @@
                 "sig_r": "79f079c4db54be4041329e096f8b47c04bcbc8ef6d5684748436ea8ff16ed5e5",
                 "sig_s": "32cf560c7fc98ac1041b27154b0787170ac93181124c41d7310323b40ca46cf1"
             }
+        },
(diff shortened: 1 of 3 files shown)


core/.changelog.d/5134.fixed
```diff
@@ -0,0 +1 @@
+Allow using Ethereum mainnet addresses on all non-Ethereum networks. This enables access to networks like Hyperliquid that use conflicting chain IDs and cannot obtain official SLIP-44 registration.
```
matejcik: nit: this should go into `changed`

obrusvit: Done on rebase

PR #5625, opened 2025-08-24
fix(python): improve `trezorctl` THP error handling

romanz:
Following https://github.com/trezor/trezor-firmware/pull/5613.

A few error examples:
```md
$ trezorctl ping 123 
## no device is connected
Failed to connect: TransportException (No Trezor device found)

$ trezorctl ping 123
## don't confirm THP pairing
Failed to connect: Cancelled

$ trezorctl btc get-address -n m/44h/0h/0h/0/0 
## fail to get address on uninitialized device
Failed to connect: TrezorFailure (NotInitialized: Device is not initialized)

$ trezorctl ping 123
## enter invalid pairing code
Enter code from Trezor: 000000
Failed to connect: TrezorFailure (FirmwareError: Unexpected Code Entry Tag)

$ trezorctl ping 123 -b
## don't confirm Ping
Action was cancelled.
```

### DIFF
--- a/python/src/trezorlib/transport/thp/protocol_v2.py
+++ b/python/src/trezorlib/transport/thp/protocol_v2.py
@@ -20,7 +20,6 @@ import logging
 import os
 import typing as t
 from binascii import hexlify
-from enum import IntEnum
 
 from noise.connection import Keypair, NoiseConnection
 
@@ -44,11 +43,6 @@ if t.TYPE_CHECKING:
 MT = t.TypeVar("MT", bound=protobuf.MessageType)
 
 
-class TrezorState(IntEnum):
-    UNPAIRED = 0x00
-    PAIRED = 0x01
-
-
 class ProtocolV2Channel(Channel):
     channel_id: int
     sync_bit_send: int
@@ -56,9 +50,8 @@ class ProtocolV2Channel(Channel):
     handshake_hash: bytes
     device_properties: bytes
 
-    _has_valid_channel: bool = False
     _features: messages.Features | None = None
-    trezor_state: int = TrezorState.UNPAIRED
+    _is_paired: bool = False
 
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


python/src/trezorlib/transport/thp/protocol_v2.py
```diff
@@ -72,11 +65,12 @@ def __init__(
...
             self._do_channel_allocation(retries=MAX_RETRANSMISSION_COUNT)
-            self.trezor_state = self._do_handshake(credential=credential)
+            LOG.debug("THP channel allocated: %04x", self.channel_id)
+            self._do_handshake(credential=credential)
+            LOG.debug("THP handshake done: is_paired=%s", self._is_paired)
 
     def get_channel(self) -> ProtocolV2Channel:
-        if not self._has_valid_channel:
-            raise RuntimeError("Channel is invalidated")
+        assert self._is_paired
```
M1nd3r: Q: What is the rationale behind replacing `RuntimeException` raising with an assert?

tests/device_tests/thp/test_pairing.py
```diff
@@ -487,8 +488,7 @@ def test_credential_request_in_encrypted_transport_phase(client: Client) -> None
     protocol._send_message(ThpEndRequest())
     protocol._read_message(ThpEndResponse)
 
-    session = client.get_seedless_session()
-
+    session = SessionV2.seedless(client)  # doesn't do pairing
```
M1nd3r: nit: This comment is a bit misleading. It took me a short while to realize that it "doesn't do pairing" not because we are calling the message **without** pairing, but because the protocol is already paired.

I would rather set the state explicitly/correctly, i.e. put 
`protocol._is_paired = True` above the seedless session creation and create the seedless session from the client as it was before. (If for some reason the `client` has different `protocol` instance, set the `client.protocol=protocol`).

python/src/trezorlib/transport/thp/protocol_v2.py
romanz: Oops... I changed it during debugging, and forgot to change it back (with the correct error message).

tests/device_tests/thp/test_pairing.py
romanz: Sounds good - fixed in f9a6e75fb0.

python/src/trezorlib/transport/thp/protocol_v2.py
romanz: Fixed in f9a6e75fb0

PR #5574, opened 2025-08-18
Implement LED switching

bieleluk:
Enable LED switching off, similar to the haptic feedback. 
Required by [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=9814-10190&t=UmICvytuX7Xhh8xy-0)

### LED driver:
- new `enabled` field in the driver struct
- new functions
  - `rgb_led_set_enabled`
  - `rgb_led_get_enabled`

### Device menu
- new input param for `show_device_menu` FwUI function
- new `DeviceMenuResult` variant
- new device menu item for LED switching

### Syscalls
- new syscalls:
  - `SYSCALL_RGB_LED_SET_ENABLED`
  - `SYSCALL_RGB_LED_GET_ENABLED`

### Python
- new trezorio module `rgb_led` with `rgb_led_set_enabled` function
- new constant `USE_RGB_LED`

### Storage
- new entry for storing whether the LED is enabled or not
- new `set_rgb_led` and `get_rgb_led()` functions

TODO

- [x] rename commit message
- [ ] add changelog entry

PR #5561, opened 2025-08-15
prepare BLE api for bond management

TychoVrahe:
fixes #5537 

switch on/off is already available.

PR #5547, opened 2025-08-13
WIP eckhart menu actions

bieleluk:
Description

This PR implements the device menu for Eckhart according to [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=9814-10122&t=IBG8ZcDd9BcuUDBx-0)

Steps

- [x] https://github.com/trezor/trezor-firmware/pull/5574
- [x] https://github.com/trezor/trezor-firmware/pull/5561
- [x] https://github.com/trezor/trezor-firmware/pull/5627
- [x] https://github.com/trezor/trezor-firmware/pull/5628
- [x] https://github.com/trezor/trezor-firmware/pull/5629

TODO:
- [x] Replace the Bluetooth setting mocked parameter and implement the Bluetooth turning on/off action.
- [x] Replace the mocked paired/connected devices and implement all connection actions: https://github.com/trezor/trezor-firmware/pull/5681.
- [x] Replace the mocked Bluetooth version parameter.
- [x] Auto-lock delay for battery/USB charged device: [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=11777-2996&t=IBG8ZcDd9BcuUDBx-0)
- [x] Long-name Bluetooth devices in
  - `Button` content - temporary fix: https://github.com/trezor/trezor-firmware/pull/5665
  - `VerticalMenuScreen` subtitle
- [x] Long label name in
  - `Button` subtext - temporary fix: https://github.com/trezor/trezor-firmware/pull/5665
- [x] Click tests
- [x] Menu refreshing after incoming Bluetooth connect/disconnect event: https://github.com/trezor/trezor-firmware/pull/5691

PR #5628, opened 2025-08-24
Eckhart device menu settings

bieleluk:
This PR is a part of https://github.com/trezor/trezor-firmware/pull/5547
It correctly shows and implements actions for device menu settings according to [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=9814-10122&t=6cHKR3uDmwRDqZ8f-0)

Settings

- Bluetooth - mocked until https://github.com/trezor/trezor-firmware/pull/5561 is merged
- Security settings - fully functional
  - Pin code settings - fully functional
  - Wipe code settings - fully functional 
- Device settings - fully functional

Notifications

- backup failed
  - 1st place in the root, red color
  - triggers repeated backup
- pin unset
  - root menu, yellow color (to match notifications)
  - shortcut for setting a pin in the security settings

Prerequisities:

- https://github.com/trezor/trezor-firmware/pull/5574
- https://github.com/trezor/trezor-firmware/pull/5627


core/src/apps/homescreen/device_menu.py
```diff
@@ -105,23 +96,74 @@ async def handle_device_menu() -> None:
...
+        from trezor.ui.layouts import confirm_action
+
+        turned_on = ble.is_connected()
+
+        await confirm_action(
+            "ble__settings",
+            TR.words__bluetooth,
+            TR.ble__enable if turned_on else TR.ble__disable,
+        )
         pass  # TODO implement bluetooth handling
```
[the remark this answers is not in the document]
bieleluk: It is currently mocked because the underlying BLE firmware is not implemented yet, as the intro describes

core/src/apps/homescreen/device_menu.py
```diff
@@ -105,23 +96,74 @@ async def handle_device_menu() -> None:
...
     # Bluetooth
     elif menu_result is DeviceMenuResult.Bluetooth:
+        from trezor.ui.layouts import confirm_action
+
+        turned_on = ble.is_connected()
+
+        await confirm_action(
+            "ble__settings",
+            TR.words__bluetooth,
+            TR.ble__enable if turned_on else TR.ble__disable,
```
romanz: nit: shouldn't it be the other way around?

```suggestion
            TR.ble__disable if turned_on else TR.ble__enable,
```

core/src/apps/homescreen/device_menu.py
```diff
@@ -105,23 +96,74 @@ async def handle_device_menu() -> None:
             raise RuntimeError(f"Unknown menu {result_type}, {index}")
     # Bluetooth
     elif menu_result is DeviceMenuResult.Bluetooth:
+        from trezor.ui.layouts import confirm_action
+
+        turned_on = ble.is_connected()
```
romanz, line 101: Can the device be disconnected from other peers, but the BLE will still be enabled?
(i.e. is it the right method? or do we need another one?)

core/src/apps/homescreen/device_menu.py
```diff
@@ -3,52 +3,27 @@
...
-    else:
-        raise ActionCancelled  # user cancelled request number prompt
-
-
 async def handle_device_menu() -> None:
     from trezor import strings
 
-    # TODO: unify with notification handling in `apps/homescreen/__init__.py:homescreen()`
     failed_backup = (
         storage_device.is_initialized() and storage_device.unfinished_backup()
```
romanz: nit: consider calling `storage_device.is_initialized()` once - and caching its result
(same for the other `storage_device` methods)

core/src/apps/homescreen/device_menu.py
```diff
@@ -84,7 +71,11 @@ async def handle_device_menu() -> None:
     )
     # Root menu
     if menu_result is DeviceMenuResult.BackupFailed:
-        pass  # TODO implement backup failed handling
+        from apps.management.backup_device import perform_backup
+
+        assert storage_device.unfinished_backup()
```
romanz: nit: consider adding `and failed_backup` to the above `if` condition (to avoid the `assert`)

(same for the rest of the `assert`s below)

core/src/apps/homescreen/device_menu.py
bieleluk: We need to use the method from https://github.com/trezor/trezor-firmware/pull/5561.
The current implementation is mocked, and a _TODO_ below says that it should be updated.
Everything regarding the BLE functionality is mocked.

core/src/apps/homescreen/device_menu.py
bieleluk: I did that intentionally because I'm implementing the click tests that should stress test the whole implementation: https://github.com/trezor/trezor-firmware/pull/5630
The point is that if we include combined statements in the if statement above, the test will silently finish without failing.

On the other hand, the assert should never occur because the device buttons are rendered only if the conditions are already met, i.e.
`failed_backup = storage_device.is_initialized() and storage_device.unfinished_backup()`

If you insist on extending the conditions, I can do that, but then I need a different mechanism for ensuring the validity of click tests.

core/src/apps/homescreen/device_menu.py
bieleluk: :face_holding_back_tears:  https://github.com/trezor/trezor-firmware/pull/5628/commits/70cff2b3d77c1f75f9e3f9af930c4ed12398e338

core/src/apps/homescreen/device_menu.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5628/commits/fe83d74e99ee3c8d3fb37c7d4cb2bb2e8b145dc8

core/src/apps/homescreen/device_menu.py
```diff
@@ -70,13 +63,14 @@ async def handle_device_menu() -> None:
         "device_menu",
     )
     # Root menu
+    is_initialized = storage_device.is_initialized()
```
romanz: nit: do we need to fetch it again here?
(not sure it can change when interacting with the menu above)

core/src/apps/homescreen/device_menu.py
```diff
@@ -70,13 +63,14 @@ async def handle_device_menu() -> None:
...
     if menu_result is DeviceMenuResult.BackupFailed:
         from trezor.messages import WipeDevice
         from trezor.ui.layouts import raise_if_cancelled
 
         from apps.management.wipe_device import wipe_device
 
-        assert storage_device.unfinished_backup() and storage_device.is_initialized()
+        assert storage_device.unfinished_backup() and is_initialized
```
romanz: nit: can we replace it by `assert failed_backup`?

core/src/apps/homescreen/device_menu.py
romanz: >  if we include combined statements in the if statement above, the test will silently finish without failing.

Wouldn't it fail due to the `raise` at the last `else` clause?
https://github.com/trezor/trezor-firmware/blob/fe83d74e99ee3c8d3fb37c7d4cb2bb2e8b145dc8/core/src/apps/homescreen/device_menu.py#L229

core/src/apps/homescreen/device_menu.py
bieleluk: We don't
https://github.com/trezor/trezor-firmware/pull/5628/commits/8f15b740ff0f49749c311dccecdd8ae5c743646c

core/src/apps/homescreen/device_menu.py
bieleluk: Reworked it completely
1. moved assert conditions into the if statementt
https://github.com/trezor/trezor-firmware/pull/5628/commits/b2c48f3c91347c209e521732acc64b47c038bb01
2. cached repeated conditions
https://github.com/trezor/trezor-firmware/pull/5628/commits/7496c283552e7cc37825543d7e1d8f5d5d1a4795

romanz (approved):
Many thanks!

PR #5635, opened 2025-08-25
fix progress bar delay when installing firmware

TychoVrahe:
The loader was inhibited until for first block, now it starts after user confirmation (or automatic confirmation)


cepetr (approved):
Tested on T3W1

PR #5636, opened 2025-08-25
add T3W1 signed vendorheaders

TychoVrahe:
just adding the vendorheaders


obrusvit (approved):
I checked that the vendorheaders added match the ones provided on a separate channel.

M1nd3r (approved):
sha256sum of added files are equal to the checksums of expected files

PR #5639, opened 2025-08-25
expose currently connected BLE host address to application

TychoVrahe:
[no changelog[

### DIFF
--- a/core/embed/rust/src/trezorhal/ble/micropython.rs
+++ b/core/embed/rust/src/trezorhal/ble/micropython.rs
@@ -142,6 +142,23 @@ extern "C" fn py_connection_flags() -> Obj {
     unsafe { util::try_or_raise(block) }
 }
 
+// Return (addr_bytes, addr_type) if connected, otherwise None.
+// addr_bytes: bytes of length 6
+// addr_type: integer as provided by bt_le_addr_t (e.g., 0=public, 1=random)
+extern "C" fn py_connected_addr() -> Obj {
+    let block = || {
+        if !is_connected() {
+            return Ok(Obj::const_none());
+        }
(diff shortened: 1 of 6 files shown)


core/embed/rust/src/trezorhal/ble/micropython.rs
```diff
@@ -130,6 +130,25 @@ extern "C" fn py_connection_flags() -> Obj {
...
+// Return (addr_bytes, addr_type) if connected, otherwise None.
+// addr_bytes: bytes of length 6
+// addr_type: integer as provided by bt_le_addr_t (e.g., 0=public, 1=random)
+extern "C" fn py_connected_addr() -> Obj {
+    if !is_connected() {
+        return Obj::const_none();
+    }
+
+    let a = connected_addr();
+    unsafe {
```
romanz: nit: can we rewrite it without `unsafe`?

```rust
extern "C" fn py_connected_addr() -> Result<Obj, Error> {
    if !is_connected() {
        return Ok(Obj::const_none());
    }

    let a = connected_addr();
    let addr_obj = Obj::try_from(&a.addr[..])?;
    let type_obj = Obj::from(a.type_);
    (addr_obj, type_obj).try_into()
}
```

romanz (commented):
Thanks!

core/embed/rust/src/trezorhal/ble/micropython.rs
TychoVrahe: i think we need to return Obj here so maybe like this https://github.com/trezor/trezor-firmware/commit/942055366ecda087d3db4688e2fbc609b6e4cddf ? as it matches the stylo of other functions here

PR #5640, opened 2025-08-25
WIP [Eckhart]: Fuse passphrase and label keyboards

bieleluk:
- Fuse the passphrase and the label keyboards into one  full-screen component:
  - `StringKeyboard`
- The input component is now a generic `StringInput` trait
- Two implemented input types:
  - `PassphraseInput`
  - `LabelInput`
  
  There should be no major UI differences; just the repeated code has been reduced.


core/embed/rust/src/ui/layout_eckhart/firmware/keyboard/string.rs
```diff
@@ -24,41 +14,26 @@ use super::super::{
...
-    OneLine,
-    /// One line with the last pending character.
-    OneLineWithMarker,
-    /// The complete string is shown in the input area.
-    Complete,
-}
-
 pub enum StringKeyboardMsg {
-    Confirmed,
+    Confirmed(ShortString),
```
mmilata, line 25: IIRC we intentionally didn't include the string in the message but I don't recall why. To keep the msg structs small? To not make unnecessary copies of the passphrase? Just thinking out loud ...

bieleluk: The `ShortString` is there due to the `request_passphrase` flow. If we used just a screen for that, as we do for PIN, we could have gotten rid of that.

mmilata (approved):
LGTM, maybe let's merge after freeze?

PR #5641, opened 2025-08-25
Allow preempting stale THP channels

romanz:
In case an existing channel becomes unresponsive, we allow another channel on the same interface to preempt the active workflow (by restarting the event loop).

If there are no other active channels, the existing channel will not be preempted.

In the future, we should use [the newly introduced `Ping`/`Pong` messages](https://github.com/trezor/trezor-firmware/pull/5563) - but since they are not supported by Suite, a channel can be preempted only 1 second after the last THP device->host write (giving the host enough time to respond).

# Notes: 
- the "preempting" message will be responded with `TRANSPORT_BUSY` so the host needs to retry to "acquire" its channel (after the stale workflow is cancelled and the event loop is restarted).
- initially it was done by time-based buffers locking, but was dropped in #5546.

### DIFF
--- a/core/src/trezor/wire/__init__.py
+++ b/core/src/trezor/wire/__init__.py
@@ -104,10 +104,16 @@ if utils.USE_THP:
         if __debug__:
             _THP_CHANNELS.append(ctx._channels)
         try:
-            channel = await ctx.get_next_message()
+            while (channel := await ctx.get_next_message()) is None:
+                if __debug__:
+                    # happens if another interface is active and using THP buffers.
+                    log.error(__name__, "Another interface is active", iface=iface)
+
             while await received_message_handler.handle_received_message(channel):
                 pass
         finally:
+            if __debug__:
+                log.debug(__name__, "Finished THP session", iface=iface)
             # Wait for all active workflows to finish.
             await workflow.join_all()
             if __debug__:
--- a/core/src/trezor/wire/thp/channel.py
+++ b/core/src/trezor/wire/thp/channel.py
@@ -1,4 +1,5 @@
 import ustruct
+import utime
 from micropython import const
 from typing import TYPE_CHECKING
 
@@ -22,6 +23,7 @@ from storage.cache_thp import (
 )
 from trezor import protobuf, utils, workflow
 from trezor.loop import Timeout, race, sleep
+from trezor.wire.context import UnexpectedMessageException
 
 from ..protocol_common import Message
 from . import (
@@ -62,6 +64,10 @@ _MIN_RETRANSMISSION_COUNT = const(2)
 # It allows restarting the event loop to handle other THP channels.
 _WRITE_TIMEOUT_MS = const(5_000)
 _WRITE_TIMEOUT = sleep(_WRITE_TIMEOUT_MS)
+
+# Preempt a stale channel if another channel becomes active and we allowed enough time for the host to respond.
+# It allows interrupting a "stuck" THP workflow using a different channel on the same interface.
+_PREEMPT_TIMEOUT_MS = const(1_000)
 
 
 class Reassembler:
@@ -134,6 +140,13 @@ def verify_checksum(buffer: memoryview) -> memoryview | None:
     return None
 
 
+class ChannelPreemptedException(UnexpectedMessageException):
+    """Raising this exception should restart the event loop."""
+
+    def __init__(self) -> None:
+        super().__init__(msg=None)
+
+
 class Channel:
     """
     THP protocol encrypted communication channel.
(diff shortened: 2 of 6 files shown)


core/src/trezor/wire/__init__.py
```diff
@@ -104,7 +104,10 @@ async def handle_session(iface: WireInterface) -> None:
         if __debug__:
             _THP_CHANNELS.append(ctx._channels)
         try:
-            channel = await ctx.get_next_message()
+            while (channel := await ctx.get_next_message()) is None:
+                if __debug__:
```
M1nd3r, line 108: please. make sure that this passes on non-debug build - as the body of the while cycle might get removed

core/src/trezor/wire/thp/channel.py
```diff
@@ -134,6 +140,13 @@ def verify_checksum(buffer: memoryview) -> memoryview | None:
     return None
 
 
+class ChannelPreempted(UnexpectedMessageException):
```
M1nd3r: nit: I would like to see that `ChannelPreempted` is an exception. Meaning either renaming it to `ChannelPreemptedException` or moving it to some `exceptions` module, so it can be called as `exceptions.ChannelPreempted`.

M1nd3r (approved):
utACK

core/src/trezor/wire/__init__.py
romanz: I think it should work - tested on emulator and on device.

> as the body of the while cycle might get removed

Is it a MicroPython issue? How can I reproduce it?

core/src/trezor/wire/thp/channel.py
romanz: Sounds good - changed in ab932cb8cd4588eb9b70a1f89cd020e6416994f2.
Also fixed the logging in e0ac2dc09be26705b4ab3a924ff7d03ad971dca4.

core/src/trezor/wire/__init__.py
M1nd3r: Thanks for checking

IIRC the `if __debug__` is resolved before the actual compilation - and sometimes this leads to the removal of the code completely - which could result in some (mostly compile time) errors. If such issue arises, it  can be easily fixed e.g. by adding a `pass`  keyword at the end. Example of this is in `wipe_device.py - line 56`.

I did not check the behaviour in this place  as I did not have time to test it. I just want to make sure that someone does test it before merging 😄

core/src/trezor/wire/thp/channel.py
M1nd3r: > Sounds good - changed in https://github.com/trezor/trezor-firmware/commit/ab932cb8cd4588eb9b70a1f89cd020e6416994f2.

✅ Thanks!

> Also fixed the logging in https://github.com/trezor/trezor-firmware/commit/e0ac2dc09be26705b4ab3a924ff7d03ad971dca4.

✅ ❓ It is ok. However, I am not sure if it cannot be misunderstood  - when a channel has multiple sessions (e.g. multiple wallets) that this refers to another session **on the same channel**. I am not sure how exactly is it now, but not long ago the "buffer ownership" was to a channel and not to a session. If it is still the case, I would rather see a message  "Device in use by other channel." or "Session on another channel is in progress." or something (explicit) like that. Feel free to keep the current solution, as the potential for misunderstanding is small (imo).

core/src/trezor/wire/__init__.py
M1nd3r: Note: I have tried to reproduce the issue (in other places) but I was unsuccessful. It is possible that it might have been fixed in the meantime.

I am quite sure that it was happening when creating frozen modules - see `micropython/__init__.py::generate::generate_frozen_module` - where it was happening to me - in cases of blocks of `TYPE_CHECKING` and `USE_THP`

core/src/trezor/wire/thp/channel.py
romanz: > I am not sure how exactly is it now, but not long ago the "buffer ownership" was to a channel and not to a session. If it is still the case, <...>

Yes - a single channel is allowed to "take" the buffers (via `THP_BUFFERS_PROVIDER`).
https://github.com/trezor/trezor-firmware/blob/e0ac2dc09be26705b4ab3a924ff7d03ad971dca4/core/src/trezor/wire/thp/interface_context.py#L81-L87

romanz: WDYT about `"Another interface is active"` (ea9bda03442a82a16c8d4c518a3a117abd6da0c6)?

M1nd3r: LGTM

romanz:
Squashed the fixups -> 131d2dd50d.

PR #5646, opened 2025-08-27
add battery SoC to features message

TychoVrahe:
[no changelog]


common/protob/messages-management.proto
```diff
@@ -126,6 +126,7 @@ message Features {
     optional bool haptic_feedback = 52;         // haptic feedback is enabled
     optional RecoveryType recovery_type = 53;   // what type of recovery we are in. NB: this works in conjunction with recovery_status
     optional uint32 optiga_sec = 54;            // Optiga's security event counter.
+    optional uint32 soc = 55;                   // Battery state of charge
```
obrusvit:
```suggestion
    optional uint32 soc = 55;                   // Battery state of charge (0 - 100%)
```

Alternatively, we can name it `soc_pct`

TychoVrahe: fixed: https://github.com/trezor/trezor-firmware/commit/c01648e4cd0e96802dc02f686efa5e981d4e7694

PR #5648, opened 2025-08-27
add model identifier into nrf app image

TychoVrahe:
Adds model identifier into nrf app image. Updates mcuboot submodule. Some debug settings is cleaned a bit.

After mcuboot update, i also prepare for embedding production/dev nrf app binary (supersedes https://github.com/trezor/trezor-firmware/pull/5616)


cepetr (approved):
LGTM

As we discussed in person:
1) `MODEL_IDENTIFIER` is defined in multiple places. It would be cleaner to introduce a new module that provides this identifier and use it in mcuboot and other apps, but that seems like overkill for now.
2) mcu-tools has a bug https://github.com/mcu-tools/mcuboot/issues/1846
. That is why we need to pass the model identifier to imgtool as a big-endian value (`--custom-tlv 0x00A3 0x54335731`). We can remove this workaround once mcu-tools is fixed.

PR #5651, opened 2025-08-27
feat(core): allow preempting stale THP channels on all interfaces

romanz:
Following #5641.

PR #5656, opened 2025-08-27
test(core): reimplement some THP transport-related tests

romanz:
Following #5651.


tests/device_tests/thp/test_basic.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # There should be no response for continuation packet (starts with "?" only)
+    write_padded(transport, b"? Cont packet")
+
+
+def test_v2_unallocated(client: Client):
+    transport = client.protocol.transport
+
+    # A message to unallocated THP channel 0x789a should result in an error
+    write_padded(transport, bytes.fromhex("04789a000c001122334455667796643c6c"))
```
M1nd3r, line 38: small nit: personally, I am not a huge fan of magic strings. I would like to see the constants/variables that are used explicitly - and then concatenated into the payload.
- 04 - ctrl_byte means ENCRYPTED_TRANSPORT
- 789a - means channel id (written in comment above, so better)
- 000c  - payload len (including CRC checksum)
- 0011223344556677 - actual test payload
- 96643c6c - is the CRC checksum 

Constants for control_bytes are in `trezorlib/transport/thp/message_header.py`,

```
ENCRYPTED_TRANSPORT = 0x04
_ERROR = 0x42
```
CRC and payload_len can be imo computed dynamically, or can be set as constants...

M1nd3r (approved):
Tests are ok -> Approved 👍🏼 

Small nit regarding clarity for future readers.

tests/device_tests/thp/test_basic.py
romanz: Sounds good, will fix in a separate PR (https://github.com/trezor/trezor-firmware/issues/5675).

PR #5657, opened 2025-08-27
Implement tropic-keyfido-read.

andrewkozlik:
Untested. @onvej-sl please test.

### DIFF
--- a/core/embed/projects/prodtest/cmd/prodtest_tropic.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_tropic.c
@@ -29,7 +29,9 @@
 #include <sec/secret.h>
 #include <sec/secret_keys.h>
 
+#include "ecdsa.h"
 #include "memzero.h"
+#include "nist256p1.h"
 
 #include "common.h"
 #include "fw_CPU.h"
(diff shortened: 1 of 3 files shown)


core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -1577,6 +1579,50 @@ static void prodtest_tropic_certdev_read(cli_t* cli) {
...
+  ecc_key_origin_t origin = 0;
+  lt_ret_t ret =
+      lt_ecc_key_read(tropic_get_handle(), slot,
+                      &public_key[1], &curve_type, &origin);
+  if (ret != LT_OK || curve_type != CURVE_P256) {
+    cli_error(cli, CLI_ERROR, "lt_ecc_key_read error %d.", ret);
+    return;
+  }
+
+#ifdef SECRET_KEY_MASKING
```
onvej-sl: I expect the build to fail if `SECRET_KEY_MASKING` is not defined, because `masking_key` is not used.

In fact, I am in favor of completely removing the ifdef because it's unintuitive to me that defining some symbol can cause a function argument to be ignored. An alternative would be to remove `masking_key` from the function header if `SECRET_KEY_MASKING` is not defined.

PR #5659, opened 2025-08-27
end pairing mode when a bonded device connects

TychoVrahe:
fixes issue described in https://github.com/trezor/trezor-firmware/pull/5592#discussion_r2300911817

### DIFF
--- a/core/embed/io/ble/stm32/ble.c
+++ b/core/embed/io/ble/stm32/ble.c
@@ -279,7 +279,7 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
 
   ble_mode_t prev_mode = drv->mode_current;
   if ((msg.advertising && !msg.advertising_whitelist) ||
-      (msg.connected && drv->pairing_allowed)) {
+      (msg.connected && drv->pairing_allowed && !msg.flags.bonded_connection)) {
     drv->mode_current = BLE_MODE_PAIRING;
   } else if (msg.advertising) {
     drv->mode_current = BLE_MODE_CONNECTABLE;
@@ -297,6 +297,12 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
 
   drv->busy_flag = msg.busy_flag;
   drv->peer_count = msg.peer_count;
+
+  if (msg.connected && msg.flags.bonded_connection &&
+      drv->mode_requested == BLE_MODE_PAIRING) {
+    // bonded device connected in pairing mode - end pairing
+    ble_pairing_end(drv);
+  }
 
   if (prev_mode == BLE_MODE_PAIRING && drv->mode_current != BLE_MODE_PAIRING) {
     if (drv->mode_requested == BLE_MODE_PAIRING) {
--- a/core/embed/io/ble/stm32/ble_comm_defs.h
+++ b/core/embed/io/ble/stm32/ble_comm_defs.h
@@ -31,7 +31,10 @@ typedef struct {
 
   uint8_t peer_count;
   uint8_t busy_flag;
-  uint8_t reserved;
+  struct {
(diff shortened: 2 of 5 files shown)


core/embed/io/ble/stm32/ble.c
```diff
@@ -298,6 +299,12 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
   drv->busy_flag = msg.busy_flag;
   drv->peer_count = msg.peer_count;
 
+  if (msg.connected && msg.flags.bonded_connection != 0 &&
```
kopecdav: Nitpick: bonded connection is a flag so maybe using simple `msg.flags.bonded_connection` would be better then comparing it with non-zero value. For the first sight it evokes that bonded_connection may gain larger values then 0 or 1.

core/embed/io/ble/stm32/ble.c
```diff
@@ -279,7 +279,8 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
 
   ble_mode_t prev_mode = drv->mode_current;
   if ((msg.advertising && !msg.advertising_whitelist) ||
-      (msg.connected && drv->pairing_allowed)) {
+      (msg.connected && drv->pairing_allowed &&
+       msg.flags.bonded_connection == 0)) {
```
kopecdav: Here as well, `!msg.flags.bonded_connection` looks more clear to me.

core/embed/io/ble/stm32/ble.c
```diff
@@ -279,7 +279,8 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
 
   ble_mode_t prev_mode = drv->mode_current;
   if ((msg.advertising && !msg.advertising_whitelist) ||
-      (msg.connected && drv->pairing_allowed)) {
+      (msg.connected && drv->pairing_allowed &&
```
kopecdav: Its still bit unclear to me what is a purpose of `drv->pairing_allowed` flag here. It looks like the BLE connection process has two stages.

1. T3W1 starts pairing, got connected to the external device but the device is not bonded yet => So we have to keep pairing. 
2. T3W1 establish a secured connection with the external device (got bonded) so it can switch to `BLE_MODE_KEEP_CONNECTION`.

If my understanding is correct, do we really need `pairing_allowed` in this condition? If you are connected but not bonded, the only option would be to continue pairing, right? So, `(msg.connected && !msg.flags.bonded_connection)` should be sufficient, shouldn't it?

kopecdav (changes requested):

core/embed/io/ble/stm32/ble.c
TychoVrahe: if you are not in pairing mode, and non-bonded device connects, you'd end up in pairing mode which is not correct. This should probably be prevented directly on nRF by refusing such connection, but from perspective of this driver i think its safer to assume it might happen

core/embed/io/ble/stm32/ble.c
TychoVrahe: here https://github.com/trezor/trezor-firmware/commit/31412571bc89ac3b8d88b43924809e50bafe8b5f  i changed the flag type to bool so that the comparison is type-correct.

core/embed/io/ble/stm32/ble.c
TychoVrahe: dtto https://github.com/trezor/trezor-firmware/commit/31412571bc89ac3b8d88b43924809e50bafe8b5f

core/embed/io/ble/stm32/ble.c
kopecdav: Great, thanks!

kopecdav (approved):

PR #5661, opened 2025-08-28
feat(core): use MAC addresses in BLE menu

romanz:
THP `hostName` support will be added in a following PR.


core/embed/rust/src/trezorhal/ble/micropython.rs
```diff
@@ -349,19 +346,17 @@ pub static mp_module_trezorble: Module = obj_module! {
     ///     """
     Qstr::MP_QSTR_connection_flags => obj_fn_0!(py_connection_flags).as_obj(),
 
-    /// def get_bonds() -> list[tuple[bytes, int], ...]:
+    /// def get_bonds() -> list[bytes], ...]:
```
obrusvit:
```suggestion
    /// def get_bonds() -> list[bytes]:
```

romanz: Thanks -> fixed in https://github.com/trezor/trezor-firmware/commit/38e08315f2d0b7d285c13d75329233c6ef41dd59.

core/src/apps/homescreen/device_menu.py
```diff
@@ -8,16 +8,37 @@
 MAX_PAIRED_DEVICES = 4
 
 
+def _format_mac(ble_addr: bytes) -> str:
+    return ":".join(f"{byte:02x}" for byte in ble_addr)
```
bieleluk: nit: The upper-case representation is a bit more common IMO.
```suggestion
    return ":".join(f"{byte:02X}" for byte in ble_addr)
```

romanz: Sounds good -> 2bea581576

romanz:
When testing with `A1:B2:C3:D4:E5:F6` MAC, the `PYOPT=0` emulator build failed with:
```
=== FATAL ERROR
Location: src/ui/util.rs:66
Message: Text too long: 'A1:B2:C3:D4:E5:F6'
===
ERROR: Text too long: 'A1:B2:C3:D4:E5:F6'
reboot (with timeout)
```
@bieleluk @ibz WDYT?

EDIT: will rebase over #5665.

romanz:
Squashing and rebasing.

PR #5662, opened 2025-08-28
set BLE connection parameters differently in bootloader and firmware

TychoVrahe:
No need for high speed transfers in firmware, firmware can't handle fast BLE communication anyway.


core/embed/io/ble/stm32/ble.c
```diff
@@ -336,6 +344,17 @@ static void ble_process_rx_msg_status(const uint8_t *data, uint32_t len) {
     drv->mode_requested = BLE_MODE_OFF;
   }
 
+  // in case connection speed differs, send a command
+#ifdef BOOTLOADER
```
kopecdav: I would maybe consider to put this decision to the upper level, and use the connection_set_high_speed/low_speed api instead to be more visible. Also, I think it would be good to put a comment why this is actually needed.

kopecdav (changes requested):

core/embed/io/ble/stm32/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/d67a5cc21ba590d987372aa2e5efd42f28c5dfdf

core/embed/io/ble/stm32/ble.c
```diff
@@ -1167,6 +1164,15 @@ bool ble_unpair(const bt_le_addr_t *addr) {
   return result;
 }
 
+void ble_set_high_speed(bool enable) {
+  ble_driver_t *drv = &g_ble_driver;
+  if (!drv->initialized) {
+    return;
+  }
+
+  drv->high_speed = enable;
```
kopecdav, line 1174: I think that this should be locked, since there could be an asynchronous access from `ble_process_rx_msg_status()`

core/embed/io/ble/stm32/ble.c
kopecdav: Thanks!

core/embed/io/ble/stm32/ble.c
TychoVrahe: fixed: https://github.com/trezor/trezor-firmware/commit/ba721acd799313ae4c458ad60730584aca320f0d

PR #5663, opened 2025-08-28
Add nrf keys to ModelKeys

M1nd3r:


python/src/trezorlib/firmware/models.py
```diff
@@ -69,6 +69,13 @@ def hash_params(self) -> FirmwareHashParameters:
         return MODEL_HASH_PARAMS_MAP[self]
 
 
+def get_nrf_keys(model: Model, dev_keys: bool = False) -> t.Sequence[bytes]:
+    keys = model.model_keys(dev_keys).nrf_keys
+    if len(keys) < 1:
```
romanz:
```suggestion
    if not keys:
```

M1nd3r: changed here: https://github.com/trezor/trezor-firmware/pull/5663/commits/ef9d3e102db8b23c1193cacd7a94c742b3df9903

python/src/trezorlib/firmware/models.py
```diff
@@ -82,6 +89,7 @@ class ModelKeys:
     firmware_sigs_needed: int
     secmon_keys: t.Sequence[bytes]
     secmon_sigs_needed: int
+    nrf_keys: t.Sequence[bytes]
```
romanz, line 92: Should we also specify `nrf_sigs_needed`? (similar to the fields above)

M1nd3r: It should not be necessary. There are some reasons (some stronger, some weaker):
- the verification of `nrf_fw` signatures is done differently than for other software (boardloader/bootloader/secmon/fw), as the `nrf_fw_header` is created in format specified not by us but by the zephyr
- the value of `nrf_sigs_needed` would not be used anywhere else - it brings no additional value
- the "sigs_needed" is basically hardcoded in the verification function - the verification is not dynamic (cannot handle variable number of signatures)
- potential programmer's error (trying to get nRF keys for model without nRF chip) is handled by the `get_nrf_keys(...)` function
- if the signing scheme ever changes in the future, this variable can be easily added

romanz (approved):

python/src/trezorlib/firmware/models.py
romanz: Sounds good, thanks for the explanation!

PR #4907, opened 2025-04-11
Allow multiline text in Eckhart buttons

ibz:

Issue #4897, opened 2025-04-09
Detect overflowing text and crash during tests


ibz:
Text not fitting on the screen is currently hard to detect and we only notice this during manual testing, which is often very late - especially for languages other than English, which are less tested during development.

The goal of this issue is to have the tests and the emulator crash whenever some text does not fit on the screen. Production builds should be unaffected.

Some work has been already done but there are additional places where text can overflow.

* [x] #4226 did it for `Button`s that consist of image + text (menu items in Delizia UI basically)
* [x] #4767 does it for `Label`
* [x] #4907 does it for Eckhart menu items (including text and subtext)
  * [ ] however it doesn't do it for buttons that specifically ask for single line text
* [ ] #4918 does it for `shape::Text`
* [ ] Enable the tweaks from #4767 on Delizia also (see [this comment](https://github.com/trezor/trezor-firmware/pull/4767#discussion_r2044576778))
* [ ] Buttons without image on Delizia/Bolt
* [ ] get rid of `text_multiline` - this can probably be replaced with something else now 
* [ ] look into rendering of `Paragraphs`
* [ ] Eckhart
  * [ ] [hint](https://github.com/trezor/trezor-firmware/pull/4775#discussion_r2097541862)
  * [ ] [text screen](https://github.com/trezor/trezor-firmware/pull/4775#discussion_r2097664840)

PR #5665, opened 2025-08-28
Allow device name to overflow

ibz:
1. In the device menu, the device name could be longer than what fits in the subtitle, so we should let it overflow rather than crash!
2. In the pair & connect menu, the device name could be longer than what can fit in the text, so we should let that overflow as well!

Related to #4897

Basically we need to revert the debug-only check for the subtext introduced by #4907 in the case of the device name, while letting it crash in any other case to catch copy errors...

And also we should allow *menu items with text and subtext* have `single_line` text - which does not use `split_two_lines` which checks for overflow...

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
@@ -97,7 +97,7 @@ pub enum DeviceMenuMsg {
 struct MenuItem {
     text: TString<'static>,
     subtext: Option<(TString<'static>, Option<&'static TextStyle>)>,
-    subtext_overflow: bool,
+    allow_overflow: bool,
     stylesheet: &'static ButtonStyleSheet,
     connection_status: Option<bool>,
     action: Option<Action>,
@@ -112,7 +112,7 @@ impl MenuItem {
         Self {
             text,
             subtext: None,
-            subtext_overflow: false,
+            allow_overflow: false,
             stylesheet: MENU_ITEM_NORMAL,
             action,
             connection_status: None,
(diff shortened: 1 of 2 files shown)


romanz:
Tested on #5661 - it seems that the device name may still overflow:
- `PYOPT=0` https://drive.google.com/file/d/1AoVpDGqcJxXc5s5jDdjRCS-yUXX9tdpM
- `PYOPT=1` https://drive.google.com/file/d/1QWAf3Gtr_8A5fTeKbCc98PyMkeP21HHX

IIUC, the device name is the `text` (and the device connection status is the `subtext`).

ibz:
> the device name is the `text` (and the device connection status is the `subtext`)

Discussed on Slack, leaving a summary here for posterity as well.

There are two different cases:
 1. in the Settings menu where the device name is a subtext and it can possibly get very long
 2. in the Pair & Connect menu, where the device name is the main button text and again it can get very long

Both should be allowed to extend longer than the screen permits without crashing, and probably both should become marquees in this case.

In the case of the main button title, we cannot split it in two lines because it has no spaces when it is a MAC address. One solution would be to have `:` as a separator.

For now we just prevent it from crashing, and we might go either with that (although it is not a long term solution, since we will not be using MAC addresses in the future) or with marquee.

Either way, just making it not crash in these two cases is a step forward for now.

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -606,12 +616,21 @@ impl DeviceMenuScreen {
                     } else if let Some((subtext, subtext_style)) = item.subtext {
                         let subtext_style =
                             subtext_style.unwrap_or(&theme::TEXT_MENU_ITEM_SUBTITLE);
-                        Button::new_menu_item_with_subtext(
-                            item.text,
-                            *item.stylesheet,
-                            subtext,
-                            subtext_style,
-                        )
+                        if item.allow_overflow {
```
bieleluk: Nit: the following would be a bit more compact:
```suggestion
                        let ctor = if item.allow_overflow {
                            Button::new_single_line_menu_item_with_overflowing_subtext
                        } else {
                            Button::new_menu_item_with_subtext
                        };
                        ctor(item.text, *item.stylesheet, subtext, subtext_style)
```

PR #5666, opened 2025-08-28
 `Union` busting

M1nd3r:
[Trezor Core coding style](https://github.com/trezor/trezor-firmware/blob/9f7df35d7ecc64fa1ec7f7a4527fbc827201931f/docs/core/misc/codestyle.md?plain=1#L18) prefers (Python 3.10) `|` notation over `Union[...]`. This PR unifies all core under `|`. (+ removes the last and only mention of `Union` in `/tests`)

### DIFF
--- a/tests/click_tests/common.py
+++ b/tests/click_tests/common.py
@@ -10,7 +10,7 @@ from .. import translations as TR
 if t.TYPE_CHECKING:
     from trezorlib.debuglink import DebugLink, LayoutContent
 
-    AllActionsType = t.List[t.Union[str, t.Tuple[str, ...]]]
+    AllActionsType = list[str | tuple[str, ...]]
 
 
 # Passphrases and addresses for both models


core/src/apps/cardano/helpers/chunks.py
```diff
@@ -6,12 +6,9 @@
...
 
 if TYPE_CHECKING:
-    from typing import Generic, TypeVar, Union
+    from typing import Generic, TypeVar
 
-    Chunk = Union[
-        messages.CardanoTxInlineDatumChunk,
-        messages.CardanoTxReferenceScriptChunk,
-    ]
+    Chunk = messages.CardanoTxInlineDatumChunk | messages.CardanoTxReferenceScriptChunk
```
matejcik, line 11: just a note here that this will be invalid at run-time, i.e., outside a `TYPE_CHECKING` block / `.pyi` type stub. this was also, i believe, the reason why we kept `Union` in the pattern "top-level variable = Union[thing, thing2]"

far as I can tell all cases in this PR are fine

PR #5668, opened 2025-08-28
Prodtest certificate checks

andrewkozlik:
This PR implements the following checks for device certificate chains in prodtest:
- Ensure that the device S/N is the same as the one set in the end-entity device certificate subject name.
- Validate the MCU device certificate chain, like we already do for Optiga and Tropic.

As usual, untested :see_no_evil:.


core/embed/util/unit_properties/stm32/unit_properties.c
```diff
@@ -160,6 +160,32 @@ void unit_properties_get(unit_properties_t* props) {
...
+  // The OTP block should contain a null-terminated string when set.
+  if (sectrue !=
+          flash_otp_read(FLASH_OTP_BLOCK_DEVICE_SN, 0, block, sizeof(block)) ||
+      block[0] == 0xFF) {
+    return false;
+  }
+
+  size_t len = 0;
+  for (len = 0; len < sizeof(block); len++) {
+    if (block[len] == 0 || block[len] == 0xFF) {
```
onvej-sl: Why do you break if `block[len] == 0xFF`?

core/embed/projects/prodtest/cmd/common.c
```diff
@@ -333,14 +346,41 @@ bool check_cert_chain(cli_t* cli, const uint8_t* chain, size_t chain_size,
...
           common_name_size != sizeof(SUBJECT_COMMON_NAME) ||
           memcmp(common_name, SUBJECT_COMMON_NAME,
                  sizeof(SUBJECT_COMMON_NAME)) != 0) {
         cli_error(cli, CLI_ERROR,
                   "check_device_cert_chain, invalid common name.");
         return false;
       }
+
+      // Check that the serial number of the subject, matches the device.
+      uint8_t device_sn[32] = {0};
```
onvej-sl: Maybe `FLASH_OTP_BLOCK_SIZE` instead of `32` or define something like `DEVICE_SN_MAX_SIZE`

core/embed/util/unit_properties/stm32/unit_properties.c
andrewkozlik: The unused bytes in OTP are padded with 0xff. I was reproducing the behavior in this function:
https://github.com/trezor/trezor-firmware/blob/f01a75cfd29cd2ff9f0d7ee56905eea14ee14667/core/embed/projects/prodtest/cmd/prodtest_otp_batch.c#L52-L59
which terminates when it encounters the first 0xff byte and then searches for the first 0x00 byte to determine the length (in strlen). So my code should be equivalent. However, I think that `strnlen(block, sizeof(block))` would probably work just as well in this case, since the string should be null-terminated.

andrewkozlik: Simplified in https://github.com/trezor/trezor-firmware/pull/5668/commits/ec067061e498a1a85be97bfe0440741996cbb749

core/embed/projects/prodtest/cmd/common.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/5668/commits/ec067061e498a1a85be97bfe0440741996cbb749

core/embed/projects/prodtest/cmd/common.c
```diff
@@ -333,14 +346,41 @@ bool check_cert_chain(cli_t* cli, const uint8_t* chain, size_t chain_size,
...
+        cli_error(cli, CLI_ERROR,
+                  "check_device_cert_chain, device_sn not set.");
+      }
+
+      if (subject_sn_size != device_sn_size ||
+          memcmp(subject_sn, device_sn, device_sn_size) != 0) {
+        cli_error(cli, CLI_ERROR,
+                  "check_device_cert_chain, serial number mismatch.");
+        return false;
+      }
```
onvej-sl: This breaks compatibility with older models. Therefore, we either need to add an ifdef here or backport `device_sn` to older models.

andrewkozlik: https://github.com/trezor/trezor-firmware/pull/5668/commits/5a3d805ec6867eff6cfc1ab95eeba7a714a4a58f

PR #5670, opened 2025-08-28
Set HSM public keys

onvej-sl:
This pull request is a generalization of https://github.com/trezor/trezor-firmware/pull/5669.


core/embed/projects/prodtest/cmd/secure_channel.c
```diff
@@ -61,15 +72,24 @@ bool secure_channel_handshake_2(
     return false;
   }
 
-  if (!noise_handle_handshake_response(&noise_context, prodtest_private_key,
-                                       hsm_public_key,
-                                       (const noise_response_t*)input)) {
-    return false;
+  curve25519_key ephemeral_key_backup = {0};
+  memcpy(ephemeral_key_backup, noise_context.initiator_ephemeral_private_key,
```
onvej-sl: Maybe we should backup the whole context.

tsusanka (commented):
Just confirming the public keys are the ones I provided.

core/embed/projects/prodtest/cmd/secure_channel.c
andrewkozlik: Both solutions should work. However they violate the principle of encapsulation. The only functions that access the context variables should be the ones in noise.c. I see two possible solutions:
1. Allow `noise_handle_handshake_response()` to be called repeatedly on the same context until it succeeds. Just remove the memzero of `initiator_ephemeral_private_key` in `noise_handle_handshake_response()`, move the split chaining_key further in the function and document that the function doesn't alter the context when it fails.
2. Introduce `noise_clone_context(const noise_context_t* src, noise_context_t* dest)`.

PR #5676, opened 2025-08-29
improve reliability of nrf update

TychoVrahe:
After the final acknowledge from nRF, we still need to wait some time before resetting the chip as the ack is sent before the actual write to flash happens.


kopecdav (approved):
utACK

PR #5678, opened 2025-08-31
Use a shorter timeout when rebooting

### LINKED ISSUE #5667: Trezor do not reboot to bootloader after acknowledging request from Suite
When trying to install custom FW on **Mac** machine. Suite will promt the device to reboot-to-bootloader. After confirming with button by user, device send `Success` message to Suite which is expected to send "ACK". 

The "ACK" obviously did not make it back to device, so device tries to Retransmit the message while Suite already continues with the installation flow. After aprx 2 seconds, suite send another message over new THP channgel (in log under ID 34f3) which eventually kills the previous device channel and cause to restart the whole UI back to homescreen.

Device LOG.
Image

Problems:
1. There is a strange behavior with mac machines which need to enable "WriteWithResponse" parameter on BLE level to successfully deliver the "ACK", If we wont find a systematical solution, we should probably keep this parameter enabled on MAC. 
2. Device should probably not Retransmit repeatedly when tearing the channel down and rebooting to bootloader.

romanz:
In general, Trezor should not wait for too long for a THP ACK when sending the `Success` response (since the last ACK may be lost, similar to [Two generals' agreement and TCP handshake](https://stackoverflow.com/q/36352236)).

This PR will allow sending `Success` response 3 times (due to THP retransmissions) and proceed to reboot the device.

Fixes #5667.

### DIFF
--- a/core/src/apps/management/reboot_to_bootloader.py
+++ b/core/src/apps/management/reboot_to_bootloader.py
@@ -1,4 +1,5 @@
 import utime
+from micropython import const
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -6,6 +7,9 @@ if TYPE_CHECKING:
 
     from trezor.enums import BootCommand
     from trezor.messages import RebootToBootloader
+
+
+_REBOOT_SUCCESS_TIMEOUT_MS = const(500)
 
 
 async def install_upgrade(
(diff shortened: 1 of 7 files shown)


core/src/apps/management/reboot_to_bootloader.py
```diff
@@ -94,9 +94,12 @@ async def reboot_to_bootloader(msg: RebootToBootloader) -> NoReturn:
         boot_args = None
 
     ctx = get_context()
-    await ctx.write(Success(message="Rebooting"))
-    # make sure the outgoing USB buffer is flushed
-    await loop.wait(ctx.iface.iface_num() | io.POLL_WRITE)
+    # After ACK-ing the `Success` message (over THP), the host may already be waiting for the bootloader to start.
+    # In case this THP ACK packet is lost, the device should stop retransmissions, and reboot anyway.
+    res = await loop.race(ctx.write(Success(message="Rebooting")), loop.sleep(500))
```
M1nd3r: nit: maybe instead of 500 use properly named constant, e.g.  `REBOOT_SUCCESS_TIMEOUT_MS=const(500)`

romanz: 59deb3df9d

M1nd3r (approved):

PR #5679, opened 2025-08-31
ML-DSA device attestation

andrewkozlik:
I changed the MCU device key and certificate from Ed25519 to ML-DSA-44. I am still considering using the stronger ML-DSA-87, but there would be a little more work involved. Firstly, there were some out-of-memory errors during testing, which I don't know what they were exactly about. Also we would need to store the device certificate without the public key. Note that there is no intermediate certificate for ML-DSA due to size constraints. We would need to allocate another sector for the secrets, which I'd rather not do at this point for fear of unexpectedly breaking something.

I picked `mldsa-native` because it's the only C library that I found that exposes a key generation function in its API which lets us input the random seed. All other C libraries call a RNG. Another benefit of this library is that it is written to be constant-time for secret-dependent control flows.

There is some unnecessary code duplication, like `keccakf1600.c`, which should be replaced with our implementation, but we can deal with that later.


M1nd3r:
> Firstly, there were some out-of-memory errors during testing, which I don't know what they were exactly about.

IIRC, those were out-of-stack errors. According to documentation (and some extrapolation from benchmarks), ML-DSA-87 could need **approximately** 123 kB  stack for signing and 93 kB of stack for verifying signatures. These figures were calculated for the `pqm4` implementation - `mldsa-native` might differ.

Current size of stack for U5 devices is 32 kB
see:
[core/embed/sys/linker/stm32u5g/firmware.ld](https://github.com/trezor/trezor-firmware/blob/c855dbfb36c63a75ff9467377e025804a4ef4ad6/core/embed/sys/linker/stm32u5g/firmware.ld#L59)
[core/embed/sys/linker/stm32u58/firmware.ld](https://github.com/trezor/trezor-firmware/blob/main/core/embed/sys/linker/stm32u58/firmware.ld#L69)

@onvej-sl increased the stack size to 64 kB ib the PQ benchmarks branch for `STM32 F4` here: https://github.com/trezor/trezor-firmware/commit/545e14acbea68eb265b655dc2de4641b0ac9e29a

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -43,21 +49,26 @@
...
   for (size_t i = 0; i < length; i++) {
-    secret[i] ^= optiga_secret[i];
+    secret[i] ^= buffer[i];
   }
-  memzero(optiga_secret, sizeof(optiga_secret));
 #endif
 
 #ifdef USE_TROPIC
-  // TODO: Generate randomness using tropic and xor it with `secret`.
+  if (LT_OK != lt_random_value_get(tropic_get_handle(), buffer, length)) {
```
onvej-sl, line 64: I think this won't work in all cases because the secure channel with the tropic might not have been established yet.

andrewkozlik:
BTW, now that we are embedding the HSM secure channel public key in prodtest, it would be nice to embed the root public keys into prodtest, so that we can check the device certificates fully.

core/embed/models/T3W1/secret_layout.h
```diff
@@ -45,7 +45,7 @@
 #define SECRET_KEY_SLOT_2_PUBLIC 1
 
 #define SECRET_MCU_DEVICE_CERT_OFFSET 0x870
-#define SECRET_MCU_DEVICE_CERT_SIZE 0x400
+#define SECRET_MCU_DEVICE_CERT_SIZE 0x1000
```
onvej-sl, line 48: @TychoVrahe Does this change require a bootloader update?

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -141,15 +143,20 @@ static void prodtest_secrets_get_mcu_device_key(cli_t* cli) {
...
+  uint8_t seed[MLDSA_SEEDBYTES] = {0};
+  if (secret_key_mcu_device_auth(seed) != sectrue) {
     cli_error(cli, CLI_ERROR, "`secret_key_mcu_device_auth()` failed.");
-    return;
+    goto cleanup;
+  }
+
+  uint8_t mcu_public[CRYPTO_PUBLICKEYBYTES] = {0};
+  uint8_t mcu_private[CRYPTO_SECRETKEYBYTES] = {0};
+  if (crypto_sign_keypair_internal(mcu_public, mcu_private, seed) != 0) {
```
onvej-sl, line 177: Is this function specified in the standard, so we can be certain it won't change in the future?

andrewkozlik: The NIST ML-DSA spec specifies key generation from a 32 byte seed, which is why I am even more surprised that the reference implementation does not expose it in its API.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -160,31 +167,54 @@ static void prodtest_secrets_get_mcu_device_key(cli_t* cli) {
...
+  random_buffer(rnd, sizeof(rnd));
+
   // The challenge is intentionally constant zero.
   uint8_t challenge[CHALLENGE_SIZE] = {0};
-  ed25519_signature signature = {0};
-  ed25519_sign(challenge, sizeof(challenge), mcu_private, signature);
-  memzero(mcu_private, sizeof(mcu_private));
+  uint8_t signature[CRYPTO_BYTES] = {0};
+  size_t siglen = 0;
+  if (crypto_sign_signature_internal(signature, &siglen, challenge,
```
onvej-sl: Why do you use `crypto_sign_signature_internal` and not `crypto_sign_signature`?

andrewkozlik: I think the only reason is so that we can supply the randomness directly. Otherwise we have to define some "randombytes()" function. (I don't remember the name exactly now.)

core/embed/projects/prodtest/cmd/prodtest_secrets.c
andrewkozlik: I think all we need to do is ensure in `prodtest_secrets_init()` that the connection is established using the factory key, because that's the only place where this function is called currently and because in `prodtest_secrets_init()` only the factory key should be active. This is actually related to disabling `prodtest_secrets_init()` if Tropic is already paired.

Optionally, we can ignore the failure, so we get randomness from the Tropic chip only if there is a Tropic connection.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
andrewkozlik: Actually now that I checked the NIST specification it defines both functions:
> The key generation algorithm ML-DSA.KeyGen takes no input and outputs a public key and a private key,
which are both encoded as byte strings.
The algorithm uses an approved RBG to generate a 256-bit (32-byte) random seed 𝜉 that is given as input
to ML-DSA.KeyGen_internal (Algorithm 6), which produces the public and private keys.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -34,28 +31,44 @@
...
+#include <../vendor/mldsa-native/mldsa/sign.h>
+
 secbool generate_random_secret(uint8_t* secret, size_t length) {
   random_buffer(secret, length);
 
+  uint8_t buffer[length];
 #ifdef USE_OPTIGA
-  uint8_t optiga_secret[length];
-  if (OPTIGA_SUCCESS != optiga_get_random(optiga_secret, length)) {
+  if (OPTIGA_SUCCESS != optiga_random_buffer(buffer, length)) {
```
onvej-sl: `optiga_random_buffer()` returns `bool` instead of `optiga_result`.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
onvej-sl: I'm slightly in favor of defining `mld_randombytes()`. In the opposite case, please comment on why the prefix looks the way it does.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
onvej-sl: > I think all we need to do is ensure in prodtest_secrets_init() that the connection is established using the factory key, because that's the only place where this function is called currently and because in prodtest_secrets_init() only the factory key should be active.

I agree.

.gitmodules
```diff
@@ -41,3 +41,7 @@
 	path = vendor/sphincsplus
 	url = https://github.com/sphincs/sphincsplus/
 	branch = consistent-basew
+[submodule "vendor/mldsa-native"]
+	path = vendor/mldsa-native
+	url = https://github.com/pq-code-package/mldsa-native.git
```
onvej-sl, line 46: According to their readme, it is not recommended to use the library in a production environment yet. Are we okay with that?

core/embed/models/T3W1/secret_layout.h
TychoVrahe: it does not

.gitmodules
andrewkozlik: As long as the library passes test vectors, I think we can live with that.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
andrewkozlik: The `pre` parameter (pointer to prefix string) `(const uint8_t*)"\0\0"` represents the encoding of an empty context string and the 2 its length. If FIPS 204, Algortithm 2, line 10:
> 𝑀 ′ ← BytesToBits(IntegerToBytes(0, 1) ∥ IntegerToBytes(|𝑐𝑡𝑥|, 1) ∥ 𝑐𝑡𝑥) ∥ 𝑀

It corresponds to the part BytesToBits(IntegerToBytes(0, 1) ∥ IntegerToBytes(|𝑐𝑡𝑥|, 1) ∥ 𝑐𝑡𝑥).

andrewkozlik: Defined `ENCODED_EMPTY_CONTEXT_STRING` in https://github.com/trezor/trezor-firmware/pull/5679/commits/1add0ef2e1f378c9d1dfb6e4980af2781b69ae7a

andrewkozlik:
@cepetr @TychoVrahe I increased the stack size in prodtest to 95 kB in https://github.com/trezor/trezor-firmware/pull/5679/commits/da33cf35a4d68364382971b12628dc3e7f65c9c6, which meant decreasing the size of AUX1 RAM by 32 kB. However, the change in `core/embed/models/T3W1/memory.h` affect non-prodtest builds as well. Is it a problem?

The reason for this change is that we need about 67 kB of stack in prodtest for ML-DSA. So I need to cut off at least 4 kB somewhere, but I chose to cut off 32 kB to be extra safe in avoiding stack overflows. Originally the stack was 24 kB, so +67 kB gives 91 kB, which is more than enough, because those 67 kB already include ML-DSA + whatever was on stack before the ML-DSA functions were called. So we should be able to get away with less, if reducing AUX1 RAM is a problem.

Eventually we will need to figure out how to deal with this problem in firmware as well to implement ML-DSA device attestation. One way to solve it is to move the attestation to bootloader, where we can have a bigger stack. But let's discuss that some other time.

core/embed/models/T3W1/memory.h
```diff
@@ -94,13 +94,13 @@
 #define FB1_RAM_SIZE (768 * 1024 - 512)
 
 #define MAIN_RAM_START (0x300C0000)
-#define MAIN_RAM_SIZE (64 * 1024)
```
TychoVrahe: @cepetr can you also take a look at the stack changes here? i am not convinced that modification of `MAIN_RAM` is the way to go: it potentially affects placement of stack+variables in bootloader and boardloader too, perhaps even placing them outside of SRAM2 - losing ECC protection. Also the `memory_secmon` probably would need to be adjusted too.

Maybe simpler way to do this would be to move prodtest stack to `AUX1_RAM` in prodtest linker script?

andrewkozlik: Thank you @cepetr for pushing https://github.com/trezor/trezor-firmware/pull/5679/commits/0f5a50a655e6b61f98593dc14d56ac31ad469bcb. @TychoVrahe please review this commit.

TychoVrahe: yes this looks good

core/embed/projects/prodtest/cmd/prodtest_secrets.c
andrewkozlik: https://github.com/trezor/trezor-firmware/pull/5679/commits/bfcf1ed4b4c3766f795ad3a128e233cc20a14334

onvej-sl (approved):

PR #5680, opened 2025-09-01
Power device menu for Eckhart

bieleluk:
Temporary power menu for Eckhart with the following actions:
- restart
- reboot to bootloader
- turn off (hibernate)

This should be reverted when we are able to implement the design in [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=11722-3609&t=eLZgwJFpPqTYQnfa-0).

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
@@ -40,7 +40,8 @@ use heapless::Vec;
 //       - pin code
 //       - wipe code
 //     - device
-const MAX_SUBMENUS: usize = 7;
+//   - power
+const MAX_SUBMENUS: usize = 8;
 const MAX_DEPTH: usize = 3;
 // submenus, device screens, regulatory and about screens
 const MAX_SUBSCREENS: usize = MAX_SUBMENUS + MAX_PAIRED_DEVICES + 2;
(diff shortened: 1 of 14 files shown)


core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -358,6 +366,25 @@ impl DeviceMenuScreen {
         self.add_subscreen(Subscreen::Submenu(submenu_index))
     }
 
+    fn add_power_menu(&mut self) -> usize {
+        let mut items: Vec<MenuItem, MEDIUM_MENU_ITEMS> = Vec::new();
```
romanz, line 370: nit: `SHORT_MENU_ITEMS` (5) should fine here, no?

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -40,7 +40,8 @@ use heapless::Vec;
 //       - pin code
 //       - wipe code
 //     - device
-const MAX_SUBMENUS: usize = 7;
+//   - power
```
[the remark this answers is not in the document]
bieleluk: it's correct

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: The power screen can have a size of `SHORT_MENU_ITEMS`, but it is also a `Submenu` that must have a size od  `MEDIUM_MENU_ITEMS`:
```
struct Submenu {
    show_battery: bool,
    items: Vec<MenuItem, MEDIUM_MENU_ITEMS>,
}
```
Because the screen is only temporary, I don't see much added value in creating another struct: `ShortSubmenu`.
WDYT?

PR #5682, opened 2025-09-01
A few various fixups

romanz:

### DIFF
--- a/core/src/storage/cache_thp.py
+++ b/core/src/storage/cache_thp.py
@@ -1,5 +1,6 @@
 import builtins
 from micropython import const
+from typing import Sequence
 
 from storage.cache_common import (
     CHANNEL_HOST_STATIC_PUBKEY,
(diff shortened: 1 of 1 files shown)


core/src/storage/cache_thp.py
```diff
@@ -11,6 +12,11 @@
     DataCache,
 )
 
+if TYPE_CHECKING:
+    from typing import Sequence
+
+    pass
```
bieleluk: This is unnecessary

romanz: I think it is (https://github.com/trezor/trezor-firmware/pull/5682/commits/c13e193cfc9a0b2cb7568febf4d7568882d8dd48 comment): 
> Since `from typing import` are "commented out" by `generate_frozen_module()`,
we need to add a "trailing" `pass` - otherwise the resulting Python code will
be invalid, and frozen builds will fail.

matejcik: instead, put `from typing import Sequence` on top level, not conditional on TYPE_CHECKING

romanz: [Force-pushed a fix](https://github.com/trezor/trezor-firmware/compare/c13e193cfc9a0b2cb7568febf4d7568882d8dd48..fe8c742a3c057732f5338256d2b5885ace898c18).

Issue #5557, opened 2025-08-15
Fix THP pairing screen acording to figma


Hannsek:
More info [here](https://www.notion.so/satoshilabs/BT-THP-Testing-1e0dc5260606806aaf2fff32eb1e79b7?source=copy_link#1e0dc526060680649720d654a10949a1).

STew790:
If it is [this](https://www.notion.so/satoshilabs/BT-THP-Testing-1e0dc5260606806aaf2fff32eb1e79b7#1e0dc526060680649720d654a10949a1:~:text=Trezor%20nesed%C3%AD%20na%20Figmu%3A) issue, then it looks fixed to me because it says "Allow Trezor Suite Dev on 'my-pc-name' to pair with this Trezor"
cc @Hannsek

Hannsek:
Text in yellow color still missing.

PR #5013, opened 2025-05-06
core(rust): format translated strings

PrisionMike:
For Issue: #4338 
These changes add a new `Op` to `OpTextLayout` called `text_formatted` that accepts a template string and a vector of arguments as a function. Related function needed to be added to `StringBuffer` and `TranslatedStrings`. The later now has an `end_at` attribute for slicing.

PR #5683, opened 2025-09-01
feat(core): support basic string interpolation in Rust

romanz:
This PR allows using `'$'`-escaped digits for string interpolation (similar to [Python's `string.Template`](https://docs.python.org/3/library/string.html#template-strings) and [bash](https://tldp.org/LDP/abs/html/parameter-substitution.html)).

Such format string can be split into text and interpolation arguments, e.g. to build a `OpTextLayout` in #5557.

### DIFF
--- a/core/embed/rust/src/strutil.rs
+++ b/core/embed/rust/src/strutil.rs
@@ -308,6 +308,68 @@ impl ufmt::uDebug for TString<'_> {
     }
 }
 
+pub fn interpolate(fmt: &str) -> impl Iterator<Item = Part> {
+    Interpolate::new(fmt).filter(|part| !matches!(part, Part::Text("")))
+}
+
+pub struct Interpolate<'a> {
+    text: &'a str,
+    escape: bool,
+}
+
+impl<'a> Interpolate<'a> {
+    fn new(text: &'a str) -> Self {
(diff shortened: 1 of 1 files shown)


matejcik:
could we instead go with https://github.com/trezor/trezor-firmware/pull/5013, esp. now that we have @PrisionMike on board? :)
it would be nicer to have the same formatting syntax between Rust and Python

romanz:
I agree - will adapt this PR to support `{0}` interpolation syntax.

I prefer to use a "splitting" iterator approach, since #5557 needs to apply custom formatting on the interpolated values:
```rust
    let style = theme::firmware::TEXT_REGULAR;
    let mut ops = OpTextLayout::new(style);
    for part in description.interpolate() {
        match part {
            Part::Text(s) => {
                ops.add_text(s, style.text_font);
            }
            Part::Arg(i) => match args.get(i) {
                Some(&s) => {
                    ops.add_color(theme::YELLOW);
                    ops.add_text(s, style.text_font);
                    ops.add_color(style.text_color);
                }
                None => return Err(error::Error::OutOfRange),
            },
        };
    }
```

romanz:
Closing in favor of another PR.

PR #5685, opened 2025-09-02
feat(core): add `TString::prefix()` method 

romanz:
It will be used to implement `TString`-based interpolation in #5687.

Also, add unit-tests for `TString` slicing operation.

Rebased over #5689.

### DIFF
--- a/core/embed/rust/src/strutil.rs
+++ b/core/embed/rust/src/strutil.rs
@@ -148,7 +148,7 @@ pub enum TString<'a> {
     #[cfg(feature = "translations")]
     Translation {
         tr: TR,
-        offset: u16,
+        offset_len: Option<(u16, u16)>,
     },
     Str(&'a str),
 }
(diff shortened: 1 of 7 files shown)


core/embed/rust/src/strutil.rs
```diff
@@ -177,33 +178,69 @@ impl TString<'_> {
             #[cfg(feature = "micropython")]
             Self::Allocated(buf) => fun(buf.as_ref()),
             #[cfg(feature = "translations")]
-            Self::Translation { tr, offset } => tr.map_translated(|s| fun(&s[*offset as usize..])),
+            Self::Translation { tr, offset_len } => tr.map_translated(|s| {
+                let slice = match offset_len {
+                    Some((offset, len)) => &s[usize::from(*offset)..usize::from(*offset + *len)],
```
matejcik, line 183: this is actually wrong (and was wrong before) because in theory the translation could change between the call to `prefix()` and `map()`
presumably you should store offsets in terms of characters and calculate the right thing here?

romanz: To be sure - the problem here is that `TString::Translation { tr, offset_len }` points to the translation data, which is can be modified "under the hood", right?

In other words, this issue shouldn't happen in `TString::Allocated(StrBuffer)` and `TString::Str(&'a str)` variants - right?

matejcik: > points to the translation data

it doesn't "point to" anything is the issue. it's an enum value, and the actual string data for it is loaded inside the `map_translated` function and discarded afterwards.
you have no programmatic guarantee that TOC data will come up the same as TOU data -- the user might have changed languages in the meantime.

(in practice it's highly unlikely to happen because UI objects typically don't survive across translation change.... "typically" being the keyword)

> In other words, this issue shouldn't happen in TString::Allocated(StrBuffer) and TString::Str(&'a str) variants - right?

correct; `Str(&str)` holds immutable via lifetime guarantees, and `Allocated(StrBuffer)` is immutable per `StrBuffer`'s  own safety parameters (namely, it's either a `'static` string that can't change, or a Python `str` which is also immutable).

romanz: Let's see if translated strings are sliced in CI: https://github.com/trezor/trezor-firmware/pull/5696 -> :heavy_check_mark:

PR #5687, opened 2025-09-02
feat(core): support `StrBuffer` interpolation in Rust

romanz:
Inspired by https://github.com/trezor/trezor-firmware/pull/5013, but currently implemented only for `StrBuffer` which will be used to implement #5557.

Note: we should avoid slicing `TString` [due to its "interior mutability"](https://github.com/trezor/trezor-firmware/pull/5685#pullrequestreview-3176826356).


romanz:
Implemented slicing only for `StrBuffer` and added a unit test.

PR #5688, opened 2025-09-02
update T3W1 embedded binaries

TychoVrahe:
updated nRF and bootloader dev binaries


romanz:
Tested locally on T3W1.

PR #5691, opened 2025-09-03
Eckhart menu refresh

bieleluk:
This PR creates a new device menu feature, Refresh:

- refresh after the incoming BLE connect/disconnect event to show
- refresh to a parent sub-menu after a non-destructive  menu action

Technical details:
- new statically built device submenus to identify each screen
- new `Refresh`  device menu return type with the index of the current/closest device submenu
- new init menu screen `FwUi` parameter with the submenu that should be shown first
- new static back navigation

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
@@ -54,9 +54,9 @@ pub enum DeviceMenuId {
     Power,
 }
 
-impl TryFrom<usize> for DeviceMenuId {
+impl TryFrom<u8> for DeviceMenuId {
     type Error = ();
-    fn try_from(v: usize) -> Result<Self, Self::Error> {
+    fn try_from(v: u8) -> Result<Self, Self::Error> {
         match v {
             0 => Ok(DeviceMenuId::Root),
             1 => Ok(DeviceMenuId::PairAndConnect),
@@ -78,6 +78,13 @@ impl From<DeviceMenuId> for u8 {
     }
 }
 
+impl From<DeviceMenuId> for usize {
+    #[inline]
+    fn from(id: DeviceMenuId) -> Self {
+        usize::from(id as u8)
+    }
+}
+
 // FIXME: use mem::variant_count when it becomes stable
 const MAX_SUBMENUS: usize = 8;
 // submenus, device screens, regulatory and about screens
@@ -90,7 +97,7 @@ enum Action {
     /// Go to a registered submenu by id (static)
     GoToSubmenu(DeviceMenuId),
     /// Go to an arbitrary subscreen index (kept for device/about/regulatory)
-    GoTo(usize),
+    GoTo(u8),
     /// Return a DeviceMenuMsg to the caller
     Return(DeviceMenuMsg),
 }
@@ -104,7 +111,7 @@ pub enum DeviceMenuMsg {
     DevicePair,       // pair a new device
     DeviceDisconnect, // disconnect a device
     DeviceUnpair(
-        usize, /* which device to unpair, index in the list of devices */
+        u8, /* which device to unpair, index in the list of devices */
     ),
     DeviceUnpairAll,
 
@@ -133,6 +140,20 @@ pub enum DeviceMenuMsg {
     Close,
 }
 
+trait VecExt {
+    fn add(&mut self, item: MenuItem) -> &mut Self;
+}
+
+impl<const N: usize> VecExt for Vec<MenuItem, N> {
+    fn add(&mut self, item: MenuItem) -> &mut Self {
+        if self.push(item).is_err() {
+            #[cfg(feature = "ui_debug")]
+            fatal_error!("Menu item list is full");
+        }
+        self
+    }
+}
+
 struct MenuItem {
     text: TString<'static>,
     subtext: Option<(TString<'static>, Option<&'static TextStyle>)>,
@@ -147,7 +168,7 @@ const MENU_ITEM_WARNING: &ButtonStyleSheet = &theme::menu_item_title_orange();
 const MENU_ITEM_ERROR: &ButtonStyleSheet = &theme::menu_item_title_red();
 
 impl MenuItem {
-    pub fn new(text: TString<'static>, action: Option<Action>) -> Self {
+    fn new(text: TString<'static>, action: Option<Action>) -> Self {
         Self {
             text,
             subtext: None,
@@ -158,26 +179,46 @@ impl MenuItem {
         }
     }
 
+    pub fn go_to_submenu(title: TString<'static>, id: DeviceMenuId) -> Self {
+        Self::new(title, Some(Action::GoToSubmenu(id)))
+    }
+    pub fn go_to_subscreen(title: TString<'static>, idx: u8) -> Self {
+        Self::new(title, Some(Action::GoTo(idx)))
+    }
+    pub fn return_msg(title: TString<'static>, msg: DeviceMenuMsg) -> Self {
+        Self::new(title, Some(Action::Return(msg)))
+    }
+
     pub fn with_subtext(
-        &mut self,
+        mut self,
         subtext: Option<(TString<'static>, Option<&'static TextStyle>)>,
-    ) -> &mut Self {
+    ) -> Self {
         self.subtext = subtext;
         self
     }
 
-    pub fn with_subtext_marquee(&mut self) -> &mut Self {
+    pub fn with_subtext_marquee(mut self) -> Self {
         self.subtext_marquee = true;
         self
     }
 
-    pub fn with_connection_status(&mut self, connection_status: Option<bool>) -> &mut Self {
+    pub fn with_connection_status(mut self, connection_status: Option<bool>) -> Self {
         self.connection_status = connection_status;
         self
     }
 
-    pub fn with_stylesheet(&mut self, stylesheet: &'static ButtonStyleSheet) -> &mut Self {
-        self.stylesheet = stylesheet;
+    pub fn error(mut self) -> Self {
+        self.stylesheet = MENU_ITEM_ERROR;
+        self
+    }
+
+    pub fn warn(mut self) -> Self {
+        self.stylesheet = MENU_ITEM_WARNING;
+        self
+    }
+
+    pub fn light_warn(mut self) -> Self {
+        self.stylesheet = MENU_ITEM_LIGHT_WARNING;
         self
     }
 }
@@ -204,13 +245,13 @@ impl Submenu {
 // Each subscreen of the DeviceMenuScreen is one of these
 enum Subscreen {
     // A registered submenu
-    Submenu(usize, DeviceMenuId),
+    Submenu(u8, DeviceMenuId),
 
     // A screen allowing the user to to disconnect a device
     DeviceScreen(
         TString<'static>, /* device name */
         bool,             /* is the device connected? */
-        usize,            /* index in the list of devices */
+        u8,               /* index in the list of devices */
     ),
 
     // The about screen
@@ -244,19 +285,19 @@ pub struct DeviceMenuScreen {
     subscreens: Vec<Subscreen, MAX_SUBSCREENS>,
 
     // Sparse map from SubmenuId -> subscreen index
-    submenu_index: [Option<usize>; MAX_SUBMENUS],
+    submenu_index: [Option<u8>; MAX_SUBMENUS],
 
     // index of the current subscreen in the list of subscreens
-    active_subscreen: usize,
+    active_subscreen: u8,
 }
 
 impl DeviceMenuScreen {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
-        init_submenu: Option<usize>,
+        init_submenu: Option<u8>,
         failed_backup: bool,
         paired_devices: Vec<TString<'static>, MAX_PAIRED_DEVICES>,
-        connected_idx: Option<usize>,
+        connected_idx: Option<u8>,
         pin_code: Option<bool>,
         auto_lock_delay: Option<TString<'static>>,
         wipe_code: Option<bool>,
@@ -284,12 +325,12 @@ impl DeviceMenuScreen {
         screen.register_settings_menu();
         screen.register_power_menu();
 
-        let is_connected = connected_idx.is_some_and(|idx| idx < paired_devices.len());
+        let is_connected = connected_idx.is_some_and(|idx| usize::from(idx) < paired_devices.len());
         let connected_subtext: Option<TString<'static>> =
             is_connected.then_some(TR::words__connected.into());
 
-        let mut submenu_indices: Vec<usize, MAX_PAIRED_DEVICES> = Vec::new();
-        for (i, device) in paired_devices.iter().enumerate() {
+        let mut submenu_indices: Vec<u8, MAX_PAIRED_DEVICES> = Vec::new();
+        for (i, device) in (0u8..).zip(paired_devices.iter()) {
             let connected = connected_idx == Some(i);
             unwrap!(submenu_indices
                 .push(screen.add_subscreen(Subscreen::DeviceScreen(*device, connected, i))));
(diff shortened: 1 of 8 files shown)


core/embed/rust/librust_qstr.h
```diff
@@ -55,6 +55,7 @@ static void _librust_qstrs(void) {
   MP_QSTR_RX_PACKET_LEN;
   MP_QSTR_Reboot;
   MP_QSTR_RebootToBootloader;
+  MP_QSTR_Refresh;
```
romanz: nit: rename to `MenuRefresh` or `DeviceMenuRefresh`?

core/src/apps/homescreen/device_menu.py
```diff
@@ -24,247 +24,252 @@ def _find_device(connected_addr: bytes | None, bonds: list[bytes]) -> int | None
...
-        strings.format_autolock_duration(storage_device.get_autolock_delay_ms())
-        if config.has_pin()
-        else None
-    )
-
-    if __debug__:
-        log.debug(
-            __name__,
-            f"device menu, BLE state: {ble.connection_flags()} (peers: {ble.peer_count()})",
+    # Remain in the device loop if the until the menu is explicitly closed
```
romanz: nit: `if the until` -> `until`

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -36,14 +36,19 @@ use super::{
...
 
-// - root
-//   - pair & connect
-//   - settings
-//     - security
-//       - pin code
-//       - wipe code
-//     - device
-//   - power
+#[derive(Copy, Clone, Debug)]
```
romanz: Consider deriving `Debug` only for tests? (https://github.com/trezor/trezor-firmware/commit/4c34bb2228ae26ed5cbc6233876db8fece9c4808)

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -53,7 +58,9 @@ const DISCONNECT_DEVICE_MENU_INDEX: usize = 0;
 
 #[derive(Clone)]
 enum Action {
-    /// Go to another registered subscreen
+    /// Go to a known submenu by id (static)
```
romanz: nit: `known` -> `registered`?

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -207,6 +214,10 @@ pub struct DeviceMenuScreen {
     // Information needed to construct any subscreen on demand
     submenus: GcBox<Vec<Submenu, MAX_SUBMENUS>>,
     subscreens: Vec<Subscreen, MAX_SUBSCREENS>,
+
+    // Sparse map from SubmenuId -> subscreen index
+    submenu_index: [Option<usize>; MAX_SUBMENUS],
```
romanz: nit: consider using `u8` or `u16` (instead of `usize`)?

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -272,29 +269,42 @@ impl DeviceMenuScreen {
...
-            settings,
-            connected_subtext,
-            power,
-        );
+        screen.register_root_menu(failed_backup, pin_unset, connected_subtext);
 
+        // Activate root via the mapping
+        let root = screen
+            .try_resolve_submenu(DeviceMenuId::Root)
+            .expect("Root submenu must be registered");
```
romanz: Return an error / use `unwrap!()`.

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -701,6 +719,17 @@ impl DeviceMenuScreen {
...
+                        if let Some(menu) = self.try_resolve_submenu(id) {
+                            unwrap!(self.parent_subscreens.push(self.active_subscreen));
+                            self.set_active_subscreen(menu);
+                            self.place(self.bounds);
+                            if let ActiveScreen::Menu(screen) = self.active_screen.deref_mut() {
+                                screen.initialize_screen(ctx);
+                            }
+                        }
+                        return None;
+                    }
```
romanz, line 774: Can it be merged with the logic below?

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -800,7 +836,16 @@ impl Component for DeviceMenuScreen {
             event,
             Event::BLE(BLEEvent::Connected | BLEEvent::Disconnected | BLEEvent::ConnectionChanged)
         ) {
-            return Some(DeviceMenuMsg::Refresh);
+            let submenu_idx = match self.active_screen.deref_mut() {
+                ActiveScreen::Menu(_) => self
+                    .try_resolve_menu_id(self.active_subscreen)
+                    .unwrap_or(DeviceMenuId::Root),
+                ActiveScreen::Device(_) => DeviceMenuId::PairAndConnect,
+                ActiveScreen::Regulatory(_) | &mut ActiveScreen::About(_) => DeviceMenuId::Device,
```
romanz: why `&mut`?

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -299,6 +327,14 @@ impl DeviceMenuScreen {
         self.try_resolve_submenu(id).is_some()
     }
 
+    #[inline]
+    fn try_resolve_menu_id(&self, subscreen_idx: usize) -> Option<DeviceMenuId> {
+        self.submenu_index
+            .iter()
+            .position(|&x| x == Some(subscreen_idx))
```
romanz: indices are confusing

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -301,11 +303,13 @@ impl DeviceMenuScreen {
...
-        let root = screen
-            .try_resolve_submenu(DeviceMenuId::Root)
-            .expect("Root submenu must be registered");
-        screen.set_active_subscreen(root);
+        // Activate the init submenu
+        let init_submenu_id = init_submenu
+            .and_then(|v| DeviceMenuId::try_from(v).ok())
+            .unwrap_or_default();
+
+        let init_submenu_idx = unwrap!(screen.try_resolve_submenu(init_submenu_id));
```
romanz: Return an error (instead of panic)? or maybe just go to "root" menu?

core/src/apps/homescreen/device_menu.py
```diff
@@ -1,13 +1,27 @@
...
 from trezor import TR, config, log, utils
-from trezor.ui.layouts import interact
+from trezor.ui.layouts import interact, raise_if_cancelled
+from trezor.wire import ActionCancelled, PinCancelled
 from trezorui_api import DeviceMenuResult
 
 BLE_MAX_BONDS = 8
 
 
+class SubmenuId:
```
romanz, line 14: nit: maybe export from Rust?

core/src/apps/homescreen/device_menu.py
```diff
@@ -151,125 +188,208 @@ async def handle_device_menu() -> None:
...
+                break
         elif menu_result is DeviceMenuResult.Reboot:
             from trezor.utils import reboot_to_bootloader
 
-            # Empty boot command results to a normal reboot
-            reboot_to_bootloader()
+            try:
+                # Empty boot command results to a normal reboot
+                reboot_to_bootloader()
+            except ActionCancelled:
```
romanz: Not sure `reboot_to_bootloader` can raise `ActionCancelled`...
(same for `hibernate`)

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -792,16 +786,29 @@ impl DeviceMenuScreen {
...
 
     fn go_back(&mut self, ctx: &mut EventCtx) -> Option<DeviceMenuMsg> {
-        if let Some(parent) = self.parent_subscreens.pop() {
-            self.set_active_subscreen(parent);
-            self.place(self.bounds);
-            if let ActiveScreen::Menu(screen) = self.active_screen.deref_mut() {
-                screen.initialize_screen(ctx);
+        let parent = match self.subscreens[self.active_subscreen] {
+            Subscreen::Submenu(..) => {
+                match unwrap!(self.try_resolve_menu_id(self.active_subscreen)) {
```
romanz: Instead of `unwrap!()`, go to `DeviceMenuId::Root`?

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -268,29 +295,52 @@ impl DeviceMenuScreen {
...
+        self.submenu_index[id as usize]
+    }
+
+    #[inline]
+    fn has_submenu(&self, id: DeviceMenuId) -> bool {
+        self.try_resolve_submenu(id).is_some()
+    }
+
+    #[inline]
+    fn try_resolve_menu_id(&self, subscreen_idx: usize) -> Option<DeviceMenuId> {
```
romanz: Maybe store the `Option<DeviceMenuId>` as part of the `Subscreen::Submenu` variant?
(in order to avoid the loop here...)

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: The `expect` is no longer used there.

core/embed/rust/librust_qstr.h
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/f747912d3c29ced1fd32e26a8886322998e44d33

core/src/apps/homescreen/device_menu.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/c3036967e16502c053e4a80d685e2767b06c3ec8

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/fc81f7dd0681b98fe9d67b796f087db2686d33ad

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/9ee630339762bf3bd0039a20f9a9d60f02fd8a45

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: There are even more similar spots:
https://github.com/trezor/trezor-firmware/pull/5691/commits/ab53a670ffd9505648edbcf51e6c88313210cacc

martykan:
I have a small request, can we add `ble.disconnect()` before `reboot_to_bootloader()`?
Because when it remains connected, Suite doesn't know that the state of the device changed.

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/5aef67507cdc3f318b08f80c0890d40516761392

core/src/apps/homescreen/device_menu.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/3758ea2e85b556a0d0a72df2369bfc801a5e8b94

core/src/apps/homescreen/device_menu.py
bieleluk: @romanz, is there any reasonable way to do it? We could do something similar to `DeviceMenuResult`, but because we need it as an enum in Rust, it would be more complicated to do the full mapping than having two copies (in Rust and micropython).

core/src/apps/homescreen/device_menu.py
```diff
@@ -24,247 +38,323 @@ def _find_device(connected_addr: bytes | None, bonds: list[bytes]) -> int | None
...
+        elif menu_result is DeviceMenuResult.TurnOff:
+            from trezor import io
+
+            io.pm.hibernate()
+            break
+        elif menu_result is DeviceMenuResult.Reboot:
+            from trezor.utils import reboot_to_bootloader
+
+            # Empty boot command results to a normal reboot
+            reboot_to_bootloader()
```
romanz, line 361: Will calling it on the emulator succeed/fail (here and below)?
Asking since we have a bit different behaviour here:
https://github.com/trezor/trezor-firmware/blob/49f4537355813ef3343426177f5299a2137686c2/core/src/apps/management/reboot_to_bootloader.py#L113

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/5f9e293b5ae092ba995b2e1d08e724d40b73a5cf

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/5f9e293b5ae092ba995b2e1d08e724d40b73a5cf

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: With the current implementation, it's not possible to go to the "root" menu without using `unwrap`

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -318,12 +318,12 @@ impl DeviceMenuScreen {
         screen.register_settings_menu();
         screen.register_power_menu();
 
-        let is_connected = connected_idx.is_some_and(|idx| idx < paired_devices.len());
+        let is_connected = connected_idx.is_some_and(|idx| (idx as usize) < paired_devices.len());
```
romanz: nit: `(idx as usize)` => `usize::from(idx)` here and below.

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/3de8517c413e7c72633cf6816ed5914e279bfe79
It's an all-or-nothing approach, because mixing `u8` and `usize` was a nightmare.
Which one do you prefer?

core/src/apps/homescreen/device_menu.py
bieleluk: I replicated the behavior for non-returnable power actions:
https://github.com/trezor/trezor-firmware/pull/5691/commits/827f958b34aa5e7e0edb870a63cd391d691c4672

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5691/commits/827f958b34aa5e7e0edb870a63cd391d691c4672

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
romanz: LGTM, thanks!

bieleluk:
> I have a small request, can we add `ble.disconnect()` before `reboot_to_bootloader()`? Because when it remains connected, Suite doesn't know that the state of the device changed.

@martykan 
If we implemented `ble.disconnect()` before `reboot_to_bootloader()` only in the device menu, we would get inconsistent behavior between the reboot triggered from the device and from the host. If we want to implement this functionality, it should go directly to the `reboot_to_bootloader` function, which is out of the scope of this PR. I  suggest creating a separate issue for that.
CC: @TychoVrahe @romanz

romanz:
> Because when it remains connected, Suite doesn't know that the state of the device changed.

https://satoshilabs.slack.com/archives/CL1D61PQF/p1757600544409979 is probably relevant.

PR #5692, opened 2025-09-03
Charging in suspend rework

kopecdav:
This PR reworks the way we charge the device in suspend mode. The reason for this rework is that we have been requested to use the RGB LED during charging in suspend mode, and we also want to continuously monitor the battery SOC with the fuel gauge.

### Previous Implementation
When the device was suspended during charging, it immediately entered `STOP2` mode, but the RTC timer was set to wake up the device at short intervals (60s) and compensate the soc with a single reading of current, voltage, and ntc_temp in a background task. This approach was essentially close to coulomb counting and was quite imprecise.

### Current Implementation
Now, if the device is requested to go into suspend while charging, it will suspend most drivers but keep the power manager active (including npm1300 and stwlc38), as well as the RGB LED. The device remains in a loop where it signals charging with an RGB LED effect until:

1. Wakeup flag arrives and the device will leave suspend as usual.
2. The device becomes fully charged, it will stop charging, suspend the remaining drivers and enter STOP2 mode.

#### Other things to mention:

- The RGB LED was extended with wakeup_params. If the device goes into suspend while any RGB LED effect was started for infinite cycles, the effect is restarted again on resume. This is mainly used in the bootloader, where the device can suspend during BLE pairing and the RGB LED is breathing in blue.
- If the device was suspended while there was an ongoing effect started for a finite number of cycles, the effect is dropped and will not resume after wakeup.
- PM_AUTO_HIBERNATE_TIMEOUT_S is measured from the time the device enters STOP2 mode, so if the device is suspended while charging, the countdown starts after the device is fully charged.


core/embed/io/rgb_led/inc/io/rgb_led.h
```diff
@@ -47,6 +47,14 @@ typedef enum {
   RGB_LED_NUM_OF_EFFECTS,
 } rgb_led_effect_type_t;
 
+/**
+ * @brief RGB LED wakeup parameters
```
TychoVrahe, line 53: this struct, functions rgb_led_set_wakeup_params, suspend and resume can and behind ifdef KERNEL_MODE

core/embed/sys/suspend/stm32u5/suspend.c
```diff
@@ -64,12 +68,40 @@ wakeup_flags_t system_suspend(void) {
...
+  // If the device is requested to go in suspend, but the battery is charging,
+  // Keep in this loop until the the external power got diconnected or the
+  // device is woke up. This state is signaled with RGB LED charging effect
+  bool charging_in_suspend;
+  do {
+    if (pm_is_charging()) {
+      charging_in_suspend = true;
+
+#ifdef USE_RGB_LED
+      if (!rgb_led_effect_ongoing()) {
```
TychoVrahe, line 83: we call set_color in suspend_drivers, which ends any ongoing effect, so this should always be false an can be omitted?

core/embed/sys/power_manager/stm32u5/power_manager.c
```diff
@@ -628,6 +616,26 @@ bool pm_schedule_rtc_wakeup(void) {
   return true;
 }
 
+bool pm_is_charging(void) {
```
TychoVrahe, line 619: should we also use this in get_state function?

core/embed/sys/power_manager/stm32u5/power_monitoring.c
```diff
@@ -86,27 +86,19 @@ void pm_pmic_data_ready(void* context, pmic_report_t* report) {
...
-        // remains active. Since the device is very likely to stay in suspend
-        // mode for limited time, for now we decided to neglect this. but in
-        // the future we may want to distinguish between suspend mode
-        // with/without BLE and use different self-discharge rates.
-      }
+      // Use known battery self-discharge rate to compensate the fuel gauge
+      // estimation during the suspend period. Since this period may be very
+      // long and the battery temperature may vary, use the average ambient
+      // temperature.
+      pm_compensate_fuel_gauge(&drv->fuel_gauge.soc, drv->time_in_suspend_s,
```
TychoVrahe, line 93: we can now be suspended even when on power source even when on power source (i.e . after being fully charged), so maybe we should look at the power source state here

core/embed/io/rgb_led/inc/io/rgb_led.h
kopecdav: Right, that make sense: https://github.com/trezor/trezor-firmware/pull/5692/commits/e8d1d216e749799f47fbf3c407d839ab508cc8d5

core/embed/sys/suspend/stm32u5/suspend.c
kopecdav: Not really, you can overwrite an ongoing effect by calling `rgb_led_effect_start` function, so this is rather to latch the charging effect and not to keep restarting it over again. I could introduce the `effect_started` variable but this was bit simpler.

core/embed/sys/power_manager/stm32u5/power_manager.c
kopecdav: Right, I was about to make a new PR for this but its a simple change so lets put it here. 
https://github.com/trezor/trezor-firmware/pull/5692/commits/0a61167052f91daebb8bb1296888a7fa338a9f40

core/embed/sys/power_manager/stm32u5/power_monitoring.c
kopecdav: This is unfortunately bit more complicated. You cant easily recognize that device was connected to the external power during its time in suspend, since you can wake it up by attaching an external power. 

the same applies for drv->fully_charged flag, where you could recognize that device was suspended right after the device get fully_charged, but if somebody plug out the external power when the device get suspended, you would not compensate even if you should, or you dont know how much.

You may be suspended only for max ~ 2.hours, so compensation should be something like 64uAh. In case of fully charged device, this will probably have no impact. I would propose to keep it as it is.

TychoVrahe: ok agree. 

about the suspended time limit though. currently the timer is started after we leave the suspend-charging, but maybe we should adjust the time by the time spent in suspend-charging state? otherwise it prolongs the suspended state from application PoV

kopecdav: well .. `drv->suspend_timestamp` is captured in `pm_suspend()` and `time_to_hibernate` in `pm_driver_suspend()` is calculated based on it.   

soo .. obviously its already there. I just missed that when I wrote the PR. 

Only side effect is, that device do not AUTOHIBERNATE until is not fully charged. If that's fine, we should be good to go.

TychoVrahe: i think its acceptable for now at least

TychoVrahe (approved):

core/embed/sys/suspend/stm32u5/suspend.c
```diff
@@ -64,12 +68,40 @@ wakeup_flags_t system_suspend(void) {
...
 
+  wakeup_flags_t wakeup_flags = 0;
+  wakeup_flags_get(&wakeup_flags);
+
+  // If the device is requested to go in suspend, but the battery is charging,
+  // Keep in this loop until the the external power got diconnected or the
+  // device is woke up. This state is signaled with RGB LED charging effect
+  bool charging_in_suspend;
+  do {
+    if (pm_is_charging()) {
```
cepetr, line 79: Although not strictly necessary, it would be better to add `__WFI()` here. For example:

```c
do {
  if (pm_is_charging()) { 
  ...
  } else {
  ...
  }
  
  __WFI();
  
  wakeup_flags_get(&wakeup_flags);
  
  ...
    
    
  
```

core/embed/sys/suspend/stm32u5/suspend.c
```diff
@@ -64,12 +68,40 @@ wakeup_flags_t system_suspend(void) {
...
+      charging_in_suspend = true;
+
+#ifdef USE_RGB_LED
+      if (!rgb_led_effect_ongoing()) {
+        rgb_led_effect_start(RGB_LED_EFFECT_CHARGING, 0);
```
cepetr, line 84: Is it really necessary to enable the charging LED effect here? Shouldn't it already be enabled if `pm_is_charging()` is true?

I agree we need a two-step suspend, but to keep things simple here, it would be better to just wait (and not start the effects) if possible.

cepetr (changes requested):

core/embed/sys/suspend/stm32u5/suspend.c
cepetr: As we discussed in person, this would require quite a lot of refactoring, so we will do it together with the future LED controller/manager that will consolidate system events/states and control the RGB LED.

core/embed/sys/suspend/stm32u5/suspend.c
kopecdav: fixed here: https://github.com/trezor/trezor-firmware/pull/5692/commits/ab4b9dd1f60ea2759c1acc98c0ff2a7cca8a1d5e

cepetr (approved):
utACK

PR #5693, opened 2025-09-03
Migrate from Poetry to uv

matejcik:


matejcik:
force-pushed to add `[no changelog]` annotations
a fixup removes a leftover in pyproject.toml

matejcik:
this breaks `build-docker.sh` with an error that really doesn't make any sense

matejcik:
```
solved 97 packages in 576ms
  x Failed to build `trezor @ file:///reproducible-build/trezor-firmware/python`
  |-> The build backend returned an error
  `-> Call to `uv_build.build_editable` failed (exit status: 1)

      [stderr]
        File "<string>", line 11, in <module>
          wheel_filename = backend.build_editable("/root/.cache/uv/builds-v0/.tmplK2siH", {}, None)
        File "/root/.cache/uv/builds-v0/.tmpG3MfGy/lib/python3.13/site-packages/uv_build/__init__.py", line 125, in build_editable
          return call(args, config_settings)
        File "/root/.cache/uv/builds-v0/.tmpG3MfGy/lib/python3.13/site-packages/uv_build/__init__.py", line 54, in call
          result = subprocess.run(
              [uv_bin, *build_backend_args, *args], stdout=subprocess.PIPE
          )
        File "/nix/store/sd81bvmch7njdpwx3lkjslixcbj5mivz-python3-3.13.4/lib/python3.13/subprocess.py", line 554, in run
          with Popen(*popenargs, **kwargs) as process:
               ~~~~~^^^^^^^^^^^^^^^^^^^^^^
        File "/nix/store/sd81bvmch7njdpwx3lkjslixcbj5mivz-python3-3.13.4/lib/python3.13/subprocess.py", line 1039, in __init__
          self._execute_child(args, executable, preexec_fn, close_fds,
          ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                              pass_fds, cwd, env,
                              ^^^^^^^^^^^^^^^^^^^
          ...<5 lines>...
                              gid, gids, uid, umask,
                              ^^^^^^^^^^^^^^^^^^^^^^
                              start_new_session, process_group)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/nix/store/sd81bvmch7njdpwx3lkjslixcbj5mivz-python3-3.13.4/lib/python3.13/subprocess.py", line 1857, in _execute_child
          self._posix_spawn(args, executable, env, restore_signals, close_fds,
          ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                            p2cread, p2cwrite,
                            ^^^^^^^^^^^^^^^^^^
                            c2pread, c2pwrite,
                            ^^^^^^^^^^^^^^^^^^
                            errread, errwrite)
                            ^^^^^^^^^^^^^^^^^^
        File "/nix/store/sd81bvmch7njdpwx3lkjslixcbj5mivz-python3-3.13.4/lib/python3.13/subprocess.py", line 1801, in _posix_spawn
          self.pid = os.posix_spawn(executable, args, env, **kwargs)
                     ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      FileNotFoundError: [Errno 2] No such file or directory: '/root/.cache/uv/builds-v0/.tmpG3MfGy/bin/uv-build'

      hint: This usually indicates a problem with the package or the build environment.
```

matejcik:
resolved the build-docker problem by switching to hatchling build backend

.github/workflows/core.yml
```diff
@@ -470,7 +470,7 @@ jobs:
       - run: chmod +x core/build/unix/trezor-emu-core*
       - uses: ./.github/actions/environment  # XXX poetry maybe not needed
       - run: nix-shell --run "git clone --depth=1 https://github.com/bitcoin-core/HWI.git"
-      - run: nix-shell --arg fullDeps true --run "cd HWI && poetry install && poetry run ./test/test_trezor.py --model_t ../core/build/unix/trezor-emu-core bitcoind"
+      - run: nix-shell --arg fullDeps true --run "cd HWI && poetry install && uv run ./test/test_trezor.py --model_t ../core/build/unix/trezor-emu-core bitcoind"
```
romanz: Do we need both `poetry` and `uv` here (and in `legacy.yml`)?

romanz (approved):
Wow :open_mouth:   

```
$ time uv sync
Using CPython 3.13.4 interpreter at: /nix/store/sd81bvmch7njdpwx3lkjslixcbj5mivz-python3-3.13.4/bin/python3
Creating virtual environment at: .venv
Resolved 125 packages in 1ms
Installed 116 packages in 22ms
...
real    0m0.050s
user    0m0.044s
sys     0m0.108s

```

.github/workflows/core.yml
mmilata: Right, `uv run` fails on my system with `error: No "project" table found in: /home/mmilata/work/HWI/pyproject.toml`. Doesn't really matter since the HWI jobs are disabled.

shell.nix
```diff
@@ -80,14 +80,6 @@ stdenvNoCC.mkDerivation ({
   name = "trezor-firmware-env";
   buildInputs = lib.optionals fullDeps [
```
mmilata, line 81: Nit: we can now possibly move poetry under fullDeps as it's only used by `core_unit_python_test`, also the shell plugin is no longer needed.

.github/workflows/core.yml
matejcik: yeah good catch, HWI is using poetry so that's what we stick with

shell.nix
matejcik: poetry also runs `ts-tvl` which is used in T3W1 tests, no?

matejcik: ah, it's not used at this time.
anyway i'd keep poetry around, it's not a heavy dependency and it might come in useful. "fullDeps" is currently literally only bitcoind so :woman_shrugging:

.github/workflows/core.yml
matejcik: ad983e290776c5e8cf369b47464be611772ef750

shell.nix
matejcik: 52c76cba4a5655061d428240ab6ad2468f63c913 for shell plugin

matejcik:
no idea wtf is up with style-check failing. it's decidedly not failing on the followup PR https://github.com/trezor/trezor-firmware/pull/5707 which has the same commits

shell.nix
mmilata: > "fullDeps" is currently literally only bitcoind

Also the monero tests binary. Otherwise we should probably drop fullDeps since we no longer build bitcoind from source (and HWI tests are disabled).

mmilata (approved):

PR #5694, opened 2025-09-03
adjust advertisment intervals

TychoVrahe:

### DIFF
--- a/nordic/trezor/trezor-ble/src/ble/advertising.c
+++ b/nordic/trezor/trezor-ble/src/ble/advertising.c
@@ -37,9 +37,25 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 #define ADV_FLAG_DEV_CONNECTED 0x04
 #define ADV_FLAG_USER_DISCONNECT 0x08
 
+#define ADV_INTERVAL_FAST_MIN_MS 20
+#define ADV_INTERVAL_FAST_MAX_MS 25
+#define ADV_INTERVAL_SLOW_MIN_MS 152.5
+#define ADV_INTERVAL_SLOW_MAX_MS 211.25
+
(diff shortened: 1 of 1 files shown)


nordic/trezor/trezor-ble/src/ble/advertising.c
```diff
@@ -37,9 +37,16 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 #define ADV_FLAG_DEV_CONNECTED 0x04
 #define ADV_FLAG_USER_DISCONNECT 0x08
 
+#define ADV_INTERVAL_FAST ((int)(20 / 0.625))
```
kopecdav: This calculation is quite unclear, can you add a comment or units?

kopecdav (changes requested):

nordic/trezor/trezor-ble/src/ble/advertising.c
TychoVrahe: fixed in https://github.com/trezor/trezor-firmware/commit/8a9f7a8f876f34c06d726db229c7d41e7d62ea34

I also separated interval min and max as these should not be the same as i found in this comment:

	/** Maximum Advertising Interval (N * 0.625 milliseconds)
	 * Minimum Advertising Interval shall be less than or equal to the
	 * Maximum Advertising Interval. The Minimum Advertising Interval and
	 * Maximum Advertising Interval should not be the same value (as stated
	 * in Bluetooth Core Spec 5.2, section 7.8.5)
	 * Range: 0x0020 to 0x4000
	 */

kopecdav (approved):

nordic/trezor/trezor-ble/src/ble/advertising.c
kopecdav: Make sense, thanks!

PR #5696, opened 2025-09-03
test(core): avoid slicing `TString::Translation` variant

romanz:
Following https://github.com/trezor/trezor-firmware/pull/5685#discussion_r2316266438.

[no changelog]

### DIFF
--- a/core/embed/rust/src/strutil.rs
+++ b/core/embed/rust/src/strutil.rs
@@ -192,10 +192,15 @@ impl TString<'_> {
             #[cfg(feature = "micropython")]
             Self::Allocated(s) => Self::Allocated(s.skip_prefix(skip_bytes)),
             #[cfg(feature = "translations")]
-            Self::Translation { tr, offset } => Self::Translation {
-                tr: *tr,
-                offset: offset + skip_bytes as u16,
-            },
+            Self::Translation { tr, offset } => {
+                // Avoid slicing translated strings, since they are loaded from flash and may
+                // change later. See https://github.com/trezor/trezor-firmware/pull/5685#discussion_r2316266438 for more details.
+                debug_assert_eq!(skip_bytes, 0);
+                Self::Translation {
+                    tr: *tr,
+                    offset: offset + skip_bytes as u16,
+                }
+            }
             Self::Str(s) => Self::Str(&s[skip_bytes..]),
         }
     }


romanz:
The good news are that it seems that we don't slice translated strings (at least not in CI).

However, it would be better to disallow such slicing in compile-time :thinking:

matejcik:
why don't you want to implement slicing in terms of char offsets instead?

romanz:
> why don't you want to implement slicing in terms of char offsets instead?

It should help to avoid non-char boundary panic, but if the translated string becomes shorter (or empty) after the slice has been created, accessing the slice will also panic, right?

romanz:
In general, I think it should be fine to use `TString` slice as long as the original translation data hasn't changed.

The question is what should we do in case it does change - IIUC, current callers of `skip_prefix()` don't expect it to fail, so we can't just return an error...

matejcik:
> It should help to avoid non-char boundary panic, but if the translated string becomes shorter (or empty) after the slice has been created, accessing the slice will also panic, right?

the idea was that if you run out of bounds with the given char offset, you'll return an empty string.
the logic is, we ~never want the `TString::map` operation to fail, given a successful slice operation; and we never want a slice operation to fail given a good offset "at that time". so programmers can safely slice any TString without thinking about it, and get back _something_. ideally something close enough to what they meant
(very stupid example, if language magically changes between one page and the next, the old string was sliced at 5 chars, so those ~same 5 chars "remain" on the previous page.)

so we'd calculate char offset from byte offset at slice time, then apply the char offset with Python's bounding logic, that is you can slice outside the bounds of the string, and if you're _completely_ outside then you'll get an empty slice

an alternative but also valid approach is to say that if the translation changed, the old offsets are meaningless anyway.
then instead of `[]` slicing, use a fallible slice (`.get()` ?) and `.unwrap_or("")`

romanz:
> why don't you want to implement slicing in terms of char offsets instead?

P.S.
I think that it would be less efficient ([since `Chars` doesn't support random access](https://doc.rust-lang.org/std/primitive.str.html#method.chars)).

matejcik:
> I think that it would be less efficient

true but we don't really care on translations because they're generally very short

matejcik:
you could also save byte offsets and then at `map()` time just correct them to the nearest char boundary :woman_shrugging:

PR #5697, opened 2025-09-03
Trusted anchors for device certificates in prodtest

andrewkozlik:
This PR adds device authenticity root public keys to prodtest so that we can fully check the device certificate chain. The main reason for this check is to ensure integrity of the certificate chain.

I also changed one behavior which was inconsistent between Optiga and Tropic, vs. MCU certificate storing. We should first check the device certificate chain before storing it. In Optiga and Tropic we first stored and then checked. There it didn't matter that much, because we can overwrite the certificate, but with the MCU flash storage it's more complicated. Nevertheless to make development and testing easier I propose that in non-production mode we write the certificate even if the root key is unknown and just emit a warning in that case. In production it shouldn't happen, and if it does, then it probably means that the authority key ID is corrupted, so we should be strict in production.


core/embed/projects/prodtest/cmd/prodtest_tropic.c
```diff
@@ -1504,6 +1504,12 @@ static void cert_write(cli_t* cli, uint16_t first_slot, uint16_t slots_count) {
     goto cleanup;
   }
 
+  if (first_slot == TROPIC_DEVICE_CERT_FIRST_SLOT &&
+      !check_device_cert_chain(cli, certificate, certificate_length)) {
+    // Error returned by check_device_cert_chain().
+    return;
```
onvej-sl: Not introduced in this pull request:

```suggestion
    goto cleanup;
```

andrewkozlik: Fixed and force-pushed.

PR #5701, opened 2025-09-04
fix(core): correct BLE MAC address display

romanz:
Otherwise, the device displays BLE MAC address bytes in reversed order.


romanz:
Tested locally with `hcitool dev` - the reported MAC address is displayed correctly after this fix.

obrusvit (approved):
utACK

PR #5704, opened 2025-09-04
Erase BLE bonds when wiping the device

romanz:
[Slack](https://satoshilabs.slack.com/archives/C082Q0Y6WP3/p1756967548060319)

Also, don't fail when cancelling recovery via UI:
```
12.414 trezor.loop ERROR exception:
  File "trezor/loop.py", line 162, in _step
  File "apps/management/recovery_device/homescreen.py", line 33, in recovery_homescreen
  File "apps/management/recovery_device/homescreen.py", line 65, in recovery_process
  File "trezor/wire/context.py", line 154, in try_get_ctx_ids
  File "trezor/wire/context.py", line 114, in get_context
NoWireContext: 
```

### DIFF
--- a/core/src/apps/management/wipe_device.py
+++ b/core/src/apps/management/wipe_device.py
@@ -1,5 +1,6 @@
 from typing import TYPE_CHECKING
 
+from trezor import utils
 from trezor.wire.context import get_context, try_get_ctx_ids
 
 if TYPE_CHECKING:
@@ -58,5 +59,12 @@ async def wipe_device(msg: WipeDevice) -> NoReturn:
 
     # reload settings
     reload_settings_from_storage()
+
+    if utils.USE_BLE:
+        from trezorble import erase_bonds
+
+        # raise an exception if bonds erasing fails
+        erase_bonds()
+
     if __debug__:
         log.debug(__name__, "Device wipe - finished")
--- a/core/src/trezor/wire/context.py
+++ b/core/src/trezor/wire/context.py
@@ -218,6 +218,6 @@ def cache_delete(key: int) -> None:
 def _get_cache_for_key(key: int) -> DataCache:
     if key & SESSIONLESS_FLAG:
         return cache.get_sessionless_cache()
-    if CURRENT_CONTEXT:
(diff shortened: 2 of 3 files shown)


core/src/apps/management/wipe_device.py
```diff
@@ -55,6 +56,10 @@ async def wipe_device(msg: WipeDevice) -> NoReturn:
             log.debug(__name__, "Failed to send Success message after wipe.")
         pass
     storage.wipe_cache()
+    if utils.USE_BLE:
+        from trezorble import erase_bonds
+
+        erase_bonds()
```
M1nd3r, line 67: Could the `erase_bonds` fail?

If it can, I guess we should handle it somehow - at least log an error.

Other question is, if it is blocking or not - should we proceed with the rest of the wipe (`reload_settings_from_storage`), similarly to the situation when sending `Success` message fails - or should we inform the user that the `wipe` was not successful?

M1nd3r (changes requested):

M1nd3r (approved):

core/src/apps/management/wipe_device.py
romanz: > Could the erase_bonds fail?

`erase_bonds` can fail, but since it drops the BLE connections, we can do it only after the last write operation (in case it uses BLE).

> If it can, I guess we should handle it somehow - at least log an error.

Sounds good - logged the error in b3aae2a2c5.

> should we proceed with the rest of the wipe (reload_settings_from_storage), similarly to the situation when sending Success message fails - or should we inform the user that the wipe was not successful?

For now, we don't have an error layout for notifying the user about such issues, but I've moved `erase_bonds()` to the end of the function (b3aae2a2c5) - so in case it fails, the error will not be ignored. 
If it happens, the user will be able to see that the bonds were not deleted (via device menu).

M1nd3r (approved):
LGTM - the solution where the `erase_bonds` is the last operation is imo the best option.

romanz:
Tested wiping from UI and via `trezorctl`. In both cases, BLE bonds are deleted.

PR #5707, opened 2025-09-04
Improve nRF reproducible build process

matejcik:
builds on top of #5693 but if necessary, we can do it separately (it's just gonna be a little bit of a pain to port over the dependency changes)

* introduce the ability to construct and parse mcuboot headers
* introduce `nrftool` to replace `imgtool` and three separate scripts that worked with `imgtool` outputs
* modify `build_sign_flash.sh` to:
  * respect `$PRODUCTION` envvar
  * use `nrftool` commands to build and sign headers, correct values are taken from configs in our repository
  * always build & sign together, because one without the other makes very little sense
* simplify manual build steps
* simplify reproducible build steps


matejcik:
force-pushed to solve conflicts and make uv-related commits go away from the diff

matejcik:
@SocketSecurity ignore pypi/construct-typing@0.6.2

obrusvit (commented):
bin merging is prolly wrong. See [slack discussion](https://satoshilabs.slack.com/archives/C064H6J3S0H/p1760348601972709)

mmilata:
Anyone remembers what's up here? Slack link's dead for me ...

PR #5710, opened 2025-09-04
fix device not accepting BLE messages after wakeup

TychoVrahe:
device did not start accepting messages in case it was suspended without BLE connection. This caused busy errors in Suite


kopecdav (approved):
utACK

PR #5713, opened 2025-09-04
Add type annotation to `loop.py`

M1nd3r:
Added type annotation to `race`, `sleep`, and `spawn` functions in `loop.py`.

Type annotation for `wait` and `mailbox` would be nice too, but it is harder to do because the type cannot be (easily) inferred from the constructor. It either requires implementing "true" Generic, or passing the types to the constructor directly. There might be other solutions. However, I have not found anything nice-enough/user-friendly that would be without impact on PROD, yet.

EDIT:
Both `mailbox` and `wait` are annotated, too. However, `wait` annotation does not work properly yet as it would require additional changes mentioned here: https://github.com/trezor/trezor-firmware/pull/5713#issuecomment-3257947079

### DIFF
--- a/core/src/trezor/loop.py
+++ b/core/src/trezor/loop.py
@@ -14,13 +14,18 @@ from typing import TYPE_CHECKING
 from trezor import io, log
 
 if TYPE_CHECKING:
-    from typing import Any, Awaitable, Callable, Coroutine, Generator
+    from typing import Any, Awaitable, Callable, Coroutine, Generator, Generic, TypeVar
 
     from .loop import wait
 
-    Task = Coroutine | Generator | wait
-    AwaitableTask = Task | Awaitable
+    T = TypeVar("T")
+    Task = Coroutine[Any, Any, T] | Generator[Any, Any, T] | wait
+    AwaitableTask = Task[T] | Awaitable[T]
     Finalizer = Callable[[Task, Any], None]
+else:
+    # typechecker cheat: Generic[T] will be `object` which is a valid parent type
+    Generic = [object]
+    T = 0
 
 # tasks scheduled for execution in the future
 _queue = utimeq.utimeq(64)
@@ -122,7 +127,7 @@ def run() -> None:
             # timeout occurred, run the first scheduled task
             if _queue:
                 _queue.pop(task_entry)
-                _step(task_entry[1], task_entry[2])  # type: ignore [Argument of type "int" cannot be assigned to parameter "task" of type "Task" in function "_step"]
+                _step(task_entry[1], task_entry[2])  # type: ignore [Argument of type "int" cannot be assigned to parameter "task" of type "Task[Unknown]" in function "_step"]
                 # error: Argument 1 to "_step" has incompatible type "int"; expected "Coroutine[Any, Any, Any]"
                 # rationale: We use untyped lists here, because that is what the C API supports.
 
@@ -171,7 +176,7 @@ def _step(task: Task, value: Any) -> None:
             log.exception(__name__, e)
         finalize(task, e)
     else:
-        if isinstance(result, Syscall):
+        if isinstance(result, _RealSyscall):
             result.handle(task)
         elif result is None:
             schedule(task)
@@ -180,7 +185,7 @@ def _step(task: Task, value: Any) -> None:
                 log.error(__name__, "unknown syscall: %s", result)
 
 
-class Syscall:
+class _RealSyscall(Generic[T]):
     """
     When tasks want to perform any I/O, or do any sort of communication with the
     scheduler, they do so through instances of a class derived from `Syscall`.
@@ -192,18 +197,26 @@ class Syscall:
 
     if TYPE_CHECKING:
 
-        def __await__(self) -> Generator[Any, Any, Any]:
+        def __await__(self) -> Generator[Any, Any, T]:
             return self.__iter__()
 
     def handle(self, task: Task) -> None:
         pass
+
+    ...
+
+
+if TYPE_CHECKING:
+    Syscall = _RealSyscall
+else:
+    Syscall = {T: _RealSyscall, int: _RealSyscall}
 
 
 class Timeout(Exception):
     pass
 
 
-class sleep(Syscall):
+class sleep(Syscall[int]):
     """Pause current task and resume it after given delay.
 
     Result value is the calculated deadline.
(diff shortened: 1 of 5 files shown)


matejcik:
any reason not to make `Syscall` itself generic?

iirc you can use the following trick to do runtime correctly:
```
class _RealSyscall(Generic[T]):
    ...

if TYPE_CHECKING:
    Syscall = _RealSyscall
else:
    Syscall = [_RealSyscall]

class wait(Syscall[T]):
    ...
```

M1nd3r:
@matejcik 
> any reason not to make `Syscall` itself generic?
> 
> iirc you can use the following trick to do runtime correctly:
> 
> ```
> class _RealSyscall(Generic[T]):
>     ...
> 
> if TYPE_CHECKING:
>     Syscall = _RealSyscall
> else:
>     Syscall = [_RealSyscall]
> 
> class wait(Syscall[T]):
>     ...
> ```

I have tried this (and forgot to mention it in rationale).

Tl;dr: Generic `Syscall` brings other problems and not that many solutions. The current version is IMO better.

1) If I make `Syscall` itself generic (as described above), I get the following error in the child classes: 
```
************* Module trezor.loop
core/src/trezor/loop.py:237:4: W0231: __init__ method from base class '_RealSyscall' is not called (super-init-not-called)
core/src/trezor/loop.py:264:4: W0231: __init__ method from base class '_RealSyscall' is not called (super-init-not-called)
core/src/trezor/loop.py:417:4: W0231: __init__ method from base class '_RealSyscall' is not called (super-init-not-called)
```

2) Making `wait` or `mailbox` generic - either by using `def wait(Syscall, Generic[T])` or by using `def wait(Syscall[T])` for generic `Syscall` - does not help. IIUC, it is because there is no way for the type-checker to know, what `T` actually is.

example from `homescreen.py`
```
        usbcheck = loop.wait(io.USB_EVENT)
        while True:
            event = await usbcheck
```
the `event` is still anotated as `Any`, the type-checker does not know what it should be, even when `wait` is annotated (by `Syscall[T]`, or `Syscall, Generic[T]` - both yield the same result) (note that `Syscall[T]` has `__await__` annotated too) 
```
class wait(Syscall[T]):
...
    if TYPE_CHECKING:

        def __await__(self) -> Generator[Any, None, T]:
            return self.__iter__()
```            




Alternatively, I can use, e.g. 
`my_box = mailbox[int|None]()`
and then the `my_box` will accept only `int` or `None` (otherwise the type-checker will get angry). It works GREAT, until the TYPE_CHECKING is False. When `TYPE__CHECKING` is False, the `mailbox[...]` does not work, because it is not "indexable".
- It would need to have `__class_getitem__` implemented.
- It would still be not user-friendly for cases when I would want to pass types that should not be available without TYPE_CHECKING, e.g. `mailbox[Layout]`.


To mitigate problem number 1 - it might be possible to have generic Syscall, and add some TYPE_CHECKING gated constructor to each of the subclasses that use it. IIUC, it could work the same as it works now, I would just add gated constructor instead of gated `__await__`. However, I think that I still might need a specific `__await__` for sleep, where the return type is `int`.

matejcik:
i wouldn't mind just silencing the W0231 on those lines, pretty sure that's just pylint (or whatever) being confused



> usbcheck = loop.wait(io.USB_EVENT)
>         while True:
>             event = await usbcheck

this one may be solvable by annotating `io.USB_EVENT` with a type
```
# trezorio.pyi
class EventInterface(int, Generic[T]):
    pass

USB_EVENT: EventInterface[int]

# loop.py
class wait(Syscall[T]):
    def __init__(self, iface: io.EventInterface[T]) -> None:
        ...
```



> Alternatively, I can use, e.g.
> `my_box = mailbox[int|None]()`

how about `my_box: loop.mailbox[int | None] = mailbox()` ?

matejcik:
the advantage of generic `Syscall` is that you don't need to repeat the `__await__` declaration in every subclass

M1nd3r:
@matejcik 
Changes implemented
- changed `Syscall` to `Generic` to reduce the number of `__await__` definitions
- annotated `mailbox`

Not implemented:
- changes to `wait` - the `EventInterface[int]` approach.

core/src/trezor/loop.py
```diff
@@ -190,18 +204,26 @@ def __iter__(self) -> Generator:
...
     def handle(self, task: Task) -> None:
         pass
 
+    ...
+
+
+if TYPE_CHECKING:
+    Syscall = _RealSyscall
+else:
+    Syscall = [_RealSyscall]
```
matejcik:
```suggestion
    Syscall = {T: _RealSyscall, int: _RealSyscall}
```
and then...

core/src/trezor/loop.py
```diff
@@ -190,18 +204,26 @@ def __iter__(self) -> Generator:
...
+
 
 class Timeout(Exception):
     pass
 
 
-class sleep(Syscall):
+class sleep(Syscall[T]):
```
matejcik:
```suggestion
class sleep(Syscall[int]):
```
...and then...

core/src/trezor/loop.py
```diff
@@ -219,8 +241,15 @@ def handle(self, task: Task) -> None:
         deadline = utime.ticks_add(utime.ticks_ms(), self.delay_ms)
         schedule(task, deadline, deadline)
 
+    if TYPE_CHECKING:
+
+        def __await__(self) -> Generator[Any, None, int]:
+            return (  # type: ignore Expression of type "Generator[Any, Any, T@sleep]" is incompatible with return type "Generator[Any, None, int]"
+                super().__await__()
+            )
+
```
matejcik: ...then you can drop this

core/src/trezor/loop.py
M1nd3r: Nice solution, thanks!
Changes here: https://github.com/trezor/trezor-firmware/pull/5713/commits/9e338b2d7fd18e7b4ec40a0ded857b8e381fadea

core/src/trezor/loop.py
M1nd3r: Changed here https://github.com/trezor/trezor-firmware/pull/5713/commits/9e338b2d7fd18e7b4ec40a0ded857b8e381fadea

core/src/trezor/loop.py
M1nd3r: Dropped here: https://github.com/trezor/trezor-firmware/pull/5713/commits/9e338b2d7fd18e7b4ec40a0ded857b8e381fadea

PR #5714, opened 2025-09-05
bootloader led effects

kopecdav:
This PR add RGB LED effects in bootloader workflows to indicate CHARGING, PAIRING and critical states.

### DIFF
--- a/core/embed/io/rgb_led/unix/rgb_led.c
+++ b/core/embed/io/rgb_led/unix/rgb_led.c
@@ -100,9 +100,14 @@ void rgb_led_effect_stop(void) {
   return;
 }
 
-bool rgb_led_get_effect_ongoing(void) {
+bool rgb_led_effect_ongoing(void) {
   // RGB effect not supported in unix yet
   return false;
+}
+
+rgb_led_effect_type_t rgb_led_effect_get_type(void) {
+  // RGB effect not supported in unix yet
+  return RGB_LED_EFFECT_NONE;
 }
 
 #endif /* KERNEL_MODE */


core/embed/io/rgb_led/stm32u5/rgb_led_effects.c
```diff
@@ -105,7 +105,7 @@ static void rgb_led_linear_gc_effect(uint32_t c0, uint32_t c1,
 // Assign effect callback from the lookup table
 bool rgb_led_assign_effect(rgb_led_effect_t *effect,
                            rgb_led_effect_type_t effect_type) {
-  if (effect_type >= RGB_LED_NUM_OF_EFFECTS) {
+  if (effect_type >= RGB_LED_NUM_OF_EFFECTS && effect_type < 0) {
```
TychoVrahe: || should be used

TychoVrahe:
This solution has one additional problem.

When you are in 'host control' and start some flow, like installing FW, the LED effect will not be updated during the flow even if you i.e. disconnect the charging cable and so on. Kinda feels like a blocker (for the usage of charging effect), and i don't see a simple solution yet. cc @Hannsek

core/embed/io/rgb_led/stm32u5/rgb_led_effects.c
kopecdav: Good catch, thanks! fixup here: https://github.com/trezor/trezor-firmware/pull/5714/commits/6831aebe12c4c381dc96175f46bd9124106b48d8

kopecdav:
According to team agreement, charging effect in active bootloader was reverted, so now the charging effect is used only in suspend, or hibernation mode.

PR #5722, opened 2025-09-07
T3W1 startup adjustment

TychoVrahe:
This PR adjust startup of empty T3W1 device. Since nRF is starting slow, its undesirable to immediately show controls to user as clicking these will lead to failures. Let just show logo and wait for nRF being ready

### DIFF
--- a/core/embed/projects/bootloader/workflow/wf_empty_device.c
+++ b/core/embed/projects/bootloader/workflow/wf_empty_device.c
@@ -33,6 +33,10 @@
 #include <sys/backup_ram.h>
 #endif
 
+#ifdef USE_BLE
+#include <io/ble.h>
+#endif
+
 #include "bootui.h"
 #include "rust_ui_bootloader.h"
 #include "workflow.h"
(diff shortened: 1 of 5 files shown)


core/embed/projects/bootloader/workflow/wf_empty_device.c
```diff
@@ -48,6 +52,18 @@ workflow_result_t workflow_empty_device(void) {
   ensure(backup_ram_erase_protected() * sectrue, NULL);
 #endif
 
+#ifdef USE_BLE
+  screen_boot_empty();
+  uint32_t timeout = ticks_timeout(5000);
```
obrusvit, line 57: Any reason for the value 5s worth commenting?

obrusvit (approved):
utACK, saw the video, looks good.

core/embed/projects/bootloader/workflow/wf_empty_device.c
TychoVrahe: not really, should be enough for nRF to start but doesn't take forever in case of the chip malfunction

PR #5724, opened 2025-09-08
refactor BLE interface

TychoVrahe:


core/embed/io/ble/inc/io/ble.h
```diff
@@ -33,18 +33,6 @@
 
 #define BLE_MAX_BONDS 8
```
kopecdav, line 39: Is this still 8?

core/embed/io/ble/inc/io/ble.h
kopecdav, line 1: Can you also put some brief comment to this enum? 

```
typedef enum {
  BLE_MODE_OFF,
  BLE_MODE_KEEP_CONNECTION,
  BLE_MODE_CONNECTABLE,
  BLE_MODE_PAIRING,
  BLE_MODE_DFU,
} ble_mode_t;
```

kopecdav (changes requested):
This is a good refactor, cuts out lots of duplicit code and makes the driver much more readable. Leaving just two nitpicks. 

I think it would also make sense to review and refactor the overall multipoint/singlepoint logic, feels like there are some gaps there, but that is definitely for another PR.

core/embed/io/ble/inc/io/ble.h
TychoVrahe: yes this matches `CONFIG_BT_MAX_PAIRED` in `nordic/trezor/trezor-ble/prj.conf`

core/embed/io/ble/inc/io/ble.h
TychoVrahe: added https://github.com/trezor/trezor-firmware/commit/f883114ca9e96cfd0800dd390826cc36b8494f03

PR #5728, opened 2025-09-08
Turn long menu subtexts into marquee

ibz:


core/embed/rust/src/ui/component/marquee.rs
```diff
@@ -24,6 +24,7 @@ enum State {
 
 pub struct Marquee {
     area: Rect,
+    y_offset: i16,
```
bieleluk: Isn't there a better way, i.e., to restrict the area to the exact size so there is no need to add the offset?

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -45,6 +45,8 @@ pub struct Button {
     long_press_danger: bool,
     long_timer: Timer,
     haptic: bool,
+
+    subtext_marquee: Marquee,
```
bieleluk: The subtext is divided between the `button_content` and a separate entry `subtext_marquee`. I would prefer that the marquee be part of the `ButtonContent`. On the other hand, that would require implementing the `Component` trait, which might not be feasible.
 In any case, the `subtext_marquee` should be an `Option`.

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -59,6 +61,24 @@ impl Button {
...
+                subtext,
+                subtext_style,
+                ..
+            } => Marquee::new(
+                subtext,
+                subtext_style.text_font,
+                subtext_style.text_color,
+                subtext_style.background_color,
+            ),
+            _ => Marquee::new(
```
bieleluk: If `subtext_marquee` was an `Option`, this might be the `None` case.

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -257,6 +273,14 @@ impl Button {
     pub fn enable_if(&mut self, ctx: &mut EventCtx, enabled: bool) {
         if enabled {
             self.enable(ctx);
+            if let ButtonContent::TextAndSubtext {
```
bieleluk: After a discussion with @lapohoda, the following is a desired behavior when the button is disabled:
- When the button is disabled, the marquee should reset its state and should not move.

When it's re-enabled, it should start moving from the initial position again.

Here is a diff that does that
[diff_marquee.txt](https://github.com/user-attachments/files/22252199/diff_marquee.txt)

core/embed/rust/src/ui/layout_eckhart/firmware/vertical_menu_screen.rs
```diff
@@ -222,6 +222,7 @@ impl<T: MenuItems> Component for VerticalMenuScreen<T> {
         // This is needed to enable swipe detection only when the menu does not fit
         if let Event::Attach(_) = event {
             self.initialize_screen(ctx);
+            self.menu.event(ctx, event);
```
bieleluk: If the new `marquee` behavior from [diff_marquee.txt](https://github.com/user-attachments/files/22252199/diff_marquee.txt)  is implemented, this line is redundant.

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -624,10 +652,32 @@ impl Component for Button {
...
     fn place(&mut self, bounds: Rect) -> Rect {
         self.area = bounds;
+        let subtext_start =
+            self.baseline_text_height() * 2 + constant::LINE_SPACE + self.baseline_subtext_height();
+        self.subtext_marquee.set_y_offset(subtext_start);
+        self.subtext_marquee.place(self.area.inset(Insets::new(
+            0,
+            self.content_offset.x,
+            0,
+            self.content_offset.x / 2,
```
bieleluk: It's questionable, whether the initial position should be aligned with the main text or whether the subtext should not go further left than the main text. Personally, I would be for the second one but haven't strong opinion about it.
```suggestion
            self.content_offset.x,
```

bieleluk (commented):
I have several code suggestions and one critical test finding: 
When testing the label setting in the device menu, I encountered the `Timer queue` error. This didn't happen on the `main` branch.
```
=== FATAL ERROR
Location: src/ui/component/base.rs:603
Message: Timer queue is full
===
ERROR: Timer queue is full
```

core/embed/rust/src/ui/layout_eckhart/component/button.rs
ibz: Yeah, I don't think `ButtonContent` should be a `Component` - it just represents the data that is used to configure a `Button`, the component is still the `Button` itself. But I agree with the `Option` and did as you suggested: https://github.com/trezor/trezor-firmware/pull/5728/commits/99eab2bf1ceb9d452849ee031faa0832f09d2696#diff-751893ff48c24c322109d4391c8d9b7ff7a10a27fb2605ff09565fbea8d48f14R49

core/embed/rust/src/ui/layout_eckhart/component/button.rs
ibz: Looks good, thank you! https://github.com/trezor/trezor-firmware/pull/5728/commits/99eab2bf1ceb9d452849ee031faa0832f09d2696#diff-751893ff48c24c322109d4391c8d9b7ff7a10a27fb2605ff09565fbea8d48f14R284

core/embed/rust/src/ui/layout_eckhart/component/button.rs
ibz: I think it looks worse with that change. Not only the misalignment with the button text, but also the fact that once it scrolls until the end, the last letter is not fully visible...

core/embed/rust/src/ui/component/marquee.rs
ibz: As you suggested, using `in_clip` in the marquee fixed the issue I was having with the scrolling menu and the marquee restricted to an area. 4ce137796e489815796f150fd9cb118e4adda0f8

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -624,10 +645,23 @@ impl Component for Button {
...
     fn place(&mut self, bounds: Rect) -> Rect {
         self.area = bounds;
+        let subtext_start =
+            self.baseline_text_height() * 2 + constant::LINE_SPACE + self.baseline_subtext_height();
+        if let Some(m) = &mut self.subtext_marquee {
+            m.place(self.area.inset(Insets::new(
+                subtext_start,
+                self.content_offset.x,
+                0,
+                self.content_offset.x / 2,
```
bieleluk: After changing the `marquee` rendering function, the placement has changed, so there is no gap at the beginning:
```suggestion
                self.content_offset.x,
```
It might have the same effect in other models as well.

ibz: Yep, looks great now! 6e7df70e6ae1f4127fb096cbc598e04a251f7c35

PR #5732, opened 2025-09-09
Enabling dev signed translation for QA testing

PrisionMike:
Dev signed translation blobs will now be accepted even in optimised build, as long as PRODUCTION flag is not set.


matejcik:
please flip the feature logic
that is: we must **enable** feature (like `dev_keys`) in order to get this functionality, that is **disabled in vanilla build**

matejcik:
then `debug` feature can also imply `dev_keys` (but doesn't have to)

keying it to `PRODUCTION` is ~fine, though my personal preference would be to set a special flag for it; _or_ key it to `BOOTLOADER_DEVEL` or _something_, so that this behavior is not implicit but rather a deliberate choice.

but leaving both of those decisions to @obrusvit

PrisionMike:
[no changelog]

obrusvit:
Agreed with @matejcik 

Let's introduce a new Rust feature called `dev_keys` which will be set only if `PRODUCTION==0` in build. I believe that should cover all cases.

PrisionMike:
[no changelog]

core/Makefile
```diff
@@ -50,6 +50,7 @@ TREZOR_DISABLE_ANIMATION ?= $(if $(filter 0,$(PYOPT)),1,0)
 STORAGE_INSECURE_TESTING_MODE ?= 0
 UI_PERFORMANCE_OVERLAY ?= 0
 DBG_CONSOLE ?=
+DEV_KEYS ?= 0
```
obrusvit: Let's not introduce a new compilation flag and let's tie it to `PRODUCTION`. Please remove `DEV_KEYS` flag

core/SConscript.firmware
```diff
@@ -874,6 +875,8 @@ if EVERYTHING:
     features.append('universal_fw')
 if UI_PERFORMANCE_OVERLAY:
     features.append('ui_performance_overlay')
+if DEV_KEYS and not PRODUCTION:
```
obrusvit:
```suggestion
if not PRODUCTION:
```

equivalent change to other parts pls

obrusvit:
@PrisionMike In order to pass the Changelog check, put "[no changelog]" to the last line of the commit message.

FYI, alternative is making a changelog entry as described [here](https://docs.trezor.io/trezor-firmware/misc/changelog.html) but that is not relevant for this PR. Changelogs are supposed to be more user oriented.

PR #5733, opened 2025-09-09
chore(core/eckhart): enable going back in the 1st recovery word

bieleluk:
### 1st-share back behavior
- [x] The flow goes back to the word number selection

### Higher-share back behavior
- [x] go to the previous checkpoint

### TODO
- [x] device/click tests
- [ ] Find out why the fixtures for other models changed.

### DIFF
--- a/core/src/apps/management/recovery_device/homescreen.py
+++ b/core/src/apps/management/recovery_device/homescreen.py
@@ -134,6 +134,8 @@ async def _continue_recovery_process() -> Success:
 
         # if they were invalid or some checks failed we continue and request them again
         if not words:
+            if not is_first_step:
+                await _request_share_next_screen()
             continue
 
         try:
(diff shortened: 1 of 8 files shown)


core/src/trezor/ui/layouts/eckhart/recovery.py
```diff
@@ -36,7 +36,7 @@ async def request_word(
     prefill_word: str = "",
 ) -> str:
     prompt = TR.recovery__word_x_of_y_template.format(word_index + 1, word_count)
-    can_go_back = word_index > 0
+    can_go_back = True
```
ibz: Variable became useless..

ibz:
Tested and it works.

core/src/trezor/ui/layouts/eckhart/recovery.py
bieleluk: https://github.com/trezor/trezor-firmware/pull/5733/commits/898034c512bc61190d4fdf0b453fe47ca8949977

core/src/apps/management/recovery_device/homescreen.py
```diff
@@ -134,6 +134,8 @@ async def _continue_recovery_process() -> Success:
 
         # if they were invalid or some checks failed we continue and request them again
         if not words:
+            if not is_first_step:
```
mmilata, line 137: We can only end up if the backup is SLIP39, right?

bieleluk: yes

PR #5734, opened 2025-09-09
build(core): fix CI build error

romanz:
Use https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-nonstring-variable-attribute to avoid the following build issue:
```
chacha20poly1305/chacha_merged.c:26:31: error: initializer-string for character array is too long, array size is 16 but initializer has size 17 (including the null terminating character); did you mean to use the 'nonstring' attribute? [-Werror,-Wunterminated-string-initialization]
   26 | static const char sigma[16] = "expand 32-byte k";
      |                               ^~~~~~~~~~~~~~~~~~
chacha20poly1305/chacha_merged.c:27:29: error: initializer-string for character array is too long, array size is 16 but initializer has size 17 (including the null terminating character); did you mean to use the 'nonstring' attribute? [-Werror,-Wunterminated-string-initialization]
   27 | static const char tau[16] = "expand 16-byte k";
      |                             ^~~~~~~~~~~~~~~~~~
2 errors generated.
```
https://github.com/trezor/trezor-firmware/actions/runs/17579981560/job/49933897339


romanz:
It also fails on `vendor/secp256k1-zkp/src/modules/schnorrsig/main_impl.h`:
```
In file included from ../vendor/secp256k1-zkp/src/secp256k1.c:830:
../vendor/secp256k1-zkp/src/modules/schnorrsig/main_impl.h:48:46: error: initializer-string for character array is too long, array size is 13 but initializer has size 14 (including the null terminating character); did you mean to use the 'nonstring' attribute? [-Werror,-Wunterminated-string-initialization]
   48 | static const unsigned char bip340_algo[13] = "BIP0340/nonce";
      |                                              ^~~~~~~~~~~~~~~
1 error generated.
```
IIUC, it was fixed in https://github.com/bitcoin-core/secp256k1/commit/fa67b6752d8ba3e4c41f6c36b1c6b94a21770419 and released here:
- https://github.com/bitcoin-core/secp256k1/releases/tag/v0.6.0
- https://github.com/bitcoin-core/secp256k1/releases/tag/v0.7.0

crypto/chacha20poly1305/chacha_merged.c
```diff
@@ -23,8 +23,8 @@ void ECRYPT_init(void)
   return;
 }
 
-static const char sigma[16] = "expand 32-byte k";
-static const char tau[16] = "expand 16-byte k";
+static const char sigma[16] __attribute__((nonstring)) = "expand 32-byte k";
+static const char tau[16] __attribute__((nonstring)) = "expand 16-byte k";
```
andrewkozlik: This a valid solution, but why not use the fully portable solution which they used in https://github.com/bitcoin-core/secp256k1/blob/36e76952cbf1cf54ddd2d8756cc31a486e2ba1d9/src/modules/schnorrsig/main_impl.h#L48?
```suggestion
static const char sigma[] = {'e', 'x', 'p', 'a', 'n', 'd', ' ', '3', '2', '-', 'b', 'y', 't', 'e', ' ', 'k'};
static const char tau[] = {'e', 'x', 'p', 'a', 'n', 'd', ' ', '1', '6', '-', 'b', 'y', 't', 'e', ' ', 'k'};
```

romanz: Sounds good - fixed in 50e47f3059.

PR #5738, opened 2025-09-09
feat(core): trim strings to specific bytes' limit

romanz:
Will be used for trimming cached THP host names (https://satoshilabs.slack.com/docs/T0J8V2YBY/F09DU00DW0P).

### DIFF
--- a/core/src/trezor/strings.py
+++ b/core/src/trezor/strings.py
@@ -153,3 +153,16 @@ def format_autolock_duration(auto_lock_ms: int) -> str:
         auto_lock_label = TR.plurals__lock_after_x_seconds
 
     return format_plural("{count} {plural}", auto_lock_num, auto_lock_label)
+
+
+def trim_str(s: str, max_bytes: int) -> str:
+    """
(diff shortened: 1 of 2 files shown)


core/src/trezor/strings.py
```diff
@@ -153,3 +153,18 @@ def format_autolock_duration(auto_lock_ms: int) -> str:
...
+    """
+    assert max_bytes >= 0
+    chars = 0
+    for char in s:
+        char_size = len(char.encode())
+        if max_bytes < char_size:
+            break
+        max_bytes -= char_size
+        chars += 1
+    return s[:chars]
```
bieleluk: nit: The approach with enumerate is more readable, IMO.
```suggestion
    for i, char in enumerate(s):
        char_size = len(char.encode())
        if max_bytes < char_size:
            return s[:i]
        max_bytes -= char_size
    return s
```

romanz: Thanks! Changed in 5385bc5662.

romanz:
Squashed into 5385bc5662 into fc443f2a4.

PR #5742, opened 2025-09-10
Upgrade pyright to 1.1.404

matejcik:
Of course it's not .405 because of course [there's a blocker bug](https://github.com/microsoft/pyright/issues/10906)

The **overwhelming majority** of this PR is renaming specific instances of `bytes` to a newly introduced type `AnyBytes = bytes | bytearray | memoryview` 
and in fact if you run search&replace for `AnyBytes` -> `bytes`, you'll get a significantly more manageable review task.

Why?
To resolve **byte types promotion problem**. A long standing issue in Python typing ecosystem was that the types `bytes`, `bytearray` and `memoryview` are similar enough that programmers are using them interchangeably in practice ... but in actuality, there's no inheritance relationship between them and they are in fact incompatible in a lot of ways.
So in other words a function annotated with `foo(buf: bytes)` may do an operation that is, e.g., not supported on a `memoryview`, and fail at run-time.
Or, you know. Use the bytes as a key in a dict, then later mutate it, and then act surprised when the value can't be looked up.

At the same time, a massive base of existing code (including ours) relies on the similarities so much that making the distinction breaking would turn everyone off from using type annotations.
Well, as it happens, Pyright 1.1.400 chose to flip the switch and detect the distinction.

This can be disabled via a config option ... but interestingly, doing _that_ brings other problems! (which i don't remember right now because I investigated them several months ago and decided that keeping this switch on is a better choice for our repo)

So this PR replaces `bytes` with `AnyBytes`, forcing users of the "AnyBytes" values to _only_ use the common subset of available operations.
This is actually perfectly good most of the time.
One sticky point is that I turned all protobuf bytes fields to AnyBytes, meaning we can't rely on them being `bytes | bytearray` and supporting e.g. `.startswith()`. That means in a small number of places, a "useless" conversion to bytes is done on the value coming from protobuf.
As an advantage, this will allow us, at any time, to change the protobuf impl to return memoryviews, which seems like a useful property.

---

Another change worth mentioning is that ANN warnings from `flake8-annotations` are now enabled for trezorlib, i.e., it's no longer ok to leave out function return type annotations.

The rest of it is mostly uninteresting.


matejcik:
review set to @obrusvit, but please assign whoever you feel should actually do it

mmilata:
i'll take it

python/src/trezorlib/protobuf.py
```diff
@@ -341,7 +340,7 @@ def __init__(self, reader: Reader, limit: int) -> None:
         self.reader = reader
         self.limit = limit
 
-    def readinto(self, buf: bytearray) -> int:
+    def readinto(self, buf: bytearray, /) -> int:
```
mmilata, line 343: Just curious, why the `/` here but not other places?

vendor/secp256k1-zkp
romanz, line 1: it seems to undo [`vendor/secp256k1-zkp` update](https://github.com/trezor/trezor-firmware/commit/9491ab65dc49a2633ddf1bd452f16d0a01c422bf):
```
$ git show 9e3ff5b4f9 -- vendor
commit 9e3ff5b4f9f6eecbe7feeabf0cf72469e978d87b
Author: matejcik <ja@matejcik.cz>
Date:   Wed Sep 10 11:43:05 2025 +0200

    style(core): update types to pyright 1.1.404
    
    Majority by volume is resolving the byte types promotion problem, which
    pyright stopped ignoring

diff --git a/vendor/secp256k1-zkp b/vendor/secp256k1-zkp
index a660a4976e..f8c1b0e0e6 160000
--- a/vendor/secp256k1-zkp
+++ b/vendor/secp256k1-zkp
@@ -1 +1 @@
-Subproject commit a660a4976efe880bae7982ee410b9e0dc59ac983
+Subproject commit f8c1b0e0e665d81be9074d9702d97ef209cca32d
```
`vendor/secp256k1-zkp` was updated in https://github.com/trezor/trezor-firmware/pull/5734:
```
$ git show 9491ab65dc49a2633ddf1bd452f16d0a01c422bf -- vendor
commit 9491ab65dc49a2633ddf1bd452f16d0a01c422bf
Author: Roman Zeyde <roman.zeyde@satoshilabs.com>
Date:   Tue Sep 9 14:22:18 2025 +0300

    chore(crypto): update secp256k1 submodule to 0.7.0
    
    [no changelog]

diff --git a/vendor/secp256k1-zkp b/vendor/secp256k1-zkp
index f8c1b0e0e6..a660a4976e 160000
--- a/vendor/secp256k1-zkp
+++ b/vendor/secp256k1-zkp
@@ -1 +1 @@
-Subproject commit f8c1b0e0e665d81be9074d9702d97ef209cca32d
+Subproject commit a660a4976efe880bae7982ee410b9e0dc59ac983
```

romanz (changes requested):

python/src/trezorlib/protobuf.py
matejcik: it's a protocols thing. protocols in pyright are sensitive to names of arguments that are not position-only. this is a protocol function and i didn't want to force implementers to always use the same name.

vendor/secp256k1-zkp
matejcik: fixed in rebase

romanz (approved):

PR #5743, opened 2025-09-10
Use THP hostname in BLE device menu

romanz:


core/src/apps/thp/pairing.py
```diff
@@ -474,3 +478,19 @@ def _check_method_is_allowed(ctx: PairingContext, method: ThpPairingMethod) -> N
...
+
+def _cache_host_name(mac_addr: bytes, host_name: str) -> None:
+    from trezor.messages import ThpPairedCacheEntry
+    from trezor.strings import trim_str
+
+    from . import paired_cache
+
+    entries = paired_cache.load()
+    if any(mac_addr == e.mac_addr for e in entries):
+        # skip writing to flash if this MAC address is already cached
```
mmilata, line 491: So we always keep the first host_name seen. I suppose this is better if there are two (or more) apps that present different hostnames for some reason, but probably suboptimal if there's only one app and the user changes the hostname at some point ... Can't think of anything better so no action needed here.

romanz: I agree - we probably should somehow let the user know that the displayed hostname is "sticky" (i.e. doesn't change after the first THP pairing).
CC: @Hannsek

romanz:
> But everything should fit fine in the 128KiB we have if I'm reading this right.

Yes - since we limit the cache entries to [match the current bonds](https://github.com/trezor/trezor-firmware/blob/d7e97cd23bc3efa74128956ec68dfcc8621bdc4a/core/src/apps/thp/paired_cache.py#L30-L31), there should be ≤8 such entries, each taking ~40 bytes[^1], so the max serialized size should be ≤400 bytes (d7e97cd23bc3efa74128956ec68dfcc8621bdc4a). 

IIUC, only the [`_HOMESCREEN`](https://github.com/trezor/trezor-firmware/blob/d7e97cd23bc3efa74128956ec68dfcc8621bdc4a/core/src/storage/device.py#L21) entry can be larger.

[^1]: one entry = 32 bytes for hostname + 6 bytes for MAC addres + some protobuf overhead

romanz:
Pushed https://github.com/trezor/trezor-firmware/commit/d7e97cd23bc3efa74128956ec68dfcc8621bdc4a to exclude THP storage-related functions from non-THP builds.

romanz:
Squashing ~and rebasing~ over latest `main`.

core/src/apps/homescreen/device_menu.py
```diff
@@ -25,6 +27,8 @@ def _find_device(connected_addr: bytes | None, bonds: list[bytes]) -> int | None
 async def handle_device_menu() -> None:
     from trezor import strings
 
+    from ..thp import paired_cache
```
bieleluk, line 31: Let's make a THP a requirement for the device menu

romanz: https://github.com/trezor/trezor-firmware/pull/5743/commits/09f8b460b6ebf0ab423c2871178bf0764cdf98a5

romanz:
Tested locally with two hosts.
The THP paired host name appears in Bluetooth menu, which the unpaired host entry is showing its BLE MAC address:
https://satoshilabs.slack.com/archives/C07JWCTDWEA/p1757585313632649?thread_ts=1756369084.988519&cid=C07JWCTDWEA

romanz:
I have reverted https://github.com/trezor/trezor-firmware/pull/5734/commits/b90aa10303eee542bf2b9d42e08733c50198b77b by mistake, pushing a fix: bf545ec68d

romanz:
Squashed and force-pushed.

PR #5748, opened 2025-09-11
introduce push notifications 

TychoVrahe:


core/embed/sys/syscall/stm32/syscall_stubs.c
```diff
@@ -150,6 +150,16 @@ void reboot_device(void) {
     ;
 }
 
+// =============================================================================
+// notify.h
+// =============================================================================
+
+#include <sys/notify.h>
+
+void notify_send(notification_event_t event) {
```
kopecdav, line 159: Should this function + `notification_event_t` also be allowed in `build.rs` ?

nordic/trezor/trezor-ble/src/ble/service.c
```diff
@@ -89,6 +94,26 @@ int service_send(struct bt_conn *conn, trz_packet_t *data) {
...
+  params.attr = attr;
+  params.data = buf;
+  params.len = len;
+  params.func = on_sent;
+  params.user_data = (void *)buf;
+
+  if (conn && bt_gatt_is_subscribed(conn, attr, BT_GATT_CCC_NOTIFY)) {
+    return bt_gatt_notify_cb(conn, &params);
+  } else {
+    return -EINVAL;
```
kopecdav: you should probably k_free(buf) here, in case you wont make it to `bt_gatt_notify_cb` or maybe allocate the buffer within the `if`

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/692a4efb1371f802547a6250cd6ed744959f85e8

core/embed/sys/syscall/stm32/syscall_stubs.c
TychoVrahe: when we need it rust, but i think its premature

PR #5749, opened 2025-09-11
Tutorial tweaks

ibz:


core/embed/rust/src/ui/layout_eckhart/firmware/tutorial_screen.rs
```diff
@@ -170,3 +173,96 @@ impl crate::trace::Trace for TutorialWelcomeScreen {
...
+        debug_assert_eq!(bounds.width(), SCREEN.width());
+
+        let (rest, action_bar_area) = bounds.split_bottom(theme::ACTION_BAR_HEIGHT);
+        let content_area = rest.inset(theme::SIDE_INSETS).inset(Insets::top(TOP_INSET));
+
+        self.text.place(content_area);
+        self.action_bar.place(action_bar_area);
+
+        let text_height = theme::firmware::TEXT_REGULAR.text_font.line_height();
+        self.power_button_indicator_bounds = Rect::from_bottom_right_and_size(
```
bieleluk: Nit: isn't it better to hard-code the rectangle bounds? We are aiming at aligning it with the button position, which doesn't depend on the text height. The current [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd?node-id=13074-1586#1424702665) is 10px misaligned from the real button, so we should adapt the implementation accordingly, i.e.:
```
const BUTTON_WIDTH: i16 = 10;
const BUTTON_HEIGHT: i16 = 96;
const BUTTON_TOP_PADDING: i16 = 75;
let rect = Rect::from_center_and_size(
    Point::new(SCREEN.width(), BUTTON_TOP_PADDING + BUTTON_HEIGHT / 2),
    Offset::new(2 * BUTTON_WIDTH, BUTTON_HEIGHT),
);
Bar::new(rect)
    .with_radius(7)
    .with_bg(theme::GREEN_LIME)
    .render(target);
```

bieleluk:
@ibz 
There is a possibility that the power screen will change once again due to this [comment](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd?node-id=13074-1586#1424700525). We shouldn't merge the PR until we have a definitive answer.

core/embed/rust/src/ui/layout_eckhart/firmware/tutorial_screen.rs
ibz: OK: f79e160e0956a86f22796141a70ce854dd11b1c7

PR #5750, opened 2025-09-11
fix(core): use translation strings instead of literals

romanz:
Also, add a description format argument to `show_pairing_device_name()` to use #5687.


core/translations/en.json
```diff
@@ -1228,6 +1236,18 @@
       "Delizia": "",
       "Eckhart": "Allow {0} on {1} to pair with this Trezor?"
     },
+    "thp__pair_name": {
+      "Bolt": "",
```
bieleluk, line 1240: Does it make sense to use the empty string for models where it's not used?
If yes, then we should make a huge refactor of the translations.

romanz: > If yes, then we should make a huge refactor of the translations.

Do you mean refactoring the translation mechanism / blob format? or something else?

romanz: Opened a related improvement task: https://github.com/trezor/trezor-firmware/issues/5751

obrusvit: It makes sense to make the translation blobs smaller for other layouts.

I'd introduce some mechanism to signify it's not used for the layout but let's not deal with it now

Issue #4624, opened 2025-02-12
split touch driver FT6c36 and FT3168


TychoVrahe:
The drivers are different enough for using single implementation to be dangerous enough. See i.e. https://github.com/trezor/trezor-firmware/pull/4614/files

We should split it into two, so that changes made to one don't affect the other

PR #5752, opened 2025-09-12
Touch drivers' source split

bleska:
This PR resolves ISSUE https://github.com/trezor/trezor-firmware/issues/4624 .

The purpose is to split the currently shared source code for touch drivers FT6x36 and FT3168. There are differences between these two touch drivers e.g. FT3168 has a monitor mode feature from which it has to be waken up e.g. by I2C dummy read, FT6x36 doesn't have this which implies #ifdef... and more complicated code.


core/embed/io/touch/ft6x36/ft6x36.c
```diff
@@ -31,8 +31,6 @@
 #include "panels/lx154a2422cpt23.h"
 #elif defined TOUCH_PANEL_LHS200KB_IF21
 #include "panels/lhs200kb-if21.h"
-#elif defined TOUCH_PANEL_LX250A2410A
-#include "panels/lx250a2410a.h"
 #endif
```
TychoVrahe, line 34: the panel should be removed in ft6x36_panel_correction function too

core/embed/io/touch/ft3168/ft3168.c
```diff
@@ -0,0 +1,530 @@
...
+      return secfalse;
+    }
+  }
+
+  return sectrue;
+}
+
+static void ft3168_panel_correction(uint16_t x, uint16_t y, uint16_t* x_new,
+                                    uint16_t* y_new) {
+#ifdef TOUCH_PANEL_LX154A2422CPT23
```
TychoVrahe: only TOUCH_PANEL_LX250A2410A exists

core/embed/io/touch/ft6x36/ft6x36.c
bleska: fixed here: https://github.com/trezor/trezor-firmware/pull/5752/commits/ac30a1feb2751eaf0c384cc63ad6d69821e7027f

core/embed/io/touch/ft3168/ft3168.c
bleska: fixed here: https://github.com/trezor/trezor-firmware/pull/5752/commits/ac30a1feb2751eaf0c384cc63ad6d69821e7027f

TychoVrahe (approved):

PR #5756, opened 2025-09-12
chore(core/eckhart): Align power buttons with bootloader

bieleluk:


bieleluk:
The alignment will be done the other way round
https://github.com/trezor/trezor-firmware/pull/5758

PR #5757, opened 2025-09-12
send lock&unlock notifications when hard-locking

TychoVrahe:
[no changelog]


core/embed/projects/bootloader/workflow/wf_empty_device.c
```diff
@@ -100,6 +102,7 @@ workflow_result_t workflow_empty_device(void) {
       return res;
```
kopecdav: There are several places within the while loop from which you can return right away, shouldn't we also send `NOTIFY_UNLOCK` there? Or maybe instead of doing return we should probably break from loop, since we already keep the return value in `res` variable?

Also on line 88, there is a return which will escape the function without `workflow_ifaces_deinit()`, that seems incorrect.

core/src/main.py
```diff
@@ -53,6 +53,8 @@
         import ble  # noqa: F401
         del ble
 
+# send unlock notification
+utils.notify_send(1)
```
kopecdav: This seems really dangerous, any chance we can propagate the enum to micropython?

core/embed/projects/bootloader/workflow/wf_auto_update.c
```diff
@@ -49,6 +52,7 @@ workflow_result_t workflow_auto_update(const vendor_header *const vhdr,
     return WF_OK_REBOOT_SELECTED;
```
kopecdav: Probably better to break here with `res=WF_OK_REBOOT_SELECTED` or add `NOTIFY_UNLOCK`

kopecdav (changes requested):

core/embed/projects/bootloader/workflow/wf_empty_device.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/b8eb016d4301ce2eef4e1abb3db2030e61b22bc0

core/src/main.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/a5532d89e2247bfb061dc5bcb949b9eb73b99cf0

core/embed/projects/bootloader/workflow/wf_auto_update.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/b8eb016d4301ce2eef4e1abb3db2030e61b22bc0

kopecdav: Probably better to also remove the deinit on line 51?

TychoVrahe: https://github.com/trezor/trezor-firmware/commit/39a7111eaaec25b1e569d6082241726423fc5f96

PR #5759, opened 2025-09-12
Improve state handling around Tropic locking

andrewkozlik:
Changes:
- Return `OK NO` in `tropic-lock-check` if Tropic hasn't been paired.
- Fail `secrets-init` if Tropic locking has already started so that we don't overwrite MCU's pairing secrets.
- Changed global static variable `tropic_is_paired` to a local static variable. It seems confusing to me that it can and was being read, when we don't know whether it has been set correctly.
- Make sure that Tropic is paired before locking. Probably not necessary, since if the privileged key works, everything is probably OK, and if it is partially paired, e.g. factory slot was not erased, then access to it's functionality should be blocked by the config. But better safe than sorry.

### DIFF
--- a/core/embed/projects/prodtest/cmd/prodtest_secrets.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_secrets.c
@@ -33,6 +33,7 @@
 
 #ifdef USE_OPTIGA
 #include <sec/optiga.h>
+#include "prodtest_optiga.h"
 #endif
 
 #ifdef USE_TROPIC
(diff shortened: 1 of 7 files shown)


onvej-sl:
You can adopt https://github.com/trezor/trezor-firmware/commit/76f9b36e268df064649d66888b6c03a6cbcf487d, where I fixed more places where `return` is called instead of `goto cleanup`.

core/embed/projects/prodtest/cmd/prodtest_secrets.c
```diff
@@ -117,6 +117,14 @@ static void prodtest_secrets_init(cli_t* cli) {
...
 
 #ifdef USE_TROPIC
+  // Make sure that Tropic pairing hasn't started so that we don't overwrite the
+  // MCU's pairing secrets.
+  curve25519_key tropic_public = {0};
+  if (secret_key_tropic_public(tropic_public) == sectrue) {
+    cli_error(cli, CLI_ERROR, "Tropic pairing has already started.");
+    return;
+  }
+
```
onvej-sl, line 153: Is NRF pairing reversible? Could it happen that you pair with the NRF chip, call `secrets-init` and won't be able to pair again with the NRF chip?

andrewkozlik: @TychoVrahe?

andrewkozlik: Currently `nrf-pair` is not allowed by prodtest if Optiga is locked or the serial number has been written. The nRF chip does not impose any restrictions on writing the secret.
- The Optiga-unlocked requirement does not make sense, because an attacker could replace the Optiga with one that is not locked, read the nRF secret via prodtest and then put the original Optiga back in.
- The SN-not-written requirement is quite unintuitive IMO. It implies that nRF must be paired before writing the SN, which is a surprising requirement IMO.

I removed these two checks and replaced them with a secrets-not-locked requirement in https://github.com/trezor/trezor-firmware/pull/5759/commits/e21e2e71b309e7e2bc63cb4762bad07103e7fe72. I think that this is not perfect either, but it is more intuitive than the SN-not-written requirement. The ideal solution would probably be to have a separate flag in the flash or OTP for "nRF locked". But I don't think we want yet another flag.

andrewkozlik:
> Fail `secrets-init` if Tropic locking has already started so that we don't overwrite MCU's pairing secrets.

Added an additional check if Optiga is already locked here: https://github.com/trezor/trezor-firmware/pull/5759/commits/5853c021a8260b2760ee524f3e1897b73be9dc88

core/embed/projects/prodtest/cmd/prodtest_secrets.c
TychoVrahe: sorry, missed this completely. I prepared similar solution to https://github.com/trezor/trezor-firmware/commit/e21e2e71b309e7e2bc63cb4762bad07103e7fe72 in https://github.com/trezor/trezor-firmware/pull/5780 before i noticed this, it goes a bit further so please check it out

andrewkozlik: So the change https://github.com/trezor/trezor-firmware/pull/5759/commits/e21e2e71b309e7e2bc63cb4762bad07103e7fe72 that I made in prodtest_secrets.c is redundant. I can revert that, or should I keep it anyway? The rest can be rebased over this branch.

onvej-sl (approved):

PR #5766, opened 2025-09-14
feat(core/eckhart): backup needed menu item

bieleluk:
Add "Backup Device" option to Device Menu when backup is needed.


ibz (approved):

core/src/apps/homescreen/device_menu.py
```diff
@@ -129,19 +131,30 @@ async def handle_device_menu() -> None:
                 )
                 await wipe_device(WipeDevice())
             except ActionCancelled:
-                init_submenu = SubmenuId.ROOT
+                init_submenu_idx = SubmenuId.ROOT
```
obrusvit, line 168: Is there a reason why sometimes the `ini_submenu_idx` is in the `except` block and sometimes in the `finally` block? Can you unify?

bieleluk: The reason is that there are different cases:
- different screens after different results: `except` and `else` are used for setting the new screen
- similar screen after any result: `finally` is used

We can, however, change the logic if some transition doesn't make sense.

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -269,6 +268,8 @@ enum Subscreen {
     AboutScreen,
     // A screen showing the regulatory information
     RegulatoryScreen,
+    // A screen showing information about a backup
```
ibz: Nit: not about "a backup". It's literally *the* backup.

That being said, the comment doesn't add much value as the enum member explains it all.

ibz (approved):

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: Left a comment there to be consistent with other enum entries:
https://github.com/trezor/trezor-firmware/pull/5766/commits/1f8628b1bfc365a867d4ab40d25d9737bfc413d4

PR #5765, opened 2025-09-14
Allow backup check only when the backup exists

bieleluk:

PR #5767, opened 2025-09-14
Firmware auto-suspend

obrusvit:
This PR implements auto-suspend in firmware and also improves the behavior of regular suspend (initiated by power button) following https://github.com/trezor/trezor-firmware/pull/5606

Targeting https://github.com/trezor/trezor-firmware/pull/5765


core/src/apps/base.py
```diff
@@ -609,6 +636,14 @@ def reload_settings_from_storage() -> None:
     workflow.idle_timer.set(
         storage_device.get_autolock_delay_ms(), lock_device_if_unlocked
     )
+
+    if utils.USE_POWER_MANAGER:
+        # autodim setting is not from storage but keeping it here for simplicity
+        workflow.idle_timer.set(10_000, autodim_display)  # TODO: change back to 30s
```
matejcik: TODO when

core/src/boot.py
```diff
@@ -117,8 +137,8 @@ async def bootscreen() -> None:
 config.init(show_pin_timeout)
 translations.init()
 
-if __debug__ and not utils.EMULATOR:
-    config.wipe()
+# if __debug__ and not utils.EMULATOR:
+#     config.wipe()
```
matejcik: shouldn't get commited

core/src/trezor/ui/__init__.py
```diff
@@ -105,18 +106,14 @@ def set_current_layout(layout: "Layout | ProgressLayout | None") -> None:
 
     def _handle_power_button_press() -> None:
         """Handle power button press event during firmware operation."""
-        from trezor import config
+        from apps.base import lock_device_if_unlocked
```
matejcik: this tells me that `apps.base` is not the right place for `lock_device_if_unlocked`.
you shouldn't be importing `apps` here, but i'm guessing you want this function to trigger e.g. from the boot pin dialog, correct? so that's now pulling `apps.base` which is undesirable 

something something, at least make it `apps.common` that i think is already pulled by the bootup pin screen?
(not super important to resolve in this PR today, but it _should_ be resolved soonish

core/src/trezor/ui/__init__.py
```diff
@@ -105,18 +106,14 @@ def set_current_layout(layout: "Layout | ProgressLayout | None") -> None:
...
 
-        from apps.base import lock_device
-        from apps.management.pm.suspend import suspend_device
+        will_close_workflow = config.has_pin() and workflow.autolock_interrupts_workflow
+        lock_device_if_unlocked()
 
-        if config.has_pin() and config.is_unlocked():
-            lock_device(interrupt_workflow=True)
+        if will_close_workflow:
+            # prevent further layout interaction
```
matejcik: this is somewhat weird: you'll go to Suspend with the layout still running, and after wakeup you resume here and raise `Shutdown`.

for that matter: you lock & suspend even if `workflow.autolock_interrupts_workflow is False`, means you _do_ in fact interrupt a long-running operation before hitting this.

not sure what the correct resolution is here though. should uninterruptible workflows continue running in the background?
should `lock_device_if_unlocked()` be NoReturn??

in general it feels like this should not be part of the UI event handling, but instead run a separate handler somewhere in the background.
rough sketch: a "suspended homescreen" is installed as the default workflow, and pressing the power button will (1) in any case turn off the display and (2) `if autolock_interrupts_workflow is True: workflow.close_others()`, causing the device to go to "suspend homescreen", and only _that_ will actually invoke the suspend, and then handle the resumption

matejcik: actually this solution meshes well with how "locked homescreen" works on non-PM models

core/src/apps/base.py
obrusvit: https://github.com/trezor/trezor-firmware/pull/5767/commits/7faf51f9e336246557ae312e17faf458738297a5

forgotten, thanks

core/src/boot.py
obrusvit: https://github.com/trezor/trezor-firmware/pull/5767/commits/f7aee752a7d1555c668252d2635d2b212c8965dc

core/src/trezor/ui/__init__.py
obrusvit: The desired behavior is that auto-suspend also works in the bootscreen where you're welcomed with "Tap to start" lockscreen (if PIN is configured). That's why `IdleTimer` is utilized to register `autodim_display` and `lock_device_if_unlocked_on_battery`

core/src/trezor/ui/__init__.py
obrusvit: The current implementation behaves as follows: 
- if you don't have the PIN set and you press the power button, the device only suspends and after waking up (e.g. by a repeated press of the power button), you are placed exactly where you were. This is for example during creating of a wallet. Then the user can continue if e.g. their device was suspended by auto-suspend. Note that this is not the case when the setup is done on USB -> in that case, the flow would be interrupted.
- if you have the PIN set and you press the power button, the workflow will be closed and the subsequent call of the `_event` would crash because the `Layout` is already gone.

Once we call `pm_suspend()` the execution stops there and we wait for the wakeup_flag

STew790:
**QA NOK**
If the device goes to suspend, Suite can't wake it up.

**Info:**
 - Suite version: desktop 25.10.0 (1160cfd53e2405511eea0ac122886d0bc19cf3ce)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.10.0 Chrome/140.0.7339.41 Electron/38.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1800x1169
 - Device: T3W1 938939a
 - Transport: BluetoothTransport, BridgeTransport 3.1.0

PR #5768, opened 2025-09-15
add BLE battery service

TychoVrahe:
Just level is propagated for now.


kopecdav (approved):
Looks good, propagating also the bls_batter_charge_state seems like an easy step forward which we can add later.

PR #5771, opened 2025-09-15
Fuel gauge driven battery_critical flag

kopecdav:
This PR fixes the bug in power manager where we forget to recover the battery critical_flag from the backup RAM during `pm_init()` and then it changes the way how the battery is being recovered from UVLO. 

Reason for this is ~ it showed up, that UVLO battery of turned off device will relax from 3.0V to around 3.1-3.15V based on the battery temperature and thus choosing right threshold to recover the battery from UVLO was almost impossible. At the same time, fuel gauge seems to be mature enough to takeover the responsibility.

So now, device will turn off when it reaches the UVLO voltage threshold of 3.0V and it also set the battery critical flag. The battery critical flag is then reset if the device is charged to at least 2% or if the USB power is connected. until then, the device will reject to turn ON.

### DIFF
--- a/core/embed/sys/power_manager/stm32u5/power_manager.c
+++ b/core/embed/sys/power_manager/stm32u5/power_manager.c
@@ -451,12 +451,7 @@ pm_status_t pm_store_data_to_backup_ram() {
 
   pm_recovery_data_t recovery = {.version = PM_RECOVERY_DATA_VERSION};
 
-  // Fuel gauge state
-  if (drv->battery_critical) {
-    recovery.soc = 0;
-  } else {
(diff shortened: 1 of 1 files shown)


kopecdav:
Added one more fix https://github.com/trezor/trezor-firmware/pull/5771/commits/d4a116703ed8ecf22c765a91a618c926ff43178e which removes the forced store of zero SoC to backup RAM in case the critical battery flag was on.

This caused that if you charged device on the wireless charger, had less then 2% and tried to turn on. Device rejected it, but also set SOC back to 0 .. so inpatient users, constantly smashing the power button would be waiting forever.

TychoVrahe (approved):

PR #5774, opened 2025-09-15
feat(core): introduce lock_manager

matejcik:
it eats some functions from apps.base, and introduces an asynchronous notification mechanism for suspending via `notify_suspend()`

TODO:
* [x] `suspend_homescreen` to defer suspending after a workflow is finished -> in such case, power button will just be unresponsive
* [x] darkening the screen immediately in response


matejcik:
still not there (maybe unwanted?):
shorter timeout if you wake by button but don't do anything. 5secs or so, we talked about this a couple months ago

obrusvit (approved):
Tested ACK

improves https://github.com/trezor/trezor-firmware/pull/5767

PR #5775, opened 2025-09-15
check nRF FW authenticity in firmware

TychoVrahe:


core/embed/io/nrf/stm32u5/nrf.c
```diff
@@ -551,6 +551,15 @@ bool nrf_authenticate(void) {
     return false;
   }
 
+  uint32_t timeout = ticks_timeout(5000);
+
+  // check that nRF communication is running prior to auth check
+  while (!ticks_expired(timeout)) {
```
kopecdav, line 557: Do you want to continue if the timeout expires, or rather return false here?

core/site_scons/models/T3W1/trezor_t3w1_revC.py
```diff
@@ -137,6 +137,7 @@ def configure(
             "vendor/stm32u5xx_hal_driver/Src/stm32u5xx_hal_uart.c",
             "vendor/stm32u5xx_hal_driver/Src/stm32u5xx_hal_uart_ex.c",
         ]
+    defines += [("USE_NRF_AUTH", "1")]
```
kopecdav: This should be probably conditioned by: 

`if "nrf_auth" in features_wanted:
     defines += [("USE_NRF_AUTH", "1")]
`

 This comment applies for all three revisions.

core/embed/sec/secret/inc/sec/secret_keys.h
```diff
@@ -56,7 +56,7 @@ secbool secret_key_tropic_masking(uint8_t dest[ECDSA_PRIVATE_KEY_SIZE]);
 
 #endif  // USE_TROPIC
 
-#ifdef USE_NRF
+#ifdef USE_NRF_AUTH
```
kopecdav, line 59: There is a `secret_key_nrf_pairing()` function call in `nrf_test.c` file which should be probably conditioned by #ifdef USE_NRF_AUTH as well. 

```
bool nrf_test_pair(void) {
  nrf_register_listener(NRF_SERVICE_PRODTEST, nrf_test_cb);

  g_nrf_test.success = false;
  g_nrf_test.failure = false;

  uint8_t data[NRF_PAIRING_SECRET_SIZE + 1] = {PRODTEST_CMD_PAIR};

  if (sectrue != secret_key_nrf_pairing(&data[1])) {           <--------------------------- HERE
    return false;
  }
```

kopecdav (changes requested):

core/embed/io/nrf/stm32u5/nrf.c
TychoVrahe: let continue for now, shouldn't be an issue

core/site_scons/models/T3W1/trezor_t3w1_revC.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/90b7990e4d15ca7d3a6531fbc0ada3cacdf9fbf1

core/embed/sec/secret/inc/sec/secret_keys.h
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/90b7990e4d15ca7d3a6531fbc0ada3cacdf9fbf1

Issue #4939, opened 2025-04-22
T3T1 Backlight driver buzzing


kopecdav:
T3T1 backlight driver TPS61043 generate a noise in audible spectrum. 

Single device is barely audible, but multiple devices makes quite annoying and loud sound. 

The noise is clearly generated by backlight driver which is driven by PWM at same frequency. 

![Image](https://github.com/user-attachments/assets/8f52ee5a-de47-47ec-a444-eb3ccecb8b87)

PR #5776, opened 2025-09-15
T3T1 Backlight driver buzzing fix

bleska:
This PR resolves ISSUE https://github.com/trezor/trezor-firmware/issues/4939 .

The purpose is to increase the PWM frequency from 12.5 kHz to 25 kHz as it's above the human audible range. It's compliant to the TPS61043 DS which allows PWM frequency range up to 50 kHz.


bleska:
We need to check what situations can occur in a device e.g. old boot loader and new firmware. Since the new FW shall have higher PWM frequency (25 kHz) than the old boot loader (12.5 kHz), how to safely reconfigure the PWM TIM so that no backlight glitch happens?

UPDATE: 17.9.2025
It doesn't seem to be an issue because inside "void backlight_init(backlight_action_t action) " function in "tps61043.c" file, the respective timer peripheral undergoes full re-initialization including its reset and clock enable. Therefore, there is always a discontinuity at PWM signal, no matter of the frequency change.

PR #5779, opened 2025-09-15
stop charging indication when battery is fully charged

TychoVrahe:
In off mode.

### DIFF
--- a/core/embed/projects/bootloader/main.c
+++ b/core/embed/projects/bootloader/main.c
@@ -236,21 +236,22 @@ static secbool boot_sequence(void) {
     pm_state_t state;
     pm_get_state(&state);
 
-    if (state.charging_status == PM_BATTERY_CHARGING) {
-      // charing screen
+    if (pm_is_charging()) {
+      // charging indication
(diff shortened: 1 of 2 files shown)


core/embed/projects/bootloader/main.c
```diff
@@ -236,21 +236,20 @@ static secbool boot_sequence(void) {
...
-      // charing screen
+    if (pm_is_charging()) {
+      // charging indication
 #ifdef USE_RGB_LED
       if (!rgb_led_effect_ongoing()) {
         rgb_led_effect_start(RGB_LED_EFFECT_CHARGING, 0);
       }
 #endif
     } else {
+      rgb_led_set_color(0);
```
kopecdav, line 248: This should be surrounded by #ifdef USE_RGB_LED

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/fbfefb6accc532a3d90e00aae71295049a7198e5

kopecdav (approved):

PR #5782, opened 2025-09-16
Differentiate wipe code from pin set up

ibz:
The immediate motive behind this is to have a separate copy when canceling wipe code set up in Eckhart, but after these changes we can probably stop passing a bunch of strings around and have the layout decide which strings to use depending on whether a pin code or a wipe code is being set.

Also, having the functions/flows named `_pin` when they were dealing with pin *or* wipe code was confusing to start with.


core/embed/rust/src/ui/api/firmware_micropython.rs
```diff
@@ -1688,13 +1689,14 @@ pub static mp_module_trezorui_api: Module = obj_module! {
 
     // TODO: supply more arguments for Wipe code setting (delizia)
     ///
-    /// def flow_confirm_set_new_pin(
+    /// def flow_confirm_set_new_code(
```
obrusvit, line 1692: TBH I don't like the new name either. How about longer but more descriptive `flow_confirm_set_new_pin_or_wipecode`

core/embed/rust/src/ui/api/firmware_micropython.rs
```diff
@@ -1688,13 +1689,14 @@ pub static mp_module_trezorui_api: Module = obj_module! {
 
     // TODO: supply more arguments for Wipe code setting (delizia)
```
obrusvit: Seems like this `TODO:` is now obsolete

core/embed/rust/src/ui/api/firmware_micropython.rs
obrusvit: You know what, it's probably fine 🤷‍♂️

obrusvit (commented):
1 comment

core/embed/rust/src/ui/api/firmware_micropython.rs
ibz: And there was another one too! :) 84c7fc3c66e21661dfb2994cd01579320578db96

core/translations/order.json
obrusvit, line 1: What happened here? why is the order different? 🤔

ibz: Shouldn't matter... or?

ibz: Fixed anyway...

obrusvit (approved):
Thanks. Please rebase

PR #5787, opened 2025-09-17
chore(core): additional Crowdin sync for Eckhart

obrusvit:
[no changelog]


obrusvit:
Force pushed a formatting mistake in JSON which caused a havoc

PR #5790, opened 2025-09-17
Don't request marquee animation frames if text fits

ibz:
Optimization: when a marquee is not needed due to the text fitting we should not request animation frames!

### DIFF
--- a/core/embed/rust/src/ui/layout_caesar/component/title.rs
+++ b/core/embed/rust/src/ui/layout_caesar/component/title.rs
@@ -15,7 +15,6 @@ pub struct Title {
     area: Rect,
     title: TString<'static>,
     marquee: Marquee,
-    needs_marquee: bool,
     centered: bool,
 }
 
@@ -24,7 +23,6 @@ impl Title {
         Self {
             title,
             marquee: Marquee::new(title, theme::FONT_HEADER, theme::FG, theme::BG),
-            needs_marquee: false,
             area: Rect::zero(),
             centered: false,
         }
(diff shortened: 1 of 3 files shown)


ibz:
> Why not making it the property of the `Marquee` instead? Something like this should do, can you test? [4644490](https://github.com/trezor/trezor-firmware/commit/464449092f085e9d33d385a90afaf16b46b4a50b)

I actually tried something like that, but then decided to optimize even further and avoid the width calculation in every event. Probably not worth the more complicated code though, since width calculation is quite cheap....

ibz:
> Why not making it the property of the `Marquee` instead? Something like this should do, can you test? [4644490](https://github.com/trezor/trezor-firmware/commit/464449092f085e9d33d385a90afaf16b46b4a50b)

Was trying to avoid the text width calculation on render, but that's a cheap operation and not worth the extra complexity in code. This works and also simplified TS3 code! 946d4d2533ea175bc4781f451bf3c7f885a7b72e

core/embed/rust/src/ui/layout_caesar/component/title.rs
```diff
@@ -91,24 +84,20 @@ impl Component for Title {
...
+            self.marquee.start(ctx, Instant::now());
+        } else {
+            return self.marquee.event(ctx, event);
         }
         None
     }
 
     fn render<'s>(&'s self, target: &mut impl Renderer<'s>) {
-        if self.needs_marquee {
+        if self.title.map(|s| theme::FONT_HEADER.text_width(s)) > self.area.width() {
```
obrusvit, line 100: Yeah, I think we can parametrize the `Marquee` to render the text with `Alignment::Center`. Not blocking this PR on this.

PR #5793, opened 2025-09-17
Remove useless change_code params

ibz:
Following #5782 some parameters became useless...

Also added a test that cancels the wipe code setup on all models to exercise the new code path.

### DIFF
--- a/tests/input_flows.py
+++ b/tests/input_flows.py
@@ -91,6 +91,43 @@ class InputFlowBase:
         return self.debug.read_layout().title()
 
 
+class InputFlowNewCodeCancel(InputFlowBase):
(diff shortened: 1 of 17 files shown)


tests/input_flows.py
```diff
@@ -91,6 +91,43 @@ def title(self) -> str:
         return self.debug.read_layout().title()
 
 
+class InputFlowNewCodeCancel(InputFlowBase):
```
bieleluk: nit: The flow is specific for the wipe code, so the flow name should reflect that.

bieleluk (approved):
LGTM with one minor suggestion

PR #5798, opened 2025-09-17
THP: make app_name required & send it from trezorlib

mmilata:
Related to #5540. I think Suite has been [sending it](https://github.com/trezor/trezor-suite/blob/de01a0302e5b8f4f6edfe1ea7f36861cf3e6e2e7/packages/connect/src/device/thp/pairing.ts#L264) for some time but we should probably test it first.

### DIFF
--- a/core/src/apps/thp/pairing.py
+++ b/core/src/apps/thp/pairing.py
@@ -111,9 +111,10 @@ async def handle_pairing_request(
     if not ThpPairingRequest.is_type_of(message):
         raise UnexpectedMessage("Unexpected message")
 
-    # TODO: make app_name required eventually
     if not message.host_name:
         raise DataError("Missing host_name.")
+    if not message.app_name:
+        raise DataError("Missing app_name.")
 
     peer_addr = ctx.channel_ctx.iface_ctx.connected_addr()
     await ui.show_pairing_dialog(message.host_name, message.app_name)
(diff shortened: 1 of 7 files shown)


vdovhanych (approved):

romanz (approved):
Thanks - please also update the UI fixtures.

core/src/apps/thp/pairing.py
```diff
@@ -111,9 +111,10 @@ async def handle_pairing_request(
     if not ThpPairingRequest.is_type_of(message):
         raise UnexpectedMessage("Unexpected message")
 
-    # TODO: make app_name required eventually
     if not message.host_name:
         raise DataError("Missing host_name.")
+    if not message.app_name:
```
matejcik, line 116: both should be required at protobuf level

mmilata: d90503e1876b97722303a20609da50446ff26753

Leaving the check in `pairing.py` to make sure they are non-empty.

M1nd3r:
I see an issue. This change might result in failing UI tests when run locally, as the device name is taken from the "real device". Example: `ci-runner` vs `petrs-T16`.


It is also possible that this issue is there already since the change of the screens 🤔 

IMO we can postpone it for now and solve it separately later.

Screenshot from 2025-09-18 15-31-21

mmilata:
Good catch, thanks. I think I'll add optional `host_name` and `app_name` parameters somewhere to the trezorlib API and make tests use it instead of environment variable.

PR #5808, opened 2025-09-18
feat(core): use tropic chip entropy in RNG

cepetr:
This PR refactors the secmon/kernel API for random number generation.

- `rng_get()` and `optiga_random_buffer()` were removed from syscalls/smcalls, and two more generic functions were introduced: `rng_fill_buffer()` and `rng_fill_buffer_strong()`.

- `rng_fill_buffer_strong()` now uses the Tropic random generator in addition to the Optiga random generator when available.

- there are no longer two random generators in the emulator (`rng_get()` and `random32()` are now equivalent).

- this PR also enables the use of the MCU TRNG in the prodtest firmware (previously, it relied only on the OPTIGA TRNG).

There is still a `random_buffer()` in the crypto library that does the same thing as `rng_fill_buffer()`. We should reconsider it in the future and maybe remove the rand module from the crypto library entirely.


core/embed/sec/rng/inc/sec/rng.h
```diff
@@ -17,19 +17,54 @@
...
+ *
+ * The function requires that Optiga and/or Tropic to be initialized
+ * if they are enabled by USE_OPTIGA/USE_TROPIC.
+ *
+ * @param buffer Buffer to fill with random bytes.
+ * @param buffer_size Size of the buffer in bytes.
+ *
+ * @return True on success, false on failure.
+ */
+bool rng_fill_buffer_strong(void* buffer, size_t buffer_size);
```
andrewkozlik: Let's add the `__wur` attribute to this function (warn_unused_result).
```suggestion
bool __wur rng_fill_buffer_strong(void* buffer, size_t buffer_size);
```

core/embed/sec/tropic/tropic.c
```diff
@@ -80,13 +88,27 @@ bool tropic_init(void) {
...
   curve25519_key tropic_pubkey = {0};
   secbool pubkey_ok = secret_key_tropic_public(tropic_pubkey);
+
+#if !PRODUCTION
+  // Allow running with default factory keys in non-production fw
+  if (pubkey_ok != sectrue || privkey_ok != sectrue) {
+    static const curve25519_key factory_private = {
+        0x28, 0x3f, 0x5a, 0x0f, 0xfc, 0x41, 0xcf, 0x50, 0x98, 0xa8, 0xe1,
+        0x7d, 0xb6, 0x37, 0x2c, 0x3c, 0xaa, 0xd1, 0xee, 0xee, 0xdf, 0x0f,
+        0x75, 0xbc, 0x3f, 0xbf, 0xcd, 0x9c, 0xab, 0x3d, 0xe9, 0x72};
```
andrewkozlik: This key already appears in prodtest. It would be good to have these keys defined in one place. Also, there is another factory key which is used in some older Tropic01 models that is used when `TROPIC_TESTING_KEYS` is set, so that can be handled in that one place.

core/embed/sec/tropic/tropic.c
```diff
@@ -184,6 +206,52 @@ bool tropic_data_read(uint16_t udata_slot, uint8_t *data, uint16_t *size) {
   return res == LT_OK;
 }
 
+#if !PRODUCTION
+static bool tropic_get_tropic_pubkey(curve25519_key pubkey) {
+  tropic_driver_t *drv = &g_tropic_driver;
+
+  if (drv->initialized) {
+    return true;
+  }
```
andrewkozlik: Shouldn't this be like this?
```suggestion
  if (!drv->initialized) {
    return false;
  }
```

core/embed/sec/tropic/tropic.c
```diff
@@ -80,13 +88,27 @@ bool tropic_init(void) {
 
   curve25519_key tropic_pubkey = {0};
   secbool pubkey_ok = secret_key_tropic_public(tropic_pubkey);
+
+#if !PRODUCTION
+  // Allow running with default factory keys in non-production fw
+  if (pubkey_ok != sectrue || privkey_ok != sectrue) {
```
andrewkozlik: I am wondering whether this `if` statement shouldn't be split into two `if` statements. Logically, the fact that the pubkey isn't written, doesn't necessarily imply that we should use the factory pairing key. It is very likely that the two conditions are either both false or both true, because the device is either a provisioned device or an unprovisioned development device. But theoretically, we could encounter a semi-provisoned device, especially among development devices.

core/embed/sec/tropic/tropic.c
cepetr: Yes, of course, fixed fc55c5b55dc983181ac6875d4895e03c254e3dc1

core/embed/sec/tropic/tropic.c
cepetr: split, see ad50b08ce8480ce63f11bdcc510e3c6bfcc0b379

core/embed/sec/rng/inc/sec/rng.h
cepetr: I agree, added, see 6eb8bb870c65932603a7aa0fa705656be2838d52

core/embed/sec/rng/rng_common.c
```diff
@@ -0,0 +1,110 @@
...
+
+#ifndef USE_INSECURE_PRNG
+// Re-implementation of random32() function declared in crypto/rand.h
+// to use MCU TRNG instead of crypto library PRNG.
+uint32_t random32(void) { return rng_get(); }
+#endif
+
+// Re-implementation of weak random_buffer() function defined in crypto/rand.c
+// to be the same as rng_fill_buffer() function.
+void random_buffer(uint8_t* buf, size_t len) { rng_fill_buffer(buf, len); }
```
andrewkozlik: I am feeling a bit uneasy about these functions. It's hard to see where the randomness is coming from, which makes it prone to bugs. We need to be very careful around random number generations, because bugs are so hard to uncover. I don't have a solution in mind yet, because I still haven't wrapped my head around this, but here are some comments:
- Do we really need `rng_get()` in insecure mode? That should be a very internal function and everybody should be using `rng_fill_buffer()`. Right now it looks a lot like a circular definition. (I realize it's not.)
- The duplicity of `random_buffer()` and `rng_fill_buffer()` in our codebase is not good. It should be the same thing, but is it?
- Similarly, the coexistence of `random32()` and `rng_get()` is not good. Here we have `random32(void) { return rng_get();}` if not using the insecure RNG and in rng.c we have `rng_get(void) {return random32();` if using the insecure RNG. Again looks like a circular definition, even though it's not.
- I think as a rule of thumb, we should aim to minimize the number of instances of `USE_INSECURE_PRNG` in the code. Ideally there should be just one point where we reference it.

core/embed/sec/tropic/tropic.c
cepetr: how about this  3cf86214e58d864747d2818006a8c30647828d31?

core/embed/sec/rng/rng_common.c
cepetr: Yes, I was aware of all the things you mentioned. I didn't want to modify the crypto lib for now, so the solution is somewhere halfway.

Definitely we should have one source of random numbers. I think it should be the `sec/rng` module rather than `crypto/rand`. Right now it's mixed for insecure/secure modes, which makes the code complicated.

Some ideas:

1) We could implement a PRNG in `sec/rng` and remove it from `crypto/rand` - so both secure/insecure mode random  numbers comes from a single place.

2) We should not have two sets of random functions. It's really confusing:
- We can simply get rid of `random32()` - it's used only in one place (ecdsa.c) and we can replace it with a call to random_buffer().
- I have no idea how to get rid of `random_buffer()` for now since it's extensively used in the crypto lib. Maybe we can leave it where it is and keeps its implementation on crypto lib users.

3) I agree `rng_get()` should probably be internal to `sec/rng`. Right now it's designed so it can be used anywhere as a helper that may simplify some types of code. But we might remove it and use `rng_fill_buffer(&var, sizeof(var))` instead.

it would be better to discuss this in person.

andrewkozlik: Some more concrete thoughts:
  - Let's replace `random_buffer()` with `rng_fill_buffer()` wherever possible. At least everywhere in `core`, but ideally throughout our code base.
  - We should have three elementary functions:
    1. Get random uint32 from the MCU's TRNG.
    2. Get pseudo-random uint32 from the insecure RNG.
    3. Get random uint32 from 1 or 2, depending on `USE_INSECURE_PRNG`.

    Right now we have `random32()` which takes on the roles of functions (ii) and (iii), and we have `rng_get()` which takes on the roles of function (i) and (iii). How about (i) will be called `trng_get()`, (ii) will be called `prng_get()` and (iii) will be called either `rng_get()` or `random32()`? Then `rng_fill_buffer()` always uses function (iii).
  - It would be good to show at the very end of the build that the code was built with the insecure RNG. Not sure how best to do that. We can "just" detect `USE_INSECURE_PRNG` in the build script, but ideally the warning should reflect that the pseudo-random generator function was linked into the binary.

TychoVrahe:
Somewhat related. We discussed some time ago possibility of including nRF RNG into randomness generation. Is that still on the table? We don't have nRF driver in secmon so it might be a bit cumbersome to implement

andrewkozlik:
> including nRF RNG into randomness generation

That would certainly be a nice addition.

core/embed/sec/rng/rng_common.c
cepetr: I tried to improve it in 054858e1b3518d6a130dd699c9dcccf7aeacb498. I am not entirely satisfied with the changes, but I beleive they move the project forward.

- The code still a little spagetti-like because of having both `random_buffer()` and `rng_fill_buffer()` exist. The only real fix I see is consolidating then into a single function.

- The key update is that now `rng_fill_buffer()` has become the main source of randomness.  `rng_get()`, `random_buffer()` and `random32()` functions now relay on it. _(But in the emulator, when using insecure PRNG, `rng_fill_buffer()` is calling `radom_buffer()` and that might be confusing.)_

- Insecure random generator is now placed in rand_insecure.c. This way, if the  file isn't compiled/linked, insecure rnadom numbers won't appear in the firmware.  I think it's a bit safer then marking random_buffer() as weak.

- I decided to not rename `random_buffer()` to `insecure_random_buffer()`, since that would mean updating all tests. Now just adding rand_insecure.c into test Makefiles is sufficient.

- Finally I replaced calls to `random_buffer()` with `rng_fill_buffer()` in the core folder.  However, random_buffer() is still invoked directly withing crypto lib.

storage/tests/c/Makefile
```diff
@@ -54,9 +55,17 @@ build/%.o: $(BASE)%.c $(BASE)%.h
 	mkdir -p $(@D)
 	$(CC) $(CFLAGS) -DFLASH_BIT_ACCESS -DFLASH_BLOCK_WORDS=1 $(INC) -c $< -o $@
 
+build/%.o: $(BASE)%.c
+	mkdir -p $(@D)
+	$(CC) $(CFLAGS) -DFLASH_BIT_ACCESS -DFLASH_BLOCK_WORDS=1 $(INC) -c $< -o $@
+
```
andrewkozlik, line 63: I don't understand this rule. It looks like a duplicate of the previous one, except with one missing prerequisite. Same below.

crypto/rand.h
```diff
@@ -27,9 +27,18 @@
...
+
 void random_buffer(uint8_t *buf, size_t len);
+
+static inline uint32_t random32(void) {
+  uint32_t r = 0;
+  random_buffer((uint8_t *)&r, sizeof(r));
+  return r;
+}
+
 void random_xor(uint8_t *buf, size_t len);
```
andrewkozlik: We should be able to get rid of `random_xor()` and replace it with `rng_fill_buffer_strong()`.

storage/tests/c/Makefile
cepetr: Yes, it's strange. The first rule applies if there is both a `.c` and a `.h` file, while the second applies if there is only a `.c` file (with no corresponding `.h`). Without this rule, `rand_insecure.c` is not compiled. I have no idea how to simplify it.

crypto/rand.h
cepetr: There are only two places where we use `random_xor()`:

- optiga.c
- storage/tests/c3/storage.c

IMHO, in neither case can we use `rng_fill_buffer_strong()`.

storage/tests/c/Makefile
andrewkozlik: We can simplify it a little bit by factoring out the recipe using https://www.gnu.org/software/make/manual/make.html#Canned-Recipes

andrewkozlik: However, it makes me wonder, how much it makes sense to include the header files in the dependencies.
```mk
build/%.o: $(BASE)%.c $(BASE)%.h
```
What could possibly change in the header file to warrant rebuilding the object file? Typically a `#define` macro or constant. But if that happens, then we want to rebuild all files which include the header, not just the one which shares its stem name.

core/embed/sec/rng/stm32/rng.c
```diff
@@ -47,17 +50,36 @@ uint32_t rng_read(const uint32_t previous, const uint32_t compare_previous) {
...
+
+  if (remaining > 0) {
+    uint32_t r = rng_get_u32();
+    memcpy(dst, &r, remaining);
+  }
+}
+
 #endif  // SECURE_MODE
+
+// Implements random_buffer() function defined in crypto/rand.c
```
andrewkozlik:
```suggestion
// Implements random_buffer() function declared in crypto/rand.h
```

core/embed/sec/rng/stm32/rng.c
```diff
@@ -47,17 +50,36 @@ uint32_t rng_read(const uint32_t previous, const uint32_t compare_previous) {
...
   } else {
     previous = current;
   }
-  current = rng_read(previous, 1);
+  current = rng_read_u32(previous, 1);
   return current;
 }
 
+void rng_fill_buffer(void* buffer, size_t buffer_size) {
+  uint32_t* dst = (uint32_t*)buffer;
```
andrewkozlik, line 67: Beware that unaligned access can lead to some unexpected behavior. It's probably supported on our processors, I am not sure though if and under what conditions. On processors where this leads to unexpected behavior, it may be problematic to detect, since we have no expectations of the random numbers. Is there any static assert that we could use to check that unaligned access is OK to do?

core/embed/sec/rng/rng_common.c
andrewkozlik: > But in the emulator, when using insecure PRNG, `rng_fill_buffer()` is calling `random_buffer()` and that might be confusing.)

I think this is acceptable, since the comments explain it well enough. The changes in this PR are a huge step forward in clarifying the behavior.

Nevertheless, how about implementing the insecure random buffer in `insecure_random_buffer()`, exposing that function in the header file and then in the emulator, when using the insecure PRNG, `rng_fill_buffer()` will explicitly call `insecure_random_buffer()`, and `random_buffer()` can be defined like it is in rand_insecure.c except that it will just be calling `insecure_random_buffer()`. Functionally it's of course all the same. I am just trying to maximize clarity of what is happening and as I stated the present implementation is fine, so judge for yourself.

core/embed/sec/rng/stm32/rng.c
andrewkozlik: However, IIRC we might already be using unaligned access somewhere else. So it's no big deal. Just feels a little scarier with the RNG and difficulty of detecting unexpected behavior. Feel free to ignore this comment if you are confident it's not a problem for us.

crypto/rand.h
cepetr: Finally, I fully eliminated `random_xor()` from crypto/rand.

I replaced `random_xor()` with `rng_fill_buffer_strong()` in optiga.c. As we discussed in person, there is now a hidden dependency of Optiga on Tropic, but this is acceptable because the PIN function is called only after both the Tropic and Optiga drivers are fully initialized.

In c3/storage.c, `random_xor()` remains behind `#ifdef USE_OPTIGA` and is not compiled, so I left it there.

see 6e508223e1f94f1828924c3d05b37eff94f01baf

storage/tests/c/Makefile
cepetr: I've commented this, see 82c2bbf1088bf68c3d34e9c7c86c5441834af646

core/embed/sec/rng/stm32/rng.c
cepetr: fixed, see dea83225582e0d6d17bd8a40b8bb5db54de1f7c6

core/embed/sec/rng/stm32/rng.c
cepetr: Yes, we previously agreed to use unaligned accesses since Cortex-M permits them.

core/embed/sec/rng/stm32/rng.c
cepetr: and finally 09a542ff3ce7b89a84c59f115a19da61611e9b8b

PR #5809, opened 2025-09-19
allow unit variant rework

TychoVrahe:


core/embed/projects/prodtest/cmd/prodtest_otp_variant.c
```diff
@@ -156,7 +172,35 @@ static void prodtest_otp_variant_write(cli_t* cli) {
   }
 #endif
 
-  if (sectrue == flash_otp_is_locked(FLASH_OTP_BLOCK_DEVICE_VARIANT)) {
+  uint8_t block_num = FLASH_OTP_BLOCK_DEVICE_VARIANT;
+
+  if (rework) {
+    if (sectrue == flash_otp_is_locked(block_num)) {
+      cli_error(cli, CLI_ERROR_LOCKED,
+                "OTP rework block is locked and cannot be written again.");
```
kopecdav, line 182: According to error message it looks like you tried to unlock the rework block, but `block_num` variable still contains `FLASH_OTP_BLOCK_DEVICE_VARIANT` instead.

TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/0f102424457a25686a40fe046c75334c1817c8b0

core/embed/sys/smcall/stm32/smcall_dispatch.c
```diff
@@ -18,6 +18,8 @@
  */
 
 #ifdef SECMON
+#include <stddef.h>
```
cepetr: Aren't these headers already included with <trezor_rtl.h>?

core/embed/sys/smcall/stm32/smcall_verifiers.c
```diff
@@ -29,6 +29,8 @@
 #include "smcall_probe.h"
 #include "smcall_verifiers.h"
 
+#include "util/unit_properties.h"
```
cepetr: This include should be incorporated in `"smcall_verifiers.h"`

core/embed/util/unit_properties/inc/util/unit_properties.h
```diff
@@ -25,47 +25,91 @@
...
+/**
+ * @brief Gets the device serial number
+ *
+ * @param device_sn Buffer to store the device serial number
+ * @param max_device_sn_size Maximum size of the device_sn buffer
+ * @param device_sn_size Pointer to store the actual size of the serial number
+ * @return true if the serial number was successfully retrieved
+ * @return false otherwise
+ */
+bool unit_properties_get_sn(uint8_t* device_sn, size_t max_device_sn_size,
```
cepetr, line 114: Do we really need a new function/syscall/smcall for this? Wouldn't it be better to put the serial number into the `unit_properties_t` structure - since it's just a few bytes and would also benefit from caching?

cepetr (changes requested):

core/embed/sys/smcall/stm32/smcall_verifiers.c
TychoVrahe: fixed http://github.com/trezor/trezor-firmware/commit/c819c4b4bacfdb14cdf182bb490c32081f9d3e74

core/embed/sys/smcall/stm32/smcall_dispatch.c
TychoVrahe: fixed http://github.com/trezor/trezor-firmware/commit/c819c4b4bacfdb14cdf182bb490c32081f9d3e74

core/embed/util/unit_properties/inc/util/unit_properties.h
TychoVrahe: I agree in principle, however i decided against this at this moment - the function is used in provisioning code which i don't want to touch this close release, especially because my ability to test it is limited.

I created an issue and lets do this when there is more time to test: https://github.com/trezor/trezor-firmware/issues/5915

cepetr (approved):
utACK

PR #5810, opened 2025-09-19
Simplify marquee use in device menu

ibz:
* menu subtext cannot overflow - either it fits, or it becomes a marquee
* main text single / double line independent of subtext being a marquee


core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -157,8 +157,8 @@ impl<const N: usize> VecExt for Vec<MenuItem, N> {
 
 struct MenuItem {
     text: TString<'static>,
+    single_line: bool,
```
bieleluk, line 160: nit: `single_line_text`

PR #5814, opened 2025-09-19
add T3W1 signed production binaries

TychoVrahe:
bootloader & secmon


TychoVrahe:
closing: bootloader is obsolete and secmon no longer compatible, lets put them after next signing

PR #5783, opened 2025-09-16
feat(core/eckhart): add host info screen to device menu

bieleluk:
- add host info screen for each paired device according to [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=9814-10168&t=O7WnxVj0YCDeyoeN-0)
- add the THP `app_name` ~to the cache and use it~ in the connection main text if present according to [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=12331-41764&t=yj3Wzd2q2Pm7mznN-0)

PR #5817, opened 2025-09-19
feat(core/eckhart): cache THP app/host name along with MAC

mmilata:
Part of #5783 that contains incompatible storage changes and should make it to the freeze.

### DIFF
--- a/core/src/apps/homescreen/device_menu.py
+++ b/core/src/apps/homescreen/device_menu.py
@@ -1,4 +1,5 @@
 from micropython import const
+from typing import TYPE_CHECKING
 
 import storage.device as storage_device
 import trezorble as ble
@@ -7,6 +8,9 @@ from trezor import TR, config, log, utils
 from trezor.ui.layouts import interact, raise_if_cancelled
 from trezor.wire import ActionCancelled, PinCancelled
 from trezorui_api import CANCELLED, DeviceMenuResult
+
+if TYPE_CHECKING:
+    from trezor.messages import ThpPairedCacheEntry
 
 BLE_MAX_BONDS = 8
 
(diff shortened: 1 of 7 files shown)


core/src/apps/homescreen/device_menu.py
```diff
@@ -24,11 +24,21 @@ class SubmenuId:
...
 
-def _get_hostname(ble_addr: bytes, hostname_map: dict[bytes, str]) -> str:
-    if (hostname := hostname_map.get(ble_addr)) is None:
+def _get_hostinfo(
+    ble_addr: bytes, hostname_map: dict[bytes, tuple[str, str]]
+) -> tuple[str, tuple[str, str] | None]:
+    mac = ":".join(f"{byte:02X}" for byte in reversed(ble_addr))
+
+    hostinfo = hostname_map.get(ble_addr)
+    if hostinfo is None:
```
romanz: Do we need the `if` here?
(it seems that returning `(mac, hostinfo)` is enough...)

core/src/apps/homescreen/device_menu.py
```diff
@@ -24,11 +24,21 @@ class SubmenuId:
...
         # Internal MAC address representation is using reversed byte order.
```
romanz: nit: move the comment to the mac formatting line above

core/src/apps/homescreen/device_menu.py
```diff
@@ -77,7 +87,9 @@ async def handle_device_menu() -> None:
         connected_idx = _find_device(connected_addr, bonds)
         if __debug__:
             log.debug(__name__, "connected: %s (%s)", connected_addr, connected_idx)
-        hostname_map = {e.mac_addr: e.host_name for e in paired_cache.load()}
+        hostname_map = {
+            e.mac_addr: (e.host_name, e.app_name) for e in paired_cache.load()
```
romanz: Since we need both `host_name` and `app_name`, it would be probably simpler to keep the following map:

```python
entry_map = {e.mac_addr: e for e in paired_cache.load()}
```

WDYT?

core/src/apps/homescreen/device_menu.py
mmilata: nope, fixed in c1d3e195c9889027790178a87dde8e3448164e75

core/src/apps/homescreen/device_menu.py
mmilata: c1d3e195c9889027790178a87dde8e3448164e75

core/src/apps/homescreen/device_menu.py
mmilata: c1d3e195c9889027790178a87dde8e3448164e75

romanz (approved):

PR #5832, opened 2025-09-22
enable controlling TX power level from main MCU

TychoVrahe:
[no changelog]


TychoVrahe:
introduced power level enum and adjusted tx power for connection in https://github.com/trezor/trezor-firmware/commit/11bac081a673bac04369814a3f8132b4419e0da1

PR #5834, opened 2025-09-22
fix(core): adapt THP retransmission timeouts using recent ACK latency

romanz:
Otherwise, the default device retransmissions are expected to timeout when BLE connection experiences high latency.

Let's use THP ACKs to estimate the channel latency.
This estimate will be used to increase the timeout for the next transmission, to avoid "overloading" high-latency BLE connection.

This PR shouldn't affect USB-based THP communication, since it usually has ~1 ms latency.


obrusvit:
@marekrjpolak did you have a chance to test this?

marekrjpolak:
@obrusvit I did, and while I have no clue whether it's the best we can do, it's much much better than before. :+1:

romanz:
Rebasing to resolve a conflict:

image

core/src/trezor/wire/thp/channel.py
```diff
@@ -441,15 +443,21 @@ async def write_encrypted_payload(self, ctrl_byte: int, payload: AnyBytes) -> No
...
+            timeout_ms = ack_latency_ms + round(10300 - 1010000 / (100 + i))
             try:
                 # wait and return after receiving an ACK, or raise in case of an unexpected message.
                 await self.recv_payload(expected_ctrl_byte=None, timeout_ms=timeout_ms)
             except Timeout:
                 if __debug__:
                     log.warning(__name__, "Retransmit after %d ms", timeout_ms)
                 continue
+
+            ack_latency_ms = utime.ticks_diff(utime.ticks_ms(), self.last_write_ms)
```
M1nd3r, line 456: EDIT: Resolved. Lessons learned - do not start your day without a cup of coffee ☕ 

If the call `self.recv_payload` fails on `Timeout`, there is a `continue` statement in the except branch. 

Thus, if there are multiple retransmissions of a single message, we adjust the value `ack_latency_ms` for each of them. We also write the value to the `cache` for each.

To me, it seems more reasonable to update it not on per-attempt basis, but per-message bases. We should calculate and update it only once at the end. It should not cause a problem as the "regular" timeout is increasing between two attempts from 200ms for the first attempt to the ~3.5 seconds for the 50th attempt.

If there is a significant latency, we can increase the value for "all" of the timeouts by the latency. 

Example:
- If the latency is 500 ms, first retransmission attempt should have 700 ms timeout, and the 50th attempt should have ~4 s timeout. 



Note: in case of multiple retransmission attempts, we max-out the `ack_latency_ms` on 800 ms after about three of them. (Another reason to **not** update it for every retransmission attempt).
```
ack_latency_ms = utime.ticks_diff(utime.ticks_ms(), self.last_write_ms)
```
It is like this because the `self.last_write_ms` is updated before the retransmission starts - and the sum of the first three retransmission timeouts is 898 ms.

core/src/trezor/wire/thp/channel.py
```diff
@@ -441,15 +443,21 @@ async def write_encrypted_payload(self, ctrl_byte: int, payload: AnyBytes) -> No
                     log.error(__name__, "Sending is stuck for %d ms", _WRITE_TIMEOUT_MS)
                 break
 
-            # starting from 200ms till ~3.52s
-            timeout_ms = round(10300 - 1010000 / (100 + i))
+            # Channel's estimated latency + a variable delay (from 200ms till ~3.52s)
+            timeout_ms = ack_latency_ms + round(10300 - 1010000 / (100 + i))
```
M1nd3r, line 447: I am thinking whether we should decrease/change the `timeout_ms` variable delay - `round(10300 - 1010000 / (100 + i))` - as the `ack_latency_ms` is being added. 🤷🏼 

For instance, we could do it like this: ```round(8000  - 790000 / (100 + i))```

```
 1: OLD (0.200 s,   0.200 s) | NEW (0.100 s,   0.100 s)
50: OLD (3.521 s, 103.792 s) | NEW (2.698 s,  78.361 s)
```
- It changes the default first delay to `100 ms`
- It decreases the 50th delay about `800 ms` - keeping the max interval at approx. 3.5 seconds when the `ack_latency_ms` is maxed-out.


We can tweak the variable delay series as we want - or we might keep it as it is now. I am not sure which option/value is better.

romanz: I would prefer to use the current variable delay series, so this PR will only **increase** the timeouts in case of an increased ACK latency (but not by too much), to prevent "overloading" the BLE host.

CC: @szymonlesisz WDYT?

M1nd3r: >  in case of an increased ACK latency

I just want to say that, IIUC, changes in this PR will result in increase even for low-ish latency. Previously, the first retransmission timeout was fixed to 200 ms.
After the change, even if the delay is only 50 ms (no retransmission is triggered), the `ack_timeout` is set to 50 ms -> resulting in the first retransmission having 250 ms (200 + 50 ms) timeout. 

Imagine a valid case - message is damaged or lost. If the latency is "only" 120 ms, the retransmission will take place after 320 ms instead of current 200 ms.

**To be clear** - I am not against increasing the value. I just want to emphasize that we had 100 ms before and the "first iteration" solution (https://github.com/trezor/trezor-firmware/pull/5807) was to increase the timeout by default to 200 ms. So this PR could either be a "second increase", or a dynamic replacement of the static "first solution".

M1nd3r: > l only increase the timeouts

Only a nit: the proposed change of variable timeout series does not decrease the  timeout value (except for the first message in the channel)

- In case of latency being >= 100 ms, the fist retransmission timeout will be set to >= 200 ms (100ms from the first variable timeout and >= 100 ms from the `ack_latency` value) - which is >= current value.
- If the latency is lower than 100 ms -> even by changing the series as proposed above, retransmission will not be triggered.

PR #5837, opened 2025-09-23
Add progress when requesting extra data

ibz:

### DIFF
--- a/core/src/apps/ethereum/sign_typed_data.py
+++ b/core/src/apps/ethereum/sign_typed_data.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable
 
 from trezor.enums import EthereumDataType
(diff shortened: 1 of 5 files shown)


core/src/apps/ethereum/sign_typed_data.py
```diff
@@ -46,13 +48,19 @@ async def sign_typed_data(
     # Display address so user can validate it
     await require_confirm_address(address_bytes)
 
+    progress_obj = progress(indeterminate=True)
+    progress_obj.report(30)
```
obrusvit: The progress is reported in the range from 0 to 1000 so if you want to go by thirds, put 300 or so.

ibz: Good catch, fixed it in the normal ETH flow as well, which is where I took the bad inspiration from.

bieleluk (approved):
utACK

PR #5838, opened 2025-09-23
Task/applet improvements

cepetr:
This PR brings several improvements and fixes that make it possible to run multiple unprivileged apps in parallel.

- Separates the applet logic from coreapp-specific header
- adds systask static data base register settings (SB) for relocatable executables (RWPI)
- corrects MPU reconfiguration when switching between multiple unprivileged task 
   (after event polling and a callback from kernel)
- removes unnecessary stack-protector usage in syscalls stubs as a small optimization
- moves certain constant in coreapp from .data to .rodata segment 
- removes irq locking from unprivileged (micropython) code
- increases MAIN_RAM area from 24KB to 28KB

### DIFF
--- a/core/embed/projects/firmware/mpconfigport.h
+++ b/core/embed/projects/firmware/mpconfigport.h
@@ -194,10 +194,8 @@ typedef int mp_int_t; // must be pointer size
 typedef unsigned int mp_uint_t; // must be pointer size
 typedef long mp_off_t;
 
-#include <sys/irq.h>
-
-#define MICROPY_BEGIN_ATOMIC_SECTION()     irq_lock()
-#define MICROPY_END_ATOMIC_SECTION(state)  irq_unlock(state)
+#define MICROPY_BEGIN_ATOMIC_SECTION()     (0)
+#define MICROPY_END_ATOMIC_SECTION(state)  (void)(state)
 #define MICROPY_EVENT_POLL_HOOK \
     do { \
         extern void mp_handle_pending(bool); \


core/embed/sys/mpu/stm32u5/mpu.c
```diff
@@ -337,16 +337,33 @@ void mpu_set_active_applet(applet_layout_t* layout) {
...
+      }
+    }
     // clang-format on
+
   } else {
     DIS_REGION(2);
     DIS_REGION(3);
     DIS_REGION(4);
+
+    drv->app_tls = (mpu_area_t){0, 0};
```
matejcik: i'll need a little explanation here

(a) this function is always only touching regions 2, 3, 4 and optionally 7
what is with the other regions? how do we ensure those are known good?

(b) why is region 7 only touched conditionally? i'd expect `DIS_REGION` to be unconditional

core/embed/sys/task/stm32/systask.c
```diff
@@ -58,6 +58,7 @@ typedef struct {
   systask_t* waiting_task;
   // Bitmap of used task IDs
   uint32_t task_id_map;
+
```
matejcik, line 61: stray newline

core/embed/sys/task/sysevent.c
```diff
@@ -195,8 +199,6 @@ void sysevents_poll(const sysevents_t *awaited, sysevents_t *signalled,
                     uint32_t deadline) {
   sysevent_dispatcher_t *dispatcher = &g_sysevent_dispatcher;
 
-  memset(signalled, 0, sizeof(*signalled));
```
matejcik, line 198: any reason for this removal?
it seems cleaner to always clear the `signalled` array on entry, in case we do an early return somewhere before setting the result

core/embed/projects/firmware/mpconfigport.h
```diff
@@ -194,10 +194,8 @@ typedef int mp_int_t; // must be pointer size
 typedef unsigned int mp_uint_t; // must be pointer size
 typedef long mp_off_t;
 
-#include <sys/irq.h>
-
-#define MICROPY_BEGIN_ATOMIC_SECTION()     irq_lock()
-#define MICROPY_END_ATOMIC_SECTION(state)  irq_unlock(state)
+#define MICROPY_BEGIN_ATOMIC_SECTION()     (0)
+#define MICROPY_END_ATOMIC_SECTION(state)  (void)(state)
```
matejcik, line 198: can you explain this change?

cepetr: Since we run micropython in a single thread and it never executes in interrupt context, we do not need this kind of synchronization.

When calling `irq_lock()` and `irq_unlock()` in unprivileged mode, interrupts are not disabled or enabled - they are just no-ops (writes to PRIMASK are completely ignored by the CPU).

It was effectively a bug that has now been fixed.

core/embed/sys/task/sysevent.c
cepetr: yes, you are right, i reverted this. See 373612313f0fa9db673ab41c5726c29bcd311d03

core/embed/sys/task/stm32/systask.c
cepetr: In most cases, I add an empty line after the last definition in a struct. Our clang-format doesn't enforce this rule, so it might be something we could unify.

core/embed/sys/mpu/stm32u5/mpu.c
cepetr: Yes, this is getting tricky (and not easy to explain).

Regions 2, 3, and 4 are used exclusively for the applet's unprivileged memory (code or data) and nothing else, so the function can set them directly. The changes apply immediately in any MPU mode. 

Region 7 is shared (it has multiple purposes), so it can be changed only when the MPU is in APP mode, where this region is used for the unprivileged TLS section. When the MPU is in different mode, we store the value and use later.

The function `mpu_set_active_applet()` only enables the memory regions used by the applet and is called when the kernel needs to access applet memory or switch to this applet. It should not change any other regions, since those are controlled by other functions in different situations.

matejcik: tagging in @TychoVrahe for reviewing this piece.

I dislike the way this function is organized, esp. wrt the region 7 that is only configured sometimes and sometimes not. But Petr tells me that yes, this is somewhat less than ideal, but also we'll need to live with it until we can fully overhaul the MPU config.
As such, I want someone familiar with MPU code to review and approve this.

cepetr: Finally, I decided to make the change. I hope it makes things much cleaner (and I think now it's as you proposed). See 6986f0ecc85ada81adaf4387882f91a3aa4eb621

PR #5760, opened 2025-09-12
Use Tropic in AuthenticateDevice

andrewkozlik:
Successfully tested on a provisioned TS7.

TODOs:
- In a separate PR: Add certificates to the Tropic emulator and add Tropic to the AuthenticateDevice device test, see https://github.com/trezor/trezor-firmware/issues/5801
- @matejcik has some followup improvements for the changes in trezorctl. https://github.com/trezor/trezor-firmware/pull/5802

PR #5845, opened 2025-09-23
Tropic PIN protection

onvej-sl:


onvej-sl:
TODO:

- [x] Fix time estimates.
- [x] Consider decreasing `PIN_STRETCH_ITERATIONS` and PIN_ITER_COUNT=1.

core/embed/sec/optiga/inc/sec/optiga.h
```diff
@@ -62,13 +62,31 @@ void optiga_set_sec_max(void);
...
 bool __wur optiga_random_buffer(uint8_t *dest, size_t size);
 
-bool __wur optiga_pin_set(optiga_ui_progress_t ui_progress,
-                          uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]);
+bool __wur optiga_pin_init(optiga_ui_progress_t ui_progress);
 
-uint32_t optiga_estimate_time_ms(storage_pin_op_t op);
+optiga_pin_result __wur
+optiga_stretch_pin(optiga_ui_progress_t ui_progress,
+                   uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]);
```
andrewkozlik: How about making the return value the same type as `optiga_stretch_pin_offline()`?

core/embed/sec/optiga/inc/sec/optiga.h
```diff
@@ -62,13 +62,31 @@ void optiga_set_sec_max(void);
...
+
+bool __wur optiga_stretch_pin_offline(
+    optiga_ui_progress_t ui_progress,
+    const uint8_t stretching_secret[OPTIGA_PIN_SECRET_SIZE],
```
andrewkozlik: Let's rename this to something more specific like `hmac_stretching_secret` or `hmac_key`.

core/embed/sec/secret/stm32u5/secret_keys.c
```diff
@@ -156,6 +156,13 @@ secbool secret_key_tropic_masking(uint8_t dest[ECDSA_PRIVATE_KEY_SIZE]) {
                                      KEY_INDEX_TROPIC_MASKING, dest);
 }
 
+secbool secret_key_is_privileged_accessible(void) {
+  curve25519_key key = {0};
+  secbool result = secret_key_tropic_pairing_privileged(key);
+  memzero(key, sizeof(key));
+  return result;
+}
```
andrewkozlik: We should probably replace this function with a static global variable `pkey_index_t g_pairing_key_slot` in `core/embed/sec/tropic/tropic.c`, that will indicate which one of the pairing key slots was used in `tropic_init()`.

core/embed/sec/optiga/unix/optiga.c
```diff
@@ -99,15 +99,38 @@ bool optiga_read_sec(uint8_t *sec) {
...
+
+bool optiga_stretch_pin_offline(
+    optiga_ui_progress_t ui_progress,
+    const uint8_t stretching_secret[OPTIGA_PIN_SECRET_SIZE],
+    uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]) {
+  return true;
+}
+
+bool optiga_reset_hmac_counter(optiga_ui_progress_t ui_progress,
+                               const uint8_t reset_key[OPTIGA_PIN_SECRET_SIZE]);
```
andrewkozlik: This seems redundant.

core/embed/sec/tropic/inc/sec/tropic.h
```diff
@@ -33,6 +36,16 @@
...
 #define TROPIC_DEVICE_KEY_SLOT 0  // ECC_SLOT_0
 
+#ifdef USE_STORAGE
+// KEK masks used in PIN verification
+#define TROPIC_KEK_MASKS_PRIVILEGED_SLOT 32
+#define TROPIC_KEK_MASKS_UNPRIVILEGED_SLOT 64
+
+// Mac-and-destroy slots used in PIN verification
+#define TROPIC_FIRST_MAC_AND_DESTROY_SLOT_PRIVILEGED 0
+#define TROPIC_FIRST_MAC_AND_DESTROY_SLOT_UNPRIVILEGED 32
```
andrewkozlik:
```suggestion
#define TROPIC_FIRST_MAC_AND_DESTROY_SLOT_UNPRIVILEGED 64
```

core/embed/sec/tropic/inc/sec/tropic.h
```diff
@@ -33,6 +36,16 @@
 #define TROPIC_DEVICE_CERT_SLOT_COUNT 3
 #define TROPIC_DEVICE_KEY_SLOT 0  // ECC_SLOT_0
 
+#ifdef USE_STORAGE
+// KEK masks used in PIN verification
+#define TROPIC_KEK_MASKS_PRIVILEGED_SLOT 32
+#define TROPIC_KEK_MASKS_UNPRIVILEGED_SLOT 64
```
andrewkozlik: `TROPIC_KEK_MASKS_PRIVILEGED_SLOT` should be in the range 128-255 and `TROPIC_KEK_MASKS_UNPRIVILEGED_SLOT` should be in the range 256-511.
```suggestion
#define TROPIC_KEK_MASKS_PRIVILEGED_SLOT 128
#define TROPIC_KEK_MASKS_UNPRIVILEGED_SLOT 256
```

core/embed/sec/tropic/inc/sec/tropic.h
```diff
@@ -57,6 +70,10 @@ lt_handle_t* tropic_get_handle(void);
 
 #endif
 
+#include "libtropic_common.h"
```
andrewkozlik: We don't want to include libtropic in our headers files, cf. https://github.com/trezor/trezor-firmware/pull/5760#discussion_r2354665973. It's probably better to redefine `TROPIC_MAC_AND_DESTROY_SIZE` here.

core/embed/sec/tropic/inc/sec/tropic.h
```diff
@@ -75,3 +92,27 @@ bool tropic_data_multi_read(uint16_t first_slot, uint16_t slot_count,
...
+
+bool tropic_reset_slots(tropic_ui_progress_t ui_progress, uint16_t pin_index,
+                        const uint8_t reset_key[MAC_AND_DESTROY_DATA_SIZE]);
+
+bool tropic_pin_set(
+    tropic_ui_progress_t ui_progress,
+    uint8_t stretched_pins[PIN_MAX_TRIES][MAC_AND_DESTROY_DATA_SIZE],
+    uint8_t reset_key[MAC_AND_DESTROY_DATA_SIZE]);
+
+bool tropic_set_kek(
```
andrewkozlik: Let's rename this to `tropic_set_kek_masks()` to better document what the function is doing. Otherwise it gives the impression that we are writing the KEK to the Tropic chip.

core/embed/sec/tropic/inc/sec/tropic.h
```diff
@@ -75,3 +92,27 @@ bool tropic_data_multi_read(uint16_t first_slot, uint16_t slot_count,
...
+    tropic_ui_progress_t ui_progress,
+    const uint8_t kek[MAC_AND_DESTROY_DATA_SIZE],
+    const uint8_t stretched_pins[PIN_MAX_TRIES][MAC_AND_DESTROY_DATA_SIZE]);
+
+bool tropic_get_kek(tropic_ui_progress_t ui_progress,
+                    const uint8_t stretched_pin[MAC_AND_DESTROY_DATA_SIZE],
+                    uint16_t pin_index, uint8_t kek[MAC_AND_DESTROY_DATA_SIZE]);
```
andrewkozlik: Let's rename this to `tropic_unmask_kek()` and move the `pin_index` parameter before the `stretched_pin` parameter like it is in the other functions.

core/embed/sec/optiga/inc/sec/optiga_commands.h
```diff
@@ -239,6 +239,8 @@ optiga_result optiga_derive_key(optiga_key_derivation deriv, uint16_t oid,
                                 size_t key_size);
 optiga_result optiga_set_trust_anchor(void);
 optiga_result optiga_set_priv_key(uint16_t oid, const uint8_t priv_key[32]);
+optiga_result optiga_clear_all_auto_states(void);
+optiga_result optiga_set_counter(uint16_t oid, uint8_t value);
```
andrewkozlik: Let's change the name to `optiga_reset_counter()` and rename the `value` parameter to `limit`. Otherwise it gives the impression that the function sets the counter to a particular value, when it actually clears the counter.

storage/storage.c
```diff
@@ -1257,7 +1434,9 @@ static secbool storage_set_encrypted(const uint16_t key, const void *val,
 
   // Write the IV to the flash.
   uint8_t buffer[CHACHA20_BLOCK_SIZE] = {0};
-  random_buffer(buffer, CHACHA20_IV_SIZE);
+  if (!rng_fill_buffer_strong(buffer, CHACHA20_IV_SIZE)) {
```
andrewkozlik: I wonder whether we need strong entropy for the IV, since the main expectation from the IV is avoidance of collisions, not secrecy. I am hesitant in this case, because I don't know what kind of UX impact this may have on the firmware. We should first find all the places which lead to `storage_set_encrypted()` being called and see how long `rng_fill_buffer_strong()` takes to execute.

storage/storage.c
```diff
@@ -74,6 +80,17 @@
 // NOTE: This should always equal the value in VERSION_KEY.
 #define UNAUTH_VERSION_KEY ((APP_STORAGE << 8) | 0x08)
 
+#if USE_TROPIC
+// Key that is used to reset the M&D slots in Tropic after successfull unlock.
+#define TROPIC_HMAC_RESET_KEY ((APP_STORAGE << 8) | 0x09)
```
andrewkozlik:
```suggestion
#define MAC_AND_DESTROY_RESET_KEY ((APP_STORAGE << 8) | 0x09)
```

storage/storage.h
```diff
@@ -44,7 +44,23 @@ extern const uint8_t *PIN_EMPTY;
...
+#endif
+
+// The number of slots configured as stretched PINs.
+#if USE_OPTIGA
+#if USE_TROPIC
+#define OPTIGA_STRETCHED_PINS_COUNT PIN_MAX_TRIES
+#else
+#define OPTIGA_STRETCHED_PINS_COUNT 1
+#endif
+#endif
```
andrewkozlik: This constant is not necessarily Optiga-specific. If we had only Tropic, then this would still be used. In fact it is used in `core/embed/sec/tropic/tropic.c`. So I think:
- We can rename `OPTIGA_STRETCHED_PINS_COUNT` -> `STRETCHED_PIN_COUNT`.
- Move `#if USE_OPTIGA` deeper inside, or maybe even remove the condition altogether.
```suggestion
#if USE_TROPIC
#define STRETCHED_PIN_COUNT PIN_MAX_TRIES
#else
#if USE_OPTIGA
#define STRETCHED_PIN_COUNT 1
#endif
#endif
```

andrewkozlik: Whenever we have `<something>_count`, let's please keep `<something>` singular. I don't think this is just my OCD, because I don't see anyone suggesting the use of a plural in this case. For reference a few pages that pop up in a Google search:
- `nodeCount` https://stackoverflow.com/questions/548646/naming-a-count-variable
- `errorCount` https://google.github.io/styleguide/tsguide.html#descriptive-names
- `failureCount`, `orderCount` https://medium.com/better-programming/useful-tips-for-naming-your-variables-8139cc8d44b5
- A general English usage discussion, e.g. "body count" https://www.usingenglish.com/forum/threads/plural-or-singular-when-using-with-the-word-count.242655/
- Reddit discussion, `treeCount` https://www.reddit.com/r/readablecode/comments/21tvjj/variable_naming_count_vs_number_vs_numberof_vs_n/

The same applies to `*mac_and_destroys_count` below.

storage/storage.c
```diff
@@ -663,17 +693,91 @@ static void derive_kek_optiga_v4(
...
+  uint8_t optiga_hmac_stretching_secret[OPTIGA_PIN_SECRET_SIZE] = {0};
+  if (!rng_fill_buffer_strong(optiga_hmac_stretching_secret,
+                              sizeof(optiga_hmac_stretching_secret))) {
+    goto cleanup;
+  }
+  if (!optiga_pin_init(ui_progress)) {
+    goto cleanup;
+  }
+  if (!optiga_stretch_pin_offline(ui_progress, optiga_hmac_stretching_secret,
+                                  stretched_pin)) {
```
andrewkozlik: It would be nice if we could generate `optiga_hmac_stretching_secret` inside `optiga_stretch_pin_offline()` and set it in Optiga already there. That way it wouldn't even have to leave the context of that function, because the only reason why it is needed here is for `optiga_pin_set()` to set it in Optiga, which would no longer be necessary. It would also make the arguments of `optiga_pin_set()` identical with `tropic_pin_set()`, which is a nice touch. Of course we would need to tweak the name `optiga_stretch_pin_offline()`, but first I want to make sure that this proposal is even feasible. I haven't thought yet about what auto states need to be active to be able to set the `optiga_hmac_stretching_secret` in Optiga.

andrewkozlik: I think the idea as described above doesn't work because we first need to authorize using the `optiga_hmac_reset_key` in order to write the `optiga_hmac_stretching_secret` to Optiga. There are still ways to hide the `optiga_hmac_stretching_secret` inside `optiga.c` using a static variable there, but that's not very pretty.

core/embed/sec/optiga/inc/sec/optiga.h
onvej-sl: When I look at it now, I think we should wipe the storage if `optiga_stretch_pin()` returns `OPTIGA_PIN_COUNTER_EXCEEDED`, which is the same as what we do with `optiga_pin_verify()`.

core/embed/sec/optiga/inc/sec/optiga.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/a5065fc0a98a2a17fdaaba53d6b93108dbaba2e1.

core/embed/sec/optiga/unix/optiga.c
onvej-sl: Removed in https://github.com/trezor/trezor-firmware/pull/5845/commits/231066b1430489c963f3d0574d43d9c8c77657f0.

core/embed/sec/tropic/inc/sec/tropic.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/a8df47ae8f741343eb7043188ae93cbee8894e00.

core/embed/sec/tropic/inc/sec/tropic.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/a8df47ae8f741343eb7043188ae93cbee8894e00.

core/embed/sec/optiga/inc/sec/optiga_commands.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/d93226e9218bb3f959f647a5d4d4d1dd41d6d51a.

storage/storage.c
onvej-sl: Renamed to `TROPIC_MAC_AND_DESTROY_RESET_KEY` in https://github.com/trezor/trezor-firmware/pull/5845/commits/867deb6169ef41ef8cae824c1f7260c61ae1864f.

core/embed/sec/optiga/inc/sec/optiga.h
andrewkozlik: I agree, we should do that. In that case keeping the present return value makes sense.

storage/storage.h
onvej-sl: > This constant is not necessarily Optiga-specific. If we had only Tropic, then this would still be used. In fact it is used in core/embed/sec/tropic/tropic.c.

I don't agree with that. I intended for the constant to be optiga-specific. In tropic, I use `PIN_MAX_TRIES` instead. The only occurrence of `OPTIGA_STRETCHED_PINS_COUNT` in `tropic.c` is a mistake and should be replaced with `PIN_MAX_TRIES`.

andrewkozlik: The reason why it doesn't seem Optiga-specific to me is that we have pieces of code even in storage.c that are independent of Optiga and they still work with `OPTIGA_STRETCHED_PINS_COUNT`. For example if we had a Tropic-only device, we would still need the `pin_index` or `pin_fails` to estimate time and to know which M&D slot to access in Tropic, and for that storage.c has these two pieces of code:
```c
#if OPTIGA_STRETCHED_PINS_COUNT > 1
  if (sectrue != pin_get_fails(&pin_index)) {
```
```c
#if OPTIGA_STRETCHED_PINS_COUNT > 1
  uint32_t pin_fails = 0;
  if (sectrue != pin_get_fails(&pin_fails)) {
```
Similarly when setting up the PIN we work with an array of stretched PINs in storage.c which is declared like this:
```c
#if USE_TROPIC || USE_OPTIGA
  uint8_t stretched_pins[OPTIGA_STRETCHED_PINS_COUNT][SHA256_DIGEST_LENGTH] = {
```

storage/storage.c
onvej-sl: I used `rng_fill_buffer()` in https://github.com/trezor/trezor-firmware/pull/5845/commits/9eb1c66cbc0ab2ab8eec94b37b75fd3f8b828c26.

core/embed/sec/optiga/inc/sec/optiga.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/24e2b9db74c653dd3a9e4e9782e1234fc7b04cd7.

storage/storage.h
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/cf1c01d00e7c66fd53d0b4cc5a028124168f9ff4, https://github.com/trezor/trezor-firmware/pull/5845/commits/e2a5f30ee53bec90cd87722cdb31be695d7e3c2d and https://github.com/trezor/trezor-firmware/pull/5845/commits/c01557a4b26219b5c0e6af5d4e2e02014ea7d287.

core/embed/sec/secret/stm32u5/secret_keys.c
onvej-sl: https://github.com/trezor/trezor-firmware/pull/5845/commits/dc606e8c1dc4f3cbd6acfc90fd60d64cf8579aad aims to implement what you suggested and to fix the issue with tropic pairing on a partially provisioned trezor.  However, I have not tested it yet. I would also like to use `session_start` in prodtest in the future.

andrewkozlik: So we can remove `secret_key_is_privileged_accessible()`, right?

core/embed/sec/tropic/tropic.c
```diff
@@ -53,6 +55,49 @@ static bool tropic_get_tropic_pubkey(lt_handle_t *handle,
...
+  if (pairing_key_index == TROPIC_PRIVILEGED_PAIRING_KEY_SLOT) {
+    if (secret_key_tropic_pairing_privileged(trezor_private) != sectrue) {
+      return false;
+    }
+  }
+  if (pairing_key_index == TROPIC_UNPRIVILEGED_PAIRING_KEY_SLOT) {
+    if (secret_key_tropic_pairing_unprivileged(trezor_private) != sectrue) {
+      return false;
+    }
+  }
```
andrewkozlik: nit: I'd make this into a `switch` statement for better legibility and add a `default` clause.

core/embed/sec/tropic/tropic.c
```diff
@@ -74,49 +119,19 @@ bool tropic_init(void) {
...
+  if (session_start(drv, TROPIC_FACTORY_PAIRING_KEY_SLOT)) {
+    return true;
   }
-
-  memzero(trezor_privkey, sizeof(trezor_privkey));
-
-  drv->initialized = true;
+#endif
 
   return true;
```
andrewkozlik: I think this shouldn't be here. If all session starts fail, then we should fall through the cleanup and return false.

core/embed/sec/tropic/tropic.c
```diff
@@ -269,6 +284,231 @@ bool tropic_random_buffer(void *buffer, size_t length) {
...
+  tropic_driver_t *drv = &g_tropic_driver;
+
+  if (!drv->initialized) {
+    return false;
+  }
+
+  int first_slot_index =
+      drv->pairing_key_index == TROPIC_UNPRIVILEGED_PAIRING_KEY_SLOT
+          ? TROPIC_FIRST_MAC_AND_DESTROY_SLOT_UNPRIVILEGED
+          : TROPIC_FIRST_MAC_AND_DESTROY_SLOT_PRIVILEGED;
```
andrewkozlik: This gets repeated a lot. Would be nice to factor it out into a static function.

storage/storage.c
```diff
@@ -714,24 +818,65 @@ static secbool __wur derive_kek_unlock(
...
+#endif
+#if USE_TROPIC
+  if (!tropic_stretch_pin(ui_progress, pin_index, stretched_pin)) {
+    goto cleanup;
+  }
+#endif
+#if USE_OPTIGA
+  optiga_ret = optiga_pin_verify(ui_progress, pin_index, stretched_pin);
+  if (optiga_ret != OPTIGA_PIN_SUCCESS) {
+    memzero(stretched_pin, SHA256_DIGEST_LENGTH);
```
andrewkozlik, line 830: ISTM that this should go into the `cleanup:` section. Otherwise the `stretched_pin` doesn't get cleaned up in case of later failures. Also same memzero on line 837 doesn't need to be there.

storage/storage.c
```diff
@@ -714,24 +818,65 @@ static secbool __wur derive_kek_unlock(
...
     memzero(stretched_pin, SHA256_DIGEST_LENGTH);
 
-    if (ret == OPTIGA_PIN_COUNTER_EXCEEDED) {
+    if (optiga_ret == OPTIGA_PIN_COUNTER_EXCEEDED) {
       // Unreachable code. Wipe should have already been triggered in unlock().
       storage_wipe();
       show_pin_too_many_screen();
     }
-    ensure(ret == OPTIGA_PIN_INVALID ? sectrue : secfalse,
+    ensure(optiga_ret == OPTIGA_PIN_INVALID ? sectrue : secfalse,
```
andrewkozlik: `optiga_stretch_pin()` can't return `OPTIGA_PIN_INVALID`, can it? So does that mean we always RSOD?
- If yes, then let's be more explicit about it.
- Should we RSOD? I realize we do below with `optiga_pin_verify()`, so that's consistent, but we don't RSOD for any of the other functions when they fail e.g. `tropic_stretch_pin()`, so that doesn't seem consistent. I don't know what the right answer is, but we should make the behavior consistent unless there is a reason not to.

core/embed/sec/tropic/tropic.c
```diff
@@ -40,6 +41,7 @@
 typedef struct {
   bool initialized;
   bool sec_chan_established;
```
andrewkozlik: `sec_chan_established` seems unused, so either remove or set accordingly in `session_start()`

core/embed/sec/tropic/tropic.c
```diff
@@ -53,6 +55,49 @@ static bool tropic_get_tropic_pubkey(lt_handle_t *handle,
...
+#endif
+    {
+      memzero(trezor_private, sizeof(trezor_private));
+      return false;
+    }
+  }
+
+  if (lt_session_start(&drv->handle, tropic_public, pairing_key_index,
+                       trezor_private, trezor_public) != LT_OK) {
+    memzero(trezor_private, sizeof(trezor_private));
```
andrewkozlik: We also want to memzero the private key upon success. So add another memzero below, or maybe add a `cleanup` section?

core/embed/sec/tropic/tropic.c
```diff
@@ -74,49 +119,19 @@ bool tropic_init(void) {
...
-  secbool pubkey_ok = secret_key_tropic_public(tropic_pubkey);
-
-#if !PRODUCTION
-  // Allow running with default factory keys in non-production fw
-  if (privkey_ok != sectrue) {
-    tropic_get_factory_privkey(trezor_privkey);
-    pairing_key_slot = TROPIC_FACTORY_PAIRING_KEY_SLOT;
-    privkey_ok = sectrue;
+#ifndef TREZOR_EMULATOR
+  if (session_start(drv, TROPIC_UNPRIVILEGED_PAIRING_KEY_SLOT)) {
```
andrewkozlik: We should start with the privileged key and then try the unprivileged.

core/embed/sec/tropic/tropic.c
```diff
@@ -269,6 +284,231 @@ bool tropic_random_buffer(void *buffer, size_t length) {
...
+  for (int i = 0; i <= pin_index; i++) {
+    ui_progress();
+
+    res = lt_mac_and_destroy(&drv->handle, first_slot_index + i, reset_key,
+                             output);
+    if (res != LT_OK) {
+      goto cleanup;
+    }
+
+    ui_progress();
```
andrewkozlik, line 371: nit: this ui_progress() can go after the `for`-cycle, because we are unnecessarily calling the function twice in a row

core/embed/sec/tropic/tropic.c
```diff
@@ -269,6 +284,231 @@ bool tropic_random_buffer(void *buffer, size_t length) {
...
+
+  int first_slot_index =
+      drv->pairing_key_index == TROPIC_UNPRIVILEGED_PAIRING_KEY_SLOT
+          ? TROPIC_FIRST_MAC_AND_DESTROY_SLOT_UNPRIVILEGED
+          : TROPIC_FIRST_MAC_AND_DESTROY_SLOT_PRIVILEGED;
+
+  ui_progress();
+
+  for (int i = 0; i < PIN_MAX_TRIES; i++) {
+    ui_progress();
```
andrewkozlik: nit: this ui_progress() is redundant

core/embed/sec/tropic/tropic.c
```diff
@@ -269,6 +284,231 @@ bool tropic_random_buffer(void *buffer, size_t length) {
...
+
+  switch (op) {
+    case STORAGE_PIN_OP_SET:
+      return pin_set_mac_and_destroy_count * mac_and_destroy_time_ms;
+    case STORAGE_PIN_OP_VERIFY:
+      return pin_verify_mac_and_destroy_count * mac_and_destroy_time_ms;
+    case STORAGE_PIN_OP_CHANGE:
+      return pin_change_mac_and_destroy_count * mac_and_destroy_time_ms;
+    default:
+      return 0;
```
andrewkozlik, line 539: Ideas for improvement. We could also include the time it takes:
- For SET to erase and write the masks slot, which will be roughly another 30ms.
- For SET to generate the reset key. Not sure whether that should be here though.
- For VERIFY to read the masks slot, probably over 12 ms.

core/embed/sec/optiga/optiga.c
```diff
@@ -69,12 +67,16 @@
...
-// Value of the PIN counter when it is reset.
-static const uint8_t COUNTER_RESET[] = {0, 0, 0, 0, 0, 0, 0, PIN_MAX_TRIES};
-
-// Value of the PIN counter with one extra attempt needed in optiga_pin_set().
-static const uint8_t COUNTER_RESET_EXTRA[] = {0, 0, 0, 0,
-                                              0, 0, 0, PIN_MAX_TRIES + 1};
+// Stretched PINs
+// The first stretched PIN is OPTIGA_OID_DATA + 4 to preserve compatiblity with
+// Trezors without Tropics.
+// OPTIGA_OID_DATA + 4 and OPTIGA_OID_DATA + 8 are not used since they are
```
andrewkozlik:
```suggestion
// OPTIGA_OID_DATA + 0 and OPTIGA_OID_DATA + 8 are not used since they are
```

core/embed/sec/optiga/optiga.c
```diff
@@ -69,12 +67,16 @@
...
+// occupied by the PIN secret and PIN HMAC secret.
+static const uint16_t OID_STRETCHED_PINS[] = {
+    OPTIGA_OID_DATA + 4, OPTIGA_OID_DATA + 1, OPTIGA_OID_DATA + 2,
+    OPTIGA_OID_DATA + 3, OPTIGA_OID_DATA + 5, OPTIGA_OID_DATA + 6,
+    OPTIGA_OID_DATA + 7, OPTIGA_OID_DATA + 9, OPTIGA_OID_DATA + 10,
+    OPTIGA_OID_DATA + 11};
```
andrewkozlik, line 83: I think we should have here a static_assert ensuring that the number of array elements is STRETCHED_PIN_COUNT

core/embed/sec/optiga/optiga.c
```diff
@@ -357,40 +384,92 @@ bool optiga_set_metadata(uint16_t oid, const optiga_metadata *metadata) {
...
   metadata.execute = OPTIGA_META_ACCESS_ALWAYS;
   metadata.data_type = TYPE_AUTOREF;
   if (!optiga_set_metadata(OID_PIN_SECRET, &metadata)) {
     return false;
   }
 
-  // Set metadata for stretched PIN.
+  ui_progress();
+
+#if OPTIGA_STRETCHED_PINS == 1
```
andrewkozlik: This symbol doesn't seem to exist. Probably `STRETCHED_PIN_COUNT`.

core/embed/sec/optiga/optiga.c
```diff
@@ -602,134 +707,137 @@ static bool optiga_pin_stretch_cmac_ecdh(
   return ret;
 }
 
-bool optiga_pin_set(optiga_ui_progress_t ui_progress,
-                    uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]) {
+bool optiga_pin_init(optiga_ui_progress_t ui_progress) {
+  return optiga_pin_init_metadata(ui_progress) &&
+         optiga_pin_init_stretch(ui_progress);
```
andrewkozlik: Missing `set_ui_progress()`.

andrewkozlik: Same problem also in `optiga_stretch_pin()`, `optiga_stretch_pin_offline()` and `optiga_reset_hmac_counter()`. We should always register the `ui_progress` callback before executing commands and deregister it before exiting. Alternatively, we could make storage responsible for registering and deregistering the callback, but I am not sure it's wise to place the responsibility for deregistering it on the storage code.

andrewkozlik: On a related note, I see this PR is adding lots of `ui_progress()` calls, when actually I think we can remove all of them, because `ui_progress()` is now being called in the `optiga_read()` loop. For context, the callback was added there so that the progress bar does not freeze when Optiga is enforcing a throttling delay. I am not sure why I didn't remove all the other `ui_progress()` calls when I introduced this feature.

core/embed/sec/optiga/optiga.c
```diff
@@ -602,134 +707,137 @@ static bool optiga_pin_stretch_cmac_ecdh(
...
     goto end;
   }
 
-  // Generate the key for the HMAC-SHA256 PIN stretching step.
-  uint8_t pin_hmac[OPTIGA_PIN_SECRET_SIZE] = {0};
-  if (!rng_fill_buffer_strong(pin_hmac, sizeof(pin_hmac))) {
+  ui_progress();
+
+  // Authorise using OID_PIN_SECRET so that we can write to the last stretched
+  // PIN
```
andrewkozlik:
```suggestion
  // PIN and to OID_STRETCHED_PIN_CTR.
```

core/embed/sec/optiga/optiga.c
```diff
@@ -602,134 +707,137 @@ static bool optiga_pin_stretch_cmac_ecdh(
...
+    // reduce the security of their device any more than if the Optiga was not
+    // integrated into the device in the first place.
+    hmac_sha256(stretched_pins[i], OPTIGA_PIN_SECRET_SIZE, pin_secret,
+                sizeof(pin_secret), stretched_pins[i]);
+
+    ui_progress();
+
+    // Authorise using OID_STRETCHED_PINS[i] so that we can write to
+    //  * OID_STRETCHED_PINS[i + 1], if i < STRETCHED_PIN_COUNT - 1;
+    //  * OID_PIN_HMAC and OID_PIN_HMAC_CTR, if i == STRETCHED_PIN_COUNT - 1.
```
andrewkozlik:
```suggestion
    //  * OID_STRETCHED_PINS[i - 1], if i > 0;
    //  * OID_PIN_HMAC and OID_PIN_HMAC_CTR, if i == 0.
```

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/4bfb81f38d23d17515b67559d7114f8ae2c9d51f.

core/embed/sec/tropic/tropic.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/4bfb81f38d23d17515b67559d7114f8ae2c9d51f.

core/embed/sec/tropic/tropic.c
onvej-sl: Removed in https://github.com/trezor/trezor-firmware/pull/5845/commits/4bfb81f38d23d17515b67559d7114f8ae2c9d51f.

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/4bfb81f38d23d17515b67559d7114f8ae2c9d51f.

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/4bfb81f38d23d17515b67559d7114f8ae2c9d51f.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/8a1c754f2faad77a20f6795fa7970f1ae0a8f7c7.

core/embed/sec/optiga/optiga.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/8a1c754f2faad77a20f6795fa7970f1ae0a8f7c7.

core/embed/sec/tropic/tropic.c
```diff
@@ -53,6 +54,50 @@ static bool tropic_get_tropic_pubkey(lt_handle_t *handle,
...
+    goto cleanup;
+  }
+
+  drv->pairing_key_index = pairing_key_index;
+  drv->initialized = true;
+
+cleanup:
+  memzero(trezor_private, sizeof(trezor_private));
+
+  return drv->initialized;
```
andrewkozlik, line 106: If we use `drv->initialized` as the return value, then we should set it to `false` at the beginning of the function.

core/embed/sec/optiga/optiga.c
```diff
@@ -69,12 +67,19 @@
 // The throttling delay when the security event counter is at its maximum.
```
andrewkozlik, line 71: Commenting on line 65 above. We should consider setting `PIN_STRETCH_ITERATIONS` to 1 if `USE_TROPIC` or `STRETCHED_PIN_COUNT > 1`. It would slightly reduce the number of security events and it would also slightly reduce the overall time for the PIN verification, both of which have grown now.

core/embed/sec/optiga/optiga.c
```diff
@@ -858,32 +975,81 @@ static optiga_pin_result optiga_pin_stretch_hmac(
...
   uint8_t digest[OPTIGA_PIN_SECRET_SIZE] = {0};
   hmac_sha256(stretched_pin, OPTIGA_PIN_SECRET_SIZE, NULL, 0, digest);
 
-  // Authorise using OID_STRETCHED_PIN so that we can read from OID_PIN_SECRET
-  // and reset OID_PIN_HMAC_CTR.
-  optiga_result res = optiga_set_auto_state(
-      OPTIGA_OID_SESSION_CTX, OID_STRETCHED_PIN, digest, sizeof(digest));
+  ui_progress();
+
+  // Authorise using OID_STRETCHED_PINS[pin_index]
```
andrewkozlik:
```suggestion
  // Authorise using OID_STRETCHED_PINS[pin_index] so that we can read from
  // OID_STRETCHED_PINS[pin_index + 1] and eventually from OID_PIN_SECRET.
```

core/embed/sec/optiga/optiga.c
```diff
@@ -934,24 +1134,63 @@ optiga_pin_result optiga_pin_verify(
...
+    if (optiga_reset_counter(OID_STRETCHED_PIN_CTR, PIN_MAX_TRIES + 1) !=
+        OPTIGA_SUCCESS) {
+      ret = OPTIGA_PIN_ERROR;
+      goto end;
+    }
   }
 
   ui_progress();
 
+  optiga_clear_all_auto_states();
```
andrewkozlik: nit: redundant, since the function is called again in cleanup

core/embed/sec/optiga/optiga.c
```diff
@@ -934,24 +1134,63 @@ optiga_pin_result optiga_pin_verify(
...
+  ui_progress();
+
+  // Reset the counter.
+  if (optiga_reset_counter(OID_PIN_HMAC_CTR, PIN_MAX_TRIES) != OPTIGA_SUCCESS) {
+    return false;
+  }
+
+  ui_progress();
+
+  return true;
```
andrewkozlik: I think it would be good to have a cleanup section here, where we clear all auto states, e.g. if `optiga_reset_counter()` fails, then we leave the auto state open. Not a major security issue I suppose, but not OK.

core/embed/sec/tropic/tropic.c
onvej-sl: In https://github.com/trezor/trezor-firmware/pull/5845/commits/04a818b279c3ac4fc42bf01e2fe1c9283a280ded, I removed all occurrences of `ui_progress()`.

core/embed/sec/tropic/tropic.c
onvej-sl: In https://github.com/trezor/trezor-firmware/pull/5845/commits/04a818b279c3ac4fc42bf01e2fe1c9283a280ded, I removed all occurrences of `ui_progress()`.

core/embed/sec/optiga/optiga.c
onvej-sl: In https://github.com/trezor/trezor-firmware/pull/5845/commits/04a818b279c3ac4fc42bf01e2fe1c9283a280ded, I removed all occurrences of `ui_progress()` and added registration and deregistration of the `ui_progress` callback.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/14f041a48ff222208eca8c397b3e3a9e24a67d00.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/14f041a48ff222208eca8c397b3e3a9e24a67d00.

onvej-sl:
https://github.com/trezor/trezor-firmware/pull/5845/commits/04a818b279c3ac4fc42bf01e2fe1c9283a280ded swaps the optiga hmac stretching with tropic stretching.

core/embed/sec/optiga/optiga.c
andrewkozlik: registration and deregistration still missing in `optiga_reset_hmac_counter()`

core/embed/sec/optiga/optiga.c
```diff
@@ -516,8 +587,6 @@ static bool optiga_pin_stretch_common(
     goto end;
   }
 
-  ui_progress();
```
andrewkozlik, line 519: This change means that `ui_progress()` is an unused parameter of `optiga_pin_stretch_common()` now, so it should be removed from the function signature. I would expect the compiler to complain about this.

If I am not mistaken, this also means that it will become an unused parameter in `optiga_pin_stretch_secret_v4()`

core/embed/sec/optiga/optiga.c
```diff
@@ -816,12 +898,13 @@ optiga_pin_result optiga_pin_verify_v4(
...
 end:
   memzero(stretched_pin, sizeof(stretched_pin));
-  optiga_clear_auto_state(OID_STRETCHED_PIN);
+  optiga_clear_all_auto_states();
   optiga_set_ui_progress(NULL);
   return ret;
 }
 
 static optiga_pin_result optiga_pin_stretch_hmac(
+    optiga_ui_progress_t ui_progress,
```
andrewkozlik: `ui_progress` - unused parameter

core/embed/sec/optiga/optiga.c
```diff
@@ -859,31 +942,29 @@ static optiga_pin_result optiga_pin_stretch_hmac(
...
-    goto end;
+  if (pin_index >= STRETCHED_PIN_COUNT) {
+    return OPTIGA_PIN_ERROR;
   }
 
-  ret = optiga_pin_stretch_hmac(stretched_pin);
-  if (ret != OPTIGA_PIN_SUCCESS) {
-    goto end;
+  if (optiga_pin_stretch_hmac(ui_progress, stretched_pin)) {
+    return OPTIGA_PIN_ERROR;
```
andrewkozlik: Incorrect error handling. We should test `!= OPTIGA_PIN_SUCCESS` and if the function failed, then return the error from `optiga_pin_stretch_hmac()`, but not immediately, we have to go through `end:` first

core/embed/sec/optiga/optiga.c
```diff
@@ -859,31 +942,29 @@ static optiga_pin_result optiga_pin_stretch_hmac(
...
     uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]) {
   optiga_set_ui_progress(ui_progress);
-  optiga_pin_result ret = OPTIGA_PIN_SUCCESS;
 
-  // Stretch the PIN more with stretching secrets from the Optiga.
-  if (!optiga_pin_stretch_cmac_ecdh(ui_progress, stretched_pin)) {
-    ret = OPTIGA_PIN_ERROR;
-    goto end;
+  if (pin_index >= STRETCHED_PIN_COUNT) {
+    return OPTIGA_PIN_ERROR;
```
andrewkozlik:
```suggestion
    ret = OPTIGA_PIN_ERROR;
    goto end;
```

core/embed/sec/optiga/unix/optiga.c
```diff
@@ -99,15 +99,35 @@ bool optiga_read_sec(uint8_t *sec) {
...
+    uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]) {
+  return OPTIGA_PIN_SUCCESS;
+}
+
+bool optiga_stretch_pin_offline(
+    optiga_ui_progress_t ui_progress,
+    const uint8_t hmac_stretching_secret[OPTIGA_PIN_SECRET_SIZE],
+    uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]) {
+  return true;
+}
```
andrewkozlik: unused functions

andrewkozlik:
In optiga.c we have functions that deal with signing, certificates and PIN authentication. I named the PIN authentication functions `optiga_pin_*` to make it easy to immediately tell that the function is related to that part of Optiga functionality. We have these:
- `optiga_pin_init_metadata()`
- `optiga_pin_init_stretch()`
- `optiga_pin_stretch_cmac_ecdh()`
- `optiga_pin_stretch_hmac()`
- `optiga_pin_stretch_common()`
- `optiga_pin_stretch_secret_v4()`
- `optiga_pin_set()`
- `optiga_pin_verify()`
- `tropic_pin_set()`

Then we have some functions which don't respect this convention. In some cases respecting it would make the names rather long, but in most cases I think it helps consistency and clarity, e.g. `tropic_reset_slots()` is not immediately obvious that it's PIN related.
- `optiga_reset_hmac_counter()` -> `optiga_pin_reset_hmac_counter()`
- `optiga_stretch_hmac_offline()` -> `optiga_pin_stretch_hmac_offline()`
- `tropic_stretch_pin()` -> `tropic_pin_stretch()`
- `tropic_reset_slots()` -> `tropic_pin_reset_slots()`
- `tropic_set_kek_masks()` -> `tropic_pin_set_kek_masks()` (maybe remove "kek", if it feels too long)
- `tropic_unmask_kek()` -> `tropic_pin_unmask_kek()`
- `mcu_stretch_pin()` -> `mcu_pin_stretch()` (for consistency with the other stretching functions)

WDYT?

There are also the `*_estimate_time_ms()` functions, but we might decide to use those for other operations in the future, such as signing. So I wouldn't necessarily put those into the `pin_` namespace.

core/embed/sec/optiga/unix/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/801de050a39db0cb071631af82e47298bcc96f71.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/801de050a39db0cb071631af82e47298bcc96f71.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/801de050a39db0cb071631af82e47298bcc96f71.

core/embed/sec/optiga/optiga.c
onvej-sl: > I would expect the compiler to complain about this.

It looks like gcc sometimes fails to warn about unused parameters. :man_shrugging: 

Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/801de050a39db0cb071631af82e47298bcc96f71.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/801de050a39db0cb071631af82e47298bcc96f71.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/a7940ff0ae52e4b4a3e50e9f12989d64522e865d.

onvej-sl:
> WDYT?

The renaming you suggested makes sense to me.

core/embed/sec/secret/stm32u5/secret_keys.c
onvej-sl: Removed in https://github.com/trezor/trezor-firmware/pull/5845/commits/c97b2bfcc75c54e1f7441b7201063dacc97f894a. I will remove the reverted commit during squashing.

storage/storage.c
onvej-sl: This part of code was significantly changed in https://github.com/trezor/trezor-firmware/commit/04a818b279c3ac4fc42bf01e2fe1c9283a280ded and I believe it has fixed the issue.

storage/storage.c
onvej-sl: I think https://github.com/trezor/trezor-firmware/commit/04a818b279c3ac4fc42bf01e2fe1c9283a280ded should fix this as well. I decided to show RSOD in the event of any unexpected error during verification.

core/embed/sec/tropic/tropic.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/6d5f79ee24f2f275c0f088b4424283d353f6435a.

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/6d5f79ee24f2f275c0f088b4424283d353f6435a.

andrewkozlik (commented):
utACK. I am very happy with how the code turned out, especially considering the complexity of dealing with the different combinations of SEs. :rocket: 

I am not hitting "Approve" just yet, because I would like to re-review the PR one more time if I can, ideally after rebasing over main, but I also don't want to hold it back from being merged unnecessarily. Given the importance of this code, I am rather nervous that there isn't a second reviewer to look for potential issues. @M1nd3r can you have a look too?

Next steps:
- @onvej-sl please squash the fixups now, rebase over `main` and force-push. After that @M1nd3r and I will go for a second round of review.
- I haven't reviewed the time estimates at all, because as I understand, that's not done yet. Shall we deal with the time estimates in this PR or defer them to a later PR?
- Consider reducing `PIN_ITER_COUNT` and `PIN_STRETCH_ITERATIONS`. We can do that in a separate PR too, but should deal with it before Tuesday next week. It would be helpful to have some timing statistics on for the current setup vs. `PIN_ITER_COUNT=1` and `PIN_STRETCH_ITERATIONS=1`.
- Let's do some quick upgrade tests on T1 or TT (one model with no Optiga) and TS3 or TS5 (one model with Optiga) to make sure that the storage isn't broken. How about @M1nd3r? Build firmware before this change and after this change. Initialize the storage and set up a PIN with the old firmware. Upgrade to new firmware and make sure that the PIN unlocks the device correctly. QA should do this thoroughly for all models later on including scenarios with failed PIN attempts before and after upgrade.
- Think about tweaking the design to parallelize Tropic and Optiga operations. IMO not worth the effort, given that Optiga operations are going to be the bottleneck, and I doubt that parallelizing would actually improve the execution time in any significant way (150ms?).

core/embed/sec/tropic/tropic.c
```diff
@@ -269,6 +288,231 @@ bool tropic_random_buffer(void *buffer, size_t length) {
...
+  if (!drv->initialized) {
+    return false;
+  }
+
+  mac_and_destroy_slot_t first_slot_index = get_first_mac_and_destroy_slot(drv);
+
+  ui_progress();
+
+  lt_ret_t res = lt_mac_and_destroy(&drv->handle, first_slot_index + pin_index,
+                                    stretched_pin, stretched_pin);
```
andrewkozlik: We are missing the HMAC envelope here and in all places where `lt_mac_and_destroy()` is called. By HMAC envelope I mean:
```
digest = HMAC-SHA256(stretched_pin, "")
out = MAC-AND-DESTROY(slot, digest)
stretched_pin = HMAC-SHA256(stretched_pin, out)
```

storage/storage.c
```diff
@@ -663,17 +693,76 @@ static void derive_kek_optiga_v4(
...
+#endif
+#if USE_TROPIC
+  if (!rng_fill_buffer_strong(kek, SHA256_DIGEST_LENGTH)) {
+    goto cleanup;
+  }
+  if (tropic_pin_set_kek_masks(ui_progress, kek, stretched_pins) != true) {
+    goto cleanup;
+  }
+#else
+  memcpy(kek, stretched_pins[0], SHA256_DIGEST_LENGTH);
```
onvej-sl, line 752: TODO

core/embed/sec/optiga/optiga.c
andrewkozlik: I did some manual testing and the PIN authentication seems to take a pretty long time ~4.5 seconds (didn't do a precise measurement). I am strongly in favor of setting `PIN_ITER_COUNT=1` and `PIN_STRETCH_ITERATIONS=1` when `STRETCHED_PIN_COUNT > 1`, or maybe when `USE_OPTIGA && USE_TROPIC`. This should reduce the time by 1 to 1.5 seconds.

core/embed/sec/optiga/optiga_commands.c
```diff
@@ -515,6 +533,14 @@ optiga_result optiga_set_auto_state(uint16_t nonce_oid, uint16_t key_oid,
   hmac_sha256(key, key_size, nonce, sizeof(nonce), ptr);
 
   OPTIGA_LOG(__func__, tx_buffer, tx_size)
+  // The auto-state is added before the command is actually executed. Otherwise,
+  // it could happen that the command succeeded, but an error occurred during
+  // the receipt of the response. In such a case, the auto-state couldn't be
+  // cleared using optiga_clear_all_auto_states().
+  if (!auto_states_add(key_oid)) {
+    // This should never happen.
+    return OPTIGA_ERR_UNEXPECTED;
```
andrewkozlik: It can happen if too many auto states are requested. We should probably return what Optiga would return `OPTIGA_ERR_CODE_MEMORY`.

core/embed/sec/tropic/tropic.c
```diff
@@ -269,6 +288,231 @@ bool tropic_random_buffer(void *buffer, size_t length) {
...
+  uint16_t masked_kek_slot = get_kek_masks_slot(drv);
+
+  ret = lt_r_mem_data_erase(&drv->handle, masked_kek_slot);
+  if (ret != LT_OK) {
+    goto cleanup;
+  }
+
+  ui_progress();
+
+  ret = lt_r_mem_data_write(&drv->handle, masked_kek_slot, (uint8_t *)masks,
```
andrewkozlik: nit: `masks` shouldn't need a type cast

core/embed/sec/tropic/tropic.c
andrewkozlik: actually not addressed

storage/storage.c
```diff
@@ -1110,6 +1226,44 @@ static secbool unlock(const uint8_t *pin, size_t pin_len,
...
+    if (!optiga_pin_reset_hmac_counter(ui_progress, optiga_hmac_reset_key)) {
+      memzero(optiga_hmac_reset_key, sizeof(optiga_hmac_reset_key));
+      return secfalse;
+    }
+    memzero(optiga_hmac_reset_key, sizeof(optiga_hmac_reset_key));
+  }
+#endif
+
+#if USE_TROPIC
+  uint8_t tropic_mac_and_destroy_reset_key[SHA256_DIGEST_LENGTH] = {0};
```
andrewkozlik: nit: SHA256_DIGEST_LENGTH -> TROPIC_MAC_AND_DESTROY_SIZE

Same in the `if`-statement below. Actually, it would be better to check `tropic_mac_and_destroy_reset_key_len != sizeof(tropic_mac_and_destroy_reset_key)`.

core/embed/sec/optiga/optiga.c
```diff
@@ -905,13 +987,33 @@ optiga_pin_result optiga_pin_verify(
...
+  if (pin_index == 0) {
+    //  If this is the first PIN attempt or there is only one stretched PIN
+    //  slot, the HMAC counter can be reset immediately. Otherwise, the counter
+    //  is reset in optiga_pin_reset_hmac_counter().
+    if (optiga_reset_counter(OID_PIN_HMAC_CTR, PIN_MAX_TRIES) !=
+        OPTIGA_SUCCESS) {
+      ret = OPTIGA_PIN_ERROR;
+      goto end;
+    }
+  }
```
andrewkozlik, line 1036: How about adding an `else` branch here as shown in the suggestion? This will
1. simplify code the code 40 lines later (see my second suggestion), and
2. localize the logic which deals with the optimization to this one place.
```suggestion
  uint8_t stretched_pin_ctr_limit = PIN_MAX_TRIES;
  if (pin_index == 0) {
    //  If this is the first PIN attempt or there is only one stretched PIN
    //  slot, the HMAC counter can be reset immediately. Otherwise, the counter
    //  is reset in optiga_pin_reset_hmac_counter().
    if (optiga_reset_counter(OID_PIN_HMAC_CTR, PIN_MAX_TRIES) !=
        OPTIGA_SUCCESS) {
      ret = OPTIGA_PIN_ERROR;
      goto end;
    }
  } else {
    // An extra attempt will be needed to authorize using OID_STRETCHED_PIN[0]
    // in optiga_pin_reset_hmac_counter().
    stretched_pin_ctr_limit += 1;
  }
```

core/embed/sec/optiga/optiga.c
```diff
@@ -934,24 +1038,58 @@ optiga_pin_result optiga_pin_verify(
...
+      goto end;
+    }
+  } else {
+    // The one extra attempt is used to authorize using the first stretched PIN
+    // in optiga_pin_reset_hmac_counter().
+    if (optiga_reset_counter(OID_STRETCHED_PIN_CTR, PIN_MAX_TRIES + 1) !=
+        OPTIGA_SUCCESS) {
+      ret = OPTIGA_PIN_ERROR;
+      goto end;
+    }
```
andrewkozlik: See my other code improvement suggestion above.
```suggestion
  if (optiga_reset_counter(OID_STRETCHED_PIN_CTR, stretched_pin_ctr_limit) !=
      OPTIGA_SUCCESS) {
    ret = OPTIGA_PIN_ERROR;
    goto end;
```

andrewkozlik (commented):
I finished my second round of review, added a few more code improvement comments and noticed two previous minor comments that were actually not resolved. The only thing I haven't reviewed is `optiga_estimate_time_ms()`, because that's not finished yet.

The only critical things are:
1. <s>the missing HMAC envelope around the MAC-and-destroy calls and</s>
2. <s>the reduction in the number of iterations.</s>

Although the estimates are a priority too of course, they can technically be done later.

Some general thoughts that came across my mind while reading this PR:
- In tropic.c I wouldn't mind changing `PIN_MAX_TRIES` -> `STRETCHED_PIN_COUNT`, because it is always used to represent the size of `stretched_pins` array. Both constants are the same thing of course and I think it doesn't cause any confusion, so I don't have a strong preference either way. Leaving it up to you.
- Changing `pin_index` -> `stretched_pin_index` everywhere would seem more comprehensive, because it represents the index in the `stretched_pins` array or in `OID_STRETCHED_PINS`, whereas `pin_index` gives the impression that it represents the number of the current PIN attempt, which is not true for Tropic-less devices.
- The strong RNG calls in optiga.c, tropic.c and storage.c should be added to the time estimate. I don't know how significant its contribution to the overall time is, but the tiny pieces add up. I also don't know whether Optiga imposes throttling on the RNG calls. If yes, then it would be very significant. For this it would probably make sense to add `uint32_t time_estimate_rng_strong_ms(void)` to `core/embed/sec/time_estimate/inc/sec/time_estimate.h`, which may just return a constant based on `USE_TROPIC` and `USE_OPTIGA` or call something like `optiga_estimate_time_ms(OP_RNG)` if throttling is an issue.

core/embed/sec/optiga/inc/sec/optiga.h
```diff
@@ -62,13 +62,25 @@ void optiga_set_sec_max(void);
...
+bool __wur optiga_pin_init(optiga_ui_progress_t ui_progress);
 
-uint32_t optiga_estimate_time_ms(storage_pin_op_t op);
+bool optiga_pin_stretch_cmac_ecdh(
+    optiga_ui_progress_t ui_progress,
+    uint8_t stretched_pin[OPTIGA_PIN_SECRET_SIZE]);
+
+bool __wur optiga_pin_set(
+    optiga_ui_progress_t ui_progress,
+    uint8_t stretched_pins[STRETCHED_PIN_COUNT][OPTIGA_PIN_SECRET_SIZE],
```
M1nd3r, line 73: super small nit:
```suggestion
    uint8_t stretched_pins[STRETCHED_PINS_COUNT][OPTIGA_PIN_SECRET_SIZE],
```

andrewkozlik: I actually had an opposite nit [here](https://github.com/trezor/trezor-firmware/pull/5845#discussion_r2381450627). I'd like to open a discussion about some variable naming conventions in tech_firmware once our workload goes down a bit. IMO this is just a czechism, which unfortunately appeared a few times in our code at the beginning (`inputs_count` in Bitcoin protobuf) and has lately started spreading heavily.

M1nd3r: I agree with the language pow. Nevertheless, I think that `STRETCHED_PINS_COUNT` is slightly better because it describes something like `stretched_pins.count()`. Simply: keeping the ROOT of the variable (array/list) name and appending `_COUNT` at the end.

I agree that the language pow is valid. I'll leave it up to @onvej-sl - whether to go with the _formulaic approach_ (append `_COUNT` -> `STRETCHED_PINS_COUNT`) or the grammatically correct one (keep as-is).

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/8dd713d390f64d63b1df1e5e7cf5955c10e0dbbb.

core/embed/sec/optiga/inc/sec/optiga.h
M1nd3r: @andrewkozlik I retract my previous comment - the consistence among all variables is better (from my pow) than adherence to the formulaic "append `_COUNT`" approach.

onvej-sl: I understand both approaches and don't have a strong opinion on them. However, I do not want to rename it again.

core/embed/sec/optiga/optiga_commands.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/87ada3df40cbc82071756f4493ce2bc198c71483.

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/7998bda7339b75e3c0593b3e37987c9c2b121840.

core/embed/sec/tropic/tropic.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/7998bda7339b75e3c0593b3e37987c9c2b121840.

core/embed/sec/tropic/tropic.c
onvej-sl: Removed in https://github.com/trezor/trezor-firmware/pull/5845/commits/7998bda7339b75e3c0593b3e37987c9c2b121840.

M1nd3r:
Review status update:
- I have not found any new issues, yet. However, I was not able to verify everything thoroughly enough. I will have a look at it in the following days (possibly also over the weekend).
- The upgrade tests (done yesterday) were successful on TT and TS5 devices ✅

core/embed/sec/optiga/optiga.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/8fa89074293c2786180132b73e43295d5098bdcf.

storage/storage.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/9c96d670f10b909bd46c80faa450259d2be2d120.

core/embed/sec/optiga/optiga.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/8fa89074293c2786180132b73e43295d5098bdcf.

core/embed/sec/optiga/optiga.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/8fa89074293c2786180132b73e43295d5098bdcf.

storage/storage.c
```diff
@@ -106,7 +106,7 @@ const uint32_t V0_PIN_EMPTY = 1;
 // up constant storage space.
 #define MAX_WIPE_CODE_LEN 50
 
-#if STORAGE_INSECURE_TESTING_MODE && !PRODUCTION
+#if (STORAGE_INSECURE_TESTING_MODE && !PRODUCTION) || STRETCHED_PIN_COUNT > 1
 #define PIN_ITER_COUNT 1
```
onvej-sl, line 110: TODO: This results in zero iterations, since `10 * (PIN_ITER_COUNT / 10)` iterations are actually performed.

andrewkozlik: It's actually OK. Our PBKDF2 implementation always guarantees one iteration, but it's undocumented, so I agree not very nice.

andrewkozlik (approved):

core/embed/sec/tropic/tropic.c
```diff
@@ -269,6 +290,252 @@ bool tropic_random_buffer(void *buffer, size_t length) {
...
+  for (int i = 0; i < TROPIC_MAC_AND_DESTROY_SIZE; i++) {
+    kek[i] =
+        masks[pin_index * TROPIC_MAC_AND_DESTROY_SIZE + i] ^ stretched_pin[i];
+  }
+  return true;
+}
+
+uint32_t tropic_estimate_time_ms(storage_pin_op_t op, uint16_t pin_index) {
+  const int pin_verify_time = 50 + 100 + (pin_index + 1) * 50;
+  const int pin_set_time = 65 + PIN_MAX_TRIES * 155 + 130;
```
andrewkozlik: Can we assign some simple names to the constants? Like:
```c
const int pin_stretch = 50;
const int unmask_kek = 100;
const int reset_slot = pin_stretch;
const int init_slot = 155;
const int set_kek_masks = 130;
...
```

core/embed/sec/optiga/optiga.c
```diff
@@ -259,13 +272,45 @@ uint32_t optiga_estimate_time_ms(storage_pin_op_t op) {
   // To estimate the overall time of the PIN operation we multiply the
   // throttling delay by the number of protected Optiga commands and add the
   // time required to execute all Optiga commands without throttling delays.
+
+  const int init_metadata_time = 550;
```
andrewkozlik: I'd expect `init_metadata_time` to depend on `STRETCHED_PIN_COUNT`. Something like `(7 + STRETCHED_PIN_COUNT) * set_metadata_time`.

core/embed/sec/optiga/optiga.c
```diff
@@ -259,13 +272,45 @@ uint32_t optiga_estimate_time_ms(storage_pin_op_t op) {
...
+  const int init_metadata_protected_command_count = 0;
+
```
andrewkozlik, line 279: Aren't the time parameters for `optiga_pin_init_stretch()` missing?

core/embed/sec/optiga/optiga.c
```diff
@@ -259,13 +272,45 @@ uint32_t optiga_estimate_time_ms(storage_pin_op_t op) {
...
+  const int stretch_cmac_ecdh_time = PIN_STRETCH_ITERATIONS * 275;
+  const int stretch_cmac_ecdh_protected_command_count =
+      PIN_STRETCH_ITERATIONS * 2;
+
+  const int set_time = STRETCHED_PIN_COUNT * 180 + 400;
```
andrewkozlik, line 284: I assume that about 220 ms of the 400 here are mainly the two `rng_fill_buffer_strong()` calls. This time is probably going to be less for devices without Tropic. (Not sure if this is significant though.)

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/a6da1ca5aad729450d18f7d8ad28bc39d5cb1f16.

core/embed/sec/optiga/optiga.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5845/commits/a6da1ca5aad729450d18f7d8ad28bc39d5cb1f16. 550 ms is the combined time of  `optiga_pin_init_metadata` and `optiga_pin_init_stretch`.

core/embed/sec/optiga/optiga.c
onvej-sl: `rng_fill_buffer_strong()` takes 66 ms, from which 16 ms is spent on optiga and 50 ms on tropic. I think this can be fixed later.

core/embed/sec/tropic/tropic.c
onvej-sl: Done in https://github.com/trezor/trezor-firmware/pull/5845/commits/aefa781389627d2fac666556e35d789ae9199311.

core/embed/sec/optiga/optiga.c
```diff
@@ -259,13 +272,46 @@ uint32_t optiga_estimate_time_ms(storage_pin_op_t op) {
...
+  const int set_protected_command_count = 2 * (STRETCHED_PIN_COUNT + 1);
+
+  const int verify_time = (STRETCHED_PIN_COUNT - pin_index - 1) * 170 + 470;
+  const int verify_protected_command_count =
+      1 + 2 * (STRETCHED_PIN_COUNT - pin_index + 1);
```
andrewkozlik, line 289:
```suggestion
      2 * (STRETCHED_PIN_COUNT - pin_index + 1 + (pin_index != 0));
```

andrewkozlik: Wait, no. I don't understand what the `1 +` is at the beginning. I thought it was related to `reset_hmac_counter_protected_command_count`, but no.

onvej-sl: Which operation is represented by ` (pin_index != 0)`?

andrewkozlik (approved):

M1nd3r (approved):

PR #5847, opened 2025-09-23
feat(core): poke IdleTimer on THP BLE msgs

obrusvit:
- prevent long running workflows on bluetooth from auto-locking the device

[no changelog]

### DIFF
--- a/core/src/trezor/wire/thp/interface_context.py
+++ b/core/src/trezor/wire/thp/interface_context.py
@@ -27,6 +27,11 @@ from .checksum import CHECKSUM_LENGTH
 if __debug__:
     from trezor import log
 
+
+if utils.USE_BLE:
+    import trezorble as ble
+    from trezor.workflow import idle_timer
+
 if TYPE_CHECKING:
     from buffer_types import AnyBuffer, AnyBytes
     from trezorio import WireInterface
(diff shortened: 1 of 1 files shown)


core/src/trezor/wire/thp/interface_context.py
```diff
@@ -94,6 +99,9 @@ def _wait_for_packet(self) -> Generator[wait, int, tuple["InterfaceContext", int
         """
         # Uses `yield` instead of `await` to avoid allocations.
         packet_len = yield self._read
+        if utils.USE_BLE and self._iface == ble.interface.iface_num():
```
romanz: `self._iface` is `WireInterface`, so you probably meant:
```suggestion
        if utils.USE_BLE and self._iface is ble.interface:
```

romanz (approved):
utACK

core/src/trezor/wire/thp/interface_context.py
obrusvit: fixed on rebase.

mmilata:
I think this is not quite right - see #7522. Do you remember which workflows were interrupted this way? We should probably do it differently, like introducing a loader to the affected workflows.

obrusvit:
@mmilata IIRC, this was introduced specifically for Suite discovery.

I think the right way how to fix that would be changing our `BusyScreen` to have configurable message (possibly via enum, so that we can choose an appropriate string in FW). Then if Suite starts discovery, it can send `SetBusy(reason=PLEASE_WAIT)`. The busyscreen can poke idle_timer instead.

mmilata:
I see, discovery is a bunch of `GetPublicKey` right? Not a multimessage workflow, so merging 294dabbc9e04a851057c3166bfee0732a1d77574 would break it. I suppose extending `SetBusy` could work.

Do we need access to storage after discovery is started? If not perhaps we can turn discovery into a workflow and use the `workflow.autolock_interrupts_workflow` mechanism?

Or we could poke the timer whenever we receive discovery-related message over BLE. At least until proper solution is in place.

PR #5850, opened 2025-09-23
fix(core): wait for `ButtonRequest` task to avoid THP desync

romanz:

### DIFF
--- a/core/src/trezor/ui/__init__.py
+++ b/core/src/trezor/ui/__init__.py
@@ -26,7 +26,7 @@ if TYPE_CHECKING:
     from trezorui_api import LayoutObj, UiResult  # noqa: F401
 
     T = TypeVar("T", covariant=True)
-    ButtonRequestTuple = tuple[ButtonRequestType, str]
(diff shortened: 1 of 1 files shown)


core/src/trezor/ui/__init__.py
```diff
@@ -249,11 +256,27 @@ async def get_result(self) -> T:
...
+                self.button_request_task = self._handle_button_requests(is_done)
+                self._start_task(self.button_request_task)
+            result = await self.result_box
+
+            if is_done is not None:
+                # Make sure ButtonRequest is ACKed, before the result is returned.
+                # Otherwise, THP channel may become desynced (due to two consecutive writes).
+                try:
+                    if self.button_request_box.is_empty():
+                        self.button_request_box.put(None)
```
romanz, line 272: Use `put(None, replace=None)`, otherwise the `await` below may never return.

romanz: Should be fixed in https://github.com/trezor/trezor-firmware/pull/5862.

PR #5852, opened 2025-09-24
connection interval fix

TychoVrahe:
This PR adjust connection intervals for better reaction times and fixes setting high speed connection for newly connected devices.


kopecdav (approved):
utACK

PR #5857, opened 2025-09-24
signal firmware corrupted in features

TychoVrahe:


core/embed/projects/bootloader/protob/pb/messages.pb.h
```diff
@@ -114,8 +94,36 @@ typedef struct _Features {
     uint32_t unit_packaging;
     bool has_soc;
     uint32_t soc;
+    bool has_firmware_corrupted;
```
obrusvit, line 97: `has_firmware_corrupted` is unused?

TychoVrahe: this is auto-generated and used inside this macro: 

```
#define MSG_SEND_ASSIGN_VALUE(FIELD, VALUE) \
  {                                         \
    msg_send.has_##FIELD = true;            \
    msg_send.FIELD = VALUE;                 \
  }
```

PR #5858, opened 2025-09-24
Font kerning

kopecdav:
This PR updates the text rendering routines to take into account the font kernings + updates fonts export scripts to extract the kerning pairs from .odt files.

_Note: Currently the fonts for T3W1 do not contain any kerning pairs probably due to incorrect export from its font parent, this will be most likely fixed in near feature._


TychoVrahe:
Before continuing, we will also need to implement support for kerning with translations/language specific chars. I imagine we should add secondary kerning table to the translations blob where we'd look for these special chars.

TychoVrahe:
superseded by https://github.com/trezor/trezor-firmware/pull/6620

PR #5859, opened 2025-09-24
timeout FW installation when the next message doesn't arrive in time

TychoVrahe:

### DIFF
--- a/core/embed/projects/bootloader/workflow/wf_firmware_update.c
+++ b/core/embed/projects/bootloader/workflow/wf_firmware_update.c
@@ -43,6 +43,8 @@
 #include "emulator.h"
 #endif
 
+#define MESSAGE_RX_TIMEOUT 10000
+
 typedef enum {
   UPLOAD_OK = 0,
   UPLOAD_IN_PROGRESS = 1,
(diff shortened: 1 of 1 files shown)


core/embed/projects/bootloader/workflow/wf_firmware_update.c
```diff
@@ -643,6 +647,12 @@ workflow_result_t workflow_firmware_update(protob_io_t *iface) {
 
     sysevents_poll(&awaited, &signalled, ticks_timeout(100));
 
+    if (ticks_expired(msg_deadline)) {
+      // timeout
+      ui_screen_fail();
+      return WF_ERROR;
```
kopecdav: Isn't WF_ERROR too harsh? This will ends up in RSOD even if the communication breaks on host side. Maybe cancelling the upload saying "FW installation failed" and going back to bootloader makes more sense?

TychoVrahe: We can definitely reconsider, but for now this is same result as when the communication breaks in the middle of receiving message, you'll also end up with RSOD. There are bunch of things we would need to solve too:: if installing on fresh device fails in the middle, the state after reboot 'firmware corrupted', but we don't have transition from workflow 'empty device' to regular bootloader, so we'd need to implement that too.

kopecdav (approved):

Issue #5623, opened 2025-08-23
Chunkified text doesn't properly work in the TextLayout


bieleluk:
##  Incomplete chunks at the end of the line

When using the `TextLayout` to render the left-aligned chunkified text, the text doesn't break only at the entire chunks, i.e.: 

<img width="414" height="129" alt="Image" src="https://github.com/user-attachments/assets/dbecc793-abad-481c-aadd-86c792ba1545" />

### Expected layout

```
8208  8811  1111
1111  11
```

### Root cause

`TextLayout` uses `Spans` to render the individual lines, in particular, the function `fit_horizontally`. The issue is that the function goes only character-by-character and doesn't look at the following characters to see if they would complete a chunk at the beginning of another line.

### Fix

We would have to predict if the entire chunk fits the line in the function `Span::fit_horizontally`

## Center and Right alignment do not work at all with chunkified text

When using the center alignment, the first chunk is rendered in the middle of the line, the second in the middle of the right half line, ..., until the text doesn't fit.

<img width="403" height="108" alt="Image" src="https://github.com/user-attachments/assets/a50b84a8-7981-4aee-aac3-0ac120951595" />

### Severity

The `TextLayout` for the centered chunkified text is needed for the `Eckhart`'s PIN Keyboard screen: [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=11094-5812&t=7Xm7nQBSWalF82SG-0). For now, the left alignment is used.

### Root cause

The function`TextLayout::layout_text` uses `Spans` to render the individual lines. In the normal text, the `Span` corresponds to one line, and the alignment works properly. After the end of the span, the span is aligned in the remaining space. In the normal text, the remaining space is the entire line.
But in the case of the chunkified text, one span corresponds to one chunk because the chunk gap has to be rendered. This is also the reason for the strange rendering because:
The remaining space for the first chunk is the entire line, the remaining space for the second chunk is the right half of the line, ...

### Fix

The alignment should be done once per line, not after the each `Span` is processed.

PR #5864, opened 2025-09-25
Discard chunk when it is too big

PrisionMike:
Fixes the first part of #5623 

Ensures that chunk boundaries respect line width constraints by peeking ahead to check if the next chunk fits before breaking the line.

[no changelog]


PrisionMike:
Updating fixtures required. The failing test had a small extra offset that has been addressed now.
Screenshot From 2025-09-28 10-08-50

PrisionMike:
Fixes #5882

mmilata:
This shouldn't happen:
Screenshot 2025-10-01 at 17-06-14 Master differing screens

core/embed/rust/src/ui/component/text/layout.rs
```diff
@@ -400,9 +420,6 @@ impl TextLayout {
                         if should_append_ellipsis {
                             sink.ellipsis(*cursor, self);
                         }
-                        // TODO: This does not work in case we are the last
```
mmilata, line 403: Very historical TODO😅 2703d714c2dc1f3ba907d223825a58e2a914b173

PrisionMike:
- `Caesar` click test resolved. Was caused by incorrect use of `is_last_line` in inserting linebreak.
- Fixes #5882 . No ellipses overflow in `Eckhart`
- Partially fixes #5623 . Partial chunks won't be rendered if there is more text available to complete it.
- `End` and `Center` alignment issue mentioned in #5623 still need major rework.
- [CI run](https://data.trezor.io/dev/firmware/ui_report/18203020143/T3W1-en-core_device_test-index.html) with old fixtures for reference.

obrusvit (commented):
As of the offline discussion, we will go with this solution https://github.com/trezor/trezor-firmware/pull/5934

romanz:
https://github.com/trezor/trezor-firmware/pull/5934 was merged, so we can close this PR - right?

PR #5865, opened 2025-09-25
add notifications on important state changes

TychoVrahe:
[no changelog]


core/src/apps/management/change_pin.py
```diff
@@ -65,6 +65,7 @@ async def change_pin(msg: ChangePin) -> Success:
         msg_wire = "PIN removed"
 
     await success_pin_change(curpin, newpin)
+    utils.notify_send(utils.NOTIFY_PIN_CHANGE)
```
obrusvit: maybe we can call it before `success_pin_change`? 🤔

core/embed/sys/notify/inc/sys/notify.h
```diff
@@ -39,9 +39,16 @@
...
 typedef enum {
-  NOTIFY_BOOT = 0,   /**< Device boot/startup notification */
-  NOTIFY_UNLOCK = 1, /**< Device unlocked and ready to accept messages */
-  NOTIFY_LOCK = 2    /**< Device hard-locked and won't accept messages */
+  NOTIFY_BOOT = 0,       /**< Device boot/startup notification */
+  NOTIFY_UNLOCK = 1,     /**< Device unlocked and ready to accept messages */
+  NOTIFY_LOCK = 2,       /**< Device hard-locked and won't accept messages */
+  NOTIFY_DISCONNECT = 3, /**< User-initiated disconnect from host */
+  NOTIFY_SETTING_CHANGE = 4, /**< Change of settings */
+  NOTIFY_SOFTLOCK =
```
obrusvit, line 47: should we use `SOFTLOCK` and `SOFTUNLOCK` also in bootloader when you (auto-)suspend? 

@szymonlesisz suggests it's good for consistency.

TychoVrahe: i dont think so, if we want that we should introduce SUSPEND & WAKEUP notifications, locking is different thing

obrusvit: Right, I suggest not adding those if suite doesn't need it.

core/src/apps/management/change_pin.py
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/7d5288de334eb16859ecf1c81d039e5237157c3c

PR #5867, opened 2025-09-25
fix(core): cache host and app name also at credential-based connection 

### LINKED ISSUE #21886: THP host name is not displayed in menu after pairing over BLE
**Steps to reproduce:**
1. Pair over BLE
2. Go to BLE device menu
3. BLE MAC is shown (instead of THP host name)

**Info:**
 - Suite version: desktop 25.10.0 (c2e8b42781f1de33a91be6e06aa2124aebaca120)
 - Browser: Trezor Suite 25.10.0
 - OS: Linux x86_64
 - Screen: 1920x1200
 - Device: Trezor T3W1 2.9.2 universal (revision c07fc0f2098d0448d2221abdabf7937bb425d53b)
 - Transport: BluetoothTransport, BridgeTransport 3.1.0

romanz:
If THP pairing is done over USB (which doesn't have a MAC address) the cache won't be populated.

In order to fix this, we will also update the cache on credential-based connection.

Fixes https://github.com/trezor/trezor-suite/issues/21886.

Rebased over #5869.

### DIFF
--- a/core/.changelog.d/5867.fixed
+++ b/core/.changelog.d/5867.fixed
@@ -1,0 +1,1 @@
+[T3W1] Cache THP host info also during credential-based pairing.
--- a/core/src/apps/thp/pairing.py
+++ b/core/src/apps/thp/pairing.py
@@ -50,6 +50,7 @@ if __debug__:
     from trezor import log
 
 if TYPE_CHECKING:
+    from buffer_types import AnyBytes
     from typing import Any, Callable, Concatenate, ParamSpec
 
     from trezorui_api import UiResult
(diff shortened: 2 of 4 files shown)


M1nd3r (approved):
utACK

One thing I do not like is that I needed to check two functions 
```
ctx.iface_ctx.connected_addr()
```
and 
```
cache_host_info
```
to verify that USB connections are not affected. (Because `mac_addr = ctx.iface_ctx.connected_addr()` will set `mac_addr` to `None` and `cache_host_info` will abort when `mac_addr` is `None`.)

IMO it could use some comment, documentation or something. But the minor lack of clarity is definitely a non-blocking issue that can be solved later / put into backlog, if needed.

romanz:
Rebasing over `main` to resolve a conflict.

romanz:
> IMO it could use some comment, documentation or something. But the minor lack of clarity is definitely a non-blocking issue that can be solved later / put into backlog, if needed.

Added a comment & more explicit typing in c8a074d4b9.

PR #5868, opened 2025-09-25
Fix tropic deinitialization

onvej-sl:
This pull request addresses an issue where trezor returns `LT_L1_SPI_ERROR` after waking up.

### DIFF
--- a/core/embed/sec/tropic/stm32/tropic01.c
+++ b/core/embed/sec/tropic/stm32/tropic01.c
@@ -139,6 +139,8 @@ lt_ret_t lt_port_deinit(lt_l2_state_t *s2) {
   HAL_GPIO_DeInit(TROPIC01_SPI_MOSI_PORT, TROPIC01_SPI_MOSI_PIN);
   HAL_GPIO_DeInit(TROPIC01_PWR_PORT, TROPIC01_PWR_PIN);
 
+  drv->initialized = false;
+
   return LT_OK;
 }
 


core/embed/sec/tropic/stm32/tropic01.c
```diff
@@ -139,6 +139,8 @@ lt_ret_t lt_port_deinit(lt_handle_t *h) {
   HAL_GPIO_DeInit(TROPIC01_SPI_MOSI_PORT, TROPIC01_SPI_MOSI_PIN);
   HAL_GPIO_DeInit(TROPIC01_PWR_PORT, TROPIC01_PWR_PIN);
 
+  drv->initialized = true;
```
TychoVrahe: false?

TychoVrahe (changes requested):

core/embed/sec/tropic/stm32/tropic01.c
onvej-sl: Fixed in https://github.com/trezor/trezor-firmware/pull/5868/commits/70dbae58f62bae640215ca9fdfd77196199ff31a.

TychoVrahe:
I am still kinda failing to understand what this solves though, the initialized flag is set to false anyway in `tropic_deinit` by this memset: `memset(drv, 0, sizeof(*drv));`.

onvej-sl:
> the initialized flag is set to false anyway in tropic_deinit by this memset: memset(drv, 0, sizeof(*drv));

This is right, however the driver in `tropic.c` and the driver in `tropic01.c` are two separate drivers. What happens after wakeup is that `tropic_init()` is called, which calls `lt_init()`, which in turn calls `lt_port_init()`. And `lt_port_init()` is skipped because the flag has not been cleared.

TychoVrahe:
ah right, makes sense now

PR #5869, opened 2025-09-25
chore(core): make `host_name` & `app_name` required also in `ThpCredentialMetadata`

romanz:
[Slack discussion](https://satoshilabs.slack.com/archives/C078GRAK58U/p1758798602530199)

>> How to handle when 3rd party does not provide either `HostName` or `AppName`?
>
> I think that currently it can happen only if Suite uses an "old" credential (generate by the current firmware), and tries to reuse it with the "new" firmware. In this case, `decode_credential()` will fail to decode (due to a missing required field), and the user will have to re-pair THP (similar to the case when a corrupted credential is used to connect).
This can be also resolved by clicking `Forget all`, to force THP re-pairing.

>> there could be environments where you cant really tell the machine name (browsers/webextensions)
>
> If the host name is unknown, we can set it to an empty string (instead of `None`), so the confirmation message will be `"Allow {0} to pair with this Trezor?"` instead of `"Allow {0} on {1} to pair with this Trezor?"`.
https://github.com/trezor/trezor-firmware/blob/01eb84edd2d78aa225487ccb9b0addcb004cf546/core/src/trezor/wire/thp/ui.py#L24


mmilata (approved):

romanz:
Tested both on BLE & USB.

PR #5870, opened 2025-09-25
Yet another redesign of the connected item

bieleluk:
- move hostname to the main text
  - break the words at the end of the 1st line
  - trim and add ellipsis at the end of the second line
- remove connection icon
- change subtext into connection status

HostInfoScreen
- add app name to host info screen

### DIFF
--- a/core/embed/rust/src/ui/display/font.rs
+++ b/core/embed/rust/src/ui/display/font.rs
@@ -342,13 +342,34 @@ impl FontInfo {
                     // Another character would not fit => split at the previous word boundary
                     return &text[0..prev_word_boundary];
                 }
-                if c == ' ' || c == ':' {
+                if c == ' ' {
                     prev_word_boundary = i;
                 }
                 text_width += c_width;
             }
             text // the whole text fits
         })
+    }
+
+    /// Get the longest prefix of a given `text` (breaking at letter boundaries)
+    /// that will fit into the area `width` pixels wide.
+    pub fn longest_prefix_break_words<'a>(&'static self, width: i16, text: &'a str) -> &'a str {
+        let mut prefix = text;
+        loop {
+            if self.text_width(prefix) <= width {
+                return prefix;
+            }
+            // remove exactly one UTF-8 char from the end
+            if let Some((i, _)) = prefix.char_indices().next_back() {
+                if i == 0 {
+                    return "";
+                }
+                debug_assert!(prefix.is_char_boundary(i));
+                prefix = &prefix[..i];
+            } else {
+                return ""; // empty string
+            }
+        }
     }
 
     /// Get the length of the longest suffix from a given `text`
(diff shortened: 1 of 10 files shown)


core/translations/en.json
```diff
@@ -109,6 +109,7 @@
     "bitcoin__unverified_external_inputs": "The transaction contains unverified external inputs.",
     "bitcoin__valid_signature": "The signature is valid.",
     "bitcoin__voting_rights": "Voting rights to",
+    "ble__app_connected": "App connected",
```
obrusvit: shouldn't it be "Apps connected"?

core/embed/rust/src/ui/display/font.rs
```diff
@@ -351,6 +351,18 @@ impl FontInfo {
         })
     }
 
+    /// Get the longest prefix of a given `text` (breaking at letter boundaries)
+    /// that will fit into the area `width` pixels wide.
+    pub fn longest_prefix_break_words<'a>(&'static self, width: i16, text: &'a str) -> &'a str {
+        for (i, _c) in text.char_indices() {
```
romanz: Here `i` iterates over the characters' indices in ascending order - but in the `if` below it is used to access the characters in reversed order (`text[..text.len() - i]`).

I guess it works for ASCII characters, but I am not sure it will also work for Unicode...

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
```diff
@@ -264,9 +264,10 @@ enum Subscreen {
     ),
 
     HostInfoScreen(
-        TString<'static>, /* host name */
-        TString<'static>, /* MAC address */
-        u8,               /* parent screen index */
+        TString<'static>,         /* host name */
```
romanz: (really small) nit: we should use a separate struct to allow naming the individual components of this screen (mainly from maintainability point of view).

Can be definitely done in a separate PR.

core/embed/rust/src/ui/layout_eckhart/component/button.rs
```diff
@@ -519,39 +518,96 @@ impl Button {
...
+                            )),
                         );
                         true
+                    } else if *break_words {
+                        let first = stylesheet
+                            .font
+                            .longest_prefix_break_words(available_width, t);
+
+                        if first == t {
+                            // The first line fits
```
romanz, line 540: It reminds me a similar logic in `TextLayout::layout_text()` - can we reuse it here?

(just curious - can be definitely done in a separate PR)

core/embed/rust/src/ui/display/font.rs
romanz: BTW, can we adapt `FontInfo::longest_prefix` to handle this case?

Maybe we can use a similar approach to https://doc.rust-lang.org/std/primitive.str.html#method.split

core/translations/en.json
bieleluk: Fixed
https://github.com/trezor/trezor-firmware/pull/5870/commits/a1ca010d14e71e51d400774001f8e5bcd4143c3c

core/embed/rust/src/ui/layout_eckhart/component/button.rs
bieleluk: It is definitely a good candidate for the `TextLayout`, and I created a PR for that: https://github.com/trezor/trezor-firmware/issues/5880. This is just a quick fix that should be merged ASAP.
The entire `button.rs` should use `TextLayout` instead of splitting lines manually.

core/embed/rust/src/ui/display/font.rs
bieleluk: I agree with you; this is not a particularly stable approach.  This is a temporary solution until the `Textlayout` is used there.

Furthermore, many more functions also rely on this assumption:`longest_prefix`, `split_two_lines`, `longest_suffix`.
I created an issue that should solve that at once: https://github.com/trezor/trezor-firmware/issues/5879

core/embed/rust/src/ui/layout_eckhart/firmware/device_menu_screen.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5870/commits/c216b2785a57d580be545ca1e2b10ff6fd9df6ee

core/embed/rust/src/ui/display/font.rs
romanz: > Here `i` iterates over the characters' indices in ascending order - but in the `if` below it is used to access the characters in reversed order (`text[..text.len() - i]`).

WDYT about:
```rust
    /// Get the longest prefix of a given `text` (breaking at letter boundaries)
    /// that will fit into the area `width` pixels wide.
    pub fn longest_prefix_break_words<'a>(&'static self, width: i16, text: &'a str) -> &'a str {
        let mut prefix = text;
        while !prefix.is_empty() {
            if self.text_width(prefix) <= width {
                return prefix;
            }
            let mut prefix_len = prefix.len();
            // NOTE: replace with floor_char_boundary when stable
            while !prefix.is_char_boundary(prefix_len) {
                prefix_len -= 1;
            }
            prefix = &prefix[..prefix_len];
        }
        ""
    }
```

romanz: Please also add a unit-test for the new function.

romanz (approved):

core/embed/rust/src/ui/display/font.rs
bieleluk: Your proposed change would have gotten stuck in an infinite loop, so I had to tweak it a bit:
https://github.com/trezor/trezor-firmware/pull/5870/commits/3968296fa3655832cdcb0f35fd009459c0309424

core/embed/rust/src/ui/display/font.rs
```diff
@@ -354,13 +354,22 @@ impl FontInfo {
...
+            if self.text_width(prefix) <= width {
+                return prefix;
+            }
+            // remove exactly one UTF-8 char from the end
+            if let Some((i, _)) = prefix.char_indices().next_back() {
+                if i == 0 {
+                    return "";
+                }
+                prefix = &prefix[..i];
+                debug_assert!(prefix.is_char_boundary(i));
```
romanz, line 367: IIUC, the `debug_assert!` should happen before `prefix[..i]` slicing operation - right?

core/embed/rust/src/ui/display/font.rs
```diff
@@ -413,3 +422,66 @@ impl GlyphMetrics for Font {
...
+        assert_eq!(font.longest_prefix_break_words(w_h, text), "H");
+
+        // Just below "Hello" fits "Hell"
+        assert_eq!(font.longest_prefix_break_words(w_hello - 1, text), "Hell");
+        assert_eq!(font.longest_prefix_break_words(w_hello, text), "Hello");
+
+        // Exact full width and bigger should return the whole string
+        assert_eq!(font.longest_prefix_break_words(w_full, text), text);
+        assert_eq!(
+            font.longest_prefix_break_words(w_full.saturating_add(999), text),
```
romanz: nit: is there any reason for using 999 here (and 500 below)?

IIUC, `w_full + 1` should also work - right?

romanz (approved):
Thanks!

core/embed/rust/src/ui/display/font.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5870/commits/2eae69901d4c45164586092e952466b5b608faf9

core/embed/rust/src/ui/display/font.rs
bieleluk: https://github.com/trezor/trezor-firmware/pull/5870/commits/2eae69901d4c45164586092e952466b5b608faf9

PR #5873, opened 2025-09-25
fix(prodtest): fix unhandled return code of secret_write()

cepetr:
This PR fixes an unhandled return code of `secret_write()` in the `secret-certdev-write` command, which caused misleading behavior when writing a certificate a second time, since `secret_write()` always fails in that case.


TychoVrahe:
duplicate: https://github.com/trezor/trezor-firmware/pull/5872

PR #5876, opened 2025-09-25
feat(core/eckhart): LED effect in HoldToConfirm

obrusvit:
This PR adds `Finalizing` phase to `HoldToConfirmAnim` for Eckhart UI. The controlling component `ActionBar` was adjusted accordingly. 

The Finalizing phase turns on the LED, shows the `ICON_DONE`, progressively dims the border and icon. `ActionBar` only waits for the result in this phase.

In the process, I slightly refactored the code, especially the animation state is now contained in one enum.

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/hold_to_confirm.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/hold_to_confirm.rs
@@ -2,9 +2,9 @@ use crate::{
     strutil::TString,
     time::{Duration, Stopwatch},
     ui::{
-        component::{Component, Event, EventCtx, Never},
+        component::{Component, Event, EventCtx},
         display::Color,
-        geometry::{Offset, Rect},
+        geometry::{Alignment2D, Insets, Offset, Rect},
         lerp::Lerp,
         shape::{self, Renderer},
     },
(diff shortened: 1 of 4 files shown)


core/embed/rust/src/ui/layout_eckhart/firmware/hold_to_confirm.rs
```diff
@@ -34,16 +37,30 @@ pub struct HoldToConfirmAnim {
...
+pub enum HoldToConfirmMsg {
+    /// The hold to confirm action was completed
+    Finalized,
+}
+
+enum AnimState {
+    Idle,
+    /// Growing border from start
+    Growing {
+        timer: Stopwatch,
```
bieleluk: nit: It might be a bit confusing to use the name `timer` for the `Stopwatch` struct when there is another struct `Timer` with a different functionality.

obrusvit: https://github.com/trezor/trezor-firmware/pull/5876/commits/c95f7fd65d28ba2a170e3a522e476c1d1b08567a

you're right, we can even remove the `timer` from `struct HoldToConfirmAnim`

PR #5881, opened 2025-09-26
chore(core/eckhart): use marquee only for the button subtext rendering

bieleluk:
Continuation of https://github.com/trezor/trezor-firmware/pull/5810


romanz (approved):
utACK

PR #5892, opened 2025-09-26
Avoid `ThpTransportBusy` error when pairing with a locked device

romanz:

### DIFF
--- a/core/src/trezor/wire/thp/received_message_handler.py
+++ b/core/src/trezor/wire/thp/received_message_handler.py
@@ -75,6 +75,7 @@ async def handle_received_message(channel: Channel) -> bool:
         await channel.iface_ctx.write_error(
             channel.get_channel_id_int(), ThpErrorType.DEVICE_LOCKED
         )
+        channel.clear()
     return False
 
 


romanz:
Tested with Suite Desktop:
image

romanz:
CC: @marekrjpolak @szymonlesisz

marekrjpolak:
@romanz I can confirm this works as expected.

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -224,29 +225,6 @@ async def _handle_pairing(ctx: Channel) -> None:
     await ctx.connection_context.handle(message)
 
 
-def _should_have_ctrl_byte_encrypted_transport(ctx: Channel) -> bool:
```
romanz, line 227: Note to reviewers: before c32002082225d3eab6e74670000daba62242729d this method was used to make sure the control byte matches the channel state - and was replaced by the `recv_payload(control_byte.is_encrypted_transport)` call:
https://github.com/trezor/trezor-firmware/blob/c32002082225d3eab6e74670000daba62242729d/core/src/trezor/wire/thp/channel.py#L288-L300

M1nd3r (approved):
utACK

PR #5901, opened 2025-09-30
Notify host about channel replacement

M1nd3r:


romanz:
Just from curiousity - how will it be used by Suite?

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -177,6 +177,9 @@ async def _handle_state_handshake(
             if paired:
                 trezor_state = _TREZOR_STATE_PAIRED
                 ctx.credential = credential
+                if ctx.is_channel_to_replace():
+                    # When replacing existing channel, user confirmation is not needed
+                    trezor_state = _TREZOR_STATE_PAIRED_AUTOCONNECT
```
romanz, line 182: Will it also be sent to the client in case of an "auto-connect" credential?
If not - maybe the constant (here & in `trezorlib`) should be renamed?

szymonlesisz:
@romanz , yes, we are already waiting for it here https://github.com/trezor/trezor-suite/pull/22029

@M1nd3r tested with emulator, works as expected :+1:

PR #5906, opened 2025-09-30
introduce BLE disable/enable

TychoVrahe:


core/embed/rust/src/trezorhal/ble/micropython.rs
```diff
@@ -424,4 +439,17 @@ pub static mp_module_trezorble: Module = obj_module! {
     ///     Raises exception if BLE driver reports an error.
     ///     """
     Qstr::MP_QSTR_reject_pairing => obj_fn_0!(py_reject_pairing).as_obj(),
+
+    /// def get_enabled(bool):
```
obrusvit:
```suggestion
    /// def set_enabled(bool):
```

core/embed/io/ble/unix/ble.c
```diff
@@ -58,3 +58,7 @@ uint8_t ble_get_bond_list(bt_le_addr_t *bonds, size_t count) { return 0; }
 void ble_set_high_speed(bool enable){};
 
 void ble_notify(const uint8_t *data, size_t len){};
+
+void ble_set_enabled(bool enabled) {}
+
+bool ble_get_enabled(void) { return true; }
```
obrusvit:
```suggestion
bool ble_get_enabled(void) { return false; }
```

obrusvit (commented):
Two comments for now.

core/embed/rust/src/trezorhal/ble/micropython.rs
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/108fdb3db125566c27184389ab4360172b4a745e

core/embed/io/ble/unix/ble.c
TychoVrahe: fixed https://github.com/trezor/trezor-firmware/commit/108fdb3db125566c27184389ab4360172b4a745e

obrusvit (approved):

PR #5908, opened 2025-09-30
chore(core): fix type hints for logging functions

M1nd3r:


core/src/trezor/log.py
```diff
@@ -6,7 +6,7 @@
     from typing import Any
 
 
-def _no_op(*args: Any, **kwargs: Any) -> None:
+def _no_op(name: str, msg: str, *args: Any, iface: WireInterface | None = None) -> None:
```
romanz, line 9: nit (for consistency with `core/embed/rust/src/micropython/logging.rs`):
```suggestion
def _no_op(name: str, msg: str, *args: Any, *, iface: WireInterface | None = None) -> None:
```

M1nd3r: Actually, with added `*, `, I am getting the following error(s) (locally):
```
[BLACK]
error: cannot format core/src/trezor/log.py: Cannot parse for target version Python 3.13: 9:44: def _no_op(name: str, msg: str, *args: Any, *, iface: WireInterface | None = None) -> None:
```
Hints:
- Keyword-only parameter separator not allowed after '*' parameter
- Keyword-only argument separator not allowed after '*' parameter


So I will keep it as-is (the `trezorlog.pyi` is not formatted, so it is not affected).

PR #5910, opened 2025-09-30
Fix disconnection while waiting for the bootloader

romanz:
Tested locally:
```
$ TREZOR_MODEL=T3T1 QUIET_MODE=1 make -C core build_firmware upload
...
Waiting for bootloader...
Failed to close <usb1.USBDeviceHandle object at 0x7254f9adbb10>: LIBUSB_ERROR_NO_DEVICE [-4]
Uploading  [####################################]  100%
make: Leaving directory '/home/rzeyde/src/trezor-firmware/core'
```

### DIFF
--- a/python/src/trezorlib/cli/__init__.py
+++ b/python/src/trezorlib/cli/__init__.py
@@ -203,11 +203,20 @@ class TrezorConnection:
         )
         return session
 
-    def get_transport(self) -> "Transport":
+    def get_transport(self, _clear_cache: bool = False) -> "Transport":
         global _TRANSPORT
         if _TRANSPORT is not None:
-            return _TRANSPORT
+            if not _clear_cache:
+                return _TRANSPORT
 
+            # remove previously cached transport
+            try:
+                atexit.unregister(_TRANSPORT.close)
+                _TRANSPORT.close()
+            except Exception as e:
+                self._print_exception(e, "Failed to close transport")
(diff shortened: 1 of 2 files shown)


python/src/trezorlib/cli/__init__.py
```diff
@@ -203,10 +203,16 @@ def get_session(
...
         global _TRANSPORT
         if _TRANSPORT is not None:
-            return _TRANSPORT
+            if not _clear_cache:
+                return _TRANSPORT
+
+            # remove previously cached transport
+            atexit.unregister(_TRANSPORT.close)
+            _TRANSPORT.close()
+            _TRANSPORT = None
```
mmilata: Nit: this could prevent infinite loop in case of misbehaving transport (which probably shouldn't be the case with our current transports). Alternatively the bootloader reconnect loop could have a limit/timeout.
```suggestion
            try:
                atexit.unregister(_TRANSPORT.close)
                _TRANSPORT.close()
            except Exception as e:
                _print_exception(e)
            finally:
                _TRANSPORT = None
```

romanz: Thanks! https://github.com/trezor/trezor-firmware/pull/5910/commits/739120579d887d9eb16d7b95f4bd8bf614d40f2a

mmilata (approved):

PR #5913, opened 2025-10-01
[Eckhart]: Update chunkfied text style

bieleluk:
- add starting ellipsis to Eckhart mono fonts: [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=10052-8010&t=xmQcMyYNLhTLkO4c-0)
- when the address doesn't fit on one screen, add the hint with the counter
- update spacing between chunks and lines: [Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=6191-1259&t=xmQcMyYNLhTLkO4c-0)


core/embed/rust/src/ui/layout_eckhart/firmware/text_screen.rs
```diff
@@ -167,24 +174,38 @@ where
...
 
-        // Introduce side insets + top padding if the header is not present
-        content_area = if self.header.is_none() {
-            content_area.inset(CONTENT_INSETS_NO_HEADER)
-        } else {
-            content_area.inset(SIDE_INSETS)
+        let compute_content_area = |area: Rect, has_header: bool| {
+            let mut area = area.inset(SIDE_INSETS);
+            if !has_header {
+                area = area.inset(CONTENT_INSETS_NO_HEADER);
```
obrusvit: `CONTENT_INSETS_NO_HEADER` contain also `PADDING` from the sides so this is probably wrong.

bieleluk: https://github.com/trezor/trezor-firmware/pull/5913/commits/9a41b9de330ca90335b5720fd47fef42cbbbfa56

bieleluk: Had to fix it once again.
The `CONTENT_INSETS_NO_HEADER` also includes the bottom part for the action bar, which was already used in the previous step. The correct closure is:

```
        let compute_content_area = |area: Rect, has_header: bool| {
            let mut area = area.inset(SIDE_INSETS);
            if !has_header {
                area = area.inset(Insets::top(CONTENT_INSETS_NO_HEADER.top));
            }
            area
        };
```

Issue #5801, opened 2025-09-18
AuthenticateDevice device test with Tropic


andrewkozlik:
At present the `AuthenticateDevice` device test only checks the signature coming from Optiga. For T3W1 it should ideally also check the signature coming from Tropic. However, the Trezor emulator is built by default with `DISABLE_TROPIC=1`, which is probably intentional, because if the Tropic emulator were needed for the device tests, then devs would need to run it when running device tests.

So we need to:
1. Decide on whether to use the Tropic emulator for Trezor emulator in device tests or some alternative, like mocking the Tropic functions.
2. Initialize the Tropic emulator with a device key and device certificate or set up the mock functions with an emulator key and certificate.
3. Check the Tropic signature and certificate chain in the `AuthenticateDevice` device test.

matejcik:
i'll note here that in the past we had to provision the emulator with a patched build of staging HSM
i believe that step is unnecessary -- we can instead generate a non-secret private key, put its public key into `authenticity-dev.json`, and sign the emulator certificate with that. much easier than pulling hardware out of the server room

PR #5916, opened 2025-10-01
Test Tropic authenticity check

ibz:
* we have the following cases:
  1. default build (`DISABLE_TROPIC=1`) => test expects some certificates to be returned by Tropic, so it will fail
  2. build with Tropic (`DISABLE_TROPIC=0`) but the Tropic model is not running => test will fail `ProcessError: Tropic signing failed.`
  3. build with Tropic (`DISABLE_TROPIC=0`) and the Tropic model is running => test will verify the certificate chain and signature

As pointed in the comments, we should probably invert the `DISABLE_TROPIC` flag (so we build with Tropic by default) and start `ts-tvl` always as part of starting the Trezor emulator. Will do that in a future PR. For now let's stay consistent with what we have.

About @matejcik 's [comment](https://github.com/trezor/trezor-firmware/issues/5801#issuecomment-3306102089) - that part is not addressed in this PR at all. Will address the `trezorlib` part after this.


.github/workflows/core.yml
```diff
@@ -297,7 +298,14 @@ jobs:
           path: core/build
       - run: chmod +x core/build/unix/trezor-emu-core*
       - uses: ./.github/actions/environment
+      - name: Start Tropic model
+        if: ${{ env.TREZOR_MODEL == 'T3W1' }}
+        run: nix-shell --arg fullDeps true --run "cd vendor/ts-tvl && mkdir -p logs && poetry env use 3.12 && poetry install && poetry run model_server tcp -c ../../tests/tropic_model.yml >logs/server.log 2>&1 &"
```
mmilata: Nit: we depend on ts-tvl in top-level pyproject.toml so following should work fine. Eventually I'd like emu.py to launch Tropic emulator automatically.
```suggestion
        run: nix-shell --arg fullDeps true --run "cd vendor/ts-tvl && mkdir -p logs && uv run model_server tcp -c ../../tests/tropic_model.yml >logs/server.log 2>&1 &"
```

tests/device_tests/test_authenticate_device.py
```diff
@@ -46,48 +86,47 @@ def test_authenticate_device(session: Session, challenge: bytes) -> None:
...
-            cert.tbs_certificate_bytes,
-            cert.signature_algorithm_parameters,
-        )
+    certs = [x509.load_der_x509_certificate(cert) for cert in proof.tropic_certificates]
 
-    # Verify that the common name matches the Trezor model.
-    common_name = cert.subject.get_attributes_for_oid(x509.oid.NameOID.COMMON_NAME)[0]
-    assert common_name.value.startswith(session.model.internal_name)
+    if session.model.internal_name in MODELS_WITHOUT_TROPIC:
+        pytest.xfail(f"{session.model.internal_name} does not support Tropic")
```
mmilata: Seems to me the test should rather succeed in this case. But I'd rather split it in two testcases and use pytest.mark.models to determine which to run.

mmilata (commented):
> where is a better place to store the generated config used to provision the model

I'd create directory under tests, something like `tests/tropic_model`.

> should we store the script used to generate the config (see below) somewhere instead?

`core/tools` probably

.github/workflows/core.yml
ibz: > Eventually I'd like emu.py to launch Tropic emulator automatically.

Maybe this is the way to go eventually. Would also solve @andrewkozlik 's dislike for the way we do things now. Basically it would mean... if you use a T3W1 emulator it always includes Tropic, there aren't two ways of doing things. Guess I was just too afraid to disrupt others back when I introduced Tropic, but now that the thing has matured, it should just be on by default. Thanks for mentioning this!

ibz: > following should work fine

Seems that `ts-tvl`'s `pyproject.toml` is missing `[project]` which makes it incompatible with `uv` :(

tests/device_tests/test_authenticate_device.py
ibz: Looks much cleaner indeed, thank you! 974d29c7e347b2c1bf6991551255401041c87edf

.github/workflows/core.yml
mmilata: Right, sorry, you need to run `uv` in repo top-level, but no idea how to tell `uv run` to run `cd vendor/ts-tvl && model_server ...`

mmilata: > you use a T3W1 emulator it always includes Tropic

Yeah, sounds like something we should do unless it significantly slows down unrelated tests. I may look into this as part of my BLE emulator work which is kinda same but also kinda different in some regards.

.github/workflows/core.yml
```diff
@@ -297,7 +298,18 @@ jobs:
           path: core/build
       - run: chmod +x core/build/unix/trezor-emu-core*
       - uses: ./.github/actions/environment
+      - name: Start Tropic model
+        if: ${{ env.TREZOR_MODEL == 'T3W1' }}
+        run: |
+          mkdir -p vendor/ts-tvl/logs && cp vendor/ts-tvl/model_configs/example_config/tropic01_ese_* tests/tropic_model
+          python core/tools/generate_tropic_model_config.py -k tests/tropic_model/tropic_key.pem -c tests/tropic_model/tropic_cert.pem -r tests/tropic_model/root_cert.pem > tests/tropic_model/config.yml
```
mmilata: If the generated config is always the same, can we commit it to the repo and hook it up to the `make gen` and `make gen_check`? Or does the emulator require absolute file paths?

.github/workflows/core.yml
```diff
@@ -297,7 +298,18 @@ jobs:
...
+          nix-shell --arg fullDeps true --run "cd vendor/ts-tvl && poetry env use 3.12 && poetry install && poetry run model_server tcp -c ../../tests/tropic_model/config.yml >logs/server.log 2>&1 &"
```
mmilata: For `test_emu_ui_multicore` we get one Tropic emulator for multiple Trezor emulators, can this cause any problems?

.github/workflows/core.yml
```diff
@@ -297,7 +298,18 @@ jobs:
...
       - run: nix-shell --run "uv run make -C core ${{ env.ACTIONS_DO_UI_TEST == 'true' && 'test_emu_ui_multicore' || 'test_emu' }}"
+      - name: Show Tropic model logs
+        if: ${{ env.TREZOR_MODEL == 'T3W1' }}
+        run: |
+          echo "==== MODEL SERVER LOG ===="
+          cat vendor/ts-tvl/logs/server.log
```
mmilata: I'd rather ask the TS emulator to log into `tests/trezor-tvl.log`, then the `upload-artifact` job below will match the file name and you can download it in the archive with other logs.

tests/tropic_model/root_cert.pem
```diff
@@ -0,0 +1,9 @@
+-----BEGIN CERTIFICATE-----
```
mmilata, line 1: Can you perphaps also include short README in the directory with a paragraph describing what are these files, and the openssl commands used to generate them that are in the top post of this PR? References to Tropic docs appreciated.

mmilata (commented):
Minor comments, will continue tomorrow.

core/tools/generate_tropic_model_config.py
```diff
@@ -0,0 +1,92 @@
...
+    root_cert_der_bytes = root_cert.public_bytes(serialization.Encoding.DER)
+
+    # certificate chain with the length prefix
+    all_cert_bytes = (
+        (len(tropic_cert_der_bytes) + len(root_cert_der_bytes)).to_bytes(2)
+        + tropic_cert_der_bytes
+        + root_cert_der_bytes
+    )
+
+    SLOT_LEN = 444
```
mmilata, line 75: FWIW ts-tvl master has this = 475. But it doesn't really matter if all certs fit in 3 slots.

https://github.com/tropicsquare/ts-tvl/commit/700c83db256f29ca32054bbcd167ff96c0bc59af

core/tools/generate_tropic_model_config.py
```diff
@@ -0,0 +1,92 @@
...
+    tropic_cert_der_bytes = tropic_cert.public_bytes(serialization.Encoding.DER)
+
+    root_cert = x509.load_pem_x509_certificate(root_cert.read())
+    root_cert_der_bytes = root_cert.public_bytes(serialization.Encoding.DER)
+
+    # certificate chain with the length prefix
+    all_cert_bytes = (
+        (len(tropic_cert_der_bytes) + len(root_cert_der_bytes)).to_bytes(2)
+        + tropic_cert_der_bytes
+        + root_cert_der_bytes
```
mmilata, line 72: Optiga currently uses 3 level PKI so there's also additional intermediate CA certificate. I'm assuming it will be the same with Tropic (but maybe let's confirm with someone). Their pubkeys would then go in `authenticity-dev.json` under `ca_pubkeys` (or similar list but for Tropic).

.github/workflows/core.yml
ibz: Much better! 1d88a3f83b80985ae2ae54eca47fe79d84f745b6

.github/workflows/core.yml
ibz: Looks much better! Thanks for pointing! 75dc55f5d41e02c817859a05720c668b43cd929f

.github/workflows/core.yml
ibz: Sure it did, running them separately now, see this: 644ca7c66b38eaa071d5daf37f7357bfcf1ad8ba

core/tools/generate_tropic_model_config.py
ibz: Interesting. OK. Actually I think the test will fail once we update to `master` then. Because of this check: https://github.com/trezor/trezor-firmware/blob/main/core/embed/sec/tropic/tropic.c#L616

I'll update to `master` now and add a comment here to point to the constant in `libtropic`!

mmilata: Ok, this is probably a bit trickier as it depends on Tropic firmware version, see also tropicsquare/libtropic#301. So let's keep the value consistent with libtropic which now has 444.

python/src/trezorlib/_internal/emulator.py
```diff
@@ -180,6 +188,8 @@ def start(
                 # process is running, no need to start again
                 return
 
+        self.start_tropic_model()
```
mmilata, line 282: We should only do this for T3W1.

Can you please put all the tropic model handling to a separate class not descended from Emulator so that it's clearly separate?

mmilata: Also we should make sure that emu.py works with emulator built with DISABLE_TROPIC=1 even if tropic model fails to start 🤔

core/tools/generate_tropic_model_config.py
```diff
@@ -0,0 +1,160 @@
...
+        "s_t_priv": "tropic01_ese_private_key_1.pem",
+        "s_t_pub": "tropic01_ese_public_key_1.pem",
+        "x509_certificate": "tropic01_ese_certificate_1.pem",
+        "debug_random_value": b"\x00\xc0\xff\xee",
+        "r_user_data": user_data,  # certificate chain
+        "r_ecc_keys": {  # signing key at index 0
+            0: {
+                "a": tropic_a,
+                "s": tropic_s,
+                "prefix": tropic_prefix,
```
mmilata, line 105: I suppose it might be worth opening an issue in ts-tvl to either make loading privkeys easier (like `s_t_priv`) or at least document the key preparation in `tvl/targets/model/README.md` ...

ibz: Yep: https://github.com/tropicsquare/ts-tvl/issues/4

python/src/trezorlib/_internal/emulator.py
ibz: > make sure that emu.py works with emulator built with DISABLE_TROPIC=1 even if tropic model fails to start 🤔

`emu.py` does not currently start the Tropic model, although we should make it do that soon, I suppose. Note that it instantiates `CoreEmulator` without passing `launch_tropic_model` so it defaults to `False`.

ibz: > Can you please put all the tropic model handling to a separate class not descended from Emulator so that it's clearly separate?

Done: 1a485fa1ba44dc67d6ac173d673adf40c30bd323

tests/tropic_model/root_cert.pem
ibz: Yep: 7d7ee01b3ac858dab5055e4efe7dd89dc9ee448f

core/embed/projects/prodtest/main.c
```diff
@@ -191,9 +191,13 @@ static void drivers_init(void) {
   ble_init();
 #endif
 #ifdef USE_TROPIC
+#if TREZOR_EMULATOR
+  tropic_init(28992);
+#else
   tropic_init();
   tropic_wait_for_ready();
 #endif
```
mmilata: Not sure
```suggestion
#endif
  tropic_wait_for_ready();
```

ibz: :100:

ibz:
Another issue I found in `ts-tvl`while working on this PR: https://github.com/tropicsquare/ts-tvl/issues/5

PR #5919, opened 2025-10-01
feat(core/eckhart): remove fuel gauge from homescreen

obrusvit:
[no changelog]


obrusvit:
This will be reworked differently.

PR #5922, opened 2025-10-02
feat(core+python): support device unlocking during THP handshake

romanz:
Add a flag to the first THP handshake message, so the device can:

- fail the handshake if the device is locked (current behavior)
- prompt the user to enter the PIN (similar to other non-pairing flows)

:red_circle: This is a backward-incompatible change, so Suite THP implementation **MUST** be adapted as well :red_circle: 
:point_right: https://github.com/trezor/trezor-suite/pull/22146

In case an old host performs a handshake with a new device, the device will ignore the message and emit the following log:
```
trezor.wire.thp.received_message_handler ERROR Message received is not a valid handshake init request: 32 bytes
```

- the old hanshake init request is ***32*** bytes long
- the new hanshake init request is ***33*** bytes long

### DIFF
--- a/core/src/trezor/wire/thp/received_message_handler.py
+++ b/core/src/trezor/wire/thp/received_message_handler.py
@@ -89,19 +89,48 @@ async def _handle_state_handshake(
 
     payload = await ctx.recv_payload(control_byte.is_handshake_init_req)
 
-    if len(payload) != PUBKEY_LENGTH:
-        log.error(__name__, "Message received is not a valid handshake init request!")
+    if len(payload) != PUBKEY_LENGTH + 1:
+        if __debug__:
+            log.error(
+                __name__,
+                "Message received is not a valid handshake init request: %d bytes",
+                len(payload),
+            )
         return
 
-    if not config.is_unlocked():
+    host_ephemeral_public_key = payload[:PUBKEY_LENGTH]
+    # show the PIN keyboard to allow the user to unlock the device
+    try_to_unlock = payload[PUBKEY_LENGTH] & 0x01 == 1
+
+    async def _check_unlocked() -> None:
+        if config.is_unlocked():
+            return
+
+        if try_to_unlock:
+            from trezor import workflow
+
+            from apps.common.lock_manager import unlock_device
+
+            # Register the unlock prompt with the workflow management system
+            # (in order to avoid immediately respawning the lockscreen task)
+            try:
+                return await workflow.spawn(unlock_device())
+            except Exception as e:
+                if __debug__:
+                    log.exception(__name__, e)
+
+        # Fail pairing if still locked
         raise ThpDeviceLockedError
+
+    await _check_unlocked()
 
     handshake = Handshake()
 
     trezor_ephemeral_public_key, encrypted_trezor_static_public_key, tag = (
         handshake.handle_th1_crypto(
             get_encoded_device_properties(ctx.iface),
-            host_ephemeral_public_key=payload,
+            host_ephemeral_public_key=host_ephemeral_public_key,
+            payload=payload[PUBKEY_LENGTH:],
         )
     )
 
@@ -130,8 +159,7 @@ async def _handle_state_handshake(
     # will be `None` on USB interface, to be ignored by `cache_host_info()`
     mac_addr: AnyBytes | None = ctx.iface_ctx.connected_addr()
 
-    if not config.is_unlocked():
-        raise ThpDeviceLockedError
+    await _check_unlocked()
 
     host_encrypted_static_public_key = payload[: KEY_LENGTH + TAG_LENGTH]
     handshake_completion_request_noise_payload = payload[KEY_LENGTH + TAG_LENGTH :]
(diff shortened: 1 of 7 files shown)


romanz:
CC: @marekrjpolak

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -89,19 +89,46 @@ async def _handle_state_handshake(
...
+            from trezor import workflow
+
+            from apps.common.lock_manager import unlock_device
+
+            # Register the unlock prompt with the workflow management system
+            # (in order to avoid immediately respawning the lockscreen task)
+            try:
+                return await workflow.spawn(unlock_device())
+            except Exception as e:
+                log.exception(__name__, e)
```
romanz: add `if __debug__`:

romanz:
Tested `38554c20b5` locally on a locked device using this PR's `trezorctl`:
```
$ trezorctl get-features
(PIN entry dialog is shown on device)
Features (257 bytes) {
...
}
```

If the user cancels PIN entry, the pairing fails with:
```
$ trezorctl get-features
Failed to connect: DeviceLocked
```

core/src/trezor/wire/thp/received_message_handler.py
romanz: Fixed in 38554c20b5.

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -89,19 +89,47 @@ async def _handle_state_handshake(
...
+                __name__,
+                "Message received is not a valid handshake init request: %d bytes",
+                len(payload),
+            )
         return
 
-    if not config.is_unlocked():
+    host_ephemeral_public_key = payload[:PUBKEY_LENGTH]
+    # a non-zero value will show the PIN keyboard to allow the user to unlock the device
+    try_to_unlock = payload[PUBKEY_LENGTH] != 0
```
M1nd3r: Maybe change this to 
```suggestion
    try_to_unlock = payload[PUBKEY_LENGTH] & 0x1 == 1
```

so that in the future we can use the remaining 7 bits for other purpose/ other flags

romanz: Sounds good - changed in 75011c118 + 835e435a9e.

romanz:
[Squashing](https://github.com/trezor/trezor-firmware/compare/ac3a7c625f4a48bf5d2a540c816b842d3f2bd3ef..0540693f29b93cf31feed65833b958edcf15333e) into a single commit.

core/src/trezor/wire/thp/received_message_handler.py
```diff
@@ -89,19 +89,48 @@ async def _handle_state_handshake(
...
+
+        if try_to_unlock:
+            from trezor import workflow
+
+            from apps.common.lock_manager import unlock_device
+
+            # Register the unlock prompt with the workflow management system
+            # (in order to avoid immediately respawning the lockscreen task)
+            try:
+                return await workflow.spawn(unlock_device())
```
mmilata, line 117: `_check_unlocked` returns `None` but `workflow.spawn()` returns class `spawn`, interesting that pyright doesn't mind

tests/device_tests/thp/test_handshake.py
```diff
@@ -49,3 +50,68 @@ def test_handshake(client: Client) -> None:
...
+
+    protocol._do_channel_allocation()
+    protocol._init_noise(
+        randomness_static=randomness_static,
+    )
+    # the device should show the PIN keyboard
+    protocol._send_handshake_init_request(try_to_unlock=True)
+    protocol._read_ack()
+    debug.synchronize_at("PinKeyboard")
+    debug.input(PIN4)
```
mmilata, line 90: Nit: testcase with wrong pin would be nice.

mmilata:
Btw have you considered bumping the version in `ThpDeviceProperties`? May on may not help Suite.

core/src/trezor/wire/thp/received_message_handler.py
romanz: `workflow.spawn(unlock_device())` returns a `spawn`, but `await`ing it should result in a `None` - right?

tests/device_tests/thp/test_handshake.py
romanz: Added in 65905f5c41

romanz:
> Btw have you considered bumping the version in `ThpDeviceProperties`? May on may not help Suite.

Not sure, since the old handshake format will not be released...
@marekrjpolak WDYT?

mmilata:
Fair enough, unless we want to test 2.9.2->2.9.3 upgrade on production hardware.

core/src/trezor/wire/thp/received_message_handler.py
mmilata: ah right, makes sense

mmilata (approved):

romanz:
Squashed into a single commit - let's merge it together with https://github.com/trezor/trezor-suite/pull/22146.

obrusvit (approved):
Let's merge.

PR #5924, opened 2025-10-02
chore: reserve message type range

M1nd3r:


common/protob/messages.proto
```diff
@@ -183,7 +183,7 @@ enum MessageType {
     MessageType_LiskGetPublicKey = 121 [(wire_in) = true];
     MessageType_LiskPublicKey = 122 [(wire_out) = true];
     */
-    reserved 114 to 122;
```
andrewkozlik, line 188: We should explicitly reserve 123-149 and add a comment saying that it's for internal use, I'm not sure how much detail we want to give, but I think the usage is nothing secret. Whenever we rebase the internal branch over main we will remove the `reserved 123 to 149;` line.

M1nd3r: My idea was to give as little detail as possible. Nevertheless, I was also thinking whether to have separate reserved statements.

Adjusted here to separate statements with "reserved for internal use" comment: https://github.com/trezor/trezor-firmware/pull/5924/commits/a2204eec4fab654a62df1df8b4e0b0d8ed447673

PR #5925, opened 2025-10-02
Adjust page counter position

ibz:

### DIFF
--- a/core/embed/rust/src/ui/layout_eckhart/firmware/hint.rs
+++ b/core/embed/rust/src/ui/layout_eckhart/firmware/hint.rs
@@ -36,8 +36,6 @@ impl<'a> Hint<'a> {
     pub const HEIGHT_MINIMAL: i16 = 40;
     /// height of the multi line component [px]
     pub const HEIGHT_MAXIMAL: i16 = 66;
-    /// margins from the edges of the screen [px]
-    const HINT_INSETS: Insets = Insets::new(16, 24, 24, 24);
 
     fn from_content(content: HintContent<'a>) -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


core/embed/rust/src/ui/layout_eckhart/firmware/hint.rs
```diff
@@ -117,12 +123,18 @@ impl<'a> Component for Hint<'a> {
         debug_assert!(bounds.width() == screen().width());
         debug_assert!(bounds.height() == self.height());
 
+        let insets = if matches!(self.content, HintContent::PageCounter(_)) {
```
obrusvit: Please do `let insets = self.content.insets()` and make the distinction internally in the `HintContent`

obrusvit (changes requested):

core/embed/rust/src/ui/layout_eckhart/firmware/hint.rs
ibz: d292299e1bef723f00056997546b783b8dea4fa6

core/embed/rust/src/ui/layout_eckhart/firmware/hint.rs
```diff
@@ -106,7 +104,8 @@ impl<'a> Hint<'a> {
     /// Returns the height of the content including padding. In case of the
     /// instruction, the height is calculated based on the text length.
     pub fn height(&self) -> i16 {
-        self.content.height() + Self::HINT_INSETS.top + Self::HINT_INSETS.bottom
+        let insets = self.content.insets();
+        self.content.height() + insets.top + insets.bottom
```
obrusvit, line 108: And now you can adjust the content.height() function to use the internal Insets constant instead of doing it here :)

PR #5928, opened 2025-10-02
feat(core): allow exporting unit serial number on T3W1

romanz:
[Figma](https://www.figma.com/design/Dkl7W5PLqpr3TnXJ5hbnfd/--Future-Prod---Safe-7?node-id=14047-44614&t=E8Bd3B1sghxiRzfm-4)

It can be exported using CLI:
```
$ trezorctl device serial-number
12345678901234
```


obrusvit (commented):
I think we want to gate this feature for T3W1.

romanz:
> I think we want to gate this feature for T3W1.

Sounds good - https://github.com/trezor/trezor-firmware/pull/5928/commits/1e4449762184b6bda60233b24c9c45ecf61a427d

Is it OK to gate only the `GetSerialNumber` request, or also the rest of the implementation?

romanz:
Will gate also the new C & Python modules.

core/.changelog.d/5922.added
```diff
@@ -0,0 +1 @@
+Allow exporting device serial number.
```
mmilata: Nit

```suggestion
[T3W1] Allow exporting device serial number.
```

mmilata (approved):
LGTM

core/.changelog.d/5922.added
romanz: 170e0d9f7ad6d63bea20f9fa41014ef3ea4cd743

obrusvit (approved):

PR #5934, opened 2025-10-03
Proper rendering of default ellipsis

bieleluk:
- proper rendering of default leading ellipsis
- do not show incomplete chunks
- proper evaluation of the last chunk
- render trailing ellipsis within the text bound


obrusvit (commented):
Please add changelog entry to core `5934.fixed`

```
Fix incorrect chunkified address rendering.
```

PR #5937, opened 2025-10-03
fix(core): use handshake message payload in handshake hash

M1nd3r:


romanz:
Will be squashed into https://github.com/trezor/trezor-firmware/pull/5922 with updated UI fixtures.

PR #5939, opened 2025-10-03
wipe BLE bonds after entering wipe code

TychoVrahe:

### DIFF
--- a/core/.changelog.d/5939.fixed
+++ b/core/.changelog.d/5939.fixed
@@ -1,0 +1,1 @@
+[T3W1] Erase BLE bonds too after entering wipe code.
--- a/core/embed/io/ble/inc/io/ble.h
+++ b/core/embed/io/ble/inc/io/ble.h
@@ -394,3 +394,10 @@ void ble_set_enabled(bool enabled);
  * @return true if enabled, false otherwise
  */
 bool ble_get_enabled(void);
+
+/**
+ * @brief Wait until BLE driver is ready or timeout occurs
+ *
+ * @return true if BLE driver is ready, false if timeout occurred
+ */
+bool ble_wait_until_ready(void);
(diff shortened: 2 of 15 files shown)


core/embed/sys/task/stm32/system.c
```diff
@@ -247,3 +254,12 @@ void system_exit_fatal(const char* message, const char* file, int line) {
   size_t file_len = file != NULL ? strlen(file) : 0;
   system_exit_fatal_ex(message, message_len, file, file_len, line);
 }
+
+void system_exit_wipe(const char* title, const char* message,
```
cepetr: I wonder if we could do it in a more generic way. For example, instead of adding `system_exit_wipe`, we could use `system_exit()` with a specific exit code to signal a wipe request:

How about this:
- Rename `reboot_with_rsod()` -> `reboot_with_pminfo()`.
- `reboot_and_wipe()` would then call `reboot_with_pminfo()` with `pminfo` set to `TASK_TERM_REASON_EXIT` and `exit_code` set to a specific value.
- Then we would not need a new `BOOT_COMMAND_xxx` or any new `system_exit_xxx()` routines.

Edited: 
With `system_exit()`, we cannot exchange a text message between the app and the bootloader. We could either add this to `pminfo`, or we might be able to live without it for this use case.

cepetr (changes requested):

core/embed/sys/task/stm32/system.c
TychoVrahe: as discussed offline, reworked in https://github.com/trezor/trezor-firmware/commit/b470c124bac826b3110c980dd04ca858f47adc1c

core/embed/projects/bootloader/wire/wire_iface_ble.h
```diff
@@ -28,3 +28,5 @@ void ble_iface_deinit(void);
 bool ble_iface_start_pairing(void);
 
 void ble_iface_end_pairing(void);
+
+bool ble_wait_until_ready(void);
```
cepetr: Nit: this new function might fit better in `ble.h`.

cepetr (approved):
utACK

PR #5940, opened 2025-10-03
USB VCP fixes

cepetr:
This PR fixes several issues in USB VCP communication:

1) RX side fixes - completely refactored the USB VCP driver ring buffer

    - when the buffer is full, incoming data from the host is suspended
    - ring buffer synchronization between interrupt and main task
    - the buffer is no longer a true ring buffer

2) TX side fixes
    - `syshandle_write_blocking()` now uses the entire timeout to send data
    - prodtest console writes to USB with a longer timeout

3) Improved the prodtest ping command to support an optional text parameter to return - so the ping command can be used for vcp stress test.

### DIFF
--- a/core/embed/projects/prodtest/cmd/prodtest_ping.c
+++ b/core/embed/projects/prodtest/cmd/prodtest_ping.c
@@ -22,13 +22,13 @@
 #include <rtl/cli.h>
 
 static void prodtest_ping(cli_t* cli) {
-  if (cli_arg_count(cli) > 0) {
+  if (cli_arg_count(cli) > 1) {
     cli_error_arg_count(cli);
     return;
   }
 
   // Respond with an OK message
-  cli_ok(cli, "");
+  cli_ok(cli, "%s", cli_arg(cli, "text"));
 }
 
 // clang-format off
@@ -37,5 +37,5 @@ PRODTEST_CLI_CMD(
   .name = "ping",
   .func = prodtest_ping,
   .info = "Send a ping to the device",
-  .args = ""
+  .args = "[<text>]"
 );
--- a/core/embed/sys/task/sysevent.c
+++ b/core/embed/sys/task/sysevent.c
@@ -325,10 +325,35 @@ ssize_t syshandle_read_blocking(syshandle_t handle, void *buffer,
 
 ssize_t syshandle_write_blocking(syshandle_t handle, const void *data,
                                  size_t data_size, uint32_t timeout) {
-  if (timeout > 0) {
-    sysevents_t awaited = {.write_ready = 1 << handle};
-    sysevents_t signalled = {0};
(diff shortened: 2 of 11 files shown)


core/embed/sys/task/sysevent.c
```diff
@@ -325,10 +325,35 @@ ssize_t syshandle_read_blocking(syshandle_t handle, void *buffer,
...
+      sysevents_t awaited = {.write_ready = 1 << handle};
+      sysevents_t signalled = {0};
+      sysevents_poll(&awaited, &signalled, deadline);
+
+      if (signalled.write_ready == 0) {
+        // Timeout
+        break;
+      }
+
+      ssize_t written = syshandle_write(handle, ptr, remaining);
```
TychoVrahe, line 347: iiuc a consequence of solving this here is that other task could also be trying to write and there is no guarantee of atomicity of the write, which can potentially be a problem for some of the drivers.. probably not something we need to solve now, not even sure how that should be solved

core/embed/projects/prodtest/cmd/prodtest_ping.c
```diff
@@ -22,13 +22,13 @@
...
 static void prodtest_ping(cli_t* cli) {
-  if (cli_arg_count(cli) > 0) {
+  if (cli_arg_count(cli) > 1) {
     cli_error_arg_count(cli);
     return;
   }
 
   // Respond with an OK message
-  cli_ok(cli, "");
+  cli_ok(cli, "%s", cli_arg(cli, "text"));
```
TychoVrahe, line 31: in case of empty string inserted, this results in returning OK with a trailing space - i gues this is unexpected for the user but not very sure whether this can cause any problems

cepetr: I don't think it's a big issue since the message parser on the other side should ignore this space. If we decide to fix it, I’d prefer to make the changes in `cli_ok()`, `cli_verror()`, and `cli_vtrace()`, not here. Do you think it's worth it?

core/embed/sys/task/sysevent.c
cepetr: Yes, I'm also not sure whether this behavior of `syshandle_write_blocking()` is correct, but I tend to say yes. I'm not sure `syshandle_write_blocking()` needs to be atomic - maybe only `syshandle_write()` does. I agree we should be aware of this and maybe redefine the behavior if this logic doesn't fit in the future.

core/embed/projects/prodtest/cmd/prodtest_ping.c
TychoVrahe: no lets not do that now, lets just hope that the message parsers on the other side are implemented as expected

PR #5943, opened 2025-10-03
Fix PIN keyboard clearance

obrusvit:
This PR restarts the PIN keyboard after suspend. The desired behavior is that the inputted (but not confirmed) PIN code is not readable after the device is woken up.

### DIFF
--- a/core/src/apps/common/lock_manager.py
+++ b/core/src/apps/common/lock_manager.py
@@ -168,6 +168,9 @@ def lock_device_if_unlocked() -> None:
 
     if utils.USE_POWER_MANAGER and not utils.EMULATOR:
         if workflow.autolock_interrupts_workflow:
+            if not config.is_unlocked():
+                # locking during PIN entering should restart the workflow, otherwise the keyboard input field is not cleared
+                workflow.close_others()
             # suspend immediately
             suspend_and_resume()
         else:
--- a/core/src/apps/common/lock_manager.py
+++ b/core/src/apps/common/lock_manager.py
@@ -29,6 +29,7 @@ else:
 
     _SHOULD_SUSPEND = False
     _notify_power_button: loop.mailbox[None] = loop.mailbox()
+    notify_bootscreen: loop.mailbox[None] = loop.mailbox()
 
     def _schedule_suspend_after_workflow() -> None:
         """Signal that the device should be suspended by the default task after the
@@ -47,6 +48,7 @@ else:
 
         Notifies an asynchronous task to perform the suspend in a separate thread.
         """
+        notify_bootscreen.put(None, replace=True)
         _notify_power_button.put(None, replace=True)
 
     async def _power_handler() -> None:
--- a/core/src/boot.py
+++ b/core/src/boot.py
@@ -17,6 +17,7 @@ from trezor.pin import (
     show_pin_timeout,
 )
 from trezor.ui.layouts.homescreen import Lockscreen
+from micropython import const
 
 from apps.common.request_pin import can_lock_device, verify_user_pin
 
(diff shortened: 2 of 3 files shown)


core/src/boot.py
```diff
@@ -50,6 +51,35 @@ def enforce_welcome_screen_duration() -> None:
...
+
+    async def pin_unlock_sequence() -> None:
+        while True:
+            lockscreen = Lockscreen(label=storage.device.get_label(), bootscreen=True)
+            await lockscreen.get_result()
+            lockscreen.__del__()
+            res = await loop.race(verify_user_pin(), wait_for_suspend())
+            if res is _SUSPEND_MARKER:
+                # make some delay for the suspend
+                await loop.sleep(100)
```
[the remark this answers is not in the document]
romanz: What may happen if there is no `sleep` here?

core/src/apps/common/lock_manager.py
```diff
@@ -168,6 +168,9 @@ def lock_device_if_unlocked() -> None:
 
     if utils.USE_POWER_MANAGER and not utils.EMULATOR:
         if workflow.autolock_interrupts_workflow:
+            if not config.is_unlocked():
+                # locking during PIN entering should restart the workflow, otherwise the keyboard input field is not cleared
+                workflow.close_others()
```
romanz, line 175: Can we add a test for this case (auto-locking during PIN entry in FW main loop)?

core/src/apps/common/lock_manager.py
```diff
@@ -168,6 +170,9 @@ def lock_device_if_unlocked() -> None:
...
```
romanz, line 175: `workflow.close_others()` is also called by the `lock_device()` method above (if `workflow.autolock_interrupts_workflow` is true). 
- Is it related to this issue?
- Will `workflow.close_others()` called twice?

core/src/boot.py
```diff
@@ -25,7 +26,7 @@
 
 if utils.USE_POWER_MANAGER:
     from trezor import workflow
-    from apps.common.lock_manager import configure_autodim, boot as boot_power_manager
+    import apps.common.lock_manager as lock_manager
```
romanz: nit (for consistency with the `import` above):
```suggestion
    from apps.common import lock_manager
```

core/src/boot.py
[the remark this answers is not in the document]
obrusvit: If we don't `sleep` here then the `lockscreen` is shown very briefly in the next iteration before the device actually suspends.
- with `sleep`: 
    PIN keyboard -> PWR button -> black screen
- without `sleep`:
    PIN keyboard -> PWR button -> Lockscreen -> black screen

core/src/apps/common/lock_manager.py
romanz: Maybe it can be simplified by modifying this line:
```python
elif config.is_unlocked():
```
to be:
```python
elif config.is_unlocked() or workflow.autolock_interrupts_workflow:
```

obrusvit: you're right that on the 1st pass through `lock_device_if_unlocked`, the `workflow.close_others()` is called twice. On the repeated passes (i.e. pressing the power button on the lockscreen), it should be called only once. Your proposal would work but maybe we can introduce a new bool `workflow.lockscreen_active` or something like that :thinking:

core/src/boot.py
obrusvit: https://github.com/trezor/trezor-firmware/pull/5943/commits/3c1c85e31f1d71fbc464189c819b567d5eeb0f51

core/src/apps/common/lock_manager.py
romanz: (not urgent for now, can be done in a separate PR)

obrusvit:
Merging amidst failing coverage report for T3B1, will investigate later.

PR #4531, opened 2025-01-23
WIP: Increase passphrase size

romanz:

PR #5946, opened 2025-10-06
Fix: Center align shown pin for Eckhart

### LINKED ISSUE #5623: Chunkified text doesn't properly work in the TextLayout

PrisionMike:
Center align shown pin for Eckhart, as per design. Adding spaces instead of chunkification for spacing out the pin.
[no_changelog]
Fixes #5623


core/embed/rust/src/ui/component/text/layout.rs
PrisionMike, line 1: Just a formatting commit. Not relevant to fix. Keeping `struct` and `impl` together.

core/embed/rust/src/strutil.rs
```diff
@@ -84,6 +84,19 @@ pub fn format_pairing_code(code: u32, width: usize) -> ShortString {
     formatted_code
 }
 
+/// Adds spaces between characters. Output length is capped at 99 since
+/// ShortString is 50 characters long max. Example: "12345" becomes "1 2 3 4 5"
+pub fn space_out_str(pin: &str) -> String<99> {
```
obrusvit: The bad thing about this is that it generates new variant for `String<T>` which takes some precious flash space. That's why we want to keep the generic type instantiations at minimum. We should think about a solution.
- are we able to make `space_out_str` return `'str` somehow? We may have to deal with some lifetime issues
- making `ShortString` to be `String<99>` instead of `String<50>`. That's bad because now we carry ~twice as much bytes everywhere.
- making `space_out_str` to be `Eckhart` specific so that it consumes flash space only there where we're not that limited
- ??

obrusvit: Maybe you can compare binsizes of T3W1 before and after your change.

obrusvit: Maybe we can revive and merge https://github.com/trezor/trezor-firmware/pull/4531 
cc @romanz and thanks for the suggestion.

romanz: > making `space_out_str` to be `Eckhart` specific so that it consumes flash space only there where we're not that limited

I think it would be the simplest solution (for now), since rebasing & testing https://github.com/trezor/trezor-firmware/pull/4531 will probably take a while.

obrusvit: Let's do so @PrisionMike 

I just tested on the device and it works nicely :+1:

PrisionMike: Will need some reference but I think I can do that. I also have another idea of copying this routine to `pin.rs` where it is used. Since we only need it while rendering the pin, I can possibly get away with just using the RAM instead of invoking another type instantiation.

obrusvit: @PrisionMike
> I also have another idea of copying this routine to pin.rs where it is used.

That's exactly what we meant. That will do the trick for now. Just remove your function `space_out_str` from `strutil.rs` and put it to `layout_eckhart/.../pin.rs`.

PrisionMike:
no_changelog

obrusvit:
After updating `fixtures.json` please squash the commits. Especially, we don't want "Revert" commits in the history.

PrisionMike:
[no changelog]

PR #5947, opened 2025-10-06
Eckhart UI: change loaders

obrusvit:
This PR changes Eckhart loaders, i.e. `ProgressScreen`:
- removes the percentage indication
- moves the `title` text into the upper area, similarly as is done in Bootloader `screen_progress(..)`


core/translations/en.json
bieleluk, line 1: Two strings do not have the ellipsis added for Eckhart:
- Coinjoin in progress
- Changing language

core/embed/rust/src/ui/layout_eckhart/theme/mod.rs
```diff
@@ -212,3 +212,8 @@ pub const TEXT_SMALL_GREY_EXTRA_LIGHT: TextStyle = TextStyle {
     text_color: GREY_EXTRA_LIGHT,
     ..TEXT_SMALL
 };
+
+/// Where to place main text of progress screens of bootloader and firmware
+pub const PROGRESS_TEXT_ORIGIN: Point = super::constant::SCREEN
+    .top_left()
+    .ofs(Offset::new(PADDING, 38));
```
bieleluk: nit: The value `38`  is "content top padding" when the header is not present. It is also hard-coded for the  `TextScreen` and `Homescreen`. Can we create a constant in the `theme` for all of them at once?

core/embed/rust/src/ui/layout_eckhart/firmware/progress_screen.rs
```diff
@@ -93,16 +83,26 @@ impl Component for ProgressScreen {
...
+            Offset::new(
+                SCREEN.width() - 2 * theme::PADDING,
+                4 * theme::TEXT_NORMAL.text_font.text_max_height(),
+            ),
+        );
         let middle_text_area = Rect::snap(
-            SCREEN.center(),
-            Offset::new(bounds.width(), max_text_area),
-            Alignment2D::CENTER,
+            main_text_area.bottom_center(),
```
bieleluk, line 94: nit:
This approach avoids text overlaps but is prone to vertical shifts. I would personally stick with the central placement and ensure that there is no overlap between areas.

core/translations/en.json
obrusvit: https://github.com/trezor/trezor-firmware/pull/5947/commits/8259ce577d2dbd887d30eafe93c87967d315f459

core/embed/rust/src/ui/layout_eckhart/theme/mod.rs
obrusvit: https://github.com/trezor/trezor-firmware/pull/5947/commits/8e4f11cb6bde6b71f9a121475a360f6bd3929665

core/embed/rust/src/ui/layout_eckhart/firmware/progress_screen.rs
obrusvit: You're right but it's only used for Coinjoin loaders so let's not complicate it for now.

bieleluk (approved):
LGTM

PR #5948, opened 2025-10-06
chore(core): remove Japan from regulatory screen

obrusvit:
- temporarily remove the regulatory screen for Japan until the certification process is completed

[no changelog]


bieleluk (approved):
utACK

PR #5949, opened 2025-10-06
feat(core): host afuera

obrusvit:
- removes the occurences of "host" and instead using something more user-friendly like "app" or "connected device"

[no changelog]


core/embed/rust/src/ui/layout_eckhart/bootloader/connect.rs
bieleluk, line 1: `thp__continue_on_host` should be changed as well

core/translations/en.json
bieleluk, line 1: `thp__continue_on_host` should be changed as well or completely removed if it's unused

obrusvit: Done on rebase

bieleluk (approved):

PR #5952, opened 2025-10-06
respect BLE settings in bootloader

TychoVrahe:
This PR implements a system that enables bootloader to respect BLE on/off settings as set by user in firmware.

This setting is copied to backup RAM, so that bootloader can read it after booting.

In bootloader, when user starts new pairing in menu, BLE is re-enabled automatically, but is not respected for next firmware run. 

Using backup RAM for this means that if the device is entirely discharged, and user enters bootloader on next switch on before running FW, the bluetooth will be enabled. Should be very rare situation.

Few fixes added:
- start advertising after re-enabling BLE in fimrware device menu
- wipe backup RAM along with storage/bonds in bootloader after receiving WIPE boot command.

### DIFF
--- a/core/embed/io/ble/stm32/ble.c
+++ b/core/embed/io/ble/stm32/ble.c
@@ -27,6 +27,7 @@
 
 #include <io/ble.h>
 #include <io/nrf.h>
+#include <sys/backup_ram.h>
 #include <sys/irq.h>
 #include <sys/sysevent_source.h>
 #include <sys/systick.h>
@@ -50,6 +51,11 @@
 
 #define BLE_DATA_HEADER_SIZE 7
 #define BLE_DATA_SIZE (BLE_RX_PACKET_SIZE + BLE_DATA_HEADER_SIZE)
+
+typedef struct {
+  uint8_t version;
+  bool enabled;
+} ble_recovery_data_t;
 
 typedef struct {
   ble_mode_t mode_requested;
@@ -659,8 +665,16 @@ bool ble_init(void) {
     goto cleanup;
   }
 
-  drv->power_level = BLE_TX_POWER_PLUS_4_DBM;
   drv->enabled = true;
+  ble_recovery_data_t backup_data;
+  if (backup_ram_read(BACKUP_RAM_KEY_BLE_SETTINGS, &backup_data,
+                      sizeof(backup_data), NULL)) {
+    if (backup_data.version == 1) {
+      drv->enabled = backup_data.enabled;
+    }
+  }
+
+  drv->power_level = BLE_TX_POWER_PLUS_4_DBM;
   drv->initialized = true;
   return true;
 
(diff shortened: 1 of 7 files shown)


core/embed/io/ble/stm32/ble.c
```diff
@@ -51,6 +52,11 @@
 #define BLE_DATA_HEADER_SIZE 7
 #define BLE_DATA_SIZE (BLE_RX_PACKET_SIZE + BLE_DATA_HEADER_SIZE)
 
+typedef struct {
+  uint8_t version;
+  bool enabled;
+} ble_backup_data_t;
```
kopecdav: nitpick: to be consistent with power manager, can we rename this to `ble_recovery_data_t`?

core/embed/projects/bootloader/workflow/wf_ble_pairing_request.c
```diff
@@ -45,6 +45,8 @@ static bool encode_pairing_code(uint32_t code, uint8_t *outbuf) {
 }
 
 workflow_result_t workflow_ble_pairing_request(const fw_info_t *fw) {
+  ble_set_enabled(true);
```
kopecdav, line 48: Iam not sure how this should work. So if I have a device on which i disabled the `BLE` in FW device menu, and then jump too bootloader. Bootloader will initiate `BLE` module disabled, but clicking on `Pair_new_device` will completelly takeover this settings and enable the `BLE` again. Then if i do not pair anything and cancel the pairing by pressing `X` -> `BLE` remain enabled even if I jump to FW. 

Woudn't be better to disable the BLE again if user cancel the pairing in bootloader?

kopecdav (changes requested):

core/embed/projects/bootloader/workflow/wf_ble_pairing_request.c
TychoVrahe: In firmware, first thing done is that the BLE enabled setting is read from storage and this function is called again, overriding the setting stored in RAM changed by bootloader. So this is only sort of temporary enable.

Whether this is correct is still question, maybe we should rather disable the 'pair new device' in bootloader menu, maybe there should be 'enable BLE' item instead, not sure. But this is probably easiest option, we can redo this later if we get product direction

core/embed/io/ble/stm32/ble.c
TychoVrahe: https://github.com/trezor/trezor-firmware/commit/9ee830eb0a4f014fdf624803be17d88771c4fd03

core/embed/projects/bootloader/workflow/wf_ble_pairing_request.c
kopecdav: I dont think that this is temporary because `ble_set_enabled()` will rewrite the backup RAM, so in case the bootloader will enable the BLE, firmware will start also enabled doesn't it? 

Or is there any other storage which will takover the BLE settings in backup RAM?

core/embed/io/ble/stm32/ble.c
kopecdav: thanks!

core/embed/projects/bootloader/workflow/wf_ble_pairing_request.c
TychoVrahe: yes there is 'the Storage', in flash, where firmware stores this setting and it takes precedence over backup RAM. The backup RAM is used by bootloader because it cannot access the Storage.

kopecdav: ahh ok, now its clear. thanks

kopecdav (approved):

PR #5954, opened 2025-10-06
ci(core): always upload coverage report

romanz:
Otherwise, we don't have the coverage report in case the job fails:
https://github.com/trezor/trezor-firmware/actions/runs/18283216173/job/52053655433


M1nd3r:
LGTM - the code coverage artefact is generated for the failing test in this PR (the change works).

romanz:
Opened https://github.com/trezor/trezor-firmware/issues/5957 using this PR's findings.

